PR #1, opened 2017-04-12
Add actions to the script to enable/disable the repository

albertodonato:
Note that the script is not yet fully functional, since the GPG key is not imported.

Also, it currently points at staging.

### DIFF
--- a/tests/test_ubuntu_advantage.py
+++ b/tests/test_ubuntu_advantage.py
@@ -1,13 +1,21 @@
+from __future__ import print_function
 import imp
 import os
-from unittest import TestCase
+
+from fixtures import TestWithFixtures, TempDir
 
 script_path = os.path.join(
     os.path.dirname(os.path.dirname(__file__)), 'ubuntu-advantage')
 ubuntu_advantage = imp.load_source('ubuntu_advantage', script_path)
 
 
-class UbuntuAdvantageTests(TestCase):
+class UbuntuAdvantageTest(TestWithFixtures):
+
+    def setUp(self):
+        super(UbuntuAdvantageTest, self).setUp()
+        self.tempdir = self.useFixture(TempDir())
+        self.messages = []
+        self.print = lambda message: self.messages.append(message)
 
     def test_get_args_enable(self):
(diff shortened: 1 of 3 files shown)


albertodonato:
[Write script to configure the ESM repo](https://trello.com/c/l89ynCFo/242-write-script-to-configure-the-esm-repo)

ubuntu-advantage
```diff
@@ -13,15 +25,53 @@ def get_args(argv):
...
+        ' Run "sudo apt-get update" to update lists.')
+
+
+def disable_esm(lists_dir=None, print=print):
+    """Disable ESM archive."""
+    list_file = get_list_file(lists_dir=lists_dir)
+    if os.path.exists(list_file):
+        list_file_disabled = list_file + '.save'
+        os.rename(list_file, list_file_disabled)
+        print('Ubuntu ESM repository disabled.')
```
fginther: Who is the consumer of this message? Do we not need to add 'Run "sudo apt-get update" to update lists.' here as well to avoid installing packages already in apt cache?

albertodonato: This script is intented to be included in a package in the base distro, and to be run manually, so the message is intended for the user directly.

We could of course run apt-get update as well, I went with the original sketch of the script which didn't have that (same way by default apt-add-repository doesn't update lists).

tests/test_ubuntu_advantage.py
```diff
@@ -19,3 +27,74 @@ def test_get_args_disable(self):
...
+            lists_dir=self.tempdir.path, print=self.print)
+        # the file has moved
+        self.assertFalse(os.path.exists(list_file))
+        # the original content is still there
+        with open(list_file + '.save') as fh:
+            content = fh.read()
+        self.assertEqual(original_content, content)
+        # a message is printed out
+        self.assertIn(
+            'Ubuntu ESM repository enabled.'
```
fginther: Oops, I missed this the first time. This should be looking for 'Ubuntu ESM repository disabled.'. This test passes because it calls enable too so both messages are in self.message. My quick fix for this is https://pastebin.canonical.com/185715/.

PR #3, opened 2017-04-12
Import GPG key for the repository

albertodonato:
Import the GPG key used to sign the ESM repository.

Currently, the public key is the one for staging.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 flake8
+mock
 fixtures
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: python
+sudo: required
 python:
   - "2.7"
 install: pip install flake8 tox
 script: tox
+addons:
+  apt:
+    packages:
+      - gnupg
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,2 +1,3 @@
 flake8
+mock
```
ericsnowcurrently: If this project were Python 3, you'd have mock for free in the stdlib. :)

albertodonato: I know, but precise is quite old, and this file will be included in an existing package, so we don't want to add dependencies (python3)

albertodonato:
[Write script to configure the ESM repo on machines](https://trello.com/c/l89ynCFo/242-write-script-to-configure-the-esm-repo-on-machines)

blackboxsw:
Should be remove the apt-key upon disable?

albertodonato:
@blackboxsw I think it's fine to leave it there. It won't do any harm, and removing the repo is not something the customer would normally do (since you'd end up with no updates at all after EOL)

PR #9, opened 2017-04-26
Esm motd

dpb1:


dustinkirkland:
Looks great!  +1

update-motd.d/99-esm
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+ESM_ENABLED="This Ubuntu 12.04 ESM system is actively receiving extended security updates
```
sparkiegeek: Does it change identity just because ESM is enabled? It's still called Ubuntu 12.04 LTS last time I checked?

update-motd.d/99-esm
```diff
@@ -0,0 +1,17 @@
...
+from Canonical:
+ * http://www.ubuntu.com/esm"
```
sparkiegeek: https

update-motd.d/99-esm
```diff
@@ -0,0 +1,17 @@
...
+
+ESM_DISABLED="This Ubuntu 12.04 LTS system is past its End of Life, and is no longer
+receiving security updates.  To protect the integrity of this system, it’s
+critical that you enable Extended Security Maintenance updates:
+ * http://www.ubuntu.com/esm"
```
sparkiegeek: https

update-motd.d/99-esm
```diff
@@ -0,0 +1,17 @@
...
+
+if `apt-cache policy | grep -q 'esm\.ubuntu\.com'`; then
```
albertodonato: You can use 
```
if ubuntu-advantage is-esm-enabled; then
    ...
```

now (I added the command to the new script version). Note that no ` is used (you want the return code, not the output of the command)

dustinkirkland:
Yes, we are referring to it as 12.04 ESM, rather than 12.04 LTS.

update-motd.d/99-esm
albertodonato: Fixed

update-motd.d/99-esm
albertodonato: Fixed

albertodonato:
@dustinkirkland can you please re-review? I made some changes to the PR

update-motd.d/99-esm
basak: "actively receiving" would be inaccurate, for example if the token has been revoked.

dustinkirkland:
"registered to receive"


:-Dustin
@dustinkirkland

PR #10, opened 2017-04-27
Rewrite script in POSIX shell

albertodonato:


ubuntu-advantage
```diff
@@ -1,166 +1,141 @@
...
+ enable-esm <token>     enable the ESM repository
+ disable-esm            disable the ESM repository
+
+the <token> argument must be in the form "user:password"
+
+EOF
+}
+
+
+if [ "$USER" != "root" ]; then
```
sparkiegeek: instead of $USER, use `id -u` as per https://askubuntu.com/a/30157/1102

ubuntu-advantage
```diff
@@ -1,166 +1,141 @@
...
+hjyMm0AWNnQShlt+ChQn9Uv/GgcrhSU3KiM/RkiAyGuW1BKSC7BW3IroRWilSHVQ
+2Df3PX3pNQ==
+=FHLy
+-----END PGP PUBLIC KEY BLOCK-----
+EOF
+}
+
+
+write_list_file() {
+    cat > "$REPO_LIST" <<EOF
```
sparkiegeek: Should we guard this on $REPO_LIST not existing? and/or consider setting noclobber ?

i.e. What's the expected behaviour of two enables one after the other?

ubuntu-advantage
```diff
@@ -1,166 +1,141 @@
...
+#!/bin/sh -e
+
+SCRIPTNAME=$(basename "$0")
+
+REPO_URL="esm.ubuntu.com"
+REPO_LIST=${REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-esm-precise.list"}
+APT_KEY_ADD=${APT_KEY_ADD:-apt-key add -}
+
+import_gpg_key() {
+    $APT_KEY_ADD >/dev/null <<EOF
```
sparkiegeek: luckily EOF doesn't appear in the key :+1:

ubuntu-advantage
```diff
@@ -1,166 +1,141 @@
...
+    disable-esm)
+        disable_esm
+        ;;
+
+    is-esm-enabled)
+        is_esm_enabled
+        ;;
+
+    *)
+        usage
```
sparkiegeek: hmm, would be nice to see usage without being root, perhaps call this before bailing out on that check?

ubuntu-advantage
albertodonato: it would have to be on a line by itself, which can't happen

ubuntu-advantage
albertodonato: what's the issue if $REPO_LIST doesn't exist? that's the normal case when the repo is not enabled.
Two enables would be effectively no-op, unless you change the password (which is actually a valid use case, in case you typo'd it the first time)

ubuntu-advantage
albertodonato: Done, thanks.

ubuntu-advantage
sparkiegeek: My concern was with it existing. e.g. I enable it, then I edit the file and add comments, or version it with Puppet and add a header or something like that, do we expect/want the script to clobber what I did? I would vote no.

ubuntu-advantage
albertodonato: done

ubuntu-advantage
albertodonato: as discussed on IRC, if you're using a config management tool you should configure the repository through it (for instance Landscape managing APT sources would override local changes anyway). This is meant for users configuring the repository locally.

Issue #11, opened 2017-04-27
Update GPG key with new email uid


panlinux:
From the [SRU comment #6](https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/1686183/comments/6), we should update the GPG key with a more descriptive email address.

[RT #101190](https://portal.admin.canonical.com/102190) was filed requesting this change, and as soon as we have the new key contents the copy in the ubuntu-advantage script should be updated.

panlinux:
Updated key at https://pastebin.canonical.com/187000/plain/

Issue #13, opened 2017-04-27
Do not use apt-key to install the ESM gpg key


panlinux:
From [SRU comment #7](https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/1686183/comments/7):

- the key should be shipped as a key fragment in `/usr/share/keyrings/ubuntu-keyring-extended-security-maintainance.gpg`
- the shell script should simply copy that key fragment into `/etc/apt/trusted.gpg.d/` upon enablement of the ESM repository
- there should not be encoded binary in the shell script, and no need to call `apt-key`; just a `cp`.

Additionally, the key should be removed from `/etc/apt/trusted.gpg.d/` when esm is disabled.

PR #14, opened 2017-04-27
Rework script.

### LINKED ISSUE #12: Install apt-transport-https if needed
From [SRU comment #7](https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/1686183/comments/7), the ubuntu-advantage script should install the apt-transport-https package if it's not installed, so that apt can access the ESM archive which is behind https.

albertodonato:
- copy the keyring file instead of importing with apt-key.
- install apt-transport-https if not installed.
- rework tests

Closes #12 and #13


ubuntu-advantage
```diff
@@ -71,9 +16,16 @@ deb https://${1}@${REPO_URL}/ubuntu precise main
...
+is_package_installed() {
+    dpkg-query -s "$1" >/dev/null 2>&1
+}
+
 
 enable_esm() {
-    import_gpg_key
+    cp "${KEYRINGS_DIR}/${REPO_KEY_FILE}" "$APT_KEYS_DIR"
+    if ! is_package_installed apt-transport-https; then
+        apt-get install apt-transport-https
```
panlinux: I think we should use `apt-get install -y` here. As an improvement, we could wrap this install in a nice message saying "Installing missing dependency <foo>..." and only display apt's output in the case of an error, or else the screen gets too polluted in my opinion: see https://pastebin.canonical.com/187004/

ubuntu-advantage
```diff
@@ -82,6 +34,7 @@ enable_esm() {
 disable_esm() {
     if [ -f "$REPO_LIST" ]; then
         mv "$REPO_LIST" "${REPO_LIST}.save"
+        rm  "$APT_KEYS_DIR/$REPO_KEY_FILE"
```
panlinux: This `rm` should be guarded with `-f`, or the script can fail if that file doesn't exist by any chance:

```
ubuntu@precise-esm:~$ sudo rm /etc/apt/trusted.gpg.d//ubuntu-keyring-esm.gpg
ubuntu@precise-esm:~$ sudo ubuntu-advantage disable-esm
rm: cannot remove `/etc/apt/trusted.gpg.d/ubuntu-keyring-esm.gpg': No such file or directory
ubuntu@precise-esm:~$ 
```

ubuntu-advantage
```diff
@@ -71,9 +16,16 @@ deb https://${1}@${REPO_URL}/ubuntu precise main
 EOF
 }
 
+is_package_installed() {
+    dpkg-query -s "$1" >/dev/null 2>&1
```
panlinux: Suggestion: `dpkg-query -W` has less output to get rid of than `-s`:
```
root@precise-esm:~# dpkg-query -W base-files; echo $?
base-files	6.5ubuntu6.9
0
root@precise-esm:~# dpkg-query -W asdasdasd; echo $?
No packages found matching asdasdasd.
1
root@precise-esm:~# 
```

ubuntu-advantage
panlinux: Should we run `apt-get update` here perhaps?

dpb1: I'd rather not do an apt-get update.  perhaps... print a message.

albertodonato: yeah, that's a corner case (you have no reason to remove the file yourself), but fixed

ubuntu-advantage
albertodonato: Changed

ubuntu-advantage
albertodonato: we do print a message when enabling and disabling telling the use to run `apt-get update`

ubuntu-advantage
albertodonato: Actually, in precise `-W` will print the package name (without version) if it's not installed. So back to `-s`

ubuntu-advantage
albertodonato: Done

albertodonato:
Closes #11 too

panlinux:
Shipit

PR #26, opened 2017-06-30
Install ca-certificates (Fixes #25).

### LINKED ISSUE #25: enable-esm should install ca-certificates on trusty+
https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1690270

Just as apt-transport-https is not in the minimal seed and must be installed by ubuntu-advantage enable-esm to ensure its availability, ca-certificates is not part of minimal, only standard. Without ca-certificates, apt-transport-https does not have SSL CAs to verify the esm.ubuntu.com site against.

While this is non-fatal in precise, later versions of apt-transport-https (rightly!) enforce SSL certificate verification, so we should fix this to be future-proof.

dpb1:
- Add testing instructions to README
- unit tests
- Fix dep8 text matching issue on precise.

### DIFF
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -2,13 +2,14 @@
 
 SCRIPTNAME=$(basename "$0")
 
-SERIES=`lsb_release -cs`
+SERIES=$(lsb_release -cs)
 REPO_URL="esm.ubuntu.com"
 REPO_KEY_FILE="ubuntu-esm-keyring.gpg"
 REPO_LIST=${REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-esm-${SERIES}.list"}
 KEYRINGS_DIR=${KEYRINGS_DIR:-"/usr/share/keyrings"}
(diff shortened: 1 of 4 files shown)


ubuntu-advantage
```diff
@@ -2,13 +2,14 @@
 
 SCRIPTNAME=$(basename "$0")
 
-SERIES=`lsb_release -cs`
+SERIES=$(lsb_release -cs)
```
panlinux: Please keep using backslashes. Even though Ubuntu's /bin/sh (-> dash) supports $(), the rest of the script uses backslashes.

dpb1: Keeping this one, there are no other backticks used in the script (and there are other instances of `$()`, so I'm actually unifying things).  Also, `shellcheck` doesn't like it:
```
dpb@helo:ubuntu-advantage-script[(25-ca-certs)]$ shellcheck -s sh ubuntu-advantage

In ubuntu-advantage line 5:
SERIES=`lsb_release -cs`
       ^-- SC2006: Use $(..) instead of legacy `..`.

```

I want to add shellcheck as a make target, but... another branch I think.

PR #32, opened 2017-08-08
Check the running kernel version before trying to enable livepatch

panlinux:
This branch adds a running kernel check just before enabling livepatch. If running an unsupported kernel for snaps, we print a message instructing the user to reboot into a supported kernel and then run the enable-livepatch command again. This will catch the trusty fresh install case, where you get a 3.x kernel, and also people who for some reason have downgraded their kernel on xenial.

This is not a livepatch issue: any snap will break if running on a kernel less than 4.4.0. At first it will seem to be fine (snap list, snap install, snap remove, all work), but the moment you try to use a snap, it fails.

### DIFF
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -12,7 +12,7 @@ APT_KEYS_DIR=${APT_KEYS_DIR:-"/etc/apt/trusted.gpg.d"}
 APT_METHOD_HTTPS=${APT_METHOD_HTTPS:="/usr/lib/apt/methods/https"}
 CA_CERTIFICATES=${CA_CERTIFICATES:="/usr/sbin/update-ca-certificates"}
 SNAPD=${SNAPD:="/usr/lib/snapd/snapd"}
-
+KERNEL_VERSION=$(uname -r)
 
 install_livepatch_prereqs() {
     if [ ! -f "$SNAPD" ]; then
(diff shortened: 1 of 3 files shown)


ubuntu-advantage
```diff
@@ -30,8 +30,20 @@ install_livepatch_prereqs() {
...
+        else
+            echo "Your currently running kernel ($(uname -r)) is too old to"
+            echo "support snaps. Version 4.4.0 or higher is needed."
+            echo
+            echo "Please reboot your system into a supported kernel version"
+            echo "and run the following command one more time to complete the"
+            echo "installation:"
+            echo
+            echo "sudo ubuntu-advantage enable-livepatch $1"
+            exit 5
```
albertodonato: Is it worth to document return values in the manpage?

ubuntu-advantage
```diff
@@ -30,8 +30,20 @@ install_livepatch_prereqs() {
 enable_livepatch() {
     install_livepatch_prereqs
     if ! is_livepatch_enabled; then
-        echo 'Enabling Livepatch with the given token, stand by...'
-        canonical-livepatch enable "$1"
+        if check_snapd_kernel_support; then
+            echo 'Enabling Livepatch with the given token, stand by...'
+            canonical-livepatch enable "$1"
+        else
+            echo "Your currently running kernel ($(uname -r)) is too old to"
```
albertodonato: I'd add a global variable like `KERNEL_VERSION=$(uname -r)` and use it here and below

ubuntu-advantage
```diff
@@ -109,6 +121,19 @@ validate_livepatch_token() {
     echo "$1" | grep -q -E '^[0-9a-fA-F]{32}$'
 }
 
+# snapd needs a 4.4.x *running* kernel
+check_snapd_kernel_support() {
+    kernel_release=$(uname -r)
+    v1=$(echo "$kernel_release" | cut -d . -f 1)
+    v2=$(echo "$kernel_release" | cut -d . -f 2)
```
albertodonato: please declare `v1` and `v2` local

ubuntu-advantage
```diff
@@ -109,6 +121,19 @@ validate_livepatch_token() {
...
+    if [ "$v1" -ge "4" ]; then
+        if [ "$v2" -ge "4" ]; then
+            return 0
+        fi
+    fi
```
albertodonato: You can write this simply as 
```
test "$v1" -ge 4 -a "$v2" -ge 4
```

ubuntu-advantage
panlinux: Yes, I've been thinking about that. It should be a new branch, though.

ubuntu-advantage
```diff
@@ -109,6 +122,15 @@ validate_livepatch_token() {
     echo "$1" | grep -q -E '^[0-9a-fA-F]{32}$'
 }
 
+# snapd needs a 4.4.x *running* kernel
+check_snapd_kernel_support() {
+    local v1 v2
+    kernel_release=$KERNEL_VERSION
```
albertodonato: Drop this and just use $KERNEL_VERSION below?

panlinux: ops :)

PR #35, opened 2017-08-08
Only display apt output in the case of errors.

panlinux:
Collect all the output of apt calls and save the result code. If it indicates there was an error, then display it and exit.

This was mainly done for the trusty case, where apt-get install snapd pulls in a new kernel and that package's postinst prints out a lot of info to stderr. Other packages we might install are cleaner and only output to stderr when there is an error. So while not really needed for these other apt-get calls, I decided to wrap them in the say way for consistency, both in terms of coding and script output.

Previous Trusty output: http://pastebin.ubuntu.com/25271508/
Output with the proposed changes: http://pastebin.ubuntu.com/25271514/

### DIFF
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -112,7 +112,7 @@ disable_esm() {
         mv "$REPO_LIST" "${REPO_LIST}.save"
         rm -f "$APT_KEYS_DIR/$REPO_KEY_FILE"
         echo -n 'Running apt-get update... '
-        check_result apt-get update 2>&1
+        check_result apt-get update
         echo 'Ubuntu ESM repository disabled.'
     else
         echo 'Ubuntu ESM repository was not enabled.'
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -14,10 +14,25 @@ CA_CERTIFICATES=${CA_CERTIFICATES:="/usr/sbin/update-ca-certificates"}
 SNAPD=${SNAPD:="/usr/lib/snapd/snapd"}
 KERNEL_VERSION=$(uname -r)
 
+check_result() {
+    local result output
+    result=0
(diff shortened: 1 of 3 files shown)


ubuntu-advantage
```diff
@@ -14,10 +14,25 @@ CA_CERTIFICATES=${CA_CERTIFICATES:="/usr/sbin/update-ca-certificates"}
...
+        echo "ERROR"
+        if [ -n "$2" ]; then
+            echo "$2" >&2
+        fi
+        exit $1
+    else
+        echo "OK"
+    fi
+}
+
```
albertodonato: You could actually put all the logic in this function:

```bash
check_result() {
    local result output
    result=0
    output=$("$@" 2>&1) || result=$?
    if [ $result -ne 0 ]; then
        echo "ERROR"
        if [ "$output" ]; then
            echo "$output"
        fi
    else
        echo "OK"
    fi   
}
```

Then the call site would be just ```check_call apt-get install -y snapd``` (and you can drop `result` and `output` there)

ubuntu-advantage
```diff
@@ -79,24 +95,24 @@ enable_esm() {
...
 }
 
 disable_esm() {
     if [ -f "$REPO_LIST" ]; then
         mv "$REPO_LIST" "${REPO_LIST}.save"
         rm -f "$APT_KEYS_DIR/$REPO_KEY_FILE"
-        echo 'Running apt-get update...'
-        apt-get update >/dev/null
+        echo -n 'Running apt-get update... '
+        check_result apt-get update 2>&1
```
albertodonato: is the `2>&1` needed?

panlinux: No, that's a left over, thanks for catching it

PR #36, opened 2017-08-10
Add shellcheck linter

albertodonato:
Run shellcheck in Travis builds.
Also, a couple of minor cleanups in the script

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,14 @@
 language: python
 python:
   - "3.5"
+dist: precise
 install: pip install tox
-script: tox
+script:
+  - tox
+  - shellcheck -s dash ubuntu-advantage update-motd.d/*
+addons:
+  apt:
+    sources:
+      - debian-sid
+    packages:
+      - shellcheck
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -1,4 +1,5 @@
 #!/bin/sh -e
+# shellcheck disable=SC2039
 
 SCRIPTNAME=$(basename "$0")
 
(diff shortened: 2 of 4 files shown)


ubuntu-advantage
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh -e
+# shellcheck disable=SC2039
```
albertodonato: This is to disable the warning about non-POSIX use of `local` (which is fine with dash)

.travis.yml
```diff
@@ -1,5 +1,15 @@
...
-script: tox
+script:
+  - tox
+  - shellcheck -s dash ubuntu-advantage update-motd.d/*
+addons:
+  apt:
+    sources:
+      - debian-sid
+    packages:
+      - shellcheck
```
panlinux: Why do we need the package from debian?

ubuntu-advantage
```diff
@@ -126,7 +127,7 @@ is_esm_enabled() {
 is_livepatch_enabled() {
     # it's fine if it fails because the snap isn't installed. It's still
     # a non-zero return value
-    canonical-livepatch status > /dev/null 2>&1
+    canonical-livepatch status >/dev/null 2>&1
```
panlinux: Is this just cosmetic?

ubuntu-advantage
```diff
@@ -9,16 +10,16 @@ REPO_KEY_FILE="ubuntu-esm-keyring.gpg"
...
+CA_CERTIFICATES=${CA_CERTIFICATES:-"/usr/sbin/update-ca-certificates"}
+SNAPD=${SNAPD:-"/usr/lib/snapd/snapd"}
 KERNEL_VERSION=$(uname -r)
 
 check_result() {
     local result output
     result=0
     output=$("$@" 2>&1) || result=$?
-    if [ "$result" -ne "0" ]; then
+    if [ "$result" -ne 0 ]; then
```
panlinux: Shouldn't we also remove the quotes from $result then?

ubuntu-advantage
albertodonato: yes

.travis.yml
albertodonato: shellcheck is not in precise

ubuntu-advantage
albertodonato: yeah, I can do that too

albertodonato:
@panlinux yeah the version in xenial seems to be quite old. I don't get them here (with 0.4.4).
We're pulling 0.4.6 from debian, FWIW

panlinux:
ok, +1, merge away

PR #38, opened 2017-08-14
Document return codes on the manpage

### LINKED ISSUE #33: Document the exit codes in the manpage
The manpage should have a section describing the possible exit codes from the ubuntu-advantage script.

albertodonato:
Closes #33


panlinux:
Can you add a comment that we can also exit with apt-get status codes? Currently check_result is only wrapping apt-get.

albertodonato:
@panlinux done

panlinux:
If apt commands run by the tool fail, the return code from apt is returned.

Suggestion:

       If apt commands run by the tool fail, the exit status from apt is returned.

+1 regardless

albertodonato:
Good suggestion, updated

PR #46, opened 2017-08-16
Only enable/disable ESM on precise

### LINKED ISSUE #43: ubuntu-advantage, restrict esm to precise only
```
dpb@helo:~[]$ sudo ubuntu-advantage enable-esm foobar:foobar
Running apt-get update... ERROR
...
```

(Testing on Xenial)

We should probably keep a known good whitelist of releases and check that up front.

albertodonato:
Closes #43


panlinux:
+1, nice

PR #48, opened 2017-08-16
Add status command

### LINKED ISSUE #40: Status would be nice
Mockup:

```
$ ubuntu-advantage status
* livepatch: enabled
* esm: disabled (not available)

$ ubuntu-advantage livepatch status
<if enabled returns `canonical-livepatch status`>
<if disabled returns disabled>

$ ubuntu-advantage esm status
disabled (not available)
```

albertodonato:
Closes #40


ubuntu-advantage
```diff
@@ -199,11 +220,12 @@ Currently available are:
 - Canonical Livepatch Service (https://www.ubuntu.com/server/livepatch)
 
 Commands:
+ status                    show current status of Ubuntu advantage offerings
```
sparkiegeek: a/advantage/Advantage/

ubuntu-advantage
```diff
@@ -187,6 +187,31 @@ check_user() {
...
+        livepatch_output=$(canonical-livepatch status)
+    else
+        livepatch_status="disabled"
+        is_supported_series "$LIVEPATCH_SUPPORTED_SERIES" || livepatch_status="$livepatch_status (not available)"
+    fi
+    echo "livepatch: $livepatch_status"
+    if [ "$livepatch_output" ]; then
+        echo "$livepatch_output"
+    fi
+
```
panlinux: Please add an extra blank line in this output between the statuses of both services so that we get a clear separation between them, or else, at first glance, when livepatch is enabled for example, the "esm:" line might look as being another item of the livepatch status:
```
ubuntu@15-89:~$ ubuntu-advantage status
livepatch: enabled
kernel: 4.4.0-89.112-generic
fully-patched: false
version: "28.1"
esm: disabled (not available)
```
+1

panlinux:
+1 with a comment about adding an extra blank line between the services status output

panlinux:
Hm, almost forgot, the manpage needs an update about this new status command.

ubuntu-advantage
albertodonato: Done

panlinux:
+1

PR #51, opened 2017-08-16
Split tests file

albertodonato:
This split the tests file so that we have
 - `fakes.py` containing fake helper commands run by the script in tests
 - `testing.py` with the base test class (containing common setup and helper functions)
 - separate `test_*.py` files with tests for ESM, livepatch and other script commands

I plan to also rework the common `setUp` so that it's easier to set up tests for different series and they're more consistent (currently the default setup is for precise byt has livepatch enabled).
This branch only moves stuff around, there's no change in tests code.


panlinux:
Looks great, thanks

PR #56, opened 2017-09-01
Apt non interactive

panlinux:
Create a wrapper for apt-get that sets options for it to be non-interactive. Change all apt-get calls to use that wrapper, and if apt-get is called directly without setting these options, fail the tests.

I welcome comments about my approach to inject the APT_GET preamble using make_fake_binary(), I kind of felt it was too implicit. Also the error message isn't seen when the test fails, because the exit status is checked first. I could have changed that in all tests, but thought it best to submit this first and get some feedback.

### DIFF
--- a/tests/fakes.py
+++ b/tests/fakes.py
@@ -1,6 +1,6 @@
 # Fake for commands invoked by the script.
 
-SNAP_LIVEPATCH_INSTALLED = """#!/bin/sh
+SNAP_LIVEPATCH_INSTALLED = """\n
 if [ "$1" = "list" ]; then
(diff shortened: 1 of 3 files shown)


albertodonato:
Code changes look good, +1

WRT testing, I agree adding the automatic wrapper is a bit too implicit. My suggestion would be something like https://github.com/CanonicalLtd/ubuntu-advantage-script/pull/57 (based on this branch), which adds an extra fake for apt-get and specific tests for this behavior, so that other tests are unaffected by changes in the fakes.

PR #57, opened 2017-09-04
Apt not inteactive (based on Andreas' #56)

albertodonato:
Rework tests from https://github.com/CanonicalLtd/ubuntu-advantage-script/pull/56 to add explicit tests for apt-get being called with right parameters.


panlinux:
It looks like you added the shebang lines back, but that's ok, I can remove them in a follow-up branch.

Regarding your changes, I just cringe a bit that we have to now test the same code path twice: once to check that the dependency is installed, and another time to check that apt_get (and not apt-get) was used. That's the only thing I liked about my branch :)
+1

PR #58, opened 2017-09-07
Fipsua

joylatten:
Please review modifications to include enabling fips in ubuntu-advantage script.


panlinux:
Hi, could you take a look at the shell lint errors? I think they are all about double quoting variables. These checks do not run locally because the version of shellcheck in Ubuntu is too old :/

ubuntu-advantage
```diff
@@ -141,6 +154,109 @@ disable_esm() {
...
+                       strongswan strongswan-hmac
+
+    echo "Configuring FIPS..."
+    configure_fips
+    echo "Successfully configured FIPS. PLEASE REBOOT to complete FIPS enablement."
+}
+
+fips_enabled_check() {
+    if [ -f "$FIPS_ENABLED_FILE" ]; then
+                cat $FIPS_ENABLED_FILE
```
albertodonato: Why do you need to cat the file?

ubuntu-advantage
```diff
@@ -151,7 +267,7 @@ is_livepatch_enabled() {
     canonical-livepatch status >/dev/null 2>&1
 }
 
-validate_esm_token(){
+validate_esm_or_fips_token(){
```
albertodonato: I'd rename this to `validate_user_pass_token`, so it's more explicit

joylatten:
Ok, I will take a look at this today and get it fixed.

albertodonato:
@joylatten also, could you please add a new testfile with tests for FIPS (similarly to `test_esm.py` and `test_livepatch.py`)?

joylatten:
hmmm... one of the commits did contain a tests/test_fips.py for fips testcases.

albertodonato:
Also, please don't use `dpkg-query -W`, since it can give false results (see https://github.com/CanonicalLtd/ubuntu-advantage-script/pull/14#discussion_r113871684)

README.md
```diff
@@ -8,7 +8,7 @@ Currently it supports the following:
 
 - [Ubuntu Extended Security Maintenance](https://ubuntu.com/esm) archive.
 - [Canonical Livepatch](https://www.ubuntu.com/server/livepatch) service for managed live kernel patching.
-
+- [Canonical FIPS 140-2 Certified Modules] Install Configure and Enable FIPS modules.
```
albertodonato: This is missing a link to a web page

albertodonato:
@joylatten sorry, for some reason I was only looking at a single commit and didn't see the test file

tests/test_fips.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class FIPSTest(UbuntuAdvantageTest):
+
+    SERIES = 'xenial'
+
+    def test_enable_fips(self):
+        """The enable-fips option enables the FIPS repository."""
+        p = self.boot_cfg
+        p.write_text("")
```
albertodonato: `p` is unneded, please just `self.boot_cfg.write_text("")` (same for other tests below)

ubuntu-advantage
```diff
@@ -141,6 +154,109 @@ disable_esm() {
     fi
 }
 
+fips_prep_check() {
+    # check to see if fips pkgs already installed.
+    fips_hmacs="openssh-client-hmac openssh-server-hmac libssl1.0.0-hmac \
```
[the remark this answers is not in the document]
joylatten: ok

ubuntu-advantage
```diff
@@ -141,6 +154,109 @@ disable_esm() {
...
+                linux-fips strongswan-hmac"
```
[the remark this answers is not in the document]
joylatten: linux-fips contains the kernel cryptoapi's hmac. Each fips module has its own individual hmac deb package. These act like the keys to fips mode, thus I check each one.

ubuntu-advantage
```diff
@@ -141,6 +154,109 @@ disable_esm() {
...
+    for pkg in $fips_hmacs; do
+        dpkg-query -W $pkg > /dev/null 2>&1
+        if [ $? -eq 0 ]; then
+            echo "ERROR: Looks like FIPS already installed."
+            return 1
+        fi
+    done
+
+    # sanity check
+    case "$ARCH" in
```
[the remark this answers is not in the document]
joylatten: Ah yes, true. Moving this check to be done first.

ubuntu-advantage
```diff
@@ -141,6 +154,109 @@ disable_esm() {
...
+            ;;
+    esac
+    return 0
+}
+
+configure_fips() {
+    # if /boot has its own partition, then get the bootdevice
+    # Note: /boot/efi  does not count
+    bootdev=$(awk '!/^\s*#/ && $2 ~ /^\/boot\/?$/ { print $1 }' $FSTAB)
+    fips_params="fips=1"
```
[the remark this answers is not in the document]
joylatten: ok

ubuntu-advantage
```diff
@@ -141,6 +154,109 @@ disable_esm() {
...
+        check_result zipl
+    else
+        if ! test -d $FIPS_BOOT_CFG_DIR; then
+            mkdir $FIPS_BOOT_CFG_DIR > /dev/null 2>&1
+            if [ $? -ne 0 ]; then
+                echo "Failed to make directory, $FIPS_BOOT_CFG_DIR."
+                return 1
+            fi
+        fi
+        cat > "$FIPS_BOOT_CFG" <<EOF
```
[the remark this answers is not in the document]
joylatten: ok

ubuntu-advantage
joylatten: /proc/sys/crypto/fips_enabled contains either a "0" (fips mode is disabled) or a "1" (fips mode is enabled). Cat the file will tell me which one it is...

ubuntu-advantage
joylatten: ok

README.md
joylatten: ok, used "dpkg-query -s" instead. 
Unfortunately, we do not have a web page yet for FIPS. As soon as we do, I will add it here.

joylatten: I will remove the reference from here until we do get a web page

joylatten: hmmmm... is it ok to leave the reference showing the script supports this? Or should i remove it until we get a web page?

joylatten:
General question. I ran make lint after making all the suggested changes. I now only see these, 

In ubuntu-advantage line 210:
        echo -n 'Updating grub to enable fips ...'
             ^-- SC2039: In POSIX sh, echo flags are not supported.

These are everywhere we use "echo -n" in the script. Should I just ignore them?

panlinux:
Yes, ignore that particular lint error. You can see that the check that travis runs doesn't report SC2039. But it's reporting other errors, like "                    ^-- SC2086: Double quote to prevent globbing and word splitting."

joylatten:
Ok, cool! Thanks! I did notice the checks in info posted below did not have these, so was a bit confused. Thanks for clearing up.

README.md
albertodonato: You can leave the reference to FIPS, but remove the brackets until there's a link (otherwise they show in the output).
I'd just write it as:

- Canonical FIPS 140-2 Certified modules management

ubuntu-advantage
```diff
@@ -141,6 +154,116 @@ disable_esm() {
...
+    # if /boot has its own partition, then get the bootdevice
+    # Note: /boot/efi  does not count
+    bootdev=$(awk '!/^\s*#/ && $2 ~ /^\/boot\/?$/ { print $1 }' $FSTAB)
+    fips_params="fips=1"
+    if [ -n "$bootdev" ]; then
+        fips_params="$fips_params bootdev=$bootdev"
+    fi
+
+    if [ "$ARCH" = "s390x" ]; then
+        sed -i -e 's/^parameters\s*=.*/& '"$fips_params"'/' "$FIPS_BOOT_CFG"
```
albertodonato: I think you missed my previous comment about this:

You should not use `/` as sed s command separator since `$bootdev` can also contain `/`.
I usually use `s,<re>,<sub>,` in these cases.

joylatten: Oh gosh, my apologies, I did miss that one. Let me fix it now. And I also need to remove the brackets from README since fips does not yet have a website.

albertodonato:
Thanks for the PR, approved

PR #62, opened 2017-09-13
add a few code helpers, and cleanups

albertodonato:


panlinux:
Thanks. Would you mind adding another small set of changes to your branch, since it's a generic "cleanup" one?

All the  check_result calls should all follow this pattern:
```
echo -n "something... "
check_result some_command
```
Note the empty space after the "...": that's because check_result will either print an OK or an ERROR text.
A few such calls were added by the fips branch that don't follow that pattern: they, for example, have a space before the "..." and no space after it.

Would you mind fixing those? :)

Thanks, +1 after that.

PR #63, opened 2017-09-18
Check if FIPS just needs reboot to enable

### LINKED ISSUE #60: enable-fips should check the running kernel
Right now, if you run enable-fips twice, the second time will tell you that fips is already enabled, even though that's not yet true because the machine is still running the old kernel:
```
root@15-89:~# ubuntu-advantage enable-fips <redacted>
Running apt-get update... OK
Ubuntu FIPS PPA repository enabled.
Installing FIPS packages (this may take a while)...OK
Configuring FIPS...
Updating grub to enable fips ...OK
Successfully configured FIPS. PLEASE REBOOT to complete FIPS enablement.

root@15-89:~# ubuntu-advantage enable-fips <redacted>
ERROR: Looks like FIPS already installed.
root@15-89:~# echo $?
6
```

It could check the running kernel, and if it's not the fips kernel, repeat the reboot message.
```
$ uname -r
4.4.0-1002-fips
```

albertodonato:
Fixes #60

### DIFF
--- a/tests/test_fips.py
+++ b/tests/test_fips.py
@@ -27,6 +27,26 @@ class FIPSTest(UbuntuAdvantageTest):
             'Installing missing dependency apt-transport-https',
             process.stdout)
 
+    def test_enable_fips_already_enabled(self):
+        """If fips is already enabled, an error is returned."""
+        self.make_fake_binary('dpkg-query')
+        p = self.fips_enabled_file
+        p.write_text('1')
+        process = self.script('enable-fips', 'user:pass')
+        self.assertEqual(6, process.returncode)
+        self.assertEqual(
+            'FIPS is already enabled.', process.stdout.strip())
+
+    def test_enable_fips_installed_not_enabled(self):
+        """If fips is installed but not enabled an error is returned."""
+        self.make_fake_binary('dpkg-query')
+        process = self.script('enable-fips', 'user:pass')
+        self.assertEqual(6, process.returncode)
+        self.assertEqual(
+            'FIPS is already installed. '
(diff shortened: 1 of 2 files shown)


tests/test_fips.py
```diff
@@ -27,6 +27,25 @@ def test_enable_fips(self):
...
+        self.make_fake_binary('dpkg-query')
+        p = self.fips_enabled_file
+        p.write_text('1')
+        process = self.script('enable-fips', 'user:pass')
+        self.assertEqual(6, process.returncode)
+        self.assertEqual(
+            'FIPS is already enabled.', process.stdout.strip())
+
+    def test_enable_fips_installed_not_enabled(self):
+        """If fips is already installed but not enabled an is returned."""
```
panlinux: what? :)

ubuntu-advantage
```diff
@@ -180,7 +180,11 @@ fips_prep_check() {
                 linux-fips strongswan-hmac"
     for pkg in $fips_hmacs; do
         if is_package_installed "$pkg"; then
-            echo "ERROR: Looks like FIPS already installed."
+            if is_fips_enabled; then
+                echo 'FIPS is already enabled.'
+            else
+                echo 'FIPS is already installed. Please reboot to enable it.'
```
panlinux: "Please reboot into the fips kernel to enable it." better?

tests/test_fips.py
albertodonato: oops, left out "error" :)

ubuntu-advantage
albertodonato: Sure, fixed

PR #68, opened 2017-09-30
Rename script to advantage (Closes #67)

### LINKED ISSUE #67: Add a symlink so advantage -> ubuntu-advantage
It's been requested that we make the script available as just "advantage". The simple way seems to be to add a symlink advantage -> ubuntu-advantage, and also for the manpage and other documentation.

panlinux:
This closes #67


advantage.1
```diff
@@ -1,9 +1,8 @@
-.TH UBUNTU-ADVANTAGE 1  "28 April 2017" "" ""
+.TH ADVANTAGE 1  "28 April 2017" "" ""
 .SH NAME
-ubuntu-advantage \- Enable or disable Ubuntu Advantage offerings from
-Canonical.
+advantage \- Enable or disable Ubuntu Advantage offerings from Canonical.
 .SH SYNOPSIS
-.B ubuntu-advantage
+.B advantage
 <command> [parameters]
```
albertodonato: Could you add another line here to show both usage with both names?

panlinux:
How about the above change?

albertodonato:
+1!

PR #71, opened 2017-10-04
d/changelog update for v11

panlinux:


dpb1:
lgtm, let's figure out a way to not have to do this next time.  Maybe ack can tell us how.

PR #73, opened 2017-10-13
revert rename, add link as "ua"

albertodonato:


ubuntu-advantage
```diff
@@ -109,7 +109,7 @@ enable_livepatch() {
             echo "and run the following command one more time to complete the"
             echo "installation:"
             echo
-            echo "sudo advantage enable-livepatch $1"
+            echo "sudo ubuntu-advantage enable-livepatch $1"
```
panlinux: Can we use $(basename $0) here perhaps, instead of "ubuntu-advantage"?

panlinux: Better yet, ${SCRIPTNAME} is already defined

albertodonato: Good point, done

PR #74, opened 2017-10-13
Debian changelog for v12

panlinux:


panlinux:
Self-approving as checks have passed and this is just a d/changelog update.

Issue #79, opened 2017-10-27
configure_fips() doesn't have a validation mechanism if bootdev isn't a UUID.


slashdd:
Current code:

```
configure_fips() {
    local bootdev fips_params result

    # if /boot has its own partition, then get the bootdevice
    # Note: /boot/efi  does not count
    bootdev=$(awk '!/^\s*#/ && $2 ~ /^\/boot\/?$/ { print $1 }' "$FSTAB")
    fips_params="fips=1"
    if [ -n "$bootdev" ]; then
        fips_params="$fips_params bootdev=$bootdev"
    fi
```

It look if a separate /boot partition exist in FSTAB, if nothing is return (nonzero value) it skip the bootdev portion but what if the $bootdev is nonzero and doesn't contain a UUID  ?

The first field of /boot line could be a device name in the form of (/dev/sda) or a label (LABEL=Boot).
I think iff UUID is needed, then we should add a validation mechanism.

```
# if /boot has its own partition, then get the bootdevice
    # Note: /boot/efi  does not count
    bootdev=$(awk '!/^\s*#/ && $2 ~ /^\/boot\/?$/ { print $1 }' "$FSTAB")
    fips_params="fips=1"
    if [[ -n "$bootdev" && "$bootdev" == "UUID"* ]]; then
        fips_params="$fips_params bootdev=$bootdev"
```

albertodonato:
As discussed on IRC, device names, `LABEL` and `UUID` identifiers are all accepted as FIPS `bootdev`

PR #80, opened 2017-10-27
[fips] configure_fips() doesn't have a validation mechanism if bootde…

slashdd:
…v isn't a UUID.

Bug#79

Current code look if a separate /boot partition exist in FSTAB, if nothing is return (nonzero value) it skip the bootdev portion but what if the $bootdev is nonzero and doesn't contain a UUID ?

The first field of /boot line could be a device name in the form of (/dev/sdX,...) or a label (LABEL=Boot).
I think iff UUID is needed, then we should add a validation mechanism.

Signed-off-by: Eric Desrochers eric.desrochers@canonical.com

### DIFF
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -200,7 +200,7 @@ configure_fips() {
     # Note: /boot/efi  does not count
     bootdev=$(awk '!/^\s*#/ && $2 ~ /^\/boot\/?$/ { print $1 }' "$FSTAB")
     fips_params="fips=1"
-    if [ -n "$bootdev" ]; then
+    if [[ -n "$bootdev" && "$bootdev" == "UUID"* ]]; then
         fips_params="$fips_params bootdev=$bootdev"
     fi
 


albertodonato:
What are the acceptable values for FIPS `bootdev=` parameter?
I was assuming that anything that's valid in fstab (partition, `LABEL=` and `UUID=`)  would work. Is it not the case?

ubuntu-advantage
```diff
@@ -200,7 +200,7 @@ configure_fips() {
     # Note: /boot/efi  does not count
     bootdev=$(awk '!/^\s*#/ && $2 ~ /^\/boot\/?$/ { print $1 }' "$FSTAB")
     fips_params="fips=1"
-    if [ -n "$bootdev" ]; then
+    if [[ -n "$bootdev" && "$bootdev" == "UUID"* ]]; then
```
albertodonato: I think this could just be:

```
if [[ "$bootdev" == UUID=* ]; then
```

albertodonato:
Fixes #79

albertodonato:
Closing this as #79 has been closed

PR #84, opened 2017-10-31
check for CPU hardware support when enabling FIPS

### LINKED ISSUE #82: fips-enable should fail when the AES instruction is not found
On machines without the AES instruction set (tested on libvirt/KVM VM) the FIPS kernel fails to boot with fips=1 on the kernel cmdline. Changing the CPU profile to include the AES instruction set (with no other changes) allows the kernel to boot.

The AES instruction's availability can be checked by "grep -m1 -o aes /proc/cpuinfo" outputting nothing if AES is not available and "aes" if it is.

The ubuntu-advantage-tools should detect if AES instruction set is present and if not output an error instead of installing, and then the host failing to boot.

albertodonato:
Closes #82

### DIFF
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -26,6 +26,8 @@ SNAPD=${SNAPD:-"/usr/lib/snapd/snapd"}
 
 FIPS_ENABLED_FILE=${FIPS_ENABLED_FILE:-"/proc/sys/crypto/fips_enabled"}
 FSTAB=${FSTAB:-"/etc/fstab"}
+CPUINFO=${CPUINFO:-"/proc/cpuinfo"}
+
 if [ "$ARCH" = "s390x" ]; then
     FIPS_BOOT_CFG=${FIPS_BOOT_CFG:-"/etc/zipl.conf"}
 else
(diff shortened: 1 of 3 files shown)


ubuntu-advantage
```diff
@@ -165,11 +177,28 @@ disable_esm() {
...
+            if ! check_cpu_flag aes; then
+                echo 'FIPS requires AES CPU extensions'
+                return 1
+            fi
+            ;;
+
+        ppc64le)
+            power_cpu_ver="$(power_cpu_version)"
+            if [ -z "$power_cpu_ver" ] || [ "$power_cpu_ver" -lt 8 ]; then
+                echo 'FIPS not supported on this CPU'
```
panlinux: I think we could say here that we require POWER8 or higher. Something like this perhaps: "Unsupported POWER cpu: 8 or higher is required", or "FIPS requires POWER version 8 or higher", or "FIPS requires POWER8 or later". Better ask the power people how this "version" is usually referred to.

cpaelzer: https://en.wikipedia.org/wiki/POWER8
Non Authorative answer, but "FIPS requires POWER8 or later" sounds good to me.

PR #86, opened 2017-11-06
display installed tools package version

albertodonato:


ubuntu-advantage.1
```diff
@@ -17,6 +17,10 @@ It must be run with root privileges.
 .B
 status
 Show the status of Ubuntu Advantage offerings.
+.TP
+.B
+version
+Print tools version.
```
panlinux: Better to use the same text as in the help output: "Show the tool version". Or just "Show version".

PR #88, opened 2017-11-09
a few cleanups

albertodonato:
- group and reorder global variables
- add an `error_msg` helper, write all error messages to stderr
- make the function to create an APT source file generic

### DIFF
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -35,6 +35,10 @@ else
     FIPS_BOOT_CFG=${FIPS_BOOT_CFG:-"${FIPS_BOOT_CFG_DIR}/99-fips.cfg"}
 fi
 
+error_msg() {
+    echo "$@" >&2
+}
+
 check_result() {
     local result output
     result=0
@@ -42,7 +46,7 @@ check_result() {
     if [ $result -ne 0 ]; then
         echo "ERROR"
         if [ -n "$output" ]; then
-            echo "$output" >&2
+            error_msg "$output"
         fi
         exit $result
     else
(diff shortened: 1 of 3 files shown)


ubuntu-advantage
```diff
@@ -148,16 +167,9 @@ disable_livepatch() {
...
-deb https://${1}@${REPO_URL}/ubuntu ${SERIES} main
-# deb-src https://${1}@${REPO_URL}/ubuntu ${SERIES} main
-EOF
-}
-
 enable_esm() {
-    cp "${KEYRINGS_DIR}/${REPO_KEY_FILE}" "$APT_KEYS_DIR"
-    write_esm_list_file "$1"
+    cp "${KEYRINGS_DIR}/${ESM_REPO_KEY_FILE}" "$APT_KEYS_DIR"
+    write_apt_list_file "$ESM_REPO_LIST" "$ESM_REPO_URL" "$1"
```
panlinux: Maybe use a local var called $credentials here instead of $1

ubuntu-advantage
```diff
@@ -253,19 +265,11 @@ configure_fips() {
...
-EOF
-}
-
 enable_fips() {
     fips_prep_check || exit 6
 
     cp "${KEYRINGS_DIR}/${FIPS_REPO_KEY_FILE}" "$APT_KEYS_DIR"
-    write_fips_list_file "$1"
-
+    write_apt_list_file "$FIPS_REPO_LIST" "$FIPS_REPO_URL" "$1"
```
panlinux: Same here, use a local $credentials

panlinux: Or $token

ubuntu-advantage
panlinux: Or $token

PR #89, opened 2017-11-24
check credentials before adding repositories

### LINKED ISSUE #85: enabling repositories (FIPS,ESM) should only add repos for valid credentials
From https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1730361:

Hi,
while testing for an SRU in ua I found this which I think is less of a nice experience than it should be.

If you try FIPS via ua you might have no credentials, but want to try.
What happens is:

```
sudo ubuntu-advantage enable-fips xxx:xxx
Running apt-get update... ERROR
[...]
W: The repository 'https://private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu xenial Release' does not have a Release file.
E: Failed to fetch https://private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu/dists/xenial/main/binary-amd64/Packages 401 Unauthorized
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

Fine, I get why xxx:xxx isn't working.
But then it leaves my system in a bad state.

```
# apt update
[...]
Err:7 https://private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu xenial Release
  401 Unauthorized
Reading package lists... Done
E: The repository 'https://private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu xenial Release' does not have a Release file.
```

I'd ask you to check the credentials somehow and only add the repo IF those are good.
An alternative would be that if "on enablement" the fail is at the step "Running apt-get update... ERROR" then remove the repo you configured (the one in /etc/apt/sources.list.d/ubuntu-fips-xenial.list).

So it could look either like:
```
sudo ubuntu-advantage enable-fips xxx:xxx
Checking Credentials... ERROR
```
(no add repo happening)

or:
```
$ sudo ubuntu-advantage enable-fips xxx:xxx
Running apt-get update... ERROR
Rolling back repository... Ok
```

albertodonato:
Closes #85

### DIFF
--- a/tests/test_esm.py
+++ b/tests/test_esm.py
@@ -49,6 +49,49 @@ class ESMTest(UbuntuAdvantageTest):
         self.assertIn(
             'DEBIAN_FRONTEND=noninteractive', self.read_file('apt_get.env'))
 
+    def test_enable_esm_invalid_token(self):
+        """If token is invalid, an error is returned."""
+        message = (
+            'E: Failed to fetch https://esm.ubuntu.com/'
+            '  401  Unauthorized [IP: 1.2.3.4]')
+        self.make_fake_binary(
+            'apt-helper', command='echo "{}"; exit 1'.format(message))
+        process = self.script('enable-esm', 'user:pass')
+        self.assertEqual(3, process.returncode)
+        self.assertIn('Checking token... ERROR', process.stdout)
+        self.assertIn('Invalid token', process.stderr)
+
+    def test_enable_esm_invalid_token_trusty(self):
+        """Invalid token error is caught with apt-helper in trusty."""
+        message = 'E: Failed to fetch https://esm.ubuntu.com/  HttpError401'
+        self.make_fake_binary(
+            'apt-helper', command='echo "{}"; exit 1'.format(message))
+        process = self.script('enable-esm', 'user:pass')
+        self.assertEqual(3, process.returncode)
+        self.assertIn('Checking token... ERROR', process.stdout)
+        self.assertIn('Invalid token', process.stderr)
+
+    def test_enable_esm_error_checking_token(self):
+        """If token check fails, an error is returned."""
+        message = (
+            'E: Failed to fetch https://esm.ubuntu.com/'
+            '  404  Not Found [IP: 1.2.3.4]')
+        self.make_fake_binary(
+            'apt-helper', command='echo "{}"; exit 1'.format(message))
+        process = self.script('enable-esm', 'user:pass')
+        self.assertEqual(3, process.returncode)
+        self.assertIn('Checking token... ERROR', process.stdout)
+        self.assertIn(
+            'Failed checking token (404  Not Found [IP: 1.2.3.4])',
+            process.stderr)
+
+    def test_enable_esm_skip_token_check_no_helper(self):
+        """If apt-helper is not found, the token check is skipped."""
+        self.apt_helper.unlink()
+        process = self.script('enable-esm', 'user:pass')
+        self.assertEqual(0, process.returncode)
+        self.assertIn('Checking token... SKIPPED', process.stdout)
+
     def test_enable_esm_install_apt_transport_https_fails(self):
         """Stderr is printed if apt-transport-https install fails."""
         self.apt_method_https.unlink()
@@ -97,7 +140,7 @@ class ESMTest(UbuntuAdvantageTest):
             'Invalid token, it must be in the form "user:password"',
             process.stderr)
 
-    def test_enable_esm_invalid_token(self):
+    def test_enable_esm_invalid_token_format(self):
         """The ESM token must be specified as "user:password"."""
         process = self.script('enable-esm', 'foo-bar')
         self.assertEqual(3, process.returncode)
--- a/tests/test_fips.py
+++ b/tests/test_fips.py
@@ -123,13 +123,56 @@ class FIPSTest(UbuntuAdvantageTest):
             'Invalid token, it must be in the form "user:password"',
             process.stderr)
 
-    def test_enable_fips_invalid_token(self):
+    def test_enable_fips_invalid_token_format(self):
         """The FIPS token must be specified as "user:password"."""
         process = self.script('enable-fips', 'foo-bar')
         self.assertEqual(3, process.returncode)
         self.assertIn(
             'Invalid token, it must be in the form "user:password"',
             process.stderr)
+
+    def test_enable_fips_invalid_token(self):
+        """If token is invalid, an error is returned."""
+        message = (
+            'E: Failed to fetch https://esm.ubuntu.com/'
+            '  401  Unauthorized [IP: 1.2.3.4]')
+        self.make_fake_binary(
+            'apt-helper', command='echo "{}"; exit 1'.format(message))
+        process = self.script('enable-fips', 'user:pass')
+        self.assertEqual(3, process.returncode)
+        self.assertIn('Checking token... ERROR', process.stdout)
+        self.assertIn('Invalid token', process.stderr)
+
+    def test_enable_fips_invalid_token_trusty(self):
+        """Invalid token error is caught with apt-helper in trusty."""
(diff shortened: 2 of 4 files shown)


ubuntu-advantage
```diff
@@ -57,6 +57,16 @@ check_result() {
     fi
 }
 
+http_request_code() {
+    curl -ILs "$1" | awk '/^HTTP\// { code=$2; }; END { print code; }'
```
panlinux: We should check if we can add curl as a dependency of ubuntu-advantage-tools. If yes, then we need to update debian/control. We are already pulling in the gnutls library because of apt-transport-https, so if we go ahead with this then we should make sure we are using the gnutls version of libcurl3

ubuntu-advantage
```diff
@@ -313,10 +327,27 @@ is_livepatch_enabled() {
...
+
+    case $(http_request_code "$url") in
+        200)
+            return 0
+            ;;
+        401)
+            error_msg 'Invalid credentials'
+            ;;
+        *)
+            error_msg 'Failed checking credentials'
```
panlinux: We should include the http code we got here in this message.

ubuntu-advantage
```diff
@@ -85,12 +95,12 @@ install_package_if_missing_file() {
 # Write an APT sources file
 write_apt_list_file() {
     local repo_file="$1"
-    local repo_url="$2"
-    local credentials="$3"
 
+    local url
+    url=$(private_repo_url "$2" "$3")
```
panlinux: What is $2 and $3 now? I kind of like the idea of local meaningful variables, because then we probably don't need docstrings ;)

tests/test_esm.py
```diff
@@ -49,6 +49,22 @@ def test_enable_esm_install_apt_transport_https_apt_get_options(self):
         self.assertIn(
             'DEBIAN_FRONTEND=noninteractive', self.read_file('apt_get.env'))
 
+    def test_enable_esm_invalid_credentials(self):
+        """If credentials are invalid, an errors is returned."""
```
panlinux: "errors" typo

tests/test_esm.py
```diff
@@ -49,6 +49,22 @@ def test_enable_esm_install_apt_transport_https_apt_get_options(self):
...
+        self.make_fake_binary('curl', command='echo HTTP/1.1 401 Unauthorized')
+        process = self.script('enable-esm', 'user:pass')
+        self.assertEqual(1, process.returncode)
+        self.assertIn('Checking credentials... ERROR', process.stdout)
+        self.assertIn('Invalid credentials', process.stderr)
+
+    def test_enable_esm_error_checking_credentials(self):
+        """If credentials check fails, an error is returned."""
+        self.make_fake_binary('curl', command='exit 1')
```
panlinux: curl won't exit with non-zero if it encounters a 404, for example. You would get the same result in this test if it exited 0. We only check what it prints, right?

tests/test_fips.py
```diff
@@ -131,6 +131,22 @@ def test_enable_fips_invalid_token(self):
             'Invalid token, it must be in the form "user:password"',
             process.stderr)
 
+    def test_enable_fips_invalid_credentials(self):
+        """If credentials are invalid, an errors is returned."""
```
panlinux: "errors" typo

tests/test_esm.py
albertodonato: fixed

ubuntu-advantage
albertodonato: done

tests/test_esm.py
albertodonato: Correct, I've updated the test (so it also tests the printed HTTP code)

tests/test_fips.py
albertodonato: fixed

ubuntu-advantage
albertodonato: good point :) fixed

ubuntu-advantage
albertodonato: They don't seem to be alternatives. apt-transport-https depends on libcurl3-gnutls, whereas curl depends on libcurl3.
FWIW I don't see a reason we can't depend on curl.

panlinux: After a discussion in IRC (See https://irclogs.ubuntu.com/2017/11/27/%23ubuntu-devel.html#t11:34), we decided to abandon curl as it indeed pulls in too much for ubuntu-minimal. It was suggested that we try these:
 * from trusty onwards: `/usr/lib/apt/apt-helper download-file <url> <path>` (not available in precise)
 * or talk to the https method directly, like shown in http://paste.ubuntu.com/26057904/
    $ /usr/lib/apt/methods/http
    100 Capabilities
    Version: 1.2
    Pipeline: true
    Send-Config: true
    
    600 URI Acquire
    URI: https://jak-linux.org
    FileName: /tmp/test

panlinux: I'm fine with skipping this auth-check feature for precise. Talking to the https method over stdin/stdout/stderr via shell seems icky.

ubuntu-advantage
```diff
@@ -313,10 +325,50 @@ is_livepatch_enabled() {
...
+        rm -f "$tempfile" "$log"
+        return 0
+    fi
+
+    local error_line
+    error_line=$(sed -n 's/^E: Failed to fetch [^ ]\+ \+//p' "$log")
+    if echo "$error_line" | grep -q '^\(HttpError\)\?401'; then
+        error_msg 'Invalid credentials'
+    else
+        error_msg "Failed checking credentials ($error_line)"
```
panlinux: We should use the term "token" instead of credentials, that's how we refer to the pair foo:bar in all other communications with the user. This could also be changed in the code (functions, variables) or else it might get confusing.

ubuntu-advantage
```diff
@@ -313,10 +325,50 @@ is_livepatch_enabled() {
...
+}
+
+check_credentials() {
+    local repo_url="$1"
+    local credentials="$2"
+
+    # skip the check if the helper is not found
+    [ -x "$APT_HELPER" ] || return 0
+
+    echo -n 'Checking credentials... '
```
panlinux: What do you think about doing the APT_HELPER check after this echo, and output "SKIP" or "SKIPPED (no apt-helper found)", or some combination of these, if we don't find it?

ubuntu-advantage
```diff
@@ -313,10 +325,50 @@ is_livepatch_enabled() {
...
+    local file="$3"
+
+    local url
+    url=$(private_repo_url "$repo_url" "$credentials" "$file")
+
+    local tempfile log
+    tempfile="$(mktemp)"
+    log="$(mktemp)"
+    if "$APT_HELPER" download-file "$url" "$tempfile" >"$log" 2>&1; then
+        rm -f "$tempfile" "$log"
```
panlinux: We only need this tempfile because apt-helper won't write to /dev/null, right? Perhaps add a quick comment here about that, and that tempfile will just be discarded? Maybe even rename it to "discardfile" or something like that :)

albertodonato: done

ubuntu-advantage
albertodonato: Good point, done

ubuntu-advantage
albertodonato: done

ubuntu-advantage
```diff
@@ -313,10 +326,54 @@ is_livepatch_enabled() {
...
+    local error_line
+    error_line=$(sed -n 's/^E: Failed to fetch [^ ]\+ \+//p' "$log")
+    if echo "$error_line" | grep -q '^\(HttpError\)\?401'; then
+        error_msg 'Invalid token'
+    else
+        error_msg "Failed checking token ($error_line)"
+    fi
+
+    rm -f "$discard" "$log"
+    return 1
```
panlinux: Should we return 3 here perhaps? From the manpage:
   3      Invalid or missing token when enabling a service

PR #94, opened 2017-12-15
use shellcheck docker image for travis builds

albertodonato:
- add makefile target to run shellcheck from docker, so it's always the latest (this needs to run on trusty as precise images don't support docker)
- add separate lint Travis stage (which runs before the test one)
- run tests with python 3.4 and 3.6 (both on precise)

This PR depends on #97

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,25 @@
 language: python
+python:
+  - 3.4
+  - 3.6
+dist: precise
+
+services:
+  - docker
 
 matrix:
-  include:
-    - os: linux
-      dist: precise
-      python: 3.4
-    - os: linux
-      dist: trusty
-      python: 3.5
-    - os: linux
-      dist: trusty
-      python: 3.6
-      
-install: make testdep
-script:
-  - make lint
-  - make test
+  fast_finish: true
 
-addons:
-  apt:
-    sources:
-      - debian-sid
-    packages:
-      - shellcheck
+stages:
+  - lint
+  - test
+
+jobs:
+  include:
+    - stage: lint
+      script: make lint/docker
+      python: 3.6
+      dist: trusty
+
+install: make testdep
+script: make test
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -9,12 +13,23 @@ test:
...
 	@tox -e lint
 
-lint-sh:
-	@shellcheck -s bash ubuntu-advantage modules/*
-	@shellcheck -s dash update-motd.d/*
+lint-sh: SHELLCHECK = shellcheck
+lint-sh: BASH_SCRIPTS = $(UA_SCRIPTS)
+lint-sh: DASH_SCRIPTS = $(MOTD_SCRIPTS)
+lint-sh: _lint-sh-command
+
```
panlinux: Is this grouping of makefile targets the same as having just one target and the command lines below it, indented with a tab? Or what is its purpose?

.travis.yml
```diff
@@ -1,25 +1,25 @@
...
+
+services:
+  - docker
 
 matrix:
+  fast_finish: true
+
+stages:
+  - lint
+  - test
```
panlinux: This "test" stage is some sort of default? I don't see it referenced in a job later on

albertodonato: yes, it's the default stage. I need to declare the order here because by default stages are run after the test one.

Makefile
albertodonato: Target-specific variable declarations need to be on a target line on their own.
You can declare the same target multiple times, they're all run in the order of declaration.
What you can't do is declare a variable as part of a target command, IOW you can't
```
lint-sh:
     BASH_SCRIPTS = $(UA_SCRIPTS)
     ....
     <commands>
```

panlinux: Nice trick

panlinux:
Ok, in general, what are we trying to fix with this PR? I don't see changes in the shell code or in the SC checks, are we just getting ready for future changes?

albertodonato:
This is just to make use of build stages, so we only run lint once, and only if that passed, tests are run (in parallel) on the different python versions

albertodonato:
Also, it allows to use the latest shellcheck version, regardless of ubuntu release.

PR #95, opened 2017-12-18
Check that arch is supported when enabling livepatch

### LINKED ISSUE #93: Livepatch is only available for x86_64/amd64
Since livepatch is only available for x86_64/amd64 (64 bits, therefore), the ua-tools script could detect that and prevent the user from trying to enable it on an unsupported architecture.

For example, on s390x it's listed as available:
```
ubuntu@xenial-andreas:~$ ua status
esm: disabled (not available)
livepatch: disabled
fips: disabled
```

But won't install:
```
ubuntu@xenial-andreas:~$ sudo ua enable-livepatch <redacted>
Installing the canonical-livepatch snap.
This may take a few minutes depending on your bandwidth.
error: snap "canonical-livepatch" not found
```

uname:
```
ubuntu@xenial-andreas:~$ uname -a
Linux xenial-andreas 4.4.0-103-generic #126-Ubuntu SMP Mon Dec 4 16:22:13 UTC 2017 s390x s390x s390x GNU/Linux
```

albertodonato:
Fixes #93

### DIFF
--- a/modules/service-esm.sh
+++ b/modules/service-esm.sh
@@ -1,6 +1,8 @@
 # shellcheck disable=SC2039
 
 ESM_SUPPORTED_SERIES="precise"
+ESM_SUPPORTED_ARCHS=""
+
 ESM_REPO_URL="esm.ubuntu.com"
 ESM_REPO_KEY_FILE="ubuntu-esm-keyring.gpg"
 ESM_REPO_LIST=${ESM_REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-esm-${SERIES}.list"}
@@ -35,5 +37,7 @@ esm_is_enabled() {
 }
 
 esm_check_support() {
-    check_service_support "Extended Security Maintenance" "$ESM_SUPPORTED_SERIES"
+    check_service_support \
+        "Extended Security Maintenance" "$ESM_SUPPORTED_SERIES" \
+        "$ESM_SUPPORTED_ARCHS"
 }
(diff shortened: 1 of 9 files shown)


modules/service-esm.sh
```diff
@@ -1,6 +1,8 @@
 # shellcheck disable=SC2039
 
 ESM_SUPPORTED_SERIES="precise"
+ESM_SUPPORTED_ARCHS=""
```
panlinux: I think using a special key here is better than an empty string. How about "ALL", or "@all@", or something like that?

tests/test_livepatch.py
```diff
@@ -42,6 +43,15 @@ def test_enable_livepatch_invalid_token(self):
         self.assertEqual(3, process.returncode)
         self.assertIn('Invalid or missing Livepatch token', process.stderr)
 
+    def test_enable_livepatch_unsupported_on_non_x86_64(self):
+        """Livepatch is only supported on x86_64."""
```
panlinux: A more appropriate test name and docstring would be that livepatch is not supported on i686.

modules/service-esm.sh
panlinux: Also, I don't think the case of an empty list is tested. You could write a test that sets the arch to something exotic and that esm is still enabled (or using the new ALL special name, if you decide to accept that suggestion).

albertodonato: "ALL" sounds good to me, as it matches deb archs. Done

tests/test_livepatch.py
albertodonato: fixed

modules/service-esm.sh
albertodonato: added a test

PR #96, opened 2017-12-18
Use APT auth.conf for storing credentials

### LINKED ISSUE #22: sources.list file created for ESM is world-readable, leaks subscriber token to all local users
Filed in launchpad as https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1700611

The sources.list.d entry for esm is created with the default umask, which means that all local users on the system have access to the token. Being able to read globally-readable files on the filesystem does not necessarily mean you are an ESM subscriber who should have access to this token and be able to access the ESM archive.

We should probably create this file mode 0600. (Though it is too late to fix this for precise.)

albertodonato:
Fixes #22

### DIFF
--- a/modules/apt.sh
+++ b/modules/apt.sh
@@ -12,16 +12,24 @@ package_version() {
     dpkg-query -W -f '${Version}\n' "$1" 2>/dev/null
 }
 
-# Write an APT sources file
-write_apt_list_file() {
+apt_add_repo() {
     local repo_file="$1"
     local repo_url="$2"
     local credentials="$3"
+    local keyring_file="$4"
 
-    cat >"$repo_file" <<EOF
-deb https://${credentials}@${repo_url}/ubuntu ${SERIES} main
-# deb-src https://${credentials}@${repo_url}/ubuntu ${SERIES} main
-EOF
+    _apt_write_list_file "$repo_file" "$repo_url"
+    _apt_add_auth "$repo_url" "$credentials"
+    cp "$keyring_file" "$APT_KEYS_DIR"
+}
+
+apt_remove_repo() {
+    local repo_file="$1"
+    local repo_url="$2"
+    local keyring_file="$3"
+
+    rm -f "$repo_file" "$keyring_file"
+    _apt_remove_auth "$repo_url"
 }
 
 apt_get() {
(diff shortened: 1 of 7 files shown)


modules/service-esm.sh
```diff
@@ -9,8 +9,8 @@ esm_enable() {
     local token="$1"
 
     check_token "$ESM_REPO_URL" "$token"
```
panlinux: It's not a problem introduced in this branch, but it will show up only once this lands. We should check if esm is already enabled before enabling it again, or else we will pile up machine lines in auth.conf, not to mention running apt update multiple times needlessly.

modules/apt.sh
```diff
@@ -43,3 +51,38 @@ install_package_if_missing_file() {
...
+    local repo_url="$1"
+    local credentials="$2"
+
+    local login password
+    login=$(echo "$credentials" | cut -d: -f1)
+    password=$(echo "$credentials" | cut -d: -f2)
+    [ -f "$APT_AUTH_FILE" ] || touch "$APT_AUTH_FILE"
+    chmod 600 "$APT_AUTH_FILE"
+    echo "machine ${repo_url}/ubuntu/ login ${login} password ${password}" \
+         >>"$APT_AUTH_FILE"
```
panlinux: Once we check if esm is already enabled before calling this, we should be safe(r) and be able to avoid multiple identical lines in auth.conf, but maybe we should add a second check and only add our line if it doesn't exist already. Maybe just check for the url (careful with substring matches).

modules/apt.sh
```diff
@@ -43,3 +51,38 @@ install_package_if_missing_file() {
...
+
+_apt_remove_auth() {
+    local repo_url="$1"
+
+    local tempfile
+    tempfile=$(mktemp)
+    chmod 600 "$tempfile"
+    awk -v url="$repo_url" \
+        'BEGIN { pattern = "^machine " url "/ubuntu/ "; }; $0 !~ pattern;' \
+        "$APT_AUTH_FILE"  >"$tempfile"
```
panlinux: awk showoff :)
How about:
sed -i "/^machine $url\/ubuntu\/ login/d" $APT_AUTH_FILE

No need to use temp files.

That being said, I think I will file a bug about the suffix "url" being wrong for some of these variables. We are using $foo_url in many places where it's actually a hostname, not an url.

panlinux:
```diff
diff --git a/modules/apt.sh b/modules/apt.sh
index 742d471..955027c 100644
--- a/modules/apt.sh
+++ b/modules/apt.sh
@@ -76,13 +76,7 @@ _apt_add_auth() {
 }
 
 _apt_remove_auth() {
-    local repo_url="$1"
+    local repo_hostname="$1"
 
-    local tempfile
-    tempfile=$(mktemp)
-    chmod 600 "$tempfile"
-    awk -v url="$repo_url" \
-        'BEGIN { pattern = "^machine " url "/ubuntu/ "; }; $0 !~ pattern;' \
-        "$APT_AUTH_FILE"  >"$tempfile"
-    mv "$tempfile" "$APT_AUTH_FILE"
+    sed -i "/^machine $repo_hostname\/ubuntu\/ login" "$APT_AUTH_FILE"
 }
```

modules/service-esm.sh
albertodonato: Right, we should check if the service is enabled as first thing for each service. I'd defer this to a separate branch, since it's an issue already and I plan to unify the enable/disable logic.
I opened #99

modules/apt.sh
albertodonato: done, thanks

modules/apt.sh
albertodonato: I don't think we should go too deep with validating the content of the file. The user could always put crap in the file, we should just expect what we put there. As you said, this will be safer once we check before adding. Note that disable would remove all lines for the host, if there are many

tests/test_esm.py
```diff
@@ -174,6 +186,8 @@ def test_disable_esm(self):
         # the keyring file is removed
         keyring_file = self.trusted_gpg_dir / 'ubuntu-esm-keyring.gpg'
         self.assertFalse(keyring_file.exists())
+        # credentials are removed
+        self.assertEqual(self.apt_auth_file.read_text(), '')
```
panlinux: I think you should test that only our entry is removed, and not that the file is empty. I.e., start the test with some dummy unrelated entry and make sure that at the end it's the only thing remaining.

albertodonato: There's a separate test (`test_enable_fips_auth_if_other_entries`) for this

panlinux: That test doesn't check for removal, because there is no disabling of fips

albertodonato: Ah, you're right, sorry. Fixed the test.

PR #101, opened 2017-12-28
fix return value for FIPS arch check

### LINKED ISSUE #98: split _fips_prep_check() checks so that the exit status is correct
Currently the `_fips_prep_check` function performs two very different checks:

- if the current CPU has the minimum flags
- if the packages are already installed

In both cases, it returns 1, which is interpreted as a failure. The whole script then, however, returns 6, which according to the man page means:

`6      It was determined that FIPS has already been installed`

That is incorrect if what failed was the cpu flags check. We should probably move the first check into the new `check_service_support` changed in PR #95 so that we can exit 7 (also introduced in that PR).

albertodonato:
Fixes #98

### DIFF
--- a/modules/service-fips.sh
+++ b/modules/service-fips.sh
@@ -13,11 +13,13 @@ else
     FIPS_BOOT_CFG_DIR=${FIPS_BOOT_CFG_DIR:-"/etc/default/grub.d"}
     FIPS_BOOT_CFG=${FIPS_BOOT_CFG:-"${FIPS_BOOT_CFG_DIR}/99-fips.cfg"}
 fi
+FIPS_HMAC_PACKAGES="openssh-client-hmac openssh-server-hmac libssl1.0.0-hmac \
+        linux-fips strongswan-hmac"
 
-enable_fips() {
+fips_enable() {
     local token="$1"
 
-    _fips_prep_check || exit 6
+    _fips_check_packages_installed || exit 6
 
     check_token "$FIPS_REPO_URL" "$token"
     apt_add_repo "$FIPS_REPO_LIST" "$FIPS_REPO_URL" "$token" \
@@ -36,7 +38,7 @@ enable_fips() {
                  strongswan strongswan-hmac
 
     echo "Configuring FIPS... "
-    _configure_fips
+    _fips_configure
     echo "Successfully configured FIPS. Please reboot into the FIPS kernel to enable it."
 }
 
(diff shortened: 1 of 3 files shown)


modules/service-fips.sh
```diff
@@ -107,11 +109,11 @@ _fips_prep_check() {
...
 
-    # check to see if fips pkgs already installed.
-    fips_hmacs="openssh-client-hmac openssh-server-hmac libssl1.0.0-hmac \
-                linux-fips strongswan-hmac"
-    for pkg in $fips_hmacs; do
+_fips_check_packages_installed() {
+    local pkg
+    for pkg in $FIPS_HMAC_PACKAGES; do
         if is_package_installed "$pkg"; then
             if fips_is_enabled; then
```
panlinux: I know this wasn't changed in this branch, but I think this check should be better. It should check that all packages are installed before moving on to fips_is_enabled, because if there is a missing package, we will want to fix it by installing it. And we definitely shouldn't tell the user to reboot if we find out just one package is installed and not the rest.

modules/service-fips.sh
```diff
@@ -47,10 +49,11 @@ fips_is_enabled() {
 fips_check_support() {
     check_service_support \
         "Canonical FIPS 140-2 Modules" "$FIPS_SUPPORTED_SERIES" \
-        "$FIPS_SUPPORTED_ARCHS"
+        "$FIPS_SUPPORTED_ARCHS" || return 1
```
panlinux: check_service_support() just exits (4 or 7), this return 1 is for a "just in case" scenario? Maybe we should change check_service_support() to return 4 or 7 instead if exiting?

modules/service-fips.sh
```diff
@@ -88,9 +91,8 @@ _fips_enabled_check() {
     echo 0
 }
 
-_fips_prep_check() {
-    local fips_hmacs pkg power_cpu_ver
-
+_fips_check_arch() {
+    local power_cpu_ver
```
panlinux: I'm thinking _fips_check_arch() should return 7 in the case of failure, or its call in fips_check_support() should || return 7

modules/service-fips.sh
albertodonato: Can you file a separate bug for this?

panlinux: Sure, issue #102

modules/service-fips.sh
albertodonato: good catch, done

modules/service-fips.sh
albertodonato: ah, I think this was a leftover, removed the `|| return 1`

PR #103, opened 2018-01-02
Refactor enable-* and is-*-enabled commands

albertodonato:
Unify the logic for `enable-<service>` and `is-<service>-enabled` commands, reducing duplication.
A followup branch will take care of `disable-<service>` as well

### DIFF
--- a/modules/service-fips.sh
+++ b/modules/service-fips.sh
@@ -1,6 +1,6 @@
 # shellcheck disable=SC2034,SC2039
 
-FIPS_SERVICE_NAME="Canonical FIPS 140-2 Modules"
+FIPS_SERVICE_TITLE="Canonical FIPS 140-2 Modules"
 FIPS_SUPPORTED_SERIES="xenial"
 FIPS_SUPPORTED_ARCHS="x86_64 ppc64le s390x"
 
--- a/modules/service-fips.sh
+++ b/modules/service-fips.sh
@@ -54,6 +54,15 @@ fips_check_support() {
     _fips_check_arch
 }
 
+fips_validate_token() {
+    local token="$1"
+
+    if ! validate_user_pass_token "$token"; then
+        error_msg 'Invalid token, it must be in the form "user:password"'
+        return 1
+    fi
+}
+
 _fips_configure() {
     local bootdev fips_params result
 
--- a/modules/service-fips.sh
+++ b/modules/service-fips.sh
@@ -1,8 +1,8 @@
-# shellcheck disable=SC2039
+# shellcheck disable=SC2034,SC2039
 
-FIPS_SERVICE_NAME="Canonical FIPS 140-2 Modules" # shellcheck disable=SC2034
-FIPS_SUPPORTED_SERIES="xenial" # shellcheck disable=SC2034
-FIPS_SUPPORTED_ARCHS="x86_64 ppc64le s390x" # shellcheck disable=SC2034
+FIPS_SERVICE_NAME="Canonical FIPS 140-2 Modules"
+FIPS_SUPPORTED_SERIES="xenial"
+FIPS_SUPPORTED_ARCHS="x86_64 ppc64le s390x"
 
 FIPS_REPO_URL="private-ppa.launchpad.net/ubuntu-advantage/fips"
 FIPS_REPO_KEY_FILE="ubuntu-fips-keyring.gpg"
--- a/modules/service-fips.sh
+++ b/modules/service-fips.sh
@@ -47,13 +47,6 @@ fips_is_enabled() {
     is_package_installed fips-initramfs && [ "$(_fips_enabled_check)" -eq 1 ]
 }
 
-fips_check_support() {
-    check_service_support \
-        "Canonical FIPS 140-2 Modules" "$FIPS_SUPPORTED_SERIES" \
-        "$FIPS_SUPPORTED_ARCHS"
-    _fips_check_arch
-}
-
 fips_validate_token() {
     local token="$1"
 
@@ -61,6 +54,26 @@ fips_validate_token() {
         error_msg 'Invalid token, it must be in the form "user:password"'
         return 1
     fi
+}
+
+fips_check_support() {
+    local power_cpu_ver
+    case "$ARCH" in
+        x86_64)
+            if ! check_cpu_flag aes; then
+                error_msg 'FIPS requires AES CPU extensions'
+                return 7
+            fi
+            ;;
+
+        ppc64le)
+            power_cpu_ver="$(power_cpu_version)"
+            if [ -z "$power_cpu_ver" ] || [ "$power_cpu_ver" -lt 8 ]; then
+                error_msg 'FIPS requires POWER8 or later'
+                return 7
+            fi
+            ;;
+    esac
 }
 
 _fips_configure() {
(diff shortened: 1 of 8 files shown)


ubuntu-advantage
```diff
@@ -94,102 +94,76 @@ Commands:
...
-        livepatch_check_support
-        remove_snap="no"
-        if [ -n "$2" ]; then
-            if [ "$2" = "-r" ]; then
-                remove_snap="yes"
+    local service
+    service=$(service_from_command "$command")
+    # if the command contains a service name, check that it's valid
+    if [ "$service" ] && ! name_in_list "$service" "$SERVICES"; then
+        usage
```
panlinux: before displaying the usage(), we should print why we are doing it. i.e., "echo invalid comand $command"

ubuntu-advantage
```diff
@@ -94,102 +94,76 @@ Commands:
  enable-fips <token>       enable the FIPS PPA repository and install,
                            configure and enable FIPS certified modules
  disabe-fips               currently not supported
+
 EOF
+    exit 1
 }
 
+main() {
+    local command="$1"
```
panlinux: Since this is a function now (main), we should also declare the "remove_snap" local variable used later on.

modules/service.sh
```diff
@@ -0,0 +1,52 @@
...
+         /^is-.*-enabled$/ {
+             match($0, /^is-(.*)-enabled$/, m);
+             print m[1];
+         }
+         /^(enable|disable)-/ {
+             match($0, /^(enable|disable)-(.*)/, m);
+             print m[2];
+         }
+    '
+}
```
panlinux: This does the same, FTR:
echo "$command" |  sed -r 's,^is-([^-]+)-enabled$,\1,;s,^(enable|disabled)-([^-]+)$,\2,'

modules/utils.sh
```diff
@@ -38,3 +38,13 @@ name_in_list() {
...
 }
+
+expand_var() {
+    local var="$1"
+
+    echo "${!var}"
+}
+
+uppercase() {
+    echo "$@" | tr '[:lower:]' '[:upper:]'
```
panlinux: Since we are using bash, this can be:
```
local parameters="$@"
echo ${parameters^^}
```
Or something along those lines. The trick is ${var^^}

I think we have just one caller for this, in:
./modules/service.sh:    service_upper=$(uppercase "$service")

That could be just "service_upper=${service^^} then.

modules/service.sh
```diff
@@ -0,0 +1,52 @@
...
+    "${service}_enable" "$token"
+}
+
+service_is_enabled() {
+    local service="$1"
+
+    "${service}_is_enabled"
+}
+
+service_check_support() {
```
panlinux: We have check_service_support() and service_check_support() now, this is getting confusing.

modules/service.sh
```diff
@@ -0,0 +1,52 @@
...
+    local service="$1"
+
+    local service_upper
+    service_upper=$(uppercase "$service")
+    local name series archs
+    name=$(expand_var "${service_upper}_SERVICE_NAME")
+    series=$(expand_var "${service_upper}_SUPPORTED_SERIES")
+    archs=$(expand_var "${service_upper}_SUPPORTED_ARCHS")
+    check_service_support "$name" "$series" "$archs"
```
panlinux: This check_service_support() function doesn't really care about the service, does it? The "$name" parameter is there just for it to echo it out.

modules/service.sh
panlinux: Maybe it's time to better distinguish really internal functions from functions a module writer should use. Use a "_" prefix maybe. That being said, this confusion between check_service_support and service_check_support can only be sorted with a new name I think.

modules/service-fips.sh
```diff
@@ -1,5 +1,6 @@
-# shellcheck disable=SC2039
+# shellcheck disable=SC2034,SC2039
 
+FIPS_SERVICE_NAME="Canonical FIPS 140-2 Modules"
```
panlinux: I think these variables <service>_SERVICE_NAME should be called <service>_SERVICE_TITLE, because we are already using "esm", "livepatch" and "fips" as names elsewhere.

modules/service.sh
```diff
@@ -0,0 +1,52 @@
...
+    call_if_defined "${service}_check_support"
```
panlinux: Isn't this a bit obscure? Shouldn't we just require that all services have a $service_check_support function, even if it's a no-op?

modules/utils.sh
```diff
@@ -19,11 +19,11 @@ check_result() {
...
 
-check_user() {
-    if [ "$(id -u)" -ne 0 ]; then
-        error_msg "This command must be run as root (try using sudo)"
-        exit 2
-    fi
+call_if_defined() {
+    local command="$1"
+
+    type -t "$command" >/dev/null || return 0
```
panlinux: I'd make sure type -t is returning "function", or else if someone puts a $service_check_support script in $PATH, or shell alias, or just a file in the current directory, we will call out to that.

modules/service.sh
albertodonato: and wins the obfuscated sed contest in the process :)

ubuntu-advantage
albertodonato: good catch, done

modules/utils.sh
albertodonato: Yes, I actually fix this (along with other things that can be simplfied in bash) in a followup branch (since we switched to bash after I started this)

modules/service.sh
albertodonato: I agree, `check_service_support` is badly named, I'm not sure if it's better, but I'm renaming it to `check_support` (which matches `check_user`).

The idea is to have functions from each module to use the module name prefix (hence `service_`, `esm_`, `apt_`, ... ) but some generic functions don't really follow that naming.

I've been using the `_` prefix for functions that are only meant for use inside the file where they're declared.

Hopefully the rename above should be enough to make the distinction clearer

modules/service.sh
albertodonato: Yes, hence I dropped the `service` part in the function name now :)

modules/utils.sh
albertodonato: This is actually intentional, so it's possible to use `call_if_defined` to run a command only if it exists (whithout having to run `which`). Note that a function takes precedence over a command in path.

modules/service-fips.sh
albertodonato: Done

ubuntu-advantage
albertodonato: done

modules/service.sh
albertodonato: Yeah I didn't really want to end up with a lot of empty functions if they're not required.
I use this for another case (printing the service extended status) as well in a followup branch.

modules/utils.sh
panlinux: Ok, I'm a bit worried about outside manipulation of what we run. I tried setting an alias for fips_check_support,  because an alias takes precedence to a function:
```
# fips_check_support() {
> echo function busted
> }

# type -t fips_check_support
function

# alias fips_check_support="echo alias busted"

# type -t fips_check_support
alias

# fips_check_support
alias busted
```
But neither get "injected" into the script.

PR #105, opened 2018-01-03
Check if a service is already enabled before enabling

### LINKED ISSUE #99: enable-* commands should check if the service is enabled
All `enable-<service>` commands should check first if the service is already enabled.

albertodonato:
Fixes #99

### DIFF
--- a/modules/service.sh
+++ b/modules/service.sh
@@ -21,7 +21,7 @@ service_enable() {
 
     check_user
     service_check_support "$service"
-    _service_check_enabled "$service"
+    _service_check_enabled "$service" || exit 6
     "${service}_validate_token" "$token" || exit 3
     "${service}_enable" "$token"
 }
--- a/modules/service.sh
+++ b/modules/service.sh
@@ -21,6 +21,7 @@ service_enable() {
 
     check_user
     service_check_support "$service"
+    _service_check_enabled "$service"
     "${service}_validate_token" "$token" || exit 3
     "${service}_enable" "$token"
 }
@@ -36,4 +37,17 @@ service_check_support() {
 
     check_series_arch_supported "$service"
     call_if_defined "${service}_check_support"
+}
+
+_service_check_enabled() {
+    local service="$1"
+
+    local service_upper title
+    service_upper=$(uppercase "$service")
+    title=$(expand_var "${service_upper}_SERVICE_TITLE")
+
+    if service_is_enabled "$service"; then
+        error_msg "$title is already enabled"
+        return 1
+    fi
 }
--- a/tests/test_livepatch.py
+++ b/tests/test_livepatch.py
@@ -87,6 +87,12 @@ class LivepatchTest(UbuntuAdvantageTest):
 
     def test_is_livepatch_enabled_false(self):
         """is-livepatch-enabled returns 1 if the service is not enabled."""
+        self.setup_livepatch(installed=False)
+        process = self.script('is-livepatch-enabled')
+        self.assertEqual(1, process.returncode)
+
+    def test_is_livepatch_enabled_false_not_instaled(self):
+        """is-livepatch-enabled returns 1 if the service is not installed."""
         process = self.script('is-livepatch-enabled')
         self.assertEqual(1, process.returncode)
 
@@ -94,7 +100,7 @@ class LivepatchTest(UbuntuAdvantageTest):
         """enable-livepatch when it's already enabled is detected."""
         self.setup_livepatch(installed=True, enabled=True)
         process = self.script('enable-livepatch', self.livepatch_token)
-        self.assertEqual(1, process.returncode)
+        self.assertEqual(6, process.returncode)
         self.assertIn('Canonical Livepatch is already enabled', process.stderr)
 
     def test_enable_livepatch(self):
--- a/tests/test_livepatch.py
+++ b/tests/test_livepatch.py
@@ -12,7 +12,7 @@ class LivepatchTest(UbuntuAdvantageTest):
 
     def setUp(self):
         super().setUp()
-        self.setup_livepatch(installed=True, enabled=True)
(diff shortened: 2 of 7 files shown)


modules/service.sh
```diff
@@ -21,6 +21,7 @@ service_enable() {
 
     check_user
     service_check_support "$service"
+    _service_check_enabled "$service"
```
panlinux: I think we should `|| exit 6` here and change the manpage accordingly. Currently 6 is only for "fips is already enabled", it can now be "$service is already enabled". Or, if we don't want to change existing meaning, we could create a new exit status.

tests/testing.py
```diff
@@ -58,6 +62,7 @@ def setUp(self):
         self.apt_auth_file = Path(self.tempdir.join('auth.conf'))
         self.apt_method_https = self.bin_dir / 'apt-method-https'
         self.ca_certificates = self.bin_dir / 'update-ca-certificates'
+        self.snap = self.bin_dir / 'snap'
```
panlinux: What is this used for? Removing it doesn't make any tests fail.

tests/testing.py
```diff
@@ -130,17 +135,21 @@ def script(self, *args, env_update=None):
...
         """Setup livepatch-related fakes."""
-        if installed is not None:
-            command = (
-                SNAP_LIVEPATCH_INSTALLED if installed
-                else SNAP_LIVEPATCH_NOT_INSTALLED)
-            self.make_fake_binary('snap', command=command)
+        command = (
+            SNAP_LIVEPATCH_INSTALLED if installed
+            else SNAP_LIVEPATCH_NOT_INSTALLED)
+        self.make_fake_binary('snap', command=command)
```
panlinux: Can we make the same simplification for the enabled=None logic? I don't remember why we had these default to None. Tests still pass, but am I trusting them too much? :)
```diff
--- a/tests/testing.py
+++ b/tests/testing.py
@@ -135,19 +135,18 @@ class UbuntuAdvantageTest(TestWithFixtures):
         process.stderr.close()
         return result
 
-    def setup_livepatch(self, installed=False, enabled=None,
+    def setup_livepatch(self, installed=False, enabled=False,
                         livepatch_command=None):
         """Setup livepatch-related fakes."""
         command = (
             SNAP_LIVEPATCH_INSTALLED if installed
             else SNAP_LIVEPATCH_NOT_INSTALLED)
         self.make_fake_binary('snap', command=command)
-        if enabled is not None:
-            if livepatch_command is not None:
-                command = livepatch_command
-            else:
-                command = LIVEPATCH_ENABLED if enabled else LIVEPATCH_DISABLED
-            self.make_fake_binary('canonical-livepatch', command=command)
+        if livepatch_command is not None:
+            command = livepatch_command
+        else:
+            command = LIVEPATCH_ENABLED if enabled else LIVEPATCH_DISABLED
+        self.make_fake_binary('canonical-livepatch', command=command)
 
     def setup_esm(self, enabled=False):
         """Setup the ESM repository."""
```

tests/testing.py
albertodonato: Probably a leftover, I think I had it referenced before, dropping it

modules/service.sh
albertodonato: yeah that makes sense, I'll change it to `exit 6`

tests/testing.py
albertodonato: IIRC the reason for `None` there is that no `canonical-livepatch` command is present. I think we do check if the binary is present, so maybe tests pass because we still default to disabled.

tests/test_livepatch.py
```diff
@@ -72,46 +73,47 @@ def test_enable_livepatch_apt_get_options(self):
 
     def test_enable_livepatch_installs_snap(self):
         """enable-livepatch installs the livepatch snap if needed."""
-        self.setup_livepatch(installed=False)
+        self.setup_livepatch(enabled=False)
```
panlinux: Can this setup_livepatch() be dropped, since the default for enabled is False already?

modules/service.sh
panlinux: Should we perhaps move the enabled check to after the token check? Just noticed this:
```
ubuntu@ubuntu:~$ sudo ua enable-livepatch 
Canonical Livepatch is already enabled
ubuntu@ubuntu:~$ 
```

albertodonato: Maybe, the only downside is that in the case of ESM the token is checked with an HTTP get, which takes a bit more time. I think either way is fine, maybe file a separate bug?

tests/test_livepatch.py
albertodonato: Well, that depends on the comment above, since the default is `None` at the moment

modules/service.sh
panlinux: Ah, right, I thought it was just a simple presence check, or syntax. Didn't realize the validity of the token was included. Leave it like this then.

tests/test_livepatch.py
```diff
@@ -72,46 +73,47 @@ def test_enable_livepatch_apt_get_options(self):
...
 
     def test_is_livepatch_enabled_true(self):
         """is-livepatch-enabled returns 0 if the service is enabled."""
+        self.setup_livepatch(installed=True, enabled=True)
         process = self.script('is-livepatch-enabled')
         self.assertEqual(0, process.returncode)
 
     def test_is_livepatch_enabled_false(self):
         """is-livepatch-enabled returns 1 if the service is not enabled."""
-        self.setup_livepatch(enabled=False)
```
panlinux: So here we are effectivelly calling setup_livepatch(enabled=None), aren't we changing the test a little bit? It still passes either way.

albertodonato: yeah I think this can be reverted (or we could have 2 tests)

PR #106, opened 2018-01-08
add repository pinning for FIPS packages

### LINKED ISSUE #61: Fips packages should be pinned or held to avoid fips-breaking upgrades
After enabling FIPS, several packages are replaced by their FIPS counterparts. These come from a special FIPS PPA provided by Canonical.

Normal system upgrades, however, could replace these FIPS packages with the non-FIPS ones, thus rendering the installation uncertified.

It sounds reasonable that the ubuntu-advantage script, after having enabled fips and installed these extra needed packages, should somehow mark these as having a higher priority than the normal ubuntu repositories. This could be achieved either via "holds", or by pinning the fips PPA at a higher priority.

The PPA pinning sounds more attractive at first glance because it would rank any package that comes from that PPA as having a higher priority, so it's future proof.

albertodonato:
Fixes #61


tests/testing.py
```diff
@@ -50,10 +50,13 @@ class UbuntuAdvantageTest(TestWithFixtures):
     def setUp(self):
         super(UbuntuAdvantageTest, self).setUp()
         self.tempdir = self.useFixture(TempDir())
-        self.repo_list = Path(self.tempdir.join('repo.list'))
         self.boot_cfg = Path(self.tempdir.join('boot.cfg'))
         self.fstab = Path(self.tempdir.join('fstab'))
         self.cpuinfo = Path(self.tempdir.join('cpuinfo'))
+        self.esm_repo_list = Path(self.tempdir.join('esm-repo.list'))
+        self.fips_repo_list = Path(self.tempdir.join('esm-repo.list'))
```
joylatten: Should this be self.tempdir.join('fips-repo.list')?

albertodonato: good catch, fixed

modules/service-fips.sh
```diff
@@ -25,6 +26,8 @@ fips_enable() {
     check_token "$FIPS_REPO_URL" "$token"
     apt_add_repo "$FIPS_REPO_LIST" "$FIPS_REPO_URL" "$token" \
                  "${KEYRINGS_DIR}/${FIPS_REPO_KEY_FILE}"
+    apt_add_repo_pinning "$FIPS_REPO_PREFERENCES" \
+                         LP-PPA-ubuntu-advantage-fips 1001
```
panlinux: Since this is in the `fips_enable()` path, it doesn't look like we will get this feature by merely upgrading ubuntu-advantage-fips in a fips-enabled system? People should disable (which is not possible with fips) and re-enable?

albertodonato: True, but it's the same with other things (like authentication credentials in a separate auth file).
Supporting upgrades gets complicated, I would not worry about it in this case

albertodonato: I guess we could support `fips-disable` (by removing the repositories as we do for esm)?

vineethapai: fips-enable does more than enabling repositories unlike esm. It sets FIPS parameters kernel command line, configures boot loader etc. "fips-disable" as a feature may need to undo all that, or else the system will continue to run in fips mode.
For new users, pinning repos through fips_enable may work when this feature is made available through ubuntu-advantage tool. For existing users who already enabled fips, needs a different way to pin their repos.

panlinux: We could create the pinning config file in the package's postinst, if we detect that fips is enabled already.

vineethapai: That may be a good way to do it when ubuntu-advantage package undergoes update.

panlinux: We should just probably not use `ubuntu-advantage is-fips-enabled` in a maintainer script, because that calls `apt-cache policy`, and there could be locks in play here. We would be calling that inside an apt transaction. I'm thinking a simple check for config files plus perhaps running kernel for this.

panlinux: Ok, let's leave this for another time. It's not worth handling the upgrade case, and it can be sorted via documentation and instructions.

panlinux: Let me just give this branch a run and then I'll approve it.

albertodonato: @panlinux +1 on considering adding the pinning in the postints, but it seems we can defer it

panlinux: Found a bug in modules/apt.sh in master, which I fixed as a `[trivial]`. That's why I merged master back into this.

panlinux:
I'm having trouble correctly enabling fips on a xenial vm. After the packages are installed and I reboot, it stays stuck:
image

I'll debug this further. I don't think it's related to this branch. I have noticed boot delays in the past in fips-enabled systems, but not this long.

panlinux:
Ok, managed to login:
```
root@04-57:~# systemd-analyze blame | head
    5min 21.823s cloud-init-local.service
          1.798s cloud-config.service
          1.262s dev-vda1.device
          1.115s cloud-init.service
          1.023s strongswan.service
           766ms cloud-final.service
```

That's quite a delay

PR #109, opened 2018-01-20
Refactor disable commands

albertodonato:
This depends on #105

### DIFF
--- a/modules/service.sh
+++ b/modules/service.sh
@@ -1,12 +1,5 @@
 # shellcheck disable=SC2039
 
-check_user() {
-    if [ "$(id -u)" -ne 0 ]; then
-        error_msg "This command must be run as root (try using sudo)"
-        exit 2
-    fi
-}
-
 service_from_command() {
     local command="$1"
 
@@ -19,11 +12,21 @@ service_enable() {
     local service="$1"
     local token="$2"
 
-    check_user
-    service_check_support "$service"
+    _service_check_user
+    _service_check_support "$service"
     _service_check_enabled "$service" || exit 6
     "${service}_validate_token" "$token" || exit 3
     "${service}_enable" "$token"
+}
+
+service_disable() {
+    local service="$1"
+
+    _service_check_user
+    _service_check_support "$service"
+    _service_check_disabled "$service" || exit 8
+    shift 1
+   "${service}_disable" "$@"
 }
 
 service_is_enabled() {
@@ -32,7 +35,14 @@ service_is_enabled() {
     "${service}_is_enabled"
 }
 
-service_check_support() {
+_service_check_user() {
+    if [ "$(id -u)" -ne 0 ]; then
+        error_msg "This command must be run as root (try using sudo)"
+        return 2
+    fi
+}
+
+_service_check_support() {
     local service="$1"
 
     check_series_arch_supported "$service"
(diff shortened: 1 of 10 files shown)


modules/service.sh
```diff
@@ -19,20 +12,37 @@ service_enable() {
...
     "${service}_validate_token" "$token" || exit 3
     "${service}_enable" "$token"
 }
 
+service_disable() {
+    local service="$1"
+
+    _service_check_user
+    _service_check_support "$service"
+    _service_check_disabled "$service"
```
panlinux: Like we did for service-already-enabled, should we add a "|| exit 8"' here if the service is already disabled, and add this new exit code to the manpage? ("8" was the next available one)

tests/test_fips.py
```diff
@@ -60,7 +58,7 @@ def test_enable_fips_already_enabled(self):
 
     def test_enable_fips_installed_not_enabled(self):
         """If fips is installed but not enabled an error is returned."""
-        self.make_fake_binary('dpkg-query')
+        self.setup_fips()
```
panlinux: We may not gain much, but I would think it more logical to call setup_fips() in the test setUp() with some reasonable default. ESM and livepatch already do this.
I counted the setup_fips() calls in the existing tests and we call it about 3 times to enable fips, and 3 times to disable it, so whatever our default state should be is a 50-50 chance of being correct most of the time :)

tests/test_fips.py
```diff
@@ -226,37 +224,29 @@ def test_enable_fips_ppc64le_older_power(self):
...
+        self.setup_fips(enabled=True)
         process = self.script('disable-fips')
         self.assertEqual(1, process.returncode)
-        self.assertEqual(
-            'Disabling FIPS is currently not supported.',
-            process.stderr.strip())
+        self.assertIn(
+            'Disabling FIPS is currently not supported.', process.stderr)
 
     def test_disable_fips_warns_if_not_enabled(self):
```
panlinux: I think the name of this test, and its docstring, should be changed now that this condition is an error and not a warning, even though we exited non-zero before. Maybe `test_disable_fips_fails_if_not_enabled()`.

tests/test_livepatch.py
```diff
@@ -121,21 +121,23 @@ def test_disable_livepatch_invalid_remove_snap_option(self):
     def test_disable_livepatch_already_disabled(self):
         """disable-livepatch when it's already disabled is detected."""
```
panlinux: same as above, `test_disable_livepatch_fails_if_not_enabled()`

tests/test_fips.py
albertodonato: Yeah but we only want to call `setup_fips()` if there's something to set up. The FIPS file must not always be there, and always mocking `dpkg-query` migth cover failures

modules/service.sh
```diff
@@ -19,20 +12,37 @@ service_enable() {
     local service="$1"
     local token="$2"
 
-    check_user
-    service_check_support "$service"
-    _service_check_enabled "$service" || exit 6
+    _service_check_user
+    _service_check_support "$service"
+    _service_check_enabled "$service"  || exit 6
```
panlinux: extra space before the ||

tests/testing.py
```diff
@@ -152,3 +152,11 @@ def setup_esm(self, enabled=False):
         """Setup the ESM repository."""
         command = ESM_ENABLED if enabled else ESM_DISABLED
         self.make_fake_binary('apt-cache', command=command)
+
+    def setup_fips(self, enabled=None):
+        """Setup FIPS."""
+        if enabled is None:
+            return
+
```
panlinux: Was this blank line on purpose?

PR #110, opened 2018-01-31
"update-fips" support to allow FIPS package updates

vineethapai:


vineethapai:
The patchset contains changes to install updates to all the FIPS packages from ubuntu-advantage/fips-updates ppa.

albertodonato:
Why do we need a command here to update packages? Isn't a standard `apt update && apt upgrade` sufficient?

vineethapai:
albertodonato - The updates to FIPS packages are hosted on a separate "fips-updates" ppa as they are uncertified. The "update-fips" command is to enable access to that ppa by adding credentials to sources.list,  install the updates, configure boot loader for a kernel update, and also to configure fips on the machine.  For subsequent updates after initial update, it does more than apt update as updating boot loader is required for kernel updates.

albertodonato:
I'm not sure I have the full picture here. What's the use case for `update-fips` if doing so takes the system out of compliance? Where will certified updates for FIPS packages be hosted?

vineethapai:
At this point there is no plan to certify the FIPS updates. There are a number of customer queries (infact almost all of them) to run the packages in just "FIPS" mode meaning to execute FIPS code path without being strictly certified. They would like updates to be applied to modules. For example landing the recent spectre, meltdown fixes to FIPS kernel.

panlinux:
This is ready for rebasing

PR #111, opened 2018-02-02
Livepatch motd

panlinux:
This branch adds a MOTD script for Livepatch, and tests. Lots of tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ testdep:
 	pip install tox
 
 test:
-	@tox -e python
+	@tox -e py3
 
 lint: lint-py lint-sh
 
@@ -19,17 +19,14 @@ lint-py:
 	@tox -e lint
 
 lint-sh: SHELLCHECK = shellcheck
-lint-sh: BASH_SCRIPTS = $(UA_SCRIPTS)
-lint-sh: DASH_SCRIPTS = $(MOTD_SCRIPTS)
+lint-sh: BASH_SCRIPTS = $(UA_SCRIPTS) $(MOTD_SCRIPTS)
 lint-sh: _lint-sh-command
 
 lint-sh/docker: SHELLCHECK = docker run --rm -it -v $(PWD):/repo koalaman/shellcheck
-lint-sh/docker: BASH_SCRIPTS = $(patsubst %,/repo/%,$(UA_SCRIPTS))
-lint-sh/docker: DASH_SCRIPTS = $(patsubst %,/repo/%,$(MOTD_SCRIPTS))
+lint-sh/docker: BASH_SCRIPTS = $(patsubst %,/repo/%,$(UA_SCRIPTS) $(MOTD_SCRIPTS))
 lint-sh/docker: _lint-sh-command
 
 _lint-sh-command:
 	@$(SHELLCHECK) -s bash $(BASH_SCRIPTS)
-	@$(SHELLCHECK) -s dash $(DASH_SCRIPTS)
 
 .PHONY: build testdep test lint lint/docker lint-py lint-sh lint-sh/docker _lint-sh-command
--- a/tests/fakes.py
+++ b/tests/fakes.py
@@ -33,9 +33,22 @@ fi
 LIVEPATCH_ENABLED = """\n
 if [ "$1" = "status" ]; then
     cat <<EOF
-kernel: 4.4.0-87.110-generic
-fully-patched: true
-version: "27.3"
+client-version: "7.23"
+architecture: x86_64
+cpu-model: QEMU Virtual CPU version 2.5+
+last-check: 2018-01-30T16:50:01.99308582Z
+boot-time: 2018-01-30T12:49:32Z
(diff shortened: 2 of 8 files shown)


tests/test_livepatch_motd.py
```diff
@@ -0,0 +1,238 @@
...
+    LIVEPATCH_NEEDS_CHECK,
+    LIVEPATCH_CHECK_FAILED,
+    UA_STATE_ELSEWHERE)
+from random import randrange
+
+LIVEPATCH_IS_ENABLED = 'Canonical Livepatch is enabled.'
+LIVEPATCH_STATE_MESSAGES = {
+    'checked': {
+        'applied':
+            [LIVEPATCH_IS_ENABLED,
```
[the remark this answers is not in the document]
panlinux: What do I gain?

tests/testing.py
```diff
@@ -88,12 +88,13 @@ def read_file(self, path):
         with open(self.tempdir.join(path)) as fh:
             return fh.read()
 
-    def script(self, *args):
-        """Run the script."""
-        command = ['./ubuntu-advantage']
+    def script(self, *args, script='./ubuntu-advantage', env_update={}):
```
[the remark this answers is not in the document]
panlinux: flake8 in trusty is giving me this now :/
```
tests/testing.py:92:39: E901 SyntaxError: invalid syntax
    def script(self, *args, env_update=None):
                                      ^
```

tests/test_livepatch_motd.py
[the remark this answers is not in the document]
albertodonato:
```
'applied': (
     LIVEPATCH_IS_ENABLED, ...)
```

As said, just a nitpick, there's no need to use lists since they're not modified

Makefile
```diff
@@ -19,17 +19,14 @@ lint-py:
 	@tox -e lint
 
 lint-sh: SHELLCHECK = shellcheck
-lint-sh: BASH_SCRIPTS = $(UA_SCRIPTS)
-lint-sh: DASH_SCRIPTS = $(MOTD_SCRIPTS)
+lint-sh: BASH_SCRIPTS = $(UA_SCRIPTS) $(MOTD_SCRIPTS)
 lint-sh: _lint-sh-command
 
 lint-sh/docker: SHELLCHECK = docker run --rm -it -v $(PWD):/repo koalaman/shellcheck
 lint-sh/docker: BASH_SCRIPTS = $(patsubst %,/repo/%,$(UA_SCRIPTS))
```
albertodonato: This should become
```
BASH_SCRIPTS = $(patsubst %,/repo/%,$(UA_SCRIPTS) $(MOTD_SCRIPTS))
```

tests/fakes.py
```diff
@@ -68,3 +81,47 @@
...
 """
+
+LIVEPATCH_ENABLED_STATUS = """
+cat <<EOF
+client-version: "7.23"
+status:
+- kernel: 4.4.0-87.110-generic
+  running: true
+  livepatch:
+    checkState: {}
```
albertodonato: please use named subtitutions, e.g.:

```
checkState: {check_state}
```
so that the .format() call is a bit more explicit (see below)

tests/fakes.py
```diff
@@ -68,3 +81,47 @@
...
+    patchState: {}
+"""
+
+LIVEPATCH_CHECKED_UNAPPLIED = LIVEPATCH_ENABLED_STATUS.format(
+    'checked', 'unapplied')
```
albertodonato: Per comment above, this would be something like `LIVEPATCH_ENABLED_STATUS.format(check_state='checked, patch_state='unapplied)`

Same below

update-motd.d/99-livepatch
```diff
@@ -0,0 +1,94 @@
...
+            ;;
+        *)
+            echo "Unknown check status. Please see /var/log/syslog for more information."
+            echo "     Status: \"$check_state\""
+            ;;
+    esac
+}
+
+
+if ! [[ ":$PATH:" == *":/snap/bin:"* ]]; then
```
albertodonato: you can use `!=` rather than the extra `!` before

Makefile
panlinux: Actually, I think this is the right fix, as the docker run was missing the motd scripts. Do you agree?
```diff
@@ -23,7 +23,7 @@ lint-sh: BASH_SCRIPTS = $(UA_SCRIPTS) $(MOTD_SCRIPTS)
 lint-sh: _lint-sh-command
 
 lint-sh/docker: SHELLCHECK = docker run --rm -it -v $(PWD):/repo koalaman/shellcheck
-lint-sh/docker: BASH_SCRIPTS = $(patsubst %,/repo/%,$(UA_SCRIPTS))
+lint-sh/docker: BASH_SCRIPTS = $(patsubst %,/repo/%,$(UA_SCRIPTS) $(MOTD_SCRIPTS))
 lint-sh/docker: _lint-sh-command
 
 _lint-sh-command:
```

PR #115, opened 2018-02-05
refactor function for service status

albertodonato:


modules/service.sh
```diff
@@ -52,10 +75,8 @@ _service_check_support() {
 _service_check_enabled() {
     local service="$1"
 
-    local service_upper title
-    service_upper=$(uppercase "$service")
-    title=$(expand_var "${service_upper}_SERVICE_TITLE")
-
+    local title
+    title=$(expand_var "${service}_SERVICE_TITLE" true)
```
panlinux: I suggest to keep `expand_var` defined as it was, and call it like this:
```bash
title=$(expand_var "${service^^}_SERVICE_TITLE")
```

ubuntu-advantage
```diff
@@ -38,37 +38,10 @@ load_modules() {
...
-    if fips_is_enabled; then
-        fips_status="enabled"
-    else
-        fips_status="disabled"
-        is_supported_series "$FIPS_SUPPORTED_SERIES" || fips_status="disabled (not available)"
-    fi
-    echo "fips: $fips_status"
+    local service
+    for service in $SERVICES; do
+        service_print_status "$service"
```
panlinux: SERVICES is defined as `SERVICES="esm fips livepatch"`. That means the new ordering of services in the status output is the order in which they appear in that variable, which is different to what `print_status` did up until now, which was esm, livepatch and fips. I remember Steve Langasek making a remark about this order, but I don't recall what it was. It doesn't seem to be alphabetical in the old print_status, but it is in the SERVICES variable. I suggest to add a comment above the SERVICES definition at the top of the file asking to keep the services in there ordered alphabetically.

PR #117, opened 2018-02-05
A few bash-powered cleanups

### LINKED ISSUE #104: Replace exit with return
Replace `exit` calls in functions with `return`, since the script is executed with `-e` anyway.

albertodonato:
Fixes #104

### DIFF
--- a/modules/snap.sh
+++ b/modules/snap.sh
@@ -1,9 +1,8 @@
 # shellcheck disable=SC2039
 
-# snapd needs a 4.4.x *running* kernel
 check_snapd_kernel_support() {
-    local v1 v2
-    v1=$(echo "$KERNEL_VERSION" | cut -d . -f 1)
-    v2=$(echo "$KERNEL_VERSION" | cut -d . -f 2)
-    test "$v1" -ge "4" -a "$v2" -ge "4"
+    declare -a version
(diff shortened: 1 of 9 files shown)


modules/snap.sh
```diff
@@ -1,9 +1,8 @@
...
-# snapd needs a 4.4.x *running* kernel
 check_snapd_kernel_support() {
-    local v1 v2
-    v1=$(echo "$KERNEL_VERSION" | cut -d . -f 1)
-    v2=$(echo "$KERNEL_VERSION" | cut -d . -f 2)
-    test "$v1" -ge "4" -a "$v2" -ge "4"
+    declare -a version
+    mapfile -t version < <(echo -e "${KERNEL_VERSION//[.-]/\\n}")
+    # snapd needs a 4.4.x *running* kernel
+    test "${version[0]}" -ge "4" -a "${version[1]}" -ge "4"
```
panlinux: I'm surprised you didn't try to awk'ify this :)

Just a curiosity, do you know why this won't work:
```bash
echo -e "${KERNEL_VERSION//[.-]/\\n}" | mapfile -t version
```

albertodonato: Stop passing me for an awk fan :)

WRT mapfile, it's because if you run `cmd | mapfile`, that mapfile is actually setting `version` in a different shell's env, not in the current one.
To have it set in the current shell, you have to do `mapfile <(cmd)`

PR #118, opened 2018-02-06
Cleanups to repository URLs

### LINKED ISSUE #100: Rename `_url` variables to `_host` where appropriate

albertodonato:
Include schema in repository URLs. This also makes the match a bit safer when removing repository credentials from the auth file.

Fixes #100 

Note: depends on #117

### DIFF
--- a/modules/apt.sh
+++ b/modules/apt.sh
@@ -5,7 +5,7 @@ private_repo_url() {
     local credentials="$2"
     local file="$3"
 
-    echo "https://${credentials}@${repo_url}/ubuntu${file}"
+    echo "https://${credentials}@${repo_url##https://}/ubuntu${file}"
 }
 
 package_version() {
@@ -21,7 +21,7 @@ apt_add_repo() {
     local keyring_file="$4"
 
     _apt_write_list_file "$repo_file" "$repo_url"
-    _apt_add_auth "$repo_url" "$credentials"
+    _apt_add_auth "${repo_url##https://}" "$credentials"
     cp "$keyring_file" "$APT_KEYS_DIR"
 }
 
@@ -31,7 +31,7 @@ apt_remove_repo() {
     local keyring_file="$3"
 
     rm -f "$repo_file" "$keyring_file"
-    _apt_remove_auth "$repo_url"
+    _apt_remove_auth "${repo_url##https://}"
 }
 
 apt_get() {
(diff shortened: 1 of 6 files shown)


modules/apt.sh
```diff
@@ -61,26 +61,32 @@ _apt_write_list_file() {
...
-deb https://${repo_url}/ubuntu ${SERIES} main
-# deb-src https://${repo_url}/ubuntu ${SERIES} main
+deb ${repo_url}/ubuntu ${SERIES} main
+# deb-src ${repo_url}/ubuntu ${SERIES} main
 EOF
 }
 
 _apt_add_auth() {
-    local repo_host="$1"
+    local repo_path="$1"
```
panlinux: This is actually getting the hostname plus a path, just not the protocol prefix, right? `repo_path` doesn't sound like it would contain a hostname.

Since this branch is mostly about making sure the variable name (`repo_url`) matches what it contains, I think we should either rename `repo_path` to something like, ugh, `repo_host_path`, or even better perhaps, strip the protocol prefix inside this method (instead of in the caller) and call it `repo_url` again. Then we can strip the protocol further down in that echo call.

Care must be taken that we are only stripping https, not any other protocol, though, but the caller is making that assumption already.

modules/apt.sh
```diff
@@ -5,7 +5,7 @@ private_repo_url() {
     local credentials="$2"
     local file="$3"
 
-    echo "https://${credentials}@${repo_url}/ubuntu${file}"
+    echo "https://${credentials}@${repo_url##https://}/ubuntu${file}"
```
panlinux: We should probably make it clear somewhere that we are only supporting the https protocol, we won't be stripping anything else. Maybe add a note to where these URLs are defined.

modules/apt.sh
```diff
@@ -61,26 +61,32 @@ _apt_write_list_file() {
...
-    echo "machine ${repo_host}/ubuntu/ login ${login} password ${password}" \
+    echo "machine ${repo_path}/ubuntu/ login ${login} password ${password}" \
          >>"$APT_AUTH_FILE"
 }
 
 _apt_remove_auth() {
-    local repo_host="$1"
-
-    # shellcheck disable=SC1117
-    sed -i "/^machine ${repo_host}\/ubuntu\/ login/d" "$APT_AUTH_FILE"
```
panlinux: This would still work:
```
sed -i "/^machine ${repo_host//\//\\/}\/ubuntu\/ login/d" auth.conf
```

But ugly, yes. Let me try to parse your awk.

modules/apt.sh
```diff
@@ -61,26 +61,32 @@ _apt_write_list_file() {
...
+    local repo_path="$1"
```
panlinux: Like my other comment, should we make this function take a full url as an argument, and strip the protocol as needed? repo_path should just be a path, not a hostname plus a path, in my mind.

modules/apt.sh
```diff
@@ -61,26 +61,32 @@ _apt_write_list_file() {
...
+
+    local tempfile
+    tempfile=$(mktemp)
+    chmod 600 "$tempfile"
+    # don't use pattern matching as the repo path contains slashes and dots
+    awk -v repo_path="${repo_path}/ubuntu/" \
+        '! ($1 == "machine" && $2 == repo_path)' \
+        "$APT_AUTH_FILE"  >"$tempfile"
+    mv "$tempfile" "$APT_AUTH_FILE"
```
panlinux: Have you tried this with mawk?

modules/apt.sh
```diff
@@ -5,7 +5,8 @@ private_repo_url() {
     local credentials="$2"
     local file="$3"
 
-    echo "https://${credentials}@${repo_url}/ubuntu${file}"
+    local repo_schema="${repo_url/:\/\/*/}}"
```
panlinux: I think you meant `repo_scheme`, or `url_scheme`: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

modules/apt.sh
```diff
@@ -61,26 +62,33 @@ _apt_write_list_file() {
...
-    local repo_host="$1"
+    local repo_url="$1"
     local credentials="$2"
 
     local login="${credentials%:*}"
     local password="${credentials#*:}"
     [ -f "$APT_AUTH_FILE" ] || touch "$APT_AUTH_FILE"
     chmod 600 "$APT_AUTH_FILE"
-    echo "machine ${repo_host}/ubuntu/ login ${login} password ${password}" \
+    echo "machine ${repo_url/*:\/\//}/ubuntu/ login ${login} password ${password}" \
```
panlinux: Suggestion: use another local variable here, like in the awk code further down, called `repo_host_path`, or `repo_without_scheme` :), to make it more obvious what we are adding to the auth file. The sequence of `/` and `\` is never immediately obvious about what it does, and easy to typo.

modules/apt.sh
albertodonato: yes, good catch, fixed

modules/apt.sh
albertodonato: done

PR #120, opened 2018-02-07
check that all prerequisite packages are installed when enabling FIPS

### LINKED ISSUE #102: fips pkg installed check should check all of them before making a decision
The check for installed fips packages in _fips_prep_check(), later changed to _fips_check_packages_installed() in PR #101, should iterate over all packages before proceeding. It should check that all packages are installed before moving on to fips_is_enabled, because if there is a missing package, we will want to fix it by installing it. And we definitely shouldn't tell the user to reboot if we find out just one package is installed and not the rest.

albertodonato:
Fixes #102 

Also, prefix all package-related functions with `apt_` (for namespacing)

### DIFF
--- a/modules/service-fips.sh
+++ b/modules/service-fips.sh
@@ -16,12 +16,14 @@ else
     FIPS_BOOT_CFG=${FIPS_BOOT_CFG:-"${FIPS_BOOT_CFG_DIR}/99-fips.cfg"}
 fi
 FIPS_HMAC_PACKAGES="openssh-client-hmac openssh-server-hmac libssl1.0.0-hmac \
-        linux-fips strongswan-hmac"
+    linux-fips strongswan-hmac"
+FIPS_OTHER_PACKAGES="openssh-client openssh-server openssl libssl1.0.0 \
+    fips-initramfs strongswan"
 
 fips_enable() {
     local token="$1"
 
-    _fips_check_packages_installed || error_exit service_already_enabled
+    _fips_check_installed || error_exit service_already_enabled
 
     check_token "$FIPS_REPO_URL" "$token"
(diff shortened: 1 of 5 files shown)


modules/service-fips.sh
```diff
@@ -121,16 +121,22 @@ _fips_enabled_check() {
     echo 0
 }
 
+_fips_check_installed() {
+    if ! _fips_check_packages_installed; then
+        return
```
panlinux: Shouldn't we return 1 here? What happens with such an empty return?

tests/test_fips.py
```diff
@@ -72,6 +72,14 @@ def test_enable_fips_installed_not_enabled(self):
             'Please reboot into the FIPS kernel to enable it.',
             process.stderr.strip())
 
+    def test_enable_fips_not_all_packages_installed(self):
+        # one of the packages is not installed
+        self.make_fake_binary(
+            'dpkg-query', command='[ $2 != openssh-client-hmac ]')
+        process = self.script('enable-fips', 'user:pass')
+        self.assertEqual(process.returncode, 0)
+        self.assertIn('Successfully configured FIPS', process.stdout)
```
panlinux: I suggest to also check for `Installing FIPS packages`.

modules/service-fips.sh
albertodonato: This is the same as `return 0`.
The logic here is inverted, the function returns nonzero if packages are installed, as it's used as
`_fips_check_installed || error_exit service_already_enabled`

tests/test_fips.py
albertodonato: Done

PR #123, opened 2018-02-09
Switch back to dash (Fixes #121)

### LINKED ISSUE #121: Rewrite motd script in dash
Following the review comment from https://code.launchpad.net/~ahasenack/ubuntu/+source/ubuntu-advantage-tools/+git/ubuntu-advantage-tools/+merge/337213/comments/884844, we should adapt the livepatch motd script to dash, and return the esm one to dash as well.

panlinux:


update-motd.d/99-livepatch
```diff
@@ -54,8 +54,8 @@ print_status() {
 }
 
 
-if [[ ":$PATH:" != *":/snap/bin:"* ]]; then
-    PATH=:/snap/bin:$PATH
+if ! echo "$PATH" | grep -qE "/snap/bin(:|$|/:/$)"; then
```
albertodonato: I think the regexp you want here is: `grep -E "(^|:)/snap/bin/?(:|$)"`

The initial group is needed to avoid matching `/foo/snap/bin`, and at the end you want either a colon or the end of the string, after the optional final `/`.

PR #124, opened 2018-02-14
Support returning the status for a single service

albertodonato:


ubuntu-advantage
```diff
@@ -58,13 +65,14 @@ Currently available are:
 
 Commands:
  version                   show the tool version
- status                    show current status of Ubuntu Advantage offerings
- enable-esm <token>        enable the ESM repository
+ status <OFFERING>         show current status of Ubuntu Advantage offerings
```
panlinux: This should be `[OFFERING]`, since it's not mandatory. I'm not very keen on the name `OFFERINGS`, even though it was used in `status` before. Perhaps `name`?

tests/test_script.py
```diff
@@ -79,6 +79,17 @@ def test_status_i686_livepatch_not_avaiable(self):
...
         self.assertIn("livepatch: disabled (not available)", process.stdout)
 
+    def test_status_with_one_service(self):
+        """The status for a single service can be returned."""
+        process = self.script('status', 'fips')
+        self.assertEqual(process.returncode, 0)
+        self.assertEqual(process.stdout, 'fips: disabled (not available)\n')
+
+    def test_status_with_one_service_unknown(self):
+        """The script exits with error a single service can be returned."""
```
panlinux: This docstring needs work :)

albertodonato: fixed

ubuntu-advantage
albertodonato: yeah, name sounds better, fixed

PR #125, opened 2018-02-14
Implement a cache for the livepatch motd (fixes #122)

### LINKED ISSUE #122: Cache status output
Following https://code.launchpad.net/~ahasenack/ubuntu/+source/ubuntu-advantage-tools/+git/ubuntu-advantage-tools/+merge/337213/comments/884844, we want to remove "bash" and anything else that could cause delays from the login path. The best solution is to probably cache the output of the `ubuntu-advantage status` output, and maybe even update it out-of-band, so that the script run at login time at most prints existing text instead of running potentially costly code.

Brain storm:

a) Have the motd script itself manage the `ua status` cache.

- if the cache is too old, call the `ua status` command, use its output, and update the cache at the same time
- otherwise, use the cache value
Script might still "hickup" when updating the cache

b) have the motd script always use the cached value of `ua status`

- another process, like a cron job, updates that cache
- the motd script could check if the cache is too old and warn about it
- alternatively, or in addition to the above, the cache timestamp could be shown in the motd output

c) something entirely different: cache the whole output of the motd script

- the motd script would check for a cached output and print that, instead of reading a cached `ua status` output and parsing it
- another process, like the cron job, would run what is today the motd script and cache its output, i.e., those lines about the livepatch status, instead of caching just the output of `ua status`

panlinux:
Ideas for follow-up branches:

 - update the cache every time the user calls "ua status"
 - update the cache right after an enable or disable call

PPA with (manual) test builds: https://launchpad.net/~ahasenack/+archive/ubuntu/ua-livepatch-motd-cache/+packages

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+set -e
+
+STAMP_FILE=/var/cache/ubuntu-advantage-tools/ubuntu-advantage-status.stamp
+
+if [ "$1" = purge -a -f $STAMP_FILE ]; then
+    rm $STAMP_FILE
+fi
+
+#DEBHELPER#
+
--- a/debian/postrm
+++ b/debian/postrm
@@ -1,6 +1,4 @@
-#!/bin/sh
-
-set -e
+#!/bin/sh -e
 
 STAMP_FILE=/var/cache/ubuntu-advantage-tools/ubuntu-advantage-status.stamp
 
--- a/debian/postrm
+++ b/debian/postrm
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 
-CACHE_FILE=/var/cache/ubuntu-advantage-tools/ubuntu-advantage-status.stamp
+CACHE_FILE=/var/cache/ubuntu-advantage-tools/ubuntu-advantage-status.cache
 
 if [ "$1" = purge -a -f "$CACHE_FILE" ]; then
     rm "$CACHE_FILE"
--- a/debian/postrm
+++ b/debian/postrm
@@ -1,9 +1,9 @@
 #!/bin/sh -e
 
-STAMP_FILE=/var/cache/ubuntu-advantage-tools/ubuntu-advantage-status.stamp
+CACHE_FILE=/var/cache/ubuntu-advantage-tools/ubuntu-advantage-status.stamp
 
-if [ "$1" = purge -a -f "$STAMP_FILE" ]; then
-    rm "$STAMP_FILE"
+if [ "$1" = purge -a -f "$CACHE_FILE" ]; then
+    rm "$CACHE_FILE"
 fi
 
 #DEBHELPER#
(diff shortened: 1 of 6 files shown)


dpb1:
Hi Andreas -- Thanks for this patch, nice to have livepatch guarded since it will be in the default login path.  I tested install/remove upgrade on Xenial, Artful and Bionic, and everything is working as expected.

debian/postrm
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
```
albertodonato: you can set `/bin/sh -e` in the shebang

debian/postrm
```diff
@@ -0,0 +1,12 @@
...
+
+STAMP_FILE=/var/cache/ubuntu-advantage-tools/ubuntu-advantage-status.stamp
+
+if [ "$1" = purge -a -f $STAMP_FILE ]; then
+    rm $STAMP_FILE
```
albertodonato: please quote `"$STAMP_FILE"` (both uses)

debian/ubuntu-advantage-tools.cron.hourly
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
```
albertodonato: same here

debian/postrm
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
+
+STAMP_FILE=/var/cache/ubuntu-advantage-tools/ubuntu-advantage-status.stamp
```
albertodonato: I'd s/STAMP/CACHE/ everywhere (including filename suffix), since it's not just an empty stamp file, it actually contains data

update-motd.d/99-livepatch
```diff
@@ -59,7 +60,10 @@ if ! echo "$PATH" | grep -qE "(^|:)/snap/bin/?(:|$)"; then
 fi
 
 service_name="livepatch"
-ua_status=$($UA status 2>&1)
+# if there is no cache file yet (the cron job hasn't run yet), create a
+# fresh one. We also want to ignore an empty cache file.
+[ -s "$UA_STATUS_CACHE" ] || "$UA" status > "$UA_STATUS_CACHE" 2>&1
```
albertodonato: why are you redirecting stderr to stdin? this means any error message would end up in the cache

panlinux: Because it's what the former `$($UA status 2>&1)` did. I guess I should also do that in the cron job (or not do it, if we decide to change that here as well).

debian/postrm
panlinux: Funy, I copied this syntax/idea from other similar scripts, I guess "stamp" got abused indeed

update-motd.d/99-livepatch
albertodonato: I don't think you should do it, because if for some reason some stderr text goes in, it could be parsed in place of what's expected

Issue #130, opened 2018-02-16
Do not ever call ubuntu-advantage in the motd, ever.


panlinux:
If there is no cache file, the motd script will call `ua status` to create one and use it. It must not do that, ever. If there is no cache, bail.

PR #132, opened 2018-02-16
Don't update cache inside motd script (Fixes: #130)

panlinux:
I can squash the commits before merging.


tests/fakes.py
```diff
@@ -123,14 +123,41 @@
 LIVEPATCH_CHECK_FAILED = LIVEPATCH_ENABLED_STATUS.format(
     check_state='check-failed', patch_state='does not matter')
 
-UA_STATE_ELSEWHERE = """
-cat <<EOF
-fips: disabled
-esm: enabled
+STATUS_CACHE_LIVEPATCH_ENABLED = """esm: disabled (not available)
```
albertodonato: please use `"""\` on the first line so you can keep the text aligned (same below)

PR #133, opened 2018-03-01
Add "update-fips" feature support to ubuntu-advantage tool

vineethapai:
The patchset adds support for updating FIPS packages installed on a system. The updated packages are hosted in a separate "fips-update" ppa under ubuntu-advantage. The updates are not FIPS certified. They include CVEs and bug fixes. Adding "update-fips" support allows an automated way to deliver fixes and updates to interested FIPS customers.


modules/service-fips.sh
```diff
@@ -8,6 +8,10 @@ FIPS_REPO_URL="https://private-ppa.launchpad.net/ubuntu-advantage/fips"
 FIPS_REPO_KEY_FILE="ubuntu-fips-keyring.gpg"
 FIPS_REPO_LIST=${FIPS_REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-fips-${SERIES}.list"}
 FIPS_REPO_PREFERENCES=${FIPS_REPO_PREFERENCES:-"/etc/apt/preferences.d/ubuntu-fips-${SERIES}"}
+FIPS_UPDATES_REPO_URL="https://private-ppa.launchpad.net/ubuntu-advantage/fips-updates"
+FIPS_UPDATES_REPO_KEY_FILE="ubuntu-fips-updates-keyring.gpg"
+FIPS_UPDATES_REPO_LIST=${FIPS_UPDATES_REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-fips-updates-${SERIES}.list"}
+FIPS_UPDATES_REPO_PREFERENCES=${FIPS_UPDATE_REPO_PREFERENCES:-"/etc/apt/preferences.d/ubuntu-fips-updates-${SERIES}"}
```
albertodonato: Typo here: `UPDATE` vs `UPDATES`

modules/service-fips.sh
```diff
@@ -50,6 +54,62 @@ fips_disable() {
...
+    local bypass_prompt="$2"
+    local fips_configured=0
+
+    check_token "$FIPS_UPDATES_REPO_URL" "$token"
+
+    echo -n "Updating FIPS packages will take the system out of FIPS compliance."
+    if [ "$bypass_prompt" -ne 1 ]; then
+        if ! prompt_user 'Do you want to proceed?'; then
+            error_msg "Aborting updating FIPS packages..."
+            error_exit no_updates_selected
```
albertodonato: I think this should just `return`, since the user aborting is not really an error.

modules/service-fips.sh
```diff
@@ -50,6 +54,62 @@ fips_disable() {
...
+        fi
+    fi
+
+    #add the fips-updates repo if the system is undergoing updates the first time
```
albertodonato: please add a blank space after `#` (same for other cases below)

modules/service.sh
```diff
@@ -5,7 +5,7 @@ service_from_command() {
 
      echo "$command" | sed -n -r \
          's,^is-(.+)-enabled$,\1,p;
-          s,^(enable|disable)-(.+)$,\2,p'
+          s,^(enable|disable|update)-(.+)$,\2,p'
```
albertodonato: We shouldn't add a global `update-<service>` command, otherwise all services would get it (and we'd have to add `<service>_update` functions for each of the others with an unsupported message.
I'd suggest adding a `update-fips` entry to the `case` in main, and having needed code there

modules/service.sh
```diff
@@ -29,6 +29,17 @@ service_disable() {
...
 
+service_update() {
+    local service="$1"
+    local token="$2"
+    local bypass_prompt="$3"
+
+    _service_check_user
+    _service_check_support "$service"
+    "${service}_validate_token" "$token" || error_exit invalid_token
+    "${service}_update" "$token" "$bypass_prompt"
```
albertodonato: as said above, this wouldn't exist for `esm` or `livepatch`, causing an error

modules/utils.sh
```diff
@@ -16,10 +16,30 @@ error_exit() {
...
     )
     exit "${codes[$code]}"
 }
 
+# prompt the user to confirm
+#
+prompt_user() {
+    local answer
+
+    echo >&2
```
albertodonato: why echoing to stderr? (same below)

ubuntu-advantage
```diff
@@ -75,7 +75,9 @@ Commands:
  enable-livepatch <TOKEN>  enable the Livepatch service
  disable-livepatch [-r]    disable the Livepatch service. With "-r", the
                            canonical-livepatch snap will also be removed
-
+ update-fips <token> [-y]  Install and enable updated non-certified FIPS
+                           modules. With "-y" or "--yes" the user prompt
+                           will be bypassed.
```
albertodonato: Please keep this close to other fips command entries (and in the same order as the manpage)

tests/testing.py
```diff
@@ -76,6 +80,8 @@ def setUp(self):
         self.trusted_gpg_dir.mkdir()
         (self.keyrings_dir / 'ubuntu-esm-keyring.gpg').write_text('GPG key')
         (self.keyrings_dir / 'ubuntu-fips-keyring.gpg').write_text('GPG key')
+        (self.keyrings_dir / 'ubuntu-fips-updates-keyring.gpg').write_text(
+                                                                     'GPG key')
```
albertodonato: indentation is weird here

modules/service-fips.sh
panlinux: This fixes the test errors, btw :)

vineethapai: Thanks and fixed.

modules/service-fips.sh
vineethapai: Agree. Fixed.

modules/service-fips.sh
vineethapai: Done.

modules/service.sh
vineethapai: As indicated above, I removed it from that file.

modules/utils.sh
vineethapai: Removed this.

tests/testing.py
vineethapai: Fixed it.

ubuntu-advantage
vineethapai: Fixed.

ubuntu-advantage
```diff
@@ -109,6 +111,26 @@ main() {
             service_disable "$service" "$@"
             ;;
 
+        update-fips)
+            local token="$1"
+            local bypass_prompt=0
+            while true ; do
+               case "$2" in
+                  --yes | -y)
```
panlinux: I think we can drop `--yes`, supporting `-y` is enough. No other command has `--option`.

ubuntu-advantage.1
```diff
@@ -55,6 +55,25 @@ cryptoapi.
 disable-fips
 It's currently not possible to disable FIPS after it has been enabled.
 
+.TP
+.B
+update-fips \fItoken\fR[\fB\-y|--yes\fR]
```
panlinux: There should be an exta space after the first `\fR` and before `[`

ubuntu-advantage
```diff
@@ -109,6 +111,26 @@ main() {
...
+                     bypass_prompt=1
+                     break;
+                     ;;
+                  *)
+                     break;
+                     ;;
+               esac
+            done;
```
panlinux: It doesn't look like this while/true/done loop plus a case/esac is necessary. Also, you are accepting `--anything`. You should probably reject anything other than `-y` (and `--yes` if you think it should be kept).

modules/service-fips.sh
```diff
@@ -50,6 +54,62 @@ fips_disable() {
...
 }
 
+fips_update() {
+    local token="$1"
+    local bypass_prompt="$2"
+    local fips_configured=0
+
+    check_token "$FIPS_UPDATES_REPO_URL" "$token"
+
+    echo -n "Updating FIPS packages will take the system out of FIPS compliance."
```
panlinux: Please drop the `-n` from this echo, or else the next statement will be added right next to it:
```
Checking token... OK
Updating FIPS packages will take the system out of FIPS compliance.Do you want to proceed? [N/y] 
```

modules/service-fips.sh
```diff
@@ -50,6 +54,62 @@ fips_disable() {
...
+        apt_add_repo "$FIPS_UPDATES_REPO_LIST" "$FIPS_UPDATES_REPO_URL" "$token" \
+                 "${KEYRINGS_DIR}/${FIPS_UPDATES_REPO_KEY_FILE}"
+        apt_add_repo_pinning "$FIPS_UPDATES_REPO_PREFERENCES" \
+                         LP-PPA-ubuntu-advantage-fips-updates 1001
+        apt_install_package_if_missing_file "$APT_METHOD_HTTPS" apt-transport-https
+        apt_install_package_if_missing_file "$CA_CERTIFICATES" ca-certificates
+        echo -n 'Running apt-get update... '
+        check_result apt_get update
+        echo 'Ubuntu FIPS-UPDATES PPA repository enabled.'
+    fi
```
panlinux: Do we want to keep going if the updates repo is already there? What is the user story here where I would run `ua update-fips <token>` multiple times?

modules/service-fips.sh
vineethapai: ok. There will be an extra blank line when I remove -n. Is that OK?
Checking token... OK
Updating FIPS packages will take the system out of FIPS compliance.

Do you want to proceed? [N/y]

panlinux: Do you know what adds that extra line? `prompt_user()`? In any case, go ahead and remove `-n` please, it's used for when we want another subsequent `echo` to finish up a sentence, like in the case of reporting `OK` or `ERROR`.

ubuntu-advantage
albertodonato: +1

ubuntu-advantage
```diff
@@ -109,6 +111,26 @@ main() {
...
+            service_check_user
+            service_check_support "fips"
```
albertodonato: shouldn't we also check that fips is enabled? otherwise you could run `ua update-fips <token>`  when you haven't first enabled it

modules/service-fips.sh
vineethapai: I think we would need to do updates via the ua tool itself as 
a) kernel update will need update-grub to be run
b) Not all customers may want updates.

This brings me to another question. I am open to suggestions here. Currently the implementation does not check if the packages installed are latest. Invoking the update switch multiple times just tries to update all packages and asks to reboot. I am thinking we may need to add some apt checking (apt-cache policy <pkg> or apt list -a --upgradable) to make sure if all installed packages are latest and just return if so. Ask to reboot only if kernel has been updated.

ubuntu-advantage
vineethapai: fips_update function takes care of it.  if no initramfs fips package is found on the system "fips_configured" will be 0 and it assumes fips was never enabled on the system and does everything that fips_enable does after installing the updated packages.

ubuntu-advantage
vineethapai: Fixed

ubuntu-advantage.1
vineethapai: Fixed.

ubuntu-advantage
vineethapai: Fixed.

modules/service-fips.sh
vineethapai: Removed -n. Fixed.

modules/service-fips.sh
panlinux: `ua update-fips <token>` shouldn't be needed anymore after it was enabled. Package updates, kernel included, should be handled by the normal ubuntu upgrade process for deb packages. Unless the fips kernel package does or requires something different.

Which updates to install, and how, is always a customer's choice. Or are you talking about `unattended-upgrades`? AFAIK that is enabled by default for the security pocket.

panlinux:
Please also update `debian/changelog`, like this:
```
* New upstream release:
  - added update-fips command
```
The `* New upstream release:` text is already there, so just add the second line, or some text along those lines (feel free to expand it a bit). And if you have a public launchpad bug about it, mention it as well using the syntax `(LP: #xxxxxx)`. No need to change the rest of the changelog file, that we can do when it's time for a new release. This is just so we don't forget to mention this new feature in the changelog.

modules/service-fips.sh
vineethapai: Agree. After offline discussion via email on this, I understand now that ua update-fips <token> is meant to be used only once to enable fips-updates repo and install updates. The future updates are going to be installed via apt upgrade.

vineethapai: I did see ESM added to unattended-upgrades ( /etc/apt/apt.conf.d/50unattended-upgrades) and it installs where applicable in addition to the security pocket. 
For FIPS, I prefer leaving it to customer which updates to install as certification is involved.

albertodonato:
Given that the purpose of the command is to enabled updates (not really for updating packages once the updates repo is enabled) I wonder if the command shouldn't be called `enable-fips-updates` or something similar.

vineethapai:
First invocation of "update-fips", it enables fips-updates repo and also installs updates if available. I do like the suggestion of using "enable-fips-updates". Is there a two word command convention (enable-*, disable-*etc) followed in the tool? If so, will that go against it?

panlinux:
I'm +1 for `enable-fips-updates`, as just `update-fips` already generated some confusion where we thought that's how the system would be kept up-to-date, as opposed to a regular `apt update`. Too bad it's a 3-word command. If someone has a good suggestion, fine, but also let's not bike shed too much :)

vineethapai:
I made the change to use "enable-fips-updates" to enable the fips-updates repo. Thanks.

debian/changelog
```diff
@@ -1,6 +1,8 @@
 ubuntu-advantage-tools (15) bionic; urgency=medium
 
   * New upstream release
+    - added update-fips command. This command installs FIPS modules updated
```
panlinux: s/update-fips/enable-fips-updates/

vineethapai: Done. Updated debian/changelog.

panlinux:
There is a missing space after the `:`:
```
$ ua status
esm: disabled (not available)
fips: enabled
  fips-updates (uncertified):disabled
livepatch: disabled
```

modules/service-fips.sh
```diff
@@ -50,6 +54,63 @@ fips_disable() {
...
 
+fips_updates_enable() {
+    local token="$1"
+    local bypass_prompt="$2"
+    local fips_configured=0
+    local fips_kernel_version=0
+
+    check_token "$FIPS_UPDATES_REPO_URL" "$token"
+
+    echo "Updating FIPS packages will take the system out of FIPS compliance."
```
panlinux: Wording suggestion. Instead of `Updating FIPS packages will take the system out of FIPS compliance`, use something like `Enabling the fips-updates repository will take the system out of FIPS compliance`. The problem isn't updating the fips packages, but updating from fips-updates, right? Essentially we have two channels:
 * fips: for the **certified** packages
 * fips-updates: for these **uncertified** updates

There is nothing preventing `fips` from eventually receiving a new set of certified packages, in which case updating the packages should be just fine.

That being said, I don't want to bikeshed, whichever way you prefer is fine :)

vineethapai: Only updates from fips-updates ppa is uncertified. Enabling fips-updates repository by itself will not make the system non-compliant, but installing updates from that repo will. So may be I can change it to "Installing updates from fips-updates repository will take the system out of FIPS compliance". Do you agree?

panlinux: yep

vineethapai: fixed both formatting with status and the user prompt message.

modules/service-fips.sh
```diff
@@ -50,6 +54,63 @@ fips_disable() {
...
+    # shellcheck disable=SC2086
+    check_result apt_get install $FIPS_HMAC_PACKAGES $FIPS_OTHER_PACKAGES
+
+    # if fips was never configured before and is enabled for the
+    # first time, configure fips
+    if [ "$fips_configured" -eq 0 ]; then
+        echo "Configuring FIPS... "
+        _fips_configure
+    fi
+    echo -n "Successfully updated FIPS packages."
```
panlinux: I think this `-n` here is incorrect. Look at the last line, how my shell prompt is showing up:
```
ubuntu@xenial-ua:~$ sudo ua enable-fips-updates  <redacted> 
Checking token... OK
Updating FIPS packages will take the system out of FIPS compliance.
Do you want to proceed? [N/y] y
Updating FIPS packages (this may take a while)... OK
Successfully updated FIPS packages.ubuntu@xenial-ua:~$ 
```

modules/service-fips.sh
```diff
@@ -50,6 +54,63 @@ fips_disable() {
...
+
+    # show the message to reboot only if fips kernel has undergone an upgrade or fips is
+    # being configured the first time
+    if [ "$fips_kernel_version" != "$(package_version linux-fips)" ] || [ "$fips_configured" -eq 0 ]; then
+        echo " Please reboot into the new FIPS kernel."
```
panlinux: If you drop that previous `-n`, then you can remove the leading space here I believe.

vineethapai: agree.

modules/service-fips.sh
panlinux: Ok, so what's the end result here? `enable-fips-updates` should check if `fips-updates` is already enabled? Right now it's not doing that, maybe this change was forgotten, or I misunderstood it.

Right now `fips_updates_enable()` does, in this order:
 * checks the token (network call)
 * optionally prompts the user for confirmation (bypassed with `-y`)
 * checks if the updates repo file exists and, if not, adds it and installs needed dependencies
 * installs (new) fips packages. Will be noop if the updates were already installed
 * configures fips (bootloader, etc) if this is the first time fips is installed, i.e., the user isn't coming from `enable-fips` but is going straight to `enable-fips-updates`

I was expecting that, if `fips-updates` is already enabled, all the steps above would be skipped.

vineethapai: Earlier this function was understood to be used every time an update was to be installed. It should skip all the steps if fips-updates is already configured manually or via tool. I will add in that check before user prompt.

panlinux:
That formatting change broke a simple test:
```
'Successfully updated FIPS packages. Please reboot into the new FIPS kernel'
```
vs
```
'Successfully updated FIPS packages.
Please reboot into the new FIPS kernel'
```
This fixes it I think:
```diff
--- a/tests/test_fips.py
+++ b/tests/test_fips.py
@@ -294,7 +294,7 @@ class FIPSTest(UbuntuAdvantageTest):
             'Configuring FIPS...',
             process.stdout)
         self.assertIn(
-            'Successfully updated FIPS packages. '
+            'Successfully updated FIPS packages.\n'
             'Please reboot into the new FIPS kernel',
             process.stdout)
```

modules/service-fips.sh
```diff
@@ -50,6 +54,69 @@ fips_disable() {
...
 }
 
+fips_updates_enable() {
+    local token="$1"
+    local bypass_prompt="$2"
+    local fips_configured=0
+    local fips_kernel_version=0
+    local result=0
+
+    check_token "$FIPS_UPDATES_REPO_URL" "$token"
```
panlinux: I think the token check should come after the `_fips_updates_is_enabled` check. It's a network check, it takes a while, and has no benefit in the case of fips updates being enabled already:
```
ubuntu@xenial-ua:~$ sudo ua enable-fips-updates <redacted>
Checking token... OK
FIPS-UPDATES repository is already enabled.
```

modules/service-fips.sh
```diff
@@ -50,6 +54,69 @@ fips_disable() {
...
+    _fips_updates_is_enabled || result=$?
+    if [ $result -eq 0 ]; then
+        error_msg "FIPS-UPDATES repository is already enabled."
+        return;
+    fi
```
panlinux: Instead of just returning, shouldn't we perhaps `error_exit service_already_enabled`? Like we do for the non-updates fips case a few lines up in the same module:
```
fips_enable() {
    local token="$1"

    _fips_check_installed || error_exit service_already_enabled

    check_token "$FIPS_REPO_URL" "$token"
...
````

File: modules/service-fips.sh (hunk 19 above)
Comment (panlinux) [reply modules/service-fips.sh]: The `fips_enable()` function in the same file also does the token check after the is-already-enabled one.

File: modules/service-fips.sh [hunk 21]
```diff
@@ -83,6 +150,21 @@ fips_check_support() {
...
 
+fips_print_status() {
+    local result=0
+
+    _fips_updates_is_enabled || result=$?
+    if [ $result -eq 0 ]; then
+        echo "fips-updates (uncertified): enabled"
+    else
+        echo "fips-updates (uncertified): disabled"
+    fi
```
panlinux: Hm, with this change, the output isn't valid yaml anymore. For example:
```
ubuntu@xenial-ua:~$ ua status
esm: disabled (not available)
fips: enabled
  fips-updates (uncertified): enabled
livepatch: enabled
  client-version: "7.23"
  architecture: x86_64
  cpu-model: Intel Core Processor (Skylake)
  last-check: 2018-03-20T19:23:21.463449773Z
  boot-time: 2018-03-20T19:18:34Z
  uptime: 4m51s
  status:
  - kernel: 4.4.0-1003.3-fips
    running: true
    livepatch:
      checkState: checked
      patchState: nothing-to-apply
      version: ""
      fixes: ""
```
If I feed that to python3-yaml, I get:
```
yaml.scanner.ScannerError: mapping values are not allowed here
  in "<unicode string>", line 3, column 29:
      fips-updates (uncertified): enabled
                                ^
```

Any suggestions?

panlinux: Hmm, scratch that, it's already a problem with the others.

modules/service-fips.sh
vineethapai: fips-update is not a separate service. FIPS_SERVICE_TITLE is defined only for FIPS service. On a fips enabled machine trying to enable-fips again shows "Canonical FIPS 140-2 Modules is already enabled". Since FIPS-UPDATES is not a service and title is not defined, it does not show any error message on exit. So I added a separate error_msg and return.

modules/service-fips.sh
vineethapai: OK

modules/service-fips.sh
vineethapai: fixed

panlinux:
Taking another look, thanks for the changes and comments.

modules/utils.sh
```diff
@@ -20,6 +20,26 @@ error_exit() {
...
+    while true; do
+        read -r -p "$* [N/y] " answer 2>&1
+        case "$answer" in
+        y|Y)
+            return 0
+            ;;
+        ''|n|N)
+            return 1
+            ;;
+        * ) echo "Please answer y or n."
```
albertodonato: extra space before parenthesis

modules/service-fips.sh
```diff
@@ -50,6 +54,70 @@ fips_disable() {
     not_supported 'Disabling FIPS'
 }
 
+fips_updates_enable() {
+    local token="$1"
+    local bypass_prompt="$2"
+    local fips_configured=0
```
albertodonato: please add an empty line between parameters and other local vars

modules/service-fips.sh
albertodonato: I think you should be able do do what @panlinux suggests, IOW `_fips_updates_is_enabled || error_exit service_already_enabled`. Even if `fips-updates` is not a separate service, the error still makes sense to me in this context.
If you want to show the error you can:

```
if _fips_updates_is_enabled; then
    error_msg "FIPS-UPDATES repository is already enabled."
    error_exit service_already_enabled
fi
```

vineethapai:
I made all the changes ack requested.

modules/service-fips.sh
```diff
@@ -50,6 +54,71 @@ fips_disable() {
...
+        return;
+    fi
+
+    check_token "$FIPS_UPDATES_REPO_URL" "$token"
+
+    echo "Installing updates from FIPS-UPDATES repository will take the system out of FIPS compliance."
+    if [ "$bypass_prompt" -ne 1 ]; then
+        if ! prompt_user 'Do you want to proceed?'; then
+            error_msg "Aborting updating FIPS packages..."
+            error_exit service_already_enabled
```
panlinux: I think you meant for this error_exit to be elsewhere,  not when bypassing the prompt

vineethapai: oops. fixing it

vineethapai: Done

panlinux:
Thanks. Taking the whole thing for a spin now.

albertodonato:
+1

PR #144, opened 2018-06-15
Add support to common criteria EAL2 artifacts installation

vineethapai:
The patchset adds a new command "install-cc" to allow installation of Ubuntu Common Criteria EAL2 artifacts on a intermediary machine at the customer site. The artifacts are packaged in a .DEB file "ubuntu-commoncriteria" and will be hosted on a ppa under ubuntu-advantage (https://launchpad.net/~ubuntu-advantage/+archive/ubuntu/commoncriteria/). Upon installation of ubuntu-commoncriteria package, the artifacts are installed in /usr/lib/common-criteria directory.  It’s up to the administrator at the customer site, following our documentation or local policies, to decide how to share these files with the computers that will actually install and setup CC.

We are currently pending Common Criteria certification from the certification body (CSEC). Since there is at least one active customer request waiting on the certification, we are trying to get this supported via ubuntu-advantage-tools in the mean time. The final version of the ubuntu-commoncriteria package will be uploaded to the ppa under ubuntu-advantage after we get certification. We hope to be certified in a couple of weeks. Currently in this patchset, I have used an internal ppa (https://launchpad.net/~fips-cc-stig/+archive/ubuntu/fipsdevppa) which has a temporary version of the .DEB package to implement and test this feature. After we receive the certification, I will update the ppa and the ppa keyring in a separate patch.


ubuntu-advantage
```diff
@@ -80,6 +80,8 @@ Commands:
  enable-livepatch <TOKEN>          enable the Livepatch service
  disable-livepatch [-r]            disable the Livepatch service. With "-r", the
                                    canonical-livepatch snap will also be removed
+ install-cc <TOKEN>                enable the commoncriteria PPA repository and
+                                   install the ubuntu-commoncriteria .DEB package
```
panlinux: Please add a common-criteria one-line description under "Currently available are" in the usage() function, after livepatch.

modules/service-cc.sh
```diff
@@ -0,0 +1,66 @@
...
+    check_token "$CC_REPO_URL" "$token"
+    apt_add_repo "$CC_REPO_LIST" "$CC_REPO_URL" "$token" \
+                 "${KEYRINGS_DIR}/${CC_REPO_KEY_FILE}"
+    apt_install_package_if_missing_file "$APT_METHOD_HTTPS" apt-transport-https
+    apt_install_package_if_missing_file "$CA_CERTIFICATES" ca-certificates
+    echo -n 'Running apt-get update... '
+    check_result apt_get update
+    echo 'Ubuntu Common Criteria PPA repository enabled.'
+
+    echo -n 'Installing Common Criteria artifacts (this may take a while)...'
```
panlinux: Missing a whitespace after ..., so that when check_result below prints either OK or ERROR it looks like `... OK` and not `...OK`

modules/service-cc.sh
```diff
@@ -0,0 +1,66 @@
...
+    # shellcheck disable=SC2086
+    check_result apt_get install $CC_UBUNTU_COMMONCRITERIA
+
+    echo "Successfully installed Common Criteria artifacts. Please check in /usr/lib/common-criteria."
```
panlinux: Can the next steps be a bit clearer? Maybe follow instructions from a specific file in `/usr/lib/common-criteria`, or a URL.
I'm still thinking about the whole enable versus install difference. so I may come back to this point later on. Maybe we should use "prepare":
`Successfully prepared this machine to host the Common Criteria artifacts. Please ....`.

modules/service-cc.sh
```diff
@@ -0,0 +1,66 @@
+# shellcheck disable=SC2034,SC2039
+ 
+CC_SERVICE_TITLE="Canonical Common Criteria"
+CC_SUPPORTED_SERIES="xenial"
+CC_SUPPORTED_ARCHS="x86_64 ppc64le s390x"
+
+CC_REPO_URL="https://private-ppa.launchpad.net/fips-cc-stig/fipsdevppa"
```
panlinux: This needs to be the final PPA, even if it's not ready yet. For testing we can switch locally.

panlinux: That being said, can I get access to your testing ppa? :)

vineethapai: I just added you to the test ppa. Can you please check access

panlinux: Got it, thanks

tests/test_cc.py
```diff
@@ -0,0 +1,142 @@
...
+        self.assertIn('Checking token... SKIPPED', process.stdout)
+
+    def test_install_cc_only_supported_on_xenial(self):
+        """The install-cc option fails if not on Xenial."""
+        self.SERIES = 'zesty'
+        process = self.script('install-cc', 'user:pass')
+        self.assertEqual(4, process.returncode)
+        self.assertIn(
+            'Canonical Common Criteria is not supported on zesty',
+            process.stderr)
```
panlinux: Please also add a test for the (un)supported architectures. Looks like arm is unsupported, so I think it's ok to just add that negative test.

ubuntu-advantage
panlinux: Ok, I thought a bit more about this, and what do you think about calling this new command `enable-cc-provisioning`. I think it represents better what is happening, and will allow us to use all the service-related functions we already have in the code.

modules/service-cc.sh
vineethapai: I updated the ppa to UA/common-criteria and also updated that ppa key in keyrings.

tests/test_cc.py
vineethapai: Added 2 tests for arm64 and i686 for unsupported architectures.

ubuntu-advantage
vineethapai: I'd have to agree. That makes more sense. Should the service be called cc-provisioning then?

panlinux: yes

modules/service-cc.sh
```diff
@@ -0,0 +1,67 @@
...
+    echo -n 'Running apt-get update... '
+    check_result apt_get update
+    echo 'Ubuntu Common Criteria PPA repository enabled.'
+
+    echo -n 'Installing Common Criteria artifacts (this may take a while)... '
+    # shellcheck disable=SC2086
+    check_result apt_get install $CC_UBUNTU_COMMONCRITERIA
+
+    echo -n "Successfully prepared this machine to host the Common Criteria artifacts. "
+    echo "Please follow instructions in Ubuntu-1604-Configuration-Guide in /usr/lib/common-criteria to configure EAL2 on the target machine(s)."
```
panlinux: Just give the full path to the pdf: `/usr/lib/common-criteria/Ubuntu-1604-Configuration-Guide.pdf`

vineethapai:
I have updated the flag to enable-cc_provisioning. The service command is "cc_provisioning" to keep with the variable naming conventions in the service-cc.sh file. The feature works. I see a few assertion failures on the tests. I will check them tomorrow.

modules/service-cc.sh
vineethapai: I added a README in the .DEB file and gave a path to that.

ubuntu-advantage
vineethapai: Done. Changed the flag to "cc_provisioning" to keep with the bash variable naming convention. The feature is tested and working. I see some assertion failures on the updated python tests with new command (also committed in this patch). I will check them again today.

panlinux:
Does it really have to be cc_provisioning, with an underscore? fips-updates is an example of a "long" name and uses "-", but granted it's not treated as a service in the code.

panlinux:
Ah, it's because of the function names.

vineethapai:
yes because of function names being invoked via $SERVICE. I could change it ccprovisioning if that looks better with no underscore.

panlinux:
We could add a quick translation before invoking the functions. Replace "-" with "_" in the service variable

panlinux:
```
andreas@nsnx:~$ service=cc-provisioning
andreas@nsnx:~$ echo ${service//-/_}
cc_provisioning
```

vineethapai:
I changed the flag to cc-provisioning. I updated the service function invocation using the translation suggested. Thanks for the suggestion.  I also fixed the test assertion failures with the latest commit.

panlinux:
I thought we could get away with doing the substitution just once, at the beginning, right after it was validated as a valid service the script can handle. Did you try that, or did you go directly with the option of doing the substitution everywhere the service variable was used?

vineethapai:
In some places it needs the service as "cc-provisioning"  for example in function service_print_status. I could change it back when it is printing $service: $status with a - instead of underscore.

panlinux:
Like this, based on your 5690cd005011c654a56856d55e5773e102c885d2 commit:
```diff
diff --git a/tests/testing.py b/tests/testing.py
index 9c11461..36833bd 100644
--- a/tests/testing.py
+++ b/tests/testing.py
@@ -119,7 +119,7 @@ class UbuntuAdvantageTest(TestWithFixtures):
             'ESM_REPO_LIST': str(self.esm_repo_list),
             'FIPS_REPO_LIST': str(self.fips_repo_list),
             'FIPS_UPDATES_REPO_LIST': str(self.fips_updates_repo_list),
-            'CC_REPO_LIST': str(self.cc_repo_list),
+            'CC_PROVISIONING_REPO_LIST': str(self.cc_repo_list),
             'FIPS_BOOT_CFG': str(self.boot_cfg),
             'FIPS_BOOT_CFG_DIR': str(self.etc_dir),
             'FIPS_ENABLED_FILE': str(self.fips_enabled_file),
diff --git a/ubuntu-advantage b/ubuntu-advantage
index 17b127b..5b3861d 100755
--- a/ubuntu-advantage
+++ b/ubuntu-advantage
@@ -81,7 +81,7 @@ Commands:
  enable-livepatch <TOKEN>          enable the Livepatch service
  disable-livepatch [-r]            disable the Livepatch service. With "-r", the
                                    canonical-livepatch snap will also be removed
- enable-cc_provisioning <TOKEN>    enable the commoncriteria PPA repository and
+ enable-cc-provisioning <TOKEN>    enable the commoncriteria PPA repository and
                                    install the ubuntu-commoncriteria .DEB package
 EOF
     error_exit invalid_command
@@ -93,9 +93,10 @@ main() {
 
     local service
     service=$(service_from_command "$command")
+    service=${service//-/_}
     # if the command contains a service name, check that it's valid
     if [ "$service" ] && ! name_in_list "$service" "$SERVICES" \
-           && [ "$service" != "fips-updates" ]; then
+           && [ "$service" != "fips_updates" ]; then
         error_msg "Invalid command: \"$command\""
         usage
     fi

```

vineethapai:
I was just trying that. In print_status function, where it checks for name_in_list "$service" "$SERVICES" , I will need to use original flag and also in service_print_status function where the $service : $status is printed. So far, I see two of those instances where original command is needed. It is better to revert in just two places than to may be change every where in service.sh. I agree.

vineethapai:
How does this look? 

@@ -42,12 +42,12 @@ print_status() {
 
     local services="$SERVICES"
     if [ "$service" ]; then
-        name_in_list "$service" "$SERVICES" || error_exit invalid_command
+        name_in_list "${service//_/-}" "$SERVICES" || error_exit invalid_command
         services="$service"
     fi
 
     for service in $services; do
-        service_print_status "$service"
+        service_print_status "${service//-/_}"
     done
 }
 
@@ -99,6 +99,10 @@ main() {
         error_msg "Invalid command: \"$command\""
         usage
     fi
+    # replace -(hyphen) in service commands with _(underscore) (eg: cc-provisioning) to
+    # use in generic service function invocations. Adding it here so the name_in_list
+    # function call above uses the original command.
+    service=${service//-/_}
 
     case "$command" in
         status)

vineethapai:
How about this?

@@ -42,12 +42,12 @@ print_status() {
 
     local services="$SERVICES"
     if [ "$service" ]; then
-        name_in_list "$service" "$SERVICES" || error_exit invalid_command
+        name_in_list "${service//_/-}" "$SERVICES" || error_exit invalid_command
         services="$service"
     fi
 
     for service in $services; do
-        service_print_status "$service"
+        service_print_status "${service//-/_}"
     done
 }
 
@@ -99,6 +99,10 @@ main() {
         error_msg "Invalid command: \"$command\""
         usage
     fi
+    # replace -(hyphen) in service commands with _(underscore) (eg: cc-provisioning) to
+    # use in generic service function invocations. Adding it here so the name_in_list
+    # function call above uses the original command.
+    service=${service//-/_}
 
     case "$command" in
         status)

vineethapai:
Ok, I reverted the service file. I made the necessary changes for replacing hypen with underscore from ubuntu-advantage script. The latest commit has the changes.

modules/service-cc.sh
```diff
@@ -0,0 +1,62 @@
...
+    # shellcheck disable=SC2086
+    check_result apt_get install $CC_PROVISIONING_UBUNTU_COMMONCRITERIA
+
+    echo -n "Successfully prepared this machine to host the Common Criteria artifacts. "
+    echo "Please follow instructions in /usr/lib/common-criteria/README to configure EAL2 on the target machine(s)."
+}
+
+cc_provisioning_disable() {
+    not_supported 'Disabling Common Criteria EAL2 Provisioning'
+}
```
panlinux: Since this is now just installing and removing a single deb, we can probably think about allowing disabling. It's a topic for another branch, though.

modules/service-cc.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo -n 'Running apt-get update... '
+    check_result apt_get update
+    echo 'Ubuntu Common Criteria PPA repository enabled.'
+
+    echo -n 'Installing Common Criteria artifacts (this may take a while)... '
+    # shellcheck disable=SC2086
+    check_result apt_get install $CC_PROVISIONING_UBUNTU_COMMONCRITERIA
+
+    echo -n "Successfully prepared this machine to host the Common Criteria artifacts. "
+    echo "Please follow instructions in /usr/lib/common-criteria/README to configure EAL2 on the target machine(s)."
```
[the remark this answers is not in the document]
vineethapai: output is printed to two lines in latest commit.

Yes, the files are all in one place to allow it to be easily tarred up and served to target machines. It is more for convenience providing all files related to CC at one place.

modules/service-cc.sh
```diff
@@ -0,0 +1,62 @@
...
+    apt_install_package_if_missing_file "$CA_CERTIFICATES" ca-certificates
+    echo -n 'Running apt-get update... '
+    check_result apt_get update
+    echo 'Ubuntu Common Criteria PPA repository enabled.'
+
+    echo -n 'Installing Common Criteria artifacts (this may take a while)... '
+    # shellcheck disable=SC2086
+    check_result apt_get install $CC_PROVISIONING_UBUNTU_COMMONCRITERIA
+
+    echo "Successfully prepared this machine to host the Common Criteria artifacts. "
```
panlinux: You can drop the whitespace at the end

vineethapai:
Added the entry for supporting CC provisioning to debian/changelog

PR #155, opened 2018-11-13
Add new flag enable-cisaudit to support cis audit

vineethapai:
The patchset introduces new flags enable-cisaudit and disable-cisaudit to
install automated tooling package for cis auditing on a machine running xenial. 
The new flag enable-cisaudit enables the security-benchmarks
PPA and installs the cis audit tooling package and the flag disable-cisaudit uninstalls
the cis audit tooling package, removes the repository and clears the apt entries.

Signed-off-by: Vineetha Kamath <vineetha.hari.pai@canonical.com>

### DIFF
--- a/modules/service-cis.sh
+++ b/modules/service-cis.sh
@@ -42,10 +42,14 @@ cisaudit_disable() {
                         "$APT_KEYS_DIR/$CISAUDIT_REPO_KEY_FILE"
         echo -n 'Running apt-get update... '
         check_result apt_get update
-        check_result apt_get remove $CISAUDIT_UBUNTU_CISBENCHMARK
-        echo "Canonical CIS Benchmark 16.04 Audit Tool Disabled."
+        echo "Canonical CIS Benchmark 16.04 Audit Tool Repository Disabled."
     else
-        echo 'Canonical CIS Benchmark 16.04 Audit Tool is not enabled.'
+        echo 'Canonical CIS Benchmark 16.04 Audit Tool Repository is not Enabled.'
+    fi
+
+    if apt_is_package_installed $CISAUDIT_UBUNTU_CISBENCHMARK; then
+        check_result apt_get remove $CISAUDIT_UBUNTU_CISBENCHMARK
+        echo 'Canonical CIS Benchmark 16.04 Audit Tool Removed.'
     fi
 }
 
--- a/modules/service-cis.sh
+++ b/modules/service-cis.sh
@@ -1,13 +1,13 @@
 # shellcheck disable=SC2034,SC2039
  
-CISAUDIT_SERVICE_TITLE="Canonical CIS 16.04 Benchmark Audit Tool"
+CISAUDIT_SERVICE_TITLE="Canonical CIS Benchmark 16.04 Audit Tool"
 CISAUDIT_SUPPORTED_SERIES="xenial"
 CISAUDIT_SUPPORTED_ARCHS="x86_64 ppc64le s390x"
 
 CISAUDIT_REPO_URL="https://private-ppa.launchpad.net/ubuntu-advantage/security-benchmarks"
 CISAUDIT_REPO_KEY_FILE="ubuntu-securitybenchmarks-keyring.gpg"
 CISAUDIT_REPO_LIST=${CISAUDIT_REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-cis-${SERIES}.list"}
-CISAUDIT_UBUNTU_CISBENCHMARK="cisbenchmark-16.04"
+CISAUDIT_UBUNTU_CISBENCHMARK="ubuntu-cisbenchmark-16.04"
 
 cisaudit_enable() {
     local token="$1"
@@ -15,7 +15,7 @@ cisaudit_enable() {
 
     _cisaudit_is_installed || result=$?
     if [ $result -eq 0 ]; then
-        error_msg "CIS audit package is already installed and files are available in /usr/lib/ubuntu-securityguides/cisbenchmark-16.04."
+        error_msg "CIS benchmark audit package is already installed and files are available in /usr/share/ubuntu-securityguides/$CISAUDIT_UBUNTU_CISBENCHMARK."
         error_exit service_already_enabled
     fi
 
@@ -33,11 +33,20 @@ cisaudit_enable() {
     check_result apt_get install $CISAUDIT_UBUNTU_CISBENCHMARK
 
     echo "Successfully installed the CIS audit tool."
-    echo "Please follow instructions in /usr/share/doc/cisbenchmark-16.04/README to run the CIS audit tool on the target machine(s)."
+    echo "Please follow instructions in /usr/share/doc/$CISAUDIT_UBUNTU_CISBENCHMARK/README to run the CIS audit tool on the target machine(s)."
 }
 
 cisaudit_disable() {
-    not_supported 'Disabling CIS Audit tool install'
+    if [ -f "$CISAUDIT_REPO_LIST" ]; then
+        apt_remove_repo "$CISAUDIT_REPO_LIST" "$CISAUDIT_REPO_URL" \
+                        "$APT_KEYS_DIR/$CISAUDIT_REPO_KEY_FILE"
+        echo -n 'Running apt-get update... '
+        check_result apt_get update
+        check_result apt_get remove $CISAUDIT_UBUNTU_CISBENCHMARK
+        echo "Canonical CIS Benchmark 16.04 Audit Tool Disabled."
+    else
+        echo 'Canonical CIS Benchmark 16.04 Audit Tool is not enabled.'
+    fi
 }
 
 cisaudit_is_enabled() {
@@ -45,7 +54,7 @@ cisaudit_is_enabled() {
 }
 
 cisaudit_print_status() {
-    echo "cisaudit: files are in /usr/lib/ubuntu-securityguides/cisbenchmark-16.04"
+    echo "cisaudit: files are in /usr/share/ubuntu-securityguides/$CISAUDIT_UBUNTU_CISBENCHMARK"
 }
 
 _cisaudit_is_installed() {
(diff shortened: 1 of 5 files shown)


ubuntu-advantage.1
```diff
@@ -101,10 +101,16 @@ package which has the common criteria artifacts. The artifacts include a
 configure script, a tarball with additional packages and post install scripts
 and an evaluated configuration guide. The artifacts will be installed in
 /usr/lib/common-criteria directory.
+.SH CIS (Canonical CIS Audit tooling)
+Enable CIS Auditing PPA and install CIS audit tool package
 .TP
 .B
-disable-cc-provisioning
-Disabling common criteria provisioning is not supported
```
panlinux: Did you remove disable-cc-provisioning from the manpage on purpose?

modules/service-cis.sh
```diff
@@ -0,0 +1,62 @@
...
+    # shellcheck disable=SC2086
+    check_result apt_get install $CISAUDIT_UBUNTU_CISBENCHMARK
+
+    echo "Successfully installed the CIS audit tool."
+    echo "Please follow instructions in /usr/share/doc/cisbenchmark-16.04/README to run the CIS audit tool on the target machine(s)."
+}
+
+cisaudit_disable() {
+    not_supported 'Disabling CIS Audit tool install'
+}
```
panlinux: Isn't this simple to disable? We can just remove the private-ppa (modules/service-esm.sh has an example) and them remove the single package we installed?

ubuntu-advantage.1
vineethapai: That was an unintentional. I am thinking to add support to disable-cc-provisioning flag along with disable-cisaudit flag since they are very similar. Can I do it in this same PR?

modules/service-cis.sh
vineethapai: yes, I added disable-cisaudit. We may change the package name and the path it is installed. Waiting for that to be finalized before I update the PR.

ubuntu-advantage.1
panlinux: I think disable-cisaudit can be done here, but disable-cc-provisioning needs its own branch.

modules/service-cis.sh
```diff
@@ -0,0 +1,62 @@
...
+CISAUDIT_REPO_LIST=${CISAUDIT_REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-cis-${SERIES}.list"}
+CISAUDIT_UBUNTU_CISBENCHMARK="cisbenchmark-16.04"
+
+cisaudit_enable() {
+    local token="$1"
+    local result=0
+
+    _cisaudit_is_installed || result=$?
+    if [ $result -eq 0 ]; then
+        error_msg "CIS audit package is already installed and files are available in /usr/lib/ubuntu-securityguides/cisbenchmark-16.04."
```
panlinux: If the directory `cisbenchmark-16.04` will be named after the package, then perhaps it's better to use `$CISAUDIT_UBUNTU_CISBENCHMARK` here

modules/service-cis.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo -n 'Running apt-get update... '
+    check_result apt_get update
+    echo 'Ubuntu Security Benchmarks PPA repository enabled.'
+
+    echo -n 'Installing CIS audit benchmark tool (this may take a while)... '
+    # shellcheck disable=SC2086
+    check_result apt_get install $CISAUDIT_UBUNTU_CISBENCHMARK
+
+    echo "Successfully installed the CIS audit tool."
+    echo "Please follow instructions in /usr/share/doc/cisbenchmark-16.04/README to run the CIS audit tool on the target machine(s)."
```
panlinux: If the directory `cisbenchmark-16.04` will be named after the package, then perhaps it's better to use `$CISAUDIT_UBUNTU_CISBENCHMARK` here

modules/service-cis.sh
vineethapai: Done

modules/service-cis.sh
vineethapai: Done

vineethapai:
Please use ubuntu-cisbenchmark-16.04 from "https://private-ppa.launchpad.net/fips-cc-stig/fipsdevppa"  to test the PR.

ubuntu-advantage.1
vineethapai: OK

panlinux:
Sorry, I don't have access to that private ppa.

panlinux:
Thanks for the changes, they all look good. I would just like to give it a run with the real package to exercise the code paths in a real environment and then we can merge this.

vineethapai:
You are listed as one of the members with access to the fipsdevppa. I also added you to the fips-cc-stig team.

modules/service-cis.sh
```diff
@@ -0,0 +1,71 @@
...
+        apt_remove_repo "$CISAUDIT_REPO_LIST" "$CISAUDIT_REPO_URL" \
+                        "$APT_KEYS_DIR/$CISAUDIT_REPO_KEY_FILE"
+        echo -n 'Running apt-get update... '
+        check_result apt_get update
+        check_result apt_get remove $CISAUDIT_UBUNTU_CISBENCHMARK
+        echo "Canonical CIS Benchmark 16.04 Audit Tool Disabled."
+    else
+        echo 'Canonical CIS Benchmark 16.04 Audit Tool is not enabled.'
+    fi
+}
```
panlinux: I was trying this disable option, and I think it can be improved a bit. For example, if there is a transient network error while running `apt_get update`, the script will flag that and stop. The `ubuntu-cisbenchmark` package won't be removed (that would have been the next step), but the repository is gone already.

The user just tries again then. This time, however, since the repository is gone already, `cisaudit_disable()` will just print that the tool is not enabled, when in fact it's still installed. I think we should decouple the repository removal with the package removal. Since `cisaudit_disable()` is only called if the package is installed, we can remove the package outside of the repository conditional.

modules/service-cis.sh
```diff
@@ -0,0 +1,71 @@
+# shellcheck disable=SC2034,SC2039
+ 
+CISAUDIT_SERVICE_TITLE="Canonical CIS Benchmark 16.04 Audit Tool"
+CISAUDIT_SUPPORTED_SERIES="xenial"
+CISAUDIT_SUPPORTED_ARCHS="x86_64 ppc64le s390x"
+
+CISAUDIT_REPO_URL="https://private-ppa.launchpad.net/ubuntu-advantage/security-benchmarks"
+CISAUDIT_REPO_KEY_FILE="ubuntu-securitybenchmarks-keyring.gpg"
+CISAUDIT_REPO_LIST=${CISAUDIT_REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-cis-${SERIES}.list"}
+CISAUDIT_UBUNTU_CISBENCHMARK="ubuntu-cisbenchmark-16.04"
```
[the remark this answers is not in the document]
vineethapai: ubuntu-cisbenchmark-16.04 is a binary package build from the source package ubuntu-securityguides. The other package you see on the fipsdevppa is an old one.

modules/service-cis.sh
```diff
@@ -0,0 +1,71 @@
...
+
+cisaudit_enable() {
+    local token="$1"
+    local result=0
+
+    _cisaudit_is_installed || result=$?
+    if [ $result -eq 0 ]; then
+        error_msg "CIS benchmark audit package is already installed and files are available in /usr/share/ubuntu-securityguides/$CISAUDIT_UBUNTU_CISBENCHMARK."
```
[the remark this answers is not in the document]
vineethapai: When I used CISAUDIT_REPO_URL="https://private-ppa.launchpad.net/fips-cc-stig/fipsdevppa", and ran the ubuntu-advantage tool with the creds, it installed the ubuntu-cisbenchmark-16.04 package with files in /usr/share/ubuntu-securityguides/ubuntu-cisbenchmark-16.04 directory.

modules/service-cis.sh
```diff
@@ -0,0 +1,71 @@
...
+    echo -n 'Running apt-get update... '
+    check_result apt_get update
+    echo 'Ubuntu Security Benchmarks PPA repository enabled.'
+
+    echo -n 'Installing CIS audit benchmark tool (this may take a while)... '
+    # shellcheck disable=SC2086
+    check_result apt_get install $CISAUDIT_UBUNTU_CISBENCHMARK
+
+    echo "Successfully installed the CIS audit tool."
+    echo "Please follow instructions in /usr/share/doc/$CISAUDIT_UBUNTU_CISBENCHMARK/README to run the CIS audit tool on the target machine(s)."
```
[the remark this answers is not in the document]
vineethapai: Yes README is yet to be created. It is currently not available in the package.

modules/service-cis.sh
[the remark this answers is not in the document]
panlinux: Indeed it is correct. In fact, I now think this code is never reached, because before _service_check_enabled() is called before cisaudit_enable(). I understand it was copied from another module and can be left as is.

modules/service-cis.sh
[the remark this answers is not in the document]
panlinux: ok

modules/service-cis.sh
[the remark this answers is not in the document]
panlinux: Ok.

modules/service-cis.sh
```diff
@@ -0,0 +1,75 @@
...
+        check_result apt_get update
+        echo "Canonical CIS Benchmark 16.04 Audit Tool Repository Disabled."
+    else
+        echo 'Canonical CIS Benchmark 16.04 Audit Tool Repository is not Enabled.'
+    fi
+
+    if apt_is_package_installed $CISAUDIT_UBUNTU_CISBENCHMARK; then
+        check_result apt_get remove $CISAUDIT_UBUNTU_CISBENCHMARK
+        echo 'Canonical CIS Benchmark 16.04 Audit Tool Removed.'
+    fi
```
panlinux: I suggest an `else` clause here for the package removal, just like above for the repository removal, saying that the tool is not installed.

panlinux: ...except `cisaudit_disable()` wouldn't even be called if the package were not installed. Well, the suggestion is more about symmetry then, up to you if you want to add it.

vineethapai: yes, 'disable' checks if the package is installed first and wouldn't even go there, so I didn't add the else.

Issue #139, opened 2018-04-03
Why is "strongswan" mandatory?


panlinux:
This came up as a review comment in https://code.launchpad.net/~ahasenack/ubuntu/+source/ubuntu-advantage-tools/+git/ubuntu-advantage-tools/+merge/342208

_Umm, don't you want to push the FIPS version of any package being installed (or that will be later installed via the pins) instead of unconditionally installing them?
You force e.g. strongswan onto all the fips systems without any need. Most might just use kernel and libssl.
I personally think it is wrong to just install all of them._

The suggestion is to force install what is really needed for fips, but to just upgrade to the fips version in the case of other pcakages like strongswan. If the user doesn't have strongswan already installed, then we shouldn't install the fips one. Similarly for other packages.

joylatten:
We cannot anticipate which fips modules a customer wants, so the enable-fips installs all the certified fips modules. Thus naturally assumed the update-fips should update all of them. Manual install instructions are published and provided for customers that prefer to "customize" their fips installations. See, https://assets.ubuntu.com/v1/f35fe02e-Canonical+FIPS+Installation+Instructions.pdf?utm_source=ubunteu&utm_medium=url_shortner&utm_term=fips-install&utm_campaign=shortner

Given that after initial installation, whether done manually or via enable-fips, a customer may have removed fips modules not desired, the update perhaps can loop through and see which fips modules are installed, and only update those.

If that sounds good, then perhap an issue can be opened to do that for xenial.

cpaelzer:
Hi @joylatten I'm unsure if I followed correctly, but I'm totally with @panlinux on this one.

Installing almost all packages just for enabling FIPS increases the attack surface and the install footprint - that seems wrong.

It is right to force install what is needed always (like the kernel), but only modify the apt sources so that if a user installs other packages which have FIPS support he will get the FIPSified version of them. 

Quoting and modifying your statement:
You: "We cannot anticipate which fips modules a customer wants, so the enable-fips installs all the certified fips modules"
Suggestion: "We cannot anticipate which fips modules a customer wants, so the enable-fips updates the apt information of all the certified fips modules, sets mandatory packages to be installed and then - by triggering apt upgrade - installs all of those. Unused packages will not be installed, but if ever installed will fetch the FIPS version of the packages."

panlinux:
There was a PR to fix this in the old shell client. We didn't land it in time, and it wasn't adopted for the new client yet.

[...]

PR #157, opened 2018-11-30
Update only installed FIPS packages for a fips-update

vineethapai:
There is currently an issue open - "Why is "strongswan" mandatory? #139"
This patch updates only the installed user space FIPS packages on a fips-update. The fips kernel and initramfs are always updated. The packages openssl, openssh-server, openssh-client and strongswan are only updated if they are installed on the machine.
Signed-off-by: Vineetha Kamath <vineetha.hari.pai@canonical.com>


powersj:
I'm going to close this merge as it will no longer apply to the new client code. The issue (#139) will remain open during the course of further development.

PR #160, opened 2019-01-11
Revert "Merge pull request #147 from panlinux/switch-kernel-for-lp"

panlinux:
This reverts commit 77606776c83d118264d4bc7efe898b937d28febd, reversing
changes made to df02ea5f24e43cc395df878269895b2ad8d6cb0d.

Dropping the switch-kernel-for-lp merge, as that feature is incomplete and
the way forward is not clear. What's missing is a way to keep the machine on the selected kernel, for the next boot and future kernel upgrades.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -2,13 +2,11 @@ ubuntu-advantage-tools (18) UNRELEASED; urgency=medium
 
   * Have ua status cope with the additional livepatch of running a kernel
     that is not supported for livepatches.
-  * Have an option for enable-livepatch to install a compatible kernel if
-    needed.
 
-  [ Vineetha Kamath ]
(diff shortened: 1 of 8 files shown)


basak:
lgtm

One note. Not a -1 or anything as I don't have the context, and I'll leave it to your discretion. ef4a688 introduced a new exit code 9 I think? Depending on if you think you'll ever see that in the wild, you might want to preserve that as no longer used somehow so it doesn't ever get used for anything else, for disambiguation in case you ever get a report from a point prior to this revert.

panlinux:
Thanks for the note. This code wasn't released, apart from the daily ppa devel builds, which I doubt has external users other than me and perhaps other developers.

PR #163, opened 2019-01-29
Python ua client v1

blackboxsw:
Adapt conact api calls and request handling to some v1 paths present in full contract.yaml schema

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -9,6 +9,9 @@ API_PATH_TMPL_CONTRACT_MACHINES = '/contracts/{contract}/context/machines'
 API_PATH_TMPL_MACHINE_CONTRACT = '/machines/{machine}/contract'
 API_PATH_TMPL_RESOURCE_MACHINE_ACCESS = '/resources/{resource}/context/machine'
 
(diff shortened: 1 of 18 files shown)


uaclient/contract.py
```diff
@@ -25,7 +29,7 @@ def __init__(self, e, error_response):
         else:
             self.api_errors = [error_response]
         for error in self.api_errors:
-            error['code'] = error['title']
+            error['code'] = error.get('title', error.get('code'))
```
raharper: why would the error have and why are we replacing it with title if present?
if 'title in error:
error['code'] = error['title']
?

uaclient/contract.py
```diff
@@ -103,8 +107,7 @@ def request_machine_contract_status(
         self.cfg.write_cache('machine-contracts', contracts)
         return contracts
 
-    def request_contract_machine_attach(self, contract_id, user_token,
-                                        machine_id=None):
+    def request_contract_machine_attach(self, user_token, machine_id=None):
         """Requests machine attach to the provided contact_id.
 
         @param contract_id: Unique contract id provided by contract service.
```
raharper: docstring still mentions contract_id, but that's no longer in the signature

PR #167, opened 2019-03-08
tox: use nose2, speed up run, and add coverage

powersj:
This utilizes nose2 so we can also turn on coverage for the python
files. By setting the envdir the tests also run considerably faster.
The file was cleaned up to make it easier to read.

The setup.py was added to tox and corrections made to clean up lint errors.

The *requirements.txt files were alphabetized and cleaned up.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,55 +1,56 @@
+# Copyright (C) 2019 Canonical Ltd.
+# This file is part of ubuntu-advantage-client.  See LICENSE file for license.
+
 import glob
 import setuptools
-import subprocess
-import os
 
 from uaclient import config
 from uaclient.util import subp
 
 NAME = 'ubuntu-advantage-tools'
 
+raw_requirements, _err = subp(['./dev/read-dependencies'])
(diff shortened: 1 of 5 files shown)


powersj:
Example output:
https://paste.ubuntu.com/p/x8xHJCghz4/

setup.py
```diff
@@ -1,55 +1,56 @@
...
 from uaclient import config
 from uaclient.util import subp
 
 NAME = 'ubuntu-advantage-tools'
 
+raw_requirements, _err = subp(['./dev/read-dependencies'])
+INSTALL_REQUIRES = raw_requirements.rstrip('\n').split('\n')
+
+raw_test_requirements, _err = subp([
+    './dev/read-dependencies', '-r', 'test-requirements.txt']
```
blackboxsw: Thanks for this fixup. No sense in having two copies around. Will have to clean up read-dependencies at some point to gut py2 stuff. I'll look over your other branches and if you haven't already done this I'll add a low-weight issue.

PR #173, opened 2019-03-11
d/control: fix build-depends

### LINKED ISSUE #170: FTBFS in Xenial, Bionic, Cosmic, and Disco
Build Logs:
- [Xenial](https://launchpadlibrarian.net/414294440/buildlog.txt.gz)
- [Bionic](https://launchpadlibrarian.net/414294433/buildlog.txt.gz)
- [Cosmic](https://launchpadlibrarian.net/414294432/buildlog.txt.gz)
- [Disco](https://launchpadlibrarian.net/414294439/buildlog.txt.gz)

Same traceback in all releases:
```shell
  File "setup.py", line 6, in <module>
    from uaclient import config
  File "/home/buildd/work/tree/recipe/uaclient/config.py", line 7, in <module>
    import yaml
ImportError: No module named 'yaml'
E: pybuild pybuild:274: clean: plugin distutils failed with: exit code=1: python3.5 setup.py clean 
```

powersj:
This adds the missing python3-setuptools and python3-yaml depedencies
needed for building on Xenial and newer.

I also re-ordered the depends to be more alphabetical.

Closes #170

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,9 +2,9 @@ Source: ubuntu-advantage-tools
 Section: misc
 Priority: important
 Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
-Build-Depends: debhelper (>=9), dh-python,
-               flake8 | python3-flake8, python3-mock,
-               python3 (>= 3.4), python3-nose, python3-unittest2
+Build-Depends: debhelper (>=9), dh-python, python3 (>= 3.4),
+               flake8 | python3-flake8, python3-mock, python3-nose,
+               python3-setuptools, python3-unittest2, python3-yaml
 Standards-Version: 3.9.6
 Homepage: https://buy.ubuntu.com
 Vcs-Git: https://github.com/CanonicalLtd/ubuntu-advantage-script.git


powersj:
Example sbuild with this patch for bionic:
https://paste.ubuntu.com/p/9rBg9yJYnH/

powersj:
Why do we depend on flake8 or python3-flake8? To be consistent should we always use python3-flake8 to pull down the python3 package?

panlinux:
> Why do we depend on flake8 or python3-flake8? To be consistent should we always use python3-flake8 to pull down the python3 package?
> 
> For what it is worth, Trusty does not have flake8, only python3-flake8.

That was the reason: trusty

powersj:
> That was the reason: trusty

I am confused how this was working on the existing client. When I run sbuild I get the following error:

> The following packages have unmet dependencies:
>  sbuild-build-depends-ubuntu-advantage-tools-dummy : Depends: flake8 but it is not installable
> E: Unable to correct problems, you have held broken packages.

Does the `|` syntax it is satisfied by one or the other packages, yet when it goes to install flake8 fails since it does not exist?

panlinux:
sbuild might be limited on trusty. I know that, for example, `apt-get build-dep ./` doesn't work on trusty, just in later releases.

The latest build log on trusty[1] does show python3-flake8 being installed, so the `|` syntax does work at least there.

That being said, I just noticed that slashd changed the trusty packaging to drop the tests at build time:
ubuntu-advantage-tools (10ubuntu0.14.04.2) trusty; urgency=medium

```
  * Skipping tests during package builds.
    - d/rules: Disabling the tests.
    - d/control: Removing tests dependencies.

 -- Eric Desrochers <eric.desrochers@canonical.com>  Tue, 07 Nov 2017 10:07:49 -0500

I have no recollection of why that was needed.

```
1. https://launchpadlibrarian.net/407743998/buildlog_ubuntu-trusty-i386.ubuntu-advantage-tools_18~122~git.bef5b2c~ubuntu14.04.1_BUILDING.txt.gz

basak:
On Mon, Mar 11, 2019 at 04:27:35PM -0700, Joshua Powers wrote:
> Does the `|` syntax it is satisfied by one or the other packages, yet when it goes to install flake8 fails since it does not exist?

I'm not sure of the context, but this sounds like it might need
--resolve-alternatives to sbuild maybe?

powersj:
> --resolve-alternatives to sbuild maybe?

This worked, however, now I run into an issue with the way flake8 is run. On Trusty, it is expected to be run via `flake8 ...` and on xenial and newer via `python3 -m flake8 ...`. I can see why he would want to pull it out at this point.

panlinux:
I remember changing that explicitly to a line that would work on all releases, trusty included. In old master, `d/rules` had this change:
```diff
 override_dh_auto_test:
        python3 -m unittest discover tests
-       flake8 tests
+       python3 /usr/bin/flake8 tests
```

panlinux:
The above was changed back in commit 90407726:
```diff
 override_dh_auto_test:
-       python3 -m unittest discover tests
-       python3 /usr/bin/flake8 tests
+       python3 -m unittest discover uaclient
+       python3 -m flake8 uaclient
```

powersj:
Closing this one as my Trusty fix supersedes this.

Thanks for the help panlinux and basak.

PR #174, opened 2019-03-12
Enable Trusty build, remove python-six, fix flake8

### LINKED ISSUE #169: FTBFS in Trusty
https://launchpadlibrarian.net/414294435/buildlog.txt.gz

E: Build-Depends dependency for ubuntu-advantage-tools cannot be satisfied because the package python3-unittest2 cannot be found

### LINKED ISSUE #172: dev/bddeb assumes Trusty is python2
The bddeb script has the following line:

```python
is_python2 = bool(args.series == 'trusty')
```

This is a false assumption. This script should remove any and all references to Python 2 as the client should be Python 3 only.

powersj:
Trusty is now buildable in an sbuild.

This removes dependencies on python-six, which was getting used
throughout. This is meant to be a Python 3 only application so there is
no need for python-six. Here is how various usage was updated:

python-six     --> python-3
string_types   --> str
binary_types   --> bytes
StringIO       --> io.StringIO
six.moves      --> import urllib or input()
@add_metaclass --> function signature

Due to differences in the way that flake8 can be installed across Trusty
to Bionic, this simplifies the dependency list and deals with the
different mechanisms for calling it. The trusty found flake8 errors are
fixed and one error is ignored.

Closes #169 #170 #172

### DIFF
--- a/dev/bddeb
+++ b/dev/bddeb
@@ -61,31 +61,23 @@ def run_helper(helper, args=None, strip=True):
     return stdout
 
 
-def write_debian_folder(root, templ_data, is_python2=False):
+def write_debian_folder(root, templ_data):
     """Create a debian package directory with all rendered template files."""
     print("Creating a debian/ folder in %r" % (root))
-    if is_python2:
-        pyver = "2"
-        python = "python"
-    else:
-        pyver = "3"
-        python = "python3"
-
-
 
     # Write out the control file template
     reqs_output = run_helper(
-        'read-dependencies', args=['--python-version', pyver, '--system-pkg-names'])
+        'read-dependencies', args=['--python-version', '3', '--system-pkg-names'])
     reqs = reqs_output.splitlines()
     test_reqs = run_helper(
         'read-dependencies',
         ['--requirements-file', 'test-requirements.txt',
-         '--system-pkg-names', '--python-version', pyver]).splitlines()
+         '--system-pkg-names', '--python-version', '3']).splitlines()
 
     requires = []
     # We consolidate all deps as Build-Depends as our package build runs all
     # tests so we need all runtime dependencies anyway.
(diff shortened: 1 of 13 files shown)


powersj:
Tests failed because they still run Python 2 linting. The metaclass syntax is not supported in Python 2.

dev/bddeb
```diff
@@ -61,7 +61,7 @@ def run_helper(helper, args=None, strip=True):
     return stdout
 
 
-def write_debian_folder(root, templ_data, is_python2):
+def write_debian_folder(root, templ_data, is_python2=False):
     """Create a debian package directory with all rendered template files."""
     print("Creating a debian/ folder in %r" % (root))
     if is_python2:
```
OddBloke: Can we just drop this code path entirely?

OddBloke:
One inline question; this looks good to me otherwise.

dev/bddeb
powersj: Agreed, I pushed another commit to remove the rest of the python2 verbage/conditionals

uaclient/util.py
```diff
@@ -3,8 +3,8 @@
 import logging
 import os
 import re
-import six
 import subprocess
+import urllib
```
blackboxsw: in python3.7 (disco) maybe even 3.4 and 3.5 urllib doesn't surface submodule names directly on the urllib module so It doesn't have to load all submodules I believe.
As a result we need 
```
from urllib import request

and below change the call sites to 

request.Request
&& 
request.urlopen

powersj:
ok I cleaned up the urllib imports in the serivceclient.py, util.py, and found another call in sso.py. I am 100% find with using requests instead, as you can tell I don't have much experience with using urllib directly :D

powersj:
2x +1s and tests pass, so I am merging to get some builds going

PR #191, opened 2019-03-19
test_base: rename TestConcreteEntitlement to not have Test prefix

OddBloke:
pytest attempts to collect TestConcreteEntitlement as a test class, and
emits a warning when it fails to do so.

(This change is fully compatible with continued use of nose2.)


OddBloke:
No oopsie, it's not a warning on nose2! Just something I noticed when seeing what `pytest` did, and the fix was easy. :-)

PR #200, opened 2019-03-20
 add flake8-{trusty,xenial,bionic} tox targets (and make them pass)

OddBloke:


tox.ini
```diff
@@ -1,15 +1,25 @@
...
+commands =
+    !flake8: nose2 --with-coverage --coverage uaclient/ uaclient
+    flake8: flake8 uaclient setup.py
 
+[flake8]
+ignore =
+    # Older versions of flake8 et al don't permit the now-recommended-by-PEP-8
+    # parameter spacing for annotated function arguments with defaults (e.g.
+    # `def spam(ham: str = "eggs"):`).
+    E251,
```
powersj: In debian/rules we call nose and flake8 directly instead of running it via tox, so we use the in-distro versions of packages. That means this ignore statement will not have any effect when building the package.

tox.ini
```diff
@@ -1,15 +1,25 @@
...
-commands = flake8 uaclient setup.py
-
-[testenv:pycodestyle]
-commands = pycodestyle {posargs:uaclient setup.py}
+    trusty: -cdev/trusty-constraints.txt
+    xenial: -cdev/xenial-constraints.txt
+    bionic: -cdev/bionic-constraints.txt
+commands =
+    !flake8: nose2 --with-coverage --coverage uaclient/ uaclient
+    flake8: flake8 uaclient setup.py
```
powersj: This didn't work for me with tox 2.5.0: `ERROR: InvocationError: could not find executable '!flake8:'` I'm not sure why the conditional formatting wasn't working.

In either case shouldn't the first line by "py3:" instead so it only runs on py3 and that way it also simplifies your N.B. comment for adding new tests?

tox.ini
```diff
@@ -1,15 +1,25 @@
...
-commands = nose2 --with-coverage --coverage uaclient/ uaclient
-
-[testenv:flake8]
-commands = flake8 uaclient setup.py
-
-[testenv:pycodestyle]
-commands = pycodestyle {posargs:uaclient setup.py}
+    trusty: -cdev/trusty-constraints.txt
+    xenial: -cdev/xenial-constraints.txt
+    bionic: -cdev/bionic-constraints.txt
```
powersj: I leave this up to you... but I prefer constraints-{release}.txt so they are all stacked near each other, but that is personal preference

tox.ini
OddBloke: Per [the flake8 documentation](http://flake8.pycqa.org/en/3.1.1/user/configuration.html#configuration-locations), flake8 will look in `setup.cfg`, `tox.ini` or `.flake8` for its configuration (regardless of how its being run). Given that we already have a `tox.ini`, I opted for that.

I've also just confirmed that this will work in trusty (using a container):

```
# apt install python3-flake8
...
# cat << EOF > foo.py
> def foo(x = "y"):
>     pass
> EOF
# flake8 foo.py 
foo.py:1:10: E251 unexpected spaces around keyword / parameter equals
foo.py:1:12: E251 unexpected spaces around keyword / parameter equals
# cat << EOF > tox.ini
> [flake8]
> ignore =
>     E251,
> EOF
# cat << EOF > tox.ini
> [flake8]
> ignore =
>     # This is a comment
>     E251,
> EOF
# flake8 foo.py 
# rm tox.ini
# flake8 foo.py 
foo.py:1:10: E251 unexpected spaces around keyword / parameter equals
foo.py:1:12: E251 unexpected spaces around keyword / parameter equals
```

I'm happy to switch to one of the other two if you think that would be preferable.

tox.ini
OddBloke: Yep, looks like the negative conditional factors were introduced in [this commit](https://github.com/tox-dev/tox/commit/2ce35c0020ff22401ec7b8207032f5f2257aa6f3) and so are only available from 3.0 onwards. The docs suggest that all the conditional factor syntax was introduced in 1.8 (which would be fine everywhere but trusty, and we don't need to run tox on trusty); I've filed https://github.com/tox-dev/tox/issues/1201 to address that.

I will switch to the syntax you suggested.

(I used `!flake8` so invocations with other Python versions (e.g. `py34`, `py37`, `py2`) would work. This would only really be necessary if we needed to support Python 2, which we don't. If we want other such invocations to work, we can explicitly add them in future. For reference, `tox -e py37` with the new `tox.ini` gives: `ERROR: Commands not specified. Please update relevant section of <snip>/ubuntu-advantage-client/tox.ini`.)

tox.ini
OddBloke: Yep, definitely preferable.

tox.ini
powersj: TIL... thanks for that!

PR #207, opened 2019-03-25
config: reset cached class attributes during config.delete_cache

### LINKED ISSUE #196: ua detach causes traceback in get_motd_status
`get_motd_summary` should properly return empty string on detached status.

possibly `cfg.is_attached` is not properly reporting false during 'detach' due to caching of `Config.machine_token`.

```pytb
[DEBUG]: File does not exist: /var/lib/ubuntu-advantage/account-contracts.json
  File "/usr/bin/ua", line 11, in <module>
    load_entry_point('ubuntu-advantage-tools==19.1', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 373, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 197, in action_detach
    ua_status.write_motd_summary(cfg)
  File "/usr/lib/python3/dist-packages/uaclient/status.py", line 114, in write_motd_summary
    ua_motd_status = get_motd_summary(cfg)
  File "/usr/lib/python3/dist-packages/uaclient/status.py", line 139, in get_motd_summary
    contractInfo = cfg.contracts['contracts'][0]['contractInfo']
TypeError: list indices must be integers or slices, not str
```

blackboxsw:
Avoid tracebacks when referencing deleted configuration which
was incorrectly persisted in memory even after a `ua detach`
operation.

Since our only current caller is Config.delete_cache, drop the key-specific
parameter/logic from Config.delete_cache.

Fixes #196

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -50,7 +50,10 @@ class UAConfig(object):
     @property
     def accounts(self):
         """Return the list of accounts that apply to this authorized user."""
-        return self.read_cache('accounts')['accounts']
+        accounts = self.read_cache('accounts')
+        if not accounts:
+            return []
+        return accounts['accounts']
 
     @property
     def contract_url(self):
@@ -116,7 +119,7 @@ class UAConfig(object):
             self._machine_token = self.read_cache('machine-token')
         return self._machine_token
 
-    def data_path(self, key):
+    def data_path(self, key=None):
         """Return the file path in the data directory represented by the key"""
         if not key:
             return self.cfg['data_dir']
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -53,6 +53,20 @@ class UAConfig(object):
         accounts = self.read_cache('accounts')
         if not accounts:
             return []
+        warning_msg = None
+        if not isinstance(accounts, dict):
+            warning_msg = ('Unexpected type %s in cache %s' %
+                           (type(accounts), self.data_path('accounts')))
+        elif 'accounts' not in accounts:
+            warning_msg = ("Missing 'accounts' key in cache %s" %
+                           self.data_path('accounts'))
+        elif not isinstance(accounts['accounts'], list):
+            warning_msg = (
+                "Unexpected 'accounts' type %s in cache %s" %
+                (type(accounts['accounts']), self.data_path('accounts')))
+        if warning_msg:
+            LOG.warning(warning_msg)
+            return []
         return accounts['accounts']
 
     @property
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -91,11 +91,18 @@ class UAConfig(object):
 
     @property
     def entitlements(self):
-        """Return the machine-token if cached in the machine token response."""
+        """Return a dictionary of entitlements keyed by entitlement name.
+
+        Return an empty dict if no entitlements are present.
+        """
         if self._entitlements:
             return self._entitlements
+        machine_token = self.machine_token
+        if not machine_token:
+            return {}
+
         self._entitlements = {}
-        contractInfo = self.machine_token['machineTokenInfo']['contractInfo']
+        contractInfo = machine_token['machineTokenInfo']['contractInfo']
         ent_by_name = dict(
             (e['type'], e) for e in contractInfo['resourceEntitlements'])
         for entitlement_name, ent_value in ent_by_name.items():
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -124,13 +124,12 @@ class UAConfig(object):
             return os.path.join(self.cfg['data_dir'], self.data_paths[key])
         return os.path.join(self.cfg['data_dir'], key)
 
-    def delete_cache(self, key=None):
-        if key is None:
-            for key in self.data_paths.keys():
-                cache_path = self.data_path(key)
-                if os.path.exists(cache_path):
-                    os.unlink(cache_path)
-        else:
+    def delete_cache(self):
+        """Remove all configuration cached response files class attributes."""
+        self._contracts = None
+        self._entitlements = None
+        self._machine_token = None
+        for key in self.data_paths.keys():
             cache_path = self.data_path(key)
             if os.path.exists(cache_path):
                 os.unlink(cache_path)
--- a/uaclient/tests/test_config.py
+++ b/uaclient/tests/test_config.py
@@ -18,6 +18,8 @@ KNOWN_DATA_PATHS = (('bound-macaroon', 'bound-macaroon'),
 
 class TestAccounts(TestCase):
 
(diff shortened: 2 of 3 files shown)


blackboxsw:
to test:

make deb
# in some bionic/xenial/trusty etc container
container $ dpkg -i ubuntu-advantage-tools*deb

container $ sudo cat > /etc/ubuntu-advantage/uaclient.conf <<EOF
# Ubuntu-Advantage client config file.
sso_auth_url: 'https://login.ubuntu.com'
contract_url: 'https://contracts.staging.canonical.com'
data_dir: /var/lib/ubuntu-advantage
log_level: info
EOF

container $ sudo ua attach
# SSO 2FA
container $ sudo ua detach
# should have no traceback 



sudo ua attach
# sso

uaclient/tests/test_config.py
```diff
@@ -0,0 +1,172 @@
...
+        with open(data_path, 'w') as f:
+            f.write(json.dumps(expected))
+
+        assert expected == cfg.read_cache(key)
+
+
+class TestDeleteCache(TestCase):
+
+    @params('_contracts', '_entitlements', '_machine_token')
+    def test_delete_cache_unsets_any_local_cached_class_attributes(self, attr):
```
OddBloke: In my opinion, this test is tied more closely to the implementation than is really necessary. I think we'd be better testing the invariants of the delete_cache operation (i.e. if things are set before, they are unset afterwards, and corresponding cache files are removed), so that if we decide to add an in-memory cache for another attribute (or, for an example that would require more than extending @params on this test, if we decide to move the in-memory caching to be in a single dict rather than separate attributes), we don't have to modify these tests.

uaclient/tests/test_config.py
```diff
@@ -0,0 +1,172 @@
...
+        """The delete_cache unsets any locally cached class attributes."""
+        tmp_dir = self.tmp_dir()
+        cfg = UAConfig({'data_dir': tmp_dir})
+        assert None is getattr(cfg, attr)
+        setattr(cfg, attr, attr)
+        assert None is not getattr(cfg, attr)
+        assert None is cfg.delete_cache()
+        assert None is getattr(cfg, attr)
+
+    def test_delete_cache_removes_any_cached_data_path_files(self):
```
OddBloke: Along the lines of above, I believe this _is_ an appropriate test; we want to be sure that the old values aren't sitting on-disk for our code to pick up in future.

uaclient/tests/test_config.py
```diff
@@ -0,0 +1,172 @@
...
+            assert expected_json_content == stream.read()
+        assert value == cfg.read_cache(key)
+
+
+class TestReadCache(TestCase):
+
+    @params(*KNOWN_DATA_PATHS)
+    def test_read_cache_returns_none_when_data_path_absent(
+            self, key, path_basename):
+        """The data_path method returns the data_dir when key is absent."""
```
OddBloke: This docstring doesn't look like it matches the test.

uaclient/config.py
```diff
@@ -50,7 +50,10 @@ def __init__(self, cfg=None):
     @property
     def accounts(self):
         """Return the list of accounts that apply to this authorized user."""
-        return self.read_cache('accounts')['accounts']
+        accounts = self.read_cache('accounts')
+        if not accounts:
+            return []
+        return accounts['accounts']
```
OddBloke: I wonder if we should be catching `KeyError` here and raising an exception with a better error, in case the accounts cache file ends up incorrect (it could be empty if it was written when the disk was full, for example).

(We could also consider checking that `accounts['accounts']` is a list, to avoid potentially very confusing errors elsewhere in the codebase.)

uaclient/tests/test_config.py
```diff
@@ -0,0 +1,172 @@
...
+
+from nose2.tools.params import params
+
+from uaclient.config import UAConfig
+from uaclient.testing.helpers import TestCase
+
+
+# These are in a variable rather than inline to work around
+# https://github.com/nose-devs/nose2/issues/433
+SIMPLE_PARAMS = (('machine_token', 'machine-token', None),
```
OddBloke: This is no longer in use.

uaclient/tests/test_config.py
```diff
@@ -0,0 +1,172 @@
...
+        cfg.write_cache('accounts', {'accounts': ['acct1', 'acct2']})
+
+        assert ['acct1', 'acct2'] == cfg.accounts
+
+
+class TestDataPath(TestCase):
+
+    def test_data_path_returns_data_dir_path_without_key(self):
+        """The data_path method returns the data_dir when key is absent."""
+        cfg = UAConfig({'data_dir': '/my/dir'})
```
OddBloke: We should consider using a temporary directory here (and below); it wouldn't be surprising for `UAConfig` to ensure its `data_dir` exists at instantiation (rather than on first write as is currently the case), and that could cause some interesting problems.

blackboxsw: "it wouldn't be surprising for UAConfig to ensure its data_dir exists at instantiation".  

I guess I don't have a strong case here, but I guess from the Landscape projects we had tried avoiding side-effects in __init__  just to avoid work on instantiation and testing related to that init. I *think* that suggestion really comes from old C++ constructor concerns and python style choices more than actual defensible python programming practices. So, I'm game for moving it into __init__. Though I'd generally like to avoid complexity in __init__ if we can. (This case doesn't really suggest complexity).

OddBloke: Yep, I broadly agree that we shouldn't do it; this was me overthinking how robust our tests need to be.

uaclient/tests/test_config.py
blackboxsw: +1

uaclient/tests/test_config.py
blackboxsw: Changed up the tests a bit to validate cached propert cfg.entitlements and cfg.contracts public properties with previous settings get cleared by delete_cache

blackboxsw:
> I'd like to see some tests added for this; I started playing around with this last week, so you could potentially use [these tests](http://paste.ubuntu.com/p/nWMGffjg2f/) as a basis.

Thanks for the unit tests. I've added coverage based on your suggestions and review comments

blackboxsw:
oops

uaclient/tests/test_config.py
blackboxsw: reused in a later commit. thanks for that sample unit test

powersj:
The Python: 3.7-dev travi-ci build has been attempting to run for hours, I am going to merge this anyways since the other tasks have passed successfully.

Issue #206, opened 2019-03-25
'UA attach <token>' to accept Contract token only instead of macaroon


blackboxsw:
Per https://github.com/CanonicalLtd/ua-contracts/issues/171 the token that is surfaced is a contract token meant for the add machine token route. Not a top-level user-token. ua client <token> currently expects a macaroon for access to accounts routes, We need to limit ua client to using the contract-related routes when the providing a contract token on the commandline.

PR #208, opened 2019-03-25
cli: only call contract.get_contract_token_for_account during SSO auth

blackboxsw:
When ua attach <contractToken> is provided, ua client should only talk
to the /v1/context/machines/token route with a Bearer token auth.
  
Only in SSO auth do we have a bound macaroon with which ua client can
talk to ua-contracts endpoints to query account info and create a
contract token via the following SSOAuth routes:
  /v1/accounts,
  /v1/accounts/{account}/contracts,
  /v1/contracts/{contract}/token

Rework existing ua_attach_cli unit happy path unit test to assert macaroon auth isn't being
called when token is provided.
  
Fixes: #206

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -223,6 +223,7 @@ def action_attach(args, cfg):
             return 1
         cfg.write_cache('bound-macaroon', bound_macaroon)
         try:
+            contract_client.request_accounts(bound_macaroon)
             contract_token = contract.get_contract_token_for_account(
                 contract_client, bound_macaroon, cfg.accounts[0]['id'])
         except (sso.SSOAuthError, util.UrlError) as e:
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -212,33 +212,30 @@ def action_attach(args, cfg):
         print(ua_status.MESSAGE_NONROOT_USER)
         return 1
     contract_client = contract.UAContractClient(cfg)
-    bound_macaroon = None
     if not args.token:
-        bound_macaroon_bytes = sso.discharge_root_macaroon(
-            contract_client)
+        bound_macaroon = None
+        bound_macaroon_bytes = sso.discharge_root_macaroon(contract_client)
         if bound_macaroon_bytes is not None:
             bound_macaroon = bound_macaroon_bytes.decode('utf-8')
+        if not bound_macaroon:
+            print('Could not attach machine. Unable to obtain authenticated'
+                  ' user token')
+            return 1
+        cfg.write_cache('bound-macaroon', bound_macaroon)
+        try:
+            contract_token = contract.get_contract_token_for_account(
+                contract_client, bound_macaroon, cfg.accounts[0]['id'])
+        except (sso.SSOAuthError, util.UrlError) as e:
+            logging.error(str(e))
+            print('Could not attach machine. Unable to obtain authenticated'
+                  ' contract token')
+            return 1
     else:
-        bound_macaroon = args.token
-    if not bound_macaroon:
-        print('Could not attach machine. Unable to obtain authenticated user'
-              ' token')
-        return 1
-    bound_macaroon = bound_macaroon
-    cfg.write_cache('bound-macaroon', bound_macaroon)
-    try:
-        contract_client.request_accounts(bound_macaroon)
-        contracts = contract_client.request_account_contracts(
-            bound_macaroon, cfg.accounts[0]['id'])
-        contract_id = contracts['contracts'][0]['contractInfo']['id']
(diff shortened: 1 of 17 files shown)


blackboxsw:
This branch probably needs some unit tests specifically on get_contract_token_for_account.

uaclient/cli.py
```diff
@@ -209,41 +209,38 @@ def action_attach(args, cfg):
         print(ua_status.MESSAGE_NONROOT_USER)
         return 1
     contract_client = contract.UAContractClient(cfg)
-    bound_macaroon = None
     if not args.token:
-        bound_macaroon_bytes = sso.discharge_root_macaroon(
-            contract_client)
+        bound_macaroon = None
+        bound_macaroon_bytes = sso.discharge_root_macaroon(contract_client)
         if bound_macaroon_bytes is not None:
```
OddBloke: I think this could be simplified now that we don't need `bound_macaroon` outside of this `if` branch:

```py
bound_macaroon_bytes = sso.discharge_root_macaroon(contract_client)
if bound_macaroon_bytes is None:
    print('Could not attach machine. Unable to obtain authenticated user token')
    return 1
bound_macaroon = bound_macaroon_bytes.decode('utf-8')
```

uaclient/tests/test_cli_attach.py
```diff
@@ -86,23 +88,26 @@ def test_happy_path_without_arg(self, action_status, contract_client,
...
+        token = 'contract-token'
         args = mock.MagicMock(token=token)
         cfg = TestConfig.with_account()
 
         ret = action_attach(args, cfg)
 
         assert 0 == ret
         assert 1 == action_status.call_count
-        assert token == cfg._cache_contents['bound-macaroon']
+        contract_machine_attach.called_once_with(contract_token=token)
```
OddBloke: The assertion methods on mocks are best avoided, because typos in the names of the methods are silently swallowed:

```
>>> my_mock = mock.Mock()
>>> my_mock()
<Mock name='my_mock()' id='139900840650568'>
>>> my_mock.assert_not_called()
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/mock/mock.py", line 897, in assert_not_called
    raise AssertionError(msg)
AssertionError: Expected 'my_mock' to not have been called. Called 1 times.
>>> my_mock.asert_not_called()
<Mock name='mock.asert_not_called()' id='139900841774720'>
```

This is such a problem, in fact, that mock tries to catch some of these cases, but it obviously won't be perfect: https://github.com/python/cpython/blob/master/Lib/unittest/mock.py#L595

For these two, I would propose:

```py
assert [mock.call(contract_token=token)] == contract_machine_attach.call_args_list
assert 0 == discharge_root_macaroon.call_count
```

In both of these cases, if you typo the attribute on the mock, the assertion will fail.

blackboxsw: +1 changed unit tests to avoid error-prone typos on mock asserts. Thanks for this +1

uaclient/cli.py
blackboxsw: Fixed with your suggestion here too. Thanks Dan.

uaclient/tests/test_cli_attach.py
OddBloke: Perhaps GH is slow to update, but it looks like we're still using `called_once_with`?

uaclient/tests/test_config.py
```diff
@@ -27,14 +27,25 @@ def test_accounts_returns_empty_list_when_no_cached_account_value(self):
...
         assert ['acct1', 'acct2'] == cfg.accounts
 
+    def test_accounts_extracts_accounts_key_from_machine_token_cache(self):
+        """Use machine_token cached accountInfo when no accounts cache."""
+        tmp_dir = self.tmp_dir()
+        cfg = UAConfig({'data_dir': tmp_dir})
+        accountInfo = {'id': '1', 'name': 'accountname'}
+
+        cfg.write_cache('machine-token',
+                {'machineTokenInfo': {'accountInfo': accountInfo}})
```
powersj: uaclient/tests/test_config.py:45:17: E128 continuation line under-indented for visual indent

uaclient/tests/test_cli_attach.py
powersj: still not resolved

blackboxsw: thanks . just pushed a fix for this unit test

uaclient/tests/test_config.py
blackboxsw: fd05dced5aecb16bffaa09b5c5437f7d13f75a7afixed lint in commit

PR #216, opened 2019-03-27
apt: fips, cis and esm now use aptURL instead of serviceURL directive

### LINKED ISSUE #215: apt: use aptURL directive  instead of serviceURL
for repo-based entitlements use aptUrl  instead of serviceURL

Example entitlement response:
{
   "entitlement" : {
      "obligations" : {
         "enableByDefault" : true
      },
      "affordances" : {
         "series" : [
            "xenial"
         ]
      },
      "directives" : {
         "aptKey" : "A166877412DAC26E73CEBF3FF6C280178D13028C",
         "aptURL" : "http://35.231.183.48/"
      },
      "type" : "fips",
      "entitled" : true
   },
   "resourceToken" : "<READACTED>npU8YmN9VT0wnA"
}

blackboxsw:
ua-contracts surfaces an aptURL directive to specify the entitlement's production/staging URL.
Fix the remaining cases where serviceURL was still being used in esm, fips and cis entitlements.

Fixes #215

### DIFF
--- a/uaclient/entitlements/tests/test_cis.py
+++ b/uaclient/entitlements/tests/test_cis.py
@@ -1,0 +1,87 @@
+"""Tests related to uaclient.entitlement.base module."""
+
+import mock
+from io import StringIO
+
+from uaclient import config
+from uaclient.entitlements.cis import CISEntitlement
+from uaclient.testing.helpers import TestCase
+
+
+CIS_MACHINE_TOKEN = {
+    'machineSecret': 'blah',
+    'machineTokenInfo': {
+        'contractInfo': {
+            'resourceEntitlements': [
+                {'type': 'cis-audit'}]}}}
+
+
+CIS_RESOURCE_ENTITLED = {
+    'resourceToken': 'TOKEN',
+    'entitlement': {
+        'obligations': {
+            'enableByDefault': True
+        },
+        'type': 'cis-audit',
+        'entitled': True,
+        'directives': {
+            'aptURL': 'http://CIS',
+            'aptKey': 'APTKEY'
+        },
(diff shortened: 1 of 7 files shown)


uaclient/entitlements/tests/test_cis.py
```diff
@@ -0,0 +1,87 @@
...
+            with mock.patch('uaclient.apt.add_ppa_pinning') as m_add_pinning:
+                self.assertTrue(entitlement.enable())
+
+        add_apt_calls = [
+            mock.call('/etc/apt/sources.list.d/ubuntu-cis-audit-xenial.list',
+                      'http://CIS', 'TOKEN', None, 'APTKEY')]
+
+        subp_apt_cmds = [
+            mock.call(['apt-get', 'update'], capture=True),
+            mock.call(['apt-get', 'install', 'ubuntu-cisbenchmark-16.04'])]
```
powersj: If the package name changes, and I hope it does, will this start failing?

blackboxsw: If the package name changes, we need to release a new UA-client, currently packages being installed are a static hardcoded list of packages for fips/fips-updates  and cis, We could ask the ua-contract backend to surface a new 'packages' directive which would allow us to react to a dynamic package dependency change in the future. There is nothing that the PPA itself can tell us about what needs to be installed (or run post-ppa enablement)

blackboxsw: We can also bake in the option to look for a 'packages' directive in ua client now and use that instead of the static list if ua-contracts surfaces it in the future.

PR #218, opened 2019-03-27
apt: handle macaroon resourceToken from ua-contract instead of username:passwd

blackboxsw:
ua-contracts service provides a macaroon for a resourceToken instead of generally PPA credentials of the format username:password.

When obtaining a single auth token from ua-contracts service, apt authentication will bearer:.

Handle a ValueError if credentials are not of the format username:password and instead use bearer as the username and as the password.

This diff is only big while we await prerequisite #216.

Here is the incremental diff
http://paste.ubuntu.com/p/byvKGBg4c6/


powersj:
local tests passed, merging

PR #219, opened 2019-03-27
livepatch: honor config directives caCerts and remoteServer

### LINKED ISSUE #211: livepatch: observe entitlement configuration directives caCerts and remoteServer
ua-contracts surfaces two config directives caCerts and remoteServer that should be passed to livepatch before enable

blackboxsw:
contracts.staging.canonical.com surfaces remoteServer and caCerts directives for livepatch,

Allow ua client to pass those config options to canonical-livepatch config X=y

Fixes #211

### DIFF
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -59,7 +59,7 @@ class LivepatchEntitlement(base.UAEntitlement):
         'Canonical Livepatch Service'
         ' (https://www.ubuntu.com/server/livepatch)')
 
-    # Use a lamda so we can mock util.is_container in tests
+    # Use a lambda so we can mock util.is_container in tests
     static_affordances = (
         ('Cannot install Livepatch on a container',
          lambda: util.is_container(),
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -61,6 +61,7 @@ class LivepatchEntitlement(base.UAEntitlement):
 
     def __init__(self, cfg=None):
         super().__init__(cfg)
+        # Set static_affordances in init so we can mock is_container in tests
         self.static_affordances = (
             ('Cannot install Livepatch on a container', util.is_container,
              False),)
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -139,16 +139,19 @@ class LivepatchEntitlement(base.UAEntitlement):
         return operational_status
 
 
-def process_directives(entitlement_cfg):
+def process_directives(cfg):
     """Process livepatch configuration directives.
 
     @raises: ProcessExecutionError if unable to configure livepatch.
     """
-    remote_server = entitlement_cfg.get('remoteServer')
+    if not cfg:
+        return
+    directives = cfg.get('entitlement', {}).get('directives', {})
+    remote_server = directives.get('remoteServer')
     if remote_server:
         util.subp(['/snap/bin/canonical-livepatch', 'config',
-                   'remote-server=%s' % remote_server])
-    ca_certs = entitlement_cfg.get('caCerts')
+                   'remote-server=%s' % remote_server], capture=True)
+    ca_certs = directives.get('caCerts')
     if ca_certs:
         util.subp(['/snap/bin/canonical-livepatch', 'config',
-                   'ca-certs=%s' % ca_certs])
+                   'ca-certs=%s' % ca_certs], capture=True)
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -72,6 +72,13 @@ class LivepatchEntitlement(base.UAEntitlement):
             print('Installing canonical-livepatch snap...')
             util.subp(['snap', 'install', 'canonical-livepatch'], capture=True)
         entitlement_cfg = self.cfg.entitlements.get(self.name)
+        try:
+            process_directives(entitlement_cfg)
+        except util.ProcessExecutionError as e:
+            msg = 'Unable to configure Livepatch: '
+            print(msg)
+            logging.error(msg)
+            return False
         livepatch_token = entitlement_cfg.get('resourceToken')
         if not livepatch_token:
             logging.debug(
@@ -126,3 +133,18 @@ class LivepatchEntitlement(base.UAEntitlement):
             logging.debug('Livepatch not enabled. %s', str(e))
             operational_status = (status.INACTIVE, str(e))
         return operational_status
+
+
+def process_directives(entitlement_cfg):
+    """Process livepatch configuration directives.
+
+    @raises: ProcessExecutionError if unable to configure livepatch.
+    """
+    remote_server = entitlement_cfg.get('remoteServer')
+    if remote_server:
+        util.subp(['/snap/bin/canonical-livepatch', 'config',
+                   'remote-server=%s' % remote_server])
+    ca_certs = entitlement_cfg.get('caCerts')
+    if ca_certs:
+        util.subp(['/snap/bin/canonical-livepatch', 'config',
+                   'ca-certs=%s' % ca_certs])
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -75,7 +75,7 @@ class LivepatchEntitlement(base.UAEntitlement):
         try:
             process_directives(entitlement_cfg)
         except util.ProcessExecutionError as e:
-            msg = 'Unable to configure Livepatch: '
+            msg = 'Unable to configure Livepatch: ' + str(e)
             print(msg)
             logging.error(msg)
             return False
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -59,12 +59,11 @@ class LivepatchEntitlement(base.UAEntitlement):
         'Canonical Livepatch Service'
         ' (https://www.ubuntu.com/server/livepatch)')
 
-    def __init__(self, cfg=None):
-        super().__init__(cfg)
-        # Set static_affordances in init so we can mock is_container in tests
-        self.static_affordances = (
-            ('Cannot install Livepatch on a container', util.is_container,
-             False),)
+    # Use a lamda so we can mock util.is_container in tests
+    static_affordances = (
+        ('Cannot install Livepatch on a container',
+         lambda: util.is_container(),
+         False),)
 
     def enable(self):
         """Enable specific entitlement.
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -58,8 +58,12 @@ class LivepatchEntitlement(base.UAEntitlement):
     description = (
         'Canonical Livepatch Service'
         ' (https://www.ubuntu.com/server/livepatch)')
-    static_affordances = (
-        ('Cannot install Livepatch on a container', util.is_container, False),)
+
+    def __init__(self, cfg=None):
+        super().__init__(cfg)
+        self.static_affordances = (
+            ('Cannot install Livepatch on a container', util.is_container,
+             False),)
 
     def enable(self):
         """Enable specific entitlement.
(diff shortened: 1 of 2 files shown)


uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -0,0 +1,84 @@
...
+from uaclient.testing.helpers import TestCase
+
+
+class TestLivepatchContractStatus(TestCase):
+
+    def test_contract_status_entitled(self):
+        """The contract_status returns ENTITLED when entitled is True."""
+        tmp_dir = self.tmp_dir()
+        cfg = config.UAConfig(cfg={'data_dir': tmp_dir})
+        machineToken = {
```
OddBloke: This looks like it's the same in every test; could we pull it out to a shared variable somewhere? (If we don't expect it to be mutated by the tests or the code, we could also make it a [MappingProxyType](https://docs.python.org/3/library/types.html#types.MappingProxyType) to ensure that they don't.)

OddBloke:
Oh, I just saw your messages in IRC about this not yet being complete; if you could include such notes in the PR too in future, that'd be handy!

OddBloke:
(Functionally this LGTM.)

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: +1 will address this with a shared var and use deepcopy if we need to manipulate it.

uaclient/entitlements/livepatch.py
```diff
@@ -59,8 +59,12 @@ class LivepatchEntitlement(base.UAEntitlement):
     description = (
         'Canonical Livepatch Service'
         ' (https://www.ubuntu.com/server/livepatch)')
-    static_affordances = (
-        ('Cannot install Livepatch on a container', util.is_container, False),)
+
+    def __init__(self, cfg=None):
+        super().__init__(cfg)
+        self.static_affordances = (
```
OddBloke: Why the move?

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -0,0 +1,220 @@
...
+from nose2.tools.params import params
+
+from uaclient import config
+from uaclient.entitlements.livepatch import (
+    LivepatchEntitlement, process_directives)
+from uaclient import status
+from uaclient.testing.helpers import TestCase
+
+
+LIVEPATCH_MACHINE_TOKEN = {
```
OddBloke: I would still suggest wrapping these with `types.MappingProxyType`, so we can be sure that there isn't any accidental modification happening.

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -0,0 +1,220 @@
...
+            'kernelFlavors': ['generic', 'lowlatency'],
+            'tier': 'stable'
+        }
+    }
+}
+
+M_PATH = 'uaclient.entitlements.livepatch.'  # mock path
+
+
+class TestLivepatchContractStatus(TestCase):
```
OddBloke: We could reduce some of the repetition in several of these test classes by adding a super-class with a `setUp` that does:

```py
tmp_dir = self.tmp_dir()
self.cfg = config.UAConfig(cfg={'data_dir': tmp_dir})
self.cfg.write_cache('machine-token', LIVEPATCH_MACHINE_TOKEN)
self.cfg.write_cache('machine-access-livepatch', LIVEPATCH_RESOURCE_ENTITLED)
```

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -0,0 +1,220 @@
...
+        cfg.write_cache('machine-access-livepatch',
+                        {'entitlement': {'entitled': False}})
+        entitlement = LivepatchEntitlement(cfg)
+        self.assertEqual(status.NONE, entitlement.contract_status())
+
+
+class TestLivepatchProcessDirectives(TestCase):
+
+    @params(('remoteServer', 'remote-server=%s'), ('caCerts', 'ca-certs=%s'))
+    def test_process_directives_call_livepatch_config_command(
```
OddBloke: An additional test that covers both directives being specified at once would be good.

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -0,0 +1,220 @@
...
+        cfg.write_cache('machine-token', LIVEPATCH_MACHINE_TOKEN)
+        cfg.write_cache('machine-access-livepatch',
+                        LIVEPATCH_RESOURCE_ENTITLED)
+        entitlement = LivepatchEntitlement(cfg)
+        with mock.patch('sys.stdout', new_callable=StringIO) as m_stdout:
+            self.assertTrue(entitlement.can_enable())
+        self.assertEqual('', m_stdout.getvalue())
+
+    @mock.patch('uaclient.util.is_container', return_value=True)
+    @mock.patch('os.getuid', return_value=0)
```
OddBloke: It's not particularly important, but as you don't use `m_getuid` in either of these tests, you could avoid it being passed in at all by invoking the decorator like this:

```py
@mock.patch('os.getuid', Mock(return_value=0))
```

Also, given that it's on each test in this class, you could also move it to be a class decorator; it would then apply to any additional tests written.

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -0,0 +1,220 @@
...
+    @mock.patch(M_PATH + 'LivepatchEntitlement.can_enable', return_value=False)
+    def test_enable_false_when_can_enable_false(self, m_can_enable):
+        """When can_enable returns False enable returns False."""
+        tmp_dir = self.tmp_dir()
+        cfg = config.UAConfig(cfg={'data_dir': tmp_dir})
+        cfg.write_cache('machine-token', LIVEPATCH_MACHINE_TOKEN)
+        cfg.write_cache('machine-access-livepatch',
+                        LIVEPATCH_RESOURCE_ENTITLED)
+        entitlement = LivepatchEntitlement(cfg)
+        with mock.patch('sys.stdout', new_callable=StringIO) as m_stdout:
```
OddBloke: Should this test be asserting something about `m_stdout`? (If not, we don't need `as m_stdout` here.)

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -0,0 +1,220 @@
...
+            self.assertTrue(entitlement.enable())
+        assert self.mocks_install + self.mocks_config in m_subp.call_args_list
+        msg = ('Installing canonical-livepatch snap...\n'
+               'Canonical livepatch enabled.\n')
+        assert msg == m_stdout.getvalue()
+
+    @mock.patch('uaclient.util.subp')
+    @mock.patch('uaclient.util.which', return_value='/found/livepatch')
+    @mock.patch(M_PATH + 'LivepatchEntitlement.can_enable', return_value=True)
+    def test_enable_installs_and_configures_livepatch_snap_when_present(
```
OddBloke: This test name is a little off, we're testing that it _doesn't_ install it if it's already present.

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: thanks for this. I wrapped in MappingProxyType. This requires our unit tests to also cast back to dict because we can't json serialize (for config.write_cache) and we can't pickle this type for copy.deepcopy calls in unit tests.

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: Good suggestion added BaseEnabledLivepatchConfig class for that setup

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: ahh right, good point moved to class decorator.

blackboxsw: held off on the mock.Mock for the moment.

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: fixed name and docstr

uaclient/entitlements/livepatch.py
blackboxsw: Moved this into init so that we could mock out util.is_container at init timeframe in unit tests. If it remains a predefined class attr, mocking doesn't help because cls.static_affordances is already initialized to the real util.is_container.

blackboxsw: Adding a comment to that affect in the code

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: +1 added the multi-directive test.

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: Validated m_stdout isn't being saturated w/ print statements.

uaclient/entitlements/livepatch.py
OddBloke: I don't love the fact that `static_affordances` is no longer, well, static.

If we instead change the function in `static_affordances` to a `lambda` which calls `util.is_container` then we can successfully mock (because the `util.is_container` name lookup now happens at call time and not at class definition time):

```diff
diff --git a/uaclient/entitlements/livepatch.py b/uaclient/entitlements/livepatch.py
index b8d86ab..23b2e16 100644
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -58,13 +58,12 @@ class LivepatchEntitlement(base.UAEntitlement):
     description = (
         'Canonical Livepatch Service'
         ' (https://www.ubuntu.com/server/livepatch)')
-
-    def __init__(self, cfg=None):
-        super().__init__(cfg)
-        # Set static_affordances in init so we can mock is_container in tests
-        self.static_affordances = (
-            ('Cannot install Livepatch on a container', util.is_container,
-             False),)
+    static_affordances = (
+        ('Cannot install Livepatch on a container',
+         # Use a lambda so we can mock util.is_container in tests
+         lambda: util.is_container(),
+         False),
+    )
 
     def enable(self):
         """Enable specific entitlement.
```

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -0,0 +1,204 @@
...
+                                           'caCerts': 'value2',
+                                           'ignored': 'ignoredvalue'}}}
+        with mock.patch('uaclient.util.subp') as m_subp:
+            process_directives(cfg)
+        expected_calls = [
+            mock.call(['/snap/bin/canonical-livepatch', 'config',
+                       'remote-server=value1'], capture=True),
+            mock.call(['/snap/bin/canonical-livepatch', 'config',
+                       'ca-certs=value2'], capture=True)]
+        assert expected_calls == m_subp.call_args_list
```
OddBloke: Thanks for adding the test!

Given that the order of these calls doesn't matter, we could make this `assert set(expected_calls) == set(m_subp.call_args_list)` which would make this test a little less sensitive to future changes.

(Or, really, if we're doing that then we should just construct `expected_calls` as a set.)

blackboxsw: I appreciate the suggestion, though I can't cast mock.call to a set  because: 
```
    'ca-certs=value2'], capture=True)])
TypeError: unhashable type: '_Call'
````

Normally I'd agree that the unit test should be flexible regarding ordering (especially when dealing with py2 vs py3  ordering of dict.keys()) but in this case process_directives explicitly processes remoteServer key before caCerts, so the test is correct. 

I'll look around for an alternative for handling the set cast of mock.call. But otherwise I'd be for passing on this suggestion.

uaclient/entitlements/livepatch.py
blackboxsw: Thanks for the Dan, just pushed a change and tested livepatch enable on container and non-container and success.

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: We've decided it's not worth in on this case but will attempt to use the approach for other list types where ordering doesn't matter.

PR #221, opened 2019-03-27
motd: remove motd content

powersj:
This remove MOTD related items from the package. We will not be updating
MOTD directly anymore as we have the status subcommand. We can re-add
this if needed, otherwise this will be handled by update-notifier and
other scripts.


powersj:
Do I need to keep the cache directory if we get rid of all the MOTD stuff? namely `/var/cache/ubuntu-advantage-tools`?

PR #225, opened 2019-03-27
cli: proper error message when no subcommand given

### LINKED ISSUE #182: UA client with no args show show full help usage
This output is not helpful:

$ ua
usage: ubuntu-advantage-client [command] [flags]
ubuntu-advantage-client: error: the following arguments are required:

powersj:
Fixes #182


uaclient/cli.py
```diff
@@ -364,7 +364,12 @@ def main(sys_argv=None):
     if not sys_argv:
         sys_argv = sys.argv
     parser = get_parser()
-    args = parser.parse_args(args=sys_argv[1:])
+    cli_arguments = sys_argv[1:]
+    if not cli_arguments:
+        parser.print_usage()
```
blackboxsw: What do you think about parser.print_help()
print('Try \'ubuntu-advantage\' status\' for more information.')


?

powersj: Well my fear is if they haven't attached yet, so then they get an error message saying go see attach. That feels like we are bouncing between error messages.

blackboxsw: +1.

PR #229, opened 2019-03-29
integrate apt esm hook

julian-klode:
This integrates the hook for apt into the package. What's missing is the whole gettext shenanigans, and maybe you want to do things slightly differently in places, but this feels like a good start.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 
 build:
-	@echo Nothing to be done for build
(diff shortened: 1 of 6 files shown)


julian-klode:
> Also I think we need to adapt setup.py to add more data_files for your apt hook config and the binary built for the apt-esm-hook right?

They are installed by the makefile, so I don't think so

PR #233, opened 2019-03-29
Replace nose2 with pytest

OddBloke:


OddBloke:
I've opened this as a draft pull request because I want to be absolutely certain that we're all happy with this change before it lands.

uaclient/conftest.py
```diff
@@ -0,0 +1,41 @@
+import io
+import logging
+
+import pytest
+
+
+@pytest.fixture
+def caplog_text(request):
```
OddBloke: This function works around missing functionality in the versions of pytest in trusty and xenial; I recommend skipping over it on an initial review and coming back to it once you see how it's used.

(Note that in xenial we will be able to replace this with `pytest-catchlog` which is packaged in Ubuntu as `python3-pytest-catchlog`; that project was integrated in to pytest between xenial and bionic.)

uaclient/conftest.py
```diff
@@ -0,0 +1,41 @@
+import io
```
OddBloke: N.B. https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -43,59 +43,64 @@
 M_PATH = 'uaclient.entitlements.livepatch.'  # mock path
 
 
-class BaseEnabledLivepatchConfig(TestCase):
-    """Base class which sets up an enabled livepatch entitlement."""
+@pytest.fixture
+def livepatch_entitlement(tmpdir):
```
OddBloke: This is a pytest "fixture". If a test includes its name as a parameter, then it will be injected in to that test as that parameter. (You can see this fixture being used in most of the tests below.)

This allows for more modular setup/teardown than provided by unittest (or, by extension, nose2), because test functions can rely on _multiple_ fixtures.

See https://docs.pytest.org/en/latest/fixture.html for more information about pytest fixtures.

OddBloke: (You can see that this fixture itself depends on another fixture, the built-in `tmpdir` fixture.)

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -43,59 +43,64 @@
...
+    """
+    cfg = config.UAConfig(cfg={'data_dir': tmpdir.strpath})
+    cfg.write_cache('machine-token', dict(LIVEPATCH_MACHINE_TOKEN))
+    cfg.write_cache('machine-access-livepatch',
+                    dict(LIVEPATCH_RESOURCE_ENTITLED))
+    return LivepatchEntitlement(cfg)
 
 
-class TestLivepatchContractStatus(BaseEnabledLivepatchConfig):
+class TestLivepatchContractStatus:
```
OddBloke: `pytest` generally leaves unittest.TestCase derived classes alone, as the interactions between `pytest` and `unittest` could cause complex issues.

Note that this means that all the assertion methods on `self` aren't available; pytest instruments the `assert` statement to provide more useful output than most of those assertion methods anyway, so they generally aren't missed.

(This particular class doesn't use any `pytest` functionality that requires us to drop the `TestCase` inheritance, but as I was removing `BaseEnabledLivepatchConfig` I was modifying its definition anyway, so I dropped the inheritance regardless.)

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -43,59 +43,64 @@
...
+        livepatch_entitlement.cfg.write_cache(
+            'machine-access-livepatch', {'entitlement': {'entitled': False}})
+        assert status.NONE == livepatch_entitlement.contract_status()
 
 
-class TestLivepatchProcessDirectives(TestCase):
+class TestLivepatchProcessDirectives:
 
-    @params(('remoteServer', 'remote-server=%s'), ('caCerts', 'ca-certs=%s'))
+    @pytest.mark.parametrize('directive_key,livepatch_param_tmpl', (
```
OddBloke: pytest allows us to generically "mark" tests with flags. These can be freeform for our own use, but there are a few flags that have a built-in meaning (e.g. @pytest.mark.skip will skip the decorated test), including `parametrize`, which specifies that this test should be run multiple times with the given parameters.

See https://docs.pytest.org/en/latest/mark.html for some pointers to the built-in marks, and https://docs.pytest.org/en/latest/example/markers.html#mark-examples for some information about (and examples of) custom marks.

uaclient/tests/test_config.py
```diff
@@ -1,202 +1,195 @@
...
 
-class TestWriteCache(TestCase):
+class TestWriteCache:
 
-    @params(('unknownkey', 'content1'), ('another-one', 'content2'))
+    @pytest.mark.parametrize('key,content', (
+        ('unknownkey', 'content1'), ('another-one', 'content2')))
     def test_write_cache_write_key_name_in_data_dir_when_data_path_absent(
-            self, key, content):
+            self, tmpdir, key, content):
```
OddBloke: This test is a useful example of the `tmpdir` fixture; it doesn't inject a string containing a path, it injects a `py.path` object. These have a rich set of methods that allow for more expressive tests.

In this test, for example:
* we use `tmpdir.strpath` to get a string containing the path to the temporary directory
* we use `tmpdir.join` to produce a new `py.path` object for a sub-path of our temporary directory
* we use `.check()` on that constructed object to check whether or not it exists

OddBloke:
(Moved out of draft because Travis doesn't run on draft PRs; otherwise please consider this still a draft.)

uaclient/tests/test_config.py
blackboxsw: Excellent inline comments here for context. Thanks for the good examples and doc references

blackboxsw:
Land pending successful CI build on travis

PR #248, opened 2019-04-01
test_cc: expand test to mock apt-transport-https/ca-certificates 

OddBloke:
Currently it doesn't mock os.path.exists, so the behaviour of the
function under test depends on whether these packages are installed in
the test environment (which they aren't during package build but
generally are on developer systems).

### DIFF
--- a/uaclient/entitlements/tests/test_cc.py
+++ b/uaclient/entitlements/tests/test_cc.py
@@ -6,7 +6,6 @@ from io import StringIO
 from uaclient import config
 from uaclient import status
 from uaclient.entitlements.cc import CommonCriteriaEntitlement
-from uaclient.testing.helpers import TestCase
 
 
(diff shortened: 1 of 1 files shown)


uaclient/entitlements/tests/test_cc.py
```diff
@@ -58,34 +61,64 @@ def test_can_enable_true_on_entitlement_inactive(
         assert '' == m_stdout.getvalue()
 
 
-class TestCommonCriteriaEntitlementEnable(TestCase):
+class TestCommonCriteriaEntitlementEnable:
 
+    # Paramterize True/False for apt_transport_https and ca_certificates
```
blackboxsw: minor spelling nit Parametrize

blackboxsw: will sort that comment in next branch.

PR #251, opened 2019-04-02
apt: pinning to use hardcoded origin for each entitlement

### LINKED ISSUE #238: fips: invalid pin origin on staging fips for repo urls that do not contain private-ppa.launchpad.net
The staging repositories for fips are fips.staging.canonical.com instead of the production urls //private-ppa.launchpad.net/ubuntu-advantage/fips.

uaclient/apt makes an assumption on the origin of the PPAs based on replacing private-ppa.launchpad.net with LP-PPA- and hyphenating full repoi url path.   This results in a broken ppa file for staging For fips ppas origin like the following:

cat /etc/apt/preferences.d/ubuntu-fips-xenial 
Package: *
Pin: release o=fips.staging.ubuntu.com-, n=xenial
Pin-Priority: 1001

Origin in this case should be the Origin field from the PPA, which is present in /var/lib/apt/lists/fips.staging.ubuntu.com_ubuntu_dists_xenial_Release
 Origin: UbuntuFIPS

blackboxsw:
add_apt_pinning incorrectly assumed that FIPS repositories would live behind a private-ppa.launchpad.net/ url and would incorrectly calculate origin based on the expected LP_PPA origin line.

Staging and production FIPS/ESM/CC services now live behind a separate haproxy service that is not hosted as a Launchpad private-ppa. So, aptURLs related to those services live at something like esm.staging.canonical.com/fips instead of private-ppa.launchpad.net.

As a result, the entitlements should reference a expected PPA Origin which will be common across either production and staging services instead of calculating that origin.

Fixes #238

### DIFF
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -12,7 +12,7 @@ class RepoEntitlement(base.UAEntitlement):
 
     repo_list_file_tmpl = '/etc/apt/sources.list.d/ubuntu-{name}-{series}.list'
     repo_pref_file_tmpl = '/etc/apt/preferences.d/ubuntu-{name}-{series}'
-    origin = 'UNSET'   # The repo Origin value for setting pinning
+    origin = None   # The repo Origin value for setting pinning
 
     repo_url = 'UNSET'
     repo_key_file = 'UNSET'  # keyfile delivered by ubuntu-cloudimage-keyring
@@ -55,6 +55,14 @@ class RepoEntitlement(base.UAEntitlement):
             logging.error(str(e))
             return False
         if self.repo_pin_priority:
+            if not self.origin:
+                logging.error(
+                    "Cannot setup apt pin. Empty apt repo origin value '%s'." %
+                    self.origin)
+                logging.error(
+                    status.MESSAGE_ENABLED_FAILED_TMPL.format(
+                        title=self.title))
+                return False
             repo_pref_file = self.repo_pref_file_tmpl.format(
                 name=self.name, series=series)
             apt.add_ppa_pinning(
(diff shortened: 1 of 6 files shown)


uaclient/entitlements/repo.py
```diff
@@ -12,6 +12,7 @@ class RepoEntitlement(base.UAEntitlement):
 
     repo_list_file_tmpl = '/etc/apt/sources.list.d/ubuntu-{name}-{series}.list'
     repo_pref_file_tmpl = '/etc/apt/preferences.d/ubuntu-{name}-{series}'
+    origin = 'UNSET'   # The repo Origin value for setting pinning
```
OddBloke: `None` would be a more normal value to indicate that something is unset; why the string?

blackboxsw: It's a good point, None would still be cast to 'None' in a print string. I just wanted something that would be a bit more visible/understandable given that RepoEntitlement.origin is written to user-facing files like /etc/apt/preferences.d/<some_pin_file>.

blackboxsw: That said; 'None' is still understandable from a config file, so I'm ok changing this to None. Thoughts

OddBloke: If we write either `o=UNSET` or `o=None` to a config file, I would expect that config file to be useless. Neither of those origins are ever going to exist, so the pinning will never apply. Am I correct?

If so, I think we should ensure that we don't write out a config file if origin isn't something meaningful, and it's easier/clearer to check for None in the code that writes the file than it is 'UNSET'.

blackboxsw: +1 pushed changes to avoid emitting an o=None value. All entitlements have this origin hardcoded so we should be able to get here but unit tests are added for this logic.

Issue #261, opened 2019-04-03
fips, cc, livepatch: honor architectures affordance from ua-contracts entitlement response


blackboxsw:
ua-contracts could provide an affordance 'architectures' as a list

fips entitment should present:
 architechtures: ['x86_64', 'ppc64le', 's390x']

cc should be:
 architechtures: ['x86_64', 'ppc64le', 's390x']

livepatch should be:
 architechtures: ['x86_64']

PR #270, opened 2019-04-04
entitlements: honor architectures affordance from ua-contracts

blackboxsw:
Also refactor check_affordances to limit white-space indents.
 
Fixes: #261


blackboxsw:
To test:

$sudo UA_CONTRACT_URL=https://contracts.staging.canonical.com ua attach

# edit /var/lib/ubuntu-advantage/machine-access-livepatch.json and replace x86_64 with x86_65
$ sudo ua enable livepatch
Livepatch is not available for platform x86_64.
Supported platforms are: x86_65

Issue #271, opened 2019-04-04
missing python3-package-resources dep on trusty build


panlinux:
The trusty deb is missing a dependency on `python3-package-resources`:
```ubuntu@trusty-ua:~$ sudo apt install ubuntu-advantage-tools 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following extra packages will be installed:
  libsodium18 python3-libnacl python3-pymacaroons python3-six python3-yaml
The following NEW packages will be installed:
  libsodium18 python3-libnacl python3-pymacaroons python3-six python3-yaml
The following packages will be upgraded:
  ubuntu-advantage-tools
1 upgraded, 5 newly installed, 0 to remove and 7 not upgraded.
...
Setting up ubuntu-advantage-tools (19.1-0ubuntu1~365~git.7f5e517~ubuntu14.04.1) ...
Processing triggers for libc-bin (2.19-0ubuntu6.15) ...
ubuntu@trusty-ua:~$ ua
Traceback (most recent call last):
  File "/usr/bin/ua", line 5, in <module>
    from pkg_resources import load_entry_point
ImportError: No module named 'pkg_resources'
```
Specifically:
```
ubuntu@trusty-ua:~$ dpkg -s ubuntu-advantage-tools |grep Depends
Depends: python3:any (>= 3.3.2-2~), python3, python3-pymacaroons, python3-yaml, libapt-pkg4.12 (>= 0.8.0), libc6 (>= 2.4), libgcc1 (>= 1:4.1.1), libstdc++6 (>= 4.4.0)
```
It's from our daily build ppa, and I have proposed enabled:
```
ubuntu@trusty-ua:~$ apt-cache policy ubuntu-advantage-tools 
ubuntu-advantage-tools:
  Installed: 19.1-0ubuntu1~365~git.7f5e517~ubuntu14.04.1
  Candidate: 19.1-0ubuntu1~365~git.7f5e517~ubuntu14.04.1
  Version table:
 *** 19.1-0ubuntu1~365~git.7f5e517~ubuntu14.04.1 0
        500 http://ppa.launchpad.net/canonical-server/ua-client-daily/ubuntu/ trusty/main amd64 Packages
        100 /var/lib/dpkg/status
     10ubuntu0.14.04.2 0
        500 http://br.archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
```

panlinux:
[...]

PR #272, opened 2019-04-04
Depend on python3-pkg-resources

panlinux:
pkg_resources is imported when the /usr/bin/ua script is run.

This fixes issue #271


powersj:
This looks like the right answer per other Debian bugs:

- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=839075
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=789822
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=789822

Issue #276, opened 2019-04-05
/etc/update-motd.d/99-esm hardcodes precise, expects old ua tool


panlinux:
We may have a task somewhere to handle motd, but at the moment at least the `/etc/update-motd.d/99-esm` file is still the old one from the bash version of ua and hardcodes precise.

Not only that, it uses syntax from the old ua tool: `if ubuntu-advantage is-esm-enabled; then`

Which fails:
```
root@trusty-ua:~# /etc/update-motd.d/99-esm
usage: ubuntu-advantage [command] [flags]
ubuntu-advantage: error: argument : invalid choice: 'is-esm-enabled' (choose from 'status', 'attach', 'detach', 'enable', 'disable', 'version')
This Ubuntu 14.04.6 LTS system is past its End of Life, and is no longer
receiving security updates.  To protect the integrity of this system, it’s
critical that you enable Extended Security Maintenance updates:
 * https://www.ubuntu.com/esm
```

panlinux:
This file is not part of the new client, so it should be removed in postinst by a maintainer script.

panlinux:
Files and directories that can be removed:
- `/etc/update-motd.d/99-esm`: we are not doing that here anymore?
- `/etc/update-motd.d/80-livepatch`: we are not doing that here anyore?
- `/var/cache/ubuntu-advantage-tools`: was used by the livepatch motd, present in bionic and later
- `/etc/cron.daily/ubuntu-advantage-tools`: was used to generate a `ua status` output daily, to be consumed by the MOTD scripts

powersj:
marking closed

PR #278, opened 2019-04-05
Remove obsolete files on upgrade with rm conf

panlinux:
This takes care of removing files we don't use anymore, and that are not automatically removed by dpkg itself.

In this branch I'm treating all files under `/etc/*` as conf files, and using `dpkg-maintscript-helper`'s `rm_conf`. I have another branch, remove-obsolete-files-on-upgrade, where I was just doing all that manually.

I'm not sure these files we are removing should be considered config files: it's a cron job, and a bunch of motd scripts. The drawback is that, if the user changed any of these, dpkg will leave backup files lying around. But looks like all tools that parse `update-motd.d/` and `cron.daily/` are aware of that and won't use the dpkg-renamed scripts.

This fixes issue #276


panlinux:
I'm about to push a more condensed set of commits, stand by

panlinux:
pushed

PR #293, opened 2019-04-07
Clean up dev directory

### LINKED ISSUE #255: Clean up dev directory
Now that we have a staging service that is live we should be using that for development. Can we please clean out any and all scripts that are no longer used from the directory?

Mark for possible removal:

- contracts-controller.patch
- contracts-go.patch
- contracts-implemented-controller.patch
- contracts-schema.patch
- demo-contract-service
- entitlement-creds.json (?)
- install-contracts-server
- pkg-deps.json
- run-uaclient
- runserver.sh
- uaclient

That would leave the following, which are there to help with development and packaging:

- bddeb
- changelog.in
- constraints-bionic.txt
- constraints-trusty.txt
- constraints-xenial.txt
- control.in
- make-tarball
- pkg-deps.json
- read-dependencies
- README.txt

Final changes:

- I would like README.txt renamed to README.md
- If we do not remove entitlement-creds.json, then we should update it for our package and not re-use cloud-init data as-is.

powersj:
This breaks the dev directory into 'tools' and 'demo'. Where the tools
directory has building and test related files and the demo directory has
files related to demoing and development of the package itself.

Python files in tools were all made pylint clean and verified to be
working correctly.

This also updates the related Makefile, tox.ini, and README.md files
with the changes in directories.

Fixes #255


powersj:
Patch applied and pushed

tools/read-dependencies
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""List pip dependencies or system package dependencies for cloud-init."""
```
blackboxsw: hah, copy paste fail anyone?

PR #297, opened 2019-04-07
keyrings: remove entire keyrings directory

### LINKED ISSUE #256: Remove keyrings directory
Do we need the keyrings directory? I think this comes from the contract service details

powersj:
Keyrings no longer need to be shipped with the package, therefore we
should remove them from the source code of the UA client.

Fixes #256


powersj:
ok removed all usage of keyrings from setup.py and the various services files. We will rely entirely on the fingerprint and if we are unable to get it fail.

uaclient/entitlements/cc.py
```diff
@@ -32,14 +29,13 @@ def disable(self, silent=False, force=False):
...
             name=self.name, series=series)
-        keyring_file = os.path.join(apt.APT_KEYS_DIR, self.repo_key_file)
         entitlement_cfg = self.cfg.read_cache(
             'machine-access-%s' % self.name)['entitlement']
         access_directives = entitlement_cfg.get('directives', {})
         repo_url = access_directives.get('aptURL', self.repo_url)
         if not repo_url:
             repo_url = self.repo_url
-        apt.remove_auth_apt_repo(repo_filename, repo_url, keyring_file)
+        apt.remove_auth_apt_repo(repo_filename, repo_url)
```
panlinux: Is this lacking a `fingerprint` arg now? It's defined as `remove_auth_apt_repo(repo_filename, repo_url, fingerprint)`

panlinux: Indeed, this needs fixing in all such cases, not just cc:
```
ubuntu@trusty-ua:~/git$ sudo ua disable esm
    load_entry_point('ubuntu-advantage-tools==18.1', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 371, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 162, in action_disable
    if entitlement.disable():
  File "/usr/lib/python3/dist-packages/uaclient/entitlements/esm.py", line 35, in disable
    apt.remove_auth_apt_repo(repo_filename, repo_url)
TypeError: remove_auth_apt_repo() missing 1 required positional argument: 'fingerprint'
ubuntu@trusty-ua:~/git$ 
```

panlinux: I also wonder why tests didn't fail.

We can also break this PR up into two: one to just handle the more pressing case of adding/removing the esm keyring, and the other about removing the superfluous keyrings.

powersj: Tests passed because I correctly updated the mocked calls. When I rebased on master after the apt changes yesterday I accepted what was in master and did not update my calls in each of the entitlements.

setup.py
```diff
@@ -28,7 +28,7 @@ def _get_version():
     ),
     data_files=[
         ('/etc/ubuntu-advantage', ['uaclient.conf']),
-        ('/usr/share/keyrings', glob.glob('keyrings/*')),
+        ('/usr/share/keyrings', glob.glob('debian/keyrings/*')),
```
panlinux: any particular reason for this change? If we ever make this package a non-native package, we will have to revert this, because setup.py would be referencing something that is not part of the tarball anymore.

powersj: My thought process was since the keyring is getting installed at install time and not something the python code, I would put it in the debian dir.

I can pull it out if that makes more sense?

panlinux: It's not incorrect, just not needed. We can leave it as is.

powersj:
After numerous discussions around keyrings and how we want to ship things, I am closing this merge and opening a new one with the much smaller set of changes.

PR #301, opened 2019-04-07
logrotate: add basic config

### LINKED ISSUE #241: Missing logrotate configuration
We should have a logrotate configuration for `/var/log/ubuntu-advantage.log`.

Also, is this file prone to contain secrets? Maybe when `--debug` is used? In that case, we should also change its permissions to `0600` or something like that.

powersj:
Keep 6 months worth of logs, while compressing old ones.

Fixes #241


panlinux:
It's not working and I'm not yet sure why that permissions check is in place, at least on trusty:
```
root@trusty-ua:~# logrotate -f -v /etc/logrotate.d/ubuntu-advantage-tools 
reading config file /etc/logrotate.d/ubuntu-advantage-tools

Handling 1 logs

rotating pattern: /var/log/ubuntu-advantage.log  forced from command line (6 rotations)
empty log files are not rotated, old logs are removed
considering log /var/log/ubuntu-advantage.log
error: skipping "/var/log/ubuntu-advantage.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
root@trusty-ua:~# l /var/log/ubuntu-advantage.log 
-rw-r--r-- 1 root root 390 Apr  8 15:46 /var/log/ubuntu-advantage.log
root@trusty-ua:~# cat /etc/logrotate.d/ubuntu-advantage-tools 
/var/log/ubuntu-advantage.log {
    rotate 6
    monthly
    compress
    delaycompress
    missingok
	notifempty
}
```

panlinux:
Oh, my mistake I think. I called logrotate with just our config file, but it works if I call it with `/etc/logrotate.conf`:
```
root@trusty-ua:~# logrotate -f /etc/logrotate.conf 
root@trusty-ua:~# l /var/log/ubuntu-advantage.log*
-rw-r--r-- 1 root root   0 Apr  8 15:55 /var/log/ubuntu-advantage.log
-rw-r--r-- 1 root root 390 Apr  8 15:46 /var/log/ubuntu-advantage.log.1
```

debian/ubuntu-advantage-tools.logrotate
```diff
@@ -0,0 +1,8 @@
+/var/log/ubuntu-advantage.log {
+    rotate 6
+    monthly
+    compress
+    delaycompress
+    missingok
+	notifempty
```
panlinux: extra indent?

powersj: yep, fixed in latest

PR #302, opened 2019-04-07
Add bash-completion to install for 'ua'

### LINKED ISSUE #165: Add bash completion scripts
This would be a nice-to-have for users.

powersj:
Fix #165

### DIFF
--- a/tools/ua.bash
+++ b/tools/ua.bash
@@ -1,0 +1,33 @@
+# bash completion for ubuntu-advantage-tools
+
+_ua_complete()
+{
+    local cur_word prev_word
+    cur_word="${COMP_WORDS[COMP_CWORD]}"
+    prev_word="${COMP_WORDS[COMP_CWORD-1]}"
+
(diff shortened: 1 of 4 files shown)


tools/ua.bash
```diff
@@ -0,0 +1,33 @@
+# bash completion for ubuntu-advantage-tools
+
+_ua_complete()
+{
+    local cur_word prev_word
+    cur_word="${COMP_WORDS[COMP_CWORD]}"
+    prev_word="${COMP_WORDS[COMP_CWORD-1]}"
+
+    subcmds="attach detach disable enable status version"
```
blackboxsw: We could dynamically source the optional flags and subcommands with the following: 

subcmds=$(ua --help | awk '/^\s*$|Available|Use/ {next;} /Flags:/{flag=1;next}/Use ubuntu-avantage/{flag=0}flag{ if ( $1 ~ /,/ ) { print $2}  else print $1}')


This only tab-completes the first element, but it's a lot better than nothing.

PR #303, opened 2019-04-07
d/control: update standards to 4.3.0

### LINKED ISSUE #279: lintian: ancient-standards-version
W: ubuntu-advantage-tools source: ancient-standards-version 3.9.6 (released 2014-09-17) (current is 4.3.0)

powersj:
Fix #279


powersj:
I went through the checklist and I did not find anything that stood out for the project. There were comments about reproducible builds, which would be interesting to verify for the project, otherwise I think we are good to move up.

PR #305, opened 2019-04-07
util: remove decode binary

### LINKED ISSUE #179: util.decode_binary can (probably) be removed
As we're now Python 3 only, we should know unambiguously whether we're getting text or bytes back from whatever we're calling; `util.decode_binary` can probably be removed.

powersj:
As the client is Python 3 only we should know what we are dealing with
in terms of bytes of strings.

Fixes #179


blackboxsw:
This breaks on json.loads handling of url responses from ua-contracts. will have to sort what specifically falls over on 'ua attach'

Issue #275, opened 2019-04-05
esm enabled, status shows disabled


panlinux:
On trusty, when enabling esm, status will still show it's disabled afterwards. This is because the pinning of -32768 is still there, and that is caused by an incorrect url in the auth.conf file.

In trusty, only a bare hostname entry in auth.conf will be considered for changing the pinning. That, plus how it looks up the entries in auth.conf and decides which one to use, means the file needs to have the following careful structure:
- for ESM purposes, we need one entry with a bare hostname but ending in a slash:
```
machine esm.staging.ubuntu.com/ login bearer password the-password
```
- for other services that share the same hostname but differ in the path, they must come *before*, i.e., more specific entries need to come first. So for example:
```
machine esm.staging.ubuntu.com/otherservice/ login bearer password another-password
machine esm.staging.ubuntu.com/ login bearer password a-password
```
Right now the passwords are the same, but this could change in the future, and the above rules make sure apt will select the right entry.

UA currently creates an entry like this:
```
machine esm.staging.ubuntu.com/ubuntu/ login bearer password a-password
```
That will leave the pinning at -32768 and `ua status` will claim that esm is disabled.

So to recap:
- more specific entries at the top
- for esm purposes, there needs to be one entry with just the hostname, but ending with a trailing slash, at the bottom

PR #307, opened 2019-04-08
apt: fix apt auth format, insert specific repo subpaths before base path

blackboxsw:
- separate function add_apt_auth_conf_entry from add_auth_apt_repo
- Drop trailing ubuntu/ from repo apt auth conf format to avoid magic
  pin -32768 due to invalid auth config for esm
- Parse apt auth file repo urls:
    - If our repo_url is a specific subpath of an
      existing repo url, insert our repo url before the matched
      base url entry
    - Otherwise, append to the end of the auth entries

Fixes: #275

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -1,10 +1,12 @@
 import glob
 import logging
 import os
+import re
 import shutil
 
 from uaclient import util
 
+APT_AUTH_COMMENT = '  # ubuntu-advantage-tools'
 APT_CONFIG_AUTH_FILE = 'Dir::Etc::netrc/'
 APT_CONFIG_AUTH_PARTS_DIR = 'Dir::Etc::netrcparts/'
 APT_CONFIG_LISTS_DIR = 'Dir::State::lists/'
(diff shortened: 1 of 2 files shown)


uaclient/apt.py
```diff
@@ -38,16 +35,16 @@ def valid_apt_credentials(repo_url, username, password):
         return True   # Do not validate
     try:
         util.subp(['/usr/lib/apt/apt-helper', 'download-file',
-                   '%s://%s:%s@%s/ubuntu/pool/' % (
-                       protocol, username, password, repo_path),
+                   '%s://%s@%s/ubuntu/dists/%s/Release' % (
+                       protocol, credentials, repo_path, series),
                    '/tmp/uaclient-apt-test'],
```
powersj: The ESM repo will have 'dists' open to everyone, which is why I thought you had gone with pool originally. Is this handled for ESM somewhere else?

blackboxsw: wow, invalid merge conflict resolution by  me against master. This should not have leaked into this branch when I rebased

Issue #329, opened 2019-04-09
uaclient will only add *-updates apt sources list if the system already has *-updates  enabled apt policy


blackboxsw:
ua-contracts server reports supported suites for  apt entitlements:

They look something like this for esm:
"suites": ["precise", "trusty-security"]}

ua client should only install the suites that apply to the running series

Issue #315, opened 2019-04-08
Final ESM pockets and components


powersj:
Place holder bug for finalizing what the ESM pockets and components will be.

PR #330, opened 2019-04-09
apt: obtain supported suite directives for apt entitlements

blackboxsw:
ua-contracts service provides a 'suites' directive which can
be a list or None and describes the supported apt source suites
that can be enabled for a given entitlement.

Also adds default trusty-security and trusty-updates to package postinst.

Enable any supported suites that match the current running series

Fixes: #329
Fixes: #315

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -7,11 +7,12 @@ ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
 configure_esm_source() {
     if [ ! -e "$ESM_APT_SOURCE_FILE" ]; then
         cat > $ESM_APT_SOURCE_FILE <<EOF
-# Written by ubuntu-advantage-tools during package postinst
-deb https://esm.ubuntu.com/ubuntu trusty security
-deb https://esm.ubuntu.com/ubuntu trusty updates
-# deb-src https://esm.ubuntu.com/ubuntu security
-# deb-src https://esm.ubuntu.com/ubuntu updates
+# Written by ubuntu-advantage-tools
+deb https://esm.ubuntu.com/ubuntu trusty-security main
+# deb-src https://esm.ubuntu.com/ubuntu trusty-security main
+
+deb https://esm.ubuntu.com/ubuntu trusty-updates main
+# deb-src https://esm.ubuntu.com/ubuntu trusty-updates main
 EOF
     fi
 }
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -48,8 +48,8 @@ def valid_apt_credentials(repo_url, username, password):
     return False
 
 
-def add_auth_apt_repo(repo_filename, repo_url, credentials, keyring_file=None,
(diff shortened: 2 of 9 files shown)


uaclient/apt.py
```diff
@@ -66,12 +66,25 @@ def add_auth_apt_repo(repo_filename, repo_url, credentials, keyring_file=None,
...
+
     logging.info('Enabling authenticated repo: %s', repo_url)
     content = ''
-    for pocket in pockets:
-        content += ('deb {url}/ubuntu {series} {pocket}\n'
-                    '# deb-src {url}/ubuntu {series} {pocket}\n'.format(
-                        url=repo_url, series=series, pocket=pocket))
+    if not suites:
+        logging.debug('No apt source suite provided, assuming "%s".', series)
+        suites = [series]  # The default suite should be the series name
```
powersj: I debate whether suites should always be required, as we wouldn't ever have a xenial for esm

powersj:
and one test error to fix:

```
>       assert apt_cmds == m_subp.call_args_list
E       AssertionError: assert [call(['apt-k...capture=True)] == [call(['apt-ca...capture=True)]
E         At index 0 diff: call(['apt-key', 'adv', '--keyserver', 'keyserver.ubuntu.com', '--recv-keys', 'APTKEY'], capture=True) != call(['apt-cache', 'policy'])
E         Right contains more items, first extra item: call(['apt-key', 'adv', '--keyserver', 'keyserver.ubuntu.com', '--recv-keys', 'APTKEY'], capture=True)
E         Use -v to get the full diff
uaclient/tests/test_apt.py:161: AssertionError
```

uaclient/apt.py
blackboxsw: Sure let's complain loudly. I'm seeing  suites for esm, cc and fips defined by ua-contracts, so I'm good with that

debian/postinst
```diff
@@ -8,10 +8,11 @@ configure_esm_source() {
     if [ ! -e "$ESM_APT_SOURCE_FILE" ]; then
         cat > $ESM_APT_SOURCE_FILE <<EOF
 # Written by ubuntu-advantage-tools during package postinst
```
powersj: drop the "during package postinst"

debian/postinst
```diff
@@ -8,10 +8,11 @@ configure_esm_source() {
...
-deb https://esm.ubuntu.com/ubuntu trusty security
-deb https://esm.ubuntu.com/ubuntu trusty updates
-# deb-src https://esm.ubuntu.com/ubuntu security
-# deb-src https://esm.ubuntu.com/ubuntu updates
+deb https://esm.ubuntu.com/ubuntu trusty-security main
+# deb-src https://esm.ubuntu.com/ubuntu trusty-security main
+
+deb https://esm.ubuntu.com/ubuntu trusty-updates
```
powersj: missing main at the end

debian/postinst
blackboxsw: done.

Issue #343, opened 2019-04-10
Don't enable *-updates suites if the server isn't opted in to trusty-updates


blackboxsw:

powersj:
As discussed in the cross-team sync, we will not enable ESM trusty-updates if someone has disabled trusty-updates in their sources.list

PR #345, opened 2019-04-10
apt: ignore *-updates suite if system not configured for series-updates

blackboxsw:
ua-contracts sends a 'suites' directive for each entitlement which is a
list of potential suites to enable. For esm, those suites are:
 suites: ['precise', 'trusty-security', 'trusty-updates']

The client will only enable any '*-updates' apt source, if the machine
has already opted into <series>-updates via an apt-cache policy check.

Fixes: #343

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -67,7 +67,9 @@ def add_auth_apt_repo(repo_filename, repo_url, credentials, suites,
         raise InvalidAPTCredentialsError(
             'Invalid APT credentials provided for %s' % repo_url)
 
-    updates_enabled = True   # TODO determine from apt-cache policy
+    # Does this system have updates suite enabled?
+    policy, _err = util.subp(['apt-cache', 'policy'])
+    updates_enabled = bool(' %s-updates/' % series in policy)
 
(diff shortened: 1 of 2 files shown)


uaclient/apt.py
```diff
@@ -67,7 +67,9 @@ def add_auth_apt_repo(repo_filename, repo_url, credentials, suites,
         raise InvalidAPTCredentialsError(
             'Invalid APT credentials provided for %s' % repo_url)
 
-    updates_enabled = True   # TODO determine from apt-cache policy
+    # Does this system have updates suite enabled?
+    policy, _err = util.subp(['apt-cache', 'policy'])
+    updates_enabled = bool(' %s-updates/' % series in policy)
```
blackboxsw: Question I have here is should we also consider  <series>-proposed as a valid state where we should setup esm trusty-updates suite?

panlinux: You mean if the server ever sends down a %s-proposed line? No.

powersj: no the server will never send proposed and we should never be adding proposed

Issue #342, opened 2019-04-10
ua enable should refresh ua-contracts service to update machine-access credential details


blackboxsw:

powersj:
Agreed we would:

* Add ua refresh subcommand to reget the contract
* Refresh contract at each ua enable

PR #346, opened 2019-04-10
refresh: add subcommand to refresh all contract details

blackboxsw:
- create separate function contract.request_updated_contract since this
logic is common to both attach, enable and refresh
- Add new UAContractClient.request_machine_token_refresh which use existing machine_token to request updated machine-token
 - call request_updated_contract prior to enable operation
 - call request_updated_contract in attach and refresh operations 

Fixes: #342

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -316,9 +316,11 @@ def action_refresh(args, cfg):
         print(ua_status.MESSAGE_NONROOT_USER)
         return 1
     if contract.request_updated_contract(cfg):
-        print('Refreshed Ubuntu Advantage contracts.')
-        logging.debug('Refreshed Ubuntu Advantage contracts.')
+        print(ua_status.MESSAGE_REFRESH_SUCCESS)
+        logging.debug(ua_status.MESSAGE_REFRESH_SUCCESS)
         return 0
+    print(ua_status.MESSAGE_REFRESH_FAILURE)
+    logging.debug(ua_status.MESSAGE_REFRESH_FAILURE)
     return 1
 
 
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -315,10 +315,10 @@ def action_refresh(args, cfg):
     if not cfg.is_attached:
         print(ua_status.MESSAGE_UNATTACHED)
         return 1
-    contract_token = cfg.read_cache('contract-token')
-    if contract_token:
-        contract_token = contract_token['contractToken']
-    if contract.request_contract_updates(cfg, contract_token=contract_token):
+    if os.getuid() != 0:
+        print(ua_status.MESSAGE_NONROOT_USER)
+        return 1
+    if contract.request_contract_updates(cfg):
         print('Refreshed Ubuntu Advantage contracts.')
         logging.debug('Refreshed Ubuntu Advantage contracts.')
         return 0
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -18,7 +18,6 @@ import json
 import logging
 import os
 import sys
-import textwrap
 
 from uaclient import config
 from uaclient import contract
@@ -178,6 +177,9 @@ def action_enable(args, cfg):
 
     @return: 0 on success, 1 otherwise
     """
+    if cfg.is_attached and os.getuid() == 0:
+        # Refresh contracts prior to enable
+        contract.request_contract_updates(cfg, contract_token=None)
     ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[args.name]
     entitlement = ent_cls(cfg)
     return 0 if entitlement.enable() else 1
@@ -189,10 +191,7 @@ def action_detach(args, cfg):
     @return: 0 on success, 1 otherwise
     """
     if not cfg.is_attached:
-        print(textwrap.dedent("""\
-            This machine is not attached to a UA Subscription, sign up here:
-                  https://ubuntu.com/advantage
-        """))
+        print(ua_status.MESSAGE_UNATTACHED)
         return 1
     if os.getuid() != 0:
         print(ua_status.MESSAGE_NONROOT_USER)
@@ -224,9 +223,10 @@ def action_attach(args, cfg):
         bound_macaroon = bound_macaroon_bytes.decode('utf-8')
         cfg.write_cache('bound-macaroon', bound_macaroon)
         try:
-            contract_client.request_accounts(bound_macaroon)
+            contract_client.request_accounts(macaroon_token=bound_macaroon)
             contract_token = contract.get_contract_token_for_account(
-                contract_client, bound_macaroon, cfg.accounts[0]['id'])
+                contract_client, cfg.accounts[0]['id'],
+                macaroon_token=bound_macaroon)
         except (sso.SSOAuthError, util.UrlError) as e:
             logging.error(str(e))
             print('Could not attach machine. Unable to obtain authenticated'
@@ -235,17 +235,13 @@ def action_attach(args, cfg):
     else:
         contract_token = args.token
 
-    machine_token_response = contract_client.request_contract_machine_attach(
-        contract_token=contract_token)
-    contractInfo = machine_token_response['machineTokenInfo']['contractInfo']
-    for entitlement in contractInfo['resourceEntitlements']:
-        if entitlement.get('entitled'):
-            # Obtain each entitlement's accessContext for this machine
-            entitlement_name = entitlement['type']
-            contract_client.request_resource_machine_access(
-                cfg.machine_token['machineToken'], entitlement_name)
+    if not contract.request_contract_updates(cfg, contract_token):
+        print(
+            "Could not attach machine. Error contacting server %s" %
+            cfg.contract_url)
+        return 1
     print("This machine is now attached to '%s'.\n" %
-          machine_token_response['machineTokenInfo']['contractInfo']['name'])
+          cfg.machine_token['machineTokenInfo']['contractInfo']['name'])
     action_status(args=None, cfg=cfg)
     return 0
 
@@ -286,6 +282,10 @@ def get_parser():
         help='disable a specific support entitlement on this machine')
     disable_parser(parser_disable)
     parser_disable.set_defaults(action=action_disable)
+    parser_refresh = subparsers.add_parser(
+        'refresh', help=(
+            'Refresh ubuntu-advantage entitlements from contracts server.'))
+    parser_refresh.set_defaults(action=action_refresh)
     parser_version = subparsers.add_parser(
         'version', help='Show version of ua-client')
     parser_version.set_defaults(action=print_version)
@@ -306,6 +306,17 @@ def action_status(args, cfg):
 
 def print_version(_args=None, _cfg=None):
     print(version.get_version())
+
+
+def action_refresh(args, cfg):
+    if not cfg.is_attached:
+        print(ua_status.MESSAGE_UNATTACHED)
+        return 1
+    if contract.request_contract_updates(cfg, contract_token=None):
+        print('Refreshed Ubuntu Advantage contracts.')
+        logging.debug('Refreshed Ubuntu Advantage contracts.')
+        return 0
+    return 1
 
 
 def setup_logging(level=logging.INFO, log_file=None):
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -235,8 +235,7 @@ def action_attach(args, cfg):
         try:
             contract_client.request_accounts(macaroon_token=bound_macaroon)
             contract_token = contract.get_contract_token_for_account(
-                contract_client, cfg.accounts[0]['id'],
-                macaroon_token=bound_macaroon)
+                contract_client, bound_macaroon, cfg.accounts[0]['id'])
         except (sso.SSOAuthError, util.UrlError) as e:
             logging.error(str(e))
             print('Could not attach machine. Unable to obtain authenticated'
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -45,8 +45,20 @@ UA_STAGING_DASHBOARD_URL = 'https://contracts.staging.canonical.com'
 DEFAULT_LOG_FORMAT = (
     '%(asctime)s - %(filename)s:(%(lineno)d) [%(levelname)s]: %(message)s')
 
-
 STATUS_FORMATS = ['tabular', 'json']
+
+
+def assert_attached_root(func):
+    """Decorator asserting root user and attached config."""
+    def wrapper(args, cfg):
+        if not cfg.is_attached:
+            print(ua_status.MESSAGE_UNATTACHED)
+            return 1
+        if os.getuid() != 0:
+            print(ua_status.MESSAGE_NONROOT_USER)
+            return 1
+        return func(args, cfg)
+    return wrapper
 
 
 def attach_parser(parser=None):
@@ -159,6 +171,7 @@ def status_parser(parser=None):
     return parser
 
 
+@assert_attached_root
 def action_disable(args, cfg):
     """Perform the disable action on a named entitlement.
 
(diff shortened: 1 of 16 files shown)


uaclient/contract.py
```diff
@@ -68,30 +70,41 @@ def request_root_macaroon(self):
...
 
-    def request_accounts(self, macaroon_token):
+    def request_accounts(self, macaroon_token=None, machine_token=None):
         """Request list of accounts this user has access to."""
         headers = self.headers()
-        headers.update({'Authorization': 'Macaroon %s' % macaroon_token})
+        if macaroon_token:
+            headers.update({'Authorization': 'Macaroon %s' % macaroon_token})
+        else:
+            headers.update({'Authorization': 'Bearer %s' % machine_token})
```
OddBloke: If we aren't passed either `macaroon_token` or `machine_token`, we will add `Authorization: Bearer None` as a header; we should probably, instead, raise an error (rather than getting an inevitable error after a round-trip to the server).

uaclient/contract.py
```diff
@@ -68,30 +70,41 @@ def request_root_macaroon(self):
...
+    def request_account_contracts(
+            self, account_id, macaroon_token=None, machine_token=None):
         """Request a list of contracts authorized for account_id."""
         url = API_V1_TMPL_ACCOUNT_CONTRACTS.format(account=account_id)
         headers = self.headers()
-        headers.update({'Authorization': 'Macaroon %s' % macaroon_token})
+        if macaroon_token:
+            headers.update({'Authorization': 'Macaroon %s' % macaroon_token})
+        else:
+            headers.update({'Authorization': 'Bearer %s' % machine_token})
```
OddBloke: Same issue as above.

uaclient/contract.py
```diff
@@ -68,30 +70,41 @@ def request_root_macaroon(self):
...
+            self, contract_id, macaroon_token=None, machine_token=None):
         """Create a contract token for use when adding a machine to a contract
 
         """
         headers = self.headers()
-        headers.update({'Authorization': 'Macaroon %s' % macaroon_token})
+        if macaroon_token:
+            headers.update({'Authorization': 'Macaroon %s' % macaroon_token})
+        else:
+            headers.update({'Authorization': 'Bearer %s' % machine_token})
```
OddBloke: Same issue as above.

OddBloke: (Given this appears in multiple places, a function which handles this logic, including the appropriate error messaging might make sense?)

uaclient/cli.py
```diff
@@ -170,6 +169,12 @@ def action_enable(args, cfg):
 
     @return: 0 on success, 1 otherwise
     """
+    if cfg.is_attached and os.getuid() == 0:
+        # Refresh contracts prior to enable
+        contract_token = cfg.read_cache('contract-token')
```
OddBloke: This logic is replicated in `action_refresh()` below; should it move in to a helper?

(Should we at least be debug logging that we're performing this?)

uaclient/cli.py
```diff
@@ -320,6 +323,20 @@ def print_version(_args=None, _cfg=None):
     print(version.get_version())
 
 
+def action_refresh(args, cfg):
+    if not cfg.is_attached:
```
OddBloke: This doesn't check if we're root, but the similar code in `action_enable` does; is that an intentional difference?

uaclient/cli.py
```diff
@@ -320,6 +323,20 @@ def print_version(_args=None, _cfg=None):
...
+        print(ua_status.MESSAGE_UNATTACHED)
+        return 1
+    contract_token = cfg.read_cache('contract-token')
+    if contract_token:
```
OddBloke: If this is false, should we `return 1` at this point, rather than attempting a (presumably doomed?) `request_contract_updates` call?

uaclient/cli.py
```diff
@@ -227,17 +230,13 @@ def action_attach(args, cfg):
...
-    contractInfo = machine_token_response['machineTokenInfo']['contractInfo']
-    for entitlement in contractInfo['resourceEntitlements']:
-        if entitlement.get('entitled'):
-            # Obtain each entitlement's accessContext for this machine
-            entitlement_name = entitlement['type']
-            contract_client.request_resource_machine_access(
-                cfg.machine_token['machineToken'], entitlement_name)
+    if not contract.request_contract_updates(cfg, contract_token):
+        print(
+            "Could not attach machine. Error contacting server %s" %
```
OddBloke: Should this string be in `ua_status` instead?

uaclient/cli.py
```diff
@@ -320,6 +323,20 @@ def print_version(_args=None, _cfg=None):
...
 
+def action_refresh(args, cfg):
+    if not cfg.is_attached:
+        print(ua_status.MESSAGE_UNATTACHED)
+        return 1
+    contract_token = cfg.read_cache('contract-token')
+    if contract_token:
+        contract_token = contract_token['contractToken']
+    if contract.request_contract_updates(cfg, contract_token=contract_token):
+        print('Refreshed Ubuntu Advantage contracts.')
```
OddBloke: Should this string be in `ua_status` instead?

uaclient/contract.py
```diff
@@ -228,3 +244,41 @@ def redact_sensitive(contract_response):
...
+def request_contract_updates(cfg, contract_token):
+    """Request contract updates from ua-contracts service.
+
+    @param cfg: Instance of UAConfig for this machine.
+    @param contract_token: Token of the contract we are updating.
+
+    @return: True on success False otherwise.
+    """
+    contract_client = UAContractClient(cfg)
+    if not contract_token:  # Crawl the accounts contract routes to get details
```
OddBloke: As written, the function signature and documentation don't suggest `contract_token` can be False-ish; can we be more explicit about that?

uaclient/contract.py
```diff
@@ -228,3 +244,41 @@ def redact_sensitive(contract_response):
...
+        machine_token = cfg.machine_token['machineToken']
+        try:
+            contract_client.request_accounts(machine_token=machine_token)
+            contract_token = get_contract_token_for_account(
+                contract_client, cfg.accounts[0]['id'],
+                machine_token=machine_token)
+        except util.UrlError as e:
+            logging.error(str(e))
+            print('Could not update contract info from %s' %
```
OddBloke: Should this string be in `ua_status`?

uaclient/contract.py
blackboxsw: +1 OddBloke, sorry on this branch it was very WorkInProgress to give us something to talk about with ua-conracts folks (I should have indicated as such on the description). The contracts team went a significantly different direction per https://github.com/CanonicalLtd/ua-contracts/pull/238  which involves a big overhaul here. I'm working that now and will ping when this needs re-review.

uaclient/contract.py
blackboxsw: + 1 will resolve these as ua-contracts won't support both types of auth, dropping all related

uaclient/cli.py
blackboxsw: read_cache does a util.load_file which adds a debug log entry about the file read. That said, I had wanted to drop the load_file debug logs as they are super noisy. So, maybe it's worth putting the log here instead.

uaclient/cli.py
blackboxsw: good point. that's a bug, will fix thanks.

uaclient/cli.py
blackboxsw: I presume you're thinking about internationalization of messages... We probably should keep all printed statements, errors etc in one place. We'll probably have to separate thes out of ua/status.py as it's going to get fairly big I presume.

uaclient/cli.py
blackboxsw: added.

uaclient/cli.py
blackboxsw: added a check and returned 1 if contract_token is falsish

uaclient/contract.py
blackboxsw: moved a couple of strings into ua_status

uaclient/contract.py
blackboxsw: I think we now validate contract_token falseish and exit 1

uaclient/cli.py
blackboxsw: tried to move some strings to ua_status

uaclient/contract.py
blackboxsw: ok think I've addressed most comments

blackboxsw:
to test:
```
# update /var/lib/ua/private/machine-token.json to represent type: 'fips-updates' as 'entitled': true
cp /var/lib/ubuntu-advantage/private/machine-access-fips.json /var/lib/ubuntu-advantage/private/machine-access-fips-updates.json
# see ua status report entitled
ua refresh 
ua status  # will report unentitled again for fips-updates
```

uaclient/cli.py
blackboxsw: Added a decorator for this check  and applied it to detach, enable disable operations.

uaclient/cli.py
blackboxsw: moved a few more strings to ua_status.MESSAGE_*

uaclient/cli.py
```diff
@@ -165,36 +178,33 @@ def action_disable(args, cfg):
         return 1
 
 
+@assert_attached_root
 def action_enable(args, cfg):
     """Perform the enable action on a named entitlement.
 
     @return: 0 on success, 1 otherwise
     """
+    if cfg.is_attached and os.getuid() == 0:
```
OddBloke: We no longer need this check due to `assert_attached_root`

uaclient/cli.py
```diff
@@ -165,36 +178,33 @@ def action_disable(args, cfg):
         return 1
 
 
+@assert_attached_root
```
OddBloke: Does this mean we can drop the checks that `assert_attached_root` now performs from `UAEntitlement.can_enable`?

uaclient/config.py
```diff
@@ -160,14 +161,18 @@ def data_path(self, key=None, private=True):
             return os.path.join(data_dir, self.data_paths[key])
         return os.path.join(data_dir, key)
 
-    def delete_cache(self):
-        """Remove all configuration cached response files class attributes."""
+    def delete_cache(self, key=None):
```
OddBloke: I think a clearer API would be to have separate methods for "delete everything" and "delete this one thing"; very easy to accidentally delete everything by passing something False-ish as `key` to this current one.

(This doesn't need to happen in this PR; once it lands let's open up a techdebt issue.)

uaclient/contract.py
OddBloke: My comment here doesn't appear to have been addressed. (I've also unresolved the related comments below.)

panlinux:
From an UX perspective, I'm against issuing a refresh with every enable command. That is adding a non-trivial delay to simple commands, even when they are supposed to fail. For example:
```
root@trusty-ua:~# time ua enable cc
Canonical Common Criteria EAL2 Provisioning is not available for Ubuntu trusty.

real	0m6.565s
```
I think it's reasonable to expect the user to run `ua refresh` when an entitlement he just bought isn't showing as available yet.

uaclient/contract.py
blackboxsw: My fault. I thought I backed out the Bearer Auth optional param when ua-contracts went a different direction. dropped content related to this change from account and contracts request methods.

uaclient/contract.py
blackboxsw: really dropped

uaclient/contract.py
blackboxsw: really dropped

uaclient/cli.py
blackboxsw: cgood point, dropped, we had discussion about potentially not performing the pre-enable refresh check because it costs 7 seconds
@powersj will weigh in on whether this branch should drop the pre-enable refresh check.
  - Client could handle stale auth or url errors on setup here and emit 'Try ua refresh' to the user if enable didn't work for instance.

uaclient/cli.py
blackboxsw: I think we should be able to drop those checks from Entitlement.can_enable. no sense duplicating

blackboxsw: dropped from entitlements. made for a good number of mock drops on getuid there. Thanks for thinking of this.

uaclient/config.py
```diff
@@ -160,14 +161,29 @@ def data_path(self, key=None, private=True):
...
+
     def delete_cache(self):
-        """Remove all configuration cached response files class attributes."""
+        """Remove configuration cached response files class attributes."""
         self._contracts = None
         self._entitlements = None
         self._machine_token = None
-        for key in self.data_paths.keys():
+        for path_key in self.data_paths.keys():
             for private in (True, False):
```
OddBloke: Should we just be calling `self.delete_cache_key` in this for loop?

uaclient/config.py
```diff
@@ -160,14 +161,29 @@ def data_path(self, key=None, private=True):
             return os.path.join(data_dir, self.data_paths[key])
         return os.path.join(data_dir, key)
 
+    def delete_cache_key(self, key):
+        """Remove specific cache file."""
+        if not key:
+            return
```
OddBloke: Are there any cases where we would intentionally call `delete_cache_key` with a False-ish value? If not, then hitting this `return` implies that we have introduced a bug; could we either raise an exception or at least log an error here?

blackboxsw:
Created separate issues to track followup items/discussions:
* [apt service refreshes should look at aptURL/resourceToken/suites ](https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/385)
*[ livepatch service refresh should look at remoteServer, CACerts, series. kernel etc. ](https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/386)

uaclient/cli.py
blackboxsw: We won't address the separationg of pre-enable apt-get update in this branch. we will talk about it in ua status weekly checkin meeting

uaclient/config.py
blackboxsw: separate config.delete_cache_key method is now published.

blackboxsw:
> From an UX perspective, I'm against issuing a refresh with every enable command. That is adding a non-trivial delay to simple commands, even when they are supposed to fail. For example:
> 
> ```
> root@trusty-ua:~# time ua enable cc
> Canonical Common Criteria EAL2 Provisioning is not available for Ubuntu trusty.
> 
> real	0m6.565s
> ```
> I think it's reasonable to expect the user to run `ua refresh` when an entitlement he just bought isn't showing as available yet.

We will follow up on the time cost of ua enable performing an apt-get update in the weekly UA checkin meeting to determine if it needs dropping.

uaclient/config.py
blackboxsw: Dan, correct, we shouldn't ever call this with a non-string or empty key. I've dropped that early exit and we can expect a RuntimeError or TypeError if we get invalid values passed in. Let's break loudly if this happens as it is a programming error.

uaclient/config.py
blackboxsw: Good point. we should be calling delete_cache_key to avoid duplication.

uaclient/cli.py
```diff
@@ -173,36 +185,32 @@ def action_disable(args, cfg):
...
 def action_enable(args, cfg):
     """Perform the enable action on a named entitlement.
 
     @return: 0 on success, 1 otherwise
     """
+    logging.debug(ua_status.MESSAGE_REFRESH_ENABLE)
+    if not contract.request_updated_contract(cfg):
```
OddBloke: Do we need to file an issue to capture the fact we were considering not doing this here?

uaclient/config.py
```diff
@@ -161,16 +162,29 @@ def data_path(self, key=None, private=True):
...
+            self._machine_token = None
+        for private in (True, False):
+            cache_path = self.data_path(key, private)
+            if os.path.exists(cache_path):
+                os.unlink(cache_path)
+
     def delete_cache(self):
-        """Remove all configuration cached response files class attributes."""
+        """Remove configuration cached response files class attributes."""
         self._contracts = None
```
OddBloke: These lines are redundant now, I believe?

uaclient/cli.py
blackboxsw: already filed. except that the issue morphed into, we are doing this, but we will make it more performant. https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/399

uaclient/config.py
blackboxsw: dropped

powersj:
I'm merging as comments have been addressed

PR #349, opened 2019-04-10
deb: prerm and postrm cleanup no need to remove esm list file in postrm

blackboxsw:
Since prerm calles migrate_apt_sources(clean=True) all sources list
files are already removed in prerm.

Also prerm doesn't take a purge param


powersj:
@panlinux should we revert this given we found it doesn't cover purge?

panlinux:
We need to better understand what `migrate_apt_sources()` does, it shouldn't remove the config files unless it's a purge. And since this is already post 19.2 tag, I would revert it yes, until we have that understanding.

Issue #339, opened 2019-04-09
Only root sees esm enabled


panlinux:
I can't say if this is realted to commit 476598c.
After I enable esm, I see this behavior:
```
ubuntu@trusty-ua:~$ sudo ua status|grep esm
esm           entitled         active

ubuntu@trusty-ua:~$ ua status|grep esm
esm           entitled         inactive
```

Same with `apt-cache policy`:
```
ubuntu@trusty-ua:~$ apt-cache policy | grep esm
-32768 https://esm.staging.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
     origin esm.staging.ubuntu.com

ubuntu@trusty-ua:~$ sudo apt-cache policy | grep esm
 500 https://esm.staging.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
     origin esm.staging.ubuntu.com
```

panlinux:
I don't see how `ua` would interfere with `apt-cache policy` not working. I'll troubleshoot more tomorrow

panlinux:
If I make `/etc/apt/auth.conf.d/90ubuntu-advantage` mode `0644`, then both `ua status` and `apt-cache policy` will say esm is enabled (and pinning 500 instead of -32768 in the case of `apt-cache`).

I can't tell right now if this is new behavior, or if it was like this since the beginning.

blackboxsw:
This is new behavior because apt not magically checks auth perms to a repo to set this magic pin

blackboxsw:
UAClient could capture when an entitlement was enabled to help out the non-root user case, But I'm wondering if there is another besides checking apt-cache policy  for the magic pin for non-root user. Maybe Julian has thoughts here.

julian-klode:
I mean, you could just write individual files per entitlement and then check for their presence. This does not help apt as user though, as it will still see ESM being disabled because apt can't read the file.

If this is a severe limitation we want to get rid of, we have to disable writing Packages-Require-Authorization to the Release file and manually pin the ESM repo down with "Pin-Priority: never" (and ua needs a Depends apt (>= version with that feature).

panlinux:
This is somewhat new behavior. In the old bash ua client, we used to write sources.list.d files with creds in there, mode 0644, exactly because making them 0600 would break apt-cache policy. That's when we were directed at auth.conf, which *can* be made 0600 and apt-cache policy would still work, and it did.

panlinux:
[...]

PR #350, opened 2019-04-10
repo: persist a localEnabled flag in machine-acces for non-root status

blackboxsw:
Any repo-based entitlement can not check apt-cache policy to determine
if the entitlement is enabled because apt will automatically pin -32768
for a non-root user of an authenticated apt source because the non-root
user does not have permission on /etc/apt/auth.conf.d/<secureconffile>
due to mode 600 on those files.

As a result uaclient will fallback to our localEnabled flag in public
cache files which is set True after enable and False at disable.

Contained in this merge
- move redact_sensitive function out of contract and into util
- cis, fips*, esm, cis-audit will amend public redacted cache files with localEnabled True|False on enable|disable operations
- RepoEntitlement.operational_status  checks  to localEnabled == True value for non-root user if magic apt pin -32768 was found.
- Fixed  trusty util.is_container fails to report container for non-root user because running-in-container fails to report 'lxc'  (filed bug https://bugs.launchpad.net/upstart/+bug/1824270)
    - This is_container fix avoids an invalid 'inactive' setting reported to non-root users for entitlements that can't be installed on containers (livepatch/fips)

Fixes: #339

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -39,11 +39,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                     ['apt-get', 'remove', '--assume-yes'] + self.packages)
             except util.ProcessExecutionError:
                 pass
-            public_cache = self.cfg.read_cache('machine-access-%s' % self.name)
-            public_cache['localEnabled'] = False
-            redacted_cache = util.redact_sensitive(public_cache)
-            self.cfg.write_cache(
-                'machine-access-%s' % self.name, redacted_cache, private=False)
+            self._set_local_enabled(False)
         if not silent:
             print('Warning: no option to disable {title}'.format(
                 title=self.title)
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -39,6 +39,11 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                     ['apt-get', 'remove', '--assume-yes'] + self.packages)
             except util.ProcessExecutionError:
                 pass
+            public_cache = self.cfg.read_cache('machine-access-%s' % self.name)
+            public_cache['localEnabled'] = False
+            redacted_cache = util.redact_sensitive(public_cache)
+            self.cfg.write_cache(
+                'machine-access-%s' % self.name, redacted_cache, private=False)
         if not silent:
             print('Warning: no option to disable {title}'.format(
                 title=self.title)
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -108,11 +108,7 @@ class RepoEntitlement(base.UAEntitlement):
                     status.MESSAGE_ENABLED_FAILED_TMPL.format(
                         title=self.title))
                 return False
-        public_cache = self.cfg.read_cache('machine-access-%s' % self.name)
-        public_cache['localEnabled'] = True
-        redacted_cache = util.redact_sensitive(public_cache)
-        self.cfg.write_cache(
-            'machine-access-%s' % self.name, redacted_cache, private=False)
+        self._set_local_enabled(True)
         print(status.MESSAGE_ENABLED_TMPL.format(title=self.title))
         for msg in self.messaging.get('post_enable', []):
             print(msg)
@@ -140,3 +136,11 @@ class RepoEntitlement(base.UAEntitlement):
             # policy will show APT_DISABLED_PIN for authenticated sources
             return status.ACTIVE, '%s is active' % self.title
         return status.INACTIVE, '%s is not configured' % self.title
+
+    def _set_local_enabled(self, value):
+        """Set local enabled flag true or false."""
+        public_cache = self.cfg.read_cache('machine-access-%s' % self.name)
+        public_cache['localEnabled'] = value
+        redacted_cache = util.redact_sensitive(public_cache)
+        self.cfg.write_cache(
+            'machine-access-%s' % self.name, redacted_cache, private=False)
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -108,6 +108,11 @@ class RepoEntitlement(base.UAEntitlement):
                     status.MESSAGE_ENABLED_FAILED_TMPL.format(
                         title=self.title))
                 return False
+        public_cache = self.cfg.read_cache('machine-access-%s' % self.name)
+        public_cache['localEnabled'] = True
+        redacted_cache = util.redact_sensitive(public_cache)
+        self.cfg.write_cache(
+            'machine-access-%s' % self.name, redacted_cache, private=False)
         print(status.MESSAGE_ENABLED_TMPL.format(title=self.title))
         for msg in self.messaging.get('post_enable', []):
             print(msg)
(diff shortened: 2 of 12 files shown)


uaclient/util.py
```diff
@@ -72,16 +68,12 @@ def encode_text(text, encoding='utf-8'):
...
-            # try to run a helper program. if it returns true/zero
-            # then we're inside a container. otherwise, no
-            subp(helper)
-            return True
-        except (IOError, OSError):
-            pass
+    for filename in ('container_type', 'systemd/container'):
+        path = os.path.join(run_path, filename)
+        if os.path.exists(path):
+            return bool(load_file(path).strip())
```
OddBloke: I _think_ with both of these files, their existence is sufficient to confirm we're in a container; do we need to read them?

uaclient/entitlements/tests/test_cc.py
```diff
@@ -49,6 +50,33 @@
...
 
+class TestCommonCriteriaEntitlementOperationalStatus:
+
+    @pytest.mark.parametrize(
+        'arch,series,details',
+        (('arm64', 'xenial', 'Canonical Common Criteria EAL2 Provisioning is'
+          ' not available for platform arm64.\nSupported platforms are:'
+          ' x86_64, ppc64le, s390x\n'),
+         ('s390x', 'trusty', 'Canonical Common Criteria EAL2 Provisioning'
+          ' is not available for Ubuntu trusty.')))
```
OddBloke: Why does one of these have a trailing newline and the other doesn't? Do we have inconsistent output for these two cases?

uaclient/entitlements/repo.py
```diff
@@ -130,4 +135,8 @@ def operational_status(self):
         match = re.search(r'(?P<pin>(-)?\d+) %s' % repo_url, out)
         if match and match.group('pin') != APT_DISABLED_PIN:
             return status.ACTIVE, '%s is active' % self.title
+        if os.getuid() != 0 and entitlement_cfg.get('localEnabled', False):
```
OddBloke: What's the rationale for only using the cache for non-root users? If it's correct for non-root users (which hopefully we're confident of!), then can we use it for root also, and save ourselves the `apt-cache` call?

blackboxsw:
to test:
```
# update /var/lib/ua/private/machine-token.json to represent type: 'fips-updates' as 'entitled': true
cp /var/lib/ubuntu-advantage/private/machine-access-fips.json /var/lib/ubuntu-advantage/private/machine-access-fips-updates.json
# see ua status report entitled
ua refresh 
ua status  # will report unentitled again for fips-updates
```

uaclient/util.py
blackboxsw: fixed and pushed is_container changes to first attempt systemd-detect-virt and fall back to presence of said files per your review.

uaclient/entitlements/tests/test_cc.py
blackboxsw: good point, it is inconsistent spacing on the output for those cases. it probably should be the same. I'll drop the trailing newline

uaclient/entitlements/repo.py
blackboxsw: Ahh the reason I thought of limiting this case to just non-root users is because it is not the real configuration on the system and could become outdated in two cases:
 - someone manually removes  apt sources or apt auth config for a entitlement outside of ua client tooling
- the backend retires the current auth token for some reason and the invalid token is still being used. In that case ua status would lie saying enabled, but apt-get update would be choking with 403s.

blackboxsw: my thought is we at least want to insulate root user from such potentially stale inconsistency.

uaclient/entitlements/fips.py
```diff
@@ -39,6 +39,11 @@ def disable(self, silent=False, force=False):
                     ['apt-get', 'remove', '--assume-yes'] + self.packages)
             except util.ProcessExecutionError:
                 pass
+            public_cache = self.cfg.read_cache('machine-access-%s' % self.name)
```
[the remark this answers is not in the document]
blackboxsw: Good call, added _set_local_enabled

uaclient/util.py
```diff
@@ -65,16 +61,17 @@ def encode_text(text, encoding='utf-8'):
     return text.encode(encoding)
 
 
-def is_container():
+def is_container(run_path='/run'):
```
OddBloke: This shouldn't block this landing, but what is `run_path` intended to be used for?  It doesn't look like it's utilised anywhere in this PR.

uaclient/entitlements/repo.py
OddBloke: If we are root in `operational_status`, should we update the cache once we have determined what the status is? That would increase our chances of having a cache that represents reality.

blackboxsw: +1 I like it. it would solve some corner cases.

uaclient/util.py
blackboxsw: run_path is used when we fall through and don't have systemd-detect-virt utility

uaclient/entitlements/repo.py
OddBloke: Are you happy including that in this PR, or shall we raise a separate (but still blocker) issue?

uaclient/util.py
OddBloke: Sorry, I didn't mean `run_path` as a local variable, I meant `run_path` as a parameter; why does it need to be modifiable by callers (and if it does need to be, why aren't any of them doing so now)?

uaclient/entitlements/repo.py
blackboxsw: actually let's hold off on this PR, I think that needs broader discussion Let's file a blocker issue and discuss there

uaclient/util.py
blackboxsw: ahh right, it was intended to allow me to add a unit test that seeded tmpdir fixture in there, but I never committed a new unit test for it.

uaclient/tests/test_util.py
```diff
@@ -68,6 +68,39 @@
 """
 
 
+class TestIsContainer:
```
OddBloke: I think this needs some negative tests; something like:

```py
def is_container(run_path=None):
    try:
        util.subp(['systemd-detect-virt', '--quiet', '--container'])
    except:
        pass
    return True
```

would pass this testing, and is obviously incorrect. :-)

uaclient/tests/test_util.py
```diff
@@ -68,6 +68,39 @@
...
+        m_subp.return_value = '', ''
+        assert True is util.is_container()
+        calls = [mock.call(['systemd-detect-virt', '--quiet', '--container'])]
+        assert calls == m_subp.call_args_list
+
+    @mock.patch('uaclient.util.subp')
+    def test_true_on_run_container_type(self, m_subp, tmpdir):
+        """Return True when /run/container_type exists."""
+        m_subp.side_effect = OSError('No systemd-detect-virt utility')
+        util.write_file(tmpdir.join('container_type').strpath, '')  # touch
```
OddBloke: This could be `tmpdir.join('container_type').write('')`

uaclient/tests/test_util.py
```diff
@@ -68,6 +68,39 @@
...
+        assert True is util.is_container(run_path=tmpdir)
+        calls = [mock.call(['systemd-detect-virt', '--quiet', '--container'])]
+        assert calls == m_subp.call_args_list
+
+    @mock.patch('uaclient.util.subp')
+    def test_true_on_run_systemd_container(self, m_subp, tmpdir):
+        """Return True when /run/systemd/container exists."""
+        m_subp.side_effect = OSError('No systemd-detect-virt utility')
+        path = tmpdir.join('systemd/container').strpath
+        os.makedirs(os.path.dirname(path))
```
OddBloke: I think this could be `tmpdir.join('systemd/container').write('', ensure=True)`

uaclient/tests/test_util.py
blackboxsw: good point. fixed and pushed

uaclient/tests/test_util.py
blackboxsw: done.

uaclient/tests/test_util.py
blackboxsw: done and thanks

OddBloke:
The Travis failure looks like a temporary infra problem on their side, so I've restarted it.

uaclient/entitlements/repo.py
OddBloke: Did this issue get filed?

PR #362, opened 2019-04-12
apt-hook: Do not crash/fail if we can't read /proc/self/status

julian-klode:
Emergency hotfix, as disco image builds are broken:

Only run the self-test if we can access the file, and ignore any
errors from reading status files, so that things fail silently
where we can't read them for whatever reasons.

This broke live image building for server.

### DIFF
--- a/apt-hook/hook.cc
+++ b/apt-hook/hook.cc
@@ -60,9 +60,6 @@ static std::string getppid_of(std::string pid)
       return line;
    }
 
-   if (stream.fail())
-      _error->Error("Could not parse proc status of %s", pid.c_str());
-
    return "";
 }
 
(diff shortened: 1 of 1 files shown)


panlinux:
https://launchpadlibrarian.net/418949790/buildlog_ubuntu_disco_amd64_ubuntu-server-live_BUILDING.txt.gz
```
(...)
Processing triggers for libc-bin (2.29-0ubuntu1) ...
Processing triggers for initramfs-tools (0.131ubuntu18) ...
apt-esm-hook: hook.cc:186: int main(int, char**): Assertion `ppid == getppid_of("self")' failed.
Aborted (core dumped)
E: Problem executing scripts APT::Install::Post-Invoke-Success '[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook'
E: Sub-process returned an error code
E: config/hooks/032-installer-squashfs.binary failed (exit non-zero). You should check for errors.
P: Begin unmounting filesystems...
P: Saving caches...
P: Begin unmounting filesystems...
P: Saving caches...
Build failed
(...)
```

PR #378, opened 2019-04-15
esm: only remove the apt auth entry on disable

### LINKED ISSUE #351: enable esm doesn't work a second time
When one runs `ua disable esm`, that is also mistakenly removing the `sources.list.d/` file and the keyring file. This has at least two implications:
- you don't get back to the pre-enable situation with the ESM advertisement, because the `sources.list.d/` file is gone
- if you enable esm again, apt will fail because the keyring file is gone from `/etc/apt/trusted.gpg.d/` directory

OddBloke:
We only remove the repo from the apt auth file, because ESM is a
special-case: we want to be able to report on the available ESM updates
even when it's disabled.

Fixes #351

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -131,15 +131,8 @@ def add_apt_auth_conf_entry(repo_url, login, password):
     util.write_file(apt_auth_file, '\n'.join(new_lines), mode=0o600)
 
 
-def remove_auth_apt_repo(repo_filename, repo_url, keyring_file=None,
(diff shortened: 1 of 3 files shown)


OddBloke:
This needs some test fixes; I wasn't sure if I was only seeing issues locally because of Python 3.7, so I wanted a Travis run. Will reopen once I expect Travis to pass.

OddBloke:
I couldn't reopen this, so see #382.

PR #382, opened 2019-04-15
esm: remove only the apt auth entry on disable

### LINKED ISSUE #351: enable esm doesn't work a second time

OddBloke:
We only remove the repo from the apt auth file, because ESM is a
special-case: we want to be able to report on the available ESM updates
even when it's disabled.

Fixes #351

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -131,15 +131,8 @@ def add_apt_auth_conf_entry(repo_url, login, password):
     util.write_file(apt_auth_file, '\n'.join(new_lines), mode=0o600)
 
 
-def remove_auth_apt_repo(repo_filename, repo_url, keyring_file=None,
-                         fingerprint=None):
-    """Remove an authenticated apt repo and credentials to the system"""
-    logging.info('Removing authenticated apt repo: %s', repo_url)
-    util.del_file(repo_filename)
-    if keyring_file:
-        util.del_file(keyring_file)
-    elif fingerprint:
-        util.subp(['apt-key', 'del', fingerprint], capture=True)
+def remove_repo_from_apt_auth_file(repo_url):
+    """Remove a repo from the shared apt auth file"""
     _protocol, repo_path = repo_url.split('://')
     if repo_path.endswith('/'):  # strip trailing slash
         repo_path = repo_path[:-1]
@@ -154,6 +147,18 @@ def remove_auth_apt_repo(repo_filename, repo_url, keyring_file=None,
             os.unlink(apt_auth_file)
         else:
             util.write_file(apt_auth_file, content, mode=0o600)
+
+
+def remove_auth_apt_repo(repo_filename, repo_url, keyring_file=None,
+                         fingerprint=None):
+    """Remove an authenticated apt repo and credentials to the system"""
+    logging.info('Removing authenticated apt repo: %s', repo_url)
+    util.del_file(repo_filename)
+    if keyring_file:
+        util.del_file(keyring_file)
+    elif fingerprint:
+        util.subp(['apt-key', 'del', fingerprint], capture=True)
+    remove_repo_from_apt_auth_file(repo_url)
 
 
 def add_ppa_pinning(apt_preference_file, repo_url, origin, priority):
--- a/uaclient/entitlements/esm.py
+++ b/uaclient/entitlements/esm.py
@@ -25,16 +25,16 @@ class ESMEntitlement(repo.RepoEntitlement):
         if not self.can_disable(silent, force):
             return False
         series = util.get_platform_info('series')
-        repo_filename = self.repo_list_file_tmpl.format(
-            name=self.name, series=series)
-        keyring_file = os.path.join(apt.APT_KEYS_DIR, self.repo_key_file)
         entitlement_cfg = self.cfg.read_cache(
             'machine-access-%s' % self.name)['entitlement']
         access_directives = entitlement_cfg.get('directives', {})
         repo_url = access_directives.get('aptURL', self.repo_url)
         if not repo_url:
             repo_url = self.repo_url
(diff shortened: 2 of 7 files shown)


OddBloke:
(Turns out if you force-push to a branch, you can't reopen a pull request for that branch, so here's a new one.)

uaclient/apt.py
```diff
@@ -131,15 +131,8 @@ def add_apt_auth_conf_entry(repo_url, login, password):
...
-def remove_auth_apt_repo(repo_filename, repo_url, keyring_file=None,
-                         fingerprint=None):
-    """Remove an authenticated apt repo and credentials to the system"""
-    logging.info('Removing authenticated apt repo: %s', repo_url)
-    util.del_file(repo_filename)
-    if keyring_file:
-        util.del_file(keyring_file)
-    elif fingerprint:
-        util.subp(['apt-key', 'del', fingerprint], capture=True)
+def remove_repo_from_apt_auth_file(repo_url):
```
blackboxsw: Thanks for adding the unit test for esm, is it work a minimal unit test for remove_repo_from_apt_auth_file and/or remove_auth_repo too?

uaclient/entitlements/esm.py
```diff
@@ -25,16 +25,16 @@ def disable(self, silent=False, force=False):
...
             'machine-access-%s' % self.name)['entitlement']
         access_directives = entitlement_cfg.get('directives', {})
         repo_url = access_directives.get('aptURL', self.repo_url)
         if not repo_url:
             repo_url = self.repo_url
-        apt.remove_auth_apt_repo(repo_filename, repo_url, keyring_file)
+        # We only remove the repo from the apt auth file, because ESM is a
+        # special-case: we want to be able to report on the available ESM
+        # updates even when it's disabled
+        apt.remove_repo_from_apt_auth_file(repo_url)
```
blackboxsw: I think we also need some additional logic here to revert the original /etc/apt/sources.list.d/ubuntu-esm-trusty.list to the content which was added in our postinst script which adds the following
```
deb https://esm... trusty-security main 
and 
deb https://esm... trusty-updates main 
```

During ua enable esm the trusty-updates suite can be dropped if the trusty machine was not configured for 'trusty-updates'. This results in us dropping visibility to ESM packages in ESM's trusty-updates suite. We need to reconstitute that file (or at least re-add the trusty-updates suite if needed on auth disable.

uaclient/apt.py
OddBloke: Yep, will add.

uaclient/entitlements/esm.py
OddBloke: Do you think we need to resolve this in this PR, or shall I file an issue so we can cover it in a follow-up?

(This PR still gets us to a better place; without it, we'll never get any info about ESM updates in MOTD.)

uaclient/entitlements/tests/test_esm.py
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+}
+
+M_PATH = 'uaclient.entitlements.esm.ESMEntitlement.'
+M_REPOPATH = 'uaclient.entitlements.repo.'
+M_GETPLATFORM = M_REPOPATH + 'util.get_platform_info'
+
+
+@pytest.fixture
```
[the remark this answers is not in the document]
OddBloke: This is used in `test_disable_removes_apt_config` in this file. If you want to double-check if it's used in the other places it's defined, we can open a techdebt issue to address those other cases?

uaclient/apt.py
OddBloke: @blackboxsw I've just pushed up one half of these tests because writing them raised a couple of questions about the implementation of `remove_auth_apt_repo`.

Firstly, it will only use the `fingerprint` for deletion if `keyring_file` is not given; should we unconditionally perform both actions if both are given?

Secondly, the `fingerprint` parameter doesn't appear to be passed in by any consumer. Do we actually want to keep it, or can we simplify the codebase a little?

(Whatever the answer to each of these questions is, I think we can open an issue to address it; I've already written tests for the existing behaviour, so we won't save any time by addressing it in this PR.)

uaclient/entitlements/esm.py
blackboxsw: Agreed. let's deal with that separately. /me needs to get in an agile/iterative space

uaclient/apt.py
blackboxsw: @OddBloke I think the way uaclient is headed, we ultimately want to drop static keyrings from uaclient if we can. The intent of the new client talking to ua-contracts backend was to primarily support the fingerprint or short key that the backend provides instead of these static keys.  I don't think we should perform the install operation on both keys. They probably should be mutually exclusive or at least  prefer the backend advertised key if present.

blackboxsw: +1 on separate PRs

OddBloke: I just opened #394 to cover the removal of the parameter, and #395 to cover unconditional deletion.

uaclient/entitlements/esm.py
OddBloke: I've opened #396 for this.

OddBloke:
I've realised that this causes failures on trusty that our tests don't catch yet; fixing that now.

OddBloke:
OK, that last commit should have failed in Travis, but #398 means we're getting false positive results; I'll fix the issue just by testing locally.

blackboxsw:
+1 on the followup changeset and tests here. Thanks!

PR #383, opened 2019-04-15
apt-hook: Add basic i18n (good enough for lang packs)

julian-klode:
This just generates the .pot at build time, so launchpad picks it
up and we can have translations in language packs.

We should probably move that into the parent Makefile and/or have
it be done by setup.py, but this should work for now.


julian-klode:
Not that it's urgent - nobody will be building language packs anyway now.

apt-hook/Makefile
```diff
@@ -1,6 +1,9 @@
-all: hook
+all: build
 
-build: hook
+build: hook ubuntu-advantage.pot
+
+ubuntu-advantage.pot: hook.cc
+	xgettext hook.cc -o ubuntu-advantage.pot
```
powersj: Excuse my ignorance here, but once the file is generated do we need to do something with it?

julian-klode: Normally you'd ship the .pot and translated .po files in the tarball; but this is more of a hack, as we don't really care I guess, since we do translations on the package via langpacks anyway. So we can rely on launchpad picking the file up after the build and then shipping translations in langpacks.

Obviously, we'd want to have one pot file for python and C++ code, so that might need different work.

PR #387, opened 2019-04-15
OC: ua trusty enable esm

panlinux:
Support enable ESM on trusty


tests/test_esm.py
```diff
@@ -10,19 +10,33 @@ class ESMTest(UbuntuAdvantageTest):
...
+            '# deb-src https://user:pass@esm.ubuntu.com/ubuntu '
+            'trusty-updates main\n')
+        for series in ['precise', 'trusty']:
+            self.SERIES = series
+            process = self.script('enable-esm', 'user:pass')
+            self.assertEqual(0, process.returncode)
+            self.assertIn('Ubuntu ESM repository enabled', process.stdout)
+            self.assertEqual(expected_repo_list[series],
+                             self.repo_list.read_text())
+            keyring_file = self.trusted_gpg_dir / 'ubuntu-esm-keyring.gpg'
```
powersj: os.path.join(self.trusted_gpg_dir, 'ubuntu-esm-keyring.gpg')

of course we will need something different for trusty

panlinux: I have another branch to add the unattended-upgrades config, it's https://github.com/panlinux/ubuntu-advantage-client/tree/oc-unattended-upgrades-pockets

I wanted to land things piece by piece.

About the gpg key, do we have it already, or should I just make one up and replace it when the time comes? I also wanted to do a separate branch for that.

powersj: If you fix the keyring_file creation I'll +1 this.

Let's ask about the GPG key in standup tomorrow.

I'll take al ook at the UU config next.

panlinux: Ah, I see what you mean now about the keyring. Looks like I will have to handle that in this branch.

panlinux:
I'm installing both keyrings in the case of an ESM supported distro. Should I split that up? On trusty, only install the trusty keyring? I haven't fully considered the precise to trusty release upgrade scenario.

Issue #390, opened 2019-04-15
OC: install new unattended-upgrades config for trusty ESM


panlinux:
Trusty ESM is the first ESM that will have `-updates` and `-security` pockets, and these need to be allowed origins in the unattended upgrades configuration.

panlinux:
This was fixed in PR #392

PR #392, opened 2019-04-15
OC: Install unattended-upgrades config

panlinux:
Install an unattended-upgrades config for the ESM -security pocket.

Fixes issue #390

### DIFF
--- a/apt.conf.d/51ubuntu-advantage-esm
+++ b/apt.conf.d/51ubuntu-advantage-esm
@@ -1,0 +1,3 @@
+Unattended-Upgrade::Allowed-Origins {
+  "${distro_id}ESM:${distro_codename}-security";
+};
--- a/debian/install
+++ b/debian/install
@@ -1,3 +1,4 @@
 ubuntu-advantage usr/bin/
 keyrings/*.gpg usr/share/keyrings/
 update-motd.d etc/
+apt.conf.d etc/apt/


panlinux:
Nothing else to do, these are taken care by the package itself. We may have gone overboard with postrm in the disco package, have to check. These config files here are not generated in postinst, so dpkg takes care of them because it knows about them:
```
$ cat /var/lib/dpkg/info/ubuntu-advantage-tools.conffiles 
/etc/apt/apt.conf.d/51ubuntu-advantage-esm
/etc/update-motd.d/99-esm

$ sudo apt remove ubuntu-advantage-tools
...
$ l /etc/apt/apt.conf.d/51ubuntu-advantage-esm /etc/update-motd.d/99-esm
-rw-r--r-- 1 root root  90 Apr 15 21:02 /etc/apt/apt.conf.d/51ubuntu-advantage-esm
-rwxr-xr-x 1 root root 604 Apr 15 17:08 /etc/update-motd.d/99-esm

$ sudo apt purge ubuntu-advantage-tools
...
$ l /etc/apt/apt.conf.d/51ubuntu-advantage-esm /etc/update-motd.d/99-esm
ls: cannot access /etc/apt/apt.conf.d/51ubuntu-advantage-esm: No such file or directory
ls: cannot access /etc/update-motd.d/99-esm: No such file or directory
```

Issue #389, opened 2019-04-15
OC: New GPG key for trusty ESM


panlinux:
I'm told trusty esm will have a new GPG key. We need to add it to the package.

panlinux:
I have the pubkey.

powersj:
Closing

PR #400, opened 2019-04-17
OC: ua trusty new gpg key

panlinux:
This installs the new ESM signing key v2 we got from #is.

Fixes issue #389


ubuntu-advantage
```diff
@@ -15,7 +15,7 @@ FIPS_REPO_URL="private-ppa.launchpad.net/ubuntu-advantage/fips"
 FIPS_REPO_KEY_FILE="ubuntu-fips-keyring.gpg"
 FIPS_REPO_LIST=${FIPS_REPO_LIST:-"/etc/apt/sources.list.d/ubuntu-fips-${SERIES}.list"}
 REPO_URL="esm.ubuntu.com"
-REPO_KEY_FILES="ubuntu-esm-keyring.gpg ubuntu-trusty-esm-keyring.gpg"
+REPO_KEY_FILES="ubuntu-esm-keyring.gpg ubuntu-esm-v2-keyring.gpg"
```
powersj: You need to remove the original ESM key file from this list. You deleted it from the package so later on it cannot be used and this variable is used to cp files into apt's keyring.

panlinux: I did not delete it from the package. I deleted an extra copy that was in the root directory of the repository. The one we are installing is in the keyrings/ directory.

PR #405, opened 2019-04-17
OC: support auth.conf.d for ESM

panlinux:
This is not changing FIPS to use auth.conf.d, because FIPS does not exist in trusty. If we want to do that work, to keep things consistent, then that is a matter for another branch.

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -27,6 +27,13 @@ case "$1" in
         # remove ESM keyring files created by enable-esm
         rm -f /etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg
         rm -f /etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg
+        # ESM sources.list.d snippet
+        series=$(grep ^DISTRIB_CODENAME= /etc/lsb-release | cut -d = -f 2)
+        if [ -n "${series}" ]; then
+            rm -f /etc/apt/sources.list.d/ubuntu-esm-${series}.list
+        fi
+        # ubuntu-advantage auth.conf.d snippet
+        rm -f /etc/apt/auth.conf.d/90ubuntu-advantage
     ;;    
 
     *)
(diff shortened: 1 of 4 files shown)


tests/testing.py
```diff
@@ -99,6 +102,8 @@ def script(self, *args):
             'FIPS_BOOT_CFG_DIR': str(self.etc_dir),
```
blackboxsw: I think you missed a REPO_URL setting -> ESM_REPO_URL in testing.py

debian/postrm
```diff
@@ -27,6 +27,13 @@ case "$1" in
         # remove ESM keyring files created by enable-esm
         rm -f /etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg
         rm -f /etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg
+        # ESM sources.list.d snippet
+        series=$(grep ^DISTRIB_CODENAME= /etc/lsb-release | cut -d = -f 2)
+        if [ -n "${series}" ]; then
+            rm -f /etc/apt/sources.list.d/ubuntu-esm-${series}.list
```
blackboxsw: Should we also remove or reinstate /etc/apt/sources.list.d/ubuntu-esm-${series}.list.save if present?

panlinux: I wonder. The point of keeping a save file around is because previously the creds would be in this file. But now they are not, so keeping this file around is a bit useless in my opinion, and even generating it is a bit useless. In the spirit of doing the least amount of changes possible, I kept it, that's all.

panlinux: I think I'll remove it, since I'm also removing the auth.conf.d snippet

panlinux: Although... even when removing ppas, a .save file is kept

tests/testing.py
panlinux: Fixed. Sorry I had to force push due to other reasons, but you can see the specific commit that fixed this (3d189615f3ebf705c8577be82d8a23027e686c73)

Issue #411, opened 2019-04-18
Only attempt to enable enableByDefault entitlements if they are applicable to the current system


OddBloke:
Entitlements can have the `enableByDefault` obligation, but not be applicable to the current system (e.g. livepatch is `enableByDefault` but inapplicable to a system running the stock trusty kernel, because we don't publish livepatches for it). In those cases, instead of attempting to enable and displaying error messages to the user, we should simply skip the enablement step (whilst reporting that it was skipped to the user).

Open Questions:
- [x] should we report that we skipped enabling to the user?

powersj:
> should we report that we skipped enabling to the user?

:+1:

OddBloke:
Cool, description updated.

PR #412, opened 2019-04-18
honour enableByDefault obligations from contract server

### LINKED ISSUE #239: uaclient honor enableByDefault entitlement obligation from ua-contracts to auto-enable an entitlement
For an entitlement (esm|livepatch) those entitlements should automatically be enabled after `ua attach` if the entitlement specifies obligations: enableByDefault: true. This true value means:
   if the system meets all affordances (constraints) then it *should* be enabled upon 'attach' instead of manually enabled after the fact.

      "obligations" : {
         "enableByDefault" : true
      },

OddBloke:
This implements `enableByDefault` support so that ubuntu-advantage-client will enable the appropriate entitlements at attach time.

(More code will follow to address #411, so that we only attempt to enable entitlements if they are applicable to the current system.)

Fixes #239.

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -185,6 +185,19 @@ def action_disable(args, cfg):
         return 1
 
 
+def _perform_enable(entitlement_name: str, cfg: config.UAConfig) -> bool:
+    """Perform the enable action on a named entitlement.
+
+    (This helper excludes any messaging, so that different enablement code
+    paths can message themselves.)
+
+    @return: True on success, False otherwise
+    """
+    ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[entitlement_name]
+    entitlement = ent_cls(cfg)
+    return entitlement.enable()
+
+
 @assert_attached_root
 def action_enable(args, cfg):
     """Perform the enable action on a named entitlement.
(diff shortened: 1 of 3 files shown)


uaclient/cli.py
```diff
@@ -194,9 +207,7 @@ def action_enable(args, cfg):
     logging.debug(ua_status.MESSAGE_REFRESH_ENABLE)
     if not contract.request_updated_contract(cfg):
         logging.debug(ua_status.MESSAGE_REFRESH_FAILURE)
-    ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[args.name]
-    entitlement = ent_cls(cfg)
-    return 0 if entitlement.enable() else 1
+    return 0 if _perform_enable(args.name, cfg) else 1
```
blackboxsw: Why do we want to switch to this call instead of calling the entitlement instance that we have locally in this function?

OddBloke: The code here was refactored in to `_perform_enable` so it could be used from both the new code in `action_attach` and in `action_enable` (which also does refreshing, and could feasibly also introduce extra output in the future that `action_attach` shouldn't output).

blackboxsw: Makes sense. Thanks

OddBloke: Just to round this out with our IRC conversation: I agree that `.enable(silent=True)` would be a reasonable path forward if we do end up wanting to suppress that output. During attach, however, we _do_ want the output that `.enable()` produces.

(This separate function would still be necessary so that we don't perform a spurious refresh (or two) during attach, and to avoid wasting effort by running through `assert_attached_root` multiple times.)

PR #414, opened 2019-04-18
Review fixes

panlinux:
Addresses review comments from https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1825239/comments/1

In the esm-is-already-enabled case, I opted to exit 1, which is documented in the manpage as:
```
1
Invalid command or option
```
Only in later releases did we add an exit code for `<service> is already enabled`, and that was 8.

enable-fips already checks for "fips is already enabled", and also exits non-zero (6 in this case).


panlinux:
Waiting for CI...

Issue #404, opened 2019-04-17
Install packaged gpg keys for cc, fips and fips-updates instead of using aptKey directive from ua-contracts service


blackboxsw:
Currentlly uaclient looks at contract service response for an aptKey directive. If present, uaclient ignores the packaged gpg key ring for any apt-based service.

Per security team suggestion it may be better to just use our packaged keys instead of relying on the external response from ua-contracts service.

The code to look at is entitlements/repo.py:enable which has a mutual exclusion if aptKey is present:
```
        ppa_fingerprint = directives.get('aptKey')
        if ppa_fingerprint:
            keyring_file = None
        else:
            keyring_file = os.path.join(apt.KEYRINGS_DIR, self.repo_key_file)
```

powersj:
In my mind a fix for this would involve:

- remove calls to apt-key (not considered stable interface and not something we should be using)
- ship the correct keys

However, we need to have a discussion about how we change keys. Using the entitlements would be one way to rotate keys.

PR #415, opened 2019-04-18
esm: add esm-v2 gpg keyring, drop old keyring, ignore aptKey directive

blackboxsw:
There are security concerns with honoring aptKey directive from
ua-contracts service.

ubuntu-advantage-tools will now:
 - install the specific entitlement keyring by default
 - log a debug message about skipping aptKey configuration directive
 - adjust postinst and postrm to install/remove v2 ringfile

There will be a followup discussion with security about the best
approach for using aptKey directive to rotate keys for each entitlement.

Fixes: #404


OddBloke:
Will this remove `ubuntu-esm-keyring.gpg` on upgrade? If not, we're leaving a key we don't intend to use on the system, which is an unnecessary attack vector.

debian/preinst
```diff
@@ -28,6 +28,8 @@ case "$1" in
                 rm -f $CACHE_DIR/$CACHE_FILE || true
             fi
         fi
+	# Remove old key on upgrade path
+        rm -f /etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg
```
OddBloke: Will this cause problems if the installation of a new version fails? Could we end up deleting `ubuntu-esm-keyring.gpg` but still be on the old version that requires it?

debian/postrm
```diff
@@ -4,41 +4,21 @@ set -e
...
-        rm "$esm_apt_source_file"
-    fi
-
-    esm_unattended_upgrades_file="/etc/apt/apt.conf.d/51ubuntu-advantage-esm"
-    if [ -f "$esm_unattended_upgrades_file" ]; then
-        rm "$esm_unattended_upgrades_file"
-    fi
+    rm -f /etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg
+    rm -f /etc/apt/sources.list.d/ubuntu-esm-trusty.list
+    rm -f /etc/apt/apt.conf.d/51ubuntu-advantage-esm
```
powersj: why are we removing all the tests?

debian/preinst
blackboxsw: per side conversation, I'll look at using dpkg-maintscript-helper's rm_conffile to handle this to avoid permanently deleting this file across failed package upgrade boundaries

debian/postrm
blackboxsw: I was removing the file tests for a few reasons:
- single operation rm -f which doesn't fail if file or directory does not exist
- less code, less vertical content to and scoping review
- checked with andreas on his preference, and he thought this option looked good.

debian/preinst
blackboxsw: Ok, I looked over using dpk-maintscript cm_conffile and TLDR, I don't think we can use it because the old deb doesn't actually deliver this file in it's catalog. it is a side-effect of an enable-esm operation. 
As a result, the upgrade path won't see /etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg as belonging to the upgraded deb package and so we can't rely on rm_conffile to remove it

blackboxsw: I think I'll leave this as is with an rm -f in preinst and we can chat about improvements for this next week with packaging folks.

OddBloke: I think we're better leaving the file in place than risking breaking people's installations. Can we file an issue for the follow-up proper fix and then drop this `rm` for now?

PR #418, opened 2019-04-19
Add _perform_enable tests and refactor UAEntitlement tests

OddBloke:
Also expand the `_perform_enable` documentation.

These changes will be prerequisite for my fix for #412, but can readily be reviewed/landed separately so I'm proposing them early.

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -191,6 +191,9 @@ def _perform_enable(entitlement_name: str, cfg: config.UAConfig) -> bool:
     (This helper excludes any messaging, so that different enablement code
     paths can message themselves.)
 
+    :param entitlement_name: the name of the entitlement to enable
+    :param cfg: the UAConfig to pass to the entitlement
+
     @return: True on success, False otherwise
     """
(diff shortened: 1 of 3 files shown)


uaclient/cli.py
```diff
@@ -191,6 +191,9 @@ def _perform_enable(entitlement_name: str, cfg: config.UAConfig) -> bool:
     (This helper excludes any messaging, so that different enablement code
     paths can message themselves.)
 
+    :param entitlement_name: the name of the entitlement to enable
```
blackboxsw: What markup should we be using for docstrings. I had been trying to follow epytext per http://epydoc.sourceforge.net/manual-epytext.html   but that's only because that's what landscape team did. Should we use a different docstring markup syntax in the event that we auto-generate docs at some point?

OddBloke: This is [the Sphinx syntax](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html?highlight=param#info-field-lists), which is what RTD uses for documentation generation.

(Until we're publishing API docs for others to consume, if ever, I don't think the exact structure matters too much; we'll almost always be consuming it in the source files.)

Issue #385, opened 2019-04-15
repos: refresh does not update service if aptURL or resourceToken changes


blackboxsw:
For apt-based services, aptURLs or resourceTokens could change for any number of reasons. UA Client should react to those changes upon refresh and update the apt-based service.

possible config changes that could affect availability of a repo in a given environment could be:
- resourceToken 
-  architectures
- aptURL
- suites
- aptKey
- additionalPackages
- series

blackboxsw:
David confirmed we want uaclient to react to refreshed changes on enabled entitlements.

PR #425, opened 2019-04-20
contract: process_entitlement_delta to react to apt directive deltas

blackboxsw:
Log and handle changes to resourceToken, aptURL and suites directives during a ua
refresh operation. When an active entitlement transitions to unentitled, that entitlement will be disabled.

Refactor Fips and ESMEntitlement.disable methods up into a common RepoEntitlement.disable() method which calls a new remove_apt_config() to manage removal of apt config files. RepoEntitlement.process_contract_deltas calls remove_apt_configto ensure stale apt config information is removed prior to writing updated apt config content.

While livepatch delta disable is handled, handling livepatch token and service configuration deltas will be handled in a separate issue.

Fixes: #385
Fixes: #436

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -185,7 +185,6 @@ class UAConfig:
         """Remove configuration cached response files class attributes."""
         for path_key in self.data_paths.keys():
             self.delete_cache_key(path_key)
-        self.flush_cache()
 
     def read_cache(self, key, silent=False):
         cache_path = self.data_path(key)
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -127,8 +127,10 @@ class UAConfig:
         ent_by_name = dict(
             (e['type'], e) for e in contractInfo['resourceEntitlements'])
         for entitlement_name, ent_value in ent_by_name.items():
-            entitlement_cfg = self.read_cache(
-                'machine-access-%s' % entitlement_name, silent=True)
+            entitlement_cfg = {}
+            if ent_value.get('entitled'):
+                entitlement_cfg = self.read_cache(
+                    'machine-access-%s' % entitlement_name, silent=True)
             if not entitlement_cfg:
                 # Fallback to machine-token info on unentitled
                 entitlement_cfg = {'entitlement': ent_value}
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -170,12 +170,11 @@ class UAConfig:
         if not key:
             raise RuntimeError(
                 'Invalid or empty key provided to delete_cache_key')
-        if key.startswith('machine-access'):
+        if key.startswith('machine-access') or key == 'machine-token':
             self._entitlements = None
+            self._machine_token = None
         elif key == 'account-contracts':
             self._contracts = None
-        elif key == 'machine-token':
-            self._machine_token = None
         for private in (True, False):
             cache_path = self.data_path(key, private)
             if os.path.exists(cache_path):
@@ -206,6 +205,11 @@ class UAConfig:
         data_dir = os.path.dirname(filepath)
         if not os.path.exists(data_dir):
             os.makedirs(data_dir)
+        if key.startswith('machine-access') or key == 'machine-token':
+            self._machine_token = None
+            self._entitlements = None
+        elif key == 'account-contracts':
+            self._contracts = None
         if not isinstance(content, str):
             content = json.dumps(content)
         if private:
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -159,12 +159,6 @@ class UAConfig:
             return os.path.join(data_dir, self.data_paths[key])
         return os.path.join(data_dir, key)
 
-    def flush_cache(self):
-        """Clear any cached locals so new cache values are read."""
-        self._contracts = None
-        self._entitlements = None
-        self._machine_token = None
-
     def delete_cache_key(self, key):
         """Remove specific cache file."""
         if not key:
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -159,6 +159,12 @@ class UAConfig:
             return os.path.join(data_dir, self.data_paths[key])
         return os.path.join(data_dir, key)
 
+    def flush_cache(self):
+        """Clear any cached locals so new cache values are read."""
+        self._contracts = None
+        self._entitlements = None
+        self._machine_token = None
+
     def delete_cache_key(self, key):
         """Remove specific cache file."""
         if not key:
@@ -179,6 +185,7 @@ class UAConfig:
         """Remove configuration cached response files class attributes."""
         for path_key in self.data_paths.keys():
             self.delete_cache_key(path_key)
+        self.flush_cache()
 
     def read_cache(self, key, silent=False):
         cache_path = self.data_path(key)
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -1,5 +1,4 @@
 import logging
-import os
 
 from uaclient import serviceclient
 from uaclient import util
@@ -8,7 +7,6 @@ from uaclient import util
 API_PATH_TMPL_ACCOUNT_USERS = '/accounts/{account}/users'
 API_PATH_TMPL_CONTRACT_MACHINES = '/contracts/{contract}/context/machines'
 API_PATH_TMPL_MACHINE_CONTRACT = '/machines/{machine}/contract'
-API_PATH_TMPL_RESOURCE_MACHINE_ACCESS = '/resources/{resource}/context/machine'
 
 API_V1_ACCOUNTS = '/v1/accounts'
 API_V1_TMPL_ACCOUNT_CONTRACTS = '/v1/accounts/{account}/contracts'
@@ -246,6 +244,52 @@ def get_contract_token_for_account(contract_client, macaroon, account_id):
     return contract_token_response['contractToken']
 
 
+def get_dict_deltas(orig_dict, new_dict, path=''):
+    """Return a dictionary of delta between orig_dict and new_dict."""
+    deltas = {}
+    for key, value in orig_dict.items():
+        if isinstance(value, dict):
+            if path:
+                sub_path = path + '.' + key
+            else:
+                sub_path = key
+            sub_delta = get_dict_deltas(
+                value, new_dict.get(key, {}), path=sub_path)
+            if sub_delta:
+                deltas[key] = sub_delta
+        elif value != new_dict.get(key):
+            key_path = key if not path else path + '.' + key
+            logging.debug(
+                "Contract value for '%s' changed to '%s'",
+                key_path, new_dict.get(key))
+            deltas[key] = new_dict.get(key)
+    for key, value in new_dict.items():
+        if key not in orig_dict:
+            deltas[key] = value
+    return deltas
+
+
+def process_entitlement_delta(orig_access, new_access):
+    """Process a entitlement access dictionary deltas if they exist.
+
+    :param orig_access: Dict with original entitlement access details before
+        contract refresh deltas
+    :param orig_access: Dict with updated entitlement access details after
+        contract refresh
+    """
+    from uaclient import entitlements
+
+    if not orig_access or orig_access == new_access:
+        return {}
+    deltas = get_dict_deltas(orig_access, new_access)
+    if deltas:
+        name = orig_access['entitlement']['type']
+        ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[name]
+        entitlement = ent_cls()
+        entitlement.process_contract_deltas(orig_access, deltas)
+    return deltas
+
+
 def request_updated_contract(cfg, contract_token=None):
     """Request contract refresh from ua-contracts service.
 
@@ -255,13 +299,12 @@ def request_updated_contract(cfg, contract_token=None):
 
     @return: True on success False otherwise.
     """
-    contract_client = UAContractClient(cfg)
     orig_token = cfg.machine_token
     if orig_token and contract_token:
         raise RuntimeError(
             'Got unexpected contract_token on an already attached machine')
+    contract_client = UAContractClient(cfg)
     if contract_token:  # We are a mid ua-attach and need to get machinetoken
-        refresh_operation = False
         try:
             new_token = contract_client.request_contract_machine_attach(
                 contract_token=contract_token)
@@ -270,7 +313,6 @@ def request_updated_contract(cfg, contract_token=None):
                 'Could not obtain machine token. %s', str(e))
             return False
     else:
-        refresh_operation = True
         machine_token = orig_token['machineToken']
         contract_id = orig_token['machineTokenInfo']['contractInfo']['id']
         try:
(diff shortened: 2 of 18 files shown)


uaclient/contract.py
```diff
@@ -246,6 +244,92 @@ def get_contract_token_for_account(contract_client, macaroon, account_id):
...
 
 
+def process_repo_entitlement_deltas(orig_access, deltas) -> None:
+    """Process repo-based entitlement deltas.
+
+    :param orig_access: Dict with original entitlement access details before
+        refresh deltas.
+    :param deltas: Dict containing discovered entitlement access deltas.
+    """
+    from uaclient import entitlements
```
OddBloke: These imports and the `issubclass` check below suggest to me that this should be a method on `RepoEntitlement`; does that make sense?

uaclient/contract.py
```diff
@@ -246,6 +244,92 @@ def get_contract_token_for_account(contract_client, macaroon, account_id):
...
+    orig_affordances = orig_access['entitlement'].get('affordances', {})
+    for affordance, value in orig_affordances.items():
+        if value != new_affordances.get(affordance):
+            logging.debug(
+                "Affordance value for '%s.%s' changed to '%s'",
+                name, affordance, new_affordances.get(affordance))
+            if 'affordances' not in deltas:
+                deltas['affordances'] = {}
+            deltas['affordances'][affordance] = new_affordances.get(affordance)
+    new_directives = new_entitlement.get('directives', {})
```
OddBloke: The entitlements code and the directives code are almost identical; could it be refactored in to a single helper that we call twice from this function?

uaclient/contract.py
powersj: I think this is comment is still outstanding

uaclient/entitlements/repo.py
```diff
@@ -144,6 +151,43 @@ def operational_status(self):
...
+        """Process any contract access deltas for this entitlement.
+
+        :param orig_access: Dictionary containing the original
+            resourceEntitlement access details.
+        :param deltas: Dictionary which contains only the changed access keys
+        and values.
+        """
+        if not deltas:
+            return
+        super().process_contract_deltas(orig_access, deltas)
```
OddBloke: It looks to me like this super-call could end up disabling the entire entitlement, in which case there isn't much point doing anything else. I don't have a particularly good idea of how to short-circuit here, but is it something we should consider?

uaclient/util.py
```diff
@@ -61,6 +62,31 @@ def encode_text(text, encoding='utf-8'):
...
+            if path:
+                sub_path = path + '.' + key
+            else:
+                sub_path = key
+            sub_delta = get_dict_deltas(
+                value, new_dict.get(key, {}), path=sub_path)
+            if sub_delta:
+                deltas[key] = sub_delta
+        elif value != new_value:
+            key_path = key if not path else path + '.' + key
```
OddBloke: This is the same logic as `sub_path` in the other branch of this if; can we refactor it out so we only calculate it once?

uaclient/util.py
```diff
@@ -61,6 +62,31 @@ def encode_text(text, encoding='utf-8'):
...
+    deltas = {}
+    for key, value in orig_dict.items():
+        new_value = new_dict.get(key, DROPPED_DICT_KEY)
+        if isinstance(value, dict):
+            if path:
+                sub_path = path + '.' + key
+            else:
+                sub_path = key
+            sub_delta = get_dict_deltas(
+                value, new_dict.get(key, {}), path=sub_path)
```
OddBloke: If `new_dict` doesn't have `key`, do we need to recurse to determine the outcome here? Doesn't this means it should be a `DROPPED_DICT_KEY`?

OddBloke: I think this is also causing a logical error ATM:

```py
>>> from uaclient.util import get_dict_deltas

>>> get_dict_deltas({}, {})
{}

>>> get_dict_deltas({'a': {}}, {})
{}

>>> get_dict_deltas({'a': ''}, {})
Out[6]: {'a': 'DROPPED'}
```

uaclient/entitlements/repo.py
```diff
@@ -144,6 +151,43 @@ def operational_status(self):
...
+        creds = 'bearer:%s' % new_token
+        if new_url:
+            # Remove original aptURL and auth and rewrite
+            apt.remove_auth_apt_repo(
+                repo_filename, orig_access.get('directives', {}).get('aptURL'))
+        else:
+            new_url = orig_access.get('directives', {}).get('aptURL')
+        if not new_suites:
+            new_suites = orig_access.get('directives', {}).get('suites', [])
+        apt.add_auth_apt_repo(repo_filename, new_url, creds, suites=new_suites)
```
OddBloke: In `enable`, we wrap this in error handling; should we be doing the same here?

uaclient/entitlements/repo.py
```diff
@@ -144,6 +151,43 @@ def operational_status(self):
...
+            name=self.name, series=series)
+        if not new_token:
+            new_token = orig_access['resourceToken']
+        creds = 'bearer:%s' % new_token
+        if new_url:
+            # Remove original aptURL and auth and rewrite
+            apt.remove_auth_apt_repo(
+                repo_filename, orig_access.get('directives', {}).get('aptURL'))
+        else:
+            new_url = orig_access.get('directives', {}).get('aptURL')
```
OddBloke: In `enable` we default to `self.repo_url` which we aren't doing here.

uaclient/entitlements/repo.py
OddBloke: After we configure the new repo in `enable`, we call `apt-get update` so that the new repo is immediately available for use. Should we do the same here?

uaclient/entitlements/repo.py
```diff
@@ -144,6 +151,43 @@ def operational_status(self):
             return status.ACTIVE, '%s is active' % self.title
         return status.INACTIVE, '%s is not configured' % self.title
 
+    def process_contract_deltas(
```
OddBloke: I have a number of comments in here relating to behavioural differences from `enable`. To avoid always having to manually reconcile changes to the two, I wonder if we should do some refactoring so that `enable` and `process_contract_deltas` use the same code?

(A more radical idea would be to refactor things entirely, so that "enable" is really just processing a delta from {} to our initial entitlement.)

What do you think?

uaclient/entitlements/base.py
```diff
@@ -194,6 +200,30 @@ def is_access_expired(self):
...
+        :param orig_access: Dictionary containing the original
+            resourceEntitlement access details.
+        :param deltas: Dictionary which contains only the changed access keys
+        and values.
+        """
+        if not deltas:
+            return
+        entitled_delta = deltas.get('entitlement', {}).get('entitled')
+        if orig_access and entitled_delta in (util.DROPPED_DICT_KEY, False):
+            if self.can_disable(silent=True):
```
OddBloke: Should we do anything if we _can't_ disable? It might be helpful to inform the user that their status has changed? I'm not sure.

uaclient/config.py
```diff
@@ -179,6 +185,7 @@ def delete_cache(self):
         """Remove configuration cached response files class attributes."""
         for path_key in self.data_paths.keys():
             self.delete_cache_key(path_key)
+        self.flush_cache()
```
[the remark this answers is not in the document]
blackboxsw: agreed, I was debating about removing the side-effect inside delete_cache_key, originally and realized we still need to specific clearing of local instance attributes inside delete_cache_key anyway. I'll  revert this.

uaclient/entitlements/repo.py
blackboxsw: (A more radical idea would be to refactor things entirely, so that "enable" is really just processing a delta from {} to our initial entitlement.) 

I was leaning toward this suggestion and couldn't decide whether or not to do that in this branch.
I left the comment in the code to imply that we should be doing just that. # TODO: handle initial install deltas where orig_access is empty

I was uncertain whether or not to bite off that chunk of work in this branch, but given your inline comments, this branch will end up ultimately duplicating a lot of that common logic anyway to ensure the delta process behaves the same as green-field enable(). I'll refactor to use the same logic so we don't have to deconcile across different logical paths.

uaclient/entitlements/repo.py
blackboxsw: We now call enable in process_contract_deltas so logic will be the same.

blackboxsw: We will call enable now from process_contract_deltas, so that we can reuse this logic.

uaclient/entitlements/repo.py
blackboxsw: Dan, let me know what you think here. I've changed the operation to return True if it has already performed operations. repo.process_contract_deltas will exit early of True returned from the super call.

uaclient/entitlements/repo.py
blackboxsw: + 1 we are calling enable directly now, the only separate action that process_contract_deltas performs directly is the cleanup/removal or old_url related-content from apt config files. And calling enable directly to do all the setup/writing of updated apt config files

uaclient/entitlements/base.py
blackboxsw: This check serves two purposes:
1. don't try to disable something upon transition to unentitled if it isn't currently enabled
2. don't disable fips* entitlements which don't have a disable option.

I'll add an else here that also checks operational_status == ACTIVE to print an info message saying why it can't be disabled.

uaclient/entitlements/base.py
```diff
@@ -194,6 +200,30 @@ def is_access_expired(self):
...
+
+        :param orig_access: Dictionary containing the original
+            resourceEntitlement access details.
+        :param deltas: Dictionary which contains only the changed access keys
+        and values.
+        """
+        if not deltas:
+            return
+        entitled_delta = deltas.get('entitlement', {}).get('entitled')
+        if orig_access and entitled_delta in (util.DROPPED_DICT_KEY, False):
```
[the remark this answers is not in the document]
blackboxsw: I've reworked this logic to check that 
orig_access is non-empty and deltas has an 'entitled' key with a value (representing a change in entitlement) and that entitled is either DROPPED_KEY or  False. The entitled value will only be None if ua-contracts provides an 'entitlement': null in their response.

blackboxsw:
To test:
follow these steps to setup a local ua-contract service:

#setup a local ua-contracts container per 
https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/195

create an bootstrapped admin user with password and manipulate contract entitlements between ua refresh calls
https://pastebin.canonical.com/p/23N7ZK5stT/

uaclient/util.py
```diff
@@ -12,6 +12,7 @@
 
 ETC_MACHINE_ID = '/etc/machine-id'
 DBUS_MACHINE_ID = '/var/lib/dbus/machine-id'
+DROPPED_DICT_KEY = 'DROPPED'
```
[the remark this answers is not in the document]
blackboxsw: Good suggestion and using object() allowed me to see an error in handling of said deltas.u thanks.

uaclient/config.py
```diff
@@ -159,6 +159,12 @@ def data_path(self, key=None, private=True):
             return os.path.join(data_dir, self.data_paths[key])
         return os.path.join(data_dir, key)
 
+    def flush_cache(self):
```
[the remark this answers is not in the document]
blackboxsw: I went this direction, write_cache will now unset UAConfig locals _entitlement, _machine_token etc when written. It changed some of the logic that we relying on the stale in-memory cache, but I admin it is much cleaner to just force dirty reads instead of relying on the stale config cache.

uaclient/entitlements/tests/test_base.py
```diff
@@ -78,17 +87,11 @@ def test_init_accepts_a_uaconfig(self):
 
     def test_can_disable_false_on_unentitled(
```
OddBloke: I think this test name needs updating.

uaclient/tests/test_contract.py
```diff
@@ -1,13 +1,54 @@
...
+M_REPO_PATH = 'uaclient.entitlements.repo.RepoEntitlement.'
+
+
+class TestProcessEntitlementDeltas:
+
+    def test_error_on_missing_entitlement_type(self):
+        """Raise an error when neither dict contains entitlement type."""
+        error_msg = ('Could not determine contract delta service type %s %s' %
+                     ({}, {'something': 'non-empty'}))
+        with pytest.raises(RuntimeError, match=error_msg):
```
OddBloke: TIL about the match kwarg to raises, thanks!

Unfortunately, it looks like it's only been in pytest since 3.1.0: https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst#310-2017-05-22

This isn't showing up as an error because older versions of `pytest.raises` will silently ignore any kwargs they don't understand (this is on trusty):

```py
>>> import pytest
>>> pytest.raises()
  File "<stdin>", line 1, in <module>
TypeError: raises() missing 1 required positional argument: 'ExpectedException'
>>> pytest.raises('foo')
<_pytest.python.RaisesContext object at 0x7fb1765b6eb8>
>>> pytest.raises('foo', bar='baz')
<_pytest.python.RaisesContext object at 0x7fb1755e3128>
```

So we're silently not actually asserting anything about the error message on older versions of pytest.

uaclient/util.py
OddBloke: @blackboxsw This comment appears unaddressed (as does the one below it):

```py
In [2]: get_dict_deltas({'a': ''}, {})
Out[2]: {'a': <object at 0x7fc624a9c660>}

In [3]: get_dict_deltas({'a': {}}, {})
Out[3]: {}
```

OddBloke:
@blackboxsw I don't believe this fixes #436. This PR only covers disabling entitlements when they move from entitled to unentitled, not from applicable to inapplicable.

uaclient/entitlements/esm.py
```diff
@@ -16,31 +11,4 @@ class ESMEntitlement(repo.RepoEntitlement):
         ' (https://ubuntu.com/esm)')
     repo_url = 'https://esm.ubuntu.com'
     repo_key_file = 'ubuntu-esm-v2-keyring.gpg'
-
-    def disable(self, silent=False, force=False):
```
blackboxsw: Moved ESMEntitlement.disable and FIPSEntitlement.disable to common parent RepoEntitlement.disable.

uaclient/entitlements/fips.py
```diff
@@ -12,39 +9,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                 'libssl1.0.0-hmac', 'linux-fips', 'strongswan-hmac',
                 'openssh-client', 'openssh-server', 'openssl', 'libssl1.0.0',
                 'fips-initramfs', 'strongswan']
-
-    def disable(self, silent=False, force=False):
```
blackboxsw: Moved ESMEntitlement.disable and FIPSEntitlement.disable to common parent RepoEntitlement.disable.

uaclient/util.py
blackboxsw: Just addressed dropped keys of type dict as value in commit 6bc337c8b611d2fa0ab49fd8948e8b59dcf92415

uaclient/util.py
blackboxsw: refactored logic to calculate key_path in one place.

uaclient/entitlements/base.py
```diff
@@ -194,6 +198,44 @@ def is_access_expired(self):
             return False
         return True
 
+    def process_contract_deltas(
+            self, orig_access: 'Dict[str, Any]',
+            deltas: 'Dict[str, Any]') -> None:
```
OddBloke: This returns a `bool`

Issue #377, opened 2019-04-15
`ua` doesn't detect repo entitlements as active until an apt update has happened


OddBloke:
To set up this reproducer, I launched a lxd container, enabled proposed and the daily PPA, and installed ubuntu-advantage-tools 19.4\~406\~git.fb8f7cd. I then copied the staging configuration from the repo to `conf`.

```
# UA_CONFIG_FILE=conf ua status
This machine is not attached to a UA subscription.
See `ua attach` or https://ubuntu.com/advantage

# UA_CONFIG_FILE=conf ua attach
Email: daniel.watkins@canonical.com
Password: 
Second-factor auth: <snip>
This machine is now attached to 'daniel.watkins@canonical.com'.

Account: daniel.watkins@canonical.com
Subscription: daniel.watkins@canonical.com
Valid until: 0001-01-01
Technical support level: not included

cc            entitled         n/a
cis-audit     none             inactive
esm           entitled         inactive
fips          entitled         n/a
fips-updates  entitled         n/a
livepatch     entitled         n/a

Enable entitlements with `ua enable <service>`

# UA_CONFIG_FILE=conf ua enable esm
Extended Security Maintenance enabled.
# UA_CONFIG_FILE=conf ua status
Account: daniel.watkins@canonical.com
Subscription: daniel.watkins@canonical.com
Valid until: 0001-01-01
Technical support level: not included

cc            entitled         n/a
cis-audit     none             inactive
esm           entitled         inactive
fips          entitled         n/a
fips-updates  entitled         n/a
livepatch     entitled         n/a

Enable entitlements with `ua enable <service>`
```

At this point, ESM should be displayed as `active`, but it isn't until I do an `apt update`:

```
# apt update
<snip>
Fetched 66.4 kB in 2s (26.1 kB/s)
Reading package lists... Done

# UA_CONFIG_FILE=conf ua status
Account: daniel.watkins@canonical.com
Subscription: daniel.watkins@canonical.com
Valid until: 0001-01-01
Technical support level: not included

cc            entitled         n/a
cis-audit     none             inactive
esm           entitled         active
fips          entitled         n/a
fips-updates  entitled         n/a
livepatch     entitled         n/a

Enable entitlements with `ua enable <service>`
```

OddBloke:
This also means that the following doesn't work:

```
$ ua attach
...
$ ua enable esm
...
$ ua disable esm
Extended Security Maintenance is not currently enabled.
[...]

PR #427, opened 2019-04-22
esm: all repo entitlements now call apt-get update on enable

blackboxsw:
Expectation is that a machine that is enabling a service would need to
access that services packages from the repo they just enabled.

Fixes: #377


blackboxsw:
To test:
make deb
lxc launch ubuntu-daily:trusty t1
lxc file push ubuntu-advantage-tools_all.deb t1/
lxc exec t1 bash
t1 $  dpkg -i /ubuntu-advantage-tools_all.deb
t1 $ ua attach
t1 $ ua enable esm
t1 $ ua status   # should show active

PR #430, opened 2019-04-22
Fix/empty config values

### LINKED ISSUE #406: Traceback if a config file without a `log_level` used
With this config file:

```yaml
# Development UA Client config file. YAML
sso_auth_url: 'https://login.ubuntu.com'
contract_url: 'https://contracts.staging.canonical.com'
data_dir: /tmp/uaclient
log_file: ubuntu-advantage-devel.log
```

I see this traceback:

```pytb
$ UA_CONFIG_FILE=conf2 ua attach
    load_entry_point('ubuntu-advantage-tools==19.5', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 396, in main
    cfg = config.UAConfig()
  File "/usr/lib/python3/dist-packages/uaclient/config.py", line 54, in __init__
    self.cfg = parse_config()
  File "/usr/lib/python3/dist-packages/uaclient/config.py", line 271, in parse_config
    cfg['log_level'] = cfg['log_level'].upper()
AttributeError: 'int' object has no attribute 'upper'
```

This also occurs if `UA_CONFIG_FILE` points at a non-existent config file (i.e. `$ UA_CONFIG_FILE=nope ua attach`).

powersj:
The default value for logging when it was missing was becoming an
integer (logging.INFO == 20) and then being passed around between
functions that expected a string value for the level.

Rather than storing logging.INFO as the default value for logging, use
the same value/string that we print to the file, namely 'INFO' and then
do the right thing with the string.

Fixes #406

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -399,7 +399,7 @@ def main(sys_argv=None):
     args = parser.parse_args(args=cli_arguments)
     cfg = config.UAConfig()
     try:
-        log_level = int(cfg.log_level)
+        log_level = cfg.log_level
     except TypeError:
         log_level = getattr(logging, '%s' % cfg.log_file.upper())
     console_level = logging.DEBUG if args.debug else logging.INFO
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -398,10 +398,7 @@ def main(sys_argv=None):
         sys.exit(1)
     args = parser.parse_args(args=cli_arguments)
     cfg = config.UAConfig()
-    try:
-        log_level = cfg.log_level
-    except TypeError:
-        log_level = getattr(logging, '%s' % cfg.log_file.upper())
+    log_level = cfg.log_level
     console_level = logging.DEBUG if args.debug else logging.INFO
     setup_logging(console_level, log_level, cfg.log_file)
     return args.action(args, cfg)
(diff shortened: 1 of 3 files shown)


uaclient/cli.py
```diff
@@ -399,7 +399,7 @@ def main(sys_argv=None):
     args = parser.parse_args(args=cli_arguments)
     cfg = config.UAConfig()
     try:
-        log_level = int(cfg.log_level)
+        log_level = cfg.log_level
```
OddBloke: If we're no longer calling `int()`, nothing here should be raising a `TypeError`; can we drop the try/except here too? (And, presumably, just refer to `cfg.log_level` directly where we're currently using `log_level` in this function?)

powersj: I can catch that, if I catch something like this I was going to suggest using the default value (info). does that work?

OddBloke: So you're suggesting that these lines in `setup_logging`: https://github.com/CanonicalLtd/ubuntu-advantage-client/blob/e74b4c1ad81b1c77a709459d2581a85e1a620c64/uaclient/cli.py#L370-L371 will change to fallback to INFO if we get a `ValueError` from `setLevel`?

That sounds reasonable to me.

powersj:
ready for re-review

Issue #440, opened 2019-04-24
`ua enable fips*` should install linux-fips and upgrade any FIPS packages that are already installed


OddBloke:
(Rough TL;DR: We should replicate the behaviour that Vineetha implemented for the old client in https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/157/commits/dc43d8dbfb70c7a21099e26ac4a908cdaae72109)

I just had a conversation with security folks regarding precisely what packages we need to install on FIPS enablement. The summary is that on enable we should do the following:

1. Install `linux-fips`
1. For each FIPS package, if it is installed on the system, upgrade it (and if a corresponding `$package-hmac` exists for it, ensure we install that too)
    * The existence of `$package-hmac` doesn't need to be determined dynamically; we know which ones exist, and should make sure we install them explicitly we find that their sibling package is installed

Once this change is made, we should confirm that, post-enable, packages that weren't installed during `enable` will be pulled from the FIPS PPA; this shouldn't represent a change in behaviour from the current code.

(Side note: we concluded that we should continue to prohibit users from running `ua disable fips`; it remains the case that removing FIPS from a system isn't supported.)

OddBloke:
In a future re-certification of xenial (and the in-progress certification for bionic), the packages which have a corresponding `-hmac` package have had a Recommends or Depends added so that the `-hmac` package will be pulled in by apt. Until the xenial re-certification is complete, we:

(a) still need to handle installing the appropriate -hmac packages during enable, and
(b) should consider emitting a message so that users know that they will need to manually install the `-hmac` packages if needed.

(Vineetha has confirmed on IRC that once the recertification is complete, newly-enabled systems will use the new packages; we will be able to simplify our handling once that is the case.)

PR #448, opened 2019-04-30
test_fips: refactoring

OddBloke:
This is preparatory refactoring work for #440. The next step will be to parametrize the entitlement fixture so that we're applying all of these tests to both FIPS and FIPSUpdates, and then I'll write tests for the existing installation behaviour.

I'm submitting now because (a) I'm about to finish for the day, and (b) this is self-contained work that can land by itself.

### DIFF
--- a/uaclient/entitlements/tests/test_fips.py
+++ b/uaclient/entitlements/tests/test_fips.py
@@ -60,15 +60,8 @@ def entitlement(tmpdir):
 
 class TestFIPSEntitlementCanEnable:
 
-    def test_can_enable_true_on_entitlement_inactive(self, tmpdir):
+    def test_can_enable_true_on_entitlement_inactive(self, entitlement):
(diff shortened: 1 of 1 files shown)


uaclient/entitlements/tests/test_fips.py
```diff
@@ -60,36 +60,25 @@ def entitlement(tmpdir):
...
-        entitlement.static_affordances = ()
         with mock.patch('sys.stdout', new_callable=StringIO) as m_stdout:
             assert True is entitlement.can_enable()
         assert '' == m_stdout.getvalue()
 
 
 class TestFIPSEntitlementEnable:
 
-    def test_enable_configures_apt_sources_and_auth_files(self, tmpdir):
+    @mock.patch(M_PATH + 'util.is_container', return_value=False)
```
blackboxsw: I don't think we need this mock here since the test mocks can_enable below inline (which is what called is_container)

Issue #449, opened 2019-05-01
trusty: fresh ua attach results in traceback on enableByDefault: True


blackboxsw:
`process_contract_deltas` results in a traceback on trusty for esm auto-enablement due to an attempted run of  `remove_auth_apt_repo` passing in a None value for aptURL.

```pytb
Traceback (most recent call last):
  File "/usr/bin/ua", line 11, in <module>
    load_entry_point('ubuntu-advantage-tools==19.5', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 408, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 266, in action_attach
    if not contract.request_updated_contract(cfg, contract_token):
  File "/usr/lib/python3/dist-packages/uaclient/contract.py", line 320, in request_updated_contract
    orig_entitlements.get(name, {}), new_access)
  File "/usr/lib/python3/dist-packages/uaclient/contract.py", line 274, in process_entitlement_delta
    entitlement.process_contract_deltas(orig_access, deltas)
  File "/usr/lib/python3/dist-packages/uaclient/entitlements/repo.py", line 166, in process_contract_deltas
    apt.remove_auth_apt_repo(repo_filename, old_url)
  File "/usr/lib/python3/dist-packages/uaclient/apt.py", line 162, in remove_auth_apt_repo
    remove_repo_from_apt_auth_file(repo_url)
  File "/usr/lib/python3/dist-packages/uaclient/apt.py", line 137, in remove_repo_from_apt_auth_file
    _protocol, repo_path = repo_url.split('://')
AttributeError: 'NoneType' object has no attribute 'split'
```

PR #450, opened 2019-05-01
bug: do not try remove_apt_auth_apt_repo if old_url is None

blackboxsw:
On fresh ua attach deltas, old aptURL does not exist.
    
process_contract_deltas will not attempt apt_auth_apt_repo in this case.
Drive by add unit test for contract ordering.
    
Fixes: #449

### DIFF
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -158,12 +158,14 @@ class RepoEntitlement(base.UAEntitlement):
                 name=self.name)
             logging.info(msg)
         if delta_entitlement.get('directives', {}).get('aptURL'):
-            # Remove original aptURL and auth and rewrite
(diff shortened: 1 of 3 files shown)


blackboxsw:
Open question for product:

**Product owner question**: If someone uses the client to manually ua disable X, and backend transitions X from enableByDefault: False -> True. This branch will automatically re-enable service X on next call to ua enable anything or ua refresh. This behavior could be undesired for existing machines where the disable operation was called. Is this a valid use-case to be concerned with?

OddBloke:
@blackboxsw Are the fixes for #449 and #452 tied together in this PR because they're more easily fixed together? If not, could we split this in to two PRs, one for each fix? I'm struggling to work out what's motivating some of these changes, because there are a couple of different things happening.

Issue #452, opened 2019-05-01
manually disabled services could be automatically re-enabled when processing contract deltas


blackboxsw:
During `ua refresh`, `ua enable` and `ua attach` operations, ua client processes any contract deltas and enables any inactive services that represent a transition to enableByDefault: True. 

There are a couple of corner cases which may neglect that service was previsouly manually disabled by a user via ua disable:

1. person tried a not auto-enabled service and didn't want it running.
```
ua attach
ua enable X
ua disable X
# ua-contracts changes X to be enableByDefault: True
ua refresh (X gets re-enabled)
```

2. person manually disabled an auto-enabled service and detach/reattached
```
ua attach
ua disable livepatch
ua detach
ua attach       # livepatch re-enabled
```
3. if ua client refreshes get performed daily auto-enabled service gets disabled by the user, backend transitions enableByDefault -> False; enableByDefault ->True between refreshes
```
ua attach
# backend livepatch transition enableByDefault -> False
ua refresh
# backend livepatch transition enableByDefault -> True
ua refresh   # or ua enable of a different service (which triggers a refresh)
```

4. Backend introduces and new auto-enabled service, existing ua clients would auto-enable without customer intervention/agreement/knowledge.

```
ua attach
# backend adds new support service with enableByDefault:True 
ua refresh
# ua client sees 'new support service' enableByDefault == True and auto-enables it
```

## Decision

Due to discussion below, scenario 1-3 will be rejected and scenario 4 will be addressed In the following manner:
* ua attach operation will be the only operation that will auto-enable services
* ua refresh operation will print a message to the user about enabling the new recommended service

powersj:
Scenario 1: There is a very low chance of this occurring. Not concerned about it as we do not have a huge list of services that are enabled by default.
Scenario 2: If someone re-attaches then they should expect things to be re-enabled by default. Not an issue.
Scenario 3: Similar to 1, not considering an issue.

OddBloke:
I think using livepatch (or any existing entitlement) as a concrete example minimises the potential impact here. My concern is this: if we were to introduce a completely new entitlement that we wanted enabled by default, the current client code would enable that entitlement on any UA attached machine next time we enabled or refreshed.

[...]

Issue #340, opened 2019-04-10
Make `caplog_text` log level capturing consistent across Ubuntu versions


OddBloke:
On bionic, we see DEBUG log messages in `caplog_text` which we don't see elsewhere. We should ensure that we are configuring our log capturing consistently, and then remove the workaround/TODO comment introduced in #277.

OddBloke:
This is done, but we have some tests that are still working around the old behaviour which can be simplified.

OddBloke:
I've opened up #527 for the remaining clean up, but I think we can close this out now.

PR #457, opened 2019-05-02
Feature/delta add livepatch auto enable

blackboxsw:
Relocate enableByDefault processing up from RepoEntitlement into UAEntitltment so logic can be shared with livepatch.

Drop batch enableByDefault processing from cli.py because individual entitlements will handle this in  process_contract_deltas.

Add log-olny option via allow_enable=False to process_contract_deltas which will only log that a new default entitlement was seen and how to manually enable it. Only ua attach will provide allow_enable=True which will automatically call UAEntitlement.enable () if enableByDefault is True in the contract

Fixes: #452


uaclient/cli.py
```diff
@@ -263,7 +263,8 @@ def action_attach(args, cfg):
     if not contract_token:
         print('No valid contract token available')
         return 1
-    if not contract.request_updated_contract(cfg, contract_token):
+    if not contract.request_updated_contract(
+            cfg, contract_token, allow_enable=True):
```
blackboxsw: ua enable is the only path that allows UAEntitlement.enable() call if the procesed deltas contain an enableByDefault: True obligation.

uaclient/cli.py
```diff
@@ -273,17 +274,6 @@ def action_attach(args, cfg):
         ua_status.MESSAGE_ATTACH_SUCCESS_TMPL.format(
             contract_name=contract_name))
 
-    # sort the list, so that we get consistent attach behaviour regardless of
```
blackboxsw: Dropped this because individual services will be handling enableByDefault during process_contract_deltas.

uaclient/entitlements/base.py
```diff
@@ -245,6 +251,26 @@ def process_contract_deltas(
             # responses on unentitled services.
             self.cfg.delete_cache_key('machine-access-%s' % self.name)
             return True
+
+        resourceToken = orig_access.get('resourceToken')
```
blackboxsw: Shuffled this up into base and out of RepoEntitlement so that livepatch can be auto-enabled by process_contract_deltas.

uaclient/conftest.py
```diff
@@ -33,12 +33,14 @@ def caplog_text(request):
             # Older versions of pytest only have getfuncargvalue, which is now
             # deprecated in favour of getfixturevalue
             caplog = request.getfuncargvalue('caplog')
+        caplog.set_level(logging.INFO)
```
OddBloke: I think the changes to this file fix #340?

(For reference, in future, ideally I'd like to see changes like this in a separate commit. It makes it easier to understand what the motivation for a change was if it has a more specific commit message, and it also means that we won't regress #340 if we ever wanted to revert the commit that introduces these changes. To be clear, I *don't* think it's worth splitting it in to a separate commit in this PR.)

uaclient/entitlements/base.py
```diff
@@ -211,22 +211,28 @@ def is_access_expired(self):
...
         """
         if not deltas:
             return True  # We processed all deltas that needed processing
-        transition_to_unentitled = False
-        if orig_access and 'entitled' in deltas.get('entitlement', {}):
-            transition_to_unentitled = (
-                deltas['entitlement']['entitled'] in (False, util.DROPPED_KEY))
+
+        delta_entitlement = deltas.get('entitlement', {})
+        transition_to_unentitled = bool(delta_entitlement == util.DROPPED_KEY)
```
OddBloke: No need for the `bool` call here; `==` is going to return a `bool` (unless someone does something _very_ weird in a custom `__eq__` method).

uaclient/status.py
```diff
@@ -72,8 +72,9 @@ class TxtColor:
 This machine is now attached to '{contract_name}'.
 """
 
-MESSAGE_ATTACH_ENABLING_DEFAULTS = 'Enabling default entitlements...'
-MESSAGE_ENABLE_BY_DEFAULT_TMPL = 'Enabling default entitlement {name}'
+MESSAGE_ENABLE_BY_DEFAULT_TMPL = 'Enabling default service {name}'
+MESSAGE_ENABLE_BY_DEFAULT_MANUAL_TMPL = """\
+Service {name} is recommended by default. To enable run `ua enable {name}`"""
```
OddBloke: I'd put a comma after enable here: "To enable, run \`...`"

PR #458, opened 2019-05-02
fips: only install/upgrade optional packages that are already on the system

### LINKED ISSUE #440: `ua enable fips*` should install linux-fips and upgrade any FIPS packages that are already installed

OddBloke:
The effect of this is to upgrade the optional FIPS packages that a user already has present (as well as installing their HMACs).  Previously, we would install _every_ optional FIPS package, whether or not the user was likely to want them.

Fixes #440

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -256,3 +256,11 @@ def migrate_apt_sources(clean=False, cfg=None, platform_info=None):
                 'Disabled %s after package upgrade/downgrade. %s',
                 entitlement.title, details)
         entitlement.enable()  # Re-enable on current series
+
+
+def is_pkg_installed(pkg_name: str) -> bool:
+    try:
+        util.subp(['dpkg', '-s', pkg_name])
+    except util.ProcessExecutionError:
+        return False
+    return True
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -1,8 +1,8 @@
 from uaclient.entitlements import repo
-from uaclient import util
+from uaclient import apt, util
 
 try:
-    from typing import List  # noqa
+    from typing import Dict, List, Set  # noqa
 except ImportError:
     # typing isn't available on trusty, so ignore its absence
     pass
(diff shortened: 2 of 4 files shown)


uaclient/apt.py
```diff
@@ -256,3 +256,11 @@ def migrate_apt_sources(clean=False, cfg=None, platform_info=None):
                 'Disabled %s after package upgrade/downgrade. %s',
                 entitlement.title, details)
         entitlement.enable()  # Re-enable on current series
+
+
+def is_pkg_installed(pkg_name: str) -> bool:
```
blackboxsw: given that there are ~11 packages  that coule be installed, do we want to batch this with a single dpkg -l and parsing instead for shelling out for each package?

OddBloke: Good question; I'll do some profiling locally. Would you be happy with us changing that in a separate PR?

blackboxsw: Yes that works fine techdebt etc. I was wondering about something like get_installed_packages(pkg_names=())

OddBloke: Given that this will happen as a follow-up, is this approved, or are you still working through your review?

Issue #486, opened 2019-05-08
ua status: technical support level not reporting proper supportLevel


blackboxsw:
ua status currently shows

Technical support level: n/a

This should be one of the values (n/a|essential|standard|advanced) which is provided by ua-contract backend resourceEntitlement -> support  -> affordances -> supportLevel

blackboxsw:
## update: per previous ua status checkin, we decided to leave free contract at 'n/a' instead of 'not included'   This will be omitted from ua status output anyway when #501 is addressed**

PR #487, opened 2019-05-08
status: technical support level comes from ua-contract service

blackboxsw:
The backend service returns a 'supportLevel' affordance value that
should be reported by `ua status` as the Techincal support level.

When no supportLevel value is present in the machine token response,
return 'n/a' (status.INAPPLICABLE) from ua client.

Fixes: #486

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -259,6 +259,12 @@ class UAConfig:
                 'name': ent.name, 'entitled': contract_status,
                 'status': op_status, 'statusDetails': op_details}
             response['services'].append(service_status)
+        support = self.entitlements.get('support', {}).get('entitlement')
+        if support:
+            supportLevel = support.get('affordances', {}).get('supportLevel')
+            if not supportLevel:
+                supportLevel = DEFAULT_STATUS['techSupportLevel']
+            response['techSupportLevel'] = supportLevel
         return response
 
     def status(self) -> 'Dict[str, Any]':
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -241,7 +241,7 @@ class UAConfig:
     def _status(self) -> 'Dict[str, Any]':
         """Return configuration status as a dictionary."""
         from uaclient.entitlements import ENTITLEMENT_CLASSES
-        response = DEFAULT_STATUS
+        response = copy.deepcopy(DEFAULT_STATUS)
         response['attached'] = self.is_attached
         if not self.is_attached:
             return response
--- a/uaclient/tests/test_config.py
+++ b/uaclient/tests/test_config.py
@@ -308,6 +308,8 @@ class TestStatus:
             'techSupportLevel': status.INAPPLICABLE,
         }
         assert expected == cfg.status()
+        # cfg.status() idempotent
+        assert expected == cfg.status()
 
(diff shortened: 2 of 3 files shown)


blackboxsw:
To test:
make deb
lxc launch ubuntu-daily:bionic b1
lxc file push ubuntu-advantage-tools_all.deb b1/
lxc exec b1 -- dpkg -i /ubuntu-advantage-tools_all.deb
lxc exec b1 ua attach
lxc exec b1 ua status

powersj:
@blackboxsw can you resolve the conflicts please

blackboxsw:
Unblocking this per UA checkin meeting. We will allow this behavior in client and expect minor salesforce tickets for corrections on individual contracts if customers see discrepancy for invalid supportLevel reported.

uaclient/config.py
```diff
@@ -241,7 +241,7 @@ def write_cache(self, key: str, content: 'Any') -> None:
     def _status(self) -> 'Dict[str, Any]':
         """Return configuration status as a dictionary."""
         from uaclient.entitlements import ENTITLEMENT_CLASSES
-        response = DEFAULT_STATUS
+        response = copy.deepcopy(DEFAULT_STATUS)
```
OddBloke: There doesn't appear to be a test that covers this change. (Ideally, it would also be in a separate commit.)

uaclient/config.py
```diff
@@ -259,6 +259,11 @@ def _status(self) -> 'Dict[str, Any]':
                 'name': ent.name, 'entitled': contract_status,
                 'status': op_status, 'statusDetails': op_details}
             response['services'].append(service_status)
+        support = self.entitlements.get('support', {}).get('entitlement')
+        if support:
+            response['techSupportLevel'] = support.get(
+                'affordances', {}).get('supportLevel', status.INAPPLICABLE)
```
OddBloke: This leaves us with `status.INAPPLICABLE` as the default in two different places (here and in `DEFAULT_STATUS`); can we refactor things so that if/when this default changes (looking at you #501) we don't have to ensure that both are kept in sync?

OddBloke:
Oh, another thought: does this mean we can get rid of some of the constants in status.py?

uaclient/config.py
blackboxsw: added an idempotency check to an existing unit test

uaclient/config.py
blackboxsw: Using the DEFAULT_STATUS['techicalSupportLevel'] we have defined... and will grab #501 next

uaclient/tests/test_config.py
```diff
@@ -306,6 +310,7 @@ def test_nonroot_attached(self, _m_getuid):
             'techSupportLevel': status.INAPPLICABLE,
         }
         assert expected == cfg.status()
+        # cfg.status is idempotent
```
OddBloke: This addition seems more aspirational than functional. Did you accidentally miss a line? :-P

blackboxsw: This is showing up in my local branch per commitish 60744219dd54fc413cdb82b1c6cb1dccfee86d22
+        # cfg.status() idempotent
+        assert expected == cfg.status()

Issue #489, opened 2019-05-09
can't ua attach in 14.04 docker container


gaughen:
Here's what I did: 
 
```
$ docker run -i -t ubuntu:trusty /bin/bash
root@556ef802a9c8:/# apt update
root@556ef802a9c8:/# apt install software-properties-common
root@556ef802a9c8:/# add-apt-repository ppa:canonical-server/ua-client
root@556ef802a9c8:/# apt update
root@556ef802a9c8:/# apt install ubuntu-advantage-tools
root@556ef802a9c8:/# ua attach <token>
Traceback (most recent call last):
  File "/usr/bin/ua", line 9, in <module>
    load_entry_point('ubuntu-advantage-tools==19.5', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 397, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 267, in action_attach
    cfg, contract_token, allow_enable=True):
  File "/usr/lib/python3/dist-packages/uaclient/contract.py", line 300, in request_updated_contract
    contract_token=contract_token)
  File "/usr/lib/python3/dist-packages/uaclient/contract.py", line 149, in request_contract_machine_attach
    machine_id = util.get_machine_id(self.cfg.data_dir)
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 298, in get_machine_id
    write_file(fallback_machine_id_file, machine_id)
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 237, in write_file
    content = encode_text(content)
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 80, in encode_text
    return text.encode(encoding)
AttributeError: 'UUID' object has no attribute 'encode'
```

PR #491, opened 2019-05-09
docker: fix persisting generated uuid on images without machine-id files

blackboxsw:
On docker ubuntu images, neither /etc/machine-id nor
/var/lib/dbus/machine-id exist. Fix the path where uaclient generates
and persists its own machine-id using uuid4().

Fixes: #489


blackboxsw:
I see this in docker images for xenial too.

Issue #308, opened 2019-04-08
livepatch: retry snap install operation on intermittent errors


cpaelzer:
Hi,
in a discussion about #299 it came up that long term (i.e. not for the release now) we might want to wrap some actions e.g. those that are network dependent or others that are just known to potentially be flaky for whatever reason into retry actions.
That should not mean we should write and allow flaky code ourselves, but only that e.g. "snap install" which can fail due to network issues but often works in a retry then could provide a better experience if retried transparently.

Update:
We should be able to surface an optional retries param to util.subp to cope with most use cases.

blackboxsw:
Thanks Christian for this issue. I'm changing this to directly something specific that is actionable in uaclient code. For this particular issue.

Changing the title "Retry some activities that are known to potentially be flaky" to  'Retry "snap install" and apt update on intermittent errors'.

OddBloke:
On Mon, May 06, 2019 at 11:40:35AM -0700, Chad Smith wrote:
> Changing the title "Retry some activities that are known to
> potentially be flaky" to  'Retry "snap install" and apt update on
> intermittent errors'. 

I would suggest that those are two distinct issues; we don't have a
shared code path for these two commands.

blackboxsw:
Created separate apt retries issue #496   This issue will address only livepatch snap retries

PR #497, opened 2019-05-13
Feature/retry snap install

blackboxsw:
livepatch: add 3 retries for snap install command

Add optional retry_sleeps list param to subp function. The retry_sleeps param is a list of sleep durations to perform between retries to allow for a simple backoff algorithm when retrying a command. 

Add the following retry_sleeps when installing the livepatch snap, [05, 1, 5] which will allow for 3 retries if the snap install fails. 

Fixes: #308

### DIFF
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -167,19 +167,13 @@ def readurl(url: str, data: 'Optional[bytes]' = None,
     return content, resp.headers
 
 
-def subp(args: 'Sequence[str]', rcs: 'Optional[List[int]]' = None,
-         capture: bool = False,
-         retry_sleeps: 'Optional[List[int]]' = None) -> 'Tuple[str, str]':
+def _subp(args, rcs=None, capture=False):
     """Run a command and return a tuple of decoded stdout, stderr.
 
     @param subp: A list of arguments to feed to subprocess.Popen
     @param rcs: A list of allowed return_codes. If returncode not in rcs
         raise a ProcessExecutionError.
-    @param capture: Boolean set True to log the command and response.
-    @param retry_sleeps: Optional list of sleep lengths to apply between
-        retries. Specifying a list of [0.5, 1] instructs subp to retry twice
-        on failure; sleeping half a second before the first retry and 1 second
-        before the next retry.
+     @param capture: Boolean set True to log the command and response.
 
     @return: Tuple of utf-8 decoded stdout, stderr
     @raises ProcessExecutionError on invalid command or returncode not in rcs.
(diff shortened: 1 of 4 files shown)


uaclient/entitlements/tests/test_fips.py
```diff
@@ -177,7 +180,7 @@ def test_enable_errors_on_repo_pin_but_invalid_origin(
 
         assert 0 == m_add_apt.call_count
         assert 0 == m_add_pinning.call_count
-        assert 'ERROR    Cannot setup apt pin' in caplog_text()
+        assert 'Cannot setup apt pin' in caplog_text()
```
OddBloke: I think we're actually relaxing the test here; this would no longer catch if we started logging at a higher log level (e.g. CRITICAL, or accidentally logging at an undefined level > logging.ERROR).

I don't think this is that important for this test (because the difference between ERROR and CRITICAL isn't that huge given how we use the logging levels), but for tests where we are currently ensuring we emit WARNING I think it probably is an issue worth addressing (because changing from WARNING to ERROR is a substantial difference in how we would expect users to respond).

(If this change was solely to address the log message matching errors, perhaps we can just drop it?)

uaclient/util.py
```diff
@@ -183,25 +188,39 @@ def subp(args, rcs=None, capture=False):
...
+            proc = subprocess.Popen(
+                bytes_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            (out, err) = proc.communicate()
+        except OSError:
+            try:
+                message = ('Failed running cmd: %s, rc: %s stderr: %s' %
+                           (' '.join(args), proc.returncode, err))
+            except UnboundLocalError:
+                message = 'Failed running cmd: %s' % ' '.join(args)
+            if not retry_sleeps:
```
OddBloke: I don't have time to dig in to this further, because I'm about to run to a session, but we shouldn't have the retrying logic duplicated both here and below; can we do something to remove the repetition?

uaclient/entitlements/tests/test_fips.py
blackboxsw: Dan, yes let's drop for now, I  wanted to get someone's eyes on the approach in case we wanted to construct something that filters caplogs to a specific level. There might be something separate we can iterate on if the need arises, but your fix to tip settles the original issue of CI falling over.

uaclient/util.py
blackboxsw: good thought, I was wondering how to do this. Spent a bit more time and teased out this logic into a separate _raise_or_retry helper function.

OddBloke:
Could we make this a little less of an invasive change to subp's code if we wrapped a `try`/`except ProcessExecutionError` around the entire current body of subp and implemented the retry logic in that except body?

(Apologies for the disjointed sprint week review!)

powersj:
@blackboxsw can you resolve the conflicts and resolve the conversations with @OddBloke if they are fixed?

uaclient/entitlements/tests/test_fips.py
blackboxsw: Just pushed a changeset that is less invasive for subp by creating a wrapper around _subp

uaclient/util.py
blackboxsw: this has been superceded by a change where we use subp as a retrying wrapper around _subp

blackboxsw:
finally addressed review comments

uaclient/util.py
```diff
@@ -207,6 +202,40 @@ def subp(args: 'Sequence[str]', rcs: 'Optional[List[int]]' = None,
...
+    while True:
+        try:
+            out, err = _subp(args, rcs, capture)
+            break
+        except ProcessExecutionError as e:
+            message = str(e)
+            if retry_sleeps:
+                logging.debug(
+                    message + " Retrying %d more times.", len(retry_sleeps))
+                time.sleep(retry_sleeps.pop(0))
```
OddBloke: This will mutate the list that's passed in, which would break consumer code that did something like:

```
sleeps = [1, 2, 3]
subp(..., retry_sleeps=sleeps)  # Would retry
subp(..., retry_sleeps=sleeps)  # Would not, because sleeps is now []
```

uaclient/util.py
```diff
@@ -166,14 +167,13 @@ def readurl(url: str, data: 'Optional[bytes]' = None,
     return content, resp.headers
 
 
-def subp(args: 'Sequence[str]', rcs: 'Optional[List[int]]' = None,
-         capture: bool = False) -> 'Tuple[str, str]':
+def _subp(args, rcs=None, capture=False):
```
OddBloke: Not having type annotations here means that the body of the function won't be type-checked (because `args`, `rcs` and `capture` will all be Any).

Issue #504, opened 2019-05-16
Clean up 401 messages on bad password/username


powersj:
When logging in via SSO, do not print all the massive messages. Simply say something like "Invalid username or password". I would check what `snapcraft login` shows or something similar.

The 3 large messages and the 401 message were too much

OddBloke:
Hey Josh,

When I fail to attach (I've tried invalid email, invalid password and invalid 2FA token; all present with the same format, just a different message after "Bad Request:"), I don't see 3 large messages (unless, I guess, "ERROR", "HTTP Error 400" and "Bad Request" are the messages?):

```
# ua attach
Email: lolno
Password: 
ERROR: HTTP Error 400: Bad Request: Enter a valid email address.
Could not attach machine. Unable to obtain authenticated user token
```

Regardless, I think we can still do better by condensing this to a single error line; this is what `snapcraft login` errors look like:

```
$ snapcraft login
Enter your Ubuntu One e-mail address and password.
If you do not have an Ubuntu One account, you can create one at https://snapcraft.io/account
Email: lolno
Password: 
Authentication error: Failed to get unbound discharge: Invalid request data
```

PR #525, opened 2019-05-27
Prune unused functions and methods

OddBloke:
I noticed one apparently unused function when looking at error messaging for #504 and then used https://github.com/jendrikseipp/vulture to detect the others.

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -10,7 +10,6 @@ except ImportError:
     pass
 
 
-API_PATH_TMPL_ACCOUNT_USERS = '/accounts/{account}/users'
 API_PATH_TMPL_CONTRACT_MACHINES = '/contracts/{contract}/context/machines'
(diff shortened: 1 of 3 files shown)


uaclient/contract.py
```diff
@@ -163,21 +125,6 @@ def request_contract_machine_attach(self, contract_token, machine_id=None):
         self.cfg.write_cache('machine-token', machine_token)
         return machine_token
 
-    def request_contract_machine_detach(self, contract_id, user_token):
```
blackboxsw: I believe we **should** call this and it is an error of omission that we aren't. We should check in UA checkin meeting today that the ua-contracts service expects this call to close the loop for accounting on active 'attached' machines.

OddBloke: @blackboxsw I've dropped the commit that removed this method.

PR #529, opened 2019-05-29
Refactor internal status to be stored across three axes

OddBloke:
This refactors the internal state of the codebase so that we're passing around three types of enum: `ContractStatus`, `ApplicabilityStatus` and `ApplicationStatus`. Entitlements have methods that match these names, which return the entitlement's status for that axis.

operational_status remains (albeit moved to a shared implementation for all entitlements), but will be removed in favour of direct use of the `contract_status()`, `applicability_status()` and `application_status()` methods in a future PR.

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -14,6 +14,7 @@ from uaclient import config
 from uaclient import contract
 from uaclient import status
 from uaclient import util
+from uaclient.status import ContractStatus
 
 RE_KERNEL_UNAME = (
     r'(?P<major>[\d]+)[.-](?P<minor>[\d]+)[.-](?P<patch>[\d]+\-[\d]+)'
@@ -101,7 +102,7 @@ class UAEntitlement(metaclass=abc.ABCMeta):
             contract_client = contract.UAContractClient(self.cfg)
             contract_client.request_resource_machine_access(
                 token, self.name)
-        if not self.contract_status() == status.ENTITLED:
+        if not self.contract_status() == ContractStatus.ENTITLED:
             if not silent:
                 print(status.MESSAGE_UNENTITLED_TMPL.format(title=self.title))
             return False
@@ -188,14 +189,14 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         """
         pass
 
-    def contract_status(self) -> str:
-        """Return whether contract entitlement is ENTITLED or NONE."""
+    def contract_status(self) -> ContractStatus:
+        """Return whether the user is entitled to the entitlement or not"""
         if not self.cfg.is_attached:
-            return status.NONE
+            return ContractStatus.UNENTITLED
         entitlement_cfg = self.cfg.entitlements.get(self.name, {})
         if entitlement_cfg and entitlement_cfg['entitlement'].get('entitled'):
-            return status.ENTITLED
-        return status.NONE
+            return ContractStatus.ENTITLED
+        return ContractStatus.UNENTITLED
 
     def is_access_expired(self) -> bool:
         """Return entitlement access info as stale and needing refresh."""
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -118,16 +118,17 @@ class UAEntitlement(metaclass=abc.ABCMeta):
             return False
         return True
 
-    def check_affordances(self) -> 'Tuple[ApplicabilityStatus, str]':
+    def applicability_status(self) -> 'Tuple[ApplicabilityStatus, str]':
         """Check all contract affordances to vet current platform
 
         Affordances are a list of support constraints for the entitlement.
         Examples include a list of supported series, architectures for kernel
         revisions.
 
-        @return: Tuple (ApplicabilityStatus, detailed_message).  APPLICABLE if
-        platform passes all defined affordances, INAPPLICABLE if it doesn't
-        meet all of the provided constraints.
+        :return:
+            tuple of (ApplicabilityStatus, detailed_message).  APPLICABLE if
+            platform passes all defined affordances, INAPPLICABLE if it doesn't
+            meet all of the provided constraints.
         """
         entitlement_cfg = self.cfg.entitlements.get(self.name)
         if not entitlement_cfg:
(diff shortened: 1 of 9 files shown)


uaclient/entitlements/base.py
```diff
@@ -273,7 +278,26 @@ def process_contract_deltas(
 
         return False
 
-    @abc.abstractmethod
     def operational_status(self) -> 'Tuple[str, str]':
         """Return whether entitlement is ACTIVE, INACTIVE or UNAVAILABLE"""
+        applicability, details = self.applicability_status()
```
blackboxsw: Always good to drop the duplicated code. thanks.

uaclient/entitlements/base.py
```diff
@@ -273,7 +278,26 @@ def process_contract_deltas(
...
+        application_status, explanation = self.application_status()
+        return application_status.to_operational_status(), explanation
+
+    @abc.abstractmethod
+    def application_status(self) -> 'Tuple[status.ApplicationStatus, str]':
+        """
+        The current status of application of this entitlement
+
+        :return:
+            A tuple of (ApplicationStatus, human-friendly reason)
```
blackboxsw: human-friendly reason... are you saying humans are inherently unreasonable?

uaclient/status.py
```diff
@@ -1,14 +1,53 @@
...
+
 class TxtColor:
     OKGREEN = '\033[92m'
     DISABLEGREY = '\033[37m'
     FAIL = '\033[91m'
     ENDC = '\033[0m'
 
 
+@enum.unique
+class ApplicationStatus(enum.Enum):
```
blackboxsw: I'm missing something. What is the benefit of this implementation versus the following?

class ApplicationStatus(enum.Enum):
    """
    An enum to represent the current application status of an entitlement
    """
    ENABLED = ACTIVE
    DISABLED = INACTIVE


It looks like we could drop the to_operational_status method by moving this enum definition below the ACTIVE/INACTIVE global declarations.

    https://paste.ubuntu.com/p/f5ZnssXHJ7/

uaclient/status.py
```diff
@@ -1,14 +1,53 @@
...
+@enum.unique
+class ApplicabilityStatus(enum.Enum):
+    """
+    An enum to represent whether an entitlement could apply to this machine
+    """
+    APPLICABLE = object()
+    INAPPLICABLE = object()
+
+
 ACTIVE = 'active'
```
blackboxsw: Can we drop ACTIVE and INACTIVE now and use the ApplicationStatus enum as the reference for those static stings?

uaclient/status.py
blackboxsw: alternately, we could drop ACTIVE/INACTIVE altogether right and just us the 'active'/'inactive' strings here couldn't we?

uaclient/status.py
```diff
@@ -19,7 +58,6 @@ class TxtColor:
     ACTIVE: TxtColor.OKGREEN + ACTIVE + TxtColor.ENDC,
```
blackboxsw: could instead reference ApplicationStatus.ENABLED.value right?

uaclient/entitlements/base.py
OddBloke: :-)

uaclient/entitlements/base.py
OddBloke: Are you saying they aren't?! :-P

uaclient/status.py
OddBloke: My intent is to remove `operational_status()` entirely once this work is complete (replaced by use of the new/renamed axes directly in its current call sites). I expect to remove `to_operational_status()` at the same time.

I used a method instead of the enum values directly here because I wanted to have a clean interface that I could just remove along with `operational_status()`. If I find that it makes sense in the next tranche of work to have ENABLED/DISABLED have useful values (and I expect it will!) then I'll certainly add them, but I didn't want to use them just to support the method I plan to remove in the future.

uaclient/status.py
```diff
@@ -1,14 +1,53 @@
...
+
+@enum.unique
+class ContractStatus(enum.Enum):
+    """
+    An enum to represent whether a user is entitled to an entitlement
+
+    (The value of each member is the string that will be used in status
+    output.)
+    """
+    ENTITLED = 'entitled'
```
OddBloke: Further to my above thinking, this was actually something of a compromise; I like the idea of keeping data and presentation separate, whereas this does muddle up that split (because you could index in to ContractStatus using the "presentation" value and get back something meaningful).

The value of `ContractStatus` is only used in one place though (and that's a place we will be keeping, in status output), so I didn't think it was worth the extra code here.

uaclient/status.py
OddBloke: As I mentioned above, something will happen to ACTIVE/INACTIVE once `operational_status()` is gone, but it's not worth refactoring them before that.

uaclient/status.py
OddBloke: Yep, as above, this will come.

OddBloke:
This fix was squashed in to the commit that broke `ua status` output, this is the diff from the previously reviewed PR:

```diff
diff --git a/uaclient/status.py b/uaclient/status.py
index 7877388..8525ef3 100644
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -58,6 +58,8 @@ STATUS_COLOR = {
     ACTIVE: TxtColor.OKGREEN + ACTIVE + TxtColor.ENDC,
     INACTIVE: TxtColor.FAIL + INACTIVE + TxtColor.ENDC,
     INAPPLICABLE: TxtColor.DISABLEGREY + INAPPLICABLE + TxtColor.ENDC,
+    ContractStatus.ENTITLED.value: (
+        TxtColor.OKGREEN + ContractStatus.ENTITLED.value + TxtColor.ENDC),
     NONE: TxtColor.DISABLEGREY + NONE + TxtColor.ENDC,
     ESSENTIAL: TxtColor.OKGREEN + ESSENTIAL + TxtColor.ENDC,
     STANDARD: TxtColor.OKGREEN + STANDARD + TxtColor.ENDC,
```

PR #531, opened 2019-05-29
Remove a bunch of unused code (as determined by vulture)

OddBloke:


uaclient/sso.py
```diff
@@ -9,23 +9,12 @@
...
-API_PATH_USER_KEYS = API_PATH_V2 + '/keys/'
-API_PATH_OAUTH_TOKEN = API_PATH_V2 + '/tokens/oauth'
 API_PATH_TOKEN_DISCHARGE = API_PATH_V2 + '/tokens/discharge'
-API_PATH_TOKEN_REFRESH = API_PATH_V2 + '/tokens/refresh'
 
 
 # Some Ubuntu SSO API responses use UNDERSCORE_DELIMITED codes, others use
 # lowercase-hyphenated. We'll standardize on lowercase-hyphenated
 API_ERROR_2FA_REQUIRED = 'twofactor-required'
-API_ERROR_2FA_FAILURE = 'twofactor-failure'
```
blackboxsw: I feel like this and API_ERROR_INVALID_DATA are premature too as I'd expect them to be handled by a branch that should be fixing issue: #504

uaclient/contract.py
```diff
@@ -88,7 +84,6 @@ def request_account_contracts(self, macaroon_token, account_id):
         headers = self.headers()
         headers.update({'Authorization': 'Macaroon %s' % macaroon_token})
         account_contracts, _headers = self.request_url(url, headers=headers)
-        self.cfg.write_cache('account-contracts', account_contracts)
```
blackboxsw: I'd like to leave this cache file in place for a couple of reasons:
 - I find value in having a cached reference for all ua-contracts responses to aid in debug/triage
 - we are only a few days from starting mulit-contract support and we'll need to be referencing this response file during ua attach to select/reselect the proper contract during attach.

I suppose these are both circumstantial requests, and could be resolved other ways, but minimally I'd like to wait when dropping cache files until we have a better way of preserving ua-contracts response than log files. We also may have more fodder to drop this once multi-contract feature is started.

uaclient/status.py
```diff
@@ -1,34 +1,23 @@
...
 
 # Colorized status output for terminal
 STATUS_COLOR = {
     ACTIVE: TxtColor.OKGREEN + ACTIVE + TxtColor.ENDC,
     INACTIVE: TxtColor.FAIL + INACTIVE + TxtColor.ENDC,
     INAPPLICABLE: TxtColor.DISABLEGREY + INAPPLICABLE + TxtColor.ENDC,
     ENTITLED: TxtColor.OKGREEN + ENTITLED + TxtColor.ENDC,
-    EXPIRED: TxtColor.FAIL + EXPIRED + TxtColor.ENDC,
     NONE: TxtColor.DISABLEGREY + NONE + TxtColor.ENDC,
-    STANDARD: TxtColor.OKGREEN + STANDARD + TxtColor.ENDC,
```
blackboxsw: Good to know, this is a bug. These should have been used by my supportLevel branch which is landed. 

I've filed #538 to capture that.

uaclient/sso.py
OddBloke: These aren't used in the codebase today, and I already get out relevant error messages:

```
$ UA_CONFIG_FILE=uaclient-devel.conf fakeroot ua attach
Email: daniel.watkins@canonical.com
Password: 
Second-factor auth: 123456
ERROR: HTTP Error 403: Forbidden: The provided 2-factor key is not recognised.
Could not attach machine. Unable to obtain authenticated user token
```

The issue in #504 isn't that we don't have useful content in the error, it's that the errors are needlessly verbose/noisy. So it isn't clear to me why we _would_ need them (and if I'm wrong, the branch that introduces the code that needs them can always reintroduce them!).

uaclient/contract.py
OddBloke: OK, I'm not sure I agree with the reasoning here, but in the interests of landing everything else I'll un-drop this and we can pick it up separately.

uaclient/sso.py
blackboxsw: The branch you are working for this may not actually have to use them.  It's probably a cleaner approach than the attempt I made with https://paste.ubuntu.com/p/HdZXjHHCrx/ which tried to map error types from SSO endpoints a bit better.  As you mention, it's simple enough to reintroduce if needed.

Issue #501, opened 2019-05-16
On personal accounts hide 'valid till' and 'technical support' fields


powersj:
In a review with Mark we decided that the "valid till" and "technical support" levels should be hidden from view on personal accounts. This is because they will always show "n/a" and it does not lead to a good experience.

Keep in mind though, that non-personal accounts should still show these fields, even if they are empty/null/n/a.

OddBloke:
Looking through the data we currently pull down from the contract server, it's not clear to me how we would identify a "personal" account. I believe we'll need changes to the contract server to clearly surface this information.

(I only have ready access to one Ubuntu SSO account, so I haven't been able to check if e.g. the contract ID for the free tier is static across users. Regardless, I don't think we would want to hardcode such an ID in to the client, as that will make changing it in future fraught.)

OddBloke:
This may be closely related to #486; I'm going to remove my face from this and see where #486 gets us to before spending any more time on this.

powersj:
I want this fixed on the backend before we make a change. Please file a bug on contracts backend to track the work required there.

PR #532, opened 2019-05-29
status: redact technical support and expiry for free contracts

blackboxsw:
Fixes: #501

### DIFF
--- a/uaclient/tests/test_status.py
+++ b/uaclient/tests/test_status.py
@@ -7,7 +7,6 @@ from uaclient.status import format_tabular, TxtColor
 class TestFormatTabular:
 
     @pytest.mark.parametrize('support_level,expected_colour', [
-        ('n/a', TxtColor.DISABLEGREY),
         ('essential', TxtColor.OKGREEN),
         ('standard', TxtColor.OKGREEN),
         ('advanced', TxtColor.OKGREEN),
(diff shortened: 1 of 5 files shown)


OddBloke:
In #501, @powersj said "non-personal accounts should still show these fields, even if they are empty/null/n/a" which this branch doesn't handle.

Technically, this looks good to me, but I would like confirmation from Josh that we're happy to modify the behaviour for all users before we merge it.

uaclient/tests/test_status.py
```diff
@@ -7,7 +7,6 @@
 class TestFormatTabular:
 
     @pytest.mark.parametrize('support_level,expected_colour', [
-        ('n/a', TxtColor.DISABLEGREY),
```
OddBloke: Why did we drop this line? I believe it's possible for a non-free contract to not have technical support.

powersj: Hmm I paused here and briefly thought the same. It was my understanding that if you buy a contract, then you are either buying Essential, Standard, or Advanced and one of those three should appear.

However, if we do come across a case of a contract without one of those three, this will fail?

OddBloke: This shouldn't fail, we'll just produce uncoloured output (which will be inconsistent with other "n/a"s in the same output). I assume that someone could pay us for just, say, livepatch for $kernel without receiving technical support?

powersj: That is UA essential :)

OddBloke: We only provide the essential support people need, i.e. none? Sweet naming.

OddBloke: (Fair enough!)

Issue #535, opened 2019-05-30
Avoid traceback on ctrl-C during ua attach if user cancels


blackboxsw:
```
multipass@ua-t:~$ sudo ua attach
Email: ^CTraceback (most recent call last):
  File "/usr/bin/ua", line 11, in <module>
    load_entry_point('ubuntu-advantage-tools==19.5', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 399, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 247, in action_attach
    bound_macaroon_bytes = sso.discharge_root_macaroon(contract_client)
  File "/usr/lib/python3/dist-packages/uaclient/sso.py", line 215, in discharge_root_macaroon
    discharge_macaroon = prompt_request_macaroon(cfg, caveat_id)
  File "/usr/lib/python3/dist-packages/uaclient/sso.py", line 180, in prompt_request_macaroon
    email = input('Email: ')
KeyboardInterrupt
```

Let's make it a cleaner exit

PR #536, opened 2019-05-30
attach: no tracebacks when ctrl-c is used during email/password prompts

blackboxsw:
Fixes: #535


blackboxsw:
Closing, Dan is working a more general approach for this

PR #540, opened 2019-05-30
status: colour standard/advanced support levels correctly

### LINKED ISSUE #538: ua client tech support level essential, standard and advanced  should be green
Per uaclient spec https://docs.google.com/document/d/1ZUjKyErwJftQnT800PIWgspT9P_QgvRKhd-jnrAPycg/edit

support level (which comes from ua-contract service supportLevel value) should be green if it is one of  essential, standard or advanced

OddBloke:
Fixes #538

### DIFF
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -10,6 +10,8 @@ INACTIVE = 'inactive'
 INAPPLICABLE = 'n/a'
 ENTITLED = 'entitled'
 NONE = 'none'
+STANDARD = 'standard'
+ADVANCED = 'advanced'
 
 # Colorized status output for terminal
 STATUS_COLOR = {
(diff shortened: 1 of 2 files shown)


uaclient/status.py
```diff
@@ -10,6 +10,8 @@ class TxtColor:
 INAPPLICABLE = 'n/a'
 ENTITLED = 'entitled'
 NONE = 'none'
+STANDARD = 'standard'
+ADVANCED = 'advanced'
```
blackboxsw: Just missing an 'essential' here too.

uaclient/status.py
```diff
@@ -18,6 +20,8 @@ class TxtColor:
     INAPPLICABLE: TxtColor.DISABLEGREY + INAPPLICABLE + TxtColor.ENDC,
     ENTITLED: TxtColor.OKGREEN + ENTITLED + TxtColor.ENDC,
     NONE: TxtColor.DISABLEGREY + NONE + TxtColor.ENDC,
+    STANDARD: TxtColor.OKGREEN + STANDARD + TxtColor.ENDC,
+    ADVANCED: TxtColor.OKGREEN + ADVANCED + TxtColor.ENDC
```
blackboxsw: ditto on ESSENTIAL

Issue #423, opened 2019-04-19
Upgrade path: remove old keyring files on upgrade to new uaclient


blackboxsw:
Old uaclient delivered keyring files for esm which have now been replaced with ubuntu-esm-v2-keyring.gpg. 

Given that v1 of the esm keyring file was replaced because it was viewed as compromised, uaclient should either remove that keyring file on upgrade from oldclient to newclient. As of 4/16/2019, production and staging repositories are signed only by esm-v2 gpgkey.

powersj:
Marking blocker for next release

PR #543, opened 2019-05-30
packaging: trusty should remove v1 esm key if present after upgrade

blackboxsw:
v1 was viewed as compromised and has been retired in favor of
ubuntu-esm-v2-keyring.gpg. Remove this from the machines list of
acceptable keys if present as the esm repo is signed with v2.

Fixes: #423

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -3,9 +3,11 @@
 set -e
 
 ESM_APT_GPG_KEY="/etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg"
+ESM_APT_GPG_KEY_OLD="/etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg"
 ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
 
 configure_esm() {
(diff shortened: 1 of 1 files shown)


debian/postinst
```diff
@@ -3,9 +3,11 @@
 set -e
 
 ESM_APT_GPG_KEY="/etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg"
+ESM_APT_GPG_KEY_V1="/etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg"
```
powersj: bikeshedding, but can we call it something else? like old?

debian/postinst
```diff
@@ -3,9 +3,11 @@
...
 ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
 
 configure_esm() {
+    rm -f $ESM_APT_GPG_KEY_V1;   # Remove retired key from key list
```
panlinux: no need for a `;`

debian/postinst
blackboxsw: funny I had OLD but changed to V1 in case we end up retiring V2 too.
I don't mind switching it.

Issue #520, opened 2019-05-22
Write preferences.d to disable ESM repository rather than magic


julian-klode:
Since Packages-Require-Authorization did not work very well, let's do manual pinning instead when adding the ESM repo:

We'd want to have

```
Package: *
Pin: ...
Pin-Priority: never
```

where ... describes the repo as we do now for some I think.

and an appropriate ua client Depends on apt (>= in -updates) so apt understands it.

julian-klode:
Oh it just popped into my mind: The Depends probably needs to be on `libapt-pkg<abi> (>= version)`, not on `apt`.

panlinux:
@julian-klode , given that libapt-pkg<abi> is different across ubuntu releases, is there a way to make this automatic, by having the apt hook perhaps use a specific symbol of that library, or are we better off using substvars in the packaging for now, and have a static mapping between libapt-pkg<abi> and ubuntu releases in the debian/rules file?

julian-klode:
We do not have any specific symbol introduced in that version, so hardcoding that stuff for now seems to be the easiest option.

PR #545, opened 2019-05-31
esm: use Pin-Priority never apt preference file to disable esm initially

blackboxsw:
Instead of relying on Packages-Require-Authorization header for the ESM
repo, disable the repo initially on package install setting
Pin-Priority: never. Upon enabling, we'll remove that preference file thus
enabling package installs from the repo.
    
Fixes: #520


debian/postinst
```diff
@@ -22,6 +23,14 @@ deb https://esm.ubuntu.com/ubuntu trusty-updates main
 # deb-src https://esm.ubuntu.com/ubuntu trusty-updates main
 EOF
     fi
+    if [ ! -e "$ESM_APT_PREF_FILE" ]; then
+	cat > $ESM_APT_PREF_FILE <<EOF
+# Written by ubuntu-advantage-tools
+Package: *
+Pin: release o=UbuntuESM, n=xenial
```
OddBloke: Is `xenial` right here?

OddBloke:
Code looks good with the exception of my one inline question.

debian/postinst
blackboxsw: wow, oops. fixed.

PR #546, opened 2019-05-31
config: drop unused UAConfig.contracts property

OddBloke:
We never actually read the account-contracts cached file (as we only
need it when we're performing an attach, at which point we obviously
won't have it cached), so we can drop it entirely.

(In 3fd37d3 we switched to using the contract information in the machine
token, and in 9cf48da we dropped the last of the MOTD support; that was
the point at which we stopped using it.)

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -39,7 +39,6 @@ class UAConfig:
     data_paths = {
         'bound-macaroon': DataPath('bound-macaroon', True),
         'accounts': DataPath('accounts.json', True),
-        'account-contracts': DataPath('account-contracts.json', True),
         'account-users': DataPath('account-users.json', True),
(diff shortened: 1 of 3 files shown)


OddBloke:
Reintroducing this, because I still don't think it's worth keeping around. The only time we write it is at attach, which means (a) it's of limited utility for debugging, because it's likely to be out-of-date, and (b) if any code does start using it, it will almost certainly be broken because it will only ever operate on the state of the system/contract server at attach time.

((b) is particularly insidious, IMO, because we probably wouldn't notice it in our testing in short-lived machines/instances.)

If it truly is needed in its current form for the multi-contract work, we can always revert this commit (but I suspect we'll find that it's deficient in other ways, as it isn't exercised by anything currently).

PR #547, opened 2019-05-31
contract: remove unused UAContractClient.request_contract_machine_detach

OddBloke:
There was a concern that we should be calling this on detach; I'm opening up this PR to serve as a place for us to work that out. We can either close or merge it depending on our determination.

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -10,8 +10,6 @@ except ImportError:
     pass
 
 
-API_PATH_TMPL_CONTRACT_MACHINES = '/contracts/{contract}/context/machines'
-
 API_V1_ACCOUNTS = '/v1/accounts'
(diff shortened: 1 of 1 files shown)


blackboxsw:
I chatted with #jaas on IRC Wed May 29th:

<cmars> blackboxsw: just my opinion, but i think a DELETE like that can only ever be considered a courtesy and not necessarily reliable, since VMs can be terminated in many wayssay 
12:05 
<urulama> blackboxsw, so, consider it "nice to have", but we won't count on it

I'd let's keep it in and wire it to the detach operation with a timeout (incase of network unavailable) as there is already an existing backend api endpoint and this can only help us on the backend measuring utilization of services and attach/detach rates

blackboxsw:
We've decided to land this as we need to fixup this method a bit anyway

PR #552, opened 2019-06-03
packaging: depend on libapt-pkg4.12 version to use pin-priority never

blackboxsw:
Bug #1821308 fix is required to honor pin-priority: never in apt
preferences files. This fix comes in different libapt-pkg<abi> versions on trusty, xenial, bionic and disco.
- trusty: libapt-pkg4.12 version 1.0.1ubuntu2.23 
- xenial: libapt-pkg5.0 version 1.2.30
- bionic: ibapt-pkg5.0 (>= 1.6.9)
- cosmic: libapt-pkg5.0 (>= 1.7.3)
- disco: libapt-pkg5.0 (>= 1.8.0~rc1)
- eoan ++ libapt-pkg5.0 (>= 1.8.1)

UA client depends on this version to prevent installing packages from esm repositories until esm is enabled.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -4,6 +4,19 @@ export DH_VERBOSE=1
 include /usr/share/dpkg/pkg-info.mk
 FLAKE8 := $(shell flake8 --version 2> /dev/null)
 
+include /etc/os-release
+ifeq (${VERSION_ID},"14.04")
+LIBAPT_PKG_DEP="libapt-pkg4.12 (>= 1.0.1ubuntu2.23)"
+else ifeq (${VERSION_ID},"16.04")
+LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.2.31)"
+else ifeq (${VERSION_ID},"18.04")
+LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.6.9)"
+else ifeq (${VERSION_ID},"18.10")
+LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.7.4)"
+else
+LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.8.1)"
+endif
+
 %:
 	dh $@ --with python3,bash-completion --buildsystem=pybuild
 
@@ -22,6 +35,10 @@ else
 	python3 -m flake8 uaclient
 endif
 endif
+
+override_dh_gencontrol:
+	echo extra:Depends=$(LIBAPT_PKG_DEP) >> debian/ubuntu-advantage-tools.substvars
+	dh_gencontrol
 
 override_dh_auto_install:
 	dh_auto_install --destdir=debian/ubuntu-advantage-tools
--- a/debian/control
+++ b/debian/control
@@ -25,7 +25,8 @@ Architecture: any
 Depends: ${misc:Depends},
          ${python3:Depends},
          ${shlibs:Depends},
-         python3-pkg-resources
+         python3-pkg-resources,
+         ${extra:Depends},
 Description: management tools for Ubuntu Advantage
  Ubuntu Advantage is the professional package of tooling, technology
  and expertise from Canonical, helping organisations around the world


OddBloke:
We currently generate a dependency of `libapt-pkg4.12 (>= 0.8.0)`; is there something we should be doing instead so that whatever generates that dependency generates the correct lower bound without manual specification?

panlinux:
If we used a symbol that was only declared in the newer libapt-pkg symbol file, then we would get the correct dependency automatically. I don't know if the fix to libapt introduced such a new symbol, probably not.

OddBloke:
OK, makes sense to me. Next question: on eoan, that (`dh_shlibdeps`?) generates `libapt-pkg5.0 (>= 0.8.0)` instead. `libapt-pkg4.12` isn't present in any release past trusty. So I think this change will immediately cause our PPA builds on post-trusty releases to become uninstallable.

OddBloke:
(I just tested a disco build locally, and can confirm it wasn't installable in a lxd container.)

panlinux:
Indeed, we can't have it static like this. If there is nothing we "link" to in libapt-pkg, so that the dependency gets automatically added, then plan B is to generate a [subsvars](https://www.debian.org/doc/debian-policy/ch-source#variable-substitutions-debian-substvars) file so that we can generate the right dependency name depending on the ubuntu release where the package is being built.

panlinux:
Here is a *quick* example:
d/rules:
landscape_common_substvars = debian/landscape-common.substvars
...
# from quantal onwards, we don't want python-gnupginterface anymore (#1045237)
ifneq (,$(filter $(dist_release),lucid precise))
    echo "extra:Depends=python-gnupginterface" >> $(landscape_common_substvars)
endif

d/control:
Package: landscape-common
Architecture: any
Depends: ${python:Depends}, ${misc:Depends}, ${extra:Depends},
...

That `${extra:Depends}` will include the dynamically generated depends added in d/rules.

panlinux:
Since there is only one package here, we can create `debian/substvars` directly, and populate it in `d/rules`, I think in the `override_dh_auto_build` section. Let me try some stuff.

panlinux:
Ok, this worked as a proof of concept:
```diff
diff --git a/debian/control b/debian/control
index 9925206..025a18a 100644
--- a/debian/control
+++ b/debian/control
@@ -25,6 +25,7 @@ Architecture: any
 Depends: ${misc:Depends},
          ${python3:Depends},
          ${shlibs:Depends},
+         ${extra:Depends},
          python3-pkg-resources
 Description: management tools for Ubuntu Advantage
  Ubuntu Advantage is the professional package of tooling, technology
diff --git a/debian/rules b/debian/rules
index ca39cc9..5661381 100755
--- a/debian/rules
+++ b/debian/rules
@@ -11,6 +11,10 @@ override_dh_auto_build:
        dh_auto_build
        make -C apt-hook build
 
+override_dh_gencontrol:
+       echo "extra:Depends=libapt-foo-bar" >> debian/ubuntu-advantage-tools.substvars
+       dh_gencontrol $@
+
 override_dh_auto_test:
 ifeq (,$(filter nocheck,$(DEB_BUILD_OPTIONS)))
        python3 -m pytest
```

It's more dynamic, in the sense that it can be made to change depending on the Ubuntu release where the package is being built, but also still "static" in the sense that we will have a static mapping between libapt-pkg versions and the Ubuntu release in `d/rules`, so not optimal yet.

debian/rules
```diff
@@ -3,6 +3,7 @@ export DH_VERBOSE=1
 
 include /usr/share/dpkg/pkg-info.mk
 FLAKE8 := $(shell flake8 --version 2> /dev/null)
+DIST_RELEASE := $(shell lsb_release -cs)
```
panlinux: This needs an extra build-depends. Can't we parse `/etc/os-release`? It's not identical in its structure in all releases, though, one would have to check all the way back from trusty.

debian/rules
```diff
@@ -23,6 +24,21 @@ else
 endif
 endif
 
+override_dh_gencontrol:
+ifeq ($(DIST_RELEASE),trusty)
+	echo "extra:Depends=libapt-pkg4.12 (>= 1.0.1ubuntu2.23)" >> debian/ubuntu-advantage-tools.substvars
+else ($(DIST_RELEASE),xenial))
+	echo "extra:Depends=libapt-pkg5.0 (>= 1.2.30)" >> debian/ubuntu-advantage-tools.substvars
```
panlinux: We could perhaps depend on 1.2.31: https://launchpad.net/ubuntu/+source/apt/1.2.31
There is 1.2.32 in proposed, but doesn't look like it's related to what we are doing here.

debian/rules
```diff
@@ -23,6 +24,21 @@ else
...
+else ($(DIST_RELEASE),cosmic))
+	echo "extra:Depends=libapt-pkg5.0 (>= 1.7.3)" >> debian/ubuntu-advantage-tools.substvars
```
panlinux: Maybe 1.7.4? https://launchpad.net/ubuntu/+source/apt/1.7.4 same fix as above for xenial (auth.conf.d)

debian/rules
```diff
@@ -23,6 +24,21 @@ else
...
+else ($(DIST_RELEASE),disco))
+	echo "extra:Depends=libapt-pkg5.0 (>= 1.8.0~rc1)" >> debian/ubuntu-advantage-tools.substvars
```
panlinux: Just 1.8.0 perhaps, instead of ~rc1? https://launchpad.net/ubuntu/+source/apt/1.8.0 I don't see rc1 published anywhere.

debian/rules
```diff
@@ -23,6 +24,21 @@ else
...
+else ($(DIST_RELEASE),eoan))
+	echo "extra:Depends=libapt-pkg5.0 (>= 1.8.1)" >> debian/ubuntu-advantage-tools.substvars
+endif
+	dh_gencontrol $@
```
panlinux: I'm checking if `$@` is necessary, it looks like it isn't (I know I suggested it).

debian/rules
```diff
@@ -23,6 +24,21 @@ else
 endif
 endif
 
+override_dh_gencontrol:
+ifeq ($(DIST_RELEASE),trusty)
+	echo "extra:Depends=libapt-pkg4.12 (>= 1.0.1ubuntu2.23)" >> debian/ubuntu-advantage-tools.substvars
+else ($(DIST_RELEASE),xenial))
```
panlinux: You need another `ifeq` after `else` in all of these.

debian/rules
panlinux: Just confirmed, `$@`  is unecessary. `dh_gencontrol` by itself is sufficient.

debian/rules
blackboxsw: Can do this instead with the following (relying on our own code which is tested on each release we support)
https://paste.ubuntu.com/p/srMkS7DgR5/

blackboxsw: updated this to use VERSION_ID from /etc/os-release as that avoids python in debian/rules and is an available field on all supported series.

debian/rules
blackboxsw: done will use -updates version on each series since we  know it carries the fix (and we can test it for compatibility)

debian/rules
blackboxsw: moving disco to 1.8.1 since that's in disco-updates  (and dropping the ifeq disco because eoan is on the same minimim version so we don't need to break out disco as a separate conditional from >= eoan.

debian/rules
blackboxsw: resolved in a separate commit before this review.

debian/rules
blackboxsw: dropped in 833263fe1909e4ad865a85da9a5b5cd470c8f641

debian/rules
blackboxsw: resolved in commitish e798ea2a0817f421a486b84fba0fe1231de4da8f.   moved to the published version in <series>-updates for each series.

panlinux:
Try this: https://pastebin.ubuntu.com/p/5cd5JBvhSq/ I tested with bionic and it worked, didn't test the others (specially trusty).

debian/rules
```diff
@@ -1,9 +1,22 @@
 #!/usr/bin/make -f
 export DH_VERBOSE=1
+include /etc/os-release
```
panlinux: Nit picking, put this include next to the ifeq block, so it's clearer that the VERSION_ID variable comes from there

blackboxsw: done

powersj:
Let's clean up the extra lines and merge this!

PR #557, opened 2019-06-05
Start improving discharge_root_macaroon error handling

OddBloke:
This is the first PR towards addressing #504. It puts the framework in place for the new error reporting, and addresses the errors that are handled in `discharge_root_macaroon`.

(The following PR(s) will update the exceptions produced by the interaction with SSO itself, to directly address #504.)

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -363,12 +363,14 @@ def setup_logging(console_level, log_level, log_file=None):
             if handler.stream.name == '<stderr>':
                 handler.setLevel(console_level)
                 handler.setFormatter(console_formatter)
+                handler.set_name('console')  # Used to disable console logging
                 stderr_found = True
                 break
     if not stderr_found:
         console = logging.StreamHandler(sys.stderr)
         console.setFormatter(console_formatter)
         console.setLevel(console_level)
+        console.set_name('console')  # Used to disable console logging
         root.addHandler(console)
     if os.getuid() == 0:
         # Setup debug file logging for root user as non-root is read-only
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -241,7 +241,8 @@ def action_attach(args, cfg):
         raise exceptions.NonRootUserError()
     contract_client = contract.UAContractClient(cfg)
     if not args.token:
-        bound_macaroon_bytes = sso.discharge_root_macaroon(contract_client)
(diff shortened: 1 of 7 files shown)


uaclient/sso.py
```diff
@@ -198,10 +200,13 @@ def discharge_root_macaroon(contract_client):
         caveat_id = extract_macaroon_caveat_id(root_macaroon['macaroon'])
         discharge_macaroon = prompt_request_macaroon(cfg, caveat_id)
     except (util.UrlError) as e:
-        logging.error("Could not reach url '%s' to authenticate.", e.url)
-        return None
+        msg = "Could not reach URL %s to authenticate"
+        logging.error(msg, e.url)
```
blackboxsw: It seems like in most cases we would want to be logging UserFactingErrors anyway, why aren't we perform the logging.error in cli.py:main_error_handler line 392?

OddBloke: Great point; I'll look at addressing that.

blackboxsw: Yeah I think if you had the with disable_log_to_console context manager you could also drop it from the discharge_root_macaroon call site

blackboxsw: > Yeah I think if you had the with disable_log_to_console context manager in the exceptions.UserFacingError of main_error_handler you could also drop that ctx_mgr from the discharge_root_macaroon call site

OddBloke: I've changed where the logging happens; I don't think we're yet ready to remove the context manager, though we might be by the end of this series of PRs.

uaclient/cli.py
```diff
@@ -241,7 +241,8 @@ def action_attach(args, cfg):
         raise exceptions.NonRootUserError()
     contract_client = contract.UAContractClient(cfg)
     if not args.token:
-        bound_macaroon_bytes = sso.discharge_root_macaroon(contract_client)
+        with util.disable_log_to_console():
```
blackboxsw: This is the context manager call that I think we can avoid  now that you do that in cli.py

blackboxsw: invalid, as you mentioned in IRC. we aren't handling SSOAuth errors etc in : https://github.com/CanonicalLtd/ubuntu-advantage-client/blob/master/uaclient/sso.py#L172-L176. So this stands

PR #558, opened 2019-06-05
travis: use Python 3.7 for mypy runs

OddBloke:
mypy currently fails to install on Python 3.8 (due to it relying on a
version of typed-ast that doesn't yet include this fix:
https://github.com/python/typed_ast/pull/119).  There's no reason to be
running it on 3.8, so just run it on 3.7 instead.


OddBloke:
(Please manually check that the mypy Travis run succeeded; it won't cause overall check failure.)

OddBloke:
(It passed!)

PR #573, opened 2019-06-06
apt: enable -updates iff -updates from the Ubuntu archive is enabled

OddBloke:
Before this, we would match against the -updates stanza for ESM that's
in-place at install time; this would effectively mean that -updates was
always enabled.

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -75,7 +75,16 @@ def add_auth_apt_repo(repo_filename: str, repo_url: str, credentials: str,
 
     # Does this system have updates suite enabled?
     policy, _err = util.subp(['apt-cache', 'policy'])
-    updates_enabled = bool(' %s-updates/' % series in policy)
+    updates_enabled = False
+    for line in policy.splitlines():
+        # We only care about $suite-updates lines
+        if 'a={}-updates'.format(series) not in line:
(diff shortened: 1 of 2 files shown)


OddBloke:
I've tested this locally in a trusty VM and it now behaves as we expect.

uaclient/apt.py
```diff
@@ -75,7 +75,16 @@ def add_auth_apt_repo(repo_filename: str, repo_url: str, credentials: str,
 
     # Does this system have updates suite enabled?
     policy, _err = util.subp(['apt-cache', 'policy'])
-    updates_enabled = bool(' %s-updates/' % series in policy)
+    updates_enabled = False
+    for line in policy.splitlines():
+        # We only care about $suite-updates lines
+        if 'a={}-updates'.format(series) not in line:
```
blackboxsw: Can we combine these two conditionals

# We only care about $suite-updates from the Ubuntu archive
if 'o=Ubuntu,a={}-updates'.format(series) not in line:
    updates_enabled = True
    break

OddBloke: I considered this, but I'm not 100% sure that `o=` and `a=` will always be next to one another in that order, so I went with a more forgiving approach.

Issue #492, opened 2019-05-10
Exit on network connectivity issues during enable


blackboxsw:
If there is no network connection on a machine, ua enable should not proceed to try enabling a support service:

```
 sudo ua enable livepatch
ERROR: Could not refresh machine token. <urlopen error [Errno 101] Network is unreachable>
Installing canonical-livepatch snap... 
ERROR: Failed running cmd: snap install canonical-livepatch, rc: 1 stderr: b'error: cannot install "canonical-livepatch": Post\n       https://api.snapcraft.io/v2/snaps/refresh: dial tcp: lookup\n       api.snapcraft.io on 10.162.96.1:53: server misbehaving\n'
```

blackboxsw:
Ua client should stop and exit non-zero on failure to 'refresh' prior to enabling the service as that will also fail.

OddBloke:
As mentioned in my comment on #574, I'm 100% sure that this is actually desirable behaviour.

blackboxsw:
As agreed in #574, we **want** to proceed if there is an intermittent outage of ua-contracts sever as the contract will most likely not have changed anyway. So, no need to interrupt the user enabling their service if the rest of the service endpoints could be up and running appropriately.

PR #574, opened 2019-06-06
enable: exit non-zero on contract refresh failure

blackboxsw:
We don't want to attempt enabling if we get a failure that contracts
service is unavailable for a couple reasons:
 - stale contract information from ua-contracts service which could have
disabled our services
 - network is down for the whole system which would result in failure to
   enable both the repo-based or livepatch services

Fixes: #492

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -213,7 +213,8 @@ def action_enable(args, cfg):
     """
     logging.debug(ua_status.MESSAGE_REFRESH_ENABLE)
     if not contract.request_updated_contract(cfg):
-        logging.debug(ua_status.MESSAGE_REFRESH_FAILURE)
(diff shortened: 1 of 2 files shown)


OddBloke:
I'm not entirely convinced by this. I agree that if networking is down for the whole system then there's no point proceeding any further, but if the issue is just that the contract server is down then I'm less convinced. People on free contracts, for example, are very unlikely to have an entitlement taken away from them, so using the previous, cached responses would result in the same outcome almost all the time (and even if they had become unentitled, we'd just disable the entitlement next time they interacted with `ua` when the contract server was up).

blackboxsw:
killing this as @OddBloke  is right, typical scenario would really just be that the user would still have an unchanged contract so we only cause them pain if ua-contracts service is intermittently down

PR #575, opened 2019-06-06
config: allow contract_url and sso_auth_url to have a trailing slash

### LINKED ISSUE #566: uaclient.conf should not break on trailing '/' for contract_url value.
If a user added a trailing forward slash to etc/ubuntu-advantage/uaclient.conf for contract_url: 'https://contracts.canonical.com/' **ua attach** will fallover due to a redirect 307 from the contract server and fail to attach due to referencing urls containing two forward slashes like https://contracts.canonical.com//v1/contracts/token.  

The contract server probably shouldn't be redirecting, but UA Client should not 
append a leading '/' if contract_url already contains the slash.

blackboxsw:
Fixes #566

### DIFF
--- a/uaclient/defaults.py
+++ b/uaclient/defaults.py
@@ -7,11 +7,11 @@ any of our dependencies installed.
 
 DEFAULT_CONFIG_FILE = '/etc/ubuntu-advantage/uaclient.conf'
 BASE_AUTH_URL = 'https://login.ubuntu.com'
-BASE_SERVICE_URL = 'https://uaservice.canonical.com'
+BASE_CONTRACT_URL = 'https://contracts.canonical.com'
 
 CONFIG_DEFAULTS = {
     'sso_auth_url': BASE_AUTH_URL,
-    'service_url': BASE_SERVICE_URL,
+    'contract_url': BASE_CONTRACT_URL,
     'data_dir': '/var/lib/ubuntu-advantage',
     'log_level': 'INFO',
     'log_file': '/var/log/ubuntu-advantage.log'
--- a/uaclient/serviceclient.py
+++ b/uaclient/serviceclient.py
@@ -1,6 +1,7 @@
 import abc
 import json
 from urllib import error
+from posixpath import join as urljoin
 
 from uaclient import config
 from uaclient import util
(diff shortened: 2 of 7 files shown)


uaclient/defaults.py
```diff
@@ -7,11 +7,11 @@
 
 DEFAULT_CONFIG_FILE = '/etc/ubuntu-advantage/uaclient.conf'
 BASE_AUTH_URL = 'https://login.ubuntu.com'
-BASE_SERVICE_URL = 'https://uaservice.canonical.com'
+BASE_SERVICE_URL = 'https://contracts.canonical.com'
```
OddBloke: Should we rename this to `BASE_CONTRACT_URL`?

OddBloke:
I'm happy for us to land this, but I'm not sure this is addressing the root cause of #566; we should be constructing our URLs in a way that wouldn't introduce unneeded slashes even if they were present in configuration. (For example, https://stackoverflow.com/a/15279799 shows a way of using `posixpath.join` to do that.)

uaclient/defaults.py
blackboxsw: yes; and done.

blackboxsw:
@OddBloke I agree with your hesitation about how we are constructing url paths. I've added checking user config valuesvia urlparse as well as using urljoin to make our routes. Hopefully this addresses the concern there. 
Thank you for the thoughtful review.

uaclient/defaults.py
blackboxsw: done. good thought.

uaclient/serviceclient.py
```diff
@@ -45,7 +46,7 @@ def request_url(self, path, data=None, headers=None, method=None):
             headers = self.headers()
         if headers.get('content-type') == 'application/json' and data:
             data = json.dumps(data).encode('utf-8')
-        url = getattr(self.cfg, self.cfg_url_base_attr) + path
+        url = urljoin(getattr(self.cfg, self.cfg_url_base_attr), path)
```
OddBloke: Using `urljoin` will stop us from being able to configure base URLs that aren't at the root of a domain:

```
>>> urljoin('http://base/path', '/uh/oh')
'http://base/uh/oh'
```

I think using `posixpath.join` should avoid it.

blackboxsw: though posixpath.join has another iteresting behavior in that if second param starts with leading forward slash, you only get the second value

>>> posixpath.join('https://asdf/', '/oops')
'/oops'

blackboxsw: will sort it to use posixpath appropriately

PR #579, opened 2019-06-07
Refresh keyrings to match current production (and add script to make it easier in future)

### LINKED ISSUE #576: Refresh keyrings to correct keys
These are the keys that we should be using:

```
esm-new:
  precise: 67C7A026
  trusty-security: 13CB4B13
  trusty-updates: 13CB4B13
fips:
  xenial: EDE22758
fips-updates:
  xenial-updates: EDE22758
cc:
  xenial: 3F491C46
```

I've checked and we have the right key for ESM (as we don't need the precise key), but the others are all currently incorrect.

OddBloke:
Fixes #576


tools/refresh-keyrings.sh
```diff
@@ -0,0 +1,43 @@
...
+
+KEYRING_ESM="ubuntu-esm-v2-keyring.gpg"
+KEYRING_FIPS="ubuntu-fips-keyring.gpg"
+KEYRING_FIPS_UPDATES="ubuntu-fips-updates-keyring.gpg"
+KEYRING_CC="ubuntu-cc-keyring.gpg"
+
+KEYS_ESM="13CB4B13"
+KEYS_FIPS="EDE22758"
+KEYS_FIPS_UPDATES="EDE22758"
+KEYS_CC="3F491C46"
```
powersj: Can/should we use the longer key ids?

panlinux: +1, longer keys are a must I believe.

OddBloke: Ack, I'll request the longer key IDs from IS.

powersj: You should confirm these :)

- ESM v2: 4067E40313CB4B13
- FIPS: C1997C40EDE22758
- FIPS Updates: C1997C40EDE22758
- CC: 3A186E733F491C46

OddBloke: Done, and updated.

OddBloke: @powersj  If we don't trust the short key IDs for lookup, I can't exactly use them to look up the keys myself now, can I? :-P

PR #580, opened 2019-06-07
logs: log message about contract refresh now info level

blackboxsw:


uaclient/cli.py
```diff
@@ -211,7 +211,7 @@ def action_enable(args, cfg):
 
     @return: 0 on success, 1 otherwise
     """
-    logging.debug(ua_status.MESSAGE_REFRESH_ENABLE)
+    logging.info(ua_status.MESSAGE_REFRESH_ENABLE)
```
OddBloke: We should be using `print` for messages we explicitly want users to see, so that we can eventually disable logging output to the console entirely.

uaclient/status.py
```diff
@@ -120,7 +120,8 @@ class UserFacingStatus(enum.Enum):
 
 MESSAGE_REFRESH_ENABLE = 'Refreshing contracts prior to enable'
 MESSAGE_REFRESH_SUCCESS = 'Refreshed Ubuntu Advantage contracts.'
-MESSAGE_REFRESH_FAILURE = 'Failure to refresh Ubuntu Advantage contracts.'
+MESSAGE_REFRESH_FAILURE = """\
+Ignoring failure to refresh Ubuntu Advantage contracts."""
```
powersj: Why prompted this change?

uaclient/cli.py
blackboxsw: done! thanks

uaclient/status.py
blackboxsw: only a passing comment from andreas about maybe reformatting the text we log which agreed with my misgiving about starting a log message with 'Failure' in our logs that really is not treated as an operational failure because we ignore it

PR #587, opened 2019-06-07
postrm: remove log files on package purge

### LINKED ISSUE #586: Remove log files on purge
According to https://www.debian.org/doc/debian-policy/ch-files.html#log-files, log files should be removed when the package is purged:

> Log files should be removed when the package is purged (but not when it is only removed). This should be done by the postrm script when it is called with the argument purge (see Details of removal and/or configuration purging).

panlinux:
If purging, remove the log file(s). Globbing there is to catch rotated and compressed or not logs.

Fixes #586

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -15,6 +15,10 @@ remove_cache_dir(){
     rm -rf /var/lib/ubuntu-advantage
 }
 
+remove_logs(){
+    rm -f /var/log/ubuntu-advantage.log*
+}
+
 remove_esm(){
     rm -f /etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg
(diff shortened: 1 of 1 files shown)


debian/postrm
```diff
@@ -15,6 +15,10 @@ remove_cache_dir(){
     rm -rf /var/lib/ubuntu-advantage
 }
 
+remove_logs(){
+    rm -f /var/log/ubuntu-advantage.log*
```
blackboxsw: Why the asterisk?

OddBloke: Log rotation?

panlinux: Globbing there is to catch rotated and compressed or not logs.

powersj: log rotate right?

PR #592, opened 2019-06-07
Docs/debug update

blackboxsw:
update manpage and add cli help documentation for --debug

### DIFF
--- a/ubuntu-advantage.1
+++ b/ubuntu-advantage.1
@@ -76,43 +76,41 @@ Show version of the Ubuntu Advantage package.
 
 
 .SH CONFIGURATION
-By default, ubuntu-advantage client configuration options are read from
-(/etc/ubuntu-advantage/uaclient.conf). If a (uaclient.conf) is present in
-the current working directory it will be used instead of the default conf file.
+By default, Ubuntu Advantage client configuration options are read from
+\fB/etc/ubuntu-advantage/uaclient.conf\fB.
 
 The following configuration options are available:
 .TP
 .B
-\fBcontract_url\fP
-The ubuntu advantage contract server URL. (\fBhttps://contacts.canonical.com\fP)
+\fBsso_auth_url\fP
+The SSO authentication service URL
 .TP
 .B
-\fBsso_auth_url\fP
-The SSO authentication service URL. (\fBhttps://login.ubuntu.com\fP)
+\fBcontract_url\fP
+The ubuntu advantage contract server URL
 .TP
 .B
 \fBdata_dir\fP
-Where ubuntu-advantage stores its data files
-.TP
-.B
-\fBlog_file\fP
-The log file for ubuntu-advantage-tools (\fB/var/log/ubuntu-advantage.log\fP)
+Where Ubuntu Advantage client stores its data files
 .TP
 .B
 \fBlog_level\fP
-The logging level used when writing to log_file. (\fBdebug\fP)
-
+The logging level used when writing to \fBlog_file\fP
+.TP
+.B
+\fBlog_file\fP
+The log file for the Ubuntu Advantage client
 
 .P
 Additionally, any configuration option can be overridden in the environment
-by setting an environment variable prefaced by \fBUA_<uppercase_option_name>\fP.
+by setting an environment variable prefaced by \fBUA_<option_name>\fP. Both
+uppercase and lowercase environment variables are allowed.
 
-For example, the following overrides the contract_url found in uaclient.conf:
+For example, the following overrides the log_level found in uaclient.conf:
 .PP
 .nf
 .fam C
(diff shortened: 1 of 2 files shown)


blackboxsw:
to test:
man ./ubuntu-advantage.1

and 

python3 -m uaclient.cli --help

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
 Show version of the Ubuntu Advantage package.
 
 
+.SH CONFIGURATION
+By default, ubuntu-advantage client configuration options are read from
```
powersj: We typically say "Ubuntu Advantage" above, so let's stick with that

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
...
+(/etc/ubuntu-advantage/uaclient.conf). If a (uaclient.conf) is present in
```
powersj: Add i.e.: `(i.e. /etc/ubuntu-advantage/uaclient.conf). `

Why do we tell them about a local configuration file? I think we should omit that.

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
...
+the current working directory it will be used instead of the default conf file.
+
+The following configuration options are available:
+.TP
+.B
```
powersj: can you sort these the same as the config file itself?

I show:

- sso_auth_file
- contract_url
- data_dir
- log_level
- log_file

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
...
+\fBcontract_url\fP
+The ubuntu advantage contract server URL. (\fBhttps://contacts.canonical.com\fP)
```
powersj: Ubuntu Advantage

Can we also remove the defaults from each option? One less thing to update when we change things.

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
...
+.TP
+.B
+\fBsso_auth_url\fP
+The SSO authentication service URL. (\fBhttps://login.ubuntu.com\fP)
+.TP
+.B
+\fBdata_dir\fP
+Where ubuntu-advantage stores its data files
```
powersj: Where the Ubuntu Advantage client stores data files.

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
...
+.TP
+.B
+\fBlog_file\fP
+The log file for ubuntu-advantage-tools (\fB/var/log/ubuntu-advantage.log\fP)
```
powersj: The log file for the Ubuntu Advantage client.

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
...
+.B
+\fBlog_level\fP
+The logging level used when writing to log_file. (\fBdebug\fP)
+
+
+.P
+Additionally, any configuration option can be overridden in the environment
+by setting an environment variable prefaced by \fBUA_<uppercase_option_name>\fP.
+
+For example, the following overrides the contract_url found in uaclient.conf:
```
powersj: can you show overriding the log level instead? It seems weird to override the contract URL

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
...
+.PP
+.nf
+.fam C
+  UA_CONTRACT_SERVER=https://mycontracts.canonical.com ua attach
+
```
powersj: extra line here

ubuntu-advantage.1
```diff
@@ -75,6 +75,48 @@ Possible values are: active, inactive, or n/a
...
+.TP
+.B
+\fBlog_file\fP
+The log file for ubuntu-advantage-tools (\fB/var/log/ubuntu-advantage.log\fP)
+.TP
+.B
+\fBlog_level\fP
+The logging level used when writing to log_file. (\fBdebug\fP)
+
+
```
powersj: extra line?

ubuntu-advantage.1
blackboxsw: will omit it. We probably shouldn't tell them thx.

ubuntu-advantage.1
blackboxsw: done.

ubuntu-advantage.1
blackboxsw: funny, I had that first, but put in contract_url due to our current testing need for internal canonical. makes sense to do log_level as an example though. fixed

Issue #561, opened 2019-06-06
Loop on invalid 2FA during sso login


blackboxsw:
Upon invalid 2fa, re-prompt for 2FA instead of re-requesting email and password.

One such error that should be handled is redirects on expired valid 2FA (if you miss the 2FA refresh window:

```
$ ua attach
Email: chad.smith@canonical.com
Password: 
Second-factor auth: 708053
ERROR: HTTP Error 307: Temporary Redirect
Could not attach machine. Unable to obtain authenticated contract token
```

Expectation on invalid 2fa:

```
$ ua attach
Email: chad.smith@canonical.com
Password: 
Second-factor auth: invalid
Invalid second-factor auth. Try again: 398413
```

blackboxsw:
The temporary redirect error was a side-effect of having a uaclient.conf which appended two forward slashes on the ua-contracts routes due to bug fixed by #575

PR #598, opened 2019-06-10
attach: two-factor auth reprompt only two-factor auth on failed 2fa

blackboxsw:
Sometimes the 2fa token expires before typing it, or the user mistypes
the 2fa. No longer exit non-zero, forcing the person to re-enter login
and password. Just re-prompt for two factor authentication.

Fixes: #561

### DIFF
--- a/uaclient/sso.py
+++ b/uaclient/sso.py
@@ -18,6 +18,8 @@ API_PATH_TOKEN_DISCHARGE = API_PATH_V2 + '/tokens/discharge'
 # Some Ubuntu SSO API responses use UNDERSCORE_DELIMITED codes, others use
 # lowercase-hyphenated. We'll standardize on lowercase-hyphenated
 API_ERROR_2FA_REQUIRED = 'twofactor-required'
+API_ERROR_INVALID_CREDENTIALS = 'invalid-credentials'
+TWOFACTOR_RETRIES = 1  # Number of times we allow retyping on failed 2FA
 
 
 class MacaroonFormatError(RuntimeError):
@@ -49,7 +51,7 @@ class SSOAuthError(util.UrlError):
     def __contains__(self, error_code):
         return error_code in [error['code'] for error in self.api_errors]
 
-    def __get__(self, error_code, default=None):
+    def __getitem__(self, error_code, default=None):
         for error in self.api_errors:
             if error['code'] == error_code:
                 return error['message']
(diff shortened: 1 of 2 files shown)


uaclient/tests/test_sso.py
```diff
@@ -85,6 +85,24 @@ def __str__(self):
...
+        'error_key,value',
+        ((sso.API_ERROR_2FA_REQUIRED, '2FA required'),
+         (sso.API_ERROR_INVALID_CREDENTIALS, 'Invalid pwd')))
+    def test_api_errors_as_dict_keyed_on_error_code(self, error_key, value):
+        api_errors = [{'code': sso.API_ERROR_2FA_REQUIRED,
+                       'message': '2FA required'},
+                      {'code': sso.API_ERROR_INVALID_CREDENTIALS,
+                       'message': 'Invalid pwd'}]
+        exception = SimpleSSOAuthError("Invalid 2fa", api_errors=api_errors)
+        if value is None:
```
OddBloke: `value` is never None.

uaclient/sso.py
```diff
@@ -173,10 +174,14 @@ def prompt_request_macaroon(cfg: UAConfig, caveat_id: str) -> dict:
...
         except SSOAuthError as e:
-            if API_ERROR_2FA_REQUIRED not in e:
-                raise exceptions.UserFacingError(str(e))
-            args['otp'] = input('Second-factor auth: ')
-            continue
+            if API_ERROR_2FA_REQUIRED in e:
+                args['otp'] = input('Second-factor auth: ')
+                continue
+            elif API_ERROR_INVALID_CREDENTIALS in e:
+                if '2-factor' in e[API_ERROR_INVALID_CREDENTIALS]:
```
OddBloke: Is this comparing against the human-readable message provided by the contract server? Are we sure that message is considered stable by the contract server team?

uaclient/tests/test_sso.py
```diff
@@ -122,6 +140,37 @@ def test_2fa_errors_dont_raise_and_request_extra_input(
...
 
+    @pytest.mark.parametrize(
+        'error_message,reprompt',
+        (('The provided 2-factor key is not recognised.', True),
+         ('Invalid email or password', False)))
+    @mock.patch('uaclient.sso.getpass')
+    @mock.patch('builtins.input')
+    @mock.patch('uaclient.sso.UbuntuSSOClient')
+    def test_2fa_credential_errors_reprompt_for_second_factor_auth_input(
+            self, m_sso_client, m_input, _m_getpass, error_message, reprompt):
```
OddBloke: `reprompt` is never used.

blackboxsw: I forgot to use it; instead performing a string search in the unit test. I've changed the test to make use of reprompt now.

uaclient/tests/test_sso.py
blackboxsw: I forgot to add the third parameterized tuple. just added it.

uaclient/sso.py
blackboxsw: It's comparing thea human-readable message for login.ubuntu.com which is publicly documented as error values. 
https://canonical-identity-provider.readthedocs.io/en/latest/resources/token.html?highlight=invalid-credentials


Just to be sure I confirmed with roadmr in #snapstore channel today that those are expected stable values and he said there are no plans in the future  to change value strings and that part of the code hasn't changed in years.

OddBloke: That documentation appears to suggest that we should be seeing `TWOFACTOR_FAILURE` as the error, not `INVALID_CREDENTIALS`; am I missing something?

blackboxsw: @OddBloke I checked with roadmr about it. He says while it may be an ommision/documentation error, they will not plan to change the error code as they might break existing clients. For the same reason they will also not change the error message because clients may currently be parsing it.

OddBloke: This makes me nervous, because we have to hope that this decision (which is some combination of undocumented and incorrectly documented) doesn't change in the ~8.5 years for which we will be supporting bionic images with the client code that relies on this behaviour.

That said, it's not like we really have a choice of upstream here so I don't see what other option we have, but I'd like @powersj to be aware of this, and sign-off before we land this.

blackboxsw: Sounds fair @OddBloke, though we could future-proof ourselves in the event that they did add an error 'code': 'twofactor-failure' in the future as properly documented.

powersj: Two changes:

1) Only allow the user to re-try 2fa once.
2) I want a comment explainging the use of `INVALID_CREDENTIALS` over `TWOFACTOR_FAILURE`

blackboxsw: Done and done. Thanks gents.

PR #607, opened 2019-06-12
Remove the force parameter from .disable()

### LINKED ISSUE #371: FIPSCommonEntitlement.disable() always returns False
Even when force is passed and it successfully performs all the force actions.

(I'm not sure if this is a bug, techdebt or intentional; I noticed it when reading the code. I've tagged it as both for now.)

### LINKED ISSUE #434: Failure to `apt-get install` during `ua enable fips*` should not attempt to remove packages
If we fail to install packages during `ua enable fips`, we should not attempt to remove any packages we have installed. Doing so is unlikely to appropriately restore the system to its previous state, and carries some amount of risk of breaking the system further. Instead, we should clearly report the errors and allow the user to resolve it as they see fit.

OddBloke:
The first commit fixes #434 by not performing any actions on disable of a FIPS entitlement (other than messaging).

This meant that the only place passing `force=True` to a `disable()` call was the exception handler for failed installs in `RepoEntitlement.enable()`. As this is happening _during_ enable, it can't logically be a "disable" because we aren't enabled yet. So to clean up the logic, I then refactored the cleanup code to a `_cleanup()` method which both the exception handler and `.disable()` call. (I stopped short of introducing `.cleanup()` on `UAEntitlement` because at the moment we only have one caller of it, in a sub-class.)

This then meant that nothing was calling `.disable()` with `force=True`, so we could remove it from the definition of `disable()`.

This, in turn, meant that nothing was calling `can_disable()` with `force=True`, so we could also remove it from there (and then simplify the implementation).

(This also fixes #371, because `False` is now the appropriate thing to return at all time.)

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -68,23 +68,20 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         """
         pass
 
-    def can_disable(self, silent: bool = False, force: bool = False) -> bool:
+    def can_disable(self, silent: bool = False) -> bool:
         """Report whether or not disabling is possible for the entitlement.
 
         @param silent: Boolean set True to silence printed messages/warnings.
-        @param force: Boolean set True to allow disable even if entitlement
-            doesn't appear 'enabled'.
         """
         message = ''
         retval = True
         application_status, _ = self.application_status()
 
-        if not force:
-            if application_status == status.ApplicationStatus.DISABLED:
-                message = status.MESSAGE_ALREADY_DISABLED_TMPL.format(
-                    title=self.title
-                )
-                retval = False
+        if application_status == status.ApplicationStatus.DISABLED:
+            message = status.MESSAGE_ALREADY_DISABLED_TMPL.format(
+                title=self.title
+            )
+            retval = False
         if message and not silent:
             print(message)
         return retval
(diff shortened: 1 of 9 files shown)


uaclient/entitlements/base.py
```diff
@@ -68,26 +68,19 @@ def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
...
-        if not force:
-            if application_status == status.ApplicationStatus.DISABLED:
-                message = status.MESSAGE_ALREADY_DISABLED_TMPL.format(
-                    title=self.title
-                )
-                retval = False
-        if message and not silent:
-            print(message)
-        return retval
+        if application_status == status.ApplicationStatus.DISABLED:
```
blackboxsw: good to drop all those extra local vars too. Who would do such a thing in the first place? :)

uaclient/entitlements/repo.py
```diff
@@ -82,7 +79,7 @@ def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
                     ['apt-get', 'install', '--assume-yes'] + self.packages,
                     capture=True, retry_sleeps=APT_RETRIES)
             except util.ProcessExecutionError:
-                self.disable(silent=True, force=True)
+                self._cleanup()
```
blackboxsw: Isn't this _cleanup  still called from fips on failure to install one of fips packages?

Issue #352, opened 2019-04-11
Remove cache dir/file on postinst, not preinst


panlinux:
We should remove the cache dir and file, from old ua, in postinst, not preinst:
```
<vorlon> andreas: I don't see any reason for the preinst bits to happen in preinst instead of postinst, can you explain?
<andreas> vorlon: without that, dpkg was issuing a warning that it couldn't remove a non-empty directory
<vorlon> oh
<andreas> when upgrading from a package that had the cache file in there, i.e., cron.daily ran at least once in the past
<vorlon> those warnings are no big deal as long as your maintainer script takes care of it afterwards
<andreas> ok, but by removing the file, and leaving the directory, dpkg removed the (now empty) directory in the end
<vorlon> so I would remove the cache file and the cache dir in the postinst, not in the preinst, because preinst scripts make apt/dpkg more rittle
<vorlon> brittle
<vorlon> (this is also not a blocker)
```

powersj:
Can you confirm this is for the old ua client? Was this done?

I'm inclined not to do this unless we absolutely must since we don't really want to be uploading the old client.

panlinux:
This is for the new client that was uploaded to disco.

panlinux:
By moving it to postinst, we get the dpkg warning as expected:
```
$ sudo dpkg -i ../*.deb
(Reading database ... 95947 files and directories currently installed.)
Preparing to unpack .../ubuntu-advantage-tools_19.5_amd64.deb ...
Unpacking ubuntu-advantage-tools (19.5) over (17) ...
dpkg: warning: unable to delete old directory '/var/cache/ubuntu-advantage-tools': Directory not empty
Setting up ubuntu-advantage-tools (19.5) ...
Removing obsolete conffile /etc/update-motd.d/80-esm ...
Removing obsolete conffile /etc/update-motd.d/80-livepatch ...
Removing obsolete conffile /etc/cron.daily/ubuntu-advantage-tools ...
Processing triggers for man-db (2.8.3-2ubuntu0.1) ...
```
But the directory is gone in the end:
```$ l /var/cache/ubuntu-advantage*
ls: cannot access '/var/cache/ubuntu-advantage*': No such file or directory
```

powersj:
Closed due to 14ad98d

PR #609, opened 2019-06-12
Remove the old CACHE_DIR in postinst

panlinux:
We can just rm -rf the directory, it's quicker and simpler.

`d/preinst` had only this content, and is no longer necessary.

I was trying to remember why I was so careful with `dpkg --compare-versions`, but couldn't come up with a reasonable explanation. Maybe because it was in `preinst`, or I wanted to avoid a useless `rm -rf` on every upgrade, but `dpkg --compare-versions` doesn't look cheaper. Feel free to point out flaws in this thinking.

Note that `d/preinst` is still needed because of `d/ubuntu-advantage-tools.maintscript`, which adds code to the `#DEBHELPER#` marker all over the place.

Fixes issue #352

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -50,6 +50,10 @@ case "$1" in
       if dpkg --compare-versions "$2" lt-nl "19.5~"; then
           upgrade_to_status_cache
       fi
+
+      # CACHE_DIR is no longer present or used since 19.1
+      rm -rf /var/cache/ubuntu-advantage-tools
+
       grep -iq trusty /etc/os-release && configure_esm
       if [ ! -f /var/log/ubuntu-advantage.log ]; then
           touch /var/log/ubuntu-advantage.log
--- a/debian/preinst
+++ b/debian/preinst
@@ -20,16 +20,7 @@ case "$1" in
     install)
     ;;
     upgrade)
-        # CACHE_DIR is no longer present or used since 19.1
-        # dpkg will take care of removing CACHE_DIR, as long
-        # as it's empty
-        if dpkg --compare-versions "$2" lt-nl "19.1~"; then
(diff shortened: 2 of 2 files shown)


basak:
> Note that d/preinst is still needed because of d/ubuntu-advantage-tools.maintscript, which adds code to the #DEBHELPER# marker all over the place.

I believe that if you remove a maintainer script entirely, debhelper will still do the right thing, so I think you should remove it and it will be fine.

> I was trying to remember why I was so careful with dpkg --compare-versions, but couldn't come up with a reasonable explanation.

I'd say that it makes it really easy to verify later when it is that you can drop the upgrade path in the future from the guard alone. Also, it stops the mildly surprising behaviour of deleting the file if the user happens to recreate it or something (eg. backup restore). The upgrade path itself is equally surprising but is necessary, so I would keep the surprise to a minimum by guarding it as you had done before. However, in this particular case all that seems unlikely and perhaps not relevant, so it's just a style thing and so entirely up to you.

+1 in that it looks to me that this does what you want it to do, though I think you should remove the preinst.

I'm not sure it's worth changing, but I also don't understand Steve's "more brittle" comment. Did Steve intend that you change it, or was he just advising for future reference? I don't see how the previous preinst would cause a problem in this particular case - it looks trivially correct to me (though I would have quoted the `rm` argument), unless the lack of unwinding is a problem (which I suspect it isn't given that it's a cache).

basak:
> ...though I think you should remove the preinst.

I mean if you are moving the code to the postinst that is, rather than not changing anything.

powersj:
Merging based on +1 from @basak

panlinux:
We talked about that preinst-to-postinst move again in a meeting following that upload, and we confirmed it should be done, and the dpkg warning message was ok as long as the postinst really took care of the removal in the end.

PR #610, opened 2019-06-12
Be consistent when using Ubuntu release names

### LINKED ISSUE #314: Be consistent when using Ubuntu release names
We should probably capitalize the Ubuntu series name:

```
$ sudo ua enable fips
FIPS is not available for Ubuntu trusty.
```
We could have a nice mapping between code names ("trusty") and visible names, something like:
`trusty -> Ubuntu 14.04 LTS ("trusty")`

`/etc/lsb-release` has a description field. For trusty, for example, it is `DISTRIB_DESCRIPTION="Ubuntu 14.04.6 LTS"`

OddBloke:
This does the following:

* remove some unused members from the return value of `get_platform_info` and cleans up its regexes
* adds determination of the user-facing version string for our current platform as 'version' to the return value of `get_platform_info`
    * The normalisation performed to get a consistent user-facing string also means we can simplify down to just having a single regex which operates against the normalised version
* uses that value when informing users that their current Ubuntu release is unsupported

Fixes #314

### DIFF
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -324,6 +324,12 @@ REGEX_OS_RELEASE_VERSION = r'(?P<version>\d+\.\d+) (LTS )?\((?P<series>\w+).*'
 
 
 def get_platform_info() -> 'Dict[str, str]':
+    """
+    Returns a dict of platform information.
+
+    N.B. This dict is sent to the contract server, which requires the
+    distribution, type and release keys.
+    """
     os_release = parse_os_release()
     platform_info = {
         'distribution': os_release.get('NAME', 'UNKNOWN'),
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -335,6 +335,8 @@ def get_platform_info() -> 'Dict[str, str]':
     if ', ' in version:
         # Fix up trusty's version formatting
         version = '{} ({})'.format(*version.split(', '))
+    # Strip off an LTS point release (14.04.1 LTS -> 14.04 LTS)
+    version = re.sub(r'\.\d LTS', ' LTS', version)
     platform_info['version'] = version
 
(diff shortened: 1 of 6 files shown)


OddBloke:
(https://twitter.com/Odd_Bloke/status/1138896327344054272)

OddBloke:
Oops, I missed a use of `get_platform_info`, so some of that refactoring is incorrect; fixing now.

OddBloke:
OK, we're back to this being reviewable.

OddBloke:
Example output:

```
$ sudo ua enable esm
Refreshing contracts prior to enable
Extended Security Maintenance is not available for Ubuntu 16.04 LTS (Xenial Xerus).
```

uaclient/util.py
```diff
@@ -319,28 +319,35 @@ def parse_os_release(release_file: 'Optional[str]' = None) -> 'Dict[str, str]':
...
         'distribution': os_release.get('NAME', 'UNKNOWN'),
         'type': 'Linux'}
 
     version = os_release['VERSION']
-    match = re.match(REGEX_OS_RELEASE_VERSION_1, version)
-    if not match:
-        match = re.match(REGEX_OS_RELEASE_VERSION_2, version)
+    if ', ' in version:
+        version = '{} ({})'.format(*version.split(', '))
+    version = re.sub(r'\.\d LTS', ' LTS', version)
```
blackboxsw: Do we need a '+'        r'\.\d+ LTS'   ?

OddBloke: We don't, this is just stripping off the point release, so there should be exactly one digit. (I don't know what pre-point-release LTS's look like, but if they are "20.04 LTS" then this would be fine but \d+ would give us "20 LTS", I think.)

I'll add a comment to make what this is doing clearer.

PR #615, opened 2019-06-13
attach: add --no-auto-enable option

### LINKED ISSUE #366: Support disabling auto-enablement of entitlements on attach
Once #360 is fully implemented, we will want a way to disable the automatic enablement of entitlements at `ua attach` time. This will allow users who know precisely what entitlements they want enabled/disabled to only enable those they want enabled.

OddBloke:
This will skip enabling any services that are marked as autoEnable.

Fixes #366


OddBloke:
(I've tested this on xenial with both token and SSO attach.)

Issue #613, opened 2019-06-13
valid_apt_credentials check doesn't properly message about repo timeouts


blackboxsw:
When apt-helper is unable to contact the configured repo aptURL on curl timeout it emits the following incorrect error message:
ERROR: Invalid APT credentials provided for https://esm.staging.ubuntu.com/cc

Since staging apt repos are behind the VPN, having the VPN off will result in a timeout trying to access those repos and valid_apt_credentials should have a more correct error message in this case.

apt-helper script spefically has an exit code 100 and the followig error message that uaclient should better handle

Turns out, invalid credentials also emits and exit code of 100. So we'll need to match 401 Unauthorized vs Connection timed out'

E: Failed to fetch https://esm.staging.ubuntu.com/cc/ubuntu/pool/  Connection timed out after 120001 milliseconds\n\nE: Download Failed\n

Steps to reproduce:
 0. disable Canonical vpn
 1. deploy xenial vm
 2. configure uaclient to talk to staging contracts service
 3. sudo ua attach
 4. sudo ua enable cc-eal
  ... wait 2 mins for a timeout

PR #619, opened 2019-06-14
apt: failure to acccess repo url will result in timeout error message

blackboxsw:
Refactor and valid_apt_credentials to assert_valid_apt_credentials.
The new function will raise different UserFacingError messages on
repo url timeout versus credentials/authentication errors.

Fixes: #613

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -4,6 +4,7 @@ import os
 import re
 import shutil
 
+from uaclient import exceptions
 from uaclient import util
 
 try:
@@ -22,36 +23,38 @@ APT_METHOD_HTTPS_FILE = '/usr/lib/apt/methods/https'
 CA_CERTIFICATES_FILE = '/usr/sbin/update-ca-certificates'
 
 
-class InvalidAPTCredentialsError(RuntimeError):
-    """Raised when invalid token is provided for APT access"""
-    pass
-
-
-def valid_apt_credentials(repo_url, username, password):
+def assert_valid_apt_credentials(repo_url, username, password):
     """Validate apt credentials for a PPA.
 
     @param repo_url: private-ppa url path
     @param username: PPA login username.
     @param password: PPA login password or resource token.
 
-    @return: True if valid or unable to validate
+    @raises: UserFacingError for invalid credentials, timeout or unexpected
+        errors.
     """
     protocol, repo_path = repo_url.split('://')
     if not os.path.exists('/usr/lib/apt/apt-helper'):
-        return True   # Do not validate
+        return
     try:
         util.subp(['/usr/lib/apt/apt-helper', 'download-file',
                    '%s://%s:%s@%s/ubuntu/pool/' % (
                        protocol, username, password, repo_path),
-                   '/tmp/uaclient-apt-test'],
-                  capture=False)  # Hide credentials from logs
-        return True
-    except util.ProcessExecutionError:
-        pass
+                   '/tmp/uaclient-apt-test'])
+    except util.ProcessExecutionError as e:
+        if e.exit_code == 100:
+            stderr = str(e.stderr).lower()
+            if re.search(r'401\s+unauthorized|httperror401', stderr):
+                raise exceptions.UserFacingError(
+                    'Invalid APT credentials provided for %s' % repo_url)
+            elif re.search(r'connection timed out', stderr):
+                raise exceptions.UserFacingError(
+                    'Timeout trying to access APT repository at %s' % repo_url)
+        raise exceptions.UserFacingError(
+            'Unexpected APT error. See /var/log/ubuntu-advantage.log')
     finally:
         if os.path.exists('/tmp/uaclient-apt-test'):
             os.unlink('/tmp/uaclient-apt-test')
-    return False
 
 
 def add_auth_apt_repo(repo_filename: str, repo_url: str, credentials: str,
@@ -69,9 +72,7 @@ def add_auth_apt_repo(repo_filename: str, repo_url: str, credentials: str,
     series = util.get_platform_info()['series']
     if repo_url.endswith('/'):
         repo_url = repo_url[:-1]
-    if not valid_apt_credentials(repo_url, username, password):
-        raise InvalidAPTCredentialsError(
-            'Invalid APT credentials provided for %s' % repo_url)
+    assert_valid_apt_credentials(repo_url, username, password)
 
     # Does this system have updates suite enabled?
     policy, _err = util.subp(['apt-cache', 'policy'])
--- a/uaclient/tests/test_apt.py
+++ b/uaclient/tests/test_apt.py
@@ -12,8 +12,8 @@ from uaclient.apt import (
     APT_AUTH_COMMENT, add_apt_auth_conf_entry, add_auth_apt_repo,
     add_ppa_pinning, clean_apt_sources, find_apt_list_files,
     get_installed_packages, remove_apt_list_files, remove_auth_apt_repo,
-    remove_repo_from_apt_auth_file, valid_apt_credentials)
-from uaclient import apt, util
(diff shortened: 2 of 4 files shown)


uaclient/apt.py
```diff
@@ -22,36 +23,37 @@
 CA_CERTIFICATES_FILE = '/usr/sbin/update-ca-certificates'
 
 
-class InvalidAPTCredentialsError(RuntimeError):
```
blackboxsw: I'm uncertain whether we want specialized exceptions for InvalidAptCredentials vs AptRepoTimeout, or just raising UserFacingError with the proper error message. Since we don't really react differently to a raised InvalidAptCreds vs Timeout, I went with using UserFacingError with the proper error message.

uaclient/apt.py
```diff
@@ -22,36 +23,37 @@
...
-    """Raised when invalid token is provided for APT access"""
-    pass
-
-
-def valid_apt_credentials(repo_url, username, password):
+def assert_valid_apt_credentials(repo_url, username, password):
```
blackboxsw: I felt like logic for raised exceptions should live inside valid_apt_credentials instead of returning True/False and then trying to parse apt-helper stderror content for 401s up in call sites. That prompted me to move the construction of the exception types inside this function.

uaclient/apt.py
```diff
@@ -22,36 +23,37 @@
...
     @param password: PPA login password or resource token.
 
-    @return: True if valid or unable to validate
+    @raises: UserFacingError for invalid credentials or timeout
+    @raises: ProcessExecutionError for unexpected errors
     """
     protocol, repo_path = repo_url.split('://')
     if not os.path.exists('/usr/lib/apt/apt-helper'):
-        return True   # Do not validate
+        return None
```
OddBloke: As we aren't ever returning a value, just `return` here would be clearer.

uaclient/apt.py
OddBloke: This could feasibly return `Tuple[bool, str]` (or even just `Optional[str]` with `None` indicating success), but as we don't have a consumer that _doesn't_ want the exceptions, I think what you're doing here is the correct choice.

uaclient/apt.py
OddBloke: I think there are a couple of cases where we would want specialised exceptions: when we want to be able to catch a particular error (and so need a distinct class for the `except` clause), and when we want to produce the same error in multiple places (and want to ensure that the text remains consistent across all error sites).

I don't believe either of those apply here, so +1 on just `UserFacingError`.

uaclient/apt.py
```diff
@@ -22,36 +23,37 @@
...
+    except util.ProcessExecutionError as e:
+        if e.exit_code == 100:
+            stderr = str(e.stderr)
+            if re.search(r'401\s+Unauthorized|HttpError401', stderr):
+                raise exceptions.UserFacingError(
+                    'Invalid APT credentials provided for %s' % repo_url)
+            elif re.search(r'Connection timed out', stderr):
+                raise exceptions.UserFacingError(
+                    'Timeout trying to access %s' % repo_url)
+        raise
```
OddBloke: Should we raise a catch-all `UserFacingError` here, so that unexpected problems don't cause a traceback?

uaclient/tests/test_apt.py
```diff
@@ -101,46 +101,63 @@ def test_valid_apt_credentials_true_when_missing_apt_helper(
...
+        'exit_code,stderr,error_msg',
+        ((1, 'something broke',
+          "Failed running command 'apt-helper ' [exit(1)]. "
+          "Message: something broke"),
+         (100, 'E: Failed to fetch ... HttpError401 on trusty',
+          'Invalid APT credentials provided for http://fakerepo'),
+         (100, 'E: Failed to fetch ... 401 Unauthorized on xenial',
+          'Invalid APT credentials provided for http://fakerepo'),
+         (100, 'E: Failed to fetch ... Connection timed out',
+          'Timeout trying to access http://fakerepo')))
```
OddBloke: This is quite a generic error message. I wonder if this message might be clearer as: 'Timeout trying to access APT repository at http://fakerepo'

uaclient/apt.py
```diff
@@ -22,36 +23,37 @@
...
-        pass
+                   '/tmp/uaclient-apt-test'])
+    except util.ProcessExecutionError as e:
+        if e.exit_code == 100:
+            stderr = str(e.stderr)
+            if re.search(r'401\s+Unauthorized|HttpError401', stderr):
+                raise exceptions.UserFacingError(
+                    'Invalid APT credentials provided for %s' % repo_url)
+            elif re.search(r'Connection timed out', stderr):
+                raise exceptions.UserFacingError(
```
OddBloke: Raising a UserFacingError for this means that we won't retry, but I think a connection timeout is something we should be retrying (because it could be a transient networking issue or temporary server outage).

uaclient/apt.py
OddBloke: OK, having read through more carefully, I think we may actually want something like a `RetryableUserFacingError`, which the subp retrying logic will retry on (but which will still output nicer user output by dint of being a subclass of `UserFacingError`).

(I still don't think we need an exception per-message here, because we're only using each one once.)

uaclient/apt.py
blackboxsw: good thought. Fixing.

uaclient/apt.py
OddBloke: Oh, hold on, I think I have things mixed up here; we've already done all the retrying we would do in the util.subp call so disregard this.

blackboxsw: Hrm, in this specific case the timeout on apt-helper is 2 mins, this makes a  retry here crazy long.

blackboxsw: Ok, so we likely aren't going to retry this specific timeout case at this time. a user facing error will allow someone to ua enable esm again if they need to for the time being, I've filed #626 to capture the timeout length discussion to see if we think we should reduce on that timeout a bit.

uaclient/apt.py
```diff
@@ -22,36 +23,37 @@
...
+    except util.ProcessExecutionError as e:
+        if e.exit_code == 100:
+            stderr = str(e.stderr).lower()
+            if re.search(r'401\s+unauthorized|httperror401', stderr):
+                raise exceptions.UserFacingError(
+                    'Invalid APT credentials provided for %s' % repo_url)
+            elif re.search(r'connection timed out', stderr):
+                raise exceptions.UserFacingError(
+                    'Timeout trying to access APT repository at %s' % repo_url)
+        raise exceptions.UserFacingError(str(e))
```
OddBloke: What is `str(e)` going to look like here?  If it's more than a single-line summary then I don't think we should pipe that straight out to the user.

OddBloke: (FAOD, and to save another round-trip, if it is more verbose than we want, I think we should just emit "unknown apt problem sorry" and people can reference the logs or run with --debug to get more.)

blackboxsw: Your assumption is correct here. other errors are a bit too long for taste.

ERROR: Failed running command '/usr/lib/apt/apt-helper2 download-file https://bearer:mAgJhAwoQ4Hs8wpIhS-nr72eEdrg9gBIBMBpJCkRtdC0AAMpzfzfXBQmgTvoLGE9-Rynn_lNCNulp4v_VvVoaiD05YWEzMjlhZWE5YTc4NGUzNGUzN2NiYjQ1Y2ZmYzRmNBIBKgAABiCbND3Clmt8PKmmTgG_eYiwLbqzOKouJsIDrnr7rmfGsg@esm.staging.ubuntu.com/ubuntu/pool/ /tmp/uaclient-apt-test' [exit(1)]. Message: b'my error string\n'

PR #620, opened 2019-06-14
apt-hook: Add missing headers for APT 1.9

julian-klode:
APT 1.9 cleans up the APT headers ever so slightly, so we
need to include a few more.


panlinux:
Does this build with old apt? Trusty has 1.0.1ubuntu2.23

julian-klode:
Yeah, that should be fine. It's jsut two more includes, and those are very old headers that exist back there.

OddBloke:
I've confirmed that this still builds on trusty, thanks!

PR #625, opened 2019-06-14
livepatch: use /usr/bin/snap directly and raise an error if snapd is not installed

### LINKED ISSUE #356: Livepatch entitlement uses `which snap` to determine if snapd is installed
`snapd` is not the only package to provide a snap command (at least in trusty):

```
$ apt-file search -x bin/snap$
snap: /usr/bin/snap
snapd: /usr/bin/snap
```

(This is low priority to fix, but is nonetheless a possible failure case.)

OddBloke:
This uses the full path to `/usr/bin/snap` to ensure that we don't end up using some other binary named `snap` that happens to be in the PATH on a system. And, as there is more than one package in trusty that installs `/usr/bin/snap`, we ensure that `snapd` is installed before using `/usr/bin/snap`.

Fixes #356

### DIFF
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -2,7 +2,7 @@ import logging
 
 from uaclient.entitlements import base
 from uaclient.entitlements.repo import APT_RETRIES
-from uaclient import status
+from uaclient import apt, exceptions, status
 from uaclient import util
 from uaclient.status import ApplicationStatus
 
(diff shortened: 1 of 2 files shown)


uaclient/entitlements/livepatch.py
```diff
@@ -46,15 +46,19 @@ def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
         if not self.can_enable(silent=silent_if_inapplicable):
             return False
         if not util.which('/snap/bin/canonical-livepatch'):
-            if not util.which('snap'):
+            if not util.which('/usr/bin/snap'):
```
blackboxsw: Worth a global? 
SNAP_CMD='/usr/bin/snap'

OddBloke: Yep, good shout.

PR #627, opened 2019-06-14
tools: remove bddeb and scripts only used by it

OddBloke:
bddeb is designed to generate an upstream tarball and handle
inconsistencies across major Python versions.  As we're a native package
that only supports Python 3, people should instead be able to build
the package using just regular package-build tooling.


Makefile
```diff
@@ -1,25 +1,13 @@
-
-
 build:
 	$(MAKE) -C apt-hook build
 
-deps:
-	./tools/read-dependencies -v 3 --system-pkg-names --test-distro
```
blackboxsw: I use this target to quickly setup a clean vm that can be used as a dev/test environment, maybe we can migrate some of your multipass goodness in here to make that setup easy:
deps:
        k-build-deps -t "apt-get --no-install-recommends --yes" -i ./debian/control

Issue #359, opened 2019-04-11
test_cis should include a test that `apt-transport-https` and `ca-certificates` are installed if necessary


OddBloke:
Currently we only test one case (which, until #357 is fixed, is the case of the system on which the tests are running).

(We should also consider whether we can drop and/or consolidate the "enable" testing for the RepoEntitlement subclasses.)

powersj:
CIS is a 19.7 item, so moving milestone.

PR #634, opened 2019-06-17
entitlements/tests: refactor to a shared entitlement_factory fixture

OddBloke:
This reduces a lot of re-definition of common data structures, making it
clearer what actually differs between different test cases.


OddBloke:
When looking at #359, the inconsistent implementation of the entitlement tests was making it difficult to determine what tests were and weren't shared; this is a step towards rationalising those tests.

Issue #626, opened 2019-06-14
apt credentials validation has a 2 minute default timeout when repo is inaccessible.


blackboxsw:
UA Client runs apt-helper tool to quickly validate crednetials against an apt repository url. This tools has 2 minute default timeout.  If the repo is inaccessible, the commandline `ua enable esm|fips*|cc-eal` could sit for 2 minutes before giving up. 
Do we want to reduce this timeout more:

Options to reduce timeout wait:
   1. Provide apt config Acquire::http::Timeout  <seconds> option when calling apt-helper (via APT_CONFIG=<filename> environment variable)
   2.  Provide uaclient.util.subp a timeout parameter

PR #654, opened 2019-06-20
subp: reduce apt-helper call timeout from 2 mins to 20 seconds

blackboxsw:
apt has a default timeout of 2 minutes for tools like apt-helper.

Add a timeout parameter to util.subp allowing ua client to raise custom
timeouts from subprocess.communicate for specific commands.

Apply a 20 second timeout to the apt-helper call used when validating
credentials against an apt repo url. If we don't see a response from the
repo in that apt-helper call in 20 seconds, chances are we something is
wrong that a 2 minute timeout isn't going to resolve.

Fixes: #626

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -5,6 +5,7 @@ import re
 import shutil
 
 from uaclient import exceptions
+from uaclient import status
 from uaclient import util
 
 try:
@@ -21,6 +22,12 @@ APT_KEYS_DIR = '/etc/apt/trusted.gpg.d'
 KEYRINGS_DIR = '/usr/share/keyrings'
 APT_METHOD_HTTPS_FILE = '/usr/lib/apt/methods/https'
 CA_CERTIFICATES_FILE = '/usr/sbin/update-ca-certificates'
+
+# Since we generally have a person at the command line prompt. Don't loop
+# for 5 minutes like charmhelpers because we expect the human to notice and
+# resolve to apt conflict or try again.
+# Hope for an optimal first try.
+APT_RETRIES = [1.0, 5.0, 10.0]
 
 
 def assert_valid_apt_credentials(repo_url, username, password):
@@ -57,6 +64,21 @@ def assert_valid_apt_credentials(repo_url, username, password):
             os.unlink('/tmp/uaclient-apt-test')
 
 
+def run_apt_command(cmd, error_msg) -> str:
+    """Run an apt command, retrying upon failure APT_RETRIES times.
+
+    :return: stdout from successful run of the apt command.
+    :raise UserFacingError: on issues running apt-cache policy.
+    """
+    try:
+        out, _err = util.subp(cmd, capture=True, retry_sleeps=APT_RETRIES)
+    except util.ProcessExecutionError as e:
+        if 'Could not get lock /var/lib/dpkg/lock' in str(e.stderr):
+            error_msg += ' Another process is running APT.'
+        raise exceptions.UserFacingError(error_msg)
+    return out
+
+
 def add_auth_apt_repo(repo_filename: str, repo_url: str, credentials: str,
                       suites: 'List[str]', keyring_file: str = None) -> None:
     """Add an authenticated apt repo and credentials to the system.
@@ -75,8 +97,9 @@ def add_auth_apt_repo(repo_filename: str, repo_url: str, credentials: str,
     assert_valid_apt_credentials(repo_url, username, password)
 
     # Does this system have updates suite enabled?
-    policy, _err = util.subp(['apt-cache', 'policy'])
     updates_enabled = False
+    policy = run_apt_command(
+        ['apt-cache', 'policy'], status.MESSAGE_APT_POLICY_FAILED)
     for line in policy.splitlines():
         # We only care about $suite-updates lines
         if 'a={}-updates'.format(series) not in line:
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -3,6 +3,7 @@ import logging
 import os
 import re
 import shutil
+import subprocess
 
 from uaclient import exceptions
 from uaclient import status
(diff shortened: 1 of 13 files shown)


uaclient/tests/test_apt.py
```diff
@@ -152,7 +152,34 @@ def test_errors_on_process_execution_errors(
...
+        """Raise the appropriate user facing error from apt-helper timeout."""
+
+        # Failure apt-helper response
+        m_subp.side_effect = exceptions.CommandTimeoutError(
+            cmd=['apt-helper', 'SECRETS'], timeout=10)
+        with pytest.raises(exceptions.CommandTimeoutError) as excinfo:
+            assert_valid_apt_credentials(
+                repo_url='http://fakerepo', username='user', password='pwd')
+        error_msg = (
+            'Timeout after 20 seconds: cannot validate credentials for APT'
```
OddBloke: This error message feels the wrong way around to me; it's structured as "specific: general". I think something more like "Cannot validate credentials for APT: timeout after 20 seconds" feels clearer to me, as it's telling me what failed and then giving me the specific reason for that failure.

uaclient/util.py
```diff
@@ -227,13 +233,15 @@ def _subp(args: 'Sequence[str]', rcs: 'Optional[List[int]]' = None,
...
             bytes_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        (out, err) = proc.communicate()
+        (out, err) = proc.communicate(timeout=timeout)
     except OSError:
         try:
             raise ProcessExecutionError(
                 cmd=' '.join(args), exit_code=proc.returncode, stderr=err)
         except UnboundLocalError:
             raise ProcessExecutionError(cmd=' '.join(args))
+    except subprocess.TimeoutExpired:
```
OddBloke: I believe the only way this would be raised is if a consumer passed in a timeout. As such, I think we can let the `subprocess` exception bubble out to the callers, because they know definitively that they need to handle it.

I think this would also mean that we would no longer need a generic "timeout" `UserFacingException`; each caller can catch `subprocess.TimeoutExpired` and raise an exception with exactly the message that they want.

uaclient/util.py
```diff
@@ -245,14 +253,16 @@ def _subp(args: 'Sequence[str]', rcs: 'Optional[List[int]]' = None,
 
 
 def subp(args: 'Sequence[str]', rcs: 'Optional[List[int]]' = None,
-         capture: bool = False,
+         capture: bool = False, timeout=None,
```
OddBloke: Type annotation, please!

uaclient/util.py
```diff
@@ -208,17 +210,21 @@ def readurl(url: str, data: 'Optional[bytes]' = None,
     return content, resp.headers
 
 
-def _subp(args: 'Sequence[str]', rcs: 'Optional[List[int]]' = None,
-          capture: bool = False) -> 'Tuple[str, str]':
+def _subp(args: 'Sequence[str]',
+          rcs: 'Optional[List[int]]' = None,
+          capture: bool = False,
+          timeout: 'Optional[int]' = None) -> 'Tuple[str, str]':
```
OddBloke: Definitely not a big deal, but timeout on communicate is actually a float; no need for us to be more restrictive: https://github.com/python/typeshed/blob/9941616b9efd31181bf3e5dd1d5f375791352fec/stdlib/3/subprocess.pyi#L318-L322

uaclient/apt.py
```diff
@@ -14,6 +15,7 @@
     # typing isn't available on trusty, so ignore its absence
     pass
 
+APT_HELPER_TIMEOUT = 20.0    # 20 second timeout used for apt-helper call
```
OddBloke: Absolutely no problems defining this as a float. But as a note, to make sure my previous comment wasn't misunderstood, an `int` is an acceptable value for a variable of type `float`:

```
$ echo 'x: float = 1' > test.py
$ mypy --strict test.py
$ echo $?
0
```

though the inverse is not true:

```
$ echo 'x: int = 1.0' > test.py
$ mypy --strict test.py
test.py:1: error: Incompatible types in assignment (expression has type "float", variable has type "int")
```

uaclient/status.py
```diff
@@ -115,6 +115,9 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_ATTACH_SUCCESS_TMPL = """\
 This machine is now attached to '{contract_name}'.
 """
+MESSAGE_CMD_TIMEOUT_TMPL = """\
```
OddBloke: I think this is now unused.

uaclient/tests/test_apt.py
```diff
@@ -152,7 +153,34 @@ def test_errors_on_process_execution_errors(
...
+        """Raise the appropriate user facing error from apt-helper timeout."""
+
+        # Failure apt-helper response
+        m_subp.side_effect = subprocess.TimeoutExpired(
+            'something timed out', timeout=1000000)
+        with pytest.raises(exceptions.UserFacingError) as excinfo:
+            assert_valid_apt_credentials(
+                repo_url='http://fakerepo', username='user', password='pwd')
+        error_msg = (
+            'Cannot validate credentials for APT repo: fakerepo. Timeout'
```
OddBloke: If you're concerned about a long repo URL (which is definitely a valid concern!), we could consider making this something like:

ERROR: Cannot validate credentials for APT repo. Timeout after 20 seconds trying to reach https://esm.ubuntu.com/this/is/a/long/url/but/its/at/the/end/so/its/not/so/bad.

blackboxsw: +1

uaclient/apt.py
blackboxsw: Good point, and thanks for clarification, I wanted the float there as a hint of the type that is allowed in case we are tweaking that number at the declaration level and don't see that floats are allowed.

Issue #482, opened 2019-05-07
use columns for first set of ua status information as well


cpaelzer:
Today on `ua status` and the multiple times discussed `ua status --verbose` we have:

Not so readable
```
Account: christian.ehrhardt@canonical.com
Subscription: christian.ehrhardt@canonical.com
Valid until: n/a
Technical support level: n/a
```

Nice and readable:
```
cc            entitled         n/a
cis-audit     none             n/a
esm           entitled         active
fips          entitled         n/a
fips-updates  none             n/a
livepatch     entitled         n/a

```

I'd suggest to columnize the upper output as well to something like
```
Account                   christian.ehrhardt@canonical.com
Subscription              christian.ehrhardt@canonical.com
Valid until               n/a
Technical support level   n/a
```

dpb1:
My first reaction: I like the way key/value data is set distinct from the tabular data and feel this suggestion muddies it.  I'd like mpt's read on it though.

matthewpaulthomas:
I can see that lining up the values would make it easier to answer questions like “Are the account and the subscription using exactly the same e-mail address”.

Key-value data could still be distinct from tabular data in two other ways:
- keys could still end with colons, while table values don’t
- keys could be right-aligned, which would also avoid the problem where short keys are far away from their values.
```
                Account: christian.ehrhardt@canonical.com
           Subscription: christian.ehrhardt@canonical.com
            Valid until: n/a
Technical support level: n/a
```

PR #659, opened 2019-06-26
status: right align the upper status output

powersj:
Fixes: #482

### DIFF
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -104,9 +104,6 @@ MESSAGE_UNATTACHED = """\
 This machine is not attached to a UA subscription.
 See `ua attach` or https://ubuntu.com/advantage"""
 
-STATUS_HEADER_TMPL = """\
-Account: {account}
-Subscription: {subscription}"""
 STATUS_SERVICE_HEADER = '\nSERVICE'
 STATUS_TMPL = '{name: <14}{entitled: <26}{status}'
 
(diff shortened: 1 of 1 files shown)


uaclient/status.py
```diff
@@ -132,10 +129,14 @@ def format_tabular(status):
         return MESSAGE_UNATTACHED
     tech_support_level = status['techSupportLevel']
 
-    content = [STATUS_HEADER_TMPL.format(
-        account=status['account'], subscription=status['subscription'])]
-    if status['origin'] != 'free':  # All free accounts will have 'free' origin
-        content.append('Valid until: ' + str(status['expires']))
+    content = []
+    if status['origin'] == 'free':
+        content.append('     Account: ' + status['account'])
```
OddBloke: I think this would be preferable using the spacing that Python format templates provide (so that we can easily change the spacing in future if necessary). See `STATUS_TMPL` (in context of this diff) for a pointer towards how to do that.

powersj: I pushed a change and I am looking for thoughts on how to make that cleaner. It looks messy and you can tell I haven't dealt with fstrings very often.

OddBloke:
I have an improvement for this locally, I'll write tests tomorrow morning and submit it.

OddBloke:
(In fact, I've pushed it up, in case Josh wants to land it before demo'ing; I will still write tests tomorrow morning.)

powersj:
Thanks @OddBloke for the branch. Much cleaner :)

PR #669, opened 2019-06-26
d/rules: only hardcode the libapt-pkg<abi> dependency for known cases

### LINKED ISSUE #668: eoan now has libapt-pkg5.90
in `d/rules` we are using `LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.8.1)"`. There is no `libapt-pkg5.0` in eoan anymore, and as a result the eoan package is uninstallable because it wants both abi 5.0 and 5.90
```
$ dpkg --info ubuntu-advantage-tools_19.5~552~git.c3f7562~ubuntu19.10.1_amd64.deb |grep libapt-pkg
 Depends: python3-pymacaroons, python3-yaml, python3:any, libapt-pkg5.90 (>= 0.8.0), libc6 (>= 2.14), libgcc1 (>= 1:3.0), libstdc++6 (>= 5.2), python3-pkg-resources, libapt-pkg5.0 (>= 1.8.1)
```
We started adding this explicit dependency in `d/rules` because of the "never:" pinning handling. For eoan, I think we can skip that, since it will be released with abi 5.90 (or later) and that should have the right pinning handling.

panlinux:
All other cases will rely on dh_shlibdeps as usual.

Fixes #668

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -13,7 +13,7 @@ else ifeq (${VERSION_ID},"18.04")
 LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.6.9)"
 else ifeq (${VERSION_ID},"18.10")
 LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.7.4)"
-else
+else ifeq (${VERSION_ID},"19.04")
 LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.8.1)"
 endif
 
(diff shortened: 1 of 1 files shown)


debian/rules
```diff
@@ -37,7 +37,7 @@ endif
 endif
 
 override_dh_gencontrol:
-	echo extra:Depends=$(LIBAPT_PKG_DEP) >> debian/ubuntu-advantage-tools.substvars
+	[ -z '$(LIBAPT_PKG_DEP)' ] || echo extra:Depends=$(LIBAPT_PKG_DEP) >> debian/ubuntu-advantage-tools.substvars
```
panlinux: Could also use another `ifeq` here instead of a shell variable check. I had to use single quotes here because the makefile variable has symbols like `(` and `>` which the shell doesn't like.

panlinux:
What was wrong with cosmic? It's 18.10, and should still be handled.

OddBloke:
On Wed, Jun 26, 2019 at 01:55:37PM -0700, Andreas Hasenack wrote:
> What was wrong with cosmic? It's 18.10, and should still be handled.

I don't have a schroot set up locally (and `mk-sbuild` failed
confusingly).

Issue #647, opened 2019-06-19
`ua status` should not emit colour codes when not outputting to a console


OddBloke:
```
$ ua status > test
$ cat -v test
Account: UA Client Test
Subscription: UA Client Test Contract
Valid until: n/a
Technical support level: ^[[92madvanced^[[0m

SERVICE
cc-eal        ^[[92mentitled^[[0m         ^[[91minactive^[[0m
cis-audit     ^[[37mnone^[[0m             ^[[37mn/a^[[0m
esm           ^[[92mentitled^[[0m         ^[[37mn/a^[[0m
fips          ^[[92mentitled^[[0m         ^[[37mpending^[[0m
fips-updates  ^[[92mentitled^[[0m         ^[[91minactive^[[0m
livepatch     ^[[92mentitled^[[0m         ^[[91minactive^[[0m

Enable entitlements with `ua enable <service>`
```

powersj:
Looks like we can do something like:

```
from sys import stdout

COLOR_STRING if is_tty print else STRING
```

PR #671, opened 2019-06-27
status: do not emit color when not using a tty

powersj:
Fixes: #647


OddBloke:
@powersj  Travis has passed: https://travis-ci.org/CanonicalLtd/ubuntu-advantage-client/builds/551319129 so you can use your super-powers to land this.

PR #674, opened 2019-06-27
status: add header tests, and function type annotations

OddBloke:


uaclient/tests/test_status.py
```diff
@@ -5,6 +5,19 @@
 from uaclient.status import format_tabular, TxtColor
 
 
+@pytest.fixture
+def status_dict():
```
blackboxsw: naming nit:  status_dict_attached?

OddBloke: Good shout, thanks! Fixed now.

Issue #681, opened 2019-07-02
token link is wrong


powersj:
`ua attach -h` shows:

```
positional arguments:
  token                Optional token obtained from Ubuntu Advantage
                       dashboard: https://contracts.canonical.com
```

That URL is wrong it should be: `https://auth.contracts.canonical.com/`

We should see if that link is anywhere else in the code as well

PR #682, opened 2019-07-02
docs: help documentation link to token URL points to auth.contracts.c.c

blackboxsw:
Fixes: #681

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -42,7 +42,7 @@ Subscription: {subscription}
 Valid until: {contract_expiry}
 Technical support level: {tech_support_level}
 """
-UA_DASHBOARD_URL = 'https://contracts.canonical.com'
+UA_AUTH_TOKEN_URL = 'https://auth.contracts.canonical.com'
 
 DEFAULT_LOG_FORMAT = (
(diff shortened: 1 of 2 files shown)


uaclient/cli.py
```diff
@@ -43,6 +43,7 @@
 Technical support level: {tech_support_level}
 """
 UA_DASHBOARD_URL = 'https://contracts.canonical.com'
```
OddBloke: I think we can drop this constant now, right?

uaclient/cli.py
```diff
@@ -77,7 +78,7 @@ def attach_parser(parser=None):
     parser.add_argument(
         'token', nargs='?',
         help=('Optional token obtained from Ubuntu Advantage dashboard: %s' %
```
OddBloke: Should we still be calling it the "dashboard" in this message?

OddBloke:
Couple of inline questions.

uaclient/cli.py
blackboxsw: dropped. nice catch. sloppy of me.

uaclient/cli.py
blackboxsw: changed, also fixed the help docs on --otp

Issue #428, opened 2019-04-22
auto-enablement should exit non-zero on autoenable failure


blackboxsw:
on trusty, esm should be autoenabled. When testing against a local ua-contract server that return invalid esm credentials we see an error printed during auto-enable that should result in minimally a non-zero  exit code 

```
root@t1:~# ua attach
Email: chad.smith@canonical.com
Password: 
Second-factor auth: 066134
This machine is now attached to 'chad.smith@canonical.com'.

Enabling default entitlements...
[ERROR]: Invalid APT credentials provided for https://esm.ubuntu.com
Cannot install Livepatch on a container

Account: chad.smith@canonical.com
Subscription: chad.smith@canonical.com
Valid until: n/a
Technical support level: n/a

cc            none             inactive
cis-audit     none             inactive
esm           entitled         inactive
fips          none             n/a
fips-updates  none             n/a
livepatch     entitled         n/a

Enable entitlements with `ua enable <service>`
root@t1:~# echo $?
0
```

OddBloke:
(As background, once #411 is fixed, we should no longer see the "Cannot install Livepatch on a container" in this output.)

PR #688, opened 2019-07-11
attach: exit non-zero on failure to auto-enable services during attach

blackboxsw:
Also adjust request_updated_contract raise a UserFacingError if any
auto-enabled service fails to get enabled during the attach.

Fixes: #428

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -204,8 +204,11 @@ def action_enable(args, cfg):
     @return: 0 on success, 1 otherwise
     """
     print(ua_status.MESSAGE_REFRESH_ENABLE)
-    if not contract.request_updated_contract(cfg):
-        logging.debug(ua_status.MESSAGE_REFRESH_FAILURE)
+    try:
+        contract.request_updated_contract(cfg)
+    except (util.UrlError, exceptions.UserFacingError):
+        # Inability to refresh is not a critical issue during enable
+        logging.debug(ua_status.MESSAGE_REFRESH_FAILURE, exc_info=True)
     return 0 if _perform_enable(args.name, cfg) else 1
 
 
@@ -235,10 +238,18 @@ def action_attach(args, cfg):
     if not contract_token:
         print('No valid contract token available')
         return 1
-    if not contract.request_updated_contract(
-            cfg, contract_token, allow_enable=args.auto_enable):
+    try:
+        contract.request_updated_contract(
+            cfg, contract_token, allow_enable=args.auto_enable)
+    except util.UrlError as exc:
+        with util.disable_log_to_console():
+            logging.exception(exc.msg)
         print(
             ua_status.MESSAGE_ATTACH_FAILURE_TMPL.format(url=cfg.contract_url))
+        return 1
+    except exceptions.UserFacingError as exc:
+        logging.warning(exc.msg)
+        action_status(args=None, cfg=cfg)
         return 1
     contract_name = (
         cfg.machine_token['machineTokenInfo']['contractInfo']['name'])
@@ -316,11 +327,14 @@ def print_version(_args=None, _cfg=None):
 
 @assert_attached_root
 def action_refresh(args, cfg):
-    if contract.request_updated_contract(cfg):
-        print(ua_status.MESSAGE_REFRESH_SUCCESS)
-        logging.debug(ua_status.MESSAGE_REFRESH_SUCCESS)
-        return 0
-    raise exceptions.UserFacingError(ua_status.MESSAGE_REFRESH_FAILURE)
+    try:
+        contract.request_updated_contract(cfg)
+    except util.UrlError as exc:
+        with util.disable_log_to_console():
+            logging.exception(exc.msg)
+        raise exceptions.UserFacingError(ua_status.MESSAGE_REFRESH_FAILURE)
+    print(ua_status.MESSAGE_REFRESH_SUCCESS)
+    return 0
 
 
 def setup_logging(console_level, log_level, log_file=None):
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -1,5 +1,6 @@
 import logging
 
+from uaclient import exceptions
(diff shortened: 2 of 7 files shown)


ubuntu-advantage.1
```diff
@@ -170,6 +170,16 @@ downtime, keeping your Ubuntu LTS systems secure and compliant.
...
 
+.SH EXIT STATUS
+The Ubuntu Advantage command line tool will exit with the following codes:
+
+.BR 0: " Success"
+.br
+.BR 1: " Error occurred during the operation"
+Error messages are emitted to console and /var/log/ubuntu-advantage.log
+.br
+.BR 2: " Non-critical warning occurred during the operation"
```
OddBloke: Warnings are emitted, they don't occur. Perhaps s/warning/error/, and s/Error/Critical error/ in the `1` description?

uaclient/cli.py
```diff
@@ -349,11 +361,14 @@ def print_version(_args=None, _cfg=None):
...
 @assert_attached_root
 def action_refresh(args, cfg):
-    if contract.request_updated_contract(cfg):
-        print(ua_status.MESSAGE_REFRESH_SUCCESS)
-        logging.debug(ua_status.MESSAGE_REFRESH_SUCCESS)
-        return 0
-    raise exceptions.UserFacingError(ua_status.MESSAGE_REFRESH_FAILURE)
+    try:
+        contract.request_updated_contract(cfg)
+    except (util.UrlError, exceptions.UserFacingError) as exc:
```
OddBloke: This feels like we're discarding some potentially relevant user-facing output by not using the `UserFacingError` we catch. Is that something we could improve?

uaclient/cli.py
```diff
@@ -217,8 +217,12 @@ def action_enable(args, cfg):
     @return: 0 on success, 1 otherwise
     """
     print(ua_status.MESSAGE_REFRESH_ENABLE)
-    if not contract.request_updated_contract(cfg):
+    try:
+        contract.request_updated_contract(cfg)
+    except (util.UrlError, exceptions.UserFacingError) as exc:
+        # Inability to refresh is not a critical issue during enable
         logging.debug(ua_status.MESSAGE_REFRESH_FAILURE)
+        logging.debug(str(exc))
```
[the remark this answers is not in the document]
blackboxsw: sure. concatenated two sentences together. in a single log line.

uaclient/contract.py
```diff
@@ -238,32 +249,28 @@ def request_updated_contract(cfg, contract_token=None, allow_enable=False):
...
-    except util.UrlError as e:
-        logging.error(
-            'Could not obtain updated contract information. %s', str(e))
-        return False
-    return True
+        except exceptions.UserFacingError as e:
+            errors.append(e)
+    if errors:
+        raise exceptions.UserFacingError(
+            'Could not auto-enable recommended contract services. %s' %
```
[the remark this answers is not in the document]
blackboxsw: changed to "Failure processing Ubuntu Advantage contract changes. %s"

ubuntu-advantage.1
blackboxsw: good thought. done

uaclient/cli.py
```diff
@@ -204,8 +204,11 @@ def action_enable(args, cfg):
     @return: 0 on success, 1 otherwise
     """
     print(ua_status.MESSAGE_REFRESH_ENABLE)
-    if not contract.request_updated_contract(cfg):
-        logging.debug(ua_status.MESSAGE_REFRESH_FAILURE)
+    try:
+        contract.request_updated_contract(cfg)
+    except (util.UrlError, exceptions.UserFacingError) as exc:
+        # Inability to refresh is not a critical issue during enable
+        logging.debug(ua_status.MESSAGE_REFRESH_FAILURE + ". " + str(exc))
```
OddBloke: As this is debug level (and therefore won't go to the console by default), we can just pass `exc_info=True` to get the full traceback in the log.

uaclient/cli.py
OddBloke: This looks unaddressed, I think?

uaclient/contract.py
```diff
@@ -183,32 +194,28 @@ def request_updated_contract(cfg, contract_token=None, allow_enable=False):
...
-                new_access = entitlement
             process_entitlement_delta(
                 orig_entitlements.get(name, {}), new_access,
                 allow_enable=allow_enable)
-    except util.UrlError as e:
-        logging.error(
-            'Could not obtain updated contract information. %s', str(e))
-        return False
-    return True
+        except exceptions.UserFacingError as e:
```
OddBloke: Should we be catching all exceptions here? If we don't then they'll bubble all the way out to the user and cause the process to exit 1, I think.

OddBloke: (To be clear, I mean in an additional exception handler; we would probably want to log the exception and append a generic message to errors.)

OddBloke:
This is pretty close, I think, just a few remaining little bits to clean up.

uaclient/contract.py
```diff
@@ -183,32 +194,33 @@ def request_updated_contract(cfg, contract_token=None, allow_enable=False):
...
+            user_errors.append(e)
+        except Exception as e:
+            with util.disable_log_to_console():
+                logging.exception(str(e))
+            raise exceptions.UserFacingError(
+                'Unexpected error handling Ubuntu Advantage contract changes')
+    if user_errors:
+        raise exceptions.UserFacingError(
+            'Failure processing Ubuntu Advantage contract changes. %s' %
```
OddBloke: I think this will format a bit weirdly. With `user_errors = ['a', 'b']`, I think it would output:

```
Failure processing Ubuntu Advantage contract changes. a
b
```

which looks wonky to me. Perhaps:

```
Failure processing Ubuntu Advantage contract changes:
- a
- b
```

would be better?

blackboxsw: done.

Issue #622, opened 2019-06-14
refresh the status cache on reboot?


panlinux:
Should we refresh the `ua status` cache on a reboot? I noticed a discrepancy between the cache and the real status when I rebooted a trusty machine. It had livepatch enabled (it was on a 4.4 kernel) but I rebooted it into a 3.3 kernel, which invalidates livepatch. `ua status` kept saying livepatch was enabled, but `sudo ua status` disagreed, and also updated the cache, so after that both started saying livepatch was disabled.

OddBloke:
This will also occur after you enable FIPS{ Updates,} and reboot in to linux-fips; `status` will still be using the cache from before reboot (and therefore report "pending" for the entitlement that is really now active).

This seems reasonable to me.

powersj:
If we did not have the cache, and required everyone to run status as root, would this be an issue?

OddBloke:
Nope, root always gets fresh data, so this would not be an issue.

powersj:
If I recall correctly, we cache data for the `ua status` command, primarily due to ESM. Did the recent changes to how we pin for ESM make it so we did not need to cache anymore?

OddBloke:
Assigning to myself to double-check if we do need the status cache.

OddBloke:
So I'm not sure about the pinning, but when we switched to using the status cache we removed all other public data, so we _do_ need the cache for reporting status.

(I think you're thinking of the older iteration of the cache, where we just cached the enabled status of each entitlement. We moved to a full cache of the status JSON after that.)

powersj:
-1 on increasing what the client does. Only issue is very specific odd ball cases. I am more inclined to make someone run `sudo ua status` over adding startup scripts to the client.

PR #689, opened 2019-07-11
WIP: status: add upstart and systemd scripts to update status per boot

blackboxsw:
WIP: needs discussion on whether we want this every boot
====

Two use cases need discussion:
  1.  Trusty switch to unsupported livepatch kernel
     A. (trusty) ua enable livepatch (on hwe kernel) 
     B. reboot to non hwe kernel
     C. (non-root) ua status will still report livepatch active
     D. (root) ua status will properly update livepatch inactive
     
  2. Xenial, Bionic FIPS*
     A. sudo ua enable fips
     B. (non-root) ua status fips pending   # requires reboot to fips kernel
     C. sudo reboot # to fips kernel
     D. (non-root) ua status fips pending    # status only updated by root
     E: (root) ua status   fips active
     
1 may be a specific corner case we don't really care about. If we are concerned about hitting ua-contracts service every boot unnecessarily, we could limit the scope of this update check to only do this on systems where services are in 'pending' status on an LTS series. Because pending is generally an interim state that we expect to leave post a reboot, this would certainly limit us to systems where FIPS was activated, and have a potential to transition 

Fixes: #622

To test:
```
# launch and connect to vm on your preferred series
sudo make deps
dpkg-buildpackage -us -uc
dpkg -i ../ubuntu-advantage-tools*deb
sleep 60
sudo reboot 
# check timestamp on status,json to make sure it was written again
ls -ltr /var/lib/ubuntu-advantage/status.json  
```

### DIFF
--- a/systemd/ua-status.service
+++ b/systemd/ua-status.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Ubuntu Advantage status refresh
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ua status
+TimeoutSec=5
+
+[Install]
+WantedBy=multi-user.target
+After=network-online.target
+Want=network-online.target
--- a/debian/control
+++ b/debian/control
@@ -14,7 +14,8 @@ Build-Depends: bash-completion,
                python3-mock,
                python3-pytest,
                python3-setuptools,
-               python3-yaml
+               python3-yaml,
+               dh-systemd,
 Standards-Version: 4.3.0
 Homepage: https://buy.ubuntu.com
 Vcs-Git: https://github.com/CanonicalLtd/ubuntu-advantage-script.git
--- a/debian/rules
+++ b/debian/rules
@@ -18,7 +18,7 @@ LIBAPT_PKG_DEP="libapt-pkg5.0 (>= 1.8.1)"
 endif
 
 %:
-	dh $@ --with python3,bash-completion --buildsystem=pybuild
+	dh $@ --with python3,bash-completion,systemd --buildsystem=pybuild
 
 override_dh_auto_build:
 	dh_auto_build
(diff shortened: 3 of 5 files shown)


powersj:
hmm... this means that every single systems running UA would be checking in with the backend on a reboot right? I also don't like the idea of us making our dumb client any less dumb as this feels like a gateway drug to doing other service like changes.

powersj:
Also Trusty (upstart) will only ever support ESM and livepatch, which should not need this. So if we do end up deciding to do this I would not want the upstart script in tree.

systemd/ua-status.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Ubuntu Advantage status refresh
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ua status
```
raharper: Should we follow the apt-daily.{service,timer} units?

systemd/ua-status.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Ubuntu Advantage status refresh
```
raharper: We should add:

After=network-online.target
Want=network-online.target

as ua status is a network hit.

systemd/ua-status.service
blackboxsw: `We probably want to sync correct service status information before any non-root user logs into the system to avoid exposure for invalid satus information (because only root user *ua status* runs would attempt to  actually update status cache). As such, we probably don't want to delay this operation by a variable amount of time because we might get spurious bugs filed if that timer splay was large enough on one specific reboot.

systemd/ua-status.service
blackboxsw: thanks done.

Issue #624, opened 2019-06-14
uaclient: Allow for different GPG keys per series in ESM archive


OddBloke:
In https://github.com/CanonicalLtd/ua-contracts/issues/319, IS requested the ability to use different GPG keys for different series in ESM. As the client currently ignores the server's `aptKey` directives and installs a single keyring for each entitlement, this is something we'll need to respond to also.

OddBloke:
The easiest fix for this would be to include the keys for every series in the one keyring file we install. This, however, somewhat defeats the point of having separate keys, so is less than ideal.

OddBloke:
This is a proposal for a more robust solution, which doesn't rely on the `apt-key` command (or even keyservers), but would honour the `aptKey` directive:

Instead of shipping one keyring per-entitlement, ubuntu-advantage-client will instead ship each key that we know to be trusted separately, identified by its key ID. When we are enabling an entitlement, we will use the `aptKey` directive to select which of these shipped keys we use and import it in to the keyring at the path we would currently copy our keyring to.

(This separate key shipping could be a single keyring with every key, or we could ship the keys ASCII-armored for easier of import on the target system, or we could do something else; the important aspect is that we can index in to whatever storage we're using by key ID.)

OddBloke:
(Having said that it doesn't matter how we ship the keys, separate files on disk would mean that we would be tightly coupled to the exact key ID format that the contract server provides. If they moved between short IDs, long IDs or fingerprints, things would break. So that's probably an argument for using `gpg` to select the keys from a single keyring.)

OddBloke:
(We should ensure that whatever solution we come up with for this will support key rotation.)

powersj:
Based on another thread and discussions around what the client should do it seems:
1) The client will be the mechanism for shipping keys for the UA client
2) If keys need to be updated, it involves an update to the UA client it self
3) For rotation, the official keyring can be used to revoke bad keys and what would be required by the client is shipping the new key

As such I believe this means we want different keys per release. Is there a clear action for the client now?

OddBloke:
I believe there are two parts to this: (a) handle per-series keys when we're copying keys in to `/etc/apt/...`; currently we only handle one key(ring) per entitlement, and (b) work out how we want to use the indication from the server to message to the user if we don't have the expected key available.

[...]

PR #690, opened 2019-07-13
apt: support per-series aptKey

blackboxsw:
This branch allows ua-contracts to drive aptKey changes per-series, allowing ua client to install
the designated aptKey during the enable operation.

Adjust tools/refresh-keyrings.sh to store all keys in a since ubuntu-advantage-keyring.gpg and 
drop all individual gpg files from keyrings directory.

Add a new gpg module with helper function gpg_export_key_from_keyfile to publish a single
key from the shared keyring to a /etc/apt/trusty.gpg.d/.

Wire to gpg_export_key_from_keyfile to add_apt_auth_repo so keys are written at enable time.
Exit in error if no aptKey directive is present for a given service in machine-token.
Adapt debian/postinst to write esm source/preference files in postinst on trusty/xenial/bionic and cleanup any stale files from unmatched series.

Fixes: #624
Fixes: #692


tools/refresh-keyrings.sh
```diff
@@ -25,24 +54,9 @@ generate_keyring() {
...
-KEYS_FIPS="C1997C40EDE22758"
-KEYS_FIPS_UPDATES="C1997C40EDE22758"
-KEYS_CC="3A186E733F491C46"
+UA_KEYRING="ubuntu-advantage-keyring.gpg"
 
-generate_keyring "$TARGET_DIR/$KEYRING_ESM" "$KEYS_ESM"
-generate_keyring "$TARGET_DIR/$KEYRING_FIPS" "$KEYS_FIPS"
-generate_keyring "$TARGET_DIR/$KEYRING_FIPS_UPDATES" "$KEYS_FIPS_UPDATES"
-generate_keyring "$TARGET_DIR/$KEYRING_CC" "$KEYS_CC"
+APT_KEYS=`cat $MACHINE_TOKEN | json_pp | awk -F '"' '/aptKey.*/{printf $4 " "}'`
```
OddBloke: I'm not convinced we should be making this change; we want an easily reviewable change process for the keys we ship, and having them hard-coded in this file makes that much easier, IMO.

Also: a token doesn't necessarily have access to all the entitlements, and we may want to ship more keys than just the current `aptKey` values, to support key transitions. So I'm not sure it's possible to fully automate this using the contract server, which means we need the manual key handling anyway.

(A potential future improvement would be to check that the set of keys in the keyring after generation is a super-set of the set of keys specified by the `aptKey` directives, but that's very much a nice-to-have so if you think that's worthwhile let's file an issue for it and prioritise it separately.)

uaclient/entitlements/repo.py
```diff
@@ -173,10 +174,14 @@ def setup_apt_config(self) -> None:
                 'No specific resourceToken present. Using machine token'
                 ' as %s credentials', self.title)
             token = self.cfg.machine_token['machineToken']
-        if directives.get('aptKey'):
-            logging.debug(
-                "Ignoring aptKey directive '%s'", directives.get('aptKey'))
-        keyring_file = os.path.join(apt.KEYRINGS_DIR, self.repo_key_file)
+        aptKey = directives.get('aptKey')
+        if not aptKey:
+            logging.warning(
```
OddBloke: If we aren't going to install a key, would it be better to just error out here (rather than continue on to enable a repo which won't be useable)?

uaclient/entitlements/repo.py
```diff
@@ -173,10 +174,14 @@ def setup_apt_config(self) -> None:
...
+                "No aptKey directive provided for %s", self.name)
+        else:
+            keyring_file = os.path.join(apt.APT_KEYS_DIR, self.repo_key_file)
+            gpg.export_gpg_key_from_keyring(
```
OddBloke: Key installation was previously performed by `add_auth_apt_repo`; what's the thinking behind moving it up a layer here?

uaclient/gpg.py
```diff
@@ -0,0 +1,33 @@
...
+    logging.debug('Exporting gpg key %s from %s', key_id, source_keyring_file)
+    try:
+        out, _err = util.subp(export_cmd, decode=False)
+    except util.ProcessExecutionError as exc:
+        with util.disable_log_to_console():
+            logging.error(str(exc))
+        raise exceptions.UserFacingError(
+            "Unable to export gpg keys from keyring %s" % source_keyring_file)
+    if not out:
+        raise KeyError("gpg key '%s' not found in %s" % (
```
OddBloke: We should not be using `KeyError` for anything other than dictionary keys.

uaclient/gpg.py
```diff
@@ -0,0 +1,33 @@
...
+            key_id, source_keyring_file))
+    util.write_file(destination_keyfile, out)
```
OddBloke: Why pull the output through our process rather than just passing `--output` to `gpg`?

debian/postinst
```diff
@@ -2,15 +2,23 @@
 
 set -e
 
+ESM_KEY_ID="56F7650A24C9E9ECF87C4D8D4067E40313CB4B13"
```
OddBloke: Does this need to be different per-series?

uaclient/util.py
```diff
@@ -304,7 +311,8 @@ def which(program: str) -> 'Optional[str]':
     return None
 
 
-def write_file(filename: str, content: str, mode: int = 0o644) -> None:
+def write_file(filename: str, content: 'Union[str, bytes]',
```
OddBloke: Instead of complicating the signature of `write_file`, could we introduce a separate bytes-writing function instead (something like `write_file_bytes`)?  (`write_file` could be refactored to be a wrapper around the bytes-writer.)

(This may be moot if we get `gpg` to perform writing for us, of course, because we could also drop these changes.)

uaclient/util.py
```diff
@@ -211,7 +211,8 @@ def readurl(url: str, data: 'Optional[bytes]' = None,
 def _subp(args: 'Sequence[str]',
           rcs: 'Optional[List[int]]' = None,
           capture: bool = False,
-          timeout: 'Optional[float]' = None) -> 'Tuple[str, str]':
+          timeout: 'Optional[float]' = None,
+          decode: 'Optional[bool]' = True) -> 'Tuple[str, str]':
```
OddBloke: This type signature is inaccurate, as we can now return str or bytes as the tuple members. Similar to below, I don't love the return type of a function being dynamic based on one of its arguments, because it makes the codebase harder to reason about, but it's more of a PITA to address here than there because of the multiple layers we have going on.

OddBloke:
OK, first review pass complete. (I haven't yet reviewed the tests because I think they may change based on my review comments.)

tools/refresh-keyrings.sh
blackboxsw: +1 I think you sold me on the " values, to support key transitions" part. I'll add hardcoded values then for each series here. And we can validate with this script that the keyring generated contains the set of static keys we add here

debian/postinst
blackboxsw: Yes it does, I had flied a separate issue for this to avoid bloating this branch more, but if you agree I'll pull that work into this branch as this branch doesn't work without a followup for #692

OddBloke: I think it probably is worth doing here; how much bloat would you expect it to add? (Unless I'm missing something, this is just a case statement, right?)

uaclient/entitlements/repo.py
blackboxsw: I moved it up for two reasons:
 - it needs self.repo_key_file from the Entitlement to generate the target keyring_file path and aptKey from self.cfg directives. Instead of passing two params into add_auth_apt_repo to do one thing, I thought we could make the call up here and simplify add_auth_apt_repo call signature.

uaclient/entitlements/repo.py
blackboxsw: raising a UserFacingError here

uaclient/entitlements/repo.py
OddBloke: IMO, it would be preferable to have the apt function be "atomic" even if it requires another parameter.

(That only looks like one reason though, so perhaps there's more to it?)

uaclient/entitlements/repo.py
OddBloke: +1

uaclient/entitlements/repo.py
blackboxsw: When it was a simple file copy, it made sense to do that inside add_auth_apt_repo. Now that we have to do a gpg key extraction and then a copy to a named keyring_file destination that makes ad_auth_apt_repo grow another parameter in the function signature, one for the key_id and one for the destination keyring file to write. I thought it'd be simpler to reduce the number of params and do the simple work here, but if you prefer the other way, I can extend add_auth_apt_repo.

uaclient/gpg.py
blackboxsw: roger, using UserFacingError.

uaclient/entitlements/repo.py
blackboxsw: Will fix that to accept the additional param.

uaclient/gpg.py
blackboxsw: good point, hadn't thought of that. Using --output instead.

uaclient/util.py
blackboxsw: Dropped these changes, from util, turns out I don't need them while using gpg --output thanks for that.

uaclient/util.py
blackboxsw: dropped them. thanks again for that sentiment.

debian/postinst
blackboxsw: Just missing that we'll have to parameterize a few of the functions in postinst to handle upgrade path trusty->xenial etc. But It's not too much. I'll tease it out today.

blackboxsw:
Addressed all review comments I believe. Thanks @OddBloke

debian/postinst
```diff
@@ -2,38 +2,85 @@
...
 
-    if [ ! -e "$ESM_APT_SOURCE_FILE" ]; then
-        cat > $ESM_APT_SOURCE_FILE <<EOF
+write_esm_apt_source() {
+    ENABLE_SERIES=$1
+    for SERIES in trusty xenial bionic; do
+        APT_SOURCE=`echo $ESM_APT_SOURCE_FILE | sed -e "s/<SERIES>/$SERIES/"`
+	if [ "$SERIES" != "$ENABLE_SERIES" ]; then
+            # Delete inactive series apt source files
+	    rm -f $APT_SOURCE;   # Delete inactive series files
```
blackboxsw: This rm should fix #693

debian/postinst
OddBloke: OK, I was assuming we were only going to make the changes to install these per-series (as this change would otherwise regress that specific functionality); I think the additional changes fixing other, pre-existing issues would be best submitted as a separate PR.

uaclient/gpg.py
```diff
@@ -0,0 +1,33 @@
...
+    logging.debug('Exporting gpg key %s from %s', key_id, source_keyring_file)
+    try:
+        util.subp(export_cmd)
+    except util.ProcessExecutionError as exc:
+        with util.disable_log_to_console():
+            logging.error(str(exc))
+        raise exceptions.UserFacingError(
+            "Unable to export gpg keys from keyring %s" % source_keyring_file)
+    if not os.path.exists(destination_keyfile):
+        raise exceptions.UserFacingError("gpg key '%s' not found in %s" % (
```
OddBloke: `gpg` on trusty and eoan appears to output "`gpg: WARNING: nothing exported`" to stderr when we're in this error case; is it worth checking if that was output and emitting a more general message otherwise?

(I'm concerned there may be a class of "`gpg: WARNING: failed to frobnicate the doodad`" non-zero-exit error messages that we don't know about; if we treat them all the same, it could be quite frustrating for the user.)

What do you think?

debian/postinst
blackboxsw: Ok just backed out xenial/bionic support for esm from postinst. Filed a separate PR to fix #692. That PR is #699

uaclient/gpg.py
```diff
@@ -0,0 +1,33 @@
...
+    :param destination_keyfile: The filename created with the single exported
+        key.
+
+    :raise UserFacingError: Any GPG errors or if specific key does not exist in
+        the source_keyring_file.
+    """
+    export_cmd = ['gpg', '--output', destination_keyfile, '--yes',
+                  '--no-auto-check-trustdb', '--no-default-keyring',
+                  '--keyring', source_keyring_file, '--export', key_id]
+    logging.debug('Exporting gpg key %s from %s', key_id, source_keyring_file)
```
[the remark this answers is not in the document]
blackboxsw: done

uaclient/gpg.py
blackboxsw: Added a check for 'nothing exported' in stderr. Will raise anything else if we didn't create the gpg keyring file properly

tools/refresh-keyrings.sh
```diff
@@ -9,6 +9,22 @@
 #
 # N.B. This will rename any existing keyrings with the suffix .old.
 
+if [ $# != 1 -o ! -d $1 ]; then
+ echo "Usage: $0 <key_directory>"
+ exit 1
+fi
+if [ ! $(id -u) = 0 ]; then
+   echo "This script must be run as root"
```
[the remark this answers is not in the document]
blackboxsw: dropped thanks

tools/refresh-keyrings.sh
```diff
@@ -25,24 +41,19 @@ generate_keyring() {
...
-KEYS_FIPS="C1997C40EDE22758"
-KEYS_FIPS_UPDATES="C1997C40EDE22758"
-KEYS_CC="3A186E733F491C46"
+UA_KEYRING="ubuntu-advantage-keyring.gpg"
 
-generate_keyring "$TARGET_DIR/$KEYRING_ESM" "$KEYS_ESM"
-generate_keyring "$TARGET_DIR/$KEYRING_FIPS" "$KEYS_FIPS"
-generate_keyring "$TARGET_DIR/$KEYRING_FIPS_UPDATES" "$KEYS_FIPS_UPDATES"
-generate_keyring "$TARGET_DIR/$KEYRING_CC" "$KEYS_CC"
+generate_keyring $TARGET_DIR/$UA_KEYRING" $EAL_KEY_ID_XENIAL $ESM_KEY_ID_PRECISE $ESM_KEY_ID_TRUSTY $ESM_KEY_ID_XENIAL $ESM_KEY_ID_BIONIC $FIPS_KEY_ID_XENIAL
```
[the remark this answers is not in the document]
blackboxsw: fixed thank you

uaclient/entitlements/repo.py
```diff
@@ -173,10 +173,10 @@ def setup_apt_config(self) -> None:
...
                 ' as %s credentials', self.title)
             token = self.cfg.machine_token['machineToken']
-        if directives.get('aptKey'):
-            logging.debug(
-                "Ignoring aptKey directive '%s'", directives.get('aptKey'))
-        keyring_file = os.path.join(apt.KEYRINGS_DIR, self.repo_key_file)
+        aptKey = directives.get('aptKey')
+        if not aptKey:
+            raise exceptions.UserFacingError(
+                "No aptKey directive provided for %s", self.name)
```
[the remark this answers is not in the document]
blackboxsw: We now say 
                "Ubuntu advantage server provided no aptKey directive for %s."
                " Please file a bug at %s." % (self.name, defaults.BUG_URL))

defaults.BUG_URL = https://launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bugs

uaclient/tests/test_gpg.py
```diff
@@ -0,0 +1,70 @@
+from uaclient import exceptions, gpg, util
+from uaclient.testing import data
+
+import logging
```
OddBloke: stdlib imports [should](https://www.python.org/dev/peps/pep-0008/#imports) go above our own imports.

uaclient/tests/test_gpg.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class TestExportGPGKeyFromKeyring:
+
+    def test_key_error_on_missing_key(self, tmpdir):
+        """Raise UserFacingError when key_id is not in source_keyring_file."""
+        src_keyring = tmpdir.join('ubuntu-advantage-keyring')
+        src_keyring.write('')
+        destination_keyfile = tmpdir.join('destination_keyfile').strpath
+        key_id = 'Bueller... Bueller'
```
OddBloke: Should we make this a valid key ID, so we're sure we aren't hitting a GPG code path that tests for ID validity?

uaclient/tests/test_gpg.py
```diff
@@ -0,0 +1,70 @@
...
+        with pytest.raises(exceptions.UserFacingError) as excinfo:
+            gpg.export_gpg_key_from_keyring(
+                key_id=key_id, source_keyring_file=src_keyring.strpath,
+                destination_keyfile=destination_keyfile)
+
+        error_msg = "GPG key '%s' not found in %s" % (
+            key_id, src_keyring.strpath)
+        assert error_msg in str(excinfo.value)
+        assert False is os.path.exists(destination_keyfile)
```
OddBloke: Could this just be `assert not os...`?

uaclient/tests/test_gpg.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @pytest.mark.parametrize('caplog_text', [logging.ERROR], indirect=True)
```
OddBloke: Do we definitely need this line? I'd expect ERROR level messages to be included by default, and we do check the log level inline (by including the `ERROR    ` prefix in the string we look for in `caplog_text()`).

(It's really neither here nor there whether we include this for this test, but I would like to be sure that this isn't masking a bug in `caplog_text`.)

uaclient/tests/test_gpg.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_user_facing_error_on_invalid_keyring_packet(
+            self, caplog_text, tmpdir):
+        """Raise UserFacingError on invalid keyring packet. Log cmd error."""
+        source_keyring = tmpdir.join('keyring').strpath
+        destination_keyfile = tmpdir.join('destination_key').strpath
+        key_id = '3CB3DF682220A643B43065E9B30EDAA63D8F61D0'
+        # Create bogus keyring with invalid packet which causes gpg errors
+        with open(source_keyring, 'wb') as stream:
+            stream.write(b'\n')
```
OddBloke: Not worth changing, but for reference I _think_ we could do this with `tmpdir.join('keyring').write(b'\n', mode='wb')`.

uaclient/tests/test_gpg.py
OddBloke: (There are a few other instances of this if it can.)

uaclient/tests/test_gpg.py
```diff
@@ -0,0 +1,70 @@
...
+        with pytest.raises(exceptions.UserFacingError) as excinfo:
+            gpg.export_gpg_key_from_keyring(
+                key_id=key_id, source_keyring_file=source_keyring,
+                destination_keyfile=destination_keyfile)
+        msg = "Failed running command 'gpg --output {}".format(
+            destination_keyfile)
+        assert 'ERROR    {}'.format(msg) in caplog_text()
+        assert False is os.path.exists(destination_keyfile)
+        msg = 'Unable to export GPG keys from keyring %s' % source_keyring
```
OddBloke: https://write.wrestle.town/oddbloke/python-anti-pattern-single-element-s-formatting

uaclient/tests/test_gpg.py
```diff
@@ -0,0 +1,70 @@
...
+        assert msg == str(excinfo.value)
+
+    def test_export_single_key_from_keyring(self, tmpdir):
+        """Raise UserFacingError on invalid keyring packet. Log cmd error."""
+        source_keyring = tmpdir.join('keyring').strpath
+        destination_keyfile = tmpdir.join('destination_key').strpath
+        # Create bogus keyring with invalid packet which causes gpg errors
```
OddBloke: I think this comment is an incorrect copy/paste.

OddBloke: Oh, and the test docstring.

uaclient/entitlements/repo.py
```diff
@@ -173,10 +174,11 @@ def setup_apt_config(self) -> None:
...
             token = self.cfg.machine_token['machineToken']
-        if directives.get('aptKey'):
-            logging.debug(
-                "Ignoring aptKey directive '%s'", directives.get('aptKey'))
-        keyring_file = os.path.join(apt.KEYRINGS_DIR, self.repo_key_file)
+        aptKey = directives.get('aptKey')
+        if not aptKey:
+            raise exceptions.UserFacingError(
+                "Ubuntu advantage server provided no aptKey directive for %s."
+                " Please file a bug at %s." % (self.name, defaults.BUG_URL))
```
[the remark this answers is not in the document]
blackboxsw: good point, backing this out.

uaclient/tests/test_gpg.py
blackboxsw: I'll change it here for reference next time. I'd rather use the builtin features of pytest tmpdir anyway.

uaclient/tests/test_gpg.py
blackboxsw: thanks for the blog link, now I get to have some fun reading your blog.
fixed.

uaclient/tests/test_gpg.py
blackboxsw: dropped this

uaclient/gpg.py
```diff
@@ -0,0 +1,37 @@
...
+        with util.disable_log_to_console():
+            logging.error(str(exc))
+        raise exceptions.UserFacingError(
+            "Unable to export GPG keys from keyring %s" % source_keyring_file)
+    if 'nothing exported' in err:
+        raise exceptions.UserFacingError("GPG key '%s' not found in %s" % (
+            key_id, source_keyring_file))
+    if not os.path.exists(destination_keyfile):
+        raise exceptions.UserFacingError(
+            "Unexpected error exporting GPG key '%s' from %s: Error: %s" % (
```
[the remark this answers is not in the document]
blackboxsw: I'm wondering about other errors that we didn't account for, such as our fragile test  for substring 'nothing exported'  in stderr not working due to a gpg tooling update. At least this last condition will check if we really did export and create a keyring file. If not, we'll at least have a message raised to allow us to triage more easily.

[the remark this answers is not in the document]
OddBloke: To clarify, I think raising this error is good, it's _just_ having stderr included in it that's a sticking point.

(We should make sure it's logged so that we can ask users for it if they hit this.)

PR #702, opened 2019-08-07
test_serviceclient: fix up mocking

OddBloke:
I think this should fix the build failures we're currently seeing, though I wasn't able to reproduce them locally.


OddBloke:
OK, I reproduced by doing a full package build locally, and confirmed that this fix is good.

PR #710, opened 2019-08-13
Adds base58/checksum validation for contract tokens

jaycee:


OddBloke:
This now has conflicts, so is going to require some work in the event we do end up landing it. I'm closing it out for now, so we can aim to get our list of PRs down to zero.

PR #711, opened 2019-08-13
enable: ux rework messaging when enable on unattached

blackboxsw:


OddBloke:
Quick review from the bus on my phone. This looks good to me! One thought:
Should we refactor the duplicated assert_attached_root logic into a helper
that takes message as an argument?

uaclient/cli.py
```diff
@@ -197,12 +197,17 @@ def _perform_enable(entitlement_name: str, cfg: config.UAConfig, *,
...
-@assert_attached_root
 def action_enable(args, cfg):
     """Perform the enable action on a named entitlement.
 
     @return: 0 on success, 1 otherwise
     """
+    if os.getuid() != 0:
+        raise exceptions.NonRootUserError()
+    if not cfg.is_attached:
+        raise exceptions.UserFacingError(
```
OddBloke: As I said in my bus review, I'd like us to not duplicate this.  Could we refactor this to a helper that `assert_attach_root` could use too? We can just pass an exception instance in, instead of the message (I couldn't see the definition of `assert_attached_root` when I wrote that).

blackboxsw:
Dropping this in favor of https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/731  taking Dan's review comments  and allowing us to pass in a message template to assert_attached_root

PR #716, opened 2019-08-15
uaclient: replace % formatting with .format()

OddBloke:
Except in a couple of cases where we were mistakenly doing a %-format
for a log message, where the fix is to pass the %-format arguments as
parameters.

### DIFF
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -122,10 +122,13 @@ class RepoEntitlement(base.UAEntitlement):
         policy = apt.run_apt_command(
             ['apt-cache', 'policy'], status.MESSAGE_APT_POLICY_FAILED
         )
-        match = re.search(r'(?P<pin>(-)?\d+) %s[^-]' % repo_url, policy)
+        match = re.search(r'(?P<pin>(-)?\d+) {}[^-]'.format(repo_url), policy)
         if match and match.group('pin') != APT_DISABLED_PIN:
-            return ApplicationStatus.ENABLED, '%s is active' % self.title
-        return ApplicationStatus.DISABLED, '%s is not configured' % self.title
+            return ApplicationStatus.ENABLED, '{} is active'.format(self.title)
+        return (
+            ApplicationStatus.DISABLED,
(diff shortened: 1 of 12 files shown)


uaclient/entitlements/repo.py
```diff
@@ -192,24 +195,24 @@ def setup_apt_config(self) -> None:
         aptKey = directives.get('aptKey')
         if not aptKey:
             raise exceptions.UserFacingError(
-                "Ubuntu Advantage server provided no aptKey directive for %s."
-                % self.name
+                "Ubuntu Advantage server provided no aptKey directive for"
```
blackboxsw: What do you think about single quotes instead of double throughout? I'd like to avoid double-quotes as  as style choice unless we needed to embed single quotes within the string.

uaclient/tests/test_contract.py
```diff
@@ -21,9 +21,9 @@ class TestProcessEntitlementDeltas:
     def test_error_on_missing_entitlement_type(self):
         """Raise an error when neither dict contains entitlement type."""
         new_access = {'entitlement': {'something': 'non-empty'}}
-        error_msg = 'Could not determine contract delta service type %s %s' % (
-            {},
-            new_access,
+        error_msg = (
+            'Could not determine contract delta service type'
+            ' {} {}'.format({}, new_access)
```
blackboxsw: This just looks a bit funky. How about   
error_msg = 'Could not determine contract delta service type {{}} {}'.format(new_access)  ?

uaclient/entitlements/repo.py
OddBloke: If we want to standardise on string quoting (and bring it up in review), I think we should undo my (solo) decision to have black not normalise strings, and have it handle quotes too.

uaclient/tests/test_contract.py
OddBloke: Done.

uaclient/entitlements/repo.py
blackboxsw: nice thought. yes let's have a discussion about this tomorrow and see what opinions are on this policy.

PR #732, opened 2019-09-05
Trivial update to comment in cli.py

anthonydillon:
Update "livepatch" in the comment detailing available services in cli.py.


OddBloke:
@powersj Travis is having a normal one (i.e. not running jobs for no apparent reason), could you use your super-powers to merge this without a CI run?

PR #738, opened 2019-09-09
ux: update naming/descriptions of entitlements

blackboxsw:


uaclient/config.py
```diff
@@ -236,15 +240,43 @@ def _status(self) -> "Dict[str, Any]":
...
+        When unattached, get available resources from the contract service
+        to report detailed availability of different resources for this
+        machine.
+
+        Write the status-cache when called by root.
+        """
+        from uaclient.contract import get_available_resources
+        from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
+
+        if not self.is_attached:
```
blackboxsw: Dan, is the approach of handling unattached case up in status instead of down in _status distasteful?  I only pulled it up because non-root users running status have to do the same work a root  user does

OddBloke:
Is this WIP just because of those two open questions, or do you anticipate making further changes before this is ready for a full review? (Just trying to gauge how much time I should spend on the review now vs later. :+1:)

blackboxsw:
> n the review now vs later. )
Hi Dan, the only reason this is wip is pending you or Josh's answers to the open questions and me adding a couple more unit tests. for the status tabular format changes. I hadn't anticipated any more functional changes unless your answers suggest something.

OddBloke:
Cool, thanks for the clarification! I'm going to leave both of the open questions to @powersj and dig into the review now.

powersj:
> Josh: Do we want any colorized output of yes/no?

No, let's not colorize this output, since we are not doing any call to action.

> Josh/Dan: Did we decide to drop non-root support of ua status in the mid-cycle?

Does the un-athenticated status command require root?

OddBloke:
On Mon, Sep 09, 2019 at 09:24:13AM -0700, Joshua Powers wrote:
> > Josh/Dan: Did we decide to drop non-root support of ua status in the
> > mid-cycle?
> 
> Does the un-athenticated status command require root?

I'll let Chad confirm, but I _believe_ that root is only required if we
want to do the caching of platform and resources responses.

blackboxsw:
> No, let's not colorize this output, since we are not doing any call to action.
Great, no color.
> Does the un-athenticated status command require root?
No, as @OddBloke mentioned, non-root user will be able to run **ua status**, they will just be unable to write a platform-info cache file which would have saved them from having to talk to the contracts API server each time they run **ua status**. So non-root user won't have an optimization in ua status response time. (it takes a 1 second to run **ua status** now for uncached request)

blackboxsw:
> Hey Chad, I'm finding this a little difficult to review because it feels like three distinct changes are present:
> 
> 1. Update some strings so that our output looks correct
> 2. Introduce logic to display available resources in unattached status
> 3. Introduce caching to optimise the change in (2)
> 
> I could probably manage reviewing (1)+(2) in the same PR, but throwing (3) into the mix is making it difficult to grok the logic for (2) in a way that I'm comfortable reviewing. How much work do you think it would be to split these changes apart for review/landing?
> 
> As a side note, I'd also like them split apart because it's clear to me that we _need_ to land (1) and (2) for the UX work, whereas (3) is an optimisation that I'd like to be able to decide about separately.
> 
> (Ideally (1) and (2) would still be distinct commits, in case we want to revert one but not the other; that's just a nit though.)

Good thought @OddBloke I'll break this PR up now into 3 separate PRs.

uaclient/entitlements/tests/test_cc.py
```diff
@@ -61,15 +61,15 @@ class TestCommonCriteriaEntitlementUserFacingStatus:
...
+                "Common Criteria EAL2 Provisioning Packages is not available"
                 " for platform arm64.\n"
                 "Supported platforms are: x86_64, ppc64le, s390x",
             ),
             (
                 "s390x",
                 "trusty",
                 "14.04 LTS (Trusty Tahr)",
-                "Canonical Common Criteria EAL2 Provisioning is not available"
+                "Common Criteria EAL2 Provisioning Packages is not available"
```
OddBloke: This reads strangely to me; it _should_ be "Common Criteria EAL2 Provisioning Packages are not available".

uaclient/entitlements/tests/test_cc.py
```diff
@@ -228,11 +228,11 @@ def exists(path):
         expected_stdout += "\n".join(
             [
                 "Updating package lists",
-                "Installing Canonical Common Criteria EAL2 Provisioning"
+                "Installing Common Criteria EAL2 Provisioning Packages"
                 " packages",
```
OddBloke: And this reads extremely weirdly: "Installing Common Criteria EAL2 Provisioning Packages packages"

uaclient/entitlements/cc.py
```diff
@@ -6,10 +6,8 @@
 class CommonCriteriaEntitlement(repo.RepoEntitlement):
 
     name = "cc-eal"
-    title = "Canonical Common Criteria EAL2 Provisioning"
-    description = (
-        "Common Criteria for Information Technology Security Evaluation - EAL2"
-    )
+    title = "CC EAL2"
+    description = "Common Criteria EAL2 Provisioning Package"
```
OddBloke: Package**s**?

Issue #720, opened 2019-08-23
Unathenticated UA Status


powersj:
# Work Items

- [ ] If a user has not attached, show what features are available on this release when running `ua status`.
- [ ] Also use custom descriptions if they apply (e.g. livepatch needs HWE)

# Examples

```
$ sudo ua status
SERVICE       AVAILABLE  DESCRIPTION
cc-eal        no         Common Criteria EAL2 Provisioning Packages
cis-audit     no         Center for Internet Security Audit Tools
esm-infra     yes        UA Infra: Extended Security Maintenance
fips          no         NIST-certified FIPS modules
fips-updates  no         Uncertified security updates to FIPS modules
livepatch     yes        Canonical Livepatch service

This machine is not attached to a UA subscription.
See https://ubuntu.com/advantage
```

PR #742, opened 2019-09-09
ux: status when unattached will report available services for the mac…

blackboxsw:
…hine

- Add UAContract.request_resources method which provides the machine's arch,
  kernel and series as query string params in order to determine what
  services are available
- Add contract.get_available_resources helper function which sets up a
  UAContract client, runs the request_resources method and returns a
  list of 'resources'(services)
- Refactor UAConfig.status to get_available_resources and add resource
  available: yes/no values for an unattached machine
- Add format template to status.format_tabular to emit the proper header
  columns and content on an unattached machine

Fixes: #720


uaclient/config.py
```diff
@@ -56,6 +56,7 @@ class UAConfig:
         ),
         "machine-id": DataPath("machine-id", True),
         "machine-token": DataPath("machine-token.json", True),
+        "resources": DataPath("resources.json", False),
```
OddBloke: Why do we need to cache `resources`? Shouldn't the `status-cache` be sufficient for the non-root user case?

uaclient/config.py
```diff
@@ -236,15 +239,49 @@ def _status(self) -> "Dict[str, Any]":
...
+                    LOG.debug(
+                        "Ignoring availability of unknown service {name}"
+                        " from contract server".format(name=resource["name"])
+                    )
+                    continue
+                response["services"].append(
+                    {
+                        "name": resource["name"],
+                        "description": ent_cls.description,
+                        "available": available,
```
OddBloke: I'm a little uneasy about the fact that this service dict has different members to the service dict that attached status produces, whilst living under the same key. I'm worried that it makes parsing our JSON output quite painful (if attached: <one parsing implementation> else: <completely different implementation>). Maybe this is fine because those are two very different modes of operation, but it's making my spidey sense tingle so I'm bringing it up.

What do you think?

uaclient/status.py
```diff
@@ -122,7 +139,10 @@ class UserFacingStatus(enum.Enum):
 See: sudo ua status or https://ubuntu.com/advantage"""
 MESSAGE_UNATTACHED = """\
 This machine is not attached to a UA subscription.
-See `ua attach` or https://ubuntu.com/advantage"""
+See https://ubuntu.com/advantage"""
+
+STATUS_UNATTACHED_HEADER = "\nSERVICE       AVAILABLE  DESCRIPTION"
```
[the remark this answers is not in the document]
blackboxsw: great idea. done

uaclient/config.py
blackboxsw: Dan resources isn't strictly needed, it is just that I've UAContract client caches the unaltered response of any Contract api call in private/<some-cache> If we are relying solely on the adapted response in status-cache, then future us will have to recall that this is not strictly what the contract server returned (or what keys were provided in that response).  I *think* we decided in the past that our debug log which captures this actual response is reasonbale enough, it just makes future triage a bit harder if contract backend changes response and it breaks us.

OddBloke: Yeah, I think the log is the way to go here. Having the file on disk without any context about how we got that content is less useful than what we have in the log, so even for responses we cache, I think that would be the first place I'd go when debugging.

uaclient/config.py
blackboxsw: @OddBloke, I was uneasy about this too. Though I felt that even standardizing these keys would still require extra (is|is_not)_attached to determine whether we use 'available' vs. 'entitled' keys. But I suppose that 'logic' can be embedded in our template strings which would only contain {available} vs {entitled} subst vars.


Some alternatives to keep standardized service keys in both attached and unattached status:
  1.   a. Extend attached status to add service availability = None for each service
        b. Extend unattached status to also include service 'entitled' = None for each service
  2. a. Extend attached status to actually provide availability key values
          * incurs extra remote contract API cost because we have to get those values even though not needed for status
      b. Extend unattached status to include 'entitled' = n/a


Not sure whether we should pay the API cost of 2a given that the attached status shouldn't even be using this available key.

OddBloke: Perhaps another alternative would be to not overload the services key? Attached status would continue to use the `services` key and unattached status would use another key (`resources`?). Then at least it's clear that they will contain different data structures, and consumers can use the top-level keys to determine what's going on (rather than discovering the differences when processing the first service entry).

A completely different alternative would be to continue with the code as-is, but include a message in help that the JSON output isn't yet considered stable. I'm leaning towards this option, as polluting the top-level namespace without careful thought (as I just suggested ^_^) feels likely to just be digging a deeper hole.

uaclient/config.py
blackboxsw: good point, dropped.

uaclient/config.py
blackboxsw: I think I'd like to avoid adding another top-level key too, especially because we have formerly renamed 'entitlements' to 'services' and 'resources' are really just an alias for 'services'. So agreed, we'd probably dig a deeper hole adding a new key. I'll add a _doc key (which will surface at the top of the json response which will document that 'presence of keys under services are experimental' or something like that.

blackboxsw: We have this same _doc key in cloud-init instance-data.json.

uaclient/config.py
```diff
@@ -19,6 +19,8 @@ def cast(_, x):  # type: ignore
 
 
 DEFAULT_STATUS = {
+    "_doc": "Presence of keys services are considered Experimental and may"
```
OddBloke: Given that we haven't specified this format at all, I think we can be more general and say that the entire output should not be considered stable.

Issue #727, opened 2019-08-23
ux: `ua attach` errors


powersj:
No token given
```
$ sudo ua attach
Attach requires a token: sudo ua attach <<TOKEN>>
To obtain a token please visit: https://ubuntu.com/advantage
```

Bad token
```
$ sudo ua attach badkey
Failed to attach the machine. See https://ubuntu.com/advantage
```

No network
```
$ sudo ua attach abcdefghijklmnopqrstuvwxyz01234
Failed to connect to authentication server
Check your Internet connection and try again
```

Expired token
```
$ sudo ua attach expiredkey
Expired token. See https://ubuntu.com/advantage
```

Failure enabling esm-infra
```
$ sudo ua attach abcdefghijklmnopqrstuvwxyz01234
Enabling default service esm-infra
Updating package lists
Failed to enable default services, check: ua status
```

PR #744, opened 2019-09-11
ux: customize error message on missing required parameters

blackboxsw:
Add UAClientRequireAction to required params which need a custom
error message.
Add CLIArgParser which extends ArgumentParser.parse_args to emit
customize error messages on missing UACLientRequireAction params.

Addresses missing token param (part 1 of #727)


blackboxsw:
> did you approach solving this any other way and run into issues that required this approach?
@OddBloke  the approach you suggested I had tried as a first pass, by add the parse.error() logic directly in  action_attach. That also required us to pass in a parser param into each action_* method and I felt (not very strongly) that it decoupled our parameter declaration schema in *_parser methods from the actual implementation which would raise required message errors. I liked in the current approach that at add_argument declaration, we have a reference that this action is a custom Required parameter inline instead of having to also look down in the action_* to see if additional param handling is performed.

But, this current implementation spiraled into something that is fairly heavy and clunky, maybe it is better just to go with the decoupled approach of emitting parser.error in each action_* method, or just raising a custom UserFacingError in those cases and avoid leaking parsers into the action_* methods.

OddBloke:
> just raising a custom UserFacingError in those cases and avoid leaking parsers into the action_* methods.

I like this idea, can we try it out?

uaclient/cli.py
```diff
@@ -96,6 +96,7 @@ def attach_parser(parser=None):
     parser._optionals.title = "Flags"
     parser.add_argument(
         "token",
+        nargs="?",  # required arg custom handling in action_attach
```
OddBloke: "required arg custom handling" isn't super-clear to me; I doubt I will remember what it means in a few months. :-)

uaclient/cli.py
```diff
@@ -80,7 +80,7 @@ def new_f(args, cfg):
 
 def attach_parser(parser=None):
     """Build or extend an arg parser for attach subcommand."""
-    usage = USAGE_TMPL.format(name=NAME, command="attach [token]")
+    usage = USAGE_TMPL.format(name=NAME, command="attach <token>")
```
OddBloke: This is a total aside, but what's the history behind us setting our own usage message rather than using the default?

uaclient/util.py
```diff
@@ -53,7 +53,11 @@ def __init__(
         headers: "Optional[Dict[str, str]]" = None,
         url: "Optional[str]" = None,
     ):
-        super().__init__(str(cause))
+        if getattr(cause, "reason", None):
+            cause_error = cause.reason.strerror  # type: ignore
```
OddBloke: Why the ignore comment?

Would using cast help, a la: https://github.com/CanonicalLtd/ubuntu-advantage-client/blob/537bedf2990200f975d3aa931b777dcd4a2e2533/uaclient/entitlements/__init__.py#L26

OddBloke: (Also, how about `cause_error = getattr(cause, "reason", str(cause))`?

uaclient/status.py
```diff
@@ -127,8 +130,13 @@ class UserFacingStatus(enum.Enum):
 STATUS_SERVICE_HEADER = "\nSERVICE"
 STATUS_TMPL = "{name: <14}{entitled: <26}{status}"
 
-MESSAGE_ATTACH_FAILURE_TMPL = """\
-Could not attach machine. Error contacting server {url}"""
+MESSAGE_ATTACH_REQUIRES_TOKEN = """\
+Attach requires a token: sudo ua attach <<TOKEN>>
```
OddBloke: The doubling of < and > here looks strange to me. `<TOKEN>` looks more normal to me, and would be in line with the updated usage line (except in terms of case).

@powersj What do you think?

(Regardless of which way we decide here, I think this error message and the usage string should be consistent.)

uaclient/cli.py
```diff
@@ -293,13 +294,13 @@ def action_attach(args, cfg):
         return 0
     if os.getuid() != 0:
         raise exceptions.NonRootUserError()
-    contract_token = args.token
-    if not contract_token:
-        print("No valid contract token available")
-        return 1
+    if not args.token:
+        raise exceptions.UserFacingError(
```
OddBloke: What do you think to converting this (and the other introduced `raise`s) to their own exception classes with the message hard-coded?  (So this would be `raise exception.AttachRequiresToken()` or something along those lines.)

uaclient/contract.py
```diff
@@ -211,9 +211,21 @@ def request_updated_contract(
...
+            new_token = contract_client.request_contract_machine_attach(
+                contract_token=contract_token
+            )
+        except util.UrlError as e:
+            if isinstance(e, ContractAPIError):
+                if API_ERROR_INVALID_TOKEN in e:
+                    raise exceptions.UserFacingError(
+                        status.MESSAGE_ATTACH_FAILURE
+                    )
+            if "Temporary failure in name resolution" == str(e):
```
OddBloke: This seems fragile to me. I wonder if we can make it more general. In what cases might we catch a `UrlError` (that isn't a `ContractAPIError`) where it wouldn't be a connectivity error?

uaclient/status.py
powersj: Agreed +1 on single <TOKEN> to be more consistent with other docs as well

uaclient/cli.py
blackboxsw: historically I had tried to get the argparse output to look a bit more like lxd cli output (as was a initial ux request)

uaclient/cli.py
blackboxsw: @OddBloke I had thought about this, but avoided it only because we didn't do any specific handling of that unique error condition. If we have 5 different exceptions defined and we only print(str(e)) and exit(1) I'd like to avoid having the unique exceptions defined as it is just more code.   I would lean slightly toward only defining separate exceptions if we have unique handling behavior associated with processing them. But, it's only a weak suggestion. I can be convinced otherwise.

uaclient/contract.py
blackboxsw: this may be a bigger topic than just the fragility here if we can confirm this works on trusty and xenial for the moment, maybe we can add a tech debt task to improve error handling

powersj: Is this the message that you get when you unplug your system from the network?

powersj: I'm cool with a generic connectivity error for now as a catch all and a tech debt item for more specific messages if and only if we have identified them.

Issue #724, opened 2019-08-23
ux: `ua enable`


powersj:
Enabling esm-infra:
```
$ sudo ua enable esm-infra
One moment, checking your subscription first
Updating package lists
```

Enabling livepatch on HWE kernel:
```
$ sudo ua enable livepatch
One moment, checking your subscription first
Installing snapd and systemd
Installing canonical-livepatch snap
A reboot is required to complete the install    # TODO: is this actually required? 
```

blackboxsw:
Note: don't merge fix for this issue  until ua-contracts backend merges https://github.com/CanonicalLtd/ua-contracts/issues/390

PR #750, opened 2019-09-16
ux: esm service renamed to esm-infra

blackboxsw:
User docs, cli and contracts backend will now all refer to esm as esm-infra in preparation for the eventual esm-apps service.  Even though ESM for Apps will not be on trusty, we want to align the service name on all distributions to avoid confusion.

To support this change
 - ESMEntitlement.name is now esm-infra
 - ua --help docs now source *Entitlement.description values to avoid updating two places in source
    with the same text

Fixes: #724

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1,18 +1,6 @@
 #!/usr/bin/env python
 
-"""\
-Client to manage Ubuntu Advantage support services on a machine.
-
-Available services:
- - cc-eal: Canonical Common Criteria EAL2 Provisioning
-   (https://ubuntu.com/cc-eal)
- - cis-audit: Canonical CIS Benchmark Audit Tool (https://ubuntu.com/cis)
- - esm: Extended Security Maintenance (https://ubuntu.com/esm)
- - fips: FIPS 140-2 (https://ubuntu.com/fips)
- - fips-updates: FIPS 140-2 with updates
- - livepatch: Canonical Livepatch (https://ubuntu.com/livepatch)
-
-"""
+"""Client to manage Ubuntu Advantage support services on a machine."""
 
 import argparse
 from functools import wraps
@@ -323,10 +311,20 @@ def action_attach(args, cfg):
 
 
 def get_parser():
+    service_line_tmpl = " - {name}: {description}{url}"
+    description_lines = [__doc__]
+    sorted_classes = sorted(entitlements.ENTITLEMENT_CLASS_BY_NAME.items())
+    for name, ent_cls in sorted_classes:
+        if ent_cls.help_doc_url:
+            url = " ({})".format(ent_cls.help_doc_url)
+        else:
+            url = ""
+        description.append(service_line_tmpl.format(name=name, description=ent_cls.description, url=url)
+
     parser = argparse.ArgumentParser(
         prog=NAME,
         formatter_class=argparse.RawDescriptionHelpFormatter,
-        description=__doc__,
+        description="\n".join(description),
         usage=USAGE_TMPL.format(name=NAME, command="[command]"),
         epilog=EPILOG_TMPL.format(name=NAME, command="[command]"),
     )
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -319,12 +319,23 @@ def get_parser():
             url = " ({})".format(ent_cls.help_doc_url)
         else:
             url = ""
-        description.append(service_line_tmpl.format(name=name, description=ent_cls.description, url=url)
+        service_line = service_line_tmpl.format(
+            name=name, description=ent_cls.description, url=url
+        )
+        if len(service_line) <= 80:
+            description_lines.append(service_line)
+        else:
+            wrapped_words = []
+            line = service_line
+            while len(line) > 80:
+                [line, wrapped_word] = line.rsplit(" ", 1)
+                wrapped_words.insert(0, wrapped_word)
+            description_lines.extend([line, "   " + " ".join(wrapped_words)])
 
     parser = argparse.ArgumentParser(
         prog=NAME,
         formatter_class=argparse.RawDescriptionHelpFormatter,
-        description="\n".join(description),
+        description="\n".join(description_lines),
         usage=USAGE_TMPL.format(name=NAME, command="[command]"),
         epilog=EPILOG_TMPL.format(name=NAME, command="[command]"),
     )
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -115,8 +115,7 @@ class UAContractClient(serviceclient.UAServiceClient):
 
         @param machine_token: The authentication token needed to talk to
             this contract service endpoint.
-        @param resource: Entitlement name. One of: livepatch, esm, fips or
-            fips-updates.
+        @param resource: Entitlement name.
         @param machine_id: Optional unique system machine id. When absent,
             contents of /etc/machine-id will be used.
 
--- a/uaclient/entitlements/__init__.py
+++ b/uaclient/entitlements/__init__.py
@@ -22,6 +22,7 @@ ENTITLEMENT_CLASSES = [
     LivepatchEntitlement,
 ]  # type: List[Type[UAEntitlement]]
 
+
 ENTITLEMENT_CLASS_BY_NAME = dict(
     (cast(str, cls.name), cls) for cls in ENTITLEMENT_CLASSES
 )  # type: Dict[str, Type[UAEntitlement]]
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -28,6 +28,10 @@ RE_KERNEL_UNAME = (
 
 
 class UAEntitlement(metaclass=abc.ABCMeta):
+
+    # Optional URL for top-level product service information
+    help_doc_url = None  # type: str
+
     @property
     @abc.abstractmethod
     def name(self) -> str:
--- a/uaclient/tests/test_cli.py
+++ b/uaclient/tests/test_cli.py
@@ -1,18 +1,73 @@
+import io
 import logging
 import mock
 import os
 import stat
 import sys
+import textwrap
 
 import pytest
 
-from uaclient.cli import assert_attached_root, main, setup_logging
+from uaclient.cli import assert_attached_root, get_parser, main, setup_logging
+
 from uaclient.exceptions import (
     NonRootUserError,
     UserFacingError,
     UnattachedError,
 )
 from uaclient.testing.fakes import FakeConfig
+
+
+BIG_DESC = "123456789 " * 7 + "next line"
+BIG_URL = "http://" + "adsf" * 10
+
+
+SERVICES_WRAPPED_HELP = textwrap.dedent(
+    """
+Client to manage Ubuntu Advantage support services on a machine.
+ - cc-eal: Common Criteria EAL2 Provisioning Packages
+   (https://ubuntu.com/cc-eal)
+ - cis-audit: Center for Internet Security Audit Tools
+   (https://ubuntu.com/cis-audit)
+ - esm-infra: UA Infra: Extended Security Maintenance (https://ubuntu.com/esm)
+ - fips: NIST-certified FIPS modules (https://ubuntu.com/fips)
+ - fips-updates: Uncertified security updates to FIPS modules
+ - livepatch: Canonical Livepatch service (https://ubuntu.com/livepatch)
+"""
+)
+
+
+class TestCLIParser:
+    maxDiff = None
+
+    @mock.patch("uaclient.cli.entitlements")
+    def test_help_descr_and_url_is_wrapped_at_eighty_chars(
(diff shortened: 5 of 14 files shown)


uaclient/apt.py
```diff
@@ -48,18 +48,18 @@ def assert_valid_apt_credentials(repo_url, username, password):
...
-                    "/usr/lib/apt/apt-helper",
-                    "download-file",
-                    "{}://{}:{}@{}/ubuntu/pool/".format(
-                        protocol, username, password, repo_path
-                    ),
-                    tmpf.name,
-                ],
-                timeout=APT_HELPER_TIMEOUT,
-            )
+        tmpf = tempfile.NamedTemporaryFile(delete=False)
```
OddBloke: The use of

OddBloke: From [`NamedTemporaryFile`'s docs](https://docs.python.org/3/library/tempfile.html): 

> If delete is true (the default), the file is deleted as soon as it is closed.

So using it as a context manager should result in it being cleaned up regardless of exceptions. I've just tested this:

```py
with NamedTemporaryFile() as f:
    print(f.name)
    input('')
    raise Exception
```
and confirmed that the printed file did exist when the input prompt happened, but did not exist after.

Could we drop this change from this PR, and you can re-propose it separately if I'm misunderstanding what you were seeing was wrong? (Ideally, we would have distinct changes in different PRs regardless.)

blackboxsw: Added separate issue #755 and withdrew funky NamedTempFile workaround

uaclient/cli.py
```diff
@@ -325,10 +313,26 @@ def action_attach(args, cfg):
...
+            url = " ({})".format(ent_cls.help_doc_url)
+        else:
+            url = ""
+        service_line = service_line_tmpl.format(
+            name=name, description=ent_cls.description, url=url
+        )
+        if len(service_line) <= 80:
+            description.append(service_line)
+        else:
+            parts = service_line.rsplit(" ", 1)
```
OddBloke: This won't guarantee that we're under 80 characters. For example:

```
123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 more than one word
```

Is this something we need to worry about?

uaclient/cli.py
```diff
@@ -325,10 +313,26 @@ def action_attach(args, cfg):
 
 
 def get_parser():
+    service_line_tmpl = " - {name}: {description}{url}"
+    description = [__doc__]
```
OddBloke: A nit, but `description_lines` is often used to indicate this pattern is being used.

uaclient/config.py
```diff
@@ -214,17 +214,18 @@ def _unattached_status(self) -> "Dict[str, Any]":
                 available = status.UserFacingAvailability.AVAILABLE.value
             else:
                 available = status.UserFacingAvailability.UNAVAILABLE.value
-            ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(resource["name"])
+            alias_name = resource["name"]
```
OddBloke: What's "alias" in this case?

uaclient/contract.py
```diff
@@ -115,8 +115,8 @@ def request_resource_machine_access(
 
         @param machine_token: The authentication token needed to talk to
             this contract service endpoint.
-        @param resource: Entitlement name. One of: livepatch, esm, fips or
-            fips-updates.
+        @param resource: Entitlement name. One of: livepatch, esm-infra, fips
```
OddBloke: Is this list actually accurate? As they're dynamic, should we just drop this instead of updating it?

uaclient/entitlements/base.py
```diff
@@ -28,6 +28,15 @@
 
 
 class UAEntitlement(metaclass=abc.ABCMeta):
+    @property
+    @abc.abstractmethod
+    def help_doc_url(self) -> str:
+        """Optional URL for top-level product service information
+
+        Empty string if not-applicable.
```
OddBloke: Why not `None`? And could this just be set to the not-applicable value, instead of making it abstract and requiring every entitlement to override it?

uaclient/entitlements/__init__.py
```diff
@@ -13,6 +13,8 @@ def cast(_, x):  # type: ignore
         return x
 
 
+ENTITLEMENT_ALIASES = {"esm": "esm-infra"}
```
OddBloke: This doesn't appear to be used anywhere. (Presumably related to my above question about aliases?)

uaclient/cli.py
blackboxsw: Generally I think not because people designing this looked specifically at line size/length when defining these names. I think if we grow a customized 'description' from ua-contracts response we should add this feature. I'll leave a TODO() in the code here in case we source/override description based on ua-contracts response.

uaclient/config.py
blackboxsw: oops vestigial from the alias portion of this branch which was going to alias an type: 'esm' from the backend as 'esm-infra' in the UA client if the backend had not switched over to the new key name. Backend has already landed that component and surfaced esm-infra it in staging so we don't need it. Dropping this remaining alias-related declaration

uaclient/contract.py
blackboxsw: Good suggestion. Dropped.

uaclient/entitlements/__init__.py
blackboxsw: Also vestigial and now dropped. thanks.

uaclient/entitlements/base.py
blackboxsw: Yeah that was me not thinking clearly on this regarding mypy. tried keeping it the same type instead of the Optional declaration. Fixed.

uaclient/cli.py
blackboxsw: on second thought. Let's just guard against the condition as future me won't remember to go back to this part of the code if ua-contracts changes descriptions or URL length etc.
Added a fix.

uaclient/cli.py
```diff
@@ -325,10 +313,30 @@ def action_attach(args, cfg):
...
+        )
+        if len(service_line) <= 80:
+            description_lines.append(service_line)
+        else:
+            wrapped_words = []
+            line = service_line
+            while len(line) > 80:
+                [line, wrapped_word] = line.rsplit(" ", 1)
+                wrapped_words.append(wrapped_word)
+            description_lines.extend([line, "   " + " ".join(wrapped_words)])
```
OddBloke: This looks to me like it will reverse the order of the wrapped_words. Can we add tests to confirm this is correct, please?

uaclient/entitlements/base.py
```diff
@@ -28,6 +28,15 @@
 
 
 class UAEntitlement(metaclass=abc.ABCMeta):
+    @property
+    @abc.abstractmethod
+    def help_doc_url(self) -> "Optional[str]":
```
OddBloke: I don't think I saw a response as to why this can't just be `help_doc_url = None` on the super-class, only being overriden in sub-classes that need to. (As this has a sensible default value, I don't think we need it to be abstract to _force_ sub-classes to pick a value.)

uaclient/entitlements/tests/test_base.py
```diff
@@ -17,6 +17,7 @@
 
 class ConcreteTestEntitlement(base.UAEntitlement):
 
+    help_doc_url = ""
```
OddBloke: This should be `None` if we don't convert `help_doc_url` to have a default (in which case it should be removed).

uaclient/entitlements/tests/test_repo.py
```diff
@@ -30,6 +30,7 @@
 class RepoTestEntitlement(RepoEntitlement):
     """Subclass so we can test shared repo functionality"""
 
+    help_doc_url = ""
```
OddBloke: This should be None or removed, as above.

uaclient/entitlements/base.py
blackboxsw: oops, I misunderstood your original comment. Right, we don't need subclasses to define this property. A simple superclass class-attr would do. Correcting that now.

uaclient/entitlements/base.py
```diff
@@ -28,6 +28,10 @@
 
 
 class UAEntitlement(metaclass=abc.ABCMeta):
+
+    # Optional URL for top-level product service information
+    help_doc_url = ""
```
OddBloke: Why `""` and not `None`?

uaclient/entitlements/tests/test_base.py
```diff
@@ -95,8 +95,8 @@ def test_entitlement_abstract_class(self):
             base.UAEntitlement()
         expected_msg = (
             "Can't instantiate abstract class UAEntitlement with abstract"
-            " methods application_status, description, disable, enable, name,"
```
OddBloke: Looks like we don't need to touch these lines here any longer?

uaclient/tests/test_cli.py
```diff
@@ -15,6 +18,78 @@
...
+            "esm_desc": esm.description,
+            "esm_url": esm.help_doc_url,
+            "fips_desc": fips.description,
+            "fips_url": fips.help_doc_url,
+            "fips_updates_desc": fips_updates.description,
+            "fips_updates_url": fips_updates.help_doc_url,
+            "livepatch_desc": livepatch.description,
+            "livepatch_url": livepatch.help_doc_url,
+        }
+        expected_output = """
```
OddBloke: Could we use `textwrap.dedent` here to keep things a bit tidier?

Also, I wonder if it would be better to just hardcode the expected string here? Otherwise we won't catch mistakes in the attributes themselves. What do you think?

uaclient/tests/test_cli.py
```diff
@@ -15,6 +18,78 @@
...
+ - esm-infra: {esm_desc} ({esm_url})
+ - fips: {fips_desc} ({fips_url})
+ - fips-updates: {fips_updates_desc}
+ - livepatch: {livepatch_desc} ({livepatch_url})
+
+""".format(
+            **fmt_params
+        )
+        with open(help_file, "w") as stream:
+            parser.print_help(file=stream)
```
OddBloke: Could we just use an in-memory file-like object here, to remove the need for the temporary directory? (`io.StringIO`, perhaps?)

(This also applies to the above test case, of course.)

uaclient/entitlements/base.py
blackboxsw: ^ Lazy developer. Fixed to None and declared # type: str. I was working around mypy errors claiming invalid type str on subclasses that properly set this attribute. Turns out I only needed to declare the type on the base class.

uaclient/entitlements/tests/test_base.py
blackboxsw: I fixed this on UAEntitlement base class to be
    help_doc_url = None  # type: str


This resolved the mypy workaround I was trying to add by declaring the base class as type string. No need to set this attr on each subclass.

uaclient/tests/test_cli.py
blackboxsw: I think using textwrap.dedent make sense, but since we are already dealing with lines that are overflowing the 80 char width, trying to use dedent and print the full output will likely be harder to read that the simple structured parameterized format string.

uaclient/tests/test_cli.py
blackboxsw: done.

uaclient/tests/test_cli.py
```diff
@@ -15,6 +20,74 @@
...
+        """Help output is sourced from entitlement name and description."""
+        parser = get_parser()
+        cc_eal = ENTITLEMENT_CLASS_BY_NAME["cc-eal"]
+        cis_audit = ENTITLEMENT_CLASS_BY_NAME["cis-audit"]
+        esm = ENTITLEMENT_CLASS_BY_NAME["esm-infra"]
+        fips = ENTITLEMENT_CLASS_BY_NAME["fips"]
+        fips_updates = ENTITLEMENT_CLASS_BY_NAME["fips-updates"]
+        livepatch = ENTITLEMENT_CLASS_BY_NAME["livepatch"]
+        fmt_params = {
+            "cc_desc": cc_eal.description,
```
OddBloke: I'm still not a fan of using these attributes. If we accidentally modify one of them (either directly, or via modification of the base class, or the inheritance structure), then we won't catch that in testing. I'd prefer for us to hard-code the expected strings here, so that it forms a complete test of the help output (rather than just a test of the help formatting, which is what this covers now). Does that sounds reasonable?

(To be clear, I agree that it's way clearer using the format parameters!)

blackboxsw: Let's fully satisfy that need for strict unit test of structured output. The way I had it would only show us errors if we updated descriptions/urls to affect word wrap, which was really all I was interested in testing there. With a static representation of the entire expected string, we'll update this test if we change description, name or help_doc_url. Any maybe that *is* something that we want human review on specifically as we could easily introduce typos that don't make sense on user-facing docs.

PR #751, opened 2019-09-17
ux: add attached status description column and headers

blackboxsw:
- Add new status template for attached machines that includes description column
- Rename UserFacingStatus.(ACTIVE|INACTIVE) printed values:
    active -> "enabled"
    inactive -> "disabled"
- Add new UserFacingStatus.UNAVAILABLE -> "—"

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -233,6 +233,7 @@ class UAConfig:
 
     def _attached_status(self) -> "Dict[str, Any]":
         """Return configuration of attached status as a dictionary."""
+        from uaclient.contract import get_available_resources
         from uaclient.entitlements import ENTITLEMENT_CLASSES
 
         response = copy.deepcopy(DEFAULT_STATUS)
@@ -249,10 +250,20 @@ class UAConfig:
             response["expires"] = datetime.strptime(
                 contractInfo["effectiveTo"], "%Y-%m-%dT%H:%M:%SZ"
             )
+
+        resources = get_available_resources(self)
+        unavailable_resources = [
+            resource["name"]
+            for resource in sorted(resources, key=lambda x: x["name"])
+            if not resource["available"]
+        ]
+
         for ent_cls in ENTITLEMENT_CLASSES:
             ent = ent_cls(self)
             contract_status = ent.contract_status().value
             ent_status, details = ent.user_facing_status()
+            if ent.name in unavailable_resources:
+                ent_status = status.UserFacingStatus.UNAVAILABLE
             service_status = {
                 "name": ent.name,
                 "description": ent.description,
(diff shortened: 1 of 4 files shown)


uaclient/config.py
```diff
@@ -249,12 +250,23 @@ def _attached_status(self) -> "Dict[str, Any]":
             response["expires"] = datetime.strptime(
                 contractInfo["effectiveTo"], "%Y-%m-%dT%H:%M:%SZ"
             )
+
+        resources = get_available_resources(self)
```
OddBloke: The two places we call this we call it with `self` as the first argument from `UAConfig`; should it be a method on `UAConfig`?

(This doesn't need changing for us to land this PR, I'm just curious what you think.)

uaclient/status.py
```diff
@@ -146,8 +152,8 @@ class UserFacingStatus(enum.Enum):
 
 STATUS_UNATTACHED_TMPL = "{name: <14}{available: <11}{description}"
 
-STATUS_SERVICE_HEADER = "\nSERVICE"
-STATUS_TMPL = "{name: <14}{entitled: <26}{status}"
+STATUS_HEADER = "SERVICE       ENTITLED  STATUS    DESCRIPTION"
```
OddBloke: Can we do the same trick as before, and define the header in terms of the template?

uaclient/tests/test_config.py
```diff
@@ -27,6 +27,15 @@
 M_PATH = "uaclient.entitlements."
 
 
+RESP_ALL_RESOURCES_AVAILABLE = [
+    {"name": name, "available": True} for name in ENTITLEMENT_CLASS_BY_NAME
+]
+RESP_ONLY_FIPS_RESOURCE_AVAILABLE = [
+    {"name": name, "available": bool(name == "fips")}
```
OddBloke: Total nit, but `name == "fips"` is already a bool so the call here is a noop. No need to fix here.

uaclient/status.py
blackboxsw: We can't do the same trick here because of the unprintable characters we emit due to our colorize calls, it adds about 5 unprintable characters that mess with the .format alignment if we stuff the header strings directly into the formatted template

OddBloke: Ah, bummer. In that case, could we get a comment on the template explaining why the numbers contained within it aren't actually the widths we'll see in the output? That seems like something that would be very easy to trip over in future.

uaclient/config.py
blackboxsw: There is a smell here, and I'm not sure how we should resolve it. get_available_resources is a utility function in the contract module and really relates to only contract client operations, so I don't feel like it belongs on the UAConfig class because we aren't really managing local system configuration. But, the fact that we are passing in the entire instance makes me feel like maybe we need to think instead about maybe passing in the outputs from get_available_resources as a parameter to UAConfig.status -> _attached_status and _unattached_status  to avoid this conflict/smell.

Issue #718, opened 2019-08-23
Update ESM-infra repos and workflow


powersj:
- [ ] Update ESM sources.list for `<series>-infra-[updates|security]`
- [ ] Only install ESM sources.list on package install for Trusty, not on any other release
- [ ] Ensure ESM only enables updates, if the user previous had it enabled
- [x] Rename esm service to esm-infra

# Examples

Example Trusty ESM-infra sources list for security and updates. If updates was not already enabled, then the 3rd line in the file would also be commented out.
```
$ cat /etc/apt/sources.list.d/ubuntu-infra-trusty.list
deb https://esm.ubuntu.com/ubuntu trusty-infra-security main
# deb-src https://esm.ubuntu.com/ubuntu trusty-infra-security main
deb https://esm.ubuntu.com/ubuntu trusty-infra-updates main
# deb-src https://esm.ubuntu.com/ubuntu trusty-infra-updates main
```

At a later date we will have esm-apps, and will need to create another service and file for that.

powersj:
Removing external blocker has the required bug has been merged, we should not hold on contracts server for this work now.

Issue #693, opened 2019-07-15
avoid duplicate sources.list entries after release upgrade


panlinux:
After a do-release-upgrade from an esm-enabled precise system, to trusty, esm was still enabled at the end.

The release upgrade procedure changed the distro name in the esm `sources.list.d/` file from precise to trusty, but kept the filename. `/etc/apt/sources.list.d/ubuntu-esm-precise.list` now has:
```
deb https://username:password@esm.ubuntu.com/ubuntu trusty-updates main
```
Note how the file is still named after precise, but now contains a trusty repo, and with the old-style esm credentials. This is what is keeping this trusty system linked to esm at this point.

The problem happens once you install the new client, 19.5.1. It will, in its postinst, create `ubuntu-esm-trusty.list`, since that file doesn't exist yet. It will point at the same repository as
the already existing `ubuntu-esm-precise.list`, and we get this complaint:
```
Setting up ubuntu-advantage-tools (19.5.1~ubuntu14.04.1~ppa2) ...
W: Duplicate sources.list entry https://esm.ubuntu.com/ubuntu/
trusty-security/main amd64 Packages
(/var/lib/apt/lists/esm.ubuntu.com_ubuntu_dists_trusty-security_main_binary-amd64_Packages)
W: Duplicate sources.list entry https://esm.ubuntu.com/ubuntu/
trusty-security/main i386 Packages
(/var/lib/apt/lists/esm.ubuntu.com_ubuntu_dists_trusty-security_main_binary-i386_Packages)
W: Duplicate sources.list entry https://esm.ubuntu.com/ubuntu/
trusty-updates/main amd64 Packages
(/var/lib/apt/lists/esm.ubuntu.com_ubuntu_dists_trusty-updates_main_binary-amd64_Packages)
W: Duplicate sources.list entry https://esm.ubuntu.com/ubuntu/
trusty-updates/main i386 Packages
(/var/lib/apt/lists/esm.ubuntu.com_ubuntu_dists_trusty-updates_main_binary-i386_Packages)
W: You may want to run apt-get update to correct these problems
```
Not the end of the world, as apt still exits with status 0, but this will show up everytime apt update is run, even after ua attach is done. One cannot simply delete one of these files, as the one named after precise is the only one with the esm credentials at this point (before ua attach), and is what is keeping this system up-to-date. The proper (manual) procedure to finish the release upgrade is:
- run `ua attach`
[...]

PR #752, opened 2019-09-17
packaging: enable esm trusty-infra-updates and -security suites on trusty

### LINKED ISSUE #778: ESM failed to enable, but apt preferences file is removed
The `/etc/apt/preferences.d/ubuntu-esm-trusty` file (on trusty) is created on postinst if ESM is **disabled**:
```bash
    if [ ! -e "$ESM_APT_PREF_FILE" ]; then
        ESM_POLICY=`apt-cache policy | grep https://esm.ubuntu.com | awk 'END{print $1}'`
        if [ "500" != "$ESM_POLICY" ]; then  # we are inactive
            cat > $ESM_APT_PREF_FILE <<EOF
Package: *
Pin: release o=UbuntuESM, n=trusty
Pin-Priority: never
EOF
        fi
```
This tells apt to never consider packages from ESM for upgrades.
The current client is failing to enable ESM because of the rename work that is still in progress (esm -> esm-infra), but what I noticed is that after the failed attempt to enable ESM, that preferences file is gone:
```diff
--- after-ua-install.list	2019-10-01 14:07:22.390405496 +0000
+++ after-ua-my-attach.list	2019-10-01 14:28:33.086405496 +0000
@@ -84,7 +84,6 @@
 /etc/apt/apt.conf.d/50unattended-upgrades
 /etc/apt/apt.conf.d/51ubuntu-advantage-esm
 /etc/apt/apt.conf.d/20apt-esm-hook.conf
-/etc/apt/preferences.d/ubuntu-esm-trusty
 /etc/apt/trusted.gpg.d/ahasenack-ua-temp.gpg~
 /etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg
 /etc/apt/trusted.gpg.d/ahasenack-ua-temp.gpg
```
This means apt upgrade/dist-upgrade is considering ESM packages, but it then fails to download them:
```
...
After this operation, 427 MB of additional disk space will be used.
0% [Working]
Get:1 https://esm.ubuntu.com/ubuntu/ trusty-security/main bash amd64 4.3-7ubuntu1.8+esm1 [575 kB]
Err https://esm.ubuntu.com/ubuntu/ trusty-security/main bash amd64 4.3-7ubuntu1.8+esm1
  HttpError401
...
```
This is the output of `ua status` in this state:
```
ubuntu@trusty-ua:~$ sudo ua status
cc-eal        none      —         Common Criteria EAL2 Provisioning Packages
cis-audit     none      n/a       Center for Internet Security Audit Tools
esm           none      n/a       UA Infra: Extended Security Maintenance
fips          none      —         NIST-certified FIPS modules
fips-updates  none      —         Uncertified security updates to FIPS modules
livepatch     entitled  disabled  Canonical Livepatch service

     Account: andreas.hasenack@canonical.com
Subscription: andreas.hasenack@canonical.com
```

blackboxsw:
Ensure that esm apt list files are not installed on non-trusty or
removed across package upgrade.

Also add upgrade path from existing ubuntu-advantage-tools v8 (the shell version) which named the apt 
/etc/apt/sources.list.d/ubuntu-esm-trusty.list.

Since service-name is now esm-infra we needed to mv the old filename to the new ubuntu-esm-trusty.list
to maintain properly configured esm apt sources, as well as overwrite the settings if the new-client attaches.

Fixes: #718

Also fix #778 
- Emit Priority-Pin: Never if we are also installing the apt/sources.list.d/*esm file (because we are not enabled for esm package downloads yet)

ESM repos are essentially read-only public for package lists in order to allow non-root users to see upgradable esm packages. Change debian/postinst to automatically create an apt pref file with  Pin-Priority: never for ESM until the service is enabled. This will prevent apt from thinking it has access and credentials to actually download and install esm packages until `ua enable esm` is called (which adds credentials and removes the pin file).

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -2,6 +2,8 @@
 
 set -e
 
+. /etc/os-release  # For VERSION_ID
+
 ESM_KEY_ID_TRUSTY="56F7650A24C9E9ECF87C4D8D4067E40313CB4B13"
 ESM_APT_GPG_KEY="/etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg"
 ESM_APT_GPG_KEY_OLD="/etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg"
@@ -20,6 +22,11 @@ export_gpg_key_from_shared_keyring() {
     fi
 }
 
+unconfigure_esm() {
+    rm -f $ESM_APT_GPG_KEY $ESM_APT_GPG_KEY_OLD $ESM_APT_SOURCE_FILE
+    rm -f $ESM_APT_PREF_FILE
+}
+
 configure_esm() {
     rm -f $ESM_APT_GPG_KEY_OLD  # Remove retired key from key list
     if [ ! -f "$ESM_APT_GPG_KEY" ]; then
@@ -29,11 +36,11 @@ configure_esm() {
     if [ ! -e "$ESM_APT_SOURCE_FILE" ]; then
         cat > $ESM_APT_SOURCE_FILE <<EOF
 # Written by ubuntu-advantage-tools
-deb https://esm.ubuntu.com/ubuntu trusty-security main
-# deb-src https://esm.ubuntu.com/ubuntu trusty-security main
+deb https://esm.ubuntu.com/ubuntu trusty-infra-security main
+# deb-src https://esm.ubuntu.com/ubuntu trusty-infra-security main
 
-deb https://esm.ubuntu.com/ubuntu trusty-updates main
-# deb-src https://esm.ubuntu.com/ubuntu trusty-updates main
+deb https://esm.ubuntu.com/ubuntu trusty-infra-updates main
+# deb-src https://esm.ubuntu.com/ubuntu trusty-infra-updates main
 EOF
     fi
     if [ ! -e "$ESM_APT_PREF_FILE" ]; then
@@ -67,7 +74,11 @@ case "$1" in
       # CACHE_DIR is no longer present or used since 19.1
       rm -rf /var/cache/ubuntu-advantage-tools
 
-      grep -iq trusty /etc/os-release && configure_esm
+      if [ "14.04" = "$VERSION_ID" ]; then
+        configure_esm
+      else
+        unconfigure_esm
+      fi
       if [ ! -f /var/log/ubuntu-advantage.log ]; then
           touch /var/log/ubuntu-advantage.log
       fi
--- a/debian/postinst
+++ b/debian/postinst
@@ -7,7 +7,8 @@ set -e
 ESM_KEY_ID_TRUSTY="56F7650A24C9E9ECF87C4D8D4067E40313CB4B13"
 ESM_APT_GPG_KEY="/etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg"
 ESM_APT_GPG_KEY_OLD="/etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg"
-ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
+ESM_APT_SOURCE_FILE_OLD="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
+ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list"
 ESM_APT_PREF_FILE="/etc/apt/preferences.d/ubuntu-esm-trusty"
 UA_KEYRING_FILE="/usr/share/keyrings/ubuntu-advantage-keyring.gpg"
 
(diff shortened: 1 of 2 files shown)


debian/postinst
```diff
@@ -19,6 +21,11 @@ export_gpg_key_from_shared_keyring() {
     fi
 }
 
+unconfigure_esm() {
+    rm -f $ESM_APT_GPG_KEY $ESM_APT_GPG_KEY_OLD $ESM_APT_SOURCE_FILE
```
blackboxsw: test comment on closed PR.

panlinux:
The example in issue #718 seems to imply the sources list file should have been named `/etc/apt/sources.list.d/ubuntu-infra-trusty.list`, but this branch is leaving it untouched and still calls it `ubuntu-esm-trusty.list`. Which one is correct?

blackboxsw:
> The example in issue #718 seems to imply the sources list file should have been named `/etc/apt/sources.list.d/ubuntu-infra-trusty.list`, but this branch is leaving it untouched and still calls it `ubuntu-esm-trusty.list`. Which one is correct?

Good catch Andreas, even the issue filed is incorrect. The uaclient runtime emits /etc/apt/sources.list.d/ubuntu-<lowercase-service-name-key>-<series>.list files for any apt-based ua service.  
So, with the newly named esm-infra ua client will create /etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list.

That said, on upgrade path from old 'bash' u-a-tools package to new 'python' u-a-tools, we need to migrate any existing sources list setup for a previously enabled esm old client to ensure they retain that behavior. We also want new client to replace the old client's esm apt config file if users "ua attach" to the new client thus enabling esm.   The best solution here seems to be mv the old esm apt list file  to the new filename so if new-client is 'ua attached' the enabled esm-infra will automatically replace the new apt sources list file ubuntu-esm-infra-trusty.list

debian/postinst
```diff
@@ -19,20 +22,28 @@ export_gpg_key_from_shared_keyring() {
...
+
 configure_esm() {
     rm -f $ESM_APT_GPG_KEY_OLD  # Remove retired key from key list
     if [ ! -f "$ESM_APT_GPG_KEY" ]; then
         export_gpg_key_from_shared_keyring $ESM_KEY_ID_TRUSTY
     fi
 
+    if [ -e "$ESM_APT_SOURCE_FILE_OLD" ]; then
+	mv $ESM_APT_SOURCE_FILE_OLD $ESM_APT_SOURCE_FILE
+    fi
```
panlinux: You also need to update `debian/postrm` to remove the new name instead of the old one.

debian/postinst
```diff
@@ -66,7 +77,11 @@ case "$1" in
       # CACHE_DIR is no longer present or used since 19.1
       rm -rf /var/cache/ubuntu-advantage-tools
 
-      grep -iq trusty /etc/os-release && configure_esm
+      if [ "14.04" = "$VERSION_ID" ]; then
+        configure_esm
+      else
+        unconfigure_esm
```
panlinux: Why the unconfigure_esm addition?

blackboxsw: Only install ESM sources.list on package install for Trusty, not on any other release
So, we actually don't want to advertize any esm packages on non-trusty in general.

debian/postinst
blackboxsw: thanks added that for the purge

debian/postinst
panlinux: The point is we didn't have `unconfigure_esm` before. Are you thinking about release upgrades here, i.e., when going from trusty to xenial sometime in the future?

panlinux: Ops, now I actually read your response fully.

debian/postinst
```diff
@@ -19,32 +23,36 @@ export_gpg_key_from_shared_keyring() {
...
+}
+
 configure_esm() {
     rm -f $ESM_APT_GPG_KEY_OLD  # Remove retired key from key list
     if [ ! -f "$ESM_APT_GPG_KEY" ]; then
         export_gpg_key_from_shared_keyring $ESM_KEY_ID_TRUSTY
     fi
 
+    if [ -e "$ESM_APT_SOURCE_FILE_OLD" ]; then
+	mv $ESM_APT_SOURCE_FILE_OLD $ESM_APT_SOURCE_FILE
```
panlinux: I think you need to rename the preferences file too.

blackboxsw: We now create the apt preferences file only if a new system is installing the ESM_APT_SOURCE_FILE because only on this condition is the system both unattached and unenabled for esm.   For trusty old client with enabled ESM, we would have renamed the old apt-sources list file which contains legacy credentials already, so we do not need the preferences file pinning that source disabled to 'Pin-priority: never'.

panlinux:
man

panlinux:
I'm just clicking on the wrong buttons today

debian/postinst
```diff
@@ -67,7 +75,11 @@ case "$1" in
       # CACHE_DIR is no longer present or used since 19.1
       rm -rf /var/cache/ubuntu-advantage-tools
 
-      grep -iq trusty /etc/os-release && configure_esm
+      if [ "14.04" = "$VERSION_ID" ]; then
+        configure_esm
+      else
+        unconfigure_esm
```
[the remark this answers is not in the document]
blackboxsw: I think this sort of aligns with us  fixing #693 as we have to think about precise -> trusty upgrades in that scenario. I'll add context to that issue.

PR #753, opened 2019-09-17
ux: simplified attach console messages

blackboxsw:
- Drop quotes around service name when Enabling default service 'x'
 - Drop apt authenticated repository messages to the 

Fixes: #728

### DIFF
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -172,9 +172,9 @@ MESSAGE_ENABLE_FAILURE_UNATTACHED_TMPL = """\
 To use '{name}' you need an Ubuntu Advantage subscription.
 Personal and community subscriptions are available at no charge
 See https://ubuntu.com/advantage"""
-MESSAGE_ENABLE_BY_DEFAULT_TMPL = "Enabling default service '{name}'"
+MESSAGE_ENABLE_BY_DEFAULT_TMPL = "Enabling default service {name}"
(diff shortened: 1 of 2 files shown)


uaclient/status.py
```diff
@@ -172,9 +172,9 @@ class UserFacingStatus(enum.Enum):
 To use '{name}' you need an Ubuntu Advantage subscription.
 Personal and community subscriptions are available at no charge
 See https://ubuntu.com/advantage"""
-MESSAGE_ENABLE_BY_DEFAULT_TMPL = "Enabling default service '{name}'"
+MESSAGE_ENABLE_BY_DEFAULT_TMPL = "Enabling default service {name}"
 MESSAGE_ENABLE_BY_DEFAULT_MANUAL_TMPL = """\
-Service '{name}' is recommended by default. To enable run: ua enable {name}"""
+Service {name} is recommended by default. Run: sudo ua enable {name}"""
```
OddBloke: I don't see this message in the UX doc at all. @powersj, this is the message we emit when `--no-auto-enable` is passed to `ua attach`, can you confirm that this is what we want?

OddBloke:
This LGTM, provided @powersj confirms that the message I commented on is what we want.

uaclient/status.py
powersj: +1

PR #756, opened 2019-09-19
test_cli_detach: add more tests

OddBloke:


uaclient/tests/test_cli_detach.py
```diff
@@ -26,3 +26,71 @@ def test_unattached_error_message(self, getuid):
...
         assert status.MESSAGE_UNATTACHED == err.value.msg
+
+    @mock.patch("uaclient.cli.entitlements")
+    def test_entitlements_disabled_if_can_disable(
+        self, m_entitlements, m_getuid
+    ):
+        m_getuid.return_value = 0
+
+        def cls_mock_factory(can_disable):
+            return mock.Mock(
```
blackboxsw: Wow thanks for this I had been creating fake class definitions instead of leveraging mock for that work.

OddBloke: In some (many?) cases, a fake would be better than a mock, because it can often result in less confusing test code, as well as "automatically" inheriting changes to the interface being tested against. In this case, though, the interface of the mocked object being used is small and stable enough that I didn't think it was worthwhile.

Issue #729, opened 2019-08-23
ux: detach


powersj:
Only esm-infra enabled:

```
$ sudo ua detach
Detach will disable the following service: 
    esm-infra
Are you sure? (y/N)
Disabling esm-infra
This machine is now detached
```

Both trusty services enabled:

```
$ sudo ua detach
Detach will disable the following services: 
    esm-infra
    livepatch
Are you sure? (y/N)
Disabling esm-infra
Disabling livepatch
This machine is now detached
```

Answering anything except "y":

```
$ sudo ua detach
Detach will disable the following services: 
    esm-infra
    livepatch
Are you sure? (y/N) N
```

OddBloke:
As a user, I would normally expect this prompt to be (y/N) to indicate what the default behaviour should be. What do you think, @powersj?

PR #758, opened 2019-09-20
detach: emit list of services before disabling them

OddBloke:
This is the next step for #729. It's not very big, but the next part will be the Y/n prompting, which will warrant its own PR with no other changes to consider.

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -282,7 +282,8 @@ def action_detach(args, cfg):
         if ent.can_disable(silent=True):
             to_disable.append(ent)
     if to_disable:
-        print("Detach will disable the following services:")
+        suffix = "s" if len(to_disable) > 1 else ""
+        print("Detach will disable the following service{}:".format(suffix))
         for ent in to_disable:
             print("    {}".format(ent.name))
     for ent in to_disable:
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -276,10 +276,17 @@ def action_detach(args, cfg):
 
     @return: 0 on success, 1 otherwise
     """
+    to_disable = []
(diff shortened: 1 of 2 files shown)


uaclient/cli.py
```diff
@@ -276,10 +276,17 @@ def action_detach(args, cfg):
...
     @return: 0 on success, 1 otherwise
     """
+    to_disable = []
     for ent_cls in entitlements.ENTITLEMENT_CLASSES:
         ent = ent_cls(cfg)
         if ent.can_disable(silent=True):
-            ent.disable(silent=True)
+            to_disable.append(ent)
+    if to_disable:
+        print("Detach will disable the following services:")
```
blackboxsw: Given the original issue I think we need to handle the optional pluralization:
plural = "s" if len(to_disable) > 1 else ""
print("Detach will disable the following service{}:".fornat(plural))

uaclient/tests/test_cli_detach.py
```diff
@@ -94,3 +95,25 @@ def test_returns_zero(self, m_entitlements, m_getuid):
         ret = action_detach(mock.MagicMock(), mock.MagicMock())
 
         assert 0 == ret
+
+    @mock.patch("uaclient.cli.entitlements")
+    def test_informational_message_emitted(
+        self, m_entitlements, m_getuid, capsys
+    ):
+        m_getuid.return_value = 0
+        m_entitlements.ENTITLEMENT_CLASSES = [
```
blackboxsw: Would like to see a test as well, or parameterize the existing test to validate the plural services vs service:

uaclient/cli.py
OddBloke: Good catch, thanks!

PR #759, opened 2019-09-22
fips: generic reboot required message

blackboxsw:
FIPS: generic messaging for reboot required during post_enable. Don't mention pending state.

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -59,9 +59,7 @@ class FIPSEntitlement(FIPSCommonEntitlement):
     title = "FIPS"
     description = "NIST-certified FIPS modules"
     messaging = {
-        "post_enable": [
-            "FIPS configured and pending, please reboot to make active."
-        ]
+        "post_enable": ["A reboot is required to complete the install"]
     }
     origin = "UbuntuFIPS"
     repo_url = "https://esm.ubuntu.com/fips"


uaclient/entitlements/base.py
```diff
@@ -46,6 +46,12 @@ def description(self) -> str:
         """A sentence describing this entitlement"""
         pass
 
+    @property
+    @abc.abstractmethod
+    def reboot_on_enable(self) -> bool:
```
OddBloke: Could this just be `reboot_on_enable = False`?

uaclient/entitlements/fips.py
```diff
@@ -59,9 +60,7 @@ class FIPSEntitlement(FIPSCommonEntitlement):
     title = "FIPS"
     description = "NIST-certified FIPS modules"
     messaging = {
-        "post_enable": [
-            "FIPS configured and pending, please reboot to make active."
-        ]
+        "post_enable": ["A reboot is required to complete the install"]
```
OddBloke: If we have a generic `reboot_on_enable` attribute now, should this messaging also be made generic? Would this perhaps be better implemented as specific to the FIPS classes for now (perhaps by overriding `enable` with a super call?), until we have more than one place it's needed?

uaclient/entitlements/fips.py
```diff
@@ -38,7 +39,7 @@ def application_status(self) -> "Tuple[status.ApplicationStatus, str]":
         if running_kernel.endswith("-fips"):
             return super_status, super_msg
         return (
-            status.ApplicationStatus.PENDING,
+            status.ApplicationStatus.ENABLED,
```
OddBloke: This is a separate change from the reboot-required change. Ideally these would be split into separate commits (and perhaps separate PRs, though the PENDING->ENABLED stuff is pretty minimal).

uaclient/entitlements/base.py
blackboxsw: correct. good thought. dropped it and focused this changeset specifically on the FIPS subclasses untill we have a 2nd use-case that requires reboot post enable.

uaclient/entitlements/fips.py
blackboxsw: good thought, separated to #762 762

uaclient/entitlements/fips.py
```diff
@@ -42,6 +45,14 @@ def application_status(self) -> "Tuple[status.ApplicationStatus, str]":
             "Reboot to FIPS kernel required",
         )
 
+    def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
+        """FIPS entitlements need to set reboot-required flag upon success"""
+        enabled = super().enable(silent_if_inapplicable=silent_if_inapplicable)
+        if enabled:
+            if not os.path.exists(REBOOT_REQUIRED_FILE):
+                util.write_file(REBOOT_REQUIRED_FILE, "")
```
OddBloke: My `/var/run/reboot-required` currently contains:

```
*** System restart required ***
```

and I also have `/var/run/reboot-required.pkgs`, which contains:

```
libc6
libc6
```

Should we be inserting that string into the file, rather than empty content? Should we be adding `ubuntu-advantage-tools` or `fips` or something to the `.pkgs` file?

powersj: @OddBloke can you please ask in #ubuntu-devel the correct usage of /var/run/reboot-required?

OddBloke: @powersj It occurs to me: as it's the kernel that requires a reboot to use, why isn't the kernel packaging writing this out for us?

OddBloke: Aaaand the answer is: it is!

```
$ cat /var/run/reboot-required
*** System restart required ***
$ cat /var/run/reboot-required.pkgs 
libssl1.0.0
linux-image-4.4.0-1002-fips
linux-base
linux-base
```

uaclient/entitlements/fips.py
blackboxsw: I think since we are limiting the implementation of reboot_required feature to just fips* entitlements, it makes sense for us the make the generic post_enable messaging common in FIPSCommentEntitlement parent class. Until we have other use-cases up at the RepoEntitlement parent class, I don't think we should promote this generic messaging and handling up to that level yet. I've promoted the generic post_enable messaging to FIPSCommonEntitlement for this branch.

uaclient/entitlements/fips.py
OddBloke: I pinged in IRC, but to make sure it doesn't get lost: I still think we can drop this code.

Issue #721, opened 2019-08-23
Reboot required messaging


powersj:
When a service requires a reboot do the following:

- Tell user a reboot is required, do not reboot
- Set reboot required
- Mark service enabled, not pending

PR #762, opened 2019-09-24
Feature/drop pending state

blackboxsw:
fips: report ENABLED instead of PENDING status before rebooting

No longer set FIPS to 'pending' state. Leave it as 'enabled' because
we are touching /var/run/reboot-required to alert the system to a reboot
needed.

Fixes: #721

### DIFF
--- a/uaclient/entitlements/tests/test_fips.py
+++ b/uaclient/entitlements/tests/test_fips.py
@@ -70,7 +70,9 @@ class TestFIPSEntitlementEnable:
             stack.enter_context(
                 mock.patch(M_GETPLATFORM, return_value={"series": "xenial"})
             )
-            stack.enter_context(mock.patch(M_REPOPATH + "os.path.exists"))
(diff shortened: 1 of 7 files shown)


uaclient/entitlements/tests/test_fips.py
```diff
@@ -352,7 +352,7 @@ def test_non_enabled_passed_through(
             ),
             (
                 {"kernel": "4.4.0-148-generic"},
-                status.ApplicationStatus.PENDING,
+                status.ApplicationStatus.ENABLED,
```
OddBloke: The test this change is attached to still has "pending" in its name. Can we confirm that it's still a useful test, and fix the name if it is?

blackboxsw: Test is still useful as we do surface unique application status messages based on the running kernel. Even though ua status cli doesn't distinguish between ENABLED, 'still needs reboot' and ENABLED, 'active'.

PR #766, opened 2019-09-27
detach: prompt for user confirmation before detach

### LINKED ISSUE #729: ux: detach

OddBloke:
This is the last of the changes required to fix #729.

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -243,7 +243,7 @@ def action_detach(args, cfg):
         for ent in to_disable:
             print("    {}".format(ent.name))
     if not util.prompt_for_confirmation():
-        return 0
+        return 1
     for ent in to_disable:
         ent.disable(silent=True)
     cfg.delete_cache()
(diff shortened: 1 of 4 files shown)


uaclient/cli.py
```diff
@@ -286,6 +286,8 @@ def action_detach(args, cfg):
         print("Detach will disable the following service{}:".format(suffix))
         for ent in to_disable:
             print("    {}".format(ent.name))
+    if not util.prompt_for_confirmation():
+        return 0
```
blackboxsw: I *think* we should exit 1 here on 'n' as apt does similarly

Do you want to continue? [Y/n] n
Abort.
csmith@uptown:~$ echo $?
1

PR #767, opened 2019-09-27
Introduce integration testing driven by behave

OddBloke:
This PR introduces integration testing for the project. It leverages the [behave](https://behave.readthedocs.io/en/latest/index.html) BDD framework to allow us to write tests that look similar to the UX documentation we've been given, with the implementation of those steps written in Python code.

This PR is enough for people to start iterating on new tests locally, but there is more work to be done: #853, #854, #855, #856, #857

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,34 @@ The client also includes built-in dep8 tests. These are run as follows:
 autopkgtest -U --shell-fail . -- lxd ubuntu:xenial
 ```
 
+### Integration Tests
+
+ubuntu-advantage-client uses [behave](https://behave.readthedocs.io)
+for its integration testing.
+
+The integration test definitions are stored in the `features/`
+directory and consist of two parts: `.feature` files that define the
+tests we want to run, and `.py` files which implement the underlying
+logic for those tests.
+
+To run the tests, you can use `tox`:
+
+```shell
+tox -e behave
+```
+
+or, if you just want to run a specific file, or a test within a file:
+
+```shell
+tox -e behave features/unattached_commands.feature
+tox -e behave features/unattached_commands.feature:55
+```
+
+(If you're getting started with behave, we recommend at least reading
+through [the behave
+tutorial](https://behave.readthedocs.io/en/latest/tutorial.html) to get
+an idea of how it works, and how tests are written.)
+
 ## Building
 
 The packaging for the UA client package (ubuntu-advantage-tools) is
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -37,7 +37,7 @@ def _wait_for_boot(context: Context) -> None:
 
 
 @given("a trusty lxd container")
-def step_impl(context):
+def given_a_trusty_lxd_container(context):
     now = datetime.datetime.now()
     context.container_name = CONTAINER_PREFIX + now.strftime("%s%f")
     subprocess.run(["lxc", "launch", "ubuntu:trusty", context.container_name])
@@ -52,7 +52,7 @@ def step_impl(context):
 
 
 @given("ubuntu-advantage-tools is installed")
-def step_impl(context):
+def given_uat_is_installed(context):
     _lxc_exec(
         context,
         [
@@ -68,7 +68,7 @@ def step_impl(context):
 
 
 @when("I run `{command}` as non-root")
-def step_impl(context, command):
+def when_i_run_command(context, command):
     process = _lxc_exec(
         context, shlex.split(command), capture_output=True, text=True
     )
@@ -76,10 +76,10 @@ def step_impl(context, command):
 
 
 @then("I will see the following on stdout")
-def step_impl(context):
+def then_i_will_see_on_stdout(context):
     assert_that(context.process.stdout.strip(), equal_to(context.text))
 
 
 @then("I will see the following on stderr")
-def step_impl(context):
+def then_i_will_see_on_stderr(context):
     assert_that(context.process.stderr.strip(), equal_to(context.text))
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -1,6 +1,7 @@
 import datetime
 import shlex
 import subprocess
+import time
 from typing import List
 
 from behave import given, then, when
@@ -17,6 +18,20 @@ def _lxc_exec(context: Context, cmd: List[str], *args, **kwargs):
     )
 
 
+def _wait_for_boot(context: Context) -> None:
+    retries = [2] * 5
+    for sleep_time in retries:
+        process = _lxc_exec(
+            context, ["runlevel"], capture_output=True, text=True
+        )
+        _, runlevel = process.stdout.strip().split(" ", 2)
+        if runlevel == "2":
+            break
+        time.sleep(sleep_time)
+    else:
+        raise Exception("System did not boot in {}s".format(sum(retries)))
+
+
 @given("a trusty lxd container")
 def step_impl(context):
     now = datetime.datetime.now()
@@ -28,6 +43,8 @@ def step_impl(context):
         subprocess.run(["lxc", "delete", context.container_name])
 
     context.add_cleanup(cleanup_container)
+
+    _wait_for_boot(context)
 
 
 @given("ubuntu-advantage-tools is installed")
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -1,0 +1,39 @@
+import subprocess
+from typing import List
+
+from behave import given, then, when
+from hamcrest import assert_that, equal_to
+
+
+CONTAINER_NAME = "behave-test"
+
+
+def _lxc_exec(cmd: List[str], *args, **kwargs):
+    return subprocess.run(
+        ["lxc", "exec", CONTAINER_NAME, "--"] + cmd, *args, **kwargs
+    )
+
+
+@given("a trusty lxd container")
+def step_impl(context):
+    subprocess.run(["lxc", "launch", "ubuntu:trusty", CONTAINER_NAME])
+
+
+@given("ubuntu-advantage-tools is installed")
+def step_impl(context):
+    _lxc_exec(
+        ["add-apt-repository", "--yes", "ppa:canonical-server/ua-client-daily"]
+    )
+    _lxc_exec(["apt-get", "update", "-q"])
+    _lxc_exec(["apt-get", "install", "-q", "-y", "ubuntu-advantage-tools"])
+
+
+@when("I run `ua status` as non-root")
+def step_impl(context):
+    process = _lxc_exec(["ua", "status"], capture_output=True, text=True)
+    context.output = process.stdout.strip()
+
+
+@then("I will see the following output")
+def step_impl(context):
+    assert_that(context.output, equal_to(context.text))
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -45,9 +45,9 @@ def step_impl(context, command):
     process = _lxc_exec(
         context, shlex.split(command), capture_output=True, text=True
     )
-    context.output = process.stdout.strip()
+    context.process = process
 
 
-@then("I will see the following output")
+@then("I will see the following on stdout")
 def step_impl(context):
-    assert_that(context.output, equal_to(context.text))
+    assert_that(context.process.stdout.strip(), equal_to(context.text))
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -40,9 +40,9 @@ def step_impl(context):
             "ppa:canonical-server/ua-client-daily",
         ],
     )
-    _lxc_exec(context, ["apt-get", "update", "-q"])
+    _lxc_exec(context, ["apt-get", "update", "-qq"])
     _lxc_exec(
-        context, ["apt-get", "install", "-q", "-y", "ubuntu-advantage-tools"]
+        context, ["apt-get", "install", "-qq", "-y", "ubuntu-advantage-tools"]
     )
 
 
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -1,3 +1,4 @@
+import datetime
 import subprocess
 from typing import List
 
@@ -6,7 +7,7 @@ from behave.runner import Context
 from hamcrest import assert_that, equal_to
 
 
-CONTAINER_NAME = "behave-test"
+CONTAINER_PREFIX = "behave-test-"
 
 
 def _lxc_exec(context: Context, cmd: List[str], *args, **kwargs):
@@ -17,7 +18,8 @@ def _lxc_exec(context: Context, cmd: List[str], *args, **kwargs):
 
 @given("a trusty lxd container")
 def step_impl(context):
-    context.container_name = CONTAINER_NAME
+    now = datetime.datetime.now()
+    context.container_name = CONTAINER_PREFIX + now.strftime("%s%f")
     subprocess.run(["lxc", "launch", "ubuntu:trusty", context.container_name])
 
 
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -57,3 +57,8 @@ def step_impl(context, command):
 @then("I will see the following on stdout")
 def step_impl(context):
     assert_that(context.process.stdout.strip(), equal_to(context.text))
+
+
+@then("I will see the following on stderr")
+def step_impl(context):
+    assert_that(context.process.stderr.strip(), equal_to(context.text))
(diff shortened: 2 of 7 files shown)


OddBloke:
> Needs some documentation or pointers around a) how to run b) how to debug c) how to write tests. Since you have used it before it would be nice to be able to read a primer so we can jump in and do more.

Yep, for sure. This branch was only intended as an example of the format, I'll improve those things before it lands.

> When I ran it I got some issues since DNS was not up yet "Cannot add PPA: 'ppa:canonical-server/ua-client-daily'." and "Temporary failure resolving 'security.ubuntu.com'" This is common in our testing when using LXD, so we need to add checks for that.

Yep, I wasn't running into these on my desktop but saw them on my laptop yesterday. I'll crib from elsewhere to do this. (My kingdom for ~a horse~`cloud-init status --wait`.)

> I assume we would run this as a part of CI rather than something we run night?

On my desktop, running the tests in this PR already takes 4-5 minutes. Travis is likely to be slower, and of course we will eventually be running many more tests than just these. So I think we may want to use [behave's tag expressions](https://behave.readthedocs.io/en/latest/tag_expressions.html) to run only a subset for _every_ pull request, and then do full runs against `master`. (Until we have enough tests that run speed is a problem, I don't think this needs to be a blocker to landing it, though.)

(There are some definite low-hanging fruit optimisations to be made to test speed, such as capturing an image after we've installed the version under test and using that in all features. I still don't think these will make it so fast that we can run all the tests for every PR once we have a decent number of tests, but it would increase the number of tests we could run per PR, so still worthwhile.)

This also only pulls out of the PPA at the moment, so we will need to modify it to use some other input (presumably a deb) before it can be used for gating.

> How would you structure the features to cover lxd and kvm(multipass) test cases? That seems like that would be a lot of duplication?

I have a couple of ideas here. We could use [scenario outlines](https://behave.readthedocs.io/en/latest/gherkin.html#scenario-outlines) to parameterise tests, but that's still quite verbose and would run both lxd and VM tests every run by default.

Alternatively, we could have three styles of given:
* "Given a trusty lxd container" (a test intended to only run on a lxd container),
* "Given a trusty VM" (only run on a VM), and
* "Given a trusty system" (run on both VMs and containers).

The tests would look at an environment variable to select which backend to use (i.e. `UACLIENT_BEHAVE_BACKEND=lxd` or `UACLIENT_BEHAVE_BACKEND=multipass`), and the container-specific or VM-specific "given" steps would just [skip execution](https://behave.readthedocs.io/en/latest/new_and_noteworthy_v1.2.5.html#exclude-feature-scenario-at-runtime) if we're running on a backend to which they don't apply.

OddBloke:
OK, I've just pushed up changes which add documentation to README.md, and code changes which wait for containers to boot before we try to use them. @powersj I feel like this is enough for people to start locally iterating on tests, so I propose that we aim to land this set of changes (after review, of course!) and I can file issues for the other items that you raised. Does that sounds reasonable?

powersj:
> I propose that we aim to land this set of changes (after review, of course!) and I can file issues for the other items that you raised. Does that sounds reasonable?

Agreed thanks for the response and for the updates. Looking forward to having this. This is very, very cool!

OddBloke:
That Travis "push" failure is weird, because that's not _this_ branch, that's a branch of the same name in the main repo.  So please disregard that, and review this; once we're at the point of landing, we can work out why that's happening.

features/unattached_commands.feature
```diff
@@ -0,0 +1,63 @@
+Feature: Command behaviour when unattached
```
powersj: Let me know if I am getting hung up on something, but the filename caused me to get stuck on how we grow these test cases.

So the cases we have will be the set of (platforms * releases) right? should our features files be broken out by release as well? (e.g. unattached_trusty.feature)?

Trying to think through the test cases we would eventually get:
unattached_{trusty, xenial, bionic}
attach_{trusty, xenial, bionic}
status_{trusty, xenial, bionic}
esm_infra_{trusty, xenial, bionic}
livepatch_{trusty, xenial, bionic}
...

And inside each put the platform specific code?

features/unattached_commands.feature
```diff
@@ -0,0 +1,63 @@
...
+
+    Scenario: Unattached detach in a trusty lxd container
+        Given a trusty lxd container
+          And ubuntu-advantage-client is installed
```
powersj: are we going to have any tests where we don't have it installed? Or should that be combined with the first? I can see being more verbose is helpful, but it seems this would be a lot more cut and paste.

If we want to have ua client not installed then maybe have one that removes it?

features/unattached_commands.feature
```diff
@@ -0,0 +1,63 @@
...
+        When I run `ua detach`
+        Then I will see the following on stderr:
```
powersj: I see the error? (negative)
I see the output? (positive)

features/unattached_commands.feature
```diff
@@ -0,0 +1,63 @@
+Feature: Command behaviour when unattached
+
+    Scenario: Unattached detach in a trusty lxd container
+        Given a trusty lxd container
+          And ubuntu-advantage-client is installed
+        When I run `ua detach`
```
powersj: should we be verbose of root versus non-root usage or only call out when a non-root user does something and always assume root is doing things?

e.g. root runs or sudo runs
e.g. non-root runs or user runs

features/unattached_commands.feature
```diff
@@ -0,0 +1,63 @@
...
+            """
+            This machine is not attached to a UA subscription.
+            See https://ubuntu.com/advantage
+            """
+
+    Scenario: Unattached refresh in a trusty lxd container
+        Given a trusty lxd container
+          And ubuntu-advantage-client is installed
+        When I run `ua refresh`
+        Then I will see the following on stderr:
```
powersj: why the weird indentation between given/and and when/then?

blackboxsw:
@OddBloke have you investigated or had experience with pytest-bdd and is there reason that behave is better for this application. I'm thinking mostly about parallel test execution that pytest-bdd seems to offer above behave which could be very helpful given that these integration tests are time-costly

README.md
```diff
@@ -32,6 +32,33 @@ The client also includes built-in dep8 tests. These are run as follows:
...
+To run the tests, you can use `tox`:
+
+```shell
+tox -e behave
+```
+
+or, if you just want to run a specific test:
+
+```shell
+tox -e behave features/unattached_commands.feature:55
```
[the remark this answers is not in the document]
OddBloke: Yep, will improve this.

features/steps/steps.py
```diff
@@ -0,0 +1,85 @@
...
+    def cleanup_container():
+        subprocess.run(["lxc", "stop", context.container_name])
+        subprocess.run(["lxc", "delete", context.container_name])
+
+    context.add_cleanup(cleanup_container)
+
+    _wait_for_boot(context)
+
+
+@given("ubuntu-advantage-client is installed")
```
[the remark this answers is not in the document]
OddBloke: Good catch!

features/unattached_commands.feature
OddBloke: We'll also want this to be a single step so that we can reuse an image with the right version inside to speed things up (i.e. #854).

Can we defer addressing this until we're also making that change?

features/unattached_commands.feature
OddBloke: `Given`/`When`/`Then` all align at the same column, `And` aligns its last character with the last character of the line it attaches to.

More than happy to standardise on something else, but this isn't _completely_ arbitrary. ^_^

features/unattached_commands.feature
OddBloke: Yes, I think that makes sense.

features/unattached_commands.feature
OddBloke: This may be another case (like VM vs container) where we don't want to capture the dimensions within the test code, but in the way we run the tests. So, similar to my suggestion at the bottom of https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/767#issuecomment-542834304, we could have:

* "Given an Ubuntu system" (run on all Ubuntu releases), and
* "Given a trusty system" (run only on trusty, skip on all others).

So I think we would then have `platforms × releases` common Given definitions (with each set also including an "all of them" option). So we would have:

* Given an Ubuntu system
* Given an Ubuntu lxd container
* Given an Ubuntu VM
* Given a trusty system
* Given a trusty lxd container
* Given a trusty VM
* ... these last three repeated for each Ubuntu release (but implemented with pattern matching for the steps, so we won't have to implement this N*M different times) ...

And for tests which we want to run on a subset of releases, we could do something like (completely untested!):

```gherkin
Scenario Outline: example scenario
  Given a <release_name> system
  When I check the release version
  Then I see that it's <version>

  Examples: Ubuntu Releases
    | release_name | version |
    | xenial       | 16.04   |
    | bionic       | 18.04   |
```

to minimise repetition.

features/unattached_commands.feature
powersj: agreed

features/unattached_commands.feature
powersj: ok, but if you are saying "see the following on stderr" does that mean you are actually checking stderr and not stdout? I guess I'm wondering do we need to make the distinction.

features/unattached_commands.feature
powersj: ah! sorry, I was looking at it from the end of the line... not sure why. Ignore me

OddBloke: No worries! It's more obvious what the pattern is when there's more than one "And", for sure.

features/unattached_commands.feature
OddBloke: Yep, each one checks its own thing:

OddBloke: (Previous comment was edited, in case you're only following along via email.)

I do think maintaining a distinction between the two is important, because it does make a substantial difference to the way that users interact with the command programmatically.

That said, we could also add a more general "Then I will see the following output", but the implementation could be a little tricky, because we would (probably) have to redirect stderr to stdout (easy) which would mean that the "I will see the following on (stdout|stderr)" would no longer function correctly. Not impossible to sort out, for sure, but a wrinkle nonetheless.

powersj: ok you have resolved my concerns :) having both makes sense

OddBloke: Thanks!

features/unattached_commands.feature
OddBloke: Thanks!

features/unattached_commands.feature
OddBloke: @powersj I don't think there's anything concrete to do until we're introducing our first new {platform, release}, does the above make enough sense to you for us to proceed?

powersj: agreed

features/steps/steps.py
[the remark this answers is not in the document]
OddBloke: This is now fixed.

README.md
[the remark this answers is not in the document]
OddBloke: This has been improved.

features/unattached_commands.feature
OddBloke: Done, as:

```gherkin
When I run `ua status` as non-root
```

OddBloke:
@powersj I believe I've addressed all your feedback.

@blackboxsw Did you want to review this further?

blackboxsw:
@OddBloke thanks for pulling this together. It looks really good.  Since lxc setup, apt updates and installing ua-tools is so costly I'd like us to look at defining the most costly setup items as fixtures where tests or "@given" clauses check before re-performing work where possible.


To that end; I've put together a patch set against your branch that does a couple things that could be helpful in reducing the overhead of the integration tests:

It creates an features/environment.py with an lxc_container fixture that we can use to decorate a Feature so that all scenarios in a given Feature can share that shared container for non-destructive tests.

I also adapted ubuntu-advantage-tools to only attempt to install uat from a ppa if the ppa is missing.

Here's a PR against your branch that we can discuss:  it takes the tox -e behave run from 5 minutes to  50 seconds on my laptop

https://github.com/OddBloke/ubuntu-advantage-client/pull/1

blackboxsw:
Thanks for the comments @OddBloke.   I've closed my separate PR/discussion strawman and will carry the oustanding conversation questions back over here:

 * How can we parameterize the behave cli to only drive platforms (lxc/kvm) or only test specific series, would we tag trusty vs xenial vs bionic tests?

* What are the ways behave gives us to share common setup in a series of Scenarios given that each Scenario performs cleaunup tasks?   (fixtures seems one way, but are there other patterns?)

OddBloke:
@blackboxsw 

> * How can we parameterize the behave cli to only drive platforms (lxc/kvm) or only test specific series, would we tag trusty vs xenial vs bionic tests?

Please take a look at https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/767#issuecomment-542834304 and https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/767#discussion_r335728666 and let me know if you still have questions/concerns after that.

> * What are the ways behave gives us to share common setup in a series of Scenarios given that each Scenario performs cleanup tasks? (fixtures seems one way, but are there other patterns?)

I'm not sure I understand the question, can you rephrase it?

blackboxsw:
> Please take a look at [#767 (comment)](https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/767#issuecomment-542834304) and [#767 (comment)](https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/767#discussion_r335728666) and let me know if you still have questions/concerns after that.

Thanks @OddBloke I missed the second round of comments. I guess I was hoping behave would allow us to avoid duplication in tests a bit and maybe something like your scenario outline looks promising.  My main concern was that there is potential to drive a subset "suite" of tests and minimally it looks like we could easily do that with good use of tags.
Thanks

blackboxsw:
"What are the ways behave gives us to share common setup in a series of Scenarios given that each Scenario performs cleanup tasks? (fixtures seems one way, but are there other patterns?)"

This question was posed in response to a comment you had on my separate spike:
<OddBloke> It's not entirely clear to me that this should be done via fixture (rather than via a single scenario, for example, which may make the behaviour more obvious when reading through tests). </OddBloke>

My understanding was that each Scenario performs it's own full setup and teardown when it enters and leaves scope. I was wondering if fixtures seem unclear as a solution to share common setup tasks, is there another mechanism you had in mind, or experience with that would make that common setup shareable across multiple scenarios? I'm maybe misinterpreting your suggestion above but I'm thinking you mean we need to have separate scenarios to more clearly define the "Givens", yet that makes me think there is no room for sharing common setup across scenarios. 

Instead of:
 ```
# Suggestion A
@fixture.some_costly_setup
Feature Attached command behavior
   Scenario: 1A- LXC container can't re-attach token
      Given something costly to setup
      And ua-t is already attached
      When I run attach
      Then I will see fire
   Scenario: 2A- LXC container status shows attached
      Given ua-t is already attached
      When I run status
      Then I will see attached status content

# Suggestion B
Feature Attached command behavior
   Scenario: 1B - LXC container can't re-attach token
      Given something costly to setup
      And ua-t is already attached
      When I run attach
      Then I will see fire
   Scenario: 2B - LXC container status shows attached
      Given Something costly to setup
      And  is already attached
      When I run status
      Then I will see attached status content


Without fixtures is there a way that 2B  can somehow leverage the shared 'Something costly to setup'?

OddBloke:
I guess my point was that it should be very obvious if something is running in a dirty container so that the potential sources for test failure are clear, and I'm not sure that fixtures meet that obviousness standard. So the key part of my (half considered) counter-proposal was to have a _single_ scenario for things that are shared. So in my "Suggestion C", there would be no scenario 2C, it would be tacked onto the end of scenario 1C, something like:

```
Feature Attached command behavior
  Scenario: 1C - LXC container can't re-attach token and status works
     Given something costly to setup
     And ua-t is already attached
     When I run attach
     Then I will see fire
     When I run status  # "2C"
     Then I will see attached status content
```

Note that this way of describing the tests doesn't change the fact that 1B depends on what happens in 1A, it just makes it explicit (in the way that a tag doesn't). If that's a problem, then that's evidence that we shouldn't be sharing containers between the two.

blackboxsw:
> I guess my point was that it should be very obvious if something is running in a dirty container so that the potential sources for test failure are clear, and I'm not sure that fixtures meet that obviousness standard. So the key part of my (half considered) counter-proposal was to have a _single_ scenario for things that are shared. So in my "Suggestion C", there would be no scenario 2C, it would be tacked onto the end of scenario 1C, something like:
> 
> ```
> Feature Attached command behavior
>   Scenario: 1C - LXC container can't re-attach token and status works
>      Given something costly to setup
>      And ua-t is already attached
>      When I run attach
>      Then I will see fire
>      When I run status  # "2C"
>      Then I will see attached status content
> ```
> 


Excellent thanks. This satisfies my concern about being able to chain scenarios (or steps a of scenario) that could depend on the same basic platform.

We can go ahead with this branch. Thanks for the discussions!

Issue #726, opened 2019-08-23
ux: `ua disable`


powersj:
Disable esm-infra:
```
$ sudo ua disable esm-infra
Updating package lists
```

Disable livepatch:
```
$ sudo ua disable livepatch
Removing canonical-livepatch snap
```

OddBloke:
@powersj In the UA UX doc, we seem to suggest that FIPS should be disable-able. Is this an intentional change from the current behaviour (of refusing to disable it at all)?

powersj:
FIPS work will come in the next release not this one, but yes we will work with the team to figure out what packages to remove.

OddBloke:
@powersj At the moment, we don't check people's contracts on disable (because it doesn't really matter if they're entitled to something they are turning off, I suppose). What's the thinking behind doing so?

powersj:
What does checking during disable get us? Is it complexity to the code since we only do it during refresh/enable?

powersj:
oh.. I just looked at the UX and realized it shows us refreshing *facepalm*

I'm fine with removing that from the UX as we don't do it today and I can't come up with a reason for why we would start.

powersj:
![facepalm](https://user-images.githubusercontent.com/6453401/65784059-1b903600-e106-11e9-938a-ca11dc2d8a5d.gif)

OddBloke:
Cool, thanks! Have updated the description to match.

OddBloke:
@powersj OK, I just want to double check what we're doing here. Currently the UX for repo entitlements is:

```
$ sudo ua disable esm-infra
ESM Infra disabled.
```

And we're looking to change it to:

```
$ sudo ua disable esm-infra
Updating package lists
```

That isn't a message that we currently emit, and actually isn't something we do on disable currently. (This is less noticeable than it will be on xenial, because the lists as seen by apt don't change when ESM gets enabled or disabled, so that the MOTD messaging can still function.) Should we start doing an `apt-get update` on disable, too? (Do we want a separate issue for that?)

(For livepatch, we're already emitting the required message, so it's just a case of dropping the `<foo> disabled.` line.)

powersj:
Yes we want to start doing the update and remove the message of disabled
for ESM as well as livepatch.

OddBloke:
This was fixed by #769 and #770.

PR #770, opened 2019-09-27
repo: update package lists after disable

OddBloke:
This is part of addressing #726, along with #769.


uaclient/entitlements/repo.py
```diff
@@ -297,3 +297,7 @@ def remove_apt_config(self):
                 )
             elif os.path.exists(repo_pref_file):
                 os.unlink(repo_pref_file)
+        print("Updating package lists")
```
blackboxsw: Let's pull this text up into a uaclient/status.py message in the event that we internationalize messaging at some point.
We also have this same message emitted on line 254  in setup_apt_config so we can use that status message there too.

PR #779, opened 2019-10-01
tools/constraints-*.txt: pin version of attrs

OddBloke:
The latest version of attrs caused bionic test failures, so ensure that
we're using the version that's in our target Ubuntu release.


tools/constraints-disco.txt
```diff
@@ -1,3 +1,4 @@
+attrs==18.2
```
blackboxsw: Let's drop all the other pinnings on non-bionic as they aren't causing us trouble. Ditto n eoan/disco/xenial.

OddBloke: These versions match what is in the target Ubuntu release, so mean that our Travis test runs will more closely match what our package build test runs will see. Arguably, we should be pinning _more_ stuff, but I don't think that's worth the effort until we see an issue.

(I 100% agree that if we were seeing these issues against the latest versions of everything, then we should not just pin and move on; that isn't the case here though.)

PR #780, opened 2019-10-01
Update CLI help and manpage for new status output

### LINKED ISSUE #722: Update UA help output and manpage for status
Ensure that --help prints out info about each header and possible values for status as well as in the manpage

OddBloke:
Fixes #722

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -21,6 +21,7 @@ import logging
 import os
 import pathlib
 import sys
+import textwrap
 
 from uaclient import config
 from uaclient import contract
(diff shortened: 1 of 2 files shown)


uaclient/cli.py
```diff
@@ -145,6 +146,31 @@ def status_parser(parser):
...
+        system.
+
+        Each service status line has four columns:
+
+        * SERVICE: name of the service
+        * ENTITLED: whether the contract to which this machine is attached
+          entitles use of this service. Possible values are: yes or no
+        * STATUS: whether the service is enabled on this machine.  Possible
+          values are: enabled, disabled, or — (if inapplicable)
+        * DESCRIPTION: a brief description of the service
```
blackboxsw: Should we clarify that this set of columns applies to an attached machine (and also add a section for the unattached machine which surfaces the "AVAILABLE" column indicating that a particular service is potentially available on the current series.


Since SERVICE and DESCRIPTION are the same on unattached, maybe we just include a line about AVAILABLE in this existing table and note that AVAILABLE is unattached only and STATUS/ENTITLED are on attached machines.

ubuntu-advantage.1
```diff
@@ -58,18 +58,24 @@ support contract. When attached, the report includes the specific
...
+Each service status line has four columns:
 
-.BR "name" ":"
+.BR "SERVICE" ":"
 name of the service
 
-.BR "availability" ":"
-whether the contract entitles use of this service. Possible values are:
-none or entitled
+.BR "ENTITLED" ":"
```
blackboxsw: Same thought here, we need to capture AVAILABLE somehow for unattached machine.

Issue #783, opened 2019-10-01
Non-root user before attach and after detach sees invalid empty status


blackboxsw:
In a freshly booted instance:

```
$ ua status

SERVICE      AVAILABLE  DESCRIPTION

This machine is not attached to a UA subscription.
See https://ubuntu.com/advantage
```

In a detached instance:

```
ubuntu@dev-t:~/ubuntu-advantage-client$ sudo ua detach
Are you sure? (y/N) y
This machine is now detached
ubuntu@dev-t:~/ubuntu-advantage-client$ ua status

SERVICE      AVAILABLE  DESCRIPTION

This machine is not attached to a UA subscription.
See https://ubuntu.com/advantage
```

PR #785, opened 2019-10-01
status: non-root user can query contract-server/v1/resources to see availability

blackboxsw:
Fixes: #783

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -308,7 +308,7 @@ class UAConfig:
         if os.getuid() != 0:
             response = cast("Dict[str, Any]", self.read_cache("status-cache"))
             if not response:
-                response = DEFAULT_STATUS
+                response = self._unattached_status()
         elif not self.is_attached:
             response = self._unattached_status()
         else:
--- a/uaclient/tests/test_config.py
+++ b/uaclient/tests/test_config.py
@@ -439,11 +439,14 @@ class TestStatus:
         # cfg.status() idempotent
         assert expected == cfg.status()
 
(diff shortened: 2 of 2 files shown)


OddBloke:
I'm not convinced this is the right fix; shouldn't we instead be working out why deleting the cache doesn't result in the same status as on a never-attached machine?

blackboxsw:
> I'm not convinced this is the right fix; shouldn't we instead be working out why deleting the cache doesn't result in the same status as on a never-attached machine?

Currently the never-attached machine is also broken for non-root as well (because we don't allow non-root to talk to v1/resources). So the fresh install looks like this:
ubuntu@dev-t:~$ sudo dpkg -i ubuntu-advantage-tools_19.6_amd64.deb 
Selecting previously unselected package ubuntu-advantage-tools.
(Reading database ... 56545 files and directories currently installed.)
Preparing to unpack ubuntu-advantage-tools_19.6_amd64.deb ...
Unpacking ubuntu-advantage-tools (19.6) ...
Setting up ubuntu-advantage-tools (19.6) ...
Processing triggers for man-db (2.6.7.1-1ubuntu1) ...
ubuntu@dev-t:~$ ua status

SERVICE      AVAILABLE  DESCRIPTION

See https://ubuntu.com/advantage
ubuntu@dev-t:~$ 


So, we need to figure out a preferred way to fix this, because DEFAULT_STATUS dict lacks any information about available resources (as root user gets to see that from v1/resources)/

OddBloke:
> because we don't allow non-root to talk to v1/resources

What's the reference for that?

PR #788, opened 2019-10-02
Add package builds to CI

### LINKED ISSUE #245: Add package builds to CI
I broke package builds on at least trusty in #233, which would have been good to catch before it was merged.

To that end, we should consider adding basic package builds to our Travis runs.

OddBloke:
Fixes #245

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 dist: xenial
-matrix:
+
 install:
   # Required so `git describe` will definitely find a tag; see
   # https://github.com/travis-ci/travis-ci/issues/7422
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ script:
 matrix:
   fast_finish: true
   include:
-    - install:
+    - env:
+        PACKAGE_BUILD_SERIES=xenial
+      install:
         - make travis-deb-install
       script:
         - make travis-deb-script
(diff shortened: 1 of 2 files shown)


OddBloke:
(I cancelled the latest round of builds because Travis is having a slow one, and we're going to land other branches before this one so it's going to get rebased at least one more time.)

blackboxsw:
This content looks great. The one thing I'm uncertain about is looking over the travis runs, the ~4 min cost per package build job. I don't know if the additional cpu load of this ci run on all distros is going to adversely affect our CI throughput for this project or not. Given that a lot of today was spent waiting on travis CI jobs to complete, I'm not sure if our account/project is being artificially throttled or rate-limited due to how many jobs we spawn per commit.

OddBloke:
I certainly share the concern about Travis build times. Travis certainly does throttle jobs, [this (4 year old) tweet](https://twitter.com/travisci/status/651856122559774722) suggests it's 5 jobs for open source, but those are split across the whole _organisation_, so any other CanonicalLtd projects are also sharing that pool.

Ideally, we shouldn't ever end up with 12 in-flight PRs, so I think this week so far _is_ anomalous in terms of how much Travis waiting we'll be doing. Obviously there were extenuating circumstances in this case, but if we're on top of reviews the way we normally should be (and have been!), then it should be less painful.

I would suggest landing this, and if we find that it's causing us substantial problems then we can look at backing it out and/or modifying what it does. Does that sounds reasonable?

Issue #776, opened 2019-09-30
root owned ~/.gnupg


panlinux:
Installing `ubuntu-advantage-tools` from scratch creates a root-owned `~/.gnupg` directory, if one doesn't exist already.
This is problematic when using sudo, for example:
```
ubuntu@trusty-ua:~$ sudo rm -rf .gnupg/

ubuntu@trusty-ua:~$ sudo apt install ubuntu-advantage-tools 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages were automatically installed and are no longer required:
  libfreetype6 os-prober
Use 'apt-get autoremove' to remove them.
The following NEW packages will be installed:
  ubuntu-advantage-tools
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/59.3 kB of archives.
After this operation, 286 kB of additional disk space will be used.
Selecting previously unselected package ubuntu-advantage-tools.
(Reading database ... 31959 files and directories currently installed.)
Preparing to unpack .../ubuntu-advantage-tools_19.6~603~git.6f9a5ed~ubuntu14.04.1_amd64.deb ...
Unpacking ubuntu-advantage-tools (19.6~603~git.6f9a5ed~ubuntu14.04.1) ...
Processing triggers for man-db (2.6.7.1-1ubuntu1) ...
Setting up ubuntu-advantage-tools (19.6~603~git.6f9a5ed~ubuntu14.04.1) ...
gpg: directory `/home/ubuntu/.gnupg' created
gpg: new configuration file `/home/ubuntu/.gnupg/gpg.conf' created
gpg: WARNING: options in `/home/ubuntu/.gnupg/gpg.conf' are not yet active during this run
gpg: keyring `/home/ubuntu/.gnupg/secring.gpg' created

ubuntu@trusty-ua:~$ l .gnupg
ls: cannot open directory .gnupg: Permission denied

ubuntu@trusty-ua:~$ ls -lad .gnupg
drwx------ 2 root root 4 Sep 30 19:48 .gnupg
```

panlinux:
We also need to take care of the gpg call in gpg.py, which is doing the same thing.

PR #792, opened 2019-10-02
gpg: do not create .gnupg dir during key extraction from shared keyring

blackboxsw:
Both debian/postinst and uaclient/gpg call to gpg on the command line in
order to extract a known GPG key for a service we are enabling.

Update both of these callsites to use the same gpg parameters.
Also, add --no-options parameter which prevents custom gpg config
options from being sourced as well as preventing the creation of
.gnupg in the current user's homedir.

Fixes: #776

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -14,8 +14,9 @@ export_gpg_key_from_shared_keyring() {
     KEY_ID=$1
     if [ ! -z "$KEY_ID" ]; then
         rm -f $ESM_APT_GPG_KEY
-        gpg --no-auto-check-trustdb --options /dev/null --no-default-keyring \
-            --keyring $UA_KEYRING_FILE --export $KEY_ID > $ESM_APT_GPG_KEY
+        gpg --no-auto-check-trustdb --no-options --no-default-keyring \
+            --keyring $UA_KEYRING_FILE --export $KEY_ID \
+            --output $ESM_APT_GPG_KEY --yes
     fi
 }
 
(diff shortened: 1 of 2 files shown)


debian/postinst
```diff
@@ -14,8 +14,9 @@ export_gpg_key_from_shared_keyring() {
     KEY_ID=$1
     if [ ! -z "$KEY_ID" ]; then
         rm -f $ESM_APT_GPG_KEY
-        gpg --no-auto-check-trustdb --options /dev/null --no-default-keyring \
-            --keyring $UA_KEYRING_FILE --export $KEY_ID > $ESM_APT_GPG_KEY
+        gpg --no-auto-check-trustdb --no-options --no-default-keyring \
+            --keyring $UA_KEYRING_FILE --export $KEY_ID \
+	    --output $ESM_APT_GPG_KEY --yes
```
OddBloke: Indentation here looks off.

OddBloke: Which part of this change is it that stops the creation of the .gnupg dir?

blackboxsw: @OddBloke fixed indentation. and --no-options param as mentioned in the commit msg is the one that stops the ./gnupg dir creation :)

OddBloke: Aha, right, I should have read the _entire_ manpage entry:

> Shortcut for --options /dev/null. This option is detected before an attempt to open an option file.  Using this option will also prevent the creation of a ‘~/.gnupg’ homedir.

I read the first sentence and was like "there's no change here", oops!

PR #803, opened 2019-10-08
Bug/drop gpg cmd dependency

blackboxsw:
* tools/refresh-keyrings.sh: Update dev script to emit separate key files of the format ua-<key_id>.gpg
* keyrings/*: remove consolidated ubuntu-advantage-keyring.gpg, add separate key files for each ua-<key_id>.gpg downloaded by refresh-keyrings.sh
* uaclient/gpg.gpg: no longer use gpg CLI and just copy known ua-<key_id>.gpg to <destination_keyfile_name>
* debian/postinst: no longer use gpg CLI and just copy the $UA_KEYRING_DIR/ua-$KEY_ID.gpg $ESM_APT_GPG_KEY


tools/refresh-keyrings.sh
```diff
@@ -21,20 +21,24 @@ fi
...
-        mv "$KEYRING_FILE" "$KEYRING_FILE.old"
-    fi
-    gpg \
-        --keyring "$KEYRING_FILE" \
-        --no-default-keyring \
-        --keyserver keyserver.ubuntu.com \
-        --recv-keys $KEYS
+
+    for key in $KEYS; do
+	keyring_file=$KEYRING_DIR/ua-$key.gpg
```
OddBloke: Weird indentation here.

tools/refresh-keyrings.sh
```diff
@@ -21,20 +21,24 @@ fi
...
+        if [ -e "$keyring_file" ]; then
+            mv "$keyring_file" "$keyring_file.old"
+        fi
+        gpg \
```
OddBloke: Should we include `--no-options` in this invocation?

uaclient/gpg.py
```diff
@@ -2,54 +2,27 @@
...
-            "GPG key '{}' not found in {}".format(key_id, source_keyring_file)
-        )
-    if not os.path.exists(destination_keyfile):
-        msg = "Unexpected error exporting GPG key '{}' from {}".format(
-            key_id, source_keyring_file
-        )
-        with util.disable_log_to_console():
-            logging.error(msg + " Error: {}".format(err))
-        raise exceptions.UserFacingError(msg)
+    os.rename(source_keyring_file, destination_keyfile)
```
OddBloke: This will remove `source_keyring_file`, won't it?

uaclient/tests/test_gpg.py
```diff
@@ -24,74 +24,39 @@ def home_dir(tmpdir):
 
 class TestExportGPGKeyFromKeyring:
```
OddBloke: This class name needs updating.

uaclient/tests/test_gpg.py
```diff
@@ -24,74 +24,39 @@ def home_dir(tmpdir):
...
         """Only a single key is exported from a multi-key source keyring."""
-        source_keyring = tmpdir.join("keyring").strpath
+        source_key1 = tmpdir.join("ua-{}.gpg".format(data.GPG_KEY1_ID)).strpath
+        source_key2 = tmpdir.join("ua-{}.gpg".format(data.GPG_KEY2_ID)).strpath
         destination_keyfile = tmpdir.join("destination_key").strpath
         # Create keyring with both ESM and CC-EAL2 keys
-        with open(source_keyring, "wb") as stream:
-            stream.write(data.GPG_KEY1 + data.GPG_KEY2)
-        gpg.export_gpg_key_from_keyring(
+        with open(source_key1, "wb") as stream:
```
OddBloke: A nit, but if you don't call `strpath` on the `tmpdir.join` return, I think you could do `source_key1.write(data.GPG_KEY1, "wb")` and save a bit of vertical space.

debian/postinst
```diff
@@ -11,16 +11,13 @@ ESM_APT_SOURCE_FILE_OLD="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
 ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list"
 ESM_APT_PREF_FILE_OLD="/etc/apt/preferences.d/ubuntu-esm-trusty"
 ESM_APT_PREF_FILE="/etc/apt/preferences.d/ubuntu-esm-infra-trusty"
-UA_KEYRING_FILE="/usr/share/keyrings/ubuntu-advantage-keyring.gpg"
+UA_KEYRING_DIR="/usr/share/keyrings/"
```
OddBloke: I'm not 100% sure that `/usr/share/keyrings` is the right place for these files any longer, as we're creating a bunch of them and the names are inscrutable for users. I don't know where _is_ the right place though, it just feels a little off.

Perhaps I'm just being too cautious though, and it's really fine! What do you think? Maybe @panlinux has an opinion?

tools/refresh-keyrings.sh
OddBloke: And we should probably be quoting this, in the unlikely event that `KEYRING_DIR` or `key` ever end up with spaces in them.

tools/refresh-keyrings.sh
```diff
@@ -46,10 +50,6 @@ ESM_KEY_ID_BIONIC="2926E7D347A1955504000A983121D2531EF59819"
 FIPS_KEY_ID_XENIAL="E23341B2A1467EDBF07057D6C1997C40EDE22758"
 
 
-UA_KEYRING="ubuntu-advantage-keyring.gpg"
-
-generate_keyring $TARGET_DIR/$UA_KEYRING $EAL_KEY_ID_XENIAL $ESM_KEY_ID_PRECISE $ESM_KEY_ID_TRUSTY $ESM_KEY_ID_XENIAL $ESM_KEY_ID_BIONIC $FIPS_KEY_ID_XENIAL
+generate_keyrings $TARGET_DIR $EAL_KEY_ID_XENIAL $ESM_KEY_ID_PRECISE $ESM_KEY_ID_TRUSTY $ESM_KEY_ID_XENIAL $ESM_KEY_ID_BIONIC $FIPS_KEY_ID_XENIAL
 sed -i "s/ESM_KEY_ID_TRUSTY=.*/ESM_KEY_ID_TRUSTY=\"${ESM_KEY_ID_TRUSTY}\"/" \
-    -i "s/ESM_KEY_ID_XENIAL=.*/ESM_KEY_ID_XENIAL=\"${ESM_KEY_ID_XENIAL}\"/" \
```
OddBloke: Why are these lines being dropped?

tools/refresh-keyrings.sh
blackboxsw: +1

tools/refresh-keyrings.sh
blackboxsw: sure let's

tools/refresh-keyrings.sh
blackboxsw: I'll add a comment in the code. These sed lines are dropped because we aren't pre-seeding any ESM keys in debian/postinst (because we only manipulate esm keys on trusty and pre-install keys for esm package visibility). All other keys are added/removed only and runtime during ua enable (or auto-enablement on attach)

uaclient/gpg.py
blackboxsw: bah you are absolutely right. Good review/catch.

debian/postinst
blackboxsw: Agreed, I'm not certain either, that's why I threw this branch together to get your and @panlinux thoughts on this.

uaclient/tests/test_gpg.py
blackboxsw: excellent suggestion.

tools/refresh-keyrings.sh
```diff
@@ -21,20 +21,25 @@ fi
...
+
+    for key in $KEYS; do
+	keyring_file="$KEYRING_DIR/ua-$key.gpg"
+        if [ -e "$keyring_file" ]; then
+            mv "$keyring_file" "$keyring_file.old"
+        fi
+        gpg \
+            --keyring "$keyring_file" \
+            --no-default-keyring \
+	    --no-options \
```
OddBloke: Indentation.

tools/refresh-keyrings.sh
```diff
@@ -21,20 +21,25 @@ fi
...
-        mv "$KEYRING_FILE" "$KEYRING_FILE.old"
-    fi
-    gpg \
-        --keyring "$KEYRING_FILE" \
-        --no-default-keyring \
-        --keyserver keyserver.ubuntu.com \
-        --recv-keys $KEYS
+
+    for key in $KEYS; do
+	keyring_file="$KEYRING_DIR/ua-$key.gpg"
```
OddBloke: Still indentation.

uaclient/gpg.py
```diff
@@ -1,55 +1,29 @@
...
-            "GPG key '{}' not found in {}".format(key_id, source_keyring_file)
-        )
-    if not os.path.exists(destination_keyfile):
-        msg = "Unexpected error exporting GPG key '{}' from {}".format(
-            key_id, source_keyring_file
-        )
-        with util.disable_log_to_console():
-            logging.error(msg + " Error: {}".format(err))
-        raise exceptions.UserFacingError(msg)
+    shutil.copy(source_keyring_file, destination_keyfile)
```
OddBloke: `shutil.copy` doesn't retain file metadata. What will the permissions/ownership of the installed file be?

debian/postinst
panlinux: I had a great response, then pressed ctrl-w by mistake.

Summarizing it, we used to have this:
```
keyrings/ubuntu-esm-v2-keyring.gpg
keyrings/ubuntu-fips-keyring.gpg
keyrings/ubuntu-esm-keyring.gpg
```
And now we have `keyrings/ua-$gibberish.gpg`. Why can't we have the names back, even if it's per ubuntu release, aka, a longer list?

tools/refresh-keyrings.sh
blackboxsw: man, /me adds a vimrc script for validating that.

debian/postinst
blackboxsw: Per a side discussion on this, we decided we can drop the separate gpg files, so we will keep a single ringfile and avoid the potentially confusing ua-<key_id>.gpg files that mean nothing to an end user.

uaclient/gpg.py
blackboxsw: +1, we are no longer using shutil since we are relying on apt-key now.

uaclient/util.py
```diff
@@ -416,16 +418,27 @@ def _subp(
     @raises subprocess.TimeoutError when timeout specified and the command
         exceeds that number of seconds.
     """
+    if data is None:
+        # using devnull assures any reads get null, rather
+        # than possibly waiting on input.
+        stdin = open(os.devnull)
```
panlinux: Can https://docs.python.org/3.4/library/subprocess.html#subprocess.DEVNULL be used here?

uaclient/tests/test_gpg.py
```diff
@@ -70,6 +70,10 @@ def test_user_facing_error_on_invalid_keyring_packet(
 
     def test_export_single_key_from_keyring(self, home_dir, tmpdir):
         """Only a single key is exported from a multi-key source keyring."""
+        if os.getuid() != 0:
+            raise pytest.skip(
+                "Skipping export key test as non-root can't use apt-key"
```
panlinux: Bummer, and it fails even with fakeroot:
```
ubuntu@trusty-ua:~$ rm andreas.gpg 
ubuntu@trusty-ua:~$ apt-key  export 437D05B5 | fakeroot apt-key --keyring andreas.gpg add -
OK
ubuntu@trusty-ua:~$ l andreas.gpg 
-rw-r--r-- 1 ubuntu ubuntu 0 Oct  9 17:23 andreas.gpg
ubuntu@trusty-ua:~$ 
```

panlinux: Oh, fakeroot actually works, I just forgot to use an absolute path for the output keyring file:
```
ubuntu@trusty-ua:~$ rm andreas.gpg*
ubuntu@trusty-ua:~$ apt-key  export 437D05B5 | fakeroot apt-key --keyring ./andreas.gpg add -
OK
ubuntu@trusty-ua:~$ ll andreas.gpg
-rw-r--r-- 1 ubuntu ubuntu 3502 Oct  9 18:39 andreas.gpg
ubuntu@trusty-ua:~$ apt-key --keyring ./andreas.gpg list
./andreas.gpg
-------------
pub   1024D/437D05B5 2004-09-12
uid                  Ubuntu Archive Automatic Signing Key <ftpmaster@ubuntu.com>
sub   2048g/79164387 2004-09-12

ubuntu@trusty-ua:~$ 
```

uaclient/gpg.py
```diff
@@ -46,6 +41,16 @@ def export_gpg_key_from_keyring(
         raise exceptions.UserFacingError(
```
panlinux: Not introduced in this branch, but the line just above checks `err` for the english text `nothing exported`, and that is locale dependent:
```
root@trusty-ua:~# apt-key export 123123
gpg: AVISO: nada exportado
```

uaclient/util.py
```diff
@@ -463,6 +477,8 @@ def subp(
     """Run a command and return a tuple of decoded stdout, stderr.
 
      @param subp: A list of arguments to feed to subprocess.Popen
+     @param data: A str or bytes to pass as stdin to the command.
+     @param capture: Boolean set True to log the command and response.
```
OddBloke: Let's maintain argument order in the docstring (so move this down one line).

uaclient/gpg.py
```diff
@@ -19,21 +19,16 @@ def export_gpg_key_from_keyring(
...
-        "--no-options",
-        "--no-default-keyring",
+        "apt-key",
         "--keyring",
         source_keyring_file,
-        "--export",
+        "export",
         key_id,
     ]
+    add_ring_cmd = ["apt-key", "--keyring", destination_keyfile, "add", "-"]
```
panlinux: I just tested something and noticed that `apt-key add` will not overwrite a keyring file, but add the key to it if it doesn't exist yet. You won't get duplicates, but if you call it with key foo once, and key bar after, the target file will contain both keys. Should we guard against this, and is it safe to always remove the destination file before calling this command?

uaclient/util.py
```diff
@@ -416,16 +418,27 @@ def _subp(
     @raises subprocess.TimeoutError when timeout specified and the command
         exceeds that number of seconds.
     """
+    if data is None:
+        # using devnull assures any reads get null, rather
+        # than possibly waiting on input.
+        stdin = open(os.devnull)
+    else:
+        data = data if isinstance(data, bytes) else data.encode("utf-8")
+        stdin = subprocess.PIPE  # type: ignore
```
OddBloke: This may go away/change with `subprocess.DEVNULL`, but could we give the initial definition the right Union type definition instead of ignoring the error?

uaclient/util.py
```diff
@@ -416,16 +418,27 @@ def _subp(
     @raises subprocess.TimeoutError when timeout specified and the command
         exceeds that number of seconds.
     """
+    if data is None:
+        # using devnull assures any reads get null, rather
+        # than possibly waiting on input.
+        stdin = open(os.devnull)
+    else:
+        data = data if isinstance(data, bytes) else data.encode("utf-8")
```
OddBloke: We shouldn't need to do this; could we just accept `bytes` and place the onus on callers to encode their data appropriately?

uaclient/util.py
blackboxsw: dropped on revert of apt-key cmd calls. point taken.

debian/postinst
```diff
@@ -18,9 +18,8 @@ export_gpg_key_from_shared_keyring() {
     KEY_ID=$1
     if [ ! -z "$KEY_ID" ]; then
         rm -f $ESM_APT_GPG_KEY
-        gpg --no-auto-check-trustdb --no-options --no-default-keyring \
-            --keyring $UA_KEYRING_FILE --export $KEY_ID \
-            --yes > $ESM_APT_GPG_KEY
+        apt-key --keyring $UA_KEYRING_FILE \
```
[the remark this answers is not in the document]
blackboxsw: +1 for future. dropped from this branch

uaclient/util.py
blackboxsw: Dropped from this branch. but  next time grow a param that requires stdin for subp we can take that approach

uaclient/util.py
blackboxsw: good pt. irrelevant now with the content revert

tools/refresh-keyrings.sh
```diff
@@ -45,11 +29,40 @@ ESM_KEY_ID_BIONIC="2926E7D347A1955504000A983121D2531EF59819"
...
+    for key in $KEYS; do
+        case $key in
+            $EAL_KEY_ID_XENIAL)
+                service_name="Common-Criteria-EAL2";;
+            $ESM_KEY_ID_TRUSTY|$ESM_KEY_ID_XENIAL|$ESM_KEY_ID_BIONIC)
+                service_name="ESM-For-Infrastructure";;
+            $FIPS_KEY_ID_XENIAL)
+                service_name="FIPS";;
+        esac
+        keyring_file="$KEYRING_DIR/Ubuntu-Advantage-$service_name-$key.gpg"
```
OddBloke: Title-Case is inconsistent with all the lower-case keyrings already in `/usr/share/keyrings`

uaclient/gpg.py
OddBloke: We've changed the code to no longer do this, so I'm going to mark this resolved.

uaclient/util.py
blackboxsw: yep and it should. Disregarded this comment because we dropped that functionality from this branch. But will remember that in the future.

tools/refresh-keyrings.sh
blackboxsw: used lowercase-human-readable  now.

blackboxsw:
per hangouts and IRC discussion we are now naming the ring files ubuntu-advantage-<service>-<series>.gpg and we will only ship ubuntu-advantage-esm-infra-trusty.gpg key

tools/refresh-keyrings.sh
```diff
@@ -45,11 +30,47 @@ ESM_KEY_ID_BIONIC="2926E7D347A1955504000A983121D2531EF59819"
...
 FIPS_KEY_ID_XENIAL="E23341B2A1467EDBF07057D6C1997C40EDE22758"
 
+generate_keyrings() {
+    KEYRING_DIR="$1"
+    shift
+    KEYS="$@"
+
+    # Intentionally unquoted so GPG gets the keys as separate arguments
+
+    for key in $KEYS; do
```
OddBloke: *This is more than a nit, in that I would like to see it addressed, but I'm happy for it to happen in a follow-up PR as this is tooling rather than production code.*

I think using a for loop here doesn't make as much sense now that we need to operate on each key separately. It also serves to have the key ID and the actual name we use for that keyring several lines apart, making it harder to easily review what's going on. Would a series of calls that look something like:

```
generate_keyring fips-xenial E23341B2A1467EDBF07057D6C1997C40EDE22758
```

allow us to capture both pieces of information about a key in a single line? (I think this would also allow us to drop the variable definitions above, as the keyring name is more useful as an indicator of what the key ID is for than the variable name is.)

What do you think?

uaclient/gpg.py
```diff
@@ -1,55 +1,24 @@
...
 
 from uaclient import exceptions
-from uaclient import util
 
 
-def export_gpg_key_from_keyring(
-    key_id: str, source_keyring_file: str, destination_keyfile: str
-) -> None:
-    """Export a specific key from source_keyring_file into destination_keyfile
+def export_gpg_key(source_keyfile: str, destination_keyfile: str) -> None:
```
OddBloke: Something of a nit, but I wonder if this would be better off just taking a single `str` argument and handling constructing the source and destination paths itself? It would push a little complexity down the stack, making the already-more-complex entitlement class a little simpler.

uaclient/gpg.py
```diff
@@ -1,55 +1,24 @@
...
-        "--export",
-        key_id,
-    ]
-    logging.debug("Exporting GPG key %s from %s", key_id, source_keyring_file)
-    try:
-        out, err = util.subp(export_cmd)
-    except util.ProcessExecutionError as exc:
-        with util.disable_log_to_console():
-            logging.error(str(exc))
+    logging.debug("Exporting GPG key %s", source_keyfile)
```
OddBloke: This log message, and the function name, aren't really accurate any longer. We aren't exporting a key, we're installing a keyring.

PR #813, opened 2019-10-09
pkg: precise to trusty upgrade move valid apt creds to trusty apt source

blackboxsw:
Precise legacy creds are entitled to esm on trusty. do-release-upgrade
will have re-written existing apt sources list files replacing 'precise'
pockets with 'trusty'.

On precise machines which had ESM enabled, the debian/postinst will now
migrate valid apt sources filename during do-release-upgrade from:
/etc/apt/sources.list.d/ubuntu-esm-precise.list
to:
/etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list

This way precise credentials are retained and uaclient runtime will
update or remove /etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list if
"ua disable or enable esm" operations are performed on trusty.

Fix: #693

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -5,34 +5,40 @@ set -e
 . /etc/os-release  # For VERSION_ID
 
 APT_TRUSTED_KEY_DIR="/etc/apt/trusted.gpg.d"
-ESM_INFRA_KEY="ubuntu-advantage-esm-infra-trusty.gpg"
-ESM_APT_SOURCE_FILE_OLD="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
-ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list"
-ESM_APT_PREF_FILE_OLD="/etc/apt/preferences.d/ubuntu-esm-trusty"
-ESM_APT_PREF_FILE="/etc/apt/preferences.d/ubuntu-esm-infra-trusty"
 UA_KEYRING_DIR="/usr/share/keyrings/"
+
+ESM_INFRA_KEY_TRUSTY="ubuntu-advantage-esm-infra-trusty.gpg"
+
+APT_SRC_DIR="/etc/apt/sources.list.d"
+ESM_APT_SOURCE_FILE_PRECISE="$APT_SRC_DIR/ubuntu-esm-precise.list"
+ESM_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-trusty.list"
+ESM_INFRA_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-infra-trusty.list"
+
+ESM_APT_PREF_FILE_TRUSTY="/etc/apt/preferences.d/ubuntu-esm-trusty"
+ESM_INFRA_APT_PREF_FILE_TRUSTY="/etc/apt/preferences.d/ubuntu-esm-infra-trusty"
 
 
 unconfigure_esm() {
     rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
-    rm -f $APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY $ESM_APT_SOURCE_FILE
-    rm -f $ESM_APT_PREF_FILE_OLD $ESM_APT_PREF_FILE
+    rm -f $APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY_TRUSTY
+    rm -f $ESM_INFRA_APT_SOURCE_FILE_TRUSTY
+    rm -f $ESM_APT_PREF_FILE_TRUSTY $ESM_INFRA_APT_PREF_FILE_TRUSTY
 }
 
 configure_esm() {
     rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
-    if [ ! -f "$APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY" ]; then
-        cp $UA_KEYRING_DIR/$ESM_INFRA_KEY $APT_TRUSTED_KEY_DIR
+    if [ ! -f "$APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY_TRUSTY" ]; then
+        cp $UA_KEYRING_DIR/$ESM_INFRA_KEY_TRUSTY $APT_TRUSTED_KEY_DIR
     fi
 
-    if [ -e "$ESM_APT_SOURCE_FILE_OLD" ]; then
-        mv $ESM_APT_SOURCE_FILE_OLD $ESM_APT_SOURCE_FILE
+    if [ -e "$ESM_APT_SOURCE_FILE_TRUSTY" ]; then
+        mv $ESM_APT_SOURCE_FILE_TRUSTY $ESM_INFRA_APT_SOURCE_FILE_TRUSTY
     fi
-    if [ -e "$ESM_APT_PREF_FILE_OLD" ]; then
-        mv $ESM_APT_PREF_FILE_OLD $ESM_APT_PREF_FILE
+    if [ -e "$ESM_APT_PREF_FILE_TRUSTY" ]; then
+        mv $ESM_APT_PREF_FILE_TRUSTY $ESM_INFRA_APT_PREF_FILE_TRUSTY
     fi
-    if [ ! -e "$ESM_APT_SOURCE_FILE" ]; then
-        cat > $ESM_APT_SOURCE_FILE <<EOF
+    if [ ! -e "$ESM_INFRA_APT_SOURCE_FILE_TRUSTY" ]; then
+        cat > $ESM_INFRA_APT_SOURCE_FILE_TRUSTY <<EOF
 # Written by ubuntu-advantage-tools
 deb https://esm.ubuntu.com/ubuntu trusty-infra-security main
 # deb-src https://esm.ubuntu.com/ubuntu trusty-infra-security main
@@ -41,7 +47,7 @@ deb https://esm.ubuntu.com/ubuntu trusty-infra-updates main
 # deb-src https://esm.ubuntu.com/ubuntu trusty-infra-updates main
 EOF
         # Automatically disable esm sources via apt preferences until enabled
-        cat > $ESM_APT_PREF_FILE <<EOF
+        cat > $ESM_INFRA_APT_PREF_FILE_TRUSTY <<EOF
 # Written by ubuntu-advantage-tools
 Package: *
 Pin: release o=UbuntuESM, n=trusty
@@ -59,9 +65,23 @@ upgrade_to_status_cache() {
 
 case "$1" in
     configure)
+      PREVIOUS_PKG_VER=$2
+      # Special case: legacy precise creds allowed for trusty esm
+      # do-release-upgrade substitutes s/precise/trusty/ in all apt sources.
+      # So all we need to do is rename the precise sources file to trusty.
+      # https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/693
+      case $PREVIOUS_PKG_VER in
+          *12.04*) # upgraded from precise
+              if [ -e "$ESM_APT_SOURCE_FILE_PRECISE" ]; then
+                  mv $ESM_APT_SOURCE_FILE_PRECISE \
+                      $ESM_INFRA_APT_SOURCE_FILE_TRUSTY
+              fi
+              ;;
+      esac
+
       # We changed the way we store public files in 19.5; transition to the new
       # status cache for installs of a previous version
-      if dpkg --compare-versions "$2" lt-nl "19.5~"; then
+      if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt-nl "19.5~"; then
           upgrade_to_status_cache
       fi
 


debian/postinst
```diff
@@ -67,8 +68,31 @@ upgrade_to_status_cache() {
     ua status 2>&1 > /dev/null
 }
 
+is_release_upgrade() {
+    # Return 0 if the package version we upgrade from does not match the
+    # current series (e.g. 14.04)
+    if test "${2#*$VERSION_ID}" = ${2}; then
```
OddBloke: I don't think this line is going to work as intended, for a few reasons (I'm going to see if GH will let me do multiple comment threads). Consider:

```sh
#!/bin/sh

demo() {
    echo $2
}

demo
demo a
demo a b
```

I get:

```
$ ./foo.sh 1 2


b
```

As you can see, the function call means that `$2` is no longer referencing the arguments to the script, but the arguments to the function (whether or not there are any).

OddBloke: `${2#*$VERSION_ID}` is a pretty convoluted way of performing this check, I think; I had to double-check several times that it was doing the right thing.  Perhaps something like this would be clearer:

```sh
pkg_version=$2  # this is assuming that $2 is correct, for now
case $pkg_version in
    *${VERSION_ID}*) return 1;;
esac
return 0
```

IMO, the pattern matching makes it clearer what we're actually trying to do here (which is a substring check, basically).

OddBloke: Calling `test` directly is slightly odd; using `if [ ... ]` is more standard.

debian/postinst
```diff
@@ -67,8 +68,31 @@ upgrade_to_status_cache() {
...
+    else
+        # previous pkg version contains the same VERSION_ID as upgraded pkg ver
+        return 1
+    fi
+}
+
 case "$1" in
     configure)
+      # Special case: legacy precise creds allowed for trusty esm
+      # do-release-upgrade would have properly s/precise/trusty/ in apt source
```
OddBloke: "would have properly" here is confusing. It suggests to me that `do-release-upgrade` would be doing the right thing if we let it, which it presumably isn't if we're doing something here?

debian/postinst
```diff
@@ -67,8 +68,31 @@ upgrade_to_status_cache() {
...
+      # Issue: #693
```
OddBloke: Maybe a link would be preferable?

debian/postinst
```diff
@@ -67,8 +68,31 @@ upgrade_to_status_cache() {
...
+      if is_release_upgrade; then
+          if test "${2#*12.04}" != ${2}; then   # then upgraded from precise
```
OddBloke: Is it worth considering a `PREVIOUS_VERSION_RELEASE` variable (or one with a better name :grin:) that captures the value once and can be used in both places where we currently have this logic?

debian/postinst
```diff
@@ -7,6 +7,7 @@ set -e
 ESM_KEY_ID_TRUSTY="56F7650A24C9E9ECF87C4D8D4067E40313CB4B13"
 ESM_APT_GPG_KEY="/etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg"
 ESM_APT_GPG_KEY_OLD="/etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg"
+ESM_APT_SOURCE_FILE_PRECISE="/etc/apt/sources.list.d/ubuntu-esm-precise.list"
 ESM_APT_SOURCE_FILE_OLD="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
```
OddBloke: Do we want to rename the `_OLD` variables, as the precise versions of those files are also "old", just in a different way?

debian/postinst
blackboxsw: good point. done.

debian/postinst
blackboxsw: done

debian/postinst
blackboxsw: done. ESM_INFRA_* for the win. dropping _OLD

debian/postinst
blackboxsw: fixed. Setting a global PREVIOUS_PKG_VER=$2 and using that throughout postinst for clarity/correctness

blackboxsw: dropped test and used your case statement suggestions

debian/postinst
```diff
@@ -3,14 +3,16 @@
 set -e
 
 . /etc/os-release  # For VERSION_ID
+PREVIOUS_PKG_VER=$2
```
OddBloke: `$2` is only the previous package version in the `configure` case, per https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called

debian/postinst
```diff
@@ -3,14 +3,16 @@
...
 
 ESM_KEY_ID_TRUSTY="56F7650A24C9E9ECF87C4D8D4067E40313CB4B13"
 ESM_APT_GPG_KEY="/etc/apt/trusted.gpg.d/ubuntu-esm-v2-keyring.gpg"
 ESM_APT_GPG_KEY_OLD="/etc/apt/trusted.gpg.d/ubuntu-esm-keyring.gpg"
-ESM_APT_SOURCE_FILE_OLD="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
-ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list"
-ESM_APT_PREF_FILE_OLD="/etc/apt/preferences.d/ubuntu-esm-trusty"
-ESM_APT_PREF_FILE="/etc/apt/preferences.d/ubuntu-esm-infra-trusty"
+ESM_APT_SOURCE_FILE_PRECISE="/etc/apt/sources.list.d/ubuntu-esm-precise.list"
+ESM_APT_SOURCE_FILE="/etc/apt/sources.list.d/ubuntu-esm-trusty.list"
```
OddBloke: Nit: Should we encode trusty in these variable names too?

debian/postinst
```diff
@@ -67,11 +69,34 @@ upgrade_to_status_cache() {
...
+        * ) return 0;; #  os-release changed from PREVIOUS_PKG_VER
+    esac
+}
+
 case "$1" in
     configure)
+      # Special case: legacy precise creds allowed for trusty esm
+      # do-release-upgrade substitutes s/precise/trusty/ in apt source
+      # https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/693
+      prev_pkg_version=#2
```
OddBloke: `#2`?

debian/postinst
```diff
@@ -67,11 +69,34 @@ upgrade_to_status_cache() {
...
+        *${VERSION_ID}*)  # os-release didn't change from PREVIOUS_PKG_VER
+            return 1;;
+        * ) return 0;; #  os-release changed from PREVIOUS_PKG_VER
+    esac
+}
+
 case "$1" in
     configure)
+      # Special case: legacy precise creds allowed for trusty esm
+      # do-release-upgrade substitutes s/precise/trusty/ in apt source
```
OddBloke: Aha, I see. Perhaps `s/$/ so all we need to do is rename the file/` would make this even clearer?

debian/postinst
blackboxsw: I agree that is the case, but we don't reference this variable unless we are already down the configure path. What do you suggest we do here instead?  Only set it when we are down configure) path and pass it into is_release_upgrade?

debian/postinst
blackboxsw: I didn't only because of line length and super long variable names both in declaration and in conditionals statements. At risk of the var name being near as long as the full path name, I'd like to avoid that in this case.

debian/postinst
blackboxsw: should have been dropped earlier as I don't use it (and it was wrong)

debian/postinst
OddBloke: We don't _currently_ reference it unless we're down the `configure` path, but we shouldn't make it seem like it will be available in other states.

(You can set it at the start of the configure stanza and it will be set in exactly the same way for the functions that follow it, I believe.)

blackboxsw: done.

debian/postinst
```diff
@@ -67,11 +68,36 @@ upgrade_to_status_cache() {
...
+}
+
 case "$1" in
     configure)
+      PREVIOUS_PKG_VER=$2
+      # Special case: legacy precise creds allowed for trusty esm
+      # do-release-upgrade substitutes s/precise/trusty/ in all apt sources.
+      # So all we need to do is rename the precise sources file to trusty.
+      # https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/693
+      if is_release_upgrade; then
```
OddBloke: OK, now I've seen them refactored (and possibly with fresh eyes this morning), I'm wondering if we really need this check at all. `$PREVIOUS_PKG_VER` matching `*12.04*` is a more specific condition than `$PREVIOUS_PKG_VER` _not_ matching the current release, so I think we can just drop the if statement with no change in behaviour.

What do you think?

blackboxsw: Sounds good @OddBloke, let's drop it for now, we'll have to re-introduce it in the future I bet with Xenial/Bionic release upgrade support.... But we'll be older and wiser developers then ;)

debian/postinst
blackboxsw: Fixing this in the latest rev. As there are some other var name changes that would further muddy the waters.

OddBloke:
LGTM, thanks! Needs rebasing, and please confirm that you've performed a manual upgrade test before you merge.

debian/postinst
```diff
@@ -59,9 +65,23 @@ upgrade_to_status_cache() {
 
 case "$1" in
     configure)
+      PREVIOUS_PKG_VER=$2
+      # Special case: legacy precise creds allowed for trusty esm
+      # do-release-upgrade substitutes s/precise/trusty/ in all apt sources.
+      # So all we need to do is rename the precise sources file to trusty.
+      # https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/693
+      case $PREVIOUS_PKG_VER in
+          *12.04*) # upgraded from precise
```
panlinux: There is only one release of `ubuntu-advantage-tools` in precise and its version is `1`: https://launchpad.net/ubuntu/precise/+source/ubuntu-advantage-tools
Am I missing sometihng?

panlinux: Yeah, this didn't work in a test release-upgrade from precise, with the ppa enabled.
At the end, I had these files in `/etc/apt/`:
```
/etc/apt/preferences.d/ubuntu-esm-infra-trusty
/etc/apt/sources.list.d/ubuntu-esm-precise.list
/etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list
```
`ubuntu-esm-precise.list` has these contents:
```
deb https://lol:secret@esm.ubuntu.com/ubuntu trusty-updates main
# deb-src https://lol:secret@esm.ubuntu.com/ubuntu precise main
deb https://lol:secret@esm.ubuntu.com/ubuntu trusty-security main
```
And `ubuntu-esm-infra-trusty.list`:
```
deb https://esm.ubuntu.com/ubuntu trusty-infra-security main
# deb-src https://esm.ubuntu.com/ubuntu trusty-infra-security main

deb https://esm.ubuntu.com/ubuntu trusty-infra-updates main
# deb-src https://esm.ubuntu.com/ubuntu trusty-infra-updates main
```
The only reason we didn't get the apt warning this time is because the pockets have different names, trusty having the `infra` added term.
Furthermore, esm is disabled now, because of the pinning file in `/etc/apt/preferences.d/ubuntu-esm-infra-trusty`:
```
Package: *
Pin: release o=UbuntuESM, n=trusty
Pin-Priority: never
```
`libcurl3`, for example, I have installed from the esm repository after this `do-release-upgrade`, but the pinning effectively disables it:
```
root@p1:~# apt-cache policy libcurl3
libcurl3:
  Installed: 7.35.0-1ubuntu2.20+esm3
  Candidate: 7.35.0-1ubuntu2.20+esm3
  Version table:
 *** 7.35.0-1ubuntu2.20+esm3 0
       -32768 https://esm.ubuntu.com/ubuntu/ trusty-infra-security/main amd64 Packages
       -32768 https://esm.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
        100 /var/lib/dpkg/status
     7.35.0-1ubuntu2.20 0
        500 http://archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
        500 http://security.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
     7.35.0-1ubuntu2 0
        500 http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
root@p1:~# 
```
And if I try a `--reinstall`, it kind of fails, probably because it fetches the package from `trusty-infra-security` for which we have no credentials. The credentials we have are only for `trusty-security`, in the `ubuntu-esm-precise.list` (yes, precise) list file. Still, in the end, it downloaded something, maybe it tried both repositories (but only showed one):
```
root@p1:~# apt install --reinstall libcurl3
Reading package lists... Done
Building dependency tree       
0 upgraded, 0 newly installed, 1 reinstalled, 0 to remove and 1 not upgraded.
Need to get 173 kB of archives.
After this operation, 0 B of additional disk space will be used.
Get:1 https://esm.ubuntu.com/ubuntu/ trusty-infra-security/main libcurl3 amd64 7.35.0-1ubuntu2.20+esm3 [173 kB]
Err https://esm.ubuntu.com/ubuntu/ trusty-infra-security/main libcurl3 amd64 7.35.0-1ubuntu2.20+esm3
  HttpError401
Fetched 173 kB in 3s (55.6 kB/s)
(Reading database ... 77767 files and directories currently installed.)
Preparing to unpack .../libcurl3_7.35.0-1ubuntu2.20+esm3_amd64.deb ...
Unpacking libcurl3:amd64 (7.35.0-1ubuntu2.20+esm3) over (7.35.0-1ubuntu2.20+esm3) ...
Setting up libcurl3:amd64 (7.35.0-1ubuntu2.20+esm3) ...
Processing triggers for libc-bin (2.19-0ubuntu6.15) ...
root@p1:~# echo $?
0
```

panlinux: This confirms the release upgrade upgraded ubuntu-advantage-tools version 1 from precise:
```
Unpacking python3-pkg-resources (3.3-1ubuntu2) ...
Preparing to unpack .../ubuntu-advantage-tools_19.6~642~git.986b779~ubuntu14.04.1_amd64.deb ...
Moving obsolete conffile /etc/update-motd.d/99-esm out of the way...
Unpacking ubuntu-advantage-tools (19.6~642~git.986b779~ubuntu14.04.1) over (1) ...
```

OddBloke:
@blackboxsw Does @panlinux's comment above indicate that you didn't perform the requested manual test before merging?

blackboxsw:
@OddBloke it means I did on the earlier iteration of this branch, but then I rewrote my changes to look specifically at *12.04* prev_version when I dropped the is_release_upgrade function and failed to re-run the upgrade test. Doing that again now :/

Issue #806, opened 2019-10-09
Status alignment off by one


cpaelzer:
This isn't #69 which at first sounded similar, but that was too much an old design.

An attached status output is perfectly aligned
```
SERVICE       ENTITLED  STATUS    DESCRIPTION
cc-eal        no        —         Common Criteria EAL2 Provisioning Packages
cis-audit     no        —         Center for Internet Security Audit Tools
esm-infra     yes       enabled   UA Infra: Extended Security Maintenance
fips          no        —         NIST-certified FIPS modules
fips-updates  no        —         Uncertified security updates to FIPS modules
livepatch     yes       n/a       Canonical Livepatch service
```

But before ua attach it seems off by one (content is one char right of the headers)
```
SERVICE      AVAILABLE  DESCRIPTION
cc-eal        no         Common Criteria EAL2 Provisioning Packages
esm-infra     yes        UA Infra: Extended Security Maintenance
fips          no         NIST-certified FIPS modules
fips-updates  no         Uncertified security updates to FIPS modules
livepatch     yes        Canonical Livepatch service
```

PR #814, opened 2019-10-09
status: unattached status output headers alignment off by one

blackboxsw:
Fixes: #806

### DIFF
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -196,11 +196,12 @@ def format_tabular(status: "Dict[str, Any]") -> str:
     """Format status dict for tabular output."""
     if not status["attached"]:
         content = [
+            "",
             STATUS_UNATTACHED_TMPL.format(
-                name="\nSERVICE",
+                name="SERVICE",
                 available="AVAILABLE",
                 description="DESCRIPTION",
-            )
+            ),
         ]
         for service in status["services"]:
             content.append(STATUS_UNATTACHED_TMPL.format(**service))
--- a/uaclient/tests/test_status.py
+++ b/uaclient/tests/test_status.py
@@ -1,5 +1,6 @@
 import mock
 import pytest
+import string
 
 from uaclient import config
 from uaclient.status import format_tabular, TxtColor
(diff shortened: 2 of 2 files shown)


uaclient/status.py
```diff
@@ -198,8 +198,8 @@ def format_tabular(status: "Dict[str, Any]") -> str:
         content = [
             STATUS_UNATTACHED_TMPL.format(
                 name="\nSERVICE",
-                available="AVAILABLE",
-                description="DESCRIPTION",
+                available=" AVAILABLE",
```
OddBloke: Is the issue here actually that we have `\n` at the start of name, which takes up one "character" of the line?

blackboxsw: absolutely right. good suggestion. I've added a unit test and fix the change properly.

tox.ini
```diff
@@ -19,7 +19,7 @@ commands =
     mypy: mypy --python-version 3.5 uaclient/
     mypy: mypy --python-version 3.6 uaclient/
     mypy: mypy --python-version 3.7 uaclient/
-    black: black --check --diff uaclient/
```
OddBloke: Why?

(This should at least be a separate commit capturing the rationale.)

blackboxsw: bogus commit. removed and forgot to push.

uaclient/tests/test_cli_attach.py
```diff
@@ -17,8 +17,12 @@
...
 BASIC_MACHINE_TOKEN = {
     "machineTokenInfo": {
-        "contractInfo": {"name": "mycontract", "resourceEntitlements": []},
-        "accountInfo": {"name": "accountName"},
+        "contractInfo": {
+            "id": "contract-1",
+            "name": "mycontract",
+            "resourceEntitlements": [],
+        },
+        "accountInfo": {"id": "acct-1", "name": "accountName"},
```
OddBloke: It's not clear to me what's motivating this change being included in this PR. Am I missing some information?

blackboxsw: No motivation, I leaked it in from another branch accidentally. Removed and rechecked the full diff to confirm no stray commits from other branches.

Issue #691, opened 2019-07-13
trusty: enable livepatch on stale apt sources needs apt-get update


blackboxsw:
On a stale trusty vm which hadn't run apt-get update in  a while I ran into an inability to install snapd package. This was resolved by first running apt-get update to install snapd.

```
multipass@dev-t2:~$ sudo ua enable livepatch
One moment, checking your subscription first
Installing snapd
Traceback (most recent call last):
  File "/usr/bin/ua", line 9, in <module>
    load_entry_point('ubuntu-advantage-tools==19.5', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 398, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 427, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 60, in wrapper
    return func(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 222, in action_enable
    return 0 if _perform_enable(args.name, cfg) else 1
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 208, in _perform_enable
    ret = entitlement.enable(silent_if_inapplicable=silent_if_inapplicable)
  File "/usr/lib/python3/dist-packages/uaclient/entitlements/livepatch.py", line 52, in enable
    capture=True, retry_sleeps=apt.APT_RETRIES)
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 285, in subp
    out, err = _subp(args, rcs, capture, timeout, decode)
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 249, in _subp
    stdout=out.decode('utf-8'), stderr=err.decode('utf-8'))
uaclient.util.ProcessExecutionError: Failed running command 'apt-get install --assume-yes snapd' [exit(100)]. Message: E: There are problems and -y was used without --force-yes
multipass@dev-t2:~$ sudo apt-get install --assume-yes snapd
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following extra packages will be installed:
  cgroup-lite liblzo2-2 libseccomp2 libsystemd-journal0 squashfs-tools systemd
Suggested packages:
  systemd-ui
The following NEW packages will be installed:
[...]

PR #829, opened 2019-10-11
livepatch: call apt-get update if snapd needs to be installed

blackboxsw:
Fixes: #691

### DIFF
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -50,6 +50,17 @@ class LivepatchEntitlement(base.UAEntitlement):
         if not util.which("/snap/bin/canonical-livepatch"):
             if not util.which(SNAP_CMD):
                 print("Installing snapd")
+                print(status.MESSAGE_APT_UPDATING_LISTS)
+                try:
+                    apt.run_apt_command(
+                        ["apt-get", "update"], status.MESSAGE_APT_UPDATE_FAILED
+                    )
+                except exceptions.UserFacingError as e:
+                    logging.debug(
+                        "Trying to install snapd."
+                        " Ignoring apt-get update failure: %s",
+                        str(e),
+                    )
                 util.subp(
                     ["apt-get", "install", "--assume-yes", "snapd"],
                     capture=True,
--- a/uaclient/entitlements/tests/test_livepatch.py
+++ b/uaclient/entitlements/tests/test_livepatch.py
@@ -1,6 +1,7 @@
 """Tests related to uaclient.entitlement.base module."""
 
 import copy
+import logging
 import mock
 from types import MappingProxyType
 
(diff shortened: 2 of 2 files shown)


OddBloke:
@powersj We don't have an entry in the UX doc for this case (AFAICT), can you determine what the messaging we want to display in this case is?

uaclient/entitlements/livepatch.py
```diff
@@ -50,6 +50,18 @@ def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
...
+                    apt.run_apt_command(
+                        ["apt-get", "update"], status.MESSAGE_APT_UPDATE_FAILED
+                    )
+                except exceptions.UserFacingError as e:
+                    logging.debug(
+                        "Trying to install snapd."
+                        " Ignoring apt-get update failure: %s",
+                        str(e),
+                    )
+                    pass
```
OddBloke: No need for this.

powersj:
Hmm, from the UX doc here is the simple enable livepatch scenario:

```
$ sudo ua enable livepatch
Installing canonical-livepatch snap
A reboot is required to complete the install
```

However, in cases where we update the package lists we include the message, `Updating package lists`. Therefore, I propose:

```
$ sudo ua enable livepatch
Updating package lists
Installing canonical-livepatch snap
A reboot is required to complete the install
```

Or is there a concern we should be calling out the fact that we are installing systemd and snapd packages as well?

OddBloke:
We already do emit "Installing snapd". That's what lead me to suspect that we maybe hadn't considered this case properly.

I'm not sure my opinion of what we should do here matters all that much, because I don't think we should be installing systemd on people's trusty systems full stop. So the best impression I can do of a user is "yes, please tell me that you're installing an entirely different init system so I can Ctrl-C out of this as fast as possible and reconsider the choices that brought me to this point", which presumably isn't the user experience we're optimising for. (And I do think people will (rightly) be *angry* if we install systemd on a trusty system without telling them. So... :man_shrugging:)

We did discuss at some point prompting for these installations, so perhaps we should emit the messages and they'll turn into prompts at some point in the future?

powersj:
The existing behavior is to do the install without warning so I do not feel we are regressing or changing behavior for the worse. I do agree that adding an entirely new init system is a large change, which is why you successfully negotiated that we would not do this automatically. However, I say we merge this without the confirmation for now and we can loop back.

Please file a bug.

OddBloke:
Yep, that sounds totally reasonable. I'll file two issues: one to cover the fact we don't mention that `systemd` is getting installed (with or without this PR), and one to cover prompting the user for permission.

OddBloke:
OK, so @blackboxsw I think this just need that `pass` dropping and a rebase and we're good to go.

uaclient/entitlements/livepatch.py
blackboxsw: done.

Issue #807, opened 2019-10-09
Check for livepatch availability/enablement has odd edge cases


cpaelzer:
Issues to fix:

- [X]  check container first before minKernelVersion as you can't 'fix' being in a container so don't tell people they need to upgrade
- [X] minKernelVersion check is broken as it separately compares kernMajorVer > minKernMajor && kernMinorVer > minKernelMinor   5.0 is > 4.3 :) 
- [X] Availability of livepatch should always be "no" for containers, this is really a ua-contract backend issue that I've filed for discussion @ https://github.com/CanonicalLtd/ua-contracts/issues/469
Filed separate UA Client issue for the third item above: #831 

I think we all agree that the kernel is part of the Host and livepatch will never work from inside a container.

But before subscription it shows "yes" for availability.
`livepatch     yes        Canonical Livepatch service`

Interestingly on a Trusty kernel 3.13 system it is Availability "no".
After upgrade to kernel 4.4 it becomes Availability "yes".
Shouldn't the container also report "no" here no matter what kernel version is active?

Furthermore after attaching it to my token I've seen this in the Container:
That seems to be reported ok in `ua status`
`livepatch     yes       n/a       Canonical Livepatch service`

Ok, my contract is entitled, but it isn't applicable here (that makes sense).
But when I ignore that statement and try to enable it the error is misleading:

```
ua enable livepatch
One moment, checking your subscription first
Livepatch is not available for kernel 5.0.0-29-generic.
Minimum kernel version required: 4.4
```

First: The error says "minimum 4.4" which 5.0 clearly satisfies.
Seconds: On top even if we'd have 4.4 or whatever else, it just would not work from inside a container.

I wonder if on `ua enable livepatch` there should be something like:
```
if system-detect-virt --container
   "Livepatch is unavailable, it  has to be enabled at the host where the kernel resides not in containers"
```

Note: I've found in further tests that you seem to have code for the second already.
If run on a valid kernel (Host is Trusty with 4.4):
```
root@t:~# ua enable livepatch
One moment, checking your subscription first
Cannot install Livepatch on a container
```
Maybe the just ordering of those checks needs to be improved to first check for container and only afterwards for kernel version?

Summary:
 * Availability of livepatch in unattached `ua status`  in a container should always be "no" (it is ok to be called entitled, but not available)
 * On trying to enable livepatch it it should be made clear that trying to do so in a container is in vain (code reordering?)

powersj:
[...]

PR #832, opened 2019-10-11
livepatch: check is_container before recommending kernel upgrade

blackboxsw:
livepatch: check is_container before recommending kernel upgrade

On a container, livepatch isn't every going to be supported, so
there is little sense in checking kernel version first and telling
the user to upgrade their kernel only to find out after reboot that
they still can't enable livepatch because livepatch doesn't run on the
container.
    
Reorder static_affordances (which tend to be uncorrectable on a
platform) before more dynamic environment affordances like <series>
or kernel revision which can up updated on a machine to bring it into
compliance.

Fixes: #807

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -141,6 +141,9 @@ class UAEntitlement(metaclass=abc.ABCMeta):
                 ApplicabilityStatus.APPLICABLE,
                 "no entitlement affordances checked",
             )
+        for error_message, functor, expected_result in self.static_affordances:
+            if functor() != expected_result:
+                return ApplicabilityStatus.INAPPLICABLE, error_message
         affordances = entitlement_cfg["entitlement"].get("affordances", {})
         platform = util.get_platform_info()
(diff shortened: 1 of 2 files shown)


uaclient/entitlements/base.py
```diff
@@ -194,16 +197,12 @@ def applicability_status(self) -> "Tuple[ApplicabilityStatus, str]":
...
-                [
-                    int(match.group("major")) < min_kern_major,
-                    int(match.group("minor")) < min_kern_minor,
-                ]
-            ):
+            min_kern = float("{}.{}".format(min_kern_major, min_kern_minor))
+            current_kern = float(
+                "{}.{}".format(match.group("major"), match.group("minor"))
+            )
+            if current_kern < min_kern:
```
OddBloke: This won't work for minor versions above 9:

```py
In [3]: float("4.15") < float("4.4")
Out[3]: True
```

blackboxsw: sonofagun. Right, will correct this in separate PR.

PR #834, opened 2019-10-11
livepatch: fix minKernelVersion comparison to account for major.minor

blackboxsw:
Fixes: #807

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -197,11 +197,13 @@ class UAEntitlement(metaclass=abc.ABCMeta):
 
             if not match:
                 return ApplicabilityStatus.INAPPLICABLE, invalid_msg
-            if any(
-                [
-                    int(match.group("major")) < min_kern_major,
-                    int(match.group("minor")) < min_kern_minor,
-                ]
+            kernel_major = int(match.group("major"))
+            kernel_minor = int(match.group("minor"))
+            if kernel_major < min_kern_major:
+                return ApplicabilityStatus.INAPPLICABLE, invalid_msg
+            elif (
+                kernel_major == min_kern_major
+                and kernel_minor < min_kern_minor
             ):
                 return ApplicabilityStatus.INAPPLICABLE, invalid_msg
         return ApplicabilityStatus.APPLICABLE, ""
--- a/uaclient/entitlements/tests/test_livepatch.py
+++ b/uaclient/entitlements/tests/test_livepatch.py
@@ -34,7 +34,7 @@ M_BASE_PATH = "uaclient.entitlements.base.UAEntitlement."
 def entitlement(entitlement_factory):
     affordances = {
         "architectures": ["x86_64"],
-        "minKernelVersion": "4.3",
+        "minKernelVersion": "4.4",
         "kernelFlavors": ["generic", "lowlatency"],
         "tier": "stable",
     }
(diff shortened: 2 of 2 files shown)


OddBloke:
Why submit this as a PR at all at this point? (I've cancelled the Travis run so other things can run first, hence the failure.)

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -202,6 +208,37 @@ def test_can_enable_false_on_unsupported_kernel_flavor(
         )
         assert (msg, "") == capsys.readouterr()
 
+    @pytest.mark.parametrize(
```
blackboxsw: @OddBloke I added this test to validate above and below cut line for minKernelVersion comparisons.

uaclient/entitlements/base.py
```diff
@@ -197,12 +197,14 @@ def applicability_status(self) -> "Tuple[ApplicabilityStatus, str]":
...
-                ]
-            ):
+            kernel_major = int(match.group("major"))
+            kernel_minor = int(match.group("minor"))
+            major_gt_min = kernel_major > min_kern_major
+            minor_gte_min = bool(
+                kernel_major >= min_kern_major
+                and kernel_minor >= min_kern_minor
+            )
+            if not any([major_gt_min, minor_gte_min]):
```
OddBloke: This line _reads_ strangely: "if either the major isn't greater than its minimum or the minor isn't greater than or equal to its minimum, then we are inapplicable". That's the previous logic we had, which is incorrect. I think changing `minor_gte_min` to a more accurate name would probably make this read a lot better, and make the code easier to grok.

(As a side note, I think this would be easier to grok in general if we had more than one return here, then we could do one condition at a time: `if kernel_major < min_kern_major: return ...` is clearer than defining a boolean which is referenced later in a compound if statement. This definitely doesn't need to happen in this PR or, indeed, at all.)

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -202,6 +208,37 @@ def test_can_enable_false_on_unsupported_kernel_flavor(
...
+            if meets_min_version:
+                assert entitlement.can_enable()
+            else:
+                assert not entitlement.can_enable()
+        if meets_min_version:
+            msg = ""
+        else:
+            msg = (
+                "Livepatch is not available for kernel {}.\n"
+                "Minimum kernel version required: 4.3\n".format(kernel_version)
```
OddBloke: Where does this "4.3" come from? Isn't the 4.4 the cutoff? (Did we ever even release a 4.3 Ubuntu kernel?)

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -202,6 +208,37 @@ def test_can_enable_false_on_unsupported_kernel_flavor(
         )
         assert (msg, "") == capsys.readouterr()
 
+    @pytest.mark.parametrize(
+        "kernel_version,meets_min_version",
+        (
+            ("3.5.0-00-generic", False),
+            ("4.2.0-00-generic", False),
+            ("4.10.0-00-generic", True),
+            ("5.0.0-00-generic", True),
```
OddBloke: Let's add a case for exactly the cut-off here (i.e. 4.4), to make sure that we are drawing the line on the right side of it.

uaclient/entitlements/tests/test_livepatch.py
OddBloke: Thanks!

uaclient/entitlements/base.py
blackboxsw: Good suggestion Dan.  I like the early exit/return option better.  Fixed.

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: I was using an example here, but probably best to use the expected example which should come from the live ua-contract server (4.4).

uaclient/entitlements/tests/test_livepatch.py
blackboxsw: I was just using an example here, but probably best to use the expected output that ua-contracts will be sending us and rework comparative examples related to that. Fixing now.

uaclient/entitlements/tests/test_livepatch.py
OddBloke: That wasn't what this comment was about; you're responding to the one that GH has automatically hidden because you changed that line.

This request is still open.

OddBloke: Thanks!

Issue #858, opened 2019-10-17
Helpful message should be printed when there are no connection to server


telescreen:
When the client cannot connect to the `https://contracts.canonical.com`, it prints out a python stack trace instead of a helpful message

```
resources = get_available_resources(self)
  File "/usr/lib/python3/dist-packages/uaclient/contract.py", line 291, in get_available_resources
    resources = client.request_resources()
  File "/usr/lib/python3/dist-packages/uaclient/contract.py", line 104, in request_resources
    API_V1_RESOURCES + "?" + urllib.parse.urlencode(query_params)
  File "/usr/lib/python3/dist-packages/uaclient/serviceclient.py", line 63, in request_url
    e, code=getattr(e, "code", None), headers=headers, url=url
uaclient.util.UrlError: [Errno 111] Connection refused

```

PR #860, opened 2019-10-17
status: user-facing-error on get_available_resources connectivity issues

blackboxsw:
status: better user-facing error on connectivity errors

Generalize exception handling for any util.UrlError raised by the
commandline. Log any urlerror tracebacks and emit the message:
    
Fixes: #858

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -500,6 +500,11 @@ def main_error_handler(func):
                 logging.exception("KeyboardInterrupt")
             print("Interrupt received; exiting.", file=sys.stderr)
             sys.exit(1)
+        except util.UrlError as exc:
+            with util.disable_log_to_console():
+                logging.exception(exc)
+            print(ua_status.MESSAGE_CONNECTIVITY_ERROR, file=sys.stderr)
+            sys.exit(1)
         except exceptions.UserFacingError as exc:
             with util.disable_log_to_console():
                 logging.exception(exc.msg)
--- a/uaclient/tests/test_cli.py
+++ b/uaclient/tests/test_cli.py
@@ -2,6 +2,7 @@ import io
 import logging
 import mock
 import os
+import socket
 import stat
 import sys
 import textwrap
@@ -23,6 +24,8 @@ from uaclient.exceptions import (
     UnattachedError,
 )
 from uaclient.testing.fakes import FakeConfig
+from uaclient import status
+from uaclient import util
 
 
 BIG_DESC = "123456789 " * 7 + "next line"
@@ -248,6 +251,35 @@ class TestMain:
         assert "{}\n".format(msg) == err
         error_log = caplog_text()
         assert msg in error_log
+        assert "Traceback (most recent call last):" in error_log
+
+    @mock.patch("uaclient.cli.setup_logging")
+    @mock.patch("uaclient.cli.get_parser")
+    def test_url_error_handled_gracefully(
+        self,
+        m_get_parser,
+        _m_setup_logging,
+        capsys,
+        logging_sandbox,
+        caplog_text,
+    ):
+
+        m_args = m_get_parser.return_value.parse_args.return_value
+        m_args.action.side_effect = util.UrlError(
+            socket.gaierror(-2, "Name or service not known")
+        )
+
+        with pytest.raises(SystemExit) as excinfo:
+            main(["some", "args"])
+
+        exc = excinfo.value
+        assert 1 == exc.code
+
+        out, err = capsys.readouterr()
+        assert "" == out
+        assert "{}\n".format(status.MESSAGE_CONNECTIVITY_ERROR) == err
+        error_log = caplog_text()
+        assert " [Errno -2] Name or service not known" in error_log
         assert "Traceback (most recent call last):" in error_log
 
 
--- a/uaclient/tests/test_cli_status.py
+++ b/uaclient/tests/test_cli_status.py
@@ -1,0 +1,100 @@
+import json
+import mock
+import socket
+
+import pytest
+
+from uaclient.testing.fakes import FakeConfig
+from uaclient import util
+
+from uaclient.cli import action_status
+
+M_PATH = "uaclient.cli."
+
+
+RESPONSE_LIVEPATCH_AVAILABLE = [{"name": "livepatch", "available": True}]
+UNATTACHED_STATUS = """\
+SERVICE       AVAILABLE  DESCRIPTION
+livepatch     yes        Canonical Livepatch service
+
+This machine is not attached to a UA subscription.
+See https://ubuntu.com/advantage
+"""
+
+ATTACHED_STATUS = """\
+SERVICE       ENTITLED  STATUS    DESCRIPTION
+cc-eal        no        —         Common Criteria EAL2 Provisioning Packages
+cis-audit     no        —         Center for Internet Security Audit Tools
+esm-infra     no        —         UA Infra: Extended Security Maintenance
+fips          no        —         NIST-certified FIPS modules
+fips-updates  no        —         Uncertified security updates to FIPS modules
+livepatch     no        —         Canonical Livepatch service
+
+Enable services with: ua enable <service>
+
+                Account: test_account
+           Subscription: test_contract
+            Valid until: n/a
+Technical support level: n/a
+"""
+
+
+@mock.patch(
+    M_PATH + "contract.get_available_resources",
+    return_value=RESPONSE_LIVEPATCH_AVAILABLE,
+)
+@mock.patch(M_PATH + "os.getuid", return_value=0)
+class TestActionStatus:
+    def test_attached(self, m_getuid, m_get_avail_resources, capsys):
+        """Check that root and non-root will emit attached status"""
+        cfg = FakeConfig.for_attached_machine()
+        assert 0 == action_status(mock.MagicMock(), cfg)
+        # capsys already converts colorized non-printable chars to space
+        # Strip non-printables from output
+        printable_stdout = capsys.readouterr()[0].replace(" " * 17, " " * 8)
+        assert ATTACHED_STATUS == printable_stdout
+
+    def test_unattached(self, m_getuid, m_get_avail_resources, capsys):
+        """Check that unattached status is emitted to console"""
+        cfg = FakeConfig()
+
+        assert 0 == action_status(mock.MagicMock(), cfg)
+        assert UNATTACHED_STATUS == capsys.readouterr()[0]
+
+    def test_unattached_json(self, m_getuid, m_get_avail_resources, capsys):
+        """Check that unattached status json output is emitted to console"""
+        cfg = FakeConfig()
+
+        args = mock.MagicMock(format="json")
+        assert 0 == action_status(args, cfg)
+        expected = {
+            "_doc": (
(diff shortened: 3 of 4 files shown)


uaclient/contract.py
```diff
@@ -288,5 +288,10 @@ def request_updated_contract(
 def get_available_resources(cfg) -> "List[Dict]":
     """Query available resources from the contrct server for this machine."""
     client = UAContractClient(cfg)
-    resources = client.request_resources()
+    try:
+        resources = client.request_resources()
+    except util.UrlError as exc:
```
OddBloke: Should we do this error handling further up the stack, so that _any_ `util.UrlError`s that would show a traceback end up reporting `status.MESSAGE_CONNECTIVITY_ERROR` to users?

blackboxsw: @OddBloke @powersj:   I *think* we take url error handling up a level. Though I'm not certain what messaging we *should* be doing then in the case of ua enable|disable|refresh which already could have static failure messages.   Should we supplement those messages by first emitting the connectivity error message followed by the current enable failure message?    
Examples below:

failure to attach due to connectivity:
 
Here are some examples that I think we probably need  to cleanup

- ua enable <someinapplicableservice> with network connectivity 

```
# ua enable fips
This subscription is not entitled to FIPS.
See: sudo ua status or https://ubuntu.com/advantage

# ua enable livepatch
Cannot install Livepatch on a container

```

- ua refresh on network connectivity
```
# ua refresh
Unable to refresh your subscription
```

OddBloke: I don't think we need to decide that for this change to land; regardless of what those commands' behaviour is, I still think we want a top-level handler for `util.UrlError` for the cases that we aren't currently handling.

blackboxsw: Agreed. I've moved this error handling up into uaclient/cli and added unit test coverage for this failure path

uaclient/cli.py
```diff
@@ -500,6 +500,14 @@ def wrapper(*args, **kwargs):
                 logging.exception("KeyboardInterrupt")
             print("Interrupt received; exiting.", file=sys.stderr)
             sys.exit(1)
+        except util.UrlError as exc:
+            with util.disable_log_to_console():
+                logging.exception(exc)
```
OddBloke: The argument here should be the message we want in the log attached to the traceback; is `exc` really what we want (it could be!)?

uaclient/cli.py
```diff
@@ -500,6 +500,14 @@ def wrapper(*args, **kwargs):
...
+            print(
+                "{}".format(ua_status.MESSAGE_CONNECTIVITY_ERROR),
```
OddBloke: No need to format here.

uaclient/tests/test_cli_status.py
```diff
@@ -0,0 +1,97 @@
...
+@mock.patch(M_PATH + "os.getuid", return_value=0)
+class TestActionStatus:
+    def test_attached(self, m_getuid, m_get_avail_resources, capsys):
+        """Check that root and non-root will emit attached status"""
+        cfg = FakeConfig.for_attached_machine()
+        assert 0 == action_status(mock.MagicMock(), cfg)
+        # capsys already converts colorized non-printable chars to space
+        # String non-printables from output
+        printable_stdout = capsys.readouterr()[0].replace(
+            "                 ", "        "
```
OddBloke: This would be easier to read as `" " * 3` (with the actual numbers in, of course).

uaclient/tests/test_contract.py
```diff
@@ -96,6 +97,32 @@ def test_overrides_applied_before_comparison(
 
 @mock.patch(M_PATH + "UAContractClient")
 class TestGetAvailableResources:
+    def test_request_resources_error_on_network_disconnected(self, client):
```
OddBloke: If I understand correctly, this test is just checking that `get_available_resources` doesn't catch `util.UrlError`? This seems like a lot of test code to ensure that, would it be simpler if instead of raising it specifically for this URL we just raised for any URLs? It feels like something like this could potentially replace diff lines 104-121:

```py
client.return_value.request_url.side_effect = util.UrlError(...)
```

uaclient/tests/test_contract.py
```diff
@@ -96,6 +97,32 @@ def test_overrides_applied_before_comparison(
...
+                url: util.UrlError(
+                    socket.gaierror(-2, "Name or service not known")
+                )
+            }
+            return fake_client
+
+        client.side_effect = fake_contract_client
+        with pytest.raises(util.UrlError) as exc:
+            get_available_resources(cfg)
+        assert "[Errno -2] Name or service not known" == str(exc.value)
```
OddBloke: Would it make sense to instead assert that `exc` is exactly the same exception that we put in? (Or is it not?)

uaclient/tests/test_cli.py
```diff
@@ -250,6 +253,36 @@ def test_user_facing_error_handled_gracefully(
         assert msg in error_log
         assert "Traceback (most recent call last):" in error_log
 
+    @pytest.mark.parametrize("caplog_text", [logging.ERROR], indirect=True)
```
OddBloke: Do we not capture error logging by default?

uaclient/tests/test_cli_status.py
```diff
@@ -0,0 +1,97 @@
+import mock
+import socket
+
+import pytest
+
+from uaclient.testing.fakes import FakeConfig
+from uaclient import util
+
+try:
+    from typing import Any, Dict, Optional  # noqa: F401
```
OddBloke: These imports appear unused?

uaclient/tests/test_cli_status.py
```diff
@@ -0,0 +1,97 @@
...
+        assert 0 == action_status(mock.MagicMock(), cfg)
+        assert UNATTACHED_STATUS == capsys.readouterr()[0]
+        m_getuid.return_value = 1000
+        assert 0 == action_status(mock.MagicMock(), cfg)
+        assert UNATTACHED_STATUS == capsys.readouterr()[0]
+
+    def test_error_on_connectivity_errors(
+        self, m_getuid, m_get_avail_resources, capsys
+    ):
+        """Check that root and non-root emit unattached status"""
```
OddBloke: This docstring is inaccurate.

uaclient/tests/test_cli_status.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        assert ATTACHED_STATUS == printable_stdout
+
+    def test_unattached(self, m_getuid, m_get_avail_resources, capsys):
+        """Check that root and non-root emit unattached status"""
+        cfg = FakeConfig()
+
+        assert 0 == action_status(mock.MagicMock(), cfg)
+        assert UNATTACHED_STATUS == capsys.readouterr()[0]
+        m_getuid.return_value = 1000
```
OddBloke: This isn't necessarily an effective test for non-root; you may also need to instantiate a new config (because its in-memory caches are in a state we can never reach as non-root in production).

uaclient/cli.py
blackboxsw: Ok, let's log the full traceback for this which I think is not too helpful, and follow up with a more useful log message in #868

uaclient/cli.py
blackboxsw: fixed.

uaclient/tests/test_cli_status.py
blackboxsw: done.

uaclient/tests/test_contract.py
blackboxsw: great suggestion thanks for thinking that through. I've patched the UAContractClient object's request_url method.

uaclient/tests/test_cli.py
blackboxsw: oops we do. it would only make sense to do this if we were limiting to WARNING or DEBUG messages and exact matching I guess.

uaclient/tests/test_cli_status.py
blackboxsw: dropped.

uaclient/tests/test_contract.py
blackboxsw: True. fixed to         assert urlerror == exc.value

uaclient/tests/test_cli_status.py
blackboxsw: I should have added that comment, but just wanted to make sure plumbing worked for root and non-root status, which is actually tested a level below in uaclient/tests/test_config:446 with test_nonroot_unattached_is_same_as_unattached_root.
So no need to duplicate that test up at this level as we've already unit tested root vs non-root plumbing.

blackboxsw: I've dropped the non-root part of this test.

uaclient/tests/test_cli_status.py
```diff
@@ -0,0 +1,84 @@
...
+    def test_attached(self, m_getuid, m_get_avail_resources, capsys):
+        """Check that root and non-root will emit attached status"""
+        cfg = FakeConfig.for_attached_machine()
+        assert 0 == action_status(mock.MagicMock(), cfg)
+        # capsys already converts colorized non-printable chars to space
+        # Strip non-printables from output
+        printable_stdout = capsys.readouterr()[0].replace(" " * 17, " " * 8)
+        assert ATTACHED_STATUS == printable_stdout
+
+        m_getuid.return_value = 1000
```
OddBloke: Weren't you dropping this part?

blackboxsw: ahh did from unattached, not from the attached test case. Dropping.

uaclient/cli.py
OddBloke: The call to `logging.exception` is what logs the traceback. You haven't addressed my question about what message should be attached to that traceback.

uaclient/tests/test_cli_status.py
```diff
@@ -0,0 +1,77 @@
...
+        """Check that root and non-root raise error on connectivity issues"""
+        m_get_avail_resources.side_effect = util.UrlError(
+            socket.gaierror(-2, "Name or service not known")
+        )
+
+        cfg = FakeConfig()
+
+        with pytest.raises(util.UrlError):
+            action_status(mock.MagicMock(), cfg)
+        m_getuid.return_value = 1000
```
OddBloke: As I mentioned previously, this isn't sufficient to test for a non-root user. I would suggest parameterising the test.

blackboxsw: it needed dropping and I should have attended to that when I dropped the other two instances of root vs non-root tests here. Since this is root/non-root is tested already on the config.status tests we don't need duplication of tests up at this level. I have also added a json format output test since we do that up in action_status.

uaclient/cli.py
blackboxsw: I misunderstood your intent given our earlier conversation at standup.  What I *think* for the moment (before #868 is addressed) is to log the action str representation of the exception instance which is what we are currently doing with logging.exception(exc). 
What I propose we do in #868 is no longer log the exception traceback at all, but maybe MESSAGE_CONNECTIVITY_ERROR: <str(exc)> + str(exc.url).

OddBloke: Sounds good, thanks!

PR #870, opened 2019-10-21
cli: replace UTF-8 dash with ASCII dash for non-UTF-8 output 

### LINKED ISSUE #859: ua attach or status run into "UnicodeEncodeError: 'ascii' codec can't encode character '\u2014' in position 84: ordinal not in range(128)"
Getting the following error when running `sudo ua attach` on fully updated Trusty Server.

This is a KVM instance. Ubuntu Trusty freshly installed and upgraded to 14.04.6 LTS.

```
ubuntu@trusty:\~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.6 LTS
Release:	14.04
Codename:	trusty

ubuntu@trusty:\~$ uname -a
Linux trusty 3.13.0-170-generic #220-Ubuntu SMP Thu May 9 12:40:49 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

ubuntu@trusty:\~$ sudo ua --debug attach $TOKEN_HIDDEN
(...)
DEBUG: Writing file: /var/lib/ubuntu-advantage/status.json
    load_entry_point('ubuntu-advantage-tools==19.5', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 486, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 516, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 335, in action_attach
    action_status(args=None, cfg=cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 426, in action_status
    print(ua_status.format_tabular(cfg.status()))
UnicodeEncodeError: 'ascii' codec can't encode character '\u2014' in position 84: ordinal not in range(128)
```

 After `sudo ua attach` is finished, same error will happen when issuing `sudo ua status`:

```
ubuntu@trusty:\~$ sudo ua status
    load_entry_point('ubuntu-advantage-tools==19.5', 'console_scripts', 'ubuntu-advantage')()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 486, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 516, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 426, in action_status
    print(ua_status.format_tabular(cfg.status()))
UnicodeEncodeError: 'ascii' codec can't encode character '\u2014' in position 84: ordinal not in range(128)
```

OddBloke:
When we detect that stdout's encoding isn't UTF-8, we replace \u2014 with -.

This is specifically targeted to fix #859, there may be more general changes to follow.

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -429,7 +429,16 @@ def action_status(args, cfg):
             status["expires"] = str(status["expires"])
         print(json.dumps(status))
     else:
-        print(ua_status.format_tabular(cfg.status()))
+        output = ua_status.format_tabular(cfg.status())
+        # Replace our Unicode dash with an ASCII dash if we aren't going to be
+        # writing to a utf-8 output; see
+        # https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/859
+        if (
+            sys.stdout.encoding is None
+            or "UTF-8" not in sys.stdout.encoding.upper()
+        ):
+            output = output.replace("\u2014", "-")
+        print(output)
     return 0
 
 
--- a/uaclient/tests/test_cli_status.py
+++ b/uaclient/tests/test_cli_status.py
@@ -2,6 +2,7 @@ import io
 import json
 import mock
 import socket
+import sys
 
 import pytest
 
@@ -24,12 +25,14 @@ See https://ubuntu.com/advantage
 
 ATTACHED_STATUS = """\
 SERVICE       ENTITLED  STATUS    DESCRIPTION
-cc-eal        no        —         Common Criteria EAL2 Provisioning Packages
-cis-audit     no        —         Center for Internet Security Audit Tools
-esm-infra     no        —         UA Infra: Extended Security Maintenance
-fips          no        —         NIST-certified FIPS modules
-fips-updates  no        —         Uncertified security updates to FIPS modules
-livepatch     no        —         Canonical Livepatch service
+cc-eal        no        {dash}         Common Criteria EAL2 Provisioning\
+ Packages
+cis-audit     no        {dash}         Center for Internet Security Audit Tools
+esm-infra     no        {dash}         UA Infra: Extended Security Maintenance
+fips          no        {dash}         NIST-certified FIPS modules
+fips-updates  no        {dash}         Uncertified security updates to FIPS\
+ modules
+livepatch     no        {dash}         Canonical Livepatch service
 
 Enable services with: ua enable <service>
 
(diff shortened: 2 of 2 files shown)


uaclient/cli.py
```diff
@@ -429,7 +429,16 @@ def action_status(args, cfg):
...
-        print(ua_status.format_tabular(cfg.status()))
+        output = ua_status.format_tabular(cfg.status())
+        # Replace our Unicode dash with an ASCII dash if we aren't going to be
+        # writing to a utf-8 output; see
+        # https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/859
+        if (
+            sys.stdout.encoding is None
+            or "UTF-8" not in sys.stdout.encoding.upper()
+        ):
+            output = output.replace("\u2014", "-")
```
blackboxsw: This may fall into the category of your PR description:  "This is specifically targeted to fix #859, there may be more general changes to follow."   

Food for thought: if we don't think that "\u2014" is an appropriate character sequence to show people in the console status output, shouldn't  we also replace this string in the status.json output too in general so that is is more human-friendly?

OddBloke: > This may fall into the category of your PR description: "This is specifically targeted to fix #859, there may be more general changes to follow."

Yes, though given the below I'm not sure we actually need more general changes.

> Food for thought: if we don't think that "\u2014" is an appropriate character sequence to show people in the console status output, shouldn't we also replace this string in the status.json output too in general so that it is more human-friendly?

The status value stored in `status.json` and (therefore) emitted when `--output json` is passed to `ua status` is _not_  a string containing the single Unicode character `\u2014` (i.e. —), it's the string "\u2014". It's already been encoded, specifically because `ensure_ascii` defaults to True for [`json.dump(s)`](https://docs.python.org/3/library/json.html#json.dump).

>  so that it is more human-friendly?

This is JSON output, it doesn't need to be human-friendly. (And, to be clear, we aren't avoiding emitting it in those other cases to be "human-friendly" except insofar as breaking someone's terminal isn't very friendly to that human.)

blackboxsw: I didn't realize as well that when rendering that json output in my terminal with utf-8 encoding it actually is converted back to the appropriate em-dash character. For instance: `ua status --format json` output will still contain the "\u2014" unicode character representation of the character but piping that to another tool such as jq or json_pp will result in  rendering the em-dash character "—". It makes potential script matches a bit more complex, but it's just an opinion and not a blocker to this branch especially because the status value of "—" is of dubious meaning anyway.  It's probably more likely people would be searching for strings enabled|disabled|n/a  as they aren't vague.


But here's an example of the concern, scripts that directly process `ua status --format json` ouput may incorrectly parse the status column if they are not aware of the differences. Yay unicode encoding issues.

root@dev-t:~# ua status --format json | grep —
root@dev-t:~# ua status --format json | jq "." | grep —
      "status": "—",
root@dev-t:~# ua status --format json | json_pp | grep —
Wide character in print at /usr/bin/json_pp line 81, <STDIN> chunk 1.
         "status" : "—",
root@dev-t:~#

OddBloke: This seems very clearly to be a distinct issue, as it's about the external usability of our experimental, internal-only JSON representation. Can you file one, please?

> not a blocker to this branch

So can I get an Approve?

blackboxsw: indeed.

blackboxsw:
Ok filed a followup issue for us to discuss related to whether json output should have rendered characters or an ascii string representation.  https://github.com/CanonicalLtd/ubuntu-advantage-client/issues/871

Issue #868, opened 2019-10-18
Noisy traceback logs on ua status on disconnected machine


blackboxsw:
The traceback referenced in ubuntu-advantage.log on connectivity errors is unhelpful for the following reasons:

- It doesn't properly report generic connectivity issues, but rather a symptom "[Errno -2] Name or service not known
- It's traceback is long, convoluted (with two traces of failures) making it very verbose and harder to find the actual point of failure
 - The log message doesn't match what the user sees at the console, so they won't know intuitively what to look for in the log files
- The  log doesn't actually report the url or service endpoint suffered from connectivty issues

Let's propose to log the same simple connectivity message error as seen by the user on the CLI, and the url in question as it'll be easier to read in logs what part of uaclient codebase failed in the actually failed.

Steps to reproduce:
```
cat > upgrade-uaclient.yaml <<EOF
#cloud-config
package_update: true
package_upgrade: true
apt_sources:
        - source: "deb [trusted=yes] http://ppa.launchpad.net/canonical-server/ua-client-daily/ubuntu $RELEASE main"
runcmd:
        - apt-get dist-upgrade --assume-yes --force-yes
EOF

lxc launch ubuntu-daily:trusty dev-t -c user.user-data="$(cat upgrade-uaclient.yaml )"
lxc exec dev-t ip link set down eth0
lxc exec dev-t ua status

cat /var/log/ubuntu-advantage.log



2019-10-21 17:33:27,754 - cli.py:(505) [ERROR]: [Errno -2] Name or service not known
Traceback (most recent call last):
  File "/usr/lib/python3.4/urllib/request.py", line 1182, in do_open
    h.request(req.get_method(), req.selector, req.data, headers)
  File "/usr/lib/python3.4/http/client.py", line 1125, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python3.4/http/client.py", line 1163, in _send_request
    self.endheaders(body)
  File "/usr/lib/python3.4/http/client.py", line 1121, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python3.4/http/client.py", line 951, in _send_output
    self.send(msg)
  File "/usr/lib/python3.4/http/client.py", line 886, in send
    self.connect()
  File "/usr/lib/python3.4/http/client.py", line 1260, in connect
    super().connect()
  File "/usr/lib/python3.4/http/client.py", line 863, in connect
    self.timeout, self.source_address)
[...]

PR #872, opened 2019-10-22
log: log connectivity error message and url instead of traceback

blackboxsw:
In the event of unhandled UrlErrors from ua-client operations log
the same connectivity error message seen to the console and the error
string instead of a full traceback in the logs.

Only log the exception traceback if the UrlError doesn't have a defined
url.

Fixes: #868

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -511,7 +511,12 @@ def main_error_handler(func):
             sys.exit(1)
         except util.UrlError as exc:
             with util.disable_log_to_console():
-                logging.exception(exc)
+                msg_args = {"url": exc.url, "error": exc}
+                if exc.url:
+                    msg_tmpl = ua_status.LOG_CONNECTIVITY_ERROR_WITH_URL_TMPL
+                else:
+                    msg_tmpl = ua_status.LOG_CONNECTIVITY_ERROR_TMPL
+                logging.exception(msg_tmpl.format(**msg_args))
             print(ua_status.MESSAGE_CONNECTIVITY_ERROR, file=sys.stderr)
             sys.exit(1)
         except exceptions.UserFacingError as exc:
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -120,6 +120,10 @@ MESSAGE_APT_UPDATING_LISTS = "Updating package lists"
 MESSAGE_CONNECTIVITY_ERROR = """\
 Failed to connect to authentication server
 Check your Internet connection and try again"""
+LOG_CONNECTIVITY_ERROR_TMPL = MESSAGE_CONNECTIVITY_ERROR + ". {error}"
+LOG_CONNECTIVITY_ERROR_WITH_URL_TMPL = (
+    MESSAGE_CONNECTIVITY_ERROR + ". Failed to access URL: {url}. {error}"
+)
 MESSAGE_NONROOT_USER = "This command must be run as root (try using sudo)"
 MESSAGE_ALREADY_DISABLED_TMPL = """\
 {title} is not currently enabled\nSee: sudo ua status"""
--- a/uaclient/tests/test_cli.py
+++ b/uaclient/tests/test_cli.py
@@ -224,7 +224,6 @@ class TestMain:
         error_log = caplog_text()
         assert "Traceback (most recent call last):" in error_log
 
-    @pytest.mark.parametrize("caplog_text", [logging.ERROR], indirect=True)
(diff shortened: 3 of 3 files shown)


uaclient/tests/test_cli.py
```diff
@@ -279,7 +280,13 @@ def test_url_error_handled_gracefully(
         assert "" == out
         assert "{}\n".format(status.MESSAGE_CONNECTIVITY_ERROR) == err
         error_log = caplog_text()
-        assert " [Errno -2] Name or service not known" in error_log
+        expected_log = "Check your internet connection and try again. "
```
OddBloke: A nit, but we could just have the expected log message (or message suffix) as parameters to the test.

uaclient/status.py
```diff
@@ -119,7 +119,7 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_APT_UPDATING_LISTS = "Updating package lists"
 MESSAGE_CONNECTIVITY_ERROR = """\
 Failed to connect to authentication server
-Check your Internet connection and try again"""
+Check your internet connection and try again"""
```
OddBloke: I would normally expect Internet to be capitalised. Uncapitalised usage _is_ acceptable, but the UX UA doc also has it capitalised. Any objection to changing it back?

uaclient/cli.py
```diff
@@ -511,7 +511,14 @@ def wrapper(*args, **kwargs):
             sys.exit(1)
         except util.UrlError as exc:
             with util.disable_log_to_console():
-                logging.exception(exc)
+                msg = ua_status.MESSAGE_CONNECTIVITY_ERROR
+                if exc.url:
+                    msg += ". Failed to access url: {url}. {error}".format(
```
OddBloke: s/url/URL/

uaclient/cli.py
```diff
@@ -511,7 +511,14 @@ def wrapper(*args, **kwargs):
...
+                        url=exc.url, error=exc
+                    )
+                else:
+                    msg += ". {error}".format(error=exc)
```
OddBloke: Should we have these as two different templates in `status.py` instead of concatenating?

uaclient/status.py
blackboxsw: strange. I guess it just looked wrong to me. Changing "It" back.

uaclient/cli.py

uaclient/cli.py
blackboxsw: Sure, let's do it.

uaclient/tests/test_cli.py
blackboxsw: works for me.

uaclient/cli.py
```diff
@@ -511,7 +511,12 @@ def wrapper(*args, **kwargs):
             sys.exit(1)
         except util.UrlError as exc:
             with util.disable_log_to_console():
-                logging.exception(exc)
+                msg_args = {"url": getattr(exc, "url", None), "error": exc}
```
OddBloke: Either this doesn't need to use `getattr` or the next line _does_. Looking at `util.UrlError`, we will always set `self.url` in `__init__`, so we can just make this `exc.url`.

uaclient/status.py
```diff
@@ -120,6 +120,10 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_CONNECTIVITY_ERROR = """\
 Failed to connect to authentication server
 Check your Internet connection and try again"""
+LOG_CONNECTIVITY_ERROR_TMPL = MESSAGE_CONNECTIVITY_ERROR + ". {error}"
```
OddBloke: Very much a nit, but should we use a newline instead of a `. ` here to separate the parts of the message?

uaclient/cli.py
blackboxsw: riiight. good one.

uaclient/status.py
blackboxsw: I feel like having a multiline log is hard for parsing of errors. Generally when I'm doing a quick triage/validation I expect error or warning log context  to all be on a single line (tracebacks are the only exception to this).

I was also a little uneasy about us prefacing the full MESSAGE_CONNECTIVITY_ERROR  in the error log as it already has a newline therfore making it i a bit harder to quickly grep the full context of just error messages out of the log. But in this case I thought having the same error message as seen on the console was a better breadcrumb for people trying to triage what happened in the log at the time of the error they saw in their console.

PR #875, opened 2019-10-22
d/changelog: 19.6 entries

panlinux:
I looked at the git log from 19.5.1 to 7f277d09a48dc073fda6a91f2000a6275d4fc962 and these are the important changes I think should be mentioned in `debian/changelog`.
Once this lands, there should be another MP opening up 19.7 for development, with the ubuntu release name field set to `UNRELEASED`. I don't want to do that together here because I want to be able to use a 19.6 tag in the repository for the basis of building the trusty package.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,16 @@
-ubuntu-advantage-tools (19.6) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (19.6) focal; urgency=medium
 
-  * New upstream release 
+  * New upstream release. Main changes:
+    - drop SSO interactive login support
+    - d/control: no longer depend on pymacaroons, which was only needed for
+      the SSO interactive login support
+    - drop keyrings for services not supported in trusty: cc-eal, fips,
+      fips-updates, cis audit
+    - make sure /var/lib/ubuntu-advantage/private has 0700 perms
+    - rename esm to esm-infra. Also handle upgrades
+    - don't unecessarily remove config files that are already handled by dpkg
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -1,6 +1,16 @@
-ubuntu-advantage-tools (19.6) UNRELEASED; urgency=medium
-
-  * New upstream release 
+ubuntu-advantage-tools (19.6) focal; urgency=medium
+
+  * New upstream release. Main changes:
+    - drop SSO interactive login support
+    - d/control: no longer depend on pymacaroons, which was only needed for
```
blackboxsw: Also could add
- unattached machine will now report availability of services
- dropped dependency on gpg utilities for apt key publication
- command line messages and UI rework

panlinux: `dropped dependency on gpg utilities for apt key publication` was that ever in 19.5.1?

powersj: I don't think it was

panlinux: What does `command line messages and UI rework` mean? (playing devil's advocate here, i.e., someone who is not familiar with ua reading this). Did the command-line interface change? Will existing script break?

powersj: Various output messages were changed and status output was modified.

blackboxsw: good point Andreas, this could break existing scripts if anyone used ua status --format json

powersj: no one is using the new client yet; this cannot break "existing scripts"

panlinux:
Rebased on top of master

PR #880, opened 2019-10-22
d/tests/usage: Update tests

powersj:
This updates the dep8 tests to call new commands and new subcommands
that will not reach outside of the network.

### DIFF
--- a/debian/tests/usage
+++ b/debian/tests/usage
@@ -2,5 +2,5 @@
 
 set -ex
 
-ubuntu-advantage --help | grep --silent services
-ubuntu-advantage status | grep --silent 'This machine is not attached'
+ua --help | grep --silent services
+ua version


debian/tests/usage
```diff
@@ -2,5 +2,5 @@
 
 set -ex
 
-ubuntu-advantage --help | grep --silent services
-ubuntu-advantage status | grep --silent 'This machine is not attached'
+ua --help | grep --silent services
+ua version
```
OddBloke: Nit: Should we test both spellings of the command?

powersj: All our docs going forward are going to use "ua". I treat "ubuntu-advantage" as a nice to have for previous users, even though they will have to change their processes anyway. So I'm not heavily inclined to ensure those work.

Thoughts?

OddBloke: I don't feel at all strongly about it, but it won't be particularly nice to have for them if it stops working. ;-)

Issue #790, opened 2019-10-02
Display custom descriptions from the backend server


OddBloke:
"Custom descriptions will be received from the backend server. When these are present based on different scenarios they should be presented to the user."

(This is split out from #719.)

OddBloke:
It's not 100% clear that this is _needed_ for 19.5.1, but it's split out from an issue that is in that milestone so I'm assigning it there just in case.

Issue #760, opened 2019-09-23
`ua status` attempts a call to the contract server even when attached


OddBloke:
Specifically, it fetches the resources endpoint. This is, at the very least, causing test failures during package builds (where we don't have access to the endpoint).

powersj:
It is not clear to me what the issue is here anymore.

For unattached system, we have to reach out to the contract backend to get the list of supported services and to be able to say what is available or not.

For attached systems, maybe we need to do something better? Is this still the behavior?

Please confirm, next steps here.

powersj:
#890 had comments that fixes this, however if it is not feel free to re-open with next steps.

Issue #499, opened 2019-05-14
Tests failing in master


OddBloke:
See https://travis-ci.org/CanonicalLtd/ubuntu-advantage-client/jobs/531550157

PR #890, opened 2019-10-24
status: custom descriptions for n/a service status

blackboxsw:
The response from machine attach (machine-token.json) provides an
availableResources list.  Any item in availableResources which is not
"available" will provide a "description" that should be surfaced as a
user-facing custom description for any n/a service.

The command `ua status` should surface these custom descriptions in the
DESCRIPTION column for any n/a service.

In the absence of a custom status "description" for an inapplicable
service, use the static entitlement.description.

Fixes: #790

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -245,14 +245,16 @@ class UAConfig:
 
     def _attached_service_status(
         self, ent, inapplicable_resources
-    ) -> "Dict[str, str]":
+    ) -> "Dict[str, Optional[str]]":
         details = ""
+        description_override = None
         contract_status = ent.contract_status()
         if contract_status == status.ContractStatus.UNENTITLED:
             ent_status = status.UserFacingStatus.UNAVAILABLE
         else:
             if ent.name in inapplicable_resources:
                 ent_status = status.UserFacingStatus.INAPPLICABLE
+                description_override = inapplicable_resources[ent.name]
             else:
                 ent_status, details = ent.user_facing_status()
         return {
@@ -261,6 +263,7 @@ class UAConfig:
             "entitled": contract_status.value,
             "status": ent_status.value,
             "statusDetails": details,
+            "description_override": description_override,
         }
 
     def _attached_status(self) -> "Dict[str, Any]":
@@ -286,11 +289,11 @@ class UAConfig:
             )
 
         resources = get_available_resources(self)
-        inapplicable_resources = [
-            resource["name"]
+        inapplicable_resources = {
+            resource["name"]: resource.get("description")
             for resource in sorted(resources, key=lambda x: x["name"])
             if not resource["available"]
-        ]
+        }
 
         for ent_cls in ENTITLEMENT_CLASSES:
             ent = ent_cls(self)
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -214,11 +214,15 @@ def format_tabular(status: "Dict[str, Any]") -> str:
     content = [STATUS_HEADER]
     for service_status in status["services"]:
         entitled = service_status["entitled"]
+        descr_override = service_status["description_override"]
+        description = (
+            descr_override if descr_override else service_status["description"]
+        )
         fmt_args = {
             "name": service_status["name"],
             "entitled": colorize(entitled),
             "status": colorize(service_status["status"]),
-            "description": service_status["description"],
+            "description": description,
         }
         content.append(STATUS_TMPL.format(**fmt_args))
     content.append("\nEnable services with: ua enable <service>\n")
--- a/uaclient/tests/test_status.py
+++ b/uaclient/tests/test_status.py
@@ -140,3 +140,28 @@ class TestFormatTabular:
             status_dict = status_dict_unattached
 
         assert not format_tabular(status_dict).startswith("\n")
+
+    @pytest.mark.parametrize(
+        "description_override, uf_status, uf_descr",
+        (
+            ("", "n/a", "Common Criteria EAL2 default descr"),
+            ("Custom descr", "n/a", "Custom descr"),
+            ("Custom call to action", "enabled", "Custom call to action"),
+        ),
+    )
+    def test_custom_descr(
+        self, description_override, uf_status, uf_descr, status_dict_attached
+    ):
+        """Services can provide a custom call to action if present."""
+        default_descr = "Common Criteria EAL2 default descr"
+        status_dict_attached["services"] = [
+            {
+                "name": "cc-eal",
+                "description": default_descr,
+                "available": "no",
+                "status": uf_status,
+                "entitled": True,
+                "description_override": description_override,
+            }
+        ]
+        assert uf_descr in format_tabular(status_dict_attached)
--- a/uaclient/tests/test_config.py
+++ b/uaclient/tests/test_config.py
@@ -28,8 +28,8 @@ KNOWN_DATA_PATHS = (("machine-token", "machine-token.json"),)
 M_PATH = "uaclient.entitlements."
 
 
-RESP_ALL_RESOURCES_AVAILABLE = [
-    {"name": name, "available": True} for name in ENTITLEMENT_CLASS_BY_NAME
+NO_RESOURCES_ENTITLED = [
+    {"type": name, "entitled": False} for name in ENTITLEMENT_CLASS_BY_NAME
 ]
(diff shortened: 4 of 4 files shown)


uaclient/config.py
```diff
@@ -247,12 +247,14 @@ def _attached_service_status(
         self, ent, inapplicable_resources
     ) -> "Dict[str, str]":
         details = ""
+        userFacingDetails = ""
```
OddBloke: Huh, surprised that our linting tools aren't picking this up as an invalid variable name. A nit, but this would be more consistent as `user_facing_details`.

uaclient/config.py
```diff
@@ -285,12 +287,12 @@ def _attached_status(self) -> "Dict[str, Any]":
                 contractInfo["effectiveTo"], "%Y-%m-%dT%H:%M:%SZ"
             )
 
-        resources = get_available_resources(self)
-        inapplicable_resources = [
-            resource["name"]
+        resources = self.machine_token.get("availableResources")
```
OddBloke: Does this also fix #760?

blackboxsw: This does fix #760 too... but we will likely have to sort this feature on the contract server as a result of this shift. https://github.com/CanonicalLtd/ua-contracts/issues/499

OddBloke: We could potentially switch things so that we display the custom message iff we determine based on the local system that an entitlement is inapplicable? That way we don't need to update the contract server's state.

The full solution for that ua-contracts issue would drive non-trivial client changes, I think? That's the only reason I'm wondering if there are more minimal changes we can make in the short-term to get this landed and usable in trusty.

uaclient/config.py
blackboxsw: fixed. agreed, /me got wrapped up in our json field names.

uaclient/config.py
blackboxsw: Ok, I've pushed an update but still needs unit tests for both paths. We can work around this shortcoming, but for ua-contracts #499 the refreshed token doesn't even contain availableResources at all so we'll have to fallback to get_avail_resources until 499 gets a fix. And you are right that ua-contracts change is potentially not trivial to uaclient as we'll have to pass more form data to the route

blackboxsw:
> I've only performed a partial review because I'm not convinced this PR is actually complete. Are we missing some `status.py` changes to modify `format_tabular` to display the new descriptions?

You were right @OddBloke I hadn't wired it to the actual format_tabular function. I've fixed that and added unit test coverage for that user facing output.

OddBloke:
@blackboxsw Did you mean to re-request review here?

uaclient/config.py
```diff
@@ -285,12 +288,16 @@ def _attached_status(self) -> "Dict[str, Any]":
                 contractInfo["effectiveTo"], "%Y-%m-%dT%H:%M:%SZ"
             )
 
-        resources = get_available_resources(self)
-        inapplicable_resources = [
-            resource["name"]
+        resources = self.machine_token.get("availableResources", [])
```
OddBloke: If the contract server isn't yet producing `availableResources`, how can we be sure that the logic handling it is correct? I'm concerned that there may be some difference between our expectations and reality, which we may only discover once this branch has landed (and, potentially, released which would obviously be much worse).

If we don't have reason to be very confident in the logic here (please tell me if we do!), then I think we should wait until the contract server is actually producing this content to write code that relies on it.

uaclient/status.py
```diff
@@ -214,11 +214,13 @@ def format_tabular(status: "Dict[str, Any]") -> str:
     content = [STATUS_HEADER]
     for service_status in status["services"]:
         entitled = service_status["entitled"]
+        uf_status = service_status["userFacingDetails"]
```
OddBloke: `uf_status` is a very confusing variable name for this. `service_status["status"]` is what we refer to as user-facing status everywhere else.

uaclient/tests/test_status.py
```diff
@@ -140,3 +140,28 @@ def test_no_leading_newline(
...
+        """Services can provide a custom call to action if present."""
+        default_descr = "Common Criteria EAL2 default descr"
+        status_dict_attached["services"] = [
+            {
+                "name": "cc-eal",
+                "description": default_descr,
+                "available": "no",
+                "status": uf_status,
+                "entitled": True,
+                "userFacingDetails": user_facing_details,
```
OddBloke: Seeing this dict written out like this makes me realised that `userFacingDetails` is inconsistently named compared to the other keys. (For starters, practically by definition, anything in this dict is user-facing. :grin:)

Could we find a different key name?

uaclient/config.py
blackboxsw: @OddBloke, right now the contract server **is** surfacing no availableResource key from the initial attached machine-token from the route **/v1/context/machines/token**.  But, that availableResources key is absent from the refreshed machine-token.json which we get from **/v1/resources/{resource}/context/machines/{machine}**.

The content from the initial attach 
```
   "availableResources" : [
      {
         "name" : "cc-eal",
         "available" : false
      },
      {
         "name" : "esm-infra",
         "available" : true
      },
      {
         "available" : false,
         "name" : "fips"
      },
      {
         "name" : "fips-updates",
         "available" : false
      },
      {
         "name" : "livepatch",
         "available" : true
      },
      {
         "available" : true,
         "name" : "support"
      }
   ],
```

hich gives us a traceback. So, we can't land this branch to rely on

uaclient/status.py
blackboxsw: calling it uf_details.

uaclient/tests/test_status.py
blackboxsw: good point. how about customDescription.

uaclient/config.py
OddBloke: It's still not clear to me _why_ we want to land this code before it can possibly do anything; if the contract server never implements the thing we've asked for, then we're just introducing dead code. Do we lose anything by deferring landing this specific part of this until the contract server is providing the interface we want to work against?

(The existence of an identically-named key elsewhere means we can _assume_ it will be the same, but there's no guarantee that the implementations won't be different on the backend. For example, your initial attach example doesn't include any `description` elements, so is that _definitely_ the same format as we would get? I don't think we can say that positively without seeing the (not yet completed) implementation in the refresh route in the contract server.)

I think I'd like to see us split this PR up into two pieces: one which relies on the routes we have today (and implements the display logic), and then another which drops the dependency on `get_available_resources` once we have an alternative. Is that reasonable?

uaclient/tests/test_status.py
OddBloke: I'd stick with underscored key names for a couple of reasons: firstly, I find them more aesthetically pleasing but secondly, and this might actually be a good reason, it'll give us a heuristic for knowing if we're dealing with JSON we generated vs JSON returned from the contract server.

`custom_description` sounds good to me, though I wonder if we can be more specific than "custom". Perhaps `description_override` is more descriptive of how we use it? Naming things is hard, I'll leave it to your discretion.

uaclient/config.py
blackboxsw: This code path is exercised on initial attach because initial attach machine-token.json contains availableResources so I think it's reasonable to leave it it. And, we still need the get_available_resources call for the case when we've run "ua enable <anything>" or "ua refresh" where the refreshed machine-token.json doesn't contain available_resources. I'd like the improved performance on ua attach <token> by avoiding this unnecessary v1/resources GET since we don't need it for the default install/attach case.

OddBloke: I'm still not fully sold on this. I hadn't realised that the contract server is _already_ inconsistent in its machine token responses when I wrote the above, but I think that further supports my position of "we shouldn't rely on them becoming consistent in future".

Could you split this PR up into the two separate changes it contains so we can continue this discussion without blocking most of this change from landing? (Or at least present some reason why we _can't_ split it up?)

uaclient/config.py
```diff
@@ -247,12 +247,14 @@ def _attached_service_status(
         self, ent, inapplicable_resources
     ) -> "Dict[str, str]":
         details = ""
+        description_override = ""
```
OddBloke: Would the "no override" case be better represented by `None`?

blackboxsw: None and done ;)

powersj:
uaclient/config.py:266: error: Dict entry 5 has incompatible type "str": "Optional[Any]"; expected "str": "str"

PR #892, opened 2019-10-25
sru: add sru subdirectory to track SRU verification scripts

blackboxsw:
Add a collection of SRU manual verification scripts to an ./sru subdirectory.
We can leverage these scripts in filing SRU results as well as having a tome of scripts for reference when we have other SRU verfications to exercise.


blackboxsw:
> I hate to nit, but to make this easy to view and nice for the next guy, can you format these in markdown?
> 
> Basically create headers with `#` or `##`, make the items bullet lists, and then put the script/code in actually code blocks with the "`" character.

Thanks @powersj for that. Changed all scripts to markdown

Issue #873, opened 2019-10-22
[minor] Don't ask to use sudo if I'm root already


panlinux:
Similarly to what `command-not-found` does, if instructions need to be given to a regular user, `sudo` should be shown. Otherwise, as the user is already root, there is no need for `sudo`.
Example with `command-not-found`:
```
andreas@nsnx:~$ fgfs

Command 'fgfs' not found, but can be installed with:

sudo apt install flightgear

andreas@nsnx:~$ sudo su -
[sudo] password for andreas: 
root@nsnx:~# fgfs

Command 'fgfs' not found, but can be installed with:

apt install flightgear

root@nsnx:~# 
```
In the case of `ua`, the scenario is disabling a service that is already disabled. I believe this message shows up in other cases as well, but this is the one I found:
```
root@nsnx:~# ua disable esm-infra
ESM Infra is not currently enabled.
See: sudo ua status
```
In the above case, since I'm root already, there is no need to use `sudo`.
In the case below, it's correct to suggest to use sudo, since I'm running the disable command with sudo:
```
andreas@nsnx:~$ ua disable esm-infra
This command must be run as root (try using sudo)
andreas@nsnx:~$ sudo ua disable esm-infra
ESM Infra is not currently enabled.
See: sudo ua status
andreas@nsnx:~$ 
```
Whether sudo is being used or not is possible to detect via an environment variable:
```
andreas@nsnx:~$ sudo env|grep SUDO_USER
SUDO_USER=andreas
```

PR #896, opened 2019-10-28
features: make it more explicit that we use sudo for testing

OddBloke:
Running the command with sudo vs as a true root user may start to cause
differences in light of #873, so let's be explicit that we're using
sudo.


OddBloke:
I just realised why this was sitting around without being proposed: `lxc exec`s default behaviour means that we're running _everything_ as root, so this needs some additional work.

PR #901, opened 2019-10-30
Integration tests: actually run non-root tests as non-root

OddBloke:
`lxc exec` runs as root by default, so we modify the way we call it to specify UID 1000. This also updates other code that uses `lxc_exec` to add `sudo` if needed, and fixes the feature files to reflect the actual non-root output.


OddBloke:
This is seeing behave failures in Travis which I'm currently trying to work out.

PR #902, opened 2019-10-30
Integration tests: introduce support for image reuse

OddBloke:
This should speed up local iteration substantially. See the README.md changes for further details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,42 @@ through [the behave
 tutorial](https://behave.readthedocs.io/en/latest/tutorial.html) to get
 an idea of how it works, and how tests are written.)
 
+#### Iterating Locally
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -60,6 +60,40 @@ through [the behave
...
+```
+
+You can then reuse that image by plugging its name into your next test
+run, like so:
+
+```sh
+UACLIENT_BEHAVE_REUSE_IMAGE=behave-image-1572443113978755 behave
+```
+
+If you've done this correctly, you should see test execution start
```
blackboxsw: Could note you will see the config option reporting " reuse_image = <YOUR_IMAGE_NAME>"

PR #908, opened 2019-10-30
Demo/aws premium

blackboxsw:
Sync tip of master to get all ux and behavior updates into ua premium demo branch.

Here's the full diff against master: https://paste.ubuntu.com/p/9Pv39j532t/

highlights from pastebin:
 - line 144: new attach-premium CLI subcommand added
 - line 286: get_cloud_type: use cloud-init's cloud-id if present, or extract it from /var/lib/cloud/data/result.json on trusty (which doesn't have cloud-id)
 - line 337: new Contract.request_premium_aws_attach to take pkcs7 data from aws and send it to ua-contracts
 - line 372: new retry decorator for retrying the aws metadata service to obtain pkcs7 content if metadata not yet up
 - line 406: new upstart script that runs ua attach-premium subcommand


uaclient/cli.py
```diff
@@ -29,140 +19,221 @@
...
     if not parser:
         parser = argparse.ArgumentParser(
-            prog='attach',
-            description=('Attach this machine to an existing Ubuntu Advantage'
-                         ' support subscription'),
-            usage=usage)
+            prog="attach-premium",
+            description=(
+                "For premium Ubuntu images, automatically attach"
+                " Ubuntu Advantage support subscription"
```
powersj: Update description to: "Automatically enable Ubuntu Advantage on a premium Ubuntu image"

uaclient/cli.py
```diff
@@ -197,155 +272,289 @@ def _perform_enable(entitlement_name: str, cfg: config.UAConfig, *,
...
     if os.getuid() != 0:
         raise exceptions.NonRootUserError()
-    contract_token = args.token
-    if not contract_token:
-        print('No valid contract token available')
-        return 1
+    cloud_type = identity.get_cloud_type()
+    if cloud_type != "aws":
+        print(
+            "No premium image support for cloud '%s'. Use ua attach <token>"
```
powersj: Change to:
"Premium image support is not supported on '%s'
For more information see: https://ubuntu.com/advantage"

Ensure it is on two lines.

uaclient/cli.py
```diff
@@ -197,155 +272,289 @@ def _perform_enable(entitlement_name: str, cfg: config.UAConfig, *,
...
+        help="attach this machine to an Ubuntu Advantage subscription",
+    )
     attach_parser(parser_attach)
     parser_attach.set_defaults(action=action_attach)
+    parser_attach_premium = subparsers.add_parser(
+        "attach-premium",
+        help=(
+            "for premium ubuntu images, attach this machine to an ubuntu"
+            " advantage subscription"
+        ),
```
powersj: Update help to: "automatically enable Ubuntu Advantage on a premium Ubuntu image"

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,51 @@
...
+
+def cloud_instance_factory() -> clouds.UAPremiumCloudInstance:
+    from uaclient.clouds import aws
+
+    cloud_instance_map = {"aws": aws.UAPremiumAWSInstance}
+
+    cloud_type = get_cloud_type()
+    if not cloud_type:
+        raise exceptions.UserFacingError(
+            "Could not determine cloud platform type. Unable to auto-attach"
```
powersj: Update exception to say the following on two lines:

"Unable to determine premium image platform support.
For more information see: https://ubuntu.com/advantage"

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+    cls = cloud_instance_map.get(cloud_type)
+    if not cls:
+        raise exceptions.UserFacingError(
+            "No UAPremiumCloudInstance class available for cloud type '%s'"
+            % cloud_type
```
powersj: Update exception to say the following on two lines:

"Platform not supported by premium image.
For more information see: https://ubuntu.com/advantage"

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+    instance = cls()
+    if not instance.is_viable:
+        raise exceptions.UserFacingError(
+            "This vm is not a viable Ubuntu Advantage Premium image"
+            'on cloud "%s"' % cloud_type
```
powersj: Update exception to say the following on two lines:

"Platform not supported by premium image.
For more information see: https://ubuntu.com/advantage"

upstart/ua-premium.conf
```diff
@@ -0,0 +1,14 @@
+# ua-premium - Ubuntu Advantage Premium attach
+
+# This task is run on reboot to adapt to potential change in
+# kernels/environment across boot and update Ubuntu Advantage status
+# information.
+
```
powersj: I am not asking for changes, but trying to understand why we try to attach literally on every boot. shouldn't we check to see if we are attached first?

upstart/ua-premium.conf
```diff
@@ -0,0 +1,14 @@
...
+description    "Ubuntu Advantage Premium image attach"
```
powersj: Change description to: "Ubuntu Advantage premium attach"

upstart/ua-premium.conf
```diff
@@ -0,0 +1,14 @@
...
+
+start on runlevel [2345]
+
+task
+# Refresh the status of all Ubuntu Advantage services
```
powersj: this comment doesn't seem to do what you say. We are attaching, not refreshing the status.

systemd/ua-premium.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Ubuntu Advantage Premium attach
```
powersj: "Ubuntu Advantage premium attach" to match upstart

uaclient/cli.py
blackboxsw: done

upstart/ua-premium.conf
blackboxsw: calls to `ua attach-premium` has a decorator that checks config.is_attached validate whether we're attached each call you make. it exits 0 in this case to indicate succesful end state (attached). I thought this was clearer than:
`ua status --format=json | grep -q '"attached": true'` || ua attach-premium

But maybe the above makes more sense and introduces less upstart log noise.

blackboxsw: Hrm, also I don't think I can express compound exec commands in upstart. I certainly can't in systemd.   If we surfaced a command that would exit 1 on unattached we could use something like
ExecCondtion=ua is-attached
ExecStart=ua attach-premium

blackboxsw: Additionally we could add a param --silent to attach-premium to avoid adding log messages about machine already attached (because we do noop if we are already attached)

upstart/ua-premium.conf
blackboxsw: reworded.

PR #912, opened 2019-11-09
status: dynamic available status now from refreshed machine-token

blackboxsw:
For attached machines only.

When `ua enable <SERVICE>` or `ua refresh` is called, the client
now POSTs updated arch, kernel and OS information to the contracts
server. The contract server response will contain an updated
'availableResources' key which updates the 'status' and 'description'
columns for each applicable/inapplicable service.

This is a followup branch iteratively improving the behavior introduced
by fixing #790 to add dynamic description content from the backend
server. Now uaclient performs less API calls by avoiding a round trip to
contracts.canonical.com/v1/resources to obtain updated status
information.

There will need to be a single followup branch that will address a  gap in this behavior (#913): 
On trusty vm with stock kernel:
```
sudo ua attach
sudo apt-get install --install-recommends linux-generic-lts-xenial  
sudo reboot             # boot into HWE kernel
;   # machine-token not yet refreshed, so any ua status will still report:
;  #  livepatch     yes       n/a       Available with the HWE kernel
sudo ua status
```

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -288,7 +288,9 @@ class UAConfig:
                 contractInfo["effectiveTo"], "%Y-%m-%dT%H:%M:%SZ"
             )
 
-        resources = get_available_resources(self)
+        resources = self.machine_token.get("availableResources")
+        if not resources:
+            resources = get_available_resources(self)
         inapplicable_resources = {
             resource["name"]: resource.get("description")
             for resource in sorted(resources, key=lambda x: x["name"])
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -16,7 +16,7 @@ except ImportError:
 API_ERROR_INVALID_TOKEN = "invalid token"
 API_ERROR_MISSING_INSTANCE_INFORMATION = "missing instance information"
 API_V1_CONTEXT_MACHINE_TOKEN = "/v1/context/machines/token"
-API_V1_TMPL_CONTEXT_MACHINE_TOKEN_REFRESH = (
+API_V1_TMPL_CONTEXT_MACHINE_TOKEN_UPDATE = (
     "/v1/contracts/{contract}/context/machines/{machine}"
 )
 API_V1_RESOURCES = "/v1/resources"
@@ -83,11 +83,11 @@ class UAContractClient(serviceclient.UAServiceClient):
         """
         if not machine_id:
             machine_id = util.get_machine_id(self.cfg.data_dir)
-        os = util.get_platform_info()
-        arch = os.pop("arch")
+        platform = util.get_platform_info()
+        arch = platform.pop("arch")
         headers = self.headers()
         headers.update({"Authorization": "Bearer {}".format(contract_token)})
-        data = {"machineId": machine_id, "architecture": arch, "os": os}
+        data = {"machineId": machine_id, "architecture": arch, "os": platform}
         machine_token, _headers = self.request_url(
             API_V1_CONTEXT_MACHINE_TOKEN, data=data, headers=headers
         )
@@ -154,7 +154,7 @@ class UAContractClient(serviceclient.UAServiceClient):
         )
         return resource_access
 
-    def request_machine_token_refresh(
+    def request_machine_token_update(
         self, machine_token, contract_id, machine_id=None
     ):
         """Request machine token refresh from contract server.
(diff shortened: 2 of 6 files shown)


uaclient/contract.py
```diff
@@ -149,14 +149,17 @@ def request_machine_token_refresh(
...
+        arch = os.pop("arch")
         if not machine_id:
             machine_id = util.get_machine_id(self.cfg.data_dir)
         headers = self.headers()
         headers.update({"Authorization": "Bearer {}".format(machine_token)})
         url = API_V1_TMPL_CONTEXT_MACHINE_TOKEN_REFRESH.format(
             contract=contract_id, machine=machine_id
         )
-        response, headers = self.request_url(url, headers=headers)
+        data = {"machineId": machine_id, "architecture": arch, "os": os}
```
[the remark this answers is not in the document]
blackboxsw: Our since our contract client provides content-type': 'application/json', serviceclient.py  json.dumps the dict structure and encodes it as a utf-8 bytestring which it sends to the ua-contracts backend:
 URL [POST]: https://contracts.staging.canonical.com/v1/context/machines/token, headers: {'content-type': 'application/json', 'accept': 'application/json', 'user-agent': 'UA-Client/19.6~646~git.96620f1~ubuntu14.04.1', 'Authorization': 'Bearer <REDACTED>'}, data: b'{"machineId": "<REDACTED>", "architecture": "x86_64", "os": {"distribution": "Ubuntu", "series": "trusty", "kernel": "4.4.0-148-generic", "type": "Linux", "version": "14.04 LTS (Trusty Tahr)", "release": "14.04"}}'

uaclient/config.py
```diff
@@ -288,7 +288,9 @@ def _attached_status(self) -> "Dict[str, Any]":
                 contractInfo["effectiveTo"], "%Y-%m-%dT%H:%M:%SZ"
             )
 
-        resources = get_available_resources(self)
+        resources = self.machine_token["availableResources"]
+        if not resources:
```
OddBloke: When would we take this code path?

uaclient/contract.py
```diff
@@ -171,10 +171,13 @@ def request_machine_token_refresh(
             machine_id = util.get_machine_id(self.cfg.data_dir)
         headers = self.headers()
         headers.update({"Authorization": "Bearer {}".format(machine_token)})
-        url = API_V1_TMPL_CONTEXT_MACHINE_TOKEN_REFRESH.format(
+        url = API_V1_TMPL_CONTEXT_MACHINE_TOKEN_UPDATE.format(
             contract=contract_id, machine=machine_id
         )
-        response, headers = self.request_url(url, headers=headers)
+        platform = util.get_platform_info()
```
OddBloke: This looks like a copy/paste of the code two hunks above here; should we refactor?

uaclient/config.py
blackboxsw: only on current production which hasn't updated yet

uaclient/contract.py
blackboxsw: sure we can add a _get_platform_header private method to surface those keys

uaclient/config.py
OddBloke: OK, that makes sense. I think those cases will hit a `KeyError` on the above line as things stand?

blackboxsw: ohhh right, those will hit a key error as things stand good catch.

blackboxsw: Fixed.

uaclient/contract.py
blackboxsw: Make that _get_platform_data instead. done.

uaclient/config.py
```diff
@@ -288,7 +288,9 @@ def _attached_status(self) -> "Dict[str, Any]":
                 contractInfo["effectiveTo"], "%Y-%m-%dT%H:%M:%SZ"
             )
 
-        resources = get_available_resources(self)
+        resources = self.machine_token.get("availableResources", [])
```
OddBloke: A total nit, but no need to instantiate a list here, the value won't be used if we go down that path. Look how much more efficient using `None` is:

```
$ python3 -m timeit '{}.get("foo", [])'
5000000 loops, best of 5: 75.1 nsec per loop
$ python3 -m timeit '{}.get("foo", None)'
5000000 loops, best of 5: 60.5 nsec per loop
```

:joy:

uaclient/contract.py
```diff
@@ -171,15 +169,24 @@ def request_machine_token_refresh(
...
+        data = self._get_platform_data(machine_id)
+        response, headers = self.request_url(url, headers=headers, data=data)
         if headers.get("expires"):
             response["expires"] = headers["expires"]
         self.cfg.write_cache("machine-token", response)
         return response
 
+    def _get_platform_data(self, machine_id):
+        """"Return a dict of platform-relateddata for contract requests"""
+        if not machine_id:
```
OddBloke: This logic exists in both of the callsites, we don't need to replicate it in all three functions.

blackboxsw: fixed.

Issue #911, opened 2019-11-08
Trusty ESM index should only be installed on i386/amd64


powersj:
Upstream bug: https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1851858

panlinux:
We can fix part of this in postinst, but we also need a fix for `ua attach`, as it tries to enable `esm-infra` again:
```
ubuntu@trusty-ppc64el:~$ sudo ua attach lolsecret
Enabling default service esm-infra
Updating package lists
Updating package lists
Failed to enable default services, check: sudo ua status
```
ESM remains disabled and with no advertisement, which is good since `apt-get update` keeps working.
The postinst fix I'm testing is:
```diff
--- a/debian/postinst
+++ b/debian/postinst
@@ -17,6 +17,8 @@ ESM_INFRA_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-infra-trusty.list"
 ESM_APT_PREF_FILE_TRUSTY="/etc/apt/preferences.d/ubuntu-esm-trusty"
 ESM_INFRA_APT_PREF_FILE_TRUSTY="/etc/apt/preferences.d/ubuntu-esm-infra-trusty"
 
+MYARCH="$(dpkg --print-architecture)"
+ESM_SUPPORTED_ARCHS="i386 amd64"
 
 unconfigure_esm() {
     rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
@@ -78,8 +80,15 @@ case "$1" in
       rm -rf /var/cache/ubuntu-advantage-tools
 
       if [ "14.04" = "$VERSION_ID" ]; then
-        configure_esm
+        if echo "$ESM_SUPPORTED_ARCHS" | grep -qw "$MYARCH"; then
+          # 14.04 and supported arch
+          configure_esm
+        else
+          # 14.04 and unsupported arch
+          unconfigure_esm
+        fi
       else
+        # not 14.04, regardless of arch
         unconfigure_esm
       fi
       if [ ! -f /var/log/ubuntu-advantage.log ]; then

panlinux:
Doesn't look like anything else needs to be done with the client. I asked the server devs to change the config on staging to allow only for i386/amd64 for esm-infra, and a quick test on ppc64el showed this working already, as the client didn't try to enable esm-infra on attach, nor was it advertised as being available on unattached status.

[...]

PR #915, opened 2019-11-12
Ua nonx86 fix

panlinux:
Fixes issue #911 

I'm using `dpkg --print-architecture` and not `uname -a`, but it's fine since it's contained in the postinst script. This does introduce another point where the list of supported architectures for esm-infra is stored, however. If this list ever changes, an SRU will be needed.
@blackboxsw wondered if we couldn't do something dynamic in postinst, but that would involve a network call, like `ua status` already does. It could be an option I guess, and fallback to a hardcoded list if the network isn't available, but it might be frowned upon, and will not work in our builders and DEP8 runners (egress filtering would block it).

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+ubuntu-advantage-tools (19.6~ubuntu14.04.4) trusty; urgency=medium
+
+  * d/postinst: only configure ESM on supported architectures (LP: #1851858)
+
+ -- Andreas Hasenack <andreas@canonical.com>  Mon, 11 Nov 2019 09:56:46 -0300
+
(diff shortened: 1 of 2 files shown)


panlinux:
I don't understand what travis is trying to do here. My branch (`ua-nonx85-fix`) is definitely in a remote repo (`panlinux`), as is common practice:
```
$ git clone --depth=50 --branch=ua-nonx86-fix https://github.com/CanonicalLtd/ubuntu-advantage-client.git CanonicalLtd/ubuntu-advantage-client
```
Why would it think it's in the CanonicalLtd ua repo?

I checked other PRs where tests passed, and all were about branches from other remotes, and the checkout command doesn't have a `--branch` parameter, so I'm even wondering if it really tested the proposed branch, and not just master over and over again.

PR #918, opened 2019-11-13
Demo/aws premium

blackboxsw:
- Update upstart "start on" condition to only exec once cloud-init emits /var/lib/cloud/data/result.json
- Add 'ec2' alias for cloud platform name on Trusty

### DIFF
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -30,7 +30,10 @@ def get_cloud_type() -> str:
 def cloud_instance_factory() -> clouds.UAPremiumCloudInstance:
     from uaclient.clouds import aws
 
-    cloud_instance_map = {"aws": aws.UAPremiumAWSInstance}
+    cloud_instance_map = {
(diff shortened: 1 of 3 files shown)


uaclient/clouds/identity.py
```diff
@@ -8,7 +8,7 @@
 CLOUDINIT_RESULT_FILE = "/var/lib/cloud/data/result.json"
 
 
-@util.retry(FileNotFoundError, [1, 2])
+@util.retry(FileNotFoundError, [1, 2, 3])
```
powersj: Is this better handled by the service file? We could block on the cloud-id file?

blackboxsw: Good point, added and tested the following upstart change:

# Only start once cloud-init finishes detecting the DataSource on trusty
start on file FILE=/var/lib/cloud/data/result.json

Able to see that no retries were needed then, so I dropped the added retry.

PR #919, opened 2019-11-14
WIP: High-level review of AWS premium image attach

OddBloke:
This branch is going to be split up into more reviewable chunks, but first let's iron out any high-level changes so that we aren't spending time refactoring something that won't pass muster.


uaclient/clouds/identity.py
```diff
@@ -0,0 +1,50 @@
...
+
+from uaclient import exceptions
+from uaclient import clouds
+from uaclient import status
+from uaclient import util
+
+CLOUDINIT_RESULT_FILE = "/var/lib/cloud/data/result.json"
+
+
+@util.retry(FileNotFoundError, [1, 2])
```
OddBloke: As discussed on IRC, this retry is no longer necessary (because the upstart job will ensure we never hit `FileNotFoundError` here, unless something very bad has happened and then waiting for a few seconds isn't going to help).

It also follows that we can drop the decorator defintion.

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,50 @@
...
+    except FileNotFoundError:
+        pass
+    return ""
+
+
+def cloud_instance_factory() -> clouds.UAPremiumCloudInstance:
+    from uaclient.clouds import aws
+
+    cloud_instance_map = {
+        "aws": aws.UAPremiumAWSInstance,
```
OddBloke: I think we should normalise these in `get_cloud_type`, that's already having to handle the differences between trusty and later releases.

uaclient/cli.py
```diff
@@ -297,6 +315,48 @@ def action_detach(args, cfg):
...
+        print(
+            "Premium image support is not supported on '%s'.\n"
+            "For more information see: https://ubuntu.com/advantage"
+            % cloud_type
+        )
+        return 0
+    instance = identity.cloud_instance_factory()
+    contract_client = contract.UAContractClient(cfg)
+    pkcs7 = instance.identity_doc
+    contractTokenResponse = contract_client.request_premium_aws_attach(pkcs7)
```
OddBloke: It seems strange to have some of the cloud-specific logic captured in the `UAPremiumCloudInstance` sub-classes but still have AWS hard-coded here. We don't necessarily need to address this before we land it (because we're only going to support AWS to start with), but I would like us to be sure we know how we're going to encapsulate this better in future, to be sure we aren't heading into a dead end.

uaclient/clouds/aws.py
```diff
@@ -0,0 +1,33 @@
...
+DMI_PRODUCT_SERIAL = "/sys/class/dmi/id/product_serial"
+DMI_PRODUCT_UUID = "/sys/class/dmi/id/product_uuid"
+
+
+class UAPremiumAWSInstance(UAPremiumCloudInstance):
+
+    # mypy does not handle @property around inner decorators
+    # https://github.com/python/mypy/issues/1362
+    @property  # type: ignore
+    @util.retry(HTTPError, retry_sleeps=[1, 2, 5])
```
OddBloke: Oh, I missed this usage of the retry decorator when I made my previous comment. My bad!

OddBloke:
OK, I've made some inline high-level comments. I think there are three obviously separable chunks here:

1) the basic implementation of the actual premium attach functionality
2) implementing retry
3) upstart/systemd integration

(1) should clearly come first, and then (2) and (3) can probably follow in either order. (Depending on how big (1) is, it might be best to split it up further, but there isn't an _obvious_ way to do that. Thoughts?)

What do you think @blackboxsw ?

uaclient/clouds/identity.py
blackboxsw: +1, Let's drop that from this changeset before landing. I guess we should demo/aws-premium branch proper.

uaclient/clouds/identity.py
blackboxsw: good point. +1

uaclient/clouds/identity.py
blackboxsw: This retry decorartor usage will disappear. But, we'll keep the definition because of the 2nd callsite in identity.

OddBloke:
High-level review complete, we can close this out.

PR #921, opened 2019-11-18
premium: add aws support for premium images

blackboxsw:
Provide a new attach-premium subcommand that will crawl AWS premium
metadata and provide that to new aws ua-contacts backend route to
allow for easy attach of premium images without having to obtain a
token from ubuntu.com/advantage

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -101,6 +101,24 @@ def require_valid_entitlement_name(operation: str):
     return wrapper
 
 
+def attach_premium_parser(parser):
+    """Build or extend an arg parser for init-attach subcommand."""
+    usage = USAGE_TMPL.format(name=NAME, command="attach-premium")
+    if not parser:
+        parser = argparse.ArgumentParser(
+            prog="attach-premium",
+            description=(
+                "Automatically enable Ubuntu Advantage on a premium Ubuntu"
+                " image"
+            ),
+            usage=usage,
+        )
+    else:
+        parser.usage = usage
+        parser.prog = "premium-attach"
+    return parser
+
+
 def attach_parser(parser):
     """Build or extend an arg parser for attach subcommand."""
     usage = USAGE_TMPL.format(name=NAME, command="attach <token>")
@@ -297,6 +315,48 @@ def action_detach(args, cfg):
     return 0
 
 
+def action_attach_premium(args, cfg):
+    from uaclient.clouds import identity
+
+    if cfg.is_attached:
+        print(
+            "This machine is already attached to '{}'.".format(
+                cfg.accounts[0]["name"]
+            )
+        )
+        return 0
+    if os.getuid() != 0:
+        raise exceptions.NonRootUserError()
+    cloud_type = identity.get_cloud_type()
+    if cloud_type not in ("aws",):  # TODO(avoid hard-coding supported types)
+        print(
+            "Premium image support is not supported on '%s'.\n"
+            "For more information see: https://ubuntu.com/advantage"
+            % cloud_type
+        )
+        return 0
+    instance = identity.cloud_instance_factory()
+    contract_client = contract.UAContractClient(cfg)
+    pkcs7 = instance.identity_doc
+    contractTokenResponse = contract_client.request_premium_aws_attach(pkcs7)
+    contract_token = contractTokenResponse["contractToken"]
+    if not contract.request_updated_contract(
+        cfg, contract_token, allow_enable=True
+    ):
+        print(
+            ua_status.MESSAGE_ATTACH_FAILURE_TMPL.format(url=cfg.contract_url)
+        )
+    contract_name = cfg.machine_token["machineTokenInfo"]["contractInfo"][
+        "name"
+    ]
+    action_status(args=None, cfg=cfg)
+    print(
+        ua_status.MESSAGE_ATTACH_SUCCESS_TMPL.format(
+            contract_name=contract_name
+        )
+    )
+
+
 def action_attach(args, cfg):
     if cfg.is_attached:
         print(
@@ -386,6 +446,12 @@ def get_parser():
     )
     attach_parser(parser_attach)
     parser_attach.set_defaults(action=action_attach)
+    parser_attach_premium = subparsers.add_parser(
+        "attach-premium",
+        help="automatically enable Ubuntu Advantage on a premium Ubuntu image",
+    )
+    attach_premium_parser(parser_attach_premium)
+    parser_attach_premium.set_defaults(action=action_attach_premium)
     parser_detach = subparsers.add_parser(
         "detach",
         help="remove this machine from an Ubuntu Advantage subscription",
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -78,6 +78,18 @@ def assert_attached(unattached_msg_tmpl=None):
     return wrapper
 
 
+def assert_not_attached(f):
+    """Decorator asserting unattached config."""
+
+    @wraps(f)
+    def new_f(args, cfg):
+        if cfg.is_attached:
+            raise exceptions.AlreadyAttachedError(cfg)
+        return f(args, cfg)
+
+    return new_f
+
+
 def require_valid_entitlement_name(operation: str):
     """Decorator ensuring that args.name is a valid service.
 
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -114,7 +114,7 @@ def require_valid_entitlement_name(operation: str):
 
 
 def attach_premium_parser(parser):
-    """Build or extend an arg parser for init-attach subcommand."""
+    """Build or extend an arg parser for attach-premium subcommand."""
     parser.prog = "attach-premium"
     parser.usage = USAGE_TMPL.format(name=NAME, command=parser.prog)
     parser._optionals.title = "Flags"
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -316,6 +316,35 @@ def action_detach(args, cfg):
     return 0
 
 
+def _attach_with_token(
+    cfg: config.UAConfig, token: str, allow_enable: bool
+) -> int:
+    """Common functionality to take a token and attach via contract backend"""
+    try:
+        contract.request_updated_contract(
+            cfg, token, allow_enable=allow_enable
+        )
+    except util.UrlError as exc:
+        with util.disable_log_to_console():
+            logging.exception(exc)
+        print(ua_status.MESSAGE_ATTACH_FAILURE)
+        return 1
+    except exceptions.UserFacingError as exc:
+        logging.warning(exc.msg)
+        return 1
+    contract_name = cfg.machine_token["machineTokenInfo"]["contractInfo"][
+        "name"
+    ]
+    print(
+        ua_status.MESSAGE_ATTACH_SUCCESS_TMPL.format(
+            contract_name=contract_name
+        )
+    )
+
+    action_status(args=None, cfg=cfg)
+    return 0
+
+
 @assert_not_attached
 @assert_root
 def action_attach_premium(args, cfg):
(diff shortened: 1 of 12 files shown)


uaclient/contract.py
```diff
@@ -105,6 +106,22 @@ def request_resources(self) -> "Dict[str, Any]":
         )
         return resource_response
 
+    def request_premium_aws_attach(self, pkcs7: str):
```
OddBloke: This isn't exactly requesting attach, right? I _think_ this is performing the fetch of the token (which in the normal attach path the user would provide), am I following correctly?

uaclient/contract.py
```diff
@@ -105,6 +106,22 @@ def request_resources(self) -> "Dict[str, Any]":
...
+
+        @param pkcs7: string obtained from AWS metadata service from
+            http://169.254.169.254/latest/dynamic/instance-identity/pkcs7
+            from this instance.
+
+        @return: Dict of the JSON response containing the contract-token.
+        """
+        data = {"pkcs7": pkcs7}
+        response, _headers = self.request_url(
+            API_V1_PREMIUM_AWS_TOKEN, data=data
```
OddBloke: We aren't setting `content-type` here, in what form does this end up getting sent over the wire?

uaclient/status.py
```diff
@@ -145,6 +145,9 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_UNENTITLED_TMPL = """\
 This subscription is not entitled to {title}.
 For more information see: https://ubuntu.com/advantage"""
+MESSAGE_UNSUPPORTED_PREMIUM = """\
```
OddBloke: This appears to be unused?

uaclient/cli.py
```diff
@@ -101,6 +101,24 @@ def new_f(args, cfg):
     return wrapper
 
 
+def attach_premium_parser(parser):
+    """Build or extend an arg parser for init-attach subcommand."""
+    usage = USAGE_TMPL.format(name=NAME, command="attach-premium")
+    if not parser:
```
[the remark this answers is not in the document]
blackboxsw: +1, found that while starting to write unit tests. thanks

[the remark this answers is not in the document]
OddBloke: I have addressed this locally.

uaclient/cli.py
```diff
@@ -297,6 +315,48 @@ def action_detach(args, cfg):
     return 0
 
 
+def action_attach_premium(args, cfg):
+    from uaclient.clouds import identity
```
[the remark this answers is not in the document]
OddBloke: OK, I see that we can't because the module doesn't yet exist. Let's make sure we move it up when it does exist.

uaclient/cli.py
```diff
@@ -297,6 +315,48 @@ def action_detach(args, cfg):
...
+
+    if cfg.is_attached:
```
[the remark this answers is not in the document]
OddBloke: I'm looking at this now.

uaclient/cli.py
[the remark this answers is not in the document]
OddBloke:
```py
def action_attach_premium(args, cfg, identity=None):
    if identity is None:
        from uaclient.clouds import identity
```

uaclient/cli.py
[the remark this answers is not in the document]
blackboxsw: rebased to include your changes

uaclient/cli.py
```diff
@@ -297,6 +315,48 @@ def action_detach(args, cfg):
...
+            "For more information see: https://ubuntu.com/advantage"
+            % cloud_type
+        )
+        return 0
+    instance = identity.cloud_instance_factory()
+    contract_client = contract.UAContractClient(cfg)
+    pkcs7 = instance.identity_doc
+    contractTokenResponse = contract_client.request_premium_aws_attach(pkcs7)
+    contract_token = contractTokenResponse["contractToken"]
+    if not contract.request_updated_contract(
```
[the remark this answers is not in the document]
blackboxsw: We do need that up here, fixed. We probably should refactor to avoid this duplication (and potential divergence) of behavior.

uaclient/contract.py
blackboxsw: well, the same utf-encoded byte-string. if no headers are provided, self.headers() is called which the base serviceclient defines as
    def headers(self):
        return {
            "user-agent": "UA-Client/{}".format(version.get_version()),
            "accept": "application/json",
            "content-type": "application/json",
        }

As a result, the headers settings will say we are encoding json content and so we request_url encodes a json string. We probably need to clean up the unnecessary headers being set.

uaclient/status.py
blackboxsw: Only due to an oversight for not pulling in the identity module. I have since pulled the identity module into this branch and resolved to source a common MESSAGE_UNSUPPORTED_PREMIUM_CLOUD_TYPE message which cli also uses.

uaclient/contract.py
blackboxsw: Renamed to request_premium_aws_contract_token

uaclient/cli.py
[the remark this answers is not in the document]

OddBloke:
OK, I'm going to do a review pass over this, and then start addressing my own comments until Chad is online.

uaclient/cli.py
```diff
@@ -101,10 +113,17 @@ def new_f(args, cfg):
     return wrapper
 
 
+def attach_premium_parser(parser):
+    """Build or extend an arg parser for init-attach subcommand."""
```
OddBloke: Nit: "init-attach"

uaclient/cli.py
```diff
@@ -297,23 +316,13 @@ def action_detach(args, cfg):
...
-            )
-        )
-        return 0
-    if os.getuid() != 0:
-        raise exceptions.NonRootUserError()
-    if not args.token:
-        raise exceptions.UserFacingError(
-            ua_status.MESSAGE_ATTACH_REQUIRES_TOKEN
-        )
+def _attach_with_token(
```
OddBloke: Good stuff, thank you!

uaclient/cli.py
```diff
@@ -336,6 +345,45 @@ def action_attach(args, cfg):
     return 0
 
 
+@assert_not_attached
+@assert_root
+def action_attach_premium(args, cfg):
+    # TODO(import identity from the top of the module)
```
OddBloke: Now we have that module, we should be able to do this, I think?

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,55 @@
...
+from uaclient import util
+
+CLOUDINIT_RESULT_FILE = "/var/lib/cloud/data/result.json"
+
+
+# Mapping of datasource names to cloud-id responses. Trusty compat with Xenial+
+DATASOURCE_TO_CLOUD_ID = {"ec2": "aws"}
+
+
+def get_cloud_type_from_result_file(result_file=CLOUDINIT_RESULT_FILE) -> str:
```
OddBloke: Nit: parameter type annotation

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,55 @@
...
+    result = json.loads((util.load_file(result_file)))
```
OddBloke: Nit: additional parentheses

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,55 @@
...
+    dsname = result["v1"]["datasource"].split()[0].lower()
```
OddBloke: How will this behave if cloud-init fails for some reason? I _assume_ that these keys would only be missing if something really bad happened (which would presumably mean that a premium attach either won't work or will be pointless because the machine is going to have to be decommissioned)?

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,55 @@
...
+def get_cloud_type() -> str:
+    if util.which("cloud-id"):
+        # Present in cloud-init on >= Xenial
+        out, _err = util.subp(["cloud-id"])
+        return out.strip()
+    try:
+        return get_cloud_type_from_result_file()
+    except FileNotFoundError:
+        pass
+    return ""
```
OddBloke: Does None make more sense for this case?

uaclient/clouds/identity.py
```diff
@@ -0,0 +1,55 @@
...
+
+def cloud_instance_factory() -> clouds.UAPremiumCloudInstance:
+    from uaclient.clouds import aws
+
+    cloud_instance_map = {"aws": aws.UAPremiumAWSInstance}
+
+    cloud_type = get_cloud_type()
+    if not cloud_type:
+        raise exceptions.UserFacingError(
+            "Unable to determine premium image platform support\n"
```
OddBloke: Should this be a constant in status.py?

uaclient/clouds/tests/test_aws.py
```diff
@@ -0,0 +1,56 @@
...
+            ("notec2", "ec2UUID", "ec2Serial", True),
+            ("notec2", "EC2UUID", "Ec2Serial", True),
+            ("notec2", "!EC2UUID", "Ec2Serial", False),
+            ("notec2", "EC2UUID", "1Ec2Serial", False),
+            ("notec2", "ec2UUID", "ec3Serial", False),
+            ("notec2", "ec3UUID", "ec2Serial", False),
+        ),
+    )
+    @mock.patch(M_PATH + "util.load_file")
+    def test_is_viable_based_on_sys_product_serial_and_uuid(
```
OddBloke: I really like this test, it's nicely parameterized and the body is very clear. Thanks!

uaclient/clouds/tests/test_identity.py
```diff
@@ -0,0 +1,102 @@
...
+            ("DataSourceSomeTHING", "something"),
+            ("DaTaSoUrCeMiNe", "mine"),
+            ("DataSourceEc2", "aws"),
+            ("DataSourceEc2Lookalike", "ec2lookalike"),
+        ),
+    )
+    def test_get_cloud_type_from_lowercase_v1_datasource_key(
+        self, ds_name, expected, tmpdir
+    ):
+        """The value of cloud_type is extacted from results datasource key.
```
OddBloke: Nit: "extacted"

uaclient/tests/test_cli_attach_premium.py
```diff
@@ -0,0 +1,111 @@
...
+from uaclient.cli import (
+    action_attach_premium,
+    attach_premium_parser,
+    get_parser,
+)
+from uaclient.exceptions import AlreadyAttachedError, NonRootUserError
+
+M_PATH = "uaclient.cli."
+
+BASIC_MACHINE_TOKEN = {
```
OddBloke: This looks familiar, is there a way we can reuse an existing definition of a basic machine token?

uaclient/cli.py
OddBloke: Looking at this.

OddBloke: Fix pushed.

uaclient/clouds/identity.py
OddBloke: Looking at this.

OddBloke: Fix pushed.

uaclient/clouds/identity.py
OddBloke: Fix pushed.

uaclient/clouds/tests/test_identity.py
OddBloke: Looking at this.

OddBloke: Fix pushed.

uaclient/cli.py
OddBloke: Looking at this.

OddBloke: Fix pushed.

uaclient/clouds/identity.py
OddBloke: Looking at this now.

OddBloke: Fix pushed.

uaclient/clouds/identity.py
OddBloke: Looking at this now.

OddBloke: Fix pushed.

uaclient/tests/test_cli_attach_premium.py
OddBloke: Looking at this now.

OddBloke: Fix pushed.

uaclient/cli.py
blackboxsw: thanks!

uaclient/clouds/identity.py
blackboxsw: thanks again.

uaclient/clouds/identity.py
blackboxsw: There would be a couple reasons for result.json not being the format we expect or not being present:
 - Cloud-init version is much older than what is shipped in trusty-updates
 - cloud-init failed  on a system due to missing dependencies
 - cloud-init failed on an unsupported platform or distro type

Any of these would fall into the unsupported images camp, so I don't think we need to address that.

uaclient/clouds/tests/test_aws.py
blackboxsw: thanks for working out the pytest.parameterize examples. made it easy to extend.

PR #923, opened 2019-11-19
entitlements: rely solely on contract server for repo_url

OddBloke:
Before this commit, we had a hard-coded default that we would use if the
contract server failed to provide us with an aptURL directive.  The
logic for that default becomes more complex with xenial/bionic, because
the URLs have changed.

Instead of introducing more complexity into the client, we instead drop
that defaulting, with the assumption that the contract server will
always provide an aptURL directive.

### DIFF
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -45,11 +45,6 @@ class RepoEntitlement(base.UAEntitlement):
 
     @property
     @abc.abstractmethod
-    def repo_url(self) -> str:
-        pass
-
-    @property
-    @abc.abstractmethod
     def repo_key_file(self) -> str:
         pass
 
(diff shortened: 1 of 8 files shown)


uaclient/exceptions.py
```diff
@@ -29,6 +29,17 @@ def __init__(self, cfg):
         )
 
 
+class MissingAptURLDirective(UserFacingError):
```
blackboxsw: Excellent, thanks for the clear breadcrumb for if this breaks on ua-contract backend.

uaclient/entitlements/repo.py
```diff
@@ -106,7 +101,7 @@ def application_status(self) -> "Tuple[ApplicationStatus, str]":
         )
         repo_url = directives.get("aptURL")
         if not repo_url:
-            repo_url = self.repo_url
+            raise exceptions.MissingAptURLDirective(self.name)
```
blackboxsw: At some point, not today, it might be nice for uaclient to define a set of mandatory keys required of a contract server response. If any are missing, we balk early and in one place instead of handling this in multiple places in the codebase. Now is probably not the time for us to address that. What do you think about a queuing a techdebt issue to tighten up our response validation. Is it worth it?

OddBloke: Yeah, I had the same thought. I think a class which represents the response which handles validation in its `__init__` could be a good way to go.

I'm +1 on a techdebt issue.

PR #924, opened 2019-11-19
ua premium: add upstart  and systemd boot scripts to automatically attach premium images

blackboxsw:
Add upstart script and systemd unit which will wait for cloud-init to complete before attempting to run ua attach-premium on a machine.

On non-premium images, `ua attach-premium` exits 0 and logs that attach a platform is not detected as a premium image.

On attached machines after reboot, `ua attach-premium` logs a message that the machine is already attached and exits 0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@
 import glob
 import setuptools
 
-from uaclient import defaults, version
+from uaclient import defaults, util, version
 
 NAME = "ubuntu-advantage-tools"
 
@@ -20,6 +20,21 @@ def _get_version():
     return major_minor
 
 
+def _get_data_files():
+    data_files = [
+        ("/etc/apt/apt.conf.d", ["apt.conf.d/51ubuntu-advantage-esm"]),
+        ("/etc/ubuntu-advantage", ["uaclient.conf"]),
+        ("/usr/share/keyrings", glob.glob("keyrings/*")),
+        (defaults.CONFIG_DEFAULTS["data_dir"], []),
+    ]
+    rel_major, _rel_minor = util.get_platform_info()["release"].split(".", 1)
+    if rel_major == "14":
+        data_files.append(("/etc/init", glob.glob("upstart/*")))
+    else:
+        data_files.append(("/lib/systemd/system", glob.glob("systemd/*")))
+    return data_files
+
+
 setuptools.setup(
     name=NAME,
     version=_get_version(),
@@ -33,12 +48,7 @@ setuptools.setup(
             "features.*",
         ]
     ),
-    data_files=[
-        ("/etc/apt/apt.conf.d", ["apt.conf.d/51ubuntu-advantage-esm"]),
-        ("/etc/ubuntu-advantage", ["uaclient.conf"]),
-        ("/usr/share/keyrings", glob.glob("keyrings/*")),
-        (defaults.CONFIG_DEFAULTS["data_dir"], []),
-    ],
+    data_files=_get_data_files(),
     install_requires=INSTALL_REQUIRES,
     extras_require=dict(test=TEST_REQUIRES),
     author="Ubuntu Server Team",
--- a/systemd/ua-premium.service
+++ b/systemd/ua-premium.service
@@ -1,0 +1,8 @@
+[Unit]
+Description=Ubuntu Advantage premium attach
+After=cloud-final.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ua attach-premium
+TimeoutSec=0
--- a/upstart/ua-premium.conf
+++ b/upstart/ua-premium.conf
@@ -1,0 +1,14 @@
+# ua-premium - Ubuntu Advantage Premium attach
+
+# This task is run on boot to automatically attach vms to Ubuntu Advantage
+# for select premium images. The machine does not get attached if the
+# image is not considered a premium image type.
+
+description    "Ubuntu Advantage premium attach"
+
+# Only start once cloud-init finishes detecting the DataSource on trusty
+start on file FILE=/var/lib/cloud/data/result.json
+
+task
+# Attempt to attach-premium action. Does nothing if machine is already attached
+exec /usr/bin/ua attach-premium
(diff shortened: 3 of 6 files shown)


bryceharrington:
Looks good, I made a couple suggestions on irc and they're integrated now.  +1 on the systemd approach used.

basak:
I'm not sure `/usr/bin/cloud-init status --wait` is best. cloud-init integrates with systemd itself, so maybe it should be an `After=cloud-init-thing`? That would make the systemd unit status more accurate, which will avoid some edge case issues that might rely on the correct status. Like a shu down initiated during boot avoiding a hang on a timeout and that kind of thing. Ideally cloud-init would declare some systemd targets and make that a stable API for other systemd units, but I believe there's enough there already to provide something that works right now.

systemd/ua-premium.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Ubuntu Advantage premium attach
+
+[Service]
+Type=oneshot
+ExecStartPre=/usr/bin/cloud-init status --wait
```
OddBloke: I agree with @basak, we should be using native systemd dependencies to get this to start at the appropriate time. This approach will add another invocation of cloud-init to each boot, which is a performance hit we don't need to take.

upstart/ua-premium.conf
```diff
@@ -0,0 +1,15 @@
+# ua-premium - Ubuntu Advantage Premium attach
+
+# This task is run on reboot to adapt to potential change in
```
OddBloke: This comment looks like it needs updating.

upstart/ua-premium.conf
```diff
@@ -0,0 +1,15 @@
...
+
+description    "Ubuntu Advantage premium attach"
+
+# Only start once cloud-init finishes detecting the DataSource on trusty
+start on file FILE=/var/lib/cloud/data/result.json
+
+task
+# Attempt to attach-premium action if machine is not already attached
+exec /usr/bin/ua attach-premium
+
```
OddBloke: Do we need this newline?

upstart/ua-premium.conf
```diff
@@ -0,0 +1,15 @@
...
+# kernels/environment across boot and update Ubuntu Advantage status
+# information.
+
+description    "Ubuntu Advantage premium attach"
+
+# Only start once cloud-init finishes detecting the DataSource on trusty
+start on file FILE=/var/lib/cloud/data/result.json
+
+task
+# Attempt to attach-premium action if machine is not already attached
```
OddBloke: This doesn't look accurate; we run `attach-premium` whether or not the machine is already attached.

setup.py
```diff
@@ -20,6 +20,21 @@ def _get_version():
...
 
+def _get_data_files():
+    data_files = [
+        ("/etc/apt/apt.conf.d", ["apt.conf.d/51ubuntu-advantage-esm"]),
+        ("/etc/ubuntu-advantage", ["uaclient.conf"]),
+        ("/usr/share/keyrings", glob.glob("keyrings/*")),
+        (defaults.CONFIG_DEFAULTS["data_dir"], []),
+    ]
+    rel_major, _rel_minor = util.get_platform_info()["release"].split(".", 1)
+    if rel_major == "14":
```
OddBloke: Does this work as intended if I'm building a trusty source package on a non-trusty system? Or will this select the build system based on the build host?

(This may not be a problem, I just want to understand!)

basak:
> cloud-init integrates with systemd itself, so maybe it should be an `After=cloud-init-thing`?

Now that I think about it, I believe upstart has an equivalent mechanism that should work too, and since it's legacy it shouldn't change behaviour on you either.

I have no particular opinion on what would be most appropriate for the upstart side. I didn't think of Dan's performance point. Do you care about performance on Trusty?

setup.py
blackboxsw: Correct. This will select to package files based on the underlying host platform. Since we removed the ability to cross-platform build in earlier iterations, our commitment to build on the target platform was something  that I relied upon here.

systemd/ua-premium.service
blackboxsw: +1 shifting back to the earlier implementation which goes as soon as cloud-init has detected the datasource and brought up 'final' networking.

[Install]
WantedBy=multi-user.target
# cloud-init has completed datasource detection by network-online target
After=network-online.target
Want=network-online.target

upstart/ua-premium.conf
blackboxsw: dropped.

upstart/ua-premium.conf
blackboxsw: updated to
# Attempt to attach-premium action. Does nothing if machine is already attached

setup.py
OddBloke: We haven't been able to build the _binary_ packages on different hosts for a while, but I believe we can still build the source packages on different hosts. This is a pretty important difference, both for a local sbuild workflow but also for the package builds in Travis.

I'll double-check a source package build with this branch in a couple of minutes.

OddBloke: I've just tested this locally and a source build on focal correctly installed the upstart file for trusty and the systemd file for bionic. Marking this resolved.

basak:
On Tue, Nov 19, 2019 at 12:47:50PM -0800, Chad Smith wrote:
> +1 shifting back to the earlier implementation which goes as soon as cloud-init has detected the datasource and brought up 'final' networking.
> 
> [Install]
> WantedBy=multi-user.target
> # cloud-init has completed datasource detection by network-online target
> After=network-online.target
> Want=network-online.target

Are you sure this doesn't need to refer to cloud-final.service? That
service looks like it runs after network-online.target.

systemd/ua-premium.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Ubuntu Advantage premium attach
+After=cloud-final.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ua attach-premium
+TimeoutSec=0
```
OddBloke: What does this configuration do?

raharper: Do you want to have this command retry for any reason? or are you fine with a single attempt?
For non-daemon services, the TimeoutSec controls how long systemd waits before marking success/failure for a service; in this case it's appropriate since once ua exits, we either are attached (success! return code of 0) or not attached (and return code non-zero).

This looks fine to me.  Let me know if there are other design goals here.

PR #925, opened 2019-11-19
premium: add retry decorator around aws.identity_doc to handle failure

blackboxsw:
Since metadata services could go down intermittently for maintenance,
add a retry decorator that acceepts a list of sleep times to use when
catching an exception.

Add this retry decorator and a simple backoff of three retries at
1s 2s and 5s to attempt to allow for short downtimes of
the indetity_doc route for AWS.

### DIFF
--- a/uaclient/clouds/aws.py
+++ b/uaclient/clouds/aws.py
@@ -1,5 +1,8 @@
+from urllib.error import HTTPError
+
 from uaclient.clouds import UAPremiumCloudInstance
 from uaclient import util
+
 
 IMDS_URL = "http://169.254.169.254/latest/dynamic/instance-identity/pkcs7"
 SYS_HYPERVISOR_PRODUCT_UUID = "/sys/hypervisor/uuid"
@@ -8,7 +11,11 @@ DMI_PRODUCT_UUID = "/sys/class/dmi/id/product_uuid"
 
 
 class UAPremiumAWSInstance(UAPremiumCloudInstance):
-    @property
+
+    # mypy does not handle @property around inner decorators
+    # https://github.com/python/mypy/issues/1362
+    @property  # type: ignore
+    @util.retry(HTTPError, retry_sleeps=[1, 2, 5])
     def identity_doc(self):
         response, _headers = util.readurl(IMDS_URL)
         return response
--- a/uaclient/clouds/tests/test_aws.py
+++ b/uaclient/clouds/tests/test_aws.py
@@ -1,4 +1,7 @@
+import logging
 import mock
+from io import BytesIO
+from urllib.error import HTTPError
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


uaclient/clouds/aws.py
```diff
@@ -1,14 +1,20 @@
 from uaclient.clouds import UAPremiumCloudInstance
 from uaclient import util
 
+from urllib.error import HTTPError
```
OddBloke: stdlib imports should go above our internal imports.

uaclient/clouds/tests/test_aws.py
```diff
@@ -17,6 +20,44 @@ def test_identity_doc_from_aws_url_pkcs7(self, readurl):
...
+    @mock.patch(M_PATH + "util.readurl")
+    def test_retry_backoff_on_failed_identity_doc(
+        self, readurl, sleep, fail_count, exception, caplog_text
+    ):
+        """Retry backoff is attempted before failing to get AWS.identity_doc"""
+
+        def fake_someurlerrors(url):
+            if readurl.call_count <= fail_count:
+                raise HTTPError(
+                    None, 400 + readurl.call_count, None, None, BytesIO()
```
OddBloke: A nit, but I would suggest using something other than 400 as the starting point here. When I see "404" and "HTTP Error 401" those have semantic meaning that isn't relevant here. Whereas, say, 704 and "HTTP Error 701" are more obviously just test data.

uaclient/clouds/tests/test_aws.py
```diff
@@ -17,6 +20,44 @@ def test_identity_doc_from_aws_url_pkcs7(self, readurl):
...
+            with pytest.raises(HTTPError) as excinfo:
+                instance.identity_doc
+            assert 404 == excinfo.value.code
+        else:
+            assert "pkcs7WOOT!==" == instance.identity_doc
+
+        expected_sleep_calls = [mock.call(1), mock.call(2), mock.call(5)]
+        assert expected_sleep_calls == sleep.call_args_list
+        expected_logs = [
+            "HTTP Error 401: None Retrying 3 more times.",
```
OddBloke: We pass `None` into `HTTPError`s constructor three times; should we pass an actual value for whichever of those this "None" is, to be sure that we're including the correct thing in these log messages?

uaclient/clouds/tests/test_aws.py
```diff
@@ -17,6 +20,44 @@ def test_identity_doc_from_aws_url_pkcs7(self, readurl):
...
+            if readurl.call_count <= fail_count:
+                raise HTTPError(
+                    None, 400 + readurl.call_count, None, None, BytesIO()
+                )
+            return "pkcs7WOOT!==", {"header": "stuff"}
+
+        readurl.side_effect = fake_someurlerrors
+        instance = UAPremiumAWSInstance()
+        if exception:
+            with pytest.raises(HTTPError) as excinfo:
```
OddBloke: A nit, but if we're going to hard-code `HTTPError` here (which is fine!), we can make `exception` just a `bool`. (Alternatively, we could do `pytest.raises(exception)` here, but I think that's less obvious for no real gain.)

uaclient/clouds/tests/test_aws.py
```diff
@@ -17,6 +20,44 @@ def test_identity_doc_from_aws_url_pkcs7(self, readurl):
...
+        """Retry backoff is attempted before failing to get AWS.identity_doc"""
+
+        def fake_someurlerrors(url):
+            if readurl.call_count <= fail_count:
+                raise HTTPError(
+                    None, 400 + readurl.call_count, None, None, BytesIO()
+                )
+            return "pkcs7WOOT!==", {"header": "stuff"}
+
+        readurl.side_effect = fake_someurlerrors
```
OddBloke: For reference, you could achieve roughly the same effect with:

```py
readurl.side_effect = [
    HTTPError(None, 400 + readurl.call_count, None, None, BytesIO())
] * fail_count + [("pkcs7WOOT!==", {"header": "stuff"})]
```

because `side_effect` will iterate over a given list, raising any exception subclasses and returning the value otherwise. In this case, I think what you went with is preferable, but figured I'd share an interesting tidbit. :+1:

uaclient/clouds/tests/test_aws.py
blackboxsw: very good point. fixed.

uaclient/clouds/tests/test_aws.py
blackboxsw: Changed to add the error message instead of None.

uaclient/clouds/tests/test_aws.py
blackboxsw: Smart idea, the problem is that list arithmetic is setting the error code to 400 for every HTTPError created and I wanted something to reprensent increased different error codes per error. We could do this with a bit more verbose list creation, but I figured that's gets almost as verbose as the assigned local function.

uaclient/util.py
```diff
@@ -218,6 +219,37 @@ def disable_log_to_console():
...
 
+def retry(exception, retry_sleeps):
+    """Decorator to retry on exception for retry_sleeps.
+
+     @param retry_sleeps: List of sleep lengths to apply between
+        retries. Specifying a list of [0.5, 1] tells subp to retry twice
+        on failure; sleeping half a second before the first retry and 1 second
+        before the second retry.
+     @param exception: The exception class to catch and retry for the provided
+        retry_sleeps. Any other exception types will immediately be raised.
```
OddBloke: A nit, but they won't be caught at all, which is a little different to being raised.

OddBloke:
Rebasing/updates done, this is ready to land once Travis is happy.

PR #927, opened 2019-11-20
premium: aws handle contract error for non-premium pkcs7 identity

blackboxsw:
Contracts server returns the api error "missing instance information"
when non-premium identity docs are sent to the aws token request route.
Handle this error and emit a message about being a non-premium image.

Also refactor attach_premium function to be a bit easier to test.

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -349,9 +349,9 @@ def _attach_with_token(
 def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
     """Detect cloud_type and request a contract token from identity info.
 
-    :param cfg: Parsed configuration dictionary
+    :param cfg: a ``config.UAConfig`` instance
 
-    :raise NonPremiumImageError: When not on a premium image type. Exits 0.
+    :raise NonPremiumImageError: When not on a premium image type.
     :raise UserFacingError: On unexpected connectivity issues to contract
         server or inability to access identity doc from metadata service.
 
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -346,28 +346,48 @@ def _attach_with_token(
     return 0
 
 
-@assert_not_attached
-@assert_root
-def action_attach_premium(args, cfg):
+def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
+    """Detect cloud_type and request a contract token from identity info.
+
+    :param cfg: Parsed configuration dictionary
+
+    :raise NonPremiumImageError: When not on a premium image type. Exits 0.
+    :raise UserFacingError: On unexpected connectivity issues to contract
+        server or inability to access identity doc from metadata service.
+
+    :return: contract token obtained from identity doc
+    """
     cloud_type = identity.get_cloud_type()
     if cloud_type not in ("aws",):  # TODO(avoid hard-coding supported types)
-        print(
+        raise exceptions.NonPremiumImageError(
             ua_status.MESSAGE_UNSUPPORTED_PREMIUM_CLOUD_TYPE.format(
                 cloud_type=cloud_type
             )
         )
-        return 0
-    # TODO(Clean up the attach_premium flow a bit and add error handling)
     instance = identity.cloud_instance_factory()
     contract_client = contract.UAContractClient(cfg)
     pkcs7 = instance.identity_doc
-    # TODO(Need to handle errors due to non-premium ec2 images)
-    contractTokenResponse = contract_client.request_premium_aws_contract_token(
-        pkcs7
-    )
-    return _attach_with_token(
-        cfg, token=contractTokenResponse["contractToken"], allow_enable=True
-    )
+    try:
+        # TODO(make this logic cloud-agnostic if possible)
+        tokenResponse = contract_client.request_premium_aws_contract_token(
+            pkcs7
+        )
+    except contract.ContractAPIError as e:
+        if contract.API_ERROR_MISSING_INSTANCE_INFORMATION in e:
+            raise exceptions.NonPremiumImageError(
+                ua_status.MESSAGE_UNSUPPORTED_PREMIUM
+            )
+        raise e
+    return tokenResponse["contractToken"]
+
+
+@assert_not_attached
+@assert_root
+def action_attach_premium(args, cfg):
+    token = _get_contract_token_from_cloud_identity(cfg)
+    if token is None:
+        return 0
+    return _attach_with_token(cfg, token=token, allow_enable=True)
 
 
 @assert_not_attached
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -352,8 +352,10 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
     :param cfg: a ``config.UAConfig`` instance
 
     :raise NonPremiumImageError: When not on a premium image type.
-    :raise UserFacingError: On unexpected connectivity issues to contract
+    :raise UrlError: On unexpected connectivity issues to contract
         server or inability to access identity doc from metadata service.
+    :raise ContractAPIError: On unexpected errors when talking to the contract
+        server.
 
     :return: contract token obtained from identity doc
     """
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -385,8 +385,6 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
 @assert_root
 def action_attach_premium(args, cfg):
     token = _get_contract_token_from_cloud_identity(cfg)
-    if token is None:
-        return 0
     return _attach_with_token(cfg, token=token, allow_enable=True)
 
 
--- a/uaclient/exceptions.py
+++ b/uaclient/exceptions.py
@@ -20,7 +20,6 @@ class NonPremiumImageError(UserFacingError):
     """Raised when machine doesn't appear to be premium/pro"""
 
     exit_code = 0
-    pass
 
 
 class AlreadyAttachedError(UserFacingError):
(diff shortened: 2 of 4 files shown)


uaclient/cli.py
```diff
@@ -346,28 +346,38 @@ def _attach_with_token(
...
     pkcs7 = instance.identity_doc
-    # TODO(Need to handle errors due to non-premium ec2 images)
-    contractTokenResponse = contract_client.request_premium_aws_contract_token(
-        pkcs7
-    )
-    return _attach_with_token(
-        cfg, token=contractTokenResponse["contractToken"], allow_enable=True
-    )
+    try:
+        tokenResponse = contract_client.request_premium_aws_contract_token(
```
OddBloke: I think we're still missing a TODO against this line to make it non-EC2 specific.

uaclient/cli.py
```diff
@@ -346,28 +346,38 @@ def _attach_with_token(
     return 0
 
 
-@assert_not_attached
-@assert_root
-def action_attach_premium(args, cfg):
+def _get_contract_token_from_cloud_identity(cfg):
```
OddBloke: Better documentation of what the return value will be would be good (i.e. what does `None` mean in this case). (And perhaps type annotations?)

uaclient/cli.py
```diff
@@ -346,28 +346,38 @@ def _attach_with_token(
...
     cloud_type = identity.get_cloud_type()
     if cloud_type not in ("aws",):  # TODO(avoid hard-coding supported types)
         print(
             ua_status.MESSAGE_UNSUPPORTED_PREMIUM_CLOUD_TYPE.format(
                 cloud_type=cloud_type
             )
         )
-        return 0
-    # TODO(Clean up the attach_premium flow a bit and add error handling)
+        return None
```
OddBloke: Would this and the call site be simpler if we raised a `UserFacingError` with exit_code set to 0?

uaclient/tests/test_cli_attach_premium.py
```diff
@@ -24,9 +28,118 @@ def test_non_root_users_are_rejected(getuid):
...
+class TestGetContractTokenFromCloudIdentity:
+    @pytest.mark.parametrize("cloud_type", ("awslookalike", "azure", "!aws"))
+    @mock.patch("uaclient.clouds.identity.get_cloud_type")
+    def test_non_aws_cloud_type_returns_no_token(
+        self, get_cloud_type, cloud_type, capsys
+    ):
+        """Non-aws clouds will not return a token."""
+        get_cloud_type.return_value = cloud_type
+        assert None is _get_contract_token_from_cloud_identity(FakeConfig())
+        out = capsys.readouterr()[0]
```
OddBloke: A nit: we do `out, _err = capsys.readouterr()` almost everywhere that we only want `out`, it would be good to be consistent. (This applies throughout these new tests, I think.)

uaclient/tests/test_cli_attach_premium.py
```diff
@@ -24,9 +28,118 @@ def test_non_root_users_are_rejected(getuid):
         action_attach_premium(mock.MagicMock(), cfg)
 
 
+class TestGetContractTokenFromCloudIdentity:
+    @pytest.mark.parametrize("cloud_type", ("awslookalike", "azure", "!aws"))
+    @mock.patch("uaclient.clouds.identity.get_cloud_type")
+    def test_non_aws_cloud_type_returns_no_token(
+        self, get_cloud_type, cloud_type, capsys
```
OddBloke: A nit: `get_cloud_type` is more usually spelled `m_get_cloud_type` elsewhere. (I think this applies elsewhere in these tests, but I'll only comment about it once. :+1:)

uaclient/tests/test_cli_attach_premium.py
```diff
@@ -24,9 +28,118 @@ def test_non_root_users_are_rejected(getuid):
...
+    def test_return_token_from_contract_server_using_identity_doc(
+        self,
+        _get_cloud_type,
+        cloud_instance_factory,
+        request_aws_contract_token,
+        capsys,
+    ):
+        """Return token from the contract server using the identity."""
+
+        def fake_instance_factory():
```
OddBloke: This identical definition is repeated several times, should it move to a top-level function that all the tests can share?

uaclient/cli.py
OddBloke: (FAOD, I mean a new `UserFacingError` sub-class.)

uaclient/cli.py
```diff
@@ -346,28 +346,38 @@ def _attach_with_token(
...
+        if contract.API_ERROR_MISSING_INSTANCE_INFORMATION in e:
+            print(ua_status.MESSAGE_UNSUPPORTED_PREMIUM)
+            return None
+        raise e
+    return tokenResponse["contractToken"]
+
+
+@assert_not_attached
+@assert_root
+def action_attach_premium(args, cfg):
```
OddBloke: I really like how this looks now, it's really clear what the high-level logic of a premium attach is. Thanks!

uaclient/tests/test_cli_attach_premium.py
blackboxsw: +1

uaclient/tests/test_cli_attach_premium.py
blackboxsw: agreed.

uaclient/tests/test_cli_attach_premium.py
```diff
@@ -36,51 +149,46 @@ def test_already_attached(self, _m_getuid, capsys):
...
+    @mock.patch(M_PATH + "_get_contract_token_from_cloud_identity")
+    def test_happy_path_on_aws_non_premium(
+        self,
+        get_contract_token_from_cloud_identity,
+        request_updated_contract,
+        _m_getuid,
+    ):
+        """Noop when _get_contract_token_from_cloud_identity finds no token"""
+        get_contract_token_from_cloud_identity.return_value = None
+        args = mock.MagicMock()
```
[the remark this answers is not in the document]
blackboxsw: good thought. done.

uaclient/cli.py
blackboxsw: good call. was going too quickly.

uaclient/cli.py
```diff
@@ -346,28 +346,48 @@ def _attach_with_token(
     return 0
 
 
-@assert_not_attached
-@assert_root
-def action_attach_premium(args, cfg):
+def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
+    """Detect cloud_type and request a contract token from identity info.
+
+    :param cfg: Parsed configuration dictionary
```
OddBloke: A `config.UAConfig` instance is not a dictionary. :-)

uaclient/cli.py
```diff
@@ -346,28 +346,48 @@ def _attach_with_token(
...
+
+    :raise NonPremiumImageError: When not on a premium image type. Exits 0.
```
OddBloke: I'm not sure that the exit code is relevant here; callers may catch the exception and discard it, which would mean we wouldn't exit 0.

uaclient/cli.py
```diff
@@ -346,28 +346,48 @@ def _attach_with_token(
...
+    :raise UserFacingError: On unexpected connectivity issues to contract
```
OddBloke: Should this say `ContractAPIError` instead of `UserFacingError`?

uaclient/cli.py
```diff
@@ -346,28 +346,48 @@ def _attach_with_token(
...
+            )
+        raise e
+    return tokenResponse["contractToken"]
+
+
+@assert_not_attached
+@assert_root
+def action_attach_premium(args, cfg):
+    token = _get_contract_token_from_cloud_identity(cfg)
+    if token is None:
```
OddBloke: `token` should never be `None`, we'll now raise exceptions in the cases it would have been.

uaclient/exceptions.py
```diff
@@ -16,6 +16,13 @@ def __init__(self, msg: str) -> None:
         self.msg = msg
 
 
+class NonPremiumImageError(UserFacingError):
+    """Raised when machine doesn't appear to be premium/pro"""
+
+    exit_code = 0
+    pass
```
OddBloke: No need for `pass` here.

uaclient/cli.py
OddBloke: Looking at this now.

uaclient/cli.py
OddBloke: Looking at this now.

uaclient/cli.py
OddBloke: I'm not going to make this change because I want @blackboxsw to confirm.

uaclient/cli.py
OddBloke: Fix pushed.

uaclient/cli.py
OddBloke: Fix pushed.

uaclient/cli.py
OddBloke: Looking at this now.

OddBloke: Fix pushed.

uaclient/exceptions.py
OddBloke: Fix pushed.

OddBloke:
OK, I've fixed up most of my review nits/comments, I have one remaining question at https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/927#discussion_r349113147

uaclient/cli.py
blackboxsw: UserFacingError is incorrect, it should be UrlError for connetivity issues (in serviceclient.py) and ContractAPIError if error in talking to the backend. I've updated the docstring

PR #928, opened 2019-11-20
initial implementation of esm-apps

OddBloke:
This introduces a new ESMAppsEntitlement, which will handle the enabling
and management of ESM for Apps.

This simply reuses the existing esm-infra keyring (by copying it), so
that some basic testing can be performed.


uaclient/entitlements/esm.py
```diff
@@ -1,13 +1,22 @@
 from uaclient.entitlements import repo
 
 
-class ESMInfraEntitlement(repo.RepoEntitlement):
-
+class ESMBaseEntitlement(repo.RepoEntitlement):
     help_doc_url = "https://ubuntu.com/esm"
+    origin = "UbuntuESM"
+    repo_pin_priority = "never"
+    disable_apt_auth_only = True  # Only remove apt auth files when disabling
```
blackboxsw: I think now we need to change this to a @property class method that looks at our util.get_platform_info()['series'] because on trusty is the only place we want to continue advertising esm apt sources right?

OddBloke: Great point, let me look at that.

blackboxsw:
Casey gave us a token for staging that has access to both esm-apps and infra. There is some coupling there in this current implementation that needs to be sorted as both get enabled when one is of the services is enabled. And if you detach one, you get a trace trying to attach the second

uaclient/entitlements/esm.py
blackboxsw: Also pushed the following fixes:

- unset ESM*.repo_pin_priorityand
- fixed UbuntuESM vs UbuntuESMApps origin
- more strict aptURL matching in apt-policy

blackboxsw:
Pushed 3 commits to decouple ESMApps from ESMInfra using Casey's token .

OddBloke:
@blackboxsw I've reviewed your added commits and they look good, thank you for following up on this change! Will leave this for you to land once you're in, as you're still marked as requesting changes and I want to be sure that we're not missing anything.

PR #932, opened 2019-11-21
esm-apps: allow unattended security upgrades for esm-apps

blackboxsw:


apt.conf.d/51ubuntu-advantage-esm
```diff
@@ -1,3 +1,6 @@
 Unattended-Upgrade::Allowed-Origins {
   "${distro_id}ESM:${distro_codename}-infra-security";
```
powersj: Does this need to be updated to ESMInfra? To be honest I'm not sure where the ESMApps name comes from or if it needs to be something specific.

blackboxsw: Nope, ESMInfra isn't there because of legacy support 
full successful unattended upgrade run https://pastebin.canonical.com/p/NBHyPNmVN9/

Issue #940, opened 2019-11-27
esm: trusty apt disabling is no longer pinned never


blackboxsw:
In getting bionic/xenial to work for UA Premium, we dropped repo pin handling which allowed trusty to advertise available esm package even when esm-infra is disabled. Re-introduce this behavior on trusty only for esm-infra only.  Xenial and Bionic should remove all auth and repo files when esm-* services are disabled.

PR #942, opened 2019-11-28
esm: restore trusty apt pinning behavior

blackboxsw:
Differentiate ESMInfra's repo_pin_pririority and disable_apt_auth_only

Retain Xenial and Bionic behavior for ESMInfra which will behave like
ESMApps which removes any apt artifacts (preferences, auth or list
files) when ESMInfra is disabled.

Fixes: #940

### DIFF
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -29,10 +29,14 @@ class RepoEntitlement(base.UAEntitlement):
     origin = None  # type: Optional[str]
 
     # Optional repo pin priority in subclass
-    repo_pin_priority = None  # type: Union[int, str, None]
+    @property
+    def repo_pin_priority(self) -> "Union[int, str, None]":
+        return None
 
     # disable_apt_auth_only (ESM) to only remove apt auth files on disable
-    disable_apt_auth_only = False  # Set True on ESM to only remove apt auth
+    @property
+    def disable_apt_auth_only(self) -> bool:
+        return False  # Set True on ESM to only remove apt auth
 
     # Any custom messages to emit pre or post enable or disable operations;
     # currently post_enable is used in CommonCriteria
@@ -271,9 +275,9 @@ class RepoEntitlement(base.UAEntitlement):
         if not repo_url:
             raise exceptions.MissingAptURLDirective(self.name)
         if self.disable_apt_auth_only:
-            # We only remove the repo from the apt auth file, because ESM
+            # We only remove the repo from the apt auth file, because ESM-infra
             # is a special-case: we want to be able to report on the
-            # available ESM updates even when it's disabled
+            # available ESM-infra updates even when it's disabled
             apt.remove_repo_from_apt_auth_file(repo_url)
             apt.restore_commented_apt_list_file(repo_filename)
         else:
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -275,9 +275,9 @@ class RepoEntitlement(base.UAEntitlement):
         if not repo_url:
             raise exceptions.MissingAptURLDirective(self.name)
         if self.disable_apt_auth_only:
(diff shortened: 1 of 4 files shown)


uaclient/entitlements/tests/test_esm.py
```diff
@@ -204,6 +235,8 @@ def test_disable_removes_apt_config(
         m_platform_info,
         m_remove_auth_apt_repo,
         m_remove_apt_list_files,
+        m_remove_repo_from_apt_auth_file,
```
blackboxsw: Ok, mocks are getting a bit out of control and maybe this unit test is trying to do too much. I'll mark WIP and let @OddBloke decide whether we need to make this more testable or make the unit test more concise now that we have quite distinct behaviors between esm-infra and esm-apps.

blackboxsw:
@OddBloke I  marked WIP because of the funky final unit circus I have in test_esm.py.

HEre's validation on trusty:https://pastebin.canonical.com/p/rmggSb427Z/

uaclient/entitlements/tests/test_esm.py
OddBloke: Given the enormous if/else in here, I think splitting into separate tests makes sense. It's worth considering, however, that there may be _more_ than two tests here: there may be some common behaviour we're testing in this test which can continue to be shared between the two.

blackboxsw:
> I've responded to your one comment inline, but I have a high-level comment: we need our tests to exercise more than just the trusty code paths to be sure that the behaviour implemented here is correct.
> 
> We may want to address that separately and then revisit this PR after we've done that, what do you think?

I agree, I think we may need to add a bit more coverage for non-trusty cases across the board. That generally should probably live as a tech-debt item to follow up with after this branch. But, in the meantime I've added a number of unit tests to cover ESMInfra/ESMApps specifically on trusty and non-trusty paths.

uaclient/entitlements/tests/test_esm.py
blackboxsw: now that test_repo.py covers most of setup_apt_config and remove_apt_config, we don't need to explicitly test that interaction within esm tests. I've updated unit tests and this is ready for re-review.

blackboxsw:
un "wip" ing now that unit tests have been added and refactored.

powersj:
@OddBloke ready for re-review

uaclient/entitlements/repo.py
```diff
@@ -271,9 +275,9 @@ def remove_apt_config(self):
         if not repo_url:
             raise exceptions.MissingAptURLDirective(self.name)
         if self.disable_apt_auth_only:
-            # We only remove the repo from the apt auth file, because ESM
+            # We only remove the repo from the apt auth file, because ESM-infra
```
OddBloke:
```suggestion
            # We only remove the repo from the apt auth file, because ESM Infra
```

uaclient/entitlements/repo.py
```diff
@@ -271,9 +275,9 @@ def remove_apt_config(self):
...
             # is a special-case: we want to be able to report on the
-            # available ESM updates even when it's disabled
+            # available ESM-infra updates even when it's disabled
```
OddBloke:
```suggestion
            # available ESM Infra updates even when it's disabled
```

uaclient/entitlements/tests/test_esm.py
```diff
@@ -97,6 +155,17 @@ def fake_exists(path):
         assert add_apt_calls == m_add_apt.call_args_list
         assert 0 == m_add_pinning.call_count
         assert subp_calls == m_subp.call_args_list
+        if entitlement.name == "esm-infra":  # Then remove apt pref pin never
```
OddBloke: A nit, but I initially read this as "Then remove apt pref pin, never". Perhaps just:

```suggestion
        if entitlement.name == "esm-infra":  # Remove "never" apt pref pin
```

PR #956, opened 2020-01-07
azure-pro: generalize contract.request_contract_token and identity_doc

blackboxsw:
* auto-attach: move duplicate invalid cloud_type check out of cli
* contract: generalize request_aws_contract_token for multiple cloud_types
* clouds: generalize identity_doc to return dict instead of string


blackboxsw:
wrong target branch, resubmitting

PR #957, opened 2020-01-07
contract: generalize request-contract-token method and aws identity_doc

blackboxsw:
Groundwork for enabling Azure's auto-attach images. We generalize common methods so that
any cloud can call them with the appropriate identity doc data.

In this PR:
*  auto-attach: move duplicate invalid cloud_type check out of cli
*  contract: generalize request_aws_contract_token for multiple cloud_types
*  clouds: generalize identity_doc to return dict instead of string


uaclient/cli.py
```diff
@@ -356,22 +356,18 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
         server or inability to access identity doc from metadata service.
     :raise ContractAPIError: On unexpected errors when talking to the contract
         server.
+    :raise NonAutoAttachImageError: If this cloud type does not have
+        auto-attach support.
 
     :return: contract token obtained from identity doc
     """
     cloud_type = identity.get_cloud_type()
```
OddBloke: What do you think about adding a `AutoAttachCloudInstance.cloud_type` attribute, and changing `request_auto_attach_contract_token` to only take `instance`? Now that we aren't checking `cloud_type` here, it seems a little out of place.

blackboxsw: I like it. just pushed changes

tools/constraints-trusty.txt
```diff
@@ -1,3 +1,5 @@
+# pyyaml dropped python 3.4 support Nov 2019, pin to trusty version
```
OddBloke: I don't think we need a specific explanatory comment; we should probably have been constraining its version all along, for our testing to be closer to the target system:

```suggestion
```

We should also constrain all of the other releases, for the same reason, but that can be a separate PR.

uaclient/contract.py
```diff
@@ -103,18 +104,20 @@ def request_resources(self) -> "Dict[str, Any]":
         )
         return resource_response
 
-    def request_aws_contract_token(self, pkcs7: str):
-        """Requests contract token for auto-attach images on AWS.
+    def request_auto_attach_contract_token(
+        self, instance: clouds.AutoAttachCloudInstance
```
OddBloke: We only pass this as a keyword argument, so may as well:

```suggestion
        self, *, instance: clouds.AutoAttachCloudInstance
```

blackboxsw:
Thanks for this @OddBloke I agree that the general discussion about pinning constraints to the known version active in the distro is probably desired, regardless of whether or not the upstream version happens to work. We don't want to end up broken by depedencies out on the internet in the future as we were with this branch.

PR #958, opened 2020-01-08
azure-pro: part 2, add azure auto-attach support and detection to uaclient

blackboxsw:
The final Azure Pro support branch: 

*   azure-pro: add azure cloud instance module and tests
*  azure-pro: add azure auto-attach instance as valid cloud_instance_factory

### DIFF
--- a/uaclient/clouds/azure.py
+++ b/uaclient/clouds/azure.py
@@ -2,18 +2,26 @@ import os
 
 from urllib.error import HTTPError
 
+try:
+    from typing import Any, Dict  # noqa: F401
+except ImportError:
+    # typing isn't available on trusty, so ignore its absence
+    pass
+
 from uaclient.clouds import AutoAttachCloudInstance
 from uaclient import util
 
 
 IMDS_BASE_URL = "http://169.254.169.254/metadata/"
+
+API_VERSION = "2019-06-04"  # Needed to get subscription ID in attested data
 IMDS_URLS = {
-    "pkcs7": IMDS_BASE_URL + "attested/document?api-version=2019-06-04",
-    "compute": IMDS_BASE_URL + "instance/compute?api-version=2019-06-04",
+    "pkcs7": IMDS_BASE_URL + "attested/document?api-version=" + API_VERSION,
+    "compute": IMDS_BASE_URL + "instance/compute?api-version=" + API_VERSION,
 }
 
-SYS_HYPERVISOR_PRODUCT_UUID = "/sys/hypervisor/uuid"
 DMI_CHASSIS_ASSET_TAG = "/sys/class/dmi/id/chassis_asset_tag"
+AZURE_OVF_ENV_FILE = "/var/lib/cloud/seed/azure/ovf-env.xml"
 AZURE_CHASSIS_ASSET_TAG = "7783-7084-3265-9085-8269-3286-77"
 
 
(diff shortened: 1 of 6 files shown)


uaclient/clouds/azure.py
```diff
@@ -0,0 +1,55 @@
...
+
+from urllib.error import HTTPError
+
+from uaclient.clouds import AutoAttachCloudInstance
+from uaclient import util
+
+
+IMDS_BASE_URL = "http://169.254.169.254/metadata/"
+IMDS_URLS = {
+    "pkcs7": IMDS_BASE_URL + "attested/document?api-version=2019-06-04",
```
OddBloke: Where did this API version come from?

uaclient/clouds/azure.py
```diff
@@ -0,0 +1,55 @@
...
+    "compute": IMDS_BASE_URL + "instance/compute?api-version=2019-06-04",
+}
+
+SYS_HYPERVISOR_PRODUCT_UUID = "/sys/hypervisor/uuid"
+DMI_CHASSIS_ASSET_TAG = "/sys/class/dmi/id/chassis_asset_tag"
+AZURE_CHASSIS_ASSET_TAG = "7783-7084-3265-9085-8269-3286-77"
+
+
+class UAAutoAttachAzureInstance(AutoAttachCloudInstance):
```
OddBloke: Can we get type annotations in here, please?

uaclient/clouds/azure.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # mypy does not handle @property around inner decorators
+    # https://github.com/python/mypy/issues/1362
+    @property  # type: ignore
+    @util.retry(HTTPError, retry_sleeps=[1, 2, 5])
+    def identity_doc(self):
+        response = {}
```
OddBloke: This is a nit for sure, but "response" in the context of the IMDS makes me assume that this will represent an HTTP response, whereas this is simply an internal data structure. Maybe even `s/response/responses/` would clear that up?

uaclient/clouds/azure.py
blackboxsw: This came from cmars comment in the spec "We need to use 2019-06-04 or later, so that the subscription ID is in the attested metadata." 
https://docs.google.com/document/d/1PzLLAKpz9x4FlsYFC7iAFN533d_Kl8xMxH9ahAOc5P8/edit#

OddBloke: OK, I've asked Casey in doc if he can point us at the origin of that. I'd like to see a comment explaining why we're using that version, as it's obviously _very_ important that we don't change it backwards (which we might be tempted/asked to do as not all Azure partitions have support for this API version: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#service-availability).

blackboxsw:
Thanks for the review. I believe I incorporated all your suggestions and added a comment about why we use API_VERSION="2019-06-04"

uaclient/tests/test_cli_auto_attach.py
```diff
@@ -38,7 +38,9 @@ def fake_instance_factory(self):
         m_instance.identity_doc = "pkcs7-validated-by-backend"
         return m_instance
 
-    @pytest.mark.parametrize("cloud_type", ("awslookalike", "azure2", "!aws"))
```
OddBloke: I'd actually prefer to keep this one in, as it ensures we aren't testing for a _prefix_ of "aws".

PR #966, opened 2020-01-28
fix text parameter for pre3.7 Python versions

patriciadomin:
The text parameter was added to subprocess.run, as a more understandable alias
of universal_newlines, in Python 3.7.
This allows users to run the behave tests in Pre 3.7 Python versions.

### DIFF
--- a/features/util.py
+++ b/features/util.py
@@ -34,6 +34,7 @@ def lxc_exec(
     container_name: str,
     cmd: List[str],
     capture_output: bool = False,
+    text: bool = False,
     **kwargs: Any
 ) -> subprocess.CompletedProcess:
     """Run `lxc exec` in a container.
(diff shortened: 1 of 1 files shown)


features/util.py
```diff
@@ -65,6 +71,9 @@ def lxc_exec(
                 "stdout and stderr arguments may not be used "
                 "with capture_output."
             )
+        # stdout and stderr will be opened in text mode (by default they are
+        # opened in binary mode
+        kwargs["universal_newlines"] = True
```
OddBloke: This looks like we will _always_ set `universal_newlines` if `capture_output` is given; should we instead only be setting it when the `text` parameter is `True`?

patriciadomin: Yes, I added it because if you do not set `universal_newlines = True` you'get this: `TypeError: a bytes-like object is required, not 'str'`.

OddBloke:
```suggestion
        if text:
            kwargs["universal_newlines"] = True
```

OddBloke:
```suggestion
        kwargs["universal_newlines"] = text
```

OddBloke:
Thanks!

PR #968, opened 2020-01-30
auto-attach: handle 403 errors raised by contract server for invalid vms

### LINKED ISSUE #939: Improve the error message on failed auto-attach
The current message if the contract server doesn't recognise the given product code is:

```
$ sudo ua auto-attach
```

In cases where people discover and run `auto-attach`, we could do a better job of indicating that this is never going to work, it isn't a temporary condition, and they shouldn't try again.

blackboxsw:
Note that some errors from the contracts server don't set HTTP status and raise 403.

Errors like API_ERROR_MISSING_INSTANCE_INFORMATION for AWS resulted in 500s on the server and have now been changed per https://github.com/CanonicalLtd/ua-contracts/pull/517/files to return HTTP status code 404

Fixes #939

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -15,7 +15,6 @@ except ImportError:
     pass
 
 API_ERROR_INVALID_TOKEN = "invalid token"
-API_ERROR_MISSING_INSTANCE_INFORMATION = "missing instance information"
 API_V1_CONTEXT_MACHINE_TOKEN = "/v1/context/machines/token"
 API_V1_TMPL_CONTEXT_MACHINE_TOKEN_UPDATE = (
     "/v1/contracts/{contract}/context/machines/{machine}"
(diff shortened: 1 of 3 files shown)


uaclient/tests/test_cli_auto_attach.py
```diff
@@ -63,8 +74,11 @@ def test_aws_cloud_type_non_auto_attach_returns_no_token(
         _get_cloud_type,
         cloud_instance_factory,
         request_auto_attach_contract_token,
+        http_msg,
```
OddBloke: These aren't used in the body of the test.

uaclient/contract.py
```diff
@@ -15,7 +15,6 @@
     pass
 
 API_ERROR_INVALID_TOKEN = "invalid token"
-API_ERROR_MISSING_INSTANCE_INFORMATION = "missing instance information"
```
OddBloke: Does this error now return a 4xx code?

blackboxsw: Yes this missing instance information error is specifically what was fixed with two other failure conditions to return a 404:
storeutil.ErrAmbiguousProductMapping,
storeutil.ErrMissingProductMapping,
storeutil.ErrMissingInstanceInformation:

More context:
https://github.com/CanonicalLtd/ua-contracts/pull/517/files#r342118256

Issue #952, opened 2019-12-02
Failure to enable an entitlement on auto-attach results in non-root status not being updated


OddBloke:
Not good.

PR #969, opened 2020-01-30
attach: persist any status config changes after attach failures

blackboxsw:
There can be cases where auto-enablement of a service results in an
error. Or, intermittent connectivity to ua-contracts routes end up
resulting in an error during `ua attach`.

In the event that attach succeeds, but auto-enablement actions fail
status needs to be persisted to reflect whether attach succeeded and
any services have been successfully enabled.

Fixes: #952

### DIFF
--- a/uaclient/tests/test_cli_attach.py
+++ b/uaclient/tests/test_cli_attach.py
@@ -1,3 +1,4 @@
+import copy
 import mock
 
 from uaclient.testing.fakes import FakeConfig
@@ -22,6 +23,7 @@ M_PATH = "uaclient.cli."
 # Also used in test_cli_auto_attach.py
 BASIC_MACHINE_TOKEN = {
     "availableResources": [],
+    "machineToken": "non-empty-token",
     "machineTokenInfo": {
         "contractInfo": {
             "name": "mycontract",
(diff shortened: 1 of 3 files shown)


uaclient/tests/test_cli_attach.py
```diff
@@ -62,6 +97,26 @@ def test_token_is_a_required_argument(self, _m_getuid):
             action_attach(args, FakeConfig())
         assert status.MESSAGE_ATTACH_REQUIRES_TOKEN == str(e.value)
 
+    @mock.patch(M_PATH + "contract.request_updated_contract")
+    def test_status_updated_when_auto_enable_fails(
```
OddBloke: This test passes against master as-is, so I don't think it's testing the added code appropriately.

blackboxsw: done. added a missing mock to this test which only succeeded as we reached out to the real contract server due to contract.get_available_resources()

OddBloke: This test still hasn't been updated. If I apply this patch to your branch:

```diff
diff --git a/uaclient/cli.py b/uaclient/cli.py
index 3f44c0e..3e02312 100644
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -329,11 +329,9 @@ def _attach_with_token(
         with util.disable_log_to_console():
             logging.exception(exc)
         print(ua_status.MESSAGE_ATTACH_FAILURE)
-        cfg.status()  # Persist updated status in the event of partial attach
         return 1
     except exceptions.UserFacingError as exc:
         logging.warning(exc.msg)
-        cfg.status()  # Persist updated status in the event of partial attach
         return 1
     contract_name = cfg.machine_token["machineTokenInfo"]["contractInfo"][
         "name"
```

then this test still passes:

```
$ pytest

<snip>

==================== 553 passed in 1.87 seconds =====================
```

which demonstrates that it is not testing the change proposed in this PR. Can you modify it so that it is testing the new behaviour, please?

blackboxsw: Yep, good point, I wasn't asserting deep enough. Added logic to exercise poth cfg.status() paths upon UrlError and UserFacingError.

Issue #941, opened 2019-11-27
trusty: re-add apt-hook to trusty package


blackboxsw:
In commitish 07c06893608d2198e6501a9bd1faa42f5bc8a0ff we no longer install apt-hook in tip so that Xenial and Bionic don't try to run this hook.

This breaks trusty which needs the hook to advertise esm-infra packages.

We need to find a way to package this hook in trusty and have it only run advertise esm-infra packages.

blackboxsw:
@OddBloke landed a branch related to this and I think we just need to verify on xenial/bionic that correct behavior is there without unintended side-effects

powersj:
Once #940 is landed, @OddBloke to test this on Trusty and Xenial/Bionic from the daily PPA. Then this issue can be closed if things are working.

OddBloke:
The next step: I need to perform the described testing. Then either fix issues and close this out, or simply close this out.

OddBloke:
I've tested the daily PPA on trusty and xenial; trusty shows the hook output, xenial does not.

PR #971, opened 2020-01-31
d/rules: install apt hook on trusty only

OddBloke:


OddBloke:
This is definitely related to #941, but there are likely other pieces we need to ensure that the hook actually has the data it needs to function, so I'm not marking it as closed by this.

Issue #977, opened 2020-02-05
azure-pro: trusty cloud detection needs to match "azurenet" platform


blackboxsw:
On Trusty Azure datasource is represented in `/run/cloud-init/result.json` as  "azurenet"
```
ubuntu@trusty-ubuntu-pro-test:~$ sudo cat /run/cloud-init/result.json 
{
 "v1": {
  "errors": [], 
  "datasource": "DataSourceAzureNet [seed=/dev/sr0]"
 }
}
```

This results in a invalid rejection of azure platform:

```
sudo ua auto-attach
Auto-attach image support is not available on azurenet
See: https://ubuntu.com/advantage
```

PR #978, opened 2020-02-05
azure-pro: fix detection of DatasourceAzureNet as azure on trusty

blackboxsw:
Fixes: #977


blackboxsw:
locally built deb confirmed working same failure signature as https://github.com/CanonicalLtd/ua-contracts/issues/632
09:50 and cloud-detection valid

PR #981, opened 2020-02-06
add focal builds to CI and fix focal build failure

OddBloke:
We recently saw our first in-focal regression, so now is the time to start gating on it (and to fix the regression, of course!).


OddBloke:
WIP because I'm looking to see this fail before I push up the fix, so I can be sure that we would actually have caught the regression by doing this.

OddBloke:
No longer WIP, as the fix is pushed and the focal package build in Travis is now green.

PR #982, opened 2020-02-06
pkg: introduce a separate ubuntu-advantage-pro package v. 20.2

### LINKED ISSUE #974: Standard instances (non-pro) should not run attempt to auto-attach
Currently in master, a systemd or init script calls `ua auto-attach` to attempt to auto-attach any vm to UA. 

The same ubuntu-advantage-tools package should be able to be installed on standard and Ubuntu Pro vms without resulting in an error condition at boot.

UA client needs to better discover platform type before attempting to auto-attach against ua-contracts on "standard" (non-Pro) Ubuntu instances.

OddBloke:
In 20.2, a separate ubuntu-advantage-pro package will deliver the systemd or upstart unit which performs the Ubuntu Pro auto-attach.

Package separation of ua-pro from ua-tools allows the ubuntu-advantage-tools package to avoid complexity in determining whether a platform is certified as Ubuntu Pro before running the systemd or upstart auto-attach functionality on boot.

By delivering the systemd or upstart units in a separate deb package, both the init scripts and ubuntu-advantage-tools package can remain simple because the auto-attach features will only be
present and "active" when ubuntu-advantage-pro package is installed, which will only be on Ubuntu Pro images.

Due to the migration of upstart units and systemd services out of ubuntu-advantage-tools package,
the following changes have been made to make this breaking package upgrade cleaner:
* ubuntu-advantage.maintscript performs an rm_conffile /etc/init/ua-auto-attach.conf 
* postinst to cleanup stale systemd symlinks when upgrading from 19.7-20.1 (versions on Pro images)
* postinst to emit a debconf note during attended upgrade to alert about the need to install ubuntu-advantage-pro if upgrading from a known ubuntu-advantage-tools version on Ubuntu Pro

Fixes #974

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-ubuntu-advantage-tools (20.1) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (20.1) xenial; urgency=medium
 
   * Release 20.1:
     - azure-pro, support for azure ubuntu pro auto-attach:
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,24 @@
+ubuntu-advantage-tools (20.2) UNRELEASED; urgency=medium
+
+  * d/templates: add a debconf note on upgrade from pre-ubuntu pro package
+  * d/control: create a separate ubuntu-advantage-pro package which
+      delivers the tooling and scripts necessary to auto-attach pro machines
+      This change breaks/replaces ubuntu-advantage-tools <= 20.1
+  * d/maintscript: rm_conffile /etc/init/ua-auto-attach.conf from ua-tools pkg
+  * d/postint: remove stale systemd symlinks which have migrated to ubuntu-pro
+  * d/rules: only install the apt hook on trusty
+  * d/rules: provide --no-start to debhelper to avoid auto-attach on pkg install
+  * Release 20.2:
+    - ubuntu-pro:
+      + azure: fix detection of DatasourceAzureNet as azure on trusty
+      + generalize identity_doc to return dict instead of string
+      + auto-attach: any 4XX errors during auto-attach are the result of non-Pro
+      + auto-attach: handle 403 errors raised by contract server for invalid vms
+    - attach: persist any status config changes after attach failures
+    - output: add messaging using a different subscription if attached
+
+ -- Chad Smith <chad.smith@canonical.com>  Thu, 20 Feb 2020 11:13:15 -0700
+
 ubuntu-advantage-tools (20.1) xenial; urgency=medium
 
   * Release 20.1:
--- a/debian/control
+++ b/debian/control
@@ -34,3 +34,9 @@ Description: management tools for Ubuntu Advantage
  .
  Subscribers to Ubuntu Advantage will find helpful tools for accessing
  services in this package.
+
+Package: ubuntu-advantage-pro
+Architecture: any
+Depends: ubuntu-advantage-tools
+Description: utilities and services for Ubuntu Pro images
+ FIXME
--- a/debian/control
+++ b/debian/control
@@ -37,6 +37,9 @@ Description: management tools for Ubuntu Advantage
 
 Package: ubuntu-advantage-pro
 Architecture: any
-Depends: ubuntu-advantage-tools
+Depends: ubuntu-advantage-tools (>=20.2)
+Replaces: ubuntu-advantage-tools (<<20.2)
+Breaks: ubuntu-advantage-tools (<<20.2)
 Description: utilities and services for Ubuntu Pro images
- FIXME
+ The Ubuntu Pro package delivers additional utilities for use on authorised
+ Ubuntu Pro machines.
--- a/debian/ubuntu-advantage-tools.maintscript
+++ b/debian/ubuntu-advantage-tools.maintscript
@@ -2,3 +2,4 @@ rm_conffile /etc/update-motd.d/99-esm 19.1~ ubuntu-advantage-tools
 rm_conffile /etc/update-motd.d/80-esm 19.1~ ubuntu-advantage-tools
 rm_conffile /etc/update-motd.d/80-livepatch 19.1~ ubuntu-advantage-tools
 rm_conffile /etc/cron.daily/ubuntu-advantage-tools 19.1~ ubuntu-advantage-tools
+rm_conffile /etc/init/ua-auto-attach.conf 20.2~ ubuntu-advantage-tools
--- a/debian/postinst
+++ b/debian/postinst
@@ -21,6 +21,11 @@ ESM_INFRA_APT_PREF_FILE_TRUSTY="/etc/apt/preferences.d/ubuntu-esm-infra-trusty"
 MYARCH="$(dpkg --print-architecture)"
 ESM_SUPPORTED_ARCHS="i386 amd64"
 
+SYSTEMD_WANTS_AUTO_ATTACH_LINK="/etc/systemd/system/multi-user.target.wants/ua-auto-attach.service"
+SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enabled/ua-auto-attach.service.dsh-also"
+SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
+
+
 unconfigure_esm() {
     rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
     rm -f $APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY_TRUSTY
@@ -81,6 +86,10 @@ case "$1" in
       # We broke package compatibility in 20.2 for any image with 19.7
       if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt-nl "20.2~"; then
           if dpkg --compare-versions "$PREVIOUS_PKG_VER" ge-nl "19.7~"; then
+              # Drop stale symlinks for migrated auto-attach-service
+              rm -f $SYSTEMD_WANTS_AUTO_ATTACH_LINK
+              rm -f $SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH
+              rm -f $SYSTEMD_HELPER_ENABLED_WANTS_LINK
               # Use debconf to alert the user to the additional
               # ubuntu-advantage-pro package that should be installed
               . /usr/share/debconf/confmodule
--- a/debian/postinst
+++ b/debian/postinst
@@ -4,6 +4,7 @@ set -e
 
 . /etc/os-release  # For VERSION_ID
 
+
 APT_TRUSTED_KEY_DIR="/etc/apt/trusted.gpg.d"
 UA_KEYRING_DIR="/usr/share/keyrings/"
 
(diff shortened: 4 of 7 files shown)


powersj:
@panlinux any objection to this approach?

blackboxsw:
Should ubuntu-advantage-pro package remove the init scripts by default in postrm, not just on purge?

```
apt-get remove ubuntu-advantage-pro
updatedb
locate auto-attach
/etc/init/ua-auto-attach.conf
```


Installs fine:
```
ubuntu@trusty-ubuntu-pro-test:~$ sudo dpkg -i ubuntu-advantage-tools_20.1_amd64.deb 
(Reading database ... 29107 files and directories currently installed.)
Preparing to unpack ubuntu-advantage-tools_20.1_amd64.deb ...
Unpacking ubuntu-advantage-tools (20.1) over (20.1) ...
Setting up ubuntu-advantage-tools (20.1) ...
Processing triggers for man-db (2.6.7.1-1ubuntu1) ...
ubuntu@trusty-ubuntu-pro-test:~$ sudo dpkg -i ubuntu-advantage-pro_20.1_amd64.deb 
(Reading database ... 29107 files and directories currently installed.)
Preparing to unpack ubuntu-advantage-pro_20.1_amd64.deb ...
Unpacking ubuntu-advantage-pro (20.1) ...
Setting up ubuntu-advantage-pro (20.1) ...
Processing triggers for ureadahead (0.100.0-16) ...
```

debian/control
```diff
@@ -34,3 +34,9 @@ Description: management tools for Ubuntu Advantage
  .
  Subscribers to Ubuntu Advantage will find helpful tools for accessing
  services in this package.
+
+Package: ubuntu-advantage-pro
+Architecture: any
+Depends: ubuntu-advantage-tools
+Description: utilities and services for Ubuntu Pro images
```
blackboxsw: Should we add a conflicts ubuntu-advantage-tools  19.6 since that version of ubuntu-advantage-tools in aws pro already released and has the init scripts in it.

OddBloke: Good question. Another way of addressing that would be to have `ubuntu-advantage-pro` depends on the specific version of `ubuntu-advantage-tools` it was built with, which would (I think) force the two to be upgraded in lockstep.

We may want to do that regardless, even if it's a bit of heavy hammer for the specific issue you raised.

@panlinux What do you think?

OddBloke:
> Should ubuntu-advantage-pro package remove the init scripts by default in postrm, not just on purge?

I've just done some investigation, and it looks to me like this is only a problem on trusty. It looks like this is because the upstart init script is installed as a conffile, whereas the systemd one is not. I don't know enough about upstart/trusty packaging to know if this is expected, or if this indicates that we aren't installing it correctly in the first place. @panlinux ?

basak:
Considering just Focal (ie. not older versions of dh-systemd or upstart etc), I think you have an upgrade path problem with this current approach. A user who has just ubuntu-advantage-tools installed will upgrade without getting -pro. `lib/systemd/system/ua-auto-attach.service` will vanish, but the service will remain enabled in `/etc/systemd/system`. I haven't tested this.

As far as I can tell, Debian doesn't currently have good tooling to handle this situation. So it'll need to be handled manually, and all edge cases considered. Making sure to get this right is going to be quite painful.

Consider that packaging has always provided mechanisms to enable or disable individual services shipped by them. So instead of using a package installation as a proxy for whether a service should be enabled, could you instead use those mechanisms in the existing single package? In systemd you could use a ConditionX directive, or just not enable the service by customising the `dh_systemd_enable` invocation at package build time, for example. I think the equivalent can probably be managed in upstart for Trusty.

If this isn't possible then as well as handling the various upgrade paths you'll also need Breaks/Replaces to cover the move of `lib/systemd/system/ua-auto-attach.service` from one package to another. See https://wiki.debian.org/PackageTransition for details (I think you have case 7 plus the Depends in the other direction).

OddBloke:
Thank you for the detailed review, @basak!

@powersj @blackboxsw It sounds to me like the fact we had to rush for the AWS Pro launch, and didn't have time to properly consider how to do this, has put us into this difficult position going forward[0]. We either need to stick with one package indefinitely, or we need to take on the work (probably with the assistance of a seasoned packager) of managing the migration from one to two packages. I'm not really sure what the best path forward here is. Thoughts?

[0] Who could have predicted?!

blackboxsw:
> Thank you for the detailed review, @basak!
Definitely.

> We either need to stick with one package indefinitely, 
I'm worried about the cost of us adding smarts into uaclient for ConditionalX= Ubuntu Pro support as the data coming from AWS and Azure Pro IMDS is dynamic and signed we'd have to bake in certificate verfication in the client which is a lot of duplicated logic that the ua-contracts backend has a handle on. That logic is also slightly different for each Pro cloud that we add, so there would be a lot more moving pieces in UAClient that could break or need maintenance.


> or we need to take on the work (probably with the assistance of a seasoned packager) of managing the migration from one to two packages. 
I feel like this is the cost we need to eat to keep packaging and init scripts simple and avoid burdening UA client with lots of logic/modules to vet and validate certs or environment variables to know that it is **not** running on a Pro instance during each boot.

We can fairly easily detect in debian/*preinst that we've upgraded from a known AWS Pro ubuntu-advantate-tools v 20.1 and if the auto-attach.service file is present and needs migration.

But, the biggest question/risk with the 2 package approach is if there is a way for ua-tools to conditionall yforce a pkg dependency on ubuntu-advantage-pro package that was not expressed in the static debian/control file during package build.

That's where I hit a wall with the two package approach.


> 
> [0] Who could have predicted?!
Yeah, I wish someone had raised a concern about the carrying costs of rushing this effort. :)

blackboxsw:
Also since AWS Ubuntu PRO is the only released ua-tools with auto-attach, we might be able to
inspect security_groups or the instance_identity doc for marketplaceProduct codes 
```
$ sudo cloud-init query ds.meta_data | grep security_groups
     "security_groups": "Ubuntu Pro 18-04 LTS-Ubuntu Pro 18-04 LTS 20190828-AutogenByAWSMP-2",
 "security_groups": "Ubuntu Pro 18-04 LTS-Ubuntu Pro 18-04 LTS 20190828-AutogenByAWSMP-2",
ubuntu@ip-172-31-92-120:~$ sudo cloud-init query ds.dynamic.instance_identity.document.marketplaceProductCodes
[
 "ex9l8qw4x21426rsfg7am8gym"
]
```

blackboxsw:
Proposed the following PR into this branch for review which should separate the two packages, assemble the Breaks/Requires/Replaces debian/control stuff, and rm_conffile the errant /etc/init/ua-auto-attach.conf
https://github.com/OddBloke/ubuntu-advantage-client/pull/2

blackboxsw:
We merged in a separate PR that does the following per @basak's review:

* bump package version to 20.2 so we can easily handle obsoleting conf files and breaking package dependency changes in the split
* Add the Depends/Breaks/Replaces into ubuntu-advantage-pro debian/control
* add rm_conffile /etc/init/ua-auto-attach.conf to debian/ubuntu-advantage-tools.maintscript
* update debian/changelog for functional changes to ua-tools


This packaging change has a known gap that we accept for existing Trusty AWS Ubuntu Pro vms:  
* Upgrading from ua-tools 20.1 -> 20.2 will remove the /etc/init/auto-attach conf file from the system so the system will not be able to auto-attach across reboot within manually installing `ubuntu-advantage-pro` deb.

This gap is acceptable for these reasons:
 * all AWS Ubuntu Pro Vms should have already attached so auto-attach is a noop across reboots
 * any AWS Ubuntu Pro that hadn't already attached is entitled to support and we can guide them through the manual `apt-get install ubuntu-advantage-pro` if required

debian/control
blackboxsw: I think we resolved this conversation/question with @basak's review and the subsequent changeset we just pushed.

debian/changelog
```diff
@@ -1,4 +1,23 @@
-ubuntu-advantage-tools (20.1) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (20.2) UNRELEASED; urgency=medium
+
+  * d/control: create a seprate ubuntu-advantage-pro package which
```
vorlonofportland: separate

debian/ubuntu-advantage-tools.maintscript
```diff
@@ -2,3 +2,4 @@ rm_conffile /etc/update-motd.d/99-esm 19.1~ ubuntu-advantage-tools
 rm_conffile /etc/update-motd.d/80-esm 19.1~ ubuntu-advantage-tools
 rm_conffile /etc/update-motd.d/80-livepatch 19.1~ ubuntu-advantage-tools
 rm_conffile /etc/cron.daily/ubuntu-advantage-tools 19.1~ ubuntu-advantage-tools
+rm_conffile /etc/init/ua-auto-attach.conf 20.2~ ubuntu-advantage-tools
```
vorlonofportland: This is correct for focal, but what about for trusty?  what is the version number that will be uploaded to trusty that doesn't also match this version?  (I ask, because debian/rules includes per-series handling, but there is no per-series handling in the .maintscript file.)

blackboxsw: @vorlonofportland, thanks for the review here. For the only public Pro images on trusty, the package version on trusty is 19.7-0~201911281740~ubuntu14.04.1.

There will be a second release of pro images that will contain 20.1* for Trusty Xenial and Bionic. We need to support migration from either 19.7 or 20.1 major.minor versions as viable upgrade paths and drop the init script or systemd units appropriately across that upgrade

blackboxsw:
I've addressed all review comments and uploaded trusty, xenial bionic and focal to my test PPA https://launchpad.net/~chad.smith/+archive/ubuntu/ua-client-premium-demo

# to test
1. launch an AWS ubuntu-pro instance 14.04 16.04 or 18.04
https://aws.amazon.com/marketplace/search/results?x=0&y=0&searchTerms=ubuntu+pro

2. ssh into that instance and add the following PPA:
sudo add-apt-repository ppa:chad.smith/ua-client-premium-demo

3. Upgrade ubuntu-advantage-tools
sudo apt-get install ubuntu-advantage-tools

vorlonofportland:
> I've addressed all review comments and uploaded trusty, xenial bionic and focal to my test PPA https://launchpad.net/~chad.smith/+archive/ubuntu/ua-client-premium-demo

$dpkg --compare-versions 20.2 gt 20.2-0ubuntu1~14.04.1 && echo good || echo bad
bad
$

focal version number is lower than the bionic one?

blackboxsw:
> > I've addressed all review comments and uploaded trusty, xenial bionic and focal to my test PPA https://launchpad.net/~chad.smith/+archive/ubuntu/ua-client-premium-demo
> 
> $dpkg --compare-versions 20.2 gt 20.2-0ubuntu1~14.04.1 && echo good || echo bad
> bad
> $
> 
> focal version number is lower than the bionic one?

Excellent catch @vorlonofportland thanks on the double-check there that package versioning. It was an oversight I should have corrected when I manually updated the changelog quickly to push up to my PPA for testing. But, it did shed light on an inconsistency when I checked what versions we have published to the ppa https://code.launchpad.net/~canonical-server/+archive/ubuntu/ua-client-premium which is *in* existing Pro images. 
```
$ dpkg --compare-versions  20.1~695~git.dad1cb5~ubuntu16.04.1  gt   20.1-0~202002061703~ubuntu14.04.1   && echo good || echo bad
bad
```
I think we need to fix our version string scheme for our trusty recipe as the hyphen - instead of tilde ~ has broken that pkg version comparison. Thanks again!
Also I'll fix my PPA so I'm representing the proper version string for focal as  something more like
 ubuntu-advantage-tools - 20.2~722~git.0ec0628~ubuntu20.04.1

PR #984, opened 2020-02-07
cli: log sys.argv (once logging is configured)

### LINKED ISSUE #945: Log the command being run
When examining the log, it would be very helpful to (a) know what command was being run, and (b) which log messages relate to which command.

OddBloke:
Fixes #945

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -610,6 +610,7 @@ def main(sys_argv=None):
     log_level = cfg.log_level
     console_level = logging.DEBUG if args.debug else logging.INFO
     setup_logging(console_level, log_level, cfg.log_file)
+    logging.debug("Executed with sys.argv: %r", sys_argv)
     return args.action(args, cfg)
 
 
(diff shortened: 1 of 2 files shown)


uaclient/cli.py
```diff
@@ -610,6 +610,7 @@ def main(sys_argv=None):
     log_level = cfg.log_level
     console_level = logging.DEBUG if args.debug else logging.INFO
     setup_logging(console_level, log_level, cfg.log_file)
+    logging.info("Executed with sys.argv: %r", sys_argv)
```
blackboxsw: This should be a logging.debug right? otherwise we emit to the console anything logging.info. And logging.info generally should be reserved for information that is important for the general user to see.


Console output confirms this a bit noisy:
```
root@dev-t:~/ubuntu-advantage-client# ua status
Executed with sys.argv: ['/usr/bin/ua', 'status']
SERVICE       AVAILABLE  DESCRIPTION
cc-eal        no         Common Criteria EAL2 Provisioning Packages
esm-apps      no         UA Apps: Extended Security Maintenance
esm-infra     yes        UA Infra: Extended Security Maintenance
fips          no         NIST-certified FIPS modules
fips-updates  no         Uncertified security updates to FIPS modules
livepatch     yes        Canonical Livepatch service

See https://ubuntu.com/advantage

OddBloke: Ah, right, I'd forgotten about that "feature" of our logging. Will fix!

PR #985, opened 2020-02-09
Cleanup/drop machine access

blackboxsw:
First part is to read and process service access information directly from machine-token instead of specific machine-access file.

Branch items:
 *  UAConfig.entitlements understands how to read resoureTokens list from machine-token
 * contract.request_updated_contract now pulls delta information from cfg.entitlements object
 * UAContractClient.request_machine_token_update clears cfg._entitlements cache

 The second branch drop UAContractClient.request_resource_machine_access and any callsites.

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -116,9 +116,7 @@ class UAConfig:
         self._entitlements = {}
         contractInfo = machine_token["machineTokenInfo"]["contractInfo"]
         tokens_by_name = dict(
-            # Handle Upper-case and lower-case keys until the following issue
-            # is fixed: ua-contracts #645
-            (e.get("Type", e.get("type")), e.get("Token", e.get("token")))
+            (e["type"], e["token"])
             for e in machine_token.get("resourceTokens", [])
         )
         ent_by_name = dict(
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -116,7 +116,9 @@ class UAConfig:
         self._entitlements = {}
         contractInfo = machine_token["machineTokenInfo"]["contractInfo"]
         tokens_by_name = dict(
-            (e["Type"], e["Token"])
+            # Handle Upper-case and lower-case keys until the following issue
+            # is fixed: ua-contracts #645
+            (e.get("Type", e.get("type")), e.get("Token", e.get("token")))
             for e in machine_token.get("resourceTokens", [])
         )
         ent_by_name = dict(
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -115,19 +115,19 @@ class UAConfig:
 
         self._entitlements = {}
         contractInfo = machine_token["machineTokenInfo"]["contractInfo"]
-
+        tokens_by_name = dict(
+            (e["Type"], e["Token"])
+            for e in machine_token.get("resourceTokens", [])
+        )
         ent_by_name = dict(
             (e["type"], e) for e in contractInfo["resourceEntitlements"]
         )
         for entitlement_name, ent_value in ent_by_name.items():
-            entitlement_cfg = {}
-            if ent_value.get("entitled"):
-                entitlement_cfg = self.read_cache(
-                    "machine-access-{}".format(entitlement_name), silent=True
-                )
-            if not entitlement_cfg:
-                # Fallback to machine-token info on unentitled
-                entitlement_cfg = {"entitlement": ent_value}
+            entitlement_cfg = {"entitlement": ent_value}
+            if entitlement_name in tokens_by_name:
+                entitlement_cfg["resourceToken"] = tokens_by_name[
+                    entitlement_name
+                ]
             util.apply_series_overrides(entitlement_cfg)
             self._entitlements[entitlement_name] = entitlement_cfg
         return self._entitlements
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -174,8 +174,6 @@ class UAContractClient(serviceclient.UAServiceClient):
         response, headers = self.request_url(url, headers=headers, data=data)
         if headers.get("expires"):
             response["expires"] = headers["expires"]
-        # Clear UAConfig._entitlements cache because we are refreshing
-        self.cfg.delete_cache_key("machine-token")
         self.cfg.write_cache("machine-token", response)
         return response
 
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -174,6 +174,8 @@ class UAContractClient(serviceclient.UAServiceClient):
         response, headers = self.request_url(url, headers=headers, data=data)
         if headers.get("expires"):
             response["expires"] = headers["expires"]
+        # Clear UAConfig._entitlements cache because we are refreshing
+        self.cfg.delete_cache_key("machine-token")
         self.cfg.write_cache("machine-token", response)
         return response
 
@@ -281,18 +283,11 @@ def request_updated_contract(
             )
     delta_error = False
     unexpected_error = False
-    for name, entitlement in sorted(cfg.entitlements.items()):
-        if entitlement["entitlement"].get("entitled"):
-            # Obtain each entitlement's accessContext for this machine
-            new_access = contract_client.request_resource_machine_access(
-                new_token["machineToken"], name
-            )
-        else:
-            new_access = entitlement
+    for name, new_entitlement in sorted(cfg.entitlements.items()):
         try:
             process_entitlement_delta(
                 orig_entitlements.get(name, {}),
-                new_access,
+                new_entitlement,
                 allow_enable=allow_enable,
             )
         except exceptions.UserFacingError:
@@ -300,14 +295,14 @@ def request_updated_contract(
             with util.disable_log_to_console():
                 logging.exception(
                     "Failed to process contract delta for {name}:"
-                    " {delta}".format(name=name, delta=new_access)
+                    " {delta}".format(name=name, delta=new_entitlement)
                 )
         except Exception:
             unexpected_error = True
             with util.disable_log_to_console():
                 logging.exception(
                     "Unexpected error processing contract delta for {name}:"
-                    " {delta}".format(name=name, delta=new_access)
+                    " {delta}".format(name=name, delta=new_entitlement)
                 )
     if unexpected_error:
         raise exceptions.UserFacingError(
--- a/uaclient/entitlements/tests/conftest.py
+++ b/uaclient/entitlements/tests/conftest.py
@@ -3,29 +3,34 @@ import pytest
 from uaclient import config
 
 try:
-    from typing import Any, Dict, List, Optional  # noqa
+    from typing import Any, Dict, List  # noqa
 except ImportError:
     # typing isn't available on trusty, so ignore its absence
     pass
 
 
 def machine_token(
-    type: str,
+    entitlement_type: str,
     *,
     affordances: "Dict[str, Any]" = None,
     directives: "Dict[str, Any]" = None,
-    entitled: "Optional[bool]" = True,
+    entitled: bool = True,
     obligations: "Dict[str, Any]" = None,
     suites: "List[str]" = None
 ) -> "Dict[str, Any]":
     return {
-        "resourceTokens": [{"Type": type, "Token": "%s-token" % type}],
+        "resourceTokens": [
+            {
+                "Type": entitlement_type,
+                "Token": "{}-token".format(entitlement_type),
+            }
+        ],
         "machineToken": "blah",
         "machineTokenInfo": {
             "contractInfo": {
                 "resourceEntitlements": [
                     machine_access(
-                        type,
+                        entitlement_type,
                         affordances=affordances,
                         directives=directives,
                         entitled=entitled,
@@ -39,11 +44,11 @@ def machine_token(
 
 
 def machine_access(
-    type: str,
+    entitlement_type: str,
     *,
     affordances: "Dict[str, Any]" = None,
     directives: "Dict[str, Any]" = None,
-    entitled: "Optional[bool]" = True,
+    entitled: bool = True,
     obligations: "Dict[str, Any]" = None,
     suites: "List[str]" = None
 ) -> "Dict[str, Any]":
@@ -55,14 +60,14 @@ def machine_access(
         obligations = {"enableByDefault": True}
     if directives is None:
         directives = {
-            "aptURL": "http://{}".format(type.upper()),
+            "aptURL": "http://{}".format(entitlement_type.upper()),
             "aptKey": "APTKEY",
             "suites": suites,
         }
     return {
         "obligations": obligations,
-        "type": type,
-        "entitled": True,
+        "type": entitlement_type,
+        "entitled": entitled,
         "directives": directives,
         "affordances": affordances,
     }
--- a/uaclient/entitlements/tests/conftest.py
+++ b/uaclient/entitlements/tests/conftest.py
@@ -21,8 +21,8 @@ def machine_token(
     return {
         "resourceTokens": [
             {
-                "Type": entitlement_type,
-                "Token": "{}-token".format(entitlement_type),
+                "type": entitlement_type,
+                "token": "{}-token".format(entitlement_type),
             }
         ],
         "machineToken": "blah",
--- a/uaclient/entitlements/tests/conftest.py
+++ b/uaclient/entitlements/tests/conftest.py
@@ -84,7 +84,14 @@ def entitlement_factory(tmpdir):
     entitlement.
     """
 
-    def factory_func(cls, *, affordances=None, directives=None, suites=None):
+    def factory_func(
+        cls,
+        *,
+        affordances: "Dict[str, Any]" = None,
+        directives: "Dict[str, Any]" = None,
+        entitled: bool = True,
+        suites: "List[str]" = None
+    ):
         cfg = config.UAConfig(cfg={"data_dir": tmpdir.strpath})
         cfg.write_cache(
             "machine-token",
@@ -92,6 +99,7 @@ def entitlement_factory(tmpdir):
                 cls.name,
                 affordances=affordances,
                 directives=directives,
+                entitled=entitled,
                 suites=suites,
             ),
         )
--- a/uaclient/entitlements/tests/conftest.py
+++ b/uaclient/entitlements/tests/conftest.py
@@ -3,19 +3,35 @@ import pytest
 from uaclient import config
 
 try:
-    from typing import Any, Dict, List  # noqa
+    from typing import Any, Dict, List, Optional  # noqa
 except ImportError:
     # typing isn't available on trusty, so ignore its absence
     pass
 
 
-def machine_token(entitlement_type: str) -> "Dict[str, Any]":
+def machine_token(
+    type: str,
+    *,
+    affordances: "Dict[str, Any]" = None,
+    directives: "Dict[str, Any]" = None,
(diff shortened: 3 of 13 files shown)


uaclient/entitlements/tests/conftest.py
```diff
@@ -3,51 +3,68 @@
...
     }
 
 
 def machine_access(
-    entitlement_type: str,
+    type: str,
     *,
     affordances: "Dict[str, Any]" = None,
     directives: "Dict[str, Any]" = None,
+    entitled: "Optional[bool]" = True,
```
OddBloke: I don't think this is `Optional`; it should only ever be `True` or `False`, so it's a straight-up `bool`. I just confirmed this for myself with a toy example:

```py
from typing import Optional


def optional_foo(x: Optional[bool] = True) -> None:
    pass


optional_foo(True)
optional_foo(False)
optional_foo()
optional_foo(None)


def foo(x: bool = True) -> None:
    pass


foo(True)
foo(False)
foo()
foo(None)
```

gives:

```
$ mypy foo.py
foo.py:21: error: Argument 1 to "foo" has incompatible type "None"; expected "bool"
```

uaclient/entitlements/tests/conftest.py
```diff
@@ -3,51 +3,68 @@
...
-        "entitlement": {
-            "obligations": {"enableByDefault": True},
-            "type": entitlement_type,
-            "entitled": True,
-            "directives": directives,
-            "affordances": affordances,
-        },
+        "obligations": obligations,
+        "type": type,
+        "entitled": True,
```
OddBloke: Should this be using the value of `entitled`?

uaclient/entitlements/tests/test_cc.py
```diff
@@ -168,7 +160,7 @@ def exists(path):
             mock.call(
                 "/etc/apt/sources.list.d/ubuntu-cc-eal-xenial.list",
                 "http://CC",
-                "TOKEN",
+                "%s-token" % entitlement.name,
```
OddBloke: A nit, but `.format` here would be more consistent with the rest of the codebase.

uaclient/entitlements/tests/test_cis.py
```diff
@@ -63,7 +63,7 @@ def fake_platform(key=None):
             mock.call(
                 "/etc/apt/sources.list.d/ubuntu-cis-audit-xenial.list",
                 "http://CIS-AUDIT",
-                "TOKEN",
+                "%s-token" % entitlement.name,
```
OddBloke: A nit, but `.format` here would be more consistent with the rest of the codebase.

uaclient/entitlements/tests/test_esm.py
```diff
@@ -73,7 +73,7 @@ def fake_exists(path):
                     entitlement.name
                 ),
                 "http://{}".format(entitlement.name.upper()),
-                "TOKEN",
+                "%s-token" % entitlement.name,
```
OddBloke: A nit, but `.format` here would be more consistent with the rest of the codebase.

uaclient/entitlements/tests/test_esm.py
```diff
@@ -157,7 +157,7 @@ def fake_subp(cmd, capture=None, retry_sleeps=None):
...
```
OddBloke: A nit, but `.format` here would be more consistent with the rest of the codebase.

uaclient/entitlements/tests/test_fips.py
```diff
@@ -88,7 +87,7 @@ def test_enable_configures_apt_sources_and_auth_files(self, entitlement):
                     entitlement.name
                 ),
                 repo_url,
-                "TOKEN",
+                "%s-token" % entitlement.name,
```
OddBloke: A nit, but `.format` here would be more consistent with the rest of the codebase.

uaclient/entitlements/tests/conftest.py
```diff
@@ -3,51 +3,68 @@
...
+    type: str,
+    *,
+    affordances: "Dict[str, Any]" = None,
+    directives: "Dict[str, Any]" = None,
+    entitled: "Optional[bool]" = True,
+    obligations: "Dict[str, Any]" = None,
+    suites: "List[str]" = None
+) -> "Dict[str, Any]":
     return {
+        "resourceTokens": [{"Type": type, "Token": "%s-token" % type}],
```
OddBloke: A nit, but `.format` here would be more consistent with the rest of the codebase.

uaclient/contract.py
```diff
@@ -280,18 +282,11 @@ def request_updated_contract(
                 status.MESSAGE_CONTRACT_EXPIRED_ERROR
             )
     user_errors = []
-    for name, entitlement in sorted(cfg.entitlements.items()):
-        if entitlement["entitlement"].get("entitled"):
-            # Obtain each entitlement's accessContext for this machine
-            new_access = contract_client.request_resource_machine_access(
```
OddBloke: Have we confirmed with the contract server team that they are prepared for any consequences of us not hitting this endpoint any longer?

uaclient/config.py
```diff
@@ -115,19 +115,19 @@ def entitlements(self):
 
         self._entitlements = {}
         contractInfo = machine_token["machineTokenInfo"]["contractInfo"]
-
+        tokens_by_name = dict(
+            (e["Type"], e["Token"])
```
OddBloke: So this part of the response has the `Type` key capitalised, but the one we use in the next logical line has it lower-case? Is consistency something we could ask of ua-contracts, or has the ship sailed on changing the shape of the API response?

uaclient/entitlements/tests/conftest.py
```diff
@@ -3,51 +3,68 @@
...
     pass
 
 
-def machine_token(entitlement_type: str) -> "Dict[str, Any]":
+def machine_token(
+    type: str,
+    *,
+    affordances: "Dict[str, Any]" = None,
+    directives: "Dict[str, Any]" = None,
+    entitled: "Optional[bool]" = True,
```
OddBloke: As I mention in more detail below (I missed this definition on my first read through), this isn't `Optional`.

uaclient/contract.py
```diff
@@ -174,6 +174,8 @@ def request_machine_token_update(
         response, headers = self.request_url(url, headers=headers, data=data)
         if headers.get("expires"):
             response["expires"] = headers["expires"]
+        # Clear UAConfig._entitlements cache because we are refreshing
+        self.cfg.delete_cache_key("machine-token")
```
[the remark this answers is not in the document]
blackboxsw: ahh right, and turns out it already does, this was a shortcoming in our FakeConfig which doesn't delete that cache key during a FakeConfig.write_cache call.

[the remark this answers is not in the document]
blackboxsw: Just refactored FakeConfig subclass to require a data_dir param (presumably tmpdir.strpath) and updated all unit tests. Our FakeConfig was growing too much parallel logic with existing UAConfig and that was causing the unit test failure which prompted me to incorrectly add the "delete_cache_key" call.

uaclient/contract.py
blackboxsw: We had in IRC with mhilton and cmars before starting this work. And their consensus was that the can (and should be) determinig KPI attach metrics from their backend service endpoints (like livepatch of various PPA paths during apt-update calls).  We may have to restate that we have completed this effort again to confirm that they aren't relying on information from ua client attaching via those routes.

uaclient/config.py
blackboxsw: I think it's worth a bug against ua-contracts on that, since we are the only consumer of that data. We can make this branch work with the lower() value in either case.

blackboxsw: Filed https://github.com/CanonicalLtd/ua-contracts/issues/645

uaclient/entitlements/tests/conftest.py
blackboxsw: dropped to bool

uaclient/contract.py
blackboxsw: pinged in #ua IRC again about this cmars will chat with the team tomorrow.

uaclient/entitlements/tests/conftest.py
blackboxsw: yes, nice one

uaclient/entitlements/tests/conftest.py
blackboxsw: oopsie. got excited about the Optional. Dropped

uaclient/entitlements/tests/conftest.py
```diff
@@ -3,51 +3,68 @@
...
+                    )
                 ]
             }
         },
     }
 
 
 def machine_access(
-    entitlement_type: str,
+    type: str,
```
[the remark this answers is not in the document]
blackboxsw: done

uaclient/contract.py
blackboxsw: confirmed in IRC today from alnvdl and cmars that this is not an issue.

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -51,9 +51,15 @@ def test_contract_status_entitled(self, entitlement):
 
     def test_contract_status_unentitled(self, entitlement):
         """The contract_status returns NONE when entitled is False."""
-        entitlement.cfg.write_cache(
-            "machine-access-livepatch", {"entitlement": {"entitled": False}}
+        token = entitlement.cfg.read_cache("machine-token")
```
OddBloke: I haven't fully thought this through, so please take it with a grain of salt, but might these test changes be simpler if you used `entitlement_factory` directly, rather than using the `entitlement` fixture and then modifying it after the fact? (Looking at the `entitlement` fixture, it might make sense to introduce a `livepatch_entitlement_factory` fixture, which sets affordances/directives in the same way as `entitlement`.)

Something like (completely untested):

```py
@pytest.fixture
def livepatch_entitlement_factory(entitlement_factory):

    def factory_func(entitled: bool = True, suites: List[str] = None):
        affordances = {
            "architectures": ["x86_64"],
            "minKernelVersion": "4.4",
            "kernelFlavors": ["generic", "lowlatency"],
            "tier": "stable",
        }
        directives = {"caCerts": "", "remoteServer": "https://alt.livepatch.com"}
        return entitlement_factory(
            LivepatchEntitlement, affordances=affordances, directives=directives,
            entitled=entitled, suites=suites,
        )

    return factory_func

@pytest.fixture
def entitlement(livepatch_entitlement_factory):
    return livepatch_entitlement_factory()
```

and then this test would be something like:

```py
    def test_contract_status_unentitled(self, livepatch_entitlement_factory):
        entitlement = livepatch_entitlement_factory(entitled=False)
        assert ContractStatus.UNENTITLED == entitlement.contract_status()
```

(This would require adding an `entitled` argument to `entitlement_factory` too, and passing it through to `machine_token` in `conftest.py`.)

blackboxsw: > I haven't fully thought this through, so please take it with a grain of salt, but might these test changes be simpler if you used `entitlement_factory` directly, rather than using the `entitlement` fixture and then modifying it after the fact? (Looking at the `entitlement` fixture, it might make sense to introduce a `livepatch_entitlement_factory` fixture, which sets affordances/directives in the same way as `entitlement`.)
> 
> Something like (completely untested):
> 
> ```python
> @pytest.fixture
> def livepatch_entitlement_factory(entitlement_factory):
> 
>     def factory_func(entitled: bool = True, suites: List[str] = None):
>         affordances = {
>             "architectures": ["x86_64"],
>             "minKernelVersion": "4.4",
>             "kernelFlavors": ["generic", "lowlatency"],
>             "tier": "stable",
>         }
>         directives = {"caCerts": "", "remoteServer": "https://alt.livepatch.com"}
>         return entitlement_factory(
>             LivepatchEntitlement, affordances=affordances, directives=directives,
>             entitled=entitled, suites=suites,
>         )
> 
>     return factory_func
> 
> @pytest.fixture
> def entitlement(livepatch_entitlement_factory):
>     return livepatch_entitlement_factory()
> ```
> 
> and then this test would be something like:
> 
> ```python
>     def test_contract_status_unentitled(self, livepatch_entitlement_factory):
>         entitlement = livepatch_entitlement_factory(entitled=False)
>         assert ContractStatus.UNENTITLED == entitlement.contract_status()
> ```
> 
> (This would require adding an `entitled` argument to `entitlement_factory` too, and passing it through to `machine_token` in `conftest.py`.)

I like this suggestion a lot, that whole logic I added in testing was messy and felt dirty. Thanks for the thoughts on this will implement tomorrow.

uaclient/config.py
```diff
@@ -115,19 +115,21 @@ def entitlements(self):
 
         self._entitlements = {}
         contractInfo = machine_token["machineTokenInfo"]["contractInfo"]
-
+        tokens_by_name = dict(
+            # Handle Upper-case and lower-case keys until the following issue
+            # is fixed: ua-contracts #645
+            (e.get("Type", e.get("type")), e.get("Token", e.get("token")))
```
OddBloke: I don't want to land this branch with this logic in place. We should work out what the contract server should be doing, get them to make the changes, and write against that.

blackboxsw:
@OddBloke addressed all comments. Also validated that contracts.staging has been fixes to use lowercase `type` and `token` keys in `resourceTokens` in the machine-token.json. Still can't land/release this until production has this fix so I'm leaving the external blocker label on it until I can confirm a production fix.

uaclient/tests/test_contract.py
```diff
@@ -384,26 +378,20 @@ def test_attached_config_refresh_machine_token_and_services(
                 }
             },
         }
+        new_token = copy.deepcopy(machine_token)
+        new_token["machineTokenInfo"]["contractInfo"]["resourceEntitlements"][
+            1
```
OddBloke: Haha, thanks black!

uaclient/entitlements/tests/test_fips.py
```diff
@@ -143,14 +142,15 @@ def test_enable_returns_false_on_missing_suites_directive(
...
-            fips_entitled_no_suites,
+        token = entitlement.cfg.read_cache("machine-token")
+        entitlement_cfg = entitlement.cfg.entitlements[entitlement.name].get(
+            "entitlement"
         )
+        entitlement_cfg["directives"]["suites"] = []
+        token["machineTokenInfo"]["contractInfo"]["resourceEntitlements"] = [
+            entitlement_cfg
+        ]
+        entitlement.cfg.write_cache("machine-token", token)
```
OddBloke: Similar to the other test we discussed, would this be simpler by using an entitlement factory directly? I think so:

```diff
diff --git a/uaclient/entitlements/tests/test_fips.py b/uaclient/entitlements/tests/test_fips.py
index a44672e..6f699d5 100644
--- a/uaclient/entitlements/tests/test_fips.py
+++ b/uaclient/entitlements/tests/test_fips.py
@@ -4,6 +4,7 @@ import contextlib
 import copy
 import itertools
 import mock
+from functools import partial
 
 import pytest
 
@@ -24,9 +25,14 @@ M_GETPLATFORM = M_REPOPATH + "util.get_platform_info"
 
 
 @pytest.fixture(params=[FIPSEntitlement, FIPSUpdatesEntitlement])
-def entitlement(request, entitlement_factory):
+def fips_entitlement_factory(request, entitlement_factory):
     """Parameterized fixture so we apply all tests to both FIPS and Updates"""
-    return entitlement_factory(request.param)
+    return partial(entitlement_factory, request.param)
+
+
+@pytest.fixture
+def entitlement(fips_entitlement_factory):
+    return fips_entitlement_factory()
 
 
 class TestFIPSEntitlementCanEnable:
@@ -138,19 +144,10 @@ class TestFIPSEntitlementEnable:
         "uaclient.util.get_platform_info", return_value={"series": "xenial"}
     )
     def test_enable_returns_false_on_missing_suites_directive(
-        self, m_platform_info, m_add_apt, entitlement
+        self, m_platform_info, m_add_apt, fips_entitlement_factory
     ):
         """When directives do not contain suites returns false."""
-        # Unset suites directive
-        token = entitlement.cfg.read_cache("machine-token")
-        entitlement_cfg = entitlement.cfg.entitlements[entitlement.name].get(
-            "entitlement"
-        )
-        entitlement_cfg["directives"]["suites"] = []
-        token["machineTokenInfo"]["contractInfo"]["resourceEntitlements"] = [
-            entitlement_cfg
-        ]
-        entitlement.cfg.write_cache("machine-token", token)
+        entitlement = fips_entitlement_factory(suites=[])
 
         with mock.patch.object(entitlement, "can_enable", return_value=True):
             with pytest.raises(exceptions.UserFacingError) as excinfo:
```

OddBloke: (Oh, FYI, this patch won't apply cleanly without first applying the one above.)

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -29,31 +36,51 @@
...
-        "architectures": ["x86_64"],
-        "minKernelVersion": "4.4",
-        "kernelFlavors": ["generic", "lowlatency"],
-        "tier": "stable",
-    }
-    directives = {"caCerts": "", "remoteServer": "https://alt.livepatch.com"}
-    return entitlement_factory(
-        LivepatchEntitlement, affordances=affordances, directives=directives
-    )
+def livepatch_entitlement_factory(entitlement_factory):
```
OddBloke: When writing the fips equivalent to this, I realised we could make this simpler with `functools.partial` too:

```diff
diff --git a/uaclient/entitlements/tests/test_livepatch.py b/uaclient/entitlements/tests/test_livepatch.py
index 0e36780..26872c7 100644
--- a/uaclient/entitlements/tests/test_livepatch.py
+++ b/uaclient/entitlements/tests/test_livepatch.py
@@ -3,6 +3,7 @@
 import copy
 import logging
 import mock
+from functools import partial
 from types import MappingProxyType
 
 try:
@@ -46,26 +47,13 @@ DEFAULT_AFFORDANCES = {
 
 @pytest.fixture
 def livepatch_entitlement_factory(entitlement_factory):
-    def factory_func(
-        entitled: bool = True,
-        affordances: "Dict[str, Any]" = None,
-        suites: "List[str]" = None,
-    ):
-        if not affordances:
-            affordances = DEFAULT_AFFORDANCES
-        directives = {
-            "caCerts": "",
-            "remoteServer": "https://alt.livepatch.com",
-        }
-        return entitlement_factory(
-            LivepatchEntitlement,
-            affordances=affordances,
-            directives=directives,
-            entitled=entitled,
-            suites=suites,
-        )
-
-    return factory_func
+    directives = {"caCerts": "", "remoteServer": "https://alt.livepatch.com"}
+    return partial(
+        entitlement_factory,
+        LivepatchEntitlement,
+        affordances=DEFAULT_AFFORDANCES,
+        directives=directives,
+    )
 
 
 @pytest.fixture
```

blackboxsw:
Thanks dan for the patches, I've applied all 3.

uaclient/tests/test_contract.py
blackboxsw: yes glad I don't have to worry about defending automated formatting. as this would have been a couple of cycles. I guess I could better construct locals to avoid that nastiness. But it's more fun to don't care

OddBloke: Oh yeah, this is fine, it's totally worth the occasional dumb thing like this to not have to think about it at all the rest of the time.

Issue #953, opened 2019-12-02
Failure to enable an entitlement potentially results in other entitlements not being enabled


OddBloke:
Related to #952, but a distinct issue (that can't be solved by dropping non-root status).

This doesn't always happen when a given entitlement fails. This probably isn't an indicator that the issue isn't present every time, but that we're iterating over the entitlements in an unstable order (e.g. Python dict keys).

blackboxsw:
This case where other entitlements aren't enabled is directly related to special general Exception handling (instead of potentially expected UserFacingErrors on line 297 of the same file
https://github.com/CanonicalLtd/ubuntu-advantage-client/blob/master/uaclient/contract.py#L299L304

I'm not sure what we should do with generally unexpected Exceptions and whether we should queue them up or fail hard sooner as we are currently doing.

blackboxsw:
@OddBloke does this approach sound good? Queue the unexpected error like we do UserFacingErrors and only raise at the end of the for loop?

OddBloke:
I do think we should queue up the unexpected errors until the end, so that we attempt enable on all the entitlements. _But_, I also think we want to indicate to the user that we would appreciate a bug report in these cases.

We may want to do some UX thinking before we dive into implementation, though, because this will affect the output of `attach`.

powersj:
Here is my attempt to throw something at the wall:

```
$ sudo ua attach <token>
<bad stuff happens>
<status prints out>
An unexpected error(s) occurred during attach
For more details, see the log: /var/log/...
```

Questions:
1) What kind of unexpected errors would we be talking about?
2) What is the log file they should look at?
3) Where do they file a bug? (I'm working on a decision for this one and whether we even put that in the output)

blackboxsw:
>     1. What kind of unexpected errors would we be talking about?
 Unexpected errors might be:
      * ua-contracts server being down for maintenance resulting in timeout
      * ua-contracts server returning 503 due to not handling load while ua client iterates over each entitlement performing POSTs for each entitlement
      * unhandled logic or unseen gaps in uaclient in process_entitlement_deltas for a given entitlement type, such as unexpected values or new structure of machine-access responses from the contract server that break us.

2. /var/log/ubuntu-advantage.log for Tracebacks

3. Maybe they can file a bug at 
https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+filebug

powersj:
[...]

PR #986, opened 2020-02-10
contract: handle unexpected errors when processing contract deltas

blackboxsw:
Continue to process all entitlements if hitting unexpected errors
during service enablement during attach or refresh. Previously,
if UA client hit APT failures during enablement of repo-based services
the UA client would exit with a traceback and avoid enabling any remaining
services.

Now, log any unexpected failures and only raise it once all service deltas
have been processed. Unexpected errors will now alert the user with a
message to check /var/log/ubuntu-advantage.log and file a bug.

Fixes: #953

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -294,13 +294,19 @@ def request_updated_contract(
                 new_access,
                 allow_enable=allow_enable,
             )
-        except exceptions.UserFacingError as e:
+        except exceptions.UserFacingError:
             with util.disable_log_to_console():
-                logging.exception(str(e))
+                logging.exception(
+                    "Failed to process contract delta for {name}:"
+                    " {delta}".format(name=name, delta=new_access)
+                )
             user_errors.append(status.MESSAGE_ATTACH_FAILURE_DEFAULT_SERVICES)
-        except Exception as e:
+        except Exception:
             with util.disable_log_to_console():
-                logging.exception(str(e))
+                logging.exception(
+                    "Unexpected error processing contract delta for {name}:"
+                    " {delta}".format(name=name, delta=new_access)
+                )
             user_errors.insert(  # Highest priority error comes first
                 0,
                 status.MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL.format(
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -295,17 +295,19 @@ def request_updated_contract(
                 allow_enable=allow_enable,
             )
         except exceptions.UserFacingError as e:
-            user_errors.append(e)
+            with util.disable_log_to_console():
+                logging.exception(str(e))
+            user_errors.append(status.MESSAGE_ATTACH_FAILURE_DEFAULT_SERVICES)
         except Exception as e:
             with util.disable_log_to_console():
                 logging.exception(str(e))
-            raise exceptions.UserFacingError(
-                "Unexpected error handling Ubuntu Advantage contract changes"
+            user_errors.append(
+                status.MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL.format(
+                    operation="attach"
+                )
             )
     if user_errors:
-        raise exceptions.UserFacingError(
-            status.MESSAGE_ATTACH_FAILURE_DEFAULT_SERVICES
-        )
+        raise exceptions.UserFacingError(user_errors[0])
 
 
 def get_available_resources(cfg) -> "List[Dict]":
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -279,8 +279,8 @@ def request_updated_contract(
             raise exceptions.UserFacingError(
                 status.MESSAGE_CONTRACT_EXPIRED_ERROR
             )
-    delta_error_msg = None
-    unexpected_error_msg = None
+    delta_error = False
+    unexpected_error = False
     for name, entitlement in sorted(cfg.entitlements.items()):
         if entitlement["entitlement"].get("entitled"):
             # Obtain each entitlement's accessContext for this machine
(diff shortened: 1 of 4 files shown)


uaclient/contract.py
```diff
@@ -295,17 +295,19 @@ def request_updated_contract(
                 allow_enable=allow_enable,
             )
         except exceptions.UserFacingError as e:
-            user_errors.append(e)
+            with util.disable_log_to_console():
+                logging.exception(str(e))
```
OddBloke: A nit, but `str(e)` isn't very useful here; we should get the exception message as part of the stack trace regardless. Better would be logging some additional information that _isn't_ in the message.

uaclient/contract.py
```diff
@@ -295,17 +295,19 @@ def request_updated_contract(
...
+            user_errors.append(
+                status.MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL.format(
+                    operation="attach"
+                )
             )
     if user_errors:
-        raise exceptions.UserFacingError(
-            status.MESSAGE_ATTACH_FAILURE_DEFAULT_SERVICES
-        )
+        raise exceptions.UserFacingError(user_errors[0])
```
OddBloke: As written, I think this will only display the "unexpected error" message if it is the first error we encounter. Do you think we should instead always emit that message if _any_ of the errors are unexpected?

uaclient/status.py
```diff
@@ -191,6 +193,13 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_INVALID_SERVICE_OP_FAILURE_TMPL = """\
 Cannot {operation} '{name}'
 For a list of services see: sudo ua status"""
+MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL = (
+    """\
+Unexpected error(s) occurred during {operation}
+For more details, see the log: /var/log/ubuntu-advantage.log
+Please file a bug at: """
```
OddBloke: Should we suggest using `ubuntu-bug` instead of pointing directly at the bug URL?

uaclient/tests/test_contract.py
```diff
@@ -260,6 +262,80 @@ def fake_contract_client(cfg):
...
+                    "entitlement": {
+                        "entitled": True,
+                        "type": "ent1",
+                        "new": "newval",
+                    }
+                },
+            }
+            return client
+
+        client.side_effect = fake_contract_client
```
OddBloke: Is it possible to simplify this by using `return_value` instead of `side_effect`?

blackboxsw: I could mock return values for the two UAContractClient.request_* methods that are called for self.refresh_route and self.access_route_ent1  but then I'd also have to mock the cfg.write_cache calls when seed the for loop that iterates over cfg.entitlements.items().  So it adds complexity there because we also need to properly setup the cfg.entitlements values too.

uaclient/contract.py
blackboxsw: I've adjusted this logic to insert MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL message to the first item in the user_errors list. This way we will raise the unexpected error message to the user even if both types of errors have occurred.

blackboxsw:
Think I've addressed your comments with the exception of the side_effect return_value test comment.

uaclient/contract.py
OddBloke: Do we need to put the error in `user_errors` at all? It seems to me that we could just have a bool that tracks if we've seen an "unexpected" error; if we have, we emit the appropriate message, if we haven't, we emit `user_errors[0]`.

(And, actually, in that scenario I think all the elements in `user_errors` must be `status.MESSAGE_ATTACH_FAILURE_DEFAULT_SERVICES`, so perhaps we can replace it entirely with two bools?)

uaclient/tests/test_contract.py
OddBloke: Apologies, my question was not really specific enough at all, I was really asking if this alternative, slightly simpler implementation would work:

```py
        fake_contract_client = FakeContractClient(cfg)
        fake_contract_client._responses = {
            self.refresh_route: machine_token,
            self.access_route_ent1: {
                "entitlement": {
                    "entitled": True,
                    "type": "ent1",
                    "new": "newval",
                }
            },
        }

        client.return_value = fake_contract_client
```

OddBloke:
I'm almost EOD, so I've just replied to your replies to my comments; a full re-review will have to wait for tomorrow.

uaclient/status.py
```diff
@@ -191,6 +191,10 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_INVALID_SERVICE_OP_FAILURE_TMPL = """\
 Cannot {operation} '{name}'
 For a list of services see: sudo ua status"""
+MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL = """\
+Unexpected error(s) occurred during {operation}
+For more details, see the log: /var/log/ubuntu-advantage.log
+To file a bug run: ubuntu-bug ubuntu-advantage-tools"""
```
OddBloke: I know I suggested the change, but did we confirm that this is how we want to do bug reporting? @powersj, what do you think, `ubuntu-bug` or direct link?

powersj: I'm fine with ubuntu-bug

OddBloke:
This is currently failing CI due to a long line that black keeps restoring; I played around locally, and I think this diff will fix it (and also simplify the logic a little further):

```diff
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -279,8 +279,8 @@ def request_updated_contract(
             raise exceptions.UserFacingError(
                 status.MESSAGE_CONTRACT_EXPIRED_ERROR
             )
-    delta_error_msg = None
-    unexpected_error_msg = None
+    delta_error = False
+    unexpected_error = False
     for name, entitlement in sorted(cfg.entitlements.items()):
         if entitlement["entitlement"].get("entitled"):
             # Obtain each entitlement's accessContext for this machine
@@ -296,25 +296,29 @@ def request_updated_contract(
                 allow_enable=allow_enable,
             )
         except exceptions.UserFacingError:
-            delta_error_msg = status.MESSAGE_ATTACH_FAILURE_DEFAULT_SERVICES
+            delta_error = True
             with util.disable_log_to_console():
                 logging.exception(
                     "Failed to process contract delta for {name}:"
                     " {delta}".format(name=name, delta=new_access)
                 )
         except Exception:
-            unexpected_error_msg = status.MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL.format(
-                operation="attach"
-            )
+            unexpected_error = True
             with util.disable_log_to_console():
                 logging.exception(
                     "Unexpected error processing contract delta for {name}:"
                     " {delta}".format(name=name, delta=new_access)
                 )
-    if unexpected_error_msg:
-        raise exceptions.UserFacingError(unexpected_error_msg)
-    elif delta_error_msg:
-        raise exceptions.UserFacingError(delta_error_msg)
+    if unexpected_error:
+        raise exceptions.UserFacingError(
+            status.MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL.format(
+                operation="attach"
+            )
+        )
+    elif delta_error:
+        raise exceptions.UserFacingError(
+            status.MESSAGE_ATTACH_FAILURE_DEFAULT_SERVICES
+        )
 
 
 def get_available_resources(cfg) -> "List[Dict]":

```

If you're happy with that, I can push up a commit or you can, let me know!

PR #990, opened 2020-02-10
add test for behavior of attaching an invalid_token

patriciadomin:
it adds a new feature file when it tries to attach a machine to an UA subscription using an invalid_token

### DIFF
--- a/features/attach_invalidtoken.feature
+++ b/features/attach_invalidtoken.feature
@@ -1,0 +1,15 @@
+Feature: Command behaviour when trying to attach a machine to an Ubuntu
+         Advantage subscription using an invalid token
+
+    Scenario: Attach command in a trusty lxd container
+       Given a trusty lxd container with ubuntu-advantage-tools installed
+        When I run `ua attach INVALID_TOKEN` with sudo
+        Then I will see the following on stderr:
+            """
+            Invalid token. See https://ubuntu.com/advantage
+            """
+        When I run `ua attach INVALID_TOKEN` as non-root
+        Then I will see the following on stderr:
+             """
+             This command must be run as root (try using sudo)
+             """


features/environment.py
```diff
@@ -154,3 +158,10 @@ def _install_uat_in_container(container_name: str) -> None:
         container_name,
         ["sudo", "apt-get", "install", "-qq", "-y", "ubuntu-advantage-tools"],
     )
+
+def def generate_random_token():
```
blackboxsw: Accidental invalid "def def"  method declaration.

Also probably want to move this into either features/util or steps/steps.py as it will be likely used there if at all.

features/steps/steps.py
```diff
@@ -35,6 +28,19 @@ def when_i_run_command(context, command, user_spec):
     )
     context.process = process
 
+@when('I run `{command}` "{token}" {user_spec}')
+def when_i_run_attach_token(context, command, user_spec, token=None):
```
blackboxsw: The error you see when trying to run tox -e behave tells you that this when clause is not unique when compared to 

```



What this means is your text format of of the @when string needs to be more unique in format, 
that's why I introduced a more unique decorator format @when("I attach with token `{token_name}` {user_spec}")
This pattern will not possibly match the @when function above

Full diff here with a few points:  https://paste.ubuntu.com/p/NfRhmKWVdr/

File: features/environment.py (hunk 1 above)
Comment (patriciadomin) [reply features/environment.py]: many thanks.
if we are going to use the UAClientBehaveConfig class, so we need to provide the token via `UACLIENT_BEHAVE_contract_token=...`, (same way for image_clean and image_reuse). I've changed it. I came with an idea for getting the `invalid_token` always from `generate_random_token` and getting it from CLI when we do have a valid one (for testing the attach valid_token), but we can change it, if something else is better. If we keep this way. I think generate_random_token could stay in `environments.py`

Comment (patriciadomin) [reply features/environment.py]: https://paste.ubuntu.com/p/NdYKD5CJJY/

File: features/steps/steps.py (hunk 2 above)
Comment (patriciadomin) [reply features/steps/steps.py]: Ok I've changed it, but I kept the `token` info from  `UAClientBehaveConfig` class. Please, See if you think it is ok or not.

File: features/environment.py (hunk 1 above)
Comment (blackboxsw) [reply features/environment.py]: @patriciadomin I'm not sure what you are asking, that pasted error ```
HOOK-ERROR in before_all: TypeError: from_environ() takes 1 positional argument but 2 were given```
Means that you didn't apply my whole patchset when extended from_environ class method to accept a userdata Dictionary parameter.  Specifically your current branch must be erroring because it doesn't have line 78 of the visual diff I pasted at https://paste.ubuntu.com/p/w34fKnDVwK/

Comment (blackboxsw) [reply features/environment.py]: Also I'm not sure you'd need to provide the prefix "UA_CLIENT_BEHAVE_"  because the visual diff of the paste I sent above line 87 and 88 specifically didn't check for that environment variable prefix, but I can understand that may be confusing if we require os.environ keys to all have the prefix and -D keys to *not* have the prefix
```
+        for key, value in userdata.items():  # No cls.prefix for -D keys
+            if key not in cls.all_options:
```

Comment (patriciadomin) [reply features/environment.py]: Thanks. I've pushed it with the modifications you suggested.

File: features/steps/steps.py (hunk 2 above)
Comment (blackboxsw) [reply features/steps/steps.py]: @patriciadomin It's hard for me to know exactly how you are going to use the context.token until you push of suggest the rest of the context around why context.token is needed instead of context.config.token 

Generally, as you see @OddBloke added the UAClientBehaveConfig class to features/environment.py. That class intent is to encapsulate and simplify any reference to external environment configuration in a single place/instance and surface those config options as instance attributes to simplify tests throughout the rest of the behave tests. this way behave steps can use context.config.reuse_image or context.config.image_name etc after the initial UAClientBehaveConfig was set up.  So I think we still need some breadcrumbs of contract config information inside UAClientBehaveConfig so that at least .from_environ() method can print out the config options provided to the behave test run to ease in our debugging failed test runs.



I think it could be okay if you are keeping around context.token for ease of reference.  but I'll have to see the rest of the context around your approach once you git push your next changeset to this branch. 

Another alternative could be that you extend UAClientBehaveConfig to surface a userdata property which would act just like behave's default context.config.userdata, then any steps could reference context.config.userdata as they see fit, just like stock behave tests.

```@property
def userdata(cls):
    return cls.userdata
```

features/attach_invalid_token.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Command behaviour when attach a machine to an Ubuntu Advantage subscription
+
+    Scenario: Attach command in a trusty lxd container
```
OddBloke: This scenario is specifically testing attaching with an invalid token, can we update this wording to be explicit about that?

features/steps/steps.py
```diff
@@ -44,3 +41,14 @@ def then_i_will_see_on_stdout(context):
 @then("I will see the following on stderr")
 def then_i_will_see_on_stderr(context):
     assert_that(context.process.stderr.strip(), equal_to(context.text))
+
+def get_user_spec(user_spec):
```
OddBloke: I think the naming of this method is a little misleading. It _takes_ a `user_spec`, but it doesn't return one. Perhaps `get_command_prefix_for_user_spec` would be more accurate?

features/attach_invalid_token.feature
patriciadomin: My idea for that is:  
- Generating a random token when testing the output if you try an invalid token 
- Providing the token via CLI for the next tests: when you attach to an UA subscription using a valid token.

features/steps/steps.py
patriciadomin: Sure, I'll change it. Thanks

patriciadomin:
Sending it again. Addressed @OddBloke comments.

features/attach_invalid_token.feature
blackboxsw: I think Dan's comment comment leads to the following couple questions:
* can this feature file be renamed attach_token.feature. Each feature file in my mind is related to a single 'CLI operation' such as (attach_token.feature, status.feature, enable.feature, disasble.feature, detach.feature etc). This way we have a cohesive set of scenario tests for each CLI command parameter (attach, detach, status, enable, disable) which exercise both "happy path" (correct) and "sad path" (expected failure paths).
* maybe rename  this feature to "Feature: Attach behavior with various tokens"
* As you extend the scenarios we probably need a more correct Scenario text, like "Attach command using invalid token"
* assuming with your next branch you add something like a "Scenario: Attach with a valid contract token in a trusty lxd  container" to this feature file.

OddBloke: Oh, this is my mistake actually, I hadn't grokked the file/feature naming. I don't feel strongly that we need a single file per CLI command, though that may be easier to manage until we find that one is getting too large.

OddBloke: That last comment was a little terse, let me unpack that a bit: I think leaving the naming as-is for this PR is fine. Then, in the next PR, we can discuss whether or not we want a separate file for the new scenario being added. If we decide that we don't want separate files, then we can rename things as part of that. How does that sound?

features/attach_invalid_token.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Command behaviour when trying to attach a machine to an Ubuntu
+         Advantage subscription using an invalid token
+
+    Scenario: Attach command in a trusty lxd container
+       Given a trusty lxd container with ubuntu-advantage-tools installed
+        When I attach "<contract_token>" with sudo
```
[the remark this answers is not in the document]
patriciadomin: weird, because I was printing it to make sure I had the token and then `
`['sudo', 'ua', 'attach', '0dieQQV2IFx2GvOLN6dsZJjOuIYEK']` in this case I didn't provide it via CLI. so it is a random token, but Ill double check it

features/attach_invalid_token.feature
blackboxsw: Works for me. file renaming can be resolved at a later time if needed. Good to know that one feature file per `CLI command` isn't necessarily a goal. So we can drop that first and second bullet point suggestions

features/attach_invalid_token.feature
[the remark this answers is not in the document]
blackboxsw: sorry that the content was unclear. I added the raise RuntimeError(token) to the first line inside your when_i_attach_token() method, before you performed the reset of `token = context.config.contract_token`

The issue I was trying to suggest here wasn't that you didn't override that token value, my point was instead that when_i_attach_token() method shouldn't completely ignore it's template input variable 'token' param in the first place. That `token` method param should somehow be used to "lookup" various token values off of context.config.

As you currently have written when_i_attach_token() , you have hard-coded this method to always use `context.config.contract_token` it doesn't matter if our feature file has a

 When I attach "<contract_token>" with sudo
or 
When I attach "some_valid_token_string" with sudo

both Scenario When statements will only be exercise ua attach <context.config.contract_token>

PR #993, opened 2020-02-11
util: use `dpkg --print-architecture` in get_platform_info

### LINKED ISSUE #914: Do not use `uname` to determine architecture
While addressing issue #911 , it came to my attention that there is a mismatch in architecture names as they are being used in the contract server and in the ua client.
The server seems to be using these names (taken from https://github.com/CanonicalLtd/ua-contracts/blob/master/products/resource-esm.yaml):
  - arm64
  - armhf
  - i386
  - ppc64le
  - s390x
  - x86_64

The client, however, uses the `uname` output:
```python
def get_platform_info() -> "Dict[str, str]":
...
    uname = os.uname()
    platform_info["kernel"] = uname.release
    platform_info["arch"] = uname.machine
```
And that is not a match for many of the architecture names that the server is using. Furthermore, it's not a match for the output of `dpkg --print-architecture` which is what userspace is actually running.
`uname -m` can be deceiving in many scenarios, like chroots.
In some cases, the difference between `uname` and `dpkg` is crazy subtle:
```
ubuntu@trusty-ppc64el:~$ uname -m
ppc64le
ubuntu@trusty-ppc64el:~$ dpkg --print-architecture 
ppc64el
```
So, with all that said, what matters in the end is to use the same name everywhere. Preferably, we should be using the same names as `dpkg --print-architechture`, as that represents the user space more accurately,

The only architecture where the output of `uname -m` and `dpkg --print-architecture` match is apparently `s390x`. Here is a list:

dpkg --print-architecture | uname -m | contract server config
-- | -- | --
arm64 | aarch64 | arm64
armhf | armv7l | armhf
amd64 | x86_64 | x86_64
i386 | i686 | i386
ppc64el | ppc64le | ppc64le
s390x | s390x | s390x

OddBloke:
As well as adding some patches that were missing (and started causing failures once `get_platform_info` was invoking `util.subp`).

Fixes #914

### DIFF
--- a/uaclient/entitlements/tests/test_livepatch.py
+++ b/uaclient/entitlements/tests/test_livepatch.py
@@ -446,6 +446,7 @@ class TestLivepatchEntitlementEnable:
 
     @pytest.mark.parametrize("caplog_text", [logging.DEBUG], indirect=True)
     @pytest.mark.parametrize("apt_update_success", (True, False))
+    @mock.patch("uaclient.util.get_platform_info")
     @mock.patch("uaclient.util.subp")
     @mock.patch("uaclient.apt.run_apt_command")
     @mock.patch("uaclient.util.which", return_value=False)
@@ -458,6 +459,7 @@ class TestLivepatchEntitlementEnable:
         m_which,
         m_run_apt,
         m_subp,
+        _m_get_platform_info,
         capsys,
         caplog_text,
         entitlement,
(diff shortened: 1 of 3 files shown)


powersj:
The changes are in staging, @blackboxsw is preforming validation. Once he has it complete, I am +1 on merging.

Issue #643, opened 2019-06-18
Catch tracebacks at cli and suggest that a user file a bug


blackboxsw:
Tracebacks at the command line are of limited value to the typical person at the keyboard.
The tracebacks are already recorded in /var/log/ubuntu-advantage.log. 

UA client should catch leaked tracebacks and suggest that the user file an issue so that the client can better provide a sensible error message to the user.

panlinux:
We have to be careful to not include `/var/log/ubuntu-advantage.log` in bug reports as that will have credentials.

powersj:
I believe we have started capturing tracebacks that make sense (e.g. no internet, dns issue, etc.). I am going to close this and suggest we continue to do that on an as needed basis.

OddBloke:
On Wed, Oct 23, 2019 at 02:37:40PM -0700, Joshua Powers wrote:
> I believe we have started capturing tracebacks that make sense (e.g.
> no internet, dns issue, etc.). I am going to close this and suggest we
> continue to do that on an as needed basis.

This issue is intended to cover _all_ tracebacks (i.e. `except
Exception:`), not just the ones caused by exception classes that we have
worked out need special handling.  Currently if we see an unexpected
exception, users will get the full traceback in their console, with no
friendly messaging.

I think we can do better than this without much work, it's just a case
of adding something like:

```py
except Exception:
    with util.disable_log_to_console():
        log.exception('log msg')
    print('user facing msg', file=sys.stderr)
    sys.exit(1)
```

to main_error_handler.

PR #995, opened 2020-02-13
ux: catch any unexpected tracebacks and report a user-friendly message

blackboxsw:
In the event that client hits an unexpected error, tracebacks at the
CLI don't help most users. We want to prompt users to inspect
/var/log/ubuntu-advantage.log if possible and file a bug with
ubuntu-bug ubuntu-advantage-tools

Fixes: #643

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -596,6 +596,11 @@ def main_error_handler(func):
                 logging.exception(exc.msg)
             print("{}".format(exc.msg), file=sys.stderr)
             sys.exit(exc.exit_code)
+        except Exception:
+            with util.disable_log_to_console():
+                logging.exception("Unhandled exception, please file a bug")
+            print(ua_status.MESSAGE_UNEXPECTED_ERROR, file=sys.stderr)
+            sys.exit(1)
 
     return wrapper
 
--- a/uaclient/tests/test_cli.py
+++ b/uaclient/tests/test_cli.py
@@ -230,30 +230,50 @@ class TestRequireValidEntitlementName:
 
 
 class TestMain:
+    @pytest.mark.parametrize(
+        "exception,expected_error_msg,expected_log",
+        (
+            (
+                KeyboardInterrupt,
+                "Interrupt received; exiting.\n",
+                "KeyboardInterrupt",
+            ),
+            (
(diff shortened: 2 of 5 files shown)


uaclient/cli.py
```diff
@@ -596,6 +596,16 @@ def wrapper(*args, **kwargs):
                 logging.exception(exc.msg)
             print("{}".format(exc.msg), file=sys.stderr)
             sys.exit(exc.exit_code)
+        except Exception as exc:
+            with util.disable_log_to_console():
+                logging.exception(exc)
+            print(
+                ua_status.MESSAGE_UNEXPECTED_ERROR_DURING_OP_TMPL.format(
+                    operation="command: ua {}".format(" ".join(sys.argv[1:]))
```
OddBloke: I don't know that we want to echo back the full command. It could lead to a very long line (consider that a command could be as long as `ua attach --no-auto-enable myverylongtokeninhere`), and they'll already know what they just ran. (Also, handily, we now log the arguments. ^_^)  What do you think?

uaclient/cli.py
```diff
@@ -596,6 +596,16 @@ def wrapper(*args, **kwargs):
                 logging.exception(exc.msg)
             print("{}".format(exc.msg), file=sys.stderr)
             sys.exit(exc.exit_code)
+        except Exception as exc:
+            with util.disable_log_to_console():
+                logging.exception(exc)
```
OddBloke: As we've discussed before, we can pass in a message about the _context_ of the exception handling here, as we will get the details of the exception because we call `logging.exception`.

uaclient/cli.py
blackboxsw: +1 dropped full command output in failure message.

uaclient/cli.py
blackboxsw: Added the bookend log message for context since we already log the command itself we should be in good shape.

logging.exception("Unhandled exception, please file a bug")

uaclient/tests/test_cli.py
```diff
@@ -230,30 +230,50 @@ def test_function(args, cfg):
 
 
 class TestMain:
+    @pytest.mark.parametrize(
```
OddBloke: Nice use of parameterisation!

PR #998, opened 2020-02-19
testing: override destroy_instances to avoid vm teardown by behave

blackboxsw:
Integration tests using behave will now observe the environment variable
UACLIENT_BEHAVE_DESTROY_INSTANCES=0 which will allow behave to leave
the test instance running after completion of the integration test.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -22,6 +22,9 @@ class UAClientBehaveConfig:
         A string with an image name that should be used instead of building a
         fresh image for this test run.  If specified, image_clean will be set
         to False.
+    :param destroy_instances:
+        This boolean indicates that test containers should be destroyed after
+        the completion. Set to False to leave instances running.
     """
 
     prefix = "UACLIENT_BEHAVE_"
@@ -29,7 +32,7 @@ class UAClientBehaveConfig:
     # These variables are used in .from_environ() to convert the string
     # environment variable input to the appropriate Python types for use within
     # the test framework
-    boolean_options = ["image_clean"]
+    boolean_options = ["image_clean", "destroy_instances"]
     str_options = ["reuse_image"]
 
     # This variable is used in .from_environ() but also to emit the "Config
@@ -37,10 +40,15 @@ class UAClientBehaveConfig:
     all_options = boolean_options + str_options
 
     def __init__(
-        self, *, image_clean: bool = True, reuse_image: str = None
+        self,
+        *,
+        image_clean: bool = True,
+        reuse_image: str = None,
+        destroy_instances: bool = True
     ) -> None:
         # First, store the values we've detected
         self.image_clean = image_clean
+        self.destroy_instances = destroy_instances
         self.reuse_image = reuse_image
 
         # Next, perform any required validation
--- a/features/util.py
+++ b/features/util.py
@@ -23,7 +23,10 @@ def launch_lxd_container(
     subprocess.run(["lxc", "launch", image_name, container_name])
 
     def cleanup_container() -> None:
-        subprocess.run(["lxc", "delete", "-f", container_name])
+        if not context.config.destroy_instances:
+            print("Leaving lxd container running: {}".format(container_name))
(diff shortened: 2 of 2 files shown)


features/environment.py
```diff
@@ -22,25 +22,34 @@ class UAClientBehaveConfig:
         A string with an image name that should be used instead of building a
         fresh image for this test run.  If specified, image_clean will be set
         to False.
+    :param preserve_instance:
+        This boolean indicates that a test VM should be left running after test
```
OddBloke:
```suggestion
        This boolean indicates that a test container should be left running after test
```

features/environment.py
```diff
@@ -22,25 +22,34 @@ class UAClientBehaveConfig:
...
+        completion. Default is to destroy the machine under test after test
```
OddBloke:
```suggestion
        completion. Default is to destroy the test container after test
```

features/environment.py
```diff
@@ -22,25 +22,34 @@ class UAClientBehaveConfig:
...
     # options" stanza in __init__
     all_options = boolean_options + str_options
 
     def __init__(
-        self, *, image_clean: bool = True, reuse_image: str = None
+        self,
+        *,
+        image_clean: bool = True,
+        reuse_image: str = None,
+        preserve_instance: bool = False
```
OddBloke: I'm not entirely convinced that the pre-existing code as-written supports default-to-False perfectly; if you ever set this in your environment, then you have to either set it to "0" or you have to `unset` it entirely to disable it. Unexpectedly:

```
$ UACLIENT_BEHAVE_PRESERVE_INSTANCE=yes
# Oh, cool, it's working now
$ UACLIENT_BEHAVE_PRESERVE_INSTANCE=
$ behave
```

will set `preserve_instance` to `True`, despite the (to me) clear intent being the opposite. (You can see in `.from_environ` on this class that we "Default to True", which is why this happens.)

I don't think this is worth thinking about the semantics of (or writing code to handle) defaulting to false because a simpler alternative exists: what if we invert the config option here to `UACLIENT_BEHAVE_DESTROY_INSTANCE` which then _can_ default to True in a way that we already handle. Then:


```
$ UACLIENT_BEHAVE_DESTROY_INSTANCE=0
# Oh, cool, it's working now
$ UACLIENT_BEHAVE_DESTROY_INSTANCE=
$ behave
```

_does_ behave as we would expect (i.e. the instance is destroyed).

What do you think?

features/util.py
```diff
@@ -23,7 +23,10 @@ def launch_lxd_container(
     subprocess.run(["lxc", "launch", image_name, container_name])
 
     def cleanup_container() -> None:
-        subprocess.run(["lxc", "delete", "-f", container_name])
+        if context.config.preserve_instance:
+            print("Leaving lxc container running: {}".format(container_name))
```
OddBloke: Minor one, but the step definitions refer to them as "lxd containers", and we can use `print`s behaviour to simplify a little:

```suggestion
            print("Leaving lxd container running:", container_name)
```

features/environment.py
```diff
@@ -22,25 +22,34 @@ class UAClientBehaveConfig:
         A string with an image name that should be used instead of building a
         fresh image for this test run.  If specified, image_clean will be set
         to False.
+    :param preserve_instance:
```
OddBloke: We launch many containers during test runs, not just one; should the option be plural?

features/environment.py
```diff
@@ -22,25 +22,33 @@ class UAClientBehaveConfig:
         A string with an image name that should be used instead of building a
         fresh image for this test run.  If specified, image_clean will be set
         to False.
+    :param destroy_instances:
+        This boolean indicates that test containers should be destroyed after
+        the completion. Set to 0 to leave instances running.
```
OddBloke: This is Python-level documentation, so this should be:

```suggestion
        the completion. Set to False to leave instances running.
```

PR #1002, opened 2020-02-20
testing: add integration test for behavior of attaching a valid_token

patriciadomin:
TODO list:
- figure out a way of checking the colored status
  - currently it is not working because if you are entitled to a service and if
    it is enabled - they are green.
- improve skip option: the test do not run, but didn't show the message
- maybe check if contract_token is None inside `UAClientBehaveConfig` class

### DIFF
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -1,6 +1,7 @@
 Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         subscription using a valid token
 
+    @uses.config.contract_token
     Scenario: Attach command in a trusty lxd container
        Given a trusty lxd container with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -1,0 +1,31 @@
+Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
+        subscription using a valid token
+
+    Scenario: Attach command in a trusty lxd container
+       Given a trusty lxd container with ubuntu-advantage-tools installed
+        When I attach "<contract_token>" with sudo
+        Then stdout will include:
+        """
+        ESM Infra enabled
+        """
+        And stdout will include:
+        """
+        This machine is now attached to
+        """
+        And stdout will include:
+        """
+        SERVICE       ENTITLED  STATUS    DESCRIPTION
+        cc-eal        yes       n/a       Common Criteria EAL2 Provisioning Packages
+        cis-audit     no        —         Center for Internet Security Audit Tools
+        esm-apps      no        —         UA Apps: Extended Security Maintenance
+        esm-infra     yes       enabled   UA Infra: Extended Security Maintenance
+        fips          yes       n/a       NIST-certified FIPS modules
+        fips-updates  yes       n/a       Uncertified security updates to FIPS modules
+        livepatch     yes       n/a       Canonical Livepatch service
+
+        Enable services with: ua enable <service>
+        """
+        And I will see the following on stderr:
+        """
+        Enabling default service esm-infra
+        """
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -4,26 +4,24 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
     Scenario: Attach command in a trusty lxd container
        Given a trusty lxd container with ubuntu-advantage-tools installed
         When I attach "<contract_token>" with sudo
-        Then stdout will include:
+        Then stdout matches regexp:
         """
         ESM Infra enabled
         """
-        And stdout will include:
+        And stdout matches regexp:
         """
         This machine is now attached to
         """
-        And stdout will include:
+        And stdout matches regexp:
         """
         SERVICE       ENTITLED  STATUS    DESCRIPTION
-        cc-eal        yes       n/a       Common Criteria EAL2 Provisioning Packages
-        cis-audit     no        —         Center for Internet Security Audit Tools
-        esm-apps      no        —         UA Apps: Extended Security Maintenance
-        esm-infra     yes       enabled   UA Infra: Extended Security Maintenance
-        fips          yes       n/a       NIST-certified FIPS modules
-        fips-updates  yes       n/a       Uncertified security updates to FIPS modules
-        livepatch     yes       n/a       Canonical Livepatch service
-
-        Enable services with: ua enable <service>
+        cc-eal       +yes      +n/a      +Common Criteria EAL2 Provisioning Packages
+        cis-audit    +no       +—        +Center for Internet Security Audit Tools
+        esm-apps     +no       +—        +UA Apps: Extended Security Maintenance
+        esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance
+        fips         +yes      +n/a      +NIST-certified FIPS modules
+        fips-updates +yes      +n/a      +Uncertified security updates to FIPS modules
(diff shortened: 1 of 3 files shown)


features/steps/steps.py
```diff
@@ -36,11 +36,38 @@ def when_i_run_command(context, command, user_spec):
     context.process = process
 
 
+@when('I attach "{token}" {user_spec}')
+def when_i_attach_token(context, token, user_spec):
```
blackboxsw: Either we should make use of the token parameter being passed into this function, or we probably could drop it.

alternative 1:
```
@when('I attach "{token_name}" {user_spec}')
def when_i_attach_token(context, token_name, user_spec):
    token = getattr(context.config, token_name, None)
```
The above would require that you change your  attach_validtoken.feature
```When I attach "<contract_token>" with sudo```
changed to 
```When I attach "contract_token" with sudo```

File: features/attach_validtoken.feature [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+        ESM Infra enabled
+        """
+        And stdout will include:
+        """
+        This machine is now attached to
+        """
+        And stdout will include:
+        """
+        SERVICE       ENTITLED  STATUS    DESCRIPTION
+        cc-eal        yes       n/a       Common Criteria EAL2 Provisioning Packages
```
blackboxsw: As you mentioned you are dealing with non-printable control characters being turned into spaces offsetting the stdout alignment for the cc-eal  "yes" and "n.a" columns when compare context.process.stdout to the properly aligned context.text in your "stdout will include" assertion.

One minimal solution could have been to just copy and paste the failure output text you see and replace the misaligned text including those non-printables notice the n/a doesn't align with STATUS header

```
          """
          cc-eal        yes                n/a                Common Criteria EAL2 Provisioning Packages
          cis-audit     no                 —                  Center for Internet Security Audit Tools
          esm-apps      no                 —                  UA Apps: Extended Security Maintenance
          esm-infra     yes                enabled            UA Infra: Extended Security Maintenance
          fips          yes                n/a                NIST-certified FIPS modules
          fips-updates  yes                n/a                Uncertified security updates to FIPS modules
          livepatch     yes                n/a                Canonical Livepatch service
          """        
```
The above suggestion will get you a pass because it'll be the exact misaligned content that behave was complaining about.

Better alternative solution:

I think you really need a regex "then" step

So this step should be the following: use the + to indicate greater than one space
```suggestion
And stdout matches regexp:
      """
      SERVICE       ENTITLED  STATUS    DESCRIPTION
      cc-eal       +yes      +n/a      +Common Criteria EAL2 Provisioning Packages
      cis-audit    +no       +—        +Center for Internet Security Audit Tools
      esm-apps     +no       +—        +UA Apps: Extended Security Maintenance
      esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance
      fips         +yes      +n/a      +NIST-certified FIPS modules
      fips-updates +yes      +n/a      +Uncertified security updates to FIPS modules
      livepatch    +yes      +n/a      +Canonical Livepatch service
      """
```
Your new @then step is commented below

features/steps/steps.py
```diff
@@ -36,11 +36,38 @@ def when_i_run_command(context, command, user_spec):
...
+    when_i_run_command(context, "ua attach %s" % token, user_spec)
+
+
 @then("I will see the following on stdout")
 def then_i_will_see_on_stdout(context):
     assert_that(context.process.stdout.strip(), equal_to(context.text))
 
 
+@then("stdout will include")
+def then_and_stdout_will_include(context):
```
blackboxsw: typo fix function name, can drop "_and"

features/steps/steps.py
```diff
@@ -36,11 +36,38 @@ def when_i_run_command(context, command, user_spec):
...
+    assert_that(context.process.stdout.strip(), contains_string(context.text))
+
```
blackboxsw: I think you separately need a new @then step as commented above

```suggestion
+@then("stdout matches regexp")
+def then_stdout_matches_regexp(context):
+    assert_that(context.process.stdout.strip(), matches_regexp(context.text))
```

features/steps/steps.py
```diff
@@ -19,7 +19,7 @@ def given_a_trusty_lxd_container(context):
 
 @when("I run `{command}` {user_spec}")
 def when_i_run_command(context, command, user_spec):
-    prefix = []
+    prefix = get_command_prefix_for_user_spec(user_spec)
     if user_spec == "with sudo":
```
blackboxsw: I think you can drop this if/else now as you are setting prefix properly on the line above

patriciadomin:
don't review it yet, I'm addressing the other comments

features/steps/steps.py
patriciadomin: I can drop this hook and use just the one with `matches_regexp` instead of it + `contains_string`

features/environment.py
```diff
@@ -34,19 +36,22 @@ class UAClientBehaveConfig:
     # the test framework
     boolean_options = ["image_clean", "destroy_instances"]
     str_options = ["reuse_image"]
```
blackboxsw: You probably want to continue to include "contract_token" in the str_options list as well because it is a str type attribute an we may find it confusing in the future about why it is excluded from this list if we are operating specially on all self.str_options. 

I realize at the moment we are only iterating over self.boolean_options in from_environ() to set a default boolean value to True, but that could change in the future and 'future us' would likely miss including both str_options and redact_options if we are operating on all string type attributes to set defaults

features/environment.py
```diff
@@ -34,19 +36,22 @@ class UAClientBehaveConfig:
...
+    redact_options = ["contract_token"]
 
     # This variable is used in .from_environ() but also to emit the "Config
     # options" stanza in __init__
-    all_options = boolean_options + str_options
+    all_options = boolean_options + str_options + redact_options
```
blackboxsw: If we also kept "contract_token" in the str_options list, we wouldn't need to add those items to "all_options" here and we could drop this.

features/steps/steps.py
blackboxsw: Just touching this point to check to see if it was missed.

features/environment.py
patriciadomin: ok. I've changed it.

features/steps/steps.py
patriciadomin: sorry. Please check if it is right now. I've changed it

features/steps/steps.py
```diff
@@ -36,11 +29,38 @@ def when_i_run_command(context, command, user_spec):
     context.process = process
 
 
+@when("I attach `{contrac_token}` {user_spec}")
+def when_i_attach_token(context, contract_token, user_spec):
+    token = context.config.contract_token
+    if not token:
+        context.feature.skip(
```
blackboxsw: I have an idea here using before_scenario, I'll post that once I have a working suggestion

features/attach_validtoken.feature
```diff
@@ -0,0 +1,29 @@
+Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
+        subscription using a valid token
+
+    Scenario: Attach command in a trusty lxd container
+       Given a trusty lxd container with ubuntu-advantage-tools installed
+        When I attach "contract_token" with sudo
```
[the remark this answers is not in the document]
patriciadomin: sorry. It is fixed. I've added just one commit and I forgot the other one.

features/steps/steps.py
patriciadomin: show-skipped = True is default. You can add to cli `--no-skipped` and then do not print skipped step/feature/scenario, but still, it doesn't show: 
`context.feature.skip("Skipping test because you didn't provide a token")`

PR #1004, opened 2020-02-23
contract: drop remaining support for request_resource_machine_access

blackboxsw:
Drop remaining call site of request_resource_machine_access and any UAConfig support
for machine-access data paths.

Entitlements will now call contract.request_updated_contract instead of request_resource_machine_access when entitlement is found expired.

Additionally we want to clean the unused /var/lib/ubuntu-advantage/private/machine-access*json files upon upgrade, since they will not be read again by ubuntu-advantage-tools

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -78,6 +78,8 @@ case "$1" in
 
       # CACHE_DIR is no longer present or used since 19.1
       rm -rf /var/cache/ubuntu-advantage-tools
+      # machine-access cache files no longer present or used since 20.1
+      rm -rf /var/lib/ubuntu-advantage/private/machine-access*json
 
       if [ "14.04" = "$VERSION_ID" ]; then
         if echo "$ESM_SUPPORTED_ARCHS" | grep -qw "$MYARCH"; then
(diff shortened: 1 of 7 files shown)


debian/postinst
```diff
@@ -78,6 +78,8 @@ case "$1" in
 
       # CACHE_DIR is no longer present or used since 19.1
       rm -rf /var/cache/ubuntu-advantage-tools
+      # machine-access cache files no longer present or used since 20.1
+      rm -rf /var/lib/ubuntu-advantage/private/machine-access*json
```
OddBloke: I think we could make this a little tighter (and not recursive):

```suggestion
      rm -f /var/lib/ubuntu-advantage/private/machine-access-*.json
```

Or, alternatively, should we just list out the 5-6 specific files we know will exist, so that this glob can't come back to cause us problems?

blackboxsw: +1 on dropping the recursive remove. Since this directory is private root read-only, I don't see any exposure we would care to support where other cached json files will exist. If we re-introduce these cache files in the future our packaging will have to take care to drop this cleanup.

PR #1007, opened 2020-02-26
Reuse container

patriciadomin:
changes for reusing a container with userdata config

It allows to run behave choosing an existent container to run the tests in.
I've added it via `userdata` config, so just add `-D reuse_container=existent_cont`.
From that point, it will then proceed to install UAT.
If you don't add this config it stays the same.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -104,6 +104,8 @@ def before_all(context: Context) -> None:
     then capture an image.  This image is then reused by each feature, reducing
     test execution time.
     """
+    userdata = context.config.userdata
+    context.reuse_container = userdata.get("reuse_container")
     context.config = UAClientBehaveConfig.from_environ()
     if context.config.reuse_image is None:
         create_trusty_uat_lxd_image(context)
@@ -149,13 +151,6 @@ def create_trusty_uat_lxd_image(context: Context) -> None:
     :param context:
         A `behave.runner.Context`; this will have `image_name` set on it.
     """
-    now = datetime.datetime.now()
-    context.image_name = "behave-image-" + now.strftime("%s%f")
-    build_container_name = "behave-image-build-" + now.strftime("%s%f")
-
-    launch_lxd_container(context, "ubuntu:trusty", build_container_name)
-    _install_uat_in_container(build_container_name)
-    _capture_container_as_image(build_container_name, context.image_name)
 
     def image_cleanup() -> None:
         if context.config.image_clean:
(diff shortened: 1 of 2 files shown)


features/environment.py
```diff
@@ -149,21 +151,29 @@ def create_trusty_uat_lxd_image(context: Context) -> None:
     :param context:
         A `behave.runner.Context`; this will have `image_name` set on it.
     """
-    now = datetime.datetime.now()
-    context.image_name = "behave-image-" + now.strftime("%s%f")
-    build_container_name = "behave-image-build-" + now.strftime("%s%f")
+    if context.container_to_run:
+        print(
+            "It is going to run the test in the existent container: ",
```
blackboxsw:
```suggestion
            "Reusing the existing container:",
```

features/environment.py
```diff
@@ -149,21 +151,29 @@ def create_trusty_uat_lxd_image(context: Context) -> None:
...
+            context.container_to_run,
+        )
+    else:
+
```
blackboxsw: Can drop the empty line
```suggestion
```

features/environment.py
```diff
@@ -149,21 +151,29 @@ def create_trusty_uat_lxd_image(context: Context) -> None:
...
+        now = datetime.datetime.now()
+        context.image_name = "behave-image-" + now.strftime("%s%f")
+        build_container_name = "behave-image-build-" + now.strftime("%s%f")
```
blackboxsw: Let's drop this empty line too to keep this scope closer together
```suggestion
```

features/environment.py
```diff
@@ -149,21 +151,29 @@ def create_trusty_uat_lxd_image(context: Context) -> None:
...
 
     def image_cleanup() -> None:
         if context.config.image_clean:
             subprocess.run(["lxc", "image", "delete", context.image_name])
         else:
             print("Image cleanup disabled, not deleting:", context.image_name)
 
-    context.add_cleanup(image_cleanup)
+    if not context.container_to_run:
+        context.add_cleanup(image_cleanup)
```
blackboxsw: I think you can move this context.add_cleanup up within the "else:" clause just below the line
_capture_container_as_image(build_container_name, context.image_name)

You can also move the def image_cleanup() up before your line
  if context.container_to_run:

That way you don't have to indent the declaration of def image_cleanup() any further in the

features/environment.py
```diff
@@ -104,6 +104,8 @@ def before_all(context: Context) -> None:
     then capture an image.  This image is then reused by each feature, reducing
     test execution time.
     """
+    userdata = context.config.userdata
+    context.container_to_run = userdata.get("container_to_run")
```
blackboxsw: To align with the reuse_image config option name let's call this reuse_container

blackboxsw: I'm also thinking that maybe this should be an environment config too insteead of using userdata here?

But it's worth discussion as I see this as a separate developer tool when we are writing behave tests which maybe should live outside the expected CI configuration for Behave which uses environment variables.

patriciadomin: I don't see a reason as this is a unique configuration, just for development tests. This way, I think it is simpler, don't need to add it to `UAClientBehaveConfig` class

features/environment.py
patriciadomin: Ok thanks

features/environment.py
patriciadomin: Ok

features/environment.py
patriciadomin: ok

features/environment.py
patriciadomin: right, thanks, I've changed it

features/environment.py
blackboxsw: I think that comment above was related to keeping things in userdata. but not the rename?

patriciadomin: Let me change the name for reuse_container

PR #1008, opened 2020-02-26
changelog: release 20.2~0ubuntu1~20.04.1

blackboxsw:
WIP: do not land, will discuss series-specific upstream debian/changelog commits as a team


powersj:
Wouldn't this cause the daily PPAs produce some very long, odd versions? All the builds would have focal in them. I thought the plan was to keep the master branch show the upstream version, in this case 20.2, and then the PPA could add the other values or an upload would have the version edited to the correct value?

PR #1012, opened 2020-03-10
refactor contract_token parameter

patriciadomin:
It will avoid calling this step sending `contract_token` parameter that
is not taken into account, as it is provided via CLI environment variable `UACLIENT_BEHAVE_CONTRACT_TOKEN=...`.
The current way, doesn't matter what the function has received
because `contract_token = context.config.contract_token` overrides it.


patriciadomin:
thanks @blackboxsw I've updated it, let me know if it is better/ok now.

PR #1013, opened 2020-03-10
tests: integration tests for subcommands when attached

patriciadomin:


features/attached_commands.feature
```diff
@@ -0,0 +1,85 @@
...
+            This command must be run as root (try using sudo)
+            """
+            #When The machine is attached
+        When I run `ua refresh` with sudo
+        Then I will see the following on stdout:
+            """
+            Successfully refreshed your subscription
+            """
+
+    Scenario: Attached disable of a known service in a trusty lxd container
```
blackboxsw:
```suggestion
    Scenario: Attached disable of an already disabled service in a trusty lxd container
```

features/attached_commands.feature
```diff
@@ -0,0 +1,85 @@
...
+        Then I will see the following on stdout:
+            """
+            Updating package lists
+            """
+        When I run `ua status` with sudo
+        Then stdout matches regexp:
+            """
+            esm-infra    +yes      +disabled +UA Infra: Extended Security Maintenance
+            """
+
```
blackboxsw: I would be nice to also add a step that allows us to validate that esm repositories are properly disabled on trusty to the magic -32768 because this behavior will differ on Trusty (which advertizes available esm packages) versus Xenial and Bionic (which should not have any esm repos seen when esm is disabled).

```suggestion
    When I run `apt-cache policy` with sudo
    Then stdout matches regexp
        """
        -32768 https://esm.ubuntu.com/ubuntu/ trusty-infra-updates/main amd64 Packages
          +origin esm.ubuntu.com
        -32768 https://esm.ubuntu.com/ubuntu/ trusty-infra-security/main amd64 Packages
          +origin esm.ubuntu.com
        """
```

features/attached_commands.feature
```diff
@@ -0,0 +1,85 @@
...
+            This command must be run as root (try using sudo)
+            """
+        When I run `ua detach --assume-yes` with sudo
+        Then I will see the following on stdout:
+            """
+            Detach will disable the following service:
+                esm-infra
+            Updating package lists
+            This machine is now detached
+            """
```
blackboxsw: Pleas also add a when/then step to run/check ua status output as non-root to ensure ua status is updated appropriately to unattached text here. (as we had bugs in the past about that)

patriciadomin: Ok

features/attached_commands.feature
```diff
@@ -0,0 +1,85 @@
...
+
+    Scenario: Attached refresh in a trusty lxd container
+        Given a trusty lxd container with ubuntu-advantage-tools installed
+        When I attach contract_token with sudo
+        And I run `ua refresh` as non-root
+        Then I will see the following on stderr:
+            """
+            This command must be run as root (try using sudo)
+            """
+            #When The machine is attached
```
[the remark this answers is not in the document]
patriciadomin: Chad, there is a step with that `Scenario: Attached detach in a trusty lxd container` ...  
`And I run `ua detach` as non-root`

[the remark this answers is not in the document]
patriciadomin: I agree that we could have some scenarios joined together, but we could also revisit/refactor it in the future (maybe when we are going to have VMs) because currently we just have tests running in container(s).

features/attached_commands.feature
patriciadomin: ok changed it . thanks

features/attached_commands.feature
blackboxsw: hi @patriciadomin your suggestion failure that you pinged about per https://pastebin.canonical.com/p/RMz2PM9Xcr/
is because you ran `apt-cache policy ubuntu-advantage-tools` which only tells you about the specific package.

Note the above command I mentioned was only `apt-cache policy`. Which gives you a list of all apt repositories that the machine has access to. this will give you the expected output.

PR #1014, opened 2020-03-11
add tests for behaviour of enable command when attached to an UA subscription

patriciadomin:
add tests for each one of the available services when attached to an UA subscription

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -1,0 +1,131 @@
+@uses.config.contract_token
+Feature: Enable command behaviour when attached to an UA subscription
+
+
+    Scenario: Attached enable Livepatch service in a trusty lxd container
+        Given a trusty lxd container with ubuntu-advantage-tools installed
+        When I attach contract_token with sudo
+        And I run `ua enable livepatch` as non-root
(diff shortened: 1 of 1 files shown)


features/attached_enable.feature
```diff
@@ -0,0 +1,131 @@
...
+            """
+            This command must be run as root (try using sudo)
+            """
+        When I run `ua enable livepatch` with sudo
+        Then I will see the following on stdout:
+            """
+            One moment, checking your subscription first
+            Cannot install Livepatch on a container
+            """
+
```
blackboxsw: @patriciadomin Maybe as your next PR, can we add another test to this feature that also starts with a new

Given a trusty lxd container with ubuntu-advantage-tools and HWE kernel

Installing the HWE kernel on trusty should be
 sudo apt-get install --install-recommends linux-generic-lts-xenial 
as seen here https://wiki.ubuntu.com/Kernel/LTSEnablementStack

I think the HWE kernel test can replace the above scenario when written, but it doesn't need to be this PR.
 Scenario: Attached enable Livepatch service in a trusty lxd container

PR #1016, opened 2020-03-17
Bug/reattach across instance id delta

blackboxsw:
When VMs are cloned and relaunched in Azure and AWS, instance-ids change for the new vms.
cloud-init picks up on those instance-id changes and re-runs. UA auto-attach should be cognizant of a delta in instance-id and re-attach on Pro instances.

LP: #1867573

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -299,11 +299,20 @@ def action_enable(args, cfg):
 @assert_root
 @assert_attached()
 def action_detach(args, cfg) -> int:
+    """Perform the detach action for this machine.
+
+    @return: 0 on success, 1 otherwise
+    """
     return _detach(cfg, assume_yes=args.assume_yes)
 
 
 def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
-    """Perform the detach action for this machine.
+    """Detach the machine from the active Ubuntu Advantage subscription,
+
+    :param cfg: a ``config.UAConfig`` instance
+    :param assume_yes: Assume a yes answer to any prompts requested.
+         In this case, it means automatically disable any service during
+         detach.
 
     @return: 0 on success, 1 otherwise
     """
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -298,7 +298,11 @@ def action_enable(args, cfg):
 
 @assert_root
 @assert_attached()
-def action_detach(args, cfg):
+def action_detach(args, cfg) -> int:
+    return _detach(cfg, assume_yes=args.assume_yes)
+
+
+def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
     """Perform the detach action for this machine.
 
     @return: 0 on success, 1 otherwise
@@ -313,7 +317,7 @@ def action_detach(args, cfg):
         print("Detach will disable the following service{}:".format(suffix))
         for ent in to_disable:
             print("    {}".format(ent.name))
-    if not args.assume_yes and not util.prompt_for_confirmation():
+    if not assume_yes and not util.prompt_for_confirmation():
         return 1
     for ent in to_disable:
         ent.disable(silent=True)
@@ -372,7 +376,22 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
 
     :return: contract token obtained from identity doc
     """
-    instance = identity.cloud_instance_factory()
+    try:
+        instance = identity.cloud_instance_factory()
+    except exceptions.UserFacingError as e:
+        if cfg.is_attached:
+            # We are attached on non-Pro Image, just report already attached
+            raise exceptions.AlreadyAttachedError(cfg)
+        # Unattached on non-Pro return UserFacing error msg details
+        raise e
+    current_iid = identity.get_instance_id()
+    if cfg.is_attached:
+        prev_iid = cfg.read_cache("instance-id")
+        if current_iid == prev_iid:
+            raise exceptions.AlreadyAttachedError(cfg)
+        print("Re-attaching Ubuntu Advantage subscription on new instance")
+        _detach(cfg, assume_yes=True)
+
     contract_client = contract.UAContractClient(cfg)
     try:
         tokenResponse = contract_client.request_auto_attach_contract_token(
@@ -384,10 +403,12 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
                 ua_status.MESSAGE_UNSUPPORTED_AUTO_ATTACH
             )
         raise e
+    if current_iid:
+        cfg.write_cache("instance-id", current_iid)
+
     return tokenResponse["contractToken"]
 
 
-@assert_not_attached
 @assert_root
 def action_auto_attach(args, cfg):
     token = _get_contract_token_from_cloud_identity(cfg)
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -399,8 +399,10 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
         if current_iid == prev_iid:
             raise exceptions.AlreadyAttachedError(cfg)
         print("Re-attaching Ubuntu Advantage subscription on new instance")
-        _detach(cfg, assume_yes=True)
-
+        if _detach(cfg, assume_yes=True) != 0:
+            raise exceptions.UserFacingError(
+                ua_status.MESSAGE_DETACH_AUTOMATION_FAILURE
+            )
     contract_client = contract.UAContractClient(cfg)
     try:
         tokenResponse = contract_client.request_auto_attach_contract_token(
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -1,4 +1,6 @@
 import json
+import logging
+import os
 
 from uaclient import exceptions
 from uaclient import clouds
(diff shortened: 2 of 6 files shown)


blackboxsw:
LP:1867573

uaclient/cli.py
```diff
@@ -298,7 +298,11 @@ def action_enable(args, cfg):
 
 @assert_root
 @assert_attached()
-def action_detach(args, cfg):
+def action_detach(args, cfg) -> int:
```
OddBloke: Can we add a docstring here?

uaclient/cli.py
```diff
@@ -384,10 +403,13 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
                 ua_status.MESSAGE_UNSUPPORTED_AUTO_ATTACH
             )
         raise e
+    instance_id_file = os.path.join(cfg.data_dir, "instance-id")
+    if current_iid:
+        util.write_file(instance_id_file, current_iid)
```
OddBloke: Should this be using `UAConfig.write_cache`?

uaclient/clouds/identity.py
```diff
@@ -11,13 +13,32 @@
...
 
 # Mapping of datasource names to cloud-id responses. Trusty compat with Xenial+
 DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws"}
 
 
+def get_instance_id(
+    iid_file: str = CLOUDINIT_INSTANCE_ID_FILE
+) -> "Optional[str]":
+    """Query cloud instance-id from cmdline or CLOUDINIT_INSTANCE_ID_FILE"""
+    if util.which("cloud-id"):
```
OddBloke: Using `cloud-id`s presence as a guard makes sense in `get_cloud_type` because it's then called there. Using its presence as a proxy for cloud-init version feels a little less obvious. Would either checking the Ubuntu release version, or explicitly checking for a lower-bound cloud-init version work instead?

uaclient/clouds/identity.py
```diff
@@ -11,13 +13,32 @@
...
 CLOUDINIT_RESULT_FILE = "/var/lib/cloud/data/result.json"
+CLOUDINIT_INSTANCE_ID_FILE = "/var/lib/cloud/data/instance-id"
 
 
 # Mapping of datasource names to cloud-id responses. Trusty compat with Xenial+
 DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws"}
 
 
+def get_instance_id(
+    iid_file: str = CLOUDINIT_INSTANCE_ID_FILE
```
OddBloke: This is only used in testing, so perhaps:

```suggestion
    _iid_file: str = CLOUDINIT_INSTANCE_ID_FILE
```

uaclient/clouds/identity.py
```diff
@@ -11,13 +13,32 @@
...
+    """Query cloud instance-id from cmdline or CLOUDINIT_INSTANCE_ID_FILE"""
+    if util.which("cloud-id"):
+        # Present in cloud-init on >= Xenial
+        out, _err = util.subp(["cloud-init", "query", "instance_id"])
+        return out.strip()
+    if os.path.exists(iid_file):
+        return util.load_file(iid_file)
+    else:
+        logging.warning(
+            "Unable to determine current instance-id from %s" % iid_file
```
OddBloke: No need to perform the substitution in logging calls:

```suggestion
            "Unable to determine current instance-id from %s", iid_file
```

uaclient/clouds/identity.py
```diff
@@ -11,13 +13,32 @@
...
+) -> "Optional[str]":
+    """Query cloud instance-id from cmdline or CLOUDINIT_INSTANCE_ID_FILE"""
+    if util.which("cloud-id"):
+        # Present in cloud-init on >= Xenial
+        out, _err = util.subp(["cloud-init", "query", "instance_id"])
+        return out.strip()
+    if os.path.exists(iid_file):
+        return util.load_file(iid_file)
+    else:
+        logging.warning(
```
OddBloke: A nit, but given that the `if` above returns unconditionally, we don't need `else` or an extra level of indentation here.

uaclient/cli.py
```diff
@@ -372,7 +376,22 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
...
+            raise exceptions.AlreadyAttachedError(cfg)
+        # Unattached on non-Pro return UserFacing error msg details
+        raise e
+    current_iid = identity.get_instance_id()
+    if cfg.is_attached:
+        prev_iid = util.load_file(os.path.join(cfg.data_dir, "instance-id"))
+        if current_iid == prev_iid:
+            raise exceptions.AlreadyAttachedError(cfg)
+        print("Re-attaching Ubuntu Advantage subscription on new instance")
+        _detach(cfg, assume_yes=True)
```
OddBloke: This may return `1` on failure instead of raising an exception (though currently it won't with `assume_yes=True`). Should that case be handled here?

uaclient/clouds/identity.py
```diff
@@ -11,13 +13,32 @@
...
 
+def get_instance_id(
+    iid_file: str = CLOUDINIT_INSTANCE_ID_FILE
+) -> "Optional[str]":
+    """Query cloud instance-id from cmdline or CLOUDINIT_INSTANCE_ID_FILE"""
+    if util.which("cloud-id"):
+        # Present in cloud-init on >= Xenial
+        out, _err = util.subp(["cloud-init", "query", "instance_id"])
+        return out.strip()
+    if os.path.exists(iid_file):
```
OddBloke: Should this be using `UAConfig.read_cache`?

uaclient/cli.py
blackboxsw: d'oh

uaclient/cli.py

uaclient/clouds/identity.py
blackboxsw: I agree here. Moving it to series-based trusty/non-trusty as we know that every other supported series has a recent enough cloud-init to have this query support. I had avoided calling cloud-init query specifically as trusty cloud-init query is a mangled/useless mess for that subcommand.

uaclient/clouds/identity.py
blackboxsw: done.

uaclient/clouds/identity.py
blackboxsw: done.

uaclient/clouds/identity.py
blackboxsw: yes we should. thanks for that corrected throughout

uaclient/cli.py
blackboxsw: So a detach fails during auto-attach on new-instance, we would have not disabled and re-enabled X and X is left with potentially old credentials for X if re-attach was going to change those creds.

That said, creds are per account I thought for each service X currently, so practically it doesn't matter at the moment, but could matter in the future if creds evolve to per machine per account.

uaclient/clouds/identity.py
blackboxsw: thanks grabbed it

uaclient/cli.py
OddBloke: I'm not sure about that specific case. My concern is more general: the interface of `_detach` is that on failure, it will return `1` or raise an exception. This code as-written handles the exception failure case (because it doesn't intercept them, so they will bubble out), but it silently ignores the `return 1` case.

As it happens, today, `_detach` won't `return 1` if `assume_yes=True`, but we can't necessarily assume that will continue to be the case in future, and we shouldn't expect people to check every call site of a function when changing its behaviour within its already-defined interface.

(My assumption here is that we care that the detach failed, because we aren't catching and ignoring exceptions here.)

uaclient/clouds/tests/test_identity.py
```diff
@@ -14,6 +15,35 @@
...
 
 
+class TestGetInstanceID:
+    @pytest.mark.parametrize("series", ("xenial", "bionic", "eoan", "focal"))
+    @mock.patch(M_PATH + "util.subp", return_value=("my-iid\n", ""))
+    @mock.patch(M_PATH + "util.get_platform_info")
+    def test_use_cloud_init_query_when_non_trusty(
+        self, m_get_platform_info, m_subp, series
+    ):
+        """Get instance_id from cloud-init query when non not on trusty."""
```
OddBloke:
```suggestion
        """Get instance_id from cloud-init query when not on trusty."""
```

uaclient/cli.py
blackboxsw: Is this what you are getting at @OddBloke?
```diff
diff --git a/uaclient/cli.py b/uaclient/cli.py
index bd37c5d..0282c34 100644
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -399,8 +399,10 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
         if current_iid == prev_iid:
             raise exceptions.AlreadyAttachedError(cfg)
         print("Re-attaching Ubuntu Advantage subscription on new instance")
-        _detach(cfg, assume_yes=True)
-
+        if not _detach(cfg, assume_yes=True):
+            raise exceptions.UserFacingError(
+                ua_status.MESSAGE_DETACH_AUTOMATION_FAILURE
+            )
     contract_client = contract.UAContractClient(cfg)
     try:
         tokenResponse = contract_client.request_auto_attach_contract_token(
diff --git a/uaclient/status.py b/uaclient/status.py
index de11cab..6df7d44 100644
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -203,6 +203,7 @@ MESSAGE_ENABLE_BY_DEFAULT_TMPL = "Enabling default service {name}"
 MESSAGE_ENABLE_BY_DEFAULT_MANUAL_TMPL = """\
 Service {name} is recommended by default. Run: sudo ua enable {name}"""
 MESSAGE_DETACH_SUCCESS = "This machine is now detached"
+MESSAGE_DETACH_AUTOMATION_FAILURE = "Unable to automatically detach machine"
 
 MESSAGE_REFRESH_ENABLE = "One moment, checking your subscription first"
 MESSAGE_REFRESH_SUCCESS = "Successfully refreshed your subscription"
```

OddBloke: Yep, something like that would be great (though I would do `== 0` instead of `not`, because it's an int not a bool).

blackboxsw:
Tested on trusty Azure PRO system. and got the expected behavior:
 * first reboot after upgrade  re-attaches because instance-id was None (not yet persisted)
 * 2nd reboot doesn't re-attach because current instance-id matches persisted.

uaclient/cli.py
```diff
@@ -372,7 +385,24 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
...
+            raise exceptions.AlreadyAttachedError(cfg)
+        # Unattached on non-Pro return UserFacing error msg details
+        raise e
+    current_iid = identity.get_instance_id()
+    if cfg.is_attached:
+        prev_iid = cfg.read_cache("instance-id")
+        if current_iid == prev_iid:
+            raise exceptions.AlreadyAttachedError(cfg)
+        print("Re-attaching Ubuntu Advantage subscription on new instance")
+        if _detach(cfg, assume_yes=True) == 1:
```
OddBloke: Given this is an exit code, we can future-proof ourselves fully with:

```suggestion
        if _detach(cfg, assume_yes=True) != 0:
```

PR #1021, opened 2020-04-01
focal: add integration tests for subcommands when attached to UA

patriciadomin:
focal: add integration tests for subcommands when attached to an UA subscription
adding scenarios for just one feature file: `attached_commands.feature`
This way we are going to be able to run only one specific Ubuntu series using behave tags option `--tags=series.focal`

Fixes: #1022


features/environment.py
```diff
@@ -98,36 +98,57 @@ def from_environ(cls) -> "UAClientBehaveConfig":
...
-
-    In this function, we launch a container, install ubuntu-advantage-tools and
-    then capture an image.  This image is then reused by each feature, reducing
-    test execution time.
-    """
+    """behave will invoke this before anything else happens."""
     userdata = context.config.userdata
     context.reuse_container = userdata.get("reuse_container")
+    context.series = {}
+    context.image_name = ""
```
blackboxsw, line 105: The way you have structured the test, context.series will overwrite image_name on each test run. This yields warnings during test runs because you keep overwriting that context.image_name value. 

I think it might be better if we drop context.image_name and rename context.series to ``context.series_image_names``.

```suggestion
    context.series_image_names = {}
```

features/environment.py
```diff
@@ -98,36 +98,57 @@ def from_environ(cls) -> "UAClientBehaveConfig":
...
+    """
+    In this function, we launch a container, install ubuntu-advantage-tools and
+    then capture an image. This image is then reused by each scenario, reducing
+    test execution time.
+    """
+    for tag in scenario.effective_tags:
+        parts = tag.split(".")
+        if parts[0] == "series":
+            series = parts[1]
+            if context.config.reuse_image is None:
```
blackboxsw, line 134: There is a problem now with reuse_image because any image you provide will be used for both trusty and focal. I'm not sure how best to solve this. Maybe if reuse_image is provided you have to specify only one series?

features/environment.py
```diff
@@ -141,33 +162,31 @@ def _capture_container_as_image(container_name: str, image_name: str) -> None:
...
-            print("Image cleanup disabled, not deleting:", context.image_name)
-
     if context.reuse_container:
         print(" Reusing the existent container: ", context.reuse_container)
     else:
         now = datetime.datetime.now()
-        context.image_name = "behave-image-" + now.strftime("%s%f")
-        build_container_name = "behave-image-build-" + now.strftime("%s%f")
-        launch_lxd_container(context, "ubuntu:trusty", build_container_name)
+        context.image_name = "behave-image-%s-" % series + now.strftime("%s%f")
```
blackboxsw, line 182: I think you need to avoid setting image_name here and just set context.series_image_names[series]
```suggestion
        context.series_image_names[series] = "behave-image-%s-" % series + now.strftime("%s%f")
```

features/environment.py
patriciadomin: Ok.

features/environment.py
patriciadomin: I've solved that. It will check the image series and will show it.
For instance, if you provide a trusty image and then run all tests (trusty and focal) it will use the trusty image for trusty tests and it will create a new one for focal.

features/environment.py
patriciadomin: Thanks. I changed it.

features/environment.py
```diff
@@ -141,33 +174,46 @@ def _capture_container_as_image(container_name: str, image_name: str) -> None:
...
-        context.image_name = "behave-image-" + now.strftime("%s%f")
-        build_container_name = "behave-image-build-" + now.strftime("%s%f")
-        launch_lxd_container(context, "ubuntu:trusty", build_container_name)
-        _install_uat_in_container(build_container_name)
-        _capture_container_as_image(build_container_name, context.image_name)
-        context.add_cleanup(image_cleanup)
+        container_series = lxc_get_series(context.reuse_container)
+        if container_series == series:
+            print(
+                "\n Reusing the existent container: ", context.reuse_container
```
blackboxsw, line 195: probably replace existent  with existing.

features/environment.py
```diff
@@ -141,33 +174,46 @@ def _capture_container_as_image(container_name: str, image_name: str) -> None:
...
+            )
+            print(
+                """\n You are providing a %s container.
+                Make sure you are running this series tests."""
```
blackboxsw, line 199: Can this message print "Make sure you provided -tags=series.%s" % series

features/util.py
```diff
@@ -100,8 +102,54 @@ def wait_for_boot(container_name: str) -> None:
...
+            subprocess.run(["lxc", "image", "show", name], stdout=fileoutput)
+        with open(image_output, "r") as fileoutput:
+            prop = yaml.safe_load(fileoutput)
+            print(" `lxc image show` output: ", prop)
+        fileoutput.close()
+        os.remove(image_output)
+        try:
+            series = prop["properties"]["release"]
+            print(
+                """\n You are providing a %s image.
```
blackboxsw, line 146: thoughout uaclient code, please try to use the string .format method  instead of %s. For example print("\n You are providing a  {series} image...".format(series=series))

features/util.py
```diff
@@ -100,8 +102,54 @@ def wait_for_boot(container_name: str) -> None:
...
+    if not image:
+        output = subprocess.check_output(
+            ["lxc", "config", "get", name, "image.release"],
+            universal_newlines=True,
+        )
+        series = output.rstrip()
+        return series
+    else:
+        image_output = "image_output.yaml"
+        with open(image_output, "w") as fileoutput:
```
blackboxsw, line 136: I don't think we need the redirect to fileoutput do we?
Can we instead use:
output = subprocess.check_output([lxc image, show, name]) 
image_config = yaml.safe_load(output)
try:
   return image_config["properties"]["release"]
except KeyError:
   print(...)
return None

features/util.py
```diff
@@ -100,8 +102,54 @@ def wait_for_boot(container_name: str) -> None:
...
+    :param name:
+        The name of the container or the image to check its series.
+    :param image:
+        If image==True will check image series
+        If image==False it will check container configuration to get series.
+
+    :return:
+        The series of the container or the image.
+       `None` if it could not detect it (
+           some images doesn't have this field into properties).
```
blackboxsw, line 124: minor typos
some images don't have this field in properties).

features/util.py
patriciadomin: ok. I've changed it

features/environment.py
patriciadomin: Ok. I've added `For instance: --tags=series.{series}` because it is not an obligation to run behave providing `--tags=...`. If you don't add this, it will run all tests.

features/util.py
patriciadomin: right. I've changed it. Thanks

features/util.py
patriciadomin: ouch. Sorry. Thanks

features/environment.py
```diff
@@ -98,36 +97,70 @@ def from_environ(cls) -> "UAClientBehaveConfig":
...
+    context.series_reuse_image = ""
     context.config = UAClientBehaveConfig.from_environ()
-    if context.config.reuse_image is None:
-        create_trusty_uat_lxd_image(context)
-    else:
-        context.image_name = context.config.reuse_image
+    if context.config.reuse_image:
+        series = lxc_get_series(context.config.reuse_image, image=True)
+        if series is not None:
+            context.series_reuse_image = series
```
blackboxsw, line 109: shouldn't we also set context.series_image_name[series] = context.config.reuse_image here too?

features/environment.py
```diff
@@ -98,36 +97,70 @@ def from_environ(cls) -> "UAClientBehaveConfig":
...
+        if parts[0] == "series":
+            series = parts[1]
+            if context.config.reuse_image is None:
+                if series not in context.series_image_name:
+                    create_uat_lxd_image(context, series)
+            else:
+                if series == context.series_reuse_image:
+                    context.series_image_name[
+                        series
+                    ] = context.config.reuse_image
```
blackboxsw, line 147: I don't think we should be doing this on every scenario, let's set this once up in before_all if we can.

features/environment.py
patriciadomin: No. It is in `before_scenario` because it knows which series is currently running, so it avoids, for instance: providing a trusty image and then running focal with that image (in case you do not provide --tags=series.trusty)

features/environment.py
patriciadomin: `before_all` hook doesn't have `feature` or `scenario` parameter, so it doesn't have any info about the tags. and `series` come from `scenario.effective_tags`.

features/environment.py
blackboxsw: Sorry, I must be missing something here. reuse_image is provided and will be either a focal or trusty image. Your lxc_get_series properly determines what series that reuse_image is running. If it is `trusty`, we only need to set this once context.series_image_name["trusty"] = contest.config.reuse_image.  We don't need to do this every before_scenario right because it was already set for trusty, regardless of whether the effective tag on a series is trusty or focal or xenial etc right?

blackboxsw: And if you are setting up series_image_name properly for the proper series that  reuse_image that was provided, then in your before_scenario you can simplify the logic to:

```
if series not in context.series_image_name:
    create_uat_lxd_image(context, series)
```

features/environment.py
```diff
@@ -141,33 +174,48 @@ def _capture_container_as_image(container_name: str, image_name: str) -> None:
...
+                "\n Reusing the existing container: ", context.reuse_container
+            )
+            print(
+                """\n You are providing a %s container.
+                Make sure you are running this series tests.
+                For instance: --tags=series.{series}""".format(
+                    series=series
+                )
+            )
+            return
```
blackboxsw, line 204: I think we don't need to call lxc_get_series here again right because we already set the context.series_reuse_image in before_all

```suggestion
    if context.reuse_container and context.series_reuse_image == series:
        print(
            "\n Reusing the existing container: ", context.reuse_container
        )
        print(
            """\n You are providing a %s container.
            Make sure you are running this series tests.
            For instance: --tags=series.{series}""".format(
                series=series
            )
        )
        return
```

patriciadomin: this is for reuse_container, not for reuse_image

blackboxsw: hahah, righto. Sorry about that. I crossed my wires there. Thanks!

features/environment.py
patriciadomin: Yes. Sorry. I thought you'd like to move all series info to before_all. I've changed it. Thanks

PR #1025, opened 2020-04-06
devel: add -D trace support for ipdb

blackboxsw:


features/environment.py
```diff
@@ -107,13 +110,18 @@ def before_all(context: Context) -> None:
     userdata = context.config.userdata
     context.reuse_container = userdata.get("reuse_container")
     context.config = UAClientBehaveConfig.from_environ()
+    context.pdb = userdata.get("pdb")
```
patriciadomin, line 113: Could you move it to the line below `context.reuse_container = `?  So, we have `userdata` config together

blackboxsw: Thank you @patriciadomin. Good suggestion. I have moved it up and force pushed changes

blackboxsw:
Rejecting this PR, because behave has a builtin param -w that can help here with development and automatically sets up console logging appropriately so that ipdb works

Development flow:
   * Decorate the scenario under test with a @wip tag
   * Add an ipdb.set_trace() at your desired location
   * run tox -e behave -- -w

PR #1032, opened 2020-04-17
fix reuse_container option

patriciadomin:
Currently it is not checking the correct series to run the tests in.
It fixes the issue: if you provide a container `-D reuse_container=cont_name`
it will run either series tests with the same container.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -1,6 +1,7 @@
 import datetime
 import os
 import subprocess
+import textwrap
 from typing import Dict, Union
 
 from behave.model import Feature, Scenario
@@ -99,9 +100,9 @@ class UAClientBehaveConfig:
 def before_all(context: Context) -> None:
     """behave will invoke this before anything else happens."""
     userdata = context.config.userdata
-    context.reuse_container = userdata.get("reuse_container")
     context.series_image_name = {}
     context.series_reuse_image = ""
+    context.reuse_container = {}
     context.config = UAClientBehaveConfig.from_environ()
     if context.config.reuse_image:
         series = lxc_get_series(context.config.reuse_image, image=True)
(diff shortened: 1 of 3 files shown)


patriciadomin:
Chad, I've created the reuse_container dictionary with a slight difference of what you asked for. I kept {series : cont_name} - the same way of I did for context.series_image_name dictionary.

features/environment.py
```diff
@@ -182,20 +193,12 @@ def create_uat_lxd_image(context: Context, series: str) -> None:
     """
 
     if context.reuse_container:
```
blackboxsw, line 195: I think we can consolidate this line with the if series in context.reuse_container with a single line now
```suggestion
    if series in context.reuse_container:
```

features/steps/steps.py
```diff
@@ -12,8 +12,8 @@
 
 @given("a `{series}` lxd container with ubuntu-advantage-tools installed")
 def given_a_lxd_container(context, series):
-    if context.reuse_container:
-        context.container_name = context.reuse_container
+    if context.reuse_container and series in context.reuse_container:
```
blackboxsw, line 15: We can make this simpler
```suggestion
    if series in context.reuse_container:
    # OR 
     if context.reuse_container.get(series):
```

features/environment.py
```diff
@@ -112,6 +112,17 @@ def before_all(context: Context) -> None:
             print(" Could not check image series. It will not be used. ")
             context.config.reuse_image = None
 
+    if userdata.get("reuse_container"):
+        series = lxc_get_series(userdata.get("reuse_container"))
+        context.reuse_container = {series: userdata.get("reuse_container")}
+        print(
+            """
+           You are providing a {series} container. Make sure you are running
+           this series tests. For instance: --tags=series.{series}""".format(
```
mitechie, line 121: Would it be better if this either straight errored if the series of the container don't match the tags or auto correct the tag and run such that what you're expecting is true and just warn the user that it adapted the behavior running?

patriciadomin: If the series doesn't match, it won't be used. We do not have control of the behave `--tags`. it came from CLI.For instance, if the user try to run with with `--tags=a_tag_not_in_the_feature_file` so it won't run any test (they will be skipped).

features/steps/steps.py
patriciadomin: Yes. addressed it, thanks

PR #1034, opened 2020-04-17
focal: add integration tests

patriciadomin:
focal: add integration tests for all .features' files.
This way we are going to be able to run only one specific Ubuntu series using behave tags option `--tags=series.focal`.

Closes: #1022

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -2,20 +2,26 @@
 Feature: Enable command behaviour when attached to an UA subscription
 
     @series.trusty
-    Scenario: Attached enable Livepatch service in a trusty lxd container
+    Scenario Outline:  Attached enable of non-container services in a trusty lxd container
         Given a `trusty` lxd container with ubuntu-advantage-tools installed
         When I attach contract_token with sudo
-        And I run `ua enable livepatch` as non-root
+        And I run `ua enable <service> <flag>` as non-root
         Then I will see the following on stderr:
             """
             This command must be run as root (try using sudo)
             """
-        When I run `ua enable livepatch` with sudo
+        When I run `ua enable <service> <flag>` with sudo
         Then I will see the following on stdout:
             """
             One moment, checking your subscription first
-            Cannot install Livepatch on a container
+            Cannot install <title> on a container
             """
+
+        Examples: Un-supported services in containers
+           | service      | title        | flag          |
+           | livepatch    | Livepatch    |               |
+           | fips         | FIPS         | --assume-yes  |
+           | fips-updates | FIPS Updates | --assume-yes  |
 
     @series.trusty
     Scenario: Attached enable Common Criteria service in a trusty lxd container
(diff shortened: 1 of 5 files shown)


features/attached_enable.feature
```diff
@@ -131,3 +131,135 @@ Feature: Enable command behaviour when attached to an UA subscription
...
+        Then I will see the following on stderr:
+            """
+            This command must be run as root (try using sudo)
+            """
+        When I run `ua enable livepatch` with sudo
+        Then I will see the following on stdout:
+            """
+            One moment, checking your subscription first
+            Cannot install Livepatch on a container
+            """
```
blackboxsw, line 149: To cut down on some of the test duplication, what do you think about consolidating livepatch/fips and fips-updates into a single Scenario Outline?

Maybe we can also consolidate the same non-container services tests for series.trusty too.

```suggestion
    @series.focal
    Scenario Outline: Attached enable of non-container services in a  container
        Given a `trusty` lxd container with ubuntu-advantage-tools installed
         when I attach contract_token with sudo
         and I run `ua enable <service>` as non-root
         then I will see the following on stderr:
            """
            This command must be run as root (try using sudo)
            """
         when I run `ua enable <service>` with sudo
         then I will see the following on stdout:
            """
            One moment, checking your subscription first
            Cannot install <title> on a container
            """

      Examples: Un-supported services in containers
        | service      | title        |
        | livepatch    | Livepatch    |
        | fips         | FIPS         |
        | fips-updates | FIPS Updates |
```

features/unattached_status.feature
```diff
@@ -28,6 +28,38 @@ Feature: Unattached status
             fips-updates  no         Uncertified security updates to FIPS modules
             livepatch     yes        Canonical Livepatch service
 
+            This machine is not attached to a UA subscription.
+            See https://ubuntu.com/advantage
+            """ 
+    
+    @series.focal
+    Scenario: Unattached status in a focal lxd container
```
blackboxsw, line 36: I wonder too if these tests could be consolidated in a scenario outline too?
What do you think?
```suggestion
    Scenario Outline: Unattached status in focal lxd container

        Given a `focal` lxd container with ubuntu-advantage-tools installed
￼        when I run `ua status` <userspec>
￼        then I will see the following on stdout:
￼            """
￼            SERVICE       AVAILABLE  DESCRIPTION
￼            cc-eal        no         Common Criteria EAL2 Provisioning Packages
￼            esm-apps      yes        UA Apps: Extended Security Maintenance
￼            esm-infra     yes        UA Infra: Extended Security Maintenance
￼            fips          no         NIST-certified FIPS modules
￼            fips-updates  no         Uncertified security updates to FIPS modules
￼            livepatch     yes        Canonical Livepatch service
￼
￼            This machine is not attached to a UA subscription.
￼            See https://ubuntu.com/advantage
￼            """

      Examples:
        | userspec     |
        | as non-root  |
        | with sudo    |
```

patriciadomin: I think we should not add the Examples here because it would create 2 container for each scenarios in this file, so instead of a total of 2 containers, we'd have 4 containers

features/attached_enable.feature
patriciadomin: I've addressed this one and will work on the WIP reduce duplication, PR 1047

features/unattached_status.feature
blackboxsw: Good point. +1 on separation unless we determine a set of non-destructive tests on an LXC which allow us to reuse an existing lxc image for multiple scenarios.

Issue #1031, opened 2020-04-16
FIPS should prompt for confirmation when enabling


blackboxsw:
Checklist;
  - [x] Allow `--assume-yes` param on the CLI to avoid such prompts
  - [ ] Prompt `ua enable fips`
    Installation of additional packages are required to make this system FIPS compliant
Are you sure? (y/n)
  - [x] Prompt `ua disable fips*`
     This will disable access to certified FIPS packages
Are you sure? (y/n)
  - [x] Prompt `ua enable fips-updates`
     This system will NOT be considered FIPS certified, but will include security and bug fixes to the FIPS packages.
Are you sure? (y/n)
  - [x] Message after `ua disable fips-*`
     A reboot is required to complete the disable process

PR #1043, opened 2020-04-27
fips: add console prompts during to fips and fips-updates

blackboxsw:
Add the following prompts to FIPS and FIPS with Updates:
  * ua enable fips-updates: 
```
This system will NOT be considered FIPS certified, but will include security    
and bug fixes to the FIPS packages.
Are you sure? (y/N)
```
   * ua disable fips and ua disable fips-updates:
```
This will disable access to certified FIPS packages.                            
Are you sure? (y/N)
```
This addresses parts of https://github.com/canonical/ubuntu-advantage-client/issues/1031
Expect one more branch which will add remaining prompts and messaging for post_disable messaging.

to test:
```
cat > ua-dev.yaml <<EOF
#cloud-config
packages:
 - git
 - make
runcmd:
 - git clone -b feature/fips-support https://github.com/blackboxsw/ubuntu-advantage-client.git /var/tmp/uac
 - cd /var/tmp/uac/
 - make deps
 - dpkg-buildpackage -us -uc
 - apt-get remove ubuntu-advantage-tools   # Issue: #1044
 - dpkg -i /var/tmp/ubuntu-advantage-tools_20.4_amd64.deb
 - ua attach <YOUR_TOKEN>
EOF

multipass launch daily:xenial --cloud-init ua-dev.yaml  -n dev-xx
# connect to the vm and test fips prompts
multipass connect dev-xx 
cloud-init status --wait --long
sudo ua status  
sudo ua enable fips-updates  # of fips (which doesn't have a prompt)
```

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -70,7 +70,11 @@ class FIPSEntitlement(FIPSCommonEntitlement):
         self
     ) -> "Dict[str, List[Union[str, Tuple[Callable, Dict]]]]":
         return {
-            "post_enable": [status.MESSAGE_ENABLE_REBOOT_REQUIRED],
+            "post_enable": [
+                status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+                    operation="install"
+                )
+            ],
             "pre_disable": [
                 (
                     util.prompt_for_confirmation,
@@ -78,6 +82,11 @@ class FIPSEntitlement(FIPSCommonEntitlement):
                         "assume_yes": self.assume_yes,
                         "msg": status.PROMPT_FIPS_PRE_DISABLE,
                     },
+                )
+            ],
+            "post_disable": [
+                status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+                    operation="disable operation"
                 )
             ],
         }
@@ -111,7 +120,11 @@ class FIPSUpdatesEntitlement(FIPSCommonEntitlement):
                     },
                 )
             ],
-            "post_enable": [status.MESSAGE_ENABLE_REBOOT_REQUIRED],
+            "post_enable": [
+                status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+                    operation="install"
+                )
+            ],
             "pre_disable": [
                 (
                     util.prompt_for_confirmation,
@@ -119,6 +132,11 @@ class FIPSUpdatesEntitlement(FIPSCommonEntitlement):
                         "assume_yes": self.assume_yes,
                         "msg": status.PROMPT_FIPS_PRE_DISABLE,
                     },
+                )
+            ],
+            "post_disable": [
+                status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+                    operation="disable operation"
                 )
             ],
         }
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -70,7 +70,16 @@ class FIPSEntitlement(FIPSCommonEntitlement):
         self
     ) -> "Dict[str, List[Union[str, Tuple[Callable, Dict]]]]":
         return {
-            "post_enable": ["A reboot is required to complete the install"]
+            "post_enable": [status.MESSAGE_ENABLE_REBOOT_REQUIRED],
+            "pre_disable": [
+                (
+                    util.prompt_for_confirmation,
+                    {
+                        "assume_yes": self.assume_yes,
+                        "msg": status.PROMPT_FIPS_PRE_DISABLE,
+                    },
+                )
+            ],
         }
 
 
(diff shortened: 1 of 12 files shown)


uaclient/entitlements/fips.py
```diff
@@ -93,8 +102,23 @@ def messaging(
...
+                (
+                    util.prompt_for_confirmation,
+                    {
+                        "msg": status.PROMPT_FIPS_UPDATES_PRE_ENABLE,
+                        "assume_yes": self.assume_yes,
+                    },
+                )
+            ],
+            "post_enable": [status.MESSAGE_ENABLE_REBOOT_REQUIRED],
+            "pre_disable": [
```
mitechie, line 115: so post-disable will be future coming?

uaclient/status.py
```diff
@@ -200,6 +215,7 @@ class UserFacingStatus(enum.Enum):
 Personal and community subscriptions are available at no charge
 See https://ubuntu.com/advantage"""
 MESSAGE_ENABLE_BY_DEFAULT_TMPL = "Enabling default service {name}"
+MESSAGE_ENABLE_REBOOT_REQUIRED = "A reboot is required to complete install"
```
mitechie, line 218: This is going to be true on disable as well. Just another variable or maybe make it "MESSAGE_REBOOT_REQUIRED"?

mitechie: Guess they're slightly different message strings aren't they. Ignore me.

uaclient/entitlements/fips.py
blackboxsw: @mitechie correct. post_disable will also have to handle the case where FIPS doesn't uninstall packages (which is unique to UA client defaults which attempt to remove any packages on disable)

uaclient/status.py
blackboxsw: yeah next branch will make this message a template where we'll print "install" vs "disable process"

OddBloke:
On Mon, Apr 27, 2020 at 12:08:13PM -0700, Rick Harding wrote:
> ty, curious how the typing works on trusty. Hadn't thought about it
> but the try/except around the import triggered my "huh?"

`typing` isn't available in Python 3.4, so we wrap the import in a
try/except, and any type annotations that use imported types need to be
in strings (which means they're lazily evaluated by type checkers, and
don't cause NameErrors in normal operation).

blackboxsw:
Thanks @patriciadomin for the thorough review. You're right I forgot to wire --assume-yes into UA Client API commandline. I've added that functionality and added unit tests. just validated on xenial with ua enable fips-updates --assume-yes

patriciadomin:
Thanks again. I see the `assume-yes` is now working fine, but I have these output to double check: 

Testing this on a Xenial VM (x86_64/amd64)
a) it allows me to enable fips and fips-updates (which should be mutually exclusive). Status after the required reboot: 
```
fips          yes       enabled   NIST-certified FIPS modules
fips-updates  yes       enabled   Uncertified security updates to FIPS modules
```

b) after a) (with both fips and fips-updates enabled) it shows livepatch status `n/a`, but there are 2 issues here: 
- it doesn't point this out ; 
- if after that you try to enable livepatch (it says it is already enabled:) 
```
$ sudo ua enable livepatch
Livepatch is already enabled.
See: sudo ua status
```
but the status shows: `livepatch     yes       n/a       Canonical Livepatch service`
an then it allows to disable livepatch `ua disable livepatch` keeping the same status `livepatch     yes       n/a       Canonical Livepatch service`

c) after I enable fips 'ua enable fips` I see livepatch is still enabled: (after the required reboot) 
```
fips          yes       enabled   NIST-certified FIPS modules
fips-updates  yes       disabled  Uncertified security updates to FIPS modules
livepatch     yes       enabled   Canonical Livepatch service
```

d) same as c) enabling fips-updates
```
fips          yes       disabled  NIST-certified FIPS modules
fips-updates  yes       enabled   Uncertified security updates to FIPS modules
livepatch     yes       enabled   Canonical Livepatch service
```

blackboxsw:
> Thanks again. I see the `assume-yes` is now working fine, 

Excellent , the other issue you mentioned are not intended to be resolved by this PR and they are currently unaddressed as you mentioned. It is only intended to fix components related to #1031 .
We will be addressing the mutual exclusivity of livepatch/fips and fips vs. fips-updates in later PRs that address separate user stories.

> Testing this on a Xenial VM (x86_64/amd64)
> a) it allows me to enable fips and fips-updates (which should be mutually exclusive). Status after the required reboot:


> 
> ```
> fips          yes       enabled   NIST-certified FIPS modules
> fips-updates  yes       enabled   Uncertified security updates to FIPS modules
> livepatch     yes       n/a       Canonical Livepatch service
> ```
> 
> b) after a) (with both fips and fips-updates enabled) it shows livepatch status `n/a`, but there are 2 issues here:

Once we have mutual exclusion livepatch/fips defined on xenial we won't get to the point where livepatch can be enabled at the same time as fips. We will force the user to disable livepatch first if they want to enable fips.

> c) after I enable fips 'ua enable fips` I see livepatch is still enabled: (after the required reboot)
...
> d) same as c) enabling fips-updates

Livepatch/fips* Mutual exclusion will come into play when we tackle user story [FIPS-UX-5&6](https://trello.com/c/cm1Eqf66/1453-fips-ux-5-6-prevent-livepatch-and-fips-enabled-at-the-same-time)

patriciadomin:
Great. Sorry, many thanks, +1 then

Issue #1027, opened 2020-04-14
Enabling fips does not put in place required GRUB configuration


kikoreis:
# BLOCKED awaiting and security team FIPS metapackage which will do this work as well as managing FIPS-specific packages.

If you manage to run `sudo ua enable fips` (by working around issue #1026), you'll have FIPS packages installed, but the required grub configuration to have the machine boot by default into FIPS mode will be missing.

Here's a poor man's version of what needs to be done prior to rebooting:
```
$ ROOTFS_UUID=$(blkid -t LABEL=cloudimg-rootfs -o export | grep ^UUID | cut -d= -f2)
$ cat | sudo tee /etc/default/grub.d/99-fips.cfg << __EOF__
GRUB_DEFAULT="1>gnulinux-4.4.0-1002-fips-advanced-$ROOTFS_UUID"
GRUB_CMDLINE_LINUX_DEFAULT="\$GRUB_CMDLINE_LINUX_DEFAULT fips=1" 
__EOF__
# Paste this too or just hit <enter>
$ sudo update-grub
```
(Note that SSH host keys may also need to be regenerated, but I am unable to verify if that's actually the case or not.)

kikoreis:
Not sure how robust the ROOTFS_UUID extraction is in my code above, as at least on my laptop there is no label for the rootfs; using /proc/cmdline may be more robust:
```
awk -v RS=" " -v FS="=" '/root=/ {print $3}' /proc/cmdline
```

powersj:
Hi @kikoreis thanks for the bug and trying it out.

The work to get FIPS fully integrated and supported is on deck. 

When we first took a look at adding FIPS support I was displeased with the fact that the client had to do so many changes to the system. As a result, the the security team is working on a metapackage and additional packaging to do these changes for the user. If any of these steps were to change they could issue a package update. I greatly prefer that to hard coding these types of changes in the client.

davecore82:
I'm still seeing this behaviour in ubuntu-advantage-tools version `25.0+202006031928-762~g70fe324~ubuntu18.04.1`.

I had to manually add the following line in /etc/default/grub.d/99-fips.cfg on an Ubuntu Pro 18.04 instance on AWS:

`GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT fips=1"`

And then run "sudo update-grub" to enable the change and reboot. After the reboot I see that FIPS is enabled:

```
[...]

PR #1045, opened 2020-04-27
Fips: post_disable messaging

blackboxsw:
Add post_disable hooks for repo-based entitlements.

Fixes: #1031 

Note: because fips isn't allowed to be disabled yet related to issue #1027  (waiting on fips-meta package) we won't be able to test this feature fully yet without hacking uaclient/entitlement/fips.py to implement the 'disable' method.


blackboxsw:
oops already pulled these changes into #1043

PR #1047, opened 2020-05-01
WIP integration tests: modification to reduce duplication

patriciadomin:
what do you think about adding the background scenario to our feature files?
I've added this just to attached_enable. 
Compared to PR1034 this file has 265 lines, this way it would have 166lines.
If it is ok, so I need to change all other feature files.

to test it: (I've added wip tag)
$ UACLIENT_BEHAVE_CONTRACT_TOKEN=$contract_token tox -e behave -- --tags=wip attached_enable.feature
(or remove the tags option to run all scenarios)

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -1,15 +1,21 @@
 @uses.config.contract_token
 Feature: Enable command behaviour when attached to an UA subscription
 
-    @series.trusty
-    Scenario: Attached enable Livepatch service in a trusty lxd container
-        Given a `trusty` lxd container with ubuntu-advantage-tools installed
+    Background: Attached enable a service in a trusty lxd container
+        Given a series lxd container with ubuntu-advantage-tools installed
+        |  series   |
+        |  trusty   |
+        |  focal    |
         When I attach contract_token with sudo
         And I run `ua enable livepatch` as non-root
         Then I will see the following on stderr:
             """
             This command must be run as root (try using sudo)
             """
+
+    @wip
+    @series.trusty
+    Scenario: Attached enable Livepatch service in a trusty lxd container
         When I run `ua enable livepatch` with sudo
         Then I will see the following on stdout:
             """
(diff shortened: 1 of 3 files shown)


features/attached_enable.feature
```diff
@@ -1,31 +1,31 @@
...
-    @series.trusty
-    Scenario: Attached enable Livepatch service in a trusty lxd container
-        Given a `trusty` lxd container with ubuntu-advantage-tools installed
+    Background: Attached enable a service in a trusty lxd container
+        Given a series lxd container with ubuntu-advantage-tools installed
+        |  series   |
+        |  trusty   |
+        |  focal    |
         When I attach contract_token with sudo
         And I run `ua enable livepatch` as non-root
```
blackboxsw, line 10: A background should be used for common setup for all scenarios, it should not also carry some test and verification clauses. so this "And" and "Then" should not be part of this setup, They should be a separate Scenario.

features/attached_enable.feature
```diff
@@ -1,31 +1,31 @@
 @uses.config.contract_token
 Feature: Enable command behaviour when attached to an UA subscription
 
-    @series.trusty
-    Scenario: Attached enable Livepatch service in a trusty lxd container
-        Given a `trusty` lxd container with ubuntu-advantage-tools installed
+    Background: Attached enable a service in a trusty lxd container
+        Given a series lxd container with ubuntu-advantage-tools installed
```
blackboxsw, line 5: This can be "Given a lxd container ...."

features/steps/steps.py
```diff
@@ -10,8 +10,9 @@
 CONTAINER_PREFIX = "behave-test-"
 
 
-@given("a `{series}` lxd container with ubuntu-advantage-tools installed")
-def given_a_lxd_container(context, series):
+@given("a series lxd container with ubuntu-advantage-tools installed")
```
blackboxsw, line 13: this can be @given("a lxd container...")

patriciadomin:
ok @blackboxsw thanks. I'll move back to 1034.

blackboxsw:
Thanks again for putting this together Particia. I'd like to leave it in closed state for reference as ultimately I'd like to use this common setup mechanism in BDD. But as we are moving CI I didnt want to clutter CI with new jobs that we aren't going to get to merge in the short term. Thanks again!

Issue #1041, opened 2020-04-23
integration tests: scenario to ua auto-attach 


patriciadomin:
We don't have this scenario test `ua auto-attach` . We need to add it .

PR #1052, opened 2020-05-08
integration tests: add auto-attach

patriciadomin:
add an auto-attach scenario

Fixes: #1041


patriciadomin:
Ok . Thanks. I've added it.

PR #1055, opened 2020-05-18
Update disable and enable commands

lucasmoura:
This PR creates an initial abstraction to allow multiple entitlements to be enables/disabled using the uaclient command. Right now, the idea is to try to enable/disable as much entitlements as we can. This means that we will not abort if we find an invalid entitlement or cannot disable/enable one of them.

To be more clear, this is an example of the output of the following command:

`ua disable ent1 ent2 ent3 --asume-yes true`

Where:

* ent1 is an invalid entitlement
* ent2 is not enabled
* ent3 is enabled

The command output will be the following:

```
ent2
ent2 is not currently enabled
See: sudo ua status

ent3
Updating packages lists

Disabled results:
Services not found: ent1
Services not disabled: ent2
Services disabled: ent3
```

What I am trying to achieve is a better way to give the user a report of what happened when he runs the command, but there is definitely room for improvement here.

This is a WIP because I have not created the BDD tests yet. I will start writing them, but I think I can still collect feedback for the code and the output message I am proposing.

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -103,15 +103,11 @@ def require_valid_entitlement_names(operation: str):
         def new_f(args, cfg, **kwargs):
             if hasattr(args, "names"):
                 names = args.names
-                entitlements_not_found = []
                 entitlements_found = []
                 for ent_name in names:
-                    if ent_name not in entitlements.ENTITLEMENT_CLASS_BY_NAME:
-                        entitlements_not_found.append(ent_name)
-                    else:
+                    if ent_name in entitlements.ENTITLEMENT_CLASS_BY_NAME:
                         entitlements_found.append(ent_name)
                 kwargs["entitlements_found"] = entitlements_found
-                kwargs["entitlements_not_found"] = entitlements_not_found
             return f(args, cfg, **kwargs)
 
         return new_f
@@ -164,7 +160,7 @@ def detach_parser(parser):
 
 def enable_parser(parser):
     """Build or extend an arg parser for enable subcommand."""
-    usage = USAGE_TMPL.format(name=NAME, command="enable") + " <name>"
+    usage = USAGE_TMPL.format(name=NAME, command="enable") + " []"
     parser.usage = usage
     parser.prog = "enable"
     parser._positionals.title = "Arguments"
@@ -188,7 +184,7 @@ def enable_parser(parser):
 
 def disable_parser(parser):
     """Build or extend an arg parser for disable subcommand."""
-    usage = USAGE_TMPL.format(name=NAME, command="disable") + " <name>"
+    usage = USAGE_TMPL.format(name=NAME, command="disable") + " []"
     parser.usage = usage
     parser.prog = "disable"
     parser._positionals.title = "Arguments"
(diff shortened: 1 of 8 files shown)


uaclient/cli.py
```diff
@@ -248,21 +255,108 @@ def status_parser(parser):
...
+    extra_params["assume_yes"] = args.assume_yes
+
+    return perform_actions_on_service_entitlements(
+        action_name="disabled",
+        action_fn=_perform_disable,
+        cfg=cfg,
+        entitlements_found=entitlements_found,
+        entitlements_not_found=entitlements_not_found,
+        **extra_params
+    )
```
blackboxsw, line 357: Passing the entitlements_found and entitlements_not_found down into a new function in order to run a for loop feels like a sign that we're pushing complexity into a new function that may not need to be there.

What do you think about just keeping the looping constructs here in action_disable?
we also could drop the entitlements_not_found variable from both the decorator and and this function as we can derive that locally with  **entitlements_not_found = list(set(args.names) - set(entitlements_found))**

```suggestion
    entitlements_not_found = set(args.names) - set(entitlements_found))
    for entitlement in entitlements_found:
        _perform_disable(entitlement_name, cfg, assume_yes=args.assume_yet)
    if entitlements_not_found:
       raise exceptions.UserFacingError(
￼             tmpl.format(operation="disabled, name=", ".join(entitlements_not_found))
￼       )
```

uaclient/status.py
```diff
@@ -234,6 +234,38 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_REFRESH_FAILURE = "Unable to refresh your subscription"
 
 
+def action_report(
```
blackboxsw, line 237: Adopting new UX generally would involve a bit of review with a broader audience(other teams), instead of adding a new function, we could adopt to maybe  single UserFacingError() with the related misses/invalids and point to running ua status as we did in the current master when someone specifies invalid service names

lucasmoura: No problem @blackboxsw, I will update the code

uaclient/cli.py
lucasmoura: That makes sense, I will update the code

lucasmoura:
@blackboxsw I have updated with the old message output. I have also created the behave tests for updated commands.  If you think I should create more behave tests for them, just let me know

uaclient/cli.py
```diff
@@ -91,23 +91,24 @@ def new_f(args, cfg):
     return new_f
 
 
-def require_valid_entitlement_name(operation: str):
-    """Decorator ensuring that args.name is a valid service.
+def require_valid_entitlement_names(operation: str):
+    """Decorator that verifies which of the args.names is a valid
+    service or not.
```
blackboxsw, line 96: It feels like keeping this decorator  is now overkill given that we aren't doing anything other than setting "entitlements_found".

Minimally we aren't using operation param anymore.

I think this would be better served as a standalone function like this get_valid_entitlement_names that enable and disable can call directly.
```suggestion
def get_valid_entitlement_names(                                                
    names: 'List', allow_beta: bool = False                                     
) -> 'List'                                                                     
    """Return a list of valid entitlement names.                                
                                                                                
    :param allow_beta: Consider beta entitlements as valid                      
                                                                                
    :return: List of valid entitlement names                                    
    """                                                                         
    entitlements_found = []                                                     
    for ent_name in names:                                                      
        ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME.get(ent_name)          
        if ent_cls:                                                             
            if not ent_cls.is_beta:                                             
                entitlements_found.append(ent_name)                             
            elif allow_beta:                                                    
                entitlements_found.append(ent_name)                             
    return entitlements_found   
```

uaclient/cli.py
```diff
@@ -248,21 +251,51 @@ def status_parser(parser):
...
-    ret = 0 if entitlement.disable() else 1
+    ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[entitlement_name]
+    entitlement = ent_cls(cfg, assume_yes=assume_yes)
+    ret = entitlement.disable()
     cfg.status()  # Update the status cache
     return ret
 
 
+@assert_root
+@require_valid_entitlement_names("disable")
```
blackboxsw, line 272: We can drop this decorate and call get_valid_entitlement_names (which should probably accept an allow_beta param.

uaclient/cli.py
```diff
@@ -248,21 +251,51 @@ def status_parser(parser):
...
+    """Perform the disable action on a list of entitlements.
+
+    @return: 0 on success, 1 otherwise
+    """
+    entitlements_found = kwargs.get("entitlements_found", [])
+    entitlements_not_found = sorted(set(args.names) - set(entitlements_found))
+    tmpl = ua_status.MESSAGE_INVALID_SERVICE_OP_FAILURE_TMPL
+    ret = True
+
+    for entitlement in entitlements_found:
```
blackboxsw, line 284: I think both action_disable and action_enable can look kindof like the following:
```suggestion

@assert_root
@assert_attached(ua_status.MESSAGE_ENABLE_FAILURE_UNATTACHED_TMPL)
def action_disable(args, cfg, **kwargs):
...
valid_entitlements = get_valid_entitlements(args.name, allow_beta=args.beta)
entitlement_not_found = sorted(set(args.names) - set(entitlements_found))
ret = True
for ent_name in valid_entitlements:
   ret &= _perform_disable(entitlement, cfg, assume_yes=args.assume_yes)

    if entitlements_not_found:
        raise exceptions.UserFacingError(
            tmpl.format(
                operation="enable", name=", ".join(entitlements_not_found)
            )
        )
 return 0 if ret else 1
```

lucasmoura:
@blackboxsw I have updated the logic as requested. I have also updated how the `enable` command works slightly, just to address how the `beta` functionality works.

Issue #853, opened 2019-10-16
Provide a way of integration testing a version of ubuntu-advantage-client other than the one in the daily PPA


OddBloke:
Currently, the integration tests install the latest package from the daily PPA into the test instances.

We really need to be able to test the version that's currently checked out (in addition to the PPA and/or archive versions). This will enable developers to use the tests to validate their work locally (or even to do BDD), as well as unblocking the use of the integration tests in CI (i.e. #857).

Issue #857, opened 2019-10-16
Add integration testing to Travis PR runs


OddBloke:
This would run the integration tests on proposed PRs. This depends on #856 (because it's essentially a super-set of the work there), and on #853 (because there's no point running the tests for a PR if we aren't testing the code in that PR).

PR #1057, opened 2020-05-20
add integration testing to Travis PR runs

patriciadomin:
This allows to run integration tests from the code that is currently checked out (via PR) 

modifications to get it:

I've modified travis.yml to have only the behave test - for dev/test purposes
as I was working on my own travis-ci

TRAVIS will: 
- create a .tar.gz with the source code from ubuntu-advantage-client dir
- copy this file to /tmp

Travis is running the test adding env_var UACLIENT_BEHAVE_BUILD_PR=1: 
` - sg lxd -c 'UACLIENT_BEHAVE_BUILD_PR=1 make test' `

I've added as a new option on UAClientBehaveConfig class:  `build_pr`
which we'll have via `context.config.build_pr`

When it is `True` (via travis or manually running)
the behave test won't install UA from the dailly PPA and it will: 

- create a new base image - (1st base_image) called: 
behave-image-pre-build-$series-...
 It is responsible for building the package from the source code

The new function: `lxc_build_deb` will
 - Push source code .tar.gz to the container.
 - extract it
 - Run the new script build-from-source.sh (will create the .deb)
 - Pull .deb from this container to travis-ci instance to output_deb_file (/tmp/ubuntu-advantage.deb)

- 2nd base image will be launched
 This base image do not have the durty build dependencies installed on 1st_base_img.
 - Push .deb to this instance
 - install it 

Later, as we already have for dailly PPA (publish instance as image)

This closes issues #853 and #857


features/environment.py
```diff
@@ -202,38 +207,70 @@ def create_uat_lxd_image(context: Context, series: str) -> None:
         )
         return
     now = datetime.datetime.now()
+    ubuntu_series = "ubuntu-daily:%s" % series
+
+    #if we are building the package from PR, we need a 2nd_base_image
+    #without the durty build dependencies from that 1st_base_image
```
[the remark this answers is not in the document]
patriciadomin: Thanks Lucas. Fixed it. That's the ideia: run automatically on travis-CI
I'm editing the comment - adding the related issues : 853 and 857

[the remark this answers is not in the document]
lucasmoura: Sorry, I think I have not described my idea properly. It would be to facilitate the process of running the behave test on the local machine, like you are creating a new behave tests and wants to see if it is working fine before opening the PR. I know your code already does that, but I was thinking on creating a script that the user will just run and it will run all of the behave tests for him, without worrying about creating a lxc container or creating the tarball with the current version of the code. But this is just an idea, I don't have any implementations idea for it yet

mitechie:
That sounds like a good make test target of some sort if at all possible.

tools/build-from-source.sh
```diff
@@ -0,0 +1,11 @@
...
+set -o xtrace
+apt-get update
+apt-get install make
+cd /tmp
+ls -lh
+lsb_release -a
+cd /tmp/ubuntu-advantage-client
+make deps
+dpkg-buildpackage -us -uc
+ls /tmp |grep ubuntu-advantage-tools
```
blackboxsw, line 11: I don't think we need this file as it doesn't really have purpose for typical development, we can create in on the fly in behave.

features/util.py
```diff
@@ -161,3 +161,33 @@ def lxc_get_series(name: str, image: bool = False):
...
+        [
+            "lxc",
+            "exec",
+            name,
+            "--",
+            "/tmp/ubuntu-advantage-client/tools/build-from-source.sh"
+        ],
+    )
+    print ("\n\nPull .deb from the instance to travis VM")
+    subprocess.run(["lxc", "file", "pull", name+'/tmp/ubuntu-advantage-tools_20.4_amd64.deb', "/tmp/"])
```
blackboxsw, line 193: 1. I think we can avoid having to add a file in  ./tools/build-from-source.sh  if we just do this inside behave.
2. we should use the lxc_exec function wherever we can instead lf subprocess.run directly
3. it might be worth adding an lxc_file_push function

```suggestion

    # create the script file
    with open('build-from-source.sh', 'w') as stream
        stream.write(textwrap.dedent(""\
            #!/bin/bash
            set -o xtrace
            apt-get update
            apt-get install make
            cd /tmp
            tar -zxvf *gz
            cd ubuntu-advantage-client
            make deps
            dpkg-buildpackage -us -uc
         """))
      
    lxc_file_push("build-from-source.sh", container_name + "/tmp/")
    print ('\n\n\n Run /tools/build-from-source.sh')
    lxc_exec(container_name, ["--", "bash", "/tmp/build-from-source.sh"])
    print ("\n\nPull .deb from the instance to travis VM")
    # rename the output deb to ubuntu-advantage.deb so we don't have to worry about version changes.
    subprocess.run(["lxc", "file", "pull", container_name+'/tmp/ubuntu-advantage-tools*.deb', output_deb_file])
```

features/environment.py
```diff
@@ -202,38 +207,70 @@ def create_uat_lxd_image(context: Context, series: str) -> None:
...
+    build_container_name = "behave-image-build-%s-" % series + now.strftime(
+            "%s%f"
+        )
+
+    launch_lxd_container(context, ubuntu_series, build_container_name)
+
+    #if build_pr it will install new built .deb
+    _install_uat_in_container(build_container_name, context.config.build_pr)
+
     context.series_image_name[
```
blackboxsw, line 232: Looking over this in more depth, it seems we actually don't have to create a separate build vm anyway. I hadn't thought enough about it, but we can re-start the build VM after we have already taken a snapshot then dirty the buildVM because we are no longer using it.

So I think this section could instead look like:
```
    context.series_image_name[                                                  
        series                                                                  
    ] = "behave-image-%s-" % series + now.strftime("%s%f")                      
    build_container_name = "behave-image-build-%s-" % series + now.strftime(    
        "%s%f"                                                                  
    )                                                                           
    ubuntu_series = "ubuntu-daily:%s" % series                                  
    launch_lxd_container(context, ubuntu_series, build_container_name)          
    _install_uat_in_container(build_container_name)                             
    _capture_container_as_image(                                                
        build_container_name, context.series_image_name[series]                 
    )                                                                           
    if context.config.build_pr:                                                 
        subprocess.run(["lxc", "start", build_container_name])                  
        lxc_build_deb(                                                          
            build_container_name, output_deb_path="/tmp/ubuntu-advantage.deb")  
        subprocess.run(["lxc", "stop", build_container_name])                  
```

.travis.yml
```diff
@@ -17,67 +17,17 @@ matrix:
       script:
           # bionic has lxd from deb installed, remove it first to avoid
           # confusion over versions
+          - echo $TRAVIS_BRANCH
+          - echo $TRAVIS_BUILD_DIR
+          - cd $TRAVIS_BUILD_DIR/..
+          - tar -zcvf pr_source.tar.gz ubuntu-advantage-client
```
blackboxsw, line 23: I'd like to do this work only if BUILD_PR is provided. so maybe we do it in python
we could do the following directly in features.util:lxc_push_source_pull_deb_pkg
```
if not os.path.exists('../pr_source.tar.gz'):
    orig_cwd = os.getcwd()
    os.chdir('..')
    util.subp(['tar', '-zxvf', 'pr_source.tar.gz', 'ubuntu-advantage-client'])
    os.chdir(orig_cwd)
```

.travis.yml
```diff
@@ -17,67 +17,17 @@ matrix:
...
-        PACKAGE_BUILD_SERIES=eoan
-      install:
-        - make travis-deb-install
-      script:
-        - make travis-deb-script
-    - env:
-        PACKAGE_BUILD_SERIES=focal
-      install:
-        - make travis-deb-install
-      script:
```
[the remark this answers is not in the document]
patriciadomin: Yes

tools/build-from-source.sh
patriciadomin: ok , deleted it

features/util.py
patriciadomin: @blackboxsw I cannot use the `*` here, according to this issue: https://github.com/lxc/lxd/issues/3523
it is not supported

patriciadomin: I changed its name inside the new generated script, so we do not need to worry about version changes

features/environment.py
patriciadomin: I kept both images because this way, the snapshot doesn't have the proper UA-tools.deb (from the source) installed in. So, let me know what do you think about the new modifications I've made.
If you'd like to keep as you suggested, we need the install the .deb for each test that we run the scenarios - when creating thebehave-test-*...

features/util.py
patriciadomin: Also, as lxc_exec is using --user 1000  we cannot run `lxc_exec(container_name, ["--", "bash", "/tmp/build-from-source.sh"])` or will get an error
```
+ apt-get update                                                
E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/                      
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
```
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?```
but just adding sudo does work, so moving to it, sorry.

features/environment.py
```diff
@@ -37,7 +37,7 @@ class UAClientBehaveConfig:
     # These variables are used in .from_environ() to convert the string
     # environment variable input to the appropriate Python types for use within
     # the test framework
-    boolean_options = ["image_clean", "destroy_instances"]
+    boolean_options = ["build_pr", "image_clean", "destroy_instances"]
```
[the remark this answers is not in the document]
patriciadomin: Ok I'm fixing it. Do you want to make your patch a suggestion, so I can commit it and you'll be co-author?

.travis.yml
patriciadomin: Ok, thanks but you are extracting it (zXvf) and this code is creating (-c) the tar.gz. 
I think your point it to only create the tar.gz when BUILD_PR=1 (as the next task will be - if it is a cron job, so BUILD_PR=0 and then run it from daily PPA and not from the source code.

I need to create the file via travis because after we run `make test` it is going to run inside tox virtualenv, so we'd need to send the $TRAVIS_BUILD_DIR to python/behave,  which I believe is not necessary. 
I think the new modification that I've made address what you'd like to have.
I'm adding a new condition. What do you think?

features/util.py
```diff
@@ -161,3 +163,57 @@ def lxc_get_series(name: str, image: bool = False):
...
+            #!/bin/bash
+            set -o xtrace
+            apt-get update
+            apt-get install make
+            cd /tmp
+            tar -zxvf *gz
+            cd ubuntu-advantage-client
+            make deps
+            dpkg-buildpackage -us -uc
+            cp /tmp/ubuntu-advantage-tools*.deb /tmp/ubuntu-advantage-tools.deb
```
lucasmoura, line 198: Maybe I am missing something here, but if the `dpkg-buildpackage -us -uc` produces more than one deb files, like:

* ubuntu-advantage-pro_25.0_amd64.deb
* ubuntu-advantage-tools_25.0_amd64.deb

Would the cp command work in that situation ?

patriciadomin: Yes. It does, but for PR, as we are running only on travis-CI we do not need the -pro.deb.
So, we need to install the -tools.deb.

lucasmoura: I completely missed that the pro image does not have the `tools` part in the name :facepalm: 
Sorry @patriciadomin

patriciadomin: no worries @lucasmoura

features/environment.py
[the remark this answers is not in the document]
blackboxsw: This should be your credit. patch as you see fit, we'll be squash merging anyway, so I'll make sure co-author isn't there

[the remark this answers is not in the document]
blackboxsw: no need to take my patch as suggestion. just apply to your code as you have

PR #1064, opened 2020-05-26
Use additionalPackages from service payload

lucasmoura:
Right now, ua-client is not using the `additionalPackages` field when enabling repo services, like `cc-eal` and `fips`. This PR adds the possibility for ua-client to start using these data when enabling repo services.

For fips, we are now using it understand which extra packages should be installed when enabling the service.

### DIFF
--- a/uaclient/entitlements/cis.py
+++ b/uaclient/entitlements/cis.py
@@ -8,7 +8,3 @@ class CISEntitlement(repo.RepoEntitlement):
     title = "CIS Audit"
     description = "Center for Internet Security Audit Tools"
     repo_key_file = "ubuntu-securitybenchmarks-keyring.gpg"
-
-    @property
-    def packages(self):
-        return ["ubuntu-cisbenchmark-16.04"] + super().packages
--- a/uaclient/entitlements/cis.py
+++ b/uaclient/entitlements/cis.py
@@ -8,4 +8,7 @@ class CISEntitlement(repo.RepoEntitlement):
     title = "CIS Audit"
     description = "Center for Internet Security Audit Tools"
     repo_key_file = "ubuntu-securitybenchmarks-keyring.gpg"
-    packages = ["ubuntu-cisbenchmark-16.04"]
+
+    @property
+    def packages(self):
+        return ["ubuntu-cisbenchmark-16.04"] + super().packages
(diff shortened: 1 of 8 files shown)


uaclient/entitlements/repo.py
```diff
@@ -58,7 +58,20 @@ def messaging(
     @property
     def packages(self) -> "List[str]":
         """debs to install on enablement"""
-        return []
+        packages = []
+
+        if self.cfg:
```
blackboxsw, line 63: We will never have an entitlement with an uninitialized self.cfg because [base.py:Entitlement.__init__](https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/entitlements/base.py#L68-L69) will set self.cfg = UAClientConfig() if the cfg param is not provided.

So this if can be dropped.

uaclient/entitlements/cis.py
```diff
@@ -8,4 +8,7 @@ class CISEntitlement(repo.RepoEntitlement):
     title = "CIS Audit"
     description = "Center for Internet Security Audit Tools"
     repo_key_file = "ubuntu-securitybenchmarks-keyring.gpg"
-    packages = ["ubuntu-cisbenchmark-16.04"]
+
+    @property
+    def packages(self):
+        return ["ubuntu-cisbenchmark-16.04"] + super().packages
```
blackboxsw, line 14: You should drop this method altogether as we should get it from the resource's additionalPackages directive (and looks like an update in the cis-audit resource definition changed the package name anyway to usg-*)  https://github.com/CanonicalLtd/ua-contracts/blob/develop/products/resource-cis.yaml

lucasmoura:
@blackboxsw done

uaclient/entitlements/fips.py
```diff
@@ -25,10 +27,16 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
     def packages(self) -> "List[str]":
         packages = list(self.fips_required_packages)
         installed_packages = apt.get_installed_packages()
-        for pkg_name, extra_pkgs in self.fips_packages.items():
+
+        pkg_groups = groupby(
+            filter(lambda pkg: pkg not in packages, super().packages),
```
blackboxsw, line 32: I like this use of groupby, we probably can drop the filter so we don't have to perform that work for the filter out the pkg from fips_required_packages if we just append that required pkg name  when we are processing the pkg_groups below.

lucasmoura: Just realized that I forgot to remove the fips_package dict :facepalm: 
Thanks for the review @blackboxsw. I have updated the code

PR #1067, opened 2020-05-28
Update package version on dev tools

lucasmoura:
When creating a new development environment for ua-client, there is an error when installing the deb package. Since we are now in version 25.0, we need to update this version on the `ua-dev-cloud-config` file as well.

### DIFF
--- a/tools/ua-dev-cloud-config.yaml
+++ b/tools/ua-dev-cloud-config.yaml
@@ -9,4 +9,4 @@ runcmd:
  - make deps
  - dpkg-buildpackage -us -uc
  - apt-get remove ubuntu-advantage-tools
- - dpkg -i /var/tmp/ubuntu-advantage-tools_20.4_amd64.deb
+ - dpkg -i /var/tmp/ubuntu-advantage-tools_25.0_amd64.deb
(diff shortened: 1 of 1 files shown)


tools/ua-dev-cloud-config.yaml
```diff
@@ -9,4 +9,4 @@ runcmd:
  - make deps
  - dpkg-buildpackage -us -uc
  - apt-get remove ubuntu-advantage-tools
- - dpkg -i /var/tmp/ubuntu-advantage-tools_20.4_amd64.deb
+ - dpkg -i /var/tmp/ubuntu-advantage-tools_25.0_amd64.deb
```
blackboxsw, line 12: Let's avoid pinning the dev config to a specific version of ua-t and use globbing '*' instead so we don't have to increment this in lockstep with each new version, *and* we can install ubuntu-advantage-pro*deb too

Also, I think we need an --assume-yes on the apt-get remove in case there are depedencies that will also get uninstalled (on groovy)

```suggestion
- apt-get remove ubuntu-advantage-tools --assume-yes
- dpkg -i /var/tmp/ubuntu-advantage-*deb
```

lucasmoura:
@blackboxsw done

PR #1069, opened 2020-05-28
Disable beta services to be show/enabled without flag

lucasmoura:
Right now, services that are in a beta stage, such as `fips` are shown normally when an user run `ua status`. We are now changing that behavior for `ua status` to only show production ready services. If the user wants to see all services, the command to run will be `ua status --beta`.

Also, if the user tries to enable a beta service using: `ua enable fips`, ua-client will raise an error saying that the service could not be found. To enable beta services, users will need to run the following command: `ua enable fips --beta`

Fixes: #1079

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -244,6 +244,11 @@ def status_parser(parser):
             )
         ),
     )
+    parser.add_argument(
+        "--show-beta",
+        action="store_true",
+        help="Allow the visualization of beta services",
+    )
     parser._optionals.title = "Flags"
     return parser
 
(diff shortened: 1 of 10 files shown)


uaclient/cli.py
```diff
@@ -268,7 +276,8 @@ def _perform_enable(
     cfg: config.UAConfig,
     *,
     assume_yes: bool = False,
-    silent_if_inapplicable: bool = False
+    silent_if_inapplicable: bool = False,
+    allow_beta: bool = False
```
blackboxsw, line 280: Please add a docstring reference to this param

uaclient/entitlements/fips.py
```diff
@@ -20,6 +20,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         "strongswan": {"strongswan-hmac"},
     }  # type: Dict[str, Set[str]]
     repo_key_file = "ubuntu-advantage-fips.gpg"  # Same for fips & fips-updates
+    is_beta = True
```
blackboxsw, line 23: Please add is_beta = True to uaclient/entitlements/cis.py and cc.py. too

lucasmoura:
@blackboxsw Done, I have updated the code with your suggestions

Issue #1077, opened 2020-06-02
ua auto-attach doesn't work on AWS GovCloud


cmars:
UA client should auto-attach for cloud "aws-gov", same as if it was "aws". `POST /v1/clouds/aws/token` should work the same; contract service is able to verify the signed IID from these regions.

I think the issue is that UA client is getting "aws-gov" from cloud-init as the cloud name, and failing an exact match for "aws". Might want to do a prefix match instead so this isn't an issue for other special AWS regions in the future?

```
... - cli.py:(594) [ERROR]: Auto-attach image support is not available on aws-gov
See: https://ubuntu.com/advantage
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 576, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 616, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 89, in new_f
    return f(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 52, in new_f
    return f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 387, in action_auto_attach
    token = _get_contract_token_from_cloud_identity(cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 366, in _get_contract_token_from_cloud_identity
    cloud_type=cloud_type
uaclient.exceptions.NonAutoAttachImageError: Auto-attach image support is not available on aws-gov
See: https://ubuntu.com/advantage
```

jaredledvina:
Folks here may be interested in following the AWS ARN format from https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax which has another partition for the china region.

mitechie:
Agree, we can work to make sure we don't just fix the gov issue but have it work in a more general sense. Given that this is image specific though, we should double check which of these we publish the pro images in just to be sure.

rcj4747:
> Agree, we can work to make sure we don't just fix the gov issue but have it work in a more general sense. Given that this is image specific though, we should double check which of these we publish the pro images in just to be sure.

[...]

PR #1078, opened 2020-06-02
Add AWS cloud partitions (#1077)

rcj4747:
China and US GovCloud are separate partitions of AWS' cloud.  The `cloud-id`
tool used to identify the clouds and map them to their auto-attach logic only
works in the primary aws partition.  This patch adds support for China and GovCloud.


rcj4747:
I made this change manually on an Ubuntu Pro image registered in us-gov-west-1a and rebooted.  The instance auto-attached to UA without issue.

mitechie:
Awesome thanks for the verification Rob.

rcj4747:
The travis failure is an ephemeral issue with the Ubuntu archive, not a testcase issue.  Someone with access should trigger the travis check again.

Issue #1084, opened 2020-06-05
Client blocked on config file difference prompt


jaredledvina:
Using the Ubuntu 18.04 LTS Pro, when attached to enable fips, the installer get's hung up on a prompt:
```
$ date && sudo ua --debug enable fips
Fri Jun  5 14:18:45 UTC 2020
DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-token.json
One moment, checking your subscription first                                                                                                                                                                                           DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-access-cc-eal.json
DEBUG: Reading file: /etc/os-release
DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-access-esm-apps
DEBUG: Reading file: /etc/os-release
DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-access-esm-apps-python
DEBUG: Reading file: /etc/os-release
DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-access-esm-infra.json
DEBUG: Reading file: /etc/os-release
DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-access-fips.json
DEBUG: Reading file: /etc/os-release
DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-access-fips-updates.json
DEBUG: Reading file: /etc/os-release
DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-access-livepatch.json
DEBUG: Reading file: /etc/os-release
DEBUG: Reading file: /var/lib/ubuntu-advantage/private/machine-access-support.json
DEBUG: Reading file: /etc/os-release
DEBUG: Reading file: /etc/machine-id
DEBUG: Reading file: /etc/os-release
.................................................................................
Updating package lists
DEBUG: Ran cmd: apt-get update, rc: 0 stderr: b''
Installing FIPS packages
```

From `/var/log/apt/term.log` I can see that it's stuck prompting for a diff on `/etc/ssh/sshd_config`:
```
Log started: 2020-06-05  14:22:34
(Reading database ... 84465 files and directories currently installed.)
Preparing to unpack .../00-libssl1.1_1.1.1-1ubuntu2.fips.2.1~18.04.3.1_amd64.deb ...
[...]

PR #1086, opened 2020-06-05
Ensure we don't prompt for configuration file diff

jaredledvina:
The client can become blocked if APT prompts for a configuration file difference for the user to select. 
This commit ensures we pass through force-confdef and force-confold to avoid such prompts

Closes https://github.com/canonical/ubuntu-advantage-client/issues/1084

The logic is partially copied from how Ansible currently avoids this situation: https://github.com/ansible/ansible/blob/062e780a68f9acd2ee6f824f252458b8a0351f24/lib/ansible/modules/apt.py#L517-L523

### DIFF
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -28,7 +28,6 @@ from uaclient.status import ApplicationStatus
 
 APT_DISABLED_PIN = "-32768"
 
-
 class RepoEntitlement(base.UAEntitlement):
 
     repo_list_file_tmpl = "/etc/apt/sources.list.d/ubuntu-{name}-{series}.list"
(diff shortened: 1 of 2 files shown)


uaclient/entitlements/repo.py
```diff
@@ -77,6 +77,17 @@ def packages(self) -> "List[str]":
...
+    @property
+    def dpkg_options(self) -> "List[str]":
+        """
+        Returns a list of dpkg options to pass to apt
+        """
+        options_list = ['force-confdef', 'force-confold']
+        dpkg_options = []
+        for option in options_list:
+            dpkg_options.append('-o="Dpkg::Options::=--{}"'.format(option))
+        return dpkg_options
```
blackboxsw, line 89: This looks like the right solution thank you much @jaredledvina for proposing it.

Can we adapt the method here a bit for readability and intent. What do you think of the following?
```suggestion
    @property
    def dpkg_options(self) -> "List[str]":
        """
        Return a list of dpkg options to avoid interactive prompts

        Automatically preserve local configuration files if changes detected.
        https://debian-handbook.info/browse/stable/sect.package-meta-information.html#sidebar.questions-conffiles
        """
        return [
            '-o="Dpkg::Options::=--force-confdef"',
            '-o="Dpkg::Options::=--force-confold"',
        ]
```

jaredledvina:
Superseded by https://github.com/canonical/ubuntu-advantage-client/pull/1096 which is a nicer implementation of what I was trying to do here.

Issue #1089, opened 2020-06-09
Add lxd vm support where available


blackboxsw:
Ubuntu Xenial or later have support for launching lxc launch --vm  in order to launch kvm-based instances instead of lxc containers. Use this support to drive tests for livepatch fips and other services which cannot be installed on containers.

PR #1090, opened 2020-06-09
tests: add lxd vm support for behave tests.

blackboxsw:
Allow the ability to drive behave tests on either lxd containers or kvm-based machines.

This toggle is set via environement variabe UACLIENT_BEHAVE_MACHINE_TYPE which can
be either `lxd.container` or `lxd.vm`. In the future I'd expect we may grow `machine_type` to include Ubuntu PRO instances like `aws.pro` or `azure.pro` etc.

The following work was necessary to support this:
 - Add feature.util.lxc_create_vm_profile which creates a profile which attaches a cloud-init:config
   drive to a kvm instance that is booting
 - features.util.wait_for_boot() retry changes
      - vms take a while to boot, so change our retry logic on Xenial and later to block on
         `cloud-init status --wait` until cloud-init is done
      - Adjust retry backoff seconds because lxd VM instances take about 15 seconds to be able to connect to the lxd agent. Otherwise we see "Error: Failed to connect to lxd-agent"
 - Adding machine_type to UAClientBehaveConfig in features/environment.py
 - Allowing for @uses.config.machine_type.(lxd.container|lxd.vm) to declare a specific scenario
    requires a certain machine_type to run
 - consolidate _should_skip_tags function for use at both before_feature and before_scenario
   in order to skip invalid @uses conditions
 - skip any @series.trusty scenarios if machine_type is lxd.vm
 - change steps.steps.given_a_lxd_container -> given_a_machine

To test:
```
Grab your contract token from auth.contracts.canonical.com

cat >test.patch <<EOF

diff --git a/features/attached_enable.feature b/features/attached_enable.feature
index 35d52b8..3fb80a8 100644                                                   
--- a/features/attached_enable.feature                                          
+++ b/features/attached_enable.feature                                          
@@ -135,7 +135,6 @@ Feature: Enable command behaviour when attached to an UA subscription
             FIPS Updates is not available for Ubuntu 20.04 LTS (Focal Fossa).  
             """                                                                
                                                                                
-    @wip                                                                       
     @series.bionic                                                             
     @uses.config.machine_type.lxd.vm                                           
     Scenario: Attached enable of vm-based services in a bionic lxd vm   
EOF


# to run a bionic vm test
make clean
cd ..
tar -zcvf /tmp/pr_source.tar.gz ubuntu-advantage-client

UACLIENT_BEHAVE_CONTRACT_TOKEN=<YOUR_TOKEN> UACLIENT_BEHAVE_MACHINE_TYPE=lxd.vm tox -e behave -- -w -k

# Check your lxc profile
lxc profile show behave-vm-bionic

```

Fixes: #1089

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -3,7 +3,7 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.trusty
     Scenario Outline:  Attached enable of non-container services in a trusty lxd container
-        Given a `trusty` lxd container with ubuntu-advantage-tools installed
+        Given a `trusty` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
         And I run `ua enable <service> <flag>` as non-root
         Then I will see the following on stderr:
@@ -25,7 +25,7 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.trusty
     Scenario Outline:  Attached enable of non-container beta services in a trusty lxd container
-        Given a `trusty` lxd container with ubuntu-advantage-tools installed
+        Given a `trusty` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
         And I run `ua enable <service> <flag>` as non-root
         Then I will see the following on stderr:
@@ -50,7 +50,7 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.trusty
     Scenario: Attached enable Common Criteria service in a trusty lxd container
-        Given a `trusty` lxd container with ubuntu-advantage-tools installed
+        Given a `trusty` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
         And I run `ua enable cc-eal` as non-root
         Then I will see the following on stderr:
@@ -66,7 +66,7 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.trusty
     Scenario Outline: Attached enable not entitled service in a trusty lxd container
-        Given a `trusty` lxd container with ubuntu-advantage-tools installed
+        Given a `trusty` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
         And I run `ua enable <service>` as non-root
         Then I will see the following on stderr:
@@ -88,7 +88,7 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.trusty
     Scenario: Attached enable of an unknown service in a trusty lxd container
-        Given a `trusty` lxd container with ubuntu-advantage-tools installed
+        Given a `trusty` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
         And I run `ua enable foobar` as non-root
         Then I will see the following on stderr:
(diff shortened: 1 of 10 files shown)


features/attached_enable.feature
```diff
@@ -119,9 +119,32 @@ Feature: Enable command behaviour when attached to an UA subscription
...
 
+    @wip
     @series.focal
-    Scenario Outline: Attached enable of non-container services in a focal lxd container
-        Given a `focal` lxd container with ubuntu-advantage-tools installed
+    @uses.config.machine_type.lxd.vm
+    Scenario: Attached enable of vm-based services in a focal lxd vm
+        Given a `focal` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `ua enable <service> <flag>` as non-root
```
lucasmoura, line 128: I think it is missing the examples table here. When running the tests, I get the following error:

```
Creating behave-test-vm-focal-1593107907758756
Starting behave-test-vm-focal-1593107907758756
Retrying on unexpected cloud-init status stderr:  Error: Failed to connect to lxd-agent
Retrying on unexpected cloud-init status stderr:  Error: Instance is not running
Retrying on unexpected cloud-init status stderr:  Error: Failed to connect to lxd-agent
    Given a `focal` machine with ubuntu-advantage-tools installed ... passed in 146.671s
    When I attach `contract_token` with sudo ... passed in 255.276s
    And I run `ua enable <service> <flag>` as non-root ... passed in 0.755s
    Then I will see the following on stderr ... failed in 0.000s
Assertion Failed: 
Expected: 'This command must be run as root (try using sudo)'
     but: was 'usage: ua [command] [flags]\nua: error: unrecognized arguments: <flag>'

      """
      This command must be run as root (try using sudo)
      """
```

features/steps/steps.py
```diff
@@ -12,17 +12,23 @@
 CONTAINER_PREFIX = "behave-test-"
 
 
-@given("a `{series}` lxd container with ubuntu-advantage-tools installed")
-def given_a_lxd_container(context, series):
+@given("a `{series}` machine with ubuntu-advantage-tools installed")
+def given_a_machine(context, series):
     if series in context.reuse_container:
```
lucasmoura, line 17: I don't know if we should do this, but maybe we should check if the machine the user is trying to reuse fits the testing it is trying to use. For example, we try to reuse a container on a vm test. The test will fail, but we could alert sooner that the types are not matched and just skip or abort the test.

blackboxsw: @lucasmoura you are correct that our tests do that, and they currently do. In before_scenario and now in before_feature we check via a call to features/environment:_should_skip_tags which checks whether a conflicting ```@uses or @series``` tag is declared

features/attached_enable.feature
blackboxsw: @lucasmoura sorry I forgot to push fixes for that.

blackboxsw:
> I was able to run the bionic vm test and could not spot anything wrong in the code. However, I just have one concern related to travis. When running the VM test, we need to download the vm to perform that test. On travis, will have to perform this on every run, right ?
> 
> My concern here is related to the time needed to run the tests, since we will need to download both bionic and xenial. Although the tests that will use the vms will probably not be evaluated in travis because of #1098, maybe we should verify how travis will behave with those tests.
> 

I may be wrong, but don't believe our lxd.vm support will work in the travis build environemnt, we'll likely have to run MACHINE_TYPE=lxd.vm via jenkins on hardware that we own.  As this PR currently stands, the test framework still defaults to lxd.container machine_type so it will skip any tests decorated with `@uses.config.machine_type.lxd.vm` so this shouldn't affect travis. The only thing travis should be running is lxc.container based machine_type tests (which includes any tests that are undecorated with any `@uses.config.machine_type.*`


> Also, should we wait for xenial test to be ready to land this PR ?

I can add a xenial test at the moment, it should work fine (and will look exactly like the bionic test).

Issue #930, opened 2019-11-21
Extend `hook.cc` to report on available esm-apps upgrades


OddBloke:
In addition to the esm-infra upgrades it already advertises.

(Note that for xenial+ we shouldn't be reporting these before `enable` as we do on trusty, but that is handled by not installing anything in `/etc/apt` until `enable` happens; the hook runs before `enable` but will never find anything.)

OddBloke:
(Due to this issue, https://github.com/CanonicalLtd/ubuntu-advantage-client/pull/931 temporarily backed out installation of the hook.)

OddBloke:
This work will be undertaken by the Foundations team.

PR #1092, opened 2020-06-17
Support ESM Apps

murraybd:
Add support for ESM Apps in addition to ESM Infrastructure.

I removed the 'apt-get' assertions because the hook doesn't actually run if you use 'apt-get' and I found the fact that 'apt-get' was referenced confusing. I'd actually prefer to remove it entirely.

Fixes: #930


mitechie:
@blackboxsw can you please peek at this when you get a sec?

apt-hook/hook.cc
```diff
@@ -123,9 +125,19 @@ static void check_esm_upgrade(pkgCache::PkgIterator pkg, pkgPolicy *policy, resu
...
 	    else
-	       res.enabled_esms++;
+	       res.enabled_esms_i++;
+
+	    return;
+	 }
+
+	 if (pf.File().Archive() != 0 && pf.File().Origin() == std::string("UbuntuESMApps"))
+	 {
+	    if (policy->GetPriority(pf.File()) == -32768)
```
blackboxsw, line 137: @murraybd Is this disabled check and counter needed?

 I was under the impression that only "ESM Infra" apt sources (on trusty only) would be installed without proper authentication which results in the -32768 value. In my mind, no other Ubuntu release would need this logic, and specifically not UbuntuESMApps.   The only time we would use this is if UAClient decides to install apt sources without credentials for ESM Apps to get 'visibility' to those potential packages.


That said, it probably doesn't hurt much to have this check and count in here, that way we can support advertising ESM Apps on disabled machines if we need to.

apt-hook/hook.cc
```diff
@@ -187,15 +199,18 @@ int main(int argc, char *argv[])
    //    assert(ppid == getppid_of("self"));
    // }
    assert(cmdline_eligible(make_cmdline("apt\0update\0")));
-   assert(cmdline_eligible(make_cmdline("apt-get\0update\0")));
-   assert(!cmdline_eligible(make_cmdline("apt-get\0install\0")));
-   assert(!cmdline_eligible(make_cmdline("apt\0install\0")));
+   //assert(cmdline_eligible(make_cmdline("apt-get\0update\0")));
```
blackboxsw, line 202: If the asserts don't apply, let's gut them instead of commenting them out

apt-hook/hook.cc
blackboxsw: Here's the SPEC reference that says we don't pre-advertise UA Apps or UA infra packages on Xenial+ https://docs.google.com/document/d/19_bTeagOuY9Wd727vtLskP1rkY5FwzaFXi3b_R5qa7g/edit#

blackboxsw:
@murraybd Sorry for the delay here I wanted to give Rick a change to weigh in on the direction of this messaging. 

In motd, we will be working to consolidate all ESM Infra and Apps messaging and outstanding package counts to a single message instead of differentiating Apps from Infra packages so I didn't want to go further down a specialized messaging route if we wanted to move towards consolidation.

Regardless of our 'future' direction for motd messaging, I think your changeset  aligns with what current UAClient and update-notifier currently does in motd. I think when we have time to address consolidation of messaging in update-notifier, then we can also address that in UAClient apt hook as well.


I think I'd like to see us either drop your magic -32768 counter work for Apps or at least document in comments next to those magic apt number checks that "// Xenial and later should not be advertizing unauthenticated ESM Apps and Infra apt repos".

murraybd:
I've made the changes you requested.

PR #1096, opened 2020-06-22
fips-only: force apt noninteractive prompts during package installs

blackboxsw:
Since we currently expect FIPS* package installs to run into local config customization prompts on cloud-images, emit the appropriate apt commandline options to avoid prompting to the console, and preserve the locally modified config files for openssh-server and /boot/grub/menu.lst.

Since we hope to address this config file difference properly in the openssh-server package and fips-metapackage, add a RELEASE_BLOCKER comment to fips.py to ensure we address this with the security and cloud images team before fully supporting FIPS.

Inspired by  jaredledvina's PR #1086
Fixes: #1084

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -84,14 +84,25 @@ def assert_valid_apt_credentials(repo_url, username, password):
         )
 
 
-def run_apt_command(cmd, error_msg) -> str:
+def run_apt_command(
+    cmd: 'List[str]',
+    error_msg: str,
+    env: "Optional[Dict[str, str]]" = None,
+) -> str:
     """Run an apt command, retrying upon failure APT_RETRIES times.
+
+    :param cmd: List containing the apt command to run, passed to subp.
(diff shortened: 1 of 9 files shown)


lucasmoura:
Build is failing because of this:
```
uaclient/apt.py:90:10: F821 undefined name 'Optional'
uaclient/apt.py:90:10: F821 undefined name 'Dict'
```

blackboxsw:
We currently expect FIPS* package installs to run into local config customization prompts on cloud-images, emit the appropriate apt commandline options to avoid prompting to the console, and preserve the locally modified config files for openssh-server and /boot/grub/menu.lst.

There is an expectation that we can address this config file difference properly in the openssh-server package and fips-metapackage before official FIPS support in this client.
Add a RELEASE_BLOCKER comment to fips.py until we address these configuration differences between security and cloud images packaging of openssh-serverteam before fully supporting FIPS.

Inspired by jaredledvina's PR #1086
Fixes: #1084

Issue #1095, opened 2020-06-19
esm: debian.postinst automatically unconfigures esm-infra on all package upgrades 


blackboxsw:
# description update

ubuntu-advantage-tools postinst automatically calls `unconfigure_esm` on any package upgrade regardless of whether esm-infra is enabled. This results in the removal of the keyring /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg.

To reproduce:

- enable esm-infra on a system
- sudo add-apt-repository ppa:canonical-server/ua-client-daily
- sudo apt-get update
- sudo apt-get install ubuntu-advantage-tools
- sudo apt-get update # see warnings because /etc/apt/trusted.gpg.d/*infra* is now removed.
```
ubuntu@ip-172-31-34-86:~$ dpkg-query --show ubuntu-advantage-tools 
ubuntu-advantage-tools	24.2~16.04.1
ubuntu@ip-172-31-34-86:~$ ua status 
SERVICE       ENTITLED  STATUS    DESCRIPTION
cc-eal        yes       disabled  Common Criteria EAL2 Provisioning Packages
cis-audit     no        —         Center for Internet Security Audit Tools
esm-apps      yes       enabled   UA Apps: Extended Security Maintenance
esm-infra     yes       enabled   UA Infra: Extended Security Maintenance
fips          yes       disabled  NIST-certified FIPS modules
fips-updates  yes       disabled  Uncertified security updates to FIPS modules
livepatch     yes       enabled   Canonical Livepatch service

Enable services with: ua enable <service>

                Account: 329688318733
           Subscription: 329688318733
            Valid until: n/a
Technical support level: essential
ubuntu@ip-172-31-34-86:~$ dpkg-query --show ubuntu-advantage-tools 
ubuntu-advantage-tools	24.2~16.04.1
ubuntu@ip-172-31-34-86:~$ ls /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-*
/etc/apt/trusted.gpg.d/ubuntu-advantage-esm-apps.gpg  /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
ubuntu@ip-172-31-34-86:~$ sudo apt-get install ubuntu-advantage-tools
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be upgraded:
  ubuntu-advantage-tools
[...]

PR #1097, opened 2020-06-23
Update postinst to check if esm is already enabled

lucasmoura:
As stated on #1095, when we upgrade ua-client with esm-infra enabled, we end up removing the `ubuntu-advantage-esm-infra-trusty.gpg` file, which puts the system in an incorrect state, since ua-client will state that `esm-infra` will be enabled, but we cannot update its packages.

To fix that, we have updated the `postinst` script to check if `esm-infra` is already enable and only if it is not we let the unconfigure step run.

To test that feature, the following script was used:

```
#!/bin/sh

UA_ACCESS_KEY="UNSET"

check_esm_key() {
    name=$1
    lxc exec $name -- sudo apt update
    lxc exec $name -- ls /etc/apt/trusted.gpg.d/
}

replicate_issue() {
    cat > config_error.yaml <<EOF
#cloud-config
# Setup an ubuntu-advantage-tools development environment with cloud-init
packages:
 - git
 - make
runcmd:
 - sudo ua attach $UA_ACCESS_KEY
 - sudo ua enable esm-infra
 - git clone https://github.com/lucasmoura/ubuntu-advantage-client.git /var/tmp/uac
 - cd /var/tmp/uac/
 - git fetch
 - make deps
 - dpkg-buildpackage -us -uc
 - apt-get remove ubuntu-advantage-tools --assume-yes
 - dpkg -i /var/tmp/ubuntu-advantage-tools_25.0_amd64.deb
EOF

 name=dev-f-bug
 lxc delete $name --force > /dev/null
 lxc launch ubuntu-daily:focal $name -c user.user-data="$(cat config_error.yaml)"
 lxc exec $name -- cloud-init status --wait
 lxc exec $name -- cloud-init query userdata
 check_esm_key $name
 rm config_error.yaml
}

verify_issue() {
    cat > config_fix.yaml <<EOF
#cloud-config
# Setup an ubuntu-advantage-tools development environment with cloud-init
packages:
 - git
 - make
runcmd:
 - sudo ua attach $UA_ACCESS_KEY
 - sudo ua enable esm-infra
 - git clone https://github.com/lucasmoura/ubuntu-advantage-client.git /var/tmp/uac
 - cd /var/tmp/uac/
 - git fetch
 - git checkout fix-postinst-for-esm 
 - make deps
 - dpkg-buildpackage -us -uc
 - apt-get remove ubuntu-advantage-tools --assume-yes
 - dpkg -i /var/tmp/ubuntu-advantage-tools_25.0_amd64.deb
EOF

 name=dev-f-fix
 lxc delete $name --force > /dev/null
 lxc launch ubuntu-daily:focal $name -c user.user-data="$(cat config_fix.yaml)"
 lxc exec $name -- cloud-init status --wait
 lxc exec $name -- cloud-init query userdata
 check_esm_key $name
 rm config_fix.yaml
}

replicate_issue
verify_issue
```

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -26,12 +26,26 @@ SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enab
 SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
 
 
-unconfigure_esm() {
-    rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
-    rm -f $APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY_TRUSTY
-    rm -f $ESM_INFRA_APT_SOURCE_FILE_TRUSTY
-    rm -f $ESM_APT_PREF_FILE_TRUSTY $ESM_INFRA_APT_PREF_FILE_TRUSTY
+check_esm_infra_is_installed() {
+    expected_output="enabled"
+    actual_output=$(ua status | grep esm-infra | awk '{print $3}')
+
+    if [ "$expected_output" = "$actual_output" ]; then
+        return 0
+    else
+        return 1
+    fi
 }
+
+unconfigure_esm() {
+    if ! check_esm_infra_is_installed; then
+        rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
+        rm -f $APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY_TRUSTY
+        rm -f $ESM_INFRA_APT_SOURCE_FILE_TRUSTY
+        rm -f $ESM_APT_PREF_FILE_TRUSTY $ESM_INFRA_APT_PREF_FILE_TRUSTY
+    fi
+}
+
 
 configure_esm() {
     rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
--- a/debian/postinst
+++ b/debian/postinst
@@ -26,19 +26,12 @@ SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enab
 SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
 
 
-check_esm_infra_is_installed() {
-    expected_output="enabled"
-    actual_output=$(ua status | grep esm-infra | awk '{print $3}')
-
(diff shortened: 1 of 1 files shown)


debian/postinst
```diff
@@ -26,13 +26,27 @@ SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enab
 SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
 
 
+check_esm_infra_is_installed() {
+    expected_output="enabled"
+    actual_output=$(ua status | grep esm-infra | awk '{print $3}')
```
blackboxsw, line 31: How's this look just in case we end up changing status column order for some reason.
```suggestion
    actual_output=$(ua status | egrep 'esm-infra.*enabled)
```

debian/postinst
```diff
@@ -26,13 +26,27 @@ SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enab
 SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
 
 
+check_esm_infra_is_installed() {
```
blackboxsw, line 29: function rename suggestion:
```suggestion
check_esm_infra_is_enabled() {
```

debian/postinst
```diff
@@ -26,13 +26,27 @@ SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enab
...
+    expected_output="enabled"
+    actual_output=$(ua status | grep esm-infra | awk '{print $3}')
+
+    if [ "$expected_output" = "$actual_output" ]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
```
blackboxsw, line 39: Sorry consolidating two review comments into one.
1. rename
2. What do you think about this (in case we change column ordering in the TUI at some point we don't want to print $3 to fail)

```suggestion
check_esm_infra_is_enabled() {
    return $(ua status | egrep -q 'esm-infra.*enabled')
}

```

debian/postinst
```diff
@@ -26,13 +26,20 @@ SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enab
 SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
 
 
+check_esm_infra_is_enabled() {
+    return $(ua status | egrep -q 'esm-infra.*enabled')
+}
+
```
blackboxsw, line 32: Sorry I broke your actually working code with my suggestion, we can't actually return the exit code here, it just returns the stdout.
So this works; just tested:
```suggestioncheck_esm_infra_is_enabled() {
    ua status | egrep -q 'esm-infra.*enabled' && return 0 || return 1
}

```

Issue #1080, opened 2020-06-03
config: feature overrides


blackboxsw:
Add the ability to toggle or turn on overrides in /etc/ubuntu-advantage/uaclient.conf.

The expectation is that each feature override would live as key-value pairs under
a top-level `features` key. 

After discussion with Rick,

This issue really just a mechanism by which we can override any machine-token directive, obligation or affordance via a machine-token overlay.

Features to call this issue complete:

- [x]  allow supplementing the machine-token.json response from contract server with other key/values presented by a new local override.json file 

- [x] to toggle the override on, provide the following in `/etc/ubuntu-advantage/uaclient.conf`
```yaml
features:
    machine-token-overlay: <path_to_json>
```

- [x] surface a UAConfig.features method to expose a dict of feature key/values configured

- [x] ua version should be augmented to report the following string if feature is set:
```
$ ua version
25.0 +machine-token-overlay
```

lucasmoura:
@blackboxsw So it will not be possible to override a single feature right ? For example, if an user want to enable only the fips package, that will not be possible using this config. Or we should cover that as well ?

blackboxsw:
@lucasmoura accental self-assign when I filed this.  sorry re-assigning to you.  I think the default case would be to enable operating on all --beta services. I think we want to avoid the granularity of override functionality as best we can because typical overrides should not be too flexible, easy etc. I don't want to productize this override as it feels generally like we are enabling testing, or giving early preview of functionality, not really trying to make the feature flag handling a full featured and flexible config language.

mitechie:
To help, the experience I think we'd like to see is something like

  ua status
  ......(available things)

  ua status --all
  (all possible things)

  ua enable --beta $thing_not_available

Working towards something that can work in that way is the goal.

lucasmoura:
@blackboxsw I just want to check with you an idea that I had regarding this issue. Right now, we have fields on the `machine-token.json` which are lists of dicts, such as:

```json

{
  "availableResources": [
    {
      "available": false,
      "name": "cc-eal"
    },
    {
      "available": true,
      "name": "esm-infra"
    },
    {
      "available": false,
      "name": "fips"
    },
    {
      "available": false,
      "name": "fips-updates"
    },
    {
      "available": true,
      "name": "livepatch"
    },
    {
      "available": true,
      "name": "support"
    }
  ],
}
```
[...]

PR #1100, opened 2020-06-29
Override ua config for beta services

lucasmoura:
As stated in #1080, we need a way to override some configurations for uaclient. This is an initial step in that direction, by allowing users to override the `beta` behavior in uaclient. This PR allows for users to visualize all beta service by adding the following lines to `uaclient.conf`:

```yaml
features:
  allow_beta: true
```

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -52,6 +52,15 @@ def when_i_attach_staging_token(context, token_type, user_spec):
     when_i_run_command(context, "ua attach %s" % token, user_spec)
 
 
+@when("I append the following on uaclient config")
+def when_i_append_to_uaclient_config(context):
+    # command = "echo -e '{}' | sudo tee -a {}".format(
+    #    context.text, DEFAULT_CONFIG_FILE)
+    cmd = "sed -i '$s/$/text/' {}".format(DEFAULT_CONFIG_FILE)
+    print(cmd)
+    when_i_run_command(context, cmd, "with sudo")
+
+
 @then("I will see the following on stdout")
 def then_i_will_see_on_stdout(context):
     assert_that(context.process.stdout.strip(), equal_to(context.text))
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -54,10 +54,12 @@ def when_i_attach_staging_token(context, token_type, user_spec):
 
 @when("I append the following on uaclient config")
 def when_i_append_to_uaclient_config(context):
-    # command = "echo -e '{}' | sudo tee -a {}".format(
-    #    context.text, DEFAULT_CONFIG_FILE)
-    cmd = "sed -i '$s/$/text/' {}".format(DEFAULT_CONFIG_FILE)
-    print(cmd)
+    cmd = "printf '{}\n' > /tmp/uaclient.conf".format(context.text)
+    cmd = 'sh -c "{}"'.format(cmd)
+    when_i_run_command(context, cmd, "as non-root")
+
+    cmd = "cat /tmp/uaclient.conf >> {}".format(DEFAULT_CONFIG_FILE)
+    cmd = 'sh -c "{}"'.format(cmd)
     when_i_run_command(context, cmd, "with sudo")
 
 
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -327,7 +327,9 @@ def _perform_enable(
     @return: True on success, False otherwise
     """
     ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[entitlement_name]
-    config_allow_beta = util.get_allow_beta_value_from_config(cfg.cfg)
+    config_allow_beta = util.is_config_value_true(
+        config=cfg.cfg, path_to_value="features.allow_beta"
+    )
     allow_beta |= config_allow_beta
     if not allow_beta and ent_cls.is_beta:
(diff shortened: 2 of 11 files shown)


uaclient/cli.py
```diff
@@ -327,6 +327,7 @@ def _perform_enable(
     @return: True on success, False otherwise
     """
     ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[entitlement_name]
+    allow_beta |= cfg.cfg.get("features", {}).get("allow_beta", False)
```
blackboxsw, line 330: Let's be a bit more defensive here and get allow_beta and make sure it is booleanish ("true" | "false" | "True" | False) and raise UserFacinError if not since this might be customer visible (and ripe with potential to misspell or provide invalid values).

features/steps/steps.py
```diff
@@ -52,6 +52,17 @@ def when_i_attach_staging_token(context, token_type, user_spec):
     when_i_run_command(context, "ua attach %s" % token, user_spec)
 
 
+@when("I append the following on uaclient config")
```
blackboxsw, line 55: I like this, nice work.

lucasmoura:
@blackboxsw I have created a function to read the `allow_beta` flag from the config file

uaclient/status.py
```diff
@@ -233,6 +233,11 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_REFRESH_SUCCESS = "Successfully refreshed your subscription"
 MESSAGE_REFRESH_FAILURE = "Unable to refresh your subscription"
 
+ERROR_ON_ALLOW_BETA_KEY = """\
+Misspelled value for allow_beta key: {user_key} in \
```
blackboxsw, line 237: Probably better to say "Invalid value" instead of "Misspelled value".
Also it's better not to hardcode static errors as we can give more information.

Let's also say
Expected a boolean string, but found: {invalid_value}."""

uaclient/util.py
```diff
@@ -576,3 +577,18 @@ def write_file(filename: str, content: str, mode: int = 0o644) -> None:
         fh.write(content.encode("utf-8"))
         fh.flush()
     os.chmod(filename, mode)
+
+
+def get_allow_beta_value_from_config(config_dict):
```
blackboxsw, line 582: Sorry, near "bike shedding" here  lets make this utility more general and less specific to allow_beta.

Can we make this function just is_config_true(key_path:str, value: str):
...
raise exceptions.UserFacingError(status.ERROR_INVALID_CONFIG_VALUE.format(key_path=key_path, expected_value="boolean string: true or false", value=value

And we just make the call site:
```
allow_beta = config_dict.get("features", {}).get("allow_beta", "false")
util.is_config_true(key_path="features.allow_beta", value=allow_beta)
```

lucasmoura:
@blackboxsw Done. However, I have slightly modified the solution you have proposed. My idea here was just to make the process of getting the parameter from the dict simpler. But we can discuss if it a proper solution or not.

uaclient/conftest.py
```diff
@@ -108,4 +108,7 @@ def for_attached_machine(
             config.write_cache("machine-token", machine_token)
             return config
 
+        def override_features(self, features_override):
+            self.cfg.update(features_override)
```
blackboxsw, line 112: Since we know we are overriding features, let's make this  the following to reduce the duplication when calling this function and forcing all callsites to provide a dict with a top-level  {"features": } key
```suggestion
            self.cfg.update({"features": features_override})
```

uaclient/tests/test_config.py
```diff
@@ -444,6 +452,9 @@ def test_root_unattached(
         assert expected == cfg.status(show_beta=show_beta)
 
     @pytest.mark.parametrize("show_beta", (True, False))
+    @pytest.mark.parametrize(
+        "features_override", ((None), ({"features": {"allow_beta": True}}))
```
blackboxsw, line 456: Per the above suggestion
```suggestion
        "features_override", ((None), ({"allow_beta": True}))
```

uaclient/tests/test_config.py
```diff
@@ -609,6 +624,9 @@ def test_cache_file_is_written_world_readable(
         )
 
     @pytest.mark.parametrize("show_beta", (True, False))
+    @pytest.mark.parametrize(
+        "features_override", ((None), ({"features": {"allow_beta": False}}))
```
blackboxsw, line 628:
```suggestion
        "features_override", ((None), ({"allow_beta": False}))
```

uaclient/tests/test_util.py
```diff
@@ -558,3 +558,45 @@ def test_prompt_text(self, m_input, assume_yes, message, input_calls):
...
+class TestIsConfigValueTrue:
+    @pytest.mark.parametrize(
+        "config_dict, return_val",
+        [
+            ({}, False),
+            ({"features": {}}, False),
+            ({"features": {"allow_beta": "true"}}, True),
+            ({"features": {"allow_beta": "True"}}, True),
+            ({"features": {"allow_beta": "false"}}, False),
+            ({"features": {"allow_beta": "False"}}, False),
```
blackboxsw, line 572:
```suggestion
            ({}}, False),
            ({"allow_beta": "true"}, True),
            ({"allow_beta": "True"}, True),
            ({"allow_beta": "false"}, False),
            ({"allow_beta": "False"}, False),
```

lucasmoura:
@blackboxsw done

PR #1102, opened 2020-06-30
[WIP] Add feature to override machine-token info

lucasmoura:
Currently, there is no way to override configurations in uaclient provided by the machine-token received by the uaclient server. We are now adding the feature to allow for an user to override such configs. To do that, the user will need to provided the following lines on the `uaclient.conf` file:

```yaml
features:
  machine_token_overlay:  PATH_TO_FILE
```

Where `PATH_TO_FILE` is must be a path to a `json` file with the configurations that should be override.

 There are still lacking some behave tests for this feature, since the abstraction to write lines into `uaclient.conf` is provided by #1100, but once this PR lands, I will update this PR with the necessary behave tests

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -160,7 +160,7 @@ class UAConfig:
             )
 
             return json.loads(machine_token_overlay_content)
-        except json.JSONDecodeError:
+        except ValueError:
             raise exceptions.UserFacingError(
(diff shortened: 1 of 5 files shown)


lucasmoura:
Just realized this solution is far from complete. I am closing this PR for now.

PR #1103, opened 2020-06-30
Add secure contract token for behave tests

lucasmoura:
Add test contract token for behave tests


lucasmoura:
We have decided to create the environment variable through the travis interface as stated here:
https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings

PR #1104, opened 2020-07-02
Override machine token json file

lucasmoura:
This is an initial idea on how we should merge the overlay machine token dict into the `machine-token.json` file provided by the server. Right now, to allow `machine-token.json` to be merged we would need to append the following line into `uaclient.conf`:

```yaml
features:
  machine_token_overlay: path-to-overlay-json
```

This `path-to-overlay-json` is sparse json file that should contain all the path that should be replaced, keys that should be added or items that should be added into a list. For example, let's assume the following json:

```json
{
     "new-key": "new-value",
     "availableResources": [ 
        {"available": false, "name": "livepatch"},
        {"available": true, "name": "testresource"}
     ],
     "machineTokenInfo": {
         "resourceEntitlements": [
            {
              "type": "livepatch",
              "directives": {
                 "remoteServer": "new-remote-server"
              }
           }
        ]
   }
}
```

In that situation we would update the `machine-token.json` as follows:

* `new-key`: Suppose this key is not found in `machine-token.json`, we will add it to that dict
* `list of dict`: We will search the `machine-token.json` list for a dict with the same `id-key`, (`type` for `resourceEntitlements` and `name` for `resourceEntitlements`), if found we update the corresponding dict, if not, we add the new dict into the list.
* `list of other types`: We just extend the list with the values on the machine token overlay json

We are still lacking behave test, but once we settle for what is the best solution, I will add them.


blackboxsw:
This looks like a solid approach @lucasmoura please do feel free to flesh it out.

uaclient/util.py
```diff
@@ -615,3 +615,52 @@ def is_config_value_true(config: "Dict[str, Any]", path_to_value: str):
...
+    {"a": 1, "b": {"c": 10, "d": 3}}
+
+    @param base_dict: The dict to be updated
+    @param overlay_dict: The dict with information to be added into base_dict
+    """
+
+    def get_id_key(key):
+        id_map = {"availableResources": "name", "resourceEntitlements": "type"}
+
+        return id_map.get(key)
```
blackboxsw, line 639: This doesn't need to be a method. Let's just define a global dict

```suggestion
MERGE_ID_KEY_MAP = {"availableResources": "name", "resourceEntitlements": "type"}
```

uaclient/util.py
```diff
@@ -615,3 +615,52 @@ def is_config_value_true(config: "Dict[str, Any]", path_to_value: str):
...
+
+    for key, value in overlay_dict.items():
+        base_value = base_dict.get(key)
+        if isinstance(base_value, dict) and isinstance(value, dict):
+            depth_first_merge_overlay_dict(base_dict[key], value)
+        elif isinstance(base_value, list) and isinstance(value, list):
+            if len(base_value) and isinstance(base_value[0], dict):
+                update_dict_list(base_dict[key], value, key=key)
+            else:
+                base_value.extend(value)
```
blackboxsw, line 664: Since all the other non-dict list items in the token contract as strings, I think it's probably better if we default to replace the whole list if the value of the item is not a dict.  So we'd provide a full list of strings 

base: 
                  "series" : [
                     "precise",
                     "trusty",
                     "xenial",
                     "bionic",
                     "focal"
                  ]


overlay dropping "bionic" ::
                  "series" : [
                     "precise",
                     "trusty",
                     "xenial",
                     "focal"
                  ]

lucasmoura:
@blackboxsw I have updated the code. However, it seems that the behave tests are now failing for both focal and trusty on the following scenario:

https://github.com/canonical/ubuntu-advantage-client/blob/master/features/attach_invalidtoken.feature

Upon a closes inspection, I have found that accessing the contract server is now returning:
```
Unauthorized: [https://contracts.canonical.com/v1/context/machines/token]unauthorized
```

Instead of the error we were expecting. Is that an expected behavior ?

PR #1105, opened 2020-07-03
Contract token test

lucasmoura:
Testing `UACLIENT_BEHAVE_CONTRACT_TOKEN` use in travis while also fixing some behave tests.

There is sill on scenario failing, that is is related to the problem in #1104, where the `ua enable INVALID_TOKEN` command does not raise the expected errror because the server is returning a different error than the one expected.

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -166,25 +166,13 @@ Feature: Command behaviour when attached to an UA subscription
         Given a `trusty` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
         And I run `ua version` as non-root
-        Then I will see the following on stdout:
-            """
-            25.0
-            """
+        Then I will see the uaclient version on stdout
         When I run `ua version` with sudo
-        Then I will see the following on stdout:
-            """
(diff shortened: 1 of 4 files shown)


features/attached_commands.feature
```diff
@@ -168,22 +168,22 @@ Feature: Command behaviour when attached to an UA subscription
...
         Then I will see the following on stdout:
             """
-            20.4
+            25.0
             """
         When I run `ua version` with sudo
         Then I will see the following on stdout:
             """
-            20.4
+            25.0
```
blackboxsw, line 176: I don't like that we have to keep updating this as we change version numbers.  Can we update this to import uaclient.version.__VERSION__. And use that throughout?

lucasmoura:
@blackboxsw I will try to figure something out related to the version test, because when we import the `get_version` commands it gets the version of the machine that is running the behave test, not the one being tested. But I will try to figure something out

lucasmoura:
@blackboxsw I have fixed the version tests. I will merge this PR, but if you think there is a better way to handle this issue, just let me know

PR #1106, opened 2020-07-06
contract: handle 401 on invalid token

blackboxsw:


uaclient/tests/test_contract.py
```diff
@@ -235,10 +235,10 @@ def fake_contract_client(cfg):
             fake_client._responses = {
                 API_V1_CONTEXT_MACHINE_TOKEN: ContractAPIError(
                     util.UrlError(
-                        "Server error", code=500, url="http://me", headers={}
+                        "Server error", code=401, url="http://me", headers={}
```
mitechie, line 238: Should we be doing more generic "starts with 4xx" here to be flexible?

uaclient/contract.py
```diff
@@ -263,7 +262,7 @@ def request_updated_contract(
             )
         except util.UrlError as e:
             if isinstance(e, ContractAPIError):
-                if API_ERROR_INVALID_TOKEN in e:
+                if hasattr(e, "code") and e.code == 401:
```
mitechie, line 265: Is 400 a straight invalid (too many characters or something?) vs expired vs anything else?

uaclient/tests/test_contract.py
blackboxsw: Good point @mitechie I've added 403 as a viable failure case that implies a token error, any other http status codes outside of that seem to imply something other that just invalid token. In these cases the client will still reply with the generic:

```Failed to attach machine. See https://ubuntu.com/advantage```

uaclient/contract.py
blackboxsw: yes 400 is bad request, so I don't think we should treat that as invalid token error. It could be client doesn't know how to talk to a new contract server

Issue #1029, opened 2020-04-15
Livepatch and FIPS* cannot be enabled at the same time


blackboxsw:
Two parts:

* When livepatch is already enabled:
 Emit the following message and exit non-zero upon `sudo ua enable fips`:
"FIPS[-updates] is not available when livepatch is enabled"
* When FIPS* is already enabled:
 Emit the following message and exit non-zero upon `sudo ua enable livepatch`:
 Cannot enable livepatch when FIPS[-updates] is enabled

Issue #1112, opened 2020-07-07
Cannot use Scenario Outline with vm


lucasmoura:
When trying to run a Scenario Outline with different examples on behave, the vm tests always hangs after the first test. It seems that when creating the second vm for testing, it hangs waiting for the `lxd-agent` to be found. This is a weird issue, and I do not believe that it is fully related to the `Scenario Outline` part of behave, but it was on this scenario that I have spotted the issue.

lucasmoura:
Could not replicate this issue on a second run, closing it for now

Issue #1111, opened 2020-07-07
Cannot reuse vm images created by behave


lucasmoura:
The process to create vm is slow, so therefore is useful to reuse them when developing a new a test. However, after creating a vm test we cannot reuse them later. The reason for this is that we cannot find the [series](https://github.com/canonical/ubuntu-advantage-client/blob/master/features/environment.py#L149) attribute or the [machine_type](https://github.com/canonical/ubuntu-advantage-client/blob/master/features/environment.py#L152) one.

I have spotted this issue trying to run tests with a bionic vm. To reproduce it, run the following command (Just make sure that the bionic tests on `attached_enable.feature` are marked with the `@wip` tag:
```sh
UACLIENT_BEHAVE_DESTROY_INSTANCES=0 UACLIENT_BEHAVE_BUILD_PR=1 UACLIENT_BEHAVE_MACHINE_TYPE=lxd.vm UACLIENT_BEHAVE_CONTRACT_TOKEN=<your_yoken> tox -e behave --  --tags=@wip features/attached_enable.feature
```

After that, just try to reuse the vm created here and verify if the test are actually reusing them:
```
UACLIENT_BEHAVE_CONTRACT_TOKEN=<your_token> tox -e behave --  --tags=@wip features/attached_enable.feature -D reuse_container=MY-BIONIC-VM-NAME
```

PR #1113, opened 2020-07-07
Do not enable services if blocking services is active

lucasmoura:
Users should not be able to enable `livepatch` if `fips` is active or `fips` if `livepatch` is active as stated in #1029 . We are now add the possibility for verifying if `blocking services` are enabled before activating a service. If this happens, we will not enable the service and show the user which services are responsible for that behavior.

Furthermore, we are only testing this feature now in `Bionic`. Although there are some improvements we can achieve for testing that feature, there are some potential issues that I have found (#1111 and #1112) that are posing a difficult for developing vm tests

Fixes: #1029

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -226,21 +226,29 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.focal
     @uses.config.machine_type.lxd.vm
-    Scenario Outline: Attached enable of vm-based services in a focal lxd vm
+    Scenario: Attached enable of vm-based services in a focal lxd vm
         Given a `focal` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
         And I run `ua disable livepatch` with sudo
-        And I run `ua enable <service> --assume-yes --beta` with sudo
+        And I run `ua enable fips --assume-yes --beta` with sudo
         Then I will see the following on stdout:
             """
             One moment, checking your subscription first
-            <title> is not available for Ubuntu 20.04 LTS (Focal Fossa).
+            FIPS is not available for Ubuntu 20.04 LTS (Focal Fossa).
             """
 
-        Examples: ubuntu service
-        | service      | title        |
-        | fips         | FIPS         |
-        | fips-updates | FIPS Updates |
+    @series.focal
+    @uses.config.machine_type.lxd.vm
+    Scenario: Attached enable of vm-based services in a focal lxd vm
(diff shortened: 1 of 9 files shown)


uaclient/entitlements/base.py
```diff
@@ -32,12 +32,15 @@ class UAEntitlement(metaclass=abc.ABCMeta):
...
 
-    #  Whether to assume yes to any messaging prompts
+    # Whether to assume yes to any messaging prompts
     assume_yes = False
 
     # Wheter that entitlement is in beta stage
     is_beta = False
 
+    # List of services that if enabled, will block enabling the service
+    blocking_entitlements = []  # type: List[str]
```
blackboxsw, line 42: We probably minimally want to make this a class method with the @property decorator. The reason being we would like the ability to override this option using feature overrides in configuration.  This primitive probably makes sense as a separate method. But I would like to look at integrating this method/property call into the static_affordances checks for a service. Instead of a separate call from can_enable here's a stale pr I started putting together a while back to solve this issue. But didnt have time to finish it.  Feel free to use any of it if you think it is applicable. https://github.com/canonical/ubuntu-advantage-client/pull/1114

features/attached_enable.feature
```diff
@@ -225,20 +226,22 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.focal
     @uses.config.machine_type.lxd.vm
-    Scenario: Attached enable of vm-based services in a focal lxd vm
+    Scenario Outline: Attached enable of vm-based services in a focal lxd vm
```
blackboxsw, line 229: @lucasmoura what is your opinion about avoiding a Scenario Outline for this test and just leaving the replicated tests? The reason I ask, is because this a non-destructive test to the VM (nothing changes config-wise when we pass) and it costs us the setup of a brand new VM to just test the other case

features/attached_enable.feature
```diff
@@ -266,3 +269,55 @@ Feature: Enable command behaviour when attached to an UA subscription
...
+            ESM Infra enabled
+            Installing canonical-livepatch snap
+            Canonical livepatch enabled
+            """
+        When I run `ua disable livepatch` with sudo
+        Then I will see the following on stdout:
+            """
+            Removing canonical-livepatch snap
+            """
+        When I run `ua enable fips --assume-yes --beta` with sudo
```
blackboxsw, line 290: Can we also add a check that "which canonical-livepatch" returns empty string after the removal

lucasmoura:
@blackboxsw Done, I have updated the scenarios

PR #1115, opened 2020-07-08
Adapt behave test to use xenial and bionic

lucasmoura:
Right now, our behave tests maunly cover trusty and focal releases. We are creating a new way to allow our behave test to perform test on all 4 releases we want to support (Bionic, Focal, Trusty and Xenial)

This is still a WIP, since there are still tests that need to be converted and some test that may need to be updated, as this behave [issue](https://github.com/behave/behave/issues/847) states.

However, this PR already has the initial idea for how we can reuse tests for multiple ubuntu releases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,79 @@ matrix:
   fast_finish: true
   include:
     - python: 3.7
-      env: TOXENV=behave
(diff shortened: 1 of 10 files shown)


lucasmoura:
#1116 covers this PR

PR #1117, opened 2020-07-10
Add back release info to published vms

lucasmoura:
Since the vms test are costly to create, when developing new vm behave tests, it is really useful to be able to reuse them.

We found out that the issue is that when we publish the vm images locally, we end-up losing the image information on it. This PR adds a way to bring back at least the `image.release` info back into the image, since this is the info we need to reuse it.

Fixes: #1111


lucasmoura:
@blackboxsw updated

Issue #1119, opened 2020-07-13
fips: test that enabled fips is chosen after reboot


blackboxsw:
Integration tests need to config the following when FIPS is enabled on a new image:
 - after reboot, fips kernel is active from `uname -r`
 - after reboot, the kernel represents FIPS config is active by `cat  /proc/sys/crypto/fips_enabled` == 1

Not working on Xenial at the moment

blackboxsw:
Resolved as of commit ad754d9c

Issue #1118, opened 2020-07-13
livepatch: validate `canonical-livepatch status` command on enabled images


blackboxsw:
Include a slightly deeper test that `canonical-livepatch status` actually  claims that the snap is running and applying patches (or not). This means we are relying on the snap itself to confirm that it is running appropriately.

We want to confirm:
   - canonical-livepatch is actually active
   - expected  patches are applied (or not) given the platform/release we are testing

PR #1120, opened 2020-07-13
Add test more behave tests for livepatch

lucasmoura:
As stated by #1118 and #1119 , we need to create more test for livepatc/fips for both Bionic and Xenial. This PR is an initial effort towards that direction

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -255,9 +255,10 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.bionic
     @uses.config.machine_type.lxd.vm
-    Scenario: Attached enable of vm-based services in a bionic lxd vm
-        Given a `bionic` machine with ubuntu-advantage-tools installed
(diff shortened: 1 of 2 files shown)


features/attached_enable.feature
```diff
@@ -266,3 +266,30 @@ Feature: Enable command behaviour when attached to an UA subscription
...
             A reboot is required to complete install
             """
+
+    @series.bionic
+    @series.xenial
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable livepatch on a machine with fips active
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `canonical-livepatch status` with sudo
+        Then I will see the following on stderr:
```
blackboxsw, line 276: Let's add a When/Then for running ua status and validating that stdout regexp matches fips +enabled too

lucasmoura:
@blackboxsw I have updated both tests and added the fips kernel test for bionic. After we solve the priority issue in Xenial, I can add it back there too.

Issue #1093, opened 2020-06-17
add aws ubuntu pro integration tests


blackboxsw:
Add support for running behave integration tests using aws ubuntu pro images

blackboxsw:
Fixed by badf286

PR #1121, opened 2020-07-15
tests: add aws pro integration coverage for daily PPA

blackboxsw:
This branch would be part 1 of 2 for AWS Pro support.  It addresses part of Issue: #1093

In this branch:

- fix setup.py to handle test-requirements which use remote git links for dependencies
- Add pro.aws machine_type definition and @uses decorator for feature tests
- Use pycloudlib to perform ec2 key creation, vpc creation and launch_ec2 instances from a common vpc
- Allow building ec2 Ubuntu PRO images which install the ubuntu-advantage-tools from daily PPA
- separate integration-requirements.txt file as some pycloudlib deps are not available on trusty

We will not run integration tests from trusty, so this dependency mismatch is irrelevant to our test runs

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -3,14 +3,18 @@ import os
 import itertools
 import subprocess
 import textwrap
+import logging
 from typing import Dict, Optional, Union, List
 
 from behave.model import Feature, Scenario
 
 from behave.runner import Context
 
+import pycloudlib  # type: ignore
+
 from features.util import (
     launch_lxd_container,
+    launch_ec2,
     lxc_exec,
     lxc_get_property,
     lxc_build_deb,
@@ -18,6 +22,22 @@ from features.util import (
 
 PR_DEB_FILE = "/tmp/ubuntu-advantage.deb"
 ALL_SUPPORTED_SERIES = ["bionic", "focal", "trusty", "xenial"]
+
+EC2_KEY_FILE = "uaclient.pem"
+
+DAILY_PPA = "http://ppa.launchpad.net/canonical-server/ua-client-daily/ubuntu"
+
+USERDATA_INSTALL_DAILY_PRO_UATOOLS = """\
+#cloud-config
+apt:
+  sources:
+    ua-tools-daily:
+        source: "deb {daily_ppa} $RELEASE main"
+        keyid: 8A295C4FB8B190B7
+packages: [ubuntu-advantage-tools, ubuntu-advantage-pro]
+""".format(
+    daily_ppa=DAILY_PPA
+)
 
 
 class UAClientBehaveConfig:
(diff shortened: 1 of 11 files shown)


features/environment.py
```diff
@@ -139,11 +187,41 @@ def from_environ(cls, config) -> "UAClientBehaveConfig":
...
+            keypair = ec2_api.client.create_key_pair(
+                KeyName="uaclient-integration"
+            )
+            with open("uaclient1.pem", "w") as stream:
+                stream.write(keypair["KeyMaterial"])
+            os.chmod("uaclient1.pem", 0o600)
+        ec2_api.use_key(
+            "uaclient1.pem", "uaclient1.pem", "uaclient-integration"
+        )
+        vpc = ec2_api.create_vpc("uaclient-integration")
```
lucasmoura, line 224: I don't know if I am missing something, but why do we need to create this vpc here ?

blackboxsw: ahh oversight from an earlier test. I've removed it now.

features/environment.py
```diff
@@ -139,11 +178,37 @@ def from_environ(cls, config) -> "UAClientBehaveConfig":
...
     context.series_image_name = {}
     context.series_reuse_image = ""
     context.reuse_container = {}
     context.config = UAClientBehaveConfig.from_environ(context.config)
-
+    has_aws_keys = bool(
+        context.config.aws_access_key_id
+        and context.config.aws_secret_access_key
+    )
+    if has_aws_keys:
```
lucasmoura, line 193: If the `has_aws_keys` check fails, we will not create the `ec2_api` variable in `context.config`. That will end up causing the scenario to fail when we try to create a ec2 instance in the `given_a_machine` step. Maybe we could abort the test here if the check fails and we are running pro tests

features/ubuntu_pro.feature
```diff
@@ -0,0 +1,24 @@
+@uses.config.machine_type.pro.aws
+Feature: Command behaviour when attached to an UA subscription
+
+    @series.xenial
+    @series.bionic
+    @series.focal
+    Scenario Outline: Attached refresh in a trusty machine
```
lucasmoura, line 7: I am unable to run this scenario. Right now, when running it, I receive the following error:

```
in before_scenario: ClientError: An error occurred (InvalidParameterValue) when calling the CreateSubnet operation: invalid value for parameter ipv6-cidr-block: 64
```

I have checked to see if I am using the right `pycloudlib` branch, but the problem still persists

lucasmoura: Also, these are the version of my boto libs:

```
boto3==1.14.20
botocore==1.17.20
```

blackboxsw: I think something in your specific ec2 configuration opts to avoid ipv6 setup on subnets. I bet I have configured that on my ec2 account by default so we might have to fix pycloudlib to handle non-ipv6 setups.

Can you run this script in your pycloudlib repo root
tox -e pylint
.tox/pylint/bin/python3 ./1.py
https://paste.ubuntu.com/p/Jt4G9C9Ddb/

blackboxsw: Ok this is a timing issue when creating a subnet before the vpc is ready (and includes the amazon allocated IPv6 address).
I've added a vpc.wait_until_available and vpc.reload() before creating subnets to avoid this issue. https://github.com/canonical/pycloudlib/pull/8

PR #1125, opened 2020-07-16
Overrides/disable auto attach

blackboxsw:
Add config feature override to avoid `ua auto-attach` on pro instances.

Adding the following to uaclient.conf will result in skipping auto-attach to aid in test creation of -proposed ubuntu-advantage-tools packages.

```yaml
features:
   disable_auto_enable: true
```

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -511,6 +511,13 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
 
 @assert_root
 def action_auto_attach(args, cfg):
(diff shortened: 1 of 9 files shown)


uaclient/cli.py
```diff
@@ -511,6 +511,13 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
 
 @assert_root
 def action_auto_attach(args, cfg):
+    disable_auto_attach = util.is_config_value_true(
```
lucasmoura, line 514: I think we should add a behave test for this feature

PR #1126, opened 2020-07-16
Add manual test for livepatch enable on trusty hwe

lucasmoura:
Add a manual verification for livepatch enablement on a trusty vm with hwe kernel as stated in #1126 . I have created a temp directory on the sru folder until we do not have the SRU launchpad bug number.

However, there is something we need to discuss regarding this test, how to handle the contract token, since it will be made public if we commit the output of the test as well.


lucasmoura:
@mitechie I have updated the file to contain the output of the test script. In the output, the contract token is being redacted. Also, I can update the script once we have a testing version of the new uaclient package, since right now I am building the package from master.

Also, @blackboxsw should I test livepatch enablement before upgrading the kernel to see it fails ?

blackboxsw:
> Also, @blackboxsw should I test livepatch enablement before upgrading the kernel to see it fails ?
@lucasmoura I don't think we need that in this PR, we need that in a separate PR that should be covered in per-PR daily integration tests.

PR #1131, opened 2020-07-23
Docs/link readme to travis project build status

blackboxsw:


blackboxsw:
@OddBloke @lucasmoura raised a good point here. CI for this project is still allocated to travis-ci.org (not our .com production acct). We should chat with you tomorrow about whether migration to `.com` makes sense for this project.

PR #1134, opened 2020-07-27
tests: setup_logging so we see skip reasons on skipped tests

blackboxsw:


features/environment.py
```diff
@@ -183,6 +183,7 @@ def from_environ(cls, config) -> "UAClientBehaveConfig":
 
 def before_all(context: Context) -> None:
     """behave will invoke this before anything else happens."""
+    context.config.setup_logging()
```
lucasmoura, line 186: I don't know If I am missing something, but does the `UAClientBehaveConfig` does have the `setup_logging` method ?

blackboxsw: Correct, we are calling this function on <class 'behave.configuration.Configuration'> before we actually overwrite that context.config to our "UAClientBehaveConfig" below. I think it's probably confusing that we replace context.config with our local config instance. Maybe we can think about a better way to do this.

PR #1137, opened 2020-07-30
test: travis add aws-pro job avoid 'make clean' when build_pr is set

blackboxsw:
Note: I'd like to merge this branch with separate commits instead of squash merging.

This PR activates the travis job behave-aws-pro

It also drop a 'make clean' in favor of tar --exclude-vcs* flags which ignore .tox and .git subdirs when
creating a source tar file to build from during testing.

Additional features:
 - print launched instance.id right after launch and add retries on instance.wait failuress
 - modify some print statements

### DIFF
--- a/features/util.py
+++ b/features/util.py
@@ -85,9 +85,26 @@ def launch_ec2(
     vpc = context.config.cloud_api.get_or_create_vpc(
         name="uaclient-integration"
     )
-    inst = context.config.cloud_api.launch(
-        image_name, user_data=user_data, vpc=vpc
+    try:
+        inst = context.config.cloud_api.launch(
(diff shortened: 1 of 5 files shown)


features/util.py
```diff
@@ -85,9 +85,27 @@ def launch_ec2(
...
+        )
     )
+    time.sleep(15)
+    for sleep in (5, 10, 15):
+        try:
+            inst.wait()
+            break
+        except Exception as e:
+            print("--- Retrying instance.wait on {}".format(str(e)))
+            pass
```
lucasmoura, line 108: I think we can remove the `pass` call here

PR #1138, opened 2020-08-03
Azure integration

lucasmoura:
This PR adds the Azure Pro test infrastructure into our integration tests. Besides that, it also implements the following changes:

* Refactor how cloud tests are performed
* Update how we select scenarios to skip based on the `machine_type` token

Right now, this is a WIP because I am experiencing an issue regarding how the paramiko SSH keys are being clean. In the end of the test, the following exception is being throw:

```
Exception ignored in: <function BaseInstance.__del__ at 0x7f82b93168b0>                                                                                                                                            
  File "/home/lucasmoura/projects/ubuntu-advantage-client/.tox/behave-azure-pro/lib/python3.8/site-packages/pycloudlib/instance.py", line 57, in __del__                                                           
TypeError: catching classes that do not inherit from BaseException is not allowed                                                                                                                                  
Exception ignored in: <function BaseInstance.__del__ at 0x7f82b93168b0>                                                                                                                                            
  File "/home/lucasmoura/projects/ubuntu-advantage-client/.tox/behave-azure-pro/lib/python3.8/site-packages/pycloudlib/instance.py", line 57, in __del__                                                           
TypeError: catching classes that do not inherit from BaseException is not allowed 
```
I am still investigating why this is happening, but I think the work can start to be reviewed.

Also, we still need the Azure credentials to use to run the integration tests

### DIFF
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -4,7 +4,11 @@ import pycloudlib  # type: ignore
 import time
 import yaml
 
-from typing import Optional
+try:
+    from typing import Tuple, List, Union, Optional  # noqa
+except ImportError:
+    # typing isn't available on trusty, so ignore its absence
+    pass
 
 
 class Cloud:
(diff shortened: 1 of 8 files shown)


features/cloud.py
```diff
@@ -0,0 +1,373 @@
...
+
+        :returns:
+            An AWS cloud provider instance
+        """
+        if not image_name:
+            if not series:
+                raise ValueError(
+                    "Must provide either series or image_name to launch azure"
+                )
+            with open("features/azure-ids.yaml", "r") as stream:
```
blackboxsw, line 365: Given that you have pulled the pro image choice logic into the Cloud subclasses, I think we might also need to add a machine_type param to the __init__ of these Cloud instances so that we can determine if "generic.azure" or "generic.aws" machine_type is provided. Then if "pro" not in machine_type we can source azure-ids.yaml, otherwise we call cloud.released_image(series) for the non-pro machine_types

features/cloud.py
```diff
@@ -0,0 +1,373 @@
...
+class Cloud:
+    """Base class for cloud providers that should be tested through behave.
+
+    :param tag:
+        A tag to be used when creating the resources on the cloud provider
+    """
+
+    name = ""
+
+    def __init__(self, tag):
```
blackboxsw, line 19: For any method and function we touch, let's try to add typing hints to better describe params and returns (and validated with `tox -e mypy`)

```suggestion
    def __init__(self, tag: str)  -> None:
```

features/cloud.py
```diff
@@ -0,0 +1,373 @@
...
+
+    def has_keys(self):
+        """Check if cloud has necessary keys to launch instance."""
+        raise NotImplementedError
+
+    def get_instance_id(self, instance):
+        """Return the instance identifier.
+
+        :param instance:
+            An instance created on the cloud provider
```
blackboxsw, line 31: The id of an instance created in the cloud.
```suggestion
            The string of the unique instance id.
```

features/cloud.py
```diff
@@ -0,0 +1,373 @@
...
+            The ubuntu release to be used when creating an instance. We will
+            create an image based on this value if the used does not provide
+            a image_name value
+        :param image_name:
+            The name of the image to be used when creating the instance
+        :param user_data:
+            The user data to be passed when creating the instance
+
+        :returns:
+            An cloud provider instance
```
blackboxsw, line 67:
```suggestion
            A cloud provider instance
```

features/cloud.py
```diff
@@ -0,0 +1,373 @@
...
+                tag=self.tag,
+                client_id=self.az_client_id,
+                client_secret=self.az_client_secret,
+                tenant_id=self.az_tenant_id,
+                subscription_id=self.az_subscription_id,
+            )
+
+        return self._api
+
+    def get_instance_id(self, instance):
```
blackboxsw, line 301: let's add typing hinks here too (and all other methods/functions)

lucasmoura:
@blackboxsw Done. I have updated the code.

I will still seek the Azure SSH exception raised when we run the command with `UACLIENT_BEHAVE_BUILD_PR=1`

blackboxsw:
+1 on this awaiting CI to merge.

PR #1141, opened 2020-08-05
tests: add integration tests for aws canonical cloud-images (non Ubuntu-pro)

blackboxsw:
Add AWS and Azure integration test support for non-Ubuntu PRO images.

This PR adds two new tox targets and travis jobs: behave-azure and behave-aws which run BDD scenarios decorated with 
the tags `uses.context.config.machine_type.aws.generic` or `uses.context.config.machine_type.azure.generic` for AWS and Azure respectively.

These tox targets/travis jobs will use pycloudlib's released_image(<series>) to obtain the latest public image on the cloud platform instead of specifically requesting the PRO image ids on the cloud.

Each generic (non-pro) image requires proper cloud credentials as UACLIENT_BEHAVE_* environment variables as well
as a UACLIENT_BEHAVE_CONTRACT_TOKEN in order to attach the maching to Ubuntu Advantage.

Additionally, we make the travis jobs depend on conditional inclusion only if on of the required environment variables is present to allow for private repos to run a subset of integration tests without failing if they don't wish to run Azure or AWS for instance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,9 @@ script:
 matrix:
   fast_finish: true
   include:
-    - if: env(UACLIENT_BEHAVE_AZ_CLIENT_SECRET)
+    - if: env(UACLIENT_BEHAVE_AWS_SECRET_ACCESS_KEY) AND env(UACLIENT_BEHAVE_CONTRACT_TOKEN)
       python: 3.7
-      env: TOXENV=behave-pro-azure
+      env: TOXENV=behave-aws
       script:
           - >
               if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -29,97 +59,101 @@ matrix:
...
+              if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
+                  BUILD_PR=0
+              else
                  cd $TRAVIS_BUILD_DIR/..
                  tar -zcf pr_source.tar.gz ubuntu-advantage-client
                  cp pr_source.tar.gz /tmp
                  ls -lh /tmp
                  cd $TRAVIS_BUILD_DIR
               fi
+          # bionic has lxd from deb installed, remove it first to avoid
```
lucasmoura, line 89: I think we should update this comment, since it stating that this is bionic related, but we are doing that on all of the releases

README.md
```diff
@@ -234,17 +234,29 @@ you need to add `-D reuse_container=container_name`:
...
+  tox -e behave-aws
+```
+
+In order to run EC2 tests the following environment variables are required:
+  - UACLIENT_BEHAVE_AWS_ACCESS_KEY_ID
+  - UACLIENT_BEHAVE_AWS_SECRET_ACCESS_KEY
+
+
+To specifically run non-ubuntu pro tests using canonical cloud-images an
+additiional token obtained from https://ubuntu.com/advantage needs to be set:
```
lucasmoura, line 253: s/additiional/additional/

PR #1142, opened 2020-08-07
Improve pro testing

lucasmoura:
In the pro images test, verify if the both `esm-infra` and `esm-apps` are properly configured and packages originated from those sources can be downloaded and installed without an issue.

This is a working solution, but there is still a discussion about if we should add dummy packages to both of these origins in an attempt to simplify and guarantee a more reliable test for pro images. But since we have not reached a consensus on it yes, this is a working solution nonetheless, although it suffers from some flaws, like checking the package version directly for esm, which is not mandatory for the package maintainer to add when uploading a new package.

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -183,6 +183,16 @@ def then_i_should_see_that_the_command_is_not_found(context, cmd_name):
     assert_that(expected_return, equal_to(actual_return))
 
 
+@then("I verify that running `{cmd_name}` as `{spec}` succeeds")
+def then_i_verify_that_running_cmd_with_spec_succeeds(context, cmd_name, spec):
+    spec = "with sudo" if spec == "sudo" else "as non-root"
+    when_i_run_command(context, cmd_name, spec)
+
+    expected_return = "0"
+    actual_return = str(context.process.returncode)
+    assert_that(actual_return, equal_to(expected_return))
+
+
 def get_command_prefix_for_user_spec(user_spec):
     prefix = []
     if user_spec == "with sudo":
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -183,14 +183,22 @@ def then_i_should_see_that_the_command_is_not_found(context, cmd_name):
     assert_that(expected_return, equal_to(actual_return))
 
 
-@then("I verify that running `{cmd_name}` as `{spec}` succeeds")
-def then_i_verify_that_running_cmd_with_spec_succeeds(context, cmd_name, spec):
-    spec = "with sudo" if spec == "sudo" else "as non-root"
+@then("I verify that running `{cmd_name}` `{spec}` exits `{exit_codes}`")
+def then_i_verify_that_running_cmd_with_spec_exits_with_codes(
+    context, cmd_name, spec, exit_codes
+):
     when_i_run_command(context, cmd_name, spec)
 
-    expected_return = "0"
-    actual_return = str(context.process.returncode)
-    assert_that(actual_return, equal_to(expected_return))
+    expected_codes = exit_codes.split(",")
+    assert str(context.process.returncode) in expected_codes
+
+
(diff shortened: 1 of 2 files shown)


features/steps/steps.py
```diff
@@ -183,6 +183,16 @@ def then_i_should_see_that_the_command_is_not_found(context, cmd_name):
...
 
 
+@then("I verify that running `{cmd_name}` as `{spec}` succeeds")
+def then_i_verify_that_running_cmd_with_spec_succeeds(context, cmd_name, spec):
+    spec = "with sudo" if spec == "sudo" else "as non-root"
+    when_i_run_command(context, cmd_name, spec)
+
+    expected_return = "0"
+    actual_return = str(context.process.returncode)
+    assert_that(actual_return, equal_to(expected_return))
```
blackboxsw, line 193: We can drop the extra "as" from here as it is duplicated by {spec}  which will be "as non-root" or "with sudo"
How about we parameterize the exit code a bit


What do you think about this:
```suggestion
@then("I verify that running `{cmd_name}`{spec} exits [{exit_codes}]")
def then_i_verify_that_running_cmd_with_spec_succeeds(context, cmd_name, spec, exit_codes):
    when_i_run_command(context, cmd_name, spec)

    expected_codes = ",".split(exit_codes)
    assert str(context.process.returncode) in expected_codes
```

features/ubuntu_pro.feature
```diff
@@ -27,8 +27,46 @@ Feature: Command behaviour when attached to an UA subscription
...
+        \s*500 https://esm.ubuntu.com/apps/ubuntu <release>-apps-security/main amd64 Packages
+        """
+
         Examples: ubuntu release
-           | release | cc-eal | cc-eal-s | fips | fips-s |
-           | xenial  | yes    | disabled | yes  | n/a    |
-           | bionic  | yes    | n/a      | yes  | n/a    |
-           | focal   | yes    | n/a      | yes  | n/a    |
+           | release | cc-eal | cc-eal-s | fips | fips-s | infra-pkg    | apps-pkg |
+           | xenial  | yes    | disabled | yes  | n/a    | libkrad0     | jq       |
```
blackboxsw, line 70: +1 on hardcoding these pkgs for now so we know we are good. We can make this simpler separately if PPA of each service contains a known stable/unique deb pkg in our side discussions.

features/ubuntu_pro.feature
```diff
@@ -27,8 +27,46 @@ Feature: Command behaviour when attached to an UA subscription
             fips-updates  +<fips> +<fips-s> +Uncertified security updates to FIPS modules
             livepatch     +yes      +enabled  +Canonical Livepatch service
             """
+        When I run `apt-cache policy` with sudo
+        Then stdout matches regexp:
+        """
+        https://esm.ubuntu.com/infra/ubuntu <release>-infra-updates/main amd64 Packages
```
blackboxsw, line 33: We might want to add a @then('apt-cache policy for {URL} has permission {perm_id}') to check for -32768 vs 500

lucasmoura:
@blackboxsw Done. I have updated the scenario, I just slightly modified the your `apt policy`suggestion, but let me know if you don't agree with it

features/steps/steps.py
```diff
@@ -183,6 +183,24 @@ def then_i_should_see_that_the_command_is_not_found(context, cmd_name):
     assert_that(expected_return, equal_to(actual_return))
 
 
+@then("I verify that running `{cmd_name}` `{spec}` exits `{exit_codes}`")
+def then_i_verify_that_running_cmd_with_spec_succeeds(
```
blackboxsw, line 187:
```suggestion
def then_i_verify_that_running_cmd_with_spec_exits_with_codes(
```

lucasmoura:
@blackboxsw Updated

PR #1144, opened 2020-08-10
Improve esm tests for containers

lucasmoura:
Improve `esm` test by verifying if both `esm-infra` and `esm-apps` when enabled can download and install packages from those services.

PS: This PR is still not complete because the current contract token does not support enabling `esm-apps`

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -93,13 +93,26 @@ Feature: Enable command behaviour when attached to an UA subscription
             ESM Infra is already enabled.
             See: sudo ua status
             """
+        When I run `apt-cache policy` with sudo
+        Then apt-cache policy for the following url has permission `500`
+        """
+        <esm-infra-url> <release>-infra-updates/main amd64 Packages
+        """
+        And I verify that running `apt update` `with sudo` exits `0`
+        When I run `apt install -y <infra-pkg>` with sudo
+        And I run `apt-cache policy <infra-pkg>` as non-root
+        Then stdout matches regexp:
+        """
+        \s*500 <esm-infra-url> <release>-infra-security/main amd64 Packages
+        \s*500 <esm-infra-url> <release>-infra-updates/main amd64 Packages
+        """
 
         Examples: ubuntu release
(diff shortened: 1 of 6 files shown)


uaclient/exceptions.py
```diff
@@ -16,6 +16,22 @@ def __init__(self, msg: str) -> None:
...
+        Takes a single parameter, which is the beta service error message that
+        should be emitted before exiting non-zero.
+    """
+
+    exit_code = 1
+
+    def __init__(self, msg: str) -> None:
+        self.msg = msg
+
+
```
blackboxsw, line 34: This should subclass from UserFacingError right, then you also don't have to add/redefine exit_code = 1 as a class attr here right?

```suggestion
class BetaServiceError(Exception):
    """
    An exception to be raised trying to interact with beta service
    without the right parameters.

    :param msg:
        Takes a single parameter, which is the beta service error message that
        should be emitted before exiting non-zero.
    """
    pass


```

features/attached_enable.feature
```diff
@@ -93,13 +93,26 @@ Feature: Enable command behaviour when attached to an UA subscription
...
         Examples: ubuntu release
-           | release |
-           | bionic  |
-           | focal   |
-           | trusty  |
-           | xenial  |
+           | release | infra-pkg | esm-infra-url                       |
+           | bionic  | libkrad0  | https://esm.ubuntu.com/infra/ubuntu |
+           | focal   | hello     | https://esm.ubuntu.com/infra/ubuntu |
+           | trusty  | libgit2-0 | https://esm.ubuntu.com/ubuntu/      |
```
blackboxsw, line 114: I opened a pull request about this in ua-contracts. I *think* they should all be set the same, but we'll see what contracts team says 
https://github.com/CanonicalLtd/ua-contracts/issues/1115

lucasmoura: Great, if it lands, I will make another PR to remove that

lucasmoura:
@blackboxsw Updated the exception inheritance

PR #1148, opened 2020-08-13
Update help command

lucasmoura:
Update help command to allow services to passed to it. For example, if we run `ua help esm-infra`, the command will output:

```
name: esm-infra
entitled: yes
status: enabled
help: esm-infra help
```

If the uaclient is attached to a subscription. If not attached, the command will return:

```
name: esm-infra
available: yes
help: esm-infra help
```

And if an user pass a non-existing service, such as `ua help invalid-service`, the command will output:
```
name: invalid-service
error: could not find service: invalid-service
```

There are still some thins that need to be discussed regarding this feature:

* Difference in the output for attached/anattached uaclient
* How to handle beta services
* How the output will be for invalid services

Also, we still need to create the actual help messages for each service, but this PR creates the base structure for this feature.

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -292,3 +292,29 @@ Feature: Command behaviour when attached to an UA subscription
         """
         https://esm.ubuntu.com/ubuntu/ trusty-infra-updates/main amd64 Packages
         """
+
+    @series.all
+    Scenario Outline: Help command on an attached machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `ua help esm-infra` with sudo
+        Then I will see the following on stdout:
+            """
+            name: esm-infra
+            entitled: yes
+            status: enabled
+            help: esm-infra help
+            """
+        When I run `ua help invalid-service` with sudo
+        Then I will see the following on stdout:
+            """
+            name: invalid-service
+            error: could not find service: invalid-service
+            """
+
+        Examples: ubuntu release
+           | release |
+           | bionic  |
+           | focal   |
+           | trusty  |
+           | xenial  |
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -303,13 +303,20 @@ Feature: Command behaviour when attached to an UA subscription
             name: esm-infra
             entitled: yes
             status: enabled
-            help: esm-infra help
+            help:
+            esm-infra help
+            Information about esm-infra
+            """
+        When I run `ua help esm-infra --format json` with sudo
+        Then I will see the following on stdout:
+            """
+            {"name": "esm-infra", "entitled": "yes", "status": "enabled", "help": "esm-infra help\nInformation about esm-infra"}
             """
         When I run `ua help invalid-service` with sudo
         Then I will see the following on stdout:
             """
             name: invalid-service
-            error: could not find service: invalid-service
+            help: No help available for service "invalid-service"
             """
 
         Examples: ubuntu release
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -83,22 +83,28 @@ Feature: Command behaviour when unattached
            | xenial  | enable   | unknown   |
            | xenial  | disable  | unknown   |
 
-    @wip
     @series.all
-    Scenario Outline: Help command on an anattached machine
+    Scenario Outline: Help command on an unattached machine
         Given a `<release>` machine with ubuntu-advantage-tools installed
         When I run `ua help esm-infra` as non-root
         Then I will see the following on stdout:
             """
             name: esm-infra
             available: yes
-            help: esm-infra help
+            help:
+            esm-infra help
+            Information about esm-infra
+            """
+        When I run `ua help esm-infra --format json` with sudo
+        Then I will see the following on stdout:
+            """
+            {"name": "esm-infra", "available": "yes", "help": "esm-infra help\nInformation about esm-infra"}
             """
         When I run `ua help invalid-service` with sudo
         Then I will see the following on stdout:
             """
             name: invalid-service
-            error: could not find service: invalid-service
+            help: No help available for service "invalid-service"
             """
 
         Examples: ubuntu release
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -82,3 +82,28 @@ Feature: Command behaviour when unattached
            | xenial  | disable  | livepatch |
            | xenial  | enable   | unknown   |
            | xenial  | disable  | unknown   |
+
+    @wip
+    @series.all
+    Scenario Outline: Help command on an anattached machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `ua help esm-infra` as non-root
+        Then I will see the following on stdout:
+            """
+            name: esm-infra
+            available: yes
+            help: esm-infra help
+            """
+        When I run `ua help invalid-service` with sudo
+        Then I will see the following on stdout:
+            """
+            name: invalid-service
+            error: could not find service: invalid-service
+            """
+
+        Examples: ubuntu release
+           | release |
+           | bionic  |
+           | focal   |
+           | trusty  |
+           | xenial  |
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -142,23 +142,37 @@ def detach_parser(parser):
 
 def help_parser(parser):
     """Build or extend an arg parser for help subcommand."""
-    usage = USAGE_TMPL.format(name=NAME, command="help")
+    usage = USAGE_TMPL.format(name=NAME, command="help [service_name]")
     parser.usage = usage
-    parser.prog = "enable"
+    parser.prog = "help"
     parser._positionals.title = "Arguments"
     parser.add_argument(
-        "name",
+        "service_name",
         action="store",
-        nargs="*",
+        nargs="?",
         default="",
-        help="the name of the Ubuntu Advantage service to get help info",
+        help="a service to view help output for",
+    )
+
+    parser.add_argument(
+        "--format",
+        action="store",
+        choices=STATUS_FORMATS,
+        default=STATUS_FORMATS[0],
+        help=(
+            "output help in the specified format (default: {})".format(
+                STATUS_FORMATS[0]
+            )
+        ),
     )
     return parser
 
 
 def enable_parser(parser):
     """Build or extend an arg parser for enable subcommand."""
-    usage = USAGE_TMPL.format(name=NAME, command="enable") + " []"
+    usage = USAGE_TMPL.format(
+        name=NAME, command="enable <service_name> [<service_name>]"
+    )
     parser.usage = usage
     parser.prog = "enable"
     parser._positionals.title = "Arguments"
@@ -698,22 +712,25 @@ def action_refresh(args, cfg):
 
 
 def action_help(args, cfg):
-    name = args.name
-    if name == "":
+    service_name = args.service_name
+
+    if not service_name:
         get_parser().print_help()
         return 0
 
     if not cfg:
         cfg = config.UAConfig()
 
-    if type(name) == list:
-        # Right now, we are only supporting one service per help command
-        name = name[0]
+    help_response = cfg.help(service_name)
 
-    help_response = cfg.help(name)
-
-    for key, value in help_response.items():
-        print("{}: {}".format(key, value))
+    if args.format == "json":
+        print(json.dumps(help_response))
+    else:
+        for key, value in help_response.items():
+            if "\n" in str(value):
+                print("{}:\n{}".format(key, value))
+            elif value:
+                print("{}: {}".format(key, value))
 
     return 0
 
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -140,6 +140,22 @@ def detach_parser(parser):
     return parser
 
 
+def help_parser(parser):
+    """Build or extend an arg parser for help subcommand."""
+    usage = USAGE_TMPL.format(name=NAME, command="help")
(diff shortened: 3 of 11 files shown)


features/unattached_commands.feature
```diff
@@ -82,3 +82,28 @@ Feature: Command behaviour when unattached
            | xenial  | disable  | livepatch |
            | xenial  | enable   | unknown   |
            | xenial  | disable  | unknown   |
+
+    @wip
+    @series.all
+    Scenario Outline: Help command on an anattached machine
```
blackboxsw, line 88:
```suggestion
    Scenario Outline: Help command on an unattached machine
```

uaclient/cli.py
```diff
@@ -140,6 +140,22 @@ def detach_parser(parser):
...
+def help_parser(parser):
+    """Build or extend an arg parser for help subcommand."""
+    usage = USAGE_TMPL.format(name=NAME, command="help")
+    parser.usage = usage
+    parser.prog = "enable"
+    parser._positionals.title = "Arguments"
+    parser.add_argument(
+        "name",
+        action="store",
+        nargs="*",
```
blackboxsw, line 152: Only allow optional one positional arg here
```suggestion
        nargs="?",
```

uaclient/cli.py
```diff
@@ -140,6 +140,22 @@ def detach_parser(parser):
     return parser
 
 
+def help_parser(parser):
+    """Build or extend an arg parser for help subcommand."""
+    usage = USAGE_TMPL.format(name=NAME, command="help")
+    parser.usage = usage
+    parser.prog = "enable"
```
blackboxsw, line 147:
```suggestion
    parser.prog = "help"
```

uaclient/cli.py
```diff
@@ -140,6 +140,22 @@ def detach_parser(parser):
     return parser
 
 
+def help_parser(parser):
+    """Build or extend an arg parser for help subcommand."""
+    usage = USAGE_TMPL.format(name=NAME, command="help")
```
blackboxsw, line 145:
```suggestion
    usage = USAGE_TMPL.format(name=NAME, command="help [service_name]")
```

uaclient/cli.py
```diff
@@ -680,8 +697,24 @@ def action_refresh(args, cfg):
...
+        cfg = config.UAConfig()
+
+    if type(name) == list:
+        # Right now, we are only supporting one service per help command
+        name = name[0]
+
+    help_response = cfg.help(name)
+
+    for key, value in help_response.items():
+        print("{}: {}".format(key, value))
```
blackboxsw, line 716: I think we need to align with the spec for the help property output which is 
help:
<multiline_value>
<multiline_value>

how about if '\n' in value: then change output format to put the value on the next line.


* Also we probably want to avoid printing the value if empty/None in the tabular format as that just prints irrelevant info. Leave Null/None/empty values in json format output though

features/attached_commands.feature
```diff
@@ -292,3 +292,29 @@ Feature: Command behaviour when attached to an UA subscription
...
+    Scenario Outline: Help command on an attached machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `ua help esm-infra` with sudo
+        Then I will see the following on stdout:
+            """
+            name: esm-infra
+            entitled: yes
+            status: enabled
+            help: esm-infra help
```
mitechie, line 306: let's put help as a key: but can we newline the content and not have it start out indented here so? I'm thinking more like the multi-line examples here: https://yaml-multiline.info/

features/unattached_commands.feature
```diff
@@ -82,3 +82,28 @@ Feature: Command behaviour when unattached
...
+            """
+            name: esm-infra
+            available: yes
+            help: esm-infra help
+            """
+        When I run `ua help invalid-service` with sudo
+        Then I will see the following on stdout:
+            """
+            name: invalid-service
+            error: could not find service: invalid-service
```
mitechie, line 101: let's just remove the "name:" line in this output and only have the error

error: "invalid-service" is not a valid service.

uaclient/entitlements/base.py
```diff
@@ -38,6 +38,9 @@ class UAEntitlement(metaclass=abc.ABCMeta):
     # Wheter that entitlement is in beta stage
     is_beta = False
 
+    # Help information for the service
+    help_info = None  # type: str
```
blackboxsw, line 42: I think we should make this a class @property as we likely will need to source slightly more dynamic content from either a help JSON file on disk or potentially ua-contracts machine-token response that could be updated dynamically in the future. What we would lke is one simple place to update this help content if desired (it could even live in /var/lib/ubuntu-advantage/help.json). This way we don't have any significant binary changes if we have to update help text, or dynamically overwrite that help content if ua-contracts machine-token response starts providing us dynamic help content.

uaclient/config.py
```diff
@@ -405,6 +405,43 @@ def status(self, show_beta=False) -> "Dict[str, Any]":
...
+        response_dict["name"] = name
+
+        for resource in resources:
+            if resource["name"] == name:
+                help_resource = resource
+                help_ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(name)
+                break
+
+        if help_resource is None:
+            response_dict["error"] = "could not find service: {}".format(name)
```
mitechie, line 425: If there's no help content let's just output the rest of the YAML but have the help content be a default 

    No help available for service "xxx"

uaclient/entitlements/base.py
```diff
@@ -38,6 +38,9 @@ class UAEntitlement(metaclass=abc.ABCMeta):
     # Wheter that entitlement is in beta stage
     is_beta = False
```
mitechie, line 39: let's add the "is_beta" but only if it's enabled and it's a feature that needs --beta on the cli to enable it please.

uaclient/cli.py
```diff
@@ -140,6 +140,22 @@ def detach_parser(parser):
...
 
 
+def help_parser(parser):
+    """Build or extend an arg parser for help subcommand."""
+    usage = USAGE_TMPL.format(name=NAME, command="help")
+    parser.usage = usage
+    parser.prog = "enable"
+    parser._positionals.title = "Arguments"
+    parser.add_argument(
+        "name",
```
blackboxsw, line 150: * let's call this service_name to be clearer in help docs
* Also let's add a --format option per the spec to support json output.

uaclient/entitlements/esm.py
```diff
@@ -26,6 +26,7 @@ class ESMInfraEntitlement(ESMBaseEntitlement):
     title = "ESM Infra"
     description = "UA Infra: Extended Security Maintenance"
     repo_key_file = "ubuntu-advantage-esm-infra-trusty.gpg"
+    help_info = "esm-infra help"
```
mitechie, line 29: Can we load this from a .yaml file or something that we load this from so that we can give a single file of content for the list of features to product management and have them submit something back easily?

uaclient/entitlements/base.py
lucasmoura: Makes total sense. I will update the code

lucasmoura:
@blackboxsw and @mitechie I have updated the code. Thanks for the review :)

uaclient/cli.py
```diff
@@ -140,9 +140,39 @@ def detach_parser(parser):
...
+    usage = USAGE_TMPL.format(name=NAME, command="help [service_name]")
+    parser.usage = usage
+    parser.prog = "help"
+    parser._positionals.title = "Arguments"
+    parser.add_argument(
+        "service_name",
+        action="store",
+        nargs="?",
+        default="",
+        help="the name of the Ubuntu Advantage service to get help info",
```
mitechie, line 154: Just a wording nit:

    a service to view help output for

We should double-check the help output in the status command as well. Over time I believe we'll grow things that are not Ubutun Advantage services and so I think there's value in being more generic in places like this.

uaclient/cli.py
```diff
@@ -680,8 +711,27 @@ def action_refresh(args, cfg):
...
+    if not service_name:
+        get_parser().print_help()
+        return 0
+
+    if not cfg:
+        cfg = config.UAConfig()
+
+    help_response = cfg.help(service_name)
+
+    if args.format == "json":
```
mitechie, line 726: We need to file a bug to make sure YAML is a supported format option. It's pretty standard in Canonical and we're going to want to focus on that yet still have json as a parsable option.

uaclient/config.py
```diff
@@ -4,7 +4,7 @@
 import logging
 import os
 import yaml
-from collections import namedtuple
+from collections import namedtuple, OrderedDict
```
mitechie, line 7: I assume it's ordered dict so that the help output comes out in the right order. It might be worth highlighting that at the right place so that doesn't get changed around and is a documented decision.

uaclient/entitlements/esm.py
```diff
@@ -38,3 +38,8 @@ def repo_pin_priority(self) -> "Optional[str]":
     def disable_apt_auth_only(self) -> bool:
         """Only trusty esm-infra should remove apt auth files upon disable"""
         return bool("trusty" == util.get_platform_info()["series"])
+
+    @property
+    def help_info(self) -> str:
+        """Help information for the entitlement"""
+        return "esm-infra help\nInformation about esm-infra"
```
mitechie, line 45: Are we going to move the content to a non-python file as a followup then? I expect this content to be multiple paragraphs, with links, etc and making it more trivial to update is definitely the goal.

uaclient/config.py
lucasmoura: Yes, that's exactly why I am using it. I will add comment for that reasoning

uaclient/entitlements/esm.py
lucasmoura: II though the idea was to do something like the `machine_token` property, where we parse a json file and obtain the information that we need:

https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/config.py#L134

Since we still do not have the source of that information or the format that it will come, I have not made any assumptions on that part of the code. But, if you think there are some assumptions that we can code right now, just let me know and I can do that without an issue

uaclient/cli.py
lucasmoura: I think I can add this to this PR, I don't think that should be pose any difficulty

mitechie: that's ok, let's do it in follow up because we need to make sure it works and makes sense for all the commands like status/etc.

uaclient/entitlements/esm.py
mitechie: I think asking lech to write out json is a bit much. I was hoping either YAML or individual text files perhaps? 

Maybe we can have a help.py that's just a simple dict like:

    help['esm-infra'] = """
    <Lech to fill all this out here>
    """
    help['esm-apps'] = """
    <lech to fill out>
    """

lucasmoura: I think both ideas work, I would prefer to have an YAML, but that may cause the problem that if we update that YAML and break it, we would make all of the service's help command to fail. But we can discuss that.

But yes, I can create a help.py file with that dummy info for now and read it from there

lucasmoura:
@mitechie I have updated the code

uaclient/entitlements/esm.py
blackboxsw: @lucasmoura @mitechie if we don't like JSON for non-programming types, I'd really suggest that writing a python dict is likely an equivalent barrier to entry for these strings. Let's go YAML if we really don't like JSON.

blackboxsw: @lucasmoura since we really shouldn't be updating this yaml on the fly at the moment, I don't expect us to be too concerned with breaking help. But, maybe it does make sense to return a "no help available" string if we struggle to correctly parse the help.yaml file. I *think* that file should live in /var/lib/ubuntu-advantage/help.yaml. maybe in /etc/ubuntu-advantage (but filesystem heirarchy standard says that /etc is for "configuration" files, which this is not really).

features/attached_commands.feature
```diff
@@ -292,3 +292,37 @@ Feature: Command behaviour when attached to an UA subscription
         """
         https://esm.ubuntu.com/ubuntu/ trusty-infra-updates/main amd64 Packages
         """
+
+    @wip
```
blackboxsw, line 296: let's drop the @wips

uaclient/entitlements/base.py
```diff
@@ -38,6 +41,9 @@ class UAEntitlement(metaclass=abc.ABCMeta):
...
 
+    # Help info message for the entitlement
+    _help_info = None  # type: str
```
blackboxsw, line 45: I think we can drop this as caching the value doesn't make sense given that this will only be used via a single cmdline call at the moment and this cache would be set up each time someone calls ua help <blah>

uaclient/entitlements/base.py
```diff
@@ -56,6 +62,20 @@ def description(self) -> str:
...
+            if os.path.exists(DEFAULT_HELP_FILE):
+                with open(DEFAULT_HELP_FILE, "r") as f:
+                    help_dict = yaml.load(f, Loader=yaml.FullLoader)
+            else:
+                help_dict = DEFAULT_HELP
+
+            self._help_info = help_dict.get(self.name, {}).get("help", "")
+
+        return self._help_info
+
```
blackboxsw, line 78: Probably no need to cache as mentioned above.
Let's just read the DEFAULT_HELP_FILE and parse it if it exists. We can follow up on this, with our next PR that actually adds help content, but I think it's probably unnecessary to seed DEFAULT_HELP and rely on a DEFAULT_HELP_FILE. We should go with one or the other, not two ways of doing the same thing. Again, I think we can resolve this in the next PR when we actually update the help content for some of these services.

Issue #1128, opened 2020-07-22
tests: integration tests on motd/update-notifier behavior on T, X, B, F


blackboxsw:
we need coverage of update-notified behavior related to messaging in motd.

PR #1149, opened 2020-08-13
tests: add motd update-notifier integration tests for esm

blackboxsw:
Add validation of current state of trusty, xenial, bionic and focal
update-notifier-common behavior when seeding motd messages.

Currently Trusty advertizes ESM attach status and available upgrades
before attach.

Xenial and later do not advertize ESM when unattached.
Trusty and Focal report ESM attach status when attached, while X and B
do not.

Fixes: #1128


features/attach_validtoken.feature
```diff
@@ -6,6 +6,31 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
+
+        \d+ updates can be installed immediately.
+        \d+ of these updates are provided through UA Infrastructure ESM.
+        \d+ of these updates are security updates.
+        To see these additional updates run: apt list --upgradable
+
+        Enable UA Infrastructure ESM to receive \d+ additional security updates.
+        See https://ubuntu.com/advantage or run: sudo ua status
+        """
+        Then if `<release>` in `xenial||bionic` and stdout matches regexp:
```
lucasmoura, line 23: Since we are parsing the `||` in the step definition, maybe we could use instead `or` to making reading the feature more natural

features/steps/steps.py
```diff
@@ -134,6 +134,16 @@ def then_i_will_see_on_stdout(context):
     assert_that(context.process.stdout.strip(), equal_to(context.text))
 
 
+@then("if `{value1}` in `{value2}` and stdout matches regexp")
+def then_conditional_stdout_matches_regexp(context, value1, value2):
+    """Only apply regex assertion if value1 in value2."""
+    if value1 in value2.split("||"):
+        assert_that(
```
lucasmoura, line 141: I think we can just call the `then_stdout_matches_regexp`  function here

features/attach_validtoken.feature
```diff
@@ -26,13 +51,35 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
+        UA Infrastructure Extended Security Maintenance \(ESM\) is enabled.
 
+        \d+ updates can be installed immediately.
+        \d+ of these updates are provided through UA Infrastructure ESM.
+        \d+ of these updates are security updates.
+        To see these additional updates run: apt list --upgradable
+        """
+        Then if `<release>` in `xenial||bionic` and stdout matches regexp:
+        """
+        \d+ packages can be updated.
```
lucasmoura, line 66: I if missed something in the discussion here, but why are xenial and bionic not fully covered by ESM ?

Also, is the check regarding that we should only send a message if there is a package to be upgraded be delivered in a separate PR ?

features/steps/steps.py
blackboxsw: ohh right oops, meant to and then forgot +1

features/attach_validtoken.feature
blackboxsw: I was thinking we would need a separate PR to validate whether or not we see the message if there are no esm packages, but maybe we should just refactor this PR to do that as I would like to close out the ESM motd issue fully.

features/attach_validtoken.feature
blackboxsw: done.

features/attach_validtoken.feature
blackboxsw: Ahh ok, so this is this issue that I just filed to track it. There has been limited SRUing of update-notifier functionality which would surface ESM-aware apt_check.py --human-readable. 
We will probably have to shepherd this work and maybe work with either andreas or bryce on pkg upload sponsorship.

Issue #1127, opened 2020-07-22
tests: add integration test that apt hook commandline behaves as expected


blackboxsw:
integration test should confirm that apt package counts reported on the apt CLI are expected for esm infra and esm apps counts on a recently attached machine.

The expectation is that `apt update` or `aptitude upgrade` on a system with ESM enabled will show upgradable package counts from esm

Expected steps to reproduce:
```
 # launch focal vm
  # ua attach 
  # install the public hello deb version `apt-get install hello=2.10-2ubuntu2`
  # apt-cache policy hello  (See that a newer version from esm infra updates is available
  # apt update   # should have output line reporting esm upgradeable pkg count as 1
```

PR #1153, opened 2020-08-14
Add apt update notify integration test

lucasmoura:
Add integration test to check that when running apt update, esm-infra updates will appear in the output of the command.

This PR address #1127

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -114,6 +114,39 @@ Feature: Enable command behaviour when attached to an UA subscription
            | trusty  | libgit2-0 | https://esm.ubuntu.com/ubuntu/      |
            | xenial  | libkrad0  | https://esm.ubuntu.com/infra/ubuntu |
 
+    @wip
+    @series.xenial
+    @series.bionic
+    @series.focal
+    Scenario Outline: Attached enable of a know service shows update in a ubuntu machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `ua enable esm-infra` with sudo
+        Then I will see the following on stdout:
+            """
+            One moment, checking your subscription first
+            ESM Infra is already enabled.
+            See: sudo ua status
+            """
+        When I run `apt install -y <pkg-version>` with sudo
+        And I run `apt update` with sudo
+        Then stdout matches regexp
+        """
+        \d+ of the updates (is|are) from UA Infrastructure ESM
+        """
+        When I run `ua disable esm-infra` with sudo
+        And I run `apt update` with sudo
+        Then stdout does not match regexp
+        """
+        \d+ of the updates (is|are) from UA Infrastructure ESM
+        """
+
+        Examples: ubuntu release
+           | release | pkg-version                |
+           | bionic  | libkrad0=1.16-2ubuntu0.1   |
+           | focal   | hello=2.10-2ubuntu2        |
+           | xenial  | libkrad0=1.16-2ubuntu0.1   |
+
     @series.all
     @uses.config.machine_type.lxd.container
     Scenario Outline: Attached enable a disabled, enable and unknown service in a ubuntu machine
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -3,7 +3,7 @@ import subprocess
 import shlex
 
 from behave import given, then, when
-from hamcrest import assert_that, equal_to, matches_regexp
+from hamcrest import assert_that, equal_to, matches_regexp, not_
 
 from features.environment import create_uat_image
 from features.util import (
@@ -137,6 +137,13 @@ def then_i_will_see_on_stdout(context):
 @then("stdout matches regexp")
 def then_stdout_matches_regexp(context):
     assert_that(context.process.stdout.strip(), matches_regexp(context.text))
+
+
+@then("stdout does not match regexp")
+def then_stdout_does_not_match_regexp(context):
+    assert_that(
+        context.process.stdout.strip(), not_(matches_regexp(context.text))
+    )
 
 
 @then("stderr matches regexp")


features/attached_enable.feature
```diff
@@ -114,6 +114,32 @@ Feature: Enable command behaviour when attached to an UA subscription
...
+    @series.focal
+    Scenario Outline: Attached enable of a know service shows update in a ubuntu machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `ua enable esm-infra` with sudo
+        Then I will see the following on stdout:
+            """
+            One moment, checking your subscription first
+            ESM Infra is already enabled.
+            See: sudo ua status
```
mitechie, line 128: do we need sudo in this status call? (not for this PR but sounds like a bug in our output/suggested help)

features/attached_enable.feature
```diff
@@ -114,6 +114,32 @@ Feature: Enable command behaviour when attached to an UA subscription
...
+            """
+        When I run `apt install -y <pkg-version>` with sudo
+        And I run `apt update` with sudo
+        Then stdout matches regexp
+        """
+        \d* of the updates \w* from UA Infrastructure ESM
+        """
```
mitechie, line 135: this \d is going to be 0 for all but trusty right? Does this test do much then? Should it be testing the number on those just to note if we get a different answer we should be looking into it?

lucasmoura: Originally, I though that was the case. On focal that's exactly what is happening, but for bionic, just by enable `esm-infra`, we have 6 packages that appear in the esm update output. That's why I have used the regex. But I did not check if xenial would follow the same pattern

features/attached_enable.feature
lucasmoura: That's a good point, I don't think we need that message. Do you remember with there is any context regarding that @blackboxsw ?

features/attached_enable.feature
mitechie: ah ok, interesting. Can you let me know what those were? I'm curious if I'm mistaken that there's no value in esm-infra for bionic/xenial/focal after all then.

lucasmoura: No problem, I will reproduce the scenario on Bionic and Xenial and post the packages here

lucasmoura: Bionic:
```
krb5-locales/bionic-infra-security,bionic-infra-updates 1.16-2ubuntu0.1+esm1 all [upgradable from: 1.16-2ubuntu0.1]
libgssapi-krb5-2/bionic-infra-security,bionic-infra-updates 1.16-2ubuntu0.1+esm1 amd64 [upgradable from: 1.16-2ubuntu0.1]
libk5crypto3/bionic-infra-security,bionic-infra-updates 1.16-2ubuntu0.1+esm1 amd64 [upgradable from: 1.16-2ubuntu0.1]
libkrb5-3/bionic-infra-security,bionic-infra-updates 1.16-2ubuntu0.1+esm1 amd64 [upgradable from: 1.16-2ubuntu0.1]
libkrb5support0/bionic-infra-security,bionic-infra-updates 1.16-2ubuntu0.1+esm1 amd64 [upgradable from: 1.16-2ubuntu0.1]
```

Xenial:
```
krb5-locales/xenial-infra-security,xenial-infra-updates 1.13.2+dfsg-5ubuntu2.1+esm1 all [upgradable from: 1.13.2+dfsg-5ubuntu2.1]
libgssapi-krb5-2/xenial-infra-security,xenial-infra-updates 1.13.2+dfsg-5ubuntu2.1+esm1 amd64 [upgradable from: 1.13.2+dfsg-5ubuntu2.1]
libk5crypto3/xenial-infra-security,xenial-infra-updates 1.13.2+dfsg-5ubuntu2.1+esm1 amd64 [upgradable from: 1.13.2+dfsg-5ubuntu2.1]
libkrb5-3/xenial-infra-security,xenial-infra-updates 1.13.2+dfsg-5ubuntu2.1+esm1 amd64 [upgradable from: 1.13.2+dfsg-5ubuntu2.1]
libkrb5support0/xenial-infra-security,xenial-infra-updates 1.13.2+dfsg-5ubuntu2.1+esm1 amd64 [upgradable from: 1.13.2+dfsg-5ubuntu2.1]
```

mitechie: interesting ok. It looks like it's the same version that's in main already for those just without the +esm on the version string. Maybe the security team was testing something. They're in the list though so I guess it's not 0.

features/attached_enable.feature
blackboxsw: We don't need sudo for `ua status` specifically, but when non-sudo runs of `ua status` rely on reading the cached status file @ /var/lib/ubuntu-advantage/status.json. So, if something specifically had changed state (like livepatch being disabled on the system, or rebooting into the proper hwe kernel for livepatch support on trusty) the status will lie to the non-root user. 
I think that was the reason we left 'sudo' on the status calls in messaging (there are 4 other sudo ua status messages in uaclient/status.py). 
It's not strictly required though... just that it current has caching caveats for infrequent corner cases

mitechie: Ah interesting ok. Maybe that's a bug then we can see if we can change or have the cache file permissions more free so that it's not necessary.

features/attached_enable.feature
```diff
@@ -114,6 +114,39 @@ Feature: Enable command behaviour when attached to an UA subscription
...
+            """
+            One moment, checking your subscription first
+            ESM Infra is already enabled.
+            See: sudo ua status
+            """
+        When I run `apt install -y <pkg-version>` with sudo
+        And I run `apt update` with sudo
+        Then stdout matches regexp
+        """
+        \d* of the updates \w* from UA Infrastructure ESM
```
blackboxsw, line 135: this is a more permissive regex than mine (is|are). I think it's probably better to be strict if we can (also because it's potentially easier to read what that expected message should be.

features/steps/steps.py
```diff
@@ -139,6 +139,13 @@ def then_stdout_matches_regexp(context):
     assert_that(context.process.stdout.strip(), matches_regexp(context.text))
 
 
+@then("stdout does not match regexp")
```
blackboxsw, line 142: good add once your branch lands I'll rebase mine on master and use this step too

lucasmoura:
@blackboxsw Updated

features/attached_enable.feature
blackboxsw: @mitechie right. The issue that status has a heavily couple dependence on "machine-token.json" which contains authkeys that should only be presented to root. Also, any apt-service uses auth tokens stored in /etc/apt/auth.conf,d which is also only root-readable. So even checking entitlement_status as non-root would lie to a non-root user about whether the credentials to the PPAs are enabled because I think apt reports the ppa as not having the right credentials. So, we'd have to tweak how uaclient confirms enablement of apt-based entitlements to "work around" lacking root credentials for an apt-based service.

PR #1165, opened 2020-08-25
Dist upgrade test

lucasmoura:
Add script that allow us to test upgrading from one release to another and test the effects regarding uaclient. Right now we are testing if uaclient was updated and if enabled, the functionality keeps working as expected, but we can improve the checking in that script without an issue

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -599,9 +599,12 @@ def _install_uat_in_container(
     """
     cmds = []
     if not deb_paths:
-        deb_names = (
-            " ".join(UA_DEBS) if cloud_api else "ubuntu-advantage-tools"
+        pkg_names = (
+            [deb.replace(".deb", "") for deb in UA_DEBS]
+            if cloud_api
+            else ["ubuntu-advantage-tools"]
         )
+
         cmds.extend(
             [
                 [
@@ -611,7 +614,7 @@ def _install_uat_in_container(
                     "ppa:canonical-server/ua-client-daily",
                 ],
                 ["sudo", "apt-get", "update", "-qq"],
-                ["sudo", "apt-get", "install", "-qq", "-y", deb_names],
+                ["sudo", "apt-get", "install", "-qq", "-y"] + pkg_names,
             ]
         )
     else:
--- a/sru/release-25/dist-upgrade-test.sh
+++ b/sru/release-25/dist-upgrade-test.sh
@@ -1,0 +1,163 @@
+#!/bin/bash
+
+
+cat >install-24.3-uat.yaml <<EOF
+#cloud-config
+ssh_import_id: [lamoura]
+apt:                                                                            
+  sources:                                                                      
+    testingpro:
+      source: "ppa:chad.smith/test-ua-client-do-release-upgrade"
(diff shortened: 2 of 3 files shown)


features/environment.py
```diff
@@ -599,9 +599,12 @@ def _install_uat_in_container(
     """
     cmds = []
     if not deb_paths:
-        deb_names = (
-            " ".join(UA_DEBS) if cloud_api else "ubuntu-advantage-tools"
+        pkg_names = (
+            [deb.replace(".deb", "") for deb in UA_DEBS]
+            if cloud_api
```
blackboxsw, line 604: I don't think this is quite correct because cloud_api is for both awsgeneric and awspro test targets. We don't want to add ubuntu-advantage-pro deb on awsgeneric or azuregeneric machine_types

sru/release-25/dist-upgrade-test.sh
```diff
@@ -0,0 +1,163 @@
+#!/bin/bash
+
+
+cat >install-24.3-uat.yaml <<EOF
+#cloud-config
+ssh_import_id: [lamoura]
+apt:                                                                            
+  sources:                                                                      
+    testingpro:
+      source: "ppa:chad.smith/test-ua-client-do-release-upgrade"
```
blackboxsw, line 10: We should be able to just use the daily ppa instead of my personal test ppa once #1172  merges. So let's wait on that landing to address this.

sru/release-25/dist-upgrade-test.sh
```diff
@@ -0,0 +1,163 @@
...
+    echo "##############################"
+    echo "Attach uaclient to a subscription"
+    lxc exec $VM_NAME ua attach $CONTRACT_TOKEN
+    echo "##############################"
+}
+
+function upgrade_release(){
+    echo "##############################"
+    echo "do-release-upgrade $BASE_RELEASE -> $UPGRADE_RELEASE"
+    echo "Call do-release-upgrade with AllowThirdPart=yes for PPA upgrades"
```
blackboxsw, line 78:
```suggestion
    echo "Call do-release-upgrade with AllowThirdParty=yes for PPA upgrades"
```

sru/release-25/dist-upgrade-test.sh
```diff
@@ -0,0 +1,163 @@
...
+    lxc exec $VM_NAME -- sudo mkdir -p /etc/update-manager/release-upgrades.d                            
+    lxc exec $VM_NAME -- sh -c "echo '[Sources]\nAllowThirdParty=yes' > allow.cfg"
+    lxc exec $VM_NAME -- sudo mv allow.cfg /etc/update-manager/release-upgrades.d
```
blackboxsw, line 81: Can we do this part in our #cloud-config with something like 
```
#cloud-config
apt_upgrade: true
apt_update: true
write_files:
  - path: /etc/update-manager/release-upgrades.d/allow.cfg
    permissions : '0644'
    content: |
      [Sources]
      AllowThirdParty=yes
```

lucasmoura:
Closing this #PR as we already have that manual test captured on a behave scenario. Also, the changes on `environment.py` are no longer needed

PR #1167, opened 2020-08-26
Do not unconfigure esm on postinst

lucasmoura:
Since we plan to release uaclient on `xenial` and `bionic`, we will now be supporting upgrading between releases. The idea here is that, if an user is already subscribed to uaclient and attached to some services, the upgrade should keep that and update the services to the new version.

However, currently we are `unconfiguring esm` when we do a postinst, which creates a conflict with our idea to support upgrading between releases. This PR removes that from postinst.

Furthermore, regarding `postrm`, it appears that we can remove the call for `rm -f /etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list` since this seems to be handled by this part of [prerm](https://github.com/canonical/ubuntu-advantage-client/blob/master/debian/prerm#L17) and also because this part is only covering `trusty`

PS: Also, I could not found a reason for us to the `configure_esm` function here, since we seem not to be checking if the user is attached to a subscription or even if he is entitled to `esm`  services. If that is the case. maybe we can drop it as well. But I could be missing something here.

This is the part of the code I am talking about:
https://github.com/canonical/ubuntu-advantage-client/blob/master/debian/postinst#L116

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -118,9 +118,6 @@ case "$1" in
           # 14.04 and unsupported arch
           unconfigure_esm
         fi
-      else
-        # not 14.04, regardless of arch
-        unconfigure_esm
       fi
       if [ ! -f /var/log/ubuntu-advantage.log ]; then
           touch /var/log/ubuntu-advantage.log
--- a/debian/postinst
+++ b/debian/postinst
@@ -5,12 +5,14 @@ set -e
 . /etc/os-release  # For VERSION_ID
 
 
+MYRELEASE=$(lsb_release -sc)
 APT_TRUSTED_KEY_DIR="/etc/apt/trusted.gpg.d"
 UA_KEYRING_DIR="/usr/share/keyrings/"
 
 ESM_INFRA_KEY_TRUSTY="ubuntu-advantage-esm-infra-trusty.gpg"
 
 APT_SRC_DIR="/etc/apt/sources.list.d"
+APT_PREFERENCES_DIR="/etc/apt/preferences.d"
 ESM_APT_SOURCE_FILE_PRECISE="$APT_SRC_DIR/ubuntu-esm-precise.list"
 ESM_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-trusty.list"
 ESM_INFRA_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-infra-trusty.list"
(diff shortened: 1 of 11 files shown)


debian/postrm
```diff
@@ -19,7 +19,6 @@ remove_logs(){
 remove_esm(){
     # config files created in postinst, need explicit handling on purge
     rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
-    rm -f /etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list
```
blackboxsw, line 22: So this remove_esm was trusty only in nature because it because historically it was defined that a customer would have to ua attach again on the new Ubuntu Release, plus the gpg keys used in trusty repos were to be different than those used in xenial+. Those assumptions have changed a little as we now have to handle upgrade path from trusty -> focal and remain attached.

What I think this means is that these purge functions need to be flexible enough to handle any apt config and prefs for any ua client service on any series. So 

- [ ] I think we need to generalize these file matches a bit to remove apt config for any entitled UA service upon purge. 

- [ ] Also I think we need to avoid encoding the ubuntu series in apt config files https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/entitlements/repo.py#L34-L45.

- [ ] We need to adapt debian/postinst to rename apt trusty|xenial|bionic|focal.list and preference files one more time to remove the encoded <release> name from the apt config files.

lucasmoura:
@blackboxsw I have updated the code. Also, maybe we can move the `remove_esm` function from `postrm` and add it to `prerm`, since it is now just removing the `gpg` keys.

debian/postinst
```diff
@@ -5,16 +5,28 @@ set -e
 . /etc/os-release  # For VERSION_ID
 
 
+MYRELEASE=$(lsb_relesase -sc)
```
blackboxsw, line 8: typo

lucasmoura:
@blackboxsw Thanks, that is a great improvement in the code

debian/postrm
```diff
@@ -19,8 +19,6 @@ remove_logs(){
...
-    rm -f /etc/apt/preferences.d/ubuntu-esm-infra-trusty
```
blackboxsw, line 23: My former comment was completely incorrect. 

remove_esm should either be renamed remove_postinst_created_files.  It is only there to remove any untracked package files that were created by debian/postinst (not necessarily due to ua enable <service>).
Trusty is the only case where this happens because ua-tools pkg pre-enables visibility to esm files (unauthenticated).

So, these removes should still be here. All other apt files are removed automatically due in debian/prerm calling clean_apt_files().

Also there is a bug with prerm #1170 that needs fixing separately

```suggestion
    rm -f /etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list
    rm -f /etc/apt/preferences.d/ubuntu-esm-infra-trusty
```

lucasmoura:
@blackboxsw I have updated the `postrm` script and also addressed the `prerm` issue you raised

blackboxsw:
> prerm

So prerm happens before the package removal, I think some of the clean_apt_files could and should be handling this (which we already call in prerm). We actually could do these removes directly in prerm too without adverse affects. If you want to move them there that would be reasonable as I don't think the apt_clean_files function expects those files to actually be there for any services. (it should actually be removing them)

PR #1175, opened 2020-09-01
Apply contract deltas during do-release-upgrade

lucasmoura:
Right now, when we perform a dist-upgrade, we are not taking in consideration some contract diffs that may appear between releases. For example, the aptURL for esm-infra on trusty is `https://esm.ubuntu.com/ubuntu `, while on xenial is `https://esm.ubuntu.com/infra/ubuntu`. If we do not take this aspects into considerations. we may live the system in a inconsistent state after performing a `do-release-upgrade`;

This PR creates the necessary code to apply those changes in the `postinst` script.

PS: I am testing this now in a real upgrade path. However, while I wait for those results, I think the PR can already be reviewed


debian/postinst
```diff
@@ -27,6 +27,38 @@ SYSTEMD_WANTS_AUTO_ATTACH_LINK="/etc/systemd/system/multi-user.target.wants/ua-a
...
+   new_entitlements=new_entitlements,
+   allow_enable=True
+)
+"
+}
+
+get_release_from_pkg_version(){
+    pkg_version=$1
+
+    if [[ $pkg_version =~ "14.04" ]]; then
```
blackboxsw, line 52: Could probably do this in a case statement (which makes me feel "safer" that string comparisons in POSIX shell as we already have examples of case/esac matching).

debian/postinst
```diff
@@ -27,6 +27,38 @@ SYSTEMD_WANTS_AUTO_ATTACH_LINK="/etc/systemd/system/multi-user.target.wants/ua-a
...
+        echo "trusty"
+    elif [[ $pkg_version =~ "16.04" ]]; then
+        echo "xenial"
+    elif [[ $pkg_version =~ "18.04" ]]; then
+        echo "bionic"
+    elif [[ $pkg_version =~ "20.04" ]]; then
+        echo "focal"
+    fi
```
blackboxsw, line 60: Could probably do this in a case statement (which makes me feel "safer" that string comparisons in POSIX shell as we already have examples of case/esac matching). We also should have a catch all if we don't find a matching known version.

```suggestion
    case "$pkg_version" in
         *14.04*) echo focal;;
         *16.04*) echo xenial;;
         *18.04*) echo bionic;;
         *20.04*) echo focal;;
         *) echo "unknown-previous-release";;
    esac
```

uaclient/contract.py
```diff
@@ -189,7 +189,54 @@ def _get_platform_data(self, machine_id):
...
-def process_entitlement_delta(orig_access, new_access, allow_enable=False):
+def process_entitlements_delta(
+    past_entitlements, new_entitlements, allow_enable
+):
+    """Iterate over all entitlements in new_entitlement and apply any delta
+    found according to past_entitlements.
+
+    :param past_entitlements: dict containing the last valid information
+        regarding service entitlements.
+    :param new_entitlements: dic containing the current information regarding
```
blackboxsw, line 200:
```suggestion
    :param new_entitlements: dict containing the current information regarding
```

uaclient/contract.py
```diff
@@ -189,7 +189,54 @@ def _get_platform_data(self, machine_id):
...
+    if unexpected_error:
+        raise exceptions.UserFacingError(status.MESSAGE_UNEXPECTED_ERROR)
+    elif delta_error:
+        raise exceptions.UserFacingError(
+            status.MESSAGE_ATTACH_FAILURE_DEFAULT_SERVICES
+        )
+
+
+def process_entitlement_delta(
+    orig_access, new_access, allow_enable=False, series_overrides=True
```
blackboxsw, line 238: let's add mypy type hints here too.

debian/postinst
```diff
@@ -27,6 +27,38 @@ SYSTEMD_WANTS_AUTO_ATTACH_LINK="/etc/systemd/system/multi-user.target.wants/ua-a
...
+
+    python3 -c "
+from uaclient.config import UAConfig
+from uaclient.contract import process_entitlements_delta
+
+past_entitlements=UAConfig(series='$prev_release').entitlements
+new_entitlements=UAConfig(series='$curr_release').entitlements
+
+process_entitlements_delta(
+   past_entitlements=past_entitlements,
```
blackboxsw, line 42: Good parameter renames, much clearer.

debian/postinst
```diff
@@ -27,6 +27,38 @@ SYSTEMD_WANTS_AUTO_ATTACH_LINK="/etc/systemd/system/multi-user.target.wants/ua-a
 SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enabled/ua-auto-attach.service.dsh-also"
 SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
 
+update_files_based_on_contract_delta() {
+  prev_release=$1
+  curr_release=$2
+
+    python3 -c "
+from uaclient.config import UAConfig
```
blackboxsw, line 35:
```suggestion
import logging

from uaclient.cli import setup_logging
from uaclient.config import UAConfig
```

debian/postinst
```diff
@@ -27,6 +27,38 @@ SYSTEMD_WANTS_AUTO_ATTACH_LINK="/etc/systemd/system/multi-user.target.wants/ua-a
...
+from uaclient.contract import process_entitlements_delta
+
+past_entitlements=UAConfig(series='$prev_release').entitlements
+new_entitlements=UAConfig(series='$curr_release').entitlements
+
```
blackboxsw, line 40:
```suggestion
```suggestion
setup_logging(logging.INFO, logging.DEBUG)
```

blackboxsw (changes requested):
Can we setup_logging in the postinst so we actually emit these contract config  changes to /var/log/ubuntu-advantage.log. Then we can more easily confirm that this is working.
I was able to test a sample script running the same content as postinst with:
 update_files_based_on_contract_delta trusty xenial
And it properly reconfigured the apt config files. Well done. 
I need to get through a do-release-upgrade to confirm no other nonsense is disabling the repo across upgrade paths.

debian/postinst
```diff
@@ -141,6 +153,11 @@ case "$1" in
           fi
       fi
 
+      # UA service PPAs support all ubuntu releases, no need to
+      # specialize apt config filenames per ubuntu release.
+      redact_ubuntu_release_from_ua_apt_filenames $APT_SRC_DIR
+      redact_ubuntu_release_from_ua_apt_filenames $APT_PREFERENCES_DIR
```
blackboxsw, line 159: This didn't seem to rename the expected /etc/apt/preferences.d/ubuntu-esm-infra-trusty -> /etc/apt/preferences.d/ubuntu-esm-infra. We have a bug to sort here as that "trusty" preference file is still left intact after a ua attach <token>

debian/postinst
```diff
@@ -27,12 +27,24 @@ SYSTEMD_WANTS_AUTO_ATTACH_LINK="/etc/systemd/system/multi-user.target.wants/ua-a
 SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enabled/ua-auto-attach.service.dsh-also"
 SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
 
+get_release_from_pkg_version(){
```
blackboxsw, line 30: We may be able to drop this function now that the upgrade script is outside of postinst

blackboxsw (changes requested):
minor comments, we have a couple bugs to fix I think

blackboxsw (commented):
I think we have another 2 bugs:

1. you'll need to chmod 755 process-contract-deltas.py and add the header `#!/usr/bin/env python3` 
otherwise it doesn't get called by do-release-upgrade, and sinces it is just called with subprocess.call, it wouldn't run without the python3 header.

2.  in that the contract attach is applying contract deltas for all services (entitled or not) when we compare static contract deltas. results in spitting out empty apt configuration files for unentitled and inactive services just on ua attach <mytoken> on trusty (no do-release-upgrade required).

this may be a regression in your current branch. I'll check tip of master tomorrow.

```bash
# after ua attach <mytoken> on unattached system
Updating 'cc-eal' apt sources list on changed directives.
Updating package lists
ADD AUTH APT REPO
MY CONTENT
content:  
ADD AUTH APT REPO
MY CONTENT
content:  deb https://esm.ubuntu.com/ubuntu trusty-infra-security main
# deb-src https://esm.ubuntu.com/ubuntu trusty-infra-security main
deb https://esm.ubuntu.com/ubuntu trusty-infra-updates main
# deb-src https://esm.ubuntu.com/ubuntu trusty-infra-updates main

Updating package lists
ESM Infra enabled
Updating 'fips' apt sources list on changed directives.
Updating package lists
ADD AUTH APT REPO
MY CONTENT
content:  
Updating package lists
Updating 'fips-updates' apt sources list on changed directives.
Updating package lists
ADD AUTH APT REPO
MY CONTENT
content:  
Updating package lists
Failed to enable default services, check: sudo ua status
root@sru-t:~# cat /etc/apt/sources.list
sources.list       sources.list.d/    sources.list.save  
root@sru-t:~# cat /etc/apt/sources.list.
sources.list.d/    sources.list.save  
root@sru-t:~# cat /etc/apt/sources.list.
sources.list.d/    sources.list.save  
root@sru-t:~# cat /etc/apt/sources.list.d/
chad_smith-test-ua-client-do-release-upgrade-trusty.list
ubuntu-cc-eal.list
ubuntu-esm-infra-trusty.list
ubuntu-esm-infra.list
ubuntu-esm-infra.list.save
ubuntu-fips-updates.list
ubuntu-fips.list
```

systemd/ua-update-contract.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Ubuntu Advantage auto attach
```
blackboxsw, line 2: Sorry cut-n-paste error of mine here.
```suggestion
Description=Ubuntu Advantage update contract
```

systemd/ua-update-contract.service
```diff
@@ -0,0 +1,12 @@
...
+After=cloud-final.target
```
blackboxsw, line 3: At some point you and I might have to think about systems without cloud-init installed (like bare metal or desktop) and define the right After clauses for them

uaclient/cli.py
```diff
@@ -97,6 +98,38 @@ def new_f(args, cfg):
     return new_f
 
 
+def daemon_file_exists():
+    return os.path.exists(defaults.DEFAULT_UPGRADE_CONTRACT_FLAG_FILE)
+
+
+def get_running_services():
```
blackboxsw, line 105: let's move this into util module and maybe just define a function that does this work for us instead of jumping into a function call for a one-liner

is_service_running(service_name):
    running_services, _ = util.subp(["ps", "-aux"])
    return bool(service_name in running_services)

uaclient/cli.py
```diff
@@ -97,6 +98,38 @@ def new_f(args, cfg):
...
+def assert_contract_upgrade_not_running(f):
+    """Decorator asserting contract upgrade is not running."""
+
+    @wraps(f)
+    def new_f(args, cfg):
+        if daemon_file_exists():
+            running_services = get_running_services()
+            daemon_name = defaults.DAEMON_UPGRADE_CONTRACT_NAME
+
+            if running_services.find(daemon_name) == -1:
```
blackboxsw, line 119: this could be nicely expressed it a more pythonic way as the following
```suggestion
            if daemon_name in running_services:
```

uaclient/cli.py
```diff
@@ -97,6 +98,38 @@ def new_f(args, cfg):
...
+    running_services, _ = util.subp(["ps", "-aux"])
+    return running_services
+
+
+def assert_contract_upgrade_not_running(f):
+    """Decorator asserting contract upgrade is not running."""
+
+    @wraps(f)
+    def new_f(args, cfg):
+        if daemon_file_exists():
```
blackboxsw, line 115: this is a good idea with the decorator and something we should walk through with rick today to get his thoughts on how best to expose this information to the user for this release.

I feel like this damon status check info is something that we don't want to just hide inside this decorator.

I believe we will ultimately also want to write into status.json too, somehting machine readable for two reasons:
 1. we might want to allow status to succeed and report another bit of info "in-progress" or something while still emitting seatus output 
 2. we might want to add a blocking ua status --wait call so that PRO machines which setup a lot of services on first boot may not want to start using those VMs until all security patches etc are applied.

uaclient/contract.py
```diff
@@ -189,23 +189,83 @@ def _get_platform_data(self, machine_id):
...
 
 
-def process_entitlement_delta(orig_access, new_access, allow_enable=False):
+def process_entitlements_delta(
+    past_entitlements: "Dict[str, Any]",
+    new_entitlements: "Dict[str, Any]",
+    allow_enable: bool,
+    series_overrides: bool = True,
+) -> None:
+    """Iterate over all entitlements in new_entitlement and apply any delta
```
blackboxsw, line 198: thanks for the docs. Saves a lot of brain power in reviews and future maintenance

blackboxsw (changes requested):
looking really good. first round of comments.

blackboxsw (changes requested):
This is a good approach @lucasmoura.

I think we decided in team hangout/discussion we wanted to drop dependency on python3-daemon package and keep a synchronous solution only if possible to keep things simpler.

 I have a [PR up for ubuntu-release-upgrader](https://code.launchpad.net/~chad.smith/ubuntu-release-upgrader/+git/ubuntu-release-upgrader/+merge/391183) which allows for us to run during do-release-upgrade's PostUpgradeScripts.

When tested with your branch (coupled with this suggested diff below after you run `git rebase -i` first on your current branch) I was able to get a [successful do-release-upgrade run from bionic to focal ](https://paste.ubuntu.com/p/Jqqph5Xf2d/)

```diff
diff --git a/uaclient/cli.py b/uaclient/cli.py
index d4ec73f..ea8f132 100644
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -19,7 +19,6 @@ except ImportError:
 
 from uaclient import config
 from uaclient import contract
-from uaclient import defaults
 from uaclient import entitlements
 from uaclient import exceptions
 from uaclient import status as ua_status
@@ -98,38 +97,6 @@ def assert_not_attached(f):
     return new_f
 
 
-def daemon_file_exists():
-    return os.path.exists(defaults.DEFAULT_UPGRADE_CONTRACT_FLAG_FILE)
-
-
-def get_running_services():
-    running_services, _ = util.subp(["ps", "-aux"])
-    return running_services
-
-
-def assert_contract_upgrade_not_running(f):
-    """Decorator asserting contract upgrade is not running."""
-
-    @wraps(f)
-    def new_f(args, cfg):
-        if daemon_file_exists():
-            running_services = get_running_services()
-            daemon_name = defaults.DAEMON_UPGRADE_CONTRACT_NAME
-
-            if running_services.find(daemon_name) == -1:
-                raise exceptions.UserFacingError(
-                    "error processing upgrade-lts-contract script"
-                )
-            else:
-                raise exceptions.UserFacingError(
-                    "daemon upgrade-lts-contract is still running"
-                )
-
-        return f(args, cfg)
-
-    return new_f
-
-
 def auto_attach_parser(parser):
     """Build or extend an arg parser for auto-attach subcommand."""
     parser.prog = "auto-attach"
@@ -697,7 +664,6 @@ def get_parser():
     return parser
 
 
-@assert_contract_upgrade_not_running
 def action_status(args, cfg):
     if not cfg:
         cfg = config.UAConfig()
diff --git a/uaclient/defaults.py b/uaclient/defaults.py
index d339b2b..8ec6174 100644
--- a/uaclient/defaults.py
+++ b/uaclient/defaults.py
@@ -8,7 +8,6 @@ any of our dependencies installed.
 UAC_ETC_PATH = "/etc/ubuntu-advantage/"
 DEFAULT_CONFIG_FILE = UAC_ETC_PATH + "uaclient.conf"
 DEFAULT_HELP_FILE = UAC_ETC_PATH + "help_data.yaml"
-DEFAULT_UPGRADE_CONTRACT_FLAG_FILE = UAC_ETC_PATH + "request-update-contract"
 BASE_CONTRACT_URL = "https://contracts.canonical.com"
 DAEMON_UPGRADE_CONTRACT_NAME = "upgrade-lts-contract"
 
diff --git a/uaclient/tests/test_cli_status.py b/uaclient/tests/test_cli_status.py
index 93e2223..fe4bc35 100644
--- a/uaclient/tests/test_cli_status.py
+++ b/uaclient/tests/test_cli_status.py
@@ -9,7 +9,6 @@ import pytest
 from uaclient import util
 
 from uaclient.cli import action_status
-from uaclient import exceptions
 
 M_PATH = "uaclient.cli."
 
@@ -50,12 +49,10 @@ Technical support level: n/a
 )
 @mock.patch(M_PATH + "os.getuid", return_value=0)
 class TestActionStatus:
-    @mock.patch(M_PATH + "daemon_file_exists")
     def test_attached(
-        self, m_daemon, m_getuid, m_get_avail_resources, capsys, FakeConfig
+        self, m_getuid, m_get_avail_resources, capsys, FakeConfig
     ):
         """Check that root and non-root will emit attached status"""
-        m_daemon.return_value = False
         cfg = FakeConfig.for_attached_machine()
         assert 0 == action_status(mock.MagicMock(), cfg)
         # capsys already converts colorized non-printable chars to space
@@ -72,23 +69,19 @@ class TestActionStatus:
             expected_dash = "\u2014"
         assert ATTACHED_STATUS.format(dash=expected_dash) == printable_stdout
 
-    @mock.patch(M_PATH + "daemon_file_exists")
     def test_unattached(
-        self, m_daemon, m_getuid, m_get_avail_resources, capsys, FakeConfig
+        self, m_getuid, m_get_avail_resources, capsys, FakeConfig
     ):
         """Check that unattached status is emitted to console"""
-        m_daemon.return_value = False
         cfg = FakeConfig()
 
         assert 0 == action_status(mock.MagicMock(), cfg)
         assert UNATTACHED_STATUS == capsys.readouterr()[0]
 
-    @mock.patch(M_PATH + "daemon_file_exists")
     def test_unattached_json(
-        self, m_daemon, m_getuid, m_get_avail_resources, capsys, FakeConfig
+        self, m_getuid, m_get_avail_resources, capsys, FakeConfig
     ):
         """Check that unattached status json output is emitted to console"""
-        m_daemon.return_value = False
         cfg = FakeConfig()
 
         args = mock.MagicMock(format="json")
@@ -112,12 +105,10 @@ class TestActionStatus:
         }
         assert expected == json.loads(capsys.readouterr()[0])
 
-    @mock.patch(M_PATH + "daemon_file_exists")
     def test_error_on_connectivity_errors(
-        self, m_daemon, m_getuid, m_get_avail_resources, capsys, FakeConfig
+        self, m_getuid, m_get_avail_resources, capsys, FakeConfig
     ):
         """Raise UrlError on connectivity issues"""
-        m_daemon.return_value = False
         m_get_avail_resources.side_effect = util.UrlError(
             socket.gaierror(-2, "Name or service not known")
         )
@@ -131,10 +122,8 @@ class TestActionStatus:
         "encoding,expected_dash",
         (("utf-8", "\u2014"), ("UTF-8", "\u2014"), ("ascii", "-")),
     )
-    @mock.patch(M_PATH + "daemon_file_exists")
     def test_unicode_dash_replacement_when_unprintable(
         self,
-        _m_daemon,
         _m_getuid,
         _m_get_avail_resources,
         encoding,
@@ -143,7 +132,6 @@ class TestActionStatus:
     ):
         # This test can't use capsys because it doesn't emulate sys.stdout
         # encoding accurately in older versions of pytest
-        _m_daemon.return_value = False
         underlying_stdout = io.BytesIO()
         fake_stdout = io.TextIOWrapper(underlying_stdout, encoding=encoding)
 
@@ -159,35 +147,3 @@ class TestActionStatus:
 
         expected_out = ATTACHED_STATUS.format(dash=expected_dash)
         assert expected_out == out
-
-    @pytest.mark.parametrize(
-        "running_services,expected_msg",
-        (
-            (
-                "11:00 0:00 upgrade-lts-contract.py",
-                "daemon upgrade-lts-contract is still running",
-            ),
-            (
-                "11:00 0:00 unattended-upgrades",
-                "error processing upgrade-lts-contract script",
-            ),
-        ),
-    )
-    @mock.patch(M_PATH + "daemon_file_exists")
-    @mock.patch(M_PATH + "get_running_services")
-    def test_raise_exception_when_daemon_file_exists(
-        self,
-        m_services,
-        m_daemon,
-        m_getuid,
-        m_get_avail_resources,
-        running_services,
-        expected_msg,
-    ):
-        m_daemon.return_value = True
-        m_services.return_value = running_services
-
-        with pytest.raises(exceptions.UserFacingError) as e:
-            action_status(mock.MagicMock(), mock.MagicMock())
-
-        assert e.value.args[0] == expected_msg
diff --git a/upgrade-lts-contract.py b/upgrade-lts-contract.py
index d2fe1cf..5ea6fc1 100755
--- a/upgrade-lts-contract.py
+++ b/upgrade-lts-contract.py
@@ -1,15 +1,9 @@
 #!/usr/bin/env python3
 
-import argparse
 import contextlib
 import logging
 import os
 
-try:
-    from daemon import DaemonContext
-except ImportError:
-    DaemonContext = contextlib.suppress()
-
 from uaclient.cli import setup_logging
 from uaclient.config import UAConfig
 from uaclient.contract import process_entitlements_delta
@@ -29,17 +23,6 @@ current_codename_to_past_codename = {
 }
 
 
-def parse_args():
-    parser = argparse.ArgumentParser()
-    parser.add_argument(
-        "--no-daemon",
-        action="store_true",
-        help=("Sets the script to not run on daemon mode"),
-    )
-
-    return parser.parse_args()
-
-
 def process_contract_delta_after_apt_lock():
     setup_logging(logging.INFO, logging.DEBUG)
     out, _err = subp(["lsof", "/var/lib/apt/lists/lock"], rcs=[0, 1])
@@ -77,14 +60,6 @@ def process_contract_delta_after_apt_lock():
         "upgrade-lts-contract succeeded after %d retries", retry_count
     )
 
-    os.remove("/etc/ubuntu-advantage/request-update-contract")
-
 
 if __name__ == "__main__":
-    args = parse_args()
-
-    if args.no_daemon:
-        process_contract_delta_after_apt_lock()
-    else:
-        with DaemonContext():
-            process_contract_delta_after_apt_lock()
+    process_contract_delta_after_apt_lock()

```

lucasmoura:
@blackboxsw I have updated the code removing all of the daemon uses.

However, I have keep our apt lock loop on the upgrade-lts-contract script just as safe mechanism. But if you think we can remove it, no problem at all

blackboxsw:
> @blackboxsw I have updated the code removing all of the daemon uses.
> 
> However, I have keep our apt lock loop on the `upgrade-lts-contract` script just as safe mechanism. But if you think we can remove it, no problem at all

I think having the apt retry lock is a good mechanism to keep, there are timing possibilities I think if a system is upgrading for a long time that automated apt daily updates could kick, in and steal the apt lock out from under us. Yes I think leaving those safeguards is good.

debian/rules
```diff
@@ -48,15 +48,13 @@ override_dh_auto_install:
 	make -C apt-hook DESTDIR=$(CURDIR)/debian/ubuntu-advantage-tools install
 
 ifeq (${VERSION_ID},"14.04")
-	# Move ua-auto-attach.conf to ubuntu-advantage-pro
```
blackboxsw, line 51: I don't think we need any changes now in debian/rules now that no daemon /init present.

blackboxsw (approved):
LGTM!

PR #1183, opened 2020-09-18
cli: status add blocking --wait param and lock files for config change operations

blackboxsw:


uaclient/config.py
```diff
@@ -276,12 +279,47 @@ def _remove_beta_resources(self, response) -> "Dict[str, Any]":
...
+                inactive, active, reboot-required
+            configStatusDescription will provide more details about that state.
+        """
+        userStatus = status.UserFacingConfigStatus
+        status_val = userStatus.INACTIVE.value
+        status_desc = status.MESSAGE_NO_ACTIVE_OPERATIONS
+        if os.path.exists(self.data_path("lock")):
+            lock_file = self.data_path("lock")
+            lock_content = util.load_file(lock_file)
+            [lock_pid, lock_holder] = lock_content.split(":")
```
lucasmoura, line 295: This part of the code seems really similar to the one we have in lock decorator, maybe we can move it to util module

uaclient/config.py
```diff
@@ -337,6 +375,7 @@ def _attached_status(self) -> "Dict[str, Any]":
         from uaclient.entitlements import ENTITLEMENT_CLASSES
 
         response = copy.deepcopy(DEFAULT_STATUS)
+        response.update(self._get_config_status())
```
lucasmoura, line 378: Maybe we can update the response dict directly on `def status` method

lucasmoura (commented):
Just a couple of nits. Maybe we can have an integration test for this feature as well

uaclient/config.py
blackboxsw: moved it to util.check_lock_info for use in both callsites.

.travis.yml
```diff
@@ -19,12 +19,34 @@ matrix:
   allow_failures:
     -name: azurepro-16.04
   include:
+    - if: env(UACLIENT_BEHAVE_AWS_SECRET_ACCESS_KEY)
```
lucasmoura, line 22: I think we should split these into two separate jobs, since this check only for the existence of aws keys and not azure. This means that we may have stale azure vms that will not be cleaned because we are not providing aws keys 9Even though this scenario is not the one we have right now, I think we should be able to support it)

uaclient/cli.py
```diff
@@ -779,6 +847,9 @@ def wrapper(*args, **kwargs):
             with util.disable_log_to_console():
                 logging.exception("KeyboardInterrupt")
             print("Interrupt received; exiting.", file=sys.stderr)
+            if _LOCK_FILE and os.path.exists(_LOCK_FILE):
```
lucasmoura, line 850: Since we are using the same code in three different places in this function, maybe we could move to the `util` module as well

uaclient/config.py
```diff
@@ -48,6 +50,7 @@ class UAConfig:
         "instance-id": DataPath("instance-id", True),
         "machine-id": DataPath("machine-id", True),
         "machine-token": DataPath("machine-token.json", True),
+        "lock": DataPath("lock", False),
```
lucasmoura, line 53: Why is the lock file not marked as private ?

blackboxsw: Originally I was trying to write is as non-root user too and update status based on if other things held the lock. I backed out that change as I realized that meant we had to open up world writeable perms on the directory /var/lib/ubuntu-advantage, so I backed out the non-root user manipulating this. I forgot to back out this change when I reverted that code. dropping that now

lucasmoura (approved):

PR #1185, opened 2020-09-22
Add integration test for purge command

lucasmoura:
Create behave integration test for `apt purge ubuntu-advantage-tools`

To do that we had to:

* Fix the `clean_apt_files` method
* Update the `remove_postinst_created_files` function to only delete gpg keys. The other files should be handled by `clean_apt_files`in `prerm`
* Update postinst to not recreate old apt files for esm-infra for the trusty package

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -18,6 +18,8 @@ remove_logs(){
 
 remove_gpg_files(){
     rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
+    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-apps.gpg
+    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-fips.gpg
 }
 
 case "$1" in
--- a/debian/postrm
+++ b/debian/postrm
@@ -16,11 +16,8 @@ remove_logs(){
     rm -f /var/log/ubuntu-advantage.log*
 }
 
-remove_postinst_created_files(){
-    # config files created in postinst, need explicit handling on purge
+remove_gpg_files(){
     rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
-    rm -f /etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list
-    rm -f /etc/apt/preferences.d/ubuntu-esm-infra-trusty
 }
 
 case "$1" in
@@ -28,6 +25,7 @@ case "$1" in
         remove_apt_auth
         remove_cache_dir
         remove_logs
+        remove_gpg_files
         ;;
 esac
 
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -325,3 +325,33 @@ Feature: Command behaviour when attached to an UA subscription
            | focal   |
            | trusty  |
            | xenial  |
+
+    @series.all
+    Scenario Outline: Purge package after attaching it to a machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `touch /etc/apt/preferences.d/ubuntu-esm-infra` with sudo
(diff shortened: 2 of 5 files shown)


uaclient/apt.py
```diff
@@ -320,7 +320,7 @@ def clean_apt_files(*, _entitlements=None):
         from uaclient import entitlements as _entitlements
 
     for ent_cls in _entitlements.ENTITLEMENT_CLASSES:
-        if not isinstance(ent_cls, RepoEntitlement):
+        if not isinstance(ent_cls(), RepoEntitlement):
```
blackboxsw, line 323: I think this should be issubclass test instead of doing the work to instantiate the subclass
```suggestion
        if not issubclass(ent_cls, RepoEntitlement):
```

lucasmoura:
@blackboxsw I have updated the code removing all of the daemon uses.

However, I have keep our apt lock loop on the `upgrade-lts-contract` script just as safe mechanism. But if you think we can remove it, no problem at all

lucasmoura:
@blackboxsw I have update the code to use `issubclass`

debian/postrm
```diff
@@ -16,19 +16,16 @@ remove_logs(){
     rm -f /var/log/ubuntu-advantage.log*
 }
 
-remove_postinst_created_files(){
-    # config files created in postinst, need explicit handling on purge
+remove_gpg_files(){
     rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
```
blackboxsw, line 20: I think we need to make this match any /etc/apt/trusted.gpg.d/ubuntu-advantage-*.gpg  file because other gpg files will be created here for each apt-based UA service we deliver (and enable after initial attach)

features/attached_commands.feature
```diff
@@ -325,3 +325,33 @@ Feature: Command behaviour when attached to an UA subscription
...
+
+    @series.all
+    Scenario Outline: Purge package after attaching it to a machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `touch /etc/apt/preferences.d/ubuntu-esm-infra` with sudo
+        Then I verify that running `test -f /var/log/ubuntu-advantage.log` `with sudo` exits `0`
+        And I verify that running `test -d /var/lib/ubuntu-advantage` `with sudo` exits `0`
+        And I verify that running `test -f /etc/apt/auth.conf.d/90ubuntu-advantage` `with sudo` exits `0`
+        And I verify that running `test -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg` `with sudo` exits `0`
```
blackboxsw, line 337: Can we somehow test with globbing or ls ubuntu-advantage-*.gpg we haven't leaked any other gpg files after purge.

debian/postrm
```diff
@@ -16,19 +16,16 @@ remove_logs(){
     rm -f /var/log/ubuntu-advantage.log*
 }
 
-remove_postinst_created_files(){
-    # config files created in postinst, need explicit handling on purge
-    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
-    rm -f /etc/apt/sources.list.d/ubuntu-esm-infra-trusty.list
-    rm -f /etc/apt/preferences.d/ubuntu-esm-infra-trusty
+remove_gpg_files(){
+    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-*.gpg
```
blackboxsw, line 20: I misspoke on this one earlier it's probably safer for us to do the following:
```suggestion
    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-apps.gpg
    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-fips.gpg
```

blackboxsw (approved):
Nice on the new then clauses for the purge BDD test. Thanks for that.

PR #1186, opened 2020-09-29
Save machine-token when user runs detach

lucasmoura:
Now when a user detach from a subscription we will not disable any of
the enabled services and we will create a copy of the current machine
token file on /var/lib/ubuntu-advantage/private/machine-token.json.past

When a user want to attach to a new subscription, we will compute
the contract delta based on the past machine token and the current one.
We will use that to disable services that the user is no longer
entitled.


uaclient/defaults.py
```diff
@@ -6,9 +6,12 @@
 """
 
 UAC_ETC_PATH = "/etc/ubuntu-advantage/"
+UAC_VAR_PATH = "/var/lib/ubuntu-advantage/"
```
blackboxsw, line 9: you probably also want to set the following in config defaults
"data_dir": UAC_VAR_PATH.rstrip("/")

lucasmoura:
@blackboxsw updated

uaclient/cli.py
```diff
@@ -524,6 +533,13 @@ def _attach_with_token(
     except util.UrlError as exc:
         with util.disable_log_to_console():
             logging.exception(exc)
+
+        # Check if we have used a past machine-token file.
+        # If we had, delete this file, otherwise the system
+        # will think that it attached to the past subscription.
+        if os.path.exists(defaults.MACHINE_TOKEN_FILE):
+            util.del_file(defaults.MACHINE_TOKEN_FILE)
+
```
blackboxsw, line 542: This isn't checking the PAST_MACHINE_TOKEN
but as my suggestion above, we should use a UAConfig.data_paths key for this
```suggestion
           cfg.delete_cache("machine-token-saved")   # only deletes if cached json.save file exists
```

uaclient/cli.py
```diff
@@ -541,6 +557,10 @@ def _attach_with_token(
     )
 
     action_status(args=None, cfg=cfg)
+
+    if os.path.exists(defaults.PAST_MACHINE_TOKEN_FILE):
+        util.del_file(defaults.PAST_MACHINE_TOKEN_FILE)
+
```
blackboxsw, line 563: probably should be
```suggestion
    cfg.delete_cache("machine-token-saved")

```

uaclient/config.py
```diff
@@ -254,6 +254,10 @@ def write_cache(self, key: str, content: "Any") -> None:
                 mode = 0o644
         util.write_file(filepath, content, mode=mode)
 
+    def save_file_with_suffix(self, file_name: str, suffix: str) -> None:
```
blackboxsw, line 257: I'm not sure we need this utility function on UAConfig. we can content = read_cache("machine-token");  write_cache("machine-token-saved", content)

uaclient/defaults.py
```diff
@@ -6,14 +6,17 @@
...
 DEFAULT_CONFIG_FILE = UAC_ETC_PATH + "uaclient.conf"
 DEFAULT_HELP_FILE = UAC_ETC_PATH + "help_data.yaml"
 DEFAULT_UPGRADE_CONTRACT_FLAG_FILE = UAC_ETC_PATH + "request-update-contract"
+MACHINE_TOKEN_FILE = UAC_VAR_PATH + "private/machine-token.json"
```
blackboxsw, line 13: I don't think we can/should hard-code these paths. Probably don't need the UAC_VAR_PATH either because we need to rely on the dynamic/overriding UAConfig.data_dir/data_paths methods to get to both machine-token.json and the new machine-token.json.save files

uaclient/util.py
```diff
@@ -669,3 +670,13 @@ def check_lock_info(lock_path: str) -> "Tuple[int, str]":
         )
         os.unlink(lock_path)
         return no_lock
+
+
+def copy_file(src: str, dst: str) -> None:
```
blackboxsw, line 675: I suppose we don't really need this function if you end up using cfg.read_cache -> write_cache calls. I'll let you digest the comments above and determine if it's still worth it.

blackboxsw (changes requested):
Thanks for pulling this together Lucas. I think we have a few things to tackle here but I expect we can wrap them up tomorrow.

uaclient/cli.py
```diff
@@ -92,6 +93,27 @@ def new_f(args, cfg, **kwargs):
...
+    If have an old machine-token file in the system, generated by a past
+    attach command, we want to compute the contract diff between
+    the past machine-token file and the current one. This will allows
+    us to disable services the user is no longer entitled to.
+    """
+
+    @wraps(f)
+    def new_f(args, cfg):
+        util.copy_file(
+            src=defaults.PAST_MACHINE_TOKEN_FILE,
```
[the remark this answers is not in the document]
lucasmoura: I have updated the code to use the `data_paths` from `UAConfig`, but I do not understand why we would need to write to the cache in that decorator

uaclient/cli.py
lucasmoura: Yes, that was incorrect, but we still need to delete the `machine-token` here. The scenario is that we failed to perform the attach but we have copied the `machine-token.json.save` into `machine-token.json`, which will cause uaclient to think that the system is attached, even though the attach has failed.

uaclient/util.py
lucasmoura: Yes, you are right. The combination of those cache functions is already enough

lucasmoura:
@blackboxsw I have updated the code

uaclient/tests/test_cli_attach.py
```diff
@@ -149,6 +149,7 @@ def fake_request_updated_contract(cfg, contract_token, allow_enable):
 
         request_updated_contract.side_effect = fake_request_updated_contract
         ret = action_attach(args, cfg)
+        print(cfg.is_attached)
```
blackboxsw, line 152: We should drop this.

features/attached_commands.feature
```diff
@@ -70,7 +70,7 @@ Feature: Command behaviour when attached to an UA subscription
            | xenial  |
 
     @series.all
-    Scenario Outline: Attached detach in a trusty machine
+    Scenario Outline: Attached detach in a ubuntu machine
```
blackboxsw, line 73: Since you are referencing staging_token here you need the appropriate decorator too.
```suggestion
   @uses.config.contract_token_staging
    Scenario Outline: Attached detach in a ubuntu machine
```

lucasmoura:
Closing this PR as we not longer save the machine token when dettaching

PR #1187, opened 2020-09-30
Add tests to upgrade_lts_contract script

lucasmoura:
Add tests to the `upgrade_lts_script`. To do that, we have moved the script inside a new folder named `lib`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ def _get_version():
 def _get_data_files():
     data_files = [
         ("/etc/ubuntu-advantage", ["uaclient.conf"]),
-        ("/usr/lib/ubuntu-advantage", ["upgrade-lts-contract.py"]),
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -40,7 +40,7 @@ def _get_version():
 def _get_data_files():
     data_files = [
         ("/etc/ubuntu-advantage", ["uaclient.conf"]),
-        ("/usr/lib/ubuntu-advantage", ["upgrade-lts-contract.py"]),
+        ("/usr/lib/ubuntu-advantage", glob.glob("lib/[!_]*")),
```
blackboxsw, line 43: Just tested package build and globbing does package this script and deliver it 
/usr/lib/ubuntu-advantage/upgrade_lts_contract.py. So, I think we are good here.

blackboxsw (approved):
+1 awaiting CI pass here.

PR #1188, opened 2020-10-02
Do not remove packages

lucasmoura:
When we perform a `disable` or `detach` operation, we end up removing packages that were installed to enable the services. We are now keeping those packages installed when we perform those operations.

Also, this PR is removing the `azurepro-16.04` from `allow_failures`, since the job is running without issues

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ script:
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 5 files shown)


blackboxsw (approved):
+1 looks good. Thank you Lucas

PR #1195, opened 2020-10-06
Upgrade tests

lucasmoura:
Add integration tests when upgrading between ubuntu releases. We are also creating a new stage in travis CI just to run those tests, since we believe they will be time consuming.

PS: Right now we are using work that is still not merged on `ubuntu-release-upgrader`. Once those contributions land, we can update the tests to use `do-release-upgrade` directly


blackboxsw (approved):
+1 on this status checking if any service deltas present and cache available.

PR #1196, opened 2020-10-07
Refactor travis file

lucasmoura:
Create scripts to reuse on similar integration tests scenarios on travis yaml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,26 +67,7 @@ matrix:
           - pip install boto3
           - git clone https://github.com/canonical/server-test-scripts.git
       script:
-          - mkdir deb-artifacts
-          - cp *-debs/* deb-artifacts
-          - >
-              if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
-                  BUILD_PR=0
-              else
-                 cd $TRAVIS_BUILD_DIR/..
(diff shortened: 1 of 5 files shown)


blackboxsw (changes requested):
Let's move these scripts into the existing ./tools  subdir as we may at somepoint what to use and package a ./bin directory  in the deb package. I don't want to confuse tooling with releasable content..

bin/base_travis_integration_tests.sh
```diff
@@ -0,0 +1,33 @@
...
+
+create_pr_tar_file() {
+  cd $TRAVIS_BUILD_DIR/..
+  tar -zcf pr_source.tar.gz ubuntu-advantage-client
+  cp pr_source.tar.gz /tmp
+  ls -lh /tmp
+  cd $TRAVIS_BUILD_DIR
+}
+
+copy_deb_packages
```
blackboxsw, line 22: How we have side-effect functions being called just by sourcing this directory. I'd like to see all explicit call-sites in the scripts which source it. So we are not suprised by hidden behavior

lucasmoura:
@blackboxsw updated

blackboxsw (approved):
This looks much cleaner thank you.

PR #1197, opened 2020-10-07
changelog for 25.0~beta release

blackboxsw:
Update changelog for beta release

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,12 +1,48 @@
-ubuntu-advantage-tools (25.0) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (25.0~beta) UNRELEASED; urgency=medium
 
-  * d/control: lint warning update debhelper version 9->10 avoid deprecation
+  * d/control:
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -1,12 +1,48 @@
...
+  * d/rules:
+    - drop --with systemd fix build-depends-on-obsolete-package
+    - set fix lintian warning extra:Depends even if empty
+  * d/postrm
+    - Add more gpg keys to be deleted in postrm for Xenial+ support
+  * d/postinst:
+    - do not unconfigure non-trusty esm. no series in apt filenames (GH: #1170)
+    - check if esm is already enabled (GH: #1095)
+  * New upstream release 25.0:
+    - Do not uninstall additionalPackages or livepatch when disabling services
```
lucasmoura, line 17: Just a minor nit, I think this behavior is applied to all services and not only livepatch

lucasmoura: My mistake here, it says `or` livepatch, so it is correct

lucasmoura (approved):

PR #1204, opened 2020-10-13
Add vm test commands in tox.ini

lucasmoura:
Add `tox` commands that allow us to run only vm tests. Also, we are fixing some vm tests, since we are now not uninstalling packages when we disable services.

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -288,7 +288,6 @@ Feature: Enable command behaviour when attached to an UA subscription
             FIPS Updates is not available for Ubuntu 20.04 LTS (Focal Fossa).
             """
 
-    @series.xenial
     @series.bionic
     @uses.config.machine_type.lxd.vm
     Scenario Outline: Attached enable of vm-based services in an ubuntu lxd vm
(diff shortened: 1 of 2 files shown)


features/attached_enable.feature
```diff
@@ -288,7 +288,6 @@ Feature: Enable command behaviour when attached to an UA subscription
             FIPS Updates is not available for Ubuntu 20.04 LTS (Focal Fossa).
             """
 
-    @series.xenial
```
blackboxsw, line 291: Why are we dropping @series.xenial from this test?

tox.ini
```diff
@@ -33,6 +34,9 @@ commands =
     behave-lxd-16.04: behave -v {posargs} --tags="series.xenial,series.all" --tags="~upgrade"
     behave-lxd-18.04: behave -v {posargs} --tags="series.bionic,series.all" --tags="~upgrade"
     behave-lxd-20.04: behave -v {posargs} --tags="series.focal,series.all" --tags="~upgrade"
+    behave-vm-16.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.xenial"
+    behave-vm-18.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.bionic"
```
blackboxsw, line 38: These targets should match series.all as well so we don't skip features/attach_validtoken.feature line 76 which decorates with a series.all instead of series.bionic specifically.
```suggestion
    behave-vm-18.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.all,series.bionic"
```

blackboxsw (approved):
Minor comment for a followup to inclued series.all on vm targets in a future PR.

PR #1206, opened 2020-10-14
Show error when enabling service with invalid repo

lucasmoura:
When we enable a service and a user has an invalid repository in the system, apt update commands will fail to run. Because we run apt update when we enable repo type services, we stumble upon this issue and fail to enable the service. However, we are not showing the user the reason why the enable command failed. We are now updating the logic to display this apt issue when this problem happen during an enable operation.

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -400,3 +400,29 @@ Feature: Command behaviour when attached to an UA subscription
            | focal   |
            | trusty  |
            | xenial  |
+
+    @series.all
+    Scenario Outline: Enable command with invalid repositories in user machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `ua disable esm-infra` with sudo
+        And I run `add-apt-repository ppa:canonical-server/ua-client-daily -y` with sudo
+        And I run `apt update` with sudo
+        And I run `sed -i 's/ubuntu/ubun/' /etc/apt/sources.list.d/<ppa_file>-<release>.list` with sudo
+        And I run `ua enable esm-infra` with sudo
+        Then stdout matches regexp:
+        """
+        One moment, checking your subscription first
+        Updating package lists
+        APT update failed.
+        APT update failed to read APT config for the following URL:
+        - http://ppa.launchpad.net/canonical-server/ua-client-daily/ubun
+        """
+
+        Examples: ubuntu release
+           | release | ppa_file                                |
+           | trusty  | canonical-server-ua-client-daily        |
+           | xenial  | canonical-server-ubuntu-ua-client-daily |
+           | bionic  | canonical-server-ubuntu-ua-client-daily |
+           | focal   | canonical-server-ubuntu-ua-client-daily |
+
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -84,6 +84,50 @@ def assert_valid_apt_credentials(repo_url, username, password):
         )
 
 
+def _parse_apt_update_for_invalid_apt_config(apt_error: str) -> str:
+    """Parse apt update errors for invalid apt config in user machine.
+
+    This functions parses apt update errors regarding the presence of
+    invalid apt config in the system, for example, a ppa that cannot be
+    reached, for example.
+
+    In that scenario, apt will output a message in the following formats:
+
+    The repository 'ppa 404 Release' ...
+    Failed to fetch ppa 404 ...
+
+    On some releases, both of these errors will be present in the apt error
+    message.
+
+    :param apt_error: The apt error string
+    :return: a string containing the parsed error message.
+    """
+    error_msg = ""
+    failed_repos = set()
+
+    for line in apt_error.strip().split("\n"):
+        if line:
+            pattern_match = re.search(
+                r"(Failed to fetch |The repository .)(?P<url>[^\s]+)", line
+            )
+
+            if pattern_match:
+                repo_url_match = (
+                    "- " + pattern_match.groupdict()["url"].split("/dists")[0]
+                )
+
+                failed_repos.add(repo_url_match)
+
+    if failed_repos:
+        error_msg += "\n"
+        error_msg += status.MESSAGE_APT_UPDATE_INVALID_URL_CONFIG.format(
+            "s" if len(failed_repos) > 1 else "",
+            "\n".join(sorted(failed_repos)),
+        )
+
+    return error_msg
+
+
 def run_apt_command(
     cmd: "List[str]", error_msg: str, env: "Optional[Dict[str, str]]" = {}
 ) -> str:
@@ -104,6 +148,14 @@ def run_apt_command(
     except util.ProcessExecutionError as e:
         if "Could not get lock /var/lib/dpkg/lock" in str(e.stderr):
             error_msg += " Another process is running APT."
+        else:
+            """
+            Treat errors where one of the APT repositories
+            is invalid or unreachable. In that situation, we alert
+            which repository is causing the error
+            """
+            error_msg += _parse_apt_update_for_invalid_apt_config(e.stderr)
+
         raise exceptions.UserFacingError(error_msg)
     return out
 
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -323,11 +323,15 @@ class RepoEntitlement(base.UAEntitlement):
                 ["apt-get", "update"], status.MESSAGE_APT_UPDATE_FAILED
             )
         except exceptions.UserFacingError:
-            self.remove_apt_config()
+            self.remove_apt_config(run_apt_update=False)
             raise
 
-    def remove_apt_config(self):
-        """Remove any repository apt configuration files."""
+    def remove_apt_config(self, run_apt_update=True):
+        """Remove any repository apt configuration files.
+
+        :param run_apt_update: If after removing the apt update
+            command after removing the apt files.
+        """
         series = util.get_platform_info()["series"]
         repo_filename = self.repo_list_file_tmpl.format(name=self.name)
         entitlement = self.cfg.entitlements[self.name].get("entitlement", {})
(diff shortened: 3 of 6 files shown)


features/attached_commands.feature
```diff
@@ -364,3 +364,65 @@ Feature: Command behaviour when attached to an UA subscription
...
+        And I run `add-apt-repository ppa:ua-client/proposed -y` with sudo
+        And I run `apt update` with sudo
+        And I run `sed -i 's/ubuntu/ubun/' /etc/apt/sources.list.d/ua-client-proposed-trusty.list` with sudo
+        And I run `ua enable esm-infra` with sudo
+        Then stdout matches regexp:
+        """
+        One moment, checking your subscription first
+        Updating package lists
+        APT update failed.
+        Failed to fetch http://ppa.launchpad.net/ua-client/proposed/ubun/dists/trusty/main/binary-amd64/Packages  404  Not Found .*
```
blackboxsw, line 382: I think we can cleanup an minimize the message a bit with a call to action for the person using UA CLI.

What do folks think about this mesaging

Unable to `ua enable esm-infra` due to APT update failure.
Fix apt config for http://ppa.launchpad.net/ua-client/proposed/ubun/dists/trusty/main/binary-amd64/Packages and retry.

lucasmoura: I see no problem on using that message instead of the full pat update error. What do you think @mitechie ?

lucasmoura: @blackboxsw What about updating the message to:

```
 Fix APT config for this repo and retry:
- http://ppa.launchpad.net/ua-client/proposed/ubun/dists/trusty/main/binary-amd64/Packages 
```

The rationale for that is situations were the user has more than one invalid ppas int the system

blackboxsw: Message looks pretty good, and I think a list of errant apt repos makes sense, but I'd defer to @mitechie on whether that list might get too long if we list 3 or more failing apt repo URLs.

Can the message be:
```
APT update failed to read APT config for the following URLs:
- http://ppa.launchpad.net/ua-client/proposed/ubun/dists/trusty/main/binary-amd64/Packages
- URL2
Fix any APT config for the URLs above and retry `ua enable <blah>`
```

mitechie: I am ok with noting the details of the error and if there are several that's ok. I do think that we need that summary message of "Unable to enable esm-infra due to APT update failures. Please address and retry"

Presenting the user the important details of the error to understand what's wrong is great. We should also always try to go the next step and give them the "what do I do next" as well.

lucasmoura:
@blackboxsw @mitechie I have updated the error message.

@blackboxsw right now, I am not including the `Fix any APT config for the URLs above and run: ua enable <blah>`. The reason for that is because this scenario may happen on different scenarios than `ua enable`, such as `ua refresh`. Also, since we are using the same exception, `UserFacingError` to catch that apt update issue, it would be hard to identify situation were we should append this message. Of course we could check the exception message content and append there directly, but I just want to be sure that this is a path we need to follow

uaclient/apt.py
```diff
@@ -104,6 +104,48 @@ def run_apt_command(
...
             error_msg += " Another process is running APT."
+
+        """
+        Treat errors where one of the APT repositories
+        is invalid or unreachable. In that situation, we alert
+        which repository is causing the error
+        """
+        rep_error = "The repository"
+        fetch_error = "Failed to fetch"
+        if rep_error in str(e.stderr) or fetch_error in str(e.stderr):
```
blackboxsw, line 115: 1. this bit of logic should be a separate testable function with a couple of simple unit tests.
2. We should probably just walk the lines of the stderr anyway and use re.match on it:
3. let's use named matching regex pattern so we can easily grab the matched object by keyname
```
failed_repos = set()
for line in e.stderr.splitlines():
   pattern_match = re.search(r"(Failed to fetch|The repository) (?P<url>[^\s]+)(?= )", line)
   if pattern_match:
       # Xenial emits both 'Failed to fetch' and 'The repository' errors. Set avoids url dupes.
       failed_urls.update(["- " + pattern_match.groupdict("url")])
   if failed_urls:
       error_msg += status.ERROR_APT_CONFIG_URLS.format(urls="\n".join(failed_urls))

File: uaclient/apt.py [hunk 3]
```diff
@@ -104,6 +104,48 @@ def run_apt_command(
     except util.ProcessExecutionError as e:
         if "Could not get lock /var/lib/dpkg/lock" in str(e.stderr):
             error_msg += " Another process is running APT."
+
+        """
```
blackboxsw, line 108: Shouldn't this be an else clause as we shouldn't be able to run into an APT lock error and repository update errors right?

uaclient/apt.py
```diff
@@ -104,6 +104,48 @@ def run_apt_command(
...
+                        """
+                        for repo_url in failed_repos:
+                            if repo_url in repo_url_match:
+                                append_repo = False
+                                break
+
+                        if append_repo:
+                            failed_repos.append(repo_url_match)
+
+            apt_error = "Fix APT config for {} repo{} and retry:\n{}"
```
blackboxsw, line 143: Let's make this a ERROR template in uaclient/status.py

blackboxsw (changes requested):
let's work a bit on the logic and get a couple of unit tests on the APT url error handling

uaclient/apt.py
lucasmoura: @blackboxsw That's a good point to move that logic into a separate function, however the set directive will not work here because, for Xenial, the urls are not exactly the same. For example, we will have those two lines in xenial:

```
The repository 'http://ppa.launchpad.net/ua-client/proposed/ubun 404' ...
Failed to fetch http://ppa.launchpad.net/ua-client/proposed/ubun/dists/xenial/main/binary-amd64/Packages
```

So the URLs are not exactly the same. That's why I am using the for loop here.

But I will move the logic into a separate function

lucasmoura:
@blackboxsw Code updated. However, I have not added new unit test because I think the ones from `test_run_apt_command_with_invalid_repositories`  already covers the scenarios of the new function. But if I do believe I should test the function directly, just let me know.

lucasmoura:
@blackboxsw Code updated. Thanks for the last suggestion regarding the split on `/dists`

blackboxsw (approved):
LGTM!

uaclient/apt.py
```diff
@@ -104,6 +147,18 @@ def run_apt_command(
     except util.ProcessExecutionError as e:
         if "Could not get lock /var/lib/dpkg/lock" in str(e.stderr):
             error_msg += " Another process is running APT."
+        else:
+            """
+            Treat errors where one of the APT repositories
+            is invalid or unreachable. In that situation, we alert
+            which repository is causing the error
+            """
```
blackboxsw, line 155: This docstring should be inside _parse_apt_pdate_for_invalid_apt_config function. It doesn't really belong up here.

uaclient/apt.py
```diff
@@ -104,6 +147,18 @@ def run_apt_command(
...
+            rep_error = "The repository"
+            fetch_error = "Failed to fetch"
+            if rep_error in str(e.stderr) or fetch_error in str(e.stderr):
+                error_msg += "\n"
+                error_msg += _parse_apt_update_for_invalid_apt_config(e.stderr)
```
blackboxsw, line 160: I think _parse_apt_update_for_invalid_apt_config should prepend the "\n" itself. Also, since _parse_apt_update_for_invalid_apt_config does return "" if no match, we should be able to just call that directly.

```suggestion
    error_msg += _parse_apt_update_for_invalid_apt_config(e.stderr)
```

blackboxsw (changes requested):
Sorry @lucasmoura Just a final review pass.

uaclient/apt.py
lucasmoura: Maybe we could keep it where it is, since this docstring is describing why we are doing the checks before calling the `_parse_apt_update_for_invalid_apt_config`. But if you disagree, I can move it inside the parser function

uaclient/apt.py
lucasmoura: The reason that I am appending here to `error_msg`  is that it already contains a error string that we return: 

```
APT update failed.
```

Maybe we can ignore this message and return the result directly. However, my only concern is the case that `_parse_apt_update_for_invalid_apt_config` returns empty. In that situation, we would not alert the user apt the apt update failure.

A second option is to pass the currently `error_msg` into the parse function and do the appending internally. What do you think @blackboxsw ?

blackboxsw: I still suggest we append, just let the _parse_apt_update function add the "\n" if it finds errors it can parse. In the event that it returns "" we don't end up altering error_msg because we error_msg +=  "".

Here's the full diff of whatI meant, _parse_apt_update already hasthe right logic in it for the most part (minus the leading "\n")

```diff
diff --git a/uaclient/apt.py b/uaclient/apt.py
index 20686b4..7d64a05 100644
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -119,7 +119,7 @@ def _parse_apt_update_for_invalid_apt_config(apt_error: str) -> str:
                 failed_repos.add(repo_url_match)
 
     if failed_repos:
-        error_msg = status.MESSAGE_APT_UPDATE_INVALID_URL_CONFIG.format(
+        error_msg = "\n" + status.MESSAGE_APT_UPDATE_INVALID_URL_CONFIG.format(
             "s" if len(failed_repos) > 1 else "",
             "\n".join(sorted(failed_repos)),
         )
@@ -148,17 +148,8 @@ def run_apt_command(
         if "Could not get lock /var/lib/dpkg/lock" in str(e.stderr):
             error_msg += " Another process is running APT."
         else:
-            """
-            Treat errors where one of the APT repositories
-            is invalid or unreachable. In that situation, we alert
-            which repository is causing the error
-            """
-            rep_error = "The repository"
-            fetch_error = "Failed to fetch"
-            if rep_error in str(e.stderr) or fetch_error in str(e.stderr):
-                error_msg += "\n"
-                error_msg += _parse_apt_update_for_invalid_apt_config(e.stderr)
-
+            # Add any matching APT URL config error messages
+            error_msg += _parse_apt_update_for_invalid_apt_config(e.stderr)
         raise exceptions.UserFacingError(error_msg)
     return out
```

lucasmoura:
@blackboxsw Code updated

uaclient/apt.py
blackboxsw: +1 we can leave this docstring.
I still think we should drop the 3 lines below this docstring because that logic is contained within _parse_apt_update_for_invalid_apt_config which will return "" if no matching rep_error or fetch_error.

So I think those lines below can be just
error_msg += _parse_apt_update_for_invalid_apt_config(e.stderr)

blackboxsw (approved):
Thanks! Let's land it!

PR #1217, opened 2020-10-21
Do not enable fips when fips-updates is active

lucasmoura:
We are adding a static affordance to verify the status of fips-updates before enabling fips. If fips-updates is enabled, we cannot enable fips, since that would cause package downgrades.

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -431,3 +431,36 @@ Feature: Enable command behaviour when attached to an UA subscription
             One moment, checking your subscription first
             Cannot enable FIPS when Livepatch is enabled
             """
+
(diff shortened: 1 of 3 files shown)


features/attached_enable.feature
```diff
@@ -431,3 +431,37 @@ Feature: Enable command behaviour when attached to an UA subscription
             One moment, checking your subscription first
             Cannot enable FIPS when Livepatch is enabled
             """
+
+    @wip
```
blackboxsw, line 435: Let's drop the @wip

blackboxsw (approved):
Minor change to drop @wip. And +1

PR #1220, opened 2020-10-23
Update beta info for services

lucasmoura:
This PR updates two aspects of uaclient:

* Turn `esm-apps` into a beta service
* Do not show beta services on `ua help` command, unless the user runs the `ua help --all`.  Also `ua --help` will now display the help with only the non-beta services

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -59,13 +59,13 @@ _LOCK_FILE = None
 class UAArgumentParser(argparse.ArgumentParser):
     def __init__(
         self,
-        base_desc: str = None,
-        non_beta_services_desc: "List[str]" = None,
-        beta_services_desc: "List[str]" = None,
         prog=None,
         usage=None,
         epilog=None,
         formatter_class=argparse.HelpFormatter,
+        base_desc: str = None,
+        non_beta_services_desc: "List[str]" = None,
+        beta_services_desc: "List[str]" = None,
     ):
         super().__init__(
             prog=prog,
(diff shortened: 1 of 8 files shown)


blackboxsw (changes requested):
I think you are only missing a couple integration tests for this to work across the board (and dropping the `@wip`)
```diff
diff --git a/features/attach_validtoken.feature b/features/attach_validtoken.feature
index b42dd5e..e74d808 100644
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -41,7 +41,6 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         And stdout matches regexp:
         """
         SERVICE       ENTITLED  STATUS    DESCRIPTION
-        esm-apps     +no       +—        +UA Apps: Extended Security Maintenance
         esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance
         livepatch    +yes      +n/a      +Canonical Livepatch service
         """
@@ -88,7 +87,6 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         And stdout matches regexp:
         """
         SERVICE       ENTITLED  STATUS    DESCRIPTION
-        esm-apps     +no       +—        +UA Apps: Extended Security Maintenance
         esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance
         livepatch    +yes      +<lp_status>  +<lp_desc>
         """
```

uaclient/cli.py
```diff
@@ -57,10 +57,52 @@
 
 
 class UAArgumentParser(argparse.ArgumentParser):
+    def __init__(
+        self,
+        base_desc: str = None,
+        non_beta_services_desc: "List[str]" = None,
+        beta_services_desc: "List[str]" = None,
```
blackboxsw, line 64: If we are adding new params to the subclassed __init__ method common practice for the specialization of subclasses generally seems to be adding the new subclass params to the end of the parent's params. This also avoids positional collisions/unexpected behavior if a caller unfamiliar with extended params provided positional args instead of named kwargs.

blackboxsw (approved):

PR #1226, opened 2020-10-28
Install fips metapackage when enabling service

### LINKED ISSUE #1221: Problems when running vm test
When running vm tests, sometimes the `launch` operation fails with the following error:
```
Error: Instance stop operation timed out after 30 seconds
```

This gets us stuck since our test keeps trying to connect to the vm which has not been started. This is problem related to the profile we use to create the vm, but we need to dig deeper on that.

lucasmoura:
Currently, we our fips service is expected to install a range of packages that represent the `FIPS`  service. However, we are now expecting `FIPS`  to provide a single metapackage which will install all of the necessary `FIPS` packages for us. To test this feature, we are using the staging contract environment, since the contract in production is still not updated to use the `FIPS` metapackage.

Also, the BDD test for fips are performed only for bionic at this moment. On xenial. when we install the `FIPS` kernel and reboot the machine we cannot access it anymore through `lxd`. That is because the `FIPS` kernel on xenial is missing the required modules to run the `lxd-agent` component of `lxd`, which allow us to communicate with the vm. As a temporary solution, we are testing `FIPS` on xenial through a manual test.

Finally, this PR also fixes #1221, since the error was being caused when we stopped the vm to add new information to the profile. Since this behaviour has been [fixed](https://github.com/lxc/lxd/issues/7648) on lxd, we no longer need this code.

### DIFF
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -55,3 +55,41 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | focal   | ant      |
            | trusty  | ant      |
            | xenial  | jq       |
+
+    @series.bionic
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable of vm-based services in an ubuntu lxd vm
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token_staging` with sudo
+        And I run `ua disable livepatch` with sudo
+        When I run `ua enable fips --assume-yes --beta` with sudo
+        Then stdout matches regexp:
+            """
+            Updating package lists
+            Installing FIPS packages
+            FIPS enabled
+            A reboot is required to complete install
+            """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            fips          yes                enabled
+            """
+        And I verify that running `apt update` `with sudo` exits `0`
+        And I verify that running `grep Traceback /var/log/ubuntu-advantage.log` `with sudo` exits `1`
+        When I reboot the `<release>` machine
+        And  I run `uname -r` as non-root
+        Then stdout matches regexp:
+            """
+            fips
+            """
+        When I run `cat /proc/sys/crypto/fips_enabled` with sudo
+        Then I will see the following on stdout:
+        """
+        1
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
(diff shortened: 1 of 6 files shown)


features/staging_commands.feature
```diff
@@ -57,3 +57,34 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+            Installing FIPS packages
+            FIPS enabled
+            A reboot is required to complete install
+            """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            fips          yes                enabled
+            """
+        When I reboot the `<release>` machine
```
blackboxsw, line 80: Let's check that apt update is happy before we reboot
```suggestion
        And I verify that running `apt update` `with sudo` exits `0`            
        When I reboot the `<release>` machine
```

blackboxsw (changes requested):
One significant thing we are missing in this branch is support for existing Ubuntu PRO FIPS images which we know have a number of versioned packages pinned via `apt-mark hold` calls.
I can't decide if we want to address this in this PR or a followup, but what this represents I believe is a UA client at the moment which will break existing Ubuntu PRO FIPS images on upgrade. 

The reason this will break existing Ubuntu PRO FIPS images is because there are currently newer packages in the real FIPS PPA than were originally pinned in the AWS PRO FIPS image. This will lead to the newer ubuntu-fips metapackage  requiring specific versions of dependencies that will not be installable  and will result in a broken system.


So, we can discuss either:
   1. addressing the apt-mark unhold logic in a separate PR, making sure to add an integration test for AWS (and Azure) Ubuntu PRO images .  Also ensuring not to leak this commit into our release-25 branch in case CPC ends up rolling a new Ubuntu PRO FIPS image in the meantime.

OR 
    2. address the upgrade path on AWS/Azure PRO and add an integration test for that upgrade path to this commit.

I recognize we don't have much in the way of alternative suggestions from the team about how best to handle the `apt-mark holds` on existing FIPS PRO images, but maybe we can meet on this tomorrow to discuss a workable solution.

lucasmoura:
@blackboxsw I think we can have this work in a separate PR. The reason is we still need to decide how we are gonna install the `FIPS` metapackage in the upgrade scenario on the `FIPS PRO` machines. But I definitely agree that we can discuss this as soon as possible to reach a working solution

lucasmoura:
@blackboxsw Code updated

blackboxsw (approved):
+1 looks excellent. Let's wrap up the support of ubuntu-fips metapkg install on existing FIPS PRO machines as separate PRs.

PR #1227, opened 2020-10-28
Fix azure generic integration testing

lucasmoura:
Master build is failing due to an azure generic test that expects `esm-apps` to appear on `ua status` output. We are now removing this from the test


blackboxsw (approved):
+1

PR #1229, opened 2020-10-29
Update esm-apps check on staging_commands tests

lucasmoura:
Because we had deleted the staging token variable from our travis setting, some integration tests were not running. After adding that token again, we can see that we still have some `esm-apps`  related issues on the `staging_commands`  test. This PR fixes that issue


blackboxsw (approved):
LGTM

PR #1232, opened 2020-11-03
Fix shell condition in run_azure_travis_integration_tests.sh

OddBloke:
Went to look at the output of a job to check the exact version being used by the tests, and noticed this easy-to-fix error.


blackboxsw (approved):
And that explains why build_pr is still incorrectly True on latest master cron job.
Thanks @OddBloke

PR #1233, opened 2020-11-04
Update azure jobs on travis

lucasmoura:
Test pycloudlib new dependecies specified on this [PR](https://github.com/canonical/pycloudlib/pull/42)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,11 +135,6 @@ matrix:
       install:
           - make testdeps
           - git clone https://github.com/canonical/server-test-scripts.git
-          - git clone https://github.com/canonical/pycloudlib.git
-          - cd pycloudlib
-          - pip install -r requirements.txt
-          - pip install -e .
-          - cd ..
       script:
           - ./tools/run_azure_travis_integration_tests.sh
     - if: env(UACLIENT_BEHAVE_AZ_CLIENT_SECRET)
@@ -154,11 +149,6 @@ matrix:
       install:
           - make testdeps
           - git clone https://github.com/canonical/server-test-scripts.git
-          - git clone https://github.com/canonical/pycloudlib.git
-          - cd pycloudlib
-          - pip install -r requirements.txt
-          - pip install -e .
-          - cd ..
       script:
           - ./tools/run_azure_travis_integration_tests.sh
     - if: env(UACLIENT_BEHAVE_AZ_CLIENT_SECRET) AND type = cron
@@ -173,11 +163,6 @@ matrix:
       install:
           - make testdeps
           - git clone https://github.com/canonical/server-test-scripts.git
-          - git clone https://github.com/canonical/pycloudlib.git
-          - cd pycloudlib
-          - pip install -r requirements.txt
-          - pip install -e .
-          - cd ..
       script:
           - ./tools/run_azure_travis_integration_tests.sh
     - if: env(UACLIENT_BEHAVE_AZ_CLIENT_SECRET) AND type = cron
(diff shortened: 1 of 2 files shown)


blackboxsw:
From Travis CI PR perspective we aren't really testing @OddBloke's branch because of our common pycloudlib setup we do directly in .travis.yml. You'll see in
[pycloudlib upstream/master is being used by this CI run](https://travis-ci.com/github/canonical/ubuntu-advantage-client/jobs/428040143#L235)


That said, I think we probably need:
  1. update all references in .travis.yml "install" sections  in this PR
  2. Confirm that running `tox -e behave-awsgeneric-18.04` succeds as that would download updated pycloudlib as defined in integration-tests.txt
  3. Confirm that running `tox - ebehave-azurepro-18.04` succeeds as well

I'm mid local run of `tox -e behave-awsgeneric-18.04` now and will report back here with results.

lucasmoura:
That's a great catch. But it seems that this clone step is only being performed directly on `azure`  tests and the cleanup ones. It seems that the `aws` test that we have is not cloning pycloudlib directly from master.

But still, there is no reason to keep that behavior there for `azure` tests

OddBloke:
@blackboxsw @lucasmoura I can't quite tell from this conversation whether or not you're satisfied with the testing you've performed, can you confirm?

lucasmoura:
@OddBloke I think the test now is working for both AWS and Azure. @blackboxsw do we think we could still improve in how we perforn this test ?

lucasmoura:
@blackboxsw I have updated this PR to only touch the Azure jobs. Now, we are:

* Installing pycloudlib through `tox` instead of manually cloning and installing the project.
* Update order of azure jobs. Recently, we have seen azure jobs failing because they reach our quota limit. My hypothesis here is that we may be spinning all 8 azure jobs at the same time and maybe we end up not being able to create resource groups for all of them. We are now interleaving a batch of azure jobs with aws jobs to verify if that solves the issue.

.travis.yml
```diff
@@ -192,94 +177,34 @@ matrix:
...
           - git clone https://github.com/canonical/server-test-scripts.git
-          - git clone https://github.com/canonical/pycloudlib.git
-          - cd pycloudlib
-          - pip install -r requirements.txt
-          - pip install -e .
-          - cd ..
       script:
           - ./tools/run_azure_travis_integration_tests.sh
-    - if: env(UACLIENT_BEHAVE_AZ_CLIENT_SECRET) AND env(UACLIENT_BEHAVE_CONTRACT_TOKEN) AND type = cron
+    - if: env(UACLIENT_BEHAVE_AWS_SECRET_ACCESS_KEY) AND type = cron
```
blackboxsw, line 182: Good fix.

blackboxsw (changes requested):
@lucasmoura I don't think we can pass this PR as is. removing the pre-install of pycloublib on the cloud jobs results in not doing cleanup look at lines

https://travis-ci.com/github/canonical/ubuntu-advantage-client/jobs/434777334#L549-L553

I think we minimally need to move pycloudlib clone and setup into tools/run_azure_travis_integration_tests.sh and tools/base_travis_integration_tests.sh before trying to call azure_cleanup.py or ec2_cleanup.py respectively.

lucasmoura:
@blackboxsw That's a great catch, I missed the azure clean-up scripts. Let me update that

tools/run_azure_travis_integration_tests.sh
```diff
@@ -1,6 +1,14 @@
 #!/bin/bash
 source tools/base_travis_integration_tests.sh
 
+install_pycloudlib() {
+    git clone https://github.com/canonical/pycloudlib.git
+    cd pycloudlib
+    pip install -r requirements.txt
+    pip install -e .
```
blackboxsw, line 8: I think we can drop this now because pycloudlib's requirements.txt does just that

blackboxsw (approved):

Issue #1168, opened 2020-08-26
on ua disable fips, should we remove all apt config  and leave installed packages?


blackboxsw:
Questions to answer here:
- [ ] Currently we "do nothing" related to the FIPS apt configuraiton files on FIPS because "fips can't be disabled". Should we at least turn reomve apt config files so apt-cache policy shows no access to FIPS repos and ua status shows disabled?
- [x] For services which have installed pkgs by default, should disabling the service leave those packages installed (I think yes but wanted confirmation)   # resolved in 0371ce5a91f4683e8b82826405b385ea356b148c

mitechie:
The second one is yes, if you've enabled something and gotten it running we don't keep you from having that but keep you from future updates. The product offering is the next update vs the one you already have. 

For the first one, I'm torn on this. FIPS is a kernel/package version that should be easily reversible. You should be able to install, reboot into FIPS, test something, and go back. That said, for safety sake, I'd suggest that we'd remove the report and access to future updates and helpfully output some messaging around switching the default kernel but not actively change anything.

PR #1238, opened 2020-11-10
fips: allow disable

blackboxsw:
ua disable fips or fips-update will now uninstall the ubuntu-fips
meta package. This will remove package holds and disable fips-related
settings in grub.

Add post_disable messaging mentioning reboot-required.

Fixes: #1168

### DIFF
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -88,7 +88,18 @@ Feature: Enable command behaviour when attached to an UA staging subscription
         """
         1
         """
-
+        When I run `ua disable fips --assume-yes` with sudo
+        Then stdout matches regexp:
+            """
+            Updating package lists
+            A reboot is required to complete disable operation
+            """
+        When I reboot the `<release>` machine
+        When I run `cat /proc/sys/crypto/fips_enabled` with sudo
+        Then I will see the following on stdout:
+        """
+        0
+        """
         Examples: ubuntu release
            | release |
            | xenial  |
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -1,3 +1,4 @@
+from uaclient import apt
 from uaclient.entitlements import repo
 from uaclient import status, util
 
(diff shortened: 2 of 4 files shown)


uaclient/entitlements/fips.py
```diff
@@ -62,15 +63,27 @@ def application_status(self) -> "Tuple[status.ApplicationStatus, str]":
...
-    def disable(self, silent: bool = False) -> bool:
-        """FIPS cannot be disabled, so simply display a message to the user"""
-        if not silent:
-            print("Warning: no option to disable {}".format(self.title))
-        return False
-
-    def _cleanup(self) -> None:
-        """FIPS can't be cleaned up automatically, so don't do anything"""
-        pass
+    def remove_packages(self) -> None:
```
lucasmoura, line 66: Why are we removing FIPS packages on disable ? If the problem is the package holds, shouldn't we just unhold them on disable ?

blackboxsw: We are only removing `ubuntu-fips` meta package in this operation, the meta-package (on removal) will unset the grub config options which essentially deactivates the fips -hmac packages so the pkg needs to be removed in this case. It is not an operation that we should be doing for any other entitlement.

features/staging_commands.feature
```diff
@@ -88,7 +88,19 @@ Feature: Enable command behaviour when attached to an UA staging subscription
         """
         1
         """
-
+        When I run `ua disable fips --assume-yes` with sudo
+        Then stdout matches regexp:
+            """
+            This will disable access to certified FIPS packages.
```
lucasmoura, line 94: When running locally, the test fails because the sentence:
```
This will disable access to certified FIPS packages.
```

is not being displayed during the disable operation

blackboxsw: Ahh right, that output is only presented without the --assum-yes option.

lucasmoura (approved):

PR #1241, opened 2020-11-11
Refactor integration tests to use pycloudlib LXD

lucasmoura:
Refactor integration tests to use pycloudlib LXD implementation. This also allows us to test enabling `FIPS` for xenial.

PS: This PR depends on the following pycloudlib [PR](https://github.com/canonical/pycloudlib/pull/40)

### DIFF
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -77,6 +77,7 @@ class Cloud:
     def _create_instance(
         self,
         series: str,
+        instance_name: "Optional[str]" = None,
         image_name: "Optional[str]" = None,
         user_data: "Optional[str]" = None,
     ) -> pycloudlib.instance:
@@ -86,6 +87,8 @@ class Cloud:
             The ubuntu release to be used when creating an instance. We will
             create an image based on this value if the used does not provide
             a image_name value
+        :param instance_name:
+            The name of the instance to be created
         :param image_name:
             The name of the image to be used when creating the instance
         :param user_data:
@@ -99,6 +102,7 @@ class Cloud:
     def launch(
         self,
         series: str,
+        instance_name: "Optional[str]" = None,
         image_name: "Optional[str]" = None,
         user_data: "Optional[str]" = None,
     ) -> pycloudlib.instance.BaseInstance:
@@ -108,6 +112,8 @@ class Cloud:
             The ubuntu release to be used when creating an instance. We will
             create an image based on this value if the used does not provide
             a image_name value
+        :param instance_name:
+            The name of the instance to be created
         :param image_name:
             The name of the image to be used when creating the instance
         :param user_data:
@@ -116,10 +122,15 @@ class Cloud:
         :returns:
             An cloud provider instance
         """
-        inst = self._create_instance(series, image_name, user_data)
+        inst = self._create_instance(
+            series=series,
+            instance_name=instance_name,
+            image_name=image_name,
+            user_data=user_data,
+        )
         print(
             "--- {} instance launched: {}. Waiting for ssh access".format(
-                self.name, inst.id
+                self.name, inst.name
             )
         )
         time.sleep(15)
@@ -289,6 +300,7 @@ class EC2(Cloud):
     def _create_instance(
         self,
         series: str,
+        instance_name: "Optional[str]" = None,
         image_name: "Optional[str]" = None,
         user_data: "Optional[str]" = None,
     ) -> pycloudlib.instance:
@@ -298,6 +310,8 @@ class EC2(Cloud):
             The ubuntu release to be used when creating an instance. We will
             create an image based on this value if the used does not provide
             a image_name value
+        :param instance_name:
+            The name of the instance to be created
         :param image_name:
             The name of the image to be used when creating the instance
         :param user_data:
@@ -439,6 +453,7 @@ class Azure(Cloud):
     def _create_instance(
         self,
         series: str,
+        instance_name: "Optional[str]" = None,
         image_name: "Optional[str]" = None,
         user_data: "Optional[str]" = None,
     ) -> pycloudlib.instance:
@@ -448,6 +463,8 @@ class Azure(Cloud):
             The ubuntu release to be used when creating an instance. We will
             create an image based on this value if the used does not provide
             a image_name value
+        :param instance_name:
+            The name of the instance to be created
         :param image_name:
             The name of the image to be used when creating the instance
         :param user_data:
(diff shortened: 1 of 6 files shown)


features/cloud.py
```diff
@@ -462,3 +476,108 @@ def _create_instance(
...
+        return self._api
+
+    def manage_ssh_key(self, private_key_path: "Optional[str]" = None) -> None:
+        """Create and manage ssh key pairs to be used in the cloud provider.
+
+        :param private_key_path:
+            Location of the private key path to use. If None, the location
+            will be a default location.
+        """
+        pub_key_path = "lxd-pubkey"
```
blackboxsw, line 499: As you mentioned in passing, the integration tests currently take 4 times as long running via ssh to the lxd instance instead of lxc exec. Should this method be a noop (pass) so that lxd testing doesn't use ssh by default during typical per-PR integration tests?

features/cloud.py
```diff
@@ -462,3 +476,108 @@ def _create_instance(
...
+
+        :returns:
+            An AWS cloud provider instance
+        """
+        is_vm = bool(self.machine_type == "lxd.vm")
+
+        if not image_name:
+            image_name = self.locate_image_name(series)
+
+        print("--- Launching LXD image {}({})".format(image_name, series))
```
blackboxsw, line 541: let's add a "vm" suffix to the message if vm is launched.

blackboxsw (commented):
just a quick pass on this. will get detailed review tomorrow.

lucasmoura:
@blackboxsw One idea that I had is that we can use the ssh option only for vm tests, since it is only those tests that require the use of ssh.

To make that possible, we need to first merge this [PR](https://github.com/canonical/pycloudlib/pull/45)

lucasmoura:
@blackboxsw I have update the code with the refactoring and FIPS tests.

Right now, the `FIPS Updates` upgrade test is just assuring that we end up in a scenario were we have the FIPS metapackage uninstalled because of the `do-release-upgrade` issue. Once it has been fixed, I can update the test without an issue.

features/cloud.py
```diff
@@ -462,3 +493,156 @@ def _create_instance(
...
+        :param private_key_path:
+            Location of the private key path to use. If None, the location
+            will be a default location.
+        """
+        pub_key_path = "lxd-pubkey"
+        priv_key_path = "lxd-privkey"
+        pub_key, priv_key = self.api.create_key_pair()
+
+        with open(pub_key_path, "w") as f:
+            f.write(pub_key)
```
blackboxsw, line 589: We probably want to os.chmod these as 600 so that developers could use them on the cmdline to access instances using `ssh -i lxd-pubkey ubuntu@<ip>`

features/cloud.py
```diff
@@ -462,3 +493,156 @@ def _create_instance(
...
+            public_key_path=pub_key_path, private_key_path=priv_key_path
+        )
+
+    @property
+    def api(self) -> pycloudlib.cloud.BaseCloud:
+        """Return the api used to interact with the cloud provider."""
+        if self._api is None:
+            self._api = pycloudlib.LXDVirtualMachine(tag=self.tag)
+
+        return self._api
```
blackboxsw, line 604: We might be able to define this simply in either in _LXD base class or even Cloud if we use a class attr to define the pycloublib.*Cloud callable to create

```diffdiff --git a/features/cloud.py b/features/cloud.py
index 38114a6..c2b8600 100644
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -29,6 +29,7 @@ class Cloud:
     name = ""
     pro_ids_path = ""
     env_vars: "Tuple[str, ...]" = ()
+    pycloudlib_cls = None  # type: pycloudlib.cloud.BaseCloud
 
     def __init__(
         self,
@@ -64,7 +65,13 @@ class Cloud:
     @property
     def api(self) -> pycloudlib.cloud.BaseCloud:
         """Return the api used to interact with the cloud provider."""
-        raise NotImplementedError
+        if self.pycloudlib_cls is None:
+            raise NotImplementedError
+        if self._api is None:
+            self._api = self.pycloudlib_cls(
+                tag=self.tag, timestamp_suffix=self.timestamp_suffix
+            )
+        return self.api
 
     def manage_ssh_key(self, private_key_path: "Optional[str]" = None) -> None:
         """Create and manage ssh key pairs to be used in the cloud provider.
@@ -496,8 +503,6 @@ class Azure(Cloud):
 
 
 class _LXD(Cloud):
-    name = "_lxd"
-
     def _create_instance(
         self,
         series: str,
@@ -573,6 +578,7 @@ class _LXD(Cloud):
 
 class LXDVirtualMachine(_LXD):
     name = "lxd-virtual-machine"
+    pycloudlib_cls = pycloudlib.LXDVirtualMachine
 
     def manage_ssh_key(self, private_key_path: "Optional[str]" = None) -> None:
         """Create and manage ssh key pairs to be used in the cloud provider.
@@ -606,6 +612,7 @@ class LXDVirtualMachine(_LXD):
 
 class LXDContainer(_LXD):
     name = "lxd-virtual-machine"
+    pycloudlib_cls = pycloudlib.LXDContainer
 
     def manage_ssh_key(self, private_key_path: "Optional[str]" = None) -> None:
         """Create and manage ssh key pairs to be used in the cloud provider.
```

features/cloud.py
```diff
@@ -462,3 +493,156 @@ def _create_instance(
...
+        lxd_cmd = 'su ubuntu -c "{}"'.format(base_cmd)
+        cmd = shlex.split(lxd_cmd)
+
+        return super().execute_command(instance=instance, cmd=cmd)
+
+    @property
+    def api(self) -> pycloudlib.cloud.BaseCloud:
+        """Return the api used to interact with the cloud provider."""
+        if self._api is None:
+            self._api = pycloudlib.LXDContainer(tag=self.tag)
```
blackboxsw, line 646: let's also not forget timestamp_suffix=self.timestamp_suffix

features/cloud.py
```diff
@@ -462,3 +493,156 @@ def _create_instance(
...
+
+        self.api.use_key(
+            public_key_path=pub_key_path, private_key_path=priv_key_path
+        )
+
+    @property
+    def api(self) -> pycloudlib.cloud.BaseCloud:
+        """Return the api used to interact with the cloud provider."""
+        if self._api is None:
+            self._api = pycloudlib.LXDVirtualMachine(tag=self.tag)
```
blackboxsw, line 602: don't forget timestamp_suffix=self.timestamp_suffix

features/steps/steps.py
```diff
@@ -13,19 +13,12 @@
...
-    nullcontext,
-    wait_for_boot,
-)
+from features.util import SLOW_CMDS, emit_spinner_on_travis, nullcontext
 
 from uaclient.defaults import DEFAULT_CONFIG_FILE
 
 
-CONTAINER_PREFIX = "behave-test-"
+CONTAINER_PREFIX = "ubuntu-behave-test-"
```
blackboxsw, line 21: Why change this?

blackboxsw (changes requested):
As you mentioned in chat. something is occuring with the shlex usage when trying to run cmds that emit json since we are splitting the cmdstring blindly on spaces. I've suggested a few changes inline but I think we also may have to sort to remaining checks in features/environment which check for a possible path where cloud_api is not set since I don't think we would have that path anymore right?

- features/environment: (before_all)  ` if context.config.cloud_api:`
- features/environment:583 (_install_uat_in_container)  `if not config.cloud_api`

features/steps/steps.py
lucasmoura: This is just to guarantee that the instance name will be properly parsed by this regex:
https://github.com/canonical/pycloudlib/blob/master/pycloudlib/lxd/cloud.py#L580

lucasmoura:
@blackboxsw Thanks for the review and great catch on the leftover code for checking cloud_api.

I have updated the code and I think instead of worrying about shellify at the moment, we should wait for this pycloudlib [PR](https://github.com/canonical/pycloudlib/pull/56) to land. When that happens, we will no longer need the `execute_command` method anymore

blackboxsw (approved):
Much simpler to use the pycloudlib to rule all the platforms. Thanks for this work!

PR #1247, opened 2020-11-19
Update esm-infra lxd test

lucasmoura:
On the `Attached enable of a know service shows update in a ubuntu machine` we were failing to install the `libkrad0` package on xenial, which makes the test fail. We are now confirming that the package is installed while updating the package version that is installed on xenial.


features/attached_enable.feature
```diff
@@ -127,8 +127,8 @@ Feature: Enable command behaviour when attached to an UA subscription
             ESM Infra is already enabled.
             See: sudo ua status
             """
-        When I run `apt install -y <pkg-version>` with sudo
```
blackboxsw, line 130: When unspecified exit code in our When clause, I think we should actually default to check for non-zero exit and fail otherwise. I don't think we should require any call to append the "exits '0`' suffix.

blackboxsw (approved):
+1 Thanks for tidying this up

Issue #1109, opened 2020-07-06
FIPS: unhold any held fips packages.


blackboxsw:
# BLOCKED on existing team discussion
Azure Ubuntu PRO FIPS preview image uses apt-mark hold to pin FIPS packages as part of the build process. On auto-attach uaclient will need to unhold those packages during FIPS enable.

rcj4747:
You mention Azure's image, AWS will be the same.  Just an FYI.

blackboxsw:
Thanks @rcj4747

> You mention Azure's image, AWS will be the same. Just an FYI.

Thanks @rcj4747

blackboxsw:
Moving to milestone 26.0 as milestone 25.0 will likely be just full Xenial, Bionic Eoan and Focal release support. Fips will come in 26.0

PR #1249, opened 2020-11-23
Unhold fips pkgs

blackboxsw:
* FIPSEntitlement.setup_apt_config will unhold deb packages to support Ubuntu PRO FIPS images which were built before UA client official support of FIPS.
* Separate RepoEntitlement.install_packages method out of `enable` method so it can be called during delta processing.
* Handle apt-based contract deltas that include changes in additionalPackages directives during
`repo.process_contract_deltas`. 

Fixes: #1109

### DIFF
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -189,6 +189,7 @@ class RepoEntitlement(base.UAEntitlement):
         delta_entitlement = deltas.get("entitlement", {})
         delta_directives = delta_entitlement.get("directives", {})
         delta_apt_url = delta_directives.get("aptURL")
+        delta_packages = delta_directives.get("additionalPackages")
         status_cache = self.cfg.read_cache("status-cache")
 
         if delta_directives and status_cache:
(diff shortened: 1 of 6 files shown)


uaclient/entitlements/repo.py
```diff
@@ -235,10 +213,42 @@ def process_contract_deltas(
                 apt.remove_auth_apt_repo(repo_filename, old_url)
         self.remove_apt_config()
         self.setup_apt_config()
+        if delta_packages:
```
lucasmoura, line 216: I don't know if we want to support that, but could there be a situation were we would require to uninstall the current packages and install the new ones from the delta ?

lucasmoura (approved):

uaclient/entitlements/repo.py
blackboxsw: @lucasmoura I was considering the same condition. I think if we wanted an explicity removal of packages across deltas, we would probably have to design that with the contract backend if required. Since most of our logic is only to supplement via additionalPackages directive, I'd probably want to see a "removePackages": directive or something similar to mandate that packages get removed. Right now we don't really have a use case for it, so I didn't want to put together a solution for something that "may" not be needed.

PR #1252, opened 2020-12-02
ua-auto-attach: finish auto-attach.service before cloud-config.service starts

blackboxsw:
Declare the ua-auto-attach.service unit as being run explicity Before  and WantedBy cloud-config.service to ensure auto-attach completes before any cloud-config userdata attempts to install or update deb packages.

This guarantees that any packages will be pulled from ESM PPAs if applicable on a given machine if cloud-config requests
package installs.

### DIFF
--- a/systemd/ua-auto-attach.service
+++ b/systemd/ua-auto-attach.service
@@ -1,5 +1,7 @@
 [Unit]
 Description=Ubuntu Advantage auto attach
+Before=cloud-config.service
+WantedBy=cloud-config.service
 
 [Service]
 Type=oneshot


debian/changelog
```diff
@@ -1,4 +1,4 @@
-ubuntu-advantage-tools (26.0) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (26.0~18.04.1) bionic; urgency=medium
```
lucasmoura, line 1: I think we drop this change

blackboxsw: oops dev test leak

blackboxsw:
To test:

- sbuild package locally for bionic
- launch azure bionigi ubuntu pro image with the following userdata:
```
#cloud-config 
package_update: true                                                            
package_upgrade: true  
packages: [ubuntu-dev-tools]
```
- scp ubuntu-advantage-*deb ubuntu@<AZURE_VM>:.
- ssh ubuntu@<AZURE_VM> 'sudo dpkg -i ubuntu-advantage.*deb'
- ssh ubuntu@<AZURE_VM> sudo rm -rf /var/lib/ubuntu-advantage /var/log/ubuntu-advantage.log
- ssh ubuntu@<AZURE_VM> -- sudo cloud-init clean --logs --reboot

lucasmoura (approved):
Approved. Had some troubles installing the `ubuntu-dev-tools` package on Azure, but the package was not necessary to test the feature and the failures are not related to the new feature

Issue #1253, opened 2020-12-02
travis: integration test fixes for LXD and aws cron runs


blackboxsw:
Travis builds for aws and lxd jobs incorrectly continue to set BUILD_PR=1 because of a shell syntax error

resulting in the following failure:

./tools/run_lxd_travis_integration_tests.sh: line 12: [master: command not found

Here's the failed travis build:
https://travis-ci.com/github/canonical/ubuntu-advantage-client/jobs/451552494#L238

PR #1254, opened 2020-12-02
Fix integration test setup scripts

lucasmoura:
One check on both lxd and azure setup scripts were incorrect.
We are fixing that check to allow the builds to function as expected

Fixes: #1253


blackboxsw (approved):
+1 awaiting CI

Issue #1257, opened 2020-12-03
cloud-init bug causes cloud-init status --wait to fail across upgrade trusty -> xenial


blackboxsw:
With the recent switch to using pycloudlib.instance.restart(wait=True) we now see errors in integration tests that expose that cloud-init pickled datasource from trusty causes Tracebacks when unpickling on Xenial after an upgrade during cloud-init's init-local stage.

The error from cloud-init status --wait on xenial during init-local:
 'Paths' object has no attribute 'run_dir' error: sudo: unable to resolve host ubuntu-behave-test-trusty-160

This is [LP #1899299](https://pad.lv/1899299)
Until an SRU presents a fix in cloud-init for this upgrade path, integration tests need to avoid running instance.restart(wait=True) on trusty -> xenial upgrade scenarios.

Otherwise we hit failures like https://travis-ci.com/github/canonical/ubuntu-advantage-client/jobs/452802189#L404

PR #1258, opened 2020-12-03
upgrade-test: fix upgrade path restart failures on trusty

blackboxsw:
When upgrading trusty machines to xenial, cloud-init has a bug in
re-constituting the pickled trusty-based datasource on xenial.

That cloud-init bug is LP: #1899299

This results in the following error from cloud-init status --wait:
  'Paths' object has no attribute 'run_dir'

To avoid this error until an SRU fixes cloud-init on Xenial,
integration tests can use:
   instance.shutdown(wait=True)
   instance.start()
   instance.wait(raise_on_cloudinit_failure=False)

Fixes: #1257

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 import subprocess
 import re
 import shlex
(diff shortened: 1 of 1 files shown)


blackboxsw:
Note that a cron job is running that includes this commit on the 25b3 branch https://travis-ci.com/github/canonical/ubuntu-advantage-client/builds/206183897

lucasmoura (approved):

PR #1260, opened 2020-12-03
Add init script to run remaining uaclient commands after do-release-upgrade

lucasmoura:
After `do-release-upgrade` there could still be operations that should still be performed after a user reboots the machine. For example, when upgrading from trusty to xenial, we cannot enable livepatch after `do-release-upgrade`, only after the user reboots the machine. To handle those situations, we are creating a init script that performs the remaining operations on uaclient to leave the system in a desired state.

What could be improved:

* Right now, the script is running the `upgrade_lts_contract`  script, since we need to perform the exact same operation on reboot. However, our logging can be really confusing because of that, since we will "repeat" the command and this may not be easy to understand why by just looking at the logs
* We can add an integration test for this feature as well

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -185,10 +185,11 @@ case "$1" in
           chmod 0700 "$private_dir"
       fi
 
-      trusty_pkg=$(echo $PREVIOUS_PKG_VER | grep -o 14.04)
-      if [ "$VERSION_ID" = "16.04" ] && [ ! -z "$trusty_pkg" ]; then
+      if [ "$VERSION_ID" = "16.04" ]; then
+        if echo "$PREVIOUS_PKG_VER" | grep -q "14.04"; then
           # trusty upgrade to xenial has known post-reboot operations required (livepatch)
           mark_reboot_cmds_as_needed
+        fi
       fi
       ;;
 esac
--- a/debian/postinst
+++ b/debian/postinst
@@ -29,6 +29,8 @@ SYSTEMD_WANTS_AUTO_ATTACH_LINK="/etc/systemd/system/multi-user.target.wants/ua-a
 SYSTEMD_HELPER_ENABLED_AUTO_ATTACH_DSH="/var/lib/systemd/deb-systemd-helper-enabled/ua-auto-attach.service.dsh-also"
 SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/multi-user.target.wants/ua-auto-attach.service"
 
+REBOOT_CMD_MARKER_FILE="/var/lib/ubuntu-advantage/marker-reboot-cmds-required"
+
 # Rename apt config files for ua services removing ubuntu release names
 redact_ubuntu_release_from_ua_apt_filenames() {
     DIR=$1
@@ -113,6 +115,12 @@ EOF
     fi
 }
 
+mark_reboot_cmds_as_needed() {
+    if [ ! -f "$REBOOT_CMD_MARKER_FILE" ]; then
+      touch $REBOOT_CMD_MARKER_FILE
+    fi
+}
+
 case "$1" in
     configure)
       PREVIOUS_PKG_VER=$2
@@ -175,6 +183,12 @@ case "$1" in
       private_dir="/var/lib/ubuntu-advantage/private"
       if [ -d "$private_dir" ]; then
           chmod 0700 "$private_dir"
+      fi
+
+      trusty_pkg=$(echo $PREVIOUS_PKG_VER | grep -o 14.04)
+      if [ "$VERSION_ID" = "16.04" ] && [ ! -z "$trusty_pkg" ]; then
+          # trusty upgrade to xenial has known post-reboot operations required (livepatch)
+          mark_reboot_cmds_as_needed
       fi
       ;;
 esac
--- a/lib/reboot_cmds.py
+++ b/lib/reboot_cmds.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python3
+
+"""
+Some uaclient operations cannot be fully completed by running a single
+command. For example, when upgrading uaclient from trusty to xenial,
+we may have a livepatch change in the contract, allowing livepatch to be
+enabled on xenial. However, during the upgrade we cannot install livepatch on
+the system, only after a reboot.
+
+To allow uaclient to postpone commands that need to be executed in a system boot,
+we are using this script, which will basically try to reprocess contract deltas
+on boot time.
+"""
+import logging
+import os
+import sys
+
+from uaclient import config
+
+from uaclient.util import subp, ProcessExecutionError, del_file
+from uaclient.cli import setup_logging, assert_lock_file
+
+
+def run_command(cmd, cfg):
+    try:
+        out, _ = subp(cmd.split(), capture=True)
+        logging.debug("Successfully executed cmd: {}".format(cmd))
+    except ProcessExecutionError as exec_error:
+        msg = (
+            "Failed running cmd: {}\n"
+            "Return code: {}\n"
+            "Stderr: {}\n"
+            "Stdout: {}".format(
+                cmd, exec_error.exit_code, exec_error.stderr, exec_error.stdout
+            )
+        )
+
+        reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
+        del_file(reboot_cmd_marker_file)
+
+        logging.debug(msg)
+        sys.exit(1)
+
+
+def refresh_contract(args, cfg):
+    cmd = "ua refresh"
+    run_command(cmd=cmd, cfg=cfg)
+
+
+@assert_lock_file("ua-reboot-process-deltas")
+def process_remaining_deltas(args, cfg):
+    cmd = "/usr/bin/python3 /usr/lib/ubuntu-advantage/upgrade_lts_contract.py"
+    run_command(cmd=cmd, cfg=cfg)
+
+
+def main(args, cfg):
+    setup_logging(logging.INFO, logging.DEBUG)
+    reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
+
+    if not cfg.is_attached:
+        logging.debug("Skiping reboot_cmds. Machine is unattached")
+
+        if os.path.exists(reboot_cmd_marker_file):
+            del_file(reboot_cmd_marker_file)
+
+        return
+
+    if os.path.exists(reboot_cmd_marker_file):
+        logging.debug(
+            "Running process contract deltas on reboot ...".format(
+                reboot_cmd_marker_file
+            )
+        )
+
+        refresh_contract(args, cfg)
+        process_remaining_deltas(args, cfg)
+
+        del_file(reboot_cmd_marker_file)
+
+        logging.debug(
+            "Completed running process contract deltas on reboot ..."
+        )
+
+
+if __name__ == "__main__":
+    cfg = config.UAConfig()
+    main(args=None, cfg=cfg)
--- a/lib/reboot_cmds.py
+++ b/lib/reboot_cmds.py
@@ -17,7 +17,7 @@ import sys
 
 from uaclient import config
 
-from uaclient.util import subp, ProcessExecutionError, del_file
+from uaclient.util import subp, ProcessExecutionError
 from uaclient.cli import setup_logging, assert_lock_file
 
 
@@ -35,8 +35,7 @@ def run_command(cmd, cfg):
             )
         )
 
-        reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
-        del_file(reboot_cmd_marker_file)
+        cfg.delete_cache_key("marker-reboot-cmds")
 
         logging.warning(msg)
         sys.exit(1)
@@ -61,21 +60,17 @@ def main(args, cfg):
         logging.debug("Skipping reboot_cmds. Machine is unattached")
 
         if os.path.exists(reboot_cmd_marker_file):
-            del_file(reboot_cmd_marker_file)
+            cfg.delete_cache_key("marker-reboot-cmds")
 
         return
 
     if os.path.exists(reboot_cmd_marker_file):
-        logging.debug(
-            "Running process contract deltas on reboot ...".format(
-                reboot_cmd_marker_file
-            )
-        )
+        logging.debug("Running process contract deltas on reboot ...")
 
         refresh_contract(args, cfg)
         process_remaining_deltas(args, cfg)
 
-        del_file(reboot_cmd_marker_file)
+        cfg.delete_cache_key("marker-reboot-cmds")
 
         logging.debug(
             "Completed running process contract deltas on reboot ..."
--- a/lib/reboot_cmds.py
+++ b/lib/reboot_cmds.py
@@ -58,7 +58,7 @@ def main(args, cfg):
     reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
 
     if not cfg.is_attached:
-        logging.debug("Skiping reboot_cmds. Machine is unattached")
+        logging.debug("Skipping reboot_cmds. Machine is unattached")
 
         if os.path.exists(reboot_cmd_marker_file):
             del_file(reboot_cmd_marker_file)
--- a/lib/reboot_cmds.py
+++ b/lib/reboot_cmds.py
@@ -38,7 +38,7 @@ def run_command(cmd, cfg):
         reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
         del_file(reboot_cmd_marker_file)
 
-        logging.debug(msg)
+        logging.warning(msg)
         sys.exit(1)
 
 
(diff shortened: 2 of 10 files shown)


lib/reboot_cmds.py
```diff
@@ -0,0 +1,77 @@
...
+from uaclient import config
+
+from uaclient.util import subp, ProcessExecutionError
+from uaclient.cli import setup_logging, assert_lock_file
+
+
+CMDS_FILE_LOCATION = "/etc/ubuntu-advantage/reboot-cmds-needed"
+
+
+def run_command(cmd):
```
mitechie, line 27: Just an FYI this being more generic might be useful. I know in the ua fix work coming up we'll be triggering ua commands on behalf of the user in response to the CLI input. Not required for this PR but to start pondering about.

lib/reboot_cmds.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def refresh_contract(args, cfg):
+    cmd = "ua refresh"
+    run_command(cmd=cmd)
+
+
+@assert_lock_file("ua-reboot-process-deltas")
+def process_remaining_deltas(args, cfg):
+    cmd = "/usr/bin/python3 /usr/lib/ubuntu-advantage/upgrade_lts_contract.py"
```
mitechie, line 51: Is there somewhere that this is triggered based on facts of state? This is going to be run every startup. I understand refreshing with ua refresh, but I'm missing the step that interprets the changes in the state that determines that upgrade_lts_contract.py needs to run?

lucasmoura: The approach that we followed here is that the `upgrade_lts_contract` should be robust enough to identify if the system is already in the state that we desire. And if that is the case, it will not perform any operations. For example, suppose we have two deltas:

esm-infra: aptURL changes
livepathc: enabledByDefault is now True

If esm-infra is already configured with the aptURL delta and livepatch is already enabled, we will not perform anything.

Also, if the script succeeds on running all of the commands, we will no longer run it after every reboot.

debian/postinst
```diff
@@ -176,6 +184,11 @@ case "$1" in
       if [ -d "$private_dir" ]; then
           chmod 0700 "$private_dir"
       fi
+
+      trusty_pkg=$(echo $PREVIOUS_PKG_VER | grep -o 14.04)
+      if [ "$VERSION_ID" = "16.04" ] && [ ! -z "$trusty_pkg" ]; then
+          mark_reboot_cmds_as_needed
```
blackboxsw, line 190: Let's add a comment here
```suggestion
          # trusty upgrade to xenial has known post-reboot operations required (livepatch)
          mark_reboot_cmds_as_needed
```

lib/reboot_cmds.py
```diff
@@ -0,0 +1,77 @@
...
+import os
+import sys
+
+from uaclient import config
+
+from uaclient.util import subp, ProcessExecutionError
+from uaclient.cli import setup_logging, assert_lock_file
+
+
+CMDS_FILE_LOCATION = "/etc/ubuntu-advantage/reboot-cmds-needed"
```
blackboxsw, line 24: probably under /var/lib/ubuntu-advantage
Also variable name (CMDS_FILE_LOCATION) and filename (reboot-cmds-need) may mislead us in the future that this is something more than just a marker file.
Let's call it something that included marker or flag-file

systemd/ua-reboot-cmds.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Ubuntu Advantage reboot cmds
+
```
blackboxsw, line 3: I think we need a ConditionPathExists=/var/lib/ubuntu-advantage/marker-reboot-cmds-required here so this is only included if the marker file exists?
```suggestion
ConditionPathExists=/var/lib/ubuntu-advantage/marker-reboot-cmds-required

```

blackboxsw (commented):
Looks really good @lucasmoura .  Initial round of review here. I'll play around with it on lxc and see what else comes up.

lucasmoura:
@blackboxsw I have updated the code. Also just realized that we cannot create a integration test for this feature because we cannot support trusty vms on LXD.

However, this is the bash script I am using to test the feature:
```
#!/bin/sh
set -x

series=trusty
name=$series-upgrade

multipass delete $name
multipass purge
multipass launch $series --name $name

multipass exec $name -- sudo add-apt-repository ppa:lamoura/uaclient-test -y
multipass exec $name -- sudo apt-get update
multipass exec $name -- sudo apt-get install ubuntu-advantage-tools -y
multipass exec $name -- ua version
multipass exec $name -- sudo ua attach $UACLIENT_BEHAVE_CONTRACT_TOKEN
multipass exec $name -- sudo ua status
multipass exec $name -- sudo mkdir -p /etc/update-manager/release-upgrades.d  
multipass exec $name -- sh -c "echo '[Sources]\nAllowThirdParty=yes' > allow.cfg"
multipass exec $name -- sudo mv allow.cfg /etc/update-manager/release-upgrades.d   
multipass exec $name -- sudo do-release-upgrade --frontend DistUpgradeViewNonInteractive

multipass exec $name -- sudo reboot
sleep 20
multipass exec $name -- sudo ua status --wait

set +x
```

lib/reboot_cmds.py
```diff
@@ -0,0 +1,86 @@
...
+import os
+import sys
+
+from uaclient import config
+
+from uaclient.util import subp, ProcessExecutionError
+from uaclient.cli import setup_logging, assert_lock_file
+
+
+BASE_PATH = "/var/lib/ubuntu-advantage/"
```
blackboxsw, line 24: This value should  be sourced from UAConfig.data_dir because that value could have been modified in /etc/ubuntu-advantage/uaclient.conf

lib/reboot_cmds.py
```diff
@@ -0,0 +1,86 @@
...
+    except ProcessExecutionError as exec_error:
+        msg = (
+            "Failed running cmd: {}\n"
+            "Return code: {}\n"
+            "Stderr: {}\n"
+            "Stdout: {}".format(
+                cmd, exec_error.exit_code, exec_error.stderr, exec_error.stdout
+            )
+        )
+        logging.debug(msg)
```
blackboxsw, line 41: do we want to remove REBOOT_CMD_MARKER_FILE here? or Leave it in place to retry on next reboot?

lib/reboot_cmds.py
```diff
@@ -0,0 +1,86 @@
...
+from uaclient.cli import setup_logging, assert_lock_file
+
+
+BASE_PATH = "/var/lib/ubuntu-advantage/"
+REBOOT_CMD_MARKER_FILE = BASE_PATH + "marker-reboot-cmds-required"
+
+
+def run_command(cmd):
+    try:
+        out, _ = subp(cmd.split())
```
blackboxsw, line 30: Probably want to pass capture=True here to log the command and output to /var/log/ubuntu-advantage.log

lib/reboot_cmds.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def main(args, cfg):
+    setup_logging(logging.INFO, logging.DEBUG)
+
+    if not cfg.is_attached:
+        logging.debug("Skiping reboot_cmds. Machine is unattached")
+
+        if os.path.exists(REBOOT_CMD_MARKER_FILE):
+            os.remove(REBOOT_CMD_MARKER_FILE)
```
blackboxsw, line 63: Let's use our del_file utility func to avoid this duplication of check/remove
```suggestion
        util.del_file(REBOOT_CMD_MARKER_FILE)
```

lib/reboot_cmds.py
```diff
@@ -0,0 +1,86 @@
...
+        logging.debug(
+            "Running process contract deltas on reboot ...".format(
+                REBOOT_CMD_MARKER_FILE
+            )
+        )
+
+        refresh_contract(args, cfg)
+        process_remaining_deltas(args, cfg)
+
+        os.remove(REBOOT_CMD_MARKER_FILE)
```
blackboxsw, line 77:
```suggestion
        util.del_file(REBOOT_CMD_MARKER_FILE)
```

uaclient/entitlements/repo.py
```diff
@@ -165,6 +165,26 @@ def application_status(self) -> "Tuple[ApplicationStatus, str]":
...
+        """
+        if not delta_apt_url:
+            return False
+
+        apt_file = self.repo_list_file_tmpl.format(name=self.name)
+
+        if os.path.exists(apt_file):
+            with open(apt_file, "r") as f:
+                if delta_apt_url in f.read():
+                    return False
```
blackboxsw, line 184: Let's use our utility functions
```suggestion
            return bool(delta_apt_url in util.load_file(apt_file))
```

blackboxsw (changes requested):
a couple more nits and I think this looks good.

uaclient/entitlements/repo.py
```diff
@@ -165,6 +165,26 @@ def application_status(self) -> "Tuple[ApplicationStatus, str]":
             "{} is not configured".format(self.title),
         )
 
+    def _check_apt_url_delta_should_be_applied(self, delta_apt_url):
```
blackboxsw, line 168: We may want this to be more generic than across apt deltas right? I was thinking this would be built in as a check_apt_url_is_applied and not really just dealing with deltas. Then we can check current contract state for our running ubuntu release, and validate that the aptURL specified is already applied to the self.repo_list_file_tmpl.format(name=self.name). If not applied, then apply it.

lib/reboot_cmds.py
lucasmoura: That's a good question. Ideally, this decision should be based on the error context. For example, on network error issues, I do believe that we should maintain the marker file. However, I don't know how easy it would be to have that granularity in the code.

I believe that, initially, we could treat this operation as an one-shot attempt to get the system in expected state. If we fail, we can abort.

But if you belive we can treat well known errors, I am okay with that as well

uaclient/entitlements/repo.py
lucasmoura: Yes, we could make the name of the function more generic. The reason I was so specific here is just because this method is only being used on the `process_contract_deltas` method. But I do agree we should aim at being more generic here.

lib/reboot_cmds.py
```diff
@@ -0,0 +1,87 @@
...
+            "Stderr: {}\n"
+            "Stdout: {}".format(
+                cmd, exec_error.exit_code, exec_error.stderr, exec_error.stdout
+            )
+        )
+
+        reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
+        del_file(reboot_cmd_marker_file)
+
+        logging.debug(msg)
```
blackboxsw, line 41: Minimally I think this should be a logging.warning instead of debug message to alert us that something went wrong unexpectedly.
```suggestion
        logging.warning(msg)
```

systemd/ua-reboot-cmds.service
```diff
@@ -0,0 +1,11 @@
...
+[Service]
+Type=oneshot
+ConditionPathExists=/var/lib/ubuntu-advantage/marker-reboot-cmds-required
```
blackboxsw, line 6: I think this needs to be in the [Unit] section, othewise we see this message in `journalctl -b` 
` [/lib/systemd/system/ua-reboot-cmds.service:6] Unknown lvalue 'ConditionPathExists' in section 'Service''

blackboxsw (changes requested):
couple of minor changes and I think we are there on this branch.

uaclient/entitlements/repo.py
```diff
@@ -204,17 +218,20 @@ def process_contract_deltas(
...
                 apt.remove_auth_apt_repo(repo_filename, old_url)
-        self.remove_apt_config()
-        self.setup_apt_config()
+
+            self.remove_apt_config()
+            self.setup_apt_config()
+
         if delta_packages:
             self.install_packages(package_list=delta_packages)
+
```
blackboxsw, line 234: Let's also leave a conditional message here if we are updating packages due to deltas
```suggestion
            logging.info msg ("Installing packages on change directives: {}".format(", ".join(delta_packages)) 
            self.install_packages(package_list=delta_packages)

```

uaclient/entitlements/repo.py
```diff
@@ -204,17 +218,20 @@ def process_contract_deltas(
             "Updating '%s' apt sources list on changed directives.", self.name
```
blackboxsw, line 218: As you mentioned in chat, this message shouldn't be emitted if we actually aren't rewriting apt source config files. Let's move this into scope below ```        if not self._check_apt_url_is_applied(delta_apt_url):```

lib/reboot_cmds.py
```diff
@@ -0,0 +1,87 @@
...
+            "Failed running cmd: {}\n"
+            "Return code: {}\n"
+            "Stderr: {}\n"
+            "Stdout: {}".format(
+                cmd, exec_error.exit_code, exec_error.stderr, exec_error.stdout
+            )
+        )
+
+        reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
+        del_file(reboot_cmd_marker_file)
```
blackboxsw, line 39: This should use UAconfig.delte_cache_key instead
```suggestion
        cfg.delete_cache_key("marker-reboot-cmds")
```

lib/reboot_cmds.py
```diff
@@ -0,0 +1,87 @@
...
+        logging.debug(
+            "Running process contract deltas on reboot ...".format(
+                reboot_cmd_marker_file
+            )
+        )
+
+        refresh_contract(args, cfg)
+        process_remaining_deltas(args, cfg)
+
+        del_file(reboot_cmd_marker_file)
```
blackboxsw, line 78: again let's use UAConfig.delete_cache_key we can add common logging in there at some point that let's us know when we've deleted files
```suggestion
        cfg.delete_cache_key("marker-reboot-cmds")
```

lib/reboot_cmds.py
```diff
@@ -0,0 +1,87 @@
...
+
+def main(args, cfg):
+    setup_logging(logging.INFO, logging.DEBUG)
+    reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
+
+    if not cfg.is_attached:
+        logging.debug("Skipping reboot_cmds. Machine is unattached")
+
+        if os.path.exists(reboot_cmd_marker_file):
+            del_file(reboot_cmd_marker_file)
```
blackboxsw, line 64:
```suggestion
        cfg.delete_cache_key("marker-reboot-cmds")        
```

blackboxsw (changes requested):
Looks really good @lucasmoura just a couple of requests for using common UAConfig utilities directly (delete_cache_key) instead of del_file I think. (sorry I realize I may have suggested del_file in the first place. Now that we hang that data_path off of UAConfig we should use ```delete_cache_key``` directly

blackboxsw:
Also one additional request would be unit test coverage of reboot_cmds.py if we can.

lib/reboot_cmds.py
```diff
@@ -0,0 +1,83 @@
...
+            "Failed running cmd: {}\n"
+            "Return code: {}\n"
+            "Stderr: {}\n"
+            "Stdout: {}".format(
+                cmd, exec_error.exit_code, exec_error.stderr, exec_error.stdout
+            )
+        )
+
+        reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
+        cfg.delete_cache_key(reboot_cmd_marker_file)
```
blackboxsw, line 39: we actually need to use the UAConfig data_path key here, not a file path
```suggestion
        cfg.delete_cache_key("marker-reboot-cmds")
```

lib/reboot_cmds.py
```diff
@@ -0,0 +1,83 @@
...
+
+def main(args, cfg):
+    setup_logging(logging.INFO, logging.DEBUG)
+    reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
+
+    if not cfg.is_attached:
+        logging.debug("Skipping reboot_cmds. Machine is unattached")
+
+        if os.path.exists(reboot_cmd_marker_file):
+            cfg.delete_cache_key(reboot_cmd_marker_file)
```
blackboxsw, line 64:
```suggestion
        cfg.delete_cache_key("marker-reboot-cmds")
```

uaclient/entitlements/repo.py
```diff
@@ -202,21 +216,30 @@ def process_contract_deltas(
...
                 apt.remove_auth_apt_repo(repo_filename, old_url)
-        self.remove_apt_config()
-        self.setup_apt_config()
+
+            self.remove_apt_config()
+            self.setup_apt_config()
+
         if delta_packages:
+            logging.info(
+                "Installing packages on change directives: {}".format(
```
blackboxsw, line 237:
```suggestion
                "Installing packages on changed directives: {}".format(
```

blackboxsw (changes requested):
minor fixup

lucasmoura:
@blackboxsw Thanks for the review, I have updated the code

blackboxsw (approved):
Thanks for these changes. LGTM!

PR #1263, opened 2020-12-04
Add Google Cloud Platform support.

powersj:
This adds the ability to detect if on a Google Cloud Platform system
using a combination of the cloud-init datasource and DMI data using the
product name. The client will retreive a JWT token and send it to the
contract service for verification.

A note on naming, the contract service is expect 'gcp'. cloud-id and
cloud-init datasources use 'gce'.

### DIFF
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -8,7 +8,7 @@ from uaclient import status
 from uaclient import util
 
 try:
-    from typing import Optional  # noqa: F401
+    from typing import Dict, Optional, Type  # noqa: F401
 except ImportError:
     # typing isn't available on trusty, so ignore its absence
     pass
@@ -70,7 +70,7 @@ def cloud_instance_factory() -> clouds.AutoAttachCloudInstance:
         "aws-gov": aws.UAAutoAttachAWSInstance,
         "azure": azure.UAAutoAttachAzureInstance,
         "gce": gcp.UAAutoAttachGCPInstance,
-    }
+    }  # type: Dict[str, Type[clouds.AutoAttachCloudInstance]]
 
     cloud_type = get_cloud_type()
     if not cloud_type:
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -19,7 +19,7 @@ CLOUDINIT_INSTANCE_ID_FILE = "/var/lib/cloud/data/instance-id"
 
 
 # Mapping of datasource names to cloud-id responses. Trusty compat with Xenial+
-DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws", "gce": "gce"}
+DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws", "gce": "gcp"}
 
 
 def get_instance_id(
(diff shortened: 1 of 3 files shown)


powersj:
I am not certain about the mypy error can someone help me determine what I am missing?

`uaclient/clouds/identity.py:87: error: Cannot instantiate abstract class 'AutoAttachCloudInstance' with abstract attributes 'cloud_type', 'identity_doc' and 'is_viable'`

uaclient/clouds/identity.py
```diff
@@ -19,7 +19,7 @@
 
 
 # Mapping of datasource names to cloud-id responses. Trusty compat with Xenial+
-DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws"}
+DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws", "gce": "gce"}
```
blackboxsw, line 22: This dict translates cloud-init to ua-client so it should be "gce" -> "gcp". Ultimately it's only needed if we release Gcp trusty images.  So either we drop this is we don't plan trusty pro image support, or we change it to the following:

```suggestion
DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws", "gce": "gcp"}
```

blackboxsw (commented):
> I am not certain about the mypy error can someone help me determine what I am missing?
> 
> `uaclient/clouds/identity.py:87: error: Cannot instantiate abstract class 'AutoAttachCloudInstance' with abstract attributes 'cloud_type', 'identity_doc' and 'is_viable'`

I blame mypy bad error messages, this ultimately is because our cloud_instance_map dict doesn't declare the types of the values. So, mypy grabs the common parent class, which is abstract and then says "oooh noes". I'll provide a patch in the review for my comments.


invalid token

blackboxsw (approved):
Awesome work. Thanks @powersj!   

Minor mypy fix and either drop or keep the gce -> gcp depending on whether the plan is to support trusty GCP or not.

Here's a diff of the changes if you keep the trusty support
```diff
index 262f38e..fa24c36 100644
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -8,7 +8,7 @@ from uaclient import status
 from uaclient import util
 
 try:
-    from typing import Optional  # noqa: F401
+    from typing import Dict, Optional, Type  # noqa: F401
 except ImportError:
     # typing isn't available on trusty, so ignore its absence
     pass
@@ -19,7 +19,7 @@ CLOUDINIT_INSTANCE_ID_FILE = "/var/lib/cloud/data/instance-id"
 
 
 # Mapping of datasource names to cloud-id responses. Trusty compat with Xenial+
-DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws", "gce": "gce"}
+DATASOURCE_TO_CLOUD_ID = {"azurenet": "azure", "ec2": "aws", "gce": "gcp"}
 
 
 def get_instance_id(
@@ -70,7 +70,7 @@ def cloud_instance_factory() -> clouds.AutoAttachCloudInstance:
         "aws-gov": aws.UAAutoAttachAWSInstance,
         "azure": azure.UAAutoAttachAzureInstance,
         "gce": gcp.UAAutoAttachGCPInstance,
-    }
+    }  # type: Dict[str, Type[clouds.AutoAttachCloudInstance]]
 
     cloud_type = get_cloud_type()
     if not cloud_type:
```

blackboxsw:
I was able to attach a GCP instance using this mechanism against a development contracts endpoint.
minor issues with livepatch attach for us to sort on gcp. but looks excellent!

powersj:
@blackboxsw I accepted your gce -> gcp fix and pushed the mypy patch!

powersj:
closing in favor of #1269

PR #1269, opened 2020-12-04
 Add Google Cloud Platform support

powersj:
This adds the ability to detect if on a Google Cloud Platform system
using a combination of the cloud-init datasource and DMI data using the
product name. The client will retreive a JWT token and send it to the
contract service for verification.

A note on naming, the contract service is expect 'gcp'. cloud-id and
cloud-init datasources use 'gce'.


blackboxsw (approved):
LGTM! thanks

cmars (approved):
LGTM

PR #1270, opened 2020-12-07
Fix invalid ppa integration test

lucasmoura:
We have a test to verify if uaclient is behaving as expected if we have an invalid ppa in the
system. To produce that invalid ppa, we add the uaclient daily ppa and we use a sed to edit it.
However, the path we were using on the sed command changes, breaking the test.


blackboxsw (approved):
LGTM

Issue #1085, opened 2020-06-05
CIS shown not entitled / available when actually entitled


cmars:
We've recently added CIS to Ubuntu Pro entitlements, but the client does not show CIS as entitled=yes and status is a long dash (—).

![Screenshot from 2020-06-05 14-36-48](https://user-images.githubusercontent.com/23741/83916250-74ef4100-a73a-11ea-9dcb-78505a16149b.png)

However, the actual contract has CIS entitled and available.

![Screenshot from 2020-06-05 14-36-18](https://user-images.githubusercontent.com/23741/83916707-432aaa00-a73b-11ea-9248-bc9d5ba085b5.png)
![Screenshot from 2020-06-05 14-36-36](https://user-images.githubusercontent.com/23741/83916847-8127ce00-a73b-11ea-8a57-2f8301121ee5.png)

blackboxsw:
This is because uaclient has the service name hard-coded as ``cis-audit`` which doesn't match the backend reported service name ``cis``. This could be fixed with either a backend config service rename to `cis-audit` or a ua client frontend commit changing that hardcoded [CISEntitlement.name ](https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/entitlements/cis.py#L7)to "cis"

mitechie:
Given that CIS is a known thing and isn't a fast moving target I feel the naming issue is low risk. If we do need a second offering, it would be something less standard and having the standard be CIS with a followup (like fips and fips-updates) would be reasonable.

kikoreis:
Hmmm. Isn't there at least two principal ways in which CIS was meant to be used -- one to just analyze and report, and one to actually apply changes?

HoyluBert:
Is there a work around for how to enable cis-audit? I am currently using the Ubuntu Pro FIPS 18.04 LTS on Azure and would like to test this feature.

PR #1276, opened 2020-12-10
cis: add PPA keyring for CIS and rename service from cis-audit

blackboxsw:
Add support for CIS service:
 - adjust tools/refresh-keyrings.sh to only download CIS key
 - tools/refresh-keyrings.sh drop the trusty assertion because we only care about trusty for trusty-based services (CIS is not)
 - add keyrings/ubuntu-cis.gpg (downloaded on a Xenial system using refresh-keyrings.sh)
 -  Fix CIS name "cis" instead of "cis-audit" to match contact service entitlement name
 - Add UAContract.request_resource_machine_access method which needs to be called because only services that are enableByDefault: True get their resourceTokens included in the base machine-token.json response. Any other services will have to obtain their specific resourceToken during `ua enable <service>` calls.
 -  add a call to repo.py to call UAContractClient.request_resource_machine_access if resourceTokens are absent for a repo-based service

Fixes: #1277
Fixes: #1085


lucasmoura (commented):
Looks good to me. However, we will also need to update the integration tests were `cis` is named `cis-audit` and also the help data file, which has `cis-audit`  as well.

Finally, I think we could have a behave test with us enabling `cis` and checking that everything is working as expected

lucasmoura (approved):

PR #1278, opened 2020-12-14
Fix upgrade scenario tests for fips

lucasmoura:
Some commands were failing on the fips service upgrade integration tests. The first one was related to using
a wildcard in the `egrep` command, which is not supported on the current way we shellify commands on pycloudlib.
Also, we are now covering possibilities were we need to downgrade a package on dist-upgrade or that a package
requires reboot after upgrading it

### DIFF
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -15,8 +15,7 @@ Feature: Enable command behaviour when attached to an UA staging subscription
             """
             One moment, checking your subscription first
             GPG key '/usr/share/keyrings/ubuntu-cc-keyring.gpg' not found
-            """
-
+            """ 
     @series.xenial
     @series.bionic
     @series.focal
@@ -230,7 +229,9 @@ Feature: Enable command behaviour when attached to an UA staging subscription
         """
         1
         """
-        When I run `apt-get dist-upgrade -y` with sudo
+        When I run `apt-get dist-upgrade -y --allow-downgrades` with sudo
+        # A package may need a reboot after running dist-upgrade
+        And I reboot the `<release>` machine
         And I create the file `/etc/update-manager/release-upgrades.d/ua-test.cfg` with the following
         """
         [Sources]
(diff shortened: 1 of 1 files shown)


features/staging_commands.feature
```diff
@@ -198,6 +197,7 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | xenial  | FIPS Updates | fips-updates |https://esm.staging.ubuntu.com/fips-updates/ubuntu xenial-updates/main |
            | bionic  | FIPS Updates | fips-updates |https://esm.staging.ubuntu.com/fips-updates/ubuntu bionic-updates/main |
 
+   @wip
```
blackboxsw, line 200: let's drop the @wip

features/staging_commands.feature
```diff
@@ -264,74 +266,6 @@ Feature: Enable command behaviour when attached to an UA staging subscription
         """
 
         Examples: ubuntu release
-        | release | next_release | fips-service  | fips-name    |
-        | xenial  | bionic       | fips          | FIPS         |
-
-   @series.xenial
-   @uses.config.machine_type.lxd.vm
-   Scenario Outline: Attached FIPS upgrade across LTS releases
```
blackboxsw, line 272: Was this a complete duplicate test? if so let's mention the reason for dropping it in your PR commit message

blackboxsw (changes requested):
minor minimal change to drop the @wip. and justify the dropped integration test in your suggested commit message (or the PR description)

features/staging_commands.feature
lucasmoura: It was not exactly a duplicate. This is the test that was asserting that `fips-updates` ended up with `ubuntu-fips` uninstalled after a upgrade from xenial to bionic. Since this behavior is not happening, I have integrated the test with the fips one

blackboxsw (approved):
LGTM!

PR #1280, opened 2020-12-15
Add pytest constraint for pytest on mypy tests

lucasmoura:
pytest was recently updated to a version that breaks the mypy tests. The reason for that is because the new version of pytest uses a type annotation that is only supported on python3.6 and later. Therefore we cannot run all of the mypy tests using the latest version. Because of that we are now splitting the mypy test for each release, making sure the command is respecting the release dependencies freezes, which include pytest

File on pytest with type annotation:
https://github.com/pytest-dev/pytest/blob/6.2.x/src/_pytest/assertion/__init__.py#L86

### DIFF
--- a/tools/constraints-mypy.txt
+++ b/tools/constraints-mypy.txt
@@ -1,0 +1,2 @@
+mypy
+pytest==6.1.2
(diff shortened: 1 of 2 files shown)


ubuntu-server-builder (approved):
LGTM! thanks!

PR #1283, opened 2020-12-16
Add GCE support for integration tests

lucasmoura:
Add support for launching GCE instances and running integration test scenarios on them

PS: This can only be merged after this pycloudlib [PR](https://github.com/canonical/pycloudlib/pull/79) is accepted

### DIFF
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -65,15 +65,6 @@ class Cloud:
         """Return the api used to interact with the cloud provider."""
         raise NotImplementedError
 
-    def manage_ssh_key(self, private_key_path: "Optional[str]" = None) -> None:
-        """Create and manage ssh key pairs to be used in the cloud provider.
-
-        :param private_key_path:
-            Location of the private key path to use. If None, the location
-            will be a default location.
-        """
-        raise NotImplementedError
-
     def _create_instance(
         self,
         series: str,
@@ -201,6 +192,31 @@ class Cloud:
             image_name = self.api.daily_image(release=series)
 
         return image_name
+
+    def manage_ssh_key(self, private_key_path: "Optional[str]" = None) -> None:
+        """Create and manage ssh key pairs to be used in the cloud provider.
+
+        :param private_key_path:
+            Location of the private key path to use. If None, the location
+            will be a default location.
+        """
+        cloud_name = self.name.lower().replace("_", "-")
+        pub_key_path = "{}-pubkey".format(cloud_name)
+        priv_key_path = "{}-privkey".format(cloud_name)
+        pub_key, priv_key = self.api.create_key_pair()
+
+        with open(pub_key_path, "w") as f:
+            f.write(pub_key)
+
+        with open(priv_key_path, "w") as f:
+            f.write(priv_key)
+
+        os.chmod(pub_key_path, 0o600)
+        os.chmod(priv_key_path, 0o600)
+
+        self.api.use_key(
+            public_key_path=pub_key_path, private_key_path=priv_key_path
+        )
 
 
 class EC2(Cloud):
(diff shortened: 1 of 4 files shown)


features/cloud.py
```diff
@@ -481,6 +496,97 @@ def _create_instance(
         return inst
 
 
+class GCE(Cloud):
+    name = "gce"
+
+    """Class that represents the Azure cloud provider.
```
blackboxsw, line 502:
```suggestion
    """Class that represents the Google Cloud Platform provider.
```

features/cloud.py
```diff
@@ -481,6 +496,97 @@ def _create_instance(
         return inst
 
 
+class GCE(Cloud):
```
blackboxsw, line 499: Let's replace GCE -> GCP throughout  and gce -> gcp. as that's the official abbrevation for Google Cloud Platform

features/cloud.py
```diff
@@ -481,6 +496,97 @@ def _create_instance(
...
+        tag: "Optional[str]" = None,
+        timestamp_suffix: bool = True,
+        zone: "Optional[str]" = "a",
+        gce_credentials_path: "Optional[str]" = None,
+        gce_project: "Optional[str]" = None,
+    ) -> None:
+        self.gce_credentials_path = gce_credentials_path
+        self.gce_project = gce_project
+        self.zone = zone
+        os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = str(
```
blackboxsw, line 534: Let's also set os.environ["GOOGLE_CLOUD_PROJECT"] = str(self.gce_project)

without that I get 
WARNING:google.auth._default:No project ID could be determined. Consider running `gcloud config set project` or setting the GOOGLE_CLOUD_PROJECT environment variable

features/cloud.py
```diff
@@ -481,6 +496,97 @@ def _create_instance(
...
+
+        super().__init__(
+            machine_type=machine_type,
+            region=region,
+            tag=tag,
+            timestamp_suffix=timestamp_suffix,
+        )
+
+    @property
+    def api(self) -> pycloudlib.cloud.BaseCloud:
```
blackboxsw, line 546: At some point we probably just want to parameterize this method in the base class using self.pycloudlib_cls and use a kwargs instream or replicating this boilerplate method in each subclass. I don't think you have to do that in this branch though.

features/cloud.py
```diff
@@ -481,6 +496,97 @@ def _create_instance(
...
+            An AWS cloud provider instance
+        """
+        if not image_name:
+            image_name = self.locate_image_name(series)
+
+        print("--- Launching GCE image {}({})".format(image_name, series))
+        inst = self.api.launch(image_id=image_name, user_data=user_data)
+        return inst
+
+
```
blackboxsw, line 589: Please add a 
```suggestion

    @property
    def pycloudlib_cls(self):
        """Return the pycloudlib cls to be used as an api."""
        return pycloudlib.GCE

```

blackboxsw (changes requested):
Thanks @lamoura works like a charm. minor nits. and +1

features/cloud.py
lucasmoura: We are passing that variable through the `UACLIENT_BEHAVE_GCP_PROJECT`` env var. It is weird that this warning is popping up, since the pycloudlib implementation actually requires the project named to be passed explicitly, not though an env var

features/cloud.py
lucasmoura: Once we have that api property refactoring, I do believe we will need this method, but right now, I don't think it will be adding a lot of value

lucasmoura:
@blackboxsw Thanks for the review. I have addressed the comments

ubuntu-server-builder (approved):
Thanks for the fixups and addressing the pycloudlib setup issues I ran into.

blackboxsw (approved):
Thanks for the fixups and fixing the pycloudlib config issue/warning I experienced.

PR #1284, opened 2020-12-17
Ci with jenkins

blackboxsw:
Jenkinsfile skeleton to exercise github org plugin


TheRealFalcon:
Looks good so far! Just FYI...pytest has a junit output flag: `pytest --junitxml=pytest_results.xml`. If you use that combined with the Jenkins junit support, you can get really nice output (especially using blue ocean) of individual test failures.
https://www.jenkins.io/doc/pipeline/tour/tests-and-artifacts/

Jenkinsfile
```diff
@@ -0,0 +1,81 @@
+pipeline {
+    agent any
+
+    environment {
+        VM_NAME = "gitubuntu-ci-${currentBuild.getNumber()}"
+        UACLIENT_BEHAVE_CONTRACT_TOKEN = credentials('ua-contract-token')
```
lucasmoura, line 6: I think we need the staging token as well

Jenkinsfile
```diff
@@ -0,0 +1,81 @@
...
+                    }
+                }
+            }
+        }
+        stage ('Unit Tests') {
+            steps {
+                sh '/usr/bin/tox -e py3'
+            }
+        }
+        stage ('Integration Tests') {
```
lucasmoura, line 52: Can we add all of the lxd tests here ?

Jenkinsfile
blackboxsw: +1 just added it

Jenkinsfile
blackboxsw: added 16.04, 18.04, 20.04 lxd now. We still lack the deb builder calls, but we can followup with that as another PR

paride (approved):
LGTM, thanks!

PR #1285, opened 2020-12-17
WIP: Ci with jenkins

TheRealFalcon:
Playing with jenkinsfile


blackboxsw (approved):
Thanks @TheRealFalcon This is a really good paas at getting some results in Jenkins. Let's keep this an we can iterate and improve if necessary. Much better than trying to grok the consoleFull with interleaved logs.

PR #1290, opened 2020-12-18
   jenkins: drop tox path add contract_token_staging and build_pr env vars

blackboxsw:
After cleaning up jenkins workers and removing a stale py2-based tox
    installed in .local/bin on our systems, we can now use the tox installed
    on the jenkins worker without having to specify the full path.
    
    Also add the following integration testing environment vars:
      - UACLIENT_BEHAVE_BUILD_PR=1: force local building of the PR under test.
      - UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING: allow running our staging tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -84,12 +84,12 @@ pipeline {
                         '''
                     }
                 }
-                stage("lxc vm 16.04") {
+                stage("lxc 16.04") {
                     steps {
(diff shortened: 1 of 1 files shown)


TheRealFalcon (approved):
Nice...wasn't aware of that parallel safe option

PR #1294, opened 2021-01-05
Add lxd cleanup stage on Jenkinsfile

lucasmoura:
Currently, if there is a failure in our integration tests,
we do not remove the lxd instances from the machine. We are
now adding a stage that does exactly that

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,6 +4,7 @@ pipeline {
     environment {
         TMPDIR = "/tmp/$BUILD_TAG/"
         UACLIENT_BEHAVE_JENKINS_BUILD_TAG = "${BUILD_TAG}"
+        UACLIENT_BEHAVE_JENKINS_CHANGE_ID = "${CHANGE_ID}"
         UACLIENT_BEHAVE_BUILD_PR=1
         UACLIENT_BEHAVE_CONTRACT_TOKEN = credentials('ua-contract-token')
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -151,6 +151,16 @@ pipeline {
                 }
             }
         }
+        stage ('lxd cleanup') {
+            steps {
+                sh '''
+                set +x
+                DATE=`date -d 'now+1day' +%d/%m/%Y`
```
blackboxsw, line 158: Since we are looking at tomorrow's date and saying any vm before tomorrow (that means today) we will cleanup, this could trash any other PR runs that are active outside of this PR right?

lucasmoura: Yes, that's a great catch. I will update the PR addressing this issue

features/environment.py
```diff
@@ -223,7 +223,7 @@ def __init__(
         job_suffix = os.environ.get("UACLIENT_BEHAVE_JENKINS_BUILD_TAG")
         print("--- job suffix: {}".format(job_suffix), flush=True)
         if not job_suffix:
-            job_suffix = os.environ.get("TRAVIS_JOB_NUMBER", "dev")
+            job_suffix = os.environ.get("CHANGE_ID", "dev")
```
blackboxsw, line 226: @lucasmoura I'm not sure we get to this pathm because UACLIENT_BEHAVE_JENKINS_BUILD_TAG is set in Jenkinsfile already, so we don't fall into 'if not job_suffix' scope do we?

blackboxsw (approved):
LGTM!

Issue #1251, opened 2020-12-01
ua status to report reboot required


blackboxsw:
When a reboot is required to complete a config change for a given ubuntu advantage service, ua status shall report a reboot-required message from ua status. 

Please adapt this issue description with the suggested direction for messaging about reboots required.

UA client could provide messaging about the following scenarios:
  - reboot required to complete <service_name> enable  # livepatch enable across LTS upgrade
  - reboot required to complete <service_name> disable  # FIPS disable case
  - general system reboot required                                    #  some out-of-band deb install created /var/run/reboot-required

This issue represents a usability request to define messaging from `ua status` tabular  format to handle these cases.

Options we currently have:
 1. add new reboot-required messaging somewhere in `ua status` output, the messaging will be tailored to the specific reboot case 
 2. Could use descriptionOverride options to present the affected service STATUS as "pending" and DESCRIPTION as "reboot-required to complete <operation>"

Current UA status output for reference
root@dev-b:~# ua status --all
```bash
SERVICE       ENTITLED  STATUS    DESCRIPTION
cc-eal        yes       n/a       Common Criteria EAL2 Provisioning Packages
cis-audit     no        —         Center for Internet Security Audit Tools
esm-apps      no        —         UA Apps: Extended Security Maintenance
esm-infra     yes       enabled   UA Infra: Extended Security Maintenance
fips          yes       enabled   NIST-certified FIPS modules
fips-updates  yes       disabled  Uncertified security updates to FIPS modules
livepatch     yes       n/a       Canonical Livepatch service

Enable services with: ua enable <service>

     Account: chad.smith@canonical.com
Subscription: chad.smith@canonical.com
```

blackboxsw:
Proposed examples for discussion:

* use case 1: attached system with no active UA operations in progress
```bash

SERVICE       ENTITLED  STATUS    DESCRIPTION
esm-infra     yes       enabled   UA Infra: Extended Security Maintenance (ESM)
livepatch     yes       n/a       Canonical Livepatch service

Enable services with: ua enable <service>

      Account: chad.smith@canonical.com
 Subscription: chad.smith@canonical.com
       Status: inactive
StatusDetails: No Ubuntu Advantage operations are running
```

* use case 2: system's reboot-required flag indicates system reboot needed
```bash
...
       Status: reboot-required
StatusDetails: A reboot is required due to system reboot-required flag
```
* use case 3: UA client config change operations still pending at next boot
```bash
...
[...]

PR #1296, opened 2021-01-06
Fips post update cmds and notices

blackboxsw:
Address FIPS support in two ways:
  - add debian/postinst awareness to create marker-reboot-cmd
  - add reboot_cmds.py function fix_pro_pkg_holds which will remove any FIPS apt-holds during reboot
  - add status notices to indicate on main status output that reboot is required per #1251 

Things to address, integration testing and more unit tests for reboot_cmds.py changes

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -5,7 +5,6 @@ set -e
 . /etc/os-release  # For VERSION_ID
 
 
-MYRELEASE=$(lsb_release -sc)
 APT_TRUSTED_KEY_DIR="/etc/apt/trusted.gpg.d"
 UA_KEYRING_DIR="/usr/share/keyrings/"
 
@@ -39,7 +38,7 @@ from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
 print(*ENTITLEMENT_CLASS_BY_NAME.keys(), sep=' ')
 ")
 
-    for file in `ls $DIR`; do
+    for file in "$DIR"/*; do
         release_name=""
         case "$file" in
             *-trusty*)
@@ -61,19 +60,38 @@ print(*ENTITLEMENT_CLASS_BY_NAME.keys(), sep=' ')
              for service in ${UA_SERVICES}; do
                  if [ "${file#*$service}" != "$file" ]; then
                       # Valid apt cfg file for an ubuntu-advantage service
-                      mv $DIR/$file $DIR/$new_file
+                      mv "$file" "$new_file"
                  fi
              done
          fi
     done
 }
 
-check_esm_infra_is_enabled() {
-    ua status | egrep -q 'esm-infra.*enabled' && return 0 || return 1
+
+# Check cached service status from status.json and return 0 if enabled else 1
+check_service_is_enabled() {
+    service_name=$1
+    _RET=$(python3 -c "
+import os
+import json
+from uaclient.config import UAConfig
+cfg = UAConfig()
+status = cfg.read_cache('status-cache')
+if status:
+    for service in status['services']:
+       if service['name'] == '${service_name}':
+           print(service['status'])
+")
+   if [ "${_RET}" = "enabled" ]; then
+       return  0
+   else
+       return  1
+   fi
 }
 
+
 unconfigure_esm() {
-    if ! check_esm_infra_is_enabled; then
+    if ! check_service_is_enabled esm-infra; then
         rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
         rm -f $APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY_TRUSTY
         rm -f $ESM_INFRA_APT_SOURCE_FILE_TRUSTY
@@ -115,10 +133,27 @@ EOF
     fi
 }
 
+
+# If held fips packages exist, we are on a FIPS PRO machine with FIPS enabled
+mark_reboot_for_fips_pro() {
+    FIPS_HOLDS=$(apt-mark showholds | grep -E 'fips|libssl1|openssh-client|openssh-server|linux-fips|openssl|strongswan' || exit 0)
+    if [ "$FIPS_HOLDS" ]; then
+       mark_reboot_cmds_as_needed MESSAGE_FIPS_REBOOT_REQUIRED
+    fi
+}
+
+
 mark_reboot_cmds_as_needed() {
+    msg_name=$1
     if [ ! -f "$REBOOT_CMD_MARKER_FILE" ]; then
       touch $REBOOT_CMD_MARKER_FILE
     fi
+    python3 -c "
+from uaclient.config import UAConfig
+from uaclient.status import ${msg_name}
+cfg = UAConfig()
+cfg.add_notice(label='', description=${msg_name})
+"
 }
 
 case "$1" in
@@ -151,7 +186,7 @@ case "$1" in
               # ubuntu-advantage-pro package that should be installed
               . /usr/share/debconf/confmodule
               db_input high ubuntu-advantage-tools/suggest_pro_pkg || true
-              db_go
+              db_go || true
           fi
       fi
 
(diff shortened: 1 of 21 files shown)


uaclient/status.py
```diff
@@ -181,6 +181,12 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_NO_ACTIVE_OPERATIONS = """No Ubuntu Advantage operations are running"""
 MESSAGE_LOCK_HELD = """Operation in progress: {lock_holder} (pid:{pid})"""
 PROMPT_YES_NO = """Are you sure? (y/N) """
+MESSAGE_LIVEPATCH_LTS_REBOOT_REQUIRED = (
+    "Service livepatch requires reboot across LTS upgrade"
```
mitechie, line 185: Livepatch requires a system reboot across the LTS upgrade

uaclient/status.py
```diff
@@ -181,6 +181,12 @@ class UserFacingStatus(enum.Enum):
...
+)
+MESSAGE_FIPS_REBOOT_REQUIRED = (
+    "Service FIPS requires reboot for APT configuration"
```
mitechie, line 188: FIPS support requires a system reboot for configuration.

blackboxsw:
Sample output 
```bash
cc-eal        yes       disabled  Common Criteria EAL2 Provisioning Packages
cis           no        —         Center for Internet Security Audit Tools
esm-apps      no        —         UA Apps: Extended Security Maintenance (ESM)
fips          yes       n/a       NIST-certified FIPS modules
fips-updates  yes       enabled   Uncertified security updates to FIPS modules

NOTICES
Reboot required: Service FIPS requires a reboot to apply APT changes.

```

debian/postinst
```diff
@@ -115,10 +116,28 @@ EOF
...
 mark_reboot_cmds_as_needed() {
+    msg_name=$1
     if [ ! -f "$REBOOT_CMD_MARKER_FILE" ]; then
       touch $REBOOT_CMD_MARKER_FILE
     fi
+    python3 -c "
+from uaclient.config import UAConfig
+from uaclient.status import ${msg_name}
+cfg = UAConfig()
+cfg.add_notice('Reboot required', ${msg_name})
```
mitechie, line 139: Let's lowercase this?

mitechie: nvm

debian/postinst
```diff
@@ -115,10 +116,28 @@ EOF
     fi
 }
 
+
+mark_reboot_for_fips_pro() {
```
lucasmoura, line 120: Maybe I am missing something here but this function doesn't seem to called in the postinst script

debian/postinst
```diff
@@ -115,10 +116,28 @@ EOF
...
+    if ! check_service_is_enabled fips; then
```
lucasmoura, line 121: I don't think we can perform this call here. This is because `check_service_is_enabled` will call `ua status` which will then refresh the status cache. For example, during a `do-release-upgrade` , we would update the cache, which will show `esm-infra`  as disabled because of the apt url change and then the `upgrade_lts_contract`  script will not work as expected during reboot.

However, this will not be true if `do-release-upgrader` now correclty updates the apt url on the source list file before upgrading the package.

lib/reboot_cmds.py
```diff
@@ -41,6 +41,36 @@ def run_command(cmd, cfg):
...
+            if service_status == "enabled":
+                ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[
+                    entitlement_name
+                ]
+                logging.debug(
+                    "Attempting to remove Ubuntu Pro FIPS package holds"
+                )
+                entitlement = ent_cls(cfg)
+                try:
+                    entitlement.setup_apt_config()  # Removes package holds
```
lucasmoura, line 60: I think we still need to manually install the `additionalPackages`  directive here, right ?

Maybe the `process_remaining_deltas`  function will cover that, but my concern is that if we update the contract on prod before the user upgrades the package, we may end up not installing the metapackage.

For example, consider this scenario:

1) fips contract is updated with new `additionalPackage (ubuntu-fips)`
2) User runs a `ua refresh`
3) User upgrade the `uaclient`

In that scenario we will not install the new `ubuntu-fips` metapackage, because there will be no deltas in the contract regarding it.

lib/reboot_cmds.py
```diff
@@ -65,8 +98,10 @@ def main(args, cfg):
         return
 
     if os.path.exists(reboot_cmd_marker_file):
+        import pdb; pdb.set_trace()
```
lucasmoura, line 101: We can remove that

uaclient/config.py
```diff
@@ -94,6 +104,27 @@ def log_level(self):
         except AttributeError:
             return getattr(logging, CONFIG_DEFAULTS["log_level"])
 
+    def add_notice(self, label: str, description: str):
```
lucasmoura, line 107: I think we need a docstring for this function

uaclient/config.py
```diff
@@ -94,6 +104,27 @@ def log_level(self):
...
+        notices = self.read_cache("notices") or []
```
lucasmoura, line 108: Maybe instead of a list of tuples, we could store the `notices` as a dict ? I don't think it will make the code faster, but the `remove_notice` method will be cleaner. But again, this is not a huge necessity at all.

debian/postinst
blackboxsw: just added it to the defaul call patch in configure)

debian/postinst
blackboxsw: Changed `check_service_is enabled` to  checks the cached status.json value instead of calling out and updating the status.json cache and potentially setting services to disabled as a result. Great catch

lib/reboot_cmds.py
blackboxsw: +1 we can call entitlement.install_packages() here which will DTRT with regard to the cached additionalPackages directive on disk. 

for use cases:
 1. old client + new contract (via ua refresh): additionalPackage delta containing `ubuntu-fips` was ignored (so we need to install)
 2. new_client + new_contract from ua refresh: additionalPackages will attempt to install `ubuntu-fips` on additionalPackage delta so NOOP here if we try to re-install package that is already instaled
 3. plain upgrade to new_client, postinst will create the REBOOT_CMD_REQUIRED   marker and force this logic to run and correct new contract additionalPackages after reboot.


this means we have to release the contract update on PRO images before we release the new PRO images with version 26.0 support

lib/reboot_cmds.py
blackboxsw: think I got this. forgot to push

uaclient/config.py
blackboxsw: I think we need to pass on this until next round. Rick has a couple of ideas about where he wants these notices to go  and I think we may need to massage this data format a bit.

debian/postinst
```diff
@@ -68,12 +68,32 @@ print(*ENTITLEMENT_CLASS_BY_NAME.keys(), sep=' ')
...
-    ua status | egrep -q 'esm-infra.*enabled' && return 0 || return 1
+
+# Check cached service status from status.json and return 0 if enabled else 1
+check_service_is_enabled() {
+    service_name=$1
+    _RET=$(python3 -c "
+import os
+import json
+
+status_file = '/var/lib/ubuntu-advantage/status.json'
```
lucasmoura, line 79: I think it would be better if we use `UAConfig` to access the status cache. By doing that, I think we can avoid checking if the file exists, for example

debian/postinst
```diff
@@ -115,10 +135,27 @@ EOF
     fi
 }
 
+
+# If held fips packages exist, we are on a FIPS PRO machine with FIPS enabled
+mark_reboot_for_fips_pro() {
+    FIPS_HOLDS=$(apt-mark showholds | egrep -q 'fips|libssl1|openssh-client|openssh-server|linux-fips|openssl|strongswan')
```
lucasmoura, line 141: Just to be sure here, if we don't have any of those holds, would that command return `1` ? I am just afraid if that call could fail postinst, since we are using `set -e`

lucasmoura (commented):
@blackboxsw just a couple of questions left.

Also, do you think we could add a behave test for the FIPS PRO machine ?
Or maybe we could have a manual test for it,  just to be sure that the reboot script is working as we expect

debian/postinst
blackboxsw: done

debian/postinst
blackboxsw: yeah I think that set -e is causing failures, I would want to be loud and problematic during pkg install time if we do have unexpected failures, otherwise we might not get a bug about it and we'd have very problematic behavior in a non working postinst. The "good news" is that my branch broke postinst multiple times and did generate apt cmdline errors as a result which broke our integration tests. I think we should actually have an explicity behave step that validates during our base image build that ua-tools did install properly without error so we fail loudly during testing

lucasmoura (approved):

Issue #1297, opened 2021-01-06
fips: integration testing errors. can no longer prevent auto-attach on PRO images with cloud-config


blackboxsw:
Because auto-attach now runs before cloud-config userdata is processed, we can't prevent auto-attach from running on PRO images during launch. As a result, we need either explicit test setup for PRO images to avoid installing ubuntu-advantage-pro deb, or our PRO tests should still expect that an auto-attach has occurred and cleanup, re-attach those test instances.

failures we see are tracebacks in the test logs due to auto-attach running:
error message:
   ```bash
# features/steps/steps.py:227
    And I verify that running `grep Traceback /var/log/ubuntu-advantage.log` `with sudo` exits `1`                                                # features/steps/steps.py:227
      Traceback (most recent call last):
        File "/jenkins/servers/server/workspace/uaclient-vm/UBUNTUVMRELEASE/18.04/ua-client/.tox/behave-vm-18.04/lib/python3.6/site-packages/behave/model.py", line 1329, in run
          match.run(runner.context)
        File "/jenkins/servers/server/workspace/uaclient-vm/UBUNTUVMRELEASE/18.04/ua-client/.tox/behave-vm-18.04/lib/python3.6/site-packages/behave/matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "features/steps/steps.py", line 234, in then_i_verify_that_running_cmd_with_spec_exits_with_codes
          assert str(context.process.returncode) in expected_codes
      AssertionError
```

jenkins job log: https://jenkins.ubuntu.com/server/view/ua/job/uaclient-vm/28/UBUNTUVMRELEASE=18.04/console

**Comment by lucasmoura** [comment]
I am currently working on ensuring that pro deb are not installed on non-pro machines. However, when looking at the pro machine logs to confirm if auto-attach is being run, I get this:

```
2021-01-07 15:45:09,224 - cli.py:(986) [DEBUG]: Executed with sys.argv: ['/usr/bin/ua', 'auto-attach']
2021-01-07 15:45:09,224 - util.py:(581) [DEBUG]: Writing file: /var/lib/ubuntu-advantage/private/lock
2021-01-07 15:45:09,224 - cli.py:(691) [DEBUG]: Skipping auto-attach. Config disable_auto_attach is set.
2021-01-07 15:45:09,224 - util.py:(591) [DEBUG]: Removing file: /var/lib/ubuntu-advantage/private/lock
[...]

PR #1298, opened 2021-01-07
Do not install pro debs on non-pro instances

lucasmoura:
Currently, when performing integration tests, we install both the default package and the pro package in the same
instance. However, now that we are running the auto-attach init script before cloud-init process the userdata, we are no longer ensuring that this script will not run. Since the auto-attach command will fail on non-pro instances, we must avoid installing the pro
package on those instances.

Related to bug #1297

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -46,7 +46,7 @@ USERDATA_APT_SOURCE_PPA_TRUSTY = """\
 apt_sources:  # for trusty
   - source: deb {ppa_url} $RELEASE main
     keyid: {ppa_keyid}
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
LGTM for the integration test. let's rebase from master to clear out the test failure on azure.

lucasmoura:
@blackboxsw We had another issue here regarding the mypy tests. I had to pin another lib because of typing annotations not supported by python 3.4

blackboxsw (approved):
LGTM!

PR #1299, opened 2021-01-07
Fix esm package messages test

lucasmoura:
It seems that after we released the SRU for update notifier, some of the expected messages have changed. This updates
the message we expect to see in the test


blackboxsw (approved):
Thank you for this change. let's land it and we can reenable azuregeneric testing after we sort why we can no longer connect to azure vms

PR #1303, opened 2021-01-13
Allow destroy instances on Jenkins PR build

lucasmoura:
Currently, we are not destroying the instances after running the behave tests. However, since we are now
avoiding failures if we cannot delete an instance after the test, we can bring that behavior back


blackboxsw (approved):
LGTM

PR #1304, opened 2021-01-13
Set pip version for testing older releases

lucasmoura:
When we run flake8 and pytest jobs using dependencies based on older ubuntu releases, we must ensure that
we are using an older version of pip. If we do not do that, new versions of pip will not be able to parse the specified
dependencies tree and it will fail to run the test

To pin the pip version, I had to add a dependency to a tox plugin called `tox-pip-version`. Initially I thought about using the `VIRTUALENV_PIP` variable to set this, but it is not working as I expected, since pip is always being upgrade to the latest  version anyway

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ ifneq (,$(findstring trusty,$(TOXENV)))
 	pip install virtualenv==20.0.31
 endif
 	pip install tox
+	pip install tox-pip-version
 
 travis-deb-install:
 	git fetch --unshallow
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
This tox test is like a leading indicator for us on flake compat for a given series. As such I'm okay with us pinning the pip version of tox we are runnning to help unblock local development runs of tox on a dev system. Since our PRs will have package builds anyway on every series, we also will cover series-specific comple runs flake8 runs anyway per our dh_auto_test  calls in debian/rules. So, I don't think we lose much by pinning this version here.

PR #1307, opened 2021-01-13
Remove cis from beta services

lucasmoura:
Since we are now supporting cis, we need to remove it from beta services pool and allow users to see the service normally when running `ua status`


blackboxsw:
Thanks for putting this up, after talking with Rick, I don't think we will be going "production" of CIS until addtiional discussions are had about what level of commandline/profile support would be needed in CIS in UA-client. I'll close for now and we can re-open once we make sure to sort our usability story

PR #1308, opened 2021-01-13
jenkins: add deb build stage to assert package builds

blackboxsw:


lucasmoura (approved):

Jenkinsfile
```diff
@@ -33,6 +33,14 @@ pipeline {
...
                 . $TMPDIR/bin/activate
                 pip install tox  # for tox supporting --parallel--safe-build
+
+                # Updates to allow building packages locally
+                sudo sed -Ei 's/^# deb-src /deb-src /' /etc/apt/sources.list
+                sudo apt-get update
+                sudo apt-get build-dep -y ubuntu-advantage-tools
+                sudo apt-get install -y --install-recommends sbuild ubuntu-dev-tools dh-systemd
+                # Missing build-deps
+                sudo apt-get install -y --install-recommends libapt-pkg-dev python3-mock python3-pytest
```
paride, line 43: I think we can skip this install steps? It's safe to assume that sbuild is installed on the Jenkins nodes and chroot setup for all the LTS releases, so we don't need to reinstall it. Also: if the package is built with sbuild we don't need to install any build-dep in the host system.

In general I think our goal should be to avoid `sudo` commands if possible, to ensure the CI run is safe and can't interfere too much with other things happening on the system.

Jenkinsfile
```diff
@@ -76,6 +84,74 @@ pipeline {
...
         }
+        stage ('Package build: 14.04') {
+            environment {
+                BUILD_SERIES = "trusty"
+                ARTIFACT_DIR = "${TMPDIR}${BUILD_SERIES}"
+            }
+            steps {
+                sh '''
+                set -x
+                debuild -S -uc -us
```
paride, line 95: IIUC here you want to build the source package which is then built with sbuild, and this is why you installed the build-deps in an earlier stage. This is not necessary: the source package can be built by calling `dpkg-source -b .` directly. This is actually what the dpkg-buildpackage(1) manpage suggest ("[...] using dpkg-source directly is always a better [...]"), no build-deps are needed.

But `sbuild` can also be called directly from the deb package source tree, it will call `dpkg-source -b .` by itself. So I think we can skip this step.

paride:
On the slowness of the build process: I think we can greatly speed up the build by installing `apt-cacher-ng` in the host system and `auto-apt-proxy` in the schroots. Running `sbuild --no-run-lintian` may also help a bit. Failure in lintian do not make the sbuild build fail so are not immediately visible, so if we want a lintian check it's better to add a separate lintian stage in the Jenkinsfile in any case.

Jenkinsfile
blackboxsw: Excellent thanks @paride. I'll clean this up.

blackboxsw:
> On the slowness of the build process: I think we can greatly speed up the build by installing `apt-cacher-ng` in the host system and `auto-apt-proxy` in the schroots. Running `sbuild --no-run-lintian` may also help a bit. Failure in lintian do not make the sbuild build fail so are not immediately visible, so if we want a lintian check it's better to add a separate lintian stage in the Jenkinsfile in any case.

Will look to dro p--no-run-lintian per PR as and I think we can run it as a separate daily job.  Package lintian primarily will matter for us to resolve before releases, but not necessarily per PR

PR #1310, opened 2021-01-14
Remove grep for release in fips upgrade test

lucasmoura:
Currently, after we upgrade a release with fips enabled we run a grep command on the source list files to see
if we can find a disable or the name of the past release on them. However, the deb-src files are appearing commented
out with the past release. Since this is not an issue, we can remove grepping for the release on that part of the test


blackboxsw (approved):
LGTM!

PR #1314, opened 2021-01-15
Add https_proxy to passenv in tox

lucasmoura:
To run GCP behave tests in Jenkins, we need to setup the https_proxy env variable. However, tox only pass env variables
to the test environments if they are specified on passenv. We are now adding that variable there to unblock the GCP tests


paride (approved):
+1, we tested this actually works.

PR #1315, opened 2021-01-19
Allow uaclient to disable services during enable

### LINKED ISSUE #1301: ua enable fips should offer to disable livepatch vs just erroring
Current workflow:

    $ sudo ua enable --beta fips
    Installation of additional packages are required to make this system FIPS
    compliant.
    Are you sure? (y/N) y
    Cannot enable FIPS when Livepatch is enabled

Proposed workflow:

    $ sudo ua enable --beta fips
    Installation of additional packages are required to make this system FIPS
    compliant.
    Are you sure? (y/N) y
    FIPS cannot be enabled with Livepatch. Disable Livepatch and proceed to enable FIPS? (y/N)

lucasmoura:
When we are enabling a service in uaclient, we may end up finding that we have an incompatible service
already enabled in the system. Currently, we alert the user about this issue and abort the enable operation.
We are updating that to allow the user to decide if the incompatible services should be disabled or not
before enabling the requested service.

To make that happen, we are creating a new abstraction called `incompatible_services`. Every entitlement that has an incompatible service needs to define this parameters, which is a list of strings containing the name of the incompatible services.

We are also adding the feature override named `block_disable_on_enable` to never allow a service to be disabled during an enable operation. Those flags were created due to testing purposes, since there is no easy way to simulate a user input on behave.

fix #1301

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -370,7 +370,12 @@ Feature: Enable command behaviour when attached to an UA subscription
             FIPS enabled
             A reboot is required to complete install
             """
-        And I verify that running `ua enable livepatch` `with sudo` exits `1`
+        When I append the following on uaclient config:
+            """
+            features:
+              block_disable_on_enable: true
+            """
+        Then I verify that running `ua enable livepatch` `with sudo` exits `1`
         And I will see the following on stdout
             """
             One moment, checking your subscription first
@@ -389,13 +394,59 @@ Feature: Enable command behaviour when attached to an UA subscription
             Installing canonical-livepatch snap
             Canonical livepatch enabled
             """
-        And I verify that running `ua enable fips --assume-yes` `with sudo` exits `1`
+        When I append the following on uaclient config:
+        """
+        features:
+          block_disable_on_enable: true
+        """
+        Then I verify that running `ua enable fips --assume-yes` `with sudo` exits `1`
         And I will see the following on stdout
             """
             One moment, checking your subscription first
             Cannot enable FIPS when Livepatch is enabled
             """
 
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable fips on a machine with livepatch active
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        Then stdout matches regexp:
+            """
+            Updating package lists
+            ESM Infra enabled
+            """
+        And stdout matches regexp:
+            """
+            Installing canonical-livepatch snap
+            Canonical livepatch enabled
+            """
+        When I run `ua enable fips --assume-yes` with sudo
+        Then I will see the following on stdout
+            """
+            One moment, checking your subscription first
+            Updating package lists
+            Installing FIPS packages
+            FIPS enabled
+            A reboot is required to complete install
+            """
+        When I run `ua status` with sudo
+        Then stdout matches regexp:
+            """
+            fips +yes +enabled
+            """
+        And stdout matches regexp:
+            """
+            livepatch +yes +n/a
+            """
+
+        Examples: ubuntu release
+           | release |
+           | bionic  |
+           | xenial  |
+
+    @series.xenial
     @series.bionic
     @uses.config.machine_type.lxd.vm
     Scenario Outline: Attached enable fips on a machine with fips-updates active
@@ -428,3 +479,4 @@ Feature: Enable command behaviour when attached to an UA subscription
         Examples: ubuntu release
            | release |
            | bionic  |
+           | xenial  |
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -21,6 +21,7 @@ from uaclient.status import (
     ApplicabilityStatus,
     ContractStatus,
     UserFacingStatus,
+    MESSAGE_INCOMPATIBLE_SERVICE_STOPS_ENABLE,
 )
 from uaclient.defaults import DEFAULT_HELP_FILE
 
@@ -38,11 +39,14 @@ class UAEntitlement(metaclass=abc.ABCMeta):
     # Whether to assume yes to any messaging prompts
     assume_yes = False
 
-    # Wheter that entitlement is in beta stage
+    # Whether that entitlement is in beta stage
     is_beta = False
 
     # Help info message for the entitlement
     _help_info = None  # type: str
+
+    # List of services that are incompatible with this service
+    _incompatible_services = []  # type: "List[str]"
 
     @property
     @abc.abstractmethod
@@ -82,6 +86,16 @@ class UAEntitlement(metaclass=abc.ABCMeta):
     @property
     def static_affordances(self) -> "Tuple[StaticAffordance, ...]":
         return ()
+
+    @property
+    def incompatible_services(self) -> "List[str]":
+        """
+        Return a list of packages that aren't compatible with the entitlement.
+        When we are enabling the entitlement we can directly ask the user
+        if those entitlements can be disabled before proceding.
+        Overridden in livepatch and fips
+        """
+        return self._incompatible_services
 
     def __init__(
         self, cfg: "Optional[config.UAConfig]" = None, assume_yes: bool = False
@@ -139,6 +153,7 @@ class UAEntitlement(metaclass=abc.ABCMeta):
             if not silent:
                 print(status.MESSAGE_UNENTITLED_TMPL.format(title=self.title))
             return False
+
         application_status, _ = self.application_status()
         if application_status != status.ApplicationStatus.DISABLED:
             if not silent:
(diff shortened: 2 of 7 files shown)


features/attached_enable.feature
```diff
@@ -385,13 +390,60 @@ Feature: Enable command behaviour when attached to an UA subscription
...
+            """
+            Updating package lists
+            ESM Infra enabled
+            """
+        And stdout matches regexp:
+            """
+            Installing canonical-livepatch snap
+            Canonical livepatch enabled
+            """
+        When I run `ua enable fips --assume-yes --allow-disable` with sudo
```
mitechie, line 421: what is the difference in the aasume-yes vs the allow-disable?

uaclient/cli.py
```diff
@@ -295,6 +295,14 @@ def enable_parser(parser):
         action="store_true",
         help="do not prompt for confirmation before performing the enable",
     )
+    parser.add_argument(
+        "--allow-disable",
+        action="store_true",
+        help=(
+            "do not prompt for disable confirmation if an incompatible "
```
mitechie, line 302: this seems like a double affirmative with assume-yes?

uaclient/tests/test_cli_enable.py
```diff
@@ -458,13 +461,13 @@ def test_beta_entitlement_instantiated_and_enabled_with_config_override(
...
 
-        kwargs = {"allow_beta": False}
+        kwargs = {"allow_beta": False, "allow_disable": False}
         if silent_if_inapplicable is not None:
             kwargs["silent_if_inapplicable"] = silent_if_inapplicable
         ret = _perform_enable(ent_name, m_cfg, **kwargs)
 
         assert [
-            mock.call(m_cfg, assume_yes=False)
+            mock.call(m_cfg, assume_yes=False, allow_disable=False)
```
mitechie, line 470: I'm -1 on "allow_disable" as it's very unclear what the disable/allow are speaking to here.

features/attached_enable.feature
lucasmoura: Maybe I am overthinking that, but my rationale here is that, by default, we should ask the users for confirmation on the disable operation. Even though the `--assume-yes`  already dictates that it will answer yes to all prompts that appear, it may not be clear to the user that during an enable operation we could also disable existing services.

mitechie: The reason for folks to use --assume-yes is for scripted where no one is there and they've added that to make sure it happens (e.g. enables FIPS) so I think just --assume-yes would be sufficient for users. 

I get the idea of "allow disabling of services" now as what you meant with allow-disable. I just don't think we need that separate permission.

uaclient/entitlements/base.py
```diff
@@ -83,6 +87,14 @@ def help_info(self) -> str:
     def static_affordances(self) -> "Tuple[StaticAffordance, ...]":
         return ()
 
+    # A list of packages that are not compatible with the entitlement.
+    # When we are enabling the entitlement we can directly ask the user
+    # if those entitlements can be disabled before proceding.
+    # Overridden in livepatch and fips
```
blackboxsw, line 93: Let's make this comment into a proper docstring on the method instead of a leading comment.

uaclient/entitlements/base.py
```diff
@@ -83,6 +87,14 @@ def help_info(self) -> str:
...
+    @property
+    def incompatible_services(self):
```
blackboxsw, line 95: let's your typing hints everywhere in new code

```suggestion
    def incompatible_services(self) -> 'List[str]':
```

uaclient/entitlements/base.py
```diff
@@ -153,6 +166,78 @@ def can_enable(self, silent: bool = False) -> bool:
             if not silent:
                 print(details)
             return False
+
+        if self.incompatible_services:
+            return self.handle_incompatible_services()
+
+        return True
+
+    def handle_incompatible_services(self):
```
blackboxsw, line 175: Let's do typing hints, and please add a :return: doc  at the bottom of the docstring
```suggestion
    def handle_incompatible_services(self) -> bool:
```

uaclient/entitlements/base.py
```diff
@@ -153,6 +166,78 @@ def can_enable(self, silent: bool = False) -> bool:
...
+        """
+        Handle to proceed when a incompatible service is found during enable
```
blackboxsw, line 177:
```suggestion
        Prompt user when incompatible services are found during enable.
```

uaclient/entitlements/base.py
```diff
@@ -153,6 +166,78 @@ def can_enable(self, silent: bool = False) -> bool:
...
+        if the incompatible service should be disabled before proceeding.
+        There are also different ways to configure that behavior:
+
+        We can disable removing incompatible service during enable by
+        adding the following lines into uaclient.conf:
+
+        features:
+          block_disable_on_enable: true
+
+        We can also use the --allow-disable flag during enable to
```
blackboxsw, line 190: I think this docstring doesn't reflect the code right now, but I feel like we **should** provide a separate cmdline flag for this condition because auto-disabling livepatch without prompt represents dropping security patch updates to the system and that is a significant decisision that we don't really want to apply to the system in the FIPS case because there is no easy way back from that on a system given that we've modified the openssl stack.


This use-case feels quite a bit like apt command-line's `--assume-yes` versus `--force-yes` options since livepatch -> FIPS transition is potentially a harmful/irreparable config change on the system.

```bash
  --force-yes
       Force yes; this is a dangerous option that will cause apt to
       continue without prompting if it is doing something potentially
       harmful. It should not be used except in very special situations.
   -y, --yes, --assume-yes
       Automatic yes to prompts; assume "yes" as answer to all prompts and
       run non-interactively. If an undesirable situation, such as
       changing a held package, trying to install an unauthenticated
       package or removing an essential package occurs then apt-get will
       abort. Configuration Item: APT::Get::Assume-Yes.
```

uaclient/entitlements/tests/test_fips.py
```diff
@@ -105,10 +105,12 @@ def test_messaging_passes_assume_yes(
 
 
 class TestFIPSEntitlementCanEnable:
+    @mock.patch("uaclient.util.is_config_value_true")
```
blackboxsw, line 108: Can add the return_value=False here instead of adding it into the test method below to save a line
```suggestion
    @mock.patch("uaclient.util.is_config_value_true", return_value=False)
```

uaclient/entitlements/tests/test_fips.py
```diff
@@ -329,26 +331,45 @@ def fake_subp(cmd, *args, **kwargs):
 
         assert 1 == m_remove_apt_config.call_count
 
+    @mock.patch("uaclient.util.is_config_value_true")
+    @mock.patch("uaclient.util.prompt_for_confirmation")
```
blackboxsw, line 335: If there is horizontal space without having to break lines, we can add `, return_value=X` here for these instead of adding more lines to the test below

uaclient/entitlements/base.py
```diff
@@ -153,6 +166,78 @@ def can_enable(self, silent: bool = False) -> bool:
...
+        automatically disable any incompatible service during
+        enable.
+        """
+        from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
+
+        cfg_block_disable_on_enable = util.is_config_value_true(
+            config=self.cfg.cfg,
+            path_to_value="features.block_disable_on_enable",
+        )
```
blackboxsw, line 199: I don't think we need to surface yet another config option/feature override for this do we? Shouldn't our default behavior be to block disable of livepatch on when enabling FIPS and allow a user to either interact with the prompt with "Y"  or provide a `--force-yes` option.

blackboxsw (changes requested):
Hi @lucasmoura & @mitechie  Dropping an active livepatch service without a prompt feels a bit like potentially exposing a system to vunerabilities that we really should prefer user-interaction for. Once we move a system into a FIPS kernel, it is not easy to go back and enable livepatch anymore. As such, I'd term this config operation as a bit bigger priority and potentially destructive.

It feels a bit like the `apt` commandline has prior art and behavior that kindof behaves a bit like this. `--assume-yes` for simple prompts and `--force-yes` for potentially destructive/dangerous prompts because "I know what I'm doing". I'd like to treat disabling individual support services  as a special `--force-yes` case. We should chat about that to see if we all agree here.

uaclient/entitlements/base.py
lucasmoura: The problem is related to behave tests. I couldn't find a direct way to send a user input through behave. Therefore, this flag is way to test the scenario where a user doesn't want to disable livepatch during the install

blackboxsw:
> Hi @lucasmoura & @mitechie Dropping an active livepatch service without a prompt feels a bit like potentially exposing a system to vunerabilities that we really should prefer user-interaction for. Once we move a system into a FIPS kernel, it is not easy to go back and enable livepatch anymore. As such, I'd term this config operation as a bit bigger priority and potentially destructive.
> 
> It feels a bit like the `apt` commandline has prior art and behavior that kindof behaves a bit like this. `--assume-yes` for simple prompts and `--force-yes` for potentially destructive/dangerous prompts because "I know what I'm doing". I'd like to treat disabling individual support services as a special `--force-yes` case. We should chat about that to see if we all agree here.

We talked with @mitechie at standup and decided that customers enabling FIPS on a PRO or non-pro image are customers that know specifically what they are doing (and that they don't want security updates from livepatch anyway) so an `--assume-yes` and a `--force-yes` is really unnecessary differentiation. The only folks that are using --assume-yes are folks that really know they want stability of FIPS anyway, so one flag is enough.

uaclient/entitlements/base.py
blackboxsw: We tabled this in discussion as FIPS-users providing the --assume-yes are ones that really know what they are doing anyway and wouldn't want the expectation of livepatch on the system introducing deltas anyway.

uaclient/entitlements/base.py
blackboxsw: +1 I couldn't find a good way to do this through BDD on initial glance. Let's land this iteration and take a tech-debt item #1327 to better provide input "answers" via BDD that doesn't also surface more "public-facing" config parameters in the features: section.

features/attached_enable.feature
```diff
@@ -389,13 +394,60 @@ Feature: Enable command behaviour when attached to an UA subscription
...
+            """
+            Updating package lists
+            ESM Infra enabled
+            """
+        And stdout matches regexp:
+            """
+            Installing canonical-livepatch snap
+            Canonical livepatch enabled
+            """
+        When I run `ua enable fips --assume-yes --force-yes` with sudo
```
blackboxsw, line 425: I think this --force-yes param isn't allowed anymore right?

blackboxsw (approved):
Please update the PR description at the top to reflect this changeset and a minor fix in integration tests the Approved.

blackboxsw (changes requested):
fix int tests and we are +1

blackboxsw (approved):
LGTM! Thanks @lucasmoura

PR #1316, opened 2021-01-20
UA Fix spike: New UASecurityClient for querying CVE and USNs basic CLI

blackboxsw:
Add a full UASecurityClient to allow querying applicability of CVEs against a running machine.
Add basic `ua fix CVE` subcommand which will introspect the running system against a known CVE and 
report whether or not the CVE affects the given system.

USN handling is non-functional in this PR.

This is a simple spike intended to build very basic functionality and allow us to iterate on CLI flow.
Expect additional PRs to improve functionality for prompt the user about required CVE update actions.

Test steps:
- ua fix --help
```bash
PYTHONPATH=. python3 -m uaclient.cli fix --help
```
- ua fix invalid-value
```bash
PYTHONPATH=. python3 -m uaclient.cli fix USN-bogus
```
- ua fix valid already installed CVE on Focal
```bash
PYTHONPATH=. python3 -m uaclient.cli fix CVE-2020-28196
```
- ua fix dummy "does not affect" response for USNs
```bash
PYTHONPATH=. python3 -m uaclient.cli fix USN-4539-1
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,6 +22,7 @@ pipeline {
         UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING = credentials(
             'ua-contract-token-staging'
         )
+        JOB_SUFFIX = sh(returnStdout: true, script: "basename ${JOB_NAME}| cut -d'-' -f2").trim()
     }
 
     stages {
@@ -33,7 +34,6 @@ pipeline {
                 python3 -m venv $TMPDIR
                 . $TMPDIR/bin/activate
                 pip install tox  # for tox supporting --parallel--safe-build
-                dpkg-source -b .
                 '''
             }
         }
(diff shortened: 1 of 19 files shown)


uaclient/security.py
```diff
@@ -0,0 +1,397 @@
...
+                status.MESSAGE_SECURITY_FIX_RELEASE_STREAM.format(
+                    fix_stream=fix_stream
+                )
+            )
+            try:  # get_pkg_install_status_messages
+                util.subp(
+                    [
+                        "dpkg",
+                        "--compare-versions",
+                        fixed_version,
```
lucasmoura, line 379: Maybe I am missing something here, but is seems that the `fixed_version` package will only be set if we do not have a pocket on the `pkg_status.status` object

lucasmoura (commented):
Can we also have tests for the prompt messages being generated ? I think a unittest will work just fine, but we could also have integration tests for this

uaclient/security.py
```diff
@@ -0,0 +1,397 @@
...
+
+
+class UASecurityClient(serviceclient.UAServiceClient):
+
+    cfg_url_base_attr = "security_url"
+    api_error_cls = SecurityAPIError
+
+    def get_cves(
+        self,
+        q: "Optional[str]" = None,
```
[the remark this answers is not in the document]
blackboxsw: sure let's name it that way. and rewrite it in params to the proper param name for the route

blackboxsw (commented):
thanks for the review. addressed comments.

uaclient/security.py
```diff
@@ -0,0 +1,397 @@
...
+            "limit": limit,
+            "offset": offset,
+            "order": order,
+        }
+        usns_response, _headers = self.request_url(
+            API_V1_NOTICES, query_params=query_params
+        )
+        return [USN(client=self, response=usn_md) for usn_md in usns_response]
+
+    def get_notice(self, notice_id: str):
```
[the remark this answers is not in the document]
lucasmoura: I think my explanation was not clear, sorry for that. I  understand that we have different query params for both notices and cves, but the idea would be to mantain the signature of those methods as is and do something like this:

```py
    def get_cves(
        self,
        query: "Optional[str]" = None,
        priority: "Optional[str]" = None,
        package: "Optional[str]" = None,
        limit: "Optional[int]" = None,
        offset: "Optional[int]" = None,
        component: "Optional[str]" = None,
        version: "Optional[str]" = None,
        status: "Optional[List[str]]" = None,
    ) -> "List[CVE]":
        """Query to match multiple-CVEs.
        @return: List of CVE instances based on the the JSON response.
        """
        query_params = {
            "q": query,
            "priority": priority,
            "package": package,
            "limit": limit,
            "offset": offset,
            "component": component,
            "version": version,
            "status": status,
        }
        security_cls = CVE
        request_fmt = API_V1_CVES
        return self.query_multiple_items(query_params, request_fmt, security_cls)

    def get_notices(
        self,
        details: "Optional[str]" = None,
        release: "Optional[str]" = None,
        limit: "Optional[int]" = None,
        offset: "Optional[int]" = None,
        order: "Optional[str]" = None,
    ) -> "List[USN]":
        """Query to match multiple-USNs.
        @return: List of USN instances based on the the JSON response.
        """
        query_params = {
            "details": details,
            "release": release,
            "limit": limit,
            "offset": offset,
            "order": order,
        }
        security_cls = USN
        request_fmt = API_V1_NOTICES
        return self.query_multiple_items(query_params, request_fmt, security_cls)

    def query_multiple_items(self, query_params, request_fmt, security_cls):
        response, _headers = self.request_url(
            request_fmt, query_params=query_params
        )
        return [security_cls(client=self, response=md) for md in response]
 ```
 
 And this same idea also for `get_cve` and `get_notice`. The main idea here is to have a common function that performs the querys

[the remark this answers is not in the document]
Comment (blackboxsw) [reply uaclient/security.py]: Thanks @lucasmoura. I still feel this is a bit of overkill just because we now have a level of indirection on two very simple methods.  I think we talked post-standup about this and it sounded like you were ok with the initial cut (given that we will probably need to use the batch get_cves and get_notices methods to reduce the number of API round trips when multiple CVEs are related to a single USN or multiple USNs related to a single CVE.

File: uaclient/security.py [hunk 4]
```diff
@@ -0,0 +1,397 @@
...
+def query_installed_pkg_sources() -> "Dict[str, Dict[str, Any]]":
+    """Return a dict of all source packages installed on the system.
+
+    The dict keys will be source package name: "krb5". The value of each
+    item will contain two keys:
+    - package-list: List of binary package names intalled related to the
+         source package
+    - version: the version string of the package installed
+    """
+    out, _err = util.subp(
```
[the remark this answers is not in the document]
blackboxsw: This command will return duplicate source packages here because each binary package derived from that source is listed. So for the call above we would have seen the following 6 lines for all the krb-derived binary pkgs:
```bash
krb5-locales,krb5,1.17-6ubuntu4.1
libgssapi-krb5-2,krb5,1.17-6ubuntu4.1
libk5crypto3,krb5,1.17-6ubuntu4.1
libkrb5-26-heimdal,heimdal,7.7.0+dfsg-1ubuntu1
libkrb5-3,krb5,1.17-6ubuntu4.1
libkrb5support0,krb5,1.17-6ubuntu4.1
```


All said though, we don't need to track the individual package-lists for anything at the moment because USNs and CVEs work in terms of the source package name, not binaries, so I'm modifying this query to only look at source pkg names.
I also added a check to ignore packages which are not installed. The previous dpkg-query didn't track db:Status-Status.

we now run
```bash
dpkg-query -f='${Source},${Version},${db:Status-Status}\n' -W
````

and ignore any line without "installed" status

File: uaclient/cli.py [hunk 5]
```diff
@@ -222,6 +224,32 @@ def attach_parser(parser):
...
+    parser.usage = USAGE_TMPL.format(
+        name=NAME, command="fix <CVE-yyyy-nnnn>|<USN-nnnn>"
+    )
+    parser.prog = "fix"
+    parser._optionals.title = "Flags"
+    parser.add_argument(
+        "security_issue",
+        help=(
+            "Security vulnerability ID to inspect and resolve on this system."
+            " Format: CVE-yyyy-nnnn or USN-nnnn"
```
lucasmoura, line 238: I think we should update the CVE format, since it can accept formats like  `CVE-yyyy-nnnnn`  and `CVE-yyyy-nnnnnn`

uaclient/cli.py
```diff
@@ -222,6 +224,32 @@ def attach_parser(parser):
...
+        ),
+    )
+    return parser
+
+
+def action_fix(args, cfg, **kwargs):
+    if not re.match(security.CVE_OR_USN_REGEX, args.security_issue):
+        # TODO(review of error messaging to reduce output)
+        raise exceptions.UserFacingError(
+            "Invalid issue format issue: {}".format(args.security_issue)
```
lucasmoura, line 248: I think we can drop the second `issue`

features/attached_enable.feature
```diff
@@ -127,7 +127,7 @@ Feature: Enable command behaviour when attached to an UA subscription
             ESM Infra is already enabled.
             See: sudo ua status
             """
-        And I verify that running `apt install -y <pkg-version>` `with sudo` exits `0`
+        And I run `apt install -y <pkg-version>` with sudo, retrying exit [100]
```
lucasmoura, line 130: We should change this to a `When` step. otherwise the test will fail to build

features/unattached_commands.feature
```diff
@@ -121,3 +121,33 @@ Feature: Command behaviour when unattached
...
+            A fix is available in Ubuntu standard updates.
+            The update is already installed.
+            .*✔.* CVE-2020-28196 is resolved.
+            """
+
+        Examples: ubuntu release
+           | release |
+           | bionic  |
+           | focal   |
+           | trusty  |
```
lucasmoura, line 152: Looking at the Jenkins build logs, this CVE does not affect trusty

features/steps/steps.py
```diff
@@ -93,6 +94,23 @@ def cleanup_instance() -> None:
...
+    retries = [5, 5, 10]  # Sleep times to wait between retries
+    while context.process.returncode not in exit_codes.split(","):
+        if not retries:
+            break
+        print(
+            "--- Retrying on exit {exit_code}: {command}".format(
+                exit_code=context.process.returncode, command=command
+            )
+        )
+        time.sleep(retries.pop(0))
```
lucasmoura, line 109: The error here that will abort the execution will be something like:
```
IndexError: pop from empty list
```

However, I think we should better inform that the command failed after all of the retries attempts

lucasmoura:
@blackboxsw just a couple of small issues, but after that, I think we can merge this PR

features/steps/steps.py
blackboxsw: Added a warning "Exhausted retries waiting for edit_codes: x,y" and wrapped in a try/except block

features/unattached_commands.feature
```diff
@@ -121,3 +121,59 @@ Feature: Command behaviour when unattached
...
```
lucasmoura, line 154: I think we can drop trusty from the examples table

uaclient/cli.py
lucasmoura: I think this is still an issue

uaclient/cli.py
lucasmoura: This too

lucasmoura (approved):

Jenkinsfile
```diff
@@ -33,7 +34,6 @@ pipeline {
                 python3 -m venv $TMPDIR
                 . $TMPDIR/bin/activate
                 pip install tox  # for tox supporting --parallel--safe-build
```
paride, line 36: Side note: `--parallel-safe-build` is marked as deprecated in the docs. After a quick search it's not clear to me if there's a direct replacement for it.

PR #1321, opened 2021-01-21
Show errors when behave test fails

lucasmoura:
When running the behave tests in Jenkins, it is pretty hard to debug why one of the steps failed. The reason is that we only output that the command returned a status different than 0. Now we are also going to print the stdout and stderr returned by the command in that situation, allowing us to better debug this issue.

Also, the current timeout for apt to download service credentials is 20 seconds. We are bumping that to 60 seconds in an attempt to reduce flaky errors on the Jenkins builds


blackboxsw (approved):
LGTM

PR #1323, opened 2021-01-21
Add GCP pro test for focal machine

lucasmoura:
Add GCP PRO image support to integration tests

Fixes: #1289


blackboxsw (approved):
Approved and tested with our shared credentials.

PR #1324, opened 2021-01-22
Add FIPS test to AWS and Azure bionic images

lucasmoura:
Create FIPS tests for the generic bionic images on Azure and AWS

### DIFF
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -63,7 +63,7 @@ Feature: Enable command behaviour when attached to an UA staging subscription
         And I run `ua disable livepatch` with sudo
         And I run `apt-get install openssh-client openssh-server strongswan -y` with sudo
         And I run `apt-mark hold openssh-client openssh-server strongswan` with sudo
-        When I run `ua enable <fips-service> --assume-yes` with sudo
+        And I run `ua enable <fips-service> --assume-yes` with sudo
         Then stdout matches regexp:
             """
             Updating package lists
@@ -101,6 +101,11 @@ Feature: Enable command behaviour when attached to an UA staging subscription
             Updating package lists
             A reboot is required to complete disable operation
             """
+        When I run `apt-cache policy ubuntu-fips` as non-root
+        Then stdout matches regexp:
+        """
+        .*Installed: \(none\)
+        """
         When I reboot the `<release>` machine
         When I run `cat /proc/sys/crypto/fips_enabled` with sudo
         Then I will see the following on stdout:
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -138,3 +138,204 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
            | xenial  | n/a       | disabled    |
            | bionic  | n/a       | disabled    |
            | focal   | n/a       | n/a         |
+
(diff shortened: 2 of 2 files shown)


features/staging_commands.feature
```diff
@@ -59,7 +59,29 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+                            "type": "fips",
+                            "directives": {
+                                "additionalPackages": ["ubuntu-fips"]
+                            }
+                        }
+                    ]
+                }
+            }
+        }
+        """
```
blackboxsw, line 78: Why do we need to set this when talking to staging?
If I look at fips directives on a plain staging attach: 
```json
                  "additionalPackages" : [
                     "ubuntu-fips"
                  ],
```

blackboxsw (changes requested):

features/staging_commands.feature
blackboxsw: Shouldn't this be in an integration test tagged w/ `@uses.config.machine_type.aws.generic` ?

lucasmoura: You are right here. This is a mistake. Originally, I tried to use the same fips test for all machine types, but it did not work out and I forgot to remove those steps. I will update the code

features/staging_commands.feature
```diff
@@ -126,6 +131,7 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | xenial  | FIPS         | fips         |https://esm.staging.ubuntu.com/fips/ubuntu xenial/main |
            | bionic  | FIPS         | fips         |https://esm.staging.ubuntu.com/fips/ubuntu bionic/main |
 
+
```
blackboxsw, line 134: unintended white space I think

features/staging_commands.feature
```diff
@@ -197,6 +203,207 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | xenial  | FIPS Updates | fips-updates |https://esm.staging.ubuntu.com/fips-updates/ubuntu xenial-updates/main |
            | bionic  | FIPS Updates | fips-updates |https://esm.staging.ubuntu.com/fips-updates/ubuntu bionic-updates/main |
 
+    @series.xenial
```
blackboxsw, line 206: I think this needs to be @series.bionic right because we don't expect ubuntu-azure-fips to exist on Xenial FIPS PPAs?

features/staging_commands.feature
```diff
@@ -197,6 +203,207 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+    @uses.config.machine_type.azure.generic
+    Scenario Outline: Attached enable of vm-based services in an ubuntu lxd vm
```
blackboxsw, line 208: Can we move these two scenarios into features/attach_validtoken.feature file?

blackboxsw (changes requested):
@lucasmoura thanks for this. I think long-term we should move these FIPS tests out of staging_commands.feature and into attach_validtoken.feature. staging_commands.feature should probably just be used for interaction with contract staging and/or features.

lucasmoura:
@blackboxsw Code updated

blackboxsw (approved):
LGTM! thanks Lucas

PR #1325, opened 2021-01-25
Block enabling FIPS on clouds using Xenial

lucasmoura:
Currently, Azure and GCP do not have fips specific kernel versions for the xenial kernel. Because of that, when enabling fips on
those instances, we would end up installing the stock fips kernel on them, which is problematic, since that kernel is not optimized
for those clouds and could generate problems for the user. Therefore, we are now blocking enabling FIPS on Azure and GCP cloud instances.

We are still missing the GCP integration tests and defining what the message will be when we block FIPS because of this problem

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -2,6 +2,7 @@ from itertools import groupby
 
 from uaclient import apt
 from uaclient.entitlements import repo
+from uaclient.clouds.identity import get_cloud_type
 from uaclient import status, util
 
 try:
@@ -44,13 +45,44 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
     help_doc_url = "https://ubuntu.com/security/certifications#fips"
     _incompatible_services = ["livepatch"]
 
+    def _prevent_fips_on_xenial_cloud_instances(self):
+        cfg_allow_xenial_fips_on_cloud = util.is_config_value_true(
+            config=self.cfg.cfg,
+            path_to_value="features.allow_xenial_fips_on_cloud",
+        )
+
+        if cfg_allow_xenial_fips_on_cloud:
+            return False
+
+        series = util.get_platform_info()["series"]
+
+        if series != "xenial":
+            return False
+
+        cloud_id = get_cloud_type()
+
+        if cloud_id is None:
+            return False
+
+        return bool(cloud_id.lower() in ("azure", "gce"))
+
     @property
     def static_affordances(self) -> "Tuple[StaticAffordance, ...]":
         # Use a lambda so we can mock util.is_container in tests
+        cloud_titles = {"azure": "an Azure", "gce": "a GCP"}
+        cloud_id = get_cloud_type() or ""
+
         return (
             (
                 "Cannot install {} on a container".format(self.title),
                 lambda: util.is_container(),
+                False,
+            ),
+            (
+                status.MESSAGE_FIPS_BLOCK_ON_CLOUD.format(
+                    service=self.title, cloud=cloud_titles.get(cloud_id)
+                ),
+                lambda: self._prevent_fips_on_xenial_cloud_instances(),
                 False,
             ),
         )
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -267,6 +267,10 @@ MESSAGE_INCOMPATIBLE_SERVICE_STOPS_ENABLE = """\
 Cannot enable {service_being_enabled} when {incompatible_service} is enabled
 """
 
+MESSAGE_FIPS_BLOCK_ON_CLOUD = """\
+Ubuntu Xenial does not provide {cloud} optimized FIPS kernel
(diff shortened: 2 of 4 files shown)


uaclient/entitlements/fips.py
```diff
@@ -43,6 +44,25 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
 
     help_doc_url = "https://ubuntu.com/security/certifications#fips"
 
+    def _check_for_xenial_cloud_instances(self):
```
blackboxsw, line 47: Let's name this method to describe what check we are performing. ++ Typing hints. And add a minor docstring.
```suggestion
    def _prevent_fips_on_xenial_cloud_instances(self) -> bool:
```

uaclient/entitlements/fips.py
```diff
@@ -43,6 +44,25 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
...
+        if cfg_allow_xenial_fips_on_cloud:
+            return False
+
+        cloud_id = get_cloud_type()
+
+        if cloud_id and cloud_id.lower() in ("azure", "gce"):
+            series = util.get_platform_info()["series"]
+
+            if series.lower() == "xenial":
+                return True
```
blackboxsw, line 62: Let's pull this series check out and perform that first for two reasons:
- the series check is simple and fast and doesn't rely on cloud-init differences on trusty (no cloud-id cmd on trusty)
- I think it's more likely we'll exit on seeing Bionic instead of Xenial than seeing azure/gce vs AWS.

uaclient/entitlements/fips.py
```diff
@@ -43,6 +44,25 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
...
+
+        if cfg_allow_xenial_fips_on_cloud:
+            return False
+
+        cloud_id = get_cloud_type()
+
+        if cloud_id and cloud_id.lower() in ("azure", "gce"):
+            series = util.get_platform_info()["series"]
+
+            if series.lower() == "xenial":
```
blackboxsw, line 61: series is already lower from `get_platform_info`

uaclient/entitlements/fips.py
```diff
@@ -70,6 +90,25 @@ def static_affordances(self) -> "Tuple[StaticAffordance, ...]":
...
+        is_xenial_on_blocked_cloud = self._check_for_xenial_cloud_instances()
+
+        if is_xenial_on_blocked_cloud:
+            cloud_titles = {"azure": "Azure", "gce": "GCP"}
+
+            cloud_id = get_cloud_type() or ""
+
+            return static_affordances_list + (
+                (
+                    status.MESSAGE_FIPS_BLOCKE_ON_CLOUD.format(
```
blackboxsw, line 102: typo on BLOCKE ?

uaclient/status.py
```diff
@@ -258,6 +258,9 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_REFRESH_SUCCESS = "Successfully refreshed your subscription"
 MESSAGE_REFRESH_FAILURE = "Unable to refresh your subscription"
 
+MESSAGE_FIPS_BLOCKE_ON_CLOUD = """\
```
blackboxsw, line 261: typo on BLOCKE_

uaclient/entitlements/fips.py
```diff
@@ -70,6 +90,25 @@ def static_affordances(self) -> "Tuple[StaticAffordance, ...]":
             ),
         )
 
+        is_xenial_on_blocked_cloud = self._check_for_xenial_cloud_instances()
```
blackboxsw, line 93: It seems strange we are calling this method directly instead of just addding it to static_affordances_list? that lambda in static affordances ensures we will be calling it anyway. We can instead construct the  error_msg = status.MESSAGE_FIPS_BLOCKED_ON_CLOUD.format(cloud_id), lambda: _prevent_fips_on_xenial_cloud_instance

blackboxsw (changes requested):

uaclient/entitlements/fips.py
lucasmoura: Yes, you are right on that. We could do this directly on the existing tuple. I am updating the code

uaclient/entitlements/fips.py
```diff
@@ -43,6 +44,27 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
...
+
+        if series != "xenial":
+            return False
+
+        cloud_id = get_cloud_type()
+
+        if cloud_id is None:
+            return False
+
+        return True if cloud_id.lower() in ("azure", "gce") else False
```
blackboxsw, line 66:
```suggestion
        return bool(cloud_id.lower() in ("azure", "gce"))
```

blackboxsw (approved):

ubuntu-server-builder (approved):

PR #1332, opened 2021-01-27
Fix trusty upgrade test

lucasmoura:
When rebooting the machine on trusty, we were using the raise_on_failure as False to not raise exceptions if cloud-init failed to successfully complete. However, pycloudlib does not throw those exceptions anymore and no longer support that parameter
either. We are removing it from the method call.

Also, some of our integration tests are redundant, in the sense that they are already being test by other scenario or they can be integrated into an existing test. We are now removing those redundant tests.

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -177,7 +177,7 @@ def when_i_reboot_the_machine(context, series):
         context.instance.shutdown(wait=True)
         context.instance.start(wait=False)
         # Trusty -> Xenial upgrades would raise a Paths no run_dir attr failure
-        context.instance.wait(raise_on_cloudinit_failure=False)
+        context.instance.wait()
     else:
         context.instance.restart(wait=True)
 
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import re
 import shlex
+import time
 
 from behave import given, then, when
 from hamcrest import (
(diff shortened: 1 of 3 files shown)


features/steps/steps.py
```diff
@@ -143,7 +144,23 @@ def when_i_attach_staging_token(context, token_type, user_spec):
             ),
             user_spec,
         )
-    when_i_run_command(context, "ua attach %s" % token, user_spec)
+    cmd = "ua attach {}".format(token)
+    when_i_run_command(context, cmd, user_spec, verify_return=False)
+
+    retries = [5, 5, 10]  # Sleep times to wait between retries
+    while context.process.returncode != 0:
+        when_i_run_command(context, "ua detach --assume-yes", user_spec)
```
blackboxsw, line 152: I like the retry logic on inability to attach to staging (reasonable for our integration testing against staging because  that service goes down for deployment fairly regularly).

but I do think we should not have to call `ua detach` during those failures. UA  client should be smart enough on failure to attach to clean up any artifacts it created during attach failure. This is reminiscent of #1337 which I will put up a fix for shortly and we should land that before this PR so that you can drop the `ua detach` I hope.
```suggestion
```

lucasmoura: Okay, fair enough I did that because when I was saw that issue locally, I saw that the machine kept attached. But I could be wrong on that. 

I will drop this line and keep an eye on CI to see if it surface some errors like that

blackboxsw (approved):
Thanks @lucasmoura all good here

PR #1333, opened 2021-01-27
Fix tox-lxd-runner script for xenial

lucasmoura:
When running the tox-lxd-runner script on a xenial machine, we end up downloading a pip
file that uses fstrings on its code. This is not supported on python 3.5, which is used by xenial.
To fix this issue, we are downloading the right pip version when using python 3.5


blackboxsw (approved):
LGTM

PR #1336, opened 2021-01-27
Remove expiry check from contract update

### LINKED ISSUE #1335: Client is checking for contract effectiveness even if attachment succeeded
The contracts service allows an expired contract to be used for a grace period of 14 days after its expiration:

https://github.com/CanonicalLtd/ua-contracts/blob/74d5e786513ca2b02b8297051080241cb28578c1/internal/backend/contract.go#L388

This was added because now that we sell via the new shop, an automatic renewal (e.g., for a monthly subscription) may get stuck due to lack of payment. We don't want to immediately cut off user access, but instead allow them 14 days of access while the customer sorts any payment issues. Even if the payment goes well, the contract may be briefly invalid while the shop finalizes the payment handling (payment is only triggered ~1h after the contract expires).

We don't change the effective dates of the contract, we just take this grace period into account when determining whether we should allow access. Contract effectiveness is checked in the following situations:
- When attaching a machine
- When granting access to a resource (e.g., ESM or Livepatch)

However, there's one function in the client is preventing the grace period from being used:

https://github.com/canonical/ubuntu-advantage-client/blob/a5f10705b7662fb1d5450ddf081b333a19016be2/uaclient/contract.py#L325

That function calls ua-contracts, and regardless of the HTTP status code (even if it's 200), it checks for the contract expiration:

https://github.com/canonical/ubuntu-advantage-client/blob/a5f10705b7662fb1d5450ddf081b333a19016be2/uaclient/contract.py#L375-L380

This may lead to the weird situation where a machine succeeds in attaching, but the client says it wasn't attached. If the contract were not effective, the contracts service would return 403 with a message indicating that the contract is not effective:

https://github.com/CanonicalLtd/ua-contracts/blob/74d5e786513ca2b02b8297051080241cb28578c1/internal/v1/api.go#L519-L525

I'm thinking that perhaps the client shouldn't make any checks at all, and just print a message based on whatever the contract service returns in case it's an error (401/403). But there might have been a reason for this check being here, so I'm not putting it categorically as a suggestion :)

---

On a side note: subscriptions are now an entity in ua-contracts. They are a bundling of **product listings** (a way of selling **products** which will be tied to **contracts**) that may renew automatically. So using **subscriptions** as a proxy for **contract** is now a UX issue, as it may confuse the user about what exactly expired.

Usually, if the contract expired, it *might* mean that the shop subscription having expired could have led to it. But for customers who bought UA outside the shop (e.g., direct sales), the contract may have expired and that would be unrelated to any shop subscription.

lucasmoura:
The contract backend will now allow contract token that have expired to keep working for 14 days. Also, the contract
backend already covers the situation were the contract is no longer valid, which will return a 403 error that is already
treated in the code. Because of that, we are dropping the expiry check for contracts

fix #1335

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 import logging
 import urllib
 
@@ -348,7 +347,7 @@ def request_updated_contract(
     contract_client = UAContractClient(cfg)
     if contract_token:  # We are a mid ua-attach and need to get machinetoken
         try:
-            new_token = contract_client.request_contract_machine_attach(
+            contract_client.request_contract_machine_attach(
                 contract_token=contract_token
             )
         except util.UrlError as e:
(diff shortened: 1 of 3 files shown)


alnvdl (commented):
403 is returned in case the contract token is expired OR the contract expired. Perhaps the returned message should now say: "Expired token or contract. To obtain a new token visit: https://ubuntu.com/advantage"?

blackboxsw (approved):
Thanks @lucasmoura and @alnvdl for the review. 
LGTM

Issue #1326, opened 2021-01-25
notices: auto-attach when already attached does not cleanup notices.json


blackboxsw:
An ubuntu PRO instance, the following notice is always presented even though no auto-attach functionality is running:

```bash
SERVICE       ENTITLED  STATUS    DESCRIPTION
esm-infra     yes       enabled   UA Infra: Extended Security Maintenance (ESM)
fips          yes       enabled   NIST-certified FIPS modules
fips-updates  no        —         Uncertified security updates to FIPS modules
livepatch     yes       n/a       Canonical Livepatch service

NOTICES
Operation in progress: ua auto-attach

Enable services with: ua enable <service>

                Account: 12aad61c-6de4-4e53-a6c6-5aff52a83777
           Subscription: 12aad61c-6de4-4e53-a6c6-5aff52a83777
            Valid until: n/a
Technical support level: essential
```

```/var/lib/ubuntu-advantage/notices.json``` should not contain any entries once auto-attach completes.
This seems to only happen across system reboot on PRO machines, uaclient early exists and doesn't cleanup notices.json

PR #1339, opened 2021-01-29
notices: clear lock file and notice when encountering any exception

blackboxsw:
When the commandline encounters an unexpected error, the
main_error_handler needs to call UAConfig.delete_cache_key("lock")
to ensure any Operation in progress notices messages are also
removed. This most often happens during auto-attach on PRO machines
because it raises a UserFacingError saying:

This machine is already attached to <some-account>

Since main_error_handler doesn't have direct access to an
instantiated UAConfig, we store that UAConfig.delte_cache_key
callable on _CLEAR_LOCK_FILE

Fixes: #1326

## Test Steps

* Launch an Ubuntu PRO Machine with this branch
```bash
UACLIENT_BEHAVE_BUILD_PR=1 UACLIENT_BEHAVE_DESTROY_INSTANCES=0 
UACLIENT_BEHAVE_AWS_ACCESS_KEY_ID=<YOUR_AWS_ACCESS_KEY_ID>  UACLIENT_BEHAVE_AWS_SECRET_ACCESS_KEY=<YOUR_AWS_SECRET_ACCESS_KEY> tox -e behave-awspro-18.04 
--no-capture
```
* That run should leave around a test PRO bionic instance lying around after test, you can ssh to it with your ec2-uaclient-ci-0127-dev-0126-220857.pem in the local directory
`ls -ltr ec2*pem`
* Reboot the machine and run status afterward to confirm no "Operation in progress: auto-attach" message 
```bash
ssh -i ec2.....pem ubuntu@<PRO_IP_ADDRESS>  sudo reboot
ssh -i ec2.....pem ubuntu@<PRO_IP_ADDRESS>  sudo ua status

```
## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -50,10 +50,10 @@ DEFAULT_LOG_FORMAT = (
 STATUS_FORMATS = ["tabular", "json"]
 
 
-# Set a module-level variable here so we don't have to reinstantiate
+# Set a module-level callable here so we don't have to reinstantiate
 # UAConfig in order to determine dynamic data_path exception handling of
 # main_error_handler
-_LOCK_FILE = None
+_CLEAR_LOCK_FILE = None
 
 
 class UAArgumentParser(argparse.ArgumentParser):
@@ -121,8 +121,7 @@ def assert_lock_file(lock_holder=None):
     def wrapper(f):
         @wraps(f)
         def new_f(args, cfg, **kwargs):
-            global _LOCK_FILE
-            lock_file = cfg.data_path("lock")
+            global _CLEAR_LOCK_FILE
             (lock_pid, cur_lock_holder) = cfg.check_lock_info()
             if lock_pid > 0:
                 raise exceptions.LockHeldError(
@@ -133,10 +132,10 @@ def assert_lock_file(lock_holder=None):
             cfg.write_cache("lock", "{}:{}".format(os.getpid(), lock_holder))
             notice_msg = "Operation in progress: {}".format(lock_holder)
             cfg.add_notice("", notice_msg)
-            _LOCK_FILE = lock_file  # Set _LOCK_FILE for cleanup
+            _CLEAR_LOCK_FILE = cfg.delete_cache_key
             retval = f(args, cfg, **kwargs)
-            util.remove_file(lock_file)
-            cfg.remove_notice("", notice_msg)
+            cfg.delete_cache_key("lock")
+            _CLEAR_LOCK_FILE = None  # Unset due to successful lock delete
             return retval
 
         return new_f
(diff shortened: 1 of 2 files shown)


uaclient/cli.py
```diff
@@ -133,10 +132,10 @@ def new_f(args, cfg, **kwargs):
             cfg.write_cache("lock", "{}:{}".format(os.getpid(), lock_holder))
             notice_msg = "Operation in progress: {}".format(lock_holder)
             cfg.add_notice("", notice_msg)
-            _LOCK_FILE = lock_file  # Set _LOCK_FILE for cleanup
+            _CLEAR_LOCK_FILE = cfg.delete_cache_key
             retval = f(args, cfg, **kwargs)
-            util.remove_file(lock_file)
-            cfg.remove_notice("", notice_msg)
+            cfg.delete_cache_key("lock")
```
mitechie, line 137: why a text string vs a constant or something here? Easy for typos across uses.

mitechie: should this be around a try and only _CLEAR_LOCK_FILE when we know it succeeds?

uaclient/tests/test_cli.py
```diff
@@ -340,7 +340,9 @@ def test_function(args, cfg):
         assert mock.sentinel.success == ret
         lock_msg = "Operation in progress: some operation"
         assert [mock.call("", lock_msg)] == m_add_notice.call_args_list
-        assert [mock.call("", lock_msg)] == m_remove_notice.call_args_list
+        assert [
+            mock.call("", "Operation in progress.*")
+        ] == m_remove_notice.call_args_list
```
mitechie, line 345: should there be a test that we hit the new callable?

uaclient/cli.py
blackboxsw: > why a text string vs a constant or something here? Easy for typos across uses.

This is a good point, and I've taken a small tech-debt issue here so we can address this in 26.1 because there are sprawling changes throughout the code that really require that we create a strict defintion of allowed cache_keys
for all of the [UAConfig data_paths](https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/config.py#L60-L68) . That'll be a broad and cosmetic change that I don't want to pollute this PR for 26.0 if that sounds ok?

blackboxsw: > should this be around a try and only _CLEAR_LOCK_FILE when we know it succeeds?

So, if either the retval = f(args, cfg, **kwargs) or  cfg.delete_cache_key("lock") fails with an unexpected error we immediately raise out of this context and don't get to the line where we set _CLEAR_LOCK_FILE = None. 
Once in [main_error_handler we have exception handling](https://github.com/canonical/ubuntu-advantage-client/pull/1339/files#diff-d558bdbcb7ee1bef85fd138367f1e64708ea8271c09cf8e4c19941c24c8883d6R942-R944) that looks to see if a dirty _CLEAR_LOCK_FILE callable is set. Only if _CLEAR_LOCK_FILE is still "dirty" and we didn't get there due to an existing LockHeldError, do we call _CLEAR_LOCK_FILE("lock") which removes the lock file as well as any notices related to that lock "Operation in progress.*" message.

uaclient/tests/test_cli.py
blackboxsw: Sure the `Operation in progress.*` was white box knowledge that the only place in uaclient.config.py which calls with that specific param is inside delete_cache_key("lock") but I'll make it more explicit for reviewers with a mock instead.

lucasmoura (approved):

mitechie (approved):
ty for the investigation into the questions and understand the string is ok as a future enhancement.

Issue #1341, opened 2021-02-01
`status --json` doesn't include info about ESM Apps


mthaddon:
Currently the output of `ubuntu-advantage status --json` doesn't include any info about ESM Apps, even if ESM Apps is enabled. You need to run `ubuntu-advantage status --all` to get that info. This makes it harder for machine ingestion of such information.

PR #1342, opened 2021-02-01
status: include beta services in json formatted output with --all

blackboxsw:
Add logic to show_beta=True services when --format=json --all params
are provided on the commandline.

Fixes: #1341


lucasmoura (approved):

mitechie (approved):
ty

PR #1343, opened 2021-02-01
Block enabling fips on GCP instances

### LINKED ISSUE #1317: GCP: `ua enable fips` should not install linux-fips kernel
GCP images come with a GCP-specific kernel `linux-gcp` installed. `ua enable fips` installs `linux-fips` which is not a gcp-specific kernel. Is this a regression of behavior for GCP environments?

lucasmoura:
Currently, we do not have a gcp optimized fips kernel. Installing the default fips kernel on those instances could case
performance or even system issues. Because of that, we are now blocking enabling fips on those instances until we have the
optimized kernel for gcp.

We can override this behavior but adding the following directive into `uaclient.conf`:
```
allow_default_fips_metapackage_on_gcp: true
```

Fix #1317

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -45,26 +45,45 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
     help_doc_url = "https://ubuntu.com/security/certifications#fips"
     _incompatible_services = ["livepatch"]
 
-    def _prevent_fips_on_xenial_cloud_instances(self):
-        cfg_allow_xenial_fips_on_cloud = util.is_config_value_true(
-            config=self.cfg.cfg,
-            path_to_value="features.allow_xenial_fips_on_cloud",
-        )
+    def _allow_fips_on_cloud_instance(
+        self, series: str, cloud_id: str
+    ) -> bool:
+        """Return False when FIPS is allowed on this cloud and series.
 
-        if cfg_allow_xenial_fips_on_cloud:
-            return False
+        On Xenial Azure and GCP there will be no cloud-optimized kernel so
+        block default ubuntu-fips enable. This can be overridden in
+        config with features.allow_xenial_fips_on_cloud.
 
-        series = util.get_platform_info()["series"]
+        GCP doesn't yet have a cloud-optimized kernel or metapackage so
+        block enable of fips if the contract does not specify ubuntu-gcp-fips.
+        This also can be overridden in config with
+        features.allow_default_fips_metapackage_on_gcp.
 
-        if series != "xenial":
-            return False
 
-        cloud_id = get_cloud_type()
+        :return: False when this cloud, series or config override allows FIPS.
+        """
+        if cloud_id not in ("azure", "gce"):
+            return True
 
-        if cloud_id is None:
-            return False
+        if cloud_id == "gce":
+            if util.is_config_value_true(
+                config=self.cfg.cfg,
+                path_to_value="features.allow_default_fips_metapackage_on_gcp",
+            ):
+                return True
+            return bool("ubuntu-gcp-fips" in super().packages)
 
-        return bool(cloud_id in ("azure", "gce"))
+        # Azure FIPS cloud support
+        if series == "xenial":
+            if util.is_config_value_true(
+                config=self.cfg.cfg,
+                path_to_value="features.allow_xenial_fips_on_cloud",
+            ):
+                return True
+            else:
+                return False
+
+        return True
 
     @property
     def static_affordances(self) -> "Tuple[StaticAffordance, ...]":
@@ -72,6 +91,10 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         cloud_titles = {"azure": "an Azure", "gce": "a GCP"}
         cloud_id = get_cloud_type() or ""
 
+        series = util.get_platform_info().get("series", "")
+        blocked_message = status.MESSAGE_FIPS_BLOCK_ON_CLOUD.format(
+            series=series.title(), cloud=cloud_titles.get(cloud_id)
+        )
         return (
             (
                 "Cannot install {} on a container".format(self.title),
@@ -79,11 +102,9 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                 False,
             ),
             (
-                status.MESSAGE_FIPS_BLOCK_ON_CLOUD.format(
-                    service=self.title, cloud=cloud_titles.get(cloud_id)
-                ),
-                lambda: self._prevent_fips_on_xenial_cloud_instances(),
-                False,
+                blocked_message,
+                lambda: self._allow_fips_on_cloud_instance(series, cloud_id),
+                True,
             ),
         )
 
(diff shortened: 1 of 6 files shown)


uaclient/status.py
```diff
@@ -268,6 +268,9 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_FIPS_BLOCK_ON_CLOUD = """\
 Ubuntu Xenial does not provide {cloud} optimized FIPS kernel
 For help see: https://ubuntu.com/advantage"""
+MESSAGE_FIPS_BLOCK_ON_GCP = """\
+GCP instances does not possess a optimized FIPS kernel yet
```
blackboxsw, line 272:
```suggestion
GCP does not yet provide a cloud-optimized FIPS kernel
```

uaclient/entitlements/fips.py
```diff
@@ -66,6 +66,30 @@ def _prevent_fips_on_xenial_cloud_instances(self):
 
         return bool(cloud_id in ("azure", "gce"))
 
+    def _prevent_default_fips_on_gcp_cloud(self) -> bool:
```
blackboxsw, line 69: It feels a bit like there is an opportinuty to parameterize this method and the other _prevent_fips_on_xenial_cloud_instances into a single method. We also have a bit of duplication in calls to get_platform_info and get_cloud_type that could be reduced if we pass those series and cloud_id values into this method.

blackboxsw (changes requested):
With this added logic there is a lot of repetitive calling of get_cloud_type and get_platform_info()['series'] that I bet we could reduce. 

Here's a diff of the general thoughts I had that might reduce some of these get_cloud_type calls as well as make the static affordance method names a bit less confusing.

I also felt we could keep a single MESSAGE_FIPS_BLOCK_ON_CLOUD given that our messaging should be near identical there.

here's a diff for your thoughts
```diff 
diff --git a/uaclient/entitlements/fips.py b/uaclient/entitlements/fips.py
index ea58615..75be890 100644
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -45,49 +45,40 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
     help_doc_url = "https://ubuntu.com/security/certifications#fips"
     _incompatible_services = ["livepatch"]
 
-    def _prevent_fips_on_xenial_cloud_instances(self):
-        cfg_allow_xenial_fips_on_cloud = util.is_config_value_true(
-            config=self.cfg.cfg,
-            path_to_value="features.allow_xenial_fips_on_cloud",
-        )
-
-        if cfg_allow_xenial_fips_on_cloud:
-            return False
-
-        series = util.get_platform_info()["series"]
-
-        if series != "xenial":
-            return False
+    def _allow_fips_on_cloud_instance(
+        self, series: str, cloud_id: str
+    ) -> bool:
+        """Return True when FIPS is allowed on this cloud and series.
 
-        cloud_id = get_cloud_type()
-
-        if cloud_id is None:
-            return False
+        On Xenial Azure and GCP there will be no cloud-optimized kernel so
+        block default ubuntu-fips enable. This can be overridden in
+        config with features.allow_xenial_fips_on_cloud.
 
-        return bool(cloud_id in ("azure", "gce"))
+        GCP doesn't yet have a cloud-optimized kernel or metapackage so
+        block enable of fips if the contract does not specify ubuntu-gcp-fips.
+        This also can be overridden in config with
+        features.allow_default_fips_metapackage_on_gcp.
 
-    def _prevent_default_fips_on_gcp_cloud(self) -> bool:
-        """
-        Prevent installing ubuntu-fips on GCP instances
 
-        We still do not have an optimized FIPS kernel for GCP
-        instances. Installing the default FIPS kernel on those
-        instances may cause them to have performance or system
-        issues. To avoid that, we are blocking enabling FIPS
-        if we do not have an optimized kernel on GCP.
+        :return: True when this cloud, series or config override allows FIPS.
         """
-        cfg_allow_default_fips_metapackage_on_gcp = util.is_config_value_true(
-            config=self.cfg.cfg,
-            path_to_value="features.allow_default_fips_metapackage_on_gcp",
-        )
-
-        if cfg_allow_default_fips_metapackage_on_gcp:
-            return False
-
-        cloud_id = get_cloud_type()
-        if cloud_id and cloud_id == "gce":
+        if cloud_id not in ("azure", "gce"):
+            return True
+
+        if cloud_id == "gce":
+            if util.is_config_value_true(
+                config=self.cfg.cfg,
+                path_to_value="features.allow_default_fips_metapackage_on_gcp",
+            ):
+                return True
             return bool("ubuntu-gcp-fips" not in super().packages)
 
+        # Azure FIPS cloud support
+        if series == "xenial" and util.is_config_value_true(
+            config=self.cfg.cfg,
+            path_to_value="features.allow_xenial_fips_on_cloud",
+        ):
+            return True
         return False
 
     @property
@@ -96,6 +87,10 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         cloud_titles = {"azure": "an Azure", "gce": "a GCP"}
         cloud_id = get_cloud_type() or ""
 
+        series = util.get_platform_info().get("series")
+        blocked_message = status.MESSAGE_FIPS_BLOCK_ON_CLOUD.format(
+            series=series, cloud=cloud_titles.get(cloud_id)
+        )
         return (
             (
                 "Cannot install {} on a container".format(self.title),
@@ -103,16 +98,9 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                 False,
             ),
             (
-                status.MESSAGE_FIPS_BLOCK_ON_GCP,
-                lambda: self._prevent_default_fips_on_gcp_cloud(),
-                False,
-            ),
-            (
-                status.MESSAGE_FIPS_BLOCK_ON_CLOUD.format(
-                    service=self.title, cloud=cloud_titles.get(cloud_id)
-                ),
-                lambda: self._prevent_fips_on_xenial_cloud_instances(),
-                False,
+                blocked_message,
+                lambda: self._allow_fips_on_cloud_instance(series, cloud_id),
+                True,
             ),
         )
 
diff --git a/uaclient/status.py b/uaclient/status.py
index d9b6493..19ad40c 100644
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -266,12 +266,8 @@ Cannot enable {service_being_enabled} when {incompatible_service} is enabled
 """
 
 MESSAGE_FIPS_BLOCK_ON_CLOUD = """\
-Ubuntu Xenial does not provide {cloud} optimized FIPS kernel
+Ubuntu {series} does not provide {cloud} optimized FIPS kernel
 For help see: https://ubuntu.com/advantage"""
-MESSAGE_FIPS_BLOCK_ON_GCP = """\
-GCP instances does not possess a optimized FIPS kernel yet
-For help see: https://ubuntu.com/advantage"""
-
 ERROR_INVALID_CONFIG_VALUE = """\
 Invalid value for {path_to_value} in /etc/ubuntu-advantage/uaclient.conf. \
 Expected {expected_value}, found {value}."""
```

lucasmoura:
@blackboxsw That is a great refactoring. Thanks for that. I have updated the code with it

blackboxsw (approved):
+1 thanks for this

Issue #1345, opened 2021-02-03
fips: application_status needs to rely on /proc/sys/crypto/fips_enabled == 1


blackboxsw:
We should check os,path.exists(/proc/sys/crypto/fips_enabled) and util.load_file(" /proc/sys/crypto/fips_enabled") == 1 to determine if fips is both installed and active. 
This applies to both fips and fips-updates kernels.

Just because -fips kernel is running doesn't mean it is active.

PR #1346, opened 2021-02-03
Check proc file to verify if fips is enabled in a machine

lucasmoura:
## Proposed Commit Message
Update check that verifies if fips is enabled

Currently, we to check if fips is enabled, we verify if the fips apt sources are present in the system and the fips kernel is being used. However, if the /proc/sys/crypto/fips_enabled file doesn't exist or is set to 0, the system will not be on fips mode. Because of that, we are checking the proc file instead of the running kernel to dictate if fips is enabled or not.

We are also updating our fips tests:

* We are no longer checking for the `/proc/sys/crypti/fips_enabled`  file after disabling fips
* Adding a test for enabling fips-updates on fips enabled machine

Fix: #1345 

## Test Steps
Just run the integration tests in this PR

## Desired commit type::

## Checklist:

### DIFF
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -310,3 +310,68 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | release |
            | bionic  |
            | xenial  |
+
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable fips-updates on fips enabled vm
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token_staging` with sudo
+        And I run `ua disable livepatch` with sudo
+        And I run `apt-get install openssh-client openssh-server strongswan -y` with sudo, retrying exit [100]
+        And I run `ua enable fips --assume-yes` with sudo
+        Then stdout matches regexp:
+            """
+            Updating package lists
+            Installing FIPS packages
+            FIPS enabled
+            A reboot is required to complete install
+            """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            fips +yes                enabled
+            """
+        When I reboot the `<release>` machine
+        And  I run `ua enable fips-updates --assume-yes` with sudo
+        Then stdout matches regexp:
+            """
+            Updating package lists
+            Installing FIPS Updates packages
+            FIPS Updates enabled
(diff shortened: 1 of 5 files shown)


uaclient/entitlements/fips.py
```diff
@@ -163,9 +166,16 @@ def application_status(self) -> "Tuple[status.ApplicationStatus, str]":
         super_status, super_msg = super().application_status()
         if super_status != status.ApplicationStatus.ENABLED:
             return super_status, super_msg
-        running_kernel = util.get_platform_info()["kernel"]
-        if running_kernel.endswith("-fips"):
-            return super_status, super_msg
+
+        if os.path.exists(self.FIPS_PROC_FILE):
+            if util.load_file(self.FIPS_PROC_FILE) == 1:
```
blackboxsw, line 171: load_file doesn't return an int, so this will fail to ever show ua fips status as enabled I think
```bash
python3 -c 'from uaclient.util  import load_file; print(type(load_file("/proc/sys/crypto/fips_enabled")))
<class 'str'>

```

features/staging_commands.feature
```diff
@@ -310,3 +310,68 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+            A reboot is required to complete install
+            """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            fips +yes                n/a
+            """
+        And stdout matches regexp:
+            """
+            fips-updates +yes                enabled
```
blackboxsw, line 351: I'm not certain how this test actually passes. on my multipass kvm instance which had fips-updates enabled, I install your latest branch and I see
```bash
fips          yes       disabled  NIST-certified FIPS modules
fips-updates  yes       disabled  Uncertified security updates to FIPS modules

```

blackboxsw (changes requested):
I think we need that check to look for "1" instead of 1

blackboxsw (approved):
Thanks @lucasmoura looks good. I think we might need to talk about expected behavior when FIPS is enabled, yet we aren't running a FIPS kernel as we still report that as ENABLED, "Reboot required". But this branch is good as is

blackboxsw:
@lucasmoura I added a Merge commit to this because no commit messages in your set contained a "Fixes: #1346" footer. If one of those commits does contain that Fixes: #<id> in the future I will omit the merge commit.

PR #1354, opened 2021-02-05
Fix auto-attach issue when comparing numeric iid

lucasmoura:
Given an attached machine, when we run auto-attach we verify if the current instance id matches the instance id in the uaclient
cache. If they do not match, we proceed with the auto-attach. If the instance id could be converted to a number we would have a problem, since we read the old instance id from a json file, which automatically converts the instance id to int while the current instance id always returns a string. Even if the instance id matches, we would say that they don't because of the type difference.
We are fixing that type mismatch comparison now

## Proposed Commit Message
Fix auto-attach issue when comparing numeric iid

## Test Steps
Run the new unit test

## Desired commit type::

## Checklist:


blackboxsw (approved):
Great work!

PR #1356, opened 2021-02-06
WIP cli: clean subcommand to remove artifacts and cache

blackboxsw:
# WIP looking for input 
Request for input
Hey folks I was wondering what your thoughts were about offering a `ua clean` command which would remove any logs and artifacts left on the system by UA client. It'd help folks mastering golden images clean the system prior to generating those clean images for cloning.

I just threw this together to chat about Monday. If desireable, then I can fix up integration tests for it.

## Proposed Commit Message

>
>
>

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -192,6 +192,20 @@ def assert_not_attached(f):
     return new_f
 
 
+def clean_parser(parser):
+    """Build or extend an arg parser for clean subcommand."""
+    parser.prog = "clean"
+    parser.usage = USAGE_TMPL.format(name=NAME, command=parser.prog)
+    parser._optionals.title = "Flags"
+    parser.add_argument(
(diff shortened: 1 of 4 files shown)


lucasmoura (commented):
I think it is a good idea to have a clean command, specially since we will probably need to support this type of operation on the golden images.

However, should the `clean`  command also performs a `detach` operation ? Since we would need to do that on the golden image too, maybe we could have a flag to perform this operation as well ?

blackboxsw:
We discussed this with @mitechie at standup and there is interest in having a command or script that can perform both UA cleanup and cloud-init cleanup for golden image creation. Since this is much broader than UA Client itself, it probably shouldn't be baked into the primary CLI. We'll deliver this instead as a separate tool/script/binary which can be run.

PR #1358, opened 2021-02-06
wip: ignore 403 if machine-id changes

blackboxsw:
#WIP for discussion Monday
Better handling of single 403 error code if machine-id changed on a system
We shouldn't break auto attach if we don't have perms to "detach" a machine

## Proposed Commit Message

>
>
>

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -165,12 +165,19 @@ class UAContractClient(serviceclient.UAServiceClient):
         self, machine_token: str, contract_id: str, machine_id: str = None
     ) -> "Dict":
         """Report the attached machine should be detached from the contract."""
-        return self._request_machine_token_update(
-            machine_token=machine_token,
-            contract_id=contract_id,
-            machine_id=machine_id,
-            detach=True,
-        )
+        try:
+            return self._request_machine_token_update(
+                machine_token=machine_token,
+                contract_id=contract_id,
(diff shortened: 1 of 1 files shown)


uaclient/contract.py
```diff
@@ -165,12 +165,19 @@ def detach_machine_from_contract(
...
-        )
+        try:
+            return self._request_machine_token_update(
+                machine_token=machine_token,
+                contract_id=contract_id,
+                machine_id=machine_id,
+                detach=True,
+            )
+        except ContractAPIError as e:
+            if detach and e.code == 403 and "token invalid for machine" in e:
```
lucasmoura, line 176: It seems that the `detach` variable does not exist.

After we fix that, what would happen to PRO snapshot ? It will not be able to perfom any `uaclient` operation until it cleans the `auclient` artifacts, right ?

blackboxsw:
Closing this as we will instead cache /etc/machine-id like we do instance-id and avoid calling this detach route on the contracts server if we detect that change in machine-id because we know it will fail.

uaclient/contract.py
blackboxsw: Right sorry I moved where this code from where I originally wrote it in _request_machine_token_update and decided it was better handled here in detach_machine_from_contract. So, yes just dropping that detach check would work.

PR #1362, opened 2021-02-08
integration: block auto-attach on PRO base images

lucasmoura:
## Proposed Commit Message
integration: block auto-attach on PRO base images

Currently, we only create the disable_auto_attach entry on uaclient.conf when we are launching the test image. This is not working on trusty because cloud-init is editing that file after we have already run the auto-attach script. Therefore, we are moving the creation of this directive to the base image (the one that will be snapshoted), which will guarantee that the directive will be there once the test image runs

## Test Steps
Run both AWS and Azure trusty PRO integration tests

## Desired commit type::

## Checklist:


blackboxsw (approved):
LGTM!

PR #1363, opened 2021-02-08
Check machine id on detach from machine

lucasmoura:
## Proposed Commit Message
contract: check machine id on detach

When we running a detach operation we also call the contracts backend to delete the machine from there as well. However, if we detect a machine id change in the system, that call will not succeed, failing the entire operation. To avoid that, we are now checking if the machine id in the cache and current machine id match before doing the detach call to the contracts backend

## Desired commit type::

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+artifacts/*
 *debhelper*
 debian/ubuntu-advantage-tools/
 debian/ubuntu-advantage-tools.substvars
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -165,6 +165,17 @@ class UAContractClient(serviceclient.UAServiceClient):
         self, machine_token: str, contract_id: str, machine_id: str = None
     ) -> "Dict":
         """Report the attached machine should be detached from the contract."""
+        curr_machine_id = self._get_platform_data(machine_id=None).get(
+            "machineId", ""
+        )
+        past_machine_id = self.cfg.read_cache("machine-id")
+
+        if str(curr_machine_id) != str(past_machine_id):
+            logging.debug(
+                "Found new machine-id. Do not call detach on contract backend"
+            )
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+artifacts/*
```
blackboxsw, line 1: good add

uaclient/contract.py
```diff
@@ -165,6 +165,17 @@ def detach_machine_from_contract(
         self, machine_token: str, contract_id: str, machine_id: str = None
     ) -> "Dict":
         """Report the attached machine should be detached from the contract."""
+        curr_machine_id = self._get_platform_data(machine_id=None).get(
+            "machineId", ""
+        )
+        past_machine_id = self.cfg.read_cache("machine-id")
```
blackboxsw, line 171: We need a data_paths defintion  "instance-id": DataPath("machine-id", True), otherwise delete_cache will leave files around in /var/lib/ubuntu-advantage/ that we shouldn't.

Also while we are at it, I see we are missing
        "machine-access-cis": DataPath("machine-access-fips.json", True),
and 
        "machine-access-cis": DataPath("machine-access-fips-updates.json", True),
. Can we add check in the features/attached_commands.feature  "detach integration test that asserts /var/lib/ubuntu-advantage is empty after detach

blackboxsw (changes requested):
Generally good, can we fix the UAConfig.data_paths  for this minimally. Your call about fixing data_paths for fips fips-updates

uaclient/contract.py
blackboxsw: @lucasmoura pointed out that we already had a DataPath("machine-id"..) in the module. So disregard this request. I'll open a separate PR.

blackboxsw (approved):

PR #1365, opened 2021-02-09
Fix bionic vm test when enable fips with livepatch

lucasmoura:
## Proposed Commit Message
Fix bionic vm test when enable fips with livepatch

We have a bionic vm test where we enable livepatch and we try to enable fips afterwards. On that scenario, we also create a uaclient.conf directive to block disabling services on enable operations. We want to verify if in that situation, fips is not enabled
becaused livepatch is. However, we are checking the error message on stdout, but with the new incompatible services
abstraction, that message will be on stderr if we have the uaclient.conf directive

## Test Steps
Run the modifed integration test

## Desired commit type::

## Checklist:

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -356,6 +356,9 @@ Feature: Enable command behaviour when attached to an UA subscription
         And I will see the following on stdout
             """
             One moment, checking your subscription first
+            """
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
LGTM, got over my config issues locally to confirm.

PR #1369, opened 2021-02-11
fix: make subcommand public and remove --beta required flag

blackboxsw:
## Proposed Commit Message

Since we won't release until `ua fix` is production ready,
we can remove the --beta param and allow this command to show
in ua --help.

## Test Steps
CI should cover this, just a revert of a previous accepted commit.

## Desired commit type::

 - [*] Rebase and merge, no merge commit, rely only on existing commit messages

## Checklist:

 - [*] I have updated or added any unit tests accordingly

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -226,16 +226,10 @@ def attach_parser(parser):
 def fix_parser(parser):
     """Build or extend an arg parser for fix subcommand."""
     parser.usage = USAGE_TMPL.format(
-        name=NAME, command="fix --beta <CVE-yyyy-nnnn+>|<USN-nnnn-d+>"
+        name=NAME, command="fix <CVE-yyyy-nnnn+>|<USN-nnnn-d+>"
     )
     parser.prog = "fix"
     parser._optionals.title = "Flags"
-    parser.add_argument(
-        "--beta",
-        required=True,
-        action="store_true",
-        help="allow using this beta command",
-    )
     parser.add_argument(
         "security_issue",
         help=(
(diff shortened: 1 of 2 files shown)


uaclient/cli.py
```diff
@@ -226,30 +226,25 @@ def attach_parser(parser):
...
-            " Format: CVE-yyyy-nnnn, CVE-yyyy-nnnnnnn or USN-nnnn-dd"
+            " Format: CVE-yyyy-nnnn or USN-nnnn"
         ),
     )
     return parser
 
 
 def action_fix(args, cfg, **kwargs):
     if not re.match(security.CVE_OR_USN_REGEX, args.security_issue):
+        # TODO(review of error messaging to reduce output)
```
mitechie, line 245: I've not reviewed how we handle todo in the codebase. I want to raise the question on it's cool to have a todo, but in the past I've liked projects that the todo referenced a bug/issue number that could provide more context, discussion, and progress indication.

uaclient/cli.py
```diff
@@ -837,7 +832,7 @@ def get_parser():
         help="refresh Ubuntu Advantage services from contracts server",
     )
     parser_fix = subparsers.add_parser(
-        "fix"  # BETA omit help string to hide subcommand from help
+        "fix", help="fix CVE and USN security issues on this machine"
```
mitechie, line 835: what about 

    help="check for and mitigate the impact of a CVE/USN on this system"

mitechie (approved):
ty, one comment and one suggestion on help text.

uaclient/cli.py
blackboxsw: Good suggestion here and we should make the TODO() reference an issue. Turns out, this commit wasn't a clean revert, I shouldn't have included that comment in this as Lucas had addressed that point when the former PR was in WIP status. I've reapplied the proper diff which included his review comments. So this comment goes away.

PR #1370, opened 2021-02-12
[WIP] Update pycloudlib

lucasmoura:
## Proposed Commit Message
Make lxd containers execute commands by ssh

## Test Steps
Currently, we are only using ssh to execute commands on lxd virtual machines. With recent modifications on pycloudlib, lxd instances are using ssh by default. We are now migrating our lxd containers to execute commands by ssh too.

Also, pycloudlib now handles the profile creation of lxd virtual machines. Therefore, we don't need to support that code on uaclient anymore

PS: We are still waiting on this pycloudlib [PR](https://github.com/canonical/pycloudlib/pull/117) to land, so we can bump the pycloudlib version we are using in the project

## Desired commit type::

## Checklist:

### DIFF
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -730,15 +730,3 @@ class LXDContainer(_LXD):
     def pycloudlib_cls(self):
         """Return the pycloudlib cls to be used as an api."""
         return pycloudlib.LXDContainer
-
-    def manage_ssh_key(self, private_key_path: "Optional[str]" = None) -> None:
-        """Create and manage ssh key pairs to be used in the cloud provider.
-
-        On a LXD container, we do not use ssh keys to communicate with the
-        instance. Therefore, this method should not be used.
-
-        :param private_key_path:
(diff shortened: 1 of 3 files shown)


blackboxsw (changes requested):
I think we should be tweaking integration-requirements.txt and grabbing the latest commit in pycloudlib now that your lxc-setup.sh has landed.

lucasmoura:
@blackboxsw Updated to use the latest version of pycloudlib

blackboxsw (commented):
While this drop looks good because this is in pycloudlib proper, I think we might end up rejecting this after a conversation on this tomorrow because pycloudlib might eject that functionality and let uaclient integration tests carry that burden.

blackboxsw (approved):
LGTM as that CI failure seems intermittent on 14.04. We will double check next CI runs to determine frequency of this error and address it if it shows up.

PR #1375, opened 2021-02-16
security: closer representation of cve and usn responses

blackboxsw:
## Proposed Commit Message
CVE and USN classes will now more closely represent the response from
the Security API.

- Remove CVE.title because this is an attribute of USN.
- Rename CVE._get_related_notices -> public get_notices_metadata.
- Add comparable cached USN.get_cves_metadata
  to avoid costly API round trips

Perform usn.title lookup for a given CVE up in
get_cve_affected_packages_status instead of via a CVE.title property.

## Test Steps
This is a refactor to ensure CVE and USN more cleanly represent the API response. 
Any costly "lookups" are now placed below a get_X_metadata method which will cache the response.

But, to test locally
```bash
 PYTHONPATH=. python3 -m uaclient.cli --debug fix CVE-2020-1472
 PYTHONPATH=. python3 -m uaclient.cli --debug fix USN-4510-1
```

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -125,7 +125,7 @@ class UASecurityClient(serviceclient.UAServiceClient):
 class CVEPackageStatus:
     """Class representing specific CVE PackageStatus on an Ubuntu series"""
 
-    def __init__(self, cve_response: "Dict[str, str]"):
+    def __init__(self, cve_response: "Dict[str, Any]"):
         self.response = cve_response
 
     @property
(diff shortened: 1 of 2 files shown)


uaclient/security.py
```diff
@@ -191,31 +191,21 @@ def __init__(self, client: UASecurityClient, response: "Dict[str, Any]"):
...
-    def _get_related_notices(self):
-        if hasattr(self, "notices"):
-            return self.notices
-        self.notices = {}
-        for notice_id in self.response.get("notices", []):
-            self.notices[notice_id] = self.client.get_notice(
-                notice_id=notice_id
-            )
-        return self.notices
+    def get_notices_metadata(self):
```
lucasmoura, line 194: @blackboxsw Maybe I am missing something here, but I couldn't figure out what is the new cache mechanism we are introducing here. The old `_get_related_notices` method seems to already covers the cache that we are implementing.

Also, I could not spot that cache mechanism on the USN changes as well

lucasmoura (commented):
It seems that our integration tests are failing on the fix command because of connection issues:

```
stderr: Failed to connect to authentication server
```

Maybe we could have a retry mechanism for that command ?

uaclient/security.py
blackboxsw: > @blackboxsw Maybe I am missing something here, but I couldn't figure out what is the new cache mechanism we are introducing here. The old `_get_related_notices` method seems to already covers the cache that we are implementing.
> 
> Also, I could not spot that cache mechanism on the USN changes as well
sorry about the PR incorrect description (I updated it). This PR only renames `_get_related_notices` to a public method `get_notices_metadata`. 

The `USN.get_cves_metadata` is cached on this instance as `self._cves` instance attr using the same type of mechanism

blackboxsw:
> It seems that our integration tests are failing on the fix command because of connection issues:
> 
> ```
> stderr: Failed to connect to authentication server
> Check your Internet connection and try again
> ```
> 
> Maybe we could have a retry mechanism for that command ?

I think that is warranted. I added retries on our integration tests in the next PR after this in a separate run because we wanted to let Security API folks sort this issue, but I agree we probably want a simple retry bit of logic for any remote API interaction in general. We just need a good way to measure failures/outages in those APIs so we can actually be aware of those issues instead of just hiding them in retries in the log files.

uaclient/security.py
lucasmoura: Oh okay, I missed the new method, I was focusing only on the method that were already being used on the code. My mistake

lucasmoura (approved):

PR #1393, opened 2021-02-24
Allow printing usefull cmds on uaclient operations

lucasmoura:
## Proposed Commit Message
Allow printing apt commands

For ua fix, we will need to print some useful apt commands that we are performing to the user, such as `apt update`, `apt upgrade`. We are now allowing the run_apt_command function to print those commands when needed

## Discussion

As discussed in issue #1400, we will only need to print a subset of the commands that we run on uaclient. They are:

* uaclient top level commands (`attach`, `enable`
* Apt commands like `apt update`, `apt upgrade <pkg-name>`

Because of that, I think `ua fix` will be able to control how to print those commands directly by adding this functionality

## Test Steps
Run the new unit test introduced in the PR

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -129,7 +129,10 @@ def _parse_apt_update_for_invalid_apt_config(apt_error: str) -> str:
 
 
 def run_apt_command(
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
LGTM!

Issue #1397, opened 2021-02-24
fix: provide mechanism to prompt for subscription setup when unattached.


blackboxsw:
When unattached and `ua fix USN` determines an update is available. Prompt the user to attach the machine

```bash
The update is not installed because this system is not attached to a
subscription that covers these packages.

Choose: [S]ubscribe at ubuntu.com [A]ttach existing token [C]ancel 
```

Issue #1398, opened 2021-02-24
fix: some packages not being listed as installed security.query_installed_source_pkg_versions


blackboxsw:
```bash
$ dpkg -l | grep zip
...
ii  zip                                        3.0-11build1                         amd64        Archiver for .zip files


$ PYTHONPATH=. python3 -c 'from uaclient.security import query_installed_source_pkg_versions; print('zip' in query_installed_source_pkg_versions().keys())'
False
```

Packages which don't define a Source: (like zip, xterm ubuntu-advantage-tools etc.) are ignored by query_installed_source_pkg_versions.

lucasmoura:
I think we can close this issue, right ?

PR #1399, opened 2021-02-25
fix: correct installed_pkgs query. prompt when unattached

blackboxsw:
Prompt for ua attach when unattached and `ua fix CVE-XX` reports
an available package affected.

Add a util.prompt_choices function which defines the input string
and desirable choices/responses. prompt_choices will loop until the
valid choice is provided.

In testing ua fix CVE-2021-27135 it was noted that xterm wasn't
listed as installed. Fix dpkg-query to determine package name based
on either ${Source} or ${Package} if Source is not set.

Fixes: #1397, #1398

## Proposed Commit Message
Rebase and merge

## Test Steps
```bash
cat > 1.yaml <<EOF
#cloud-config                                                                   
package_upgrade: true                                                           
apt:                                                                            
  sources:                                                                      
    ua.list:                                                                    
       source: deb http://ppa.launchpad.net/canonical-server/ua-client-daily/ubuntu $RELEASE main
       keyid: 94E187AD53A59D1847E4880F8A295C4FB8B190B7 
packages: [ubuntu-advantage-tools]
EOF

tox -e lxc launch ubuntu-daily:bionic dev-b -c user.user-data="$(cat 1.yaml)"
lxc file push uaclient/security.py dev-b/usr/lib/python3/dist-packages/uaclient/
lxc file push uaclient/status.py dev-b/usr/lib/python3/dist-packages/uaclient/
lxc file push uaclient/util.py dev-b/usr/lib/python3/dist-packages/uaclient/

# downgrade xterm
lxc exec dev-b apt-get install xterm=330-1ubuntu2
lxc exec dev-b ua fix CVE-2021-27135
```
## Desired commit type::

## Checklist:

 - [ ] I have updated or added any integration tests accordingly # maybe will add them after initial review

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -182,6 +182,11 @@ class CVEPackageStatus:
         return "UNKNOWN: {}".format(self.status)
 
     @property
+    def requires_ua(self) -> bool:
+        """Return True if the package requires an active UA subscription."""
+        return bool(self.pocket_source != "Ubuntu standard updates")
+
+    @property
     def pocket_source(self):
         """Human-readable string representing where the fix is published."""
         if self.pocket == "esm-infra":
@@ -388,16 +393,14 @@ def get_cve_affected_packages_status(
     return affected_pkg_versions
 
 
-def prompt_for_affected_packages(
-    cfg: UAConfig,
-    issue_id: str,
-    affected_pkg_status: "Dict[str, CVEPackageStatus]",
-    installed_packages: "Dict[str, str]",
-) -> None:
-    """Process security CVE dict returning a CVEStatus object.
+def print_affected_packages_header(
+    issue_id: str, affected_pkg_status: "Dict[str, CVEPackageStatus]"
+):
+    """Print header strings describing affected packages related to a CVE/USN.
 
-    Since CVEs point to a USN if active, get_notice may be called to fill in
-    CVE title details.
+    :param issue_id: String of USN or CVE issue id.
+    :param affected_pkg_status: Dict keyed on source package name, with active
+        CVEPackageStatus for the current Ubuntu release.
     """
     count = len(affected_pkg_status)
     if count == 0:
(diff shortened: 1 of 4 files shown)


uaclient/security.py
```diff
@@ -410,6 +420,8 @@ def prompt_for_affected_packages(
     )
     msg_suffix = ": " + ", ".join(affected_pkg_status.keys())
     msgs = [msg + msg_suffix]
+    upgrade_packages = []
+    cloud_type = get_cloud_type()
```
lucasmoura, line 424: It seems that we are not using this variable ? Also, why would we need the cloud_type in this function ?

uaclient/tests/test_security.py
```diff
@@ -543,8 +552,81 @@ def test_result_keyed_by_source_package_name(
...
+                {"sl": "2.0"},
+                [
+                    "1 affected package is installed: sl\n(1/1) sl:\n",
+                    "The update is not installed because this system is not"
+                    " attached to a\nsubscription that covers these packages.",
+                ],
+            ),
+        ),
+    )
+    @mock.patch("uaclient.security.get_cloud_type", return_value="aws")
```
lucasmoura, line 610: If we remove the `get_cloud_type` call, we can drop this mock

lucasmoura:
@blackboxsw Another question, when reproducing the test scenario you proposed we are downgrading a package which already has a fix on `Ubuntu Standard Updates`. However, we trigger the user in that scenario that it needs an uaclient subscription to solve that issue, which is not true.

Shouldn't we just ask that question if the updates cannot be found on the existing ubuntu repositories ?

uaclient/security.py
blackboxsw: Ahh this should be a followup PR. I accidentally left this in, it will be used later. I'll back it out of this PR

blackboxsw:
> @blackboxsw Another question, when reproducing the test scenario you proposed we are downgrading a package which already has a fix on `Ubuntu Standard Updates`. However, we trigger the user in that scenario that it needs an uaclient subscription to solve that issue, which is not true.
> 
> Shouldn't we just ask that question if the updates cannot be found on the existing ubuntu repositories ?

Thanks, I originally thought to wait on this, but it was low-hanging fruit for this branch. I refactored functions and made this part smarter so we don't have to refactor later.

lucasmoura (approved):

Issue #1389, opened 2021-02-24
ua disable fips doesn't work in ua client 27


davecore82:
I'm trying to disable FIPS from an Ubuntu Pro FIPS 18.04 image in AWS. I updated to the latest ua client in the daily PPA. I have a prompt to disable it but it fails:

```
ubuntu@ip-172-31-60-238:~$ sudo add-apt-repository ppa:canonical-server/ua-client-daily

ubuntu@ip-172-31-60-238:~$ sudo apt install ubuntu-advantage-pro ubuntu-advantage-tools

ubuntu@ip-172-31-60-238:~$ ua version
27.0-945~gedf4a7e~ubuntu18.04.1

ubuntu@ip-172-31-60-238:~$ ua status
SERVICE ENTITLED STATUS DESCRIPTION
cis-audit no — Center for Internet Security Audit Tools
esm-infra yes enabled UA Infra: Extended Security Maintenance
fips yes enabled NIST-certified FIPS modules
fips-updates no — Uncertified security updates to FIPS modules
livepatch yes n/a Canonical Livepatch service
[...]

ubuntu@ip-172-31-60-238:~$ sudo ua disable fips
This will disable access to certified FIPS packages.
Are you sure? (y/N) y
Could not enable FIPS.

ubuntu@ip-172-31-60-238:~$ ua status
SERVICE ENTITLED STATUS DESCRIPTION
cis-audit no — Center for Internet Security Audit Tools
esm-infra yes enabled UA Infra: Extended Security Maintenance
fips yes enabled NIST-certified FIPS modules
fips-updates no — Uncertified security updates to FIPS modules
livepatch yes n/a Canonical Livepatch service
[...]

```
I tried rebooting after but I'm still running the fips kernel and fips is enabled:

```
ubuntu@ip-172-31-60-238:~$ uname -a
Linux ip-172-31-60-238 4.15.0-2000-aws-fips #4-Ubuntu SMP Tue Jan 28 12:41:43 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

ubuntu@ip-172-31-60-238:~$ cat /proc/sys/crypto/fips_enabled
1
```

mitechie:
Thanks, we've got some manual instructions to help but especially with a FIPS pro image that's built and shipped with the single FIPS kernel it's a risky change that we'd prefer the user/operator be involved with. 

[...]

Issue #1374, opened 2021-02-16
Improve error message and/or retries on intermittent networking issues against Security API server


blackboxsw:
https://jenkins.canonical.com/server-team-jenkins-private/blue/organizations/jenkins/github-ci%2Fubuntu-advantage-client/detail/PR-1373/2/pipeline#step-151-log-1973

```bash
      Error executing command: ua fix CVE-2020-28196
      stdout: 
      stderr: Failed to connect to authentication server
      Check your Internet connection and try again
```

PR #1403, opened 2021-02-25
Fix FIPS message when disable operation fails

lucasmoura:
## Proposed Commit Message
Fix FIPS message when disable operation fails

As seen in #1389, FIPS is incorrectly using an enable message when the disable operations fails. We are now updating that to display the right disable message

## Test Steps
Run the new unit test added in this PR

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -232,7 +232,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                 ["apt-get", "remove", "--assume-yes"]
                 + apt_options
                 + list(remove_packages),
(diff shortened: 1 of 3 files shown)


blackboxsw (changes requested):
~Thanks @lucasmoura can you ammend your commit to add the linked issue to the commit message so when we rebase and merge github automatically closes this issue.
I'd like to resolve this issue specifically for this PR, because the other context I mentioned in that bug was only for clarification about our release process on which we wait for PRO images~
```bash

Fixes: #1389 
```

Update we talked and we'll just link this merged commitish and comment on the issue #1389 that the error message component is fixed.

blackboxsw (approved):

blackboxsw:
Known connection error w/ security api: #1374

Issue #1401, opened 2021-02-25
fix: run apt commands to install missing pkgs to fix a CVE from Ubuntu standard updates pockets security && updates


blackboxsw:
When a vm disables daily apt updates and unattended upgrades, a stale system can be compromised and report that it is affected by a CVE. 

The resolution for these systems is for `ua fix` to apt-get update &&apt-get install  --only-upgrade <thepkg>`

No attachment to UA subscription is required for `ua fix` to perform its work

PR #1404, opened 2021-02-26
Install pkg fixes when they are on standard pocket

lucasmoura:
## Proposed Commit Message
Install pkg fixes when they are on standard pocket

When running ua fix on a given USN/CVE we may need may find update that are present on the Ubuntu Standard Updates.
In that scenario, we can just install those updates. We are now allowing ua fix to do exactly that.

Fixes: #1401

## Test Steps
Run the new integration test added to this PR

## Desired commit type::

## Checklist:

### DIFF
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -179,3 +179,43 @@ Feature: Command behaviour when unattached
         Examples: ubuntu release
            | release |
            | trusty  |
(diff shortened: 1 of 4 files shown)


blackboxsw (approved):
LGTM. Unrelated CI failure on 14.04.

PR #1407, opened 2021-02-26
Create config to add extra params to security url

lucasmoura:
## Proposed Commit Message
Create config to add extra params to security url

We are now creating a feature config that will allow users to pass extra params to the security url we are using on ua fix

## Test Steps
Run the modified unit tests on this PR

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -626,6 +626,19 @@ def parse_config(config_path=None):
         if key.startswith("ua_"):
             if "ua_features_" in key:
                 key = key[12:]  # String leading UA_FEATURES_
+
+                # Users can provide a yaml file to override
+                # config behavor. If they do, we are going
+                # to load that yaml and update the config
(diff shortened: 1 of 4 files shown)


blackboxsw (changes requested):
I think we need to do this differently by updating the query_params that we pass into request_url, not adding a suffix on base_url. Othwerwise, calls we emit which might contain query_params in the future will break if extra_security override is set.

lucasmoura:
@blackboxsw You are totally right on this. I completely missed the infrastructure we already have for query params. Thanks you for spotting that. I have update the code to address that

blackboxsw (approved):
Validated works as intended.

blackboxsw:
CI failures were run on a node during which it was under investigation for LXC issues. Failures were indicative of lxc init errors.

Issue #1405, opened 2021-02-26
Product and API specific constants should be configurable


blackboxsw:
Make entitlements and any API URLs configurable or overridable

    help_doc_url = "https://ubuntu.com/cc-eal"                                  
    name = "cc-eal"                                                             
    title = "CC EAL2"                                                           
    description = "Common Criteria EAL2 Provisioning Packages"                  
    repo_key_file = "ubuntu-cc-keyring.gpg"                                     
    is_beta = True

orndorffgrant:
Tracking in SC-893

orndorffgrant:
I'm not sure we want these to be configurable anymore. Some of these are translated or need to have a specific value otherwise something will break (name, repo_key_file)

For is_beta and possibly help_url, I think the preference would be to have those move to the backend resource definitions.

So in total I think we can close this? WDYT @renanrodrigo @lucasmoura @blackboxsw ?

renanrodrigo:
I'd say close this one and open a new one with the problem we want to solve: move all things we don't really need in the client to the Contracts Server somehow :D

orndorffgrant:
Added #2825 and #2826

PR #1408, opened 2021-03-01
[WIP] config: overrides for entitlement is_beta

blackboxsw:
# todo: unit test fix and integration test.
Convert is_beta class attribute to a @property method.
Allow overriding in config with
features:
    entitlements.fips.is_beta: false

Any value not null will set the is_beta value on the entitlement.

## Proposed Commit Message
Convert is_beta class attribute to a @property method.
Allow overriding in config with
features:
    entitlements:
        esm_apps: 
            is_beta: false

Any value not null will set the is_beta value on the entitlement.

## Test Steps
```bash
# See esm-apps hidden from status
PYTHONPATH=. python3 -m uaclient.cli status   | grep esm-apps 
# add esm-apps to non-beta status
echo "features:\n  entitlements:\n    esm_apps:\n      is_beta: false" >> uaclient.conf
# See esm-apps as production
PYTHONPATH=. python3 -m uaclient.cli status   | grep esm-apps 

# Note that version will still contain +entitlements which we don't want on PRO 26.2 images
PYTHONPATH=. python3 -m uaclient.cli version
20.3-360-g5ff17394 +entitlements
```

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/tests/test_cli.py
+++ b/uaclient/tests/test_cli.py
@@ -74,11 +74,11 @@ Client to manage Ubuntu Advantage services on a machine.
 
 
 @pytest.fixture(params=["direct", "--help", "ua help", "ua help --all"])
-def get_help(request, capsys):
+def get_help(request, FakeConfig, capsys):
     if request.param == "direct":
 
         def _get_help_output():
-            parser = get_parser()
+            parser = get_parser(FakeConfig())
             help_file = io.StringIO()
             parser.print_help(file=help_file)
             return (help_file.getvalue(), "base")
@@ -86,7 +86,7 @@ def get_help(request, capsys):
     elif request.param == "--help":
 
         def _get_help_output():
-            parser = get_parser()
+            parser = get_parser(FakeConfig())
             with mock.patch("sys.argv", ["ua", "--help"]):
                 with pytest.raises(SystemExit):
                     parser.parse_args()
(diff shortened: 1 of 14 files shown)


uaclient/tests/test_cli.py
```diff
@@ -74,19 +74,19 @@
...
+            parser = get_parser(FakeConfig())
             help_file = io.StringIO()
             parser.print_help(file=help_file)
             return (help_file.getvalue(), "base")
 
     elif request.param == "--help":
 
         def _get_help_output():
-            parser = get_parser()
+            parser = get_parser(mock.Mock())
```
lucasmoura, line 89: I think we can use `FakeConfig` here too

uaclient/util.py
```diff
@@ -652,6 +643,22 @@ def is_config_value_true(config: "Dict[str, Any]", path_to_value: str):
...
+    """Return the value present at a given path_to_value"""
+    value = config
+    default_value = {}  # type: Any
+    paths = path_to_value.split(".")
+    leaf_value = paths[-1]
+    for key in paths:
+        if key == leaf_value:
+            default_value = default
+
+        value = value.get(key, default_value)
```
lucasmoura, line 658: This seems to be the same code we are using on `util.is_config_value_true`. I think we can refactor to it to avoid this duplication

blackboxsw: I agree it was duplication, which is why I pulled out the common `util.get_value_from_dict` from `util.is_config_value_true`. is_config_value_true adds a bit more "flexibility" allowing people to provide either True, TRUE, true or TruE etc in environment values.   I realize that yaml.safe_load supports conversion of YES,True TRUE etc to python's `True`, but our environment variable handling doesn't. I've filed #1411 to track that separately

blackboxsw:
We are going to hold on this generic feature for overrides until we tackle https://github.com/canonical/ubuntu-advantage-client/issues/1405

Issue #1410, opened 2021-03-01
ua status hides enabled beta services from default output


blackboxsw:
Normally ua status hides any beta services from typical tabular status output. If one of those beta services are enabled, it probably should be something ua status should report about.

mitechie:
+1 that showing enabled beta services by default would be a useful UX improvement

PR #1412, opened 2021-03-01
Show beta services in status if they are enabled

lucasmoura:
## Proposed Commit Message
Show beta services in status if they are enabled

Currently, we are always hiding beta services from ua status, even if a beta service is enabled. However, we should always show
the user which services are enabled in the machine by default, being beta services or not.

Fixes: #1410

## Test Steps

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -368,7 +368,11 @@ class UAConfig:
                 released_resources.append(resource)
                 continue
 
-            if not ent_cls.is_beta:
+            enabled_status = status.UserFacingStatus.ACTIVE.value
+            if (
+                not ent_cls.is_beta
+                or resource.get("status", "") == enabled_status
+            ):
                 released_resources.append(resource)
 
         if released_resources:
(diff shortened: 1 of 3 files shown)


uaclient/config.py
```diff
@@ -368,7 +368,7 @@ def _remove_beta_resources(self, response) -> "Dict[str, Any]":
                 released_resources.append(resource)
                 continue
 
-            if not ent_cls.is_beta:
+            if not ent_cls.is_beta or resource.get("status", "") == "enabled":
```
blackboxsw, line 371: I think intead of a static string here we want to use our defined UserFacingStatus enum
```suggestion
            if not ent_cls.is_beta or resource.get("status", "") == UserFacingStatus.ACTIVE.value
```

blackboxsw (approved):
LGTM!

blackboxsw:
Unrelated CI failure filed https://github.com/canonical/ubuntu-advantage-client/issues/1415

Issue #1386, opened 2021-02-23
ua fix: console output of any `ua attach` operations performed by uaclient for auditing


blackboxsw:
To improve auditability for `ua fix` consumers, stream the ua 'attach/enable` etc commands run by `ua fix` in order to attach and enable services on a node to remedy a CVE/USN.
```bash
$ ua fix USN-4540-1
USN-4540-1: aftpd vulnerabilities
https://ubuntu.com/security/notices/USN-4540-1
1 affected package is installed: aftpd
(1/1) aftpd:
For easiest security on AWS, use Ubuntu Pro. https://ubuntu.com/aws/pro
A fix is available in UA Infra: ESM.
The update is not installed because this system is not attached to a subscription.
Choose: [S]ubscribe at ubuntu.com [A]ttach your existing token [C]ancel
> A
Enter your token (from https://ubuntu.com/advantage) to attach this system:
> YOUR-TOKEN-HERE
{ua attach output, including displaying default services}   <-------------------- this is the output this issue refers to
Checking and installing updates…
...
```

Issue #1327, opened 2021-01-25
BDD: add mechanism by which integration tests can respond to input prompts


blackboxsw:
In #1315 we added a config `feature:: block_disable_on_enable: true` due to a shortfall in our ability to instrument stdin input within integration tests.   Ultimately this config parameter shouldn't really be surfaced to the end-user.

Let's define a mechanism by which our integration tests can provide input "answers" to various CLI input prompts that are presented by UAClient.

orndorffgrant:
Closing because this is done.

PR #1414, opened 2021-03-02
Fix/prompt for token

blackboxsw:
Prompt updates for inline error messages, wire token prompt to _attach_with_token
and fix trusty installed package via dpkg-query format ${Status} option.

This PR does not handle integration testing due to #1327

## Proposed Commit Message
Rebase from existing commits

*  fix: prompt to emit pro suggestion on pro_clouds if unattached
    
    When unattached, emit ubuntu pro suggestion before prompting to
    attach.
    
    Also prompt for token if chosen and wire the response to
    _attach_with_token.
    
    Fixes: #1386

*   util: add error prompts on invalid selection
    
    Add inline error messages during reprompts which will color
    that error text red on invalid input.

*   trusty: query_installed_source_pkg_versions support
    
    dpkg-query doesn't support ${db:Status-Status}.
    Use ${Status} instead otherwise trusty will not
    list any installed packages.

## Test Steps
```bash
# test valid prompt and retries
PYTHONPATH=. python3 -c 'from uaclient.util import prompt_choices; prompt_choices("\nChoose: [S]ubscribe at ubuntu.com [A]ttach or [C]ancel", valid_choices=["s", "a", "c"])'

# to test prompts and attach and fix 
lxc launch ubuntu-daily:trusty  dev-t
lxc exec dev-t -- apt install libkrad0=1.12+dfsg-2ubuntu5.4;
lxc exec dev-t -- apt install krb5-locales=1.12+dfsg-2ubuntu5.4;   # optional 2nd binary needing upgrades
# try both on attached an detached machine
lxc exec dev-t  ua fix CVE-2020-28196 
```

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -1,12 +1,13 @@
 import os
 import socket
 
+from uaclient import apt
 from uaclient.config import UAConfig
 from uaclient import exceptions
 from uaclient import status
 from uaclient import serviceclient
 from uaclient import util
-from uaclient import apt
+
 
 CVE_OR_USN_REGEX = (
     r"((CVE|cve)-\d{4}-\d{4,7}$|(USN|usn|LSN|lsn)-\d{1,5}-\d{1,2}$)"
@@ -310,6 +311,25 @@ class USN:
             self._cves.append(self.client.get_cve(cve_id=cve_id))
         return self._cves
 
+    def released_packages(self, series: str) -> "Dict[str, Dict[str, str]]":
+        """Return a dict of released binary packages keyed on source pkg name.
+
+        The value is a dict with binary package name as key and pkg version.
+        """
+        binary_pkgs = {}  # type: Dict[str, str]
+        src_pkg = {}
+        for released_pkg in self.response["release_packages"].get(series, []):
+            is_source = released_pkg.get("is_source")
+            if is_source is True:
+                src_pkg = released_pkg
+            elif is_source is False:
+                binary_pkgs[released_pkg["name"]] = released_pkg["version"]
+        if not all([src_pkg, binary_pkgs]):
+            raise RuntimeError(
+                "Unable to parse USN response for {}".format(self.id)
+            )
+        return {src_pkg["name"]: binary_pkgs}
+
     def get_url_header(self):
         """Return a string representing the URL for this notice."""
         return status.MESSAGE_SECURITY_URL.format(
@@ -319,28 +339,36 @@ class USN:
         )
 
 
-def query_installed_source_pkg_versions() -> "Dict[str, str]":
+def query_installed_source_pkg_versions() -> "Dict[str, Dict[str, str]]":
     """Return a dict of all source packages installed on the system.
 
-    The dict keys will be source package name: "krb5". The value will be the
-    package version.
+    The dict keys will be source package name: "krb5". The value will list of
+    dicts with keys binary_pkg and version.
     """
+    series = util.get_platform_info()["series"]
+    if series == "trusty":
+        status_field = "${Status}"
+    else:
+        status_field = "${db:Status-Status}"
     out, _err = util.subp(
         [
             "dpkg-query",
-            "-f=${Package},${Source},${Version},${db:Status-Status}\n",
+            "-f=${Package},${Source},${Version}," + status_field + "\n",
             "-W",
         ]
     )
-    installed_packages = {}  # type: Dict[str, str]
+    installed_packages = {}  # type: Dict[str, Dict[str, str]]
     for pkg_line in out.splitlines():
         pkg_name, source_pkg_name, pkg_version, status = pkg_line.split(",")
         if not source_pkg_name:
             # some package don't define the Source
             source_pkg_name = pkg_name
-        if status != "installed":
+        if "installed" not in status:
             continue
-        installed_packages[source_pkg_name] = pkg_version
+        if source_pkg_name in installed_packages:
+            installed_packages[source_pkg_name][pkg_name] = pkg_version
+        else:
+            installed_packages[source_pkg_name] = {pkg_name: pkg_version}
     return installed_packages
 
 
@@ -353,7 +381,7 @@ def fix_security_issue_id(cfg: UAConfig, issue_id: str) -> None:
             cve = client.get_cve(cve_id=issue_id)
         except SecurityAPIError as e:
             raise exceptions.UserFacingError(str(e))
-        affected_pkg_status = get_cve_affected_packages_status(
+        affected_pkg_status = get_cve_affected_source_packages_status(
             cve=cve, installed_packages=installed_packages
         )
         print(cve.get_url_header())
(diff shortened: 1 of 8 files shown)


uaclient/security.py
```diff
@@ -497,37 +522,44 @@ def upgrade_packages_and_attach(
...
+            if choice == "a":
+                print(status.PROMPT_ENTER_TOKEN)
+                token = input("> ")
+                print(
+                    "{color}{{ua attach {token}}}{end}".format(
+                        color=status.TxtColor.DISABLEGREY,
+                        token=token,
+                        end=status.TxtColor.ENDC,
+                    )
+                )
```
lucasmoura, line 558: I think we should also print the attach command here, right ?

uaclient/security.py
```diff
@@ -497,37 +522,44 @@ def upgrade_packages_and_attach(
...
+                cli._attach_with_token(cfg, token, allow_enable=True)
```
lucasmoura, line 559: I think we should run the `_verbose_apt_update_install` function after attaching

lucasmoura (commented):
That looks really good @blackboxsw

But just for testing purposes, do you any CVE or USN that we use that need an uaclient subscription to install the packages.

Also, I know this is not ideal. But maybe we could add a feature config to automatically answer the prompt. I know this is bad, polluting feature config with attributes that should only be used on development, but for testing purposes, it may be worth it.

Another approach we can follow is to add an env variable instead, but it is still not an optimal solution

blackboxsw:
CI unrelated build failure:
`      paramiko.ssh_exception.AuthenticationException: Authentication failed.`
https://jenkins.canonical.com/server-team-jenkins-private/blue/rest/organizations/jenkins/pipelines/github-ci/pipelines/ubuntu-advantage-client/branches/PR-1414/runs/6/nodes/128/steps/153/log/?start=0

uaclient/security.py
```diff
@@ -465,69 +498,109 @@ def prompt_for_affected_packages(
...
-                "Choose: [S]ubscribe at ubuntu.com [A]ttach"
-                " existing token [C]ancel",
-                valid_choices=["s", "a", "c"],
-            )
-            print("TODO react to subscription choice:{} GH: #".format(choice))
+        if not cfg.is_attached:
+            if not _prompt_for_attach(cfg):
+                return False  # User opted to cancel
+    packages = sorted(upgrade_packages + upgrade_packages_ua)
+    print(
```
lucasmoura, line 585: We are printing outside `run_apt_command` just to make use of the colorize parts, right ?

uaclient/security.py
```diff
@@ -465,69 +498,109 @@ def prompt_for_affected_packages(
...
+        )
+    )
+    apt.run_apt_command(
+        cmd=["apt-get", "update"], error_msg=status.MESSAGE_APT_UPDATE_FAILED
+    )
+    apt.run_apt_command(
+        cmd=["apt-get", "install", "--only-upgrade", "-y"] + packages,
+        error_msg=status.MESSAGE_APT_INSTALL_FAILED,
+        env={"DEBIAN_FRONTEND": "noninteractive"},
+        print_cmd=True,
```
lucasmoura, line 601: Since we are already printing the commands, we should drop the `print_cmd`  part from this function

uaclient/security.py
```diff
@@ -310,6 +316,25 @@ def get_cves_metadata(self):
             self._cves.append(self.client.get_cve(cve_id=cve_id))
         return self._cves
 
+    def released_packages(self, series: str) -> "Dict[str, Dict[str, str]]":
```
lucasmoura, line 319: I may be missing something here, but this method seem to be not called anywhere. Is that expected ?

uaclient/status.py
```diff
@@ -134,9 +134,13 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_SECURITY_UPDATE_NOT_INSTALLED = "The update is not yet installed."
 MESSAGE_SECURITY_UPDATE_NOT_INSTALLED_SUBSCRIPTION = """\
 The update is not installed because this system is not attached to a
-subscription that covers these packages.
+subscription.
 """
 MESSAGE_SECURITY_UPDATE_INSTALLED = "The update is already installed."
+MESSAGE_SECURITY_USE_PRO_TMPL = (
+    "For easiest security on {title}, use Ubuntu PRO."
+    " https://ubuntu.com/{cloud}/pro"
```
lucasmoura, line 142: Should we add a "," to the end of this message ?

uaclient/util.py
```diff
@@ -410,10 +410,16 @@ def prompt_choices(msg: str = "", valid_choices: "List[str]" = []) -> str:
...
-    while value not in valid_choices:
-        value = input(msg).lower().strip()
+    while True:
+        print(msg)
+        value = input("> ").lower()
+        if value in valid_choices:
+            break
+        print(
+            "{} is not one of: {}".format(
+                value, ", ".join([choice.upper() for choice in valid_choices])
```
lucasmoura, line 420: Even though we only have three choices, maybe it is better to create that list outside the while loop to avoid creating one each time the user types something wrong

lucasmoura:
@blackboxsw It seems that the integration tests are also failing for the fix command now. I think this is because of the color mechanism we are applying now

uaclient/security.py
```diff
@@ -465,69 +498,109 @@ def prompt_for_affected_packages(
...
-            print_cmd=True,
-        )
+    :return: True if package upgrade completed or unneeded, False otherwise.
+    """
+    if not any([upgrade_packages, upgrade_packages_ua]):
+        return True
+    if os.getuid() != 0:
+        print(status.MESSAGE_SECURITY_APT_NON_ROOT)
+        return False
     if upgrade_packages_ua:
```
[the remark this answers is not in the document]
blackboxsw: Per @mitechie's thoughts, since we can't get to a full fix, we don't want the complexity of a half-fixed solution here at the moment.  I filed https://github.com/canonical/ubuntu-advantage-client/issues/1423 so we can track this separately.

uaclient/security.py
blackboxsw: @lucasmoura I was wondering about dropping the apt print_cmd parameter in general because we want the same type of decoration here that we have for the ua attach token output.

Per the spec, the output of the messages should default to recommended human-centered calls to `apt update` and `apt install` instead of the machine-focused calls `apt-get update` and `apt-get install` which we have to call through python code because we also provide non-interactive environment variables.

Because these commands and the printed suggestions of text are not exactly the same as what ua client is calling, I think this precludes us from printing verbatim inside the run_apt_commands the strict command being run.

uaclient/security.py
blackboxsw: I would vote +1 to drop per my above response but I wanted to check with you today for your concerns here https://github.com/canonical/ubuntu-advantage-client/pull/1414#discussion_r586620039

uaclient/security.py
lucasmoura: Agreed. since we are doing that distinction, I don't see any need to support the `print_cmd` anymore

uaclient/security.py
blackboxsw: ahh left over from an earlier iteration on this PR, I'll drop it now. I had thought we needed to read USN data for specific binary package versions instead of source package,  but instead we collect all binary pkgs now in `query_installed_source_pkg_versions` and we process those installed binary pkgs one by one in `prompt_for_affected_packages` without need to extract specific versions from the USN.

uaclient/util.py
blackboxsw: +1 done.

lucasmoura (approved):

PR #1416, opened 2021-03-02
Delete lib/__init__.py

paride:
## Proposed Commit Message

Unchanged (rebase and merge).

## Test Steps
Note that the `ubuntu-advantage-tools` package currently ships the following unwanted files:

```
$ dpkg -L ubuntu-advantage-tools | grep dist-packages/lib
/usr/lib/python3/dist-packages/lib
/usr/lib/python3/dist-packages/lib/__init__.py
/usr/lib/python3/dist-packages/lib/reboot_cmds.py
/usr/lib/python3/dist-packages/lib/upgrade_lts_contract.py
```
This is because setuptools treats `lib/` as a package to install. Let's exclude it.

The two commands will still be installed in `/usr/lib/ubuntu-advantage/`, which is the desired location.

## Desired commit type::

## Checklist:

Not really relevant but:


paride:
CI fail due to:
```
botocore.exceptions.ClientError: An error occurred (InsufficientInstanceCapacity) when calling the RunInstances operation (reached max retries: 4): We currently do not have sufficient t2.micro capacity in the Availability Zone you requested (us-east-2b). Our system will be working on provisioning additional capacity. You can currently get t2.micro capacity by not specifying an Availability Zone in your request or choosing us-east-2a, us-east-2c.
```

paride:
TIL that setuptools uses the presence of `__init__.py` to decide if a directory is a Python package. Let's drop `lib/__init__.py` as `lib/` isn't a package.

blackboxsw (approved):
LGTM thanks @paride. Yes confirmed that dropping this avoids the setuptools magic and retains our desired deliverables reboot_cmds.py etc in the proper location.

Issue #1420, opened 2021-03-02
cve: when unable to find a given CVE, improve error message:


blackboxsw:
When an entered CVE  is not found. handle 404 messaging with a nice message instead of the following
```bash
NOT FOUND: [https://ubuntu.com/security/cves/CVE-2020-13757.json] CVE with id 'CVE-2020-137579' does not exist
```

PR #1434, opened 2021-03-04
Improve error message for ua fix

lucasmoura:
## Proposed Commit Message
Improve error message for ua fix

When we do not found a CVE or USN, we display an error that
is not user friendly. We are now making that message better for the
user.

Fixes: #1420

## Test Steps

## Desired commit type::

## Checklist:


blackboxsw (approved):
LGTM

Issue #1438, opened 2021-03-08
prioritize update and prompts for -updates -security packages prior to token attach


blackboxsw:
1. Upfront say affected packages that aren't resolvable, an whether fix is in ESM* with token attach
    Unfixable
    Standard updates
     UA update
2. Prompt "do you want to upgrade these package (Y/n)"
3. Standard -updates first with apt prompt
4. If unattached, prompt for tokens
5. New message for resolved as much as we can (if unable to fix needs-triage type packages)

mitechie:
I don't get your output Lucas. 2/2 should begin walking through attaching the token and enabling the service vs saying "not installed because of things..." shouldn't it?

lucasmoura:
Oh sorry, I just realized that I butchered that comment :facepalm: Let me update with the correct message

lucasmoura:
@mitechie @blackboxsw I have updated the comment to something that makes sense now

mitechie:
```
2 affected packages are installed: curl, slsrc
(1/2) curl:
A fix is available in Ubuntu standard updates.
The update is not yet installed.

I don't think we need this line above at all in the non-esm cases. It's only there in the spec to help indicate that the package isn't/cannot be installed until the subscription is at play. 


Installing packages from standard updates: curl
{ apt update && apt install --only-upgrade -y curl }
(2/2) slsrc:
A fix is available in UA Infra.
The update is not yet installed.

Again, in this case, an already attached machine, I don't think we need to say it's not already installed. In fact, if was already installed we wouldn't show it as an affected package on the system correct? 


Installing packages from standard updates: curl
{ apt update && apt install --only-upgrade -y curl }
Installing packages that requires a ubuntu-advantage-tools subscription: slsrc
The update is not installed because this system is not attached to a subscription.
```

And here I see that I misunderstood where it was going. I think this is too much noise at this point. We're knowingly failing at installing a package in order to get to the error that this needs a subscription?

lucasmoura:
```
Again, in this case, an already attached machine, I don't think we need to say it's not already installed. In fact, if was already installed we wouldn't show it as an affected package on the system correct?
```

Not entirely, we would still list that package during `ua fix`, but instead of `The update is not yet installed.` we would have `The update is already installed.`

```
We're knowingly failing at installing a package in order to get to the error that this needs a subscription?
```

[...]

Issue #1376, opened 2021-02-17
Security API: pocket and pocket_source fields of CVE package status do not represent esm-apps or esm-infra pockets


blackboxsw:
CVE responses from the security API do not report pockets that are indicative of esm-apps or esm-infra. UA client currently relies on checking "esm" in description to determine if a pkg is resident in an "esm-infra" pocket. This is waiting on Security team to fill in appropriate pocket details in the Security API db.

```json
        {                                                                       
          "component": null,                                                    
          "description": "2:4.3.11+dfsg-0ubuntu0.14.04.20+esm9",                
          "pocket": null,                                                       
          "release_codename": "trusty",                                         
          "status": "released"                                                  
        }, 
```

orndorffgrant:
@blackboxsw We're closing this as we believe it is already fixed. Please re-open if we're mistaken.

PR #1454, opened 2021-03-11
Update output for released packages in ua fix

lucasmoura:
## Proposed Commit Message
Update output for released packages in ua fix

Currently, we list every released package in the output of ua fix and only after that we start installing them. This is not optimal since depending on the source of the package we need to perform different actions to install those packages. Therefore, we are now grouping the released packages by pocket origin and acting on them as soon as we can.

Fixes: #1438

## Test Steps
This branch is not yet complete. I still need to add more unit tests and integration tests to it.
However, it is possible to verify the current expected behavior by running the modified unit tests

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -17,6 +17,7 @@ from uaclient import exceptions
 from uaclient import status
 from uaclient import serviceclient
 from uaclient import util
+from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
 
 try:
     from typing import Any, Dict, List, Optional, Tuple  # noqa: F401
@@ -680,8 +681,8 @@ def _handle_released_package_fixes(
     if src_pocket_pkgs:
         for pocket in [
             UBUNTU_STANDARD_UPDATES_POCKET,
-            UA_INFRA_POCKET,
             UA_APPS_POCKET,
+            UA_INFRA_POCKET,
         ]:
             pkg_src_group = src_pocket_pkgs[pocket]
             binary_pkgs = binary_pocket_pkgs[pocket]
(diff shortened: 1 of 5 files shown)


uaclient/security.py
```diff
@@ -753,45 +817,74 @@ def _prompt_for_attach(cfg: UAConfig) -> bool:
     return True
 
 
+def _check_subscription_for_required_service(
+    cfg: UAConfig, pocket: str
+) -> bool:
+    """Verify if the ua subscription has the required service enabled."""
+    if pocket == UA_INFRA_POCKET:
+        service_to_check = "esm-infra"
+    else:
```
blackboxsw, line 826: We probably want to specificatlly check elif pocket == UA_APPS_POCKET   and fall through to an else: for unknown pocket sources. It's likely that we have some pocket metadata gaps that could return None or unexpected pockets, and we want to warn in that case and return False or raise a  exceptions.SecurityAPIMetadataError.

uaclient/security.py
```diff
@@ -753,45 +817,74 @@ def _prompt_for_attach(cfg: UAConfig) -> bool:
...
+
+    if ent_cls:
+        ent = ent_cls(cfg)
+        ent_status, _ = ent.user_facing_status()
+
+        if ent_status == status.UserFacingStatus.ACTIVE:
+            return True
+
+        print(
+            status.MESSAGE_SECURITY_UA_SERVICE_NOT_ENABLED.format(
```
blackboxsw, line 839: Probably add a TODO comment for a future PR
```suggestion
            # TODO(GH: #1455 prompt when service not enabled)
            status.MESSAGE_SECURITY_UA_SERVICE_NOT_ENABLED.format(
```

uaclient/security.py
```diff
@@ -635,6 +642,69 @@ def group_by_usn_package_status(affected_pkg_status, usn_released_pkgs):
     return status_groups
 
 
+def _format_packages_message(
```
blackboxsw, line 645: Good cleanup and separation of functions

blackboxsw (commented):
Looks really good, forgot to hit submit on the review.

uaclient/security.py
lucasmoura: @blackboxsw I don't think we can reach a scenario like this, because of the way we retrieve the pocket information
from the `CVEPackageStatus` object. We use the `pocket_source` property:

https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/security.py#L224

You can see that there is currently no way for us to return something that is not one of:

* Ubuntu Standard Updates
* ESM Infra
* ESM Apps

However, it seems that the security api is returning the pocket information already. Should we update that property as well ?

blackboxsw: @lucasmoura I was mostly concerned with the fact that we prefer USN.release_packages.*pocket values if present before falling back to CVE.packages.statuses.*pocket values. So if USN has a unexpected value in pocket, we wouldn't handle that case here. But, you corrected me that we still call[ CVEPackageStatus.pocket_source](https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/security.py#L232)  which will properly put things into one of our known expected pockets in the "else" clause. We are still exposed to our potentially incorrect behavior of assuming that an unknown pocket means either "UA Infra" or "Ubuntu Standard Updates", but I think we can handle that in GH: #1376.

blackboxsw (approved):
LGTM. Current CI failures were load related on this last run. We are sorting the load-related known failures separately.

Issue #1455, opened 2021-03-12
fix: validate and prompt if esm-infra/apps is disabled


blackboxsw:
ua fix checks whether a token is attached and prompts for such attach.

If a fix is required from esm-apps and that service is not enabled on an attached machine. UA client should prompt for that enablement.

Example interaction
```bash
root@dev-t:~# ua fix CVE-2020-28196
CVE-2020-28196: Kerberos vulnerability
https://ubuntu.com/security/CVE-2020-28196
Related USNs: USN-4635-1
1 affected package is installed: krb5
(1/1) krb5:
A fix is available in UA Infra.
The update is not yet installed.

The esm-infra service is disabled. Choose: [E]nable to install updates [C]ancel
> E
{ ua enable esm-infra }
{ apt update && apt install --only-upgrade -y krb5-locales }
✔ CVE-2020-28196 is resolved.
```

mitechie:
We should follow the same process and output as if they were not attached and just pick up from there since the person is attached. I suppose if the user is in this specific situation we know they disabled the service since it would default to enabled otherwise. We should do a check that they're entitled to this function (esm or esm apps) and be accurate in the output we give them. 

Using the spec as an example

```
A fix is available in UA Apps: ESM.
The update is not installed because this system does not have UA Apps: ESM enabled. 
Choose: [E]nable UA Apps: ESM [C]ancel
> E
<ua enable esm-apps>
<output from the above...>
{sudo apt update && apt sudo apt-get install --only-upgrade aftpd}
Other updates are pending. To install them all: sudo apt upgrade
✔ USN-4540-1 is addressed.
```

PR #1471, opened 2021-03-22
Prompt to enable service during ua fix

lucasmoura:
## Proposed Commit Message
Prompt to enable service during ua fix

Currently, when we are running ua fix and a necessary ua service is not enabled, we abort the operation and report that the CVE/USN was not solved. However, if the user is entitled to that service through uaclient, we can prompt the user to enable that service during the fix operation. We are now adding this functionality to the ua fix command.

Fixes: #1455

## Test Steps
Run the modified integration test in this PR

## Desired commit type::

## Checklist:

### DIFF
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -276,6 +276,31 @@ Feature: Command behaviour when unattached
             .*\{ apt update && apt install --only-upgrade -y screen \}.*
             .*✔.* USN-4747-2 is resolved.
             """
+        When I run `apt-get install -y screen=4.1.0~20120320gitdb59704-9 --force-yes` with sudo
+        And I run `ua disable esm-infra` with sudo
+        And I fix `USN-4747-2` by enabling required service
+        Then stdout matches regexp:
+            """
+            USN-4747-2: GNU Screen vulnerability
+            Found CVEs: CVE-2021-26937
+            https://ubuntu.com/security/CVE-2021-26937
+            1 affected package is installed: screen
+            \(1/1\) screen:
+            A fix is available in UA Infra.
+            The update is not installed because this system does not have
+            esm-infra enabled.
+
+            Choose: \[E\]nable esm-infra \[C\]ancel
+            > .*\{ ua enable esm-infra \}.*
+            One moment, checking your subscription first
+            Updating package lists
+            ESM Infra enabled
+            """
+        And stdout matches regexp:
+            """
+            .*\{ apt update && apt install --only-upgrade -y screen \}.*
+            .*✔.* USN-4747-2 is resolved.
+            """
 
         Examples: ubuntu release
            | release |
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -845,6 +845,35 @@ def _prompt_for_attach(cfg: UAConfig) -> bool:
     return True
 
 
+def _prompt_for_enable(cfg: UAConfig, service: str) -> bool:
+    """Prompt for enable a ua service.
+
+    :return: True if enable performed.
+    """
+    import argparse
+    from uaclient import cli
+
+    print(status.MESSAGE_SECURITY_SERVICE_DISABLED.format(service=service))
+    choice = util.prompt_choices(
+        "Choose: [E]nable {} [C]ancel".format(service),
+        valid_choices=["e", "c"],
+    )
+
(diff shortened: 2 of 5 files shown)


uaclient/security.py
```diff
@@ -822,6 +822,37 @@ def _prompt_for_attach(cfg: UAConfig) -> bool:
...
+
+    print(status.MESSAGE_SECURITY_SERVICE_DISABLED.format(service=service))
+    choice = util.prompt_choices(
+        "Choose: [E]nable {} [C]ancel".format(service),
+        valid_choices=["e", "c"],
+    )
+
+    if choice == "c":
+        return False
+    elif choice == "e":
```
blackboxsw, line 841: Could consolidate this section logically dropping the if choice == "c" because we fall through and return False at end
```suggestion
    if choice == "e":
```

uaclient/security.py
```diff
@@ -822,6 +822,37 @@ def _prompt_for_attach(cfg: UAConfig) -> bool:
...
+        print(status.colorize_commands([["ua", "enable", service]]))
+        return bool(
+            0
+            == cli.action_enable(
+                argparse.Namespace(
+                    service=[service], assume_yes=True, beta=True
```
blackboxsw, line 847: I'm not sure about providing a direct beta=True here, should we rely on config.features.allow_beta here?

uaclient/security.py
```diff
@@ -840,13 +871,24 @@ def _check_subscription_for_required_service(
...
+        if applicability_status == status.ApplicabilityStatus.APPLICABLE:
+            if _prompt_for_enable(cfg, service_to_check):
+                return True
+            else:
+                print(
+                    status.MESSAGE_SECURITY_UA_SERVICE_NOT_ENABLED.format(
+                        service=ent.name
+                    )
+                )
+                return False
```
blackboxsw, line 884: Can drop this return False as we do it below

uaclient/security.py
```diff
@@ -840,13 +871,24 @@ def _check_subscription_for_required_service(
...
+        else:
+            print(
+                status.MESSAGE_SECURITY_UA_SERVICE_NOT_ENTITLED.format(
+                    service=ent.name
+                )
             )
-        )
-        return False
+            return False
```
blackboxsw, line 891: No need to return False here as we do it below

blackboxsw (changes requested):
Looks really good. Couple of questions and nits.

uaclient/security.py
lucasmoura: My concern here is just that we might need to do that for `esm-apps`, which is still a beta service. That is why I have used `True` here. But I think you are right, it is better to rely on this configuration

lucasmoura:
@blackboxsw code updated

blackboxsw (approved):
behavior looks great. Thank you, validated integration tests locally as we have a temporary Jenkins problem.
Let's separately add logging for any config operations performed by `ua fix`. I think we don't log any colorized command output since we short-circuit into cli.py functions for `ua attach X` or `ua enable y`  
- https://github.com/canonical/ubuntu-advantage-client/issues/1477

features/unattached_commands.feature
```diff
@@ -276,6 +276,31 @@ Feature: Command behaviour when unattached
...
+            https://ubuntu.com/security/CVE-2021-26937
+            1 affected package is installed: screen
+            \(1/1\) screen:
+            A fix is available in UA Infra.
+            The update is not installed because this system does not have
+            esm-infra enabled.
+
+            Choose: \[E\]nable esm-infra \[C\]ancel
+            > .*\{ ua enable esm-infra \}.*
+            One moment, checking your subscription first
```
mitechie, line 295: Do we need to check their subscription first? 

If they're not attached we'd show the enter your token attach window

if they're not entitled, we'd show they're not entitled and error

This line feels false?

lucasmoura: When we are running any action inside `ua fix` (`enable`, `attach`, ...) we are actually running then just like an user would. The line you are seeing is the result of running `ua enable`, which always checks the subscription first before enabling a service. With that said, I do agree that this line is not that useful here. I will see if I can easily remove it

mitechie: Ok, thanks appreciate it.

Issue #1470, opened 2021-03-22
fix: API calls to notices.json?details=CVE-2018-1088 is a simple match that includes unrelated CVES


blackboxsw:
Emilia @metorino found the case that pattern matching in Security API isn't exact when it comes to CVE id matching.

UA client calls `security/notices.json?details=CVE-2018-1088` looking specifically for only USNs related to CVE-2018-1088. But the detail matching is a simple string substring match  instead of a specific CVE id. 
Note that the "matched" related CVEs on the USN returned are only the prefix `CVE-2018-1088`
```json
      "cves": [
        "CVE-2018-10876",
        "CVE-2018-10877",
        "CVE-2018-10878",
        "CVE-2018-10879",
        "CVE-2018-10880",
        "CVE-2018-10882",
        "CVE-2018-10883",
    ]
```

blackboxsw:
Without this UA client attempts to fix many completely unrelated packages for a given CVE.

mitechie:
Is this the APIs issue or the ua client's one though?

lucasmoura:
It seems that this an API issue. We can reproduce that problem by using their endpoint directly. I will see if there is another query param we can use to get the expected output here

blackboxsw:
@albertkol is this a feature request we can add to the security API or should we not expect a specific matching query for CVE ids

blackboxsw:
For the moment, ua client can filter out USN responses which contain non-specific CVE matches in USN["cves"] list. And we can talk with Albert about whether a feature-request of the API is achievable.

PR #1478, opened 2021-03-23
fix: filter usns by cve details

lucasmoura:
## Proposed Commit Message
fix: filter usns by cve details

When we are querying for related USNs to fix, we use find all related USNs to a given CVE. However, the query that we are using to get the related USNs is using applying the CVE match as a substring instead of an exact match. This means that we will be return USNs that are not related to our query CVE, but have are fixing a CVE which id is a substring of our original CVE. We are filtering the returned USNs to only those that actually contain the exact CVE id in them.

Fixes: #1470

## Test Steps
Run the new unittest added

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -162,6 +162,7 @@ class UASecurityClient(serviceclient.UAServiceClient):
             [
                 USN(client=self, response=usn_md)
                 for usn_md in usns_response.get("notices", [])
+                if details is None or details in usn_md.get("cves", [])
             ],
             key=lambda x: x.id,
         )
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
Thanks for this correct fix we can revisit if Security API adds specific CVE id searches.

I can confirm failure to ignore unrelated USNs and the fix working:
```bash
# before
CVE-2018-1088: GlusterFS vulnerabilities
https://ubuntu.com/security/CVE-2018-1088
Error: USN-3752-3 metadata does not define release_packages source_link for linux-image-gke.
✘ USN-3752-3 is not resolved.

# after other unrelated USNs like USN-3752-3 are properly ignored.
CVE-2018-1088: GlusterFS vulnerabilities
https://ubuntu.com/security/CVE-2018-1088
No affected packages are installed.
✔ CVE-2018-1088 does not affect your system.
```

Issue #1475, opened 2021-03-23
ua fix prompt when subscription is expired


lucasmoura:
When running `ua fix` we may find that the user subscription is expired. If we identify that situation, we should prompt the user
to renew the subscription like this:

```
ua fix USN-4540-1
USN-4540-1: aftpd vulnerabilities
https://ubuntu.com/security/notices/USN-4540-1
1 affected package is installed: aftpd
(1/1) aftpd:
A fix is available in UA Apps: ESM.
The update is not installed because your subscription has expired.
Choose: [R]enew your subscription (at https://ubuntu.com/advantage) [C]ancel
> R
Enter your new token to attach this system:
> token
{ ua detach }
{ ua attach token }
{ sudo apt update && apt sudo apt-get install --only-upgrade aftpd }
✔ USN-4540-1 is resolved.
```

Issue #1477, opened 2021-03-23
fix: log commands run on the user's behalf


blackboxsw:
right now fix short-circuits the uaclient/cli.py logic which logs the user-provided commands on the commandline. When 'ua attach <token>' or 'ua enable <service>' is run. Make sure we are logging all configuration changes performed on the system.

orndorffgrant:
Tracking in SC-822

orndorffgrant:
This is fixed in the current version.

PR #1480, opened 2021-03-24
fix: prompt for new token when expired one is used

lucasmoura:
## Proposed Commit Message
fix: prompt for new token when expired one is used

When running ua fix we may need to install packages that depend on a UA service. Even if the user is already attached
to a UA subscription, this may not be enough, since the token used may be expired. If we encounter that situation, we would
need to ask the user to get a new token before proceeding with the package fixes. We are now adding the necessary code to allow us to handle that scenario.

Fixes: #1475

## Test Steps
Run the new unittests for this feature

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -807,14 +807,8 @@ def prompt_for_affected_packages(
         )
 
 
-def _prompt_for_attach(cfg: UAConfig) -> bool:
-    """Prompt for attach to a subscription or token.
-
-    :return: True if attach performed.
-    """
-    import argparse
-    from uaclient import cli
-
+def _inform_ubuntu_pro_existence_if_applicable() -> None:
+    """Alert the user when running UA on cloud with PRO support."""
     cloud_type = get_cloud_type()
     if cloud_type in PRO_CLOUDS:
         print(
@@ -822,6 +816,31 @@ def _prompt_for_attach(cfg: UAConfig) -> bool:
                 title=CLOUD_TYPE_TO_TITLE.get(cloud_type), cloud=cloud_type
             )
         )
+
+
+def _run_ua_attach(cfg: UAConfig, token: str) -> bool:
+    """Attach to a UA subscription with a given token.
+
+    :return: True if attach performed without errors.
+    """
+    import argparse
+    from uaclient import cli
+
+    print(status.colorize_commands([["ua", "attach", token]]))
+    return bool(
+        0
+        == cli.action_attach(
+            argparse.Namespace(token=token, auto_enable=True), cfg
+        )
+    )
+
+
+def _prompt_for_attach(cfg: UAConfig) -> bool:
+    """Prompt for attach to a subscription or token.
+
+    :return: True if attach performed.
+    """
+    _inform_ubuntu_pro_existence_if_applicable()
     print(status.MESSAGE_SECURITY_UPDATE_NOT_INSTALLED_SUBSCRIPTION)
     choice = util.prompt_choices(
         "Choose: [S]ubscribe at ubuntu.com [A]ttach existing token [C]ancel",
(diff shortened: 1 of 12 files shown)


uaclient/security.py
```diff
@@ -913,6 +918,52 @@ def _check_subscription_for_required_service(
...
+            )
+        )
+
+    return False
+
+
+def _check_subscription_is_expired(cfg: UAConfig) -> bool:
+    from uaclient import cli
+
+    try:
```
blackboxsw, line 956: Can we emit colorize commands for `ua refresh` here to represent the changes being performed on the system.
I filed a separate issue to make sure we do this in general for ua fix work https://github.com/canonical/ubuntu-advantage-client/issues/1477

blackboxsw (commented):
Thanks Lucas, couple thoughts here on first pass.

uaclient/security.py
lucasmoura: I though about doing that, however, I don't think we should surface that to the user. We are only running that command to check if the subscription is not expired, meaning that most of the times, it will probably not change the system at all.

But you think we can still output it, no problem

lucasmoura:
@blackboxsw code updated

uaclient/security.py
blackboxsw: I'm on the fence, but not fully convinced we need it, ua refresh does make a change if there are contract changes in general like changes in aptURL, packages required etc.  The only exposure I'd think about currently  is Ubuntu PRO images, once we release 26.2 to AWS/GCP/Azure, we will lock-step update contracts changing additionalPackages directives. I guess if a PRO image updated to 27.0 we might have a case where running a ua fix before reboot would attempt to apply those contract deltas and break on apt installs of ubuntu-fips metapkg due to current deb pinning. I suppose this is a small  corner-case and maybe negligible here.

uaclient/security.py
```diff
@@ -913,6 +919,53 @@ def _check_subscription_for_required_service(
...
+            == cli.action_attach(
+                argparse.Namespace(token=token, auto_enable=True), cfg
+            )
+        )
+
+    return False
+
+
+def _check_subscription_is_expired(cfg: UAConfig) -> bool:
+    from uaclient import cli
```
blackboxsw, line 956: I have a pedantic request fo add a docstring here representing the ':return' intent for True vs False.

uaclient/security.py
```diff
@@ -913,6 +919,53 @@ def _check_subscription_for_required_service(
     return False
 
 
+def _prompt_for_new_token(cfg: UAConfig) -> bool:
```
blackboxsw, line 922: What do you think: Is it worth sharing logic between _prompt_for_new_token and _prompt_for_attach?

lucasmoura: I think the only part of the code we could reuse is this one:

```
print(status.colorize_commands([["ua", "attach", token]]))
return bool(
    0    
    == cli.action_attach(
        argparse.Namespace(token=token, auto_enable=True), cfg
    )    
)  
```

I think we can move it to a separate function. Let me do that

blackboxsw (changes requested):
Some valid integration test failures:

Some points that @mitechie should comment on:
 - I recently got guidance on some of the CVE messaging that if it's not a complete sentence, no need to follow with a period "." 
- [ ] I believe "See: ua status should not have a trailing stop because we expect possibly folks to cut-n-paste that command and the highlight might bring in the "."
- [ ] I believe "See: https:/ubuntu.com/advantage" doesn't need a trailing stop "." because it's not a sentence but a directive

Here's a diff of what I changed, dropping trailing stops from "See:  blah" and correcting a case where we also added a leading "." accidentally

diff: https://paste.ubuntu.com/p/vrWWWP9pSG/

success run: https://pastebin.canonical.com/p/SCDK47S37W/

uaclient/security.py
lucasmoura: I think this is good point. Maybe we could have another way to check if the subscription is expired or not instead of running refresh. We could see if the contracts have an endpoint that allows us to do just that.

blackboxsw (approved):
Changes look good. +1 just waiting on CI before a merge

blackboxsw:
CI known intermittent lxc failure that needs slightly better erroring from pycloudlib
````bash
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1480@3/.tox/behave-lxd-14.04/lib/python3.6/site-packages/pycloudlib/lxd/instance.py", line 88, in ip
      ip_address = result.split()[0]
  IndexError: list index out of range
```


Already a pycloudlib issue for this  https://github.com/canonical/pycloudlib/issues/129

Issue #1448, opened 2021-03-10
when a CVE is not addressed at the end of UA Fix, details are lacking


mitechie:
CVE-2017-9233: Expat vulnerability
    https://ubuntu.com/security/CVE-2017-9233
    3 affected packages are installed: expat, matanza, swish-e
    \(1/3\) matanza:
    Ubuntu security engineers are investigating this issue.
    \(2/3\) swish-e:
    Ubuntu security engineers are investigating this issue.
    \(3/3\) expat:
    A fix is available in Ubuntu standard updates.
    The update is not yet installed.
    .*\{ apt update && apt install --only-upgrade 
    .*✘.* CVE-2017-9233 is not resolved.

I think it's worth highlighting the packages responsible for not letting us resolve this. Possibly a summary line:

    2 packages are still affected: matanza, swish-e
    .*✘.* CVE-2017-9233 is not resolved.

What do you think @lechsandecki ?

This would mostly be only for cases where there are multiple packages and some of them might not be addressable. I think it'll be fairly rare.

lechsandecki:
@mitechie yes, I think it's reasonable. I don't know if we need to give a status for each package. Like "CVE XXX is not resolved, Ubuntu security engineers are investigating this issue." might be enough. Otherwise, it looks very busy

mitechie:
> @mitechie yes, I think it's reasonable. I don't know if we need to give a status for each package. Like "CVE XXX is not resolved, Ubuntu security engineers are investigating this issue." might be enough. Otherwise, it looks very busy

Thanks, and yes, we've agreed to collapse the common status packages in a previous thread. This example output is definitely a bit out of date.

blackboxsw:
Confirmed consolidated output now
```bash
root@usn-x:~# ua fix CVE-2017-9233
CVE-2017-9233: Coin3D vulnerability
https://ubuntu.com/security/CVE-2017-9233
3 affected packages are installed: expat, matanza, swish-e
(1/3, 2/3) matanza, swish-e:
Ubuntu security engineers are investigating this issue.
(3/3) expat:
A fix is available in Ubuntu standard updates.
The update is already installed.
✘ CVE-2017-9233 is not resolved.
```

This was fixed in 
csmith@uptown:~/src/ubuntu-advantage-client$ git show 06f0532e30f6af41b6edb1de550c0c3787e8c5a5

PR #1484, opened 2021-03-25
fix: show packages that were not fixed

lucasmoura:
## Proposed Commit Message
fix: show packages that were not fixed

When running ua fix when may end up in a state where we cannot fix all the packages. If we get into that situation, we will not display to the user what were the packages that were not fixed and are still affected by the CVE/USN that the command tried to fix.

Fixed: #1448

## Test Steps
Run the modified integration tests in this PR

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -701,8 +701,15 @@ def _handle_released_package_fixes(
     binary_pocket_pkgs: "Dict[str, List[str]]",
     pkg_index: int,
     num_pkgs: int,
-) -> bool:
-    """Handle the packages that could be fixed and have a released status."""
+) -> "Tuple[bool, List[str]]":
+    """Handle the packages that could be fixed and have a released status.
+
+    :returns: Tuple of
+        boolean whether all packages were successfully upgraded,
+        list of strings containing the packages that were not upgraded
+    """
+    upgrade_status = True
+    unfixed_pkgs = []
     if src_pocket_pkgs:
         for pocket in [
             UBUNTU_STANDARD_UPDATES_POCKET,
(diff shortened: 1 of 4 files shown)


uaclient/exceptions.py
```diff
@@ -1,5 +1,11 @@
 from uaclient import status
 
+try:
+    from typing import List, Optional  # noqa: F401
+except ImportError:
+    # typing isn't available on trusty, so ignore its absence
+    pass
+
```
blackboxsw, line 8: This addition is unnecessary for this PR as no typing hints were added.
```suggestion
```

uaclient/security.py
```diff
@@ -712,27 +714,44 @@ def _handle_released_package_fixes(
...
-            pkg_index += len(pkg_src_group)
+            if not upgrade_status:
+                not_fixed_pkgs += [src_pkg for src_pkg, _ in pkg_src_group]
+
+    return upgrade_status, not_fixed_pkgs
 
-    return True
+
+def _format_not_fixed_packages_msg(not_fixed_pkgs: "List[str]") -> str:
+    """Format the list of not fixed packages into an message.
```
blackboxsw, line 743:
```suggestion
    """Format the list of unfixed packages into an message.
```

uaclient/security.py
```diff
@@ -712,27 +714,44 @@ def _handle_released_package_fixes(
...
+
+    :returns: A string containing the message output for the not fixed
```
blackboxsw, line 745:
```suggestion
    :returns: A string containing the message output for the unfixed
```

uaclient/security.py
```diff
@@ -701,8 +701,10 @@ def _handle_released_package_fixes(
     binary_pocket_pkgs: "Dict[str, List[str]]",
     pkg_index: int,
     num_pkgs: int,
-) -> bool:
+) -> "Tuple[bool, List[str]]":
     """Handle the packages that could be fixed and have a released status."""
+    upgrade_status = True
+    not_fixed_pkgs = []
```
blackboxsw, line 707: I think we should substitute not_fixed for unfixed throughout this file and comments.

uaclient/security.py
```diff
@@ -701,8 +701,10 @@ def _handle_released_package_fixes(
     binary_pocket_pkgs: "Dict[str, List[str]]",
     pkg_index: int,
     num_pkgs: int,
-) -> bool:
+) -> "Tuple[bool, List[str]]":
     """Handle the packages that could be fixed and have a released status."""
```
blackboxsw, line 705: Let's improve his docstr a bit describing the return tuple values and their intent.

blackboxsw (changes requested):
Fix looks good I manually tested many different paths contract attach/detached/avoid enabling a service and the output looks good. Minor comments on variable naming and docstrings

lucasmoura:
Thanks for the review @blackboxsw This are some leftovers from my past solution. Let me clean them up and update the docstrings

lucasmoura:
@blackboxsw Code updated

blackboxsw (approved):
Behaves well in partial attach/disable/attached state. Thanks!
Integration passes locally.

PR #1488, opened 2021-03-25
tests: run status wait before detach in PRO tests

lucasmoura:
## Proposed Commit Message
tests: run status wait before detach in PRO tests

When we are creating a PRO instance, we install the latest uaclient package on the machine and then snapshot it to be used as a base image for all other tests. This base image should be a clean image, meaning that it should not be attached to uaclient. To ensure that, we run a detach operation on the base image and config it to block auto-attach. However, when we run detach, the auto-attach operation may still be running. We are now running ua status --wait before detach to avoid that.

## Test Steps
Run any of the GCP PRO integration tests

## Desired commit type::

## Checklist:

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -743,6 +743,7 @@ def _install_uat_in_container(
                 conf_path
             )
         )
+        cmds.append("sudo ua status --wait")
         cmds.append("sudo ua detach --assume-yes")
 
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
Generally a good thing. Won't make things worse and probably helps us out a lot to avoid incomplete detach.

PR #1492, opened 2021-03-29
fix: parse full API responses for related CVEs and USNs

blackboxsw:
## Proposed Commit Message

Reduce the number of API calls needed to query related CVEs and USNs
by processing the keys USN["cves"] and CVE["notices"] which now
contain detailed metadata response for the CVE and USN objects.

- Drop unneeded USN.get_cves_metadata and CVE.get_notices_metadata
- add CVE.notices to cache a list of USNs based on new "notices"
  key which contains full USN metadata from API response
- adapt CVE.notices_ids to rely on new "notices_ids" key from API
- add USN.cves to cache a list of CVEs based on the new "cves"
  full metadata from API response
- adapt USN.cves_ids to rely on new "cves_ids" key from API
- add equality checks for USN and CVE instances

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -1,6 +1,5 @@
 import copy
 import itertools
-import logging
 import os
 import socket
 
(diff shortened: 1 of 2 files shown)


uaclient/security.py
```diff
@@ -256,31 +255,19 @@ def __init__(self, client: UASecurityClient, response: "Dict[str, Any]"):
         self.response = response
         self.client = client
 
+    def __eq__(self, other) -> bool:
```
lucasmoura, line 258: I am missing why we need this override here and on the USN class as well

lucasmoura (approved):

uaclient/security.py
blackboxsw: > I am missing why we need this override here and on the USN class as well

Ahh sorry, I should have noted that, it is really only for ease of unittesting. in some of our pytest.mark.parametrize setup it allows us to instead create USN() CVE() in the list values and [do a direct equality comparison here](https://github.com/canonical/ubuntu-advantage-client/pull/1492/files#diff-1510e0bbe0ff95e8aa5c9a0d886d046cb9d1823c3eaf3bc6295cce6596779bceR304). This was we don't have to assert <expected_usn>.response  == <actual_usn>.response etc.

blackboxsw:
Confirmed this is an lxd error just on the jenkins server, local runs work fine
```bash
  behave-lxd-16.04: commands succeeded
  congratulations :)
```

Issue #1476, opened 2021-03-23
ua fix indicate if restart is needed


lucasmoura:
When running `ua fix` we may end up installing packages that require a system restart. If that happens, we must alert the user about it:

```
$ ua fix USN-4525-1
USN-4525-1: Linux kernel vulnerabilities
https://ubuntu.com/security/notices/USN-4525-1
1 affected package is installed: linux-image-generic
(1/1) linux-image-generic:
A fix is available in standard updates.
The update is installed. Kernel remains vulnerable until restart or Livepatch. 
A Livepatch fix is available. For details: “ua help livepatch”
```

lucasmoura:
@mitechie regarding `Livepatch` in that case. There is no way for us to verify if `Livepatch` has a fix for that kernel without the installed snap. We should only alert that for users that have `livepatch` enable right ? And if that is true, then we could assume that there will no kernel bugs, since `livepatch` would already have fixed it in the user system. Therefore, should we really alert users about `livepatch` fixes here ?

PR #1493, opened 2021-03-29
fix: notify users if reboot is required

lucasmoura:
## Proposed Commit Message
fix: notify users if reboot is required

When upgrading packages during ua fix we may install a package that requires a system reboot to fully fix the issue at hand. We are now alerting the user if a reboot is required when running ua fix.

Fixes: #1476

PS: The approach used here some shortcomings:

* If the user has installed a package that requires reboot and then run `ua fix`, we will not known if the reboot is need because of ua fix or the package installed by the user.
* If we detect that a reboot is required, should we say that the CVE/USN was not fixed ? I am using that approach, but I am not sure if that is the intended behavior.

## Test Steps
Run the new unittest added

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -526,6 +526,16 @@ class UAConfig:
 
         Write the status-cache when called by root.
         """
+
+        # Try to remove fix reboot notices if not applicable
+        if not util.should_reboot():
+            self.remove_notice(
+                "",
+                status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+                    operation="fix operation"
+                ),
+            )
+
         if os.getuid() != 0:
             response = cast("Dict[str, Any]", self.read_cache("status-cache"))
             if not response:
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -860,11 +860,11 @@ def prompt_for_affected_packages(
         print(_format_unfixed_packages_msg(unfixed_pkgs))
 
     if util.should_reboot():
-        print(
-            status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
-                operation="fix operation"
-            )
+        reboot_msg = status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+            operation="fix operation"
         )
+        print(reboot_msg)
+        cfg.add_notice("", reboot_msg)
         print(
             status.MESSAGE_SECURITY_ISSUE_NOT_RESOLVED.format(issue=issue_id)
         )
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -859,6 +859,17 @@ def prompt_for_affected_packages(
     if unfixed_pkgs:
         print(_format_unfixed_packages_msg(unfixed_pkgs))
 
+    if util.should_reboot():
+        print(
+            status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+                operation="fix operation"
+            )
+        )
+        print(
+            status.MESSAGE_SECURITY_ISSUE_NOT_RESOLVED.format(issue=issue_id)
+        )
+        return
+
     if fix_status:
         print(fix_message)
     else:
--- a/features/environment.py
+++ b/features/environment.py
@@ -635,11 +635,11 @@ def create_uat_image(context: Context, series: str) -> None:
     ppa = context.config.ppa
     if ppa == "":
         image_name = context.config.cloud_manager.locate_image_name(series)
-        print(
-            "--- Unset UACLIENT_BEHAVE_PPA. Using ubuntu-advantage-tools from image: {}".format(
-                image_name
-            )
-        )
+        msg = (
+            "--- Unset UACLIENT_BEHAVE_PPA. "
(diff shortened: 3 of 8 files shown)


uaclient/security.py
```diff
@@ -859,6 +859,17 @@ def prompt_for_affected_packages(
     if unfixed_pkgs:
         print(_format_unfixed_packages_msg(unfixed_pkgs))
 
+    if util.should_reboot():
```
blackboxsw, line 862: @lucasmoura, we are adding a second call-site which does util.should_reboot and messaging when reboot is needed. What do you think about refactoring util.should_reboot to accept an optional message: str param which we can pass in instead of having a [Entitlement separate class method.check_for_reboot_msg](https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/entitlements/repo.py#L85)

blackboxsw:
Confirmed current CI failure on this is           raise EOFError() on lxc init. known intermittent issue which hit bionic run this time. Doesn't affect local integration run.

uaclient/security.py
blackboxsw: Also, within this conditional if true, I think we also want to call to `cfg.add_notice("", status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(operation="fix operation"))`


If we do that we also need a cfg.remove_notice in reboot_cmds.py that'll clean up the reboot required message.
When you peek at the remove_notices in reboot_cmds.py let us know if you think we need to make better use descr_regex to remove any matching reboot message from notices

blackboxsw (changes requested):
Thanks Lucas. Generally looks pretty good, I think we should also provide notices related to this too so that `ua status` reports about this reboot beyond the transative message printed to console.  Maybe it's worth us thinking a bit about handling notice cleanup a bit more generically in  reboot_cmds to cleanse any messages of type "reboot required".  I'm not really sure on that second part, but food for thought.

uaclient/security.py
lucasmoura: @blackboxsw I agree on the notices approach, since the user may not reboot immediately after running the `ua fix` operation. However, I don't think allowing `should_reboot` to print messages will make the code better.

We would change the current code from:
```python
if util.should_reboot():
    print(
        status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
            operation="fix operation"
        )    
    )    
    print(
        status.MESSAGE_SECURITY_ISSUE_NOT_RESOLVED.format(issue=issue_id)
    )    
    return
```

to:
```python
msg =  status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
            operation="fix operation"
        )
        + "\n"
       + status.MESSAGE_SECURITY_ISSUE_NOT_RESOLVED.format(issue=issue_id)
if util.should_reboot(msg):
    return
```

My point is that I think it is better to make `should_reboot` be as simple as possible and we drive want we want to do outside of it. Even though we have that same behavior on FIPS, I still think it is better to have the output separate from the `should_reboot` function

lucasmoura:
@blackboxsw I have changed the approach a little bit. Instead of clearing the notices on the `reboot_cmd` script and I am now configuring `ua status` to check if a reboot operation is needed. If it is not, we clear all reboot notices.

I think this is a better solution since it is easier to test than pushing this cleanup to the `reboot_cmd` script. However, it will make the status operation slower. But since we don't have that many notices at the moment, this new cleanup step should not be that much of a burden in the `status` operation

uaclient/security.py
blackboxsw: +1 thanks for your suggestion/response here. Agreed.

uaclient/config.py
```diff
@@ -526,6 +526,18 @@ def status(self, show_beta=False) -> "Dict[str, Any]":
 
         Write the status-cache when called by root.
         """
+
+        # Try to remove all reboot notices if not applicable
+        if not util.should_reboot():
+            self.remove_notice("", status.MESSAGE_FIPS_REBOOT_REQUIRED)
+            self.remove_notice("", status.MESSAGE_FIPS_DISABLE_REBOOT_REQUIRED)
```
blackboxsw, line 533: I think we are missing status.MESSAGE_LIVEPATCH_LTS_REBOOT_REQUIRED too and I think this also means we can drop the remove_notices logic from reboot_cmds.py too

lucasmoura: For the `FIPS entitlements` case this will not happen because we check in `ua status`  the existence of the `/var/run/reboot-required` which indeed exists as can be seen here:
https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/entitlements/fips.py#L56

Now, for the FIPS PRO case, yeah, this will happen. But we can avoid that by also checking if the marker file is not present.
In that case, we will guarantee that the FIPS PRO case will be handled by the reboot script only and the rest of the cases will be handled by status.

What do you think ?

lucasmoura:
@blackboxsw I have updated the code to only remove the `fix` notice during `ua status`. The reason for that is because we are already managing the other notices on separate places. For the FIPS ones that I have added, I have found that we are already cleaning them up here:

https://github.com/canonical/ubuntu-advantage-client/blob/master/uaclient/entitlements/fips.py#L187

PR #1497, opened 2021-03-30
d/changelog: bring changelog from release 26

lucasmoura:


blackboxsw (approved):
The only manual diff I think we are missing is a minor additional GH issue link in an older changelog entry. If you could add the GH: #1335 to that older changelog line it'll make the diff across releases negligible for changelog

-    - contract: handle 401 on invalid token, 403 on expired (GH: #1335)

Issue #1487, opened 2021-03-25
fix: output should be wrapped for lines > 80 characters


blackboxsw:
While this is an error case for matching too many unrelated CVEs, it shows a lack for formatting/wrapping in `ua fix` messaging.

Note the "Found CVEs" line in the output below

Make sure we are appropriately wrapping (and indenting where necessary) if line length is > 80 characters in messaging,

More broadly, whatever wrapping handler we define  could be applied to any messages printed by ua.

```bash
oot@dev-x:~# ua fix USN-4770-1
USN-4770-1: GlusterFS vulnerabilities
Found CVEs: CVE-2018-14661, CVE-2018-14660, CVE-2018-14659, CVE-2018-14654, CVE-2018-14653, CVE-2018-14652, CVE-2018-14651, CVE-2018-10930, CVE-2018-10929, CVE-2018-10928, CVE-2018-10927, CVE-2018-10926, CVE-2018-10924, CVE-2018-10923, CVE-2018-10914, CVE-2018-10913, CVE-2018-10911, CVE-2018-10907, CVE-2018-10904, CVE-2018-1088, CVE-2018-10841, CVE-2014-3619
https://ubuntu.com/security/CVE-2018-10929
https://ubuntu.com/security/CVE-2018-14653
https://ubuntu.com/security/CVE-2014-3619
https://ubuntu.com/security/CVE-2018-10923
https://ubuntu.com/security/CVE-2018-14660
https://ubuntu.com/security/CVE-2018-10928
https://ubuntu.com/security/CVE-2018-10926
https://ubuntu.com/security/CVE-2018-10924
https://ubuntu.com/security/CVE-2018-14651
https://ubuntu.com/security/CVE-2018-10911
https://ubuntu.com/security/CVE-2018-10930
https://ubuntu.com/security/CVE-2018-10907
https://ubuntu.com/security/CVE-2018-14659
https://ubuntu.com/security/CVE-2018-10904
https://ubuntu.com/security/CVE-2018-10913
https://ubuntu.com/security/CVE-2018-14654
https://ubuntu.com/security/CVE-2018-14652
https://ubuntu.com/security/CVE-2018-10914
https://ubuntu.com/security/CVE-2018-10841
https://ubuntu.com/security/CVE-2018-1088
https://ubuntu.com/security/CVE-2018-14661
https://ubuntu.com/security/CVE-2018-10927
No affected packages are installed.
✔ USN-4770-1 does not affect your system.
```

PR #1501, opened 2021-03-31
fix: wrap output greater than 80 chars

orndorffgrant:
## Proposed Commit Message

> fix: wrap output greater than 80 characters
>
> Ensures output doesn't pass 80 characters when listing affected
> packages and printing long commands.
>
> Fixes: #1487

## Test Steps

TODO: @blackboxsw how did you run into this problem when you created the issue #1487 ?

Also, are there any other places we know of where we should be wrapping output?

## Desired commit type::

## Checklist:


uaclient/security.py
```diff
@@ -550,7 +550,7 @@ def fix_security_issue_id(cfg: UAConfig, issue_id: str) -> None:
             usn=usn, installed_packages=installed_packages
         )
         usn_released_pkgs = merge_usn_released_binary_package_versions(usns)
-        print(usn.get_url_header())
+        print(util.wrap_text(usn.get_url_header()))
```
orndorffgrant, line 553: Two things:
1. The issue mentioned indenting: Do we want to add a hanging indent here? We can do so with the `subsequent_indent` kwarg.
2. It might make more sense to move this `util.wrap_text` call into the `get_url_header` method rather than here - thoughts?

uaclient/tests/test_security.py
```diff
@@ -1900,6 +1901,7 @@ def test_error_msg_when_usn_does_not_have_any_related_usns(
                     return_value=["cve-123"]
                 )
                 type(usn_mock).id = mock.PropertyMock(return_value="id")
+                type(usn_mock).get_url_header = USN.get_url_header
```
orndorffgrant, line 1904: These lines were necessary because otherwise the tests would fail when a `MagicMock` would get returned from `get_url_header` and passed to `wrap_text`, which couldn't handle it because its not a string/bytes.

It is surprising to me that we even are printing the url header for these cases though (purely based on the names of the tests, so I'm likely wrong). :shrug:

uaclient/tests/test_util.py
```diff
@@ -727,3 +727,74 @@ class TestRedactSensitiveLogs:
     def test_redact_all_matching_regexs(self, raw_log, expected):
         """Redact all sensitive matches from log messages."""
         assert expected == util.redact_sensitive_logs(raw_log)
+
+
+class TestWrapText:
```
orndorffgrant, line 732: This only looks like a huge test because it involves long strings :upside_down_face:

uaclient/util.py
```diff
@@ -687,3 +690,20 @@ def redact_sensitive_logs(
...
+
+    Additional kwargs are passed on to textwrap.wrap
+
+    @param text: The long string that needs to be printed but wrapped at a
+                    given width.
+    @param width: The maximum line length allowed in the result.
+    @return: The text with newlines inserted appropriately so that it never
+                    exceeds the given width.
+    """
+    return "\n".join(textwrap.wrap(text, width=width, **kwargs))
```
orndorffgrant, line 709: While docs.python.org doesn't include docs for `textwrap` for python 3.4, I did check that this is available in a trusty container on python3.4.

This helper function only really adds two things for convenience. It sets the width to 80 (70 is the default), and joins the result with newlines so we get a string ready to be printed rather than a list of strings.

orndorffgrant:
@lucasmoura The mypy jenkins step is failing - is this the same issue that you and @paride were talking about earlier today/yesterday?

lucasmoura:
@orndorffgrant Yes it is

uaclient/security.py
blackboxsw: > Two things:
> 
> 1. The issue mentioned indenting: Do we want to add a hanging indent here? We can do so with the `subsequent_indent` kwarg.

I think we'll probably need an indent there, we defintely should toss this example by @mitechie to get his 
thoughts on the look. Anytime we see a wall-o-text, I cringe and wonder if we should really be hiding some of this content with ellipsis as it becomes unfriendly. here's the example of painful output on a unique USN https://pastebin.canonical.com/p/tXF8xmQD9s/

> 2. It might make more sense to move this `util.wrap_text` call into the `get_url_header` method rather than here - thoughts?

Yes let's move that call inside. best to expect formatted output functions to return us well-formatted output.

uaclient/util.py
blackboxsw: Normally, I'd want to avoid declaring our own func for something that currently wraps a single line of code shipped by a standard library, but in uaclient case, I think we will have extra concerns to worry about such as how we want to indent across wraps and the potential of colorized character code sequences. It's because of the latter that I think we may need to have this function in util.   That said, I think we might need to look at baking some of this utility into status.colorize_commands too.

orndorffgrant: Yeah I generally agree - the only reason I opted for this function is to replace the default width so that we don't have to pass that every time we use it, but that's a really minor thing. I'm totally open to removing this entirely and we can just call `textwrap.wrap(whatever, width=util.DEFAULT_WRAP_WIDTH)` wherever we need it.

I was thinking more about the colorization, and I'm leaning towards suggesting that we always wrap before we colorize, to avoid problems like you were mentioning (as opposed to getting the wrap function to understand the color control chars). Do you think that is workable or do you know of scenarios where we definitely have to colorize a chunk of text _before_ we wrap it?

I need to look more into how `status.colorize_commands` is used, but yes it looks like we'll probably want a textwrap on that `content` var before adding the color control codes around it.

uaclient/security.py
orndorffgrant: 1. Oof I didn't notice before, but textwrap is totally mangling some of the intended newlines in there. There may be a parameter for textwrap to honor existing newlines, or we can just only call textwrap on the chunk of text that we're dynamically generating and know could go long. I'll look into it

2. Agreed! I'll move it

uaclient/util.py
blackboxsw: > Yeah I generally agree - the only reason I opted for this function is to replace the default width so that we don't have to pass that every time we use it, but that's a really minor thing. I'm totally open to removing this entirely and we can just call `textwrap.wrap(whatever, width=util.DEFAULT_WRAP_WIDTH)` wherever we need it.

Yeah that makes sense too minimally, it'd be nice to avoid reflecting the same additional params to every callsite.

> 
> I was thinking more about the colorization, and I'm leaning towards suggesting that we always wrap before we colorize, to avoid problems like you were mentioning (as opposed to getting the wrap function to understand the color control chars). Do you think that is workable or do you know of scenarios where we definitely have to colorize a chunk of text _before_ we wrap it?

Excellent thought/suggestion. yes we should be able to wrap before colorization in colorize_commands callsites. or rather inside the colorize_commands call itself.  I think generally this stance should hold +1 on the suggestion, so maybe this isn't as big as I thought.


> 
> I need to look more into how `status.colorize_commands` is used, but yes it looks like we'll probably want a textwrap on that `content` var before adding the color control codes around it.

+1 go for it.

uaclient/security.py
blackboxsw: > 1. Oof I didn't notice before, but textwrap is totally mangling some of the intended newlines in there. There may be a parameter for textwrap to honor existing newlines, or we can just only call textwrap on the chunk of text that we're dynamically generating and know could go long. I'll look into it

Thanks I notices that we were missing leading and trailing newlines on that chunk of "wrapped" text. Figured I'd get back to you with a more detailed review why, but it seems I've run out of daylight on this review today :(. Will have to look monday.

orndorffgrant:
@blackboxsw I reworked this a bit (got rid of the helper function) and added wrapping to the affected packages output as well.

@mitechie Heres what the new output looks like in the cases where its applied:

First, a lot of CVEs for a given USN:
```
USN-4510-2: Samba vulnerability
Found CVEs: CVE-2020-1473, CVE-2020-1473, CVE-2020-1473, CVE-2020-1473,
            CVE-2020-1473, CVE-2020-1473, CVE-2020-1473, CVE-2020-1473,
            CVE-2020-1472, CVE-2020-1472, CVE-2020-1472, CVE-2020-1472,
            CVE-2020-1472, CVE-2020-1472, CVE-2020-1472, CVE-2020-1472
https://ubuntu.com/security/CVE-2020-1473
https://ubuntu.com/security/CVE-2020-1472
https://ubuntu.com/security/CVE-2020-1473
https://ubuntu.com/security/CVE-2020-1472
https://ubuntu.com/security/CVE-2020-1473
https://ubuntu.com/security/CVE-2020-1472
https://ubuntu.com/security/CVE-2020-1473
https://ubuntu.com/security/CVE-2020-1472
https://ubuntu.com/security/CVE-2020-1473
https://ubuntu.com/security/CVE-2020-1472
https://ubuntu.com/security/CVE-2020-1473
https://ubuntu.com/security/CVE-2020-1472
https://ubuntu.com/security/CVE-2020-1473
https://ubuntu.com/security/CVE-2020-1472
https://ubuntu.com/security/CVE-2020-1473
https://ubuntu.com/security/CVE-2020-1472
```

Second, a few long affected package names
```
5 affected packages are installed: longpackagename1, longpackagename2,
    longpackagename3, longpackagename4, longpackagename5
(1/5, 2/5, 3/5, 4/5, 5/5) longpackagename1, longpackagename2, longpackagename3,
    longpackagename4, longpackagename5:
A fix is available in Ubuntu standard updates.
{ apt update && apt install --only-upgrade -y longpackagename1 longpackagename2 \
  longpackagename3 longpackagename4 longpackagename5 }
✔ USN-### is resolved.
```

I'm open to any tweaking, especially for the indentation.

mitechie:
This is ok to land from a UX perspective in my mind. I am wondering if the "found cves: one, two, three" is just duplicate info with the urls and just doing a 

```
Found CVEs:
https://...
https://...
```

would be cleaner

blackboxsw:
> This is ok to land from a UX perspective in my mind. I am wondering if the "found cves: one, two, three" is just duplicate info with the urls and doing a
> 
> ```
> Found CVEs:
> https://...
> https://...
> ```
> 
> would be cleaner

+1 I think we chatted in standup about this. reduced duplication of messaging looks quite a bit better here.

uaclient/status.py
```diff
@@ -392,8 +393,14 @@ def colorize_commands(commands: "List[List[str]]") -> str:
         if content:
             content += " && "
         content += " ".join(cmd)
+    # subtract 2 from print width to make room for space and backslash
+    wrapped_content = " \\\n".join(
```
blackboxsw, line 397: I don't think we need the double-escape squence here to we?
```suggestion
    wrapped_content = " \n".join(
```

orndorffgrant: Ah so I added a backslash `\\` to the end of the line there, because thats how you would typically break up a command into multiple lines so that it can be copy-pasted.

E.g.
```
apt update && apt install package1 \
    package2 package3
```

This isn't intended to be copy-pasted/run by the user though, so we can get rid of it if you think its unnecessary.

blackboxsw, line 397: Sorry I misread that escape sequent, right we totally should have the trailing backslash to allow humans to cut-n-paste the commands. Along the same traing of thought, I think we need to make our output look more like this
```bash
{
   some long \
   command
}
# instead of 
{ some long \
  command }. 
```

Here's a patch of what I was thinking. See what you think or adapt as necessary. The reason I wanted to separate multi-line commands from the  leading/trailing brances is because we don't want cut-n-paste efforts with a mouse to have to worry about accidental inclusion of leading or trailing characters unrelated to the command.


```diff
diff --git a/uaclient/status.py b/uaclient/status.py
index 07eb68b..9824a25 100644
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -1,7 +1,8 @@
 import enum
 import sys
+import textwrap
 
-from uaclient.defaults import BASE_UA_URL
+from uaclient.defaults import BASE_UA_URL, PRINT_WRAP_WIDTH
 
 try:
     from typing import Any, Dict, List, Optional, Tuple, Union  # noqa: F401
@@ -392,8 +393,21 @@ def colorize_commands(commands: "List[List[str]]") -> str:
         if content:
             content += " && "
         content += " ".join(cmd)
-    return "{color}{{ {content} }}{end}".format(
-        color=TxtColor.DISABLEGREY, content=content, end=TxtColor.ENDC
+    # subtract 4 from print width to account for leading and trailing braces
+    # and spaces.
+    wrapped_content = " \\\n".join(
+        textwrap.wrap(
+            content, width=(PRINT_WRAP_WIDTH - 4), subsequent_indent="  "
+        )
+    )
+    if "\n" in wrapped_content:
+        prefix = "{\n  "
+        suffix = "\n}"
+    else:
+        prefix = "{ "
+        suffix = " }"
+    return "{color}{prefix}{content}{suffix}{end}".format(
+        color=TxtColor.DISABLEGREY, content=wrapped_content, end=TxtColor.ENDC, prefix=prefix, suffix=suffix
     )
 

```

blackboxsw (changes requested):
Thanks @orndorffgrant let's tweak the multiline output colorize_commands a bit if we can.please see what you think makes sense.

uaclient/status.py
orndorffgrant: Very good point! I'll work this in

orndorffgrant: This is in the latest commit - ended up looking exactly like this, plus the associated test change.

orndorffgrant:
The integration test failure is a known lxd ssh issue. I confirmed that the test passed locally, which makes this ready for re-review.

blackboxsw (approved):
Looks good to me. thanks @orndorffgrant for the work here.

PR #1506, opened 2021-04-01
Make lint and style stage run sequentially

lucasmoura:
## Proposed Commit Message
Make lint and style stage run sequentially

We are experiencing some issues running mypy tests. We believe that reason is because we are running tox inside
a virtualenv. To fix that, we removing the virtualenv creation from this stage and also making the jobs in that stage run sequentially

## Test Steps
We can spot the behavior on the CI

## Desired commit type::

## Checklist:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,35 +37,28 @@ pipeline {
                 '''
             }
         }
-        stage ('Lint and Style') {
-            parallel {
-                stage("flake8") {
-                    steps {
-                        sh '''
-                        set +x
-                        . $TMPDIR/bin/activate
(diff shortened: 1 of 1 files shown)


blackboxsw (commented):
+1

blackboxsw (approved):

paride:
@lucasmoura @blackboxsw thanks!

Newer versions of tox have a `--parallel` flag which allows to run environments in parallel. That seems to be the right thing to do to speed things up here, unfortunately the version of tox in Bionic doesn't have it yet. I'll upgrade the nodes to Focal at some point, then I think we'll be able to do a parallel run.

Issue #1507, opened 2021-04-01
image build: ua fips enable set target cloud within chroot/build environments


blackboxsw:
Currently fips enablement relies on use of `cloud-id` utility from cloud-init on systems with cloud-init installed. In image build environments, cloud-init will not yet have run so cloud-id will not succeed. Even if cloud-init did run (which is undesireable), it'd likely detect the wrong cloud due to the sandbox in which it is running.

Errors will be seen like this in chroot or build environments without cloud-init having run:
```bash
uaclient.util.ProcessExecutionError: Failed running command 'cloud-id' [exit(1)]. Message: ERROR: File not found '/run/cloud-init/instance-data.json'. Provide a path to instance data json file using --instance-data
```

PR #1508, opened 2021-04-01
{WIP] config: create settings_overrides on config

lucasmoura:
## Proposed Commit Message
config: create settings_overrides on config

Currently, we need cloud-init to identify if FIPS can be correctly installed in an instance. This behavior is making
it harder to use uaclient in other scenarios, for example, when building an image using chroot. To better handle that, we are
creating a settings_overrides dict in ua config. Users will be able to override specific settings by just updating the uaclient.conf
file.  For example, to override the cloud type users will need to add this to `uaclient.conf`:

```yaml
settings_overrides:
  cloud_type: "gce"
```

To use these overrides, we have created a decorator that can modify a function return based on the override defined in the config. As of today, this is only being used in the get_cloud_type function.

Fixes: #1507

## Test Steps
This is a WIP branch

We have a new unittest for this feature, but more tests can be included

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -1,5 +1,6 @@
 import copy
 from datetime import datetime
+from functools import wraps
 import json
 import logging
 import os
@@ -47,6 +48,7 @@ MERGE_ID_KEY_MAP = {
     "availableResources": "name",
     "resourceEntitlements": "type",
 }
+UNSET_SETTINGS_OVERRIDE_KEY = "_unset"
 
 
 # A data path is a filename, and an attribute ("private") indicating whether it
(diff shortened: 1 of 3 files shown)


uaclient/config.py
```diff
@@ -660,6 +661,34 @@ def parse_config(config_path=None):
...
+     entry in the config dict. If that key is present, we will return
+     its value as the function return.
+    """
+
+    def wrapper(f):
+        @wraps(f)
+        def new_f():
+            cfg = parse_config()
+            value_override = cfg.get("settings_overrides", {}).get(
+                override_key
```
blackboxsw, line 682: @lucasmoura let's set a known unset sentinel value if override_key like 
UNSET = "_unset"                                                                

That way it would still allow someone to override "cloud_type : null" in uaclient.conf

blackboxsw, line 682:
```suggestion
                override_key, UNSET
```

blackboxsw (commented):
This looks like the right kind of approach to me. I'll poke around a bit as you are sorting whatever additional unittest coverage you have marked this WIP for.

blackboxsw (approved):
Excellent, thanks for this Lucas. works like a charm.

PR #1515, opened 2021-04-05
Update container check for chroot environments

lucasmoura:
## Proposed Commit Message
util: improve is_container check for chroot
    
When on a chroot, we don't want to say that the system is a container type. To avoid that, we are first calling the ischroot command. If it returns 0, we assume that we are on a chroot and inform that we are not in a container.

## Test Steps
Run the modified unit tests

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -352,11 +352,23 @@ def get_platform_info() -> "Dict[str, str]":
 @lru_cache(maxsize=None)
 def is_container(run_path: str = "/run") -> bool:
     """Checks to see if this code running in a container of some sort"""
+
+    # We may mistake schroot environments for containers by just relying
+    # in the other checks present in that function. To guarantee that
+    # we do not identify a schroot as a container, we are explicitly
+    # using the 'ischroot' command here.
+    try:
+        subp(["ischroot"])
+        return False
+    except ProcessExecutionError:
+        pass
+
     try:
         subp(["systemd-detect-virt", "--quiet", "--container"])
         return True
     except (IOError, OSError):
         pass
+
     for filename in ("container_type", "systemd/container"):
         path = os.path.join(run_path, filename)
         if os.path.exists(path):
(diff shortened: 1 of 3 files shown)


uaclient/util.py
```diff
@@ -352,11 +352,18 @@ def get_platform_info() -> "Dict[str, str]":
 @lru_cache(maxsize=None)
 def is_container(run_path: str = "/run") -> bool:
     """Checks to see if this code running in a container of some sort"""
```
orndorffgrant, line 354: This new chroot special case might be a little confusing to future readers. Consider adding a note to the docstring here along the lines of: "chroots can appear to be containers but ua shouldn't treat them as containers, so we explicitly check for chroot and return False if we're in one"

uaclient/util.py
```diff
@@ -352,11 +352,18 @@ def get_platform_info() -> "Dict[str, str]":
...
+    try:
+        subp(["ischroot"])
+        return False
+    except ProcessExecutionError:
```
orndorffgrant, line 358: Curious as to why here we're catching a `ProcessExecutionError` where down below we're catching `IOError` and `OSError` - do you know what is causing the difference?

orndorffgrant (commented):
Looks good to me! Just left a tiny optional suggestion and asked a question for my own curiosity.

uaclient/util.py
lucasmoura: Yes, that's an excellent idea

uaclient/util.py
lucasmoura: The `ProcessExecutionError` happens when the command return something different from the expected return codes, which by default is `0`. I don't have the context of the `systemd` call here, but I assume that the command fails for different reasons then the expected return code

lucasmoura:
@orndorffgrant comment added

uaclient/util.py
orndorffgrant: Got it - thanks!

blackboxsw (approved):
This looks good and doing some local testing here (as well as some on aws) it feels sound. I've pushed to a PPA and reflected a question to CPC on this to make sure it'll work for them too. If so, we can land this and cut a 26.3 release with it.

PR #1516, opened 2021-04-05
fix: notify users if reboot is required

lucasmoura:
## Proposed Commit Message
fix: notify users if reboot is required

When upgrading packages during ua fix we may install a package that requires a system reboot to fully fix the issue at hand. We are now alerting the user if a reboot is required when running ua fix.

Fixes: #1476

PS: The approach used here some shortcomings:

* If the user has installed a package that requires reboot and then run ua fix, we will not known if the reboot is need because of ua fix or the package installed by the user.
* If we detect that a reboot is required, should we say that the CVE/USN was not fixed ? I am using that approach, but I am not sure if that is the intended behavior.

## Test Steps
Run the new unittest added

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -526,6 +526,16 @@ class UAConfig:
 
         Write the status-cache when called by root.
         """
+
+        # Try to remove fix reboot notices if not applicable
+        if not util.should_reboot():
+            self.remove_notice(
+                "",
+                status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+                    operation="fix operation"
+                ),
+            )
+
         if os.getuid() != 0:
             response = cast("Dict[str, Any]", self.read_cache("status-cache"))
             if not response:
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -870,11 +870,11 @@ def prompt_for_affected_packages(
         print(_format_unfixed_packages_msg(unfixed_pkgs))
 
     if util.should_reboot():
-        print(
-            status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
-                operation="fix operation"
-            )
+        reboot_msg = status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+            operation="fix operation"
         )
+        print(reboot_msg)
+        cfg.add_notice("", reboot_msg)
         print(
             status.MESSAGE_SECURITY_ISSUE_NOT_RESOLVED.format(issue=issue_id)
         )
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -869,6 +869,17 @@ def prompt_for_affected_packages(
     if unfixed_pkgs:
         print(_format_unfixed_packages_msg(unfixed_pkgs))
 
+    if util.should_reboot():
+        print(
+            status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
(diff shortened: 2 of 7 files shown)


uaclient/config.py
```diff
@@ -526,6 +526,16 @@ def status(self, show_beta=False) -> "Dict[str, Any]":
 
         Write the status-cache when called by root.
         """
+
+        # Try to remove fix reboot notices if not applicable
+        if not util.should_reboot():
+            self.remove_notice(
+                "",
+                status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
+                    operation="fix operation"
```
blackboxsw, line 535: I think we probably want to clear all reboot required notices in this case right? If so, let's take advantage of globbing in the descr_regex param
```suggestion
                    operation=".*"
```

uaclient/config.py
```diff
@@ -526,6 +526,16 @@ def status(self, show_beta=False) -> "Dict[str, Any]":
 
         Write the status-cache when called by root.
         """
+
+        # Try to remove fix reboot notices if not applicable
+        if not util.should_reboot():
```
blackboxsw, line 531: Should we also check presence of os.path.exists(self.data_path("marker-reboot-cmds") too before clearing the reboot notices?
```suggestion
        if not util.should_reboot() and not os.path.exists(self.data_path("marker-reboot-cmds") :
```

blackboxsw (commented):
@lucasmoura thanks for this couple questions here. as I think we could be a bit more generic in the remove_notice case, and specific in the conditional that gets us into this part of the cleanup.

features/environment.py
```diff
@@ -635,11 +635,11 @@ def create_uat_image(context: Context, series: str) -> None:
...
-        print(
-            "--- Unset UACLIENT_BEHAVE_PPA. Using ubuntu-advantage-tools from image: {}".format(
-                image_name
-            )
-        )
+        msg = (
+            "--- Unset UACLIENT_BEHAVE_PPA. "
+            "Using ubuntu-advantage-tools from image: {}"
+        ).format(image_name)
+        print(msg)
```
orndorffgrant, line 642: I also fixed this line-too-long issue but in a slightly different way - so that's why there is a merge conflict here. Your version looks a bit better though so I'd just accept this version.

uaclient/config.py
lucasmoura: No, I don't think we can't do this because of FIPS PRO case. After we upgrade the package in a FIPS PRO machine, we will emit a notice telling that the user needs to restart the system to install the right metapackage. However, we will not have a `/var/run/reboot-required` file in that situation. This means that if the user runs `ua status` after upgrading the package in that machine, we will remove the notice.

Because of that, I am only removing the `fix` notices on that situation.

uaclient/config.py
lucasmoura: We could, but since I am only removing `fix` related notices, I don't think it is necessary. The other notices are handled in different situations, like those FIPS ones:
https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/entitlements/fips.py#L187

features/environment.py
lucasmoura: @orndorffgrant I am fine with your fix and I don't think we need an extra commit just to change that. I will drop my changes here.

blackboxsw (approved):
+1 thanks for the discussion here.

PR #1519, opened 2021-04-05
fix: update how the expired token logic works

lucasmoura:
## Proposed Commit Message
fix: update how the expired token logic works

We are now updating the logic to verify if an UA subscription has expired. Instead of requesting an ua refresh and verifying if the
contract backend has raised any errors, we are now going directly to the machine-token.json file and verifing until which date the contract is valid through the effectiveTo field.

## Test Steps
Run the modified unit tests and run the modified trusty integration test

## Desired commit type::

## Checklist:

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -18,7 +18,7 @@ from hamcrest import (
 from features.environment import create_uat_image
 from features.util import SLOW_CMDS, emit_spinner_on_travis, nullcontext
 
-from uaclient.defaults import DEFAULT_CONFIG_FILE
+from uaclient.defaults import DEFAULT_CONFIG_FILE, DEFAULT_MACHINE_TOKEN_PATH
 
 
 CONTAINER_PREFIX = "ubuntu-behave-test-"
@@ -168,6 +168,33 @@ def when_i_fix_a_issue_by_enabling_service(context, issue):
         command="ua fix {}".format(issue),
         user_spec="with sudo",
         stdin="e\n",
+    )
+
+
+@when("I fix `{issue}` by updating expired token")
+def when_i_fix_a_issue_by_updating_expired_token(context, issue):
+    token = getattr(context.config, "contract_token")
+    when_i_run_command(
+        context=context,
+        command="ua fix {}".format(issue),
+        user_spec="with sudo",
+        stdin="r\n{}\n".format(token),
+    )
+
+
+@when("I update contract to use `{contract_field}` as `{new_value}`")
+def when_i_update_contract_field_to_new_value(
+    context, contract_field, new_value
+):
+    when_i_run_command(
+        context,
+        'sed -i \'s/"{}": "[^"]*"/"{}": "{}"/g\' {}'.format(
+            contract_field,
+            contract_field,
+            new_value,
+            DEFAULT_MACHINE_TOKEN_PATH,
+        ),
+        user_spec="with sudo",
     )
 
 
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -69,6 +69,7 @@ class UAConfig:
 
     _entitlements = None  # caching to avoid repetitive file reads
     _machine_token = None  # caching to avoid repetitive file reading
+    _contract_expiry_datetime = None
 
     def __init__(
         self, cfg: "Dict[str, Any]" = None, series: str = None
@@ -211,6 +212,18 @@ class UAConfig:
             util.apply_series_overrides(entitlement_cfg, self.series)
             self._entitlements[entitlement_name] = entitlement_cfg
         return self._entitlements
+
+    @property
+    def contract_expiry_datetime(self) -> "datetime":
+        if not self._contract_expiry_datetime:
+            contractInfo = self.machine_token["machineTokenInfo"][
+                "contractInfo"
+            ]
+            self._contract_expiry_datetime = datetime.strptime(
+                contractInfo["effectiveTo"], "%Y-%m-%dT%H:%M:%SZ"
+            )
+
+        return self._contract_expiry_datetime
 
     @property
     def is_attached(self):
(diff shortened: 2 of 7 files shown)


features/steps/steps.py
```diff
@@ -171,6 +171,33 @@ def when_i_fix_a_issue_by_enabling_service(context, issue):
...
+    context, contract_field, new_value
+):
+    when_i_run_command(
+        context,
+        'sed -i \'s/"{}": "[^"]*"/"{}": "{}"/g\' {}'.format(
+            contract_field,
+            contract_field,
+            new_value,
+            DEFAULT_MACHINE_TOKEN_PATH,
+        ),
```
orndorffgrant, line 196: Fun! My first thought would've been to read in the file, parse it as json into a dict, make the update, and then serialize it back to json. But this works just as well for any string value. I manually double checked it would handle commas correctly and it did.

Will this ever need to be used for a non-string json value?

uaclient/security.py
```diff
@@ -1018,15 +1019,25 @@ def _check_subscription_is_expired(cfg: UAConfig) -> bool:
...
-        if status.MESSAGE_ATTACH_EXPIRED_TOKEN in str(e):
-            return not _prompt_for_new_token(cfg)
-        else:
-            raise e
+    if not effective_to_date:
+        msg = (
+            "Cannot verify if the current UA subscription is valid or not\n"
+            "Could not find the effectiveTo field in the UA Contract.\n"
+        )
+        raise exceptions.UserFacingError(msg)
```
orndorffgrant, line 1033: In what scenario would this happen? Asking because if we know then we might be able to give additional information here.

For example, would it make sense to suggest that the user run `ua refresh` (if that does what I think it does)?

orndorffgrant (commented):
This looks like a much better solution! I left some small comments.

~~And here's a newbie question :)
Say I just cloned this branch and want to manually run these changes, what specific commands would you suggest I run to set up the scenario and run this version of the code?~~ I'm good :)

features/steps/steps.py
lucasmoura: I don't think it should, since the sed command here would not work well for other types. I think the best way for us is to add another function that handles that if needed

uaclient/security.py
lucasmoura: That's a good point. 

No, I have never seen this case happen and the code here is just the idea of better be safe than sorry.
But yeah, maybe we should treat this corner case scenario differently. What do you think @blackboxsw ?

features/steps/steps.py
orndorffgrant: Yep that would work. Then we would probably rename this to something like `when_i_update_contract_field_to_new_str_value` and the nonstring version would become something like `when_i_update_contract_field_to_new_nonstr_value`. But I don't think we need to do that preemptively.

lucasmoura: We could even keep the function as is and just redirect in that method based on the type of `new_value`. But I agree that we must implement this when needed

uaclient/security.py
blackboxsw: I think we can drop this check here, I have not seen an absent effectiveTo field in a couple years and the ones we had were temporary test accounts.

uaclient/security.py
```diff
@@ -1039,15 +1040,25 @@ def _check_subscription_is_expired(cfg: UAConfig) -> bool:
 
     :returns: True if subscription is expired and not renewed.
     """
-    from uaclient import cli
+    effective_to_date = (
```
blackboxsw, line 1043: This doesn't feel like it belongs in security.py. Can we bake this logic into a propert decorated method on UAConfig of some sort. Ultimately we'll need something like days_until_expiry in order to allow us to emit various MOTD/apt messages based on contracts expiring soon, contracts only expired less than a grace period (14 days) and contracts that have expired longer than the grace period. Also it'd be good to keep all json extraction logic for this effectiveTo key inside UAConfig so other callsites don't have to replicate error handling (and because eventually we'll have to support multiple contracts with potentially different expiry dates which will be surfaced under something like the contract ["items" list](https://github.com/CanonicalLtd/ua-contracts/blob/master/docs/contracts.yaml#L438) which you can see in current contract responses (and currently contains a single item list with the same effectiveTo key.

blackboxsw (changes requested):
Thanks for this Lucas, I think we can drop the handling of absence of "effectiveTo". Ultimately I think we need to move this contract expiry logic into UAConfig in some way so we can leverage for the other MOTD/apt-hook needs.

uaclient/config.py
```diff
@@ -212,6 +213,18 @@ def entitlements(self):
             self._entitlements[entitlement_name] = entitlement_cfg
         return self._entitlements
 
+    @property
+    def contract_expiry_datetime(self) -> "datetime":
+        if not self._contract_expiry_datetime:
+            contractInfo = self.machine_token["machineTokenInfo"][
```
blackboxsw, line 219: @lucasmoura,this will break on unattached machines because machine_token is None
This will also break for non-root users because they won't have access to machine-token readability in general, should we instead be using expires" key from /var/lib/ubuntu-advantage/status.json if exists as that is world readable.

```bash
TypeError: 'NoneType' object is not subscriptable
```

blackboxsw (changes requested):
Thanks @lucasmoura  I think we need a touch on where that expiry is coming from so non-root users and unattached machines won't traceback. Then we can proceed.

uaclient/config.py
blackboxsw: Will take this need to a separate branch for messaging that will need to check attach status prior to calling this method.

blackboxsw:
new fun intermittent PRO image CI test failure completely unrelated to this changeset. ua-auto-attach seeing a lock held by itself. I think we might have to make sure our locks are all cleaned up in the based image we create.
https://github.com/canonical/ubuntu-advantage-client/issues/1543

blackboxsw (approved):
LGTM. holding off on UAConfig updates for unattached machines for a separate PR

Issue #1518, opened 2021-04-05
Errrors during reboot-cmds leave stale "Operation in progress: ua-reboot-cmds"


blackboxsw:
When an unexpected error occurs during reboot-cmds operations, UA status notices.json needs to update(and clean) notices to reflect the current state of the system.

NOTICES should not continue to report the folowing a failure to complete any reboot startup operations.

```bash
NOTICES
Operation in progress: ua-reboot-cmds
```

Maybe it's worth reporting that an error occurred here.

```bash
$ systemctl status ua-reboot-cmds
● ua-reboot-cmds.service - Ubuntu Advantage reboot cmds
   Loaded: loaded (/lib/systemd/system/ua-reboot-cmds.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Mon 2021-04-05 19:16:28 UTC; 18min ago
 Main PID: 1728 (code=exited, status=1/FAILURE)

Apr 05 19:16:28 ip-172-31-43-230 python3[1728]:   File "/usr/lib/python3/dist-packages/uaclient/apt.py", line 160, in run_apt_command
Apr 05 19:16:28 ip-172-31-43-230 python3[1728]:     raise exceptions.UserFacingError(error_msg)
Apr 05 19:16:28 ip-172-31-43-230 python3[1728]: uaclient.exceptions.UserFacingError: Could not enable FIPS.
Apr 05 19:16:28 ip-172-31-43-230 python3[1728]: Failed to install packages at boot: fips-initramfs, libssl1.0.0, libssl1.0.0-hmac, linux-aws-fips, openssh-client, openssh
Apr 05 19:16:28 ip-172-31-43-230 python3[1728]: Updating package lists
Apr 05 19:16:28 ip-172-31-43-230 python3[1728]: Installing FIPS packages
Apr 05 19:16:28 ip-172-31-43-230 systemd[1]: ua-reboot-cmds.service: Main process exited, code=exited, status=1/FAILURE
Apr 05 19:16:28 ip-172-31-43-230 systemd[1]: Failed to start Ubuntu Advantage reboot cmds.
Apr 05 19:16:28 ip-172-31-43-230 systemd[1]: ua-reboot-cmds.service: Unit entered failed state.
Apr 05 19:16:28 ip-172-31-43-230 systemd[1]: ua-reboot-cmds.service: Failed with result 'exit-code'.
```

PR #1533, opened 2021-04-07
lib: fix notice handling on reboot script

lucasmoura:
## Proposed Commit Message
lib: fix notice handling on reboot script

When the reboot script fails, we keep the notice that show that it is running. This happens because the assert_lock_file decorator cannot complete and remove the notice. We are now fixing that by catching the exceptions in the reboot script. Furthermore, we also issue a notice if the reboot script fails for some reason.

Fixes: #1518

## Test Steps
Run the new unit test on this PR

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -245,6 +245,9 @@ Ubuntu Advantage server provided no aptURL directive for {entitlement_name}"""
 MESSAGE_NO_ACTIVE_OPERATIONS = """No Ubuntu Advantage operations are running"""
 MESSAGE_LOCK_HELD = """Operation in progress: {lock_holder} (pid:{pid})"""
 PROMPT_YES_NO = """Are you sure? (y/N) """
+MESSAGE_REBOOT_SCRIPT_FAILED = (
+    "Failed running reboot_cmds script. See: /var/log/ubuntu-advantage.log"
+)
 MESSAGE_LIVEPATCH_LTS_REBOOT_REQUIRED = (
     "Livepatch support requires a system reboot across LTS upgrade."
 )
--- a/uaclient/tests/test_reboot_cmds.py
+++ b/uaclient/tests/test_reboot_cmds.py
@@ -2,9 +2,15 @@ import logging
 import mock
 import pytest
 
+from uaclient.status import MESSAGE_REBOOT_SCRIPT_FAILED
 from uaclient.util import ProcessExecutionError
 
-from lib.reboot_cmds import main, fix_pro_pkg_holds, run_command
+from lib.reboot_cmds import (
+    main,
+    fix_pro_pkg_holds,
+    run_command,
+    process_reboot_operations,
+)
 
 M_FIPS_PATH = "uaclient.entitlements.fips.FIPSEntitlement."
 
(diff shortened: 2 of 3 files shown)


uaclient/status.py
```diff
@@ -245,6 +245,9 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_NO_ACTIVE_OPERATIONS = """No Ubuntu Advantage operations are running"""
 MESSAGE_LOCK_HELD = """Operation in progress: {lock_holder} (pid:{pid})"""
 PROMPT_YES_NO = """Are you sure? (y/N) """
+MESSAGE_REBOOT_SCRIPT_FAILED = (
+    "Failed running reboot_cmds script.\n" "See: /var/log/ubuntu-advantage.log"
```
blackboxsw, line 249: Why the separation of sub-strings?

lucasmoura: My mistake here, I thought the string was going to be grater than 80 characters and it is not. I will update the code

uaclient/tests/test_reboot_cmds.py
```diff
@@ -122,3 +128,40 @@ def test_run_command_failure(self, m_subp, m_exit, caplog_text):
...
+        cfg = FakeConfig()
+        cfg.for_attached_machine()
+        with mock.patch("os.path.exists", return_value=True):
+            with mock.patch("uaclient.config.UAConfig.write_cache"):
+                process_reboot_operations(args, cfg)
+
+        assert (
+            mock.call("", MESSAGE_REBOOT_SCRIPT_FAILED)
+            in m_add_notice.call_args_list
+        )
```
orndorffgrant, line 161: Based on the issue description, I would also expect an assertion similar to this one but checking that `remove_notice` was called for the "Operation in progress..." notice. Or perhaps a check that the lock was released properly.

orndorffgrant (commented):
Looking good! Just did a code review here and ran the tests. I'll actually run it tomorrow.

orndorffgrant:
@lucasmoura Interestingly, all tests pass when I run `tox`, but when the tests run as a part of building the package with a `dpkg-buildpackage -us -uc`, I get this failure for the new test:
```
>       assert expected_msg in caplog_text()
E       assert "Invalid command specified 'error'.\nFailed running commands on reboot." in "util.py                    627 DEBUG    Writing file: /tmp/pytest-of-grant/pytest-7/test_process_reboot_operations0/p...UG    Invalid command specified 'error'.\n                                        Failed running commands on reboot.\n"
E        +  where "util.py                    627 DEBUG    Writing file: /tmp/pytest-of-grant/pytest-7/test_process_reboot_operations0/p...UG    Invalid command specified 'error'.\n                                        Failed running commands on reboot.\n" = <function caplog_text.<locals>._func at 0x7f76daa2b430>()
```

In this case it looks like there are a lot of spaces/indentation before each line which is preventing the `in` from returning true.

lucasmoura:
@orndorffgrant great catch. I have update the test to address those potential extra spaces in the log

uaclient/status.py
```diff
@@ -245,6 +245,9 @@ class UserFacingStatus(enum.Enum):
 MESSAGE_NO_ACTIVE_OPERATIONS = """No Ubuntu Advantage operations are running"""
 MESSAGE_LOCK_HELD = """Operation in progress: {lock_holder} (pid:{pid})"""
 PROMPT_YES_NO = """Are you sure? (y/N) """
+MESSAGE_REBOOT_SCRIPT_FAILED = (
+    "Failed running reboot_cmds script.\nSee: /var/log/ubuntu-advantage.log"
```
blackboxsw, line 249: Probably drop the newline here because this can easily fit on one-line in the NOTICES output.
```suggestion
    "Failed running reboot_cmds script. See: /var/log/ubuntu-advantage.log"
```

blackboxsw (approved):
Minor nit on debug logs and +1 otherwise. Thanks lucas.

blackboxsw (approved):
Minor nit on debug logs and +1 otherwise. Thanks lucas/

orndorffgrant (approved):
Thanks for the minor fixes! Its building for me now :+1:

lucasmoura:
@blackboxsw code updated

PR #1534, opened 2021-04-08
d/control: allow hirsute to build correctly

lucasmoura:
## Proposed Commit Message
d/control: allow hirsute to build correctly

Hirsute does not have the dh-systemd dependency. That dependency is already provided by debhelper in that release. To fix that, we are using an or clause in the control file to guarantee that dh-systemd will be installed on gropvy and earlier but not on hirsute and later

## Test Steps
Build the package for all the releases and see that they are correctly built

## Desired commit type::

## Checklist:


blackboxsw (approved):
Thanks for this @lucasmoura works great.

Issue #1529, opened 2021-04-07
postinst: add unauthenticate esm-apps sources to /etc/apt/sources.list.d/ubuntu-esm-apps.list


blackboxsw:
As ua client did on trusty, we need to enable visibility to esm-apps on Xenial and Later in postinst.
Content should be 
```bash
deb https://esm.ubuntu.com/apps/ubuntu xenial-apps-security main
# deb-src https://esm.ubuntu.com/apps/ubuntu xenial-apps-security main
deb https://esm.ubuntu.com/apps/ubuntu xenial-apps-updates main
# deb-src https://esm.ubuntu.com/apps/ubuntu xenial-apps-updates main
```

PR #1535, opened 2021-04-08
d/postinst: xenial enable unauthenticated apt source for apps/infra

blackboxsw:
Allow postinstall to pre-enable visibility to unauthenticate esm-apps and esm-infra repos via adding apt sources  files and Pin-Priority: never in /etc/apt/preferences.d/. 

Also add  python logic to uaclient/entitlements/esm.py for ESMInfra and ESMApps to support re-setting the pin-prirority: never when esm-infra or esm-apps is disabled on xenial.

--- postinst commit
Refactor a couple of functions in ubuntu-advantage-tools.postinst
to a install_esm_apt_key_and_source which will reuse common logic
to emit unauthenticated APT sources.list.d files for both esm-apps
and esm-infra.

Once APT sources are on disk, emit a APT preferences file that pins
the repo origins as Pin-Priority: never which tells apt to list
those packages as upgradeable, but not attempt to install or upgrade
during typical apt-get upgrade|dist-upgrade.

Ran shellcheck on this content and fixed up the suggestions
thoughout the file.

Fixes: #1529

--- uacllient.entitlements. esm.py commit 
    esm: add pin never when disabling esm-infra/apps on xenial via repo_pin_priority and disable_apt_auth_only property overrides

## Proposed Commit Message
Rebase and merge 

## Test Steps
Integration test should cover the artifacts leaving a magically disabled esm-infra/esm-apps APT config on unattached systems.

Otherwise one could;
```bash
sed -i 's/UNRELEASED/xenial/ debian/changelog
git commit -am 'changelog'
build-package
sbuild-it ../out/ubuntu-advantage-tools_27.0.dsc
lxc launch ubuntu-daily:xenial test-x-postinst
lxc file push ubuntu-advantage-tools_27.0_amd64.deb test-x-postinst/

# Should not see esm APT artifacts or policy before install
lxc exec test-x-postinst -- find /etc/apt | grep esm
lxc exec test-x-postinst -- apt policy | grep esm
lxc exec test-x-postinst -- dpkg -i /ubuntu-advantage-tools_27.0_amd64.deb
# Should see esm APT artifacts or policy post install
lxc exec test-x-postinst -- find /etc/apt | grep esm
lxc exec test-x-postinst -- apt policy | grep esm

# check `apt update` output related to ESM Apps when enabled and disabled
apt update..
 82 packages can be upgraded. Run 'apt list --upgradable' to see them.
9 of the updates are from UA Apps: ESM.

```

## Desired commit type::

## Checklist:


debian/ubuntu-advantage-tools.postinst
```diff
@@ -12,17 +12,20 @@ APT_TRUSTED_KEY_DIR="/etc/apt/trusted.gpg.d"
...
 ESM_INFRA_KEY_TRUSTY="ubuntu-advantage-esm-infra-trusty.gpg"
+ESM_APPS_KEY="ubuntu-advantage-esm-apps.gpg"
 
 APT_SRC_DIR="/etc/apt/sources.list.d"
 APT_PREFERENCES_DIR="/etc/apt/preferences.d"
 ESM_APT_SOURCE_FILE_PRECISE="$APT_SRC_DIR/ubuntu-esm-precise.list"
 ESM_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-trusty.list"
 ESM_INFRA_OLD_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-infra-trusty.list"
-ESM_INFRA_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-infra.list"
+ESM_INFRA_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-esm-infra.list"
```
blackboxsw, line 22: Note I renamed this variable ESM_INFRA_APT_SOURCE_FILE_TRUSTY -> ESM_INFRA_APT_SOURCE_FILE because it didnt contain "trusty" in the source file name. It was a rename oversight when we decided to rename source files to be ubuntu release-agnostic

debian/ubuntu-advantage-tools.postinst
```diff
@@ -12,17 +12,20 @@ APT_TRUSTED_KEY_DIR="/etc/apt/trusted.gpg.d"
...
+ESM_APPS_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-esm-apps.list"
 
-ESM_APT_PREF_FILE_TRUSTY="/etc/apt/preferences.d/ubuntu-esm-trusty"
-ESM_INFRA_OLD_APT_PREF_FILE_TRUSTY="/etc/apt/preferences.d/ubuntu-esm-infra-trusty"
-ESM_INFRA_APT_PREF_FILE_TRUSTY="/etc/apt/preferences.d/ubuntu-esm-infra"
```
blackboxsw, line 25: This lines come into play because these varis should have been sourcing the common $APT_PREFERENCES_DIR defined above. no functional change

debian/ubuntu-advantage-tools.postinst
```diff
@@ -95,45 +98,96 @@ if status:
 
 unconfigure_esm() {
     if ! check_service_is_enabled esm-infra; then
-        rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove previous esm keys
-        rm -f $APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY_TRUSTY
-        rm -f $ESM_INFRA_APT_SOURCE_FILE_TRUSTY
-        rm -f $ESM_INFRA_OLD_APT_SOURCE_FILE_TRUSTY
-        rm -f $ESM_APT_PREF_FILE_TRUSTY $ESM_INFRA_OLD_APT_PREF_FILE_TRUSTY
-        rm -f $ESM_INFRA_APT_PREF_FILE_TRUSTY
```
blackboxsw, line 103: only quoting "" variable changes recommended by shellcheck

debian/ubuntu-advantage-tools.postinst
```diff
@@ -95,45 +98,96 @@ if status:
...
+        cp $UA_KEYRING_DIR/$apt_key $APT_TRUSTED_KEY_DIR
     fi
 
-    if [ -e "$ESM_APT_SOURCE_FILE_TRUSTY" ]; then
-        mv $ESM_APT_SOURCE_FILE_TRUSTY $ESM_INFRA_APT_SOURCE_FILE_TRUSTY
-    fi
-    if [ -e "$ESM_APT_PREF_FILE_TRUSTY" ]; then
-        mv $ESM_APT_PREF_FILE_TRUSTY $ESM_INFRA_APT_PREF_FILE_TRUSTY
+    # Migrate trusty legacy source list and preference file names
+    if [ "14.04" = "$VERSION_ID" ]; then
```
blackboxsw, line 141: Don't need these renamed on any other Ubuntu release, beccause the only version of UA client that surfaced this was trusty.

debian/ubuntu-advantage-tools.postinst
```diff
@@ -95,45 +98,96 @@ if status:
...
-        rm -f $ESM_INFRA_APT_PREF_FILE_TRUSTY
+        rm -f "$APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg"  # Remove previous esm keys
+        rm -f "$APT_TRUSTED_KEY_DIR/$ESM_INFRA_KEY_TRUSTY"
+        rm -f "$ESM_INFRA_APT_SOURCE_FILE"
+        rm -f "$ESM_INFRA_OLD_APT_SOURCE_FILE_TRUSTY"
+        rm -f "$ESM_APT_PREF_FILE_TRUSTY" "$ESM_INFRA_OLD_APT_PREF_FILE_TRUSTY"
+        rm -f "$ESM_INFRA_APT_PREF_FILE"
+    fi
+    if ! check_service_is_enabled esm-apps; then
+        rm -f "$APT_TRUSTED_KEY_DIR/$ESM_APPS_KEY_TRUSTY"
```
lucasmoura, line 109: Maybe I missing something, but it seems that `ESM_APPS_KEY_TRUSTY` doesn't exist

debian/ubuntu-advantage-tools.postinst
```diff
@@ -95,45 +98,96 @@ if status:
...
+
+}
+
+configure_esm() {
+    rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove legacy esm keys
+    case "${VERSION_ID}" in
+        14.04)
+            enable_apps="false" enable_infra="true" release="trusty" ;;
+        16.04)
+            enable_apps="false" enable_infra="true" release="xenial" ;;
```
lucasmoura, line 179: Shouldn't `enable_apps` here be true ?

debian/ubuntu-advantage-tools.postinst
blackboxsw: +1 I've fixed this.

debian/ubuntu-advantage-tools.postinst
blackboxsw: and yes, I had noted that when preparing the integration tests. Good catch

debian/ubuntu-advantage-tools.postinst
```diff
@@ -95,45 +98,97 @@ if status:
...
+            apt_origin="UbuntuESMApps"
+            apt_pref_file=${ESM_APPS_APT_PREF_FILE};
+            apt_source_file=${ESM_APPS_APT_SOURCE_FILE};
+            apt_key=${ESM_APPS_KEY};
+            ;;
+        infra)
+            apt_origin="UbuntuESM"
+            apt_pref_file=${ESM_INFRA_APT_PREF_FILE};
+            apt_source_file=${ESM_INFRA_APT_SOURCE_FILE};
+            apt_key=${ESM_INFRA_KEY_TRUSTY};
```
orndorffgrant, line 132: It seems odd to me that this key has "trusty" in the name but we are using it for all esm-infra, including xenial now.

blackboxsw: Yes totally agree. We had opted in this particular case to leave the key name unaltered in trusty release time-frame to avoid unnecessary risk (minimal) if the postinst renaming improperly. Historically APT repos originally were going to have a different gpg key to control the repo for each Ubuntu release. Subsequent iterations dropped that desire as unmaintainable.
We could rename here as well dropping the -trusty suffix, we would also have to make sure to handle upgrade paths for previous ua-tools installs to current and make sure we don't leave a stale gpg keyring artifact around across upgrade path.

blackboxsw:
Only CI failure is intermittent "transport:paramiko.ssh_exception.SSHException: Error reading SSH protocol banner"

lucasmoura (approved):

debian/ubuntu-advantage-tools.postinst
orndorffgrant: Ah that makes sense - thanks for the context. This is pretty minor, and I defer to you on whether its worth fitting in now or putting in the backlog for next release.

blackboxsw: Thanks @orndorffgrant I added tech-debt for https://github.com/canonical/ubuntu-advantage-client/issues/1539

PR #1540, opened 2021-04-09
apt-hook: print static messages on apt upgrade/dist-upgrade

orndorffgrant:
## New behavior

* on `apt upgrade/dist-upgrade`
  * find esm packages (names and count)
  * will insert values into message template for contract-expiry message if the template exists and print the message
* when called as `hook process-templates`
  * find esm packages (names and count)
  * will insert values into hardcoded message templates that exist
  * outputs those messages to static files with the values filled in
* on `apt update`
  * updates all the template files, but no longer prints the messages it used to

## Proposed Commit Message

> apt-hook: process template files and print messages with esm stats
>
> apt-esm-hook will now process a set of .tmpl files in the ua messages
> directory, both when apt calls it and when called separately with the
> "process-templates" argument. It replaces variables in the tmpl files
> with the values of ESM Apps/Infra package names and count. It then
> writes the result of the replacement back to the ua messages
> directory.
>
> When called on Post-Invoke for upgrade/dist-upgrade, the hook will
> print out the contents of message files if they are present.
>
> Fixes: #1546 

## Test Steps

In a new xenial lxd container, preferably one with outstanding esm updates:
1. `mkdir -p /var/lib/ubuntu-advantage/messages`
2. `echo "esm-apps {ESM_APPS_PKG_COUNT} {ESM_APPS_PACKAGES}; esm-infra {ESM_INFRA_PKG_COUNT} {ESM_INFRA_PACKAGES}" > /var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade.tmpl`
3. `/usr/lib/ubuntu-advantage/apt-esm-hook process-templates`
4. `cat /var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade` and make sure it has the expected output
5. `apt upgrade` and ensure expected output appears

## Desired commit type::

## Checklist:

### DIFF
--- a/apt-hook/Makefile
+++ b/apt-hook/Makefile
@@ -6,7 +6,7 @@ ubuntu-advantage.pot: hook.cc
 	xgettext hook.cc -o ubuntu-advantage.pot
 
 hook: hook.cc
-	$(CXX) -Wall -Wextra -pedantic $(CXXFLAGS) $(CPPFLAGS) $(LDFLAGS) -g -o hook hook.cc -lapt-pkg $(LDLIBS)
+	$(CXX) -Wall -Wextra -pedantic -std=c++11 $(CXXFLAGS) $(CPPFLAGS) $(LDFLAGS) -g -o hook hook.cc -lapt-pkg $(LDLIBS)
 
 install: hook
 	install -D -m 644 20apt-esm-hook.conf $(DESTDIR)/etc/apt/apt.conf.d/20apt-esm-hook.conf
--- a/apt-hook/hook.cc
+++ b/apt-hook/hook.cc
@@ -34,16 +34,16 @@
 #include <libintl.h>
 #include <locale.h>
 
-#define CONTRACT_EXPIRY_STATUS_MESSAGE_PATH "/var/lib/ubuntu-advantage/messages/contract-expiry-status"
-
-#define CONTRACT_EXPIRED_APT_UPGRADE_MESSAGE_TEMPLATE_PATH "/var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade.tmpl"
-#define CONTRACT_EXPIRED_APT_UPGRADE_MESSAGE_STATIC_PATH "/var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade"
-#define CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_TEMPLATE_PATH "/var/lib/ubuntu-advantage/messages/contract-expired-apt-dist-upgrade.tmpl"
-#define CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_STATIC_PATH "/var/lib/ubuntu-advantage/messages/contract-expired-apt-dist-upgrade"
-#define ESM_APPS_NOT_ENABLED_MESSAGE_TEMPLATE_PATH "/var/lib/ubuntu-advantage/messages/esm-apps-not-enabled.tmpl"
-#define ESM_APPS_NOT_ENABLED_MESSAGE_STATIC_PATH "/var/lib/ubuntu-advantage/messages/esm-apps-not-enabled"
-#define ESM_INFRA_NOT_ENABLED_MESSAGE_TEMPLATE_PATH "/var/lib/ubuntu-advantage/messages/esm-infra-not-enabled.tmpl"
-#define ESM_INFRA_NOT_ENABLED_MESSAGE_STATIC_PATH "/var/lib/ubuntu-advantage/messages/esm-infra-not-enabled"
+#define CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE_PATH              "/var/lib/ubuntu-advantage/messages/contract-expiry-status.tmpl"
+#define CONTRACT_EXPIRY_STATUS_MESSAGE_STATIC_PATH                "/var/lib/ubuntu-advantage/messages/contract-expiry-status"
+#define CONTRACT_EXPIRED_APT_UPGRADE_MESSAGE_TEMPLATE_PATH        "/var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade.tmpl"
+#define CONTRACT_EXPIRED_APT_UPGRADE_MESSAGE_STATIC_PATH          "/var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade"
+#define CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_TEMPLATE_PATH   "/var/lib/ubuntu-advantage/messages/contract-expired-apt-dist-upgrade.tmpl"
+#define CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_STATIC_PATH     "/var/lib/ubuntu-advantage/messages/contract-expired-apt-dist-upgrade"
+#define ESM_APPS_NOT_ENABLED_MESSAGE_TEMPLATE_PATH                "/var/lib/ubuntu-advantage/messages/esm-apps-not-enabled.tmpl"
+#define ESM_APPS_NOT_ENABLED_MESSAGE_STATIC_PATH                  "/var/lib/ubuntu-advantage/messages/esm-apps-not-enabled"
+#define ESM_INFRA_NOT_ENABLED_MESSAGE_TEMPLATE_PATH               "/var/lib/ubuntu-advantage/messages/esm-infra-not-enabled.tmpl"
+#define ESM_INFRA_NOT_ENABLED_MESSAGE_STATIC_PATH                 "/var/lib/ubuntu-advantage/messages/esm-infra-not-enabled"
 
 #define ESM_APPS_PKGS_COUNT_TEMPLATE_VAR "{ESM_APPS_PKG_COUNT}"
 #define ESM_APPS_PACKAGES_TEMPLATE_VAR "{ESM_APPS_PACKAGES}"
@@ -212,7 +212,9 @@ static void process_template_file(
 ) {
    std::ifstream message_tmpl_file(template_file_name.c_str());
    if (message_tmpl_file.is_open()) {
+      // This line loads the whole file contents into a string
       std::string message_tmpl((std::istreambuf_iterator<char>(message_tmpl_file)), (std::istreambuf_iterator<char>()));
+
       message_tmpl_file.close();
 
       // Process all template variables
@@ -332,13 +334,15 @@ int main(int argc, char *argv[])
 
    // Execute specified subcommand
    if (subcommand == ProcessTemplates) {
-      std::array<std::string, 4> template_file_names = {
+      std::array<std::string, 5> template_file_names = {
+         CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE_PATH,
          CONTRACT_EXPIRED_APT_UPGRADE_MESSAGE_TEMPLATE_PATH,
          CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_TEMPLATE_PATH,
          ESM_APPS_NOT_ENABLED_MESSAGE_TEMPLATE_PATH,
          ESM_INFRA_NOT_ENABLED_MESSAGE_TEMPLATE_PATH
       };
-      std::array<std::string, 4> static_file_names = {
+      std::array<std::string, 5> static_file_names = {
+         CONTRACT_EXPIRY_STATUS_MESSAGE_STATIC_PATH,
          CONTRACT_EXPIRED_APT_UPGRADE_MESSAGE_STATIC_PATH,
          CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_STATIC_PATH,
          ESM_APPS_NOT_ENABLED_MESSAGE_STATIC_PATH,
@@ -408,12 +412,15 @@ int main(int argc, char *argv[])
       }
    } else if (subcommand == PreInvoke) {
       if (command_used == "upgrade" || command_used == "dist-upgrade") {
-         // Print static message if present. Used for "Expiring soon" and "Expired but in grace period" messages.
-         std::ifstream message_file(CONTRACT_EXPIRY_STATUS_MESSAGE_PATH);
-         if (message_file.is_open()) {
-            std::cout << message_file.rdbuf();
-            message_file.close();
-         }
+         // Try expiry status message
+         process_template_file(
+            CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE_PATH,
+            CONTRACT_EXPIRY_STATUS_MESSAGE_STATIC_PATH,
+            esm_a_packages_count,
+            space_separated_esm_a_packages,
+            esm_i_packages_count,
+            space_separated_esm_i_packages
+         );
 
          // Try not enabled messages
          process_template_file(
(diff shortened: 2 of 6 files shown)


apt-hook/hook.cc
```diff
@@ -271,5 +277,32 @@ int main(int argc, char *argv[])
       ioprintf(std::cout, "\n");
    }
 
+
+   if (hook == "apt-install-pre-invoke") {
```
blackboxsw, line 281: might be able to parametrize this hook name and use it to dynamically source the /var/lib/ubuntu-advantage/<hook_name>-message file if present. So we might not need to touch this much if we want to make it a simple pass through for any custom apt-hook messaging based on the arg1 passed to apt-esm-hook

apt-hook/hook.cc
```diff
@@ -201,69 +282,230 @@ int main(int argc, char *argv[])
...
+      };
+      for (uint i = 0; i < template_file_names.size(); i++) {
+         process_template_file(
+            template_file_names[i], 
+            static_file_names[i], 
+            esm_a_packages_count,
+            space_separated_esm_a_packages,
+            esm_i_packages_count,
+            space_separated_esm_i_packages
+         );
```
orndorffgrant, line 355: Here is the process-templates sub command. Let me know if I'm missing any templates that will need to be processed.

apt-hook/hook.cc
```diff
@@ -201,69 +282,230 @@ int main(int argc, char *argv[])
...
+      }
+   } else if (subcommand == PostInvokeStats || subcommand == PostInvokeSuccess || test_run) {
+      // TODO this was the existing functionality - how much of this are we keeping?
```
orndorffgrant, line 358: Just calling out this comment. This should work the same as before currently for apt update

apt-hook/hook.cc
```diff
@@ -201,69 +282,230 @@ int main(int argc, char *argv[])
...
+         ioprintf(std::cout, gettext("See https://ubuntu.com/advantage or run: sudo ua status"));
          ioprintf(std::cout, "\n");
       }
-
-      ioprintf(std::cout, gettext("To see these additional updates run: apt list --upgradable"));
-      ioprintf(std::cout, "\n");
-      ioprintf(std::cout, gettext("See https://ubuntu.com/advantage or run: sudo ua status"));
-      ioprintf(std::cout, "\n");
+   } else if (subcommand == PreInvoke) {
+      if (command_used == "upgrade" || command_used == "dist-upgrade") {
```
orndorffgrant, line 410: Here is the main new message output functionality.

Note the different files it is currently looking at and in what situations. Let me know if this doesn't jive with your understanding of what this should be doing.

apt-hook/hook.cc
orndorffgrant: We partially discussed this out-of-band yesterday - posting here for posterity.

This is a very good point! It might not be doable because this hook needs to internally calculate which template file to use in at least one circumstance (`apt upgrade` vs `apt dist-upgrade` messages). If it turns out it is doable, we should do it.

apt-hook/Makefile
```diff
@@ -6,7 +6,7 @@ ubuntu-advantage.pot: hook.cc
 	xgettext hook.cc -o ubuntu-advantage.pot
 
 hook: hook.cc
-	$(CXX) -Wall -Wextra -pedantic $(CXXFLAGS) $(CPPFLAGS) $(LDFLAGS) -g -o hook hook.cc -lapt-pkg $(LDLIBS)
+	$(CXX) -Wall -Wextra -pedantic -std=c++11 $(CXXFLAGS) $(CPPFLAGS) $(LDFLAGS) -g -o hook hook.cc -lapt-pkg $(LDLIBS)
```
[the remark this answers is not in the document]
blackboxsw: hrm does this mean a package build depedency  on libc++abi1-11? If so, we need to check rmadison
```bash
$  rmadison libc++abi1-11
 libc++abi1-11 | 1:11.0.0-2~ubuntu20.04.1 | focal-updates/universe | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
 libc++abi1-11 | 1:11.0.0-2               | groovy/universe        | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
 libc++abi1-11 | 1:11.0.1-2ubuntu2        | hirsute/universe       | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
```

I see no C++ 11 in Xenial or trusty, so that would mean we can no longer build this hook on either right?
We can test w/ `sed -i s/UNRELEASED/xenial' debian/changelog; git commit -am 'wip'; build-package; sbuild-it ../out*27.0.dsc;

apt-hook/hook.cc
```diff
@@ -121,23 +144,34 @@ static void check_esm_upgrade(pkgCache::PkgIterator pkg, pkgPolicy *policy, resu
...
-	    else
-	       res.enabled_esms_i++;
-	 }
+            {
+               res.disabled_esms_i++;
+            }
+            else
+            {
+               res.enabled_esms_i++;
+            }
```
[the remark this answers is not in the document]
blackboxsw: Cleaner, common styling, more friendly for python eyes. +1!

apt-hook/hook.cc
```diff
@@ -167,6 +201,50 @@ static int get_update_count(result &res)
...
+   std::string template_file_name,
+   std::string static_file_name,
+   std::string esm_a_pkgs_count,
+   std::string esm_a_pkgs,
+   std::string esm_i_pkgs_count,
+   std::string esm_i_pkgs
+) {
+   std::ifstream message_tmpl_file(template_file_name.c_str());
+   if (message_tmpl_file.is_open()) {
+      std::string message_tmpl((std::istreambuf_iterator<char>(message_tmpl_file)), (std::istreambuf_iterator<char>()));
```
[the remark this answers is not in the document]
blackboxsw: thanks for that note here.  let's make it a leading comment in code to remind future us.

apt-hook/hook.cc
```diff
@@ -201,69 +282,230 @@ int main(int argc, char *argv[])
...
-      ioprintf(std::cout, gettext("See https://ubuntu.com/advantage or run: sudo ua status"));
-      ioprintf(std::cout, "\n");
+   } else if (subcommand == PreInvoke) {
+      if (command_used == "upgrade" || command_used == "dist-upgrade") {
+         // Print static message if present. Used for "Expiring soon" and "Expired but in grace period" messages.
+         std::ifstream message_file(CONTRACT_EXPIRY_STATUS_MESSAGE_PATH);
+         if (message_file.is_open()) {
+            std::cout << message_file.rdbuf();
+            message_file.close();
+         }
```
blackboxsw, line 416: @orndorffgrant I was expecting this to behave the same way as other templates: lib/ua_update_messages will emit `/var/lib/ubuntu-advantage/messages/contract-expiry-status.tmpl` and expect apt hook will update and write `/var/lib/ubuntu-advantage/messages/contract-expiry-status`

```suggestion
         process_template_file(
            CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE_PATH,
            CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE,
            esm_a_packages_count,
            space_separated_esm_a_packages,
            esm_i_packages_count,
            space_separated_esm_i_packages
         );
```

apt-hook/hook.cc
```diff
@@ -201,69 +282,230 @@ int main(int argc, char *argv[])
...
+               CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_STATIC_PATH,
+               esm_a_packages_count,
+               space_separated_esm_a_packages,
+               esm_i_packages_count,
+               space_separated_esm_i_packages
+            );
+         }
+
+         /*
+         // TODO remove
```
blackboxsw, line 458: +1 on removal of this, we can intrument the various message .tmpl files you need in /v/l/ubuntu-advantage/messages and make sure hook can be a simple string formatter that creates final message files.

blackboxsw (changes requested):
Looks good @orndorffgrant agreed on your approach and a minor ask on contract-expiry. We also might have to sort the c++11 flag though.

apt-hook/hook.cc
orndorffgrant: Yep that makes sense! Then this won't be a special case.

apt-hook/hook.cc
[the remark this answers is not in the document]
orndorffgrant: Good idea!

apt-hook/Makefile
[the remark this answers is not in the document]
orndorffgrant: This did build successfully for me. But I also understand if we'd rather play it safe and remove the dependency. I can replace `std::array` with `std::vector` and replace `std::to_string` with some `stringstream` stuff - its not hard, just not as clean-looking.

apt-hook/hook.cc
```diff
@@ -201,69 +284,183 @@ int main(int argc, char *argv[])
...
+            static_file_names[i], 
+            esm_a_packages_count,
+            space_separated_esm_a_packages,
+            esm_i_packages_count,
+            space_separated_esm_i_packages
+         );
+      }
+   } else if (subcommand == PostInvokeStats || subcommand == PostInvokeSuccess || test_run) {
+      // TODO this was the existing functionality - how much of this are we keeping?
+      if (command_used == "update")
```
lucasmoura, line 363: Looking at the spec, it seems we will not have any more message on update. But I think we can double check that

blackboxsw: @orndorffgrant @lucasmoura I think we definitely need Lech and Rick to confirm this. Lech's last comment on the spec in side-channel was to go4it with ESM messaging if available. But the spec text suggestions didn't get updated/confirmed.

blackboxsw: Let's plan on adding these for `apt update` in this PR, we can drop that functoinality easily if we need to.

orndorffgrant: I went ahead and removed in this PR per our conversation earlier

apt-hook/Makefile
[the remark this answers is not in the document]
blackboxsw: @orndorffgrant thanks I confirmed too package builds on trusty/xenial succeed and the binary gerated works. So as long as we have integration tests covering the template rendering output we should be good on this as is.

features/staging_commands.feature
```diff
@@ -46,6 +48,32 @@ Feature: Enable command behaviour when attached to an UA staging subscription
         \s*\*\*\* .* 500
         \s*500 https://esm.staging.ubuntu.com/apps/ubuntu <release>-apps-security/main amd64 Packages
         """
+        When I run `mkdir -p /var/lib/ubuntu-advantage/messages` with sudo
+        When I run `sh -c 'echo "esm-apps {ESM_APPS_PKG_COUNT} {ESM_APPS_PACKAGES}; esm-infra {ESM_INFRA_PKG_COUNT} {ESM_INFRA_PACKAGES}" > /var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade.tmpl'` with sudo
+        When I run `/usr/lib/ubuntu-advantage/apt-esm-hook process-templates` with sudo
+        When I run `cat /var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade` with sudo
```
orndorffgrant, line 54: This command is failing, which presumably means the apt-hook line is failing to output the file, but I can't recreate in manual testing.

orndorffgrant: After rebooting my machine this error disappeared

apt-hook/hook.cc
blackboxsw: Thanks yes appreciated.

apt-hook/hook.cc
```diff
@@ -24,19 +24,42 @@
...
 #include <sys/stat.h>
 #include <libintl.h>
 #include <locale.h>
 
+#define CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE_PATH              "/var/lib/ubuntu-advantage/messages/contract-expiry-status.tmpl"
+#define CONTRACT_EXPIRY_STATUS_MESSAGE_STATIC_PATH                "/var/lib/ubuntu-advantage/messages/contract-expiry-status"
+#define CONTRACT_EXPIRED_APT_UPGRADE_MESSAGE_TEMPLATE_PATH        "/var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade.tmpl"
+#define CONTRACT_EXPIRED_APT_UPGRADE_MESSAGE_STATIC_PATH          "/var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade"
+#define CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_TEMPLATE_PATH   "/var/lib/ubuntu-advantage/messages/contract-expired-apt-dist-upgrade.tmpl"
+#define CONTRACT_EXPIRED_APT_DIST_UPGRADE_MESSAGE_STATIC_PATH     "/var/lib/ubuntu-advantage/messages/contract-expired-apt-dist-upgrade"
```
blackboxsw, line 42: @orndorffgrant /var/lib/ubuntu-advantage/messages/contract-expiry-status.tmpl will contain the following
```bash
CAUTION: Your UA Apps: ESM service will expire in 10 days.        
Renew UA subscription at https://ubuntu.com/advantage to ensure                                        
continued security coverage for your applications.
```

What messages to you forsee as content in /var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade.tmpl and /var/lib/ubuntu-advantage/messages/contract-expired-apt-dist-upgrade.tmpl that differs?

orndorffgrant: "/var/lib/ubuntu-advantage/messages/contract-expiry-status.tmpl" -> https://docs.google.com/document/d/1bNvQ2EveGdGyMqs9WxoLEt1U-x9nqEbUBtdNzIpGAXE/edit#heading=h.aj8kpel9nomu

These next two aren't necessary if we don't actually need different messages per upgrade/dist-upgrade
"/var/lib/ubuntu-advantage/messages/contract-expired-apt-upgrade.tmpl" -> see apt upgrade message here https://docs.google.com/document/d/1bNvQ2EveGdGyMqs9WxoLEt1U-x9nqEbUBtdNzIpGAXE/edit#heading=h.fv9w5coai2c1
"/var/lib/ubuntu-advantage/messages/contract-expired-apt-dist-upgrade.tmpl" -> see apt dist-upgrade message here https://docs.google.com/document/d/1bNvQ2EveGdGyMqs9WxoLEt1U-x9nqEbUBtdNzIpGAXE/edit#heading=h.fv9w5coai2c1

Not sure if these next two are necessary. We might be able to use just one banner message here. If we keep them separate though, it gives us the opportunity to only print if esm_apps_count > 0 for example (not currently implemented)
"/var/lib/ubuntu-advantage/messages/esm-apps-not-enabled.tmpl" -> apps version of https://docs.google.com/document/d/1bNvQ2EveGdGyMqs9WxoLEt1U-x9nqEbUBtdNzIpGAXE/edit#heading=h.jcfd4jpjjhqv
"/var/lib/ubuntu-advantage/messages/esm-infra-not-enabled.tmpl" -> infra version of https://docs.google.com/document/d/1bNvQ2EveGdGyMqs9WxoLEt1U-x9nqEbUBtdNzIpGAXE/edit#heading=h.jcfd4jpjjhqv

orndorffgrant:
@blackboxsw Alright, this is simplified to just `contract-expiry-status.tmpl` and `contract-expiry-status-apt.tmpl` and should be good to go pending jenkins

features/attached_enable.feature
```diff
@@ -90,41 +90,43 @@ Feature: Enable command behaviour when attached to an UA subscription
...
-           | release | pkg-version            |
-           | bionic  | libkrad0=1.16-2build1  |
-           | focal   | hello=2.10-2ubuntu2    |
-           | xenial  | libkrad0=1.13.2+dfsg-5 |
+# TODO Add back equivalent once these messages are in apt upgrade
+# Will be something like: "6 upgraded, including 4 esm-apps security updates and 1 esm-infra update"
+#    @series.xenial
+#    @series.bionic
+#    @series.focal
+#    Scenario Outline: Attached enable of a know service shows update in a ubuntu machine
```
orndorffgrant, line 98: Bringing this new change to your attention @blackboxsw and @lucasmoura 
I removed the apt update messages that this is testing, but we will be adding similar functionality to apt upgrade soon, so I just commented this out for now.

apt-hook/hook.cc
```diff
@@ -201,69 +310,67 @@ int main(int argc, char *argv[])
...
+      esm_a_packages_count,
+      space_separated_esm_a_packages,
+      esm_i_packages_count,
+      space_separated_esm_i_packages
+   );
+
+   // Execute specified subcommand
+   if (subcommand == PreInvoke) {
+      if (command_used == "upgrade" || command_used == "dist-upgrade") {
+         output_file_if_present(CONTRACT_EXPIRY_STATUS_APT_MESSAGE_STATIC_PATH);
```
blackboxsw, line 372: I had envisioned these two templates as separate sub-messages
any objection to the following?
```suggestion
         output_file_if_present(CONTRACT_EXPIRY_STATUS_MESSAGE_STATIC_PATH);         
         output_file_if_present(CONTRACT_EXPIRY_STATUS_APT_MESSAGE_STATIC_PATH);
```

this way I think we can use a common expiry message/banner for both motd and APT and extend that info for apt providing supplementary messaging.

orndorffgrant: Ahhhh - totally that makes sense. I'll make the change and adjust the integration test

orndorffgrant: FFR: We double checked - motd and apt don't consistently share a header, so we won't apply this change

features/attached_enable.feature
blackboxsw: +1 thanks for this. makes sense

apt-hook/hook.cc
blackboxsw: Ack, we'll close this a Chad's mistaken.

features/attached_enable.feature
blackboxsw: +1 keep it in.

blackboxsw (approved):
Thanks for the discussion back and forth. This iteration makes the most sense given our current messaging needs. If we grow more needs, we can instrument them easily within this framework now. Great work

Issue #1536, opened 2021-04-09
apt upgrade/dist-upgrade: esm messaging about expired/expiring subscriptions 


blackboxsw:
When `apt upgrade` or `apt dist-upgrade` is called on the commandline on an attached machine with an expired license. Prompt with renewal text.

```bash
CAUTION: Your UA Apps: ESM service expired on 11 Dec 2021. Renew UA subscription at https://ubuntu.com/advantage to ensure continued security coverage for your applications. Your grace period will expire in 13 days.
```

Also provide a leading message if expiry is due within the next 20 days
```bash
CAUTION: Your UA Apps: ESM service will expire in 20 days. Renew UA subscription at https://ubuntu.com/advantage to ensure continued security coverage for your applications.
```

PR #1542, opened 2021-04-10
shared MOTD and APT messaging generation

blackboxsw:
## Proposed Commit Message

shared MOTD and APT messaging generation

APT and MOTD messaging share a number of common messages.
                                                                                
To share those message template between apt-hooks and MOTD-hooks
the ua_update_messaging script will generate message templates
appropriate for both APT and MOTD and store them in
/var/lib/ubuntu-advantage/messages/*tmpl.

The ua_update_messaging script will be triggered on every apt update
and will call esm-apt-hook to render those templates replacing
template variable names with dynamic ESM Apps and Infra package counts
and/or package names.
                                                                                
Fixes: #1536
Fixes: #1561

## Test Steps

build the package from this PR
install on a test system
```bash

sed -i 's/UNRELEASED/xenial'
git commit -am 'changelog'
build-package
sbuild-it ../out/*dsc
lxc launch ubuntu-daily:xenial test-x
lxc file push *deb test-x/
lxc exec test-x bash
dpkg -i /ubuntu-advantage-tools*deb
apt update
run-parts /etc/update-motd.d
ua attach <token>
apt update
run-parts /etc/update-motd.d # to see esm related messages
```

Update messages 
```bash
rm -rf /var/lib/ubuntu-advantage/messages
sudo python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py
for file in /var/lib/ubuntu-advantage/messages/*; do 
  echo ----- $file;
  cat $file;
done

sudo run-parts /etc/update-motd.d
apt upgrade

systemctl list-timers | grep ua-messag
```

## Desired commit type::

## Checklist:

### DIFF
--- a/apt-hook/hook.cc
+++ b/apt-hook/hook.cc
@@ -34,10 +34,22 @@
 #include <libintl.h>
 #include <locale.h>
 
-#define CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE_PATH              "/var/lib/ubuntu-advantage/messages/contract-expiry-status.tmpl"
-#define CONTRACT_EXPIRY_STATUS_MESSAGE_STATIC_PATH                "/var/lib/ubuntu-advantage/messages/contract-expiry-status"
-#define CONTRACT_EXPIRY_STATUS_APT_MESSAGE_TEMPLATE_PATH          "/var/lib/ubuntu-advantage/messages/contract-expiry-status-apt.tmpl"
-#define CONTRACT_EXPIRY_STATUS_APT_MESSAGE_STATIC_PATH            "/var/lib/ubuntu-advantage/messages/contract-expiry-status-apt"
+#define MOTD_ESM_SERVICE_STATUS_MESSAGE_STATIC_PATH      "/var/lib/ubuntu-advantage/messages/motd-esm-service-status"
+#define MOTD_APPS_NO_PKGS_TEMPLATE_PATH                  "/var/lib/ubuntu-advantage/messages/motd-no-packages-apps.tmpl"
+#define MOTD_INFRA_NO_PKGS_TEMPLATE_PATH                 "/var/lib/ubuntu-advantage/messages/motd-no-packages-infra.tmpl"
+#define MOTD_APPS_PKGS_TEMPLATE_PATH                     "/var/lib/ubuntu-advantage/messages/motd-packages-apps.tmpl"
+#define MOTD_INFRA_PKGS_TEMPLATE_PATH                    "/var/lib/ubuntu-advantage/messages/motd-packages-infra.tmpl"
+#define MOTD_APPS_PKGS_STATIC_PATH                       "/var/lib/ubuntu-advantage/messages/motd-packages-apps"
+#define MOTD_INFRA_PKGS_STATIC_PATH                      "/var/lib/ubuntu-advantage/messages/motd-packages-infra"
+#define APT_PRE_INVOKE_APPS_NO_PKGS_TEMPLATE_PATH        "/var/lib/ubuntu-advantage/messages/apt-pre-invoke-no-packages-apps.tmpl"
+#define APT_PRE_INVOKE_INFRA_NO_PKGS_TEMPLATE_PATH       "/var/lib/ubuntu-advantage/messages/apt-pre-invoke-no-packages-infra.tmpl"
+#define APT_PRE_INVOKE_APPS_PKGS_TEMPLATE_PATH           "/var/lib/ubuntu-advantage/messages/apt-pre-invoke-packages-apps.tmpl"
+#define APT_PRE_INVOKE_APPS_PKGS_STATIC_PATH             "/var/lib/ubuntu-advantage/messages/apt-pre-invoke-packages-apps"
+#define APT_PRE_INVOKE_INFRA_PKGS_TEMPLATE_PATH          "/var/lib/ubuntu-advantage/messages/apt-pre-invoke-packages-infra.tmpl"
+#define APT_PRE_INVOKE_INFRA_PKGS_STATIC_PATH            "/var/lib/ubuntu-advantage/messages/apt-pre-invoke-packages-infra"
+#define APT_PRE_INVOKE_MESSAGE_STATIC_PATH               "/var/lib/ubuntu-advantage/messages/apt-pre-invoke-esm-service-status"
+#define UBUNTU_NO_WARRANTY_STATIC_PATH                   "/var/lib/ubuntu-advantage/messages/ubuntu-no-warranty"
+
 
 #define ESM_APPS_PKGS_COUNT_TEMPLATE_VAR "{ESM_APPS_PKG_COUNT}"
 #define ESM_APPS_PACKAGES_TEMPLATE_VAR "{ESM_APPS_PACKAGES}"
@@ -140,7 +152,7 @@ static void check_esm_upgrade(pkgCache::PkgIterator pkg, pkgPolicy *policy, resu
       {
          if (pf.File().Archive() != 0 && pf.File().Origin() == std::string("UbuntuESM"))
          {
-            res.esm_i_packages.push_back(pf.File().FileName());
+            res.esm_i_packages.push_back(pkg.Name());
 
             // Pin-Priority: never unauthenticated APT repos == -32768
             if (policy->GetPriority(pf.File()) == -32768)
@@ -240,29 +252,81 @@ static void process_template_file(
 }
 
 static void process_all_templates(
-   std::string esm_a_pkgs_count,
+   int esm_a_pkgs_count,
    std::string esm_a_pkgs,
-   std::string esm_i_pkgs_count,
+   int esm_i_pkgs_count,
    std::string esm_i_pkgs
 ) {
-   std::array<std::string, 2> template_file_names = {
-      CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE_PATH,
-      CONTRACT_EXPIRY_STATUS_APT_MESSAGE_TEMPLATE_PATH
+   std::array<std::string, 4> static_file_names = {
+      APT_PRE_INVOKE_APPS_PKGS_STATIC_PATH,
+      MOTD_APPS_PKGS_STATIC_PATH,
+      APT_PRE_INVOKE_INFRA_PKGS_STATIC_PATH,
+      MOTD_INFRA_PKGS_STATIC_PATH,
    };
-   std::array<std::string, 2> static_file_names = {
-      CONTRACT_EXPIRY_STATUS_MESSAGE_STATIC_PATH,
-      CONTRACT_EXPIRY_STATUS_APT_MESSAGE_STATIC_PATH
+   std::array<std::string, 3> apt_static_files = {
+      APT_PRE_INVOKE_APPS_PKGS_STATIC_PATH,
+      APT_PRE_INVOKE_INFRA_PKGS_STATIC_PATH,
+      UBUNTU_NO_WARRANTY_STATIC_PATH
    };
+   std::array<std::string, 3> motd_static_files = {
+      MOTD_APPS_PKGS_STATIC_PATH,
+      MOTD_INFRA_PKGS_STATIC_PATH,
+      UBUNTU_NO_WARRANTY_STATIC_PATH
+   };
+
+   std::vector<std::string> template_file_names;
+   if (esm_a_pkgs_count > 0) {
+      template_file_names.push_back(APT_PRE_INVOKE_APPS_PKGS_TEMPLATE_PATH);
+      template_file_names.push_back(MOTD_APPS_PKGS_TEMPLATE_PATH);
+   } else {
+      template_file_names.push_back(APT_PRE_INVOKE_APPS_NO_PKGS_TEMPLATE_PATH);
+      template_file_names.push_back(MOTD_APPS_NO_PKGS_TEMPLATE_PATH);
+   }
+   if (esm_i_pkgs_count > 0) {
+      template_file_names.push_back(APT_PRE_INVOKE_INFRA_PKGS_TEMPLATE_PATH);
+      template_file_names.push_back(MOTD_INFRA_PKGS_TEMPLATE_PATH);
+   } else {
+      template_file_names.push_back(APT_PRE_INVOKE_INFRA_NO_PKGS_TEMPLATE_PATH);
+      template_file_names.push_back(MOTD_INFRA_NO_PKGS_TEMPLATE_PATH);
+   }
    for (uint i = 0; i < template_file_names.size(); i++) {
       process_template_file(
          template_file_names[i], 
          static_file_names[i], 
-         esm_a_pkgs_count,
+         std::to_string(esm_a_pkgs_count),
          esm_a_pkgs,
-         esm_i_pkgs_count,
+         std::to_string(esm_i_pkgs_count),
          esm_i_pkgs
       );
    }
+
+   std::ofstream apt_pre_invoke_msg;
+   apt_pre_invoke_msg.open(APT_PRE_INVOKE_MESSAGE_STATIC_PATH);
+   for (uint i = 0; i < apt_static_files.size(); i++) {
+       std::ifstream message_file(apt_static_files[i]);
+       if (message_file.is_open()) {
+           apt_pre_invoke_msg << std::endl;
+           apt_pre_invoke_msg << message_file.rdbuf();
+           message_file.close();
+       };
+   }
+   apt_pre_invoke_msg << std::endl;
+   apt_pre_invoke_msg.close();
+
+   std::ofstream motd_msg;
+   motd_msg.open(MOTD_ESM_SERVICE_STATUS_MESSAGE_STATIC_PATH);
+   for (uint i = 0; i < motd_static_files.size(); i++) {
+       std::ifstream message_file(motd_static_files[i]);
+       if (message_file.is_open()) {
+           if ( i > 0 ) {
+               motd_msg << std::endl;
+           }
+           motd_msg << message_file.rdbuf();
+           message_file.close();
+       };
+   }
+   motd_msg << std::endl;
+   motd_msg.close();
 }
 
 static void output_file_if_present(std::string file_name) {
@@ -360,16 +424,16 @@ int main(int argc, char *argv[])
    std::string esm_a_packages_count = std::to_string(res.esm_a_packages.size());
 
    process_all_templates(
-      esm_a_packages_count,
+      res.esm_a_packages.size(),
       space_separated_esm_a_packages,
-      esm_i_packages_count,
+      res.esm_i_packages.size(),
       space_separated_esm_i_packages
    );
 
    // Execute specified subcommand
    if (subcommand == PreInvoke) {
       if (command_used == "upgrade" || command_used == "dist-upgrade") {
-         output_file_if_present(CONTRACT_EXPIRY_STATUS_APT_MESSAGE_STATIC_PATH);
+         output_file_if_present(APT_PRE_INVOKE_MESSAGE_STATIC_PATH);
       }
    }
 
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -8,8 +8,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
        Given a `<release>` machine with ubuntu-advantage-tools installed
         When I run `apt-get update` with sudo, retrying exit [100]
         And I run `apt-get install -y <downrev_pkg>` with sudo, retrying exit [100]
-        When I verify that running ` --assume-yes --beta` `with sudo` exits `1`
-        And I run `/usr/lib/update-notifier/apt-check  --human-readable` as non-root
+        And I run `run-parts /etc/update-motd.d/` with sudo
         Then if `<release>` in `trusty` and stdout matches regexp:
         """
         UA Infrastructure Extended Security Maintenance \(ESM\) is not enabled.
@@ -35,7 +34,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         When I attach `contract_token` with sudo
         Then stdout matches regexp:
         """
-        ESM Infra enabled
+        UA Infra: ESM enabled
         """
         And stdout matches regexp:
         """
@@ -53,7 +52,14 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         Enabling default service esm-infra
         """
-        When I run `/usr/lib/update-notifier/apt-check  --human-readable` as non-root
+        When I append the following on uaclient config:
+            """
+            features:
+              allow_beta: true
+            """
+        And I run `apt update` with sudo
+        And I run `python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py` with sudo
+        And I run `run-parts /etc/update-motd.d/` with sudo
         Then if `<release>` in `trusty` and stdout matches regexp:
         """
         UA (Infra:|Infrastructure) Extended Security Maintenance \(ESM\) is enabled.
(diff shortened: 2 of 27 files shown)


lib/ua_update_messaging.py
```diff
@@ -0,0 +1,113 @@
...
+    if not cfg.is_attached:
+        return None
+    grace_period = -defaults.CONTRACT_EXPIRY_GRACE_PERIOD_DAYS
+    pending_expiry = defaults.CONTRACT_EXPIRY_PENDING_DAYS
+    remaining_days = cfg.contract_remaining_days
+    if 0 <= remaining_days <= pending_expiry:
+        return status.MESSAGE_CONTRACT_NEAR_EXPIRED_TMPL.format(
+            remaining_days=remaining_days, url=defaults.BASE_UA_URL
+        )
+    elif grace_period <= remaining_days < 0:
```
lucasmoura, line 54: It seems `grace_period`  is positive here. Shouldn't it be a negative number ?

blackboxsw: Sorry, probably confusing how I setup the local grace_period variable above as a negative. Probably easiest to make that postive/negative conversion locally here on the line instead of [up here ](https://github.com/canonical/ubuntu-advantage-client/pull/1542/files#diff-02bd06ac4b2af43bf296d5379d9a0d3c25cf70e5f417ea5d34fc9b98ca6683d6R47)

blackboxsw: I'll change that to inflect the value of local grace_period  here

lucasmoura: :facepalm: my mistake, just saw the negative sign there

uaclient/tests/test_ua_update_messaging.py
```diff
@@ -0,0 +1,145 @@
...
+
+M_MESSAGE_PATH = "lib.ua_update_messaging."
+
+
+class TestUpdateAPTandMOTDMessages:
+    @pytest.mark.parametrize("expiry_msg", (None, "expiry"))
+    @pytest.mark.parametrize("esm_news_msg", (None, "esm_news"))
+    @mock.patch(M_MESSAGE_PATH + "format_esm_announcement_message")
+    @mock.patch(M_MESSAGE_PATH + "format_contract_expiry_message")
+    def test_(
```
lucasmoura, line 125: I think we are missing the name of the method here

features/attach_validtoken.feature
```diff
@@ -73,8 +79,29 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
         \d+ of these updates (is a|are) security update(s)?.
+        To see these additional updates run: apt list --upgradable
+
+        Enable UA Infra: ESM to receive \d+ additional security updates.
+        See https:\/\/ubuntu.com\/security\/esm or run: sudo ua status
+        """
+        When I update contract to use `effectiveTo` as `days=-20`
+        And I run `python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py` with sudo
+        And I run `cat /var/lib/ubuntu-advantage/messages/contract-expiry-status.tmpl` with sudo
+        Then if `<release>` in `xenial or bionic` and stdout matches regexp:
```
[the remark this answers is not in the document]
orndorffgrant: For this one it looks like we just need a blank line here to match the string defined in python ->
```
        \*YOUR UA APPS: ESM SUBSCRIPTION HAS EXPIRED.\*

        {ESM_APPS_PKG_COUNT} security updates NOT APPLIED.
        Renew your UA services at https:\/\/ubuntu.com\/advantage
```

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,112 @@
...
+    return None
+
+
+def format_esm_announcement_message():
+    """Return human-readable message if ESM is offered on this LTS release."""
+    series = util.get_platform_info()["series"]
+    ubuntuinfo = distro_info.UbuntuDistroInfo()
+    if ubuntuinfo.is_lts(series):
+        return status.MESSAGE_ANNOUNCE_ESM
+    return None
```
orndorffgrant, line 73: Do we need some type of `is_beta` check before showing this message?

features/unattached_commands.feature
```diff
@@ -56,6 +56,18 @@ Feature: Command behaviour when unattached
...
+        Then if `<release>` in `xenial` and stdout matches regexp:
+        """
+        \* Introducing Extended Security Maintenance for Applications.
+        Receive updates to over 30,000 software packages with
+        your Ubuntu Advantage subscription. Free for personal use.
+
+          https:\/\/ubuntu.com\/advantage
+
+        UA Infra: Extended Security Maintenance \(ESM\) is not enabled.
+        """
```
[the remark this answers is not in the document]
orndorffgrant: I think its the indentation that this one is tripping over, maybe something like this would work?
```
        \* Introducing Extended Security Maintenance for Applications.
          Receive updates to over 30,000 software packages with
          your Ubuntu Advantage subscription. Free for personal use.

            https:\/\/ubuntu.com\/advantage

        UA Infra: Extended Security Maintenance \(ESM\) is not enabled.
```

uaclient/tests/test_ua_update_messaging.py
```diff
@@ -0,0 +1,288 @@
...
+)
+from uaclient import util
+
+from lib.ua_update_messaging import (
+    format_contract_expiry_messages,
+    format_esm_announcement_message,
+    update_apt_and_motd_messages,
+)
+
+M_MESSAGE_PATH = "lib.ua_update_messaging."
```
orndorffgrant, line 28: Personal preference would be to hard-code this string in all the `mock.patch` calls so I don't have to remember what the variable is when figuring out what is mocked, but :shrug: nbd either way

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,167 @@
...
+    util.subp(["/usr/lib/ubuntu-advantage/apt-esm-hook", "process-templates"])
+
+    # Announce ESM availabilty on LTS releases
+    esm_news_msg = format_esm_announcement_message(cfg)
+    esm_news_file = os.path.join(msg_dir, ExternalMessage.ESM_ANNOUNCE.value)
+    if esm_news_msg:
+        util.write_file(esm_news_file, esm_news_msg)
+    else:
+        util.remove_file(esm_news_file)
+        util.remove_file(esm_news_file.replace(".tmpl", ""))
```
orndorffgrant, line 162: Currently esm_news_file isn't a `tmpl` file so we dont need this line
```suggestion
    else:
        util.remove_file(esm_news_file)
```

orndorffgrant (commented):
The code looks really good! Left some very small comments. Going to follow your manual test steps now

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,167 @@
...
+        apt_msg = (
+            status.MESSAGE_CONTRACT_EXPIRED_APT_TMPL.format(
+                num_pkg="{ESM_APPS_PKG_COUNT}", url=defaults.BASE_UA_URL
+            )
+            + no_warranty
+        )
+    return motd_msg, apt_msg
+
+
+def format_esm_announcement_message(cfg: config.UAConfig) -> "Optional[str]":
```
orndorffgrant, line 88: @blackboxsw Another check we need to add here. If esm-apps is already enabled, we don't want to show the announcement message, right?

orndorffgrant:
@blackboxsw Do we want to include this kind of message in this PR as well?

LTS user without ESM Apps enabled:
apt upgrade
```
...
*The following packages could receive security updates with UA Apps: ESM service enabled:
  jq monit redis slurm-llnl
 1 esm-apps security update
Learn more about UA Apps: ESM service at https://ubuntu.com/esm
...
```

We could reuse `contract-expiry-status-apt.tmpl` for it or make a separate file

Or we can save it for a later PR

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,8 +1,10 @@
 APT::Update::Post-Invoke-Stats {
+        "[ ! -f /usr/lib/ubuntu-advantage/ua_update_messaging.py ] || python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py";
 	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook post-invoke-stats";
 };
 
 APT::Install::Post-Invoke-Success {
+        "[ ! -f /usr/lib/ubuntu-advantage/ua_update_messaging.py ] || python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py";
 	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook post-invoke-success";
```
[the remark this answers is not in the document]
blackboxsw: thanks fixed.

lib/ua_update_messaging.py
blackboxsw: Excellent point. will drop that message on enabled +1

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,8 +1,10 @@
 APT::Update::Post-Invoke-Stats {
+        "[ ! -f /usr/lib/ubuntu-advantage/ua_update_messaging.py ] || python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py";
```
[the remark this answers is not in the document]
blackboxsw: Yes, now that these pieces are fitting together, it looks like we might need to rethink parts of this and see what the best solution is. I like the speed of C++ esm-apt-hook here, we probably shouldn't be calling it twice though, once from python, once with the hook. This is the rub with baking our python ua_message_update into apt hooks instead of having a stand-alone systemd timer. We should probably ping the broader team on this to see if there are some gotchas. I know folks are not happy with apt hooks doing much more than simple file reads hence I didn't try to refactor these two modules yet because I think we may have to consider a separate systemd unit or timer.

blackboxsw:
> @blackboxsw Do we want to include this kind of message in this PR as well?
> 
> LTS user without ESM Apps enabled:
> apt upgrade
> 
> ```
> ...
> *The following packages could receive security updates with UA Apps: ESM service enabled:
>   jq monit redis slurm-llnl
>  1 esm-apps security update
> Learn more about UA Apps: ESM service at https://ubuntu.com/esm
> ...
> ```
>

I was hoping to save this as a separate PR (because I think it begs to result of our conversation Monday about whether or not to yank ua_update_message calls from apt-hook.  In order for  ua_update_messaging to good job with rendering (and wrapping) the potential package names in that formatted message, it seem to require that python calls the esm-apt-hook get-stats which could just format JSON data like
```json
{"esm-apps": ["pkg1", "pkg2"], "esm-infra": ["pkg3", "pkg4"]}
```

orndorffgrant:
> I was hoping to save this as a separate PR (because I think it begs to result of our conversation Monday about whether or not to yank ua_update_message calls from apt-hook. In order for ua_update_messaging to good job with rendering (and wrapping) the potential package names in that formatted message, it seem to require that python calls the esm-apt-hook get-stats which could just format JSON data like
> 
> ```json
> {"esm-apps": ["pkg1", "pkg2"], "esm-infra": ["pkg3", "pkg4"]}
> ```

Got it, okay. Wrapping wouldn't be too hard if we had to do it in c++ also. But yeah sounds good - we can hold off on this until we have further discussions.

orndorffgrant:
Adding more thoughts that we can continue discussing on Monday:

Below, "that approach" refers to apt-hook having to modes of operation:
1. produce stats in some type of structured form
2. print static files if they're present (note that this could be done without c++ as just an `echo` in the hook conf file)

I believe the reason we didn't go with "that approach"^ before is the following scenario:
1. ua_update_messaging runs daily and creates prerendered files based on current apt cache
2. `apt update` occurs that makes the current status messages incorrect
3. `apt upgrade`just outputs the prerendered files, but they're now incorrect

Thats how we landed on always needing the apt-hook to render "on the fly" and since it has to render anyway, it might as well be the one true rendering implementation... and now we're here!

That scenario was a concern when we were running `ua_update_messaging` daily. If we start running this script on `apt update` (as this PR introduces), then it is no longer a concern, but then we're calling a whole python script on every `apt update`, which might be unpalatable. (I personally think its worth the perf hit for the drastic reduction in complexity - we could actually remove the c++ entirely).

apt-hook/20apt-esm-hook.conf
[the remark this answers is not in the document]
blackboxsw: I put up a sample set of commits for us to talk about Monday.
apt-hook pkg-status-json d61f30a513ed3de978e7df5a2b525d6d2b4db1b2 and ua status unattached messaging about 11ad82aaec4ed0851037639a8ba54c0643061037

debian/changelog
```diff
@@ -1,4 +1,4 @@
-ubuntu-advantage-tools (27.0) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (27.0) xenial; urgency=medium
```
lucasmoura, line 1: We can drop this change

debian/control
```diff
@@ -5,6 +5,7 @@ Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
 Build-Depends: bash-completion,
                debhelper (>=9),
                debianutils,
+               distro-info,
```
lucasmoura, line 8: We can also drop this change as it was introduced on #1555

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,8 +1,10 @@
 APT::Update::Post-Invoke-Stats {
+	"[ ! -f /usr/lib/ubuntu-advantage/ua_update_messaging.py ] || python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py";
```
blackboxsw, line 2: Note I would like to separate this back to a systemd timer so we don't get "heavyweight" python calls on every apt update/upgrade etc.

apt-hook/hook.cc
```diff
@@ -34,10 +34,21 @@
 #include <libintl.h>
 #include <locale.h>
 
-#define CONTRACT_EXPIRY_STATUS_MESSAGE_TEMPLATE_PATH              "/var/lib/ubuntu-advantage/messages/contract-expiry-status.tmpl"
-#define CONTRACT_EXPIRY_STATUS_MESSAGE_STATIC_PATH                "/var/lib/ubuntu-advantage/messages/contract-expiry-status"
-#define CONTRACT_EXPIRY_STATUS_APT_MESSAGE_TEMPLATE_PATH          "/var/lib/ubuntu-advantage/messages/contract-expiry-status-apt.tmpl"
-#define CONTRACT_EXPIRY_STATUS_APT_MESSAGE_STATIC_PATH            "/var/lib/ubuntu-advantage/messages/contract-expiry-status-apt"
+#define MOTD_ESM_SERVICE_STATUS_MESSAGE_STATIC_PATH      "/var/lib/ubuntu-advantage/messages/motd-esm-service-status"
```
blackboxsw, line 37: We not fracture the messages into fragments, and apt-hook renders the appropriate NO_PKGS vs PKGS template based on what esm service pkg counts are. Ultimately after rendering partial messages process_all_templates concatonates the fragments into APT_PRE_INVOKE_MESSAGE_STATIC_PATH and MOTD_ESM_SERVICE_STATUS_MESSAGE_STATIC_PATH

apt-hook/hook.cc
```diff
@@ -140,7 +151,7 @@ static void check_esm_upgrade(pkgCache::PkgIterator pkg, pkgPolicy *policy, resu
       {
          if (pf.File().Archive() != 0 && pf.File().Origin() == std::string("UbuntuESM"))
          {
-            res.esm_i_packages.push_back(pf.File().FileName());
+            res.esm_i_packages.push_back(pkg.Name());
```
blackboxsw, line 154: Found a bug here where we were actually rendering the apt policyfile names instad of pkg name.

apt-hook/hook.cc
```diff
@@ -240,29 +251,74 @@ static void process_template_file(
...
-         esm_a_pkgs_count,
+         std::to_string(esm_a_pkgs_count),
          esm_a_pkgs,
-         esm_i_pkgs_count,
+         std::to_string(esm_i_pkgs_count),
          esm_i_pkgs
       );
    }
+
+   std::ofstream apt_pre_invoke_msg;
```
blackboxsw, line 300: Here we aggregate the APPS and INFRA messages together into the final APT of MOTD messages

features/unattached_commands.feature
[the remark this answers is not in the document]
blackboxsw: thanks used better whitespace match with a leading " +" in most lines.

apt-hook/hook.cc
orndorffgrant: Oop! Thought I fixed that but clearly didn't commit it - nice catch!

apt-hook/hook.cc
```diff
@@ -240,29 +251,74 @@ static void process_template_file(
...
+   for (uint i = 0; i < apt_static_files.size(); i++) {
+       std::ifstream message_file(apt_static_files[i]);
+       if (message_file.is_open()) {
+           apt_pre_invoke_msg.open(APT_PRE_INVOKE_MESSAGE_STATIC_PATH);
+           apt_pre_invoke_msg << message_file.rdbuf();
+           apt_pre_invoke_msg.close();
+           message_file.close();
+       };
+   }
+   apt_pre_invoke_msg.close();
```
orndorffgrant, line 310: As currently written I think this `close` is extraneous.

There might be a micro-optimization opportunity here where you open `apt_pre_invoke_msg` before the loop, and close after the loop to avoid extra syscalls, but probably not worth spending too much time on.

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+def update_apt_and_motd_messages(cfg: config.UAConfig) -> None:
+    """Emit templates and human-readable status messages in msg_dir.
+
+    These structured messages will be sourced by both /etc/update.motd.d
+    and APT UA-configured hooks. APT hook content will orginate from
+    apt-hook/hook.cc
+
+    Call esm-apt-hook process-templates to render final humnan-readable
```
orndorffgrant, line 284:
```suggestion
    Call esm-apt-hook process-templates to render final human-readable
```

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,317 @@
...
+    _write_template_or_remove(pkgs_msg, os.path.join(msg_dir, pkgs_file))
+    _write_template_or_remove(
+        motd_no_pkgs_msg, os.path.join(msg_dir, motd_no_pkgs_file)
+    )
+    _write_template_or_remove(
+        motd_pkgs_msg, os.path.join(msg_dir, motd_pkgs_file)
+    )
+
+
+def write_apt_pre_invoke_templates(cfg: config.UAConfig) -> None:
```
orndorffgrant, line 204: Consider renaming this function since this does more than just apt_pre_invoke messages

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,317 @@
...
+    config_allow_beta = util.is_config_value_true(
+        config=cfg.cfg, path_to_value="features.allow_beta"
+    )
+    apps_not_beta = bool(config_allow_beta or not apps_cls.is_beta)
+    infra_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME["esm-infra"]
+    infra_inst = infra_cls(cfg)
+
+    expiry_status, remaining_days = get_contract_expiry_status(cfg)
+
+    if apps_not_beta:
```
orndorffgrant, line 229: Do we need a `series != "trusty"` check here as well or no?

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,317 @@
...
+            pkgs_msg = MESSAGE_CONTRACT_EXPIRED_GRACE_PERIOD_TMPL.format(
+                expired_date=cfg.contract_expiry_datetime.strftime("%d %b %Y"),
+                remaining_days=grace_period_remaining,
+                url=defaults.BASE_UA_URL,
+            )
+            # Same cautionary message when in grace period
+            motd_pkgs_msg = motd_no_pkgs_msg = no_pkgs_msg = pkgs_msg
+        elif expiry_status == ContractExpiryStatus.EXPIRED:
+            series = util.get_platform_info()["series"]
+            if ent.name == "esm-apps" and series in ("trusty", "xenial"):
```
orndorffgrant, line 146: Why only for "esm-apps" here?

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,317 @@
...
+
+def remove_templates_and_messages(
+    cfg: config.UAConfig, message_files: "List[str]"
+):
+    msg_dir = os.path.join(cfg.data_dir, "messages")
+    for filename in message_files:
+        file_path = os.path.join(msg_dir, filename)
+        util.remove_file(file_path)
+        if filename.endswith(".tmpl"):
+            util.remove_file(file_path.replace(".tmpl", ""))
```
orndorffgrant, line 74: This function doesn't appear to be used anywhere unless I'm missing something

uaclient/entitlements/repo.py
```diff
@@ -395,9 +395,9 @@ def remove_apt_config(self, run_apt_update=True):
         if not repo_url:
             raise exceptions.MissingAptURLDirective(self.name)
         if self.disable_apt_auth_only:
-            # We only remove the repo from the apt auth file, because ESM Infra
-            # is a special-case: we want to be able to report on the
-            # available ESM Infra updates even when it's disabled
+            # We only remove the repo from the apt auth file, because
+            # UA Infra: ESM is a special-case: we want to be able to report on
+            # the available UA Infra: ESM updates even when it's disabled
```
orndorffgrant, line 400: I'm sure it was tedious but I appreciate standardizing terminology :rocket:

lib/ua_update_messaging.py
```diff
@@ -0,0 +1,317 @@
...
+        for msg_enum in ExternalMessage:
+            msg_path = os.path.join(msg_dir, msg_enum.value)
+            util.remove_file(msg_path)
+            if msg_path.endswith(".tmpl"):
+                util.remove_file(msg_path.replace(".tmpl", ""))
+        return
+
+    # Announce ESM availabilty on active ESM LTS releases
+    write_esm_announcement_message(cfg, series)
+    write_apt_pre_invoke_templates(cfg)
```
orndorffgrant, line 312: Do we still want to run `apt-esm-hook process-templates` after this? or just let it happen when the user runs `apt update` next?

orndorffgrant (commented):
Finished looking over the implementation code and it looks really good :rocket: Just left some small comments and clarifications

I'll start looking at the test code and actually running this now/after lunch.

lib/ua_update_messaging.py
blackboxsw: @orndorffgrant I think it's probably a good idea to run process-tempates just to sync with current state and avoid strange "racy" behavior if apt pkg counts change. Good suggestion.

apt-hook/hook.cc
blackboxsw: +1 though what I found was that if I opened and closed the file, I'd always leave around stale empty files. But turns out as well, that since ofstream.open defaults to truncate I actually blow away msg parts each open. I'll move this msg open/close  outside the for loop, maybe we can check a msg.tellp() check to remove the file if empty afterward.

lib/ua_update_messaging.py
blackboxsw: Smart man, I think I actually need to separate this to a separate msg file because it needs to be the footer after both APPS and Infra expiry msgs.

apt-hook/hook.cc
orndorffgrant, line 310: @blackboxsw I think this is actually overwriting the file each iteration instead of appending to the file

uaclient/tests/test_ua_update_messaging.py
```diff
@@ -0,0 +1,427 @@
...
+M_PATH = "lib.ua_update_messaging."
+
+
+class TestGetContractExpiryStatus:
+    @pytest.mark.parametrize(
+        "contract_remaining_days,expected_status",
+        (
+            (21, ContractExpiryStatus.ACTIVE),
+            (20, ContractExpiryStatus.ACTIVE_EXPIRED_SOON),
+            (-1, ContractExpiryStatus.EXPIRED_GRACE_PERIOD),
```
orndorffgrant, line 44: We can add in some more edges here pretty easily
```suggestion
            (20, ContractExpiryStatus.ACTIVE_EXPIRED_SOON),
            (0, ContractExpiryStatus.ACTIVE_EXPIRED_SOON),
            (-1, ContractExpiryStatus.EXPIRED_GRACE_PERIOD),
            (-14, ContractExpiryStatus.EXPIRED_GRACE_PERIOD),
            (-15, ContractExpiryStatus.EXPIRED),
```

lib/ua_update_messaging.py
orndorffgrant: bump to make sure you saw this. I don't know the answer to this

orndorffgrant (commented):
Did a final lookover before signing off for the day. These are the final things I see left here to do:
* Consider moving no_warranty message to be computed separately as mentioned in the comment
* Get the timer activated on install
* make sure all Integration tests working

After that this is good to go from my perspective!

apt-hook/hook.cc
blackboxsw: Ack, I think I fixed this now including a std::endl; between each template part.

lib/ua_update_messaging.py
blackboxsw: Bummer I trashed this change in a prior rebase . Fixing now.

apt-hook/hook.cc
```diff
@@ -240,29 +252,81 @@ static void process_template_file(
...
+
+   std::ofstream apt_pre_invoke_msg;
+   apt_pre_invoke_msg.open(APT_PRE_INVOKE_MESSAGE_STATIC_PATH);
+   for (uint i = 0; i < apt_static_files.size(); i++) {
+       std::ifstream message_file(apt_static_files[i]);
+       apt_pre_invoke_msg << std::endl;
+       if (message_file.is_open()) {
+           apt_pre_invoke_msg << message_file.rdbuf();
+           message_file.close();
+       };
```
orndorffgrant, line 311: I think we need to move this line
```suggestion
       if (message_file.is_open()) {
           apt_pre_invoke_msg << std::endl;
           apt_pre_invoke_msg << message_file.rdbuf();
           message_file.close();
       };
```

to avoid adding a new line for each file unless the file exists

orndorffgrant (approved):
Thanks for all the diligence for getting this huge messaging PR through! I noticed one thing where there seem to be extra newlines in places, but we can come back through with a bugfix PR to figure out the best way to fix that - I'm not sure if my proposed fix would fix it completely or cause new problems.

apt-hook/hook.cc
blackboxsw: excellent.thanks

Issue #1544, opened 2021-04-12
fix: esm-apps (beta) Cannot enable unknown service 'esm-apps'.


blackboxsw:
Since esm-apps is a beta service, ua client doesn't know about it by default.
Environments without proper allow_beta: true esm-apps configuration will hit the following error for upgrades which originaly from esm-apps PPAs. "Cannot enable unknown service 'esm-apps'."

```bash
ubuntu@cpc-x:~$ sudo ua fix CVE-2020-28196
CVE-2020-28196: Kerberos vulnerability
https://ubuntu.com/security/CVE-2020-28196
1 affected package is installed: krb5
(1/1) krb5:
A fix is available in Ubuntu standard updates.
The update is not installed because this system does not have
esm-apps enabled.

Choose: [E]nable esm-apps [C]ancel
> e
{ ua enable esm-apps }
One moment, checking your subscription first
Cannot enable unknown service 'esm-apps'.
Try esm-infra, fips, fips-updates, livepatch.
```

PR #1549, opened 2021-04-14
fix: do not prompt user for beta services

lucasmoura:
## Proposed Commit Message
fix: do not prompt user for beta services

When running ua fix, we may need a package fix that can only be installed using a beta service. In that situation, we will
only permit that the fix is installed if the service is already enabled. This means that we will not prompt user for enabling a
beta service or even attaching to UA if we found out that the service that will provide the fixes is a beta one.

Fixes: #1544

## Test Steps
Run the new and modified unit tests

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -473,7 +473,7 @@ def query_installed_source_pkg_versions() -> "Dict[str, Dict[str, str]]":
 
 
 def merge_usn_released_binary_package_versions(
-    usns: "List[USN]"
+    usns: "List[USN]", beta_pockets: "Dict[str, bool]"
 ) -> "Dict[str,  Dict[str, Dict[str, str]]]":
     """Walk related USNs, merging the released binary package versions.
 
@@ -481,6 +481,11 @@ def merge_usn_released_binary_package_versions(
         package names and required fix version. If multiple related USNs
         require differnt version fixes to the same binary package, track the
         maximum version required across all USNs.
+
+    :param usns: List of USN response instances from which to calculate merge.
+    :param beta_pockets: Dict keyed on service name: esm-infra, esm-apps
+        the values of which will be true of USN response instances
+        from which to calculate merge.
 
     :return: Dict keyed by source package name. Under each source package will
         be a dict with binary package name as keys and binary package metadata
@@ -490,21 +495,27 @@ def merge_usn_released_binary_package_versions(
     for usn in usns:
         # Aggregate USN.release_package binary versions into usn_pkg_versions
         for src_pkg, binary_pkg_versions in usn.release_packages.items():
-            if src_pkg not in usn_pkg_versions:
-                usn_pkg_versions[src_pkg] = binary_pkg_versions
-            else:
+            public_bin_pkg_versions = {
+                bin_pkg_name: bin_pkg_md
+                for bin_pkg_name, bin_pkg_md in binary_pkg_versions.items()
+                if False
+                is beta_pockets.get(bin_pkg_md.get("pocket", "None"), False)
+            }
+            if src_pkg not in usn_pkg_versions and public_bin_pkg_versions:
+                usn_pkg_versions[src_pkg] = public_bin_pkg_versions
+            elif src_pkg in usn_pkg_versions:
                 # Since src_pkg exists, only record this USN's binary version
                 # when it is greater than the previous version in usn_src_pkg.
                 usn_src_pkg = usn_pkg_versions[src_pkg]
-                for binary_pkg, binary_pkg_md in binary_pkg_versions.items():
-                    if binary_pkg not in usn_src_pkg:
-                        usn_src_pkg[binary_pkg] = binary_pkg_md
+                for bin_pkg, binary_pkg_md in public_bin_pkg_versions.items():
+                    if bin_pkg not in usn_src_pkg:
+                        usn_src_pkg[bin_pkg] = binary_pkg_md
                     else:
-                        prev_version = usn_src_pkg[binary_pkg]["version"]
+                        prev_version = usn_src_pkg[bin_pkg]["version"]
                         current_version = binary_pkg_md["version"]
                         if not version_cmp_le(current_version, prev_version):
                             # binary_version is greater than prev_version
-                            usn_src_pkg[binary_pkg] = binary_pkg_md
+                            usn_src_pkg[bin_pkg] = binary_pkg_md
     return usn_pkg_versions
 
 
@@ -512,6 +523,13 @@ def fix_security_issue_id(cfg: UAConfig, issue_id: str) -> None:
     issue_id = issue_id.upper()
     client = UASecurityClient(cfg=cfg)
     installed_packages = query_installed_source_pkg_versions()
+
+    # Used to filter out beta pockets during merge_usns
+    beta_pockets = {
+        "esm-apps": _is_pocket_used_by_beta_service("esm-apps", cfg),
+        "esm-infra": _is_pocket_used_by_beta_service("esm-infra", cfg),
+    }
+
     if "CVE" in issue_id:
         try:
             cve = client.get_cve(cve_id=issue_id)
@@ -527,7 +545,9 @@ def fix_security_issue_id(cfg: UAConfig, issue_id: str) -> None:
             cve=cve, installed_packages=installed_packages
         )
         print(cve.get_url_header())
-        usn_released_pkgs = merge_usn_released_binary_package_versions(usns)
+        usn_released_pkgs = merge_usn_released_binary_package_versions(
+            usns, beta_pockets
+        )
     else:  # USN
         related_usns = {}
         try:
@@ -549,7 +569,9 @@ def fix_security_issue_id(cfg: UAConfig, issue_id: str) -> None:
         affected_pkg_status = get_usn_affected_packages_status(
             usn=usn, installed_packages=installed_packages
         )
-        usn_released_pkgs = merge_usn_released_binary_package_versions(usns)
+        usn_released_pkgs = merge_usn_released_binary_package_versions(
+            usns, beta_pockets
+        )
         print(usn.get_url_header())
         related_cves = set(itertools.chain(*[u.cves_ids for u in usns]))
         if not related_cves:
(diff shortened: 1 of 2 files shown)


uaclient/tests/test_security.py
```diff
@@ -1762,13 +1949,13 @@ def test_messages_for_affected_packages_when_service_kept_disabled(
...
         type(m_entitlement_obj).name = mock.PropertyMock(
-            return_value="esm-apps"
+            return_value="esm-infra"
         )
 
         cfg = FakeConfig()
         cfg.for_attached_machine()
         with mock.patch.object(
-            sec, "ENTITLEMENT_CLASS_BY_NAME", {"esm-apps": m_entitlement_cls}
+            sec, "ENTITLEMENT_CLASS_BY_NAME", {"esm-infra": m_entitlement_cls}
```
orndorffgrant, line 1958: I think I understand the new tests, but can you explain why we changed several existing tests to reference "infra" instead of "apps" like right here?

orndorffgrant (commented):
Good stuff!

Just a nit on what appears to be dead code and a question for my education

uaclient/security.py
```diff
@@ -971,40 +971,51 @@ def _prompt_for_enable(cfg: UAConfig, service: str) -> bool:
...
+
+    if ent_cls.is_beta:
+        return False if config_allow_beta else True
     else:
-        service_to_check = "esm-apps"
+        return False
 
-    ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(service_to_check)
+    if ent_cls.is_beta and not config_allow_beta:
+        return False
```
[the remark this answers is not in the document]
lucasmoura: Yep, great catch. I updated that code and forgot to remove that :facepalm:

uaclient/tests/test_security.py
lucasmoura: My rationale here is just to make the test seems more "right". I mean that we expect `esm-infra`  package related fixes to work fine, which will not happen to `esm-apps` by default now. So I am just changing the name to reflect that.

But feature wise, it will not change anything, since we are mocking the service here, as you have noted.

orndorffgrant: Ahhh I understand - it would be confusing for the tests to expect output that doesn't align with the defaults in actual usage. Yeah I think that's a good idea!

blackboxsw (changes requested):
As-is this branch emits messages like the following on a xenial system with `sudo apt-get install krb5-locales=1.13.2+dfsg-5`

```bash
ubuntu@base-xenial:~$ ua fix CVE-2020-28196
CVE-2020-28196: Kerberos vulnerabilities
https://ubuntu.com/security/CVE-2020-28196
1 affected package is installed: krb5
(1/1) krb5:
A fix is available in UA Apps.
Package fixes cannot be installed.
To install them, run this command as root (try using sudo)
1 package is still affected: krb5
✘ CVE-2020-28196 is not resolved.
ubuntu@base-xenial:~$ sudo ua fix CVE-2020-28196
CVE-2020-28196: Kerberos vulnerabilities
https://ubuntu.com/security/CVE-2020-28196
1 affected package is installed: krb5
(1/1) krb5:
A fix is available in UA Apps.
No fix released yet
1 package is still affected: krb5
✘ CVE-2020-28196 is not resolved.
```

I actually don't think we should list any pkgs from UA Apps at all if the service is in beta.
Ultimately shouldn't we be ignoring ESM Apps while in beta anyway.

Here's a diff of where I think we could perform that ignoring
```diff
        if pocket == UA_INFRA_POCKET:
            service_to_check = "esm-infra"
        else:
            service_to_check = "esm-apps"

        ent_cls = ENTITLEMENT_CLASS_BY_NAME[service_to_check]
```

lucasmoura:
@blackboxsw code updated

uaclient/security.py
```diff
@@ -740,17 +777,36 @@ def _handle_released_package_fixes(
...
+                ):
+                    # This means that we do have packages to fix and
+                    # we must update the message to show that the fixes
+                    # are not yet released.
+                    pkg_status_override = (
+                        status.MESSAGE_SECURITY_FIX_NOT_RELEASED_YET
+                    )
+                    upgrade_status = False
+                    beta_service = True
+
```
blackboxsw, line 804: I don't think we should be doing this check here because it's too late and we've already filtered out other "valid" USN binary package versions during `merge_usn_released_binary_package_versions`.

What this ends up doing if esm-apps is beta and it has a bin_pkg version > non-beta pocket, we've already ignored the "lesser" binary pkg version from the "non-beta" pocket.


specific to the krb5-locales installed as base xenial version, ualient ends up ignoring both xenial-updates/security version and esm-apps if apps is beta.

We need to still honor xenial-updates versions and disregard beta pockets during merge_usn...
Here's a path that shows this working, ultimately we want a `ua fix CVE-2020-28196` to upgrade to xenial-updates/security if uaclient.conf has security params config and no "features:allow_beta: true"



```diff
diff --git a/uaclient/security.py b/uaclient/security.py
index 6de86e42..0eb597c7 100644
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -473,7 +473,7 @@ def query_installed_source_pkg_versions() -> "Dict[str, Dict[str, str]]":
 
 
 def merge_usn_released_binary_package_versions(
-    usns: "List[USN]"
+    usns: "List[USN]", beta_pockets: "Dict[str, bool]"
 ) -> "Dict[str,  Dict[str, Dict[str, str]]]":
     """Walk related USNs, merging the released binary package versions.
 
@@ -482,6 +482,10 @@ def merge_usn_released_binary_package_versions(
         require differnt version fixes to the same binary package, track the
         maximum version required across all USNs.
 
+    :param usns: List of USN response instances from which to calculate merge.
+    :param beta_pockets: Dict keyed on service name: esm-infra, esm-apps
+        the values of which will be true of USN response instances from which to calculate merge.
+
     :return: Dict keyed by source package name. Under each source package will
         be a dict with binary package name as keys and binary package metadata
         as the value.
@@ -489,22 +493,29 @@ def merge_usn_released_binary_package_versions(
     usn_pkg_versions = {}
     for usn in usns:
         # Aggregate USN.release_package binary versions into usn_pkg_versions
+        # pocket = usn_released_pkg.get("pocket")
         for src_pkg, binary_pkg_versions in usn.release_packages.items():
-            if src_pkg not in usn_pkg_versions:
-                usn_pkg_versions[src_pkg] = binary_pkg_versions
+            public_bin_pkg_versions = {
+                bin_pkg_k: bin_pkg_v
+                for bin_pkg_k, bin_pkg_v in binary_pkg_versions.items()
+                if False
+                is beta_pockets.get(bin_pkg_v.get("pocket", "None"), False)
+            }
+            if src_pkg not in usn_pkg_versions and public_bin_pkg_versions:
+                usn_pkg_versions[src_pkg] = public_bin_pkg_versions
             else:
                 # Since src_pkg exists, only record this USN's binary version
                 # when it is greater than the previous version in usn_src_pkg.
                 usn_src_pkg = usn_pkg_versions[src_pkg]
-                for binary_pkg, binary_pkg_md in binary_pkg_versions.items():
-                    if binary_pkg not in usn_src_pkg:
-                        usn_src_pkg[binary_pkg] = binary_pkg_md
+                for bin_pkg, binary_pkg_md in public_bin_pkg_versions.items():
+                    if bin_pkg not in usn_src_pkg:
+                        usn_src_pkg[bin_pkg] = binary_pkg_md
                     else:
-                        prev_version = usn_src_pkg[binary_pkg]["version"]
+                        prev_version = usn_src_pkg[bin_pkg]["version"]
                         current_version = binary_pkg_md["version"]
                         if not version_cmp_le(current_version, prev_version):
                             # binary_version is greater than prev_version
-                            usn_src_pkg[binary_pkg] = binary_pkg_md
+                            usn_src_pkg[bin_pkg] = binary_pkg_md
     return usn_pkg_versions
 
 
@@ -512,6 +523,12 @@ def fix_security_issue_id(cfg: UAConfig, issue_id: str) -> None:
     issue_id = issue_id.upper()
     client = UASecurityClient(cfg=cfg)
     installed_packages = query_installed_source_pkg_versions()
+
+    # Used to filter out beta pockets during merge_usns
+    beta_pockets = {
+        "esm-apps": _is_pocket_used_by_beta_service("esm-apps", cfg),
+        "esm-infra": _is_pocket_used_by_beta_service("esm-infra", cfg),
+    }
     if "CVE" in issue_id:
         try:
             cve = client.get_cve(cve_id=issue_id)
@@ -527,7 +544,9 @@ def fix_security_issue_id(cfg: UAConfig, issue_id: str) -> None:
             cve=cve, installed_packages=installed_packages
         )
         print(cve.get_url_header())
-        usn_released_pkgs = merge_usn_released_binary_package_versions(usns)
+        usn_released_pkgs = merge_usn_released_binary_package_versions(
+            usns, beta_pockets
+        )
     else:  # USN
         related_usns = {}
         try:
@@ -549,7 +568,9 @@ def fix_security_issue_id(cfg: UAConfig, issue_id: str) -> None:
         affected_pkg_status = get_usn_affected_packages_status(
             usn=usn, installed_packages=installed_packages
         )
-        usn_released_pkgs = merge_usn_released_binary_package_versions(usns)
+        usn_released_pkgs = merge_usn_released_binary_package_versions(
+            usns, beta_pockets
+        )
         print(usn.get_url_header())
         related_cves = set(itertools.chain(*[u.cves_ids for u in usns]))
         if not related_cves:
@@ -730,9 +751,9 @@ def _format_packages_message(
 
 def _get_service_for_pocket(pocket: str, cfg: UAConfig):
     service_to_check = "no-service-needed"
-    if pocket == UA_INFRA_POCKET:
+    if pocket in ("esm-infra", UA_INFRA_POCKET):
         service_to_check = "esm-infra"
-    elif pocket == UA_APPS_POCKET:
+    elif pocket in ("esm-apps", UA_APPS_POCKET):
         service_to_check = "esm-apps"
 
     ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(service_to_check)
```

blackboxsw (changes requested):
Thanks for this @lucasmoura I think we still need to fallback to xenial-updates if pocket is beta. Best place to do that is in merge_usn_released.... function.

uaclient/security.py
```diff
@@ -740,17 +777,36 @@ def _handle_released_package_fixes(
...
             binary_pkgs = binary_pocket_pkgs[pocket]
+            pkg_status_override = None
 
             if upgrade_status:
+                # we don't want to proceed with services that are still
+                # in beta. The reason is that we are unable to use those
+                # services by default, meaning that the user could potentially
+                # attach to a subscription and not install the fixes.
+                beta_service = False
+                if binary_pkgs and _is_pocket_used_by_beta_service(
```
blackboxsw, line 793: I think we can drop this logic in general if we are doing the beta checks in `merge_usn_released_binary_package_versions`

blackboxsw (approved):
+1 on this. confirmed proper ignore of apps-related content when esm-apps is beta and falls back to xenial-updates in that case for attached and unattached.

Issue #1550, opened 2021-04-14
non-root: ua status fails trying to write /var/lib/ubuntu-advantage/notices.json due to permissions errors


blackboxsw:
Non-root should never be updating notices.json. We probably need a os.getuid check before trying to remove_notices

This is probably due to our [remove_notice changes here](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/config.py#L545)

Otherwise:
```bash
ubuntu@base-xenial:~$ ua --debug status
DEBUG: Executed with sys.argv: ['/usr/bin/ua', '--debug', 'status']
DEBUG: Reading file: /var/lib/ubuntu-advantage/notices.json
DEBUG: Writing file: /var/lib/ubuntu-advantage/notices.json
ERROR: Unhandled exception, please file a bug
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1001, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1059, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 886, in action_status
    status = cfg.status(show_beta=show_beta)
  File "/usr/lib/python3/dist-packages/uaclient/config.py", line 534, in status
    operation="fix operation"
  File "/usr/lib/python3/dist-packages/uaclient/config.py", line 176, in remove_notice
    self.write_cache("notices", notices)
  File "/usr/lib/python3/dist-packages/uaclient/config.py", line 347, in write_cache
    util.write_file(filepath, content, mode=mode)
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 628, in write_file
    with open(filename, "wb") as fh:
PermissionError: [Errno 13] Permission denied: '/var/lib/ubuntu-advantage/notices.json'
Unexpected error(s) occurred.
For more details, see the log: /var/log/ubuntu-advantage.log
To file a bug run: ubuntu-bug ubuntu-advantage-tools
```

PR #1551, opened 2021-04-14
status: do not remove notices on non-root call

lucasmoura:
## Proposed Commit Message
status: do not remove notices on non-root call

When we run ua status we try to remove notices that are created by fix operations. However, non-root users do not have the right permissions to remove them. We are now fixing the code to only attempt to remove the notices if ua status is run by root users.

Fixes: #1550

## Test Steps
Run the modified tests

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -540,15 +540,6 @@ class UAConfig:
         Write the status-cache when called by root.
         """
 
-        # Try to remove fix reboot notices if not applicable
-        if not util.should_reboot():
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
LGTM and tests well on affected systems.

orndorffgrant (approved):
Worked great for me

lucasmoura:
Ci is failing for non related issue

Issue #1552, opened 2021-04-15
manual sru testing: confirm the following upgrade cases from old-client to new client 


blackboxsw:
Write a manual upgrade script which confirms the following on upgrade from existing ua-tools in `xenial-updates` to ppa:ua-client/daily:

- [ ] **unattached old client** -> on upgrade to daily doesn't traceback , fail to install, incorrectly setup default state for esm-infra/esm-apps, leave old-client APT * artifacts around unmanaged.
- [ ] **attached old client to esm-infra** -> on upgrade to daily doesn't traceback , fail to install, incorrectly setup default state for esm-infra/esm-apps, `ua status` shows  unattached still

PR #1553, opened 2021-04-15
test: manual test script for xenial upgrade

orndorffgrant:
## Proposed Commit Message

> test: manual test script for xenial upgrade
>
> When v27 is introduced to xenial, it will be installed over v10 which
> is an older version of the ua client based on bash instead of python.
> This test script runs through upgrading the old version to v27 to
> validate that nothing breaks in the process.
>
> Fixes: #1552 

>  fips: add notice when installing over old fips
>
> When fips was enabled with the old client, it used a
> different repo. We are now detecting if this old repo
> is being used and if so, providing a notice for the user
> to enable fips with the new client, which will update
> the repo appropriately.

## Test Steps

`bash tools/test_xenial_upgrade.sh` and follow along with steps. Verify nothing unexpected occurs.

Run new unit tests

## Desired commit type::

## Checklist:

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -21,6 +21,9 @@ ESM_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-trusty.list"
 ESM_INFRA_OLD_APT_SOURCE_FILE_TRUSTY="$APT_SRC_DIR/ubuntu-esm-infra-trusty.list"
 ESM_INFRA_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-esm-infra.list"
 ESM_APPS_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-esm-apps.list"
+FIPS_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-fips.list"
+
+OLD_CLIENT_FIPS_PPA="private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu"
 
 ESM_APT_PREF_FILE_TRUSTY="$APT_PREFERENCES_DIR/ubuntu-esm-trusty"
 ESM_INFRA_OLD_APT_PREF_FILE_TRUSTY="$APT_PREFERENCES_DIR/ubuntu-esm-infra-trusty"
@@ -201,17 +204,22 @@ mark_reboot_for_fips_pro() {
 }
 
 
-mark_reboot_cmds_as_needed() {
+add_notice() {
     msg_name=$1
-    if [ ! -f "$REBOOT_CMD_MARKER_FILE" ]; then
-      touch $REBOOT_CMD_MARKER_FILE
-    fi
     python3 -c "
 from uaclient.config import UAConfig
 from uaclient.status import ${msg_name}
 cfg = UAConfig()
 cfg.add_notice(label='', description=${msg_name})
 "
+}
+
+mark_reboot_cmds_as_needed() {
+    msg_name=$1
+    if [ ! -f "$REBOOT_CMD_MARKER_FILE" ]; then
+      touch $REBOOT_CMD_MARKER_FILE
+    fi
+    add_notice "$msg_name"
 }
 
 case "$1" in
@@ -247,6 +255,13 @@ case "$1" in
       # specialize apt config filenames per ubuntu release.
       redact_ubuntu_release_from_ua_apt_filenames $APT_SRC_DIR
       redact_ubuntu_release_from_ua_apt_filenames $APT_PREFERENCES_DIR
+
+      # Repo for FIPS packages changed from old client
+      if [ -f $FIPS_APT_SOURCE_FILE ]; then
+        if grep -q $OLD_CLIENT_FIPS_PPA $FIPS_APT_SOURCE_FILE; then
+            add_notice MESSAGE_FIPS_INSTALL_OUT_OF_DATE
+        fi
+      fi
 
       # CACHE_DIR is no longer present or used since 19.1
       rm -rf /var/cache/ubuntu-advantage-tools
--- a/tools/test_xenial_upgrade.sh
+++ b/tools/test_xenial_upgrade.sh
@@ -1,0 +1,224 @@
+#!/usr/bin/bash
+
+set -x
+set -e
+name=test-xenial-ua-upgrade
+
+function h1() {
+    set +x
+    echo ""
+    echo ""
+    echo ""
+    echo "############################################################################"
+    echo "## $1"
+    echo "############################################################################"
+    echo ""
+    set -x
+}
+function h2() {
+    set +x
+    echo ""
+    echo ""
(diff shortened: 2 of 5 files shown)


orndorffgrant:
Hi @blackboxsw I put up a first wip version of this

Let me know if that is the kind of script we're looking for not.

Also, when I run `ubuntu-advantage enable-esm` on a xenial container, I get the message "Sorry, but Extended Security Maintenance is not supported on xenial". Does that mean we don't need to test the case where esm-infra is already enabled as described in #1552 

Are there any other cases we should include in this script?

orndorffgrant:
We discussed out of band and decided:
1. ESM-infra wasn't supported on xenial before so we don't need to test that upgrade path
2. We should test livepatch enabled on old -> livepatch enabled with new
    * ensure livepatch still enabled, token replaced with new token from contract server after attach
3. We should test fips enabled on old -> fips enabled on new

lucasmoura:
@orndorffgrant One thing that will be good to debug is understand why we need to manually enable FIPS for status to show it as disabled. Maybe we there is a gap that we can cover here or maybe not, but I think it is still worth understanding why this is happening.

orndorffgrant:
@lucasmoura @blackboxsw I dug into why ua isn't reporting fips as enabled when it was enabled by the old client.

The old client adds the following ppa when enabling fips: `https://private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu`

The new client looks for this repo in `apt-cache policy`: `https://esm.ubuntu.com/fips`, which isn't there because thats not how the old client enabled fips.

And I believe fips is not marked for enable by default, which is why `ua attach` doesn't go ahead and enable it.

We could potentially detect this scenario and auto enable fips with the new repo whenever we see the old ppa in `apt-cache policy`, or we could advise users to run `ua enable fips` to set up the new repo. What do you think?

blackboxsw:
> @lucasmoura @blackboxsw I dug into why ua isn't reporting fips as enabled when it was enabled by the old client.
> 
> The old client adds the following ppa when enabling fips: `https://private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu`
> 
> The new client looks for this repo in `apt-cache policy`: `https://esm.ubuntu.com/fips`, which isn't there because thats not how the old client enabled fips.
> 
> And I believe fips is not marked for enable by default, which is why `ua attach` doesn't go ahead and enable it.
> 
> We could potentially detect this scenario and auto enable fips with the new repo whenever we see the old ppa in `apt-cache policy`, or we could advise users to run `ua enable fips` to set up the new repo. What do you think?

We discussed out of band on this and I think the desired approach would be the following on upgrade from old-client with an attached fips service
```bash
if [ -e  /etc/apt/sources.list.d/ubuntu-fips-xenial.list]; then
    mv   /etc/apt/sources.list.d/ubuntu-fips-xenial.list > /etc/apt/sources.list.d/ubuntu-fips.list 
   emit_notice "This FIPS install is out of date, run: sudo ua enable fips"
fi
```

What this mv leverages is the following:
1. old client stores fips APT credentials in /etc/apt/sources.list.d/ubuntu-fips-xenial.list , so we can't deleted it in case the customer doesn't intend to run `ua attach <token>`
2. moving the file to /etc/apt/sources.list.d/ubuntu-fips.list ensures that whenver `ua enable fips` is called newclient will overwrite the current apt sources file and also add /etc/apt/auth.conf.d/90ubuntu-advatage.cfg authentication entries

tools/test_xenial_upgrade.sh
```diff
@@ -0,0 +1,219 @@
...
+    h2 "Enable fips on old version"
+    set +x
+    echo "+ multipass exec $name -- ubuntu-advantage enable-fips \$FIPS_CREDS"
+    multipass exec $name -- sudo ubuntu-advantage enable-fips $FIPS_CREDS
+    set -x
+
+    h2 "Reboot to finish fips activation"
+    set +e
+    multipass exec $name -- sudo reboot
+    set -e
```
blackboxsw, line 174: You don't need the set +e -e bookends if you do this:
```suggestion
    multipass exec $name -- sudo reboot || true
```

tools/test_xenial_upgrade.sh
```diff
@@ -0,0 +1,219 @@
...
+    multipass exec $name -- sudo ua status
+    multipass exec $name -- sudo cat /proc/sys/crypto/fips_enabled
+
+    h2 "Attach - only esm-infra will be auto-enabled"
+    attach_ua multipass
+
+    h2 "Status after attach - new UA will say fips is disabled and that livepatch is n/a"
+    multipass exec $name -- sudo ua status
+    multipass exec $name -- sudo cat /proc/sys/crypto/fips_enabled
+
```
blackboxsw, line 193: Can we add a check that our postinst has migrated oldclient FIPS APT sources to newclient and that apt update doesn't error
```bash
multipass exec $name -- sudo grep $FIPS_CREDS /etc/apt/sources.list.d/ubuntu-fips.list && echo "SUCCESS: found oldclient FIPS creds" || echo "FAIL: Migrated to new client creds"
multipass exec $name -- sudo apt update
```

tools/test_xenial_upgrade.sh
```diff
@@ -0,0 +1,219 @@
...
+    h2 "Enable fips on new UA - This will re-install fips packages and ask to reboot again"
+    multipass exec $name -- sudo ua enable fips --assume-yes
+
```
blackboxsw, line 196: Let's assert that apt is good and we don't have FIPS_CREDS in the new sources-list.file
```bash
multipass exec $name -- sudo grep $FIPS_CREDS /etc/apt/sources.list.d/ubuntu-fips.list && echo "FAIL: found oldclient FIPS creds" || echo "SUCCESS: Migrated to new client creds"
multipass exec $name -- sudo apt update
## also check to make sure we have a valid ubuntu-*-fips metapackage installed
grep "install" /var/log/ubuntu-advantage.log
```

blackboxsw (changes requested):
Loooks really good @orndorffgrant minor test suggestions, I'm testing now.

tools/test_xenial_upgrade.sh
orndorffgrant: Right! Good tip

tools/test_xenial_upgrade.sh
orndorffgrant: Yes good idea

debian/ubuntu-advantage-tools.postinst
```diff
@@ -248,6 +256,13 @@ case "$1" in
       redact_ubuntu_release_from_ua_apt_filenames $APT_SRC_DIR
       redact_ubuntu_release_from_ua_apt_filenames $APT_PREFERENCES_DIR
 
+      # Repo for FIPS packages changed from old client
+      if [ -f $FIPS_APT_SOURCE_FILE ]; then
```
blackboxsw, line 260: Good update

blackboxsw (approved):
This looks good and tests well on attached/unattached upgrade from old client.

PR #1556, opened 2021-04-19
Ship esm-infra pref file only on xenial and trusty

lucasmoura:
## Proposed Commit Message
Ship esm-infra pref file only on xenial and trusty

Currently, we are shipping the esm-infra preference file on all LTS releases. However, our codebase only supports that preference file for trusty and xenial. Because of that, we are updating our postinst script to onoy ship that file for trusty and xenial

## Test Steps
Run the `Attach command in a ubuntu lxd container` scenario on `attach_valitoken.feature` without this change and with this change

## Desired commit type::

## Checklist:

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -82,13 +82,6 @@ print(*ENTITLEMENT_CLASS_BY_NAME.keys(), sep=' ')
 }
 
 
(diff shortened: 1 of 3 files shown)


blackboxsw (approved):
Looks good, let's land this and allow folks who test on bionic ++ to setup their own apt pref files for unauthenticated access to esm-* repos.

PR #1559, opened 2021-04-21
Esm apps all lts

blackboxsw:
## Proposed Commit Message  

esm: enable infra when EOL LTS and apps on all LTS
    
    Enable showing unauthenticated APT package counts for the following:
     - ESM Infra on any end of life Ubuntu LTS release
     - ESM Apps when non-beta on any Ubuntu LTS release except Trusty
    
    Fixes: #1558

---
    util: add is_lts and is_active_esm funtions to support ESM
    
    Shell out to ubuntu-distro-info to determine the following:
     - is_lts: the provided Ubuntu release is an LTS release
     - is_active_esm: the provided Ubuntu release has entered EOL and is
                      entitled to ESM Infra

---

    d/postinst: configure esminfra when is_active_esm and apps on LTS
    
    use ubuntu-distro-info to check_is_lts and check_is_active_esm.
    When on LTS, setup unauthenticated ESM Apps APT sources when
    apps is not beta.
    
    Setup ESM Infra on pkg install when ESM lifecycle has started
    via ubuntu-distro-info --series -yeol.

## Desired commit type::

## Checklist:

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -85,6 +85,28 @@ print(*ENTITLEMENT_CLASS_BY_NAME.keys(), sep=' ')
 }
 
 
+# Ubuntu LTS release all support-esm
+check_is_lts() {
+    release_name=$1
+    ubuntu-distro-info --supported-esm | grep -q "${release_name}"
+}
+
+
+# Check whether this series is under active ESM
+check_is_active_esm() {
+    release_name=$1
+    # Trusty doesn't support --series param
+    if [ "${release_name}" = "trusty" ]; then
+        return 0
+    else
+        _DAYS_UNTIL_ESM=$(ubuntu-distro-info --series "${release_name}" -yeol)
+        if [ "${_DAYS_UNTIL_ESM}" -lt "1" ]; then
+            return 0
+        fi
+    fi
+    return 1
+}
+
 # Check whether a given service is beta
 check_service_is_beta() {
     service_name=$1
@@ -205,17 +227,14 @@ EOF
 
 configure_esm() {
     rm -f $APT_TRUSTED_KEY_DIR/ubuntu-esm*gpg  # Remove legacy esm keys
-    if [ "14.04" = "${VERSION_ID}" ]; then
-            enable_apps="false"
-    else
-            enable_apps="true"
+    if check_is_active_esm "${UBUNTU_CODENAME}"; then
+        install_esm_apt_key_and_source "infra" "$UBUNTU_CODENAME"
     fi
-    if [ "${enable_apps}" = "true" ]; then
-        if ! check_service_is_beta esm-apps; then
+    if ! check_service_is_beta esm-apps; then
+        if [ "${UBUNTU_CODENAME}" != "trusty" ]; then
             install_esm_apt_key_and_source "apps" "$UBUNTU_CODENAME"
         fi
     fi
-    install_esm_apt_key_and_source "infra" "$UBUNTU_CODENAME"
 }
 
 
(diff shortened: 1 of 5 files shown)


debian/ubuntu-advantage-tools.postinst
```diff
@@ -205,17 +227,12 @@ EOF
...
-            enable_apps="true"
+    if check_is_active_esm "${UBUNTU_CODENAME}"; then
+        install_esm_apt_key_and_source "infra" "$UBUNTU_CODENAME"
     fi
-    if [ "${enable_apps}" = "true" ]; then
-        if ! check_service_is_beta esm-apps; then
-            install_esm_apt_key_and_source "apps" "$UBUNTU_CODENAME"
-        fi
+    if ! check_service_is_beta esm-apps; then
+        install_esm_apt_key_and_source "apps" "$UBUNTU_CODENAME"
```
orndorffgrant, line 234: Doe we also need to check to make sure we're not on trusty before calling `install_esm_apt_key_and_souce "apps"` ?

uaclient/entitlements/esm.py
```diff
@@ -22,15 +22,30 @@ class ESMAppsEntitlement(ESMBaseEntitlement):
...
+            config=self.cfg.cfg, path_to_value="features.allow_beta"
+        )
+        if config_allow_beta or self.is_beta is False:
+            if util.is_lts(series):
+                return "never"
+        return None
 
     @property
     def disable_apt_auth_only(self) -> bool:
         """Only xenial esm-apps should remove apt auth files upon disable"""
```
orndorffgrant, line 39: I think this comment needs updating to something like
```suggestion
        """All LTS with the exception of Trusty should remove apt auth files upon disable"""
```

debian/ubuntu-advantage-tools.postinst
```diff
@@ -205,17 +227,12 @@ EOF
...
-    else
-            enable_apps="true"
+    if check_is_active_esm "${UBUNTU_CODENAME}"; then
+        install_esm_apt_key_and_source "infra" "$UBUNTU_CODENAME"
     fi
-    if [ "${enable_apps}" = "true" ]; then
-        if ! check_service_is_beta esm-apps; then
-            install_esm_apt_key_and_source "apps" "$UBUNTU_CODENAME"
-        fi
+    if ! check_service_is_beta esm-apps; then
```
orndorffgrant, line 233: So currently, if you want to use esm-apps, you'll need to configure allow_beta _before installing_ for esm-apps to get set up - is that right? I just want to call that out to make sure it is what we want here.

uaclient/entitlements/tests/test_esm.py
```diff
@@ -21,57 +21,156 @@ def entitlement(request, entitlement_factory):
...
+            # When esm non-beta pin it on non-trusty
+            ("trusty", True, False, None, None),
+            ("xenial", True, False, None, "never"),
+            ("bionic", True, False, None, "never"),
+            ("focal", True, False, None, "never"),
+            # when ESM beta don't pin
+            ("trusty", True, True, None, None),
+            ("xenial", True, True, None, None),
+            ("bionic", True, True, None, None),
+            ("focal", True, True, None, None),
```
orndorffgrant, line 68: Do we want cases where `cfg_allow_beta` is true?

orndorffgrant (commented):
Looks great! Just left some nits and clarification questions

debian/ubuntu-advantage-tools.postinst
blackboxsw: Correct assumption.
in order to test this on pkg install we'd need to add features:\n  allow_beta: true to `/etc/ubuntu-advantage/uaclient.conf` prior to this package install. This can be done easily in the pkg ugprade from existing system, or at launch time with
```yaml
#cloud-config    
write_files:
 - path: /etc/ubuntu-advantage/uaclient.conf
   content: |
        features:
            allow_beta: true
   append: true

debian/ubuntu-advantage-tools.postinst
blackboxsw: Nice catch, yes. just added.

orndorffgrant (approved):
Good stuff! My tests worked. Approving pending jenkins.

PR #1560, opened 2021-04-21
apt-hook: new json hook for security update counts

orndorffgrant:
## Proposed Commit Message

> apt-hook: new json hook for security update counts
>
> This new hook takes advantage of the newly created
> "stats" json hook in apt to print a message about the
> number of security updates to be installed on an
> apt upgrade.

### Running tests and building
```shell
# make sure it builds
make -C apt-hook json-hook
# run unit tests
make -C apt-hook test
# integration test to make sure it is installed correctly
env UACLIENT_BEHAVE_BUILD_PR=1 tox -e behave-lxd-16.04 -- features/unattached_commands.feature:103
```

### See it in action
In a fresh hirsute container, install this version of uaclient, then
```shell
add-apt-repository ppa:deity/sid # ppa with new version of apt
apt-get update
apt upgrade
apt install distro-info-data=0.46
apt upgrade
```
And you should see "1 standard security update" underneath the stats

## Desired commit type::

## Checklist:


apt-hook/json-hook.go
```diff
@@ -0,0 +1,628 @@
...
+	if esmAppsCount == 0 {
+		displayEsmApps = false
+	}
+	if esmInfraCount == 0 {
+		displayEsmInfra = false
+	}
+
+	// There are probably more clever ways of breaking down these messages, but this makes
+	// it easy to follow the minor variations
+	if displayStandard && displayEsmApps && displayEsmInfra {
```
lucasmoura, line 66: I think I have an idea to make the code a little simpler. We could built each service message separately and then just figure out how to connect them. My idea would be like this (Sorry, will use python here because I am go noob)

```python
standard_msg = build_standard_message(standardSecurityCount) # x standard security update
esm_infra_msg = build_esm_infra_message(esmInfraCount)  # x esm-infra update
esm_apps_msg = build_esm_apps_message(esmAppsCount)  # x esm-apps update

if not standard_message_msg
    standard_msg_unifier = ""
elif esm_infra_msg and esm_apps_message:
    standard_msg_unifier = ", "
elif esm_infra_msg or esm_apps_message:
    standard_msg_unifier = " and "
else:
    standard_msg_unifier = ""

if standard_msg_unifier == ", ":
    esm_unifier = " and "
elif standard_msg_unifier == " and ":
    esm_unifier = ""
elif esm_infra_msg and esm_apps_msg:
   esm_unifier = "and"

return standard_security_msg + standard_msg_unifier + esm_infra_msg + esm_msg_unifier + esm_apps_msg
```

lucasmoura: I think the logic will hold, but maybe there are gaps on it that I could not see

lucasmoura (commented):
That looks really good @orndorffgrant :clap: :clap: 

I think after we refactor the code to use back the actual rpc payload instead of the mock, we could create an integration test to verify that the hook is successfully added to the user system and it is also working as expected.

apt-hook/json-hook.go
```diff
@@ -0,0 +1,628 @@
...
+}
+
+func createUpdateMessage(standardSecurityCount int, esmAppsCount int, esmInfraCount int) string {
+	displayStandard := true
+	// TODO other lts/beta criteria for displaying this?
+	displayEsmApps := true
+	displayEsmInfra := true
+
+	if standardSecurityCount == 0 {
+		displayStandard = false
```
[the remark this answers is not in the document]
orndorffgrant: Somewhat controversially, the entire go community has decided to use tabs and enforces it by standardizing on the `gofmt` formatting tool :shrug:

apt-hook/json-hook.go
orndorffgrant: Thanks for this! I updated the logic to construct the message in a similar manner to what you suggest here.

apt-hook/json-hook-src/json-hook.go
```diff
@@ -0,0 +1,249 @@
...
+			esmAppsUpdates += "updates"
+		} else {
+			esmAppsUpdates += "update"
+		}
+	}
+
+	return standardUpdates + esmInfraUpdates + esmAppsUpdates
+}
+
+func fromOrigin(pkgVersion jsonRPCPackageVersion, origin string) bool {
```
lucasmoura, line 123: I think we can merge both of these function and always check both origin and archive just to be safe.

debian/control
```diff
@@ -9,6 +9,7 @@ Build-Depends: bash-completion,
                dh-systemd | debhelper (>= 13.3),
                gettext,
                git,
+               golang (>=1.6),
```
lucasmoura, line 12: This will probably break the trusty package, as golang there is 1.6.

Just to confirm, do we need version 1.6 specifically on 1.3 (which is present in trusty)  would work too ?

apt-hook/json-hook-src/json-hook.go
orndorffgrant: @lucasmoura I'm not sure I understand what this comment is suggesting - can you describe what parameters a combined function would take and what it would do?

lucasmoura: @orndorffgrant Sorry for that. I mean that we can combine both `fromOrigin`  and `isSecurityUpdate`. This will allows to check both `origin` and `archive` for each security service we will provide the message.

I think ti should behave similarly to this function in update-notifier:
```python
def isSecurityUpgrade(ver):
    " check if the given version is a security update (or masks one) "
    security_pockets = [("Ubuntu", "%s-security" % DISTRO),
                        (ESM_INFRA_ORIGIN, "%s-infra-security" % DISTRO),
                        (ESM_APPS_ORIGIN, "%s-apps-security" % DISTRO),
                        ("gNewSense", "%s-security" % DISTRO),
                        ("Debian", "%s-updates" % DISTRO)]
    for (file, index) in ver.file_list:
        for origin, archive in security_pockets:
            if (file.archive == archive and file.origin == origin):
                return True
    return False
```

But we can pass the expected archive and origin directly to the function

orndorffgrant: Ohhhh okay cool, yeah thats a good idea! I'll do this

debian/control
orndorffgrant: I removed the version constraint and it seems to be building fine on trusty
There seems to be an unrelated building error on focal though which I'm trying to recreate now

apt-hook/json-hook-src/json-hook.go
orndorffgrant: I added this - let me know what you think

Somewhat related: The weird thing I had to do here was get the distro codename from the package origin, which isn't ideal, but I wasn't sure of a better way to do that - any thoughts?

apt-hook/json-hook-src/json-hook.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+		if esmAppsCount > 1 {
+			esmAppsUpdates += "updates"
+		} else {
+			esmAppsUpdates += "update"
+		}
+	}
+
+	return standardUpdates + esmInfraUpdates + esmAppsUpdates
+}
```
blackboxsw, line 121: It feels like this section if fairly repetitive.  It's good for this iteration and PR, but I'm wondering if we can iterate to do something like concatonate_strings([<updates_string>, <infra_str>, <apps_str]) and just have the comma spacing additions based on number of items in a list

orndorffgrant: This section is definitely repetitive, but I tried to balance that with the readability. The nice thing about this construction is that you can read it top to bottom and see what the message will be, rather than adding things in the middle later.

I think its worth revisiting this, but I'd rather take it as a follow up item if thats okay with you

blackboxsw (commented):
Code looks good and runs well on Hirsute. thanks for the testing instructions. I think we can simplify  createUpdateMessage a bit as it's carrying around a lot of extra variables and positional checks based on the ordering we have with standard, esminfra and esm apps in order to construct a single comma separated list. 

I'm ok with us going with this iteration and capturing a small work item to reduce that function a bit into a get_pkg_update_strings() and concatonated_strings if a separate branch Monday if you think it is unneeded

apt-hook/json-hook-src/json-hook.go
blackboxsw: +1 For sure. This is good for today! Thanks again

blackboxsw (approved):

Issue #1564, opened 2021-04-23
config: fragile handling of "features" in uaclient.conf. Invalid values break all ua commands


blackboxsw:
when adding the following invalid /etc/ubuntu-advantage/uaclient.conf features, I can break all ua client commands

```bash
cat > /etc/ubuntu-advantage/uaclient.conf <<EOF
features: null
EOF
ua status
ERROR:root:Unhandled exception, please file a bug
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1000, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1044, in main
    parser = get_parser()
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 814, in get_parser
    version=get_version(),
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 917, in get_version
    return version.get_version(features=_cfg.features)
  File "/usr/lib/python3/dist-packages/uaclient/version.py", line 28, in get_version
    for key, value in sorted(features.items()):
AttributeError: 'NoneType' object has no attribute 'items'
Unexpected error(s) occurred.
For more details, see the log: /var/log/ubuntu-advantage.log
To file a bug run: ubuntu-bug ubuntu-advantage-tools



```

PR #1565, opened 2021-04-23
config: avoid tracebacks on invalid features value in uaclient.conf

blackboxsw:
## Proposed Commit Message

If /etc/ubuntu-advantage/uaclient.conf contains unexpected values
in features: settings log a warning and return an empty dict.

Avoid tracing on every ua subcommand call.

Fixes: #1564

## Test Steps
```bash
lxc launch ubuntu-daily:xenial test-ua
lxc exec test-ua -- add-apt-repository ppa:ua-client/daily -y
lxc exec test-ua -- apt update
lxc exec test-ua -- apt install ubuntu-advantage-tools --assume-yes

cat > uaclient.conf <<EOF
features: null
EOF
lxc file push uaclient.conf test-ua/etc/ubuntu-advantage
lxc exec test-ua -- ua status   # expect to see failure
# apply the fix
lxc file push uaclient/config.py test-ua/usr/lib/python3/dist-packages/uaclient
lxc exec test-ua -- ua status   # expect no failure
```

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -248,7 +248,17 @@ class UAConfig:
     @property
     def features(self):
         """Return a dictionary of any features provided in uaclient.conf."""
-        return self.cfg.get("features", {})
+        features = self.cfg.get("features")
+        if features:
+            if isinstance(features, dict):
+                return features
+            else:
+                logging.warning(
+                    "Unexpected uaclient.conf features value."
+                    " Expected dict, but found %s",
+                    features,
+                )
+        return {}
 
     @property
     def machine_token(self):
--- a/uaclient/tests/test_util.py
+++ b/uaclient/tests/test_util.py
@@ -780,6 +780,8 @@ class TestIsConfigValueTrue:
         [
             ({}, False),
             ({}, False),
+            (None, False),
+            ({None}, False),
             ({"allow_beta": "true"}, True),
             ({"allow_beta": "True"}, True),
             ({"allow_beta": "false"}, False),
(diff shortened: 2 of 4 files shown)


uaclient/config.py
```diff
@@ -248,7 +248,16 @@ def contract_remaining_days(self) -> int:
...
         """Return a dictionary of any features provided in uaclient.conf."""
-        return self.cfg.get("features", {})
+        features = self.cfg.get("features")
+        if features:
+            if isinstance(features, dict):
+                return features
+            else:
+                logging.warning(
+                    "Unexpected uaclient.conf features value. Expected dict, but found %s",
+                    features,
```
lucasmoura, line 258: should we have here `str(features)`  instead ?

blackboxsw: I thought %s formatting does that for us right?

lucasmoura: Yep, it does. My mistake here

orndorffgrant (changes requested):
This change is working great, but it is still failing for me with `features: null` in uaclient.conf

I believe we need to handle None values better in `util.is_config_value_true` as well, which operates on the cfg dict directly

uaclient/tests/test_util.py
```diff
@@ -780,6 +780,7 @@ class TestIsConfigValueTrue:
         [
             ({}, False),
             ({}, False),
+            ({None}, False),
```
orndorffgrant, line 783: Do we need the curly braces here?
```suggestion
            (None, False),
```

orndorffgrant (approved):
Thanks for the extra fix - LGTM!

Issue #1567, opened 2021-04-23
apt-hook: make sure we are running go tests on the json hook during package build


blackboxsw:
Something in debian/rules should be exercising our go tests on the apt JSON hook to ensure our pkg builds break if we broke our json hook tests

PR #1572, opened 2021-04-26
apt-hook: run new json hook tests on package build

orndorffgrant:
## Proposed Commit Message

> apt-hook: run new json hook tests on package build
> 
> The new json hook is written in golang and has some
> go unit tests. We are now running these unit tests
> when building a deb to help prevent regressions.
> 
> While the go program itself does not require a
> particular newer version of go installed, the tests
> do. The `t.Run` function we are using requires at
> least version 1.7 to run. Because of this, we are
> adjusting our Build-Depends appropriately. We will
> now use `golang-1.10` if available (as it is in older
> releases where the default version is less than 1.7),
> and just use `golang` if not (in newer releases
> where the default version is greater than 1.7).
> 
> Fixes: #1567

> apt-hook: ignore build artifacts in git 

> apt-hook: refactor json hook messaging to be dry
> 
> The message construction code in the json hook
> was pretty repetitive. This cleans it up a bit
> and also removes some unnecessary explicit struct
> names in the tests.

> apt-hook: ignore go hook on some architectures
> 
> Golang doesn't support powerpc and riscv on older versions.
> This configures our Build-Depends and Makefile to gracefully
> handle scenarios where we can't build our go hook. We will
> simply not ship the go hook in these limited scenarios.
> 
> Fixes: #1603

> build: order build-depends alternatives newer first
> 
> When using alternatives in the Build-Depends section of
> debian/control, we should always put the newer version of
> a package first. This is because sbuild doesn't resolve
> alternatives by default and we want sbuild to work on the
> latest version of ubuntu with default settings.

> build: specify debianutils min version
> 
> This fixes a lintian error where we had an
> essential package that didn't specify version.

## Test Steps

1. Build on xenial through hirsute and make sure everything succeeds
2. Manually edit a go test to fail, then try to build and make sure the build fails

## Desired commit type::

## Checklist:


Jenkinsfile
```diff
@@ -87,7 +87,7 @@ pipeline {
                         cp debian/changelog ${WORKSPACE}/debian/changelog-${SERIES_VERSION}
                         sed -i "s/${PKG_VERSION}/${NEW_PKG_VERSION}/" ${WORKSPACE}/debian/changelog-${SERIES_VERSION}
                         dpkg-source -l${WORKSPACE}/debian/changelog-${SERIES_VERSION} -b .
-                        sbuild --nolog --verbose --dist=${BUILD_SERIES} --no-run-lintian --append-to-version=~${SERIES_VERSION}  ../ubuntu-advantage-tools*${NEW_PKG_VERSION}*dsc
+                        sbuild --resolve-alternatives --nolog --verbose --dist=${BUILD_SERIES} --no-run-lintian --append-to-version=~${SERIES_VERSION}  ../ubuntu-advantage-tools*${NEW_PKG_VERSION}*dsc
```
orndorffgrant, line 90: The `golang-1.10 | golang` in Build-Depends only works when sbuild gets this flag. `sbuild-it` passes this flag by default but I'm not sure if its safe to rely on it. Thoughts?

debian/control
```diff
@@ -9,7 +9,7 @@ Build-Depends: bash-completion,
                dh-systemd | debhelper (>= 13.3),
                gettext,
                git,
-               golang,
+               golang-1.10 | golang,
```
orndorffgrant, line 12: If this works how I hope it does, this might actually fix our architecture support woes as well. Since it appears that `golang-1.10` supports all architecures on those releases where `golang` does not (aka where `golang` is a version lower than 1.10).

orndorffgrant: @lucasmoura @blackboxsw What is the best way to test the builds for the architectures that are currently failing?

orndorffgrant: I retract my claim that this would fix the arch problem. I was looking at the wrong thing :upside_down_face:

Jenkinsfile
orndorffgrant: As discussed in this bug: https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1926949, it is okay to rely on --resolve-alternatives.

debian/control
```diff
@@ -9,7 +9,7 @@ Build-Depends: bash-completion,
                dh-systemd | debhelper (>= 13.3),
                gettext,
                git,
-               golang,
+               golang-1.14 [!powerpc] | golang-1.10 [!powerpc !riscv64],
```
orndorffgrant, line 12: This is why this _should_ work (lmk if you think this should be documented outside a github comment)
* go-1.14 is on focal and later
* go-1.14 supports risc v, and risc v ubuntu is supported on focal and later
* go-1.14 doesn't support powerpc but neither does focal or after
* go-1.10 is on bionic and earlier
* go-1.10 doesn't support risc v but neither does bionic or earlier
* go-1.10 doesn't support powerpc, so we wont deliver the hook on any powerpc bionic or earlier

So the scenarios are:
1. Any release on powerpc - nothing is satisfied, golang should be ignored and we won't deliver json-hook
2. Bionic or earlier on non-powerpc - golang-1.14 won't exist so use golang-1.10
3. Focal or later on anything (including riscv) - use first option golang-1.14

To test manually:
1. Simulate (1) above by replacing "powerpc" in this line with "amd64" and then testing a build on bionic or earlier
    - Expect a successful build and install. Verify that the jsonhook was not built and not installed
2. Simulate (2) above by using this line as-is, and testing a build on bionic or earlier
    - Expect a successful build and install. Verify that the jsonhook was built with 1.10 and installed
3. Simulate (3) above by replacing "riscv64" in this line with "amd64" and then testing a build on focal or later
    - Expect a successful build and install. Verify that the jsonhook was built with 1.14 and installed

Future issues:
* When we stop packaging go-1.14 for the latest release, we'll need to add another `|` alternative to this line.

Question: What is the best way to test this "for real" on actual/virtual powerpc/riscv machines?

orndorffgrant:
@lucasmoura @blackboxsw Okay I think this is ready for review now and covers all of our bases. The main outstanding question on my end is how to test the build in the same/similar environment as to the official builds that failed on powerpc/riscv64.

debian/control
paride: I don't think we should depend on fixed versions of packages, mainly for the future issue you also noted, but also for correctness (we don't need that *specific* version). What about Build-Depending on:
```
golang-go (>= 2:1.10~) | golang-1.13-go <!nocheck>, 
```
This should ensure that a recent enough version of `golang-go` gets installed, falling back to `golang-1.13-go` on Xenial, where `golang-go` would depend on the old `golang-1.6-go`. I hope it makes sense.

The `<!nocheck>` skips installing the dependency if the the test stage of the package build process is to be skipped. I think this is correct as it's actually a test dependency here (note the `nocheck` in the `override_dh_auto_test` target in `d/rules`).

Note that the arch is called `ppc64el`, while `powerpc` is the legacy one of the old Macs (long deprecated, see https://help.ubuntu.com/community/SupportedArchitectures#PowerPC_.28legacy.29).

debian/control
```diff
@@ -9,7 +9,7 @@ Build-Depends: bash-completion,
                dh-systemd | debhelper (>= 13.3),
```
paride: Let's swap these two as agreed at standup.

paride (changes requested):
There a lintian error to fix:
```
E: ubuntu-advantage-tools source: build-depends-on-essential-package-without-using-version Build-Depends: debianutils
```
which can be fixed by setting `debianutils (>= 4.7)` where 4.7 is the version in Xenial.

debian/control
orndorffgrant: @paride Thanks for this feedback! This is helpful. I like your suggestion and I agree we should have `golang-go` as the first choice rather than a versioned package. I also think that its a good idea to use the `golang-go` packages instead of the whole `golang` packages. However, this suggestion won't quite satisfy all of our requirements. I'll list the requirements here as I understand them.

* To build on most architectures, we need >=1.4 (or so, I haven't actually tested the lower bound here).
* To test on any arch, we need >=1.7
* To build for `riscv64` we need >= 1.14 (and riscv is on focal and newer)
* We need to gracefully handle that no version of go supports `powerpc`
    * My understanding was that the issue is with the legacy `powerpc` builds failing for xenial, not the newer `ppc64el`, which I think should be fine. **I will double check this though.**

So I think with `golang-go (>= 2:1.10~) | golang-1.13-go`, we'd end up with `golang-go=2:1.13~1ubuntu2` on focal which would fail to build `riscv64`, and we would still fail the `powerpc` builds.

But based on your suggestion, I think the following would work:
```
golang-go (>= 2:1.14~) [!powerpc] | golang-1.14-go [!powerpc] | golang-1.13-go [!powerpc !riscv64],
```
So we'd get `golang-go` on groovy and newer, we'd get `golang-1.14-go` on focal, and we'd get `golang-1.13-go` on xenial. We'd completely ignore the dependency for `powerpc` (still going to double check that). And it explicitly states that 1.13 won't support riscv, though I'm not sure if that bit is necessary.

What do you think?

debian/control
orndorffgrant: Yep! Done

apt-hook/json-hook-src/json-hook_test.go
```diff
@@ -14,33 +14,33 @@ func TestCreateUpdateMessages(t *testing.T) {
 		expectedMessage       string
 	}
 	testParamsList := []params{
-		params{0, 0, 0, ""},
```
blackboxsw, line 17: any change here besides dropping "params"

blackboxsw (changes requested):
can confirm package build failures on upstrema/main w/ golang on xenial powerpc locally in my schroots
 via `sbuild --dist=xenial --arch=powerpc  --arch-all ../out/ubuntu-advantage-tools_27.0.dsc`. 

Can confirm successful build start with this branch, though powerpc has a [different error message which include /bin/ls](https://paste.ubuntu.com/p/C6v6jkxQ5f/) that causes one unit test to fail. I think we need to add the following patch to be a bit more flexible (we could use re.match instead if you wanted)

```diff
diff --git a/uaclient/tests/test_util.py b/uaclient/tests/test_util.py
index 019faf2b..2a4b783f 100644
--- a/uaclient/tests/test_util.py
+++ b/uaclient/tests/test_util.py
@@ -291,11 +291,12 @@ class TestSubp:
         with pytest.raises(util.ProcessExecutionError) as excinfo:
             util.subp(["ls", "--bogus"])
 
-        expected_error = (
-            "Failed running command 'ls --bogus' [exit(2)]."
-            " Message: ls: unrecognized option"
-        )
-        assert expected_error in str(excinfo.value)
+        expected_errors = [
+          "Failed running command 'ls --bogus' [exit(2)].",
+          "ls: unrecognized option '--bogus'",
+        ]
+        for msg in expected_errors:
+            assert msg in str(excinfo.value)
         assert 0 == m_sleep.call_count  # no retries
 
     @mock.patch("uaclient.util.time.sleep")

File: apt-hook/json-hook-src/json-hook_test.go (hunk 5 above)
Comment (orndorffgrant) [reply apt-hook/json-hook-src/json-hook_test.go:17]: Just dropped "params" (I installed a go linter and my editor was yelling at me about it, so I thought I'd squeeze it in here)

File: apt-hook/Makefile [hunk 6]
```diff
@@ -1,3 +1,21 @@
+# the go hook steps will be skipped if this is a nonempty string
+SKIP_GO_HOOK = ""
+
+GO_BIN = $(shell which go)
+# Try go-1.14 and go-1.10 if not in path
+ifeq ($(GO_BIN),)
+	ifneq ($(wildcard /usr/lib/go-1.14/bin/go),)
+		GO_BIN = /usr/lib/go-1.14/bin/go
+	else ifneq ($(wildcard /usr/lib/go-1.10/bin/go),)
```
blackboxsw, line 9: Why are we looking at go-1.10 here instead of go-1.13 per what we are requiring in d/control?

orndorffgrant: Shoot, I forgot to update this after the discussion with Paride resulted in a change to the dependency line. I will fix this

blackboxsw (changes requested):
One other nit: in the commit message for ebb96728ba03bca2724191ed058cce1e63f66098 can you add the footer: for the bug that this should fix

LP: #1926949

orndorffgrant:
@blackboxsw Okay I think I addressed all of your changes and made sure flake was happy with the pytest change

orndorffgrant:
The packaging step on jenkins failed for bionic with go-1.13. I've built it myself on bionic with 1.13, but the fact that the build failed in a different environment is concerning. I just changed the backup version to 1.10 because there's no particular reason we need 1.13 over 1.10, and 1.10 has built successfully for bionic on jenkins before.

blackboxsw:
Builds look good on powerpc xenial and riscv64 on hirsute for me in local sbuild envs.

Riscv64 build
```bash
Build Architecture: riscv64
Build Type: binary
Build-Space: 10600
Build-Time: 106
Distribution: hirsute
Host Architecture: riscv64
Install-Time: 195
Job: ../out/ubuntu-advantage-tools_27.0~21.04.2.dsc
Lintian: warn
Machine Architecture: amd64
Package: ubuntu-advantage-tools
Package-Time: 357
Source-Version: 27.0~21.04.2
Space: 10600
Status: successful
Version: 27.0~21.04.2
```

powerpc Xenial build
```bash
Build Architecture: powerpc
Build Type: binary
Build-Space: 4920
Build-Time: 260
Distribution: xenial
Host Architecture: powerpc
Install-Time: 144
Job: ../out/ubuntu-advantage-tools_27.0~16.04.2.dsc
Lintian: fail
Machine Architecture: amd64
Package: ubuntu-advantage-tools
Package-Time: 692
Source-Version: 27.0~16.04.2
Space: 4920
Status: successful
Version: 27.0~16.04.2
--------------------------------------------------------------------------------
Finished at 2021-05-07T03:04:25Z
Build needed 00:11:32, 4920k disk space

```

Jenkinsfile
```diff
@@ -108,7 +108,7 @@ pipeline {
...
```
blackboxsw, line 111: Do we need this --resolve-alternatives anymore now that you have reordered the d/control dep order?

blackboxsw (approved):
LGTM! Builds fine on previously broken "xenial powerpc" and "hirsute riscv64"
Also uploaded to my own PPA building powerpc as well and all series to confirm builds within launchpad. I don't know how to set riscv64 arch builds as that checkbox is disabled on any ppa I've touched.
I'm leaving final review to @paride on this. but +1 from me.

paride:
@blackboxsw I think we still need `--resolve-alternatives`: with the new ` debhelper (>= 13.3) | dh-systemd` order it will build fine in the newer releases but not in the older ones, where the first dep can't be satisfied.

Jenkinsfile
utkarsh2102: AFAICT, you only need `--resolve-alternatives` for building in Groovy and before. Build in Hirsute and onward should pick up debhelper (>= 13.3) and should be good!

However, a small note below..

debian/control
```diff
@@ -4,12 +4,12 @@ Priority: important
 Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
 Build-Depends: bash-completion,
                debhelper (>=9),
-               debianutils,
+               debianutils (>= 4.7),
                dh-python,
-               dh-systemd | debhelper (>= 13.3),
+               debhelper (>= 13.3) | dh-systemd,
```
utkarsh2102, line 9: Whilst you define `debhelper (>= 13.3)` here, be aware that there exists d/compat with value 9 there.
Because of this, if you inspect the build logs, you'll see the following warnings everywhere:
`dh_auto_build: warning: Compatibility levels before 10 are deprecated (level 9 in use)`

This essentially means that even in impish, the compat level 9 is being used and might defeat the whole purpose for having a strict requirement of "debhlper (>= 13.3)".

utkarsh2102 (commented):
Hello, two quick trivial comments from my end below. Thanks!

debian/control
paride: Hi Utkarsh, I think we need to live with that warning message if we want to keep a single d/compat that works across all the supported releases.

Jenkinsfile
paride: So yep we still need it here!

paride (approved):
LGTM, thanks!

debian/control
utkarsh2102: Sure, I just wanted to mention that so that it's known, at least.
OTOH, I was curious if we have a d/compat with 9 mentioned and debhlper (>= 13.3), which of the following is gonna be used, really? 😮

blackboxsw: I think it depends on which target release is being build, since dh-systemd is unavailable on (groovy and later) I think the deb-helper >= 13.3 is used. This compexity in d/control is probably something that future-us needs to fragment a bit so that each d/control file is simple and better representative of the desired depends on each Ubuntu release. We can touch on this later though I think.

blackboxsw: Note that given I can't remember which series drops the dh-systemd deps it adds more credence to our needig making this easier (or better documented) in the future.

Issue #1571, opened 2021-04-26
ua: incompatible services are being disabled by a can_enable check.


blackboxsw:
when checking FIPSEntitlement.can_enable. Livepatch is automatically disabled and re-enabled during token attach. 

The handle_incompatible services is being called during can_enable checks on FIPS (which is incompatible w/ livepatch) but we aren't trying to install FIPS on contract attach, so the client shouldn't pre-emptively disable an active livepatch service.
Expected result: no disable or enable message of livepatch, leave original livepatch token connected after attach???

```bash
ubuntu@test-1570:~$ canonical-livepatch status
last check: 3 minutes ago
kernel: 4.4.0-210.242-generic
server check-in: succeeded
patch state: ✓ no livepatches needed for this kernel yet
tier: updates (Free usage; This machine beta tests new patches.)

ubuntu@test-1570:~$ ua status
SERVICE       AVAILABLE  DESCRIPTION
esm-infra     yes        UA Infra: Extended Security Maintenance (ESM)
fips          yes        NIST-certified FIPS modules
fips-updates  yes        Uncertified security updates to FIPS modules
livepatch     yes        Canonical Livepatch service

This machine is not attached to a UA subscription.
See https://ubuntu.com/advantage

ubuntu@test-1570:~$ sudo ua attach <REDACTED>
Disabling incompatible service: Livepatch
Enabling default service livepatch
Canonical livepatch enabled.
This machine is now attached to 'chad.smith@canonical.com'

SERVICE       ENTITLED  STATUS    DESCRIPTION
esm-infra     yes       enabled   UA Infra: Extended Security Maintenance (ESM)
fips          yes       disabled  NIST-certified FIPS modules
fips-updates  yes       disabled  Uncertified security updates to FIPS modules
livepatch     yes       enabled   Canonical Livepatch service

NOTICES
Operation in progress: ua attach

Enable services with: ua enable <service>

     Account: chad.smith@canonical.com
Subscription: chad.smith@canonical.com
```

PR #1574, opened 2021-04-26
test: uncomment additional xenial upgrade tests

orndorffgrant:
I commented out these lines when testing the last couple fips changes and forgot to take them out before we merged last time :facepalm: 

It's probably a good idea to run the script with these lines uncommented again before merging. I already did and it was all good on my end.


blackboxsw (approved):
+1 confirmed it still exposes https://github.com/canonical/ubuntu-advantage-client/issues/1571 known livepatch disable/enable incorrect messaging. But, that issue #1571 is not critical to SRU.

PR #1577, opened 2021-04-27
fix: dont say reboot required when unnecessary

orndorffgrant:
## Proposed Commit Message

> fix: dont say reboot required when unnecessary
> 
> Before, we were assuming that if the system reboot
> flag was set, then it was because of our fix operation.
> Now, we are carving out a case where we can safely know
> that the reboot is not required because of our fix.
> Specifically, if our fix operation didn't actually install
> anything, then it couldn't have caused the reboot to be
> required.
>
> LP: #1926183

## Test Steps

* Run new unit tests.
* Start a vm and do something that requires reboot. Then run a fix command  that doesn't do anything. Verify it doesn't claim that you need to reboot to complete the fix

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -782,13 +782,15 @@ def _handle_released_package_fixes(
     binary_pocket_pkgs: "Dict[str, List[str]]",
     pkg_index: int,
     num_pkgs: int,
-) -> "Tuple[bool, List[str]]":
+) -> "Tuple[bool, List[str], bool]":
     """Handle the packages that could be fixed and have a released status.
 
     :returns: Tuple of
         boolean whether all packages were successfully upgraded,
-        list of strings containing the packages that were not upgraded
+        list of strings containing the packages that were not upgraded,
+        boolean whether all packages were already installed
     """
+    all_already_installed = True
     upgrade_status = True
     unfixed_pkgs = []
     if src_pocket_pkgs:
@@ -813,6 +815,11 @@ def _handle_released_package_fixes(
                     if not binary_pkgs:
                         print(status.MESSAGE_SECURITY_UPDATE_INSTALLED)
                         continue
+                    else:
+                        # if even one pocket has binary_pkgs to install
+                        # then we can't say that everything was already
+                        # installed.
+                        all_already_installed = False
 
                 pkg_index += len(pkg_src_group)
                 upgrade_status &= upgrade_packages_and_attach(
@@ -822,7 +829,7 @@ def _handle_released_package_fixes(
             if not upgrade_status:
                 unfixed_pkgs += [src_pkg for src_pkg, _ in pkg_src_group]
 
-    return upgrade_status, unfixed_pkgs
+    return upgrade_status, unfixed_pkgs, all_already_installed
 
 
 def _format_unfixed_packages_msg(unfixed_pkgs: "List[str]") -> str:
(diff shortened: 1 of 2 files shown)


uaclient/security.py
```diff
@@ -822,7 +826,7 @@ def _handle_released_package_fixes(
             if not upgrade_status:
                 unfixed_pkgs += [src_pkg for src_pkg, _ in pkg_src_group]
 
-    return upgrade_status, unfixed_pkgs
+    return upgrade_status, unfixed_pkgs, all_already_installed
```
lucasmoura, line 829: I think that's a good approach that you are using. However, we cannot assume that just because we have a binary pkgs to install that they were actually installed.

If you look at `upgrade_packages_and_attach`, if the user is not attached to a ua subscription, we will prompt the user to attach to one to install the package. If the user says no to that, we will not install anything.

uaclient/security.py
```diff
@@ -813,6 +815,8 @@ def _handle_released_package_fixes(
                     if not binary_pkgs:
                         print(status.MESSAGE_SECURITY_UPDATE_INSTALLED)
                         continue
+                    else:
+                        all_already_installed = False
```
lucasmoura, line 819: I don't think we can just set this variable to `False` here directly. We are looking at three distinct upgrade pockets in this function, `standard updates, esm-apps and esm-infra`. Maybe one of these pockets do not have binary packages to install, but the other do

orndorffgrant: Yes I agree! Note though that this boolean is supposed to only be True when all the binary packages that fix the CVE/USN were already installed.
I'm setting this to False here because in this case, there _are_ packages to be installed, which means that we're not in the special case where they were _all_ already installed.
If we reach this case for at least one of the pockets, then we can't say that "packages were already installed" and the new special case added below won't apply.

lucasmoura: That's fair, but I think that if at least one package was installed, it could require a reboot. That's why I think we should block the reboot message only if no packages were installed at all

orndorffgrant:
@lucasmoura I added what we landed on in mattermost - I _think_ this now addresses two problems:
1. Before, if the user had refused to attach, then we might still show the reboot required message.
2. Before, if the fix was already installed, then we might still show the reboot required message.

There are honestly probably more edge cases in here we haven't thought of yet. We may want to come back to this after this release and refactor. But for now I think this addresses the main bugs that people will run into.

Also the various flags were starting to confuse me, so I added a bunch of comments - hopefully they make things more clear, but let me know if you think they're unnecessary.

lucasmoura:
@orndorffgrant I completely agree that we should refactor this method. It is getting too bloated and confusing. We should do this in the next release

lucasmoura (approved):

uaclient/tests/test_security.py
```diff
@@ -1522,6 +1522,7 @@ def test_messages_for_affected_packages_when_fix_fail(
         out, err = capsys.readouterr()
         assert expected in out
 
+    @pytest.mark.parametrize("should_reboot", (False, True))
```
orndorffgrant, line 1525: @lucasmoura I added this so to ensure this test worked whether should_reboot is False or True

blackboxsw (commented):
This looks good @orndorffgrant thanks. +1

I think during merge though, we should use squash commit in order to restate your commit message, but we also need to reference this Launchpad bug as the commit mesage footer (to aid in our debian/changelog creation and launchpad closure of the bug once released)

LP: #1926183

blackboxsw (approved):

blackboxsw:
Known CI intermittent failures:
trusty:      ERROR:paramiko.transport:Exception: Error reading SSH protocol banner
bionic:       paramiko.ssh_exception.SSHException: Error reading SSH protocol banner

Issue #1578, opened 2021-04-27
apt-hook duplicate package names tracked in messaging


blackboxsw:
In testing we can see that the apt-hook adds duplicate package names when tracking the count of packages available in ESM

Not we should not double count libkrad0 nor repeat it in the "updates for following packages"

```bash
*Your UA Infra: ESM subscription has EXPIRED*
Enabling UA Infra: ESM service would provide security updates for following packages:
  libkrad0 libkrad0
2 esm-infra security updates NOT APPLIED. Renew your UA services at
https://ubuntu.com/advantage

Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
applicable law.

The following packages will be upgraded:
  libkrad0
```

PR #1580, opened 2021-04-27
Hook no dupes

blackboxsw:
## Proposed Commit Message
    apt-hook: avoid reporting and counting duplicate package names
    
    Only track unique named package upgrades in ESM.
    Both esm -updates and esm-security pockets can contain same
    packages. Avoid counting duplicates.
    
    Fixes: #1578

- messages: add optional (s) to apt messaging to include singular/plural pkgs

## Test Steps
```bash
UACLIENT_BEHAVE_DESTROY_INSTANCES=0 UACLIENT_BEHAVE_BUILD_PR=1 UACLIENT_BEHAVE_CONTRACT_TOKEN=<YOUR_CONTRACT_TOKEN> tox -e behave-lxd-16.04 features/attach_validtoken.feature:117
```
## Desired commit type::

## Checklist:

### DIFF
--- a/apt-hook/hook.cc
+++ b/apt-hook/hook.cc
@@ -23,6 +23,7 @@
 #include <apt-pkg/policy.h>
 #include <apt-pkg/strutl.h>
 
+#include <algorithm>
 #include <fstream>
 #include <iostream>
 #include <sstream>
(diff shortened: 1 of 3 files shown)


features/attach_validtoken.feature
```diff
@@ -106,8 +106,8 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         \*Your UA Infra: ESM subscription has EXPIRED\*
         Enabling UA Infra: ESM service would provide security updates for following
         packages:
-          .*libkrad.*
-        \d+ esm-infra security updates NOT APPLIED. Renew your UA services at
+          libkrad0
+        1 esm-infra security update NOT APPLIED. Renew your UA services at
```
orndorffgrant, line 110: Based on below change I think this will need to be
```suggestion
        1 esm-infra security update(s) NOT APPLIED. Renew your UA services at
```

apt-hook/hook.cc
```diff
@@ -152,30 +153,34 @@ static void check_esm_upgrade(pkgCache::PkgIterator pkg, pkgPolicy *policy, resu
...
-
-            // Pin-Priority: never unauthenticated APT repos == -32768
-            if (policy->GetPriority(pf.File()) == -32768)
-            {
-               res.disabled_esms_i++;
-            }
-            else
-            {
-               res.enabled_esms_i++;
+            if (std::find(res.esm_i_packages.begin(), res.esm_i_packages.end(), pkg.Name()) == res.esm_i_packages.end()) {
```
orndorffgrant, line 156: This looks correct. It does technically make our counting algorithm O(n^2)
We might be able to fix this problem by just returning early from this function when we find that the package meets one of our criteria, but that feels like a riskier change whereas this is clearly doing what we want.
And our `n` is so small anyway..
I think this is good

features/attach_validtoken.feature
blackboxsw: Good fixed two cases in features where that text would change.

apt-hook/hook.cc
blackboxsw: Very good point. I've added a followup issue for improvement. When we aren't under time pressure. https://github.com/canonical/ubuntu-advantage-client/issues/1581

orndorffgrant (approved):
Just manually verified this fixes the issue. LGTM!

PR #1583, opened 2021-04-27
Release 27 changelog

lucasmoura:


debian/changelog
```diff
@@ -1,3 +1,15 @@
+ubuntu-advantage-tools (27.0~21.10.1) impish; urgency=medium
```
blackboxsw, line 1: With the current devel release we don't ~XX.YY.1) so it can be just the following:
```suggestion
ubuntu-advantage-tools (27.0) impish; urgency=medium
```

blackboxsw (approved):
minor change to version number for impish since it's a devel release. then approve

blackboxsw (approved):
LGTM!

PR #1589, opened 2021-04-28
apt-hook: mitigate failures with true

orndorffgrant:
## Proposed Commit Message

> apt-hook: mitigate failures with true
>
> In case something catastrophic happens
> in our apt hooks, we now make sure the subprocess
> always returns successfully by calling `true`
> if our hook fails.

## Desired commit type::

## Checklist:


blackboxsw (approved):
Confirmed success and failure path, unexpected failures will result in messages at apt cmdline, but not result in failing the apt command.

Unpacking krb5-locales (1.17-10ubuntu0.1) over (1.17-10) ...
Setting up krb5-locales (1.17-10ubuntu0.1) ...
  File "/tmp/1.py", line 1, in <module>
    1/0
ZeroDivisionError: division by zero
root@test-g:~# echo $?
0

Issue #1592, opened 2021-04-29
incomplete message - This machine is now attached to ''


cpaelzer:
Hi,
maybe my token/ID is bad, but on all machines that I activated my token I've got

```
sudo ua attach you-wont-get-my-token-unless-you-need-for-debug
This machine is now attached to ''

SERVICE       ENTITLED  STATUS    DESCRIPTION
esm-infra     yes       n/a       UA Infra: Extended Security Maintenance (ESM)
fips          yes       disabled  NIST-certified FIPS modules
fips-updates  yes       disabled  Uncertified security updates to FIPS modules
livepatch     yes       n/a       Canonical Livepatch service

NOTICES
Operation in progress: ua attach

Enable services with: ua enable <service>

     Account: 
Subscription: 
```

Wondering about:
 - attached to ''
 - Account: 
 - Subscription: 

Some data seems to be missing here, it isn't that the attach didn't work. IF I try to attach again I get:

```
sudo ua attach you-wont-get-my-token-unless-you-need-for-debug
This machine is already attached to ''
To use a different subscription first run: sudo ua detach.
```

cpaelzer:
FYI from looking at other issues, this seems to be a more severe form of issue #1547

cpaelzer:
Further FYI - I've only tested on various s390x and ppc64el machines so far.
But a recheck on x86 xenial container showed that it isn't arch-specific (more likely to be my token/account)

cpaelzer:
FYI x86 output (since it enables a few things by default)

```
sudo ua attach you-wont-get-my-token-unless-you-need-for-debug
Enabling default service esm-infra
Updating package lists
UA Infra: ESM enabled
A reboot is required to complete install.
Enabling default service livepatch
Installing canonical-livepatch snap
Canonical livepatch enabled.
This machine is now attached to ''

SERVICE       ENTITLED  STATUS    DESCRIPTION
esm-infra     yes       enabled   UA Infra: Extended Security Maintenance (ESM)
fips          yes       disabled  NIST-certified FIPS modules
fips-updates  yes       disabled  Uncertified security updates to FIPS modules
livepatch     yes       enabled   Canonical Livepatch service

NOTICES
Operation in progress: ua attach

Enable services with: ua enable <service>

     Account: 
Subscription
```

mitechie:
[...]

PR #1597, opened 2021-04-29
status: do not show info if not on contract

lucasmoura:
## Proposed Commit Message
status: do not show info if not on contract

When running status on an attached machine, we present the account and subscription info in the the end. However, if the contract does not have that info, we will present them as empty. We are now updating the status to not show that info at the end if it is not present in the contract.

Fixes: #1592

## Test Steps
Run the modified unit test

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -677,14 +677,19 @@ def _attach_with_token(
         cfg.status()  # Persist updated status in the event of partial attach
         config.update_ua_messages(cfg)
         return 1
+
     contract_name = cfg.machine_token["machineTokenInfo"]["contractInfo"][
         "name"
     ]
-    print(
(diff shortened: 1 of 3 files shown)


blackboxsw (changes requested):
Redacting looks good, but shouldn't this branch also touch on 
addressing the fact that contract_name is empty and we still emit
[MESSAGE_ATTACH_SUCCESS_TMPL](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/cli.py#L683)

cpaelzer:
From just looking at the changes that LGTM non authoritative +0.5

orndorffgrant (approved):
This looks good and worked for me!

blackboxsw (approved):
LGTM!

Issue #1590, opened 2021-04-28
trusty upgrade to xenial will show esm-infra as disabled


lucasmoura:
When we are attached to a trusty machine and we perform a `do-release-upgrade` operation we expect to be upgraded to xenial while keeping `esm-infra`, or any other service we have enabled, still enabled. We guarantee that by running the [upgrade_lts_contract](https://github.com/canonical/ubuntu-advantage-client/blob/main/lib/upgrade_lts_contract.py) in the end of `do-release-upgrade`. This is script is run by adding the following configuration at `/etc/update-manager/release-upgrades.d/`

```
[Sources]
Pockets=security,updates,proposed,backports,infra-security,infra-updates,apps-security,apps-updates
[Distro]
PostInstallScripts=./xorg_fix_proprietary.py, /usr/lib/ubuntu-advantage/upgrade_lts_contract.py
```

In our current testing infrastructure, we are updating the tusty packaging to the development version of uaclient and then performing the upgrade. This will guarantee that will we have that file in place before running `do-release-upgrade`. Now, since we will not release the new uaclient version into trusty, we will never have that configuration in place. This means that `do-release-upgrade`  will not run the `upgrade_lts_script` in the end, which in turn, will make `esm-infra` to be disabled, as the apt sources will not be correctly updated.

lucasmoura:
We two workarounds for this:

1) Manually add the `etc/update-manager/release-upgrades.d/` config:
```
[Sources]
Pockets=security,updates,proposed,backports,infra-security,infra-updates,apps-security,apps-updates
[Distro]
PostInstallScripts=./xorg_fix_proprietary.py, /usr/lib/ubuntu-advantage/upgrade_lts_contract.py
```

Before running `do-release-upgrade`

2) After running `do-release-upgrade`  and restarting the machine, run `ua enable esm-infra`

mitechie:
Documented as an FAQ item in https://discourse.ubuntu.com/t/ubuntu-advantage-client/21788/4

orndorffgrant:
Closing as a Won't Fix because it is documented.

PR #1598, opened 2021-04-29
Trusty upgrade script

lucasmoura:
## Proposed Commit Message
tests: manual test for trusty upgrade to xenial

Add manual upgrade test to address trusty to xenial upgrade. This test will cover #1590 

## Test Steps
Run the manual test added here

## Desired commit type::

## Checklist:


sru/release-27/test_trusty_upgrade_to_xenial.sh
```diff
@@ -0,0 +1,43 @@
...
+# not enable do-release-upgrade to run a post script that handles the services
+# migration when performing an upgrade.
+# This script is adding a solution for that problem by manually adding the necessary
+# configs for do-release-upgrade (This configurations are done for release 26 forward)
+set -x
+
+series=trusty
+name=$series-upgrade-lxd
+
+function fix_upgrade_for_esm_infra() {
```
blackboxsw, line 15: Let's document this issue URL here that we are working around

sru/release-27/test_trusty_upgrade_to_xenial.sh
```diff
@@ -0,0 +1,43 @@
...
+multipass delete $name
+multipass purge
+multipass launch $series --name $name
+
+multipass exec $name -- sudo apt-get upgrade -y
+multipass exec $name -- sudo apt-get dist-upgrade -y
+multipass exec $name -- ua version
+multipass exec $name -- sudo ua attach $UACLIENT_BEHAVE_CONTRACT_TOKEN
+multipass exec $name -- sudo ua status
+multipass exec $name -- sudo add-apt-repository ppa:orndorffgrant/uaclient-staging-27 -y
```
blackboxsw, line 30: I think we should change this to actually setup -proposed on the system under test

Something like what we do for cloud-init (but installing ubuntu-advantage-tools instead) https://github.com/cloud-init/ubuntu-sru/blob/main/sru-scripts/sru-vars.template#L59-L76

blackboxsw (changes requested):
thanks lucas, let's instrument something that tests the -proposed pocket if we can

blackboxsw (approved):
This looks good  thanks for the update.

sru/release-27/test_xenial_upgrade.sh
```diff
@@ -65,7 +65,8 @@ function teardown_vm() {
 function install_new_ua() {
     tool=$1
     h2 "Set up to use daily ppa"
-    $tool exec $name -- sudo add-apt-repository ppa:ua-client/daily -y
+    $tool exec $name -- sudo sh -c "cat <<EOF >/etc/apt/sources.list.d/ubuntu-$series-proposed.list
```
orndorffgrant, line 68: @lucasmoura Really sorry I should've mentioned - this script doesn't currently have `$series` set. Can you set `series=xenial` at the top of the script?

orndorffgrant (approved):

Issue #1594, opened 2021-04-29
Do not tab-complete or suggest unknown/unavailable services


cpaelzer:
Right now `cc-eal` and `cis-audit` seem to be present in the code but not available.

Tab completion shows:
```
ubuntu@x-ua:~$ ua enable 
cc-eal        cis-audit     esm-infra     fips          fips-updates  livepatch
``` 

Only to then bail out on `cc-eal` and `cis-audit`, but not with "not yet available on xenial" or "coming soon" or such.
But with "unknown service"

```
$ sudo ua enable cc-eal
One moment, checking your subscription first
Cannot enable unknown service 'cc-eal'.
Try esm-infra, fips, fips-updates, livepatch.
```

Also first checking my subscription and then failing to try enabling it does read like my subscription would have it.

The above outlines the tab completion, the same issue is present at the suggestions in the error messages.

If you look at this (also in the above output)

```
...
Try esm-infra, fips, fips-updates, livepatch.
```

But then 
```
ubuntu@xenial-ua-ppc:~$ sudo ua enable esm-infra
One moment, checking your subscription first
UA Infra: ESM is not available for platform ppc64el.
Supported platforms are: amd64, i386, i686, x86_64.
ubuntu@xenial-ua-ppc:~$ sudo ua enable livepatch
One moment, checking your subscription first
Livepatch is not available for platform ppc64el.
Supported platforms are: amd64, x86_64.
```

Therfore please:
- reduce tab-completion to implemented features
- reduce the "Try" suggestion to features supported on the current platform

PR #1607, opened 2021-05-03
Do not show beta services in autocomplete

lucasmoura:
## Proposed Commit Message
Do not show beta services in autocomplete

When running ua enable, we don't want to show beta services by default in when users hit tab and the command autocompletes. We are now updating the logic to only show non-beta services in that situation. We will only show beta services in autocomplete if the user has a uaclient.conf configuration that allow beta services.

Fixes: #1594

## Test Steps
1. Launch a vm with this version of uaclient installed
2. Try to autocomplete when running `ua enable`. Verify that only non-beta services appear
3. Add the following lines to `/etc/ubuntu-advantage/uaclient.conf`:

```yaml
features:
  allow_beta: true
```

4. Refresh bash with `exec bash`
5. Try to autocomplete when running `ua enable`. Verify that only all services appear now

## Desired commit type::

## Checklist:


blackboxsw (changes requested):
I think we also should use valid_services function in security._is_pocket_used_by_beta_service and uaclient/entitlements/esm.py:repo_pin_priority and disable_apt_auth_only
What do you think about those 3 cases?


@lucasmoura and also https://github.com/canonical/ubuntu-advantage-client/blob/main/debian/ubuntu-advantage-tools.postinst#L111-L128 too I think can be rewritten to leverage this function right?

lucasmoura:
@blackboxsw I am not sure if we should modify those functions. The reason is that they are already acting using an instantiated entitlement. If we change those functions to use the `valid_services` we would loop on all of the entitlements to produce the list and then check if the ent.name is not on that list.

I know that by doing that we would avoid reading the `uaclient.conf` content in those methods, but we will also create the list of service unnecessarily here.

Although I am kind of doing that on `_perform_enable`. Maybe I can modify that function to receive a list of valid services or even move that beta check outside of it.

What do you think about that ?

blackboxsw:
> @blackboxsw I am not sure if we should modify those functions. The reason is that they are already acting using an instantiated entitlement. If we change those functions to use the `valid_services` we would loop on all of the entitlements to produce the list and then check if the ent.name is not on that list.
> 
> I know that by doing that we would avoid reading the `uaclient.conf` content in those methods, but we will also create the list of service unnecessarily here.
> 
> Although I am kind of doing that on `_perform_enable`. Maybe I can modify that function to receive a list of valid services or even move that beta check outside of it.
> 
> What do you think about that ?

I think allowing for a service_names param would to valid_services() would make sense. So we'd limit our setup/response to just the list of service_names provided

lucasmoura:
@blackboxsw code updated

orndorffgrant (approved):
One small comment but otherwise looks good to me!

uaclient/entitlements/esm.py
```diff
@@ -38,27 +38,27 @@ class ESMAppsEntitlement(ESMBaseEntitlement):
     @property
     def repo_pin_priority(self) -> "Optional[str]":
         """All LTS with the exception of Trusty should pin esm-apps."""
+        from uaclient.entitlements import valid_services
+
```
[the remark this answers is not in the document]
lucasmoura: Yes, unfortunately we need to have that import here to avoid a cyclic dependency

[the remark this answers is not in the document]
orndorffgrant: Ahhh got it :+1:

lucasmoura:
@blackboxsw  I will refactor my implementation of the `valid_services` function in this PR The reason for that is a gap that I found on that solution, which is also on main. If an user run `ua enable esm-apps --beta`. In that situation we will fail to set the right pin priority as can be seen here:

https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/entitlements/esm.py#L47

This happens because we don't propagate the `--beta`  into the service.

Also, I am still not sure we should use the `valid_services` function as is. It still seems odd to me that an instance will call a function that will create another instance of the same type to check if that instance is beta or not. The instance should have that knowledge about itself.

I am thinking of implementing an `is_valid_service`method into each entitlement like this:
```
def is_valid_service(self):
   # allow_beta will not be a flag that says if we should override
   # the beta behavior of the service. It can be a combination of
   # beta flag and allow_beta config
   return if not self.beta or self.allow_beta
```

With that implemented, we can turn `valid_services` into:
```
def valid_services(self, allow_beta):
    cfg = UAConfig()
    allow_beta_cfg = is_config_value_true(cfg.cfg, "features.allow_beta")
    allow_beta |= allow_beta_cfg
    
    if allow_beta:
        return list(ENTITLEMENT_CLASS_BY_NAME.keys())

    return sorted(
        [   
            ent_name
            for ent_name, ent_cls in ENTITLEMENT_CLASS_BY_NAME.items()
            if ent_cls().is_valid_service()
        ]
    ) 
 ```
 
 if we do something like this, we will only @orndorffgrant suggestion on PR #1613 to move the beta logic into `can_enable`, which I think is correct.
 
 @orndorffgrant @blackboxsw what do you think ?

**Comment by orndorffgrant** [comment]
@lucasmoura I think it makes a lot of sense for an entitlement to know about itself whether it is "valid" or not (including the config `allow_beta`). It would also eliminate the current somewhat awkward `if name in valid_services(name):` lines. That refactor sounds good to me +1.

> Digression:
> We could even add something to the entitlements constructor that does all of the beta logic. Something like (not functional at all, but this is the gist):
> ```python
> def __init__(self, cmd_beta_flag = False):
>     self.is_valid = not self.is_beta or cmd_beta_flag or self.cfg.features.allow_beta
> ```
> That would require a lot more refactoring though and not worth it right now. Also not sure if that could cause other problems.

**Comment by blackboxsw** [comment]
> @blackboxsw I will refactor my implementation of the `valid_services` function in this PR The reason for that is a gap that I found on that solution, which is also on main. If an user run `ua enable esm-apps --beta`. In that situation we will fail to set the right pin priority as can be seen here:
> 
> https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/entitlements/esm.py#L47
> 
> This happens because we don't propagate the `--beta` into the service.
> 
> Also, I am still not sure we should use the `valid_services` function as is. It still seems odd to me that an instance will call a function that will use that instance class to see if the instance is beta or not. The instance should have that knowledge about itself.
> 
> I am thinking of implementing an `is_valid_service`method into each entitlement like this:
> 
> ```python
> def is_valid_service(self):
>    # allow_beta will not be a flag that says if we should override
>    # the beta behavior of the service. It can be a combination of
>    # beta flag and allow_beta config
>    return if not self.beta or self.allow_beta
> ```y

Why not change is_beta into a @property decorated method on the class that takes into account self.allow_beta and self._is_beta? I wonder about making is_beta a @property method too because I'd also like us long-term to be able to settings_overrides is_beta on a given class too or to drive 'beta' status from something that the machine-token metaedata returns for the entitlement.

```

> if we do something like this, we will make @orndorffgrant suggestion on PR #1613 to move the beta logic into `can_enable` easier to implement, which I think is correct.
> 
> @orndorffgrant @blackboxsw what do you think ?

Generally I think consilidating the beta checks into `can_enable` logic makes sense.

lucasmoura:
@blackboxsw @orndorffgrant code updated, please let me know if there is something odd in the code and sorry for the huge change :(

@blackboxsw I have not turned `is_beta`into a property, because we use that base construct on other unrelated parts of the code, like this one on config:
https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/config.py#L411

In situation like that, I don't think we need to instantiate the entitlement to define if it is beta or not, since we are actually driving this outside of the scope of this function already:
https://github.com/canonical/ubuntu-advantage-client/blob/a4c4fab0533520e3905def7957b988764b5fbd06/uaclient/config.py#L592

However, I can update that if needed

uaclient/entitlements/__init__.py
```diff
@@ -29,11 +32,22 @@ def cast(_, x):  # type: ignore
...
+    if allow_beta:
+        return sorted(ENTITLEMENT_CLASS_BY_NAME.keys())
+
+    return sorted(
+        [
+            ent_name
+            for ent_name, ent_cls in ENTITLEMENT_CLASS_BY_NAME.items()
+            if not ent_cls.is_beta
+        ]
+    )
```
orndorffgrant, line 53: I think now with the `valid_service` method on entitlements, this can be simplified to

```suggestion
    cfg = UAConfig()
    return sorted(
        [
            ent_name
            for ent_name, ent_cls in ENTITLEMENT_CLASS_BY_NAME.items()
            if ent_cls(cfg=cfg, allow_beta=allow_beta).valid_service
        ]
    )
```

uaclient/entitlements/base.py
```diff
@@ -138,12 +157,22 @@ def can_disable(self, silent: bool = False) -> bool:
...
 
+        :param enable_by_default: If True, do not stop enable if service
+                                  is marked as beta
         :param silent: if True, suppress output
         """
+
+        # If the contract needs to enable a beta service by default,
+        # we will not stop it here
+        if not self.valid_service and not enable_by_default:
+            return False
```
orndorffgrant, line 174: It looks like the `enable_by_default` case is already handled on line 481 of this file, so I don't think we need it here.

orndorffgrant: And so I didn't realize you were going to move the beta check into can_enable here - thank you!

This is most of fixing #1588, the rest is moving this check down to be the last check in `can_enable` (so that the "already enabled" message gets printed if necessary) and making this beta check print the same message that it was before (`MESSAGE_INVALID_SERVICE_OP_FAILURE_TMPL`) when not silent - I can handle those pieces in my PR or you can add them here - either works for me!

uaclient/entitlements/__init__.py
lucasmoura: Yes, this a valid refactoring. My rationale to separate the two cases here is just to avoid creating the service objects if not needed. But maybe this optimization is not worth it at this point

uaclient/tests/test_cli_enable.py
```diff
@@ -375,120 +374,46 @@ def test_invalid_service_names(
...
-
-        assert 1 == m_cfg.status.call_count
-        assert 1 == m_user_cfg.call_count
-        assert beta_call_count == m_is_beta.call_count
-
-    @pytest.mark.parametrize("silent_if_inapplicable", (True, False, None))
-    @mock.patch("uaclient.cli.entitlements")
-    def test_beta_entitlement_not_enabled(
-        self, m_entitlements, silent_if_inapplicable
-    ):
```
orndorffgrant, line 440: I think a full fix of #1588 will still need these tests right? I can add them back in my PR if you want though

orndorffgrant (commented):
Thanks for the refactor! I think it makes a lot of things simpler. I did leave a couple comments. And we need to iron out if we should finish fixing #1588 in this PR or if we/I should finish it in a separate PR

uaclient/tests/test_cli_enable.py
lucasmoura: Since we are no longer handling the beta services on cli directly, those tests were behaving exactly as the tests we have for unknown services, that's why I removed. But if you think you will need them on #1588, no problem at all adding them back there.

uaclient/entitlements/base.py
orndorffgrant: Ahhh yes you're right!

uaclient/tests/test_cli_enable.py
orndorffgrant: Very good point! Yes they can stay removed here then

uaclient/entitlements/__init__.py
orndorffgrant: That makes sense - I think in this case my preference would be to favor readability over the small optimization from avoiding object instantiation - but that's not a strong preference here, feel free to ignore :)

blackboxsw (approved):
LGTM! works like a charm. nice cleanups.

Issue #1582, opened 2021-04-27
validate and test CIS on Focal


mitechie:
Now that Focal has supported CIS tooling we need to validate it and add a test to our suite to make sure it's working properly with the UA client.

Issue #1608, opened 2021-05-03
Error activating CIS tool while launching AWS Pro instance with IMDSv2/ IMDSv1


ariannalozano:
Ran into a problem on this command while trying to gain access to the CIS beta tool for Ubuntu Pro:

sudo ua enable cis --beta

I was running this on aws us-west-2 from an ubuntu 18 pro instance and I get this output: 

To use 'cis' you need an Ubuntu Advantage subscription
Personal and community subscriptions are available at no charge
See https://ubuntu.com/advantage

I was on 26.3~18.04.1, with those commands I got upgraded to 27.0~18.04.1

If i run ua status --all I get 

SERVICE       AVAILABLE  DESCRIPTION
cc-eal        no         Common Criteria EAL2 Provisioning Packages
cis           yes        Center for Internet Security Audit Tools
esm-apps      yes        UA Apps: Extended Security Maintenance (ESM)
esm-infra     yes        UA Infra: Extended Security Maintenance (ESM)
fips          yes        NIST-certified FIPS modules
fips-updates  yes        Uncertified security updates to FIPS modules
livepatch     yes        Canonical Livepatch service

This machine is not attached to a UA subscription.
See https://ubuntu.com/advantage

But i can still not active cis: 

 sudo ua enable cis --beta
To use 'cis' you need an Ubuntu Advantage subscription
Personal and community subscriptions are available at no charge
See https://ubuntu.com/advantage

I can also confirm that I was trying this on an Ubuntu Pro image.

Tried running "sudo ua auto-attach", here is the output: 
It gives an error (see attached), here are the logs: 

2021-05-03 10:15:02,308 - cli.py:(1060) [DEBUG]: Executed with sys.argv: ['/usr/bin/ua', 'auto-attach']
2021-05-03 10:15:02,308 - util.py:(414) [DEBUG]: Reading file: /var/lib/ubuntu-advantage/notices.json
2021-05-03 10:15:02,309 - config.py:(357) [DEBUG]: File does not exist: /var/lib/ubuntu-advantage/notices.json
2021-05-03 10:15:02,309 - util.py:(646) [DEBUG]: Writing file: /var/lib/ubuntu-advantage/notices.json
2021-05-03 10:15:02,309 - util.py:(646) [DEBUG]: Writing file: /var/lib/ubuntu-advantage/private/lock
[...]

PR #1610, opened 2021-05-04
tests: test enabling CIS on focal

orndorffgrant:
## Discussion

It doesn't appear that any changes to the uaclient code itself will be necessary for CIS on focal. However, the command doesn't yet succeed due to the packages not being available in the archive yet. We should re-assess when they are available.

***Marked draft because this test doesn't yet pass, but will as soon as the focal cis packages are available. Do not merge.***

## Proposed Commit Message

> tests: test enabling CIS on focal
> 
> Now that CIS is available on focal, we should
> start running our CIS integration test on focal
> to test for regressions.
> 
> Fixes: #1582

## Test Steps

Run the new integration test

## Desired commit type::

## Checklist:

### DIFF
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -340,6 +340,7 @@ Feature: Enable command behaviour when attached to an UA staging subscription
 
     @series.xenial
     @series.bionic
+    @series.focal
     Scenario Outline: Attached enable of cis service in a ubuntu machine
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
Staging looks like it is available now for FIPS . This test now passes +1 We should also sort the following issue  separately https://github.com/canonical/ubuntu-advantage-client/issues/1608

PR #1612, opened 2021-05-05
cis: post enable message pointing to readme

orndorffgrant:
## Discussion

The exact content here still needs discussion and approval before merging.

## Proposed Commit Message

> cis: post enable message pointing to readme
> 
> After enabling cis using ua, the user needs to
> take additional steps to use the cis packages
> that were installed. This adds messaging to point
> users in the right direction to get started.

## Test Steps

Enable CIS. Follow the messages to look at the readmes and verify that the messages pointed you to the correct locations.

## Desired commit type::

## Checklist:


features/staging_commands.feature
```diff
@@ -350,6 +350,8 @@ Feature: Enable command behaviour when attached to an UA staging subscription
             Updating package lists
             Installing CIS Audit packages
             CIS Audit enabled
+            Refer to /usr/share/doc/usg-common/README.audit.gz on how to run an audit scan of the system
+            Refer to /usr/share/doc/usg-cisbenchmark/README.hardening.gz on how to harden the system to the desired CIS compliance level
```
blackboxsw, line 354: @mitechie @lechsandecki  @vineethapai does this working upon CIS enable look good to folks? Without any message, someone who enables CIS will have limited visibility to what to do next.

lechsandecki:
Good catch!

should they first `sudo apt install usg-cisbenchmark` or would it already be done by the ua client? do you know @blackboxsw 

Some time ago we have recorded a video with @vineethapai that summarises CIS audit options https://www.youtube.com/watch?v=wyEX0eyoK88

perhaps we could point from the ua client to that video URL as a way to introduce CIS hardening?

blackboxsw:
> Good catch!
> 
> should they first `sudo apt install usg-cisbenchmark` or would it already be done by the ua client? do you know @blackboxsw
> 
> Some time ago we have recorded a video with @vineethapai that summarises CIS audit options https://www.youtube.com/watch?v=wyEX0eyoK88
> 
> perhaps we could point from the ua client to that video URL as a way to introduce CIS hardening?

@lechsandecki ua client already installs both 'usg-cisbenchmark usg-common'. It feels a bit strange to point to a video that doesn't document use of `ua` which already does some of that install and setup for you.  But, the video is a good reference, maybe we could make sure it is captured in /usr/share/doc/usg-cisbenchmark/README.hardening.gz as a reference too?

mitechie:
I think we should stick to a URL that the product folks can adjust for content over time. Let's get the PMs to confirm the URL and output that for more information and the video/docs/etc can all be linked to on that content page. 

Once that is done we can also remove the --beta flag from CIS as a feature in the UA client.

lechsandecki:
@nmav

nmav:
1. Yes, it will be at https://ubuntu.com/security/cis
2. Is that the same as https://www.brighttalk.com/webcast/6793/432536 ?
4. Hopefully by the end of this month but it depends mostly on the web team

features/staging_commands.feature
```diff
@@ -350,6 +350,7 @@ Feature: Enable command behaviour when attached to an UA staging subscription
             Updating package lists
             Installing CIS Audit packages
             CIS Audit enabled
+            Visit https://ubuntu.com/security/cis to learn how to use the CIS packages
```
orndorffgrant, line 353: @mitechie @blackboxsw I went ahead and changed the message to point to the URL even though it isn't live yet. Let me know what you think of the wording.

blackboxsw (approved):
Wording good and we decided in standup having this in while cis is_beta = True is good to ship.

blackboxsw (approved):

features/staging_commands.feature
nmav: Now that I see the message, this is not the URI to learn how to use the CIS packages. This is the URI to learn about CIS and Ubuntu. If your intention is to point to technical instructions, then a more appropriate URI is:
https://security-certs.docs.ubuntu.com/en/cis

PR #1613, opened 2021-05-05
enable: correct messaging for beta service

orndorffgrant:
## Proposed Commit Message

> enable: correct messaging for beta service
>
> Previously, if the user tried to enable a beta service
> that was already enabled without using --beta, they
> would get incorrect messaging. Now they are correctly
> told that the service is already enabled.
> 
> Fixes: #1588

## Test Steps

Enable a beta service such as esm-apps
Then run `ua enable esm-apps` without the `--beta` flag.
Verify that the output correctly states that esm-apps is already enabled.

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -515,37 +515,6 @@ def action_disable(args, cfg, **kwargs):
     return 0 if ret else 1
 
 
-def _perform_enable(
-    entitlement_name: str,
-    cfg: config.UAConfig,
-    *,
-    assume_yes: bool = False,
-    silent_if_inapplicable: bool = False,
-    allow_beta: bool = False
-) -> bool:
-    """Perform the enable action on a named entitlement.
-
-    (This helper excludes any messaging, so that different enablement code
-    paths can message themselves.)
-
-    :param entitlement_name: the name of the entitlement to enable
-    :param cfg: the UAConfig to pass to the entitlement
-    :param assume_yes:
-        Assume a yes response for any prompts during service enable
-    :param silent_if_inapplicable:
-        don't output messages when determining if an entitlement can be
-        enabled on this system
-    :param allow_beta: Allow enabling beta services
-
-    @return: True on success, False otherwise
-    """
-    ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[entitlement_name]
-    entitlement = ent_cls(cfg, assume_yes=assume_yes, allow_beta=allow_beta)
-    ret = entitlement.enable(silent_if_inapplicable=silent_if_inapplicable)
-    cfg.status()  # Update the status cache
-    return ret
-
-
 @assert_root
 @assert_attached(ua_status.MESSAGE_ENABLE_FAILURE_UNATTACHED_TMPL)
 @assert_lock_file("ua enable")
@@ -568,17 +537,24 @@ def action_enable(args, cfg, **kwargs):
     valid_services_names = entitlements.valid_services(allow_beta=args.beta)
     ret = True
 
-    for entitlement in entitlements_found:
+    for ent_name in entitlements_found:
         try:
-            ent_ret = _perform_enable(
-                entitlement,
-                cfg,
-                assume_yes=args.assume_yes,
-                allow_beta=args.beta,
+            ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[ent_name]
+            entitlement = ent_cls(
+                cfg, assume_yes=args.assume_yes, allow_beta=args.beta
             )
+            ent_ret = entitlement.enable()
+            cfg.status()  # Update the status cache
 
-            if not ent_ret and entitlement not in valid_services_names:
-                entitlements_not_found.append(entitlement)
+            if not ent_ret:
+                can_enable, reason = entitlement.can_enable(silent=True)
+                if (
+                    not can_enable
+                    and reason == ua_status.CanEnableFailureReason.IS_BETA
+                ):
+                    # if we failed because ent is in beta and there was no
+                    # allow_beta flag/config, pretend it doesn't exist
+                    entitlements_not_found.append(ent_name)
 
             ret &= ent_ret
         except exceptions.UserFacingError as e:
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -23,6 +23,7 @@ from uaclient.status import (
     ApplicabilityStatus,
     ContractStatus,
     UserFacingStatus,
+    CanEnableFailureReason,
     MESSAGE_INCOMPATIBLE_SERVICE_STOPS_ENABLE,
 )
 from uaclient.defaults import DEFAULT_HELP_FILE
@@ -157,14 +158,18 @@ class UAEntitlement(metaclass=abc.ABCMeta):
             return False
         return True
 
-    def can_enable(self, silent: bool = False) -> bool:
+    def can_enable(
+        self, silent: bool = False
+    ) -> "Tuple[bool, Optional[CanEnableFailureReason]]":
         """
         Report whether or not enabling is possible for the entitlement.
 
         :param silent: if True, suppress output
+        :return:
+            tuple of (bool, CanEnableFailureReason).
+            (True, None) if can enable
+            (False, Reason) if can't enable
         """
-        if not self.valid_service:
-            return False
 
         if self.is_access_expired():
             logging.debug(
@@ -174,7 +179,7 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         if not self.contract_status() == ContractStatus.ENTITLED:
             if not silent:
                 print(status.MESSAGE_UNENTITLED_TMPL.format(title=self.title))
-            return False
+            return (False, CanEnableFailureReason.NOT_ENTITLED)
 
         application_status, _ = self.application_status()
         if application_status != status.ApplicationStatus.DISABLED:
@@ -184,17 +189,23 @@ class UAEntitlement(metaclass=abc.ABCMeta):
                         title=self.title
                     )
                 )
-            return False
+            return (False, CanEnableFailureReason.ALREADY_ENABLED)
+
+        if not self.valid_service:
+            return (False, CanEnableFailureReason.IS_BETA)
+
         applicability_status, details = self.applicability_status()
         if applicability_status == status.ApplicabilityStatus.INAPPLICABLE:
             if not silent:
                 print(details)
-            return False
+            return (False, CanEnableFailureReason.INAPPLICABLE)
 
         if self.incompatible_services:
-            return self.handle_incompatible_services()
+            handle_incompat_ret = self.handle_incompatible_services()
+            if not handle_incompat_ret:
+                return (False, CanEnableFailureReason.INCOMPATIBLE_SERVICE)
 
-        return True
+        return (True, None)
 
     def handle_incompatible_services(self) -> bool:
         """
@@ -473,7 +484,7 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         if enable_by_default:
             self.allow_beta = True
 
-        can_enable = self.can_enable(silent=True)
+        can_enable, _ = self.can_enable(silent=True)
         if can_enable and enable_by_default:
             if allow_enable:
                 msg = status.MESSAGE_ENABLE_BY_DEFAULT_TMPL.format(
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -8,7 +8,15 @@ import yaml
 from uaclient.util import is_config_value_true
 
 try:
-    from typing import Any, Callable, Dict, List, Optional, Tuple  # noqa: F401
+    from typing import (  # noqa: F401
+        Any,
+        Callable,
+        Dict,
+        List,
+        Optional,
+        Tuple,
+        Union,
+    )
 
     StaticAffordance = Tuple[str, Callable[[], Any], bool]
 except ImportError:
@@ -100,6 +108,14 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         """
         return self._incompatible_services
 
+    # Any custom messages to emit to the console or callables which are
+    # handled at pre_enable, pre_disable, pre_install or post_enable stages
+    @property
+    def messaging(
+        self
+    ) -> "Dict[str, List[Union[str, Tuple[Callable, Dict]]]]":
+        return {}
+
     def __init__(
         self,
         cfg: "Optional[config.UAConfig]" = None,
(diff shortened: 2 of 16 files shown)


uaclient/cli.py
```diff
@@ -536,11 +536,19 @@ def _perform_enable(
...
+
     config_allow_beta = util.is_config_value_true(
         config=cfg.cfg, path_to_value="features.allow_beta"
     )
     allow_beta |= config_allow_beta
-    if not allow_beta and ent_cls.is_beta:
+    if not already_enabled and not allow_beta and ent_cls.is_beta:
         tmpl = ua_status.MESSAGE_INVALID_SERVICE_OP_FAILURE_TMPL
         raise exceptions.BetaServiceError(
             tmpl.format(
```
orndorffgrant, line 554: The solution I have currently here adding an additional check to see if something is already enabled before doing the beta check.

I think it might make more sense to instead move the beta check that is currently here, into `can_enable`. But I wanted to run this by you @lucasmoura and @blackboxsw in case there is a reason this beta check can't go into `can_enable`. Thoughts?

lucasmoura:
@orndorffgrant I think this looks really good. We just need to sort out two integration test failures on bionic that might be related to those changes:

```
features/attached_enable.feature:32  Attached enable Common Criteria service in a ubuntu machine -- @1.1 ubuntu release
features/attached_enable.feature:89  Attached enable of a service in a ubuntu machine -- @1.1 ubuntu release
```

orndorffgrant:
@lucasmoura @blackboxsw I believe the remaining integration test failures are unrelated to this PR and are at starting to be addressed in #1629 So this PR is ready for review

uaclient/cli.py
```diff
@@ -568,17 +537,24 @@ def action_enable(args, cfg, **kwargs):
     valid_services_names = entitlements.valid_services(allow_beta=args.beta)
     ret = True
 
-    for entitlement in entitlements_found:
+    for ent_name in entitlements_found:
```
lucasmoura, line 540: I think adding an unit test for the case were we try to enable a beta service already enabled would be good. I know it will be a little redundant considering we do have an integration test for it, but since the price we pay to run an unit test is really cheap, I don't see a problem having them both, since the unit test will allow us to detect problems earlier.

Let me know if you agree with that

uaclient/tests/test_cli_enable.py
```diff
@@ -327,12 +342,71 @@ def test_entitlements_not_found_and_beta(
...
         for m_ent in mock_obj_list:
             assert [expected_enable_call] == m_ent.enable.call_args_list
 
         assert 0 == m_ent1_obj.call_count
 
+    @mock.patch("uaclient.contract.get_available_resources", return_value={})
+    @mock.patch(
+        "uaclient.entitlements.is_config_value_true", return_value=False
+    )
+    def test_entitlement_beta_and_already_enabled(
```
orndorffgrant, line 355: @lucasmoura I added this test. Let me know if this covers what you were looking for.

Also I went with creating a custom entitlement class for this test because it seemed easier than getting the right combination of mocks, but let me know if there's a nicer way to do this with mocks. Specifically, we need:
1. the real `can_enable` (so that it can appropriately print the already enabled message)
2. `enable` to call `can_enable` and return False
3. `is_beta` to be True

lucasmoura: I think I have an idea to better test this, but let me know if it is an overkill here

We can split the enable operation in two steps:

1. Check if enable is possible
2. Perform enable operation

Looking at `RepoEntitlement` and `LivepatchEntitlement`, the first step is the same, they basically call the `can_enable` method. Although `RepoEntitlement` has an extra logic to ouput pre-enable messages, but this is an generic logic that could be moved into `UAEntitlement`.

Because of that, I think we can update the enable function on `UAEntitlement` to:

```python
    def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
        """Enable specific entitlement.

        :param silent_if_inapplicable:
            Don't emit any messages until after it has been determined that
            this entitlement is applicable to the current machine.

        @return: True on success, False otherwise.
        """
        msg_ops = self.messaging.get("pre_enable", []) 
        if not handle_message_operations(msg_ops):
            return False
        if not self.can_enable(silent=silent_if_inapplicable):
            return False
        
         return self._perform_enable()

    def _perform_enable() -> bool:
        raise NotImplementedError
```

We would need to update the enable method in `RepoEntitlement` and `LivepatchEntitlement`, but the update should not be problematic. Also, If we do that, we would be able to better test this code in the unittest, since we could then just create a dummy test class like this:

```python
class TestEntitlement(entitlements.base.UAEntitlement):
    pass
```

And then, just use `mock.patch.object`  to mock the other methods as needed. But this will at least ensure that we are not producing an artificial flow for the enable operation.

But again, let me know if you think this is an overkill

orndorffgrant: @lucasmoura I had also noticed that logic was duplicated and likely applicable to any entitlement but wasn't sure if a refactor was necessary. I like the structure you suggest here though! I'll take a pass at this in this PR - thanks!

uaclient/testing/fakes.py
```diff
@@ -34,3 +36,47 @@ def request_url(
         if isinstance(response, Exception):
             raise response
         return response, {"header1": ""}
+
+
+class MockEntitlement(base.UAEntitlement):
```
lucasmoura, line 41: I think the idea here is a good one, like overriding the entitlement status methods, like `application and applicability`, but we already have a similar entitlement factory as can be seen here:

https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/entitlements/tests/conftest.py#L83

I think we bring some of those changes there

uaclient/testing/fakes.py
```diff
@@ -34,3 +36,47 @@ def request_url(
...
+        allow_beta=False,
+    ):
+        super().__init__(cfg, assume_yes=assume_yes, allow_beta=allow_beta)
+        self._disable = disable
+        self._enable = enable
+        self._applicability_status = applicability_status
+        self._application_status = application_status
+        self._contract_status = contract_status
+
+    def disable(self):
```
lucasmoura, line 65: I don't think we should override by default the `disable` and `_perform_enable` method. I think since those methods perform a more complex logic, the user should explicitly mock then in test, if needed

My concern is that we can hide some testing bits by overriding those methods by default.

uaclient/entitlements/base.py
```diff
@@ -128,10 +145,53 @@ def valid_service(self):
...
+        ret = self._perform_enable(
+            silent_if_inapplicable=silent_if_inapplicable
+        )
+        if not ret:
+            return False
+
+        msg_ops = self.messaging.get("post_enable", [])
+        if not handle_message_operations(msg_ops):
+            return False
+
```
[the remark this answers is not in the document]
orndorffgrant: I left that in the `RepoEntitlement` because it didn't seem general enough to be called for every entitlement.  (Also it was causing some bizarre behavior in tests when I tried putting it here, like `os.path.exists` returning true for a file that doesn't exist...)

uaclient/entitlements/esm.py
```diff
@@ -12,8 +12,8 @@
 class ESMBaseEntitlement(repo.RepoEntitlement):
     help_doc_url = "https://ubuntu.com/security/esm"
 
-    def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
-        enable_performed = super().enable(
+    def _perform_enable(self, *, silent_if_inapplicable: bool = False) -> bool:
+        enable_performed = super()._perform_enable(
```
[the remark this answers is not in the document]
orndorffgrant: The super class here is `RepoEntitlement` which does have `_perform_enable` defined. Also, if we were to call enable here, we would re-run any pre/post enable messaging

uaclient/entitlements/fips.py
```diff
@@ -328,8 +328,10 @@ def setup_apt_config(self) -> None:
             )
         super().setup_apt_config()
 
-    def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
-        if super().enable(silent_if_inapplicable=silent_if_inapplicable):
+    def _perform_enable(self, *, silent_if_inapplicable: bool = False) -> bool:
+        if super()._perform_enable(
```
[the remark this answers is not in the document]
orndorffgrant: Same situation

uaclient/entitlements/base.py
```diff
@@ -523,3 +593,25 @@ def application_status(self) -> "Tuple[status.ApplicationStatus, str]":
             A tuple of (ApplicationStatus, human-friendly reason)
         """
         pass
+
+
+def handle_message_operations(
```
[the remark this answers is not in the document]
orndorffgrant: Yep good point, I'll move it

uaclient/entitlements/esm.py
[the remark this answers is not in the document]
lucasmoura: Oh right, my mistake, I missed the right subclass here.

uaclient/entitlements/base.py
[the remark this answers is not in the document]
lucasmoura: Sorry, I think the method is now on `base.py`. If we bring it back to `RepoEntitlement`, no problem at all

uaclient/testing/fakes.py
orndorffgrant: The reason I couldn't use that factory is that it creates an entitlement _object_, but for this test, we need the _class_.
The factory also requires an actual entitlement class to instantiate, but this test isn't testing any particular entitlement, so we'd have to just pick one of esm-apps, livepatch, etc. - which is fine, just perhaps misleading to instantiate a specific subclass when the specific subclass doesn't matter.

I based `MockEntitlement` off of the `ConcreteTestEntitlement` found in `entitlements/tests/test_base.py`. I didn't reuse `ConcreteTestEntitlement` because the tests in `test_base.py` relied on `ConcreteTestEntitlement` actually looking at the `machine-token.json` for entitlement status, which seemed like too much complexity for this test.

Ideally this test would get a class that inherits from `UAEntitlement` and implements the bare minimum so that it can be instantiated. Ideally it would have no side affects or dependencies on files, and it wouldn't actually do anything when "enabled" or "disabled". That's the role this MockEntitlement is supposed to fill.

I can try to use the entitlement factory that you point too though if you'd prefer

uaclient/testing/fakes.py
orndorffgrant: See above longer answer. The point of MockEntitlement is too be a minimal subclass of UAEntitlement that implements all the abstract methods, including disable and _perform_enable. And MockEntitlement is supposed to not do anything.  I should probably add the purpose of the class in a docstring if we stick with this

uaclient/entitlements/base.py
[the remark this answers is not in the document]
orndorffgrant: Oof! Good catch, yep I played around with moving it which is how that happened. I'll move it back

uaclient/testing/fakes.py
orndorffgrant: @lucasmoura This particular test I'm adding in `test_cli_enable.py` is just weird because, unlike all the other tests in that file, it relies on specific functionality in `UAEntitlement`, which complicates everything. I'm tempted to just remove this test since we have the integration test for this scenario. Then i could get rid of this MockEntitlement, but we would keep the `enable`/`_perform_enable` refactor - thoughts?

lucasmoura: My mistake again, I forgot that we cannot directly instantiate the `UAEntitlement`  base class, so I agree the `entitlement_factory` is not the right place for it.

Regarding the `MockEntitlement`  abstraction, I think we could use the `ConcreteTestEntitlement` code here, like move it to the `fakes.py` module but we would not need to move the `factory` method there as well. I mean, the `MockEntitlement` would be the `ConcreteTestEntitlement` and the `test_base` would just use it in the factory method.

But, I do agree that this is turning into a huge overkill over a test that will not bring us any major contribution, since, as you said, we do have an integration test for it. So, if you want to remove it, I don't see a problem here. Also, sorry for the overkill here, I honestly thought that this test was going to be simpler to implement.

orndorffgrant: @lucasmoura Thanks for all your feedback through this. I ended up drastically reducing the scope of the test, so it is just testing to make sure "One moment, checking your subscription" is printed and not printing any other result, which is handled by `can_enable` for this case. That removed the need for the MockEntitlement all together.

So if Jenkins is good, I think I'm happy with this PR (but if integration tests don't pass then I'll take a look of course).

uaclient/cli.py
```diff
@@ -568,17 +537,24 @@ def action_enable(args, cfg, **kwargs):
...
-                assume_yes=args.assume_yes,
-                allow_beta=args.beta,
+            ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME[ent_name]
+            entitlement = ent_cls(
+                cfg, assume_yes=args.assume_yes, allow_beta=args.beta
             )
-
-            if not ent_ret and entitlement not in valid_services_names:
-                entitlements_not_found.append(entitlement)
+            ent_ret = entitlement.enable()
```
lucasmoura, line 546: Just to confirm, we are not passing `silent_if_enable` here anymore because  by default this attribute is false and we are not allowing user to change its value through CLI, right ?

uaclient/tests/test_cli_enable.py
```diff
@@ -395,33 +441,36 @@ def test_entitlement_instantiated_and_enabled(
...
         m_entitlement = m_entitlement_cls.return_value
-        if silent_if_inapplicable:
-            expected_enable_call = mock.call(silent_if_inapplicable=True)
-        else:
-            expected_enable_call = mock.call(silent_if_inapplicable=False)
+        expected_enable_call = mock.call()
         assert [expected_enable_call] == m_entitlement.enable.call_args_list
-        assert ret == m_entitlement.enable.return_value
-
-        assert 1 == m_cfg.status.call_count
```
lucasmoura, line 427: Why are we dropping those asserts ?

uaclient/cli.py
orndorffgrant: Correct! And this isn't a functional change from before.
Before, the silent param was being set to the default of False by the old _perform_enable, and now it is getting the default of false by ent.enable

uaclient/tests/test_cli_enable.py
orndorffgrant: @lucasmoura Good catch! I must've dropped it while getting other asserts working and forgot to add it back. Just pushed and it should be back now

lucasmoura (approved):
Thanks for this @orndorffgrant I will land this if CI does not show any potential problems

orndorffgrant:
@lucasmoura Update on this one. Integration test failures all look unrelated to this PR. More interesting errors that look like jenkins machine issues though. For example, one container didn't get an IP address - this usually happens when there are too many lxd containers on the machine.

(Because I need to force push in a second, the jenkins results I'm referencing above are here: https://jenkins.canonical.com/server-team-jenkins-private/blue/organizations/jenkins/uaclient-ci/detail/PR-1613/14/pipeline/97)

There's a conflict since we merged #1627 - I'll fix and force push in a minute

lucasmoura (approved):

lucasmoura:

PR #1614, opened 2021-05-06
tests: fix tox for new behave tags

lucasmoura:
## Proposed Commit Message
tests: fix tox for new behave tags

We have recently added the series.lts tags to exclude groovy and hiruste tests from running some behave scenarios. However, on the azure, aws and gcp generic tests, we have replaced series.all to series.lts, which is incorrect. We should add series.lts to thoses tests, not remove series.all from them

## Test Steps
Run any of generic cloud integration tests

## Desired commit type::

## Checklist:


orndorffgrant (commented):
It looks like all the "pro" tests (`behave-awspro-*`, `behave-gcppro-*`, `behave-azurepro-*`) used to have `series.all` for them as well but don't any more. Should this PR add `series.all` back for those tests as well?

lucasmoura:
@orndorffgrant We could, but if you look at the PRO tests the are really specific to the releases we have. I mean, there is not PRO test that has a `series.all` tag. Also, I expect the PRO tests to run only on LTS instances, that's why I have not added that tag back to the PRO tests

blackboxsw (approved):
LGTM!

Issue #1600, opened 2021-04-30
Enabling fips after having enabled fips-updates seems to work but uses newer fips-updates kernel instead


davecore82:
Tested on Ubuntu Pro 18.04 in Azure with UA client version 26.3.

I have fips-updates enabled like this:
```
ubuntu@bionicpro2021-04-30-13-19-52:~$ ua status
SERVICE       ENTITLED  STATUS    DESCRIPTION
[...]
fips          yes       n/a       NIST-certified FIPS modules
fips-updates  yes       enabled   Uncertified security updates to FIPS modules
[...]
```

If I try to enable fips, I get the following:
```
ubuntu@bionicpro2021-04-30-13-19-52:~$ sudo ua enable fips
One moment, checking your subscription first
Installation of additional packages are required to make this system FIPS
compliant.
Are you sure? (y/N) y
Cannot enable FIPS when FIPS Updates is enabled
```

This is expected because fips-updates has newer packages than fips does.

But I can disable fips-updates, reboot, and then enable just fips, and reboot again. However the fips kernel is still the new one from fips-updates:

```
ubuntu@bionicpro2021-04-30-13-19-52:~$ sudo ua disable fips-updates
This will disable access to certified FIPS packages.
Are you sure? (y/N) y
Updating package lists
A reboot is required to complete disable operation

ubuntu@bionicpro2021-04-30-13-19-52:~$ ua status
[...]
fips          yes       disabled  NIST-certified FIPS modules
fips-updates  yes       disabled  Uncertified security updates to FIPS modules
[...]

ubuntu@bionicpro2021-04-30-13-19-52:~$ sudo reboot
Connection to 52.179.19.145 closed by remote host.
Connection to 52.179.19.145 closed.

# ssh back in 

ubuntu@bionicpro2021-04-30-13-19-52:~$ sudo ua enable fips
One moment, checking your subscription first
Installation of additional packages are required to make this system FIPS
compliant.
Are you sure? (y/N) y
Updating package lists
Installing FIPS packages
FIPS enabled
A reboot is required to complete install

ubuntu@bionicpro2021-04-30-13-19-52:~$ sudo reboot
Connection to 52.179.19.145 closed by remote host.
Connection to 52.179.19.145 closed.

# ssh back in 

ubuntu@bionicpro2021-04-30-13-19-52:~$ ua status
[...]
fips          yes       enabled   NIST-certified FIPS modules
[...]

PR #1623, opened 2021-05-11
Block enabling fips if fips-updates once enabled

lucasmoura:
## Proposed Commit Message
Block enabling fips if fips-updates once enabled

When fips-updates is enabled in the system, we emit a file that will allows us to kept track of this status. This will be necessary if an user disables fips-updates and tries to enable fips. In that scenario, we cannot allow users to enable fips.

Fixes: #1600

## Test Steps

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -68,6 +68,7 @@ class UAConfig:
         "status-cache": DataPath("status.json", False),
         "notices": DataPath("notices.json", False),
         "marker-reboot-cmds": DataPath("marker-reboot-cmds-required", False),
+        "services-once-enabled": DataPath("services-once-enabled", False),
     }  # type: Dict[str, DataPath]
 
     _entitlements = None  # caching to avoid repetitive file reads
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -259,6 +259,12 @@ Feature: Enable command behaviour when attached to an UA staging subscription
             """
             <fips-service> +yes                disabled
             """
+        When I verify that running `ua enable fips --assume-yes` `with sudo` exits `1`
+        Then stdout matches regexp:
+            """
+            Cannot enable FIPS because FIPS Updates was once enabled.
+            """
+        And I verify that files exist matching `/var/lib/ubuntu-advantage/services-once-enabled`
 
         Examples: ubuntu release
            | release | fips-name    | fips-service |fips-apt-source                                                        |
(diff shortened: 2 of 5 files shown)


orndorffgrant (commented):
@lucasmoura Code looks great! I'm running a manual test now but don't expect anything to go wrong.

I do think its probably worth adding an integration test for this scenario - what do you think? it should have to be a vm one, and probably a new "Scenario" instead of tacking on to an existing one

orndorffgrant:
@lucasmoura Manual test went well and thought of one small improvement. I think we should add a message when running `ua enable fips-updates` before the "Are you sure? (y/N)" prompt along the lines of: 
`Warning: You will not be able to enable fips after fips-updates has been enabled, even if you disable fips-updates."

This would make it clearer to the user _before_ they make the irreversible change, so that hopefully less people will accidentally end up in a situation where they have fips-updates but need just fips

lucasmoura:
@orndorffgrant I think the modification we have  on the existing vm test for this change should be enough, but that aspect do you think it lacking ?

Regarding the message, I don't think this is a bad idea, just to be sure, what do you think about this @blackboxsw @mitechie ?

orndorffgrant:
@lucasmoura :facepalm: ignore me - the integration test was at the top of the review and I forgot it was there when I hit the bottom - looks great :+1:

uaclient/config.py
```diff
@@ -68,6 +68,7 @@ class UAConfig:
         "status-cache": DataPath("status.json", False),
         "notices": DataPath("notices.json", False),
         "marker-reboot-cmds": DataPath("marker-reboot-cmds-required", False),
+        "services-once-enabled": DataPath("services-once-enabled", False),
```
blackboxsw, line 71: This almost feels like a simple binary auditing of a service status. It's may be a bit too early for us to invest a lot of time in fleshing that out for this one case. But, we will likely need to address this if we start getting to generic auditing or config changes. I definitely don't think we want N files in /var/lib/ubuntu-advantage/private for every attrubute we are tracking, but, since this is the first case. Let's keep it simple.

blackboxsw (approved):
Good PR @lucasmoura thanks! I think we should probably spend some time this cycle to mini-spec ua client config auditing so we can instrument this in a more generic way once we have a second or third use-case. For our needs today, this totally makes sense and if we need to in the future we can migrate to a generic framework.

Issue #1618, opened 2021-05-07
Missing ca-certificates packages causes unclear/incorrect error message


alnvdl:
**Steps to reproduce**:
1. Starting from an Ubuntu that only has basic packages (e.g., the default `ubuntu` Docker image), run:
```sh
$ sudo ua status
```

**What happens**:
The following error is displayed:
```
Failed to connect to authentication server
Check your Internet connection and try again
```

**What is expected**:
A clearer error message (e.g., "failed to validate authentication server certificate" or even a hint to install `ca-certificates` (which fixes the problem).

This is the log for this error:
```
2021-05-07 16:04:57,499 - cli.py:(659) [DEBUG]: Executed with sys.argv: ['/usr/bin/ua', 'status']
2021-05-07 16:04:57,499 - util.py:(373) [DEBUG]: Reading file: /var/lib/ubuntu-advantage/private/machine-token.json
2021-05-07 16:04:57,499 - config.py:(177) [DEBUG]: File does not exist: /var/lib/ubuntu-advantage/private/machine-token.json
2021-05-07 16:04:57,500 - util.py:(373) [DEBUG]: Reading file: /etc/os-release
2021-05-07 16:04:57,510 - util.py:(412) [DEBUG]: URL [GET]: https://contracts.develop.canonical.com/v1/resources?architecture=amd64&series=focal&kernel=5.4.0-72-generic, headers: {'user-agent': 'UA-Client/20.3', 'accept': 'application/json', 'content-type': 'application/json'}, data: None
2021-05-07 16:04:57,679 - cli.py:(627) [ERROR]: Failed to connect to authentication server
Check your Internet connection and try again. Failed to access URL: https://contracts.develop.canonical.com/v1/resources?architecture=amd64&series=focal&kernel=5.4.0-72-generic. [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1123)
Traceback (most recent call last):
  File "/usr/lib/python3.8/urllib/request.py", line 1350, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/usr/lib/python3.8/http/client.py", line 1255, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1301, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
[...]

PR #1627, opened 2021-05-12
errors: print a more helpful message when ssl fails

orndorffgrant:
## Proposed Commit Message

> errors: print a more helpful message when ssl fails
> 
> Fixes: #1618

## Test Steps

* In some way, mess with ssl's ability to verify a certificate.
    * For example, uninstall `ca-certificates`
* Try to run any `ua` command that talks to the network (attach, fix, etc.)
* Verify that a descriptive and actionable message is printed

## Desired commit type::

## Checklist:

### DIFF
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -196,6 +196,36 @@ Feature: Command behaviour when unattached
            | groovy   | no           |
            | hirsute  | no           |
 
+
+    @series.all
+    Scenario Outline: Useful SSL failure message when there aren't any ca-certs
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `apt remove ca-certificates -y` with sudo
+        When I verify that running `ua fix CVE-1800-123456` `as non-root` exits `1`
+        Then I will see the following on stderr:
+            """
+            Failed to access URL: https://ubuntu.com/security/cves/CVE-1800-123456.json
+            Cannot verify certificate of server
+            Please install "ca-certificates" and try again.
+            """
+        When I run `apt install ca-certificates -y` with sudo
+        When I run `mv /etc/ssl/certs /etc/ssl/wronglocation` with sudo
+        When I verify that running `ua fix CVE-1800-123456` `as non-root` exits `1`
+        Then I will see the following on stderr:
+            """
+            Failed to access URL: https://ubuntu.com/security/cves/CVE-1800-123456.json
+            Cannot verify certificate of server
+            Please check your openssl configuration.
+            """
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
+           | groovy  |
+           | hirsute |
+
+
     @series.focal
     Scenario Outline: Fix command on an unattached machine
         Given a `<release>` machine with ubuntu-advantage-tools installed
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1006,14 +1006,24 @@ def main_error_handler(func):
                 _CLEAR_LOCK_FILE("lock")
             sys.exit(1)
         except util.UrlError as exc:
-            with util.disable_log_to_console():
-                msg_args = {"url": exc.url, "error": exc}
-                if exc.url:
-                    msg_tmpl = ua_status.LOG_CONNECTIVITY_ERROR_WITH_URL_TMPL
-                else:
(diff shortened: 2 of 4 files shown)


orndorffgrant:
I don't believe any of the jenkins test failures are related to this PR, but I will look into them.

features/unattached_commands.feature
```diff
@@ -196,6 +196,23 @@ Feature: Command behaviour when unattached
            | groovy   | no           |
            | hirsute  | no           |
 
+
+    @series.focal
```
lucasmoura, line 200: Can we have that test for all releases ?

features/unattached_commands.feature
```diff
@@ -196,6 +196,23 @@ Feature: Command behaviour when unattached
...
+    Scenario Outline: Useful SSL failure message when there aren't any ca-certs
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `apt remove ca-certificates -y` with sudo
+        When I verify that running `ua fix CVE-1800-123456` `as non-root` exits `1`
+        Then I will see the following on stderr:
+            """
+            Cannot verify certificate of authentication server
+            Try installing `ca-certificates` if it isn't already and try again.
+            Failed to access URL: https://ubuntu.com/security/cves/CVE-1800-123456.json
```
orndorffgrant, line 209: Here is an example of the new message we'll print if there is an SSL verification error. I just realized I used backticks out of habit and not sure if thats what we want here. Also, should we add another line along the lines of "If that doesn't work, check your openssl configuration." - What do you think? @mitechie @lucasmoura @blackboxsw

features/unattached_commands.feature
orndorffgrant: Yep good catch! I had just put one for getting it working, but will change this to "@series.all"

features/unattached_commands.feature
lucasmoura: I think dropping the backticks and adding the extra line is fine.
But let's hear from @mitechie and @blackboxsw on this as well

mitechie: A few comments. I agree that backticks should be quotes instead. 

I'd prefer if the error came first "Failed to access URL:..." and then the other material is the details about the said error. 

Can we tell if the ca-certificates package is installed or not vs leaving it hanging with ambiguity?

orndorffgrant: Ack: will change to quotes and will put "Failed to access..." first.

We could tell if ca-certificates is installed or not - I just wasn't sure how far down the path of debugging the problem we think uaclient should go (e.g. if ca-certificates is installed, do we check if the openssl config file is valid? do we check to see if the cert we got from contracts.canonical.com is what we expected?).

I think checking if ca-certificates is installed or not is a reasonable step for uaclient to take, and we can print a different message accordingly ("Please install ca-certificates" vs "Check your openssl configuration"). We probably shouldn't go any further though. What do you think @mitechie @blackboxsw @lucasmoura ?

mitechie: If it's easy to check the package specifically we know we hit that case for a user and so it seems like a reasonable step to get the user as close to the problem as possible. Thanks for the updates.

lucasmoura:
@orndorffgrant That looks really good, but the new integration test is failing on CI.
After we fix that, I we can land this without an issue

orndorffgrant:
@lucasmoura Ah thanks, taking a look now

orndorffgrant:
@lucasmoura Okay I changed the test slightly and it worked locally. Lets let jenkins try it out

orndorffgrant:
@lucasmoura Alright several integration test failures, but none seem related to this PR.
The xenial ones are known failures right now.
The focal vm one is a forklimit error - likely something to do with jenkins being overworked?
The bionic failures look like its running the current version of the tests against an older version of UAClient. Several fixed things are failing beyond just the things touched in this PR. Any ideas what couldve caused this?

lucasmoura:
@orndorffgrant my hypothesis for bionic is that it is running the 27.0.2 version of uaclient already and this causes the problem #1637 is trying to solve

For the focal vm test, that's new to me. I will need to take a look on it.

However, I don't think we should block this feature because of it

lucasmoura:
@orndorffgrant just confirmed that this is the bionic issue. If we launch a daily bionic container, we can see that uaclient has version 27.0.2

lucasmoura (approved):

orndorffgrant:
Thanks for verifying that @lucasmoura! This certainly makes #1637 more urgent

Issue #1626, opened 2021-05-11
27.0.2: json hook lintian errors regarding position independent executable and ldflags


blackboxsw:
The new json hook raises some lintian warnings:

W: ubuntu-advantage-tools: hardening-no-pie usr/lib/ubuntu-advantage/apt-esm-json-hook
W: ubuntu-advantage-tools: hardening-no-relro usr/lib/ubuntu-advantage/apt-esm-json-hook

Issue #1624, opened 2021-05-11
27.0.2 lintian errors: (SC-13)


blackboxsw:
Resolve lintian errors in debian/changelog:
    W: ubuntu-advantage-pro: spelling-error-in-changelog ammend amend
    W: ubuntu-advantage-pro: spelling-error-in-changelog ommitted omitted
    W: ubuntu-advantage-tools: spelling-error-in-changelog ammend amend
    W: ubuntu-advantage-tools: spelling-error-in-changelog ommitted

. W: ubuntu-advantage-pro: systemd-service-file-refers-to-unusual-wantedby-target lib/systemd/system/ua-auto-attach.service cloud-config.service
 >>  I think that ones just an overly aggressive warning, and is appropriate to add a lintian override for.

blackboxsw:
This is tied to PR https://github.com/canonical/ubuntu-advantage-client/pull/1649

PR #1628, opened 2021-05-12
Release 27 lintian warnings

blackboxsw:
#wip will update with content for what was released to impish

Sync changelog for what we just released in packaging branches on impish for 27.0.2

- Fix a number of lintian warnings as part of upload review and apt-json-hook compile time flags
- Also ignore lintian ubuntu-advantage-pro: systemd-service-file-refers-to-unusual-wantedby-target lib/systemd/system/ua-auto-attach.service cloud-config.service

Plan is to propose this PR back up into main after landing in release-27 to sync commits back up to tip

## Proposed Commit Message

-     lintian: override ubuntu-advantage-pro wanted-by-target cloud-init
    
    Silence warning for pro wanted-by-target cloud-init.service.

- apt-json-hook: fix lintian warnings based on compiled binary
    
    Provide -ldflgas and -buildmod=pie to go build of this hook.
    
    Fixes: #1626

- changelog: sync impish released changelog 27.0.2

 Fixes: #1624

## Test Steps
```bash
build-package
sbuild-it ../out/ubuntu-advantage-tools_27.0.2.dsc
lintian ubuntu-advantage-tools_27.0.2_amd64.deb
```
## Desired commit type::

## Checklist:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -8,24 +8,27 @@ ubuntu-advantage-tools (27.0.2) impish; urgency=medium
     - apt-hook: refactor json hook messaging to be dry
     - tests: fix subp ls error case for powerpc builds
     - jenkinsfile: add --resolve-alternatives for trusty builds
+    - amend changelog: add omitted apt-hook message for 27.0.1 stanza
 
  -- Chad Smith <chad.smith@canonical.com>  Fri, 07 May 2021 11:58:03 -0600
 
 ubuntu-advantage-tools (27.0.1) impish; urgency=medium
 
   * Add .gitignore and cleanup ignored directory .pytest_cache
+  * apt-hook: mitigate failures with true
 
- -- Chad Smith <chad.smith@canonical.com>  Fri, 07 May 2021 11:55:07 -0600
+ -- Chad Smith <chad.smith@canonical.com>  Wed, 28 Apr 2021 13:55:28 -0600
 
 ubuntu-advantage-tools (27.0) impish; urgency=medium
 
   * New upstream release 27.0:
-    - apt-hook: mitigate failures with true
+    - [redacted: actually landed in 27.0.1] apt-hook: mitigate failures with
+      true
     - messages: add optional (s) to apt messaging to include
       singular/plural pkgs
     - apt-hook: avoid reporting and counting duplicate package
       names (GH: #1578)
-    - fix: dont say reboot required when unnecessary (LP: #1926183)
+    - fix: don't say reboot required when unnecessary (LP: #1926183)
     - test: uncomment additional xenial upgrade tests
 
  -- Lucas Moura <lucas.moura@canonical.com>  Tue, 27 Apr 2021 15:31:06 -0300
@@ -409,9 +412,9 @@ ubuntu-advantage-tools (25.0~20.10.1~beta) groovy; urgency=medium
 ubuntu-advantage-tools (24.4) groovy; urgency=medium
 
   * New bug-fix-only release 24.4:
-     - uaclient.version bump to 24.4
-     - fips: honor additionalPackage directive from contract for bionic
-       (GH #1173)
+    - uaclient.version bump to 24.4
+    - fips: honor additionalPackage directive from contract for bionic
+      (GH #1173)
 
  -- Chad Smith <chad.smith@canonical.com>  Tue, 01 Sep 2020 11:14:39 -0600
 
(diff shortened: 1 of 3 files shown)


debian/changelog
```diff
@@ -8,24 +8,27 @@ ubuntu-advantage-tools (27.0.2) impish; urgency=medium
...
+  * apt-hook: mitigate failures with true
 
- -- Chad Smith <chad.smith@canonical.com>  Fri, 07 May 2021 11:55:07 -0600
+ -- Chad Smith <chad.smith@canonical.com>  Wed, 28 Apr 2021 13:55:28 -0600
 
 ubuntu-advantage-tools (27.0) impish; urgency=medium
 
   * New upstream release 27.0:
-    - apt-hook: mitigate failures with true
+    - [redacted: actually landed in 27.0.1] apt-hook: mitigate failures with
```
lucasmoura, line 25: Sorry if I am missing something here, but do we need to have those updates on the changelog ? Can't we just remove keep the correct line on release 27.0.1 and remove the other ones ?

lucasmoura:
@blackboxsw after building the package, I am seeing the following lintian errors:
```
E: ubuntu-advantage-tools changes: bad-distribution-in-changes-file impish                                                                                                                                         
E: ubuntu-advantage-tools source: build-depends-on-essential-package-without-using-version Build-Depends: debianutils                                                                                              
E: ubuntu-advantage-tools: malformed-override Unknown tag ubuntu-advantage-pro: in line 8                                                                                                                          
W: ubuntu-advantage-tools source: package-uses-deprecated-debhelper-compat-version 9                                                                                                                               
W: ubuntu-advantage-pro: systemd-service-file-refers-to-unusual-wantedby-target lib/systemd/system/ua-auto-attach.service cloud-config.service                                                                     
W: ubuntu-advantage-tools changes: unknown-architecture amd64_translations
```

Looking at this output, it seems that we still have the problem `wanted-by` warning.

However, if I run the following lintian command:
`lintian -L ">=warning" --show-overrides ubuntu-advantage-tools_27.0.2_source.changes`

I see the following output:
```
E: ubuntu-advantage-tools source: build-depends-on-essential-package-without-using-version Build-Depends: debianutils
W: ubuntu-advantage-tools source: newer-standards-version 4.5.1 (current is 4.5.0)
W: ubuntu-advantage-tools source: package-uses-deprecated-debhelper-compat-version 9
```

Which looks like what we expect. So I don't know exactly what is happening here regarding the lintian diff between sbuild and lintian. @paride do you know what can cause different outputs like this ?

Also, maybe we can address this lintian error too in this PR:
E: ubuntu-advantage-tools source: build-depends-on-essential-package-without-using-version Build-Depends: debianutils

paride:
Hi Lucas, the difference is caused by *what* you lint. There are three kind of things that lintian can lint:
1. Binary packages (`.deb` files)
1. Source packages (`.dsc` files, which contain a reference to a tarball that contains the `debian/` packaging directory, `cat` one to see it)
1. Upload manifest files (`.changes` files, which can list source and/or binary packages, depending on the build tool and settings).
so:
1. When passed a `.deb` file lintian lints it (do the installed files respect the naming/permission conventions? these kind of things). It can *not* lint the source package bits (`debian/`).
1. When passed a source package that's linted (basically: the `debian/*` packaging is linted)
1. When passed a `.changes` file, whatever that file refers to is linted.

Now you passed lintian a `*_source.changes` files, which is an upload manifest for a source-only upload (it only refers to a source package). And indeed you have only `source:` lintian messages. There should also be a `_amd64.changes` files referring to the *binary* packages, and if you lint that it will output the other warnings.

When it's `sbuild` that runs lintian that's done in the schroot and the `.changes` file contains both the source and binary packages, so everything is linted.

FWIW I normally run lintian with: `lintian -EvIL +pedantic` to get all possible messages (it's somehow really evil and pedantic!).

paride:
You should be able to replicate the sbuild lintian result by running `sbuild` with `--source` and then passing lintian the `_amd64.changes`.

debian/changelog
blackboxsw: @lucasmoura since we already released a d/changelog in impish incorrectly published this line, Robie thought it was more correct and less confusing to others who may have seen the impush release changelog to avoid a complete redaction of a line item. It was a point raised in SRU that we probably shouldn't be magically manipulating old d/changelog entries if we can avoid it.

lucasmoura (approved):

blackboxsw:
Closed in favor of a PR against main https://github.com/canonical/ubuntu-advantage-client/pull/1645
Will cherry-pick those accepted commits from main into release-27 once 1645 lands.

PR #1629, opened 2021-05-12
cis: update cis help url

blackboxsw:
## Proposed Commit Message
Update of CIS URL to https://security-certs.docs.ubuntu.com/en/cis until ubuntu.com/security/cis is available

Also add a commit to fix failing integration tests due to messaging url changes for ESM on Xenial https;//ubuntu.com/16-04 

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/entitlements/tests/test_cis.py
+++ b/uaclient/entitlements/tests/test_cis.py
@@ -107,8 +107,6 @@ class TestCISEntitlementEnable:
             "Updating package lists\n"
             "Installing CIS Audit packages\n"
             "CIS Audit enabled\n"
-            "Visit {} to learn how to use the CIS packages\n".format(
-                CIS_DOCS_URL
-            )
+            "Visit {} to learn how to use CIS\n".format(CIS_DOCS_URL)
         )
         assert (expected_stdout, "") == capsys.readouterr()
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -350,7 +350,7 @@ Feature: Enable command behaviour when attached to an UA staging subscription
             Updating package lists
             Installing CIS Audit packages
             CIS Audit enabled
-            Visit https://ubuntu.com/security/cis to learn how to use the CIS packages
+            Visit https://security-certs.docs.ubuntu.com/en/cis to learn how to use CIS
             """
         When I run `apt-cache policy usg-cisbenchmark` as non-root
         Then stdout does not match regexp:
--- a/uaclient/entitlements/cis.py
+++ b/uaclient/entitlements/cis.py
@@ -6,7 +6,7 @@ except ImportError:
     # typing isn't available on trusty, so ignore its absence
     pass
 
-CIS_DOCS_URL = "https://ubuntu.com/security/cis"
+CIS_DOCS_URL = "https://security-certs.docs.ubuntu.com/en/cis"
 
 
 class CISEntitlement(repo.RepoEntitlement):
@@ -25,8 +25,6 @@ class CISEntitlement(repo.RepoEntitlement):
     ) -> "Dict[str, List[Union[str, Tuple[Callable, Dict]]]]":
         return {
             "post_enable": [
-                "Visit {} to learn how to use the CIS packages".format(
-                    CIS_DOCS_URL
-                )
+                "Visit {} to learn how to use CIS".format(CIS_DOCS_URL)
             ]
         }


uaclient/entitlements/tests/test_cis.py
```diff
@@ -107,7 +107,7 @@ def fake_platform(key=None):
             "Updating package lists\n"
             "Installing CIS Audit packages\n"
             "CIS Audit enabled\n"
-            "Visit {} to learn how to use the CIS packages\n".format(
```
blackboxsw, line 110: Reduced this text to avoid having to line wrap on > 80 characters with new URL

features/attach_validtoken.feature
```diff
@@ -98,7 +98,17 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         When I update contract to use `effectiveTo` as `days=-20`
         And I run `python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py` with sudo
         And I run `update-motd` with sudo
-        Then if `<release>` in `xenial or bionic` and stdout matches regexp:
+        Then if `<release>` in `xenial` and stdout matches regexp:
+        """
+
+        \*Your UA Infra: ESM subscription has EXPIRED\*
```
lucasmoura, line 104: CI is failing because of that test here. To we have the extra newlines between the messages ? I think that is the error here

orndorffgrant: I looked at these messages in detail. The difference seems to be that the actual output uses the URL: "https://ubuntu.com/advantage", not the "16-04" url. I remember that "advantage" is what we want here, because at the point that someone has an expired contract, they need to be pointed directly to the URL that will help them extend their contract.

features/attach_validtoken.feature
```diff
@@ -98,7 +98,17 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
+
+        \d+ additional security update\(s\) could have been applied via UA Infra: ESM.
+
+        Renew your UA services at https:\/\/ubuntu.com\/16-04
+
+        """
+        Then if `<release>` in `bionic` and stdout matches regexp:
```
orndorffgrant, line 111: This test is also failing, but because the "expired" message isn't being printed at all for bionic - not sure if this is in scope of this PR or if you want to break this fix into a different PR

features/attach_validtoken.feature
blackboxsw: Ahh bummer. @orndorffgrant I see that there is at least one bug in rendering 16-04 esm apps url on xenial.

- if esm-apps is disabled and esm-infra enabled we are still emitting https://ubuntu.com/esm instead of  https://ubuntu.com/16-04 in /v/l/u-a/messages/motd-esm-announce. I'm backing out the integration test commit which isn't working anyway and we can address both the integration test failure and that minor fix to motd-esm-announce in the same separate branch.
As far as bionic messaging. we will only see ESM-infra messages in 2 years once bionic goes into ESM. In order to see any esm-apps messages,  we have to treat esm-apps like a non-beta service. So adding the feature uaclient.conf and run ua disable esm-infra will setup APT repo files pinned: never for esm-apps once esm-apps is disabled. 
```bash
features:
   allow_beta: true
```


This will then result in the messages:
```bash
root@dev-b:~# run-parts /etc/update-motd.d/
Welcome to Ubuntu 18.04.4 LTS (GNU/Linux 5.8.0-50-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Fri May 14 15:08:52 UTC 2021

  System load:    0.06      Processes:           23
  Usage of /home: unknown   Users logged in:     0
  Memory usage:   0%        IP address for eth0: 10.195.58.87
  Swap usage:     0%

 * Pure upstream Kubernetes 1.21, smallest, simplest cluster ops!

     https://microk8s.io/

 * Introducing Extended Security Maintenance for Applications.
   Receive updates to over 30,000 software packages with your
   Ubuntu Advantage subscription. Free for personal use.

     https://ubuntu.com/esm

94 packages can be updated.
0 updates are security updates.

New release '20.04.2 LTS' available.
Run 'do-release-upgrade' to upgrade to it.
$ apt upgrade
Reading package lists... Done
Building dependency tree       
Calculating upgrade... Done

*The following packages could receive security updates with
 UA Apps: ESM service enabled:
  libkrb5-3 libgssapi-krb5-2 libk5crypto3 krb5-locales libkrad0 libkrb5support0
Learn more about UA Apps: ESM service at https://ubuntu.com/esm

The following NEW packages will be installed:
  libnetplan0 motd-news-config python3-pexpect python3-ptyprocess
The following packages will be upgraded:

```

orndorffgrant (approved):
I captured the integration test failures with references to discussion here in #1635  - they shouldn't block the URL fix here. Approved and merging

PR #1632, opened 2021-05-13
apt-esm-json-hook: exclude go APT JSON hook on i386

blackboxsw:
## Proposed Commit Message
i386 doesn't deliver golang-1.14-go on Focal++.
So, we will exclude this utility which means
not reporting esm-infra and esm-apps counts during
apt upgrade on i386.

amd64 will retain this behavior.

LP: #1928369

## Test Steps
```bash
sudo sbuid-launchpad-chroot create --architecture="i386" "--name=focal-i386" --series=focal
rm -rf ../out
build-package
sbuild --resolve-alternatives --distr=focal --arch=i386 --arch=all ../out/*dsc
```
 
## Desired commit type::

## Checklist:


debian/control
```diff
@@ -9,7 +9,7 @@ Build-Depends: bash-completion,
                debhelper (>= 13.3) | dh-systemd,
                gettext,
                git,
-               golang-go (>= 2:1.14~) [!powerpc] | golang-1.14-go [!powerpc] | golang-1.10-go [!powerpc !riscv64],
+               golang-go (>= 2:1.14~) [!powerpc !i386] | golang-1.14-go [!powerpc !i386] |  golang-1.10-go [!powerpc !i386 !riscv64],
```
orndorffgrant, line 12: not really important, but an extra space slipped in here
```suggestion
               golang-go (>= 2:1.14~) [!powerpc !i386] | golang-1.14-go [!powerpc !i386] | golang-1.10-go [!powerpc !i386 !riscv64],
```

orndorffgrant (approved):
This successfully excluded the go hook and built for me on i386.
Also confirmed it still successfully builds including the go hook on amd64.
:+1:

blackboxsw (commented):
We are going to reject this as i386 official support doesn't need to be in ua-tools. No need to carry around extra burden as we are not going to support ua-tools on i386 for Focal++.

PR #1637, opened 2021-05-17
tests: allow downgrade when installing uaclient

lucasmoura:
## Proposed Commit Message
tests: allow downgrade when installing uaclient

When testing uaclient through the integration tests, we may end up with a machine that contains a higher version of uaclient
that the one we are testing. In that situation, we should allow downgrading uaclient for testing purposes.

This is true for the current situation on the clouds, were uaclient from the ppa is on version 27.0 while the one in archives has version 27.0.2. In that situation, we will not be able to install the required version of uaclient from our PPA or using `BUILD_PR`.

## Test Steps
Run any aws/azure generic tests using BUILD_PR=1 and BUILD_PR=0

## Desired commit type::

## Checklist:

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -33,13 +33,28 @@ bootcmd:
  - chmod 755 /usr/bin/ua
 """
 
+"""
+Package: *
+Pin: release o=LP-PPA-ua-client-daily
+Pin-Priority: 1001
+"""
+UACLIENT_PPA_PIN = (
+    "UGFja2FnZTogKgpQaW46IHJlbGVhc2Ugbz1MUC1QUE"
+    "EtdWEtY2xpZW50LWRhaWx5ClBpbi1Qcmlvcml0eTogMTAwMQo="
+)
+
 USERDATA_APT_SOURCE_PPA = """\
+write_files:
+  - encoding: b64
+    content: {preference_pin}
+    owner: root:root
+    path: /etc/apt/preferences.d/uaclient
+    permissions: 0644
 apt:
   sources:
     ua-tools-ppa:
         source: deb {ppa_url} $RELEASE main
         keyid: {ppa_keyid}
-packages: [{packages}]
 """
 
 PROCESS_LOG_TMPL = """\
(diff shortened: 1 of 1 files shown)


orndorffgrant:
@lucasmoura This is a good thing to bring up. I think our daily PPA and local builds should be building version 27.1 or 28. We should also probably have the latest changelog from 27.0.2 in `main`. Would updating our changelog so that our "UNRELEASED" version is 27.1 or 28 fix this issue? (And should it be 27.1 or 28?)

lucasmoura:
I have update the code to handle the PPA installation as well.

Also @orndorffgrant yes, changing the version on changelog would definitely fix that problem as well. My concern is only if we have the same problem in the future and we see the test fail for the same reason. This solution now will always prioritize the PPA/deb we are testing.

But regardless of this solution, yes should definitely bring the already released versions changelogs into main. I think after we land #1628 we can do that

orndorffgrant (approved):
This worked great for me! Note that I tested using the lxd bionic tests - replicated the failure on `main` and ensured it worked on this branch. Thanks for figuring this out.

Just one optional suggestion in a comment

features/environment.py
```diff
@@ -33,13 +33,24 @@
  - chmod 755 /usr/bin/ua
 """
 
+UACLIENT_PPA_PIN = (
+    "UGFja2FnZTogKgpQaW46IHJlbGVhc2Ugbz1MUC1QUE"
+    "EtdWEtY2xpZW50LWRhaWx5ClBpbi1Qcmlvcml0eTogMTAwMQo="
```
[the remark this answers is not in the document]
orndorffgrant: Just noticed you added this - thanks! This PR is good to go from my perspective

features/environment.py
```diff
@@ -33,13 +33,29 @@
...
+    owner: root:root
+    path: /etc/apt/preferences.d/uaclient
+    permissions: 0644
 apt:
   sources:
     ua-tools-ppa:
         source: deb {ppa_url} $RELEASE main
         keyid: {ppa_keyid}
-packages: [{packages}]
+packages: [openssh-server]
```
blackboxsw, line 58: I forgot why we need openssh-server again. I thought it was Xenial VM related in pycloudlib. Didn't we drop that image  support in pycloudlib@96b146ee1beb99b8e44e36525e18a9a20e00c3f2 ? Maybe I'm misremembering here.

blackboxsw: Ran both bionic and xenial VM tests without error dropping the openssh-server packages directive. I think we can avoid this since pycloudlib no longer uses  XENIAL_IMAGE_VSOCK_SUPPORT = "images:ubuntu/16.04/cloud"
which didn't include openssh-server in the image.

lucasmoura: Good catch. i have updated the code to remove it

Issue #1630, opened 2021-05-12
More detailed error messages when attachment fails with a 403 (SC-3)


alnvdl:
We've recently introduced slightly more detailed error messages in ua-contracts when machine attachment fails because the contract is not valid. The reason for that was because of customers who were trying to attach machines to their contracts before they became effective. In that case, the client gives them a generic error message directing them to `/advantage`.

We'd like to have this changed to a more informative error message, possibly explaining that their contracts are not valid yet (or if they expired already). Pointing them to `/advantage` can still happen, but it'd be nice to also directly show the actual reason we are denying attachment.

The messages look like this (the wording on the messages could be changed if you'd like something different):
```
Body: {
  "code": "forbidden",
  "message": "contract \"cAPWbzTkpV9EGQnTky8QDUtSDl5KRZO5B6R6PnpHoX0M\" not effective: contract not effective",
  "traceId": "9525486a-82ea-4521-b491-47ef34517e4b"
}
```
And this:
```
Body: {
  "code": "forbidden",
  "message": "contract \"cAJsI3QwlhB7VnJZFaHgXCw-RnsvQam_coYmqz5l9Ekw\" not effective: contract will become effective on 2021-10-01 00:00:00 +0000 UTC",
  "traceId": "5eda9bff-5530-419f-8a1d-202f7193f0c6"
}
```

---

Below is a guide on how you can replicate these scenarios using [the contract client](https://github.com/CanonicalLtd/ua-contracts/tree/develop/cmd/contract):

The expired case:
```
export CONTRACTS_URL="https://contracts.staging.canonical.com/"

echo '{"name": "test-attach-expired"}' | contract call add-account -

echo '{
    "name": "test-contract",
    "products": ["free"],
    "items": [{
        "effectiveFrom": "2020-06-01T00:00:00Z",
        "effectiveTo": "2020-10-01T00:00:00Z",
        "reason": "sf_asset_created",
    }]
}' | contract call add-account-contract <account-id-from-previous-step> - --verbose

echo '{}' | contract call get-contract-token <contract-id-from-previous-step> - --verbose

echo '{}' | contract -a token:<contract-token-from-previous-step> call add-contract-machine - --verbose
```
Result:
```
RESPONSE
Status: 403 Forbidden
Cache-Control: no-store, max-age=0
Content-Length: 177
Content-Type: application/json
Date: Wed, 12 May 2021 23:28:40 GMT
Server: openresty/1.15.8.2
[...]

PR #1638, opened 2021-05-18
attach: print contract server reason for 403 (SC-11)

orndorffgrant:
## Notes
I also added a bit to our .gitignore to cover some integration test artifacts

## Proposed Commit Message

> attach: print contract server reason for 403
>
> When the contract server returns a 403 forbidden for a request to
> attach, it also includes data about the reason for denial. We now use
> this to generate a more descriptive and accurate message than our
> generic "expired" message.
>
> Fixes: #1630

## Test Steps

Try to attach with the expired staging token in lastpass.

## Desired commit type

## Checklist:


lucasmoura:
@orndorffgrant I think we can create a manual test for this. It is not ideal, but it will at least allow us to test this feature more easily on the next SRU

orndorffgrant:
This is blocked on resolving the conversation in #1630 (cc @lucasmoura @blackboxsw @mitechie)

uaclient/util.py
```diff
@@ -232,13 +232,13 @@ def disable_log_to_console():
...
-     @param exception: The exception class to catch and retry for the provided
-        retry_sleeps. Any other exception types will not be caught by the
-        decorator.
+    @param retry_sleeps: List of sleep lengths to apply between
+       retries. Specifying a list of [0.5, 1] tells subp to retry twice
+       on failure; sleeping half a second before the first retry and 1 second
+       before the second retry.
+    @param exception: The exception class to catch and retry for the provided
+       retry_sleeps. Any other exception types will not be caught by the
+       decorator.
```
orndorffgrant, line 241: Looks like my copy of `black` might've done a few extra things here...

uaclient/util.py
```diff
@@ -752,3 +752,29 @@ def handle_message_operations(
             if not functor(**args):
                 return False
     return True
+
+
+def parse_rfc3339_date(dt_str: str) -> datetime.datetime:
```
lucasmoura, line 757: I think I am missing something here, but why do we need to support this RFC here ? I understand that this would be a generic approach, where the contracts team could send different date formats to us and we would be able to parse it. but maybe we don't need it at the moment.

However, if this was already discussed with the contracts team, no problem

uaclient/util.py
```diff
@@ -752,3 +752,29 @@ def handle_message_operations(
...
+    """
+    Parse a datestring in rfc3339 format. Originally written for compatibility
+    with golang's time.MarshalJSON function.
+
+    This drops subseconds.
+
```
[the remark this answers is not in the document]
orndorffgrant: Yeah I think somehow I ended up with a different copy of black that added that. I'll remove it

uaclient/util.py
orndorffgrant: @lucasmoura 
This isn't that generic - it is just one specific date format (and honestly probably doesn't cover the whole spec). We need to support this date format because it is the date format that the golang time package outputs when serializing to json (and I think it is a de facto standard for date strings in json), and it would be unnecessarily complex to override that.

I just noticed we parse a date we got from contract-server when we look at effectiveTo at `config.py:532`, but that is only working because the server has been consistently not including subseconds and consistently been using UTC (`Z`). If either of those ever change, then that line will break. Perhaps it's not likely that will change for production, but if we ever set up a test contract server, it may default to including a local timezone, for example. Or if we start creating arbitrary contracts, we may (purposefully or not) end up with a date including subseconds. Generally though, I think its better for our date parsing to be just a bit more robust and handle these cases rather than crash.

Also, while looking for other places we parse dates just now, I noticed we have a `DatetimeAwareJSONDecoder` class that is currently not successfully parsing dates because it doesn't cover the timezone or subseconds.

So I think we should actually use this new `parse_rfc3339_date` function in `DatetimeAwareJSONDecoder`, and then we won't need to parse dates anywhere else.

features/attach_invalidtoken.feature
```diff
@@ -21,3 +21,22 @@ Feature: Command behaviour when trying to attach a machine to an Ubuntu
...
+
+    @uses.config.contract_token_staging_expired
+    @series.all
+    Scenario Outline: Attach command failure on expired token
+       Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attempt to attach `contract_token_staging_expired` with sudo
+        Then stderr matches regexp:
+             """
+             Attach denied:
+             Contract ".*" .*
```
lucasmoura, line 33: Do we need make the Contract message here this permissive or can we add a more specific message here ?

lucasmoura (approved):
Just one question about the integration test, but LGTM

features/attach_invalidtoken.feature
orndorffgrant: @lucasmoura Right now the test is written to accept any of the three scenarios (expired, not-yet-effective, never-effective), so that the exact status of the token in `UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING_EXPIRED` is more flexible. Currently, I think the one I put in jenkins has the status `never-effective`.

I'm open to making the test more specific to one of the three scenarios - we may have to ask the contract team for a custom token if we decide to test the expired or not-yet-effective scenarios. What do you think?

orndorffgrant:
Integration test failures are unrelated to this PR. Merging

PR #1649, opened 2021-05-21
Lintian fixes, changelog sync and open 28.0 for development (SC-12)

blackboxsw:
Final round of missing changelog commits which resolve:
* lintian warnings such as changelog spelling errors and lintian overrides
* separate package-specific debian/lintain-overrides
* bring in redacted line in debian/changelog from 27.0 which was released to Bionic -> Hirsute but not publishing in impish

Also open version 28.0 for active development so daily PPA is greater than released version of ua-tools

## Proposed Commit Message

* changelog: amend redacted line in 27.0 which was already released to B-H
* lintian: rename package-specific overrides for pro vs tools
* open version 28 for active development
* changelog: open 28.0 for active development
* changelog: sync from impish release of 27.0.2

# Test Instructions
```bash
sed -i 's/UNRELEASED/focal/' debian/changelog
git commit -am 'testchangelog'
build-package
sbuild-it ../out/*dsc
# spot check lintian warnings & errors on sbuild output
```
# check specifically 
## Desired commit type::

## Checklist:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,166 @@
-ubuntu-advantage-tools (27.0) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (27.0.2) impish; urgency=medium
 
-  * Open 27.0 for active development
+  * d/control:
+    - order build-depends alternatives newer first (LP: #1926949)
+    - apt-hook: do not attempt to package go APT JSON hook on some
+      architectures (GH: #1603) (LP: #1927886, LP: #1927795)
+  * Bug-fix release 27.0.2: build failures on riscv64 and powerpc
+    - apt-hook: refactor json hook messaging to be dry
+    - tests: fix subp ls error case for powerpc builds
+    - jenkinsfile: add --resolve-alternatives for trusty builds
+    - ammend changelog: add ommitted apt-hook message for 27.0.1 stanza
 
- -- Lucas Moura <lucas.moura@canonical.com>  Fri, 15 Jan 2021 10:44:55 -0300
+ -- Chad Smith <chad.smith@canonical.com>  Fri, 07 May 2021 11:58:03 -0600
+
+ubuntu-advantage-tools (27.0.1) impish; urgency=medium
+
+  * Add .gitignore and cleanup ignored directory .pytest_cache
+  * apt-hook: mitigate failures with true
+
+ -- Chad Smith <chad.smith@canonical.com>  Wed, 28 Apr 2021 13:55:28 -0600
+
+ubuntu-advantage-tools (27.0) impish; urgency=medium
+
+  * New upstream release 27.0:
+    - apt-hook: mitigate failures with true
+    - messages: add optional (s) to apt messaging to include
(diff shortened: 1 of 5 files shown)


lucasmoura (approved):
+1 on the changelog and lintian overrides changes.

However, I am unable to build the package on focal because of this error here:
```
The following packages have unmet dependencies:                                                          
 sbuild-build-depends-main-dummy : Depends: python3-flake8 but it is not going to be installed           
E: Unable to correct problems, you have held broken packages.                                            
apt-get failed.                                                                                          
E: Package installation failed                                                                           
Not removing build depends: cloned chroot in use  
```

The `python3-flake8` version in focal `-proposed`  is broken:
https://people.canonical.com/~ubuntu-archive/proposed-migration/focal/update_excuses.html

Since this is unrelated to the changes provided here, we should not block this PR on that.
Additionally, the CI errors are also not related to those changes

orndorffgrant (changes requested):
@lucasmoura I'm still getting several lintian failures when building for xenial:

```
E: ubuntu-advantage-tools source: missing-build-dependency-for-dh_-command dh_systemd_start => dh-systemd
E: ubuntu-advantage-tools source: missing-build-dependency-for-dh-addon systemd => dh-systemd
E: ubuntu-advantage-tools source: invalid-arch-string-in-source-relation riscv64 [build-depends: golang-1.10-go [!powerpc !riscv64]]
W: ubuntu-advantage-tools source: newer-standards-version 4.5.1 (current is 3.9.7)
```

I'll look into fixing/overriding these

lucasmoura:
@orndorffgrant just to be sure, some of those lintian warning are expected, for example this error here:
```
W: ubuntu-advantage-tools source: newer-standards-version 4.5.1 (current is 3.9.7)
```

Since we are building on several different releases, each one will have a `newer-standards-version` type of error.

While, this ones for example is possible related to our `or` clause for `dh-systemd` in the control file:
```
missing-build-dependency-for-dh_-command dh_systemd_start
```

Even though this one I think we can silence in lintian-overrides

orndorffgrant:
@lucasmoura I agree that they are expected, so I think we should override them with an explanation so that they don't come up during SRU review

paride:
Hi, I would not silence (override) the newer-standards-version warning, I think it's acceptable and just means that Xenial's lintian doesn't know about the new policy versions. This is fine in my opinion.

On `missing-build-dependency-for-dh_-command` and `missing-build-dependency-for-dh_-command`: I agree these happen because of the `| dh-systemd` dependency. Interestingly the same warnings are generated when the terms of the `or` dependency are swapped. Apparently lintian (at least the version in Xenial) doesn't handle build-deps with alternatives very well. I'd silence this as a false positive providing an explanation in the overrides file.

paride:
And I think it's fine to silence `invalid-arch-string-in-source-relation`.

orndorffgrant (approved):

lucasmoura (approved):

PR #1651, opened 2021-05-24
Cherry pick commits up to 27.1 from main and changelog for 27.1

orndorffgrant:
## Proposed Commit Message

>
>
>

## Desired commit type::

## Checklist:


debian/changelog
```diff
@@ -1,3 +1,60 @@
+ubuntu-advantage-tools (27.1) impish; urgency=medium
+
+  * d/control:
+    - specify debianutils min version
+  * lintian:
+    - override xenial specific errors
+    - rename package-specific overrides for pro vs tools
+    - override ubuntu-advantage-pro wanted-by-target cloud-init
+  * New upstream release 27.1:
+    - changelog: fix lintian typos amend and redact incorrect 27.0 entry
```
lucasmoura, line 10: I think this require an entry similar to the 'd/control` one

debian/changelog
```diff
@@ -1,3 +1,60 @@
...
+      (GH: #1624)
+    - apt-hook:
```
lucasmoura, line 12: I think we can sort the topics and subtopics in the changelog. However, let's keep the tests and jenkins change in the end

debian/changelog
```diff
@@ -1,3 +1,60 @@
...
+      + avoid segfault when comparing null Apt file origin to esm
+        (GH: #1929123)
```
lucasmoura, line 14: This should be a launchpad bug, right ?

debian/changelog
```diff
@@ -1,3 +1,60 @@
...
+      + fix lintian warnings based on compiled binary (GH: #1626)
+      + only add newlines for MOTD if message file length is non-zero
+      + avoid wrapping static message formats at 80 chars
+    - docs: update releases.md per SRU review feedback on branch structuring
+    - cis:
+      + make service non-beta
+      + update cis help url
+      + post enable message pointing to docs
+    - fips:
+      + Block enabling fips if fips-updates once enabled (#1623) (GH: #1600)
```
lucasmoura, line 24: I think we should keep either the PR or the issue id

debian/changelog
```diff
@@ -1,3 +1,60 @@
...
+      + clarify messaging (GH: #1631)
+    - enable: correct messaging for beta service (GH: #1588)
+    - errors: print a more helpful message when ssl fails (GH: #1618)
+    - pro: obtain AWS IMDSv2 API token before trying to grab pkcs7 doc
+      (GH: #1608)
+    - livepatch: alert when snapd does not have wait cmd (GH: #1927905)
```
lucasmoura, line 30: Same thing here for launchpad bug

debian/changelog
```diff
@@ -1,3 +1,60 @@
...
+    - messaging:
+      + esm announcement message reports u.com/16-04
+      + separate _remove_msg_template. emit no warranty on infra disabled
+      + point to u.com/16-04 instead of u.com/advantage on ESM (GH: #1584)
+      + Infra and Apps messaging is mutually exclusive (GH: #1573)
+    - Do not show beta services in autocomplete (GH: #1594)
+    - logging: remove tracebacks for UserFacingErrors (GH: #1586)
+    - attach: do not print contract name if empty
+    - status: do not show info if not on contract (GH: #1592)
+    - tests
```
lucasmoura, line 40: I think we can remove some of those messages. We should keep in the changelog changes that the affect that are meaningful to the user. I know this is a little bit subjective, but entries like:

```
fix tox for new behave tags
add ability to inject custom userdata to a CI run
```

Don't seem meaningful to the user in my opinion and we can remove them

orndorffgrant:
@lucasmoura I think I addressed all of your changelog comments. Let me know how it looks

debian/changelog
```diff
@@ -1,3 +1,55 @@
...
+    - docs: update releases.md per SRU review feedback on branch structuring
+    - enable: correct messaging for beta service (GH: #1588)
+    - errors: print a more helpful message when ssl fails (GH: #1618)
+    - fips:
+      + Block enabling fips if fips-updates once enabled (GH: #1600)
+      + clarify messaging (GH: #1631)
+    - livepatch: alert when snapd does not have wait cmd (LP: #1927905)
+    - logging: remove tracebacks for UserFacingErrors (GH: #1586)
+    - messaging:
+      + esm announcement message reports u.com/16-04
```
lucasmoura, line 33: I think this message seems a little bit redundant with:
```
point to u.com/16-04 instead of u.com/advantage on ESM (GH: #1584)
```

I think we can remove it

debian/changelog
lucasmoura: I think we are missing a `:` here

debian/changelog
```diff
@@ -1,3 +1,55 @@
...
+  * lintian:
+    - override ubuntu-advantage-pro wanted-by-target cloud-init
+    - override xenial specific errors
+    - rename package-specific overrides for pro vs tools
+  * New upstream release 27.1:
+    - apt-hook:
+      + avoid segfault when comparing null Apt file origin to esm
+        (LP: #1929123)
+      + avoid wrapping static message formats at 80 chars
+      + fix lintian warnings based on compiled binary (GH: #1626)
```
lucasmoura, line 16: I think the lintian entry should be moved to the lintian section

debian/changelog
orndorffgrant: good catch!

debian/changelog
orndorffgrant: @lucasmoura For this one, lintian pointed out the problem, but the fix was a change to the build flags we're using for the json hook, which is why I put it in the apt-hook section. Would it make more sense here if I change the message to "update go build flags based on lintian warnings"? Or do you still think it fits better in the lintian section?

lucasmoura: I think the updated message will be clearer. If we change it, I see no problem keeping this entry on that section

orndorffgrant: Cool, will do!

debian/changelog
```diff
@@ -1,3 +1,54 @@
...
+    - cis:
+      + make service non-beta
+      + post enable message pointing to docs
+      + update cis help url
+    - docs: update releases.md per SRU review feedback on branch structuring
+    - enable: correct messaging for beta service (GH: #1588)
+    - errors: print a more helpful message when ssl fails (GH: #1618)
+    - fips:
+      + Block enabling fips if fips-updates once enabled (GH: #1600)
+      + clarify messaging (GH: #1631)
```
lucasmoura, line 29: I think we can update this message with the issue title:
`Update output of fips commands`

Just so it will be more clearer

lucasmoura (approved):
I was able to reproduce that branch by running:
```
git cherry -v release-27 main | grep + | awk '{print $2}' | xargs git cherry-pick
```

We will need to manually fix changelog conflicts, but it ends with the same features as this branch

orndorffgrant:
I'll update this comment as I manually run the specific failed tests to verify that they were flaky lxd issues on jenkins:

* behave-lxd-20.04 features/staging_commands.feature:108 passed locally :heavy_check_mark:

Issue #1652, opened 2021-05-25
ua enable: beta check is calling `handle_incompatible_services` a second time


lucasmoura:
If we fail to enable a entitlement, we call the `can_enable` method after it to check the exact reason why the entitlement fail to enable. However, the `can_enable` method calls the `handle_incompatible_service` function. If the user has an incompatible service enabled and is using a config override to block disabling services during entitlement, we will see the following output when running an enable operation:

```
One moment, checking your subscription first
Cannot enable FIPS when Livepatch is enabled.

Cannot enable FIPS when Livepatch is enabled.
```

That redundant message is being generated by the second call to `can_enable`. It is possible to verify this bug by running this [integration test](https://github.com/canonical/ubuntu-advantage-client/blob/main/features/attached_enable.feature#L350).

Even worse, if the user has a incompatible service enabled during the enable operation and chooses to not disable it, we will fail the enable operation and prompt the user again because of that second call to `can_enable`

Issue #1634, opened 2021-05-14
`UACLIENT_BEHAVE_REUSE_IMAGE` and `-D reuse_container=` don't seem to be working


orndorffgrant:
The integration tests being slow affects us in a few main ways. (My use of "we" might be a bit presumptuous here - please call me out if you disagree with these).

- We are discouraged from writing new integration tests because each new scenario adds a significant amount of time to run the tests.
- We are discouraged from running the integration tests locally, because we might as well push and let jenkins start running them.
- We spend a lot of time waiting on jenkins to finish running tests before merging PRs, slowing overall velocity
- We don't even run all of our integration tests on PRs (because they're slow), so we don't notice some bugs until release time.

Most of the time seems to be spent on setting up fresh containers/vms to run the test in. Fresh systems are necessary for testing ua because the purpose of ua is to modify global system state, but there are likely ways to reduce the time it takes to set up the fresh systems. I can `lxc launch` a container and install ua from the daily ppa in under a minute manually, so it seems like it shouldn't take drastically longer than that to set up a test scenario.

I have run into problems with the documented ways of reusing images/containers for the tests (`UACLIENT_BEHAVE_REUSE_IMAGE` and `-D reuse_container=`). I don't have detailed notes, but just following the instructions in the readme wasn't enough to get them working for me. If we decide these methods are sufficient, then resolving this issue may just involve updating the documentation to make it more obvious to a newbie like me how to get these working on a fresh system.

orndorffgrant:
We've made good progress on the integration tests. Let's just leave this open to represent that `UACLIENT_BEHAVE_REUSE_IMAGE` and `-D reuse_container=` don't seem to be working. If those issues are resolved, then we can close this.

orndorffgrant:
These features were removed recently

PR #1653, opened 2021-05-25
enable: don't allow disabling services on beta check

lucasmoura:
## Proposed Commit Message

If we fail to enable a service, we call the can_enable method directly. We do that to verify if the reason we failed to enable a service is because the service is in beta mode. However, the can_enable method calls the handle_incompatible_services method, which may prompt users to disable the incompatible service again. Since this check is only used to see if the service is beta, we should not allow for disabling incompatible services when we do it.

Fixes: #1652

PS: I believe we could have a better solution for this issue if we make the [enable](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/entitlements/base.py#L148) method return the reason why it failed. In that way, we would not need to perform the second `can_enable` call. However, this will be a bigger change in the codebase, since the `enable` method is not being called only by the `cli` module. We would need to modify some unittest and potentially test more integration scenarios.

## Test Steps
Run the modified unittest and run this [integration test](https://github.com/canonical/ubuntu-advantage-client/blob/main/features/attached_enable.feature#L350)

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -547,7 +547,10 @@ def action_enable(args, cfg, **kwargs):
             cfg.status()  # Update the status cache
 
             if not ent_ret:
-                can_enable, reason = entitlement.can_enable(silent=True)
+                can_enable, reason = entitlement.can_enable(
+                    silent=True, allow_disable=False
+                )
+
                 if (
                     not can_enable
                     and reason == ua_status.CanEnableFailureReason.IS_BETA
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -206,12 +206,14 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         return True
 
     def can_enable(
-        self, silent: bool = False
+        self, silent: bool = False, allow_disable: bool = True
     ) -> "Tuple[bool, Optional[CanEnableFailureReason]]":
         """
         Report whether or not enabling is possible for the entitlement.
 
         :param silent: if True, suppress output
+        :param allow_disable: if True, allow disabling incompatible services
+
         :return:
             tuple of (bool, CanEnableFailureReason).
             (True, None) if can enable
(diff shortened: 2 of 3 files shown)


orndorffgrant:
Thanks for noticing this and putting up a quick fix @lucasmoura! This fix was going to be my first suggestion too.

Modifying `enable` to return the failure reason is also a good idea, and I don't think would be too terrible to do for this release if you think that is better.

However, I think the greater problem here is that `can_enable` has side effects at all. Such an inconspicuously named method really shouldn't be printing or prompting anything IMHO. I think the longer term fix (which is almost certainly too much to squeeze into this release) would be to refactor so that `can_enable` never prints or prompts anything. We would move the `handle_incompatible_services` call into `enable` and only add the incompatible services check to `can_enable`. This would also require moving all the `print` calls that are currently in `can_enable` probably into `enable`. This refactor seems like it would have additional effects that I'm not aware of though, which is why I think it may not be worth trying to do it for this release. What are your thoughts?

Also, would it be too much to have jenkins run this test on PRs too? (If that would be too much, and thats why it was relegated to the daily job, then thats totally okay)

lucasmoura:
@orndorffgrant I think you are right on that approach. `can_enable` should not have methods that change the status of the system and I do agree that this is the direction we should go moving forward.

Also, I agree that this may be a bigger refactor for this release, but we should definitely do that for 27.2

Regarding running the test on CI, we could, but the VM tests are really time consuming. That's why we only run the focal vm test here, which only has one test scenario. The idea for running only focal, was to guarantee that the package could be installed and run on a VM. However, I don't think we need this type of verification anymore.

With that said, I think we should take some time after release 27.1 and better design the tests we want to run on CI while also trying to address the problems you raised regarding the integration tests.

orndorffgrant:
@lucasmoura Agreed! I went ahead and added another bullet point to #1634 about this.

So I think my suggested path forward is:
1. I'll review in more detail this PR, run the integration tests locally, and we'll merge this fix for 27.1
2. I'll add a new issue explaining our plan for refactoring `can_enable` that we can address in 27.2 (@mitechie should I add this just to Jira now? or to both github and Jira?)
3. We'll spend time on optimizing our integration tests during 27.2 just like you suggest

Does that sound okay?

orndorffgrant (approved):
Code looks good, manual test passed and integration test passed locally. I think this is good for 27.1 and we can do a bigger refactor for 27.2

lucasmoura:
@orndorffgrant I think that is a great plan for us to do. Thanks for putting it up

lucasmoura:
CI error is unrelated to the proposed fix

PR #1661, opened 2021-05-28
fips: add gov clouds to metapackage override logic (SC-71)

lucasmoura:
## Proposed Commit Message
fips: add gov clouds to metapackage override logic

When overriding the metapackage to be installed on cloud instances, we are not taking into consideration the cloud-id of government clouds. This is a problem, because those clouds should also install the cloud specific fips metapackages.

## Test Steps
Run FIPS tests for aws and azure generic scenarios

## Desired commit type::

## Checklist:

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -265,6 +265,33 @@ mark_reboot_cmds_as_needed() {
     add_notice "$msg_name"
 }
 
+notify_wrong_fips_metapackage_on_cloud() {
+
+    # On xenial, we don't have FIPS optimized kernels on the clouds.
+    # Because of that, we block enabling FIPS on xenial clouds. However,
+    # we do have a config override that allows users to install the generic
+    # FIPS packages into the cloud instance. We don't want to notify those users
+    if [ "$VERSION_ID" = "16.04" ]; then
+        return
+    fi
+
+    fips_metapkg="ubuntu-fips"
+
+    cloud_id=""
+    if command -v "cloud-id" > /dev/null ; then
+      cloud_id=$(cloud-id)
+    fi
+
+    # If the package is not installed, we don't want the postinst script to fail
+    fips_installed=$(dpkg-query -W --showformat='${db:Status-Status}\n' $fips_metapkg 2>/dev/null || true)
+
+    if echo "$cloud_id" | grep -E -q "^(azure|aws)"; then
+      if echo "$fips_installed" | grep -E -q "installed"; then
+        add_notice NOTICE_WRONG_FIPS_METAPACKAGE_ON_CLOUD
+      fi
+    fi
+}
+
 case "$1" in
     configure)
       PREVIOUS_PKG_VER=$2
@@ -305,6 +332,8 @@ case "$1" in
             add_notice MESSAGE_FIPS_INSTALL_OUT_OF_DATE
         fi
       fi
+
+      notify_wrong_fips_metapackage_on_cloud
 
       # CACHE_DIR is no longer present or used since 19.1
       rm -rf /var/cache/ubuntu-advantage-tools
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -1,4 +1,5 @@
 import os
+import re
 
 from itertools import groupby
 
@@ -153,8 +154,12 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         if series != "bionic":
             return packages
 
-        cloud_id = get_cloud_type()
-        if not cloud_id or cloud_id not in ("azure", "aws"):
+        cloud_match = re.match(
+            r"^(?P<cloud>(azure|aws)).*", get_cloud_type() or ""
+        )
+        cloud_id = cloud_match.group("cloud") if cloud_match else ""
+
+        if cloud_id not in ("azure", "aws"):
             return packages
 
         cloud_metapkg = "ubuntu-{}-fips".format(cloud_id)
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -241,6 +241,15 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                 env=env,
             )
 
+    def enable(self, *, silent_if_inapplicable: bool = False) -> bool:
+        if super().enable(silent_if_inapplicable=silent_if_inapplicable):
+            self.cfg.remove_notice(
+                "", status.NOTICE_WRONG_FIPS_METAPACKAGE_ON_CLOUD
+            )
+            return True
+
+        return False
+
 
 class FIPSEntitlement(FIPSCommonEntitlement):
 
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -290,6 +290,15 @@ NOTICE_FIPS_MANUAL_DISABLE_URL = """\
 FIPS kernel is running in a disabled state.
   To manually remove fips kernel: https://discourse.ubuntu.com/t/20738
 """
+NOTICE_WRONG_FIPS_METAPACKAGE_ON_CLOUD = """\
+Warning: FIPS kernel is not optimized for your specific cloud.
+To fix it, run the following commands:
+
+    1. sudo ua disable fips
+    2. sudo apt-get remove ubuntu-fips
+    3. sudo ua enable fips --assume-yes
+    4. sudo reboot
+"""
 PROMPT_FIPS_PRE_ENABLE = (
     """\
 This will install the FIPS core packages.
--- a/uaclient/entitlements/tests/test_fips.py
+++ b/uaclient/entitlements/tests/test_fips.py
@@ -249,6 +249,30 @@ class TestFIPSEntitlementEnable:
         assert repo_enable_return_value is fips_entitlement._perform_enable()
         assert expected_remove_notice_calls == m_remove_notice.call_args_list
 
+    @pytest.mark.parametrize(
+        "repo_enable_return_value, expected_remove_notice_calls",
+        [
+            (
+                True,
+                [mock.call("", status.NOTICE_WRONG_FIPS_METAPACKAGE_ON_CLOUD)],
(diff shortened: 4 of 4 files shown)


blackboxsw (commented):
one minor unrelated fix in test_fips.py:
can we rename the test
`test_packages_are_override_when_bionic_cloud_instance`

to 
`test_packages_are_overridden_on_bionic_cloud_instances`



Package versioning looks good given that we aren't going to remove ubuntu-fips, but also install ubuntu-aws-fips. The cloud-specific FIPS version is higher than generic FIPS kernel version so reboots will prefer the cloud-specific version


```bash
$ dpkg --compare-versions 4.15.0.2000.1 gt 4.15.0-1011.12 && echo "cloud fips greater than generic fips"
cloud fips greater than generic fips
```

blackboxsw (commented):
Two questions: 
- Should we also emit a notice that a reboot is required `ua status` if we are on aws, with fips enabled, but we are not on the cloud-specific FIPS kernel?
- is there and LP bug of GH bug which should be added to the footer of the commit message?

debian/ubuntu-advantage-tools.postinst
```diff
@@ -277,6 +295,8 @@ case "$1" in
               $ESM_INFRA_APT_SOURCE_FILE
       fi
 
+
```
blackboxsw, line 298: Extra whitespace needed?

debian/ubuntu-advantage-tools.postinst
```diff
@@ -265,6 +265,24 @@ mark_reboot_cmds_as_needed() {
     add_notice "$msg_name"
 }
 
+check_wrong_fips_metapackage_on_cloud() {
+      fips_metapkg="ubuntu-fips"
+
+      cloud_id=""
+      if which "cloud-id" > /dev/null ; then
```
blackboxsw, line 272: Let's run `shellcheck -x debian/<script>` everytime we touch debian/* script files to ensure we don't erode our scripts unintentionally.

Minor nit raised here  by shellcheck:
      if which "cloud-id" > /dev/null ; then
         ^---^ SC2230: which is non-standard. Use builtin 'command -v' instead.

```suggestion
    if command -v cloud-id > /dev/null 2>&1; then
        cloud_id=$(cloud-id)
    fi
```

debian/ubuntu-advantage-tools.postinst
```diff
@@ -265,6 +265,24 @@ mark_reboot_cmds_as_needed() {
     add_notice "$msg_name"
 }
 
+check_wrong_fips_metapackage_on_cloud() {
+      fips_metapkg="ubuntu-fips"
```
blackboxsw, line 269: Note whitespace offset seems to be 6 spaces indent here instead of 4-based like the rest of the file

debian/ubuntu-advantage-tools.postinst
```diff
@@ -265,6 +265,24 @@ mark_reboot_cmds_as_needed() {
...
+
+      cloud_id=""
+      if which "cloud-id" > /dev/null ; then
+        cloud_id=$(cloud-id)
+      fi
+
+      # If the package is not installed, we don't want the postinst script to fail
+      fips_installed=$(dpkg-query -W --showformat='${Status}\n' $fips_metapkg 2>/dev/null || true)
```
blackboxsw, line 277: Might you want "${db:Status-Status}" instead of  ${Status} since we don't have to support this on Trusty?  db:Status-Status gives us a simple "installed" to check for instead of "install ok installed"

uaclient/status.py
```diff
@@ -290,6 +290,15 @@ class CanEnableFailureReason(enum.Enum):
 FIPS kernel is running in a disabled state.
   To manually remove fips kernel: https://discourse.ubuntu.com/t/20738
 """
+NOTICE_WRONG_FIPS_METAPACKAGE_ON_CLOUD = """\
+Warning: FIPS kernel is not optimized for your specific cloud.
```
blackboxsw, line 294: While I don't really like the multi-step NOTICES messaging here, the alternatives would be:
 1. bake this logic into either reboot_cmds which ultimately would require two reboots to transition out of the generic fips kernel after next reboot 
-  OR -
2. add a background process to run right after postinst which would perform ua disable fips, apt remove ubuntu-fips; ua enable fips steps (minus the reboot) and add a MESSAGE_FIPS_REBOOT_REQUIRED.

Option 2 would be nicer in that we would only tell the customer "FIPS support requires system reboot to complete configuration."

Either of these options feel a bit heavyweight for this bug fix.  But, let's discuss with @mitechie as we consider whether either of these cases are worth the effort or usability concerns.

blackboxsw (commented):
Couple of nits and we can discuss tomorrow about whether this notice alone is acceptable for this corner case.

uaclient/status.py
lucasmoura: For option 2, we will also need to modify how we remove FIPS reboot notices on the code. Even if we manually add this notice after running those commands, if the user runs `ua status` just after it finishes, we will remove it because of this [check](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/entitlements/fips.py#L186)

lucasmoura:
@blackboxsw code updated

uaclient/entitlements/fips.py
```diff
@@ -153,8 +154,12 @@ def _replace_metapackage_on_cloud_instance(
...
             return packages
 
-        cloud_id = get_cloud_type()
-        if not cloud_id or cloud_id not in ("azure", "aws"):
+        cloud_match = re.match(
+            r"^(?P<cloud>(azure|aws)).*", get_cloud_type() or ""
+        )
+        cloud_id = cloud_match.group("cloud") if cloud_match else ""
+
+        if cloud_id not in ("azure", "aws"):
```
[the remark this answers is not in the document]
blackboxsw: Since we weren't strongly on the fence one way or another.

Let's stick with the regex because it's a more strict compound check in "one" operation. Though it ultimately takes more lines to express that.

debian/ubuntu-advantage-tools.postinst
```diff
@@ -265,6 +265,24 @@ mark_reboot_cmds_as_needed() {
...
+    cloud_id=""
+    if command -v "cloud-id" > /dev/null ; then
+      cloud_id=$(cloud-id)
+    fi
+
+    # If the package is not installed, we don't want the postinst script to fail
+    fips_installed=$(dpkg-query -W --showformat='${db:Status-Status}\n' $fips_metapkg 2>/dev/null || true)
+
+    if echo "$cloud_id" | grep -E -q "^(azure|aws)"; then
+      if echo "$fips_installed" | grep -E -q "installed"; then
```
blackboxsw, line 280: @lucasmoura Don't we also need to check too if we are on bionic or greater? Xenial doesn't have a cloud-optimized kernel on most clouds.

blackboxsw (changes requested):
I think we may need to add in a <= bionic exclusion to this post-inst to avoid trying to install something that doesn't exist on Xenial.

debian/ubuntu-advantage-tools.postinst
lucasmoura: Good catch, we do have a way to unblock installing fips on xenial. Because of that, we need that bionic >= check
I will update the code

lucasmoura:
@blackboxsw code updated

debian/ubuntu-advantage-tools.postinst
```diff
@@ -265,6 +265,26 @@ mark_reboot_cmds_as_needed() {
     add_notice "$msg_name"
 }
 
+check_wrong_fips_metapackage_on_cloud() {
```
blackboxsw, line 268: Minor rename the other check_* funtions return a value and are intended for use in conditionals, this intended use different, call and 
```suggestion
notify_wrong_fips_metapackage_on_cloud() {
```

debian/ubuntu-advantage-tools.postinst
```diff
@@ -265,6 +265,26 @@ mark_reboot_cmds_as_needed() {
...
+    fips_metapkg="ubuntu-fips"
+
+    cloud_id=""
```
blackboxsw, line 271: Let's early exi if on 16.04 and avoid performing extra work and  nesting below.
```suggestion
    cloud_id=""

    if [ "$VERSION_ID" = "16.04" ]; then                                        
        return                                                                  
    fi 
```

blackboxsw (changes requested):
Last round and +1

lucasmoura:
@blackboxsw code updated

blackboxsw (approved):
@lucasmoura I walked through it excellent work!

```bash
ubuntu@ip-172-31-16-47:~$ sudo ua status
cis           yes       disabled  Center for Internet Security Audit Tools
esm-apps      yes       enabled   UA Apps: Extended Security Maintenance (ESM)
fips          yes       disabled  NIST-certified core packages
fips-updates  yes       disabled  NIST-certified core packages with priority security updates

NOTICES
Warning: FIPS kernel is not optimized for your specific cloud.
To fix it, run the following commands:

    1. sudo ua disable fips
    2. sudo apt-get remove ubuntu-fips
    3. sudo ua enable fips --assume-yes
    4. sudo reboot


                Account: 329910648901
           Subscription: 329910648901
            Valid until: 9999-12-31 00:00:00+00:00
ubuntu@ip-172-31-16-47:~$ sudo ua enable fips --assume-yes
Updating package lists
Installing FIPS packages
FIPS enabled
A reboot is required to complete install.
ubuntu@ip-172-31-16-47:~$ sudo reboot
Connection to 54.196.209.117 closed by remote host.

ubuntu@ip-172-31-16-47:~$ uname -r 
4.15.0-2000-aws-fips
ubuntu@ip-172-31-16-47:~$ dpkg -l | egrep 'ubuntu-fips|ubuntu-aws-fips'
ii  ubuntu-aws-fips                          1.1.4                                           amd64        Install and configure linux-aws-fips kernel and user space modules
rc  ubuntu-fips                              1.1.4                                           amd64        Install and configure linux-fips kernel and user space modules

```
```

Issue #1659, opened 2021-05-27
`ua enable fips` failure -- Packages Downgraded error


arossbell:
A customer opened a case against an Ubuntu image (non-PRO) on AWS GovCloud that `ua enable fips` fails.
Without debug messages, a quiet error.
```
# ua enable fips
One moment, checking your subscription first
Installation of additional packages are required to make this system FIPS
compliant.
Are you sure? (y/N) y
FIPS cannot be enabled with Livepatch.
Disable Livepatch and proceed to enable FIPS? (y/N)y
Disabling incompatible service: Livepatch
Updating package lists
Installing FIPS packages
Updating package lists
Could not enable FIPS.
```

But among debug messages, the issue is clear. (Also including more information about the system.)
```
# ua --debug enable fips
[ . . . ]
{'Authorization': 'Bearer <REDACTED>', 'accept': 'application/json', 'content-type': 'application/json', 'user-agent': 'UA-Client/27.0.2~18.04.1'}, data: {"machineId": "<REDACTED>", "architecture": "amd64", "os": {"distribution": "Ubuntu", "type": "Linux", "version": "18.04 LTS (Bionic Beaver)", "release": "18.04", "series": "bionic", "kernel": "5.4.0-1048-aws"}}
[ . . . ]
DEBUG: Failed running command 'apt-get install --assume-yes -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" ubuntu-fips openssh-client openssh-client-hmac openssh-server openssh-server-hmac openssh-client openssh-client-hmac openssh-server openssh-server-hmac' [exit(100)]. Message: E: Packages were downgraded and -y was used without --allow-downgrades.
[ . . . ]
```

mitechie:
Thank you for the bug report. We'll investigate and get back. If possible can you confirm the ua client version and ubuntu system details, such as the VM id used to launch the instance? 

    ua --version

Thank you!

arossbell:
Hi @mitechie,

The package is versioned `27.0.2~18.04.1` and the build.info serial from the server system is `20190212.1`

Thanks!

orndorffgrant:
@arossbell Thanks for the additional info. Do we also know what versions of the listed packages were before attempting to enable fips? `openssh-client` `openssh-client-hmac` `openssh-server` `openssh-server-hmac`

arossbell:
Hi @orndorffgrant,

[...]

PR #1662, opened 2021-05-28
repo: allow downgrading packages during enable (SC-69)

lucasmoura:
## Proposed Commit Message
repo: allow downgrading packages during enable

When enabling a repo entitlement, like FIPS, we are currently not allowing for package downgrades. This is problematic if users are using package from a third party, that could have versions higher than the ones we are delivering. To enforce that users will run the right packages for the services we are enabling, we are now supporting downgrading packages when enabling a service.

Fixes: #1659

## Test Steps
Since we are just adding a new apt flag, run an existing FIPS vm test to confirm that the flag is not breaking the enable step

## Desired commit type::

## Checklist:


blackboxsw (approved):
+1 works like a charm and resolves this particular failure case.

PR #1665, opened 2021-06-03
tests: remove clashing lxd user-data (SC-91)

orndorffgrant:
## Background
We need this change before we upgrade pycloudlib to incorporate https://github.com/canonical/pycloudlib/pull/148

## Proposed Commit Message

> tests: remove clashing lxd user-data
>
> lxd relies on a write_files vendor-data directive to set up the
> lxd-agent on lxd vms of some ubuntu releases. Our use of write_files was
> overwriting the vendor-data and preventing the lxd-agent from coming up.
> 
> Here we use a runcmd to accomplish the same thing we were using
> write_files for in our integration tests.

## Test Steps

Make sure integration tests still pass

## Desired commit type::

## Checklist:

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -33,29 +33,18 @@ bootcmd:
  - chmod 755 /usr/bin/ua
 """
 
-"""
-Package: *
-Pin: release o=LP-PPA-ua-client-daily
-Pin-Priority: 1001
-"""
-UACLIENT_PPA_PIN = (
-    "UGFja2FnZTogKgpQaW46IHJlbGVhc2Ugbz1MUC1QUE"
-    "EtdWEtY2xpZW50LWRhaWx5ClBpbi1Qcmlvcml0eTogMTAwMQo="
-)
-
+# we can't use write_files because it will clash with the
+# lxd vendor-data write_files that is necessary to set up
+# the lxd_agent on some releases
 USERDATA_APT_SOURCE_PPA = """\
-write_files:
-  - encoding: b64
-    content: {preference_pin}
-    owner: root:root
-    path: /etc/apt/preferences.d/uaclient
-    permissions: 0644
+runcmd:
+  - "printf \\"Package: *\\nPin: release o=LP-PPA-ua-client-daily\\nPin-Priority: 1001\\n\\" > /etc/apt/preferences.d/uaclient"
 apt:
   sources:
     ua-tools-ppa:
         source: deb {ppa_url} $RELEASE main
         keyid: {ppa_keyid}
-"""
+"""  # noqa: E501
 
 PROCESS_LOG_TMPL = """\
 returncode: {}
@@ -698,7 +687,7 @@ def create_uat_image(context: Context, series: str) -> None:
             ppa = ppa.replace("ppa:", "http://ppa.launchpad.net/") + "/ubuntu"
 
         user_data += USERDATA_APT_SOURCE_PPA.format(
-            ppa_url=ppa, ppa_keyid=ppa_keyid, preference_pin=UACLIENT_PPA_PIN
+            ppa_url=ppa, ppa_keyid=ppa_keyid
         )
     inst = context.config.cloud_manager.launch(
         instance_name=build_container_name, series=series, user_data=user_data
--- a/features/environment.py
+++ b/features/environment.py
@@ -36,15 +36,18 @@ bootcmd:
 # we can't use write_files because it will clash with the
 # lxd vendor-data write_files that is necessary to set up
 # the lxd_agent on some releases
-USERDATA_APT_SOURCE_PPA = """\
+USERDATA_RUNCMD_PIN_PPA = """\
 runcmd:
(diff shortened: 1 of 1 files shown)


features/environment.py
```diff
@@ -33,29 +33,18 @@
...
+# lxd vendor-data write_files that is necessary to set up
+# the lxd_agent on some releases
 USERDATA_APT_SOURCE_PPA = """\
-write_files:
-  - encoding: b64
-    content: {preference_pin}
-    owner: root:root
-    path: /etc/apt/preferences.d/uaclient
-    permissions: 0644
+runcmd:
```
blackboxsw, line 40: I think we need to do this in bootcmd as runcmd doesn't happen until the end of the cloud-init user-data is processed (and packages are already installed/upgraded before your pin file is written).

blackboxsw: You probably already know this  for cloud-init, but you can see when each cloud-init config module runs by checking order of module names listed in /etc/cloud/cloud.cfg cloud_init_modules vscloud_config_modules vs  cloud_final_modules. The order that config modules are listed is the order in which cloud-init runs them during initial system boot 

and those cloud.cfg section names map closely to the cloud-init boot stages where they will be run https://cloudinit.readthedocs.io/en/latest/topics/boot.html

blackboxsw: Grr, also looks like we might get into a pickle w/ bootcmd as we use it on PRO setup too.
What do you think about this logic which allows us to append to PRO no auto-attach user-data if set, handle custom user-data case with no bootcmd directive, and raise error if bootcmd were in custom userdata.


```diff
diff --git a/features/environment.py b/features/environment.py
index d3249fed..bb59b623 100644
--- a/features/environment.py
+++ b/features/environment.py
@@ -28,17 +28,19 @@ DAILY_PPA_KEYID = "6E34E7116C0BC933"
 USERDATA_BLOCK_AUTO_ATTACH_IMG = """\
 #cloud-config
 bootcmd:
- - cp /usr/bin/ua /usr/bin/ua.orig
- - 'echo "#!/bin/sh\ndate >> /root/ua-calls\n" > /usr/bin/ua'
- - chmod 755 /usr/bin/ua
+  - cp /usr/bin/ua /usr/bin/ua.orig
+  - 'echo "#!/bin/sh\ndate >> /root/ua-calls\n" > /usr/bin/ua'
+  - chmod 755 /usr/bin/ua
+"""
+
+USERDATA_BOOTCMD_PIN_PPA = """\
+  - 'echo -e "Package: *\nPin: release o=LP-PPA-ua-client-daily\nPin-Priority: 1001" > /etc/apt/preferences.d/uaclient'
 """
 
 # we can't use write_files because it will clash with the
 # lxd vendor-data write_files that is necessary to set up
 # the lxd_agent on some releases
 USERDATA_APT_SOURCE_PPA = """\
-runcmd:
-  - "printf \\"Package: *\\nPin: release o=LP-PPA-ua-client-daily\\nPin-Priority: 1001\\n\\" > /etc/apt/preferences.d/uaclient"
 apt:
   sources:
     ua-tools-ppa:
@@ -680,7 +682,14 @@ def create_uat_image(context: Context, series: str) -> None:
         ppa_keyid = context.config.ppa_keyid
         if context.config.ppa.startswith("ppa:"):
             ppa = ppa.replace("ppa:", "http://ppa.launchpad.net/") + "/ubuntu"
-
+        if "bootcmd:" not in user_data:
+            user_data += "bootcmd:\n"
+        elif user_data != USERDATA_BLOCK_AUTO_ATTACH_IMG:
+            raise ValueError(
+                "Custom userdata provided bootcmd directive. Don't know where"
+                " to add PPA pinning in:\n{}".format(user_data)
+            )
+        user_data += USERDATA_BOOTCMD_PIN_PPA
         user_data += USERDATA_APT_SOURCE_PPA.format(
             ppa_url=ppa, ppa_keyid=ppa_keyid
         )
```

orndorffgrant: I'm pretty sure I remember tests failing without the `runcmd` (due to package version issues) and succeeding afterwards (correct ua got installed), so I'm hesitant to add the `bootcmd` complexity unless we need it.

I believe it should work as a `runcmd` because we install uaclient after boot via ssh - not as part of the cloud-init user-data. Unless I'm missing something else?

That is an excellent catch that we only need this pin when we're using the Daily PPA and I'll add that to this PR

lucasmoura: I agree that we using `runcmd` to write the `pin` file here should be enough. We should be installing uaclient after cloud-init succeeds, so the pinning in the runcmd should work fine in that situation.

orndorffgrant: @lucasmoura @blackboxsw I pushed a commit that includes the `DAILY_PPA` check but leaves it as a `runcmd` let me know what you think

lucasmoura (approved):

orndorffgrant:
The extra failing integration test on focal is addressed in #1667

Issue #1654, opened 2021-05-25
`can_enable` should not have side effects (SC-64)


orndorffgrant:
Currently, `can_enable` prints things and can even disable a service via `handle_incompatible_services`. This is confusing for such an inconspicuously named method. Ideally, `can_enable` would not have any such side effects.

We could move the `handle_incompatible_services` call into `enable` and only include a incompatible services check in `can_enable`. This would also require moving all the `print` calls that are currently in `can_enable` probably into `enable`. This refactor seems like it may have additional effects that I'm not aware of though, so will need thorough testing.

Related: #1571 #1652

PR #1667, opened 2021-06-04
refactor: remove side effects from can_enable (SC-64)

orndorffgrant:
## Proposed Commit Message

> refactor: remove side effects from can_enable
> 
> Fixes: #1654
> Fixes: #1571

> refactor: remove unused silent enable param 

> refactor: return CanEnableFailure reason from enable 

## Test Steps

Everything should still work :)

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -550,11 +550,13 @@ def action_enable(args, cfg, **kwargs):
                 not ent_ret
                 and reason is not None
                 and isinstance(reason, ua_status.CanEnableFailure)
-                and reason.reason == ua_status.CanEnableFailureReason.IS_BETA
             ):
-                # if we failed because ent is in beta and there was no
-                # allow_beta flag/config, pretend it doesn't exist
-                entitlements_not_found.append(ent_name)
+                if reason.message is not None:
+                    print(reason.message)
+                if reason.reason == ua_status.CanEnableFailureReason.IS_BETA:
+                    # if we failed because ent is in beta and there was no
+                    # allow_beta flag/config, pretend it doesn't exist
+                    entitlements_not_found.append(ent_name)
 
             ret &= ent_ret
         except exceptions.UserFacingError as e:
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -547,13 +547,12 @@ def action_enable(args, cfg, **kwargs):
             cfg.status()  # Update the status cache
 
             if not ent_ret:
-                can_enable, reason = entitlement.can_enable(
-                    silent=True, allow_disable=False
-                )
+                can_enable, reason = entitlement.can_enable()
 
                 if (
                     not can_enable
-                    and reason == ua_status.CanEnableFailureReason.IS_BETA
+                    and reason is not None
+                    and reason.reason == ua_status.CanEnableFailureReason.IS_BETA
                 ):
                     # if we failed because ent is in beta and there was no
                     # allow_beta flag/config, pretend it doesn't exist
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -543,20 +543,18 @@ def action_enable(args, cfg, **kwargs):
             entitlement = ent_cls(
                 cfg, assume_yes=args.assume_yes, allow_beta=args.beta
             )
-            ent_ret = entitlement.enable()
+            ent_ret, reason = entitlement.enable()
             cfg.status()  # Update the status cache
 
-            if not ent_ret:
-                can_enable, reason = entitlement.can_enable()
-
-                if (
-                    not can_enable
-                    and reason is not None
-                    and reason.reason == ua_status.CanEnableFailureReason.IS_BETA
-                ):
-                    # if we failed because ent is in beta and there was no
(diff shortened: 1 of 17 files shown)


uaclient/cli.py
```diff
@@ -547,9 +547,7 @@ def action_enable(args, cfg, **kwargs):
             cfg.status()  # Update the status cache
 
             if not ent_ret:
-                can_enable, reason = entitlement.can_enable(
-                    silent=True, allow_disable=False
-                )
+                can_enable, reason = entitlement.can_enable()
```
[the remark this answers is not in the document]
orndorffgrant: Good catch thank you!

orndorffgrant:
Thanks for taking a look @lucasmoura !

I fixed the bug you noticed in the `refactor: remove side effects from can_enable` commit. And I added one more small refactor commit to directly address bryce's SRU comment from 27.1

uaclient/entitlements/base.py
```diff
@@ -145,45 +146,55 @@ def valid_service(self):
...
+            elif fail.reason == CanEnableFailureReason.INCOMPATIBLE_SERVICE:
+                # this is the only reason that won't necessarily stop us from
+                # enabling
+                handle_incompat_ret = self.handle_incompatible_services()
+                if not handle_incompat_ret:
+                    return False, fail
+            else:
+                # every other reason means we can't continue
+                if fail.message is not None:
+                    print(fail.message)
```
lucasmoura, line 178: Since we are now returning the `CanEnableFailure` here, I think we can move the `print` logic into the `cli` `action_enable` function

orndorffgrant:
@lucasmoura I believe I've addressed all of your comments and added a fix for the broken integration test. Let me know what you think

uaclient/entitlements/tests/test_base.py
```diff
@@ -371,24 +311,71 @@ def test_can_enable_when_incompatible_service_found(
             with mock.patch.object(
                 ent, "ENTITLEMENT_CLASS_BY_NAME", {"test": m_entitlement_cls}
             ):
-                ret, reason = base_ent.can_enable(allow_disable=allow_disable)
```
lucasmoura, line 374: I know this is a `can_enable` test, but since we moved the logic of handling incompatible services to enable, we should have an test there that is similar to this ones, were we verify that enable doesn't fail if allow disabling the incompatible service

orndorffgrant: @lucasmoura The following test does ensure that `handle_incompatible_services` is called when `can_enable` returns `INCOMPATIBLE_SERVICE`. It hard-codes `handle_incompatible_services` to return `False` though.

Do you think we should add more? We could add a test where it calls `handle_incompatible_services`, which returns `True`, and then we verify that it proceeds to call `_perform_enable`?

lucasmoura: I think we could add one directly tied to `enable`, where we call enable with an incompatible service on and see if everything is happening according to plan

orndorffgrant: Good idea I can add that

orndorffgrant: @lucasmoura I added `test_enable_when_incompatible_service_found` which is basically the same as the old `test_can_enable_when_incompatible_service_found` but calls `enable` instead of `can_enable`

orndorffgrant:
I believe all the integration test failures are from flaky tests

uaclient/entitlements/tests/test_base.py
```diff
@@ -88,6 +89,7 @@ def factory(
             applicability_status=applicability_status,
             application_status=application_status,
             allow_beta=allow_beta,
+            enable=enable,
```
lucasmoura, line 92: Why do we need this variable for the new test ?

orndorffgrant: @lucasmoura This variable is used as the return value of `_perform_enable` for `ConcreteTestEntitlement`.

The new/modified test calls `entitlement.enable()` with the expectation that it reaches the point of calling `self._perform_enable()` after it disables the incompatible services.

Since we expect the code to call `_perform_enable`, we should set the return value for `_perform_enable` in the concrete test entitlement via this argument. I'll add a comment where it is used.

uaclient/entitlements/tests/test_base.py
```diff
@@ -352,6 +341,7 @@ def test_can_enable_when_incompatible_service_found(
         m_is_config_value_true.return_value = block_disable_on_enable
         base_ent = concrete_entitlement_factory(
             entitled=True,
+            enable=True,
```
orndorffgrant, line 344: The `_perform_enable` return value is set here for this test @lucasmoura

uaclient/entitlements/tests/test_base.py
lucasmoura: My mistake here, I missed the `_perform_enable` override

orndorffgrant:
just pushed to fix merge conflicts

lucasmoura (approved):

lucasmoura:
CI failure is unrelated to this change

PR #1669, opened 2021-06-08
config: only allow some fields to be set by envvar (SC-111)

orndorffgrant:
## Proposed Commit Message

> config: only allow some fields to be set by envvar

## Test Steps

Verify `UA_CONTRACT_URL=https://contracts.staging.canonical.com ua attach STAGING_TOKEN` doesn't work
Verify `LOG_LEVEL=debug ua status` does work

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -51,6 +51,13 @@ MERGE_ID_KEY_MAP = {
 }
 UNSET_SETTINGS_OVERRIDE_KEY = "_unset"
 
+CONFIG_FIELD_ENVVAR_ALLOWLIST = [
+    "data_dir",
+    "log_file",
+    "log_level",
+    "security_url",
+]
+
 
 # A data path is a filename, and an attribute ("private") indicating whether it
 # should only be readable by root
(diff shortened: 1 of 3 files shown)


uaclient/config.py
```diff
@@ -672,8 +679,11 @@ def parse_config(config_path=None):
     for key, value in os.environ.items():
         key = key.lower()
         if key.startswith("ua_"):
-            if "ua_features_" in key:
-                key = key[12:]  # String leading UA_FEATURES_
+            # Strip leading UA_
+            field_name = key[3:]
+            if field_name.startswith("features_"):
```
orndorffgrant, line 684: I left in the `ua_features_` support. Let me know if you think we should take this out

blackboxsw: +1 on this. I think supporting features more often than not is a one-time operation during enable of something "fips" on clouds or potentially enabling a beta service.

blackboxsw (approved):
LGTM. concise straightforward. thanks.

PR #1671, opened 2021-06-09
Allow setting proxy for web requests (SC-105)

lucasmoura:
## Proposed Commit Message
Allow setting proxy for web requests

During some uaclient operations, we perform requests to communicate with the contract backend. If an user wants uaclient to respect a proxy configuration, we need to allow proxy configuration for these types of requests.

## Desired commit type::

## Checklist:

### DIFF
--- a/features/proxy_config.feature
+++ b/features/proxy_config.feature
@@ -1,0 +1,28 @@
+@uses.config.contract_token
+Feature: Attach command when proxy is configured
+
+    @series.lts
+    Scenario Outline: Attach command when proxy is configured
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I launch a `focal` `proxy` machine
+        And I run `apt install squid -y` `with sudo` on the `proxy` machine
+        And I add this text on `/etc/squid/squid.conf` on `proxy` above `http_access deny all`:
+            """
+            acl all src 0.0.0.0\/0\nhttp_access allow all
+            """
+        And I run `systemctl restart squid.service` `with sudo` on the `proxy` machine
+        And I configure uaclient `http` proxy to use `proxy` machine
+        And I configure uaclient `https` proxy to use `proxy` machine
+        And I verify `/var/log/squid/access.log` is empty on `proxy` machine
+        And I attach `contract_token` with sudo
+        And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
+        Then stdout matches regexp:
+        """
+        .*CONNECT contracts.canonical.com.*
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -38,14 +38,15 @@ def given_a_machine(context, series):
         )
         return
     if series in context.reuse_container:
+        context.instances = {}
         context.container_name = context.reuse_container[series]
-        context.instance = context.config.cloud_api.get_instance(
+        context.instances["uaclient"] = context.config.cloud_api.get_instance(
             context.container_name
         )
         if "pro" in context.config.machine_type:
-            context.instance = context.config.cloud_api.get_instance(
-                context.container_name
-            )
+            context.instances[
+                "uaclient"
+            ] = context.config.cloud_api.get_instance(context.container_name)
         return
 
     if series not in context.series_image_name:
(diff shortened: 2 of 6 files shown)


features/steps/steps.py
```diff
@@ -119,7 +197,12 @@ def when_i_retry_run_command(context, command, user_spec, exit_codes):
 
 @when("I run `{command}` {user_spec}")
 def when_i_run_command(
-    context, command, user_spec, verify_return=True, stdin=None
+    context,
+    command,
+    user_spec,
+    verify_return=True,
+    stdin=None,
+    instance_type="uaclient",
```
orndorffgrant, line 205: I think `instance_name` might be a slightly better parameter name here but :shrug:

features/proxy_config.feature
```diff
@@ -0,0 +1,28 @@
...
+    @series.lts
+    Scenario Outline: Attach command when proxy is configured
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I launch a `focal` `proxy` machine
+        And I run `apt install squid -y` `with sudo` on the `proxy` machine
+        And I add this text on `/etc/squid/squid.conf` on `proxy` above `http_access deny all`:
+            """
+            acl all src 0.0.0.0\/0\nhttp_access allow all
+            """
+        And I run `systemctl restart squid.service` `with sudo` on the `proxy` machine
```
orndorffgrant, line 13: We might want all of this as its own `When I launch a squid proxy` command if we think we'll have additional Scenarios with proxies. Though I suppose we can probably add most of the proxy integration tests to the end of this scenario.

features/steps/steps.py
```diff
@@ -64,36 +64,114 @@ def given_a_machine(context, series):
...
+    proxy_type = "http"
+    if "https" in proxy_cfg:
+        proxy_type = "https"
+
+    proxy_cfg_entry = "{}: {}://{}:{}".format(
+        proxy_cfg, proxy_type, proxy_ip, port
+    )
+    context.text = proxy_cfg_entry
+
+    when_i_append_to_uaclient_config(context)
```
orndorffgrant, line 164: The phrase for this command doesn't totally capture what it does. I think the phrase should explicitly say that we're configuring a proxy here. Just a slight change like this might be a little clearer:
```
When I configure uaclient `http` proxy to use `proxy` machine
```
The only real difference is that "proxy" becomes part of the phrase rather than one of the arguments, but I think it does help a bit

orndorffgrant (commented):
This looks really sweet, and will be useful in testing all of the proxy features :fire: 

I like that it spins up a proxy as part of the test (as opposed to us having a piece of test infrastructure managed outside the codebase).

I haven't actually run the code yet but left a couple small comments

uaclient/cli.py
```diff
@@ -1038,6 +1038,15 @@ def main(sys_argv=None):
...
     args = parser.parse_args(args=cli_arguments)
     cfg = config.UAConfig()
+
+    http_proxy = cfg.cfg.get("http_proxy")
+    https_proxy = cfg.cfg.get("https_proxy")
+
+    if http_proxy or https_proxy:
+        util.configure_web_proxy(
+            http_proxy=http_proxy, https_proxy=https_proxy
+        )
```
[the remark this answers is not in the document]
orndorffgrant: Also, if we add these as properties on UAConfig, then we could remove the lines above and this could look like:
```suggestion
    util.configure_web_proxy(
        http_proxy=cfg.http_proxy, https_proxy=cfg.https_proxy
    )
```

features/proxy_config.feature
lucasmoura: I don't think this is a bad idea at all. At my idea here was to make this generic enough that we could use it for any other case that we need a separate machine for testing.

Even though right now we could create the `I launch a squid proxy` step, I think there are still other test scenarios we may want to exercise in the future, such as toying with firewall rules. Because of that, I am inclined for now to keep those steps as is

orndorffgrant: Totally agree with the utility of launching separate machines in general. This works for me :+1:

orndorffgrant (approved):
Looks good to me as long as jenkins is happy!

orndorffgrant:
The jenkins failures are known intermittent failures unrelated to these changes, so this is good to go from my perspective :+1:

features/steps/steps.py
```diff
@@ -64,36 +64,116 @@ def given_a_machine(context, series):
         CONTAINER_PREFIX + pr_prefix + vm_prefix + series + date_prefix
     )
 
-    context.instance = context.config.cloud_manager.launch(
-        series=series,
-        instance_name=instance_name,
-        image_name=context.series_image_name[series],
-    )
+    context.instances = {
```
blackboxsw, line 67: Given that you are changing context.instance -> context.instances. I think you need to minimally handle the setup of reuse_container which stores its container detection  on `context.instance`. I bet that needs to now be context.instance['uaclient'] right?
 Here are some other cases that need handling too I think
```bash
$ fgrep -r context\.instance features/ | grep -v instances
features/environment.py:                    result = context.instance.execute(
features/environment.py:                result = context.instance.execute(cmd, use_sudo=True)
features/steps/steps.py:        context.instance = context.config.cloud_api.get_instance(
features/steps/steps.py:            context.instance = context.config.cloud_api.get_instance(
```

features/steps/steps.py
```diff
@@ -64,36 +64,116 @@ def given_a_machine(context, series):
...
     context.add_cleanup(cleanup_instance)
-    logging.warning("--- instance ip: {}".format(context.instance.ip))
-    print("--- instance ip: {}".format(context.instance.ip))
+    logging.warning(
+        "--- instance ip: {}".format(context.instances["uaclient"].ip)
+    )
+    print("--- instance ip: {}".format(context.instances["uaclient"].ip))
+
+
+@when("I launch a `{series}` `{instance_name}` machine")
```
blackboxsw, line 103: This is solid simple and flexible, at some point it feels like we could extend this to
`@when("I launch a `{series}` `{instance_name}` machine with {userdata}")`
And then we can just provide the #cloud-config needed in the `context.text` to feed into the VM to cut down on all the natural language for setting up files etc.

blackboxsw (changes requested):
The proxy vm launch in behave does look simple and good and I'm  +1 on the `ProxyHandler` global registration. Thanks for working this. minor fixups for the remaining `context.instance`  cases I think

features/steps/steps.py
lucasmoura: Completely agree, we should allow that flexibility in the future

features/steps/steps.py
lucasmoura: Great catch, yes those cases should all be handled

lucasmoura:
@blackboxsw updated

blackboxsw (approved):
+1 integration failures unrelated to this changeset proxy setup is super useful.

PR #1675, opened 2021-06-10
proxy: allow configuration to automatically set apt proxy (SC-88)

orndorffgrant:
## Notes

~~This is currently based on #1671 to use the proxy integration test work introduced there. As such, it should be reviewed/merged after #1671.~~ This is ready for review now.

### The questions in this section were resolved in conversation on 2020-06-11 and the decision was summarized in a comment below
Currently, this installs the apt proxy config when enabling an apt based service, or on a `ua refresh`. So the user flow looks like:
```shell
vim /etc/ubuntu-advantage/uaclient.conf # add proxy values
ua refresh # or ua enable esm-infra, but the user may have already enabled it
```

That doesn't necessarily feel super intuitive. Another thing we could do is install the proxy config file whenever our UAConfig object is instantiated - then it would be updated on any `ua` command. In addition, we could also install the file at the end of our upcoming `ua config set` subcommand, so that the user doesn't have to run an extra command after `ua config set apt_http_proxy=host:port`. 

@blackboxsw @lucasmoura @mitechie thoughts? We can also iron this out in the spec before coming back to this PR

## Todo list
- [x] decide when to actually update the apt proxy config
- [x] remove the apt config file when no apt_proxy values are set
- [x] unit tests for new functions
- [x] add integration tests
- [x] rebase after #1671 is merged

## Proposed Commit Message

>
>
>

## Desired commit type::

## Checklist:


lucasmoura:
@orndorffgrant I think instead of setting the proxy during the `refresh` command, we could set it on the `process_contract_deltas` [method](https://github.com/canonical/ubuntu-advantage-client/blob/a997de77fd5320465dbd975bd70fd8f9556f7b32/uaclient/entitlements/repo.py#L163) of the RepoEntitlement class. We would need a mechanism to verify if the apt proxy file is already there and that it contains the proxies we want, but this will ensure that we will only setup the proxy when we need it and that every repo entitlement will try to set it up

orndorffgrant:
@lucasmoura Those are good points and bring up two questions of mine:

1. If a user starts a Pro instance, and then edits `uaclient.conf` to add the apt proxy vars, when do we write the apt proxy file? ESM is already enabled, and the customer doesn't need to enable any other service. I see two (maybe three) options here:

    A. We document a specific command that the user needs to run that will process the new config and write the apt proxy. With this wip PR, that command is (admittedly confusing) `ua refresh`.

    B. We _always_ process the new config and make any system updates necessary on _any_ `ua` command. Then we can just say run `ua status` to update the proxy and maybe `status` should also print proxy status info?

    C. If we want to be way too fancy/complex, we might be able to watch the `uaclient.conf` file for updates and trigger a script.

2. The current implementation assumes that `write_file` is cheap (we can do it as much as we want), and that overwriting the apt.conf file won't negatively affect apt in any way. If those two assumptions are correct, then we don't need to check if the aptproxy file is already there or if it has old values - we'll instead just always overwrite the file with the values from `uaclient.conf`. This means we'll overwrite the file with the exact same contents frequently, but also makes our logic simpler. Are those reasonable assumptions though? (and am I missing some edge case where this wont work?)

lucasmoura:
1. I think that the consideration for PRO images make sense, but I don't know if we should have a specific solution for it in our codebase. I think that if users want to launch a PRO machine with proxy configuration, we should instruct them to use a cloud-init userdata that correctly configures the proxy. In that way, auto-attach should run with the proxies already configured. I know this assumes that costumers will contact us before trying that, but I don't see another way for us to correctly run auto-attach with the proxies already configured.

However, I am still unsure if we should allow users to configure proxy configurations through uaclient when no attach/enable operation is being done. What do you think @mitechie @blackboxsw ?

2.  That makes sense, my concern was just that maybe users could have a different apt proxy configuration than the one that was in `uaclient.conf`. However, this doesn't make much sense, so you are right, we can just write the file

orndorffgrant:
To summarize our out-of-band discussion:

1. We will introduce a new command for applying the current state of the config: `ua refresh config`
2. As part of the upcoming config sub command, we will apply changes after editing the config. So `ua config set apt_http_proxy=http://host:port` would update the apt proxy file before exiting succesfully
3. We will do just-in-time configuration of proxies, so we will update the apt proxy file before executing `enable` for any apt based entitlement.

Only (3) will be included in this PR

features/proxy_config.feature
```diff
@@ -0,0 +1,82 @@
...
+        .*CONNECT contracts.canonical.com.*
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
+
+    @wip
```
lucasmoura, line 30: I think we can remove this

orndorffgrant: Oops, yes for sure - thanks for noticing that!

uaclient/config.py
```diff
@@ -58,6 +58,8 @@ def cast(_, x):  # type: ignore
     "security_url",
 ]
 
+APT_PROXY_CONF_FILE_PATH = "/etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy"
```
blackboxsw, line 61: might want to relocate this declaration into uaclient.apt.py too.

uaclient/config.py
```diff
@@ -658,6 +668,26 @@ def help(self, name):
         response_dict["help"] = help_ent.help_info
         return response_dict
 
+    def setup_apt_proxy(self):
```
blackboxsw, line 671: While it makes it easier to reference set.apt_http_proxy ad apt_https_proxy hosting this method on UAConfig, I think it probably makes more sense to locate this function in the `uaclient.apt` module to do everything related to APT from one module.

It may add a bit of complexity to the function signature, but it keep separation of roles more clean. UAConfig should mostly be dealing with sugar/helper methods around parsing contract config (via machine-token reads/writes),  uaclient.conf settings and config notifications & state.  I think side-effects related to the service stacks (APT, livepatch, snap etc) should be housed in modules related to those services instead of growing UAClient utility functions more.

 I'm hoping it's a clean enough function signature to allow for setup_apt_proxy in apt.py to allow two simple params  http_proxy and https_proxy passed  from the config properties which can be called from entitlements.repo.py as you already have below.

What do you think?

uaclient/config.py
```diff
@@ -117,6 +119,14 @@ def http_proxy(self):
     def https_proxy(self):
         return self.cfg.get("https_proxy")
 
+    @property
+    def apt_http_proxy(self):
```
blackboxsw, line 123: might as well use typing for everything we touch

What do you think about drive-by bring in the uaclient.defaults too for contract_url and security_url?
```diff
diff --git a/uaclient/config.py b/uaclient/config.py
index 80c1fc8b..acd753ed 100644
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -10,7 +10,12 @@ import yaml
 from collections import namedtuple, OrderedDict
 
 from uaclient import status, util
-from uaclient.defaults import CONFIG_DEFAULTS, DEFAULT_CONFIG_FILE
+from uaclient.defaults import (
+    BASE_CONTRACT_URL,
+    BASE_SECURITY_URL,
+    CONFIG_DEFAULTS,
+    DEFAULT_CONFIG_FILE,
+)
 from uaclient import exceptions
 
 try:
@@ -104,27 +109,27 @@ class UAConfig:
         return []
 
     @property
-    def contract_url(self):
-        return self.cfg.get("contract_url", "https://contracts.canonical.com")
+    def contract_url(self) -> str:
+        return self.cfg.get("contract_url", BASE_CONTRACT_URL)
 
     @property
-    def security_url(self):
-        return self.cfg.get("security_url", "https://ubuntu.com/security")
+    def security_url(self) -> str:
+        return self.cfg.get("security_url", BASE_SECURITY_URL)
 
     @property
-    def http_proxy(self):
+    def http_proxy(self) -> "Optional[str]":
         return self.cfg.get("http_proxy")
 
     @property
-    def https_proxy(self):
+    def https_proxy(self) -> "Optional[str]":
         return self.cfg.get("https_proxy")
 
     @property
-    def apt_http_proxy(self):
+    def apt_http_proxy(self) -> "Optional[int]":
         return self.cfg.get("apt_http_proxy")
 
     @property
-    def apt_https_proxy(self):
+    def apt_https_proxy(self) -> "Optional[str]":
         return self.cfg.get("apt_https_proxy")
 
     def check_lock_info(self) -> "Tuple[int, str]":
```

Feel free to reject or tell me to put up a separate PR for this

uaclient/status.py
```diff
@@ -507,6 +507,22 @@ def __init__(
...
+MESSAGE_APT_PROXY_CONFIG_HEADER = """\
+/*
+ * Autogenerated by ubuntu-advantage-tools
+ * Do not edit this file directly
+ *
+ * To change what ubuntu-advantage-tools sets here, edit the apt_http_proxy
+ * and apt_https_proxy fields in /etc/ubuntu-advantage/uaclient.conf. Then
+ * run:
+ *
+ * ua refresh config
```
blackboxsw, line 519: Since we are documenting ua refresh cmd TBD, I think we should also document this directly in the header of the uaclient.conf we ship
Please also add something comparable mentioning `ua refresh config` to https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient.conf

blackboxsw (changes requested):
Looks really good @orndorffgrant couple of nits here and there. let me know what you think.

features/proxy_config.feature
```diff
@@ -26,3 +26,56 @@ Feature: Attach command when proxy is configured
            | xenial  |
            | bionic  |
            | focal   |
+
+    @series.lts
+    Scenario Outline: Apt proxy is configured before attach
```
lucasmoura, line 31: I think we can merge this test with the one for the web proxies. Since the proxy tests are creating two distinct instance, I think we can merge them to save on performance.

features/util.py
```diff
@@ -133,7 +133,12 @@ def build_debs(
     for filepath in (buildscript, SOURCE_PR_TGZ):
         print("--- Push {} -> {}:/tmp".format(filepath, instance.name))
         instance.push_file(filepath, "/tmp/" + os.path.basename(filepath))
-    instance.execute(["sudo", "bash", "/tmp/" + buildscript])
+    result = instance.execute(["sudo", "bash", "/tmp/" + buildscript])
+    if result.return_code != 0:
```
lucasmoura, line 137: We could use `result.failed` instead

ubuntu-advantage.1
```diff
@@ -114,6 +114,16 @@ The logging level used when writing to \fBlog_file\fP
...
+.TP
+.B
+\fBapt_http_proxy\fP
+If set, ua will configure apt to use the specified http proxy by writing a apt
+config file to /etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy
+.TP
+.B
+\fBapt_https_proxy\fP
+If set, ua will configure apt to use the specified https proxy by writing a apt
+config file to /etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy
```
lucasmoura, line 126: Good catch adding the variables description on this file. I missed that on my web proxy PR :facepalm:
Can you add them here as well ?

uaclient/config.py
```diff
@@ -658,6 +668,26 @@ def help(self, name):
         response_dict["help"] = help_ent.help_info
         return response_dict
 
+    def setup_apt_proxy(self):
+        apt_proxy_config = ""
```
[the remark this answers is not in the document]
orndorffgrant: Thanks for catching this, absolutely

uaclient/config.py
orndorffgrant: Definitely! I guess I just didn't think to put it in `uaclient.apt` because I haven't touched that file yet, but that's definitely the place to put this functionality.

uaclient/config.py
orndorffgrant: Good call! I can totally include that, maybe broken into a separate commit

uaclient/status.py
orndorffgrant: yeah on second thought it feels weird including a reference to the non-existent-yet command in this PR. Since I preemptively started the `ua refresh [config|contract]` pr over at #1678 I think I'll remove this reference to it and add this comment as well as a comment in uaclient.conf to #1678

features/proxy_config.feature
orndorffgrant: Good point, I'll try that out!

features/util.py
orndorffgrant: Sweet, I didn't realize that existed and didn't think to check - I will update

ubuntu-advantage.1
orndorffgrant: Sure thing!

uaclient/status.py
blackboxsw: +1 on that suggestion.

uaclient/entitlements/repo.py
```diff
@@ -92,6 +92,10 @@ def _perform_enable(self) -> bool:
         @return: True on success, False otherwise.
         @raises: UserFacingError on failure to install suggested packages
         """
+        apt.setup_apt_proxy(
```
lucasmoura, line 95: I think we could move this call into `setup_apt_config`. But this is just a nit, feel free to change it or not

uaclient/entitlements/tests/test_esm.py
```diff
@@ -184,10 +184,26 @@ def test_esm_apps_disable_apt_auth_only_is_true_on_lts(
 
 @mock.patch("uaclient.entitlements.esm.update_ua_messages")
 class TestESMInfraEntitlementEnable:
+    @pytest.mark.parametrize(
+        "esm_cls", [ESMAppsEntitlement, ESMInfraEntitlement]
+    )
+    @mock.patch("uaclient.apt.setup_apt_proxy")
```
lucasmoura, line 190: Just another nit. We could add this patch into the class definition, like `update_ua_messages` just to avoid repetition.

uaclient/entitlements/tests/test_fips.py
```diff
@@ -127,7 +127,10 @@ def test_can_enable_true_on_entitlement_inactive(
 
 
 class TestFIPSEntitlementEnable:
-    def test_enable_configures_apt_sources_and_auth_files(self, entitlement):
+    @mock.patch("uaclient.apt.setup_apt_proxy")
```
lucasmoura, line 130: Same mock consideration here

lucasmoura (commented):
Just some nits here. Let me know if you think they make sense

orndorffgrant:
I believe I've addressed all comments. I'm letting jenkins test the integration test updates though :crossed_fingers:

lucasmoura:
@orndorffgrant it seems that the proxy tests `Attach command when proxy is configured ` is not working as expected :(

uaclient/apt.py
```diff
@@ -391,3 +392,37 @@ def clean_apt_files(*, _entitlements=None):
...
+        )
+    if apt_proxy_config != "":
+        apt_proxy_config = (
+            status.MESSAGE_APT_PROXY_CONFIG_HEADER + apt_proxy_config
+        )
+
+    if apt_proxy_config == "":
+        util.remove_file(APT_PROXY_CONF_FILE)
+    else:
+        util.write_file(APT_PROXY_CONF_FILE, apt_proxy_config)
```
blackboxsw, line 428: If we are about to write apt proxy, do we think it's worth inspecting whether apt already has a proxy config that we are going to collide with?
we could check         ["apt-config", "shell", "key", "Acquire::http::Proxy" ]                and if non-empty we know we may have a problem worth adding a prompt for.

uaclient/config.py
```diff
@@ -102,21 +107,29 @@ def accounts(self):
...
         return self.cfg.get("http_proxy")
 
     @property
-    def https_proxy(self):
+    def https_proxy(self) -> "Optional[str]":
         return self.cfg.get("https_proxy")
 
+    @property
+    def apt_http_proxy(self) -> "Optional[str]":
+        return self.cfg.get("apt_http_proxy")
```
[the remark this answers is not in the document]
orndorffgrant: Right, I'll do that in this PR

uaclient/apt.py
```diff
@@ -391,3 +392,37 @@ def clean_apt_files(*, _entitlements=None):
...
+        apt_proxy_config += status.MESSAGE_APT_PROXY_HTTPS.format(
+            proxy_url=https_proxy
+        )
+    if apt_proxy_config != "":
+        apt_proxy_config = (
+            status.MESSAGE_APT_PROXY_CONFIG_HEADER + apt_proxy_config
+        )
+
+    if apt_proxy_config == "":
+        util.remove_file(APT_PROXY_CONF_FILE)
```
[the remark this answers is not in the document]
orndorffgrant: Yep my goal for this function was that regardless of whether an apt proxy was set/unset/not changed, you can run this function and we'll reach the correct state - even if its just overwriting the existing file with the exact same contents, or not doing anything

uaclient/apt.py
orndorffgrant: @blackboxsw I think we absolutely should do this, but do you mind if I break this piece into another PR to keep this one from growing much more?

orndorffgrant:
@lucasmoura Thanks for catching the test failure. That was due to some sloppiness on my part friday afternoon :P

uaclient/apt.py
blackboxsw: > @blackboxsw I think we absolutely should do this, but do you mind if I break this piece into another PR to keep this one from growing much more?

Absolutely. Separate PR is the way to go. I figure for any of these nits it could have been separated easily. None are blocking for this PR.

ubuntu-advantage.1
```diff
@@ -114,6 +114,24 @@ The logging level used when writing to \fBlog_file\fP
 .B
 \fBlog_file\fP
 The log file for the Ubuntu Advantage client
+.TP
+.B
+\fBhttp_proxy\fP
```
blackboxsw, line 119: In docs, we might have to define a `ua_config` section with these keys under it maybe? not quite sure, but leave the decision to you.

blackboxsw (approved):
Unrelated failure in Ubuntu PRO test [livepatch showing as disabled](https://jenkins.canonical.com/server-team-jenkins-private/blue/organizations/jenkins/uaclient-ci/detail/PR-1675/15/pipeline/102#step-127-log-176)

I'm closing out my review as approve to I don't block it. If you carry any agreeable work items into a separate PR that works for me.

orndorffgrant:
@lucasmoura I've just pushed an addition to our manfile as chad suggested above, so I believe the previous jenkins results are still valid (https://jenkins.canonical.com/server-team-jenkins-private/job/uaclient-ci/job/PR-1675/15/display/redirect).

That means this PR should be good to go if you are good with the changes

lucasmoura (approved):
@orndorffgrant approved. I think we still have that @blackboxsw doc consideration to approach. but we can do that on another PR

PR #1678, opened 2021-06-15
refresh: add target argument to control what to refresh (SC-123)

orndorffgrant:
## Notes
This introduces `ua refresh config` but that currently doesn't do anything. So we should wait until either the apt proxy PR or a future snap/livepatch proxy PR is merged before finishing and then merging this PR.

### TODO
- [x] add integration tests
- [x] wait for features that actually require config processing
- [x] actually process config
- [x] use in apt/snap/livepatch proxy integration tests
- [x] Add "run ua refresh config" to apt proxy conf file
- [x] Add "run ua refresh config" to uaclient.conf

## Proposed Commit Message

> refresh: add target to refresh command

> ua refresh now accepts an optional arg, which can be config or contract.
> Depending on the value, ua refresh will either process any config
> changes or re-fetch the contract from the server. If the arg is not
> provided, then ua refresh will do both.

> ua refresh config will also print a warning if there is no proxy set in
> uaclient.conf but there is a snap or livepatch proxy set. We can't
> actually unset a snap/livepatch proxy in this scenario because this may
> have been intended by the user. Instead of unsetting, we provide
> instructions to either remove the snap/livepatch proxy setting, or add
> the proxy config to uaclient.conf.

## Test Steps

set some proxy variables in uaclient.conf, then run `ua refresh config` or `ua refresh`
remove those proxy variables from uaclient.conf then run refresh again.

## Desired commit type::

## Checklist:


features/attached_commands.feature
```diff
@@ -24,6 +35,37 @@ Feature: Command behaviour when attached to an UA subscription
...
+            with ubuntu-advantage-tools. If you wish to remove them, run the following
+            commands:
+
+                snap unset system proxy.https
+                canonical-livepatch config http-proxy=
+
+            Or, if you want ua to stop showing this message, you can add the proxy
+            information to /etc/ubuntu-advantage/uaclient.conf
+
+            Successfully processed your ua configuration.
```
orndorffgrant, line 61: @mitechie @blackboxsw @lucasmoura 

This is my draft of what we could display in the scenario when
1. there is a snap/livepatch proxy set
2. uaclient.conf does not have any proxy information
3. the user runs `ua refresh config`

I'm open to any changes.

mitechie: I'm working to find a way to update the spec for this but I think I'm -1 on this approach. If there's nothing in the config file then refresh config would no-op and we'd not offer to clear proxies or the like. 

We did agree that when use ua config set <x> we'd note any conflicts we'd be overriding, but I think that's all. In that light, during unset we'd simply remove the proxy values in place. 

With the refresh of the config the user hasn't said if they mean "no value" or an explicit "empty value" and so I think that's just a no-op.

mitechie: Refreshing config:

No proxy settings set in config, however, there is proxy config on the livepatch service. See HTTP://... for more information on proxy config.

mitechie: or should that be less about livepatch and more about snaps and pointing to snap proxy documentation vs our own

uaclient/config.py
```diff
@@ -665,6 +664,56 @@ def help(self, name):
         response_dict["help"] = help_ent.help_info
         return response_dict
 
+    def process_config(self, assume_yes: bool = False):
+        apt.setup_apt_proxy_with_prompts(
+            self.apt_http_proxy, self.apt_https_proxy, assume_yes=assume_yes
+        )
+
+        suggested_proxy_clearing_cmds = []
+        if snap.is_installed():
```
blackboxsw, line 673: This test is suffecient for presence of snap enabling us to set snap proxy configuration. It is not sufficient for canonical-livepatch though. canonical-livepatch snap can be installed on a non-LTS release (e.g. groovy) and will block any attempt to run any command:
```bash
$ canonical-livepatch status
2021/06/28 12:03:28 error executing status: Livepatchd error: The platform Ubuntu 20.10 is not supported. exiting.
$ echo $?
1
```


For Livepatch, we actually need to check `LivepatchEntitlement.application_status()` to determine enabled or not, or read current service status from the cache via something like:
```python
    livepatch_status = "disabled"
    for service in UAConfig().read_cache("status-cache").get("services", []):
         if service["name"] == "livepatch":
               livepatch_status = service["status"]
               break
```

Only then do we know that the livepatch snap will accept our proxy config attempt.

uaclient/snap.py
```diff
@@ -0,0 +1,87 @@
...
+
+SNAP_CMD = "/usr/bin/snap"
+SNAP_INSTALL_RETRIES = [0.5, 1.0, 5.0]
+HTTP_PROXY_OPTION = "proxy.http"
+HTTPS_PROXY_OPTION = "proxy.https"
+
+
+def is_installed() -> bool:
+    """Returns whether or not snap is installed"""
+    return "snapd" in apt.get_installed_packages()
```
blackboxsw, line 13: While this test is "valid" can we more simply check util.which(SNAP_CMD)  which we do when we "enable" livepatch anyway?

blackboxsw (changes requested):
Some driveby comments on this as we were just discussing some of the impacts of trying to call canonical-livepatch commands on unsupported Ubuntu releases. I think we need a common mechanism for both `ua config` and `ua refresh config` to check whether we actually should even attempt to apply proxy settings for livepatch.

orndorffgrant:
Closing in favor of #1694

PR #1679, opened 2021-06-15
readme: add instruction to generate MOTD messages

lucasmoura:
## Proposed Commit Message
readme: add instruction to generate MOTD messages

Add instructions to force MOTD messages to be updated in the system

## Test Steps
Launch a xenial container and follow the instructions in this guide.
Verify if the MOTD message possess ESM references on it

## Desired commit type::

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,30 @@ a daily ubuntu-advantage-tools image.
 UACLIENT_BEHAVE_REUSE_IMAGE=your-custom-image-id tox -e behave-awspro-20.04
 ```
 
+### MOTD Messages
+
+Since ubuntu-advantage-tools is responsible for enabling ESM services, we advertise them on different
+applications thorough the system, such as MOTD and apt commands like upgrade.
+
+To verify that the MOTD message is advertising the ESM packages, ensure that we have ESM source list
+files in the system. If that is the case, please run the following commands to update the state of
+MOTD and display the message:
+
+```sh
+# Make sure ubuntu-advantage-tools version >= 27.0
+ua version
+# Make apt aware of the ESM source files
+sudo apt update
+# Generates ubuntu-advantage-tools messages that should be delivered to MOTD
+# This script is trigged by a systemd timer once a day. To test it, we need to
+# enforce that it was already executed.
+sudo systemctl start ua-messaging.service
+# Force updating MOTD messages related to update-notifier
+/usr/lib/update-notifier/update-motd-updates-available --force
+# Update MOTD and display the message
+run-parts /etc/update-motd.d/
+```
+
 ## Building
 
 Creating ubuntu-advantage-tools and ubuntu-advantage-pro is created from the


README.md
```diff
@@ -344,6 +344,28 @@ a daily ubuntu-advantage-tools image.
...
+Since ubuntu-advantage-tools is responsible for enabling ESM services, we advertise them on different
+applications thorough the system, such as MOTD and apt commands like upgrade.
+
+To verify that the MOTD message is advertising the ESM packages, ensure that we have ESM source list
+files in the system. If that is the case, please run the following commands to update the state of
+MOTD and display the message:
+
+```sh
+# Make apt aware of the ESM source files
+sudo apt update
```
mitechie, line 358: should we do a check for ua client version?

mitechie (approved):

README.md
lucasmoura: I think we should assume we are running the latest version here, right ? Since this goes under the test section. But if need,  I can add that command, no problem at all

mitechie:
I was just thinking of the know test cases we hit today where you did
first need to upgrade the ua client before beginning these steps.
That's a lxd image and such but why maybe a --version check vs a
forced install command is ok?

lucasmoura:
That's fair, currently we are asking in that command for updating uaclient. However, I think you are right, there is no problem on just making that check there

README.md
```diff
@@ -344,6 +344,30 @@ a daily ubuntu-advantage-tools image.
...
+# This tool allows us to more easily update the MOTD message in the sytem
+sudo apt install update-motd
+# Generates ubuntu-advantage-tools messages that should be delivered to MOTD
+python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py
+# Force updating MOTD messages related to update-notifier
+/usr/lib/update-notifier/update-motd-updates-available --force
+# Update MOTD and display the message
+update-motd
```
orndorffgrant, line 368: I think we can safely recommend running `run-parts /etc/update-motd.d/` instead of installing update-motd?

README.md
```diff
@@ -344,6 +344,30 @@ a daily ubuntu-advantage-tools image.
...
+
+```sh
+# Make sure ubuntu-advantage-tools version >= 27.0
+ua version
+# Make apt aware of the ESM source files
+sudo apt update
+# This tool allows us to more easily update the MOTD message in the sytem
+sudo apt install update-motd
+# Generates ubuntu-advantage-tools messages that should be delivered to MOTD
+python3 /usr/lib/ubuntu-advantage/ua_update_messaging.py
```
orndorffgrant, line 364: Idea: we could recommend running the systemd service instead of directly running the python script? 
```suggestion
systemctl start ua-messaging.service
```

Related: We could also mention in the paragraph above that normally these messages get updated with a systemd timer periodically, which is why they might not show up immediately.

README.md
lucasmoura: Since we are assuming this steps should be run as a one tme test, you are right.
I was using `update-motd` because I was iterating on different message scenarios and I think it is easier to remember
I will update the docs

lucasmoura:
Thanks for the suggestions here @orndorffgrant. I have updated the doc

orndorffgrant (approved):
Just ran through the steps one more time and it worked for me. Thanks for documenting this!

PR #1681, opened 2021-06-16
apt-hook: fix invalid 20apt-esm-hook.conf syntax.

xnox:
apt.conf variables must end with ';'

Fixes: ea85ef1 ("apt-hook: new json hook for security update counts")
Fixes: 8257679 ("apt-hook: print static messages on apt upgrade/dist-upgrade (#1540)")

LP: #1930741

https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1930741

## Proposed Commit Message

>
>
>

## Desired commit type::

## Checklist:


orndorffgrant:
Thanks for the fix @xnox! This looks good to me, let's just let jenkins run and then we'll merge

blackboxsw (approved):
+1 thanks for the quick fix here.

blackboxsw:
Well jenkins got all borked yesterday and were waiting on a completed run which hasn't come. We've cleaned up Jenkins, but these CI failures are completely unrelated to a near trivial changeset. Merging as is (given that it's also already released this isn't a big dealio)

PR #1685, opened 2021-06-18
proxy: allow configuring snap and livepatch (SC-87)

lucasmoura:
## Proposed Commit Message
proxy: allow configuring snap and livepatch

Allow configuring both snap and livepatch proxies during the enable operation of livepatch

PS: There are still some points on this PR that we could improve. Currently, it seems that
requests for livepatch are arriving with a delay into the squid log, that's why I am running a sleep command
in the test. I can confirm that only running `canonical-livepatch refresh` does not remove that delay.

Furthermore, I am not seeing any log entries we I run 'ua detach`. Since we hit the contract backend on that
operation, I was expecting a log entry for it.

## Test Steps
Run the new vm test

## Desired commit type::

## Checklist:

### DIFF
--- a/features/proxy_config.feature
+++ b/features/proxy_config.feature
@@ -61,3 +61,43 @@ Feature: Proxy configuration
            | xenial  |
            | bionic  |
            | focal   |
+
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attach command when proxy is configured
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I launch a `focal` `proxy` machine
+        And I run `apt install squid -y` `with sudo` on the `proxy` machine
+        And I add this text on `/etc/squid/squid.conf` on `proxy` above `http_access deny all`:
+            """
+            acl all src 0.0.0.0\/0\nhttp_access allow all
+            """
+        And I run `systemctl restart squid.service` `with sudo` on the `proxy` machine
+        And I configure uaclient `http` proxy to use `proxy` machine
+        And I configure uaclient `https` proxy to use `proxy` machine
+        And I verify `/var/log/squid/access.log` is empty on `proxy` machine
+        And I attach `contract_token` with sudo
+        And I run `canonical-livepatch config check-interval=0` with sudo
+        And I run `canonical-livepatch refresh` with sudo
+        And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
+        Then stdout matches regexp:
+        """
+        .*CONNECT contracts.canonical.com.*
+        """
+        When I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
+        Then stdout matches regexp:
+        """
+        .*CONNECT api.snapcraft.io:443.*
+        """
+        When I run `sleep 120` as non-root
+        And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
+        Then stdout matches regexp:
+        """
+        .*CONNECT livepatch.canonical.com:443.*
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -107,7 +107,7 @@ def given_a_machine(context, series):
 def launch_machine(context, series, instance_name):
     now = datetime.datetime.now()
     date_prefix = now.strftime("-%s%f")
-    name = CONTAINER_PREFIX + series + date_prefix + instance_name
+    name = CONTAINER_PREFIX + series + date_prefix + "-" + instance_name
 
     context.instances[instance_name] = context.config.cloud_manager.launch(
         series=series, instance_name=name
@@ -157,9 +157,8 @@ def when_i_configure_uaclient_using_proxy_machine(
     # But I don't think we need this at this moment.
     port = "3128"
 
+    # we are not configuring a full https proxy for the tests
     proxy_type = "http"
-    if "https" in proxy_cfg:
-        proxy_type = "https"
 
     proxy_cfg_value = "{}://{}:{}".format(proxy_type, proxy_ip, port)
 
--- a/uaclient/entitlements/tests/test_livepatch.py
+++ b/uaclient/entitlements/tests/test_livepatch.py
@@ -22,6 +22,8 @@ from uaclient import exceptions
 from uaclient.entitlements.livepatch import (
     LivepatchEntitlement,
     process_config_directives,
+    configure_snap_proxy,
+    configure_livepatch_proxy,
 )
 from uaclient.entitlements.tests.conftest import machine_token
 from uaclient import status
@@ -65,6 +67,98 @@ def livepatch_entitlement_factory(entitlement_factory):
 @pytest.fixture
 def entitlement(livepatch_entitlement_factory):
     return livepatch_entitlement_factory()
+
+
+class TestConfigureSnapProxy:
+    @pytest.mark.parametrize(
+        "http_proxy,https_proxy,snap_retries",
+        (
+            ("http_proxy", "https_proxy", [1, 2]),
+            ("http_proxy", "", None),
+            ("", "https_proxy", [1, 2]),
+            ("http_proxy", None, [1, 2]),
+            (None, "https_proxy", None),
+            (None, None, [1, 2]),
+        ),
+    )
+    @mock.patch("uaclient.util.subp")
+    def test_configure_snap_proxy(
+        self, m_subp, http_proxy, https_proxy, snap_retries
+    ):
+        configure_snap_proxy(http_proxy, https_proxy, snap_retries)
+        expected_calls = []
+        if http_proxy:
+            expected_calls.append(
+                mock.call(
+                    [
+                        "snap",
+                        "set",
+                        "system",
+                        "proxy.http={}".format(http_proxy),
+                    ],
+                    retry_sleeps=snap_retries,
+                )
+            )
+
+        if https_proxy:
+            expected_calls.append(
+                mock.call(
+                    [
+                        "snap",
+                        "set",
+                        "system",
+                        "proxy.https={}".format(https_proxy),
+                    ],
+                    retry_sleeps=snap_retries,
+                )
+            )
+
+        assert m_subp.call_args_list == expected_calls
+
+
+class TestConfigureLivepatchProxy:
+    @pytest.mark.parametrize(
+        "http_proxy,https_proxy,livepatch_retries",
+        (
+            ("http_proxy", "https_proxy", [1, 2]),
+            ("http_proxy", "", None),
+            ("", "https_proxy", [1, 2]),
+            ("http_proxy", None, [1, 2]),
+            (None, "https_proxy", None),
+            (None, None, [1, 2]),
+        ),
+    )
+    @mock.patch("uaclient.util.subp")
+    def test_configure_livepatch_proxy(
+        self, m_subp, http_proxy, https_proxy, livepatch_retries
+    ):
+        configure_livepatch_proxy(http_proxy, https_proxy, livepatch_retries)
+        expected_calls = []
+        if http_proxy:
+            expected_calls.append(
+                mock.call(
+                    [
+                        "canonical-livepatch",
+                        "config",
+                        "http-proxy={}".format(http_proxy),
+                    ],
+                    retry_sleeps=livepatch_retries,
+                )
+            )
+
+        if https_proxy:
+            expected_calls.append(
+                mock.call(
+                    [
+                        "canonical-livepatch",
+                        "config",
+                        "https-proxy={}".format(https_proxy),
+                    ],
+                    retry_sleeps=livepatch_retries,
+                )
+            )
+
+        assert m_subp.call_args_list == expected_calls
 
 
 class TestLivepatchContractStatus:
@@ -417,6 +511,8 @@ class TestLivepatchProcessContractDeltas:
         assert setup_calls == m_setup_livepatch_config.call_args_list
 
 
+@mock.patch("uaclient.entitlements.livepatch.configure_snap_proxy")
+@mock.patch("uaclient.entitlements.livepatch.configure_livepatch_proxy")
 class TestLivepatchEntitlementEnable:
 
     mocks_apt_update = [
@@ -478,6 +574,8 @@ class TestLivepatchEntitlementEnable:
         m_run_apt,
         m_subp,
         _m_get_platform_info,
+        m_livepatch_proxy,
+        m_snap_proxy,
         capsys,
         caplog_text,
         entitlement,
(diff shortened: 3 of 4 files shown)


uaclient/util.py
```diff
@@ -497,6 +497,54 @@ def configure_web_proxy(
...
+    """
+    Configure snap to use http and https proxies.
+
+    :param http_proxy: http proxy to be used by snap. If None, it will
+                       not be configured
+    :param https_proxy: https proxy to be used by snap. If None, it will
+                        not be configured
+    """
+    if http_proxy:
+        subp(["snap", "set", "system", "proxy.http={}".format(http_proxy)])
```
blackboxsw, line 512: In uaclient/entitlements.livepatch we use retry_sleeps for any snap cmdline interaction         `retry_sleeps=SNAP_INSTALL_RETRIES`. 

Maybe we add a retry_sleeps param here and pass that into subp. (I'd suggest that we use SNAP_INSTALL_RETRIES value here)
The reason I suggest this is a strong suspicion that we are going to hit race conditions on Azure and AWS waiting for for snap seeding to complete which prompted us to add the [snap install livepatch retries](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/entitlements/livepatch.py#L114-L118)

uaclient/util.py
```diff
@@ -497,6 +497,54 @@ def configure_web_proxy(
...
+
+    if https_proxy:
+        subp(["snap", "set", "system", "proxy.https={}".format(https_proxy)])
```
blackboxsw, line 515: Same retry logic request

uaclient/util.py
```diff
@@ -497,6 +497,54 @@ def configure_web_proxy(
...
+                       not be configured
+    :param https_proxy: https proxy to be used by livepatch. If None, it will
+                        not be configured
+    """
+    if http_proxy:
+        subp(
+            [
+                "canonical-livepatch",
+                "config",
+                "http-proxy={}".format(http_proxy),
```
blackboxsw, line 534: I'm not as worried about retries here, because we call this only after we've already retried and waited to install livepatch. And we haven't yet needed retries on uaclient.entitlements.livepatch.process_config_directives in the past so we are unlikely to hit a race calling other livepatch config commands.

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -557,6 +565,8 @@ def test_enable_installs_only_livepatch_snap_when_absent_but_snapd_present(
             mock.call("/usr/bin/snap"),
         ]
         assert expected_calls == m_which.call_args_list
+        assert m_snap_proxy.call_count == 1
```
blackboxsw, line 568: Do we have unittest explicit coverage of the commands passed to subp?
If not can we add assertions of m_snap_proxy.call_args_list  and m_livepatch_proxy

features/proxy_config.feature
```diff
@@ -26,3 +26,40 @@ Feature: Attach command when proxy is configured
...
+            """
+            acl all src 0.0.0.0\/0\nhttp_access allow all
+            """
+        And I run `systemctl restart squid.service` `with sudo` on the `proxy` machine
+        And I configure uaclient `http` proxy to use `proxy` machine
+        And I configure uaclient `https` proxy to use `proxy` machine
+        And I verify `/var/log/squid/access.log` is empty on `proxy` machine
+        And I attach `contract_token` with sudo
+        And I run `canonical-livepatch refresh` with sudo
+        And I run `sleep 120` as non-root
```
blackboxsw, line 46: I wonder if we can set something like this to avoid the 2 minute timeout
```bash
  check-interval: -1  # manually refresh and then run `sudo canonical-livepatch refresh`?
  
```

I've asked in ~CS channel to see if they have a suggestion

blackboxsw (changes requested):
Thank you @lucasmoura.

uaclient/util.py
```diff
@@ -497,6 +497,54 @@ def configure_web_proxy(
         request.install_opener(opener)
 
 
+def configure_snap_proxy(
```
blackboxsw, line 500: This probably should live in uaclient.entitlements.livepatch instead of uaclient.util because it really only applies to livepatch setup and config.

uaclient/util.py
```diff
@@ -497,6 +497,54 @@ def configure_web_proxy(
...
+
+    :param http_proxy: http proxy to be used by snap. If None, it will
+                       not be configured
+    :param https_proxy: https proxy to be used by snap. If None, it will
+                        not be configured
+    """
+    if http_proxy:
+        subp(["snap", "set", "system", "proxy.http={}".format(http_proxy)])
+
+    if https_proxy:
```
blackboxsw, line 514: I think this function also needs to support the following:
  1. checking existing snap config values, if there are existing proxy settings != to those requested,  prompt with the following message before changing values:
  ```bash
  Setting the https_proxy will overwrite the following system settings:
  snap config system proxy.https = https://blah
  canonical-livepatch = https//livepatch-proxy
Are you sure (Y/n):
```
  2.  Support unsetting snap proxy config values when https_proxy="" or None
  3. Support for unsetting snap http_proxy value if http_proxy == "" or None

File: uaclient/util.py [hunk 8]
```diff
@@ -497,6 +497,54 @@ def configure_web_proxy(
...
+) -> None:
+    """
+    Configure livepatch to use http and https proxies.
+
+    :param http_proxy: http proxy to be used by livepatch. If None, it will
+                       not be configured
+    :param https_proxy: https proxy to be used by livepatch. If None, it will
+                        not be configured
+    """
+    if http_proxy:
```
blackboxsw, line 529: Same comment as on configure_snap_proxy above, this function needs to support unsetting http_proxy and or https_proxy in livepatch if either of those values are None or "".

Also,we need to perform the same pre-flight check for livepatch existing proxy settings before setting them.
Since we probably want a consolidated single prompt if livepatch and/or snap proxies are already set to something non-null it might actually make sense to consolidate both of these configure*proxy functions into one function that preflight checks livepatch and snap proxy settings and provides a single prompt  that accounts for any pre-existing proxy settings which will be overridden.

blackboxsw (changes requested):
A few more suggestions to think about as well. I think we need some sort of logic to prompt if existing proxy values exist in any of snap or canonical-livepatch before we override them.

uaclient/util.py
lucasmoura: Currently, this is true. But when we have the `ua refresh config` logic we will be using this function outside the scope of livepatch. That's why I think we should keep it in `util`

uaclient/entitlements/tests/test_livepatch.py
lucasmoura: No, I will add some unittest for the functions instead

features/proxy_config.feature
lucasmoura: It seems that using this approach here `sudo canonical-livepatch config check-interval=0; sudo canonical-livepatch refresh` still doesn't remove the delay on the squid logs :(

uaclient/util.py
lucasmoura: 1. @orndorffgrant is working on that promp functionality on apt proxies. I think the logic can be generic enough to be used here.

2 and 3. Completely agree

blackboxsw: +1 the prompt can be a separate PR if needed. no need to block this PR on that.

features/proxy_config.feature
blackboxsw: bummer. ok thanks for trying it out.

uaclient/util.py
blackboxsw: Per your  comment, about  `ua refresh config`  being outside the scope of livepatch, will be doing comparable with calling `apt.setup_apt_proxy` 

Sorry, maybe my original comment was unclear; I meant a keeping this separate utility function or function(s) called `configure_snap_proxy` or configure_livepatch_proxy` not a LivepatchEntitlement class method. The reason I suggest having the functions housed in a more specific module is because adding a bunch of unrelated things in uaclient/util.py eventually makes that module harder to maintain because it becomes too generic and could grow any function we don't have a more appropriate home for.

That said, I don't feel super strongly about this because snap + livepatch config manipulation is a bit broader than just livepatch.  In the case of apt proxy, we did move the `setup_apt_proxy` function into the apt.py module because we want all things "apt" related in one place and it's a clear cut bit of APT-only functionality.

I defer to your perspective on this,  but had wanted to explain a bit more the motivation for my original comment.

blackboxsw (approved):
This branch looks like it is in pretty good shape. Minimally, I'd like to see the unittest additions and some retries on subp calls to the snap proxy set commands. I'm marking approve here to avoid blocking the land of this PR once the unittests and retry_sleeps are added. The rest of my comments take as you wish or pull into a separate PR if needed.

uaclient/util.py
lucasmoura: Oh fair enough. No problem then, I have moved those functions into the `livepatch.py` module

lucasmoura:
@blackboxsw code updated

features/proxy_config.feature
```diff
@@ -61,3 +61,43 @@ Feature: Proxy configuration
            | xenial  |
            | bionic  |
            | focal   |
+
+    @series.lts
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attach command when proxy is configured
```
orndorffgrant, line 67: We combined the apt proxy tests into the existing scenario to avoid spinning up a proxy instance twice. We may want to do that here again (combine this into the previous scenario), or maybe the previous scenario is getting too long and this should stay separate. thoughts?

uaclient/util.py
orndorffgrant: So I don't see any code that does (2) or (3). That is, we aren't calling `snap unset ...` anywhere when our config doesn't have a proxy value. I think that this current state is good. We can't safely remove a proxy config in snap/livepatch because of the following:

If our config does not have a `http_proxy` value, and snap is not configured with a proxy, then there is nothing to do. Calling `snap unset ...` won't harm anything though. This is not a problem case - the following is:

If our config does not have a `http_proxy` value, but snap is _already configured_ to use a proxy, then we could be in one of two scenarios:
1. the user previously _used UA_ to set a proxy, then _used UA_ to unset the proxy. In this scenario, we want to run `snap unset...`
2. the user previously _manually ran_ `snap set...` to set the proxy, and hasn't touched the UA config at all. In this scenario, we shouldn't `snap unset...`. It would also be confusing if `ua enable livepatch` prompted to remove the user's snap proxy (even worse: if we went with a prompt in this scenario then a `ua enable livepatch --assume-yes` would unexpectedly remove the users snap proxy).

The problem is: Given a uaclient.conf with no proxy set and seeing that snap is already configured to use a proxy, we don't know if we're in scenario 1 or 2.

We will be able to tell which scenario we're in if the user runs `ua config unset http_proxy`, so in that case, we can safely run `snap unset...`, but otherwise, I don't think we can.

All that is to say: I think this PR does not need `snap unset...` or the `canonical-livepatch` equivalent, which means that it is good as is.

> As an aside: this complication may prevent the prompting logic from being nicely shared between the apt proxy and snap/livepatch proxy configuration... we'll see as I work through it.

features/proxy_config.feature
orndorffgrant: Oops I meant to cancel this comment. I noticed that this scenario requires a vm for livepatch

features/steps/steps.py
```diff
@@ -168,12 +167,15 @@ def when_i_configure_uaclient_using_proxy_machine(
         DEFAULT_CONFIG_FILE, tmp_local_conf
     )
     cfg = yaml.safe_load(load_file(tmp_local_conf))
+    new_cfg = {}
```
orndorffgrant, line 170: Interesting, was something failing because of appending the whole config?

features/steps/steps.py
```diff
@@ -168,12 +167,15 @@ def when_i_configure_uaclient_using_proxy_machine(
...
 
     if "ua_config" not in cfg:
-        cfg["ua_config"] = {}
-    cfg["ua_config"][proxy_cfg + "_proxy"] = proxy_cfg_value
+        new_cfg["ua_config"] = {}
+        new_cfg["ua_config"][proxy_cfg + "_proxy"] = proxy_cfg_value
+        context.text = yaml.dump(new_cfg).strip()
+    else:
+        context.text = "  {}: {}".format(proxy_cfg + "_proxy", proxy_cfg_value)
```
orndorffgrant, line 177: This line assumes that the `ua_config` key is at the end of the file. We may need something like this instead:
```suggestion
        new_cfg["ua_config"] = cfg["ua_config"]
        new_cfg["ua_config"][proxy_cfg + "_proxy"] = proxy_cfg_value
        context.text = yaml.dump(new_cfg).strip()
```

features/steps/steps.py
lucasmoura: Yeah, when we run the code twice, we were appending the whole config all over again :(

orndorffgrant: True, I just didn't expect that to break anything - regardless, this implementation is better and cleaner :+1:

features/steps/steps.py
lucasmoura: Yep, you are right on that one. But this code will be problematic because on that situation, we will create two entries of `ua_config` on the `uaclient.conf` file.

An other thing we could do here is rewrite the config instead of appending. In that situation, the original code should work fine. What do you think ?

orndorffgrant: > we will create two entries of ua_config on the uaclient.conf file
My confusion stems from not expecting that to be a problem :upside_down_face: - I would expect the yaml parser to overwrite the previous value, but then again I didn't actually look at the yaml spec or test it out to see what pyyaml actually does:grimacing: 
> rewrite the config instead of appending
Yeah thats probably the best solution +1. Note that this is what I started doing and ran into permissions problems, which is why I ended up just calling the append step. You'll likely have to do something like:
```
    cmd = "printf '{}\n' > {}".format(context.text, DEFAULT_CONFIG_FILE)
    cmd = 'sh -c "{}"'.format(cmd)
    when_i_run_command(context, cmd, "with sudo")
```

uaclient/util.py
lucasmoura: That's an excellent point here. I was kind of performing that by assuming that if `http_proxy=""` then we will unset it. Even though I don't call unset directly, it seems that setting the proxy as empty produces the same expected output on both livepatch and snap.

But this is really problematic as you pointed in scenario 2. Specially if we start delivering `uaclient.conf` with `ua_config` like this:

```
ua_config:
  http_proxy: ""
  https_proxy: ""
```

Maybe the safest way for us to handle that is to just unset proxies on the `ua config unset` scenario. But let's bring this with the team and better discuss it.

features/steps/steps.py
lucasmoura: Tbh, the tests were failing after the rebase and when I looked at the `uaclient.conf` I assumed that this was the problem. Let me double check that and confirm if they do fail.

Maybe I was executing the tests with wrong parameters and assumed the error was caused by this.

uaclient/util.py
orndorffgrant: Ah and good point about the empty string values! I wasn't even thinking about that. Yes lets discuss this more today

features/steps/steps.py
lucasmoura: @orndorffgrant You are right, that was not the problem on the test. I have removed that change.

However, I still think the best way is to rewrite the config, but we can do this in another PR

lucasmoura:
@orndorffgrant PR updated to not unset proxies if config variable is empty

orndorffgrant (approved):
LGTM!

lucasmoura:
Unrelated CI errors

PR #1687, opened 2021-06-23
proxy confirm changes (SC-153)

orndorffgrant:
## Proposed Commit Message

> proxy: prompt before overwriting existing values
> 
> This applies to any time we're configuring a proxy. We are now checking
> to see if there is already a value, and if it is different from our new
> value, we prompt the user to ask them if we should overwrite.

## Test Steps

- manually set an apt proxy
- try to set the apt proxy via ua, by adding to uaclient.conf and then enabling an apt based service
- verify you are prompted
- test both answers to prompt work as expected

repeat steps for snap and livepatch proxies

## Desired commit type::

## Checklist:


uaclient/apt.py
```diff
@@ -426,3 +426,46 @@ def setup_apt_proxy(
...
+    :param protocol: can be "http" or "https", the proxy to get the value of
+    :return: the value of the apt proxy, or None if not set
+    """
+    apt_config_key = "Acquire::{}::Proxy".format(protocol)
+    out, _ = util.subp(["apt-config", "shell", "key", apt_config_key])
+    match = re.search("key='(.*)'", out)
+    if match:
+        return match.group(1)
+    else:
+        return None
```
lucasmoura, line 445: I think we can turn this into a one line return:

```python
return match.group(1) if match else None
```

uaclient/entitlements/livepatch.py
```diff
@@ -88,6 +134,56 @@ def configure_livepatch_proxy(
...
+
+    :param protocol: can be any valid livepatch config option
+    :return: the value of the livepatch config option, or None if not set
+    """
+    out, _ = util.subp(["canonical-livepatch", "config"])
+    match = re.search('{}: "(.*)"'.format(key), out)
+    if match and match.group(1):
+        return match.group(1)
+    else:
+        return None
```
lucasmoura, line 149: Same oneliner here

uaclient/tests/test_apt.py
```diff
@@ -977,10 +976,71 @@ def test_setup_apt_proxy_config(
...
+            (("", "nonsense"), None),
+            (("key='value'", ""), "value"),
+        ],
+    )
+    @pytest.mark.parametrize(
+        "protocol, expected_apt_key",
+        [("http", "Acquire::http::Proxy"), ("https", "Acquire::https::Proxy")],
+    )
+    @mock.patch("uaclient.util.subp")
+    def test_get_current_proxy_value(
```
lucasmoura, line 1001: I think we can also have that test for the snap and livepatch implementations

lucasmoura (commented):
That looks really good @orndorffgrant. I have just a couple of small nits here

Also, I think we can extend that proxy vm test to at least verify if the prompt happens for snap

uaclient/tests/test_apt.py
orndorffgrant: there should be `test_get_snap_config_option_value` and `test_get_livepatch_config_option_value` in `test_livepatch.py` unless you're referring to something else?

orndorffgrant:
Good point @lucasmoura I will add to that scenario to test the prompt for snap/livepatch

uaclient/tests/test_apt.py
lucasmoura: My mistake here, I completely missed those tests

uaclient/entitlements/livepatch.py
```diff
@@ -135,50 +232,51 @@ def _perform_enable(self) -> bool:
...
-                    capture=True,
-                    retry_sleeps=apt.APT_RETRIES,
-                )
-            elif "snapd" not in apt.get_installed_packages():
-                raise exceptions.UserFacingError(
-                    "/usr/bin/snap is present but snapd is not installed;"
-                    " cannot enable {}".format(self.title)
-                )
-
+        if not util.which(SNAP_CMD):
```
orndorffgrant, line 235: Notice that I did refactor this section a little bit to ensure that we always try to configure the snap proxy if necessary (before it would only install the snap proxy if canonical-livepatch was not installed).

I basically just moved the `if not util.which(SNAP_CMD)` out from underneath the `if not util.which("/snap/bin/canonical-livepatch")`

orndorffgrant (commented):
Okay @lucasmoura @blackboxsw I think this is ready now

blackboxsw (commented):
This is solid thanks for the refactor and the simple helper functions. Make for easy readability. 

A couple of things that maybe we can discuss about viability for subsequent PRs (or this one if desired)
   1. I don't think we can set an APT proxy url that doesn't have a protocol scheme defined. Do we want to do some minimal value check on the value and warn `if urllib.parse.urlparse(<url>).schema == ''`
   2. Do we want to add a validate_apt_proxy subp(['timeout', 5, 'apt', 'update']) to validate that apt proxy worked in `setup_apt_proxy_with_prompts` and revert to original value while still in memory if broken?   If we allow promoting a bogus value into apt proxy config, admins won't find out about it until next daily unattended upgrades, or later.
   3. Same question with livepatch on validation of proper proxy config settings before exiting. I'm not sure, but could we validate the proxy with a `livepatch refresh` run to see if it exits non-zero on invalid proxy?

uaclient/apt.py
```diff
@@ -426,3 +426,43 @@ def setup_apt_proxy(
...
+    """
+    http_proxy_to_set, https_proxy_to_set = util.prompt_for_proxy_changes(
+        "apt",
+        curr_http_proxy=get_current_proxy_value("http"),
+        curr_https_proxy=get_current_proxy_value("https"),
+        new_http_proxy=http_proxy,
+        new_https_proxy=https_proxy,
+        assume_yes=assume_yes,
+    )
+    setup_apt_proxy(http_proxy_to_set, https_proxy_to_set)
```
blackboxsw, line 468: Should  attempt to run `apt update` to validate that our promoted proxy config works and revert to original values if it fails?

uaclient/entitlements/livepatch.py
blackboxsw: Makes sense to me because we are trying to enable livepatch at this point anyway +1 not nesting that logic unnecessarily. As you mentioned, if livepatch were already installed, accidentally **not** setting up the snap proxy settings is a gap that needed addressing.

features/proxy_config.feature
```diff
@@ -56,6 +57,39 @@ Feature: Proxy configuration
         And I run `ua enable cis` with sudo
         Then I verify that no files exist matching `/etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy`
 
+        When I run `ua disable cis` with sudo
+        And I create the file `/etc/apt/apt.conf.d/50-testproxy` with the following
+        """
+        Acquire::http::Proxy "localhost:1234";
```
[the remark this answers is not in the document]
orndorffgrant: Very good point! I'm adding `http://` to the front of all the example values in this file

orndorffgrant:
I've added protocol schemes to the fake values in the integration tests and I'm letting jenkins verify the tests still pass

For validation, we decided out of band that we will do basic URL validation, but not try to verify the proxy is functioning (by running `apt update` for example). This is captured in this issue https://warthogs.atlassian.net/browse/SC-170 and will be implemented in a follow up PR

lucasmoura (approved):
LGTM

If we don't see the ipv4 dns error on Jenkins. I am good merging this without it

orndorffgrant:
I also failed to adhere to the spec on these prompts. I gave each case an individual prompt because there could be a (very niche) situation where a user needs to overwrite the livepatch proxy but not the snap proxy. Also, adding extra prompts didn't seem like too much of a burden because we have --assume-yes. We might want to meet on monday to discuss, or we could merge as-is and revise later.

orndorffgrant:
@lucasmoura I added `@series.focal` to our proxy vm test because I wanted jenkins to run the test for me, but it is failing due to livepatch not being installed - do we expect livepatch to not be enabled on focal attach?

lucasmoura:
@orndorffgrant Yes, I think livepatch should not work on focal at the moment

orndorffgrant:
@lucasmoura Got it, I should've asked first! :sweat_smile: I'll remove that and run it locally to verify it still passes, then update this PR

blackboxsw:
+1, I was incorrect in saying that this isn't a problem on focal. Turns out lxd focal kvm images use `kernel 5.4.0-1041-kvm.kernel 5.4.0-1041-kvm` but multipass focal kvm images use `kernel 5.4.0-77-generic` which livepatch **does** support. So multipass would allow this test and lxd kvm images won't allow livepatch

blackboxsw (approved):

blackboxsw:
> I also failed to adhere to the spec on these prompts. I gave each case an individual prompt because there could be a (very niche) situation where a user needs to overwrite the livepatch proxy but not the snap proxy. Also, adding extra prompts didn't seem like too much of a burden because we have --assume-yes. We might want to meet on monday to discuss, or we could merge as-is and revise later.

As mentoined in MM: "yeah I recognized the individ prompts versus banner prompt in the spec on first review.+1 the separate prompts are easy and understandable. and if people don't like typing y<enter> we have the --assume-yes param"

orndorffgrant:
Closing in favor of #1693

PR #1688, opened 2021-06-23
status: show env vars on json output (SC-155)

lucasmoura:
## Proposed Commit Message
status: show env vars on json output

We are now displaying if there is any ua specific environment variables set when the user runs ua status --format json

## Test Steps
Run the updated unittest or set a valid UA env variable locally and verify that it appears in the json output
of `ua status --format json`

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -871,9 +871,7 @@ def action_status(args, cfg):
             status = cfg.status(show_beta=show_beta)
         print("")
     if args and args.format == "json":
-        if status["expires"] != ua_status.UserFacingStatus.INAPPLICABLE.value:
-            status["expires"] = str(status["expires"])
-        print(json.dumps(status))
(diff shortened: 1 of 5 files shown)


orndorffgrant (changes requested):
Thanks this looks good! I like how you reused the envvar allowlist, and also how you separated the json status construction into its own function, which will be useful as we add more to it.

I think this is just missing two other environment things that `ua` also cares about. We also look at `UA_CONFIG_FILE` and any envvar that starts with `UA_FEATURES_`, so we should include them in the output too.

lucasmoura:
@orndorffgrant great catch. Yes, I will add all of them

lucasmoura:
@orndorffgrant code updated

orndorffgrant (approved):
Thanks for the updates, this looks good to me!

PR #1689, opened 2021-06-24
status: do not show unavailable services on json

lucasmoura:
## Proposed Commit Message
status: do not show unavailable services on json

We are now only showing available services in the status json output. Because of that, we are also dropping the available key from the json output, since now it is a redundant key.

## Test Steps
Run the modified unittests
Launch a container and verify that `cc-eal` does not appear when unattaches and attached

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -520,6 +520,7 @@ class UAConfig:
                 description_override = inapplicable_resources[ent.name]
             else:
                 ent_status, details = ent.user_facing_status()
+
         return {
             "name": ent.name,
             "description": ent.description,
@@ -527,6 +528,9 @@ class UAConfig:
             "status": ent_status.value,
             "statusDetails": details,
             "description_override": description_override,
+            "available": "yes"
+            if ent.name not in inapplicable_resources
+            else "no",
         }
 
     def _attached_status(self) -> "Dict[str, Any]":
@@ -565,6 +569,7 @@ class UAConfig:
             response["services"].append(
                 self._attached_service_status(ent, inapplicable_resources)
             )
+
         support = self.entitlements.get("support", {}).get("entitlement")
         if support:
             supportLevel = support.get("affordances", {}).get("supportLevel")
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -648,4 +648,11 @@ def format_json_status(status: "Dict[str, Any]") -> str:
         or name == "UA_CONFIG_FILE"
     ]
 
+    available_services = [
+        service
+        for service in status.get("services", [])
+        if service.get("available", "yes") == "yes"
+    ]
+    status["services"] = available_services
+
     return json.dumps(status)
--- a/uaclient/tests/test_cli.py
+++ b/uaclient/tests/test_cli.py
@@ -253,7 +253,7 @@ class TestCLIParser:
 
         status_msg = "enabled" if ent_msg == "yes" else "—"
         ufs_call_count = 1 if ent_msg == "yes" else 0
-        ent_name_call_count = 2 if ent_msg == "yes" else 1
+        ent_name_call_count = 3 if ent_msg == "yes" else 2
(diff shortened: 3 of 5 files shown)


uaclient/config.py
```diff
@@ -520,11 +520,15 @@ def _attached_service_status(
...
                 ent_status, details = ent.user_facing_status()
+
         return {
             "name": ent.name,
             "description": ent.description,
             "entitled": contract_status.value,
             "status": ent_status.value,
+            "available": status.UserFacingAvailability.UNAVAILABLE.value
+            if ent.name in inapplicable_resources
+            else status.UserFacingAvailability.AVAILABLE.value,
```
orndorffgrant, line 531: Since we're not actually going to display this value, can we just use a boolean here instead of the string values?

lucasmoura: We could, but I made that field consistent with the `unattached` status json. Since changing it there will be considered a schema change, I preferred to follow that pattern here as well.

orndorffgrant: Ah I see what you mean now, cool.

I do think we should move towards `status` returning data, and having separate formatting functions take care of replacing the data with a user-readable strings if necessary. So in this case, that would mean returning a bool here (for both attached and unattached), and then interpreting the bool as a string over in `format_tabular`. That's just my preference though and certainly doesn't have to be in scope for this PR.

lucasmoura: I think this a good suggestion. I think we can keep this as is for now. As we work on the status json spec and see other similar behaviors like this, we can have a PR that makes all of those changes at once. I think this will be easier to address when we have a better understanding of all the data changes we want to perform

orndorffgrant: Totally, sounds good to me!

orndorffgrant (approved):
Just ran a manual test to verify - looks good!

uaclient/config.py
blackboxsw: @lucasmoura @orndorffgrant also before releasing this change, we need coordinate with desktop folks RobertAncell because they currently match ["available"== "yes"](https://git.launchpad.net/ubuntu/+source/software-properties/tree/softwareproperties/gtk/utils.py#n127) to determine if is a service is applicable when unattached. If we want to release this changeset in our next SRU, we'll need to minimally [file a bug file against software-properties](https://bugs.launchpad.net/ubuntu/+source/software-properties/+filebug) and/or help provide a minor PR to fix this and wait/help with an SRU release for software-properties that would fix this behavior in order to that any service listed when unattached and lacking an "available" key in status.json can still be  determined as "available".

blackboxsw (commented):
I think landing this is probably a blocker for this release of ua-client because we need to roll out comparable SRU in software-properties to handle this before we SRU this content.

uaclient/config.py
lucasmoura: Got it. I can align that with them

orndorffgrant:
@blackboxsw @lucasmoura I'm thinking we should put the new output behind a feature flag. That way we can have the new feature and not be blocked by other SRUs. Lets talk after standup today.

lucasmoura:
@orndorffgrant @blackboxsw PR updated to not drop `available` key in unattached status. I have also removed the `available`filter for the attached status. Let me know if we should still apply that filter there.

orndorffgrant:
> I have also removed the available filter for the attached status. Let me know if we should still apply that filter there.

I thought we were still going to filter to only "available" services in attached status, unless that also breaks something?

lucasmoura:
@orndorffgrant No, it doesn't break anything (AFIK), but I would be adding a new key to the `attached`  status output.
But since this modification should not break anything, I will bring it back

lucasmoura:
@orndorffgrant code updated

orndorffgrant (approved):
Single failing integration test is unrelated and fixed in #1690 
Thanks @lucasmoura! Merging now

PR #1690, opened 2021-06-24
fips: don't install openssh hmac packages on focal

lucasmoura:
## Proposed Commit Message
fips: don't install openssh hmac packages on focal

On Focal, we don't have the openssh hmac packages. Because of that, we must exclude then when enabling any FIPS service on Focal

PS:  Originally, i though about using a different solution that check if the package can be installed before adding it to the list of installable FIPS packages. However, this could create scenarios were we partially install FIPS packages, which is not ideal. That's why I am explicit on which package should be included based on the distro version

## Test Steps
1) attach a valid fips-preview contract on a focal machine
2) manually change the name of `fips-updates'  to `fips-preview`
3) enable `fips-preview`  and test to see if everything is working as expected

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -23,24 +23,6 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
     repo_key_file = "ubuntu-advantage-fips.gpg"  # Same for fips & fips-updates
     FIPS_PROC_FILE = "/proc/sys/crypto/fips_enabled"
 
-    """
-    Dictionary of conditional packages to be installed when
-    enabling FIPS services. For example, if we are enabling
-    FIPS services in a machine that has openssh-client installed,
-    we will perform two actions:
-
-    1. Upgrade the package to the FIPS version
-    2. Install the correspinding hmac version of that package.
-    """
-    conditional_packages = [
-        "openssh-client",
-        "openssh-client-hmac",
-        "openssh-server",
-        "openssh-server-hmac",
-        "strongswan",
-        "strongswan-hmac",
-    ]
-
     # RELEASE_BLOCKER GH: #104, don't prompt for conf differences in FIPS
     # Review this fix to see if we want more general functionality for all
     # services. And security/CPC signoff on expected conf behavior.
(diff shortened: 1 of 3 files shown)


uaclient/entitlements/fips.py
```diff
@@ -49,6 +31,33 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
...
+        1. Upgrade the package to the FIPS version
+        2. Install the correspinding hmac version of that package.
+        """
+        conditional_packages = ["strongswan", "strongswan-hmac"]
+
+        series = util.get_platform_info().get("series", "")
+        # On Focal, we don't have the openssh hmac packages.
+        # Therefore, we will not try to install them during
+        # when enabling any FIPS service
+        if series != "focal":
```
blackboxsw, line 51: Given that this will likely be focal++, I think we may want to say the following:
```suggestion
        if series in ('xenial', 'bionic')
```

blackboxsw (changes requested):
I think this looks good @lucasmoura one minor suggestion for forward-compatibility with the next LTS series "J". is to include packages only on known LTS releases. This way of J release has same HMAC package support offering as focal we have not changes to make in uaclient.

The one outstanding question I have is generally if the *-hmac packages will go away over time as the upstream fixes are integrated in standard debian packages in the <release>-updates pocket. If this is the case (and we think that `strongswan-hmac` will also be unneeded in focal or later LTSes, then maybe we should check `apt-cache policy` on any -hmac pkg before including in the conditional packages.

uaclient/entitlements/fips.py
lucasmoura: Yes, using `apt-cache policy` was my first option here. My only concern with it is that it could cause incomplete FIPS installs. For example, suppose that for some reason, the `openssh-client-hmac` package disappears from the pocket. If we are using the apt-cache policy on a xenial enable. We will enable FIPS but not install that package. I know this is a corner case, but that's why I have not used that approach

blackboxsw: Yeah this is true. The case I was thinking of was in Xenial if security team does remove openssh-client-hmac it would only be because the necessary complement of "-hmac" functionality would be released in the openssh-client deb delivered in the FIPS xenial-updates pocket. Making packages disappear from a launchpad PPA is hard to do and would have to be intentional by the security team. Given that FIPS is/was certified with a certain set of packages I don't think that can ever change anyway. But that said, I think we shouldn't spend the cycles trying to make it more dynamic, since each LTS version of FIPS should be a frozen package set anyway due to the certification constraints for it. If it becomes more dynamic in the "J" LTS release, then we can revisit. What you have is good for me.

blackboxsw (commented):
With the minor tweak on if series in ("xenial", "bionic") I'm +1
If you can add a unit test that exercises conditional_packages value on focal,xenial, bionic to assert expected behavior even better.

blackboxsw (approved):
+1 pending CI pass without any related failures.

PR #1691, opened 2021-06-25
Version integration test

orndorffgrant:
## Background
We need our tests to print the version under test. We print the version of ua to stdout, but we print _a lot_ of stuff to stdout. So we probably don't want to put all the stdout noise in front of SRU reviewers just for version verification. This is an attempt to verify the version in a way that works with the natural output of behave and works (mostly) like any other integration test.

The way I got the version test to work in such a way that it uses the environment variable and prints everything in a pretty way seems a bit hacky, but wasn't sure of a better way to do it. Let me know if there is a better way.

I also modified how we're skipping tests based  on series which should reduce the amount of noisy output we're currently getting. It also seems a bit hacky... but the output is much nicer

## Proposed Commit Message

> tests: add test to check and output version being tested

> tests: filter on series earlier to avoid extra output 

## Test Steps

- Run `tox -e behave-lxd-20.04` (for example)
- Verify that the version test is the first to run
- Verify that the version test is skipped because you didn't set UACLIENT_BEHAVE_CHECK_VERSION
- Verify that all the different series version tests were skipped and did not displaying the whole scenario before being skipped

- Run `UACLIENT_BEHAVE_CHECK_VERSION=27.1 tox -e behave-lxd-20.04 -- features/_version.feature` (for example)
- Verify that the test fails because the actual version is different

## Desired commit type::

## Checklist:


features/_version.feature
```diff
@@ -0,0 +1,18 @@
+Feature: UA is expected version
+
+    @series.all
+    @check.version
```
blackboxsw, line 4: I think we can instead use the @uses.config.check_version decorator which will automatically handle the skip logic if that attribute is not provided via `UACLIENT_BEHAVE_CHECK_VERSION=blah`

features/environment.py
```diff
@@ -459,6 +463,37 @@ def before_scenario(context: Context, scenario: Scenario):
...
+    given_a_series_match = re.match(
+        "a `(.*)` machine with ubuntu-advantage-tools installed",
+        scenario.steps[0].name,
+    )
+    if filter_series and given_a_series_match:
+        series = given_a_series_match.group(1)
+        if series and series not in filter_series:
+            scenario.skip(
+                reason=(
+                    "Skipping scenario outline series `{series}`."
```
blackboxsw, line 478: I really like this additional context/messaging +1

features/environment.py
```diff
@@ -459,6 +463,37 @@ def before_scenario(context: Context, scenario: Scenario):
...
+            return
+
+    if "check.version" in scenario.effective_tags:
+        if context.config.check_version is None:
+            scenario.skip(reason="UACLIENT_BEHAVE_CHECK_VERSION not set")
+            return
+        else:
+            # before_step doesn't execute early enough to modify the step
+            # so we perform the version step surgery here
+            version_step = [s for s in scenario.steps if s.text == "VERSION"][
```
blackboxsw, line 493: Let's add bookend decorators around the VERSION variable name to more clearly indicate that this is a variable intended for substitution  like `%VERSION%` or `{VERSION}`.
Also, can we rename this more explicitly `UACLIENT_BEHAVE_CHECK_VERSION` so we know where that variable is sourced from.  

I can think of other use-cases where we might want to generally replace structured variable names text with something from the getattr(context.config, attr_name) if we come across `{SOME_VAR}` in text. Such as validating properly configured ppa based on UACLIENT_BEHAVE_PPA or UACLIENT_BEHAVE_PPA_KEYID

blackboxsw: Here's what I was thinking as an example: we'd be able to extend it a bit with a general variable-replacement if needed in the future by adding other "uses.config" decorators.
```diff
diff --git a/features/_version.feature b/features/_version.feature
index 2a32854..95c8a92 100644
--- a/features/_version.feature
+++ b/features/_version.feature
@@ -1,13 +1,13 @@
 Feature: UA is expected version
 
     @series.all
-    @check.version
+    @uses.config.check_version
     Scenario Outline: Check ua version
         Given a `<release>` machine with ubuntu-advantage-tools installed
         When I run `ua version` with sudo
         Then I will see the following on stdout
         """
-        VERSION
+        {UACLIENT_BEHAVE_CHECK_VERSION}
         """
         Examples: version
             | release |
diff --git a/features/environment.py b/features/environment.py
index 21c606a..cf12cb1 100644
--- a/features/environment.py
+++ b/features/environment.py
@@ -483,17 +483,17 @@ def before_scenario(context: Context, scenario: Scenario):
             )
             return
 
-    if "check.version" in scenario.effective_tags:
-        if context.config.check_version is None:
-            scenario.skip(reason="UACLIENT_BEHAVE_CHECK_VERSION not set")
-            return
-        else:
-            # before_step doesn't execute early enough to modify the step
-            # so we perform the version step surgery here
-            version_step = [s for s in scenario.steps if s.text == "VERSION"][
-                0
-            ]
-            version_step.text = context.config.check_version
+    if "uses.config.check_version" in scenario.effective_tags:
+        # before_step doesn't execute early enough to modify the step
+        # so we perform the version step surgery here
+        for s in scenario.steps:
+            if not getattr(s, 'text'):
+                continue
+            if "{UACLIENT_BEHAVE_CHECK_VERSION}" in s.text:
+                s.text = s.text.replace(
+                    "{UACLIENT_BEHAVE_CHECK_VERSION}",
+                    context.config.check_version,
+                )
 
 
 FAILURE_FILES = (
```

blackboxsw (commented):
Great idea. Couple thoughts about using existing integration test logic w/ uses.config.X

features/environment.py
orndorffgrant: Really good points! Yep I'll make these suggested changes.

I also need to make sure the version test has all the tags necessary to be run for each test suite

orndorffgrant:
@blackboxsw @lucasmoura I made the suggested changes. I also ended up copy-pasting the test several times with different machine type tags to get it to run for each test suite - is there a better way to do that?

lucasmoura:
@orndorffgrant Yeah, one thing we might do to avoid that is create a `machine_type.all` tag. This will avoid this need. However, this will avoid some of the redundancy. However, there will still be some redundancy, since we don't support all series on all machine types. For example, the PRO and cloud generic tests should only be supported for LTS versions.

features/_version.feature
```diff
@@ -0,0 +1,44 @@
...
+    @series.all
+    @uses.config.check_version
+    @uses.config.machine_type.lxd.container
+    @uses.config.machine_type.lxd.vm
+    @uses.config.machine_type.aws.generic
+    @uses.config.machine_type.aws.pro
+    @uses.config.machine_type.azure.generic
+    @uses.config.machine_type.azure.pro
+    @uses.config.machine_type.gcp.generic
+    @uses.config.machine_type.gcp.pro
```
lucasmoura, line 12: My only concern here is that some of the machine types don't support all releases, like the PRO ones that are only supported by LTS releases. However, since we don't have entries for those tests in `tox.ini` (like `behave-gcppro-21.04`), I don't think there is any risk here about them.

So I am fine with this approach.

orndorffgrant: Ahh right - That is a good point that reading this test, it looks like it could run a hirsute gcppro instance, which could be confusing. But like you said, in practice the other releases should be skipped based on how we use tags for each test suite to only run the relevant releases.

So yeah, not perfect, but I think good enough for this release at least.

lucasmoura (approved):

PR #1692, opened 2021-06-25
status: update contract info on status json output

lucasmoura:
## Proposed Commit Message
status: update contract info on status json output

We are now removing from root contract based information and adding that info into a contract specific object. Additionally, the products and created_at info for the contract are now present in the json output as well

## Test Steps
Launch a container and verify that the contract object appears on the json output of `ua status --format json` when unattaches/attached

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -45,7 +45,13 @@ DEFAULT_STATUS = {
     "configStatus": status.UserFacingConfigStatus.INACTIVE.value,
     "configStatusDetails": status.MESSAGE_NO_ACTIVE_OPERATIONS,
     "notices": [],
-    "techSupportLevel": status.UserFacingStatus.INAPPLICABLE.value,
+    "contract": {
+        "id": "",
+        "name": "",
+        "created_at": "",
+        "products": [],
+        "tech_support_level": status.UserFacingStatus.INAPPLICABLE.value,
+    },
 }  # type: Dict[str, Any]
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


uaclient/config.py
```diff
@@ -544,7 +544,9 @@ def _attached_status(self) -> "Dict[str, Any]":
                 "origin": contractInfo.get("origin"),
                 "subscription": contractInfo["name"],
                 "subscription-id": contractInfo["id"],
+                "contract-created-at": contractInfo["createdAt"],
                 "notices": self.read_cache("notices") or [],
+                "products": contractInfo["products"],
             }
```
orndorffgrant, line 550: I was expecting the structure change to happen up in this function I think.

Currently this function returns an internal representation of the data that `format_json_status` interprets and restructures - which feels like an extra step.

What do you think of returning the new structure from this function? @lucasmoura

uaclient/status.py
```diff
@@ -648,4 +650,18 @@ def format_json_status(status: "Dict[str, Any]") -> str:
...
+    contract_id = status.pop("subscription-id", "")
+    created_at = status.pop("contract-created-at", "")
+    tech_lvl_support = status.pop("techSupportLevel", "")
+    products = status.pop("products", [])
+
+    status["contract"] = {
+        "id": contract_id,
+        "name": contract_name,
+        "created_at": created_at,
+        "tech_level_support": tech_lvl_support,
```
orndorffgrant, line 663: I think we want to leave the word order the same as it was before, just snake case now, so `tech_support_level`

uaclient/status.py
```diff
@@ -648,4 +650,18 @@ def format_json_status(status: "Dict[str, Any]") -> str:
...
+        "products": products,
+    }
+
+    return json.dumps(status, cls=DatetimeAwareJSONEncoder)
```
orndorffgrant, line 667: Nice this will help with the other date issues as well :)

uaclient/config.py
lucasmoura: It makes sense. I was trying to avoid changing the internal representation of the data, but there is no good reason to keep it as is. I will update the code

lucasmoura:
@orndorffgrant Code updated. I have run some integration tests locally to see if status has not changed, but let's see if CI can find any gap that I missed

orndorffgrant:
Of the jenkins test failures. 1 is the recent xenial motd bug, which should be fixed in `main` now. The other is an awspro failure where livepatch is not enabled, which seems strange and worth investigating, but unrelated to this PR.

orndorffgrant (approved):
LGTM!

PR #1693, opened 2021-06-28
proxy: print message when setting proxy (SC-179)

orndorffgrant:
This replaces #1687 

Note this also includes the same bugfix as mentioned here: https://github.com/canonical/ubuntu-advantage-client/pull/1687/files#r658167087

## Proposed Commit Message

> proxy: print message when setting proxy

## Test Steps

Set the proxy variables in the config.
Enable an apt based service and livepatch.
Verify concise messages are printed that help you understand what UA is doing.

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -408,12 +408,15 @@ def setup_apt_proxy(
     :param https_proxy: the url of the https proxy apt should use, or None
     :return: None
     """
+    if http_proxy or https_proxy:
+        print(status.MESSAGE_SETTING_SERVICE_PROXY.format(service="APT"))
+
     apt_proxy_config = ""
-    if http_proxy is not None:
+    if http_proxy:
         apt_proxy_config += status.MESSAGE_APT_PROXY_HTTP.format(
             proxy_url=http_proxy
         )
-    if https_proxy is not None:
+    if https_proxy:
         apt_proxy_config += status.MESSAGE_APT_PROXY_HTTPS.format(
             proxy_url=https_proxy
         )
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -39,6 +39,9 @@ def configure_snap_proxy(
     :@param snap_retries: Optional list of sleep lengths to apply between
                           snap calls
     """
+    if http_proxy or https_proxy:
+        print(status.MESSAGE_SETTING_SERVICE_PROXY.format(service="snap"))
+
     if http_proxy:
         util.subp(
             ["snap", "set", "system", "proxy.http={}".format(http_proxy)],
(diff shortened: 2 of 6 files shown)


uaclient/entitlements/livepatch.py
```diff
@@ -67,6 +70,9 @@ def configure_livepatch_proxy(
     :@param livepatch_retries: Optional list of sleep lengths to apply between
                                snap calls
     """
+    if http_proxy or https_proxy:
+        print(status.MESSAGE_SETTING_SERVICE_PROXY.format(service="livepatch"))
```
blackboxsw, line 74: We may want to use class attr here `LivepatchEntitlement.title` as the enable/disable messaging sources that value for a Capitalized value for the service name.  This way if we change `LivepatchEntitlement.title` in the future we won't have to find this text and adapt it later

uaclient/apt.py
```diff
@@ -408,12 +408,15 @@ def setup_apt_proxy(
     :param https_proxy: the url of the https proxy apt should use, or None
     :return: None
     """
+    if http_proxy or https_proxy:
+        print(status.MESSAGE_SETTING_SERVICE_PROXY.format(service="apt"))
```
blackboxsw, line 412: This should be capitalized "APT" for human-readable representation of the APT-service stack. I think we generally have used the appropriate service capitalization when representing human-facing prompts/info.

blackboxsw (changes requested):
minor capitalization for messaging nits.

lucasmoura:
@orndorffgrant @blackboxsw Do you think we need to display those messages when setting the proxy for `urllib` as well ?

blackboxsw:
> @orndorffgrant @blackboxsw Do you think we need to display those messages when setting the proxy for `urllib` as well ?

I think not @lucasmoura because this is only setting a local proxy in UA client and not affecting "external" services.

blackboxsw:
Known unrelated CI failure due to motd changes.
```bash
      Expected: a string matching '\d+ package(s)? can be updated.
      \d+ of these updates (is a|are) security update(s)?.'
           but: was "Welcome to Ubuntu 16.04.7 LTS (GNU/Linux 5.4.0-74-generic x86_64)\n\n * Documentation:  https://help.ubuntu.com\n * Management:     https://landscape.canonical.com\n * Support:        https://ubuntu.com/advantage\n\nUA Infra: Extended Security Maintenance (ESM) is not enabled.\n\n1 update can be applied immediately.\n1 of these updates is a standard security update.\nTo see these additional updates run: apt list --upgradable\n\n17 additional security updates can be applied with UA Infra: ESM\nLearn more about enabling UA Infra: ESM service for Ubuntu 16.04 at\nhttps://ubuntu.com/16-04\n\nNew release '18.04.5 LTS' available.\nRun 'do-release-upgrade' to upgrade to it."
```

blackboxsw (approved):
+1

PR #1694, opened 2021-06-28
ua refresh config (SC-123)

orndorffgrant:
This replaces #1678 

## Proposed Commit Message

> refresh: add target to refresh command
> 
> ua refresh now accepts an optional arg, which can be config or contract.
> Depending on the value, ua refresh will either process any config
> changes or re-fetch the contract from the server. If the arg is not
> provided, then ua refresh will do both.
> 
> ua refresh config will also print a warning if there is no proxy set in
> uaclient.conf but there is a snap or livepatch proxy set. We can't
> actually unset a snap/livepatch proxy in this scenario because this may
> have been intended by the user. Instead of unsetting, we provide
> instructions to either remove the snap/livepatch proxy setting, or add
> the proxy config to uaclient.conf.

>  git: ignore more integration test artifacts 

## Test Steps

set some proxy variables in uaclient.conf, then run ua refresh config or ua refresh
remove those proxy variables from uaclient.conf then run refresh again.

## Desired commit type::

## Checklist:


uaclient/snap.py
```diff
@@ -0,0 +1,54 @@
...
+
+SNAP_CMD = "/usr/bin/snap"
+SNAP_INSTALL_RETRIES = [0.5, 1.0, 5.0]
+HTTP_PROXY_OPTION = "proxy.http"
+HTTPS_PROXY_OPTION = "proxy.https"
+
+
+def is_installed() -> bool:
+    """Returns whether or not snap is installed"""
+    return "snapd" in apt.get_installed_packages()
```
orndorffgrant, line 13: @blackboxsw moving conversation to new PR (decided to create new PR because of significant differences after rebasing off of prompt PR). 

The LivepatchEntitlement does the following on enable:

```python
if not util.which(snap.SNAP_CMD):
    # install snap
elif "snapd" not in apt.get_installed_packages():
    # raise exception
```

So I wasn't sure which check I should use here. Maybe both? What do you think?

uaclient/entitlements/livepatch.py
```diff
@@ -24,37 +24,6 @@
 }
 
 
-def configure_snap_proxy(
```
orndorffgrant, line 27: I moved the snap functions into its own file originally to solve a  circular import problem. We still have a circular import with importing configure livepatch proxy in config.py, but I ended up adding a couple snap functions as well so I decided to keep the separate snap file. Let me know if you'd rather I move all the snap functions back in here

features/attached_commands.feature
```diff
@@ -24,6 +47,44 @@ Feature: Command behaviour when attached to an UA subscription
...
+                https_proxy: ""
+            """
+        And I run `ua refresh config` with sudo
+        Then I will see the following on stdout:
+            """
+            No proxy set in config; however, proxy is configured for: snap, livepatch.
+            See https://discourse.ubuntu.com/t/ubuntu-advantage-client/21788 for more information on ua proxy configuration.
+
+            Successfully processed your ua configuration.
+            """
```
orndorffgrant, line 82: @lucasmoura @mitechie @blackboxsw - let me know what you think of the new output in this scenario

Also, is this the correct link where we will be adding the proxy related documentation?

mitechie: +1, we should probably have a header for proxy doc material in that link but just that link will work for now.

orndorffgrant:
Brought in the changes from `main` and force pushed to hopefully get a green check from jenkins :crossed_fingers:

uaclient/config.py
```diff
@@ -681,6 +680,51 @@ def help(self, name):
         response_dict["help"] = help_ent.help_info
         return response_dict
 
+    def process_config(self):
```
lucasmoura, line 683: Just to be sure, we are treating `apt` differently from `snap`  and `livepatch`  because by default, the proxy configurations will be empty and we don't want to disable any existing `snap`and `livepatch` proxies in that scenario, right ?

orndorffgrant: Right. We treat snap/livepatch differently because we don't want to accidentally remove a manually set proxy for those services. We couldn't remove a manually set apt proxy even if we tried, so we don't have to message about it and can just worry about our own apt.conf.d file.

features/attached_commands.feature
```diff
@@ -24,6 +47,44 @@ Feature: Command behaviour when attached to an UA subscription
            | hirsute |
            | groovy  |
 
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached refresh config in a ubuntu machine vm
```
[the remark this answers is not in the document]
orndorffgrant: Sure that makes sense, I'll move it

lucasmoura (approved):

lucasmoura:
CI error is unrelated to this feature

PR #1695, opened 2021-06-28
status: add account info into json output (SC-152)

lucasmoura:
## Proposed Commit Message
status: add account info into json output

We are now creating a separate account object in the status json output. We are moving all account related info into that object while also adding new fields to into, such as created_at

## Test Steps
Launch a PRO instance and check that the account info we are adding appears when running `ua status --format json`

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -52,6 +52,12 @@ DEFAULT_STATUS = {
         "products": [],
         "tech_support_level": status.UserFacingStatus.INAPPLICABLE.value,
     },
+    "account": {
+        "name": "",
+        "id": "",
+        "created_at": "",
+        "external_account_ids": [],
+    },
 }  # type: Dict[str, Any]
 
 LOG = logging.getLogger(__name__)
@@ -550,8 +556,6 @@ class UAConfig:
         response.update(
             {
                 "attached": True,
-                "account": self.accounts[0]["name"],
-                "account-id": self.accounts[0]["id"],
                 "origin": contractInfo.get("origin"),
                 "notices": self.read_cache("notices") or [],
                 "contract": {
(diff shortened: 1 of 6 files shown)


lucasmoura:
@orndorffgrant Code updated with the same refactoring proposed on the `contracts` PR

blackboxsw (changes requested):
With these types of changes where we are changing the structure of the status.json on disk, We need to probably trigger a re-write of status.json across upgrade path. Otherwise non-root users on an already attached machine will hit stuff like this if they run `ua status` and the new-code doesn't know how to handle the old keys on disk. We have the same type of upgrade path problem in cloud-init,

```bash
ubuntu@dev-x:~$ ua --debug status
DEBUG: Executed with sys.argv: ['/usr/bin/ua', '--debug', 'status']
DEBUG: Reading file: /var/lib/ubuntu-advantage/status.json
DEBUG: Reading file: /var/lib/ubuntu-advantage/notices.json
DEBUG: File does not exist: /var/lib/ubuntu-advantage/notices.json
ERROR: Unhandled exception, please file a bug
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 976, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1049, in main
    return args.action(args, cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 876, in action_status
    output = ua_status.format_tabular(status)
  File "/usr/lib/python3/dist-packages/uaclient/status.py", line 614, in format_tabular
    tech_support_level = status["contract"]["tech_support_level"]
KeyError: 'contract'
Unexpected error(s) occurred.
For more details, see the log: /var/log/ubuntu-advantage.log
To file a bug run: ubuntu-bug ubuntu-advantage-tools
```

lucasmoura:
@blackboxsw Great catch.

I think the optimal solution would be to update the `status.json` file during `postinst`. However,  if we do that we may affect the execution of `upgrade_lts_contract` since it checks `status.json` directly to assert if a service was enable. Therefore, we cannot touch it there.

I think we have some solutions that we can implement here:

1) Support both `status.json` outputs when constructing status output
2) Do not display info in the `status.json` output if we can't find that info. Maybe we could also emit a notice for this situation asking the user to run the command as root.

But if you or @orndorffgrant have any other ideas for that, let me know and we can better discuss a path moving forward here

orndorffgrant (approved):
LGTM! And the migration from old status.json will happen in a later PR

blackboxsw (approved):
LGTM! yes later PR will tackle status schema upgrade issues.

PR #1696, opened 2021-06-28
Cli config subcmd (SC-101)

blackboxsw:
Add `ua config` subcommands "show" "set" and "unset". Set and Unset can only operate on a known subset of keys http_proxy, https_proxy, apt_http_proxy, apt_https_proxy. 

To support this functionality the following additional changes are made:
  - Add `unconfigure_livepatch_proxy` and `unconfigure_snap_proxy` functions
  - Add `UAConfig.write_cfg` to persist config changes into uaclient.conf
  - Add "setters" to all UAConfig *proxy properties which call self.write_cfg to persist changes
  - Supplement existing uaclient.conf with ua_config settings to give documentation breadcrumbs about `ua config` to folks browsing uaclient.conf values
  - Alphabetize ua subcommands to improve discoverability of commands

## Proposed Commit Message
    - cli: add ua config set, unset and show subcommands
    - apt: move APT_CONFIG_PROXY_HTTPS into apt module
    - uaclient.conf: add default ua_config setting values to cfg file
    - livepatch: add unconfigure_livepatch_proxy and snap_proxy
    - config: add write_cfg and property setters for proxy attrs
    - cli: alphabetize subcommands

## Test Steps
```bash
python3 -m uaclient.cli config
python3 -m uaclient.cli config --help
python3 -m uaclient.cli config show
python3 -m uaclient.cli config show --help
python3 -m uaclient.cli config set --help
python3 -m uaclient.cli config unset --help

# run the updated integration tests
UACLIENT_BEHAVE_CONTRACT_TOKEN=<TODO> UACLIENT_BEHAVE_DESTROY_INSTANCES=0 UACLIENT_BEHAVE_BUILD_PR=1 tox -e behave-lxd-18.04 features/proxy_config.feature
```

## Desired commit type::

## Checklist:


uaclient/cli.py
```diff
@@ -477,6 +562,127 @@ def get_valid_entitlement_names(names: "List[str]"):
...
+    @return: 0 on success, 1 otherwise
+    """
+    parser = get_parser()
+    subparser = parser._get_positional_actions()[0].choices["config"]
+    valid_choices = subparser._get_positional_actions()[0].choices.keys()
+    if args.command not in valid_choices:
+        parser._get_positional_actions()[0].choices["config"].print_help()
+        raise exceptions.UserFacingError(
+            "\n<command> must be one of: {}".format(", ".join(valid_choices))
+        )
```
orndorffgrant, line 577: Digging into the parser internals in this case makes me pause here.

Can we just assume that if this action got called, then there must not have been a sub-sub-command provided? and therefore we should always raise the UserFacingError (no need to check if the command is a valid choice)?

Or not, this is just a wishful suggestion

uaclient/cli.py
```diff
@@ -477,6 +562,127 @@ def get_valid_entitlement_names(names: "List[str]"):
...
+
+def action_config_show(args, cfg, **kwargs):
+    """Perform the 'config show' action.
+
+    @return: 0 on success, 1 otherwise
+    """
+    col_width = str(max([len(x) for x in config.UA_CONFIGURABLE_KEYS]) + 1)
+    row_tmpl = "{key: <" + col_width + "}: {value}"
+    for key in config.UA_CONFIGURABLE_KEYS:
+        print(row_tmpl.format(key=key, value=getattr(cfg, key)))
```
orndorffgrant, line 588: This is cool - I haven't used python's padding formatting feature before.

On the other hand, could we get away with just printing yaml? (is that easier or harder than this?)

uaclient/cli.py
```diff
@@ -477,6 +562,127 @@ def get_valid_entitlement_names(names: "List[str]"):
...
+    """
+    from uaclient.apt import setup_apt_proxy
+    from uaclient.entitlements.livepatch import (
+        configure_livepatch_proxy,
+        configure_snap_proxy,
+    )
+
+    parser = get_parser()
+    config_parser = parser._get_positional_actions()[0].choices["config"]
+    subparser = config_parser._get_positional_actions()[0].choices["set"]
```
orndorffgrant, line 605: driveby thought that we shouldn't address now. If we need access to the parser/subparser often in our action functions, maybe we should pass them as args

uaclient/cli.py
orndorffgrant: I see now that the help message relies on having the `print_help()` and `valid_choices` which makes my suggestion not ideal, you can ignore

uaclient/cli.py
```diff
@@ -477,6 +562,127 @@ def get_valid_entitlement_names(names: "List[str]"):
...
+            )
+        )
+    if set_key not in config.UA_CONFIGURABLE_KEYS:
+        subparser.print_help()
+        raise exceptions.UserFacingError(
+            "\n<key> must be one of: {}".format(
+                ", ".join(config.UA_CONFIGURABLE_KEYS)
+            )
+        )
+    # TODO(SC-174: add prompts on proxy changes and honor --assume-yes)
```
orndorffgrant, line 622: I think we can remove this comment :upside_down_face:

uaclient/cli.py
```diff
@@ -477,6 +562,127 @@ def get_valid_entitlement_names(names: "List[str]"):
...
+    if set_key in ("http_proxy", "https_proxy"):
+        kwargs = {set_key: set_value}
+        configure_snap_proxy(**kwargs)
+        # Only set livepatch proxy if livepatch is enabled
+        ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME["livepatch"]
+        entitlement = ent_cls(cfg)
```
orndorffgrant, line 628: Since we know exactly what entilement we want here, it might be more straightforward to just do something like
```
entitlement = entitlements.livepatch.LivepatchEntitlement(cfg)
```

uaclient/cli.py
```diff
@@ -477,6 +562,127 @@ def get_valid_entitlement_names(names: "List[str]"):
...
+            "\n<key> must be one of: {}".format(
+                ", ".join(config.UA_CONFIGURABLE_KEYS)
+            )
+        )
+    if args.key in ("http_proxy", "https_proxy"):
+        protocol_type = args.key.split("_")[0]
+        unconfigure_snap_proxy(protocol_type=protocol_type)
+        # Only unset livepatch proxy if livepatch is enabled
+        ent_cls = entitlements.ENTITLEMENT_CLASS_BY_NAME["livepatch"]
+        entitlement = ent_cls(cfg)
```
orndorffgrant, line 671: Same comment as above

uaclient/cli.py
```diff
@@ -477,6 +562,127 @@ def get_valid_entitlement_names(names: "List[str]"):
...
+        if livepatch_status == ua_status.ApplicationStatus.ENABLED:
+            configure_livepatch_proxy(**kwargs)
+    elif set_key in ("apt_http_proxy", "apt_https_proxy"):
+        # setup_apt_proxy is destructive for unprovided values. Source complete
+        # current config values from uaclient.conf before applying set_value.
+        kwargs = {
+            "http_proxy": cfg.apt_http_proxy,
+            "https_proxy": cfg.apt_https_proxy,
+        }
+        kwargs[set_key] = set_value
```
orndorffgrant, line 639: This looks like a bug, `set_key` needs the "apt_" chopped off like we do below in unset.

uaclient/config.py
```diff
@@ -665,6 +700,37 @@ def help(self, name):
...
+            content += "{key}: '{value}'\n".format(
+                key=attr, value=getattr(self, attr)
+            )
+        content += "data_dir: {}\n".format(self.data_dir)
+        if "log_level" in self.cfg:
+            log_level = self.cfg["log_level"].lower()
+        else:
+            log_level = CONFIG_DEFAULTS["log_level"]
+        content += "log_level: {}\n".format(log_level)
+        content += "log_file: {}\n".format(self.log_file)
```
orndorffgrant, line 719: Can we use `yaml.dump` for everything instead of hand-crafting this section?

uaclient/config.py
```diff
@@ -665,6 +700,37 @@ def help(self, name):
...
+                "http_proxy": self.http_proxy,
+                "https_proxy": self.https_proxy,
+                "apt_http_proxy": self.apt_http_proxy,
+                "apt_https_proxy": self.apt_https_proxy,
+            }
+        }
+        content += "\n# Managed with `sudo ua config`\n" + yaml.dump(ua_config)
+        for attr in ("features", "settings_overrides"):
+            if self.cfg.get(attr):
+                content += "\n" + yaml.dump({attr: self.cfg.get(attr)})
```
orndorffgrant, line 731: Do these two need to be handled with separate `yaml.dump` calls? Can we construct a dict and call `yaml.dump` on the whole thing?

Note that we can use `default_flow_style=False` when we have nested dicts: https://pyyaml.org/wiki/PyYAMLDocumentation#dictionaries-without-nested-collections-are-not-dumped-correctly

orndorffgrant (commented):
Halfway through code review and taking a break for lunch. This is big and looking really good!

lucasmoura (commented):
@blackboxsw I think we should also add integration tests for this feature. We could reuse one of the existing scenarios instead of creating a new one just for this new command.

Also, should we allow users to configure `features_overrides`, like `allow_beta` through this cli command ?

uaclient/cli.py
```diff
@@ -198,6 +198,91 @@ def auto_attach_parser(parser):
...
+    """Build or extend an arg parser for config subcommand."""
+    parser.usage = USAGE_TMPL.format(name=NAME, command="config <command>")
+    parser.prog = "config"
+    parser.description = (
+        "Manage Ubuntu Advantage configuration on this machine"
+    )
+    parser._optionals.title = "Flags"
+    subparsers = parser.add_subparsers(
+        title="Available Commands", dest="command", metavar=""
+    )
```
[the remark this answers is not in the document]
blackboxsw: We do have the option of subparsers.required = True as we do in the main subcommands setup. But, I didn't like the output in that case. Here's the way argparse handles a missing required subparser
```bash
# With subparsers.required = True
$ python3 -m uaclient.cli config
usage: ua config <command> [flags]
the following arguments are required: 
```

# Without subparsers.required = True we get to (better handle our metavar usage output formatting in the action_config function)[https://github.com/canonical/ubuntu-advantage-client/pull/1696/files#diff-d558bdbcb7ee1bef85fd138367f1e64708ea8271c09cf8e4c19941c24c8883d6R573]\
```bash
$ python3 -m uaclient.cli config
usage: ua config <command> [flags]

Manage Ubuntu Advantage configuration on this machine

Flags:
  -h, --help  show this help message and exit

Available Commands:
  
    show      show all Ubuntu Advantage configuration settings
    set       set Ubuntu Advantage configuration settings
    unset     unset Ubuntu Advantage configuration settings

<command> must be one of: show, set, unset

```

uaclient/cli.py
blackboxsw: This is a good suggestion, I toyed with this thought too and wondered about a slight refactor for that. This approach I took  is definitely bad practice as I'm trying to reference private methods on the parser which I assume could change without warning in the future and break us.

uaclient/cli.py
blackboxsw: Good catch, and looks like I have a unittest gap. Will fix.

blackboxsw:
> @blackboxsw I think we should also add integration tests for this feature. We could reuse one of the existing scenarios instead of creating a new one just for this new command.
> 
> Also, should we allow users to configure `features_overrides`, like `allow_beta` through this cli command ?

I confirmed with @mitechie that we only want to manage common user-configurable options via `ua config` at this iteration. Any allow_beta of custom feature flag managment is likely atypical use-cases and would warrant a bit more "expert" configuration of uaclient.conf directly.

uaclient/cli.py
lucasmoura: We could instead store the valid choices in a static variable. This will avoid the need of using a private method from argparser to discover that.

uaclient/config.py
lucasmoura: I agree that we could easily run `yaml.dump` here, but this will make some things a little bit harder:

1) We will loose the comments we are making in the file. Even though the header comment is easy to add with the `yaml.dump` solution, comments in the middle may not be as easy. But maybe we don't need those comments in the middle.
2) This code is removing any invalid configuration from `uaclient.conf`, like user configurations with typos. However, since we expect users to not directly interact with the file and those incorrect configs will not affect uaclient, maybe this is not needed.
3) This code is keeping the order of the keys. Maybe we can sort the keys in an order that we want, before using it on `yaml.dump`.

Those are not strong arguments to not use `yaml.dump`, just some points. If we can address them, I think we should use `yaml.dump`.

Am I missing something here @blackboxsw ?

uaclient/config.py
blackboxsw: Switched to use default_flow_style=False here. Thanks for the tip.

uaclient/cli.py
blackboxsw: The confusing motivation of this code is actually two fold (in differnt functions):
  1. to check valid keys() which we could easily persist somewhere as a static var or lookup function
  2. to call that subparser.print_help().   
  
Maybe for case 2 we actually   collect a global of arg_subparsers keyed on the subcommand "ua.config.set" and then we don't have to look it up using private methods. Again this is probably a separate PR, but we can discuss depending on what @lucasmoura and @orndorffgrant think.

uaclient/config.py
blackboxsw: Nothing missed other than the intent leave some minor documentation breadcrumb comments in /etc/ubuntu-advantage/uaclient.conf so people editing that file directly are alerted to the fact that there are better places (more synchronous places) to make those changes. Filed and issue to track this outstanding discussion point to make sure we have a public-facing decision on what we ultimately decide to to here.
https://github.com/canonical/ubuntu-advantage-client/issues/1702

uaclient/cli.py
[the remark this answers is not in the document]
orndorffgrant: Very good point, that is much nicer!

uaclient/cli.py
orndorffgrant: Yep, that makes sense. :+1:

uaclient/cli.py
orndorffgrant: I created a Jira issue for us to come back to this here: https://warthogs.atlassian.net/browse/SC-205

Resolving this now

uaclient/cli.py
orndorffgrant: Yep the print_help and valid_choices make for a much better UX. https://warthogs.atlassian.net/browse/SC-205 will address the use of parser private methods

Resolving

orndorffgrant:
@blackboxsw looks like you refactored `configure_snap_proxy` to use kwargs instead of positional args. And that got lost when rebasing on my snap.py refactor (sorry thats causing trouble :grimacing:), which is causing errors

uaclient/config.py
```diff
@@ -737,6 +785,26 @@ def process_config(self):
...
             )
 
+    def write_cfg(self, config_path=None):
+        """Write config values back to config_path or DEFAULT_CONFIG_FILE."""
+        if not config_path:
+            config_path = DEFAULT_CONFIG_FILE
+        content = status.MESSAGE_UACLIENT_CONF_HEADER
+        cfg_dict = copy.deepcopy(self.cfg)
+        if "log_level" not in cfg_dict:
+            cfg_dict = CONFIG_DEFAULTS["log_level"]
```
orndorffgrant, line 795: @blackboxsw I think this is supposed to set the "log_level" key value
```suggestion
            cfg_dict["log_level"] = CONFIG_DEFAULTS["log_level"]
```

blackboxsw: +1 thanks

uaclient/snap.py
```diff
@@ -16,7 +18,7 @@ def is_installed() -> bool:
 def configure_snap_proxy(
     http_proxy: Optional[str],
     https_proxy: Optional[str],
```
orndorffgrant, line 20:
```suggestion
    http_proxy: Optional[str] = None,
    https_proxy: Optional[str] = None,
```

blackboxsw: Thanks for directing me here. Yes I had botched the merge conflicts and dropped that changeset in the merge.

features/proxy_config.feature
```diff
@@ -21,8 +21,8 @@ Feature: Proxy configuration
         .*CONNECT contracts.canonical.com.*
         """
         When I run `truncate -s 0 /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        And I configure uaclient `apt_http` proxy to use `proxy` machine
-        And I configure uaclient `apt_https` proxy to use `proxy` machine
+        And I run `ua config set apt_http_proxy=http://<ci-proxy-ip>:3128` with sudo
+        And I run `ua config set apt_https_proxy=http://<ci-proxy-ip>:3128` with sudo
         And I verify `/var/log/squid/access.log` is empty on `proxy` machine
         Then I verify that no files exist matching `/etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy`
```
lucasmoura, line 27: Since we are running `ua config set` now, this will actually create the apt proxy file ant the test will fail

features/proxy_config.feature
```diff
@@ -51,15 +52,11 @@ Feature: Proxy configuration
         """
         .*GET.*ubuntu.com/ubuntu/dists.*
         """
-        When I append the following on uaclient config
-        """
-        ua_config:
-            apt_http_proxy: ""
-            apt_https_proxy: ""
-        """
+        And I run `ua config unset apt_http_proxy` with sudo
```
lucasmoura, line 55: I think this should be a `When` step

features/proxy_config.feature
orndorffgrant: Aye, and related: we may need separate tests for `ua config set` vs "edit config file then run `ua refresh config`"

lucasmoura (approved):

lucasmoura:
CI error is unrelated to this change

PR #1697, opened 2021-06-29
changelog: release 27.1

orndorffgrant:
Just bringing in the changelog entry for 27.1 that is currently only on the `release-27` branch


lucasmoura (approved):

blackboxsw:
While it's a new intermittent error type, it's completely unrelated to this PR.

First time I've seen this one:
`      botocore.exceptions.ClientError: An error occurred (IncorrectInstanceState) when calling the StartInstances operation: The instance 'i-0c088d9ce75cffddc' is not in a state from which it can be started.`

PR #1698, opened 2021-06-30
proxy: validate urls before setting as proxies (SC-170)

orndorffgrant:
## Proposed Commit Message

> proxy: validate urls before setting as proxies
>
> Introduces a proxy validation function that checks two things:
>
> 1. if the proxy value is a parseable url with http or https scheme
> 2. if the proxy can be used to reach important hardcoded destinations
> 
> If either check fails, the current operation is aborted and an error
message is printed.

## Test Steps

1. Configure ua to use an invalid url as a proxy. Ensure enabling any service fails.
2. Configure ua to use a valid url that doesn't point to a functioning proxy. Ensure enabling any service fails.
3. Configure ua to use a valid url that points to a functioning proxy. Ensure enabling any service succeeds.

## Desired commit type::

## Checklist:


uaclient/util.py
```diff
@@ -806,3 +807,34 @@ def parse_rfc3339_date(dt_str: str) -> datetime.datetime:
...
+
+
+def validate_proxy(
+    protocol: str, proxy: Optional[str], test_url: str
+) -> Optional[str]:
+    if not proxy:
+        return None
+
+    if not is_service_url(proxy):
+        print("{} is not a valid url. Not setting as proxy.".format(proxy))
```
blackboxsw, line 819: When encountering invalid I'd prefer the approach of stopping the line with loud UserFacingErrorr/log errors. In cloud-init we are too flexible and try as hard as we can to ignore non-critical errors, which leads to much later recognition that something is misconfigured on the system. The more strict we are, the more time is saves a customer having invalid config recognized.

uaclient/util.py
```diff
@@ -806,3 +807,34 @@ def parse_rfc3339_date(dt_str: str) -> datetime.datetime:
     return datetime.datetime.strptime(
         dt_str_without_microseconds, "%Y-%m-%dT%H:%M:%S%z"
     )
+
+
+def validate_proxy(
+    protocol: str, proxy: Optional[str], test_url: str
+) -> Optional[str]:
```
blackboxsw, line 814: Validate proxy in my mind should succeed or raise exceptions if proxy values are provided. If no values are provided, a return None is in order. I'm welcome to persuasion otherwise.

blackboxsw (changes requested):
suggestion for discussion

uaclient/util.py
orndorffgrant: Yep this is something I wanted to bring up/talk about - this was just my first draft. I do like the "fail fast and loud" approach when appropriate.

Thinking about it more, I think it makes sense here to fail loudly, because just like you say, if we continue after this error then we will try to use entitlements without the proxies in working order. The user specifically configured ua to use a proxy, so continuing operations while effectively ignoring the proxy would be unexpected/undesirable. 

We can chat more about this but I'll probably go ahead and change it to raise a UserFacingError

lucasmoura (commented):
LGTM

Just waiting for CI to see if the tests are working as expected

orndorffgrant:
@lucasmoura Thanks for the review! We can't actually merge quite yet because I need to add calls to the validate function as part of `ua config refresh` which was just merged earlier today (thanks for that too :))

lucasmoura:
@orndorffgrant fair enough, makes total sense

orndorffgrant:
@blackboxsw @lucasmoura I added the validate calls to the `ua refresh config` function and updated the integration tests, so this should be ready for review. 

Note that we'll want to call this validation as a part of `ua config set` as well, so if we merge #1696 first, then I'll add that in this PR. Or if we merge this PR first, then we'll need to add validate calls to #1696 before merging.

lucasmoura (approved):

blackboxsw (approved):
+ 1 Let's land this now as-is since ths is great & works well, and I have some rebase fixes to re-apply to #1696 anyway. I'll sort both after this merges.

blackboxsw:
CI failures were lxd zfs destroy storage issues.

PR #1699, opened 2021-06-30
status: do not modify internal status state

lucasmoura:
## Proposed Commit Message
status: do not modify internal status state

We have modified our internal status state to better reflect the contract structure we want to output in the json format. However,
by modifying the internal we end up affecting non-root users. That is because when a non-root user runs ua status, we get the status directly from the file saved on disk. If the user has not run any root commands after installing a new version of uaclient, that file on disk will be stale and will not be able to be used by the status formatting code. Because of that, we are bringing back the original internal status representation.

## Test Steps
Launch a container and verify that the contract info appears when running `ua status --format json` for an unattached and attached system

## Desired commit type::

## Checklist:


lucasmoura:
We won't bring back the original status state anymore.

PR #1703, opened 2021-07-02
status: allow output to be in yaml format (SC-158)

lucasmoura:
## Proposed Commit Message
status: allow output to be in yaml format

We are now allowing users to specify the yaml format when looking at the status output

## Test Steps
Launch a machine and test running `ua status --format yaml` on an unattached an attached uaclient

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -655,9 +655,7 @@ def format_tabular(status: "Dict[str, Any]") -> str:
     return "\n".join(content)
 
 
-def format_json_status(status: "Dict[str, Any]") -> str:
-    from uaclient.util import DatetimeAwareJSONEncoder
-
+def _format_status_output(status: "Dict[str, Any]") -> "Dict[str, Any]":
     status["environment_vars"] = [
         {"name": name, "value": value}
         for name, value in sorted(os.environ.items())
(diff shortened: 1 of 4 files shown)


uaclient/status.py
```diff
@@ -678,5 +676,18 @@ def format_json_status(status: "Dict[str, Any]") -> str:
...
+    return json.dumps(
+        _format_status_output(status), cls=DatetimeAwareJSONEncoder
+    )
+
+
+def format_yaml_status(status: "Dict[str, Any]") -> str:
+    import yaml
 
-    return json.dumps(status, cls=DatetimeAwareJSONEncoder)
+    return yaml.dump(_format_status_output(status))
```
lucasmoura, line 693: It seems that yaml is already parsing dates in an appropriate format. However, if you think we should make the data parsing works exactly as the json definition, let me know and I can code a loader for it

uaclient/status.py
```diff
@@ -678,5 +676,18 @@ def format_json_status(status: "Dict[str, Any]") -> str:
...
+def format_json_status(status: "Dict[str, Any]") -> str:
+    from uaclient.util import DatetimeAwareJSONEncoder
+
+    return json.dumps(
+        _format_status_output(status), cls=DatetimeAwareJSONEncoder
+    )
+
+
+def format_yaml_status(status: "Dict[str, Any]") -> str:
+    import yaml
```
lucasmoura, line 691: We need to make this import here because on `setup.py` we indirectly import this module before installing the package dependencies, which breaks the build if the import is made on the beginning of the file

orndorffgrant (commented):
This looks great!

The datetime output looks good. Unlike in JSON, where theres only a "de facto" standard of using iso8601 strings for dates, YAML has an official regexp for datetimes (https://yaml.org/type/timestamp.html) and the datetime output from pyyaml is valid.

My only recommendation before merging is that we use `default_flow_style=False` as described here (https://pyyaml.org/wiki/PyYAMLDocumentation#dictionaries-without-nested-collections-are-not-dumped-correctly) so that the services list looks nicer

lucasmoura:
@orndorffgrant code updated

orndorffgrant (approved):

blackboxsw (approved):
+1 lookin' good.

PR #1704, opened 2021-07-02
status: update root level keys of json status (SC-150)

orndorffgrant:
## Proposed Commit Message

```
status: update root level keys of json status

* changes expires to null when not present instead of "n/a"
* ensures expires is always iso format
* adds _schema_version
* adds version
* adds machine_id
* adds effective
* changes configStatus to execution_status
* changes configStatusDetails to execution_details
* changes services[].statusDetails to services[].status_details
```

## Test Steps

Run `ua status --format=json` as {nonroot, root} {unattached, attached} and ensure changes described in commit message are applied appropriately.

@blackboxsw @lucasmoura some of these may be breaking changes. If they are then I can revert them for now. Do you know if desktop team (or anyone else) relies on the old camelCase keys or on the expires value "n/a"?

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -880,9 +880,9 @@ def action_status(args, cfg):
     show_beta = args.all if args else False
     status = cfg.status(show_beta=show_beta)
     active_value = ua_status.UserFacingConfigStatus.ACTIVE.value
-    config_active = bool(status["configStatus"] == active_value)
+    config_active = bool(status["execution_status"] == active_value)
     if args and args.wait and config_active:
-        while status["configStatus"] == active_value:
+        while status["execution_status"] == active_value:
             print(".", end="")
             time.sleep(1)
             status = cfg.status(show_beta=show_beta)
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -9,7 +9,7 @@ import sys
 import yaml
 from collections import namedtuple, OrderedDict
 
-from uaclient import apt, exceptions, snap, status, util
+from uaclient import apt, exceptions, snap, status, util, version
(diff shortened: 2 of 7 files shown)


uaclient/cli.py
```diff
@@ -880,9 +880,9 @@ def action_status(args, cfg):
     show_beta = args.all if args else False
     status = cfg.status(show_beta=show_beta)
     active_value = ua_status.UserFacingConfigStatus.ACTIVE.value
-    config_active = bool(status["configStatus"] == active_value)
+    config_active = bool(status["execution_status"] == active_value)
```
lucasmoura, line 883: This is another part of the schema that we will need to handle in the `postinst` schema patch script

uaclient/config.py
```diff
@@ -550,10 +556,13 @@ def _attached_status(self) -> "Dict[str, Any]":
         from uaclient.entitlements import ENTITLEMENT_CLASSES
 
         response = copy.deepcopy(DEFAULT_STATUS)
-        contractInfo = self.machine_token["machineTokenInfo"]["contractInfo"]
+        machineTokenInfo = self.machine_token["machineTokenInfo"]
```
lucasmoura, line 559: Should we redact that info on the output if a non-root user runs that command ?

orndorffgrant: We're only grabbing the `machineId` which I think is okay for non-root users? We decided not to include the `machineToken` because it is a secret.

lucasmoura: Fair enough, the `machineToken` was the one I was actually thinking about

lucasmoura (approved):

PR #1705, opened 2021-07-02
status: add config info in the status output

lucasmoura:
## Proposed Commit Message
status: add config info in the status output

We are now adding config info in the status output. This will be reflected whenener the user runs ua status with the --format flag

## Test Steps
Launch a machine and check if the output of `ua status --format json` has the config keys. Also, check if changing the path to load the config file from will reflect in ua status

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -453,9 +453,7 @@ class UAConfig:
 
         return new_response
 
-    def _get_config_status(
-        self
-    ) -> "Dict[str, Union[str, List[Tuple[str,str]]]]":
+    def _get_config_status(self) -> "Dict[str, Any]":
         """Return a dict with execution_status, execution_details and notices.
 
             Values for execution_status will be one of UserFacingConfigStatus
@@ -488,6 +486,10 @@ class UAConfig:
             "execution_status": status_val,
             "execution_details": status_desc,
             "notices": notices,
+            "config_path": self.cfg.get("config_path"),
+            "config": {
+                k: v for k, v in self.cfg.items() if k != "config_path"
+            },
         }
 
     def _unattached_status(self) -> "Dict[str, Any]":
@@ -817,6 +819,7 @@ def parse_config(config_path=None):
     cfg.update(env_keys)
     cfg["log_level"] = cfg["log_level"].upper()
     cfg["data_dir"] = os.path.expanduser(cfg["data_dir"])
+    cfg["config_path"] = config_path
     for key in ("contract_url", "security_url"):
         if not util.is_service_url(cfg[key]):
             raise exceptions.UserFacingError(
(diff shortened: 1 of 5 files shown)


uaclient/config.py
```diff
@@ -793,6 +793,7 @@ def parse_config(config_path=None):
     cfg.update(env_keys)
     cfg["log_level"] = cfg["log_level"].upper()
     cfg["data_dir"] = os.path.expanduser(cfg["data_dir"])
+    cfg["config_path"] = config_path
```
orndorffgrant, line 796: putting the config_path in here feels a little weird. If we have time, could we refactor slightly so config path is determined up in __init__ and set as an attribute on the UAConfig instance rather than in the cfg dict? Just an idea, feel free to disagree

lucasmoura: I don't think it is that weird since the `config_path` can in fact be configured from the outside. However, I can do the refactoring if needed. @blackboxsw do you share @orndorffgrant feeling that this variable should not be include in the `config` dict ?

orndorffgrant: True that it can be configured from the outside. The slight difference is that the rest of the contents of `cfg` are configurable via uaclient.conf, while `config_path` isn't.

I agree that it is not that weird though, lets not hold up this PR on this.

uaclient/status.py
```diff
@@ -682,18 +684,28 @@ def _format_status_output(status: "Dict[str, Any]") -> "Dict[str, Any]":
 
     # We don't need the origin info in the json output
     status.pop("origin", "")
+
+    status["config_path"] = cfg["config_path"]
+
+    # We could instead delete the key from the cfg dict.
+    # We could copy the dict and delete the key, but this
+    # approach is simpler
+    status["config"] = {k: v for k, v in cfg.items() if k != "config_path"}
```
orndorffgrant, line 693: I agree with the more "functional" style of creating a new dict based on the old one, rather than doing a full copy and mutating the copy. +1

> Random aside: sometimes dict/list comprehensions are hard to read, but this is one of those cases where it perfectly expresses a map, filter, and reduce all in one succinct line. Love it

My only thought to consider is that we're currently passing `cfg` to this function to do this here, but we construct most of the data in `config.py` where we already have `cfg` available. What are your thoughts on adding the config data here vs in `UAConfig._get_config_status()`?

lucasmoura: That's a good idea, I will refactor the code to use it

lucasmoura:
@orndorffgrant code updated

orndorffgrant (approved):
Thanks for the updates! This looks great and worked for me. Lets just let Jenkins approve before merging

lucasmoura:
I could not reproduce that failing test locally. I will merge this PR as is and take a look on the daily CI

PR #1706, opened 2021-07-05
mypy: add packaging constraint

lucasmoura:
## Proposed Commit Message
mypy: add packaging constraint

packaging was recently updated to a version that uses type hints. This is not supported on python 3.5, which is a version we are running mypy tests on. We are now freezing packaging to a version that doesn't break mypy checks for python 3.5

## Test Steps
Run mypy test without this change and see the python 3.5 error. Apply this change and confirm that the problem is fixed

## Desired commit type::

## Checklist:

### DIFF
--- a/tools/constraints-mypy.txt
+++ b/tools/constraints-mypy.txt
@@ -1,3 +1,4 @@
 mypy
 pytest==6.1.2
 importlib-metadata==3.3.0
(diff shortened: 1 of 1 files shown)


orndorffgrant (approved):
Thanks for noticing and fixing this!

PR #1707, opened 2021-07-06
tests: do not prompt when installing uaclient

lucasmoura:
## Proposed Commit Message
tests: do not prompt when installing uaclient

We have recently changed uaclient.conf config file. On PRO test, we have custom config to block auto-attach. Because of that,
when we install the new uaclient package, we are prompted about config changes. We are now updating our install command to not prompt under those circumstances

## Test Steps
Run the azurepro-16.04 integration test without this change and confirm that uaclient is not properly installed. Add this change and verify that the test now succeeds as expected 

## Desired commit type::

## Checklist:

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -804,8 +804,19 @@ def _install_uat_in_container(
             inst.push_file(deb_file, "/tmp/" + deb_name)
 
         cmds.append(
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
+1 pro tests succeeded here with that invocation during `apt install`. It's also the same env/cmdline we use when installing packages for any repo-based services.

PR #1709, opened 2021-07-06
Status schema patch

blackboxsw:
Draft branch for feedback on schema backward compatibility upgrade_downgrade support.
Also a mechanical driveby commit for simpifying the typing support since we now don't support trusty.

## Proposed Commit Message
```bash
    schema: wip branch to schema patch maintenance and define version 1.0
    
    Add schema.maintenance module and initial status schema version 1.0.
    
    Schema patch files will be defined as python modules named
       uaclient.schema.status_<version_float>.py for each new backward
       incompatible schema version defined.
    
    A status.json schema is considered incompatible when any of the
    following occurs:
     - an existing key in status.json is dropped
     - the value of an existing key changes type
    
    Previous PRs have broken status.json compatibility by:
     - adapting "account" key from a string to a dict
     - dropping "subscription-id" and "subscription" keys to put them
       under a new "contract" key
    
    This PR adds the first schema/status_1.py patch file and simple
    utility functions in schema.maintenance to ensure a complete
    backward-compatible schema is presented in status JSON under a new
    "schemas" key for each supported version.
    
    Subsequent PRs will be expected to support querying specific schema
    version via ua status --schema-version X.Y --format=json
    
    Each patch file at uaclient.schema.status_1.py will define:
      - an 'up' function accepting UAConfig and status_dict which accepts
        only the previous schema version, migrate incompatible keys
        and return the patched schema version dict defined in this module
      - a 'down' function which will provide a downgrade patch reversing
        the schema upgrade operations and returning the expected previous
        schema version dict
```

## Test Steps
```bash
sudo python3 -c 'from uaclient.config import UAConfig; UAConfig().status()'

# review the new backward compatible schemas key
cat /var/lib/ubuntu-advantage/status.json | jq ".schemas"
```

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -37,7 +37,7 @@ except ImportError:
 DEFAULT_STATUS = {
     "_doc": "Content provided in json response is currently considered"
     " Experimental and may change",
-    "_schema_version": "0.1",
+    "_schema_version": "1.0",
     "version": version.get_version(),
     "machine_id": None,
     "attached": False,
@@ -670,6 +670,7 @@ class UAConfig:
 
         Write the status-cache when called by root.
         """
+        from uaclient.schema.maintenance import get_compat_schema
 
         if os.getuid() != 0:
             response = cast("Dict[str, Any]", self.read_cache("status-cache"))
@@ -680,6 +681,7 @@ class UAConfig:
         else:
             response = self._attached_status()
         response.update(self._get_config_status())
+        response = get_compat_schema(self, response)
         if os.getuid() == 0:
             self.write_cache("status-cache", response)
 
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -893,7 +893,9 @@ def parse_config(config_path=None):
                 "Invalid url in config. {}: {}".format(key, cfg[key])
             )
     # log about invalid keys before ignoring
-    for key in sorted(set(cfg.keys()).difference(VALID_UA_CONFIG_KEYS)):
+    for key in sorted(
+        set(cfg.keys()).difference(VALID_UA_CONFIG_KEYS + ("config_path",))
+    ):
         logging.warning(
             "Ignoring invalid uaclient.conf key: %s=%s", key, cfg.pop(key)
         )
--- a/uaclient/schema/maintenance.py
+++ b/uaclient/schema/maintenance.py
@@ -1,0 +1,68 @@
+"""General maintenance functions to manipulate schema versions."""
+
+from typing import Dict
+
+import copy
+import glob
+import importlib
+import logging
+import os
+import re
+
+from uaclient.config import UAConfig
+
+
+def get_patches_by_version(glob_path_str: str) -> Dict[float, str]:
+    """Return a dict of patch files from dir_name keyed by patch version."""
+    patches_dict = {}
+    for patch_path in glob.glob(glob_path_str):
+        match = re.match(r".*\/status_(?P<ver>[.\d]+)\.py", patch_path)
+        if not match:
+            logging.warning("Ignoring schema patch file {}".format(patch_path))
+            continue
+        patch_file = os.path.basename(patch_path)
+        patches_dict[float(match.groupdict()["ver"])] = patch_file.replace(
+            ".py", ""
+        )
+    return patches_dict
+
+
+def get_compat_schema(cfg: UAConfig, status_dict: Dict):
+    """Read status json from status_file, apply all patches and write it."""
+    status = copy.deepcopy(status_dict)
+    patches = get_patches_by_version(
+        "{}/status*py".format(os.path.dirname(__file__))
+    )
+    # schemas key will contain backward compat schema versions for old clients
+    schemas = status.pop("schemas", {})
+    if "_schema_version" in status:
+        current_ver = float(status["_schema_version"])
+    else:
+        current_ver = 0.1
+    # Preserve current version in under "schemas" key
+    schemas[current_ver] = copy.deepcopy(status)
+
+    downgrade_callables = {}
+    # Apply patches and preserve each new schema version
+    for patch_ver in sorted(patches):
+        mod = importlib.import_module(
+            "uaclient.schema.{}".format(patches[patch_ver])
+        )
+        if patch_ver > current_ver:
+            # preserve current_ver and apply patches
+            upgrade = getattr(mod, "up")
+            schemas[patch_ver] = upgrade(cfg, schemas[current_ver])
+            current_ver = patch_ver
+        else:
+            downgrade_callables[patch_ver] = getattr(mod, "down")
+    # Preserve backwards-compatible earlier versions of status in "schemas"
+    for downgrade_ver in sorted(downgrade_callables, reverse=True):
+        import pdb
+
+        pdb.set_trace()
+        downgrade = downgrade_callables[downgrade_ver]
+        downgrade_status = downgrade(cfg, schemas[downgrade_ver])
+        schemas[downgrade_status["_schema_version"]] = downgrade_status
+
+    status["schemas"] = schemas
+    return status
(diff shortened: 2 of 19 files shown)


uaclient/schema/status_1.py
```diff
@@ -0,0 +1,132 @@
...
+    new_status["_schema_version"] = PATCH_VERSION
+    if current_status["attached"]:
+        token = cfg.machine_token["machineTokenInfo"]
+        account_info = token["accountInfo"]
+        contract_info = token["contractInfo"]
+        external_account_ids = []
+        for external_account in account_info.get("externalAccountIds", []):
+            external_account_ids.append(
+                {
+                    "ids": external_account.get("IDS", []),
```
lucasmoura, line 60: We are not making that key translation in the current code `(IDS -> ids)`, but maybe we should

uaclient/schema/status_1.py
```diff
@@ -0,0 +1,132 @@
...
+        ), "Found unexpected key: {} cannot patch status.json.".format(key)
+    for key in DROPPED_KEYS + CHANGED_KEYS:
+        assert (
+            key in current_status
+        ), "Missing expected key: {}. Cannot patch status.json".format(key)
+        new_status.pop(key)
+    new_status["_schema_version"] = PATCH_VERSION
+    if current_status["attached"]:
+        token = cfg.machine_token["machineTokenInfo"]
+        account_info = token["accountInfo"]
```
lucasmoura, line 54: we could use `cfg.accounts` here, right ?

uaclient/schema/status_1.py
```diff
@@ -0,0 +1,132 @@
...
+from typing import Any, Dict
+
+NEW_KEYS = ("contract",)
+DROPPED_KEYS = ("subscription", "subscription-id")
+CHANGED_KEYS = ("account",)  # keys where the value type has changed
+
+PATCH_VERSION = float(re.sub(r".*status_([.\d]+).py", r"\1", __file__))
+
+
+def up(cfg: UAConfig, current_status: Dict) -> Dict:
```
lucasmoura, line 35: As mentioned on this [PR](https://github.com/canonical/ubuntu-advantage-client/pull/1704#discussion_r663167762) there are more keys that we need to update on the old status json representation

uaclient/schema/status_1.py
```diff
@@ -0,0 +1,132 @@
...
+    return new_status
+
+
+def down(_cfg: UAConfig, current_status: Dict[str, Any]) -> Dict[str, Any]:
+    """Downgrade to previous status schema version.
+
+    :return: Dict of downgraded status schema version.
+    """
+    new_status = copy.deepcopy(current_status)
+    import pdb
```
lucasmoura, line 105: We can remove that

uaclient/schema/maintenance.py
```diff
@@ -0,0 +1,70 @@
...
+    for patch_ver in sorted(patches):
+        mod = importlib.import_module(
+            "uaclient.schema.{}".format(patches[patch_ver])
+        )
+        if patch_ver > current_ver:
+            # preserve current_ver and apply patches
+            upgrade = getattr(mod, "up")
+            schemas[patch_ver] = upgrade(cfg, schemas[current_ver])
+            current_ver = patch_ver
+        else:
```
lucasmoura, line 58: if `patch_ver == current_ver`, should we downgrade it ?

lucasmoura: Also, I think I am missing something here. This is the code that will run during `postinst`, right ? If so, why would we need to have all possible status representations ? Shouldn't we just sequentially apply the patches until we are in the right version and write it back to disk ?

uaclient/schema/status_1.py
blackboxsw: Yes we could. but we need to be explicit about what specific keys we are using during the patch up or down because tip of `UAConfig.accounts` could drift as we patch the schema to newer versions, so it may not be able to reference the key properly once tip moves to schema version 2 or 3 because the upstream key may have changed. We can only rely on the key as defined in the current version we are patching.

renanrodrigo:
Quick question: when do we consider it non-compatible? Is it just about the 1st order keys?
Let's say we change the dict in the `account` dict in the future - it's still a dict, though a different one; do we consider it to be the same schema?

orndorffgrant:
@renanrodrigo

I think that technically, a breaking/noncompatible change would be any field (or subfield) of the json that our python is hardcoded to to read in one version that either doesn't exist or had a different name in a previous version. For instance, we specifically look at the "name" field inside the `account` dict for our human readable status output, so that would count.

Generally though, I would say any structure change is worth a migration. Where "any structure change" could be a name change, addition, or removal of a field or subfield - or a change in the type of a value (e.g. object -> list of objects, or a string-enum with one set of values to another)

renanrodrigo:
@orndorffgrant I would consider not only our python code but the whole JSON structure, as you stated, mostly if there might be use cases where a 3rd party script, customer-side, reads this data.

orndorffgrant:
@renanrodrigo Great point!

uaclient/schema/status_1.py
```diff
@@ -0,0 +1,132 @@
...
+
+from uaclient.config import UAConfig
+
+from typing import Any, Dict
+
+NEW_KEYS = ("contract",)
+DROPPED_KEYS = ("subscription", "subscription-id")
+CHANGED_KEYS = ("account",)  # keys where the value type has changed
+
+PATCH_VERSION = float(re.sub(r".*status_([.\d]+).py", r"\1", __file__))
```
orndorffgrant, line 32: I think this is amazing and wonderful and also funny how complicated the code is to avoid writing duplicating the version number between the file name and this variable :)

uaclient/config.py
```diff
@@ -678,6 +679,7 @@ def status(self, show_beta=False) -> "Dict[str, Any]":
         else:
             response = self._attached_status()
         response.update(self._get_config_status())
+        response = get_compat_schema(self, response)
```
orndorffgrant, line 682: This is agreeing with Lucas' comment - Our migration is useful only for the following scenario.
* root user runs `ua status`
* uaclient is upgraded
* migration is run as root
* non-root user tries to run `ua status`

If we run it here, then it will either run as non-root and fail, or it is running when a root user ran `ua status` which isn't helpful

So I think we should plan on running this in `postinst`

uaclient/schema/maintenance.py
```diff
@@ -0,0 +1,70 @@
...
+    for downgrade_ver in sorted(downgrade_callables, reverse=True):
+        import pdb
+
+        pdb.set_trace()
+        downgrade = downgrade_callables[downgrade_ver]
+        downgrade_status = downgrade(cfg, schemas[downgrade_ver])
+        schemas[downgrade_status["_schema_version"]] = downgrade_status
+
+    status["schemas"] = schemas
+    return status
```
orndorffgrant, line 70: This is an interesting idea, but I'm not sure its worth it because the `status.json` file will be overwritten without the `schemas` key next time a root user runs `ua status`

uaclient/schema/maintenance.py
orndorffgrant: I agree - for implementation I think I'd expect we filter `sorted(patches)` down to only the patches that have a higher version than the current version, and run all of the `up` functions in order.

orndorffgrant (commented):
@blackboxsw @lucasmoura I'm thinking we should just focus on the `up` for now and run it in `postinst`.

When downgrading, the to-be-installed older version won't have the `down` code. In that case we should probably just delete the status.json cache file - the downside is that non-root `ua status` won't be accurate until root runs `ua status` again, but since thats only the downgrade scenario, I think its acceptable?

blackboxsw:
Thanks for the excellent reviews @orndorffgrant and @lucasmoura.  I'm closing this PR so that we can properly spec, discuss and design full schema maintenance in a later release beyond 27.2,   For 27.2 support, we will perform a simple one-off script to make sure non-root user doesn't traceback across upgrade in this PR: https://github.com/canonical/ubuntu-advantage-client/pull/1710

PR #1710, opened 2021-07-07
non-root patch status.json file to avoid tracebacks (SC- 190)

blackboxsw:
There were a couple key renames and drops in in `/var/lib/ubuntu-advantage/statu.json` which would break non-root users runnning `ua status` after upgrade because non-root tries to read the old status.json cache which is only updated once a root user runs any UA subcommand.

Add postinst script to fix status.json for non-root users for the 27.2 release  of ubuntu-advantage-tools. It will only be triggered if /var/lib/ubuntu-advantage/status.json exists and we are upgrading from a ua-tools version prior to 27.2.

Also perform a drive-by bug fix: ua status on 27.2 emits warning messages about the config_path key added in an unrelated PR.
change write_cfg to avoid persiting "config_path" to uaclient.conf and move config_path additions below valid key checks in parse_config

## Proposed Commit Message
- 
    postinst: add status.json patch script run during postinst
    
    Between 27.1 and 27.2 ua client renamed a number of keys in
    /var/lib/ubuntu-advantage/status.json.
    
    This breaks status.json schema compatibility and results in a
    Traceback for non-root users running `ua status` because non-root
    sources the stale cached status.json with the old schema whereas
    root users would re-write status.json during any ua operation.
    
    Introduce a status schema patch file, which will be called during
    package upgrade in postinst which will adapt the status.json cache
    to the new keys so non-root users don't see errors from ua status.
    
    Official status.json chema compatibility will be introduced after
    27.2 release.

- config: do not warn on config_path key in parse_config
    
    Also do not persist config_path in UAConfig.write_cfg

## Test Steps
```bash
sed -i 's/UNRELEASED/bionic/' debian/changelog
git commit -am 'build for bionic'
build-package
sbuild-it ../out/ubuntu-advantage-tools_28.0.dsc
lxc launch ubuntu-daily:bionic dev-b
lxc file push ubuntu-advantage-tools_28.0_amd64.deb dev-b/
lxc exec dev-b bash
su - ubuntu
ua version # confirm current released version 27.1
sudo ua status # to persist version 27.1 status.json file

ua status  # expect ua status failure as non-root
# upgrade to your 28.0 deb
sudo dpkg -i /ubuntu-advantage-tools_28.0_amd64.deb
ua status  # expect success as non-root after postinst ran
grep -i patch /var/log/ubuntu-advantage.log # to see schema patch messages
```

## Desired commit type::

## Checklist:

### DIFF
--- a/lib/patch_status_json.py
+++ b/lib/patch_status_json.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python3
+
+"""
+Patch status.json file to avoid tracebacks for non-root from `ua status`
+
+Between 27.1 and 27.2 status.json broke schema key backward-compatibility.
+
+Running `ua status` as non-root users does not generate a new status.json with
+the updated status.json schema so the logic which renders ua status
+information will traceback on KeyErrors.
+
+Since we haven't introduced official schema versioning in UA client yet,
+this patch is a stop-gap until official schema version handling is
+delivered.
+"""
+
+import copy
+import json
+import logging
+
+from uaclient.cli import setup_logging
+from uaclient import util
+
+
+def patch_status_json_schema_0_1(status_file: str):
+    """Patch incompatible status.json file schema to align with version 0.1."""
+    setup_logging(logging.INFO, logging.DEBUG)
+    content = util.load_file(status_file)
+    try:
+        status = json.loads(content, cls=util.DatetimeAwareJSONDecoder)
+    except ValueError as e:
+        logging.debug(
+            "Unable to patch /var/lib/ubuntu-advantage/status.json: {}".format(
+                str(e)
+            )
+        )
+        return
+    new_status = copy.deepcopy(status)
+    if float(status.get("_schema_version", 0)) >= 0.1:
+        return  # Already have version 0.1 from daily PPA
+    logging.debug("Patching /var/lib/ubuntu-advantage/status.json schema")
+    new_status["_schema_version"] = "0.1"
+    new_status["account"] = {
+        "name": new_status.pop("account", ""),
+        "id": new_status.pop("account-id", ""),
+    }
+    new_status["contract"] = {
+        "tech_support_level": new_status.pop("techSupportLevel", "n/a"),
+        "name": new_status.pop("subscription", ""),
+        "id": new_status.pop("subscription-id", ""),
+    }
+    new_status["execution_status"] = new_status.pop("configStatus")
+    new_status["execution_details"] = new_status.pop("configStatusDetails")
+    try:
+        status_content = json.dumps(
+            new_status, cls=util.DatetimeAwareJSONEncoder
+        )
+    except ValueError as e:
+        logging.debug(
+            "Unable to patch /var/lib/ubuntu-advantage/status.json: {}".format(
+                str(e)
+            )
+        )
+    util.write_file(status_file, status_content)
+
+
+if __name__ == "__main__":
+    patch_status_json_schema_0_1(
+        status_file="/var/lib/ubuntu-advantage/status.json"
+    )
--- a/uaclient/tests/test_patch_status_json.py
+++ b/uaclient/tests/test_patch_status_json.py
@@ -1,0 +1,139 @@
+import logging
+import pytest
+
+import json
+
+from lib.patch_status_json import patch_status_json_schema_0_1
+
+UNATTACHED_UNPATCHED_STATUS = {
+    "_doc": "Content provided in json response is currently considered ...",
+    "attached": False,
+    "expires": "n/a",
+    "techSupportLevel": "n/a",
+    "notices": [],
+    "services": [
+        {
+            "name": "cc-eal",
+            "available": "yes",
+            "description": "Common Criteria EAL2 Provisioning Packages",
+        }
+    ],
+    "origin": None,
+    "configStatusDetails": "No Ubuntu Advantage operations are running",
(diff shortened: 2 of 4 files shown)


renanrodrigo (approved):
lgtm

uaclient/tests/test_patch_status_json.py
```diff
@@ -0,0 +1,134 @@
...
+    "_doc": "Content provided in json response is currently considered ...",
+    "origin": "free",
+    "execution_details": "Operation in progress: ua attach (pid:33140)",
+    "execution_status": "active",
+    "expires": "9999-12-31T00:00:00+00:00",
+    "contract": {
+        "tech_support_level": "n/a",
+        "name": "chad.smith@canonical.com",
+        "id": "cAKuvrqHend2ZHxUYTyJqXhULJ-4r2lpCzd1HLC_lrJg",
+    },
```
orndorffgrant, line 81: contract is missing "created_at" and "products"

uaclient/tests/test_patch_status_json.py
```diff
@@ -0,0 +1,134 @@
...
+    "services": [
+        {
+            "description_override": None,
+            "entitled": "yes",
+            "description": "Common Criteria EAL2 Provisioning Packages",
+            "status": "disabled",
+            "statusDetails": "CC EAL2 is not configured",
+            "name": "cc-eal",
```
orndorffgrant, line 89: services have the "available" key now

lib/patch_status_json.py
```diff
@@ -0,0 +1,72 @@
...
+        "id": new_status.pop("account-id", ""),
+    }
+    new_status["contract"] = {
+        "tech_support_level": new_status.pop("techSupportLevel", "n/a"),
+        "name": new_status.pop("subscription", ""),
+        "id": new_status.pop("subscription-id", ""),
+    }
+    new_status["execution_status"] = new_status.pop("configStatus")
+    new_status["execution_details"] = new_status.pop("configStatusDetails")
+    for drop_key in ("account-id", "subscription", "subscription-id"):
```
lucasmoura, line 54: Those keys should already be dropped when reaching this part of the code, right ?

lib/patch_status_json.py
```diff
@@ -0,0 +1,72 @@
...
+                str(e)
+            )
+        )
+        return
+    new_status = copy.deepcopy(status)
+    if float(status.get("_schema_version", 0)) >= 0.1:
+        return  # Already have version 0.1 from daily PPA
+    logging.debug("Patching /var/lib/ubuntu-advantage/status.json schema")
+    new_status["_schema_version"] = "0.1"
+    new_status["account"] = {
```
lucasmoura, line 43: I don't know if we want to do that, but maybe we should consider adding the new keys too, such as `created_at` in `account`  and `contract`

orndorffgrant (commented):
Thanks for this! This is a lot more straightforward than before. I pointed out some things in the end result that are different from what `ua status` will produce, but I'm not sure if they're really necessary.

renanrodrigo:
> lgtm

clearly I missed a lot of context lol

lib/patch_status_json.py
blackboxsw: ahh right, I rewrote the logic above later and forgot to drop this section. +1

lib/patch_status_json.py
blackboxsw: I was trying to hold off on the full schema key upgrades here and only perform the minimal key updates needed to allow non-root to succeed with status calls. I wanted to explicitly avoid performing a full schema update here for anything that is unneeded for status. The reason being that any further ua commands run by root will be correcting and rewriting this file anyway with the proper full schema.

orndorffgrant (approved):
Thanks for clarifying that this is only supposed to be a minimal set of changes to allow non-root `ua status` to work. Code looks good and worked for me!

uaclient/tests/test_patch_status_json.py
```diff
@@ -0,0 +1,134 @@
...
+    "configStatusDetails": "No Ubuntu Advantage operations are running",
+    "configStatus": "inactive",
+}
+
+UNATTACHED_PATCHED_STATUS = {
+    "execution_details": "No Ubuntu Advantage operations are running",
+    "_schema_version": "0.1",
+    "origin": None,
+    "account": {"name": "", "id": ""},
+    "expires": "n/a",
```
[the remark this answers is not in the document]
blackboxsw: Probably leaving these minor discrepancies unaddressed because we are only performing the minimal set of key changes to support non-root user runs of `ua status`. It allows us to keep this script minimal until we have official schema compatibility support for status.json where we can construct each "supported" schema version of status.json via a true patching up/down set of operations.

uaclient/tests/test_patch_status_json.py
```diff
@@ -0,0 +1,134 @@
...
+            "name": "cc-eal",
+        }
+    ],
+    "_schema_version": "0.1",
+    "attached": True,
+    "notices": [["Operation in progress: ua attach"]],
+    "account": {
+        "name": "chad.smith@canonical.com",
+        "id": "aAHQlZdfWiafnWvjDZCZDqZzDUComc8WF7IJnoG6GAmA",
+    },
```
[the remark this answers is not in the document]
blackboxsw: +1 though tabled for official schema support.

PR #1711, opened 2021-07-08
config: remove config_path from config dict

lucasmoura:
## Proposed Commit Message
config: remove config_path from config dict

We are now removing the config_path entry from the config dict and pushing that logic into a dedicated function. We are using that to not clash with the write_cfg method which by default writes all content of the config dict into disk. Since the config_path is a variable that is being used only internally, we don't need it on disk.

## Test Steps
Run the modified unit tests and check that `config_path` is still working as expected when running any of the `ua status --format` commands

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -121,9 +121,11 @@ class UAConfig:
     ) -> None:
         """"""
         if cfg:
+            self.cfg_path = None
             self.cfg = cfg
         else:
-            self.cfg = parse_config()
+            self.cfg_path = get_config_path()
+            self.cfg = parse_config(self.cfg_path)
 
         self.series = series
 
@@ -534,10 +536,8 @@ class UAConfig:
             "execution_status": status_val,
             "execution_details": status_desc,
             "notices": notices,
-            "config_path": self.cfg.get("config_path"),
-            "config": {
-                k: v for k, v in self.cfg.items() if k != "config_path"
-            },
+            "config_path": self.cfg_path,
+            "config": self.cfg,
         }
 
     def _unattached_status(self) -> "Dict[str, Any]":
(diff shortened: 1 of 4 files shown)


uaclient/tests/test_cli_status.py
```diff
@@ -492,7 +492,7 @@ def test_attached_formats(
                 "created_at": account_created_at,
                 "external_account_ids": [{"IDs": ["id1"], "Origin": "AWS"}],
             },
-            "config_path": "config_path",
+            "config_path": None,
```
renanrodrigo, line 495: Question: is there any test to cover when the config path would not be `None` ?

lucasmoura: Currently not, because of the way we are using the `FakeConfig` fixture, which creates a `UAConfig` object with a config dict ready. Maybe we should better address this in the integration tests

orndorffgrant: Good catch, I think it'd be sufficient to just add a quick unit test for `get_config_path`

uaclient/config.py
```diff
@@ -828,10 +828,27 @@ def write_cfg(self, config_path=None):
...
+
+    config_path = DEFAULT_CONFIG_FILE
+    local_cfg = os.path.join(os.getcwd(), os.path.basename(config_path))
+    if os.path.exists(local_cfg):
+        config_path = local_cfg
+
+    if os.environ.get("UA_CONFIG_FILE"):
+        config_path = os.environ.get("UA_CONFIG_FILE")
+
+    return config_path
```
blackboxsw, line 849: We can simplify this function a bit by:
1.  dropping the config_path param and only calling this function when we need to.
2. Make it a bit more DRY by changing the order we look at env variable since it overrides what we setup anyway.
3. returning early when we have the value we want
```suggestion
def get_config_path() -> str:
    """Get config path to be used when loading config dict."""
    if os.environ.get("UA_CONFIG_FILE"):
        return os.environ.get("UA_CONFIG_FILE")

    local_cfg = os.path.join(os.getcwd(), os.path.basename(DEFAULT_CONFIG_FILE))
    if os.path.exists(local_cfg):
        return local_cfg
    return DEFAULT_CONFIG_FILE
```

uaclient/config.py
```diff
@@ -845,14 +862,8 @@ def parse_config(config_path=None):
...
     """
-    if not config_path:
-        config_path = DEFAULT_CONFIG_FILE
     cfg = copy.copy(CONFIG_DEFAULTS)
-    local_cfg = os.path.join(os.getcwd(), os.path.basename(config_path))
-    if os.path.exists(local_cfg):
-        config_path = local_cfg
-    if os.environ.get("UA_CONFIG_FILE"):
-        config_path = os.environ.get("UA_CONFIG_FILE")
+    config_path = get_config_path(config_path)
```
blackboxsw, line 866: This can change to 
```suggestion
    if not config_path:
        config_path = get_config_path()
```

blackboxsw (changes requested):
Thanks Lucas. minor suggestions to simplify this function

lucasmoura:
@blackboxsw code updated

blackboxsw (approved):
LGTM! thanks!

PR #1714, opened 2021-07-08
fips: do not block enable on conditional packages (SC-216)

lucasmoura:
## Proposed Commit Message
fips: do not block enable on conditional packages

When enabling fips we install FIPS version of some packages if they are already installed in the system. Currently, we
block enable if we can't install any of those conditional packages. Now, we are alerting the user about errors on installing those
packages, but we don't block the enable operation

## Test Steps
Run the new integration test added

## Desired commit type::

## Checklist:

### DIFF
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -290,6 +290,106 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            "availableResources": [
                {
                  "available": true,
+                 "name": "fips"
+               }
+            ],
+            "machineTokenInfo": {
+                "contractInfo": {
+                   "resourceEntitlements": [
+                      {
+                        "type": "fips",
+                        "entitled": true,
+                        "affordances": {
+                          "series": [
+                            "xenial",
+                            "bionic",
+                            "focal"
+                          ]
+                         },
+                        "directives": {
+                          "suites": [
+                            "xenial",
+                            "bionic",
+                            "focal"
+                          ]
+                        }
+                     }
+                   ]
+                }
+            }
+        }
+        """
+        And I append the following on uaclient config:
+        """
+        features:
+          machine_token_overlay: "/home/ubuntu/machine-token-overlay.json"
+        """
+        When I run `ua enable <fips-service> --assume-yes` with sudo
+        Then stdout matches regexp:
+            """
+            Updating package lists
+            Installing <fips-name> packages
+            FIPS strongswan-hmac package could not be installed
+            <fips-name> enabled
+            A reboot is required to complete install
+            """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            <fips-service> +yes                enabled
+            """
+        And I verify that running `apt update` `with sudo` exits `0`
+        And I verify that `openssh-server` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-client` is installed from apt source `<fips-apt-source>`
+        When I reboot the `<release>` machine
+        And  I run `uname -r` as non-root
+        Then stdout matches regexp:
+            """
+            fips
+            """
+        When I run `cat /proc/sys/crypto/fips_enabled` with sudo
+        Then I will see the following on stdout:
+        """
+        1
+        """
+        When I run `ua disable <fips-service> --assume-yes` with sudo
+        Then stdout matches regexp:
+            """
+            Updating package lists
+            A reboot is required to complete disable operation
+            """
+        When I reboot the `<release>` machine
+        Then I verify that `openssh-server` installed version matches regexp `fips`
+        And I verify that `openssh-client` installed version matches regexp `fips`
+        When I run `apt-mark unhold openssh-client openssh-server strongswan` with sudo
+        Then I will see the following on stdout:
+        """
+        openssh-client was already not hold.
+        openssh-server was already not hold.
+        strongswan was already not hold.
+        """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            <fips-service> +yes                disabled
+            """
+
+        Examples: ubuntu release
+           | release | fips-name    | fips-service |fips-apt-source                               |
+           | focal   | FIPS         | fips |https://esm.staging.ubuntu.com/fips/ubuntu focal/main |
+
+    @series.focal
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable of vm-based services in an ubuntu lxd vm
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token_staging` with sudo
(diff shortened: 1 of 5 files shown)


uaclient/entitlements/fips.py
```diff
@@ -62,6 +62,41 @@ def conditional_packages(self):
...
+            self.packages.
+        :param package_list: Optional package list to use instead of
+            self.packages.
+        """
+        print("Installing {title} packages".format(title=self.title))
+
+        # We need to guarantee that the metapackage is installed.
+        # While the other packages should still be installed, if they
+        # fail, we should not block the enable operation.
+        fips_metapackage = self.packages[0]
```
blackboxsw, line 83: This seems very brittle, if the contract ever changes the number of mandatory additionalPackages directives to something greater than one, we now will treat those packages as optional too.  I think we need to treat conditional packages uniquely from the additionalPackage directives since they are a different "type" of directive that we treat like  `addPackagesAllowInstallFailures` if the contract server we able to surface this type of directive.


What do you think about trimming down  FIPS.packages method to only fixup the proper cloud metapkg name and just treating the conditional_packages here in your sub-classed install_packages method?

I was thinking about something like this
```diff
diff --git a/uaclient/entitlements/fips.py b/uaclient/entitlements/fips.py
index 60170a17..e9692b7b 100644
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -78,14 +78,27 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         print("Installing {title} packages".format(title=self.title))
 
         # We need to guarantee that the metapackage is installed.
-        # While the other packages should still be installed, if they
-        # fail, we should not block the enable operation.
+        # The mandatory meta package is present in addtionalPackages directive
+        # and can be installed by the default install_packages method.
         fips_metapackage = self.packages[0]
-        super().install_packages(
-            package_list=[fips_metapackage], verbose=False
+        super().install_packages(verbose=False)
+
+        # Any conditional packages packages should still be installed, but if
+        # they fail to install we should not block the enable operation.
+        installed_packages = apt.get_installed_packages()
+
+        desired_packages = []
+        pkg_groups = groupby(
+            sorted(self.conditional_packages),
+            key=lambda pkg_name: pkg_name.replace("-hmac", ""),
         )
 
-        for pkg in self.packages[1:]:
+        for pkg_name, pkg_list in pkg_groups:
+            if pkg_name in installed_packages:
+                desired_packages += pkg_list
+
+
+        for pkg in desired_packages:
             try:
                 super().install_packages(
                     package_list=[pkg], cleanup_on_failure=False, verbose=False
@@ -220,17 +233,6 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
     def packages(self) -> "List[str]":
         packages = super().packages
         packages = self._replace_metapackage_on_cloud_instance(packages)
-        installed_packages = apt.get_installed_packages()
-
-        pkg_groups = groupby(
-            sorted(self.conditional_packages),
-            key=lambda pkg_name: pkg_name.replace("-hmac", ""),
-        )
-
-        for pkg_name, pkg_list in pkg_groups:
-            if pkg_name in installed_packages:
-                packages += pkg_list
-
         return packages
 
     def application_status(self) -> "Tuple[status.ApplicationStatus, str]":
```

blackboxsw (changes requested):
I like the FIPSEntitlement,install_packages method in general, I think we minimally might need to rethink our hard-coding our self.packages[1:] operations in your new method to avoid strange side-effects if addtionalPackages directives change in the future.

uaclient/entitlements/repo.py
```diff
@@ -222,16 +225,22 @@ def process_contract_deltas(
...
+        cleanup_on_failure: bool = True,
+        verbose: bool = True,
     ) -> None:
         """Install contract recommended packages for the entitlement.
 
         :param package_list: Optional package list to use instead of
             self.packages.
-        :param package_list: Optional package list to use instead of
-            self.packages.
+        :param cleanup_on_failure: Cleanup apt files if apt install faile.
```
renanrodrigo, line 237: small typo here - `fails`?

uaclient/entitlements/fips.py
```diff
@@ -62,6 +62,41 @@ def conditional_packages(self):
...
+        self,
+        package_list: "List[str]" = None,
+        cleanup_on_failure: bool = True,
+        verbose: bool = True,
+    ) -> None:
+        """Install contract recommended packages for the entitlement.
+
+        :param package_list: Optional package list to use instead of
+            self.packages.
+        :param package_list: Optional package list to use instead of
```
renanrodrigo, line 75: docstring params don't match here

uaclient/entitlements/fips.py
orndorffgrant: If we're handling errors when installs fail, do we even need to do the `desired_packages` logic (to only try to upgrade if already installed) anymore?

Or with this new approach, can we just iterate over `conditional_packages` directly, try to install each one individually, and catch any errors?

uaclient/entitlements/fips.py
```diff
@@ -62,6 +62,41 @@ def conditional_packages(self):
...
+            self.packages.
+        """
+        print("Installing {title} packages".format(title=self.title))
```
renanrodrigo, line 78: Should this also respect the `verbose` flag?

uaclient/entitlements/fips.py
lucasmoura: I think we should because otherwise we could end up unnecessary packages in the system. For example, if the user doesn't have `strongswan` in the system, we don't need to install the FIPS version of it for the system to be FIPS compliant.

orndorffgrant: Good point! Thats good with me

lucasmoura:

renanrodrigo (approved):
lgtm

features/staging_commands.feature
```diff
@@ -278,6 +278,106 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | xenial  | FIPS Updates | fips-updates |https://esm.staging.ubuntu.com/fips-updates/ubuntu xenial-updates/main |
            | bionic  | FIPS Updates | fips-updates |https://esm.staging.ubuntu.com/fips-updates/ubuntu bionic-updates/main |
 
+    @series.focal
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable of vm-based services in an ubuntu lxd vm
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token_staging` with sudo
+        And I run `DEBIAN_FRONTEND=noninteractive apt-get install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y openssh-client openssh-server strongswan` with sudo, retrying exit [100]
+        And I create the file `/home/ubuntu/machine-token-overlay.json` with the following:
```
blackboxsw, line 287: We don't need this overlay anymore now that it landed in staging yesterday , but we can drop this as a separate PR

blackboxsw (approved):
LGTM!. Let's take an action to remove the fips staging overlay. I've created https://warthogs.atlassian.net/browse/SC-217 and assigned it to @renanrodrigo

PR #1721, opened 2021-07-13
Remove changelog entry already reflected on 27.1

lucasmoura:
We have a changelog entry for fixing lintian errors on the release 27.0 changelog. However, this was already reflected on 27.1. We are now removing that changelog entry from 27.2


athos-ribeiro (approved):
LGTM

Thanks, Lucas!

mitechie (approved):

PR #1725, opened 2021-07-13
tests: add scenarios for ua status --json|yaml

renanrodrigo:
Add integration tests to verify that `ua status --format json` or `ua status --format yaml` have the correct output, with the expected keys in them.

## Test Steps
Run the integration tests, verify it passes

## Desired commit type::

## Checklist:


features/steps/steps.py
```diff
@@ -352,6 +354,21 @@ def then_stdout_matches_regexp(context):
     assert_that(context.process.stdout.strip(), matches_regexp(context.text))
 
 
+@then("stdout is formatted as `{output_format}` and has keys")
+def then_stdout_is_formatted_and_has_keys(context, output_format):
+    output = context.process.stdout.strip()
+    if output_format == "json":
```
renanrodrigo, line 360: should we have specific errors overall?
Here for instance, should the test inform that only `json` or `yaml` are valid formats given it comes from the test scenario itself?

right now, it will fail because `data` does not exist

features/steps/steps.py
```diff
@@ -352,6 +354,21 @@ def then_stdout_matches_regexp(context):
...
+        data = json.loads(output)
+    elif output_format == "yaml":
+        data = yaml.safe_load(output)
+
+    keys = context.text.split()
+    in_data = [k for k in keys if k in data]
+
+    assert_that(len(keys), equal_to(len(data)))
+    assert_that(len(keys), equal_to(len(in_data)))
+
```
renanrodrigo, line 370: Does this error message needs to be improved as well? Right now, we have just the assertion error with the numbers. Should it be clearer on why, or we leave this to whoever runs it?

renanrodrigo (commented):
quick questions on this draft:

features/unattached_status.feature
```diff
@@ -1,5 +1,32 @@
 Feature: Unattached status
 
+    @series.all
+    @wip
+    Scenario Outline: Unattached status in a ubuntu machine - formatted
```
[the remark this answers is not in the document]
lucasmoura: Yes, the keys should remain the same. I think it is worth having that test on an attached machine as well just to ensure that this is true

features/steps/steps.py
lucasmoura: No, I don't think we need to have anything like this at the moment. The test should fail loudly if we add a new format without changing the code, so we can spot that and fix it

features/steps/steps.py
lucasmoura: We could potentially show which key is missing. This could provide useful information if the test fails

renanrodrigo: Ok, so, I am currently verifying if the keys are exactly those listed on the test, no more, no less.
I can change it to point to missing keys, or extraneous ones on error.
Is that it, or we can pass on extraneous keys? (My guess is no, it should be exact the same)

features/unattached_status.feature
```diff
@@ -1,5 +1,32 @@
...
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `ua status --format json` as non-root
```
renanrodrigo, line 7: Another one: I am testing it only as non-root, because it seems that status with `sudo` will generate the same output

features/steps/steps.py
lucasmoura: I think it should exactly the same

features/unattached_status.feature
lucasmoura: I mean, having another call as root will not affect performance that much. But I do agree that this is an overkill

lucasmoura (approved):

Issue #1727, opened 2021-07-13
Document proxy authentication methods


athos-ribeiro:
It would be nice to have a sentence in the manpages (and any other documentation bit where it could be relevant) on how proxy authentication can be performed for the `https_proxy, http_proxy, apt_https_proxy, apt_http_proxy` options

PR #1729, opened 2021-07-14
Information about proxy authentication (SC-238)

renanrodrigo:
Add information to the manpage on how to authenticate to proxy servers if needed, and add an integration test to ensure this authentication is working.

Fixes: #1727

## Desired commit type::

## Checklist:

### DIFF
--- a/features/proxy_config.feature
+++ b/features/proxy_config.feature
@@ -206,3 +206,116 @@ Feature: Proxy configuration
            | release |
            | xenial  |
            | bionic  |
+
+    @series.lts
+    Scenario Outline: Attach command when authenticated proxy is configured
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I launch a `focal` `proxy` machine
+        And I run `apt install squid apache2-utils -y` `with sudo` on the `proxy` machine
+        And I run `htpasswd -bc /etc/squid/passwordfile someuser somepassword` `with sudo` on the `proxy` machine
+        And I add this text on `/etc/squid/squid.conf` on `proxy` above `http_access deny all`:
+            """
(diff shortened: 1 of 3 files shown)


features/proxy_config.feature
```diff
@@ -206,3 +206,38 @@ Feature: Proxy configuration
            | release |
            | xenial  |
            | bionic  |
+
+    @series.lts
+    Scenario Outline: Attach command when authenticated proxy is configured
```
lucasmoura, line 211: That's pretty good, but we should also verify if snap and livepatch are answering to this proxy. Furthermore, maybe we could reuse an existing scenario for this test, what do you think ?

renanrodrigo: I'd keep a single scenario for all the authenticated proxy tests for the sake of isolation (then, any authentication issue would break just this one).
I agree that snap and livepatch should be covered, and can be added to this scenario

renanrodrigo: Question:
Is there a reason for the tests in this file targetting the Snap and livepatch proxies are not executed for focal?

lucasmoura: Yes, livepatch does not support the kvm kernel we run on the lxd instances

lucasmoura: That's also why we don't have any livepatch test for focal at the moment

renanrodrigo:
Added statements to check for snap and livepatch sing the authenticated proxy, and to ensure wrong credentials are not accepted

lucasmoura (approved):

orndorffgrant (approved):
The new tests passed for me and the new docs look good!

Issue #1734, opened 2021-07-16
`KeyError: 'configStatus'` exception from postinst


iainlane:
I just did a `dist-upgrade` on my Impish system. I got this:

```
Unpacking ubuntu-advantage-tools (27.2~21.10.1) over (27.1~21.10.1.1) ...
...
Setting up ubuntu-advantage-tools (27.2~21.10.1) ...
Installing new version of config file /etc/ubuntu-advantage/uaclient.conf ...
Traceback (most recent call last):
  File "/usr/lib/ubuntu-advantage/patch_status_json.py", line 66, in <module>
    patch_status_json_schema_0_1(
  File "/usr/lib/ubuntu-advantage/patch_status_json.py", line 50, in patch_status_json_schema_0_1
    new_status["execution_status"] = new_status.pop("configStatus")
KeyError: 'configStatus'
```

Here is my `status.json` with IDs redacted:

```
{
  "_doc": "Content provided in json response is currently considered Experimental and may change",
  "attached": true,
  "expires": "n/a",
  "origin": "free",
  "services": [
    {
      "name": "cc-eal",
      "description": "Common Criteria EAL2 Provisioning Packages",
      "entitled": "yes",
      "status": "n/a",
      "statusDetails": "",
      "description_override": null
    },
    {
      "name": "cis-audit",
      "description": "Center for Internet Security Audit Tools",
      "entitled": "no",
      "status": "—",
      "statusDetails": "",
      "description_override": null
    },
    {
      "name": "esm-apps",
      "description": "UA Apps: Extended Security Maintenance",
      "entitled": "no",
      "status": "—",
      "statusDetails": "",
      "description_override": null
    },
    {
      "name": "esm-infra",
      "description": "UA Infra: Extended Security Maintenance",
      "entitled": "yes",
      "status": "enabled",
      "statusDetails": "ESM Infra is active",
      "description_override": null
    },
    {
      "name": "fips",
      "description": "NIST-certified FIPS modules",
      "entitled": "yes",
      "status": "n/a",
      "statusDetails": "",
      "description_override": null
    },
    {
      "name": "fips-updates",
      "description": "Uncertified security updates to FIPS modules",
      "entitled": "yes",
      "status": "n/a",
      "statusDetails": "",
      "description_override": null
    },
    {
      "name": "livepatch",
      "description": "Canonical Livepatch service",
      "entitled": "yes",
      "status": "enabled",
      "statusDetails": "",
      "description_override": null
[...]

PR #1735, opened 2021-07-16
postinst: ignore configStatus when not present (SC-240)

renanrodrigo:
Ignore potentially non-existing keys when patching the status JSON.

Fixes: #1734 

## Desired commit type::

## Checklist:

### DIFF
--- a/lib/patch_status_json.py
+++ b/lib/patch_status_json.py
@@ -47,8 +47,10 @@ def patch_status_json_schema_0_1(status_file: str):
         "name": new_status.pop("subscription", ""),
         "id": new_status.pop("subscription-id", ""),
     }
-    new_status["execution_status"] = new_status.pop("configStatus")
-    new_status["execution_details"] = new_status.pop("configStatusDetails")
+    if "configStatus" in new_status:
+        new_status["execution_status"] = new_status.pop("configStatus")
+    if "configStatusDetails" in new_status:
+        new_status["execution_details"] = new_status.pop("configStatusDetails")
     try:
         status_content = json.dumps(
             new_status, cls=util.DatetimeAwareJSONEncoder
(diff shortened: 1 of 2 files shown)


lib/patch_status_json.py
```diff
@@ -47,8 +47,9 @@ def patch_status_json_schema_0_1(status_file: str):
         "name": new_status.pop("subscription", ""),
         "id": new_status.pop("subscription-id", ""),
     }
-    new_status["execution_status"] = new_status.pop("configStatus")
-    new_status["execution_details"] = new_status.pop("configStatusDetails")
+    if "configStatus" in new_status:
+        new_status["execution_status"] = new_status.pop("configStatus")
+        new_status["execution_details"] = new_status.pop("configStatusDetails")
```
lucasmoura, line 52: Just to be extra safe here, I think we need to make the same check for `configStatusDetails`

renanrodrigo: Both were introduced at the same point, so one is present only when the other is too.
But I can add it just to be safe if you think it's needed

orndorffgrant:
I was able to reproduce the issue on `main` and confirm that this branch fixes it. I will approve after a test case is added to cover this conditional

renanrodrigo:
@orndorffgrant 28 seconds!

lib/patch_status_json.py
lucasmoura: Yep, I do realize that. It is just to be extra safe, but maybe it is indeed not needed

lucasmoura (approved):

lib/patch_status_json.py
renanrodrigo: too late, added hahaha

orndorffgrant (approved):
Still works after changes and test looks good. Thanks for this quick fix!

PR #1737, opened 2021-07-16
tests: always build packages using sbuild (SC-267)

lucasmoura:
## Proposed Commit Message
tests: always build packages on a LXD container
    
When running integration tests with BUILD_PR=1, we build the ua package using in the same instance type that we are going to run the test on. This is not ideal, since we run integrations tests on clouds and LXD vms. We are now building the packages through sbuild, which is a faster and more correct approach for building the packages to be used in the test

## Test Steps
Run either a vm or a cloud test with UACLIENT_BEHAVE_BUILD_PR=1 and verify that the package is built using sbuild.
To verify that run the tests with `--no-junit` and take a look at the logs to see references for the sbuild process running

## Desired commit type::

## Checklist:

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -5,6 +5,7 @@ import re
 import tempfile
 import textwrap
 import logging
+import subprocess
 import pycloudlib  # type: ignore
 
 try:
@@ -615,15 +616,14 @@ def _capture_container_as_image(
     return cloud_api.snapshot(instance=inst)
 
 
-def build_debs_from_dev_instance(context: Context, series: str) -> "List[str]":
-    """Create a development instance, instal build dependencies and build debs
+def build_debs_from_sbuild(context: Context, series: str) -> "List[str]":
+    """Create a chroot and build the package using sbuild
 
 
     Will stop the development instance after deb build succeeds.
 
     :return: A list of paths to applicable deb files published.
     """
-    time_suffix = datetime.datetime.now().strftime("%s%f")
     deb_paths = []
 
     if context.config.debs_path:
@@ -646,33 +646,35 @@ def build_debs_from_dev_instance(context: Context, series: str) -> "List[str]":
         logging.info(
             "--- Could not find any debs to reuse. Building it locally"
         )
-        logging.info(
-            "--- Launching vm to build ubuntu-advantage*debs from local source"
-        )
-        build_container_name = (
-            "ubuntu-behave-image-pre-build-%s-" % series + time_suffix
-        )
+        chroot_name = "{}-amd64".format(series)
 
-        cloud_manager = context.config.cloud_manager
-        if "pro" in context.config.machine_type:
-            user_data = USERDATA_BLOCK_AUTO_ATTACH_IMG
-        else:
-            user_data = ""
-        inst = cloud_manager.launch(
-            instance_name=build_container_name,
-            series=series,
-            user_data=user_data,
-        )
+        try:
+            subprocess.check_output(["schroot", "-i", "-c", chroot_name])
+        except subprocess.CalledProcessError:
+            cmd = (
+                "sbuild-launchpad-chroot create"
+                '--architecture="amd64" --name={}-amd64"'.format(series),
+                '--series="{}"'.format(series),
+            )
+            msgs = (
+                "-- You need a {} chroot to proceed".format(series),
+                "-- Please run the following cmd as root:",
+                " ".join(cmd),
+                "-- If you don't have sbuild-launchpad-chroot please run:"
+                "apt-get install sbuild-launchpad-chroot",
+            )
 
-        build_container_name = cloud_manager.get_instance_id(inst)
+            logging.info("\n".join(msgs))
+            exit(-1)
 
         with emit_spinner_on_travis("Building debs from local source... "):
             deb_paths = build_debs(
-                build_container_name,
+                series=series,
                 output_deb_dir=os.path.join(tempfile.gettempdir(), series),
-                cloud_api=context.config.cloud_api,
                 cache_source=context.config.cache_source,
             )
+
+        logging.info("--- sbuild has finished building the packages")
 
     if "pro" in context.config.machine_type:
         return deb_paths
@@ -713,7 +715,7 @@ def create_uat_image(context: Context, series: str) -> None:
     time_suffix = datetime.datetime.now().strftime("%s%f")
     deb_paths = []
     if context.config.build_pr:
-        deb_paths = build_debs_from_dev_instance(context, series)
+        deb_paths = build_debs_from_sbuild(context, series)
 
     logging.info(
         "--- Launching VM to create a base image with updated ubuntu-advantage"
--- a/features/environment.py
+++ b/features/environment.py
@@ -5,7 +5,6 @@ import re
 import tempfile
 import textwrap
 import logging
-import subprocess
 import pycloudlib  # type: ignore
 
 try:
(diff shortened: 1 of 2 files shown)


features/environment.py
```diff
@@ -645,26 +645,30 @@ def build_debs_from_dev_instance(context: Context, series: str) -> "List[str]":
             "ubuntu-behave-image-pre-build-%s-" % series + time_suffix
         )
 
-        cloud_manager = context.config.cloud_manager
+        package_cloud_manager = cloud.LXDContainer(
+            machine_type=context.config.machine_type
```
orndorffgrant, line 649: Can we just hardcode "lxd.container" here instead of passing the machine_type we're testing (which may not be "lxd.container")?

orndorffgrant (commented):
Thanks for putting this together! Running a "quick" test to see how much time this saves. Just one question on the code.

lucasmoura:
@orndorffgrant code updated

orndorffgrant (approved):
@lucasmoura This doesn't seem to be a significant improvement for the lxd VM tests that I was trying. It is more likely to be beneficial for the cloud tests (though I didn't try them).

Regardless, this is a cleaner approach +1

paride:
Hi @lucasmoura, I agree this change is an improvement, however as we are reconsidering how the packages are built for testing, is there any reason why we can't build them in a schroot via `sbuild`? If we're aiming for speed and (enough) isolation from the host system that's for sure the best option, especially given that the Jenkins nodes locally cache the .deb packages installed in the schroots.

features/environment.py
```diff
@@ -649,31 +649,36 @@ def build_debs_from_dev_instance(context: Context, series: str) -> "List[str]":
...
-        else:
-            user_data = ""
-        inst = cloud_manager.launch(
-            instance_name=build_container_name,
-            series=series,
-            user_data=user_data,
-        )
+        chroot_name = "{}-amd64".format(series)
+
+        try:
```
lucasmoura, line 654: Don't know if we need all of this code to check the existence of chroots or even if this is the best approach to check it

features/util.py
```diff
@@ -85,72 +85,87 @@ def lxc_get_property(name: str, property_name: str, image: bool = False):
...
-        logging.info("--- Push {} -> {}:/tmp".format(filepath, instance.name))
-        instance.push_file(filepath, "/tmp/" + os.path.basename(filepath))
-    result = instance.execute(["sudo", "bash", "/tmp/" + buildscript])
-    if result.failed:
-        logging.error("--- ERROR: Building PR failed")
-        raise Exception(
-            "stdout: {}\nstderr: {}".format(result.stdout, result.stderr)
+        os.chdir(cwd)
+
+    logging.info("--- Creating: {}".format(SOURCE_PR_UNTAR_DIR))
```
lucasmoura, line 121: @paride I wasn't able to run sbuild using a tarball directly, that's why I am using this approach were I untar the code tarball and run sbuild passing the directory as an argument

features/util.py
```diff
@@ -85,72 +85,87 @@ def lxc_get_property(name: str, property_name: str, image: bool = False):
...
         )
-    deb_artifacts = []
+
+    curr_dir = os.getcwd()
+    os.chdir(temp_dir)
+    logging.info("--- Running sbuild")
+    subprocess.run(
+        [
+            "sbuild",
+            "--no-run-lintian",
```
lucasmoura, line 143: We could speed up the process more by adding the flags `--no-apt-upgrades` and `--no-apt-distupgrades`. However, I don't know if this a good approach for building on the chroots

paride: In general it's good we want the chroot to be up-to-date, otherwise when installing the build-deps there may be packages in the local package index that are not present anymore in the archive (due to updated). However we can assume that *on the jenkins nodes* the schroots are up-to-date, because that's done automatically every time a schroot session is started (that's how `sbuild-launchpad-chroot` sets things up).

paride: Given that the schroot will be up to date the performance penalty will be rather low, I'd stick with the defaults to cover more use cases.

features/environment.py
paride: sbuild tries a few schroot names to figure out the correct one:

> If not specified, the default is the first of schroot  name  or  alias  that matches  $distribution-$arch-sbuild, $distribution-sbuild, $distribution-$arch or $distribution that exists.

so the `"{}-amd64".format(series)` is a bit strict. I'd take the simplest approach: assume that an appropriate schroot exists (= no check), and let sbuild (and therefore the build) fails if it doesn't. Basically I'd treat the "missing schroot" case as we treat build failures in general. Wdyt?

features/util.py
paride: I'll try to figure this out.

features/environment.py
lucasmoura: I think that's fair. I will just test it out how the error is shown in case we don't have the chroot environment

features/util.py
paride: I can see the issue here. The normal way to produce a .deb source package composed by a .dsc file and tarball would be:
```
dpkg-source -b .
```
which automatically ignores several stuff, namely: `-I*.a -I*.la -I*.o -I*.so -I.*.sw? -I*/*~ -I,,* -I.[#~]* -I.arch-ids -I.arch-inventory -I.be -I.bzr -I.bzr.backup -I.bzr.tags -I.bzrignore -I.cvsignore -I.deps -I.git -I.gitattributes -I.gitignore -I.gitmodules -I.gitreview -I.hg -I.hgignore -I.hgsigs -I.hgtags -I.mailmap -I.mtn-ignore -I.shelf -I.svn -ICVS -IDEADJOE -IRCS -I_MTN -I_darcs -I{arch}`.
This list doesn't cover `.tox` and there is no "smart" option to parse `.gitignore` like `tar --exclude-vcs-ignores`. There is also no way to pass raw options to tar.

We could specify what to ignore manually via `-I`, but it will get out of sync at some point. I have another suggestion:
```
git archive --prefix=ua-clean-tree/ HEAD | tar -x -C ..
```
this should export whatever is committed into git HEAD to `../ua-clean-tree/`, without any `.git` or uncommitted/untracked filed. We can then sbuild that directory.

orndorffgrant: One downside to this approach is that it will not capture any untracked changes, if i understand correctly. We would like to easily test changes that are in-progress using the integration tests. With this approach, we can do that, we just need to create a commit for every change before running the tests with `BUILD_PR=1`.

This isn't a deal-breaker, but if we go this route, then we should document this in the readme.

lucasmoura: @paride I think the problem is not generating the tarball per se, but using it directly on sbuild, like:
`sbuild uaclient.tar.gz`

It seems that sbuild does not accept the tarball directly. When I used this, it tried to download a package source with that name instead of using the tarball directly.

But I can be missing something here

paride: Yep we need the full source package (tarball + .dsc), that's what `dpkg-source -b` would do, but it isn't as smart as `tar --exclude-vcs-ignores`. Per standup discussion, I'm fine with the tarball approach at this point.

Maybe we could do `tar -c | tar -x -C` without ever writing the tarball to disk, for a *slight* performance gain. Not sure if worth it, up to you. Thanks for implementing this!

paride (approved):
This LGTM, thanks @lucasmoura for implementing it and the discussions. To be verified if the "base" schroots set up via `sbuild-launchpad-chroot` actually do get updated+upgraded.

features/environment.py
```diff
@@ -649,31 +648,16 @@ def build_debs_from_dev_instance(context: Context, series: str) -> "List[str]":
         logging.info(
             "--- Launching vm to build ubuntu-advantage*debs from local source"
```
renanrodrigo, line 649: Do this comment still apply?

lucasmoura: Nope, I will remove it

orndorffgrant (approved):
Thanks for figuring this out @lucasmoura! +1

PR #1741, opened 2021-07-19
tests: don't run apt update on userdata

lucasmoura:
## Proposed Commit Message
tests: don't run apt update on userdata

When installing ua directly from proposed, we create an userdata that create the proposed source file and runs apt update. This can cause race conditions on the start up process of an instance, since another apt command might be running. To try to avoid that, we are now running the apt update command only when we really need it

## Test Steps
Run the gcp generic xenial test without this change and verify that it fails because of a cloud-init error. Apply that change and confirm that the test works as expected

## Desired commit type::

## Checklist:


renanrodrigo (approved):

orndorffgrant (approved):
This worked for me! Since this touches code that affects other integration tests, my inclination is to wait for jenkins to pass as well

PR #1743, opened 2021-07-21
postinst: do not run cloud-id if cloud-init did not run

mwhudson:
cloud-id will only do anyting useful if cloud-init ran and fails
otherwise (e.g. in a chroot while building an image) so do not run it at
all if /run/cloud-init/instance-data.json is not present.

LP: #1936833


lucasmoura (approved):

lucasmoura:
Thanks for this patch @mwhudson

PR #1744, opened 2021-07-21
Allow running tests with ephemeral instances (SC-266)

lucasmoura:
## Proposed Commit Message
Allow running tests with ephemeral instances

We are now allowing running the integration tests with ephemeral instances. This will allow LXD to better handle deleting the instance, since we can now just stop the ephemeral instance and LXD will properly handle the deletion of the instance

## Test Steps
Run any integration test with `UACLIENT_BEHAVE_EPHEMERAL_INSTANCE=1` and verify that the instance created instance
is an ephemeral one

## Desired commit type::

## Checklist:

### DIFF
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -72,6 +72,7 @@ class Cloud:
         instance_name: "Optional[str]" = None,
         image_name: "Optional[str]" = None,
         user_data: "Optional[str]" = None,
+        ephemeral: bool = False,
     ) -> pycloudlib.instance:
         """Create an instance for on the cloud provider.
 
@@ -85,6 +86,8 @@ class Cloud:
             The name of the image to be used when creating the instance
         :param user_data:
             The user data to be passed when creating the instance
+        :param ephemeral":
+            If instance should be ephemeral
 
         :returns:
             A cloud provider instance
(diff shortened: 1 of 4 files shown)


paride (commented):
I tested this and it works fine, thanks! One thing: I noticed that the `ubuntu-behave-image-build-*` container is not created ephemeral. Should we also move it to ephemeral when the env variable is set? Or was your idea to leave it as it is for the moment as we're switching to schroot based builds?

lucasmoura:
@paride the reason why the `image-build` instances are not ephemeral is because we create snapshots from them. To create a snapshot on [pycloudlib](https://github.com/canonical/pycloudlib/blob/main/pycloudlib/lxd/instance.py#L374) when shutdown the instance, which will not work if is ephemeral.

Also, this will not change if we have the build package stage running on chroot, because those `image-build` instances are the ones that we install the built packages. We snapshot this image to speed up the running all of the tests that requires the latest version of ua

paride:
Thanks @lucasmoura for explaining. About the snapshotting I filed: https://github.com/canonical/pycloudlib/issues/159. Should that indeed work we could expand this PR to also cover the build containers, and go fully ephemeral.

paride (approved):
LGTM and ready to be merged per standup discussion.

Issue #1541, opened 2021-04-09
status: ua status fails if run proir to cloud-init detecting the platform type


blackboxsw:
If cloud-init has not yet detected the datasource, it hasn't yet persisted instance-data,json.  This results in an error from `cloud-id` :

ERROR: File not found '/run/cloud-init/instance-data.json'. Provide a path to instance data json file using --instance-data

This in turn prevents `ua status` from outputting any useful information.
```bash
root@sbqt:~# ua status
Unexpected error(s) occurred.
For more details, see the log: /var/log/ubuntu-advantage.log
To file a bug run: ubuntu-bug ubuntu-advantage-tools
```

renanrodrigo:
I like how GH points to a new commit fixing the issue after each ammend  \\\_("/)\_/
:eyes:

PR #1746, opened 2021-07-21
cli: catch errors when determining cloud type 

renanrodrigo:
`get_cloud_type` now returns a tuple of (result, failure), and if it fails because of `cloud-id`, FIPS enablement has a warning about it.

Fixes: #1541 

## Desired commit type::

## Checklist:

### DIFF
--- a/features/staging_commands.feature
+++ b/features/staging_commands.feature
@@ -542,3 +542,48 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | release | fips-apt-source                                                        |
            | xenial  | https://esm.staging.ubuntu.com/fips-updates/ubuntu xenial-updates/main |
            | bionic  | https://esm.staging.ubuntu.com/fips-updates/ubuntu bionic-updates/main |
+
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: FIPS enablement message when cloud init didn't run properly
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I delete the file `/run/cloud-init/instance-data.json`
+        And I attach `contract_token_staging` with sudo
+        And I run `ua disable livepatch` with sudo
+        And I run `ua enable fips --assume-yes` with sudo
+        Then stderr matches regexp:
+        """
+        Could not determine cloud, defaulting to generic FIPS package.
+        """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+        """
+        fips +yes                enabled
+        """
+
+        Examples: ubuntu release
+        | release |
+        | xenial  |
+        | bionic  |
+
+    @series.focal
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: FIPS enablement message when cloud init didn't run properly
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I delete the file `/run/cloud-init/instance-data.json`
+        And I attach `contract_token_staging` with sudo
+        And I run `ua enable fips --assume-yes` with sudo
+        Then stderr matches regexp:
+        """
+        Could not determine cloud, defaulting to generic FIPS package.
+        """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+        """
+        fips +yes                enabled
+        """
+
+        Examples: ubuntu release
+        | release |
+        | focal   |
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -2,6 +2,8 @@ import json
 import logging
 import os
 
+from enum import Enum
+
 from uaclient import exceptions
 from uaclient import clouds
 from uaclient import status
@@ -9,7 +11,7 @@ from uaclient import util
 from uaclient.config import apply_config_settings_override
 
 try:
-    from typing import Dict, Optional, Type  # noqa: F401
+    from typing import Dict, Optional, Tuple, Type  # noqa: F401
 except ImportError:
     # typing isn't available on trusty, so ignore its absence
     pass
@@ -33,8 +35,13 @@ CLOUD_TYPE_TO_TITLE = {
 PRO_CLOUDS = ["aws", "azure", "gcp"]
 
 
+class NoCloudTypeReason(Enum):
+    NO_CLOUD_DETECTED = 0
+    CLOUD_ID_ERROR = 1
+
+
 def get_instance_id(
-    _iid_file: str = CLOUDINIT_INSTANCE_ID_FILE
+    _iid_file: str = CLOUDINIT_INSTANCE_ID_FILE,
 ) -> "Optional[str]":
     """Query cloud instance-id from cmdline or CLOUDINIT_INSTANCE_ID_FILE"""
     if "trusty" != util.get_platform_info()["series"]:
@@ -50,7 +57,7 @@ def get_instance_id(
 
 
 def get_cloud_type_from_result_file(
-    result_file: str = CLOUDINIT_RESULT_FILE
+    result_file: str = CLOUDINIT_RESULT_FILE,
 ) -> str:
     result = json.loads(util.load_file(result_file))
     dsname = result["v1"]["datasource"].split()[0].lower()
@@ -59,16 +66,19 @@ def get_cloud_type_from_result_file(
 
 
 @apply_config_settings_override("cloud_type")
-def get_cloud_type() -> "Optional[str]":
+def get_cloud_type() -> "Tuple[Optional[str], Optional[NoCloudTypeReason]]":
     if util.which("cloud-id"):
         # Present in cloud-init on >= Xenial
-        out, _err = util.subp(["cloud-id"])
-        return out.strip()
+        try:
+            out, _err = util.subp(["cloud-id"])
+            return (out.strip(), None)
+        except util.ProcessExecutionError:
+            return (None, NoCloudTypeReason.CLOUD_ID_ERROR)
     try:
-        return get_cloud_type_from_result_file()
+        return (get_cloud_type_from_result_file(), None)
     except FileNotFoundError:
         pass
-    return None
+    return (None, NoCloudTypeReason.NO_CLOUD_DETECTED)
 
 
 def cloud_instance_factory() -> clouds.AutoAttachCloudInstance:
@@ -84,7 +94,7 @@ def cloud_instance_factory() -> clouds.AutoAttachCloudInstance:
         "gce": gcp.UAAutoAttachGCPInstance,
     }  # type: Dict[str, Type[clouds.AutoAttachCloudInstance]]
 
-    cloud_type = get_cloud_type()
+    cloud_type, _ = get_cloud_type()
     if not cloud_type:
         raise exceptions.UserFacingError(
             status.MESSAGE_UNABLE_TO_DETERMINE_CLOUD_TYPE
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -173,7 +173,9 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
     def static_affordances(self) -> "Tuple[StaticAffordance, ...]":
         # Use a lambda so we can mock util.is_container in tests
         cloud_titles = {"azure": "an Azure", "gce": "a GCP"}
-        cloud_id = get_cloud_type() or ""
+        cloud_id, _ = get_cloud_type()
+        if cloud_id is None:
+            cloud_id = ""
 
         series = util.get_platform_info().get("series", "")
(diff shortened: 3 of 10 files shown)


uaclient/clouds/identity.py
```diff
@@ -58,12 +60,20 @@ def get_cloud_type_from_result_file(
...
-        return out.strip()
+        try:
+            out, _err = util.subp(["cloud-id"])
+            return out.strip()
+        except util.ProcessExecutionError:
+            logging.warning(
+                "Warning: could not determine cloud ID. "
+                "If you didn't disable cloud-init, "
+                "check cloud-init logs for more information."
+            )
```
lucasmoura, line 76: I think we should return something here, right ? Otherwise we will run the rest of the code unnecessarily

lucasmoura: Also, I think we should have an integration test that explores that situation. We can manually delete the `instance-data.json` file and verify how the warning will be presented to the user

orndorffgrant: I would also tweak the message slightly. Maybe something like:
```
Warning: cloud-init is present but did not run. Assuming not on cloud. If you are on a cloud, you may experience unreliable behavior.
```

Open to additionaly feedback/bikeshedding

renanrodrigo: Thank you for the initial review, all three comments make all the sense.
/ me on that

renanrodrigo:
Integration tests implemented in staging so we can have the same FIPS behavior for all `lts`

mitechie:
I believe that the changes went live today to allow fips to work in
production as well.

renanrodrigo:
Great, @mitechie  - there will be more tests to update in this case.
/me on that

orndorffgrant (commented):
Changes look good! Lets get #1748 merged and rebase this to get a green check before merging.

features/staging_commands.feature
```diff
@@ -542,3 +542,31 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+    Scenario Outline: Commands when cloud init didn't run properly
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I delete the file `/run/cloud-init/instance-data.json`
+        And I attach `contract_token_staging` with sudo
+        Then stderr matches regexp:
+        """
+        Warning: cloud-init is present but did not run. Assuming not on cloud. If you are on a cloud, you may experience unreliable behavior.
+        """
+        When I run `ua status` with sudo
+        Then stderr matches regexp:
```
lucasmoura, line 558: Instead of having that warning appear on the beginning of status, we could add it as a notice and just delete after we print the status:

We should also consider this mechanism for the other commands as well

orndorffgrant: Good idea to use notices for this. Why would we delete the notice after printing the status? Instead, I would expect we'd delete the notice when this function successfully finds a cloud or definitely returns None (any other code path that doesn't result in this warning).

lucasmoura: @orndorffgrant maybe this is a minor thing, but I thought about avoiding have that particular notice be in the `json|yaml` output.

But maybe it is not a bad idea to have it there

renanrodrigo: I think it makes sense that the notice will stay until the problem is solved.

uaclient/clouds/identity.py
```diff
@@ -58,12 +60,20 @@ def get_cloud_type_from_result_file(
...
-        out, _err = util.subp(["cloud-id"])
-        return out.strip()
+        warning_text = status.NOTICE_CLOUD_ID_ERROR
+        try:
+            out, _err = util.subp(["cloud-id"])
+            cfg.remove_notice("", warning_text)
+            return out.strip()
+        except util.ProcessExecutionError:
+            cfg.add_notice("", warning_text)
+            return "cloud-id-error"
```
lucasmoura, line 76: I think we should return `None`. This is just to make this function still consistent with this part of the [code](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/clouds/identity.py#L88)

features/staging_commands.feature
```diff
@@ -542,3 +542,35 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+
+    @series.lts
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Commands when cloud init didn't run properly
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I delete the file `/run/cloud-init/instance-data.json`
+        And I attach `contract_token_staging` with sudo
+        Then stdout matches regexp:
+        """
+        Warning: cloud-init is present but did not run. Assuming not on cloud.
```
lucasmoura, line 555: The message here could be wrong right ? The `instance-data.json` file could not be generated because cloud-init failed on some of the earlier stages of the machine.

However, this is a corner case, so don't know if we should take it into consideration. What do you think @mitechie ?

features/staging_commands.feature
```diff
@@ -542,3 +542,35 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+        If you are on a cloud, you may experience unreliable behavior.
+        """
+        When I run `ua status` with sudo
+        Then stdout matches regexp:
+        """
+        Warning: cloud-init is present but did not run. Assuming not on cloud.
+        If you are on a cloud, you may experience unreliable behavior.
+        """
+        When I attempt to run `ua enable fips` with sudo
```
lucasmoura, line 564: Why are we using attempt here ? I think we should make sure that we still enable fips even if we detect that issue.

features/staging_commands.feature
renanrodrigo: Could the message be something like `did not run properly`?

features/staging_commands.feature
renanrodrigo: Running `ua enable fips --assume-yes` skips the pre-enablement messages, since there is no confirmation step. Running it without the `--assume-yes` flag leads to an error in the test, so I am attempting here to see the message.
I tried running it with `and stdin "y\n" for the command to complete, but I had no success.

uaclient/clouds/identity.py
renanrodrigo: Got it. But I think we should somehow separate the cases.
I agree that if there is an error, we could not identify the cloud type when instantiating, but returning `None` and checking it afterwards would trigger the same warning during FIPS installation when we are not running in the cloud.
Maybe there is a better way to diff between the two situations.

features/staging_commands.feature
lucasmoura: It could, but then we would take away the case that cloud-init was disabled by the user

features/staging_commands.feature
lucasmoura: I think we should run that verification passing no first and then enabling fips to see if everything works as expected

uaclient/clouds/identity.py
lucasmoura: We should probably return a tuple then, with the cloud-type and reason for failure. Similar to what we do in the `can_enable` method

renanrodrigo: good good, /me on it

features/staging_commands.feature
renanrodrigo: If it was disabled by the user, then it did not run - makes sense imho

lucasmoura: In my understanding we are talking about either:

* `cloud-init is present but did not run.`: This will cover the case that cloud-init is disabled, but fail to cover when it actually failed to run.
* `cloud-init did not run properly`: Which will cover the failed scenario but not the disabled one.

But I can be missing something about the message change you are proposing

orndorffgrant: Side note: the actual case that brought this to our attention recently was while building an image: cloud-init has not run _yet_.

How about `cloud-init is present but has not detected the cloud`. I think that is still accurate for all the scenarios where this could occur

features/staging_commands.feature
```diff
@@ -542,3 +542,36 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+        not on cloud. If you are on a cloud, you may experience unreliable behavior.
+        """
+        When I run `ua status` with sudo
+        Then stdout matches regexp:
+        """
+        Warning: cloud-init is present but has not detected the cloud. Assuming
+        not on cloud. If you are on a cloud, you may experience unreliable behavior.
+        """
+        When I run `ua disable livepatch` with sudo
+        And I run `ua enable fips` `with sudo` and stdin `y`
```
lucasmoura, line 565: I think we should only add one more step here to guarantee that fips was enabled in the system, like checking `ua status` for example

features/staging_commands.feature
```diff
@@ -542,3 +542,36 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
            | xenial  | https://esm.staging.ubuntu.com/fips-updates/ubuntu xenial-updates/main |
            | bionic  | https://esm.staging.ubuntu.com/fips-updates/ubuntu bionic-updates/main |
+
+
+    @series.lts
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Commands when cloud init didn't run properly
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I delete the file `/run/cloud-init/instance-data.json`
+        And I attach `contract_token_staging` with sudo
```
lucasmoura, line 552: Can we use just `contract_token` here ?

uaclient/clouds/identity.py
```diff
@@ -84,7 +100,7 @@ def cloud_instance_factory() -> clouds.AutoAttachCloudInstance:
         "gce": gcp.UAAutoAttachGCPInstance,
     }  # type: Dict[str, Type[clouds.AutoAttachCloudInstance]]
 
-    cloud_type = get_cloud_type()
+    cloud_type, _error = get_cloud_type()
```
lucasmoura, line 103: I think we can just ignore the variable, right ?
`cloud_type, _ = get_cloud_type()`

uaclient/entitlements/fips.py
```diff
@@ -173,7 +173,9 @@ def _allow_fips_on_cloud_instance(
     def static_affordances(self) -> "Tuple[StaticAffordance, ...]":
         # Use a lambda so we can mock util.is_container in tests
         cloud_titles = {"azure": "an Azure", "gce": "a GCP"}
-        cloud_id = get_cloud_type() or ""
+        cloud_id, _error = get_cloud_type()
```
lucasmoura, line 176: Same thing here about ignoring the error var

uaclient/entitlements/fips.py
```diff
@@ -216,9 +218,11 @@ def _replace_metapackage_on_cloud_instance(
         if series != "bionic":
             return packages
 
-        cloud_match = re.match(
-            r"^(?P<cloud>(azure|aws)).*", get_cloud_type() or ""
-        )
+        cloud_id, _error = get_cloud_type()
```
lucasmoura, line 221: Same thing here

features/staging_commands.feature
renanrodrigo: Probably yes. I would like to keep it in the staging environment so far, and then review all of the staging tests as a separate task.

blackboxsw (approved):
Thanks @renanrodrigo works for me. Let's take that followup PR to move the integration test out of staging_commands.feature and into attached_enable.feature as a separate PR. Otherwise looks good.

features/staging_commands.feature
```diff
@@ -542,3 +542,26 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | release | fips-apt-source                                                        |
            | xenial  | https://esm.staging.ubuntu.com/fips-updates/ubuntu xenial-updates/main |
            | bionic  | https://esm.staging.ubuntu.com/fips-updates/ubuntu bionic-updates/main |
+
+    @series.lts
```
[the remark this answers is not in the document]
renanrodrigo: Will do - there wll be more of them moving (:

renanrodrigo:
I have splitted integration tests to cover that livepatch disablement failure on focal. Now it should be good.

features/staging_commands.feature
```diff
@@ -542,3 +542,49 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+        Then stdout matches regexp:
+        """
+        fips +yes                enabled
+        """
+
+        Examples: ubuntu release
+        | release |
+        | xenial  |
+        | bionic  |
+        | focal   |
```
lucasmoura, line 569: I think we can remove this focal entry here

renanrodrigo: true true

renanrodrigo:
Once you go green, you never go back :D

renanrodrigo:
Failing test is unrelated

lucasmoura (approved):

lucasmoura:
It is worth looking at those failed integration tests, but I agree they are unrelated with the proposed changes

PR #1747, opened 2021-07-22
tests: only use -proposed for ua (SC-270)

orndorffgrant:
## Proposed Commit Message

```
tests: only use -proposed for ua
```

## Test Steps

This is hard to test now that our latest release is in -updates. I suggest the following in a fresh xenial container:
1. Create the files that the new `printf` commands in this PR are creating
2. Change the specific package in `/etc/apt/preferences.d/uaclient-proposed` to `autopkgtest`
3. `apt update`
4. `apt install autopkgtest`
5. `apt policy autopkgtest`
    * Ensure that -proposed is pinned at 400 but this package is pinned at 1001. Make sure the version of autopkgtest from proposed is installed
6. `apt upgrade`
    * Ensure that nothing else is installed from -proposed

Also, still run `env UACLIENT_BEHAVE_ENABLE_PROPOSED=1 tox -e behave-vm-16.04 -- features/staging_commands.feature:479 --no-junit` and ensure it passes.

## Desired commit type::

## Checklist:

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -44,7 +44,9 @@ runcmd:
 
 USERDATA_RUNCMD_ENABLE_PROPOSED = """
 runcmd:
-  - printf \"deb http://archive.ubuntu.com/ubuntu/ {series}-proposed restricted main multiverse universe\" > /etc/apt/sources.list.d/uaclient-proposed.list
+  - printf \"deb http://archive.ubuntu.com/ubuntu/ {series}-proposed main\" > /etc/apt/sources.list.d/uaclient-proposed.list
+  - "printf \\"Package: *\\nPin: release a={series}-proposed\\nPin-Priority: 400\\n\\" > /etc/apt/preferences.d/lower-proposed"
+  - "printf \\"Package: ubuntu-advantage-tools\\nPin: release a={series}-proposed\\nPin-Priority: 1001\\n\\" > /etc/apt/preferences.d/uaclient-proposed"
 """  # noqa: E501
 
 USERDATA_APT_SOURCE_PPA = """\


features/environment.py
```diff
@@ -848,6 +848,11 @@ def _install_uat_in_container(
             )
         )
 
+    if config.enable_proposed:
+        cmds.append(
+            ["sudo", "rm", "/etc/apt/sources.list.d/uaclient-proposed.list"]
```
renanrodrigo, line 853: Should we run also `apt update` after this, just to make sure?

orndorffgrant: Ah good catch - yeah we probably should.

orndorffgrant: Just added this

renanrodrigo (approved):
lgtm

lucasmoura (commented):
The only problem with this approach is that we will not upgrade ua during `do-release-upgrade`.
Therefore, the upgrade test will not fully exercise the scenario they were intended to test.
I think it is better for us to keep the proposed pocket enabled and if we find a problem in the upgrade tests regarding proposed, we use the ppas to avoid that issue

orndorffgrant:
@lucasmoura Excellent point. For example, for the xenial upgrade test, xenial-proposed-ua would not get upgraded to bionic-proposed-ua. Instead, the test ends up keeping xenial-proposed-ua on bionic after the `do-release-upgrade`.

Is there any way to only enable -proposed for a single package?

renanrodrigo:
@orndorffgrant maybe we could set cache-policies for proposal to be minimal priority, and raise it only for `ubuntu-advantage-tools`

paride:
It would make sense if `RELEASE-proposed` somehow had lower priority then `RELEASE` when added to `sources.list`, as adding `-proposed` is mostly done to test individual packages or sets of packages. Upgrading everything from `-proposed` is in general not a good idea. Unfortunately we're not there, see:
 - https://bugs.launchpad.net/launchpad/+bug/1016776
 - https://lists.ubuntu.com/archives/ubuntu-devel/2021-January/041324.html (not implemented afaik)

The way to avoid full upgrades from `-proposed` is to pin it to a lower priority than the release: https://wiki.ubuntu.com/Testing/EnableProposed. Then packages from `-proposed` can be installed via `apt -t <target_release> install ...`

orndorffgrant:
@lucasmoura @renanrodrigo @paride Thanks for all of the guidance! This is updated now to use the pinning strategy to enable proposed but only use it for uaclient by default. Let me know what you think

features/environment.py
```diff
@@ -45,6 +45,8 @@
 USERDATA_RUNCMD_ENABLE_PROPOSED = """
 runcmd:
   - printf \"deb http://archive.ubuntu.com/ubuntu/ {series}-proposed restricted main multiverse universe\" > /etc/apt/sources.list.d/uaclient-proposed.list
```
paride, line 47: `main` should be enough here, for the reset this LGTM!

paride (changes requested):
One little inline comment, not on the lines added by this PR which LGTM, but let's not waste the occasion :-)

orndorffgrant:
@paride totally agree! Just pushed a fix

paride (approved):
LGTM

renanrodrigo (approved):

lucasmoura (approved):

PR #1751, opened 2021-07-26
tests: default to installing ua for each scenario (SC-268)

orndorffgrant:
## Background

Initial investigation into the cost of the snapshot strategy is summarized here: https://docs.google.com/spreadsheets/d/1ceahTwO7wy_Kb2i2nw7oII2PaGqm_26ZNSKw5ema4FI/edit#gid=0

Please ask if anything in there doesn't make sense.

## Proposed Commit Message

```
tests: default to installing ua for each scenario

The strategy of creating a snapshot with ua installed before running any
integration test has a significant upfront cost. There are cases, such
as local development on one feature, where this upfront cost is not
worth it. This commit adds an option so that the snapshot strategy with
upfront cost can be optionally turned on, but defaults to off.
```

## Test Steps

Run some integration tests with and without snapshot_strategy.
Use `--no-junit` so you can see the timestamps and logs to understand what is happening with and without the envvar.

Examples:
```
tox -e behave-vm-18.04 -- --no-junit
env UACLIENT_BEHAVE_SNAPSHOT_STRATEGY=1 tox -e behave-vm-18.04 -- --no-junit
```

## Desired commit type::

## Checklist:

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -17,13 +17,34 @@ from hamcrest import (
     contains_string,
 )
 
-from features.environment import create_uat_image
+from features.environment import (
+    create_instance_with_uat_installed,
+    capture_container_as_image,
+)
 from features.util import SLOW_CMDS, emit_spinner_on_travis, nullcontext
 
 from uaclient.defaults import DEFAULT_CONFIG_FILE, DEFAULT_MACHINE_TOKEN_PATH
 
 
-CONTAINER_PREFIX = "ubuntu-behave-test-"
+CONTAINER_PREFIX = "ubuntu-behave-test"
+IMAGE_BUILD_PREFIX = "ubuntu-behave-image-build"
+IMAGE_PREFIX = "ubuntu-behave-image"
+
+
+def add_test_name_suffix(context, series, prefix):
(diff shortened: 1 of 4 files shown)


features/steps/steps.py
```diff
@@ -17,13 +17,42 @@
...
+
+
+def add_test_name_suffix(context, series, prefix):
+    pr_number = os.environ.get("UACLIENT_BEHAVE_JENKINS_CHANGE_ID")
+    pr_suffix = "-" + str(pr_number) if pr_number else ""
+    is_vm = bool(context.config.machine_type == "lxd.vm")
+    vm_suffix = "-vm" if is_vm else ""
+    time_suffix = datetime.datetime.now().strftime("-%s%f")
+
+    name = "{prefix}{pr_suffix}{vm_suffix}-{series}{time_suffix}".format(
```
lucasmoura, line 41: I think we can remove this var

orndorffgrant: Ah yep, I had pulled that out as a var while debugging and forgot to delete - thanks for catching it!

lucasmoura (approved):

PR #1753, opened 2021-07-29
lib: add timer script (SC-284)

lucasmoura:
## Proposed Commit Message
lib: add timer script

Add an initial implementation of the timer script. This script has the logic to verify if a job should run in a given moment and
it also persist information about all jobs that have run during its execution

## Test Steps
Run the unit tests added for this script.
However, there are a lot more tests that should be implemented

## Desired commit type::

## Checklist:

### DIFF
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -1,0 +1,114 @@
+"""
+Timer script used to run all jobs that need to be frequently run on the system
+"""
+
+import json
+import logging
+
+import datetime
+
+from collections import OrderedDict
+from typing import Any, Callable, Dict  # noqa: F401
+
+from uaclient.config import UAConfig
+from uaclient import util
+from uaclient.jobs.update_messaging import update_apt_and_motd_messages
+
+
+LOG = logging.getLogger(__name__)
+
+# We are currently storing the time frequency for each job at this dict,
+# but we should move it later into uaclient.conf
+uaclient_jobs = OrderedDict(
+    {
+        "update_messaging": {
+            "job_func": update_apt_and_motd_messages,
+            "run_interval_seconds": 14400,
+        },
+        "update_status": {
+            "job_func": lambda cfg: None,
+            "run_interval_seconds": 14400,
+        },
+    }
+)
+
+
+def run_job(
+    job_name: str,
+    job_func: "Callable[[UAConfig], None]",
+    current_time: datetime.datetime,
+    run_interval_seconds: int,
+    jobs_status: "Dict[str, Any]",
+    cfg: UAConfig,
+) -> "Dict[str, Any]":
+    """Run a job based on time constraints.
+
+    If the current time is higher than the last time
+    the job has successfully executed plus the expected
+    time interval between job runs, we need to execute this
+    job again.
+
+    :param job_name: Name of the job to run.
+    :param job_func: Job function to run
+    :param run_interval_seconds: Interval in seconds that the job needs to
+                                 wait before running again.
+    :param jobs_status: A dict containing information about the last
+                        successful run of all jobs that should run
+                        in the system.
+
+    :return: A dict containing the update last successful run of the job
+             and the time it should execute next
+    """
+    seconds_to_add = datetime.timedelta(seconds=run_interval_seconds)
+    if job_name in jobs_status:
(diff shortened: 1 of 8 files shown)


blackboxsw (changes requested):
Looks like a good general approach and I like moving things out of lib and into uaclent.jobs

lib/timer.py
```diff
@@ -0,0 +1,66 @@
...
+uaclient_jobs = {
+    "update_messaging": {
+        "job_func": update_apt_and_motd_messages,
+        "time_frequency": 4,
+    },
+    "update_status": {"job_func": lambda: None, "time_frequency": 4},
+}
+
+
+def get_current_time():
```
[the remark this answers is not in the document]
lucasmoura: Yes, this is just to make it easier to mock `datetime.datetime.now()`
It seems that mocking it directly was not working, but I can be missing something here

lib/timer.py
```diff
@@ -0,0 +1,66 @@
...
+from uaclient.config import UAConfig
+from uaclient import util
+from uaclient.jobs.update_messaging import update_apt_and_motd_messages
+
+LOG = logging.getLogger(__name__)
+
+# We are currently storing the time frequency for each job at this dict,
+# but we should move it later into uaclient.conf
+uaclient_jobs = {
+    "update_messaging": {
```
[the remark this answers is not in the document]
lucasmoura: We could also define then in an `OrderedDict`, which I think is the easier approach here

lib/timer.py
```diff
@@ -0,0 +1,126 @@
...
+        next_run = jobs_status[job_name]["next_run"]
+
+        if current_time < next_run:
+            return {}
+
+    try:
+        job_func(cfg)
+    except Exception as e:
+        LOG.warning("Error executing job: %s", job_name)
+        LOG.warning("The following exception was raised:\n%s", str(e))
```
blackboxsw, line 80: We could consolidate these two warnings into a single log
```suggestion
        LOG.warning("Error executing job: %s. %s", job_name, e)
```

lib/timer.py
```diff
@@ -0,0 +1,126 @@
...
+    :return: A dict containing the update last successful run of the job
+             and the time it should execute next
+    """
+    seconds_to_add = datetime.timedelta(seconds=run_interval_seconds)
+    if job_name in jobs_status:
+        next_run = jobs_status[job_name]["next_run"]
+
+        if current_time < next_run:
+            return {}
+
```
blackboxsw, line 75: We don't need to jump into run_job to perform this check, I think we should avoid the extra function calls with a pre-flight check in the for loop you have in main and avoid calling this function if unneeded.

lucasmoura: I agree that we could make that check on `main`, but this will make it harder to perform unittests. For example, we would bring back the `datetime` mock issue back into the table. Also, I don't know if it good to have that logic backed directly into main.

But let me know if you think the test trade-off here is worth it

lib/timer.py
```diff
@@ -0,0 +1,126 @@
...
+    :param run_interval_seconds: Interval in seconds that the job needs to
+                                 wait before running again.
+    :param jobs_status: A dict containing information about the last
+                        successful run of all jobs that should run
+                        in the system.
+
+    :return: A dict containing the update last successful run of the job
+             and the time it should execute next
+    """
+    seconds_to_add = datetime.timedelta(seconds=run_interval_seconds)
```
blackboxsw, line 69: Let's add a bookend LOG.debug("Running job: %s", job_name)  when we are actually performing a given job so we get logs when each job is started.

lib/timer.py
blackboxsw: @lucasmoura since we likely won't have more than 10 of these jobs running in serial, it's not really going to cost us much to call into `run_job` for every job listed, so I'm good with it as is.

blackboxsw (approved):
Added a commit to filter jobs before calling `run_job`. Also simplified call signature and return value of` run_job`. Dropped empty "update_status" placeholder. We will add an `update_status` or `update_metering` function in a separate PR.

renanrodrigo (approved):
lgtm

PR #1754, opened 2021-08-02
postinst: ignore any cloud-id error on postinst (SC-280)

renanrodrigo:
Cloud ID is used during postinst to warn users about potentially wrong FIPS packages on Azure or AWS.
If `cloud-id` fails on top of those, then there would be worse problems with the instance.
If `cloud-id` fails anywhere else, the message is unnecessary.

Thus, we can completely ignore `cloud-id` fails here, and make sure we don't fail.

## Desired commit type::

## Checklist:

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -290,10 +290,7 @@ notify_wrong_fips_metapackage_on_cloud() {
 
     fips_metapkg="ubuntu-fips"
 
-    cloud_id=""
-    if command -v "cloud-id" > /dev/null && [ -f /run/cloud-init/instance-data.json ] ; then
-      cloud_id=$(cloud-id)
-    fi
+    cloud_id=$(cloud-id 2>/dev/null) || cloud_id=""
 
     # If the package is not installed, we don't want the postinst script to fail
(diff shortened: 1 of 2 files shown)


debian/ubuntu-advantage-tools.postinst
```diff
@@ -290,10 +290,7 @@ notify_wrong_fips_metapackage_on_cloud() {
 
     fips_metapkg="ubuntu-fips"
 
-    cloud_id=""
-    if command -v "cloud-id" > /dev/null && [ -f /run/cloud-init/instance-data.json ] ; then
-      cloud_id=$(cloud-id)
-    fi
+    cloud_id=$(cloud-id 2>/dev/null || true)
```
paride, line 293: This byte-by-byte how I thought to implement this, however it still makes an assumption on `cloud-id`: that it won't print anything to stdout on error. If it does, we'll end up setting `cloud_id` to a bad value, like an error message. I'd go for:
```
cloud_id=$(cloud-id 2>/dev/null) || cloud_id=""
```

paride (changes requested):
Thanks! Just an online comment.

debian/ubuntu-advantage-tools.postinst
renanrodrigo: True true, nice catch
I have updated the code

paride (approved):
lgtm!

PR #1757, opened 2021-08-04
contract: try to get machine ID from the config (SC-294)

renanrodrigo:
If we receive a value for machine ID, it is persisted in config and should be preferred over any other value.

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -293,12 +293,18 @@ def get_dict_deltas(
 
 
 @lru_cache(maxsize=None)
-def get_machine_id(data_dir: str) -> str:
+def get_machine_id(cfg) -> str:
     """Get system's unique machine-id or create our own in data_dir."""
-    # Generate, cache our own uuid if not present on the system
+    # Generate, cache our own uuid if not present in config or on the system
     # Docker images do not define ETC_MACHINE_ID or DBUS_MACHINE_ID on trusty
     # per Issue: #489
-    fallback_machine_id_file = os.path.join(data_dir, "machine-id")
+
+    if cfg.machine_token:
+        cfg_machine_id = cfg.machine_token.get("machineId")
+        if cfg_machine_id:
+            return cfg_machine_id
+
+    fallback_machine_id_file = os.path.join(cfg.data_dir, "machine-id")
 
     for path in [ETC_MACHINE_ID, DBUS_MACHINE_ID, fallback_machine_id_file]:
         if os.path.exists(path):
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -86,6 +86,7 @@ class UAContractClient(serviceclient.UAServiceClient):
             API_V1_CONTEXT_MACHINE_TOKEN, data=data, headers=headers
         )
         self.cfg.write_cache("machine-token", machine_token)
+        util.get_machine_id.cache_clear()
         return machine_token
 
     def request_resources(self) -> "Dict[str, Any]":
@@ -136,7 +137,7 @@ class UAContractClient(serviceclient.UAServiceClient):
         @return: Dict of the JSON response containing entitlement accessInfo.
         """
         if not machine_id:
-            machine_id = util.get_machine_id(self.cfg.data_dir)
+            machine_id = util.get_machine_id(self.cfg)
         headers = self.headers()
         headers.update({"Authorization": "Bearer {}".format(machine_token)})
         url = API_V1_TMPL_RESOURCE_MACHINE_ACCESS.format(
@@ -219,13 +220,14 @@ class UAContractClient(serviceclient.UAServiceClient):
             response["expires"] = headers["expires"]
         if not detach:
             self.cfg.write_cache("machine-token", response)
+            util.get_machine_id.cache_clear()
             self.cfg.write_cache("machine-id", data.get("machineId", ""))
         return response
 
     def _get_platform_data(self, machine_id):
         """"Return a dict of platform-relateddata for contract requests"""
         if not machine_id:
-            machine_id = util.get_machine_id(self.cfg.data_dir)
+            machine_id = util.get_machine_id(self.cfg)
         platform = util.get_platform_info()
         arch = platform.pop("arch")
         return {"machineId": machine_id, "architecture": arch, "os": platform}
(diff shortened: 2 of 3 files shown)


uaclient/util.py
```diff
@@ -293,12 +293,18 @@ def get_dict_deltas(
 
 
 @lru_cache(maxsize=None)
-def get_machine_id(data_dir: str) -> str:
+def get_machine_id(cfg) -> str:
```
blackboxsw, line 296: Let's move this function into uaclient.contract since that's our only callsite, It doesn't make sense to pollute uaclient.util with this function and cope with nested import issues.
Also let's use typing on everything we touch.... 
so this could be either 
```suggestion
def get_machine_id(cfg: UAConfig) -> str:
```


Or maybe we don't even need to pass in cfg, but instead just accept a fallback_machine_id_file
which can be obtained at the call_site with get_machine_id(self.cfg.data_path("machine-id"))

uaclient/util.py
```diff
@@ -293,12 +293,18 @@ def get_dict_deltas(
...
-    fallback_machine_id_file = os.path.join(data_dir, "machine-id")
+
+    if cfg.machine_token:
+        cfg_machine_id = cfg.machine_token.get("machineId")
+        if cfg_machine_id:
+            return cfg_machine_id
+
+    fallback_machine_id_file = os.path.join(cfg.data_dir, "machine-id")
 
     for path in [ETC_MACHINE_ID, DBUS_MACHINE_ID, fallback_machine_id_file]:
```
blackboxsw, line 309: What do you think if we reorder this check so that we source fallback_machine_id_file first if present. The only way it would be present is if we have written it already either due to falling through to the uuid4 generation below on first call or if we add a `self.cfg.write_cache("machine-id", response.get("machineId") if "machineId" in machine_token` over in [UAContractClient.request_contract_machine_attach](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/contract.py#L88) and [UAContractClient._request_machine_token_update](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/contract.py#L221)

blackboxsw (commented):
Looks pretty good. Can we:
- change the parameter we pass into get_machine_id to fallback_machine_id_file so we can drop logic of sourcing it from data_dir anyway (because I actually see a bug that we are writing /var/lib/ubuntu-advantage/machine-id instead of /var/lib/ubuntu-advantage/private/machine-id)
- think about changing the order of files checked to place fallback_machine_id_file first
- think about writing machine-id from machine_token["machineID"] if present in the machine token response
- think about moving get_machine_id into uaclient.contract module and related get_machine_id tests out of uaclient/tests/test_util.py  and intl test_contract.py

uaclient/util.py
renanrodrigo: About moving it to `uaclient.contract`, my only question would be if we have any expectation of using it somewhere else. Getting the machine ID seems generic enough imho, but if we will keep using it only in contracts, then I agree that it makes sense.

uaclient/util.py
blackboxsw: In side discussion @renanrodrigo pointed out we can't prefer the fallback_machine_id_file above ETC and DBUS sources because those will change across cloned images on some clouds and we don't have visibliity to that change if we continue to prefer our now stale <data_path>/machine-id cache. So we can't reorder.

We decided as well that we need to clear the lru_cache for `util.get_machine_id.clear_cache()` in UAContractClient whenever a new machine_token response is persisted if the 'machineId' key exists in the machine-token and the token has changed. This avoids the cache on get_machine_id returning the previous machine_id which was pulled from ETC or DBUS.

blackboxsw:
Thanks for the discussion @renanrodrigo, I think we decided the following:


>     * change the parameter we pass into get_machine_id to fallback_machine_id_file so we can drop logic of sourcing it from data_dir anyway (because I actually see a bug that we are writing /var/lib/ubuntu-advantage/machine-id instead of /var/lib/ubuntu-advantage/private/machine-id)

  We won't change parameter from UAConfig type because we need to check cfg.machine_token in get_machine_id in order to prefer "machineId"  key if it exists from the contract server.

>     * think about changing the order of files checked to place fallback_machine_id_file first

We won't change order because it'd ignore deltas in ETC and DBUS machine IDs in the OS for clouds that **do** update machine-id on cloned images. We should still source those machine-ids if present (and not overridden by the contract server machine-token)

>     * think about writing machine-id from machine_token["machineID"] if present in the machine token response
  We also don't want to write cfg.write_cache("machine-id") as this is a backup machineID used only in two cases:
 - persisted generated machine-id  when not /etc/ or dbus machine id is present
 - persisted machine-id used with attaching to double-check locally if our image has been cloned and new /etc or dbus machine-id != to previous_machine_id.
 
We don't really want to prefer this previous machine-id cache as we will rely on the contact server metering/activity routes to change our machineID for us if it detects duplicates. At which point uaclient will persist that new value in the new machine-token.json we write.

> 
>     * think about moving get_machine_id into uaclient.contract module and related get_machine_id tests out of uaclient/tests/test_util.py  and intl test_contract.py

Busy work that we don't really have a firm need to do at the moment. We can get by without the typing hints unless we us docstr or string 'UAClient' hints in the file.


All said I think the only thing to fix then is that UAContractClient should probably run get_machine_id.clear_cache() to clear our LRU cache when we write out a new machine-token.

renanrodrigo:
> We don't really want to prefer this previous machine-id cache as we will rely on the contact server metering/activity routes to change our machineID for us if it detects duplicates. At which point uaclient will persist that new value in the new machine-token.json we write.

Indeed - keeping in mind that once this is the default for all machines, we would have to review the check-for-dupes logic on our side - if not discard it, depending on the contract server behavior.

> All said I think the only thing to fix then is that UAContractClient should probably run get_machine_id.clear_cache() to clear our LRU cache when we write out a new machine-token.

/me on it

renanrodrigo:
@blackboxsw done. Do you think we need to write a unit test for this clear_cache operation? If the answer is yes, any hint on how?

blackboxsw:
> @blackboxsw done. Do you think we need to write a unit test for this clear_cache operation? If the answer is yes, any hint on how?

Let's avoid it for this PR. We can definitely cover this aspect in testing of updated machineIds from  contract server in the next PR needed on this general feature which will allow us to fake new machineIds from the contract server.

blackboxsw (approved):
+1 LGTM!

PR #1759, opened 2021-08-09
Integration tests review

renanrodrigo:
Test pull request to check tests behavior on Jenkins CI


renanrodrigo:
@blackboxsw thanks for your patience and help on this one :upside_down_face: 
It's now ready for review

blackboxsw (approved):
This was a large chunk of work to sort out those test cases. Thanks @renanrodrigo for working through this and making it easier to visualize where tests should be added and filling some of the existing gaps.  There is always more to do here, but this is a good few leaps in the right direction.

PR #1761, opened 2021-08-09
timer: add metering job (SC-301)

lucasmoura:
## Proposed Commit Message
timer: add metering job

Add metering job to inform contract about current activity token and enabled services in the machine

## Test Steps
Run the new integration test

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -104,7 +104,7 @@ def assert_lock_file(lock_holder=None):
 
     def wrapper(f):
         @wraps(f)
-        def new_f(args, cfg, **kwargs):
+        def new_f(*args, cfg, **kwargs):
             global _CLEAR_LOCK_FILE
             (lock_pid, cur_lock_holder) = cfg.check_lock_info()
             if lock_pid > 0:
@@ -117,7 +117,7 @@ def assert_lock_file(lock_holder=None):
             notice_msg = "Operation in progress: {}".format(lock_holder)
             cfg.add_notice("", notice_msg)
             _CLEAR_LOCK_FILE = cfg.delete_cache_key
-            retval = f(args, cfg, **kwargs)
+            retval = f(*args, cfg=cfg, **kwargs)
             cfg.delete_cache_key("lock")
             _CLEAR_LOCK_FILE = None  # Unset due to successful lock delete
             return retval
@@ -157,7 +157,7 @@ def assert_attached(unattached_msg_tmpl=None):
                 else:
                     exception = exceptions.UnattachedError()
                 raise exception
-            return f(args, cfg, **kwargs)
+            return f(args, cfg=cfg, **kwargs)
 
         return new_f
 
@@ -171,7 +171,7 @@ def assert_not_attached(f):
     def new_f(args, cfg):
         if cfg.is_attached:
             raise exceptions.AlreadyAttachedError(cfg)
-        return f(args, cfg)
+        return f(args, cfg=cfg)
 
     return new_f
 
@@ -340,7 +340,7 @@ def refresh_parser(parser):
     return parser
 
 
-def action_fix(args, cfg, **kwargs):
+def action_fix(args, *, cfg, **kwargs):
     if not re.match(security.CVE_OR_USN_REGEX, args.security_issue):
         msg = (
             'Error: issue "{}" is not recognized.\n'
@@ -562,7 +562,7 @@ def get_valid_entitlement_names(names: List[str]):
     return entitlements_found, entitlements_not_found
 
 
-def action_config(args, cfg, **kwargs):
+def action_config(args, *, cfg, **kwargs):
     """Perform the config action.
 
     :return: 0 on success, 1 otherwise
@@ -577,7 +577,7 @@ def action_config(args, cfg, **kwargs):
         )
 
 
-def action_config_show(args, cfg, **kwargs):
+def action_config_show(args, *, cfg, **kwargs):
     """Perform the 'config show' action optionally limit output to a single key
 
     :return: 0 on success
@@ -608,7 +608,7 @@ def action_config_show(args, cfg, **kwargs):
 
 
 @assert_root
-def action_config_set(args, cfg, **kwargs):
+def action_config_set(args, *, cfg, **kwargs):
     """Perform the 'config set' action.
 
     @return: 0 on success, 1 otherwise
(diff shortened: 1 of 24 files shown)


uaclient/contract.py
```diff
@@ -162,6 +163,29 @@ def request_machine_token_update(
             detach=False,
         )
 
+    def request_machine_activity(
+        self,
+        contract_id: str,
+        machine_id: str,
```
blackboxsw, line 169: I think we also need a machine_token in this method which needs to be placed in headers of the request 
headers.update({"Authorization": "Bearer {}".format(machine_token)})
We can test this against contracts.develop.canonical.com. even though staging doesn't have it.

uaclient/contract.py
```diff
@@ -162,6 +163,29 @@ def request_machine_token_update(
...
+        enabled services in the system.
+        """
+        request_data = {
+            "activity_token": activity_token,
+            "resources": enabled_services,
+        }
+        url = API_V1_MACHINE_ACTIVITY.format(
+            contract=contract_id, machine=machine_id
+        )
+        headers = self.headers()
```
blackboxsw, line 185:
```suggestion
        headers = self.headers()
        headers.update({"Authorization": "Bearer {}".format(machine_token)})
```

uaclient/contract.py
```diff
@@ -162,6 +163,29 @@ def request_machine_token_update(
...
+
+        return self.request_url(url, headers=headers, data=request_data)
```
blackboxsw, line 187: This response is our machine-token.json so we probably want to persist the token updates inside this method as we do in the UAContractClient attach/update methods instead of outside this method with  self.cfg.write_cache("machine-token", response).

uaclient/contract.py
```diff
@@ -225,7 +249,7 @@ def _request_machine_token_update(
         return response
 
     def _get_platform_data(self, machine_id):
-        """"Return a dict of platform-relateddata for contract requests"""
+        """ "Return a dict of platform-relateddata for contract requests"""
```
blackboxsw, line 252:
```suggestion
        """Return a dict of platform-related data for contract requests"""
```

uaclient/contract.py
lucasmoura: I think the only problem of persisting this data here is the problem that we need to update the enabled entitlements for possible resourceToken changes. Do you think I should move that entire logic here ?

renanrodrigo: This might as well be a subset of the machine-token, so we can think about updating only the provided keys.

blackboxsw (changes requested):
Looks pretty good, thanks for the additional check and thank you for dropping consideration of resourceToken/contract deltas for this PR. While we await a response from contracts team on whether updating all resourceTokens is truly needed, it doesn't make sense for us to bake in that risk/logic to this iteration.

I think we do need to use the @assert_lock_file decoration for this report_machine_activity func because we will be persisting machine-token.json and machine-id files across this call so we don't want to get multiple writer race conditions.

uaclient/jobs/metering.py
```diff
@@ -0,0 +1,34 @@
...
+    # We only run this job if there is no other job running.
+    # The reason for that is to avoid potential conflicts with
+    # auto-attach, attach and enable operations.
+    lock_pid, curr_lock_holder = cfg.check_lock_info()
+    if lock_pid > 0:
+        raise exceptions.LockHeldError(
+            lock_request="metering job",
+            lock_holder=curr_lock_holder,
+            pid=lock_pid,
+        )
```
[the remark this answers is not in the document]
blackboxsw: Hrm, I'll leave the decision to you on this PR. @assert_lock_file currently mandates that decorated functions take a positional "args" and "cfg" param. We could make that decorator more flexible, but I'd like that work to be a separate PR so we can spend more concentrated effort on a review of that lock handling. If you can create a separate card/ticket to address lock holding on metering_enabled_resources as an item we can work on later. then I'm +1 on this branch.

blackboxsw (approved):
@lucasmoura content is good and it is a solid step toward the metering service.
One outstanding issue I see is that `metering_enabled_resources` needs to also create and hold the lock since we will be writing updates to machine-token,json. But, let's wait until a followup branch for that work.

lucasmoura:
@blackboxsw I have updated the metering job to use the `assert_lock_file` directly. I had to action functions on the cli module to allow that reuse. Let me know if that update makes sense

blackboxsw (changes requested):
@lucasmoura we had a discussion yesterday with JCSackett and there might be a change forthcoming on the POST data needed for the machine-activity endpoint that may required a new/required `activityID` key. We'll spec this out today, but this branch will be unable to land as that endpoint is in flux.

uaclient/cli.py
```diff
@@ -117,9 +117,15 @@ def new_f(args, cfg, **kwargs):
...
-            cfg.delete_cache_key("lock")
-            _CLEAR_LOCK_FILE = None  # Unset due to successful lock delete
+
+            try:
+                retval = f(*args, cfg=cfg, **kwargs)
+            except Exception as e:
+                raise e
+            finally:
+                cfg.delete_cache_key("lock")
+                _CLEAR_LOCK_FILE = None  # Unset due to successful lock delete
```
blackboxsw, line 127: Not really a blocker, but I don't think we need the except/raise here either, the exception will be raised even without just the try/finally.

```suggestion
            finally:
                cfg.delete_cache_key("lock")
                _CLEAR_LOCK_FILE = None  # Unset due to successful lock delete
```

lucasmoura:
@blackboxsw code updated

blackboxsw:
I think you have a race condition in the ua config show test in attached_commands.feature:503/504/505 etc. If the timer job runs during intial boot gcp timer will end up being 0 instead of the default 1800. To fix it I think we need to turn off the timer so it doesn't run in this test

```diff
diff --git a/features/attached_commands.feature b/features/attached_commands.feature
index 2fd2c8f1..cde0f9db 100644
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -393,7 +393,8 @@ Feature: Command behaviour when attached to an UA subscription
     @uses.config.machine_type.lxd.container
     Scenario Outline: Run timer script on an attached machine
         Given a `<release>` machine with ubuntu-advantage-tools installed
-        When I attach `contract_token` with sudo
+        When I run `systemctl stop ua-timer.timer` with sudo
+        And I attach `contract_token` with sudo
         Then I verify that running `ua config set update_messaging_timer=-2` `with sudo` exits `1`
         And stderr matches regexp:
         """
```

The other CI error is due to staging change that renamed esm-ros -> ros and added ros-updates. So not related to this PR

blackboxsw (approved):
Only failed CI jobs are unrelated to this PR
- esm-ros -> esm renaming, addressed by #1783 
- failure to shutdown vm `RuntimeError: Failure (rc=1): Error: shutting down the container failed`   intermittent LXC issue to investigate

PR #1762, opened 2021-08-09
test: update contract response to a dict

lucasmoura:
## Proposed Commit Message
test: update contract response to a dict

We now try to get the machine id info from the contract response first. On our test, the mocked response was a string. We are now updating it to be a dict to fix that issue

## Test Steps
Run the modified test

## Desired commit type::

## Checklist:


blackboxsw (approved):
LGTM. @lucasmoura thx for that.

PR #1763, opened 2021-08-10
fix: get machine ID from inside of machineTokenInfo (SC-319)

renanrodrigo:
CS is going to send us the machine ID inside of `machineTokenInfo`, and we should try to get it from there.

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -221,14 +221,14 @@ class UAContractClient(serviceclient.UAServiceClient):
         if not detach:
             self.cfg.write_cache("machine-token", response)
             util.get_machine_id.cache_clear()
-            self.cfg.write_cache(
-                "machine-id",
-                response.get("machineId", data.get("machineId", "")),
+            machine_id = response.get("machinetTokenInfo", {}).get(
+                "machineId", data.get("machineId")
             )
+            self.cfg.write_cache("machine-id", machine_id)
         return response
 
     def _get_platform_data(self, machine_id):
-        """"Return a dict of platform-relateddata for contract requests"""
+        """Return a dict of platform-related data for contract requests"""
         if not machine_id:
             machine_id = util.get_machine_id(self.cfg)
         platform = util.get_platform_info()
--- a/uaclient/tests/test_util.py
+++ b/uaclient/tests/test_util.py
@@ -495,11 +495,9 @@ class TestApplySeriesOverrides:
 
 class TestGetMachineId:
     def test_get_machine_id_from_config(self, FakeConfig):
-        cfg = FakeConfig.for_attached_machine(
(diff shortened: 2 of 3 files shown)


uaclient/util.py
```diff
@@ -299,8 +299,10 @@ def get_machine_id(cfg) -> str:
     # Docker images do not define ETC_MACHINE_ID or DBUS_MACHINE_ID on trusty
     # per Issue: #489
 
-    if cfg.machine_token:
-        cfg_machine_id = cfg.machine_token.get("machineId")
+    if isinstance(cfg.machine_token, dict):
```
lucasmoura: Why we need the `isinstance` check here ?

renanrodrigo: we don't: there was a failing test, but its related to some mock behavior. Already removed.

uaclient/contract.py
```diff
@@ -221,14 +221,14 @@ def _request_machine_token_update(
...
-                response.get("machineId", data.get("machineId", "")),
+            machine_id = response.get("machinetTokenInfo", {}).get(
+                "machineId", data.get("machineId")
             )
+            self.cfg.write_cache("machine-id", machine_id)
         return response
 
     def _get_platform_data(self, machine_id):
-        """"Return a dict of platform-relateddata for contract requests"""
+        """Return a dict of platform-relateddata for contract requests"""
```
lucasmoura, line 231: I think we can add an extra space here between related and data

uaclient/tests/test_util.py
```diff
@@ -575,9 +573,10 @@ def test_fallback_used_if_all_other_files_are_empty(
         self, FakeConfig, tmpdir, empty_value
     ):
         data_machine_id = tmpdir.join("machine-id")
-        cfg = FakeConfig.for_attached_machine()
-        # Cache it in cfg, so the load_file mock does not mess everything up
-        cfg.machine_token
+        cfg = FakeConfig()
```
lucasmoura, line 576: Can't we just do something like this here:
```python
FakeConfig().for_attached_machine(
    machine_token={"test": "blah"}
)
```

uaclient/contract.py
renanrodrigo: Oh black got this one. Will fix

uaclient/tests/test_util.py
renanrodrigo: :man_facepalming:

renanrodrigo: had to initialize the property as well because of the mock, but it's clearer for sure

lucasmoura (approved):

renanrodrigo:
LXD integration tests dying randomly and unrelated

lucasmoura:
Couldn't reproduce the failing tests locally

PR #1764, opened 2021-08-11
Allow timed jobs' intervals to be overriden by config (SC-308)

renanrodrigo:
User can now set the desired run_interval for jobs in the `uaclient.conf` file, as in:
```
ua_config:
    <job_name>_timer: <run_interval_value>
```
It has the config for the messaging and status jobs.

## Desired commit type::

## Checklist:

### DIFF
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -34,12 +34,16 @@ class TimedJob:
 
     def run(self, cfg: UAConfig):
         """Run a job in a failsafe manner, returning True on success.
+        Checks if the job is not disabled before running it.
 
         :param cfg: UAConfig instance
 
         :return: A bool True when successfully run. False if ignored
             or in error.
         """
+        if not self._should_run(cfg):
+            return False
+
         LOG.debug("Running job: %s", self.name)
         try:
             self._job_func(cfg)
@@ -60,6 +64,10 @@ class TimedJob:
             LOG.error(error_msg)
             return self._default_interval_seconds
         return configured_interval
+
+    def _should_run(self, cfg):
+        """Verify if the job has a valid (non-zero) interval."""
+        return self.run_interval_seconds(cfg) != 0
 
 
 UACLIENT_JOBS = [
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -1,13 +1,12 @@
 """
-Timer script used to run all jobs that need to be frequently run on the system
+Timer used to run all jobs that need to be frequently run on the system
 """
 
 import logging
 
 from datetime import datetime, timedelta
 
-import collections
-from typing import Any, Callable, Dict  # noqa: F401
+from typing import Callable
 
 from uaclient.cli import setup_logging
 from uaclient.config import UAConfig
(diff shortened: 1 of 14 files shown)


lib/timer.py
```diff
@@ -16,68 +15,72 @@
...
+        return True
+
+    def run_interval_seconds(self, cfg: UAConfig):
+        """Return the run_interval for the job based on config or defaults"""
+        return cfg.timer.get(self.name, {}).get(
+            "run_interval_seconds", self._default_interval_seconds
+        )
+
+
+class Timer:
```
lucasmoura, line 50: Why do we need a Timer class ?

uaclient/config.py
```diff
@@ -346,6 +347,20 @@ def features(self):
                 )
         return {}
 
+    @property
+    def timer(self):
```
lucasmoura, line 351: I think the idea here is to allow user to set the jobs time configurations through `ua config set`
We will probably need to move that logic under the `ua_config` key

lib/timer.py
renanrodrigo: We do not - I thought it would be a more organized and readable way to encapsulate the jobs and the running functionality. and since Jobs were turned into classes, I created the Timer one as well.

uaclient/config.py
renanrodrigo: Sure, can take a look into that.
Are you thinking about something like `ua cofig set timer <job name> run_interval_seconds <value>`?

renanrodrigo: Speaking about that, if we do not want to configure anything but the intervals, is there any major downside in going
```
timer:
    <job_name>: <interval_value>
```
for config?

lib/timer.py
```diff
@@ -1,13 +1,12 @@
...
+Timer used to run all jobs that need to be frequently run on the system
 """
 
 import logging
 
 from datetime import datetime, timedelta
 
-import collections
-from typing import Any, Callable, Dict  # noqa: F401
+from typing import Callable  # noqa: F401
```
blackboxsw, line 9: do we still need the noqa comment?

uaclient/config.py
blackboxsw: @lucasmoura @renanrodrigo I vote to avoid adding a `ua config set <timer> <runinterval>` for this release. I'd expect that custom setting timer run frequency to something slower than what our defaults are is not a highly used config operation that we expect lots of users to "want". But we can discuss this with @mitechie as to whether this use case is one that begs the CLI "sugar" for ease of set/unset.

lib/timer.py
blackboxsw: I like have you've ecapsulated the TimedJob.run definition, but I'm a little uncertain if this is overengineering the "Timer" class at the moment since we are only really talking about an ordered dict and a single function to run_jobs. I also have a general aversion to __init__ methods in classes which have a lot of side-effects (like initializing UAConfig and multiple TimedJob instances. Are there other compelling reasons to make this a class?

blackboxsw (changes requested):
@renanrodrigo general approach looks good. +1 on the TimedJob instrumentation, have a couple questions about the real motivation at this PR for a Timer class, but the only significant thing I really want to avoid at the moment in this PR is CLI changes to `ua config`.

Per your questions:
Opening it as a draft, for I would wait until #1761 is merged, and then I can add the metering job here as well - imho it will be easier for me to rebase this than merging this and then rebasing the metering PR.
   - yeah you can rebase off of @lucasmoura's branch as a base. I've approved that so I imagine he can land it early in the day when CI is sorted.
   - Feel free to open a doc PR separately so we can add a doc for the full timer and both existing jobs etc. will be easier to review docs sepearately
   - Should I write any integration tests for this?   Please do extend Lucas' existing integraiton test "Run timer script on an attached machine" you might be able to just set/append to uaclient.conf and allow the timer to trigger and check the next_run value in /var/lib/ubuntu-advantage/jobs-status.json to ensure override was applied.

lib/timer.py
renanrodrigo: No other compelling. Will change it back to a function.

lib/timer.py
```diff
@@ -16,68 +15,72 @@
...
+        return cfg.timer.get(self.name, {}).get(
+            "run_interval_seconds", self._default_interval_seconds
+        )
+
+
+class Timer:
+    def __init__(self, current_time: datetime, cfg: UAConfig = None):
+        self.current_time = current_time
+        self.cfg = cfg if cfg is not None else UAConfig()
+        self.jobs = [
```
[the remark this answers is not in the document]
renanrodrigo: no class anymore so no reason for this

lib/timer.py
renanrodrigo: nope (:

uaclient/config.py
renanrodrigo: Maybe we could do this iteration using only the config file, and implement the commands in another PR afterwards

blackboxsw:
Per discussion at standup today, we **want** to support `ua config set timer ...`  whether you want to do that in this branch or a separate branch let's iron that out in the spec file. I'll add a section now for you and Lucas to discuss tomorrow.

renanrodrigo:
I can do it in this one, no problem.

uaclient.conf
```diff
@@ -12,3 +12,5 @@ ua_config:
   apt_https_proxy: null
   http_proxy: null
   https_proxy: null
+  update_messaging_timer: null
```
lucasmoura, line 15: One idea that @mitechie proposed is to ship the config with the default values already in place.
If we do that, we can remove the constants we are using on the timer module

lib/timer.py
```diff
@@ -22,69 +21,81 @@
...
+            self._job_func(cfg)
+        except Exception as e:
+            LOG.warning("Error executing job %s: %s", self.name, str(e))
+            return False
+
+        return True
+
+    def run_interval_seconds(self, cfg: UAConfig):
+        """Return the run_interval for the job based on config or defaults."""
+        configured_interval = getattr(cfg, "{}_timer".format(self.name), None)
```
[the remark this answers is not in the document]
renanrodrigo: No. As the configuration has all the job timers set to null by default, `getattr` will return `None` and the default value will not be returned.

uaclient/tests/test_ua_timer.py
```diff
@@ -2,100 +2,130 @@
...
                 "jobs-status",
-                {
-                    "update_messaging": {
-                        "next_run": now - datetime.timedelta(seconds=1)
-                    },
-                    "update_status": {
-                        "next_run": now - datetime.timedelta(seconds=1)
-                    },
-                },
+                {"day_job": {"next_run": now - datetime.timedelta(seconds=1)}},
```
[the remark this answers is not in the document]
renanrodrigo: 1. As this is a unit test, a generic entry here  tests only the functionality of this piece of code, without depending on the real environment around it
2. Follows from 1: when any new job is added, there is no side effect, and no need to change the test because it is one more item to iterate over the same logic

uaclient.conf
renanrodrigo: Ack, but then, what happens if the user sets the value to `null`?
We would need to change some hereandtheres to make sure it will not surface the None value, as we would not have the constants.

So, imho, it is ok to put the defaults here, but keeping them in the file is safer.

renanrodrigo: As discussed with @mitechie, we can define the same default values in the config file, but keep those in the constants as a fallback value for any invalid or missing value in the config file.

lib/timer.py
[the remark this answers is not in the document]
renanrodrigo: Config will not be set to null in the end, but to some integer value. I will change this to ignore wrong attribute values and log occurrences, using the default when it happens,

uaclient/config.py
```diff
@@ -768,6 +792,13 @@ def process_config(self):
             "https", self.https_proxy, util.PROXY_VALIDATION_SNAP_HTTPS_URL
         )
 
+        for prop in ("update_messaging_timer", "update_status_timer"):
```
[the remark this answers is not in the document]
renanrodrigo: Will not fail here anymore, just log an error if the validation goes wrong

uaclient/tests/test_ua_timer.py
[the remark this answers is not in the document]
lucasmoura: I get the idea, but I do like the idea that for every job we run, we will need to modify that test. It should at least cover that we are properly configuring the time for each job correctly. But let me know if you think this is an overkill, as I don't have a strong opinion on that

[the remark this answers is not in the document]
renanrodrigo: I see. I think that to test the jobs themselves, with the real values, a behave test would make more sense  - can try and orchestrate a way of checking if next_run is being calculated appropriately.

lib/timer.py
```diff
@@ -22,69 +21,88 @@
...
+        if not isinstance(configured_interval, int) or configured_interval < 0:
+            error_msg = (
+                "Invalid value for the {} interval Found in config. "
```
lucasmoura, line 63: I think we can drop this `the` and replace `Found` for `found`

lucasmoura: Also, can we test that scenario on the behave test as well ? See if we are logging this part if an invalid config is provided

uaclient/cli.py
```diff
@@ -677,6 +677,16 @@ def action_config_set(args, cfg, **kwargs):
         }
         kwargs[set_key[4:]] = set_value
         setup_apt_proxy(**kwargs)
+    elif set_key in ("update_messaging_timer", "update_status_timer"):
+        try:
+            set_value = int(set_value)
+            if set_value < 0:
+                raise ValueError()
```
lucasmoura, line 684: I think we should have a message here saying why we raised the value error.
Furthermore, maybe it is a good to idea to also have this validation on the behave test

uaclient/tests/test_ua_timer.py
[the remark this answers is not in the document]
lucasmoura: I think this be harder to check on the behave tests, but if you think there is an easy way to do that, go for it

uaclient/config.py
```diff
@@ -755,6 +791,19 @@ def help(self, name):
...
+            "update_status_timer",
+            "gcp_auto_attach_timer",
+        ):
+            value = getattr(self, prop)
+            if not isinstance(value, int) or value < 0:
+                error_msg = (
+                    "Value for the {} interval must be an integer. "
+                    "Default value will be used."
+                ).format(prop)
+                logging.error(error_msg)
```
lucasmoura, line 805: I think we can abort the method if we detect this, similar to what we do with the proxy config

uaclient/tests/test_ua_timer.py
```diff
@@ -2,117 +2,122 @@
...
+    )
+    @pytest.mark.parametrize("caplog_text", [logging.DEBUG], indirect=True)
+    def test_does_not_run_if_disabled(
+        self, FakeConfig, caplog_text, disabled_by_default, interval_value
+    ):
+        cfg = FakeConfig()
+        if not disabled_by_default:
+            setattr(cfg, "day_job_timer", 0)
+
+        def disabled_job(config):
```
lucasmoura, line 67: I think we can turn this into a mock and assert that the mock was never called, just to be sure that the reason the call failed is because of the timer and not the function itself

uaclient/config.py
renanrodrigo: I remember @mitechie said to avoid failing when dealing with the configuration files. CLI fails and does not set the value, but processing the config should not fail.

lucasmoura: But this is the user manually running `ua refresh config`. In that situation, I think we should fail early to alert the user about the mistake.

uaclient/tests/test_config.py
```diff
@@ -1402,6 +1408,20 @@ def test_process_config(
...
+                "ua_config": {
+                    "update_messaging_timer": "wrong",
+                    "update_status_timer": 43200,
+                    "gcp_auto_attach_timer": 1800,
+                }
+            }
+        )
+
+        with pytest.raises(exceptions.UserFacingError):
+            cfg.process_config()
```
lucasmoura, line 1423: Can we check the exception message here to make sure we failing due to the expected exception ?

lucasmoura (approved):
LGTM, let's just wait for CI before landing it

blackboxsw (approved):
LGTM. Let's make sure we circle back to the systemd timer being run every < 30 mins to support the gcp auto-attach frequency in a separate PR.

PR #1765, opened 2021-08-12
tests: allow snapshot images to be ephemeral

lucasmoura:
## Proposed Commit Message
tests: allow snapshot images to be ephemeral

We have recently allowed pycloudlib to make snapshot from ephemeral instances. We are nos updating uaclient
to also use ephemeral instances for the base images that will be snapshoted when running the tests

## Test Steps
Run any integration test with UACLIENT_BEHAVE_SNAPSHOT_STRATEGY=1 and verify that the base images are ephemeral

## Desired commit type::

## Checklist:


paride:
This LGTM, let's see if CI agrees.

PR #1766, opened 2021-08-13
jobs: add status update job (SC-329)

lucasmoura:
## Proposed Commit Message
jobs: add status update job

Add job to update status cache on a daily basis

## Test Steps
Run the modified unittest

## Desired commit type::

## Checklist:

### DIFF
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -12,20 +12,33 @@ from typing import Any, Callable, Dict  # noqa: F401
 from uaclient.cli import setup_logging
 from uaclient.config import UAConfig
 from uaclient.jobs.update_messaging import update_apt_and_motd_messages
+from uaclient.jobs.update_state import update_status
 
 
 LOG = logging.getLogger(__name__)
+UPDATE_MESSAGING_INTERVAL = 21600  # 6 hours
+UPDATE_STATUS_INTERVAL = 43200  # 12 hours
 
 # Store the default run_interval_seconds for each job in this dict.
 # OrderedDict is used to ensure each job is run sequentially in the order
 # listed as some jobs may depend on completion of the previous job.
 UACLIENT_JOBS = collections.OrderedDict(
-    {
-        "update_messaging": {
-            "job_func": update_apt_and_motd_messages,
-            "run_interval_seconds": 43200,  # 12 hours
-        }
-    }
+    [
+        (
+            "update_messaging",
+            {
+                "job_func": update_apt_and_motd_messages,
+                "run_interval_seconds": UPDATE_MESSAGING_INTERVAL,
+            },
+        ),
+        (
+            "update_status",
+            {
+                "job_func": update_status,
+                "run_interval_seconds": UPDATE_STATUS_INTERVAL,
+            },
+        ),
+    ]
 )  # type: Dict[str, Dict[str, Any]]
 
 
--- a/uaclient/tests/test_ua_timer.py
+++ b/uaclient/tests/test_ua_timer.py
@@ -3,8 +3,14 @@ import logging
 import mock
 import pytest
 
-from lib.timer import run_job, run_jobs
+from lib.timer import (
+    run_job,
+    run_jobs,
+    UPDATE_MESSAGING_INTERVAL,
+    UPDATE_STATUS_INTERVAL,
+)
 from uaclient.jobs.update_messaging import update_apt_and_motd_messages
+from uaclient.jobs.update_state import update_status
 
 
 class TestRunJob:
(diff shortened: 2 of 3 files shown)


lib/timer.py
```diff
@@ -20,12 +21,22 @@
...
+            {
+                "job_func": update_apt_and_motd_messages,
+                "run_interval_seconds": 43200,  # 12 hours
+            },
+        ),
+        (
+            "update_status",
+            {
+                "job_func": update_status,
+                "run_interval_seconds": 43200,  # 12 hours
```
blackboxsw, line 36: Let's make this only daily. This could potentially mean lots of repeated traffic to the contract server, and 1. customers won't like a lot of pings egressing and 2. contract server might not want a lot of repeated loads.
```suggestion
                "run_interval_seconds": 60*60*24,  # 24 hours
```

blackboxsw (changes requested):
let's increase the interval to 24 hours to avoid frequent hits to contract server for no changes (99.9% of the time). 
We also probably want to minimally instrument a `systemctl status ua-timer.service` to ensure no errors  or even attempt a `systemctl stop ua-timer.service; systemctl start ua-timer,service; `  and then Grep for the order of jobs run in logs? As it is that could wait to a separate PR.

blackboxsw:
> let's increase the interval to 24 hours to avoid frequent hits to contract server for no changes (99.9% of the time).
> We also probably want to minimally instrument a `systemctl status ua-timer.service` to ensure no errors or even attempt a `systemctl stop ua-timer.service; systemctl start ua-timer,service; ` and then Grep for the order of jobs run in logs? As it is that could wait to a separate PR.

Also we could just check that /var/lib/ubuntu-advantage/jobs-state.json contains all job names (proving success)

uaclient/tests/test_ua_timer.py
```diff
@@ -64,13 +68,21 @@ def test_run_jobs_persists_job_status_on_successful_run(
...
                 "next_run": next_run.strftime("%Y-%m-%dT%H:%M:%S.%f"),
-            }
+            },
+            "update_status": {
+                "last_run": now.strftime("%Y-%m-%dT%H:%M:%S.%f"),
+                "next_run": next_run.strftime("%Y-%m-%dT%H:%M:%S.%f"),
+            },
         }
         assert jobs_status == cfg.read_cache("jobs-status")
+        print(run_job.call_args_list)
```
renanrodrigo, line 78: just a runaway print here

renanrodrigo (commented):
Overall LGTM - just Chad's comment about the frequency

renanrodrigo (approved):
lgtm (:

lib/timer.py
```diff
@@ -20,12 +21,22 @@
...
-            "job_func": update_apt_and_motd_messages,
-            "run_interval_seconds": 43200,  # 12 hours
-        }
-    }
+    [
+        (
+            "update_messaging",
+            {
+                "job_func": update_apt_and_motd_messages,
+                "run_interval_seconds": 43200,  # 12 hours
```
mitechie, line 29: can we cut these jobs in half and do 6hrs for apt/motd and 12 for status please?

uaclient/tests/test_ua_timer.py
```diff
@@ -55,22 +56,37 @@ def test_run_jobs_persists_job_status_on_successful_run(
...
-                    }
+                    },
+                    "update_status": {
+                        "next_run": now - datetime.timedelta(seconds=1)
+                    },
                 },
             )
         run_jobs(cfg=cfg, current_time=now)
-        next_run = now + datetime.timedelta(seconds=43200)
+        update_messaging_next_run = now + datetime.timedelta(seconds=43200)
```
mitechie, line 66: This time value should be from a const or config right so that it matches up with the UACLIENT_JOBS definition?

uaclient/tests/test_ua_timer.py
```diff
@@ -55,22 +56,37 @@ def test_run_jobs_persists_job_status_on_successful_run(
...
+        update_status_next_run = now + datetime.timedelta(seconds=86400)
```
mitechie, line 67: same here, we shouldn't have repeat values

mitechie (changes requested):

lib/timer.py
lucasmoura: The initial idea was for the status job to run every 12h, but @blackboxsw suggested that we increased it to 24h, since the there is not many gap situation for the root vs non-root job.

I think we have the same logic for the `update_messaging` job where the job will only potentially produce new messages if an apt update operation occurred and changed the system state.

But if you think we should make the change, np

mitechie: Yea, I'd like to take advatange of these jobs to keep things fresher. It should be pretty light and we're not running that much.

lib/timer.py
```diff
@@ -12,20 +12,33 @@
 from uaclient.cli import setup_logging
 from uaclient.config import UAConfig
 from uaclient.jobs.update_messaging import update_apt_and_motd_messages
+from uaclient.jobs.update_state import update_status
 
 
 LOG = logging.getLogger(__name__)
+TIME_FREQUENCY_6_HOURS = 21600
+TIME_FREQUENCY_12_HOURS = TIME_FREQUENCY_6_HOURS * 2
```
mitechie, line 20: these are going to go away then and be replaced by reading config values with a default set in case they're not set then?

lucasmoura: Yes, those values are the default ones. If the user intends to add a custom configuration on `uaclient.conf` they will override those values

mitechie: Right, but then why wouldn't it be job focused? Right now if you were to change the time you need a whole new variable name. 

I'd suggest just  

    UPDATE_MESSAGING_INTERVAL = xxx
    UPDATE_STATUS_INTERVAL = yyy

and then you can use those in the code and if they ever change it's all good who cares. 

Once the config is there it could be something like

    config.get('update_messaging_interval', UPDATE_MESSAGING_INTERVAL) 

and you're off to the races.

lucasmoura: Fair enough, that's a good point. I will update the code

mitechie (approved):

PR #1767, opened 2021-08-13
azure: support Azure published images bump IMDS API version

blackboxsw:
# WIP: Do not land yet. Awaiting feedback from @patviafore-canonical on whether this is a viable approach. 
## Proposed Commit Message
```
    azure: support Azure published images bump IMDS API version
    
    Update the API_VERSION to 2020-09-01 to support grabbing both
    publisher and SKU information which helps discern whether
    an image is published by Microsoft itself versus one published
    by Canonical.
    
    Azure PRO based images will require an update to the IMDS APIcw
     version so the contract server can see whether the image is
    Canonical PRO or a Microsoft derivative image based on PRO.
```

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/clouds/azure.py
+++ b/uaclient/clouds/azure.py
@@ -14,7 +14,7 @@ from uaclient import util
 
 IMDS_BASE_URL = "http://169.254.169.254/metadata/"
 
-API_VERSION = "2019-06-04"  # Needed to get subscription ID in attested data
+API_VERSION = "2020-09-01"  # Needed to get subscription ID in attested data
 IMDS_URLS = {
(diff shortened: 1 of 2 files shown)


patviafore-canonical:
I think we'll need something else on this, but not sure yet. With my test case, I'm getting the error message "Auto-attach image support is not available on this image".  I'll coordinate with Chad to get the right fix

patviafore-canonical:
Looked at it a little more - we now think that this change contains everything we need - the Auto-attach error was coming from the contracts server.

+1 from me.

mitechie (approved):

PR #1768, opened 2021-08-16
jobs: add gcp auto-attach job (SC-333)

lucasmoura:
## Proposed Commit Message
jobs: add gcp auto-attach job

Add gcp job that tries to run an auto-attach operation every 30 minutes. Before we try to run that command we verify that we are running on a GCP cloud and that we are not already attached. If we detect any of those scenarios, the job will not execute

## Test Steps
Launch a GCP PRO instance, then:

1) Run a detach command
2) Stop the systemd timer: `systemctl stop ua-timer.service`
3) Enable the systemd timer: `systemctl start ua-timer.service`
4) Wait a couple of minutes and verify that the instance is now attached to an UA subscription
5) Verify that the jobs status json file was created

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/jobs/gcp_auto_attach.py
+++ b/uaclient/jobs/gcp_auto_attach.py
@@ -1,0 +1,36 @@
+"""
+Try to auto-attach in a GCP instance. This should only work
+if the instance has a new UA license attached to it
+"""
+
+from uaclient import config
+from uaclient import exceptions
+
+from uaclient.clouds.identity import get_cloud_type
+from uaclient.cli import action_auto_attach
+
+
+def gcp_auto_attach(cfg: config.UAConfig) -> None:
+    # If the instance is already attached we will not do anything.
+    # This implies that the user may have a new license attached to the
+    # instance, but we will not perfom the change through this job.
+    if cfg.is_attached:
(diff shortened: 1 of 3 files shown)


uaclient/jobs/gcp_auto_attach.py
```diff
@@ -0,0 +1,36 @@
...
+    # We will not do anything in a non-GCP cloud
+    cloud_id, _ = get_cloud_type()
+    if not cloud_id or cloud_id != "gce":
+        return
+
+    try:
+        # This function already uses the assert lock decorator,
+        # which means that we don't need to make create another
+        # lock only for the job
+        action_auto_attach(args=None, cfg=cfg)
```
lucasmoura, line 29: I don't know if we should reuse here the direct cli command for auto-attach, but I think this is a more straightforward solution

renanrodrigo: I see no problem in that

renanrodrigo (approved):
Overall LGTM.
If landing on top of the config PR, test changes are not needed and there will be some other things.
If landing before that, looks good.

PR #1769, opened 2021-08-17
Add ros service

lucasmoura:
## Proposed Commit Message
ROS service

Add `esm-ros` service as a beta service.

The `esm-ros` service is dependent on both `esm-apps`  and `esm-infra`. When enabling it, we check if both of these services
are enabled and if not, we prompt the user to enable them.

When disabling a service that `esm-ros` depends on, we also prompt the user alerting that we must disable `esm-ros`  as well.

## Test Steps
Run the new unittests and integration tests added for this feature

## Desired commit type::

## Checklist:


uaclient/entitlements/ros.py
```diff
@@ -0,0 +1,11 @@
+from uaclient.entitlements import repo
+
+
+class ROSESMEntitlement(repo.RepoEntitlement):
+    help_doc_url = ""
```
blackboxsw, line 5: @mitechie should this doc URL link  be https://ubuntu.com/robotics/ros-esm and should we work with @lechsandecki to make sure design folks are on board with making this the documentation landing point for ROS sort of like we have https://ubuntu.com/security/esm which onboards to setting up Ubuntu Advantage?

uaclient/entitlements/ros.py
```diff
@@ -0,0 +1,11 @@
...
+    name = "esm-ros"
+    title = "ROS ESM"
+    description = "ROS ESM Service"
```
blackboxsw, line 8: @lechsandecki @mitechie do we want this discription to be a bit more "descriptive"?
We have about 46 characters to play with to fit into a 80 char wide console without wrapping on a `ua status` call. Wrapping is an option too as we already let fips-updates overflow already
```bash
root@dev-i3:~# ua status
cis           yes       n/a       Center for Internet Security Audit Tools
esm-infra     yes       n/a       UA Infra: Extended Security Maintenance (ESM)
fips          yes       n/a       NIST-certified core packages
fips-updates  yes       n/a       NIST-certified core packages with priority sec
urity updates

File: features/staging_commands.feature [hunk 3]
```diff
@@ -90,3 +90,78 @@ Feature: Enable command behaviour when attached to an UA staging subscription
            | bionic  | bundler  |
            | focal   | ant      |
            | xenial  | jq       |
+
+    @series.lts
+    Scenario Outline: Attached enable esm-apps on a machine
```
blackboxsw, line 95:
```suggestion
    Scenario Outline: Attached enable esm-ros on a machine
```

uaclient/entitlements/base.py
```diff
@@ -60,6 +62,12 @@ class UAEntitlement(metaclass=abc.ABCMeta):
     # List of services that are incompatible with this service
     _incompatible_services = []  # type: "List[str]"
 
+    # List of services that must be active before enabling this service
+    _required_services = []  # type: "List[str]"
```
blackboxsw, line 66: It's probably safer for us to use immutables like a tuple for these class attributes because if we grow methods that manipulate these class attrs in this future in subclasses, isn't it possible that the mutable list would be overridden for all subclasses?

uaclient/entitlements/base.py
```diff
@@ -273,8 +310,36 @@ def can_enable(self) -> "Tuple[bool, Optional[CanEnableFailure]]":
...
+
+        :return:
+            True if all required services are active
+            False is at least one of the required services is disabled
+        """
+        from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
+
+        for required_service in self.required_services:
+            ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(required_service)
+            if ent_cls:
```
blackboxsw, line 336: The logic here is a silent failure if we've typo'd required_services in the static class definition. It it worth us not being defensive here so we raise a noisy error or maybe if we log a warning if required_service not in ENTITLEMENT_CLASS_BY_NAME? Same question for dependent_services logic.

uaclient/entitlements/base.py
```diff
@@ -358,6 +423,51 @@ def handle_incompatible_services(self) -> bool:
...
+        that must be enabled first. In that situation, we can ask the user
+        if the required service should be enabled before proceeding.
+        """
+        from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
+
+        for required_service in self.required_services:
+            ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(required_service)
+
+            if ent_cls:
+                ent = ent_cls(self.cfg, allow_beta=True)
```
blackboxsw, line 440: We're doing a lot of instantiation of ENTITLEMENT_CLASS_BY_NAME across the check_required_services (called by can_enable) vs handle_required_services vs _disable_dependent_services. I'm wondering if it's worth us persisting the instantiated classes for reference somewhere to avoid multiple calls to ent_cls(self.cfg, allow_beta=True) etc. Or whether it's worth us reading this from status cache, so we don't have to re-inspect application_status multiple times on a single enable or disable operation. In either case, it does seem like there is a lot of boilerplate code that could be potentially shared between handle_required_services, _disable_dependent_services and handle_incompatible_services. Maybe it's just as simple as a get_service_status function that could leverage a cached _service_status somehow so that all callers could just inspect that app status value and react appropriately instead of  performing the same ent_cls().application_status setup in multiple places.

uaclient/entitlements/base.py
```diff
@@ -444,14 +554,103 @@ def applicability_status(self) -> "Tuple[ApplicabilityStatus, str]":
...
+
+        @return: True on success, False otherwise.
+
+        @param silent: Boolean set True to silence print/log of messages
+
+        @return: True on success, False otherwise.
+        """
+        pass
+
+    def _disable_dependent_services(self):
```
blackboxsw, line 571: The naming of handle_required_services vs _disable_dependent_services gives raises a question about naming. While I'd steer toward naming this handle_dependent_services, it then loses its descriptive name which clarifies what the method actually does. Should handle_required_services be renamed `enable_required_services` instead?

blackboxsw (commented):
Couple of discussion points to start. Thanks for this @lucasmoura

blackboxsw (changes requested):
I also think we need to fix integration tests and add more detailed help docs and urls with guidance from Rick and or Lech. 
Here's a diff of what I was thinking (not accounting for any of the refactor for required/dependent service logic)

```diff
diff --git a/features/attached_commands.feature b/features/attached_commands.feature
index 97d10b4b..9752e329 100644
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -124,6 +124,7 @@ Feature: Command behaviour when attached to an UA subscription
            cis           +<cis>      +Center for Internet Security Audit Tools
            esm-apps      +<esm-apps> +UA Apps: Extended Security Maintenance \(ESM\)
            esm-infra     +yes        +UA Infra: Extended Security Maintenance \(ESM\)
+           esm-ros       +yes        +ROS ESM Service
            fips          +<fips>     +NIST-certified core packages
            fips-updates  +<fips>     +NIST-certified core packages with priority security updates
            livepatch     +yes        +Canonical Livepatch service
diff --git a/help_data.yaml b/help_data.yaml
index da222a3a..38de33f3 100644
--- a/help_data.yaml
+++ b/help_data.yaml
@@ -30,7 +30,7 @@ esm-apps:
 
 esm-infra:
    help: |
-     esm-infra provides access to a private ppa which includes available high
+     esm-infra provides access to a private PPA which includes available high
      and critical CVE fixes for Ubuntu LTS packages in the Ubuntu Main
      repository between the end of the standard Ubuntu LTS security
      maintenance and its end of life. It is enabled by default with
@@ -38,6 +38,16 @@ esm-infra:
      You can find our more about the esm service at
      https://ubuntu.com/security/esm
 
+esm-ros:
+   help: |
+     esm-ros provides access to a private PPA which includes available high
+     and critical CVE fixes for Ubuntu Robotic Operating System (ROS) packages
+     between the end of the standard Ubuntu LTS security maintenance and its
+     end of life. For full access to ROS ESM and security fixes, both
+     esm-infra and esm-apps services will also be enabled.
+     You can find our more about the esm service at
+     https://ubuntu.com/robotics/ros-esm
+
 fips:
     help: |
       FIPS 140-2 is a set of publicly announced cryptographic standards
diff --git a/uaclient/entitlements/ros.py b/uaclient/entitlements/ros.py
index d8af73cf..0689d47d 100644
--- a/uaclient/entitlements/ros.py
+++ b/uaclient/entitlements/ros.py
@@ -2,10 +2,10 @@ from uaclient.entitlements import repo
 
 
 class ROSESMEntitlement(repo.RepoEntitlement):
-    help_doc_url = ""
+    help_doc_url = "https://ubuntu.com/robotics/ros-esm"
     name = "esm-ros"
     title = "ROS ESM"
-    description = "ROS ESM Service"
+    description = "UA ROS: Robotic Operating System ESM"
     repo_key_file = "ubuntu-advantage-esm-ros.gpg"
     is_beta = True
     _required_services = ["esm-infra", "esm-apps"]
```

features/staging_commands.feature
```diff
@@ -90,3 +90,78 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+        And stdout matches regexp
+        """
+        esm-infra     yes                enabled            UA Infra: Extended Security Maintenance \(ESM\)
+        """
+        When I run `apt install python3-catkin-pkg -y` with sudo
+        Then I verify that `python3-catkin-pkg` is installed from apt source `<ros-source>`
+
+        Examples: ubuntu release
+           | release | ros-source                                                    |
+           | xenial  | https://esm.staging.ubuntu.com/ros/ubuntu xenial-updates/main |
```
renanrodrigo, line 167: Do we have this only for Xenial?

uaclient/entitlements/base.py
lucasmoura: Since we are actually using a property to read those the values from these variables, I think this problem can still happen. However, I agree we should change them to a tuple regardless

renanrodrigo (changes requested):
Small picks, quick question

uaclient/entitlements/base.py
lucasmoura: Yes, good point. I think we should fail loudly here

uaclient/entitlements/base.py
```diff
@@ -358,6 +406,55 @@ def handle_incompatible_services(self) -> bool:
...
+                        print(e_msg)
+                        return False
+
+                    enable_msg = "Enabling required service: {}".format(
+                        ent.title
+                    )
+                    logging.info(enable_msg)
+
+                    ret = ent.enable()
+                    if not ret:
```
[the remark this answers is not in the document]
lucasmoura: Yes, you are right. Great catch

features/staging_commands.feature
lucasmoura: yes

uaclient/entitlements/base.py
lucasmoura: I thought about refactoring those methods, the problem is that we rely on the services through the method. We use to define the messages we are going to show to the user and we use it to perform the enable/disable operation. We could have an auxiliary function that return the entity, which will remove the import of the ENTITLEMENT_CLASS_BY_NAME dict from those methods, but I don't think it is the best refactoring we can do.

Regarding the use of status cache, we could leverage it. The problem is that we will always need to loop on the service list to find our service status, which is not ideal; Besides that, we will still need to instantiate the service anyway for the disable/enable operations, so I don't think that change will improve the code.

But I do agree that currently, those methods are sharing a lot of code and if you have any other refactoring idea or want to further discuss the proposed ones, let me know

blackboxsw: Yeah let's postpone this refactor idea until we can better spec out the implications and improvements to be had as caching would involve also potential cache-miss logic/errors etc.

uaclient/entitlements/ros.py
mitechie: I think the URL https://ubuntu.com/robotics/ros-esm is ok and works. We can see if PMs (and this falls more under Gabriel as the ROS PM) want to change it once they play with it in beta form.

uaclient/entitlements/ros.py
mitechie: I don't think it needs the UA bits to it. UA doesn't exist in the documentation at the above URL. I think "ROS Extended Security Maintenance (ESM)" works and fits with the existing text around esm-infra.

GAN-Canonical: Agree, ROS Extended Security Maintenance (ESM) should be enough. 
Once we make the announcement of UA Apps we might want to reconsider it though. (ROS ESM is part of UA A) 

In the service column, following the latest discussion on the UA channel, maybe we should have only "ros" and "ros-updates"?

mitechie:
+1 to ros and ros-updates

uaclient/entitlements/ros.py
blackboxsw: Thanks @GAN-Canonical, we'll pull this suggestion in once we have a contract update that distinguishes ros from ros-updates.

blackboxsw (approved):
Thanks for this @lucasmoura . Ran through various enable disable scenarios on xenial with ROS. This is good for --beta though we'll have to update description/product/help text once ros vs ros-updates are represented as separate products.

blackboxsw (approved):
LGTM! integration test faillure unrelated over in `ua fix`space I believe metadata is now being updated for this particular failing test by the security team. We may need to change how UA client decides to report and/or ignore certain packages.
https://pastebin.canonical.com/p/3scMsW7zD2/

PR #1770, opened 2021-08-17
cloud: rely only on cloud-init to determine cloud and instance IDs (SC-337)

renanrodrigo:
Since the Xenial release, cloud-init provides us `cloud-id` to determine on which cloud are we running, if any. We used to check for fallback files in the case cloud-id was not present. The same logic applied to instance ID.

The proposal here is that, for the future releases, this is not needed anymore.

## Desired commit type::

## Checklist:

LP: #1940131

### DIFF
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -1,6 +1,4 @@
-import json
 import logging
-import os
 
 from enum import Enum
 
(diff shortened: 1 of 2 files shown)


lucasmoura (approved):
I think this is good, but let's wait for a review from @blackboxsw to see if we are not missing anything here

blackboxsw (approved):
Thanks @renanrodrigo this works and thank you for clarifying in the commit message here on this so we can track details better in launchpad.

blackboxsw (approved):
Confirmed integration test failures are not present locally.
1 feature passed, 0 failed, 0 skipped
1 scenario passed, 0 failed, 39 skipped
25 steps passed, 0 failed, 247 skipped, 0 undefined
Took 4m11.394s
___________________________________ summary ____________________________________
  behave-lxd-16.04: commands succeeded
  congratulations :)

PR #1772, opened 2021-08-20
tests: make lxd.container machine type explicit (SC-310)

renanrodrigo:
Current behave tests can be decorated to run on top of specific platforms. If no platform is specified, lxd.container is selected by default. However, documentation points to lxd.container as a machine type and it is there for some tests, but it is the same as defining no machine_type.

Besides being ambiguous and not consistent, this behavior can be misleading: by looking at the patterns, one can think that tests without a machine_type decorator are running on all platforms, which is not true.

This PR is a refactor to explicitly tell which tests should be executed in containers.

## Desired commit type::

## Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -47,13 +47,13 @@ commands =
     mypy: mypy --python-version 3.6 uaclient/ features/ lib/
     mypy-focal: mypy --python-version 3.7 uaclient/ features/ lib/
     black: black --check --diff uaclient/ features/ lib/ setup.py
-    behave-lxd-16.04: behave -v {posargs} --tags="series.xenial,series.lts,series.all" --tags="~upgrade"
-    behave-lxd-18.04: behave -v {posargs} --tags="series.bionic,series.lts,series.all" --tags="~upgrade"
-    behave-lxd-20.04: behave -v {posargs} --tags="series.focal,series.lts,series.all" --tags="~upgrade"
-    behave-lxd-21.04: behave -v {posargs} --tags="series.hirsute,series.all" --tags="~upgrade"
-    behave-vm-16.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.xenial,series.all,series.lts"
-    behave-vm-18.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.bionic,series.all,series.lts"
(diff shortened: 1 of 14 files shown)


tox.ini
```diff
@@ -47,13 +47,13 @@ commands =
...
-    behave-vm-18.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.bionic,series.all,series.lts"
-    behave-vm-20.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.focal,series.all,series.lts"
+    behave-lxd-16.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.container" --tags="series.xenial,series.lts,series.all" --tags="~upgrade"
+    behave-lxd-18.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.container" --tags="series.bionic,series.lts,series.all" --tags="~upgrade"
+    behave-lxd-20.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.container" --tags="series.focal,series.lts,series.all" --tags="~upgrade"
+    behave-lxd-21.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.container" --tags="series.hirsute,series.all" --tags="~upgrade"
+    behave-vm-16.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.xenial,series.all,series.lts" --tags="~upgrade"
+    behave-vm-18.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.vm" --tags="series.bionic,series.all,series.lts" --tags="~upgrade"
```
lucasmoura, line 55: Why do we need to exclude the `upgrade` tags from the vm tests ?

renanrodrigo: We don't - They are excluded for everything else, and since we have a specific upgrade entry, I excluded them here as well. Is there a reason to keep those?

lucasmoura: Since all the upgrade test are running on containers at the moment, the exclusion is unnecessary. But you right, it will not hurt us to be consistent and it on the vm tests as well

renanrodrigo: FIPS upgrade tests are running on VMs actually

renanrodrigo:
@lucasmoura added a commit making sure timer.py always runs all jobs. This may fix our intermitent test. If it does not, I can debug it further in the future.

lucasmoura:
Just waiting on CI before landing it

Issue #1463, opened 2021-03-18
ua fix command will fail if we do not use a UTF-8 locale


lucasmoura:
In the end of the `ua fix` command we display a check mark indicating if the CVE/USN was fixed or not. This check mark is a unicode character. Therefore, if the user is running on a system that is not using a locale that support these unicode characters, `ua fix` will fail if the following exception in the logs:

```
UnicodeEncodeError: 'ascii' codec can't encode character '\u2714' in position 5: ordinal not in range(128)
```

We should update uaclient to not fail in this situation

PR #1776, opened 2021-08-25
fix: do not fail if non-utf8 locale is being used (SC-368)

lucasmoura:
## Proposed Commit Message
When running an ua fix command, we tell at the end if the issue is resolved or not. In that message, we have some unicode characters that are displayed. If the user is running on a system that doesn't support those characters, the printing that last message will fail. We are now redacting the code to not fail under this condition.

Fixes: #1463

## Test Steps
Run the new unittest provided in the system.
Another way to test is to launch an instance and change the locale to `ANSI_X3.4-1968`. Run an ua fix command and verify that it doesn't break on the message

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/tests/test_security.py
+++ b/uaclient/tests/test_security.py
@@ -941,13 +941,19 @@ class TestPromptForAffectedPackages:
         FakeConfig,
     ):
         with pytest.raises(exceptions.SecurityAPIMetadataError) as exc:
-            prompt_for_affected_packages(
-                cfg=FakeConfig(),
-                issue_id="USN-###",
-                affected_pkg_status=affected_pkg_status,
-                installed_packages=installed_packages,
-                usn_released_pkgs=usn_released_pkgs,
-            )
+            with mock.patch("uaclient.util.sys") as m_sys:
+                m_stdout = mock.MagicMock()
+                type(m_sys).stdout = m_stdout
+                type(m_stdout).encoding = mock.PropertyMock(
+                    return_value="utf-8"
+                )
+                prompt_for_affected_packages(
+                    cfg=FakeConfig(),
+                    issue_id="USN-###",
+                    affected_pkg_status=affected_pkg_status,
+                    installed_packages=installed_packages,
+                    usn_released_pkgs=usn_released_pkgs,
+                )
         assert (
             "Error: USN-### metadata defines no fixed version for sl.\n"
             "1 package is still affected: slsrc\n"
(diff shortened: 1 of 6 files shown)


uaclient/status.py
```diff
@@ -184,6 +184,8 @@ def __init__(
 )
 MESSAGE_SECURITY_ISSUE_RESOLVED = OKGREEN_CHECK + " {issue} is resolved."
 MESSAGE_SECURITY_ISSUE_NOT_RESOLVED = FAIL_X + " {issue} is not resolved."
+MESSAGE_SECURITY_ISSUE_RESOLVED_NON_UTF8 = "{issue} is resolved."
+MESSAGE_SECURITY_ISSUE_NOT_RESOLVED_NON_UTF8 = "{issue} is not resolved."
```
renanrodrigo, line 188: Now that we have the function, those are not needed anymore

renanrodrigo (changes requested):

lucasmoura:
@renanrodrigo code updated

blackboxsw (approved):
Two minor unrelated changes which I think can be dropped. Otherwise +1 and ship it.

renanrodrigo:
CI complains about some other unittests failing - worth a check

uaclient/tests/test_security.py
```diff
@@ -29,10 +29,10 @@
     MESSAGE_SECURITY_APT_NON_ROOT,
     MESSAGE_SECURITY_UA_SERVICE_NOT_ENABLED,
     MESSAGE_SECURITY_UA_SERVICE_NOT_ENTITLED,
-    MESSAGE_SECURITY_ISSUE_NOT_RESOLVED,
     MESSAGE_SECURITY_UPDATE_NOT_INSTALLED_SUBSCRIPTION as MSG_SUBSCRIPTION,
     MESSAGE_SECURITY_SERVICE_DISABLED,
     MESSAGE_SECURITY_UPDATE_NOT_INSTALLED_EXPIRED,
+    MESSAGE_SECURITY_ISSUE_NOT_RESOLVED,
```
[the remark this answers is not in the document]
lucasmoura: It was just because in the older implementation I had other status messages here. But I will remove that reorder

uaclient/tests/test_security.py
```diff
@@ -941,13 +941,19 @@ def test_raise_userfacing_error_on_invalid_usn_metadata(
...
-                issue_id="USN-###",
-                affected_pkg_status=affected_pkg_status,
-                installed_packages=installed_packages,
-                usn_released_pkgs=usn_released_pkgs,
-            )
+            with mock.patch("uaclient.util.sys") as m_sys:
+                m_stdout = mock.MagicMock()
+                type(m_sys).stdout = m_stdout
+                type(m_stdout).encoding = mock.PropertyMock(
+                    return_value="utf-8"
```
blackboxsw, line 948: Now I finally discern what mock @renanrodrigo was talking about. Didn't realize you were talking about this PR earlier. +1

blackboxsw (approved):
LGTM. Awaiting CI

blackboxsw:
CI failure w/ lxc launch on focal (unrelated to this PR)       RuntimeError: Failure (rc=1): Error: Failed getting disk path: Could not locate a zvol for default/virtual-machines/ubuntu-behave-test-1776-vm-focal-1630349395863675.block

renanrodrigo (approved):

blackboxsw (approved):
Integration test lockup on Xenial. Any integration tests  which would have changed would have affected Bionic and Focal test cases too. So, I'm +1 with this changeset as unittest coverage covers the differing behavior in odd cases when LANG is not yet set, or unset intentionally by an image.

PR #1778, opened 2021-08-26
tests: add script to run integration tests (SC-197)

renanrodrigo:
A CLI script to run the integration tests, organizing the credentials, storing the results (`stdin` and `stderr`).
If testing from Proposed, check that all tests pass and create a tarball with the results.
There is an example file with the yaml structure.

## Desired commit type::

## Checklist:

### DIFF
--- a/tools/run-integration-tests.py
+++ b/tools/run-integration-tests.py
@@ -1,0 +1,195 @@
+import os
+import subprocess
+import tarfile
+import time
+from datetime import datetime
+from sys import stdout
+
+import click
+import yaml
+
+PLATFORMS = (
+    "azuregeneric",
+    "azurepro",
+    "awsgeneric",
+    "awspro",
+    "gcpgeneric",
+    "gcppro",
+    "vm",
+    "lxd",
+    "upgrade",
+)
+
+SERIES_TO_VERSION = {
+    "xenial": "16.04",
+    "bionic": "18.04",
+    "focal": "20.04",
+    "hirsute": "21.04",
+}
+
+TOKEN_TO_ENVVAR = {
+    "prod": "UACLIENT_BEHAVE_CONTRACT_TOKEN",
+    "staging": "UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING",
+    "expired": "UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING_EXPIRED",
+}
+
+
+def is_compatible(platform, series):
+    if platform in ("lxd", "upgrade"):
+        return True
+    elif series not in ("hirsute",):
+        return True
+    return False
+
+
+def build_commands(
+    platform, series, proposed, check_version, token, credentials_path, wip
+):
+    with open(credentials_path) as f:
+        credentials = yaml.safe_load(f)
+    commands = []
+    for p in platform:
+        for s in series:
+            # Not all tests can run in all platforms
+            if is_compatible(p, s):
+                series_version = SERIES_TO_VERSION[s]
+                env = os.environ.copy()
+
+                # Add check_version variable
+                if check_version:
+                    env["UACLIENT_BEHAVE_CHECK_VERSION"] = "{}~{}.1".format(
+                        check_version, series_version
+                    )
+
+                # If not told to run from proposed, run from local changes
+                if proposed:
+                    env["UACLIENT_BEHAVE_ENABLE_PROPOSED"] = "1"
+                else:
+                    env["UACLIENT_BEHAVE_BUILD_PR"] = "1"
+
+                # Inject tokens from credentials
+                for t in token:
+                    envvar = TOKEN_TO_ENVVAR[t]
+                    env[envvar] = credentials["token"].get(envvar)
+
+                # Inject cloud-specific variables from credentials
+                for cloud in ("azure", "aws", "gcp"):
+                    if cloud in p:
+                        for envvar in credentials[cloud]:
+                            env[envvar] = credentials[cloud].get(envvar)
+
+                # Until we don't get sbuild to run just once per run,
+                env["UACLIENT_BEHAVE_SNAPSHOT_STRATEGY"] = "1"
+
+                # Tox command itself
+                command = "tox -e behave-{}-{}".format(
+                    p, series_version
+                ).split()
+
+                # Wip
+                if wip:
+                    command.extend("-- -w".split())
+
+                commands.append((command, env))
+
+    return commands
+
+
+@click.command()
+@click.option(
+    "-p",
+    "--platform",
+    type=click.Choice(PLATFORMS),
+    default=PLATFORMS,
+    multiple=True,
+    help="Platform to run the tests on (requires credentials file for the clouds)",
+)
+@click.option(
+    "-s",
+    "--series",
+    type=click.Choice(SERIES_TO_VERSION.keys()),
+    default=SERIES_TO_VERSION.keys(),
+    multiple=True,
+    help="Series to run the tests on",
+)
+@click.option(
+    "--token",
+    "-t",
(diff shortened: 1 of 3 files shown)


tools/run-integration-tests.py
```diff
@@ -0,0 +1,183 @@
...
+
+    for process in processes:
+        if process.wait():
+            failing.append(process.args)
+            error = True
+
+    if error:
+        print("There were errors in runs:")
+        for process_args in failing:
+            print("- {}".format(process_args))
```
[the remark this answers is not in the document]
renanrodrigo: Currently, it checks the jobs in sequence - will try to do this better.
/me on it

tools/run-integration-tests.py
```diff
@@ -0,0 +1,183 @@
...
+
+                # Add check_version variable
+                if check_version:
+                    env["UACLIENT_BEHAVE_CHECK_VERSION"] = check_version
+
+                # If not told to run from proposed, run from local changes
+                if proposed:
+                    env["UACLIENT_BEHAVE_ENABLE_PROPOSED"] = "1"
+                else:
+                    env["UACLIENT_BEHAVE_BUILD_PR"] = "1"
```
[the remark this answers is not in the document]
renanrodrigo: Checking on how to improve it, thank you for catching this (:

tools/run-integration-tests.py
```diff
@@ -0,0 +1,183 @@
...
+def run_tests(
+    platform, series, proposed, check_version, token, credentials_path, wip
+):
+    commands = build_commands(
+        platform, series, proposed, check_version, token, credentials_path, wip
+    )
+    output_dir = "test-results"
+    if os.path.exists(output_dir):
+        shutil.rmtree(output_dir)
+    os.makedirs(output_dir)
```
orndorffgrant, line 151: Drive-by thought: I'd rather not remove all previous test results. Can we instead create a new directory under `test-results` based on the datetime the tests are run maybe? Then we don't have to remove old test results and the new test results will get their own directory with a name that indicates the time they were run.

renanrodrigo:
@orndorffgrant points addressed, ready for another review

tools/run-integration-tests.py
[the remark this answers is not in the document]
renanrodrigo: As discussed, we will address this in a separate PR

tools/run-integration-tests.py
```diff
@@ -0,0 +1,182 @@
...
+    )
+    current_datetime = datetime.now().strftime(format="%b-%d-%H%M")
+    output_dir = "test-results-{}".format(current_datetime)
+    os.makedirs(output_dir, exist_ok=True)
+    processes = []
+    for command, env in commands:
+        print("Running {}".format(command))
+        with open(
+            "test-results/{}.txt".format(command[2]), "wb"
+        ) as result_file:
```
orndorffgrant, line 158: Thanks for adding this! I think it might be nice to have the datetime as a subdir under test-results so as not to pollute the root level of the project. Also we need to make sure we use the right dir later in `open`.
```suggestion
    current_datetime = datetime.now().strftime(format="%b-%d-%H%M")
    output_dir = "test-results/{}".format(current_datetime)
    os.makedirs(output_dir, exist_ok=True)
    processes = []
    for command, env in commands:
        print("Running {}".format(command))
        with open(
            "{}/{}.txt".format(output_dir, command[2]), "wb"
        ) as result_file:
```

And then later on, lets use the datetime again in the tarball name and for the tarball contents
```
        filename = "test-results-{}.tar.gz".format(current_datetime)
        with tarfile.open(filename, "w:gz") as results:
            results.add(output_dir, arcname="test-results/")
```

tools/run-integration-tests.py
```diff
@@ -0,0 +1,182 @@
...
+    while(processes):
+        for process in processes:
+            result = process.poll()
+            if result is not None:
+                if result == 0:
+                    print("{} finished sucessfully".format(process.args))
+                else:
+                    print("Failing tests for {}".format(process.args))
+                processes.remove(process)
+        time.sleep(5)
```
orndorffgrant, line 173: Love it! Polling every 5 seconds seems appropriate for this use case.

tools/run-integration-tests.py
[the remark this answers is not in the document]
orndorffgrant: All good - can we set `UACLIENT_BEHAVE_SNAPSHOT_STRATEGY=1` in this script though?

tools/run-integration-tests.py
renanrodrigo: Indeed

renanrodrigo:
@orndorffgrant thanks for the point-outs, all addressed (:

tools/run-integration-tests.py
```diff
@@ -0,0 +1,182 @@
...
+            processes.append(process)
+
+    while(processes):
+        for process in processes:
+            result = process.poll()
+            if result is not None:
+                if result == 0:
+                    print("{} finished sucessfully".format(process.args))
+                else:
+                    print("Failing tests for {}".format(process.args))
```
lucasmoura, line 171: Maybe we could rename the result file to mark it as failed

tools/run-integration-tests.py
```diff
@@ -0,0 +1,182 @@
...
+                processes.remove(process)
+        time.sleep(5)
+
+    if proposed:
```
lucasmoura, line 175: Should we do that only if all tests succeeded ?

renanrodrigo: I thought about this.
I had it this way in the start, but then I dropped it. Now that we don't delete test results automatically, I think yes, it makes sense.

tools/run-integration-tests.py
```diff
@@ -0,0 +1,182 @@
...
+    for p in platform:
+        for s in series:
+            # Not all tests can run in all platforms
+            if is_compatible(p, s):
+                series_version = SERIES_TO_VERSION[s]
+                env = os.environ.copy()
+
+                # Add check_version variable
+                if check_version:
+                    env["UACLIENT_BEHAVE_CHECK_VERSION"] = check_version
```
orndorffgrant, line 60: @renanrodrigo Okay I think this is my final request for this script :upside_down_face: 
The version is different across ubuntu releases. e.g. focal version is `27.2.2~20.04.1` and bionic version is `27.2.2~18.04.1` etc.
So if we just set one version at the command line then it won't work across all releases' tests.

It's probably good enough for this script to expect the user to just pass `27.2.2` and auto-append the `~$RELEASE.1`

renanrodrigo: Ooh you're right. Will it be always `~$RELEASE.1` though? Do we ever get `~$RELEASE.2` ?

orndorffgrant: To my knowledge, we rarely have needed to use `~$RELEASE.2` - Thats why I only said "probably good enough" :)
Alternatively, we could add another cli flag like "check-version-patch-num" that defaults to `1` that would get used there.

renanrodrigo: Looking at the history, I will leave this flag for later - if we needed, can be easily added in the future

renanrodrigo:
@lucasmoura @orndorffgrant points addressed, ready for another round

orndorffgrant:
This looks really good! Just running it again before approving

orndorffgrant (approved):
This is a great and helpful script - thank you for taking the initiative to write this!

I'm approving with just a note for posterity: Running all of the lxd/vm integration tests at once using this script overwhelmed my machine/lxd and caused invalid test failures. The tests pass when run in isolation (without the other tests running). I don't think it's worth changing the script because of this right now, but thought it was worth mentioning.

PR #1779, opened 2021-08-27
Remove trusty typing import logic and sort imports with isort (SC-367)

orndorffgrant:
## Commentary
While touching all of the imports in all of the files, it seemed like a good time to introduce isort, which will organize all of our imports for us (black doesn't do this). I like it because it is one less thing to think about, but I'm open to push-back on introducing more tooling.

Also, now that we can rely on typing being present, we no longer need to redefine custom types like StaticAffordance in several places. I added a new `types.py` where we can organize and define our custom types. We can start doing this more often now (where it makes sense) since it involves less overhead.

## Proposed Commit Message

```
types: remove typing import exception code needed for trusty
```
```
refactor: define StaticAffordance type once
```
```
tooling: sort imports with isort 
```
```
types: only use type forward refs where necessary 
```

## Test Steps

Ensure all tests still pass and no imports were mangled by me or the new isort tooling.

## Desired commit type::

## Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,3 +3,7 @@ repos:
     rev: 19.3b0  # Also stored in dev-requirements.txt; update both together!
     hooks:
     - id: black
+  - repo: https://github.com/pycqa/isort
+    rev: 5.8.0  # Also stored in dev-requirements.txt; update both together!
+    hooks:
+    - id: isort
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -53,7 +53,7 @@ pipeline {
             steps {
                 sh '''
                 set +x
-                tox -e black
+                tox -e black -e isort
                 '''
             }
         }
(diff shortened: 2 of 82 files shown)


renanrodrigo (dismissed):
Love for isort

I think now we are always importing the types, we can use them explicitly, without the comment syntax. So instead of having
```
instance_name: "Optional[str]" = None,
```
We would go 
```
instance_name: Optional[str] = None,
```

And then we don't need the `# noqa`s

orndorffgrant:
@renanrodrigo updated! We still needed the quotes in a couple places as forward references when the referenced type isn't defined yet.
I also looked into replacing variable type comments with [variable annotations](https://www.python.org/dev/peps/pep-0526/), but we can't drop support for 3.5 yet (xenial).
Hopefully jenkins likes the changes :)

blackboxsw (approved):
+1 good ideas here!

blackboxsw (approved):
Renan's comment was resolved here. I'll merge

PR #1780, opened 2021-08-27
Enable running many tests in parallel in clouds (SC-366)

renanrodrigo:
With the random suffix, there will be no conflicts even when running many tests exactly at the same time

## Test Steps
Trigger two or more integration test runs on the same cloud, at the same time. There should be no naming conflict.

## Desired commit type::

## Checklist:

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -2,7 +2,9 @@ import datetime
 import itertools
 import logging
 import os
+import random
 import re
+import string
 import tempfile
 import textwrap
 from typing import Any, Dict, List, Optional, Tuple, Union  # noqa: F401
@@ -268,6 +270,14 @@ class UAClientBehaveConfig:
             job_suffix = job_suffix.split("PR-")[-1]
         timed_job_tag += str(job_suffix)
         timed_job_tag = timed_job_tag.replace(".", "-")
+
+        # Add 8-digit random suffix to the tag, so it does not conflict when
+        # spinning many jobs at the same time
+        random_suffix = "".join(
+            random.choices(string.ascii_lowercase + string.digits, k=8)
+        )
+        timed_job_tag += "-" + random_suffix
+
         if "aws" in self.machine_type:
             self.cloud_manager = cloud.EC2(
                 aws_access_key_id,


features/environment.py
```diff
@@ -275,6 +277,14 @@ def __init__(
             job_suffix = job_suffix.split("PR-")[-1]
         timed_job_tag += str(job_suffix)
         timed_job_tag = timed_job_tag.replace(".", "-")
+
+        # Add 8-digit random suffix to the tag, so it does not conflict when
+        # spinning many jobs at the same time
+        random_suffix = "".join(
+            random.choices(string.ascii_lowercase + string.digits, k=8)
+        )
+        timed_job_tag += random_suffix
```
orndorffgrant, line 286: Currently the name ends up like this: `i0-uaclient-ci-0830-1427-dev8e1jf5s9`
It'd be nicer if there was a dash before the random suffix, like this: `i0-uaclient-ci-0830-1427-dev-8e1jf5s9`.
```suggestion
        timed_job_tag += "-" + random_suffix
```

orndorffgrant (changes requested):
Works great! I agree with the random identifier choice; shorter higher entropy alphanumeric is nicer to look at than uuids.

One tiny request and then I'll approve

features/environment.py
```diff
@@ -275,6 +277,14 @@ def __init__(
             job_suffix = job_suffix.split("PR-")[-1]
         timed_job_tag += str(job_suffix)
```
blackboxsw, line 278: I recall there was one cloud in the past that truncated tags or outright rejection when trying to use tags > some length. Google searching it looks like ec2 is 256 char, azure 512 char and gcp is 63 character limit. So, I think we have to be aware of those limits when defining our unique tags. I don't **think** this extra 8 characters will bump up against those limits. But, if it does we can trim/compress/redact where necessary

renanrodrigo: With the added 9 characters, we have 56 - so we should be good even for the lower gcp limit.

features/environment.py
renanrodrigo: Done! Looks better now.

blackboxsw (approved):

orndorffgrant (approved):
Thanks! Let's merge as soon as Jenkins approves

renanrodrigo:
Will address conflicts and hope for good Jenkins instead of Bad Bad Jenkins

PR #1785, opened 2021-08-31
Reduce the amount of logging due to timer (SC-370)

renanrodrigo:
When a job run sucessfully but noops, the timer should not log it.
Timer jobs can return True on sucessful run to indicate something was changed, and False when it noops. The timer then only adds debug logs if the job actually did something.

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/jobs/gcp_auto_attach.py
+++ b/uaclient/jobs/gcp_auto_attach.py
@@ -8,27 +8,28 @@ from uaclient.cli import action_auto_attach
 from uaclient.clouds.identity import get_cloud_type
 
 
-def gcp_auto_attach(cfg: config.UAConfig) -> None:
+def gcp_auto_attach(cfg: config.UAConfig) -> bool:
     # If the instance is already attached we will not do anything.
     # This implies that the user may have a new license attached to the
     # instance, but we will not perfom the change through this job.
     if cfg.is_attached:
-        return
+        return False
 
     # We will not do anything in a non-GCP cloud
     cloud_id, _ = get_cloud_type()
     if not cloud_id or cloud_id != "gce":
-        return
+        return False
 
     try:
         # This function already uses the assert lock decorator,
         # which means that we don't need to make create another
         # lock only for the job
         action_auto_attach(args=None, cfg=cfg)
+        return True
     except exceptions.NonAutoAttachImageError:
         # If we get a NonAutoAttachImageError we now
         # that the machine is not ready yet to perform an
         # auto-attach operation (i.e. the license may not
         # have been appended yet). If that happens, we will not
         # error out.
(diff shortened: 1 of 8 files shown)


uaclient/jobs/gcp_auto_attach.py
```diff
@@ -8,27 +8,28 @@
...
     cloud_id, _ = get_cloud_type()
     if not cloud_id or cloud_id != "gce":
-        return
+        return False
 
     try:
         # This function already uses the assert lock decorator,
         # which means that we don't need to make create another
         # lock only for the job
         action_auto_attach(args=None, cfg=cfg)
```
lucasmoura, line 27: I think we can just `return action_auto_attach(args=None, cfg=cfg)`

renanrodrigo: `action_auto_attach` will return 0 on success and 1 on failure, so it's the opposite of what the implementation expects. I preferred to put `True | False` instead of `something | None` to make clear that we are interested if the function actually executed something or not.

Which leads to a question: do we need to log every time the function tries to auto_attach, or should we watch for its return code to decide logging or not?

The attach process overall will produce logs anyway, so I think it should return `True` here in each try, so we get that the try to attach is timer related when looking at the logs.

lucasmoura: Sorry, I think I missing your point here. After a successful auto-attach this job should not be executed again because of the `cfg.is_attached` logic. Only if the user manually detaches that we will have to run it again. So why can't we just return the output of `action_auto_attach` ?

lucasmoura: Ohhh fair enough, my mistake here :facepalm:

blackboxsw:
This branch still emits two log lines every 5 minute run of our timers. Can we find a way to either avoid calling `setup_logging` unless logging is needed (due to a config changing operation being performed). Or find a way to silence the noisy logging from util.load_file/write_file functions from within the timer logging setup/config.

```bash
2021-08-31 22:57:37,473 - util.py:(418) [DEBUG]: Reading file: /var/lib/ubuntu-advantage/jobs-status.json
2021-08-31 22:57:37,478 - util.py:(675) [DEBUG]: Writing file: /var/lib/ubuntu-advantage/jobs-status.json
```

renanrodrigo:
@blackboxsw code updated - I went the `silent` way.
- It is cleaner that conditional setup_logging in job execution logic
- It was already partially present in the code
- It is easier to remove/change if we decide to log less in the future

renanrodrigo:
Please ignore this as it only solves the problem partially, and will get too intricate to tackle everything.

PR #1783, opened 2021-08-30
ros: split entitlement into ros and ros-updates (SC-372)

orndorffgrant:
## Notes
Depends on https://github.com/CanonicalLtd/ua-contracts/pull/2348 plus IS ticket to setup mirroring for ros vs ros-updates APT repos

**We decided to hide ros in this PR so that we can land it prior to the ros infrastructure changes being ready.**

## Proposed Commit Message

```
ros: split entitlement into ros and ros-updates
```
```
tests: avoid extra sbuilding
```
```
tests: fail loudly when sbuild fails
```
```
ros: hide ros until infrastructure is in place

I left comments beginning with TODO-ROS everywhere that we will need to
touch when we come back to enable this feature. These are mostly in the
tests to add "ros" and "ros-updates" back to the expected output.
```

## Test Steps

Ensure nothing is broken and ros is sufficiently hidden for now.

## Desired commit type::

 - [ ] Squash and merge: Using the "Proposed Commit Message"
 - [ ] Create a merge commit and use "Proposed Commit Message"
 - [x] Rebase and merge, no merge commit, rely only on existing commit messages

## Checklist:

 - [ ] I have updated or added any unit tests accordingly
 - [x] I have updated or added any integration tests accordingly
 - [x] I have updated or added any documentation accordingly

PR #1786, opened 2021-08-31
tests: use correct instance ids for snapshots

orndorffgrant:
## Notes
The main thing here is fixing a bug I introduced in #1751. I had just used the instance name instead of calling `cloud_manager.get_instance_id` which returns the instance id for some platforms. We now are correctly using the result of `get_instance_id`.

I also changed some parameter names to be called `id` instead of `name` to match the name of `get_instance_id`

## Proposed Commit Message

```
tests: use correct instance ids for snapshots
```

## Test Steps

Run a test on each platform with `UACLIENT_BEHAVE_SNAPSHOT_STRATEGY=1`

## Desired commit type::

## Checklist:

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -602,12 +602,12 @@ def after_all(context):
 
 
 def capture_container_as_image(
-    container_name: str, image_name: str, cloud_api: pycloudlib.cloud.BaseCloud
+    container_id: str, image_name: str, cloud_api: pycloudlib.cloud.BaseCloud
 ) -> str:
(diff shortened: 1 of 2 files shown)


renanrodrigo (approved):
worksforme

blackboxsw (approved):
Good for me too, known expected CI failure due to staging changes esm-ros -> esm
  features/staging_commands.feature:168  Attached enable esm-ros on a machine -- @1.1 ubuntu release this will be addressed in PR #1783

Issue #553, opened 2019-06-04
Duplicate console logging setup in `cli.setup_logging`


OddBloke:
We look for an existing handler or create our own, and we perform similar actions for each. Pretty low-hanging fruit for a refactor (but more than I wanted to do right now because there isn't test coverage).

powersj:
We now have 4 CLI bugs open and I think this is the biggest of the 4. 

@OddBloke could you shed some light on what you might want to see changed? Possibly giving me something to do on the flight back :)

OddBloke:
This specifically refers to the duplication here: https://github.com/CanonicalLtd/ubuntu-advantage-client/blob/925b85af49eb411cd5f49a2875061010a86ec863/uaclient/cli.py#L368-L381 So I think it's a relatively simple refactor.

PR #1789, opened 2021-09-02
Change `setup_logging` and the timer logger (SC-370)

renanrodrigo:
This PR changes the way we set up the logging, so we have better control of our handlers, and uses that change to make the timer and the timer jobs run without logging everything they see.

The `setup_logging` function now accepts a `logger` parameter, so we can configure any logger instance, and not just the root logger. It gives us control to configure module based loggers.
Also, it will wipe out existing handlers, and add the uaclient handlers (for console and log file), ensuring we have a single instance of each handler every time the loggers are configured / reconfigured. For this to work, it's important that setup_logging is inside our main entrypoints (CLI and the `lib/` scripts). When importing anything from our modules, the root logger will not be changed by default.
`lib/timer.py` will then have its own logger configured, which will not propagate to the root logger. Any message triggered by the timer will have to come from the local one.

As a curious side-effect:
Fixes: #553

## Desired commit type::

## Checklist:

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -434,7 +434,8 @@ Feature: Command behaviour when attached to an UA subscription
         update_status_timer     +43200
         gcp_auto_attach_timer   +0
         """
-        And I verify that running `grep "Disabling gcp_auto_attach job" /var/log/ubuntu-advantage.log` `with sudo` exits `0`
+        And I verify that running `grep "Disabling gcp_auto_attach job" /var/log/ubuntu-advantage-timer.log` `with sudo` exits `0`
+        And I verify that running `grep "Disabling gcp_auto_attach job" /var/log/ubuntu-advantage.log` `with sudo` exits `1`
         When I delete the file `/var/lib/ubuntu-advantage/jobs-status.json`
         And I run `ua config set update_messaging_timer=0` with sudo
         And I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
@@ -494,18 +495,9 @@ Feature: Command behaviour when attached to an UA subscription
           metering_timer: 0
         """
         And I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
-        Then stderr matches regexp:
-        """
-        Invalid value for update_messaging interval found in config.
-        """
-        And stderr matches regexp:
-        """
-        Invalid value for update_status interval found in config.
-        """
-        And stderr does not match regexp:
-        """
-        Invalid value for gcp_auto_attach interval found in config.
-        """
+        Then I verify that running `grep "Invalid value for update_messaging interval found in config." /var/log/ubuntu-advantage-timer.log` `with sudo` exits `0`
+        And I verify that running `grep "Invalid value for update_status interval found in config." /var/log/ubuntu-advantage-timer.log` `with sudo` exits `0`
+        And I verify that running `grep "Invalid value for gcp_auto_attach interval found in config." /var/log/ubuntu-advantage-timer.log` `with sudo` exits `1`
         And I verify that the timer interval for `update_messaging` is `21600`
         And I verify that the timer interval for `update_status` is `43200`
         And I verify that the timer interval for `gcp_auto_attach` is `0`
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -415,7 +415,7 @@ Feature: Command behaviour when attached to an UA subscription
         Then I verify that running `ua config set update_messaging_timer=-2` `with sudo` exits `1`
         And stderr matches regexp:
         """
-        <value> for interval must be a positive integer.
+        Cannot set update_messaging_timer to -2: <value> for interval must be a positive integer.
         """
         When I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
         And I run `cat /var/lib/ubuntu-advantage/jobs-status.json` with sudo
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -24,14 +24,14 @@ class TimedJob:
     def __init__(
         self,
         name: str,
-        job_func: Callable[..., None],
+        job_func: Callable[..., bool],
         default_interval_seconds: int,
     ):
         self.name = name
         self._job_func = job_func
         self._default_interval_seconds = default_interval_seconds
 
-    def run(self, cfg: UAConfig):
+    def run(self, cfg: UAConfig) -> bool:
         """Run a job in a failsafe manner, returning True on success.
         Checks if the job is not disabled before running it.
 
@@ -43,36 +43,30 @@ class TimedJob:
         if not self._should_run(cfg):
             return False
 
-        LOG.debug("Running job: %s", self.name)
         try:
-            self._job_func(cfg=cfg)
+            if self._job_func(cfg=cfg):
+                LOG.debug("Executed job: %s", self.name)
         except Exception as e:
             LOG.warning("Error executing job %s: %s", self.name, str(e))
             return False
 
         return True
 
-    def run_interval_seconds(self, cfg: UAConfig):
+    def run_interval_seconds(self, cfg: UAConfig) -> int:
         """Return the run_interval for the job based on config or defaults."""
         configured_interval = getattr(cfg, "{}_timer".format(self.name), None)
         if configured_interval is None:
-            debug_msg = (
-                "No config set for {}, default value will be used."
-            ).format(self.name)
-            LOG.debug(debug_msg)
             return self._default_interval_seconds
-        elif (
-            not isinstance(configured_interval, int) or configured_interval < 0
-        ):
-            error_msg = (
+        if not isinstance(configured_interval, int) or configured_interval < 0:
+            warning_msg = (
                 "Invalid value for {} interval found in config. "
                 "Default value will be used."
             ).format(self.name)
-            LOG.error(error_msg)
+            LOG.warning(warning_msg)
             return self._default_interval_seconds
         return configured_interval
 
-    def _should_run(self, cfg):
+    def _should_run(self, cfg) -> bool:
         """Verify if the job has a valid (non-zero) interval."""
         return self.run_interval_seconds(cfg) != 0
 
@@ -95,7 +89,6 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
     Persist jobs-status with calculated next_run values to aid in timer
     state introspection for jobs which have not yet run.
     """
-    LOG.debug("Trigger UA Timer jobs")
     jobs_status = cfg.read_cache("jobs-status") or {}
     for job in UACLIENT_JOBS:
         if job.name in jobs_status:
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -47,7 +47,7 @@ class TimedJob:
             if self._job_func(cfg=cfg):
                 LOG.debug("Executed job: %s", self.name)
         except Exception as e:
-            LOG.warning("Error executing job %s: %s", self.name, str(e))
+            LOG.error("Error executing job %s: %s", self.name, str(e))
             return False
 
         return True
@@ -110,5 +110,17 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
 if __name__ == "__main__":
     cfg = UAConfig()
     current_time = datetime.utcnow()
-    setup_logging(logging.INFO, logging.DEBUG)
+
+    # The ua-timer logger should log everything to its file
+    setup_logging(
+        logging.CRITICAL,
+        logging.DEBUG,
+        log_file=cfg.timer_log_file,
+        logger=LOG,
+    )
+    # Make sure the ua-timer logger does not generate double logging
+    LOG.propagate = False
+    # The root logger should log any error to the timer log file
+    setup_logging(logging.CRITICAL, logging.ERROR, log_file=cfg.timer_log_file)
+
     run_jobs(cfg=cfg, current_time=current_time)
--- a/systemd/ua-timer.service
+++ b/systemd/ua-timer.service
@@ -1,7 +1,6 @@
 [Unit]
 Description=Ubuntu Advantage Timer for running repeated jobs
 After=network.target network-online.target systemd-networkd.service ua-auto-attach.service
-Wants=ua-auto-attach.service
 
 [Service]
 Type=oneshot
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -682,8 +682,13 @@ def action_config_set(args, *, cfg, **kwargs):
                 raise ValueError("Invalid interval for {}".format(set_key))
         except ValueError:
             subparser.print_help()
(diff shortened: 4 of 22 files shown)


renanrodrigo:
One single thing that is still being logged here is when the configuration for the interval is invalid or missing. Logs are there because we need to check the config even to decide if the job should run or not.

I vote for removing the missing one.
- If you update the package but not the config file, the interval config will be missing, using the default, and then we can log it only when the job runs.
- If you put there an invalid value, then you deserve to have an ERROR log every 5min to your file, for your bad deeds.

renanrodrigo:
Another one: if gcp_auto_attach fails for any reason, no logs will be produced here. If it succeeds, just a generic "Successful Job" is being produced. All the attach process logs are lost.
Let me know if it should behave differently.

lib/timer.py
```diff
@@ -117,5 +116,8 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
 if __name__ == "__main__":
     cfg = UAConfig()
     current_time = datetime.utcnow()
-    setup_logging(logging.INFO, logging.DEBUG)
+
+    setup_logging(logging.CRITICAL, logging.DEBUG, logger=LOG)
```
lucasmoura, line 120: Since we are going to have two distinct LOG accessing the same file, can't we have that race condition issue here ?

renanrodrigo: Good question, I had the same doubt and went through it: we will not.
In the case that a different process has a logger logging to the same file (for instance, timer.py runs exactly at the same time as you type a ua command), the loggers will not write the file at the same time.
Worse that can happen is that the lines will be interspersed.

orndorffgrant:
This looks good! I haven't done any thorough testing yet but two things:

1. I agree with your assessment of the configuration value logs. Remove log for missing value. Keep log for invalid value.
2. I would expect that we also need to remove any existing handlers from the root logger, since I thought the root logger came with some default handlers. We don't want stray default stdout logs filling up syslog or something. I haven't actually tested this yet, but wanted to mention here.

renanrodrigo:
>     2. I would expect that we also need to remove any existing handlers from the root logger, since I thought the root logger came with some default handlers. We don't want stray default stdout logs filling up syslog or something. I haven't actually tested this yet, but wanted to mention here.

Setting `LOG.propagate = False` tells our logger to execute its handlers and go home, without sending anything to the root logger (:

renanrodrigo:
Thanks @lucasmoura for pointing me to the failing test - it is fixed now.

orndorffgrant:
> > ```
> > 2. I would expect that we also need to remove any existing handlers from the root logger, since I thought the root logger came with some default handlers. We don't want stray default stdout logs filling up syslog or something. I haven't actually tested this yet, but wanted to mention here.
> > ```
> 
> Setting `LOG.propagate = False` tells our logger to execute its handlers and go home, without sending anything to the root logger (:

@renanrodrigo - right, but for example, `util.load_file` will still call `logging.debug("Reading file: %s", filename)` which will use the root logger directly right? (I could totally be wrong, but this is my worry)

renanrodrigo:
@orndorffgrant yes, you are right. Thanks for noting.
In this case, as we didn't modify the root logger, only  `warning` or greater messages will be logged to `stderr`. So we don't need to care about the `debug`, but I will carefully check for warnings and their impact. Errors should be logged no matter what I guess.

orndorffgrant:
@renanrodrigo Ah TIL the root logger only logs warning and greater by default :)

I think logging warnings/errors that happen is probably okay since they _should_ only happen in weird/unexpected cases.

Maybe we should still call `setup_logging(logging.CRITICAL, logging.WARNING)` for the root logger, so that any warnings/errors that do get logged go to our log file?

Or we could only allow errors through from the root logger, so `setup_logging(logging.CRITICAL, logging.ERROR)`?

renanrodrigo:
This PR is on hold waiting for decisions about how to handle `gcp_auto_attach`, as discussed with the team.
There is a known failing test, which will be taken care of once we know what to do from here on.

renanrodrigo:
This is ready for review again.
CI failure is unrelated, bad bad Jenkins timed out on FIPS test again.

lib/timer.py
```diff
@@ -117,5 +112,17 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
 if __name__ == "__main__":
     cfg = UAConfig()
     current_time = datetime.utcnow()
-    setup_logging(logging.INFO, logging.DEBUG)
+
+    setup_logging(
+        logging.CRITICAL,
+        logging.DEBUG,
+        log_file="/var/log/ubuntu-advantage-timer.log",
```
lucasmoura, line 119: Let's add this into the config dict, the same way we have for the base log path

lib/timer.py
```diff
@@ -117,5 +112,17 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
...
+        logger=LOG,
+    )
+    LOG.propagate = False
```
lucasmoura, line 122: Do we still need this call ?

lib/timer.py
```diff
@@ -117,5 +112,17 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
...
+
+    # Root logger should still send errors to the log file,
+    # but nothing to console here
+    setup_logging(logging.CRITICAL, logging.ERROR)
```
lucasmoura, line 126: Do we need to setup the root logger as well ?

features/attached_commands.feature
```diff
@@ -494,18 +494,9 @@ Feature: Command behaviour when attached to an UA subscription
...
-        """
-        And stderr matches regexp:
-        """
-        Invalid value for update_status interval found in config.
-        """
-        And stderr does not match regexp:
-        """
-        Invalid value for gcp_auto_attach interval found in config.
-        """
+        Then I verify that running `grep "Invalid value for update_messaging interval found in config." /var/log/ubuntu-advantage.log` `with sudo` exits `0`
```
lucasmoura, line 497: We should change the path for the grep command, right ?

renanrodrigo: Right, it just failed here - will do

lib/timer.py
renanrodrigo: Yes. As we are running jobs that might have their own logging, we want to capture it when those are executed - any error would go to the general log file.
An alternative here would be configuring only one logger, and log everything (including eventual errors in executed code  inside of the jobs) to the timer file.
What do you think? Should we have it all together?

lib/timer.py
renanrodrigo: I think we can change the whole configuration to the root logger now that we are going for a separate file. Will update.

lib/timer.py
renanrodrigo: We decided to use a separate file for the timer, avoiding any risk

renanrodrigo:
@lucasmoura in the end, we needed the root log configuration and the `propagate = false` to make sure we don't duplicate logs, and that we log eventual errors when those happen.

features/attached_commands.feature
```diff
@@ -434,7 +434,8 @@ Feature: Command behaviour when attached to an UA subscription
         update_status_timer     +43200
         gcp_auto_attach_timer   +0
         """
-        And I verify that running `grep "Disabling gcp_auto_attach job" /var/log/ubuntu-advantage.log` `with sudo` exits `0`
+        And I verify that running `grep "Disabling gcp_auto_attach job" /var/log/ubuntu-advantage-timer.log` `with sudo` exits `0`
```
lucasmoura, line 437: When running the test locally, this entry does not appear on the `/var/log/ubuntu-advantage-timer.log` log or `/var/log/ubuntu-advantage.log`

renanrodrigo: This is my bad - fixed now.
I had a shared logger for the timer and job, and then dismissed it because it didn't look so right.
Added a setup_logging call to the gcp job matching the file to ensure it.

Weirdly, my local tests were passing :(

uaclient/cli.py
```diff
@@ -1223,40 +1223,40 @@ def action_help(args, *, cfg):
...
     log_formatter = logging.Formatter(DEFAULT_LOG_FORMAT)
-    root = logging.getLogger()
-    root.setLevel(log_level)
+    if logger is None:
+        # Then we configure the root logger
+        logger = logging.getLogger()
+    logger.setLevel(log_level)
+
+    # Our handlers are the best
+    logger.handlers = []
```
blackboxsw, line 1239: I think this is reasonable for use to clear out existing loggers on each call to setup_logging. That said, I think it is worth a more descriptive  comment about "# clearing all handlers so any previous handlers for this logger are replaced"

uaclient/jobs/gcp_auto_attach.py
```diff
@@ -5,51 +5,60 @@
...
+from uaclient.cli import action_auto_attach, setup_logging
 from uaclient.clouds.gcp import GCP_LICENSES, UAAutoAttachGCPInstance
 from uaclient.clouds.identity import get_cloud_type
 from uaclient.util import get_platform_info
 
 LOG = logging.getLogger(__name__)
 
 
-def gcp_auto_attach(cfg: config.UAConfig) -> None:
+def gcp_auto_attach(cfg: config.UAConfig) -> bool:
```
blackboxsw, line 16: We can drop this right because setup_logging is already called in timer.py:main. @orndorffgrant will have to sort this with his separate PR to ua-license-attach. I don't think you need this right?

blackboxsw (changes requested):
Thanks for this @renanrodrigo. I'm not quite certain our expectations of setup_logging are correct. I believe the second call to setup_logging is probably unnecessary too in lib/timer.py.

Please correct my misunderstandings inline.

systemd/ua-timer.service
```diff
@@ -1,7 +1,6 @@
 [Unit]
 Description=Ubuntu Advantage Timer for running repeated jobs
 After=network.target network-online.target systemd-networkd.service ua-auto-attach.service
-Wants=ua-auto-attach.service
```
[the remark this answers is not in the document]
renanrodrigo: I understand the idea, but it would not work like that because the `ua-auto-attach.service` is a oneshot.
So, instead of ensure _it is running_,  the `Wants=` there ensures _it runs_. With the directive, the service itself is trying to auto-attach even before the timer service runs, and then tries again (but checking all the variables we set there).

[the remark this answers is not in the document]
renanrodrigo: Without the wants, the auto-attach is initiated only by the timer service.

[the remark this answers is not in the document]
renanrodrigo: About the lock, the `After=` clause is already telling the timer to wait for the `ua-auto-attach.service` before it runs, so we're keeping it here.

uaclient/tests/test_cli.py
```diff
@@ -727,6 +727,60 @@ def test_file_log_configured_if_root(
...
+    ):
+        log_file = tmpdir.join("file.log")
+        cfg = FakeConfig({"log_file": log_file.strpath})
+        m_config.return_value = cfg
+        root_logger = logging.getLogger()
+
+        setup_logging(logging.INFO, logging.INFO)
+        assert len(root_logger.handlers) == 2
+
+        setup_logging(logging.INFO, logging.INFO)
```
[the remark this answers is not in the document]
renanrodrigo: That's a great catch, will do

uaclient/cli.py
renanrodrigo: LOL this comment was not meant to be there at all, it was just part of the development thinking process :eyes: 
Will update

uaclient/jobs/gcp_auto_attach.py
renanrodrigo: It's just matching the calls for this timer. I would make it working in this PR, and then @orndorffgrant can adapt to his approach when separate the timer. What do you think, Grant?

lib/timer.py
```diff
@@ -117,5 +111,17 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
...
+    setup_logging(
+        logging.CRITICAL,
+        logging.DEBUG,
+        log_file=cfg.timer_log_file,
+        logger=LOG,
+    )
+    # Make sure the ua-timer logger does not generate double logging
+    LOG.propagate = False
+    # The root logger should log any error to the timer log file
+    setup_logging(logging.CRITICAL, logging.ERROR, log_file=cfg.timer_log_file)
```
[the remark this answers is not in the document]
renanrodrigo: Yes, we do have the exception, but it would ignore all errors in the process, handled by the root logger. Besides that, this setup_logging call is ensuring we log the "Disabling" message (which is an INFO) when not in GCE.

uaclient/jobs/gcp_auto_attach.py
orndorffgrant: Whatever is less work since I will be editing this. As long as there is a pattern fir me to follow so that I don't accidentally lose the benefits of this PR

uaclient/conftest.py
```diff
@@ -77,8 +77,9 @@ def logging_sandbox():
 @pytest.fixture
 def FakeConfig(tmpdir):
     class _FakeConfig(UAConfig):
-        def __init__(self, features_override=None) -> None:
-            super().__init__({"data_dir": tmpdir.strpath})
+        def __init__(self, cfg_overrides={}, features_override=None) -> None:
```
lucasmoura, line 80: Why do we need this extra `cfg_overrides` param here ?

renanrodrigo: Because otherwise, I cannot set cfg read-only properties/attributes when instantiating.
I needed to define a log_file (which is a property without a setter) for the config, so I changed FakeConfig to accept the parameter - otherwise we could change only features, with the `features_override`

systemd/ua-timer.service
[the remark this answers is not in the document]
blackboxsw: Thanks for this detail @renanrodrigo and your explanation in standup and I confirmed repeated ua-auto-attach behavior as well on every timer tick and the 'fix' without Wants= to only run once. Turns out I misunderstood how the repeated timer triggers work w.r.t. oneshot services.

uaclient/jobs/gcp_auto_attach.py
blackboxsw: Thanks @renanrodrigo for explaining this aspect for inclusion of INFO level message for disabling gcp auto-attach. This makes a lot more sense now for module-specific logging preferences

blackboxsw (approved):
Code looks good to me. Waiting on CI runs. Thank you for your explanation to correct my misread of behavior here @renanrodrigo +1.

lucasmoura (approved):

blackboxsw (approved):
LGTM! CI passed fully on https://jenkins.canonical.com/server-team-jenkins-private/blue/organizations/jenkins/uaclient-ci/detail/PR-1789/9/pipeline yet is blocked on 20.04 https://jenkins.canonical.com/server-team-jenkins-private/blue/organizations/jenkins/uaclient-ci/detail/PR-1789/10
Landing this as is as because it appears our final CI run got a Java interrupt.

PR #1792, opened 2021-09-03
proof of concept: daemon to long poll gcp metadata

orndorffgrant:
Don't merge this.

This was a quick proof of concept to explore what a daemon to long poll for gcp license changes might look like. It isn't fully fleshed out.

This actually long polls for instance metadata keyvalue changes (not licenses). You can test it by:
1. installing this on a fresh gcp instance
2. `systemctl start ua-daemon`
3. `tail -f /var/log/ubuntu-advantage.log`
4. in the gcp console (or via cli) edit the instance metadata keyvalue pairs
5. see the updated metadata and a new request started in the log
6. repeat 4-5 as much as desired

I think the thing that pushes this into "next cycle" territory though is that AFAIK gcp doesn't support changing licenses without a stop-start yet. Until they do, this wouldn't be functional and we won't be able to test this.


orndorffgrant:
Closing because we now have #1887

PR #1793, opened 2021-09-03
timer: adjust interval values for gcp_auto_attach and metering (SC-390)

renanrodrigo:
Metering should be disabled,
GCP auto attach should be 5 minutes.

## Desired commit type::

## Checklist:
(this is the first PR where I checked all 3 of those I guess hahaha)


blackboxsw (approved):
LGTM. Will ensure we have a gcp license check in the gcp job to automatically avoid roundtrips to contracts.canonical.com when no viable license info seems present.

PR #1795, opened 2021-09-07
timer: check for GCP PRO License before trying to auto attach (SC-389)

renanrodrigo:
Only try to auto attach if the PRO license is detected in the instance metadata.

## Desired commit type::

## Checklist:

### DIFF
--- a/sru/release-27.3/gcp_auto_attach_test.sh
+++ b/sru/release-27.3/gcp_auto_attach_test.sh
@@ -1,0 +1,38 @@
+#!/bin/sh
+ZONE="us-east1-b"
+INSTANCE_NAME="test-auto-attach"
+INSTANCE_TYPE="n1-standard-1"
+DISK_NAME="persistent-disk-0"
+
+set -x
+
+gcloud compute instances create $INSTANCE_NAME \
+    --image="ubuntu-minimal-1604-xenial-v20210430" \
+    --image-project="ubuntu-os-cloud" \
+    --machine-type=$INSTANCE_TYPE \
+    --zone=$ZONE
+sleep 30
+
+gcloud compute ssh $INSTANCE_NAME -- "sudo apt install software-properties-common -y"
+gcloud compute scp ubuntu-advantage-tools.deb $INSTANCE_NAME:/tmp/
+gcloud compute ssh $INSTANCE_NAME -- "sudo apt update"
+gcloud compute ssh $INSTANCE_NAME -- "sudo apt install ubuntu-advantage-tools -y"
+gcloud compute ssh $INSTANCE_NAME -- "sudo dpkg -i /tmp/ubuntu-advantage-tools.deb"
+gcloud compute ssh $INSTANCE_NAME -- "sudo sh -c \"printf \\\"    else:\n        print('pro license not present')\\\" >>  /usr/lib/python3/dist-packages/uaclient/jobs/gcp_auto_attach.py\""
+# Without the license, it will not try to auto_attach
+gcloud compute ssh $INSTANCE_NAME -- "sudo rm /var/lib/ubuntu-advantage/jobs-status.json"
+gcloud compute ssh $INSTANCE_NAME -- "sudo python3 /usr/lib/ubuntu-advantage/timer.py"
+gcloud compute ssh $INSTANCE_NAME -- "sudo ua status --wait"
+
+gcloud compute instances stop $INSTANCE_NAME
+gcloud beta compute disks update $INSTANCE_NAME --zone=$ZONE --update-user-licenses="https://www.googleapis.com/compute/v1/projects/ubuntu-os-pro-cloud/global/licenses/ubuntu-pro-1604-lts"
+gcloud compute instances start $INSTANCE_NAME
+sleep 30
+
+# Now with the license, it will succeed auto_attaching
+gcloud compute ssh $INSTANCE_NAME -- "sudo rm /var/lib/ubuntu-advantage/jobs-status.json"
+gcloud compute ssh $INSTANCE_NAME -- "sudo python3 /usr/lib/ubuntu-advantage/timer.py"
+gcloud compute ssh $INSTANCE_NAME -- "sudo ua status --wait"
+gcloud compute ssh $INSTANCE_NAME -- "sudo ua detach --assume-yes"
+
+gcloud compute instances delete $INSTANCE_NAME
--- a/uaclient/clouds/gcp.py
+++ b/uaclient/clouds/gcp.py
@@ -1,5 +1,7 @@
+import base64
+import json
 import os
-from typing import Any, Dict
+from typing import Any, Dict, List
 from urllib.error import HTTPError
 
 from uaclient import util
@@ -13,6 +15,12 @@ TOKEN_URL = (
 
 DMI_PRODUCT_NAME = "/sys/class/dmi/id/product_name"
 GCP_PRODUCT_NAME = "Google Compute Engine"
+
+GCP_LICENSES = {
+    "xenial": "8045211386737108299",
+    "bionic": "6022427724719891830",
+    "focal": "599959289349842382",
+}
 
 
 class UAAutoAttachGCPInstance(AutoAttachCloudInstance):
@@ -40,3 +48,20 @@ class UAAutoAttachGCPInstance(AutoAttachCloudInstance):
                 return True
 
         return False
+
+    def get_licenses_from_identity(self) -> List[str]:
+        """Get a list of licenses from the GCP metadata.
+
+        Instance identity token (jwt) carries a list of licenses
+        associated with the instance itself.
+
+        Returns an empty list if licenses are not present in the metadata.
+        """
+        token = self.identity_doc["identityToken"]
+        identity = base64.urlsafe_b64decode(token.split(".")[1] + "===")
+        identity_dict = json.loads(identity.decode("utf-8"))
+        return (
+            identity_dict.get("google", {})
+            .get("compute_engine", {})
+            .get("license_id", [])
+        )
--- a/uaclient/jobs/gcp_auto_attach.py
+++ b/uaclient/jobs/gcp_auto_attach.py
@@ -6,7 +6,9 @@ import logging
 
 from uaclient import config, exceptions
 from uaclient.cli import action_auto_attach
+from uaclient.clouds.gcp import GCP_LICENSES, UAAutoAttachGCPInstance
 from uaclient.clouds.identity import get_cloud_type
+from uaclient.util import get_platform_info
 
 LOG = logging.getLogger(__name__)
 
@@ -27,15 +29,27 @@ def gcp_auto_attach(cfg: config.UAConfig) -> None:
     if cfg.is_attached:
         return
 
-    try:
-        # This function already uses the assert lock decorator,
-        # which means that we don't need to make create another
-        # lock only for the job
-        action_auto_attach(args=None, cfg=cfg)
-    except exceptions.NonAutoAttachImageError:
-        # If we get a NonAutoAttachImageError we know
-        # that the machine is not ready yet to perform an
-        # auto-attach operation (i.e. the license may not
-        # have been appended yet). If that happens, we will not
-        # error out.
+    series = get_platform_info()["series"]
+    if series not in GCP_LICENSES:
         return
+
+    # Only try to auto_attach if the license is found in the metadata.
+    # If there is a problem finding the metadata, do not error out.
+    try:
+        licenses = UAAutoAttachGCPInstance().get_licenses_from_identity()
+    except Exception:
+        return
+
+    if GCP_LICENSES[series] in licenses:
+        try:
+            # This function already uses the assert lock decorator,
+            # which means that we don't need to make create another
+            # lock only for the job
+            action_auto_attach(args=None, cfg=cfg)
+        except exceptions.NonAutoAttachImageError:
+            # If we get a NonAutoAttachImageError we know
+            # that the machine is not ready yet to perform an
+            # auto-attach operation (i.e. the license may not
+            # have been appended yet). If that happens, we will not
+            # error out.
+            return
--- a/uaclient/jobs/tests/test_gcp_auto_attach.py
+++ b/uaclient/jobs/tests/test_gcp_auto_attach.py
@@ -4,9 +4,16 @@ import mock
 import pytest
 
 from uaclient.exceptions import NonAutoAttachImageError
-from uaclient.jobs.gcp_auto_attach import gcp_auto_attach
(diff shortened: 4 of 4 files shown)


renanrodrigo:
Before marking this as ready, I have just a feeling that we could have an integration test for the timed auto-attach, to make sure things go as expected.
I need some bits of guidance on how to orchestrate this tough test, though. Thoughts?

uaclient/jobs/gcp_auto_attach.py
```diff
@@ -2,15 +2,25 @@
...
 from uaclient import config, exceptions
 from uaclient.cli import action_auto_attach
+from uaclient.clouds.gcp import GCP_LICENSES, UAAutoAttachGCPInstance
 from uaclient.clouds.identity import get_cloud_type
+from uaclient.util import get_platform_info
 
 LOG = logging.getLogger(__name__)
 
 
+def get_licenses_from_identity(identityToken):
```
blackboxsw, line 18: This function feels tightly coupled with other GCP metadata specific handling. It's probably best to keep this in uaclient.clouds.gcp module so "future us" can see the impact locally if we end up altering gcp.identity_doc function to return different outputs/format/key names etc.

uaclient/jobs/gcp_auto_attach.py
```diff
@@ -2,15 +2,25 @@
...
+    identity = base64.urlsafe_b64decode(identityToken.split(".")[1] + b"===")
+    identity = json.loads(identity)
+    return identity["google"]["compute_engine"]["license_id"]
```
blackboxsw, line 21: It seems we aren't doing any error handling if the IMDS data is unavailable (due to IMDS being unresponsive/timeout) or if the format of the identityToken doesn't meet our expectations with either b64decode, json.loads,  or our expectation that the keypath ["google"]["compute_engine"]["license_id"] exists. I think we probably want to better handle those types of unexpected errors right? Maybe we add a silent "pass" in these failure cases since we expect this data to be correct some point, but we don't want to fill the logs every 5 minutes with the same error (and we don't want to talk to the contract server in those cases where the format has changed).

uaclient/jobs/gcp_auto_attach.py
```diff
@@ -27,15 +37,21 @@ def gcp_auto_attach(cfg: config.UAConfig) -> None:
...
-        # auto-attach operation (i.e. the license may not
-        # have been appended yet). If that happens, we will not
-        # error out.
-        return
+    # Only try to auto_attach if the license is found in the metadata
+    token = UAAutoAttachGCPInstance().identity_doc["identityToken"]
+    licenses = get_licenses_from_identity(token)
+    series = get_platform_info()["series"]
+
+    if GCP_LICENSES[series] in licenses:
```
blackboxsw, line 45: I think we need to be a bit more defensive here and explicit  as this will break on all non-LTS with a KeyError.
```suggestion
    if series in GCP_LICENSE and GCP_LICENSES[series] in licenses:
```

blackboxsw (changes requested):
Thanks for this @renanrodrigo I think we need a couple of fixes for making sure we don't fallover in non-LTS cases. This review raised an interesting question for me.

I'm thinking we probably also want to disable gcp_auto_attach in debian/ubuntu-advantage-tools.postinst as well if we are on non-LTS Ubuntu releases and re-enabling on LTS across upgrade path. I don't think there is any reason for us to enable gcp late-bind license attach on non-LTS. If that is true on non-LTS, there is no reason to do local work every 5 mins for something that doesn't provide benefit.  I think we can discuss whether we should provide a sparat PR for this disabling on non-LTS though.

blackboxsw:
I think in this case, let's initially create a sample manual integration test in sru/release-27/test_gcp_late_bind_license.sh which will call `gcloud beta` to add license information to a running VM. I'm uncertain if our CI GCP API tokens are permissioned to run `gcloud beta` in this capacity (I recall we individually had to be added to an "accept list" via some RBAC on Google's side). I think instrumenting this in our CI via behave may not be worth the ongoing runs for this feature, but a one-off which we run during SRU time.

uaclient/jobs/gcp_auto_attach.py
renanrodrigo: We can check for it before, so we don't even need to go for metadata

renanrodrigo:
@blackboxsw thank you for the catches, code is updated. I separated the unit tests to cover for specific things. Will try to orchestrate a test script for the integration.

uaclient/clouds/gcp.py
```diff
@@ -40,3 +48,9 @@ def is_viable(self) -> bool:
                 return True
 
         return False
+
+    def get_licenses_from_identity(self):
+        token = self.identity_doc["identityToken"]
+        identity = base64.urlsafe_b64decode(token.split(".")[1] + b"===")
+        identity = json.loads(identity)
+        return identity["google"]["compute_engine"]["license_id"]
```
lucasmoura, line 56: Maybe we should be a little more defensive when accessing those fields. I know that we surrounding this method in a try except block, but maybe we should only return an empty dict here instead.

Also, I think we should add code documentation and type hints for this method

renanrodrigo:
@lucasmoura thanks for the review, updated accordingly

lucasmoura (commented):
Awesome, thanks @renanrodrigo.
Can you provide a manual test for this change just to confirm that it is working as expected ?
Let me know if you need any help with that

uaclient/jobs/tests/test_gcp_auto_attach.py
```diff
@@ -41,20 +59,88 @@ def test_gcp_auto_attach(
...
+        m_get_licenses,
+        m_cloud_type,
+        FakeConfig,
+    ):
+        m_cloud_type.return_value = ("gce", None)
+        m_get_licenses.side_effect = TypeError("error")
+        m_platform_info.return_value = {"series": "ubuntu-lts"}
+        cfg = FakeConfig()
+
+        assert gcp_auto_attach(cfg) is None
```
lucasmoura, line 105: Can we add `call_count` check for the `m_get_licenses` mock ?

uaclient/jobs/tests/test_gcp_auto_attach.py
```diff
@@ -41,20 +59,88 @@ def test_gcp_auto_attach(
...
+        m_get_licenses,
+        m_cloud_type,
+        FakeConfig,
+    ):
+        m_cloud_type.return_value = ("gce", None)
+        m_get_licenses.return_value = ["unsupported-license"]
+        m_platform_info.return_value = {"series": "ubuntu-lts"}
+        cfg = FakeConfig()
+
+        assert gcp_auto_attach(cfg) is None
```
lucasmoura, line 125: Same thing here

uaclient/jobs/tests/test_gcp_auto_attach.py
```diff
@@ -41,20 +59,88 @@ def test_gcp_auto_attach(
...
+        m_get_licenses,
+        m_cloud_type,
+        FakeConfig,
+    ):
+        m_cloud_type.return_value = ("gce", None)
+        m_get_licenses.return_value = ["test-license-id"]
+        m_platform_info.return_value = {"series": "ubuntu-non-lts"}
+        cfg = FakeConfig()
+
+        assert gcp_auto_attach(cfg) is None
```
lucasmoura, line 145: Here too

uaclient/jobs/tests/test_gcp_auto_attach.py
renanrodrigo: Yes sure

blackboxsw:
> Thanks for this @renanrodrigo I think we need a couple of fixes for making sure we don't fallover in non-LTS cases. This review raised an interesting question for me.
> 
> I'm thinking we probably also want to disable gcp_auto_attach in debian/ubuntu-advantage-tools.postinst as well if we are on non-LTS Ubuntu releases and re-enabling on LTS across upgrade path. I don't think there is any reason for us to enable gcp late-bind license attach on non-LTS. If that is true on non-LTS, there is no reason to do local work every 5 mins for something that doesn't provide benefit. I think we can discuss whether we should provide a sparat PR for this disabling on non-LTS though.

As discussed in standup, this job NOOPs immediately on the job runner function call and doesn't emit any logs about that effort. So, it is not worth the added risk of manipulating postinst for the setup and reset of this config across LTS boundaries in this iteration since we are talking about non-LTS releases on GCP for this particular job. Also, we may have a migration in the nearterm toward using a separate daemon that handles this GCP late-bind job anyway, so we may be able to drop this job from the ua-timer in the next release.

renanrodrigo:
@blackboxsw @lucasmoura So, here is the script (based on Lucas's one, thanks!) I used to test this.
It has a hacky `print` to ensure we are not calling `auto_attach` unless the license is found, and then it auto_attaches fine when we have the license.

https://gist.github.com/renanrodrigo/a6b0c5cf9253dbb4710105bffe7d2ac1

@orndorffgrant thanks again for the tips and very useful opinions on this

orndorffgrant:
@renanrodrigo The script looks good though I haven't run it yet. I'll run it first thing tomorrow. We probably want it in the repo as well. Can you maybe add it as `sru/release-27.3/gcp_auto_attach_test.sh`?

renanrodrigo:
Added

orndorffgrant:
@renanrodrigo thanks for the script it works really well overall. I have a couple suggested tweaks to it to make it just a little more friendly:

```diff
diff --git a/sru/release-27.3/gcp_auto_attach_test.sh.sh b/sru/release-27.3/gcp_auto_attach_test.sh.sh
index 7da72022..ecc482d3 100755
--- a/sru/release-27.3/gcp_auto_attach_test.sh.sh
+++ b/sru/release-27.3/gcp_auto_attach_test.sh.sh
@@ -4,6 +4,8 @@ INSTANCE_NAME="test-auto-attach"
 INSTANCE_TYPE="n1-standard-1"
 DISK_NAME="persistent-disk-0"
 
+set -x
+
 gcloud compute instances create $INSTANCE_NAME \
     --image="ubuntu-minimal-1604-xenial-v20210430" \
     --image-project="ubuntu-os-cloud" \
@@ -16,7 +18,7 @@ gcloud compute scp ubuntu-advantage-tools.deb $INSTANCE_NAME:/tmp/
 gcloud compute ssh $INSTANCE_NAME -- "sudo apt update"
 gcloud compute ssh $INSTANCE_NAME -- "sudo apt install ubuntu-advantage-tools -y"
 gcloud compute ssh $INSTANCE_NAME -- "sudo dpkg -i /tmp/ubuntu-advantage-tools.deb"
-gcloud compute ssh $INSTANCE_NAME -- "sudo sed -i 's/disable_auto_attach = /print(\"Auto Attaching\"); disable_auto_attach = /' /usr/lib/python3/dist-packages/uaclient/cli.py"
+gcloud compute ssh $INSTANCE_NAME -- "sudo sh -c \"printf \\\"    else:\n        print('pro license not present')\\\" >>  /usr/lib/python3/dist-packages/uaclient/jobs/gcp_auto_attach.py\""
 
 # Without the license, it will not try to auto_attach
 gcloud compute ssh $INSTANCE_NAME -- "sudo rm /var/lib/ubuntu-advantage/jobs-status.json"
@@ -32,5 +34,6 @@ sleep 30
 gcloud compute ssh $INSTANCE_NAME -- "sudo rm /var/lib/ubuntu-advantage/jobs-status.json"
 gcloud compute ssh $INSTANCE_NAME -- "sudo python3 /usr/lib/ubuntu-advantage/timer.py"
 gcloud compute ssh $INSTANCE_NAME -- "sudo ua status --wait"
+gcloud compute ssh $INSTANCE_NAME -- "sudo ua detach --assume-yes"
 
 gcloud compute instances delete $INSTANCE_NAME
```

renanrodrigo:
@orndorffgrant ready and updated

sru/release-27.3/gcp_auto_attach_test.sh
```diff
@@ -0,0 +1,38 @@
...
+    --machine-type=$INSTANCE_TYPE \
+    --zone=$ZONE
+sleep 30
+
+gcloud compute ssh $INSTANCE_NAME -- "sudo apt install software-properties-common -y"
+gcloud compute scp ubuntu-advantage-tools.deb $INSTANCE_NAME:/tmp/
+gcloud compute ssh $INSTANCE_NAME -- "sudo apt update"
+gcloud compute ssh $INSTANCE_NAME -- "sudo apt install ubuntu-advantage-tools -y"
+gcloud compute ssh $INSTANCE_NAME -- "sudo dpkg -i /tmp/ubuntu-advantage-tools.deb"
++gcloud compute ssh $INSTANCE_NAME -- "sudo sh -c \"printf \\\"    else:\n        print('pro license not present')\\\" >>  /usr/lib/python3/dist-packages/uaclient/jobs/gcp_auto_attach.py\""
```
orndorffgrant, line 21: @renanrodrigo minor copy/paste mistake :)
```suggestion
gcloud compute ssh $INSTANCE_NAME -- "sudo sh -c \"printf \\\"    else:\n        print('pro license not present')\\\" >>  /usr/lib/python3/dist-packages/uaclient/jobs/gcp_auto_attach.py\""
```

renanrodrigo: damn did I commit it
sheesh

renanrodrigo:
@orndorffgrant was not a copy paste, I actually typed the + there intentionally, but since you wanted me to remove it 🤔 🤥

lucasmoura (approved):

renanrodrigo:
CI has thrown a timeout party :|

uaclient/jobs/gcp_auto_attach.py
```diff
@@ -27,15 +29,27 @@ def gcp_auto_attach(cfg: config.UAConfig) -> None:
...
-        # auto-attach operation (i.e. the license may not
-        # have been appended yet). If that happens, we will not
-        # error out.
+        licenses = UAAutoAttachGCPInstance().get_licenses_from_identity()
+    except Exception:
         return
+
+    if GCP_LICENSES[series] in licenses:
+        try:
+            # This function already uses the assert lock decorator,
```
blackboxsw, line 45: Thanks for this comment here for the future.

blackboxsw (approved):
This looks great. Thanks @renanrodrigo for the code changes, error handling and comments. 
The test script works great, and I've added a followup with GCP folks on when we might expect IMDS to support license attach without also a stop/start of the instance.

uaclient/jobs/gcp_auto_attach.py
renanrodrigo: @lucasmoura ^

renanrodrigo:
CI tests related to this already executed fine. Suite is going to fail anyway, so merging it.

PR #1796, opened 2021-09-07
Release 27.3

orndorffgrant:
Created with `git rebase --onto upstream/release-27 664c0795d8ca6fe81301b17d809106e8ad4bd64b` and fixing changelog conflicts (where that commit hash is the last commit in the main branch that release-27 already had in order)

## TODO
- [x] update with new commits as last few PRs land
- [x] add changelog
    - [x] include LP references
- [x] fix line 154 in debian/changelog to match what is in main and in pkg/ubuntu/devel

## To Test/Review:
`git diff main` to make sure the contents match what we have in main

`git diff pkg/ubuntu/devel -- debian/changelog` to check the changelog diff

Review changelog using the checklist in release docs:
- [ ] Is the version correctly updated ? We must ensure that the new version on the changelog is correct and it also targets the latest Ubuntu release at the moment.
- [ ] Is the entry useful for the user ? The changelog entries should be user focused, meaning that we should only add entries that we think users will care about (i.e. we don't need entries when fixing a test, as this doesn't provide meaningful information to the user)
- [ ] Is this entry redundant ? Sometimes we may have changes that affect separate modules of the code. We should have an entry only for the module that was most affected by it
- [ ] Is the changelog entry unique ? We need to verify that the changelog entry is not already reflected in an earlier version of the changelog. If it is, we need not only to remove but double check the process we are using to cherry-pick the commits
- [ ] Is this entry actually reflected on the code ? Sometimes, we can have changelog entries that are not reflected in the code anymore. This can happen during development when we are still unsure about the behavior of a feature or when we fix a bug that removes the code that was added. We must verify each changelog entry that is added to be sure of their presence in the product.


debian/changelog
```diff
@@ -1,7 +1,68 @@
...
+        status information for proper contract metering
+      + disabled until infrastructure is ready
+    - ros:
+      + add beta support to enable ros and ros-updates
+      + add support for required services so that esm-infra and esm-apps
+        get auto-enabled when enabling ros or ros-updates
+      + add support for dependent services so that user gets prompted to
+        disable ros/ros-updates if they disable esm-infra/esm-apps
+      + hide ros and ros-updates entitlements until infrastructure is ready
+    - dev tooling:
```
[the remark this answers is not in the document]
renanrodrigo: Agreed.
I think we could summarize it in a couple entries, like
`+ improve integration testing code and environment`
or something like this, so we note it was done, but without detail

debian/changelog
```diff
@@ -22,11 +83,9 @@ ubuntu-advantage-tools (27.2~21.10.1) impish; urgency=medium
 
   * d/control:
     - add comments to explain complex build-depends
-    - remove unnecessary distro-info dependency from build-depends
```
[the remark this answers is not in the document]
orndorffgrant: These changes down here are bringing in changes that happened while cleaning up release 27.2 that never made it into the `release-27` branch somehow.

If you diff this file against pkg/ubuntu/devel the only change is the new entry for 27.3 up top.

debian/changelog
```diff
@@ -1,7 +1,70 @@
-ubuntu-advantage-tools (27.2.2~21.10) impish; urgency=medium
+ubuntu-advantage-tools (27.3~21.10.1) impish; urgency=medium
 
   * d/tools.postinst:
-    - Do not fail in postinst if cloud-init did not run. (LP: #1936833)
+    - handle any cloud-id error and treat as "none" cloud
```
renanrodrigo, line 4: Suggestion: "Consider cloud to be "none" on any cloud-id error"

debian/changelog
```diff
@@ -1,7 +1,70 @@
...
+        fips-certified kernel is ready (LP: #1939449, LP: #1939932)
+      + print warning if cloud-id fails about installing generic fips kernel
+    - cloud:
+      + rely only on cloud-id to determine cloud type (LP: #1940131)
+      + catch errors when determining cloud type (LP: #1938207) (GH: #1541)
+    - azure:
+      + bump IMDS API version to support Azure published images
+    - proxy:
+      + shows better error message when a port value in a proxy is invalid
+    - cli:
```
renanrodrigo, line 29: (maybe cli is not exactly the place for this but) I think we should include in the changelog that we fixed the caches, making some of them permanent - I think this bug has (low, but still has) relevance to the users

renanrodrigo (commented):
two minor suggestions to the changelog, otherwise looks great

debian/changelog
```diff
@@ -1,7 +1,70 @@
-ubuntu-advantage-tools (27.2.2~21.10) impish; urgency=medium
+ubuntu-advantage-tools (27.3~21.10.1) impish; urgency=medium
 
   * d/tools.postinst:
-    - Do not fail in postinst if cloud-init did not run. (LP: #1936833)
+    - handle any cloud-id error and treat as "none" cloud
+    - purge old ua-messaging.timer/service
+  * systemd:
+    - add new ua-timer.timer that runs every 2 hours
+    - add new ua-license_check.timer that runs every 5 minutes only if
```
renanrodrigo, line 8: Should not we state here that we are removing the existing `ua-messaging` timer as well? So it's clear that we are not keeping those two plus the existing one

orndorffgrant:
@renanrodrigo updated!

debian/changelog
```diff
@@ -1,7 +1,72 @@
...
+        fips-certified kernel is ready (LP: #1939449, LP: #1939932)
+      + print warning if cloud-id fails about installing generic fips kernel
+    - cloud:
+      + rely only on cloud-id to determine cloud type (LP: #1940131)
+      + catch errors when determining cloud type (LP: #1938207) (GH: #1541)
+    - azure:
+      + bump IMDS API version to support Azure published images
+    - proxy:
+      + shows better error message when a port value in a proxy is invalid
+    - cli:
```
blackboxsw, line 30: Proxy note should probably be under cli scope
```suggestion
    - proxy:
      
    - cli:
      + ua config: shows better error on invalid proxy port value
```

debian/changelog
```diff
@@ -1,7 +1,72 @@
...
+      + collect-logs command that creates a tar file with debug-relevant logs
+        and status info (GH: #463)
+      + clean locks on exceptions more thoroughly to avoid false "Operation in
+        progress" status messages
+      + retain past service state after detach
+    - non-unicode locale support:
+      + remove unicode-only characters from help file
+      + don't print unicode-only characters in ua fix if non-utf8 locale
+        (GH: #1463)
+    - timer for recurring jobs:
```
blackboxsw, line 40:
```suggestion
    - ua-timer for recurring jobs:
```

blackboxsw (commented):
couple nits. Looks great!

debian/changelog
```diff
@@ -1,7 +1,72 @@
...
+      + individual jobs are disabled if their interval is set to 0
+      - status job:
+        + update ua status every 12 hours
+      - messaging job:
+        + update APT/MOTD ESM messaging every 6 hours
+      - metering job:
+        + disabled until infrastructure is ready
+        + for attached machines only, periodically update contract server with
+          status information for proper contract metering
+    - gcp auto attach timer:
```
lucasmoura, line 53: We need to rename that entry

debian/changelog
```diff
@@ -1,7 +1,72 @@
...
+      + add beta support to enable ros and ros-updates
+      + add support for "required services" so that esm-infra and esm-apps
+        get auto-enabled when enabling ros or ros-updates
+      + add support for "dependent services" so that user gets prompted to
+        disable ros/ros-updates if they disable esm-infra/esm-apps
+    - fips:
+      + allow fips on GCO bionic now that optimized kernel is ready
+      + disallow enabling fips on focal on clouds until cloud-optimized focal
+        fips-certified kernel is ready (LP: #1939449, LP: #1939932)
+      + print warning if cloud-id fails about installing generic fips kernel
```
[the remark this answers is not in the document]
orndorffgrant: Aha I see the confusion, I think it should actually be
```suggestion
      + print warning about generic fips kernel if cloud-id fails 
```

debian/changelog
orndorffgrant: Ah right, to `ua-license-check.timer` is what you're suggesting?

orndorffgrant:
@blackboxsw @lucasmoura updated!

lucasmoura (approved):

PR #1797, opened 2021-09-09
fips: fix focal cloud block on PRO instances

lucasmoura:
## Proposed Commit Message
fips: fix focal cloud block on PRO instances

On PRO instances, the contract does return the cloud specific fips metadata to be installed. However, this package doesn't exist
yet and the check we are performing for the PRO instances is not taking that into consideration. We are updating the logic to properly handle the PRO instances

## Test Steps
1. Launch a focal PRO machine with the proposed changes
2. Try to enable fips on it and verify that it outputting the correct message stating why fips can't be enabled

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -148,7 +148,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                 config=self.cfg.cfg, path_to_value=cfg_path
             ):
                 return True
-            return "ubuntu-{}-fips".format(cloud_id) in super().packages
+            return False
 
         if cloud_id not in ("azure", "gce"):
             return True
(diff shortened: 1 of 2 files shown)


renanrodrigo (approved):
lgtm

blackboxsw (approved):
+1 confirmed failure mode and fix
```bash
# With daily build
DEBUG: Failed running command 'apt-get install --assume-yes --allow-downgrades -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" ubuntu-azure-fips' [exit(100)]. Message: E: Unable to locate package ubuntu-azure-fips
...
Could not enable FIPS.



This will install the FIPS core packages.
Are you sure? (y/N) y
Ubuntu Focal does not provide an Azure optimized FIPS kernel
For help see: https://ubuntu.com/advantage.
```

lucasmoura:
CI errors are unrelated and could find an issue with the focal fips vm tests

PR #1799, opened 2021-09-09
Separate gcp timer

orndorffgrant:
## Proposed Commit Message

>
>
>

## Desired commit type::

## Checklist:


debian/ubuntu-advantage-tools.postinst
```diff
@@ -302,6 +303,25 @@ notify_wrong_fips_metapackage_on_cloud() {
...
+        fi
+    fi
+    # TODO remove this next line, just using it for testing
+    # touch $LICENSE_CHECK_MARKER_FILE
+
+    # HEY REVIEWERS LOOK HERE this next line seems scary to have here
+    systemctl restart ua-license-check.timer
+
+    # TODO remove this next line, just using it for testing
+    # systemctl status ua-license-check.timer
```
blackboxsw, line 322: Hrm this is complicated: I had thought we didn't need to call systemctl restart ua-license-check.timer ourselves. We- instead rely on systemd to start the service for us by using a systemd.path and systemd.service file something like this
```
cat > /lib/systemd/system/ua-license-check.path <<EOF
[Unit]
Description=Ubuntu Advantage check for GCP Ubuntu Pro licenses

[Path]
PathExists=/var/lib/ubuntu-advantage/marker-license-check
Unit=ua-license-check.service

[Install]
WantedBy=multi-user.target
EOF

cat > /lib/systemd/system/ua-license-check.service <<EOF
[Unit]
Description=Ubuntu Advantage check for GCP Ubuntu Pro licenses

[Service]
Type=oneshot
ExecStart=/usr/bin/python3 /usr/lib/ubuntu-advantage/gcp_auto_attach.py
TimeoutSec=0

[Install]                 
WantedBy=multi-user.target

EOF
````



But that is a path conditional which would activate the related systemd service for us at the presence of the marker file at any time after boot.   I'm not certain offhand how we fold the systemd timer into that equation, unless maybe we set Unit=ua-license-check.timer instead of Unit=ua-license-check.service in the ua-license-check.path file.

File: lib/gcp_auto_attach.py [hunk 2]
```diff
@@ -18,7 +19,9 @@ def gcp_auto_attach(cfg: config.UAConfig) -> None:
         # If we are not running on GCP cloud, we shouldn't run this
         # job anymore
         LOG.info("Disabling gcp_auto_attach job. Not running on GCP instance")
-        cfg.gcp_auto_attach_timer = 0
+        cfg.delete_cache_key("marker-license-check")
+        # HEY REVIEWERS LOOK HERE this next line seems scary to have here
+        util.subp(["systemctl", "restart", "ua-license-check.timer"])
```
blackboxsw, line 24: If we are deleting the marker file, and our systemd PathExists declaration works to properly enable whenever that marker file appears, then I think what we need here is `systemctl stop ua-license-check.timer`.

systemd/ua-license-check.timer
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Ubuntu Advantage Timer to check for GCP Ubuntu Pro licenses
+ConditionPathExists=!/var/lib/ubuntu-advantage/marker-attached
+ConditionPathExists=/var/lib/ubuntu-advantage/marker-license-check
```
blackboxsw, line 4: Is there any way we can use a systemd.path file with a PathExists=/var/lib/marker-license-check to this timer instead of the ConditionPathExists in a timer definition (which I think is only calculated once at boot)?

blackboxsw (commented):
Thanks for putting this together.couple of thoughts inline about using systemd.path instead of just the timer ConditionPathExists declarations. I'm not certain we can tie Paths to timers though. But it's worth a check. We do use the path constructs for our auto-attach systemd service

uaclient/cli.py
```diff
@@ -922,6 +922,7 @@ def _attach_with_token(
         print(ua_status.MESSAGE_ATTACH_SUCCESS_NO_CONTRACT_NAME)
 
     config.update_ua_messages(cfg)
+    config.write_cache("marker-attached", "")
```
lucasmoura, line 925: If we are going through that route, do we want to bring the service back if the user runs a `detach` operation ? We could remove the `marker-attached`  file on the `action_detach` function, but we would need to manually restart systemd as well, right ?

uaclient/config.py
```diff
@@ -477,10 +484,15 @@ def delete_cache_key(self, key: str) -> None:
...
-        """Remove configuration cached response files class attributes."""
+    def delete_cache(self, delete_permanent: bool = False):
+        """
+        Remove configuration cached response files class attributes.
+
+        :param delete_permanent: even delete the "permanent" files
+        """
         for path_key in self.data_paths.keys():
-            self.delete_cache_key(path_key)
+            if delete_permanent or not self.data_path(path_key).permanent:
```
lucasmoura, line 494: The `self.data_path` call will return a string here, so we cannot call the `permanent` attribute like this

uaclient/util.py
```diff
@@ -801,22 +801,22 @@ def parse_rfc3339_date(dt_str: str) -> datetime.datetime:
 
     :return: datetime.datetime object of time represented by dt_str
     """
+    # remove sub-seconds
+    dt_str_without_subseconds = re.sub(
+        r"(\d{2}:\d{2}:\d{2})\.\d+", r"\g<1>", dt_str
```
lucasmoura, line 806: Can't we just combine  `dt_str_without_subseconds` with `dt_str_with_z` by using:

```python
    dt_str_without_subseconds = re.sub(
        r"(\d{2}:\d{2}:\d{2})\.\d+", r"\g<1>Z", dt_str)
```

Also, I may be missing something, but are we changing the format of the output string now in the end ? If we are, can't this affect already stored datetime strings in the system ?

debian/ubuntu-advantage-tools.postinst
orndorffgrant: Aha I totally got this mixed up. Thanks for this @blackboxsw !!!

I'll play with it to try and get the path file to activate the timer

lib/gcp_auto_attach.py
orndorffgrant: good point!

systemd/ua-license-check.timer
orndorffgrant: Yep and thats the exact problem I ran into at eod yesterday! Thanks for reminding me about the difference between the two approaches

uaclient/cli.py
orndorffgrant: detach is calling `delete_cache` which will delete the `marker-attached` file, but yes I haven't determined if we'll need to prod systemd to notice that yet. We'll see after I change approaches to using a `.path` instead of a ConditionPathExists

uaclient/config.py
orndorffgrant: Oof good catch!

uaclient/util.py
orndorffgrant: > Can't we just combine `dt_str_without_subseconds` with `dt_str_with_z` by using:
> 
> ```python
>     dt_str_without_subseconds = re.sub(
>         r"(\d{2}:\d{2}:\d{2})\.\d+", r"\g<1>Z", dt_str)
> ```
> 

We totally could combine them, but I kept them separate solely for readability. Because regexes are hard enough to come back to 6 months later and read already, I figure that keeping each step with a single purpose would help.


> Also, I may be missing something, but are we changing the format of the output string now in the end ? If we are, can't this affect already stored datetime strings in the system ?

This is just a parsing function, so this won't affect how we're formatting datestrings during serialization - that is done at `util.py:105`. The adjustment to this function just adds support for parsing another variant of a datestring.

Before this change, it was failing to parse a datestring with subseconds, but without a timezone, which is why there had to be some ad hoc date parsing in `timer.py`. But by adding support for this format, we can just rely on DateTimeAwareJsonDecoder to parse dates in this format. I added this format to the unit test cases for this function at `test_util.py:934`

lucasmoura: Okay, good points and I do agree that reading regex is a though call. One thing we can do make that easier in the feature is having some code docs with a string before and after applying the regex.

systemd/ua-license-check.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Ubuntu Advantage Timer to check for GCP Ubuntu Pro licenses
+
+[Timer]
+OnCalendar=*:0/5
+RandomizedDelaySec=5min
+Persistent=true
```
blackboxsw, line 7: We might want to drop the persistent=true. This has a side-effect that if the machine was off for > 5 minutes. It'll be run ASAP on next boot... which will likely occur before OnStartupSec=1min and increase our chances of colliding with a either lock held by ua-auto-attach, or with cloud-init trying to detect the datasouce (for cloud-id).  Also any objection to increasing OnStartupSec to > 2 mins for sort of the same reason. Cut down on risk area where collisions of multiple UA config writers in initial system boot.

blackboxsw (changes requested):
I'd prefer this approach to https://github.com/canonical/ubuntu-advantage-client/pull/1801 because  #1801 obscures the fact that we are a recurring/polling timer with a rather high frequency. I'd rather this timer object be more discoverable via `systemctl list-timers` and treated like the proper timer that it actually is.

lib/license_check.py
```diff
@@ -0,0 +1,16 @@
...
+from uaclient.cli import setup_logging
+from uaclient.config import UAConfig
+from uaclient.jobs.license_check import gcp_auto_attach
+
+LOG = logging.getLogger(__name__)
+
+if __name__ == "__main__":
+    cfg = UAConfig()
+    setup_logging(logging.INFO, logging.DEBUG)
+    gcp_auto_attach(cfg=cfg, logger=LOG)
```
renanrodrigo, line 16: You don't need to pass the logger here, because you are configuring the root logger. When you get a logger inside the job itself, it will follow the configuration from `setup_logging` (unless you are setting up a child logger)

features/license_check.feature
```diff
@@ -0,0 +1,84 @@
...
+        When I run `touch /var/lib/ubuntu-advantage/marker-license-check` with sudo
+        Then I verify the `ua-license-check` systemd timer is scheduled to run within `10` minutes
+        When I run `systemctl start ua-license-check.service` with sudo
+        When I run `sleep 1` with sudo
+        Then I verify the `ua-license-check` systemd timer is disabled
+        Examples: version
+            | release |
+            | xenial  |
+            | bionic  |
+            | focal   |
```
renanrodrigo, line 84: that missing newline

renanrodrigo (commented):
Just a comment about the logger.
I guess it would be useful to have a small talk about how we are going to do the logging from here on, just to align it.

uaclient/jobs/license_check.py
```diff
@@ -4,27 +4,30 @@
...
+        jobs.disable_license_check_if_applicable(cfg)
         return
 
     # If the instance is already attached we will not do anything.
     # This implies that the user may have a new license attached to the
     # instance, but we will not perfom the change through this job.
     if cfg.is_attached:
+        logger.info("Disabling gcp_auto_attach job. Already attached")
+        jobs.disable_license_check_if_applicable(cfg)
         return
```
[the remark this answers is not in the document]
orndorffgrant: I've brought this in on the next commit to be pushed soon

systemd/ua-license-check.timer
orndorffgrant: Done!

lib/license_check.py
orndorffgrant: Aye, I suspected I'd end up configuring a separate logger, but honestly haven't paid attention to the logging details yet as I was waiting for #1789 to become finalized

features/license_check.feature
```diff
@@ -0,0 +1,84 @@
...
+    Scenario Outline: license_check job should run periodically on gcp generic lts
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        # verify its enabled
+        Then I verify the `ua-license-check` systemd timer is scheduled to run within `10` minutes
+        # run it and verify that it didn't disable itself
+        When I run `systemctl start ua-license-check.service` with sudo
+        Then I verify the `ua-license-check` systemd timer is scheduled to run within `10` minutes
+        # verify attach disables it
+        When I attach `contract_token` with sudo
+        When I run `sleep 1` with sudo
```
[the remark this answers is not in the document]
orndorffgrant: I've moved around some of the sleeps as not all were necessary, I'll make sure the ones that are necessary are more than 1 second though

features/license_check.feature
```diff
@@ -0,0 +1,84 @@
...
+            | bionic  |
+            | focal   |
+            | hirsute |
+
+    @wip
+    @series.lts
+    @uses.config.machine_type.aws.pro
+    @uses.config.machine_type.azure.pro
+    @uses.config.machine_type.gcp.pro
+    Scenario Outline: license_check is disabled everywhere but gcp generic
```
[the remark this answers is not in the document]
orndorffgrant: The pro tests require the config setup and ua auto-attach steps to get into the state that we want to test: after auto attach, the timer is disabled.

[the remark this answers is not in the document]
lucasmoura: I was thinking that maybe we could add the config setup to all machines and instead of running `auto-attach` on the PRO types, just running a normal attach. But you are right, it is better to exercise this other path

uaclient/cli.py
orndorffgrant: this is now updated so that attach calls an `disable_license_check` function and detach calls an `enable_license_check` function

uaclient/util.py
orndorffgrant: Good idea! I added some comments with examples

uaclient/jobs/license_check.py
```diff
@@ -4,39 +4,38 @@
 """
 import logging
 
-from uaclient import config, exceptions
+from uaclient import config, exceptions, jobs, util
 from uaclient.cli import action_auto_attach, setup_logging
 from uaclient.clouds.gcp import GCP_LICENSES, UAAutoAttachGCPInstance
 from uaclient.clouds.identity import get_cloud_type
+from uaclient.config import UAConfig
```
renanrodrigo: No need to import this - `config.UAConfig` is already being used for the typing.

uaclient/jobs/license_check.py
```diff
@@ -4,39 +4,38 @@
...
 from uaclient.util import get_platform_info
 
-LOG = logging.getLogger(__name__)
+LOG = logging.getLogger("ua_lib.license_check.jobs.license_check")
```
renanrodrigo, line 14: This is a working approach for the logging. I thought about this for the other timer, and gave up because it would introduce inconsistent behavior for the other jobs. 
Now that we have a dedicated timer, picking up its logger for the job makes more sense. I think this is a good approach.

renanrodrigo: Note to self: The only unnecessary 5-min repeating logs we are suppressing now is from the `uaclient.utils.readurl` method, while checking for the license.

renanrodrigo (changes requested):

uaclient/jobs/license_check.py
renanrodrigo: (shows outdated because of the next commit, but it's still there, I swear!

orndorffgrant: Ah! You're totally right, my bad

uaclient.conf
```diff
@@ -15,5 +15,4 @@ ua_config:
   https_proxy: null
```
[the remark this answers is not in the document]
orndorffgrant: I opted not to add it here just because most people won't care about this setting in particular, so it didnt seem worth adding to the default conf.

Thats not a strong opinion at all though, if you think I should add it, I will!

blackboxsw (commented):
@orndorffgrant looks really good. Thanks for the integration test extensions. A couple of questions for you. Otherwise I'm +1.

uaclient/jobs/license_check.py
blackboxsw: Ack, looks good here for scoping the log handler definitions.

uaclient.conf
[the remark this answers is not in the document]
renanrodrigo: Makes sense - agreed

lib/license_check.py
```diff
@@ -0,0 +1,28 @@
...
+    cfg = UAConfig()
+    # The ua-license-check logger should log everything to its file
+    setup_logging(
+        logging.CRITICAL,
+        logging.DEBUG,
+        log_file=cfg.license_check_log_file,
+        logger=LOG,
+    )
+    # Make sure the ua-license-check logger does not generate double logging
+    LOG.propagate = False
```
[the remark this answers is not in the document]
renanrodrigo: Good point. Yes, I guess we can.
When we are setting up our logging, we have a clear difference between configuring loggers for the specific modules to be "silent", or to make global noise :D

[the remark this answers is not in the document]
orndorffgrant: I think we should probably have a broader logging refactor at some point that should consider this, but am hesitant to include that change in this PR.

sru/release-27.3/gcp_auto_attach_test.sh
```diff
@@ -7,21 +7,19 @@ DISK_NAME="persistent-disk-0"
 set -x
 
 gcloud compute instances create $INSTANCE_NAME \
-    --image="ubuntu-minimal-1604-xenial-v20210430" \
+    --image="ubuntu-1604-xenial-v20210429" \
```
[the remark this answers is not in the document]
orndorffgrant: I'm not sure about the original decision to use minimal, but I changed it while I was adjusting the lines lower based on a comment I remembered from this doc https://docs.google.com/document/d/1gfQp1pe7BlvRTmf6mQOoNeHb7Fzu9LFL84yt4lZECOU/edit#heading=h.u9ntmhnc8ffy

uaclient/jobs/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+def enable_license_check_if_applicable(cfg: config.UAConfig):
+    series = util.get_platform_info()["series"]
+    if "gce" in get_cloud_type() and util.is_lts(series):
+        cfg.write_cache("marker-license-check", "")
+
+
+def disable_license_check_if_applicable(cfg: config.UAConfig):
+    if cfg.cache_key_exists("marker-license-check"):
+        cfg.delete_cache_key("marker-license-check")
+        util.subp(["systemctl", "stop", "ua-license-check.timer"])
```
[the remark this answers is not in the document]
orndorffgrant: @blackboxsw good point!
First, more context on how we got here:

on everything except gcp focal:
```bash
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has started
rm /var/lib/ubuntu-advantage/marker-license-check
systemctl stop ua-license-check.timer
# timer has stopped
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has started
```

on gcp focal (the problem platform):
```bash
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has started
rm /var/lib/ubuntu-advantage/marker-license-check
systemctl stop ua-license-check.timer
# timer has stopped
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has NOT started !!!!
```
To work on gcp focal you need to put the `stop` first. This works:
```bash
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has started
systemctl stop ua-license-check.timer
rm /var/lib/ubuntu-advantage/marker-license-check
# timer has stopped
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has started
```

But of course that doesn't work anywhere else. On other platforms the following happens when you run `stop` first:
```bash
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has started
systemctl stop ua-license-check.timer
rm /var/lib/ubuntu-advantage/marker-license-check
# timer has NOT stopped !!! (because it got immediately restarted before I rm'd the marker file
```

So the following will work on all platforms including gcp focal:
```bash
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has started
systemctl stop ua-license-check.timer
rm /var/lib/ubuntu-advantage/marker-license-check
systemctl stop ua-license-check.timer
# timer has stopped
touch /var/lib/ubuntu-advantage/marker-license-check
# timer has started
```
:tada: 


Now, the wild thing is that the change you suggested is not necessary (!?). I know this because the following snippet from `license_check.feature` works on all platforms including gcp focal:
```gherkin
        # verify attach disables it
        When I run `sleep 5` with sudo
        Then I verify the `ua-license-check` systemd timer is disabled
        # verify detach enables it
        When I run `ua detach --assume-yes` with sudo
```

I'm not totally sure why and I'm open to making the change you suggested just to be safe though. Thoughts?

(I'm also going to rerun all of the integration tests to re-verify all of this)

features/license_check.feature
```diff
@@ -0,0 +1,115 @@
...
+
+    @series.lts
+    @uses.config.machine_type.gcp.generic
+    Scenario Outline: license_check job should run periodically on gcp generic lts
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        # verify its enabled
+        Then I verify the `ua-license-check` systemd timer is scheduled to run within `10` minutes
+        # run it and verify that it didn't disable itself
+        When I run `systemctl start ua-license-check.service` with sudo
+        When I run `sleep 5` with sudo
```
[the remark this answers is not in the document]
orndorffgrant: Aha yeah thats a great idea! I'll add a "when I sleep" step to do the sleep locally.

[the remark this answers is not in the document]
orndorffgrant: And yes its just that sometimes systemd doesn't seem to immediately notice the marker file. A `sleep 1` was good enough when I initially ran all the tests but @lucasmoura suggested increasing it to compensate for potentially overloaded jenkins runs, for example.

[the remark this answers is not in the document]
renanrodrigo: For comprehension, I think `when I wait x seconds` is a better phrase for the step (:

features/license_check.feature
```diff
@@ -0,0 +1,115 @@
+Feature: License check timer only runs in environments where necessary
+
+    @series.lts
+    @uses.config.machine_type.gcp.generic
```
renanrodrigo, line 4: Since we are attaching with a contract token in the non-pro tests, I think we can make explicitly `@uses.config.contract_token` on those tests.

lucasmoura (approved):

renanrodrigo (approved):

debian/rules
```diff
@@ -56,9 +56,13 @@ override_dh_systemd_enable:
 	dh_systemd_enable -pubuntu-advantage-tools ua-reboot-cmds.service
 	dh_systemd_enable -pubuntu-advantage-tools ua-timer.timer
 	dh_systemd_enable -pubuntu-advantage-tools ua-timer.service
+	dh_systemd_enable -pubuntu-advantage-tools ua-license-check.timer
+	dh_systemd_enable -pubuntu-advantage-tools ua-license-check.service
+	dh_systemd_enable -pubuntu-advantage-tools ua-license-check.path
 
 override_dh_systemd_start:
 	dh_systemd_start -pubuntu-advantage-tools ua-timer.timer
+	dh_systemd_start -pubuntu-advantage-tools ua-license-check.path
```
blackboxsw, line 65: I'm uncertain we actually want to start this on package install, I'm seeing this result in starting up the license-check.service for a one time run on non gcp systems. Checking to confirm the absence of this will still allow the systemd path to operate properly on google cloud

orndorffgrant:
Fixed one of the test failures by changing the ua-license-check.service type to `oneshot`. Some others were due to the test token changes in #1805 and will be fixed in that PR (which should probably be merged before this one). The last couple failures appear to be flakes, we will see when jenkins re-runs

debian/rules
blackboxsw: Ok, turns out this symtom was likely due to the fact that I had a stale copy of this branch. @orndorffgrant mentioned at standup that he recently shifted the timer to a oneshot service and reset some of the WantedBy config directives which were causing the license service to run on each boot. I no longer see this behavior on my bionic lxc

blackboxsw (approved):

PR #1801, opened 2021-09-10
Separate gcp timer but without the timer

orndorffgrant:
## Notes
This was supposed to be a Draft PR - please don't merge :)

Alternative to #1799 that uses a service restart every 5 minutes instead of a timer. This might be frowned upon.
The nice(?) thing about it is that the recurring job is controlled entirely by creating/deleting the marker file - no need to call `systemctl stop ua-license-check.timer` like in the timer approach.

Look at [this commit](https://github.com/canonical/ubuntu-advantage-client/pull/1801/commits/a3540b327967ef9774de1c7b948b9b88008c3dde) to see the code diff from #1799

## Proposed Commit Message

>
>
>

## Desired commit type::

## Checklist:


systemd/ua-license-check.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Ubuntu Advantage check for GCP Ubuntu Pro licenses
+After=network.target network-online.target systemd-networkd.service ua-auto-attach.service
```
lucasmoura, line 3: Why do we need to specify all of these extra targets here ? Wouldn't waiting for `ua-auto-attach.service`  enough to run this service ?

blackboxsw: @lucasmoura auto-attach service is only present on PRO images, in the absence of that service or target, we need to fallback to the next earliest time in boot we need to wait for.... which is that network is "up" on the system so we can talk to the IMDS.

blackboxsw (changes requested):
I'd prefer your alternative #1799 because it feels like the Reboot=always RebootSec=5mins is creating a "timer-like" object that is not represented in `systemctl list-timers` and therefore less discoverable. I'll let others weigh in though if they have strong opinions on this.

systemd/ua-license-check.service
lucasmoura: Ohh excelent point.

orndorffgrant:
closing in favor of #1799

PR #1802, opened 2021-09-14
postinst: remove old ua-messaging.timer services (SC-409)

lucasmoura:
## Proposed Commit Message
postinst: remove old ua-messaging.timer services

The postrm script will mask the service `ua-messaging.timer` since it will no longer be delivered by the package. Although this operation will indeed disable the service, we can still see it when running `systemctl list-timers --all`. The reason for that is because the mask operation creates a symlink on `/etc/systemd/system/timer.target.wants/` for this service that points to `/lib/systemd/system/ua-messaging.timer` . We cannot unmask the service, since the unit will not be there anymore, so one way to avoid this service to appear when running `list-timers` is by deleting this symlink and perfoming a `systemctl daemon-reload` operation.

## Test Steps
1. Launch a xenial container
2. Upgrade UA to the daily version
3. Verify that `systemctl list-timers --all`  displays `ua-messaging.timer
4. Launch a new container and upgrade to an UA version if this postinst modifiction
5. Verify that `ua-messaging.timer` no longer appears when running list-timers

## Desired commit type::

## Checklist:

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -32,6 +32,8 @@ ESM_APPS_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-esm-apps.list"
 FIPS_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-fips.list"
 
 OLD_CLIENT_FIPS_PPA="private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu"
+OLD_MESSAGING_TIMER="ua-messaging.timer"
+OLD_MESSAGING_TIMER_MASKED_LOCATION="/etc/systemd/system/timers.target.wants/$OLD_MESSAGING_TIMER"
 
 ESM_APT_PREF_FILE_TRUSTY="$APT_PREFERENCES_DIR/ubuntu-esm-trusty"
 ESM_INFRA_OLD_APT_PREF_FILE_TRUSTY="$APT_PREFERENCES_DIR/ubuntu-esm-infra-trusty"
@@ -316,6 +318,18 @@ enable_periodic_license_check() {
     fi
 }
 
+remove_old_systemd_timers() {
+    # These are the commands that are run when the package is purged.
+    # Since we actually want to remove this service from now on
+    # we have replicated that behavior here
+    if [ -L $OLD_MESSAGING_TIMER_MASKED_LOCATION ]; then
+        if [ -x "/usr/bin/deb-systemd-helper" ]; then
+            deb-systemd-helper purge ua-messaging.timer > /dev/null || true
+            deb-systemd-helper unmask ua-messaging.timer > /dev/null || true
+        fi
+    fi
+}
+
 case "$1" in
     configure)
       PREVIOUS_PKG_VER=$2
@@ -389,8 +403,8 @@ case "$1" in
         fi
       fi
       mark_reboot_for_fips_pro
-
       enable_periodic_license_check
+      remove_old_systemd_timers
       ;;
 esac
 


debian/ubuntu-advantage-tools.postinst
```diff
@@ -302,6 +304,13 @@ notify_wrong_fips_metapackage_on_cloud() {
     fi
 }
 
+remove_old_systemd_timers() {
+    if systemctl list-timers --all | grep -q "$OLD_MESSAGING_TIMER"; then
+        rm -f $OLD_MESSAGING_TIMER_MASKED_LOCATION
+        systemctl daemon-reload
+    fi
+}
```
blackboxsw, line 312: Could we just test for the presence of the OLD_MESSAGING_TIMER_MASKED_LOCATION? Also most post(inst|rm) scripts I see also silence the logs from the reload (in case other systemd units on the system  are malformed, we don't want to generate a lot of logs to the console during our pkg upgrade. We also don't want to be impacted during our package install of a failure to run systemctl daemon-reload. So let's try something like this 
`systemctl --system daemon-reload >/dev/null || true`



```suggestion
    if [ -f  $OLD_MESSAGING_TIMER_MASKED_LOCATION ]; then
        rm $OLD_MESSAGING_TIMER_MASKED_LOCATION
        systemctl --system daemon-reload >/dev/null || true
    fi
}
```

blackboxsw (changes requested):
Thanks @lucasmoura I think we can trim this a bit, add a little flexibility in the case of errors on `daemon-reload` and avoid the extra call to systemctl list-timers --all.

debian/ubuntu-advantage-tools.postinst
orndorffgrant: That does appear to be a popular pattern +1 https://codesearch.debian.net/search?q=systemctl+--system+daemon-reload+%3E%2Fdev%2Fnull+%7C%7C+true+path%3Adebian%2F&literal=1&page=1

orndorffgrant: Actually I think debhelper has what we need, its in the postrm of the last version of uaclient:
```
# Automatically added by dh_systemd_enable
if [ "$1" = "remove" ]; then
        if [ -x "/usr/bin/deb-systemd-helper" ]; then
                deb-systemd-helper mask ua-messaging.timer >/dev/null
        fi
fi

if [ "$1" = "purge" ]; then
        if [ -x "/usr/bin/deb-systemd-helper" ]; then
                deb-systemd-helper purge ua-messaging.timer >/dev/null
                deb-systemd-helper unmask ua-messaging.timer >/dev/null
        fi
fi
# End automatically added section
```

I think we need to run the "purge" section here.

lucasmoura: Indeed we have that, the problem is that we will only have that part of the code running if the user manually runs an `apt purge ubuntu-advantage-tools`

I don't think this is called during the upgrade process of the package

orndorffgrant:
@lucasmoura this looks good and is working for me!

I think we should probably include the 
```

debian/ubuntu-advantage-tools.postinst
orndorffgrant: Also fun fact: `systemctl --system daemon-reload >/dev/null || true` is already added to our postinst by debhelper, so we shouldn't have to call it ever explicitly

lucasmoura:
@orndorffgrant Yes, that's a good idea, I can check for the existence of the symlink before trying to remove the timer.
I will update the code with both suggestions

renanrodrigo (approved):
looks good, +1 w/ Grant's change

lucasmoura:

orndorffgrant (approved):
@lucasmoura This looks good and works for me!

Based on our discussion earlier today, I still think this is appropriate. I can't think of any user that would want to customize this or need to keep this around after upgrade. If anything, it might look sloppy on our part to leave it behind. So I think this is a good move.

Not sure what is up with the jenkins tests though

lucasmoura:
Could not reproduce the test failure locally

orndorffgrant:
I'm also not able to reproduce that test failure

debian/ubuntu-advantage-tools.postinst
blackboxsw: This looks really good. Thank you for fleshing this out and coming up with a better solution for this rename. +1, if we think we want to got ahead and "fix" this issue of stale/masked timers across upgrade. Doesn't break downgrade compatibility as older versions will re-create the timer/service.

blackboxsw (approved):
LGTM. Thanks for coming up with a cleaner approach for dropping the old timer/service.

rebased and resolved conflicts.

orndorffgrant:
The single jenkins failure looks like a flake and it passed locally. Everyone has approved so I'm going to go ahead and merge :partying_face:

Issue #463, opened 2019-05-03
`ua` command to ease collection of debug data relating to the client


OddBloke:
If/when a user ends up needing support or hitting a bug in the client, it would make both the development team's and the support team's lives easier if there were a clear mechanism to collect data for debugging from an affected system.

`cloud-init` has the `collect-logs` subcommand, and I think `ua collect-logs` would make sense too.

BryanQuigley:
Either way +1.  I read it two different ways:

1.  A defined way to have ua send out relevant logs for debugging the client itself.  (Sosreport would that gain the ability to run this command)
2.  A tool to help the user run sosreport to collect information before sending it to support. RH's support tool has the ability to open a case with support and upload a sosreport/specific logs to it.

If we want to do both, the only thing to consider is how to name them so they do the obvious variant.

OddBloke:
Thanks for that comment Bryan; I've modified the title and description to make it clear that this is your "1.".

PR #1803, opened 2021-09-14
cli: add collect-logs command (SC-407)

renanrodrigo:
Add the collect-logs command.
The command picks up all relevant data for debugging possible problems in UA, and creates a tarball with it.

Fixes: #463

Signed-off-by: Renan Rodrigo <renanrodrigo@canonical.com>

## Desired commit type::

## Checklist:

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -559,3 +559,49 @@ Feature: Command behaviour when attached to an UA subscription
            | xenial  |
            | bionic  |
            | focal   |
+
+    @series.lts
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Run collect-logs on an attached machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
+        And I verify that running `ua collect-logs` `as non-root` exits `1`
+        Then I will see the following on stderr:
+             """
+             This command must be run as root (try using sudo).
+             """
+        When I run `ua collect-logs` with sudo
+        Then I verify that files exist matching `ua_logs.tar.gz`
+        When I run `tar zxf ua_logs.tar.gz` as non-root
+        Then I verify that files exist matching `logs/`
+        When I run `sh -c "ls -1 logs/ | sort -d"` as non-root
+        # On Xenial, the return value for inexistent services is the same as for dead ones (3).
+        # So the -error suffix does not appear there.
+        Then stdout matches regexp:
+        """
+        cloud-id.txt
+        jobs-status.json
+        journalctl.txt
+        livepatch-status.txt-error
+        systemd-timers.txt
+        ua-auto-attach.path.txt(-error)?
+        ua-auto-attach.service.txt(-error)?
+        uaclient.conf
+        ua-license-check.path.txt
+        ua-license-check.service.txt
+        ua-license-check.timer.txt
+        ua-reboot-cmds.service.txt
+        ua-status.json
+        ua-timer.service.txt
+        ua-timer.timer.txt
+        ubuntu-advantage.log
+        ubuntu-advantage-timer.log
+        ubuntu-esm-apps.list
+        ubuntu-esm-infra.list
+        """
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -8,7 +8,10 @@ import logging
 import os
 import pathlib
 import re
+import shutil
 import sys
+import tarfile
+import tempfile
 import textwrap
 import time
 from functools import wraps
@@ -41,6 +44,18 @@ DEFAULT_LOG_FORMAT = (
 
 STATUS_FORMATS = ["tabular", "json", "yaml"]
 
+UA_COLLECT_LOGS_FILE = "ua_logs.tar.gz"
+
+UA_SERVICES = (
+    "ua-timer.service",
+    "ua-timer.timer",
+    "ua-auto-attach.path",
+    "ua-auto-attach.service",
+    "ua-reboot-cmds.service",
+    "ua-license-check.path",
+    "ua-license-check.service",
+    "ua-license-check.timer",
+)
 
 # Set a module-level callable here so we don't have to reinstantiate
 # UAConfig in order to determine dynamic data_path exception handling of
@@ -189,6 +204,24 @@ def auto_attach_parser(parser):
     )
     parser.usage = USAGE_TMPL.format(name=NAME, command=parser.prog)
     parser._optionals.title = "Flags"
+    return parser
+
+
+def collect_logs_parser(parser):
+    """Build or extend an arg parser for 'collect-logs' subcommand."""
+    parser.prog = "collect-logs"
+    parser.description = (
+        "Collect UA logs and relevant system information into a tarball."
+    )
+    parser.usage = USAGE_TMPL.format(name=NAME, command=parser.prog)
+    parser.add_argument(
+        "-o",
+        "--output",
+        help=(
+            "tarball where the logs will be stored. (Defaults to "
+            "./ua_logs.tar.gz)"
+        ),
+    )
     return parser
 
 
@@ -1010,6 +1043,81 @@ def action_attach(args, *, cfg):
     )
 
 
+def _write_command_output_to_file(
+    cmd, filename: str, return_codes: List[int] = None
+) -> None:
+    """Helper which runs a command and writes output or error to filename."""
+    try:
+        out, _ = util.subp(cmd.split(), rcs=return_codes)
+    except util.ProcessExecutionError as e:
+        util.write_file("{}-error".format(filename), str(e))
+    else:
+        util.write_file(filename, out)
+
+
+# We have to assert root here, because the logs are not non-root user readable
+@assert_root
+def action_collect_logs(args, *, cfg: config.UAConfig):
+    output_file = args.output or UA_COLLECT_LOGS_FILE
+
+    with tempfile.TemporaryDirectory() as output_dir:
+
+        _write_command_output_to_file(
+            "cloud-id", "{}/cloud-id.txt".format(output_dir)
+        )
+        _write_command_output_to_file(
+            "ua status --format json", "{}/ua-status.json".format(output_dir)
+        )
+        _write_command_output_to_file(
+            "canonical-livepatch status",
+            "{}/livepatch-status.txt".format(output_dir),
+        )
+        _write_command_output_to_file(
+            "systemctl list-timers --all",
+            "{}/systemd-timers.txt".format(output_dir),
+        )
+        _write_command_output_to_file(
+            (
+                "journalctl --boot=0 -o short-precise "
+                "{} "
+                "-u cloud-init-local.service "
+                "-u cloud-init-config.service -u cloud-config.service"
+            ).format(
+                " ".join(
+                    ["-u {}".format(s) for s in UA_SERVICES if ".service" in s]
+                )
+            ),
+            "{}/journalctl.txt".format(output_dir),
+        )
+
+        for service in UA_SERVICES:
+            _write_command_output_to_file(
+                "systemctl status {}".format(service),
+                "{}/{}.txt".format(output_dir, service),
+                return_codes=[0, 3],
+            )
+
+        ua_logs = (
+            cfg.cfg_path or "/etc/ubuntu-advantage/uaclient.conf",
+            cfg.log_file,
+            cfg.timer_log_file,
+            cfg.license_check_log_file,
+            cfg.data_path("jobs-status"),
+            *(
+                entitlement.repo_list_file_tmpl.format(name=entitlement.name)
+                for entitlement in entitlements.ENTITLEMENT_CLASSES
+                if issubclass(entitlement, entitlements.repo.RepoEntitlement)
+            ),
+        )
+
+        for log in ua_logs:
+            if os.path.isfile(log):
+                shutil.copy(log, output_dir)
+
+        with tarfile.open(output_file, "w:gz") as results:
+            results.add(output_dir, arcname="logs/")
+
+
 def get_parser():
     service_line_tmpl = " - {name}: {description}{url}"
     base_desc = __doc__
@@ -1075,6 +1183,12 @@ def get_parser():
     )
     auto_attach_parser(parser_auto_attach)
     parser_auto_attach.set_defaults(action=action_auto_attach)
+
+    parser_collect_logs = subparsers.add_parser(
+        "collect-logs", help="collect UA logs and debug information"
+    )
+    collect_logs_parser(parser_collect_logs)
+    parser_collect_logs.set_defaults(action=action_collect_logs)
 
     parser_config = subparsers.add_parser(
         "config", help="manage Ubuntu Advantage configuration on this machine"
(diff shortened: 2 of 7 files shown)


features/attached_commands.feature
```diff
@@ -584,3 +584,44 @@ Feature: Command behaviour when attached to an UA subscription
            | xenial  |
            | bionic  |
            | focal   |
+
+
+    @wip
```
lucasmoura, line 589: I think we can remove that

features/attached_commands.feature
```diff
@@ -584,3 +584,44 @@ Feature: Command behaviour when attached to an UA subscription
...
+        journalctl.txt
+        livepatch-status.txt
+        systemd-timers.txt
+        ua-auto-attach.path.txt
+        ua-auto-attach.service.txt
+        ua-reboot-cmds.service.txt
+        ua-status.json
+        ua-timer.service.txt
+        ua-timer.timer.txt
+        ubuntu-advantage.log
```
lucasmoura, line 620: Why are the `ubuntu-advantage-timer.log` not included ?

uaclient/cli.py
```diff
@@ -1004,6 +1035,68 @@ def action_attach(args, *, cfg):
...
+    output_dir = "/tmp/ua-logs-{}".format(current_datetime)
+    if os.path.exists(output_dir):
+        shutil.rmtree(output_dir)
+    os.makedirs(output_dir)
+
+    _write_command_output_to_file(
+        "cloud-id", "{}/cloud-id.txt".format(output_dir)
+    )
+    _write_command_output_to_file(
+        "ua status --format json", "{}/ua-status.json".format(output_dir)
```
lucasmoura, line 1070: Let's double check to see if there is no private information here

uaclient/cli.py
```diff
@@ -1004,6 +1035,68 @@ def action_attach(args, *, cfg):
...
+    if os.path.isfile(path):
+        shutil.copy(path, out_dir)
+
+
+# We have to assert root here, because the logs are not non-root user readable
+@assert_root
+def action_collect_logs(args, *, cfg):
+    current_datetime = datetime.now().strftime(format="%b-%d-%H%M")
+
+    output_file = args.output or "ua_logs.tar.gz"
```
lucasmoura, line 1060: I think the `ua_logs` named should be stored in a constant

uaclient/cli.py
```diff
@@ -41,6 +44,21 @@
 
 STATUS_FORMATS = ["tabular", "json", "yaml"]
 
+UA_LOGS = (
```
lucasmoura, line 47: Most of these paths are already stored on `UAConfig`. Furthermore, some of these paths can be overridden in the config. So we should rely on the `UAConfig` object to obtain those paths

renanrodrigo: Good, well noted. Will change it.

uaclient/cli.py
renanrodrigo: Will do, and post the thoughts here

features/attached_commands.feature
renanrodrigo: aaaah I started on another branch and without the logging PR it wasn't there yet. Will rebase and include it.

uaclient/cli.py
```diff
@@ -41,6 +44,21 @@
...
+    "/var/log/ubuntu-advantage.log",
+    "/var/log/ubuntu-advantage-timer.log",
+    "/var/lib/ubuntu-advantage/jobs-status.json",
+    "/etc/apt/sources.list.d/ubuntu*",
+    "/var/lib/ubuntu-advantage/private/machine-id",
+)
+
+UA_SERVICES = (
+    "ua-timer.service",
```
blackboxsw, line 56: Remember if you land this after #1799 you'll also need the following:

ua-license-check.path
ua-license-check.service
ua-license-check.timer

uaclient/cli.py
blackboxsw: I think we also want /etc/ubuntu-advantage/uaclient.conf too.

blackboxsw (changes requested):
Looks good @renanrodrigo, a couple of minor notes and I think this works. Please add docs as you alluded to, as well as pulling in uaclient.conf. At some point we might have to start specializing uaclient.cli into separate modules as this module size is getting a bit big at 1500 lines which eventually leads to tougher maintenance of an overly generic module.

blackboxsw (commented):
Also can you ammend your commit message to `Fixes: #463` as the commit itself says "Fix  #463" which won't get picked up by our log2dch tooling.

uaclient/cli.py
renanrodrigo: Perfect, thanks for the explanation @blackboxsw

uaclient/cli.py
renanrodrigo: Can include now

features/attached_commands.feature
renanrodrigo: Added also the license-check log, expecting the GCP separate timer PR

renanrodrigo:
@blackboxsw @lucasmoura @orndorffgrant ready for the next round of reviews. 

Tests will not pass now, as they are implemented expecting #1799 to land before this. After we have that PR, I will rebase and make sure everything is ok.

uaclient/cli.py
```diff
@@ -1009,6 +1037,77 @@ def action_attach(args, *, cfg):
...
+# We have to assert root here, because the logs are not non-root user readable
+@assert_root
+def action_collect_logs(args, *, cfg: config.UAConfig):
+    current_datetime = datetime.now().strftime("%b-%d-%H%M")
+
+    output_file = args.output or UA_COLLECT_LOGS_FILE
+    output_dir = "/tmp/ua-logs-{}".format(current_datetime)
+    if os.path.exists(output_dir):
+        shutil.rmtree(output_dir)
+    os.makedirs(output_dir)
```
lucasmoura, line 1066: I think it is better to use a `try..except` block here

uaclient/cli.py
```diff
@@ -1009,6 +1037,77 @@ def action_attach(args, *, cfg):
...
+
+    for service in UA_SERVICES:
+        _write_command_output_to_file(
+            "systemctl status {}".format(service),
+            "{}/{}.txt".format(output_dir, service),
+            return_codes=[0, 3],
+        )
+
+    ua_logs = (
+        "/etc/ubuntu-advantage/uaclient.conf",
```
lucasmoura, line 1096: This file can also be overridden by the user. For example, users can use a local config file to drive UA behavior

uaclient/cli.py
```diff
@@ -1009,6 +1037,77 @@ def action_attach(args, *, cfg):
...
+        "journalctl --boot=0 -o short-precise",
+        "{}/journalctl.txt".format(output_dir),
+    )
+    _write_command_output_to_file("dmesg", "{}/dmesg.txt".format(output_dir))
+
+    for service in UA_SERVICES:
+        _write_command_output_to_file(
+            "systemctl status {}".format(service),
+            "{}/{}.txt".format(output_dir, service),
+            return_codes=[0, 3],
```
lucasmoura, line 1092: Why do we need to specify these return codes here ?

renanrodrigo: `systemctl status` returns 0 only on active running services. If a service is inactive (dead), the return code is 3. We still want to get the information even in this case.

uaclient/cli.py
```diff
@@ -1009,6 +1037,75 @@ def action_attach(args, *, cfg):
...
+                return_codes=[0, 3],
+            )
+
+        ua_logs = (
+            cfg.cfg_path,
+            cfg.log_file,
+            cfg.timer_log_file,
+            cfg.license_check_log_file,
+            cfg.data_path("jobs-status"),
+            "/etc/apt/sources.list.d/ubuntu*",
```
lucasmoura, line 1099: Also, this one check here is a little bit generic.
We could potentially take more files than needed (which is not a problem per se)
@blackboxsw @orndorffgrant do you think we should only collect logs for the `RepoEntitlements` entries ?

features/attached_commands.feature
```diff
@@ -576,3 +576,45 @@ Feature: Command behaviour when attached to an UA subscription
            | xenial  |
            | bionic  |
            | focal   |
+
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Run collect-logs on an attached machine
```
lucasmoura, line 583: Can we add that test on an unattached machine ?

uaclient/cli.py
renanrodrigo: I see the point,
Please let me know if this may be problematic, so we can check for specific files.

renanrodrigo: Done - checking only for the specific `RepoEntitlements` we have

renanrodrigo:
@lucasmoura code updated.
I will fix potential errors in unit/integration tests when rebasing.

uaclient/cli.py
```diff
@@ -41,6 +44,29 @@
 
 STATUS_FORMATS = ["tabular", "json", "yaml"]
 
+UA_COLLECT_LOGS_FILE = "ua_logs.tar.gz"
+
+UA_REPO_ENTITLEMENTS = (
```
lucasmoura, line 49: This list should be dynamically generated. We can a new variable on the `__init__`  module of `entitlements`

uaclient/tests/test_cli_collect_logs.py
```diff
@@ -0,0 +1,107 @@
...
+
+from uaclient.cli import (
+    action_collect_logs,
+    collect_logs_parser,
+    get_parser,
+    main,
+)
+from uaclient.exceptions import NonRootUserError
+
+M_PATH = "uaclient.cli."
```
orndorffgrant, line 14: I'm not sure if the `M_PATH` use is necessary for any of the mocks here

orndorffgrant (commented):
Code looks good! Just a couple comments and haven't actually run it yet

uaclient/cli.py
```diff
@@ -192,6 +207,19 @@ def auto_attach_parser(parser):
...
+def collect_logs_parser(parser):
+    """Build or extend an arg parser for 'collect-logs' subcommand."""
+    parser.prog = "collect-logs"
+    parser.description = (
+        "Collect UA logs and relevant system information into a tarball."
+    )
+    parser.usage = USAGE_TMPL.format(name=NAME, command=parser.prog)
+    parser.add_argument(
+        "-o", "--output", help="tarball where the logs will be stored."
+    )
```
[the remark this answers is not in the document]
renanrodrigo: Yes we can

uaclient/cli.py
```diff
@@ -1009,6 +1037,79 @@ def action_attach(args, *, cfg):
...
+            "systemctl list-timers --all",
+            "{}/systemd-timers.txt".format(output_dir),
+        )
+        _write_command_output_to_file(
+            "journalctl --boot=0 -o short-precise",
+            "{}/journalctl.txt".format(output_dir),
+        )
+        _write_command_output_to_file(
+            "dmesg", "{}/dmesg.txt".format(output_dir)
+        )
```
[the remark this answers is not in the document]
renanrodrigo: @blackboxsw ?

uaclient/cli.py
```diff
@@ -1009,6 +1037,79 @@ def action_attach(args, *, cfg):
     )
 
 
+def _write_command_output_to_file(
+    cmd, filename: str, return_codes: List[int] = None
+) -> None:
+    """Helper which runs a command and writes output or error to filename."""
+    try:
+        out, _ = util.subp(cmd.split(), rcs=return_codes)
```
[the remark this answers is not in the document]
renanrodrigo: It is, our subp is passing it directly to `Popen` and it should be a list afaics

uaclient/config.py
```diff
@@ -304,11 +304,11 @@ def remove_notice(self, label_regex: str, descr_regex: str):
             util.remove_file(self.data_path("notices"))
 
     @property
-    def log_file(self):
+    def log_file(self) -> str:
```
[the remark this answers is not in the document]
renanrodrigo: It punched me in the face

uaclient/tests/test_cli_collect_logs.py
renanrodrigo: Well I think it is always good to mock stuff where it is imported, right?

uaclient/cli.py
```diff
@@ -1009,6 +1037,79 @@ def action_attach(args, *, cfg):
...
+    except util.ProcessExecutionError as e:
+        util.write_file("{}-error".format(filename), str(e))
+    else:
+        util.write_file(filename, out)
+
+
+def _collect_file(path: str, out_dir: str) -> None:
+    if os.path.isfile(path):
+        shutil.copy(path, out_dir)
```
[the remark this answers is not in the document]
renanrodrigo: Yes we can

uaclient/cli.py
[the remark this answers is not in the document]
orndorffgrant: You're right, I realized that I've gotten away with just passing a string to Popen in the past, but it turns out that only works with `shell=True` (and even then only because of an implementation detail really). Also our `bytes_args` line would totally mess up a string thats passed int

uaclient/cli.py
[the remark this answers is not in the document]
blackboxsw: Good point @orndorffgrant. Let's go with minimal info at the moment. So, how about we drop the  dmesg call and also limit the scope of the journalctl  call to just our known units/services/timers.
`journalctl --boot=0 -o short-precise -u ua-timer.service -u ua-reboot-cmds.service -u cloud-init-local.service -u cloud-init-config.service -u cloud-config.service`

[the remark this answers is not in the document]
blackboxsw: We can build up if we find there are gaps besides this.

blackboxsw (approved):
I'm good with this as-is, I think as well we can drop the dmesg and maybe limit the journalctl to only the units we care about.

uaclient/cli.py
```diff
@@ -1009,6 +1037,79 @@ def action_attach(args, *, cfg):
...
+        ua_logs = (
+            cfg.cfg_path or "/etc/ubuntu-advantage/uaclient.conf",
+            cfg.log_file,
+            cfg.timer_log_file,
+            cfg.license_check_log_file,
+            cfg.data_path("jobs-status"),
+            *(
+                entitlement.repo_list_file_tmpl.format(name=entitlement.name)
+                for entitlement in entitlements.ENTITLEMENT_CLASSES
+                if issubclass(entitlement, entitlements.repo.RepoEntitlement)
```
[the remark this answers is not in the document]
renanrodrigo: I would leave those out for now, and include them in the future if we happen to need them

renanrodrigo:
So, I ended up:
- Splitting the Attached version of the test to match the order and files it generates on Xenial
- Dropped the Unatached Xenial version.

Plase let me know if I should write a separate Unattached Xenial test as well, but I think it is not needed.

uaclient/cli.py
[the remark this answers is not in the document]
orndorffgrant: Works for me!

orndorffgrant (approved):
LGTM Lets bring it in today if jenkins is happy

renanrodrigo:
@orndorffgrant Bad Bad Jenkins has 3 timeouts in the VM test already :|

orndorffgrant (approved):
Thank you for the integration test update!
Since that was the only change from the last version that jenkins ran, and all the failures on jenkins from that run passed locally, I am merging now.

PR #1804, opened 2021-09-14
fix: update return code

lucasmoura:
## Proposed Commit Message
fix: update return code

Currently, ua fix will always return 0 no matter what happens during the execution of the command. We are updating the command to only return 0 if the system ends up in a non-vulnerable state. Meaning that either the CVE/USN does not affect the system, the packages required by the fix are already installed or that we have successfully installed all of the packages necessary for the fix.

## Test Steps
Run the modified integration/unit tests

## Desired commit type::

## Checklist:

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -385,7 +385,8 @@ def action_fix(args, *, cfg, **kwargs):
         ).format(args.security_issue)
         raise exceptions.UserFacingError(msg)
 
-    return security.fix_security_issue_id(cfg, args.security_issue)
+    fix_status = security.fix_security_issue_id(cfg, args.security_issue)
+    return fix_status.value
 
 
 def detach_parser(parser):
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -385,8 +385,7 @@ def action_fix(args, *, cfg, **kwargs):
         ).format(args.security_issue)
         raise exceptions.UserFacingError(msg)
 
-    security.fix_security_issue_id(cfg, args.security_issue)
-    return 0
+    return security.fix_security_issue_id(cfg, args.security_issue)
 
 
 def detach_parser(parser):
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -1,4 +1,5 @@
 import copy
+import enum
 import itertools
 import os
 import socket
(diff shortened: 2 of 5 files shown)


renanrodrigo (approved):
lgtm

uaclient/security.py
```diff
@@ -943,10 +946,12 @@ def prompt_for_affected_packages(
                     )
                 )
             )
+            return False
```
orndorffgrant, line 949: I wasn't a part of the initial discussion about this feature, so maybe this was already discussed: I think we should double check that we want to return False in this case.

My first intuition would be that this would be considered success: `ua fix` succesfully installed the packages to fix the vulnerability. The only difference is that the user also needs to restart before the new binaries/libs are loaded (or whatever reason the restart is necessary). 

IOW: Technically the system is not in a non-vulnerable state because the reboot is required, but ua fix did everything that it can to solve the problem successfully, so I would expect ua to return success.

I'm open to opinions otherwise ofc, and maybe we need to bring up in the foundations meeting tomorrow

lucasmoura: Yes, I think this is a good point to bring for the foundations meeting. My rationale behind that is exactly what you pointed out, that the system is still in a non-vulnerable state. Another issue is that if we have a package that needs reboot prior to running `ua fix` we would say that a reboot is needed and return False here in a situation that maybe the system is non-vulnerable.

So yes, let bring it up on the meeting

orndorffgrant: Result of the meeting: Let's return `2` in this case, since the system is not definitely in a nonvulnerable state, but we want to differentiate it from a normal failure (where we fail to install the packages for some reason)

orndorffgrant (changes requested):
Thanks for the update with the third return code. This looks like it works well! Just two requests:

1. Can we have an integration test for returning `2`?
2. I'd prefer to keep all the responsibility of CLI behavior in `cli.py` and let `security.py` have a pythonic interface. In this case, that would mean `security.py` would either return a result enum or throw different types of errors. Then `cli.py` would interpret either the enum or the error and translate that to the correct exit code. Thoughts?

lucasmoura:
@orndorffgrant 

1. I think I may have a USN that triggers a reboot. I will double test that, but if not I can still touch the `reboot-required`  file manually and run the command. So yes, I will add an integration test for it.
2. Yes, I agree we can make this better. I think I will go with the enum solution, even though the exceptions will work as well. My only concern is that they are not exception per se, because they are valid states that the machine can be in. But if you disagree, I can use the exception here

orndorffgrant:
@lucasmoura Yes that all makes sense to me - thank you!

lucasmoura:
@orndorffgrant code updated

uaclient/cli.py
```diff
@@ -385,8 +385,8 @@ def action_fix(args, *, cfg, **kwargs):
         ).format(args.security_issue)
         raise exceptions.UserFacingError(msg)
 
-    security.fix_security_issue_id(cfg, args.security_issue)
-    return 0
+    fix_status = security.fix_security_issue_id(cfg, args.security_issue)
+    return int(fix_status.value)
```
orndorffgrant, line 389: Curious, why is `int() needed?

orndorffgrant (approved):
LGTM! Thanks for the updates

Just one python curiosity question

uaclient/cli.py
lucasmoura: We shouldn't need this actually, I am just ensuring that if for some reason the enum value returns a string, it will be converted. But maybe this is unnecessary. Let me know what you think and I can remove it if needed

orndorffgrant: I don't think the `int()` is necessary but it doesn't hurt either. Up to you, but if you want to try removing it then as long as the integration tests added in this PR still pass, I don't think we need to wait for jenkins again as the current results look good.

lucasmoura: I have removed the `int()` part

PR #1805, opened 2021-09-15
ros: unhide ros entitlements

orndorffgrant:
## Notes

There is movement on the ticket to update the ros infrastructure. We need to be ready to test on staging and give the go ahead for moving to production and then release this asap.

**Note that now that the ros tests are using the production contract server, we need to set UACLIENT_BEHAVE_CONTRACT_TOKEN to a token that entitles esm-apps and ros. I am updating jenkins to use such a token.**

## Proposed Commit Message

```
ros: unhide ros entitlements
```

## Test Steps

Enable ros: verify it auto enables esm infra and apps

enable ros-updates: verify it auto enables esm infra and apps and ros

## Desired commit type::

## Checklist:


renanrodrigo:
When enabling ROS, we could also update the README / manpage, so docs are up-to-date. I think this PR would be a good place to do so.

orndorffgrant:
@renanrodrigo absolutely! good catch

orndorffgrant:
ROS is working! This is ready for review

lucasmoura (approved):

uaclient/tests/test_cli_status.py
```diff
@@ -75,8 +74,7 @@
 Technical support level: n/a
 """
 
-# TODO-ROS add "ros" and "ros-updates" to BETA_SVC_NAMES
-BETA_SVC_NAMES = ["cc-eal", "esm-apps"]
+BETA_SVC_NAMES = ["cc-eal", "esm-apps", "ros", "ros-updates"]
```
blackboxsw, line 77: Why are we claiming ros and ros-updates are beta? Wouldn't this be the time we drop the is_beta = True definition and just make this a fully participating service?

blackboxsw (changes requested):
@orndorffgrant don't we actually want to drop the is_beta definition on ROS and ROS-updates in this branch? I *think* it looks like the service and applicable debs can be installed without error right?

orndorffgrant:
@blackboxsw We could go straight from not-existing -> non-beta if it weren't for the fact that ros requires esm-apps to be enabled, which is still beta. I think it would be un-intuitive for a non-beta service to require enabling a beta service. I'm open to arguments otherwise though

blackboxsw:
> @blackboxsw We could go straight from not-existing -> non-beta if it weren't for the fact that ros requires esm-apps to be enabled, which is still beta. I think it would be un-intuitive for a non-beta service to require enabling a beta service. I'm open to arguments otherwise though

Thanks, forgot about that. Agreed that it would be a surprise to have a beta service enabled when enabling ROS/ROS-updates. We'll sort expectations with Lech on this given other esm-apps discussions that are in flight.

blackboxsw (approved):
+1 Release the beta hounds

orndorffgrant:
Ha ha! :dog2: :dog2: 

Will first fix the couple minor test failures around pro contracts not entitling ros, then I'll go ahead and merge

renanrodrigo (approved):
late lgtm is lgtm

orndorffgrant:
There are some test failures as a result of switching tokens, addressing now

lucasmoura (approved):

orndorffgrant:
The final set of CI tests passed on my machine. Merging now

PR #1807, opened 2021-09-20
cc-eal: make it installable via UA (SC-398)

renanrodrigo:
This PR enables Common Criteria to be installed via `ua`.
- Add the gpg key to verify the repository signature
- Set the install mode to non-interactive, so the `ubuntu-commoncriteria` package is installed without errors

Fixes: #1809

## Test Steps
- Run the modified integration / unit tests
- Pick a 16.04 container, install the ua package from this branch, attach it and enable `cc-eal` with the `--beta` flag

## Desired commit type::

## Checklist:


renanrodrigo:
Remember that the scripts are installed no matter what, but the Ubuntu version for them to work needs to be 16.04.4.
As raised before in a conversation about it, we should verify that this is the installed version before even enabling cc-eal. This PR does not intend to solve this. A future PR, removing the `--beta` flag from the entitlement, can bring this preflight check.

features/attached_enable.feature
```diff
@@ -7,6 +7,28 @@ Feature: Enable command behaviour when attached to an UA subscription
...
+        And I will see the following on stderr:
+            """
+            This command must be run as root (try using sudo).
+            """
+        When I run `ua enable cc-eal --beta` with sudo
+        Then I will see the following on stdout:
+            """
+            One moment, checking your subscription first
+            Updating package lists
+            (This will download more than 500MB of packages, so may take some time.)
```
lucasmoura, line 19: Let's see if that download will behave properly in the Jenkins node. If not we might turn this into a manual test

lucasmoura (commented):
@renanrodrigo How was the GPG key generated ?

features/attached_enable.feature
renanrodrigo: Agreed

renanrodrigo:
@lucasmoura I just downloaded the public key and exported it

features/attached_enable.feature
renanrodrigo: Besides that, for the future, it will be useful to have the manual test so we can check for the Ubuntu version without messing the testing environment too much

lucasmoura:
I think it is better for us to handle key importing in a xenial machine because of gpg versions that we may not be aware of.
Just to be safe here

renanrodrigo:
@lucasmoura sorry for my half words: I just downloaded the public key and exported it, using the Xenial machine where I ran the first tests

orndorffgrant:
@renanrodrigo @lucasmoura Interesting, our refresh-keyrings script seems to already have support for cc-eal. Can you @renanrodrigo verify that it downloads the same key and names it appropriately (it also needs to be run on xenial of course)?

renanrodrigo:
@orndorffgrant yes, it is the same key, the name is right as well

orndorffgrant:
@renanrodrigo when I run `refresh-keyrings.sh` I get a file called `ubuntu-advantage-cc-eal-xenial.gpg` that is different from the keyring uploaded here :/

This entitlement's keyring name should probably match the others that all have the pattern: "ubuntu-advantage-{name}.gpg"

renanrodrigo:
Yes, @orndorffgrant, you are right - the file name has the suffix. But in this case, I think the `refresh-keyrings` script is wrong. Will fix it just in case.

orndorffgrant:
Yes sorry for not communicating clearly at all - I'm hoping for `refresh-keyrings.sh` to work for this key as well as all the others, which looks like it requires updating `refresh-keyrings.sh` to make sure it downloads the correct key and names it appropriately. The name of the key should be `ubuntu-advantage-cc-eal.gpg` to match the others.

Related, it looks like esm-apps is not included in that script - I will create a separate bug for that

renanrodrigo:
@orndorffgrant script updated, name fixed

tools/refresh-keyrings.sh
```diff
@@ -23,7 +23,8 @@ fi
 TARGET_DIR="$1"
 
 EAL_KEY_ID="9F912DADD99EE1CC6BFFFF243A186E733F491C46"
-ESM_KEY_ID="56F7650A24C9E9ECF87C4D8D4067E40313CB4B13"
+ESM_INFRA_KEY_ID="56F7650A24C9E9ECF87C4D8D4067E40313CB4B13"
+ESM_APPS_KEY_ID="E8A443CE358113D187BEE0E6AB01A101DB53907B"
```
orndorffgrant, line 27: Thanks for throwing this in here too :)

orndorffgrant:
> Remember that the scripts are installed no matter what, but the Ubuntu version for them to work needs to be 16.04.4.
> As raised before, in a conversation about this, we should verify that this is the installed version before even enabling cc-eal. 

FYSA: I created a separate Jira card for this: https://warthogs.atlassian.net/browse/SC-450

features/attached_enable.feature
renanrodrigo: Tests passed fine

PR #1811, opened 2021-09-21
fips: allow enabling fips on GCP bionic instances

lucasmoura:
## Proposed Commit Message
fips: allow enabling fips on GCP bionic instances

GCP has now a FIPS optimized kernel on bionic. We are now updating
the code to allow enabling the service on those types of instances

## Test Steps

PS: FIPS Updates test is not passing at the moment because of openssh package releated issues that we still need to debug.

## Desired commit type::

## Checklist:

### DIFF
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -226,7 +226,27 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
     @uses.config.machine_type.gcp.generic
     Scenario Outline: Attach command in an generic GCP Ubuntu VM
        Given a `<release>` machine with ubuntu-advantage-tools installed
-        When I attach `contract_token` with sudo
+        When I create the file `/tmp/machine-token-overlay.json` with the following:
+        """
+        {
+            "machineTokenInfo": {
+                "contractInfo": {
+                    "resourceEntitlements": [
+                        {
+                            "type": "esm-apps",
+                            "entitled": false
+                        }
+                    ]
+                }
+            }
+        }
+        """
+        And I append the following on uaclient config:
+        """
+        features:
+          machine_token_overlay: "/tmp/machine-token-overlay.json"
+        """
+        And I attach `contract_token` with sudo
         Then stdout matches regexp:
         """
         UA Infra: ESM enabled
@@ -252,5 +272,5 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         Examples: ubuntu release livepatch status
            | release | lp_status | fips_status |
            | xenial  | n/a       | n/a         |
-           | bionic  | n/a       | n/a         |
+           | bionic  | n/a       | disabled    |
            | focal   | n/a       | n/a         |
--- a/features/enable_fips_cloud.feature
+++ b/features/enable_fips_cloud.feature
@@ -16,8 +16,6 @@ Feature: FIPS enablement in cloud based machines
             | release | release_title | fips_service  |
             | xenial  | Xenial        | fips          |
             | xenial  | Xenial        | fips-updates  |
-            | bionic  | Bionic        | fips          |
-            | bionic  | Bionic        | fips-updates  |
             | focal   | Focal         | fips          |
             | focal   | Focal         | fips-updates  |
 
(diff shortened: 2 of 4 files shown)


features/enable_fips_cloud.feature
```diff
@@ -458,3 +456,82 @@ Feature: FIPS enablement in cloud based machines
         Examples: ubuntu release
            | release | fips-name    | fips-service |fips-apt-source                               |
            | focal   | FIPS         | fips         |https://esm.ubuntu.com/fips/ubuntu focal/main |
+
+    @series.bionic
+    @uses.config.machine_type.gcp.generic
+    Scenario Outline: Attached enable of FIPS in an ubuntu Bionic AWS vm
```
blackboxsw, line 462:
```suggestion
    Scenario Outline: Attached enable of FIPS in an ubuntu Bionic GCP vm
```

uaclient/entitlements/fips.py
```diff
@@ -226,12 +231,13 @@ def _replace_metapackage_on_cloud_instance(
...
 
-        cloud_match = re.match(r"^(?P<cloud>(azure|aws)).*", cloud_id)
+        cloud_match = re.match(r"^(?P<cloud>(azure|aws|gce)).*", cloud_id)
         cloud_id = cloud_match.group("cloud") if cloud_match else ""
 
-        if cloud_id not in ("azure", "aws"):
+        if cloud_id not in ("azure", "aws", "gce"):
             return packages
 
+        cloud_id = "gcp" if cloud_id == "gce" else cloud_id
```
blackboxsw, line 240: Let's use our mapping from uaclient.identity.DATASOURCE_TO_CLOUD_ID
```suggestion
        cloud_id = "CLOUD_TYPE_TO_TITLE.get(cloud_id, cloud_id)
```

blackboxsw (commented):
Seems to install fine on GCP generic Bionic images for both fips and fips-updates. After a reboot on fips-updates enabled, I've lost the ability to ssh to the vm. Debugging now

uaclient/entitlements/fips.py
lucasmoura: @blackboxsw the problem with this solution is that the `CLOUD_TYPE_TO_TITLE` is not a proper translation for the other clouds, for example, we would fail to get the correct metapackage if we were on an AWS gov cloud

blackboxsw: great point.

blackboxsw: Anyhow, not a blocker, because it is a one-off specifically for gce anyway. No need to broaden that net at the moment.

features/enable_fips_cloud.feature
```diff
@@ -458,3 +456,151 @@ Feature: FIPS enablement in cloud based machines
...
+            <fips-service> +yes                disabled
+            """
+
+        Examples: ubuntu release
+           | release | fips-name    | fips-service |fips-apt-source                                |
+           | bionic  | FIPS         | fips         |https://esm.ubuntu.com/fips/ubuntu bionic/main |
+
+    @series.bionic
+    @uses.config.machine_type.gcp.generic
+    Scenario Outline: Attached enable of FIPS in an ubuntu Bionic AWS vm
```
blackboxsw, line 540:
```suggestion
    Scenario Outline: Attached enable of FIPS-updates in an ubuntu Bionic GCP vm
```

Also change AWS -> GCP in the other gcp test.

In a separate PR  we can define a step like then_conditional_I_run_command to avoid this test duplication

@then("if `{value1}` in `{value2}` then I run `{command}` `{user_spec}`")

features/attach_validtoken.feature
```diff
@@ -226,7 +226,27 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
     @uses.config.machine_type.gcp.generic
     Scenario Outline: Attach command in an generic GCP Ubuntu VM
        Given a `<release>` machine with ubuntu-advantage-tools installed
-        When I attach `contract_token` with sudo
+        When I create the file `/tmp/machine-token-overlay.json` with the following:
```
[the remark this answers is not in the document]
blackboxsw: +1 makes sense.

blackboxsw (approved):
All good here assuming CI passes or has an explanable failure. VAlidated correct behavior Xenial (error message about non-GCP-optimized kernel) and Bionic successful enable of fips and fips-updates.

orndorffgrant (approved):
Thanks for this Lucas! 
All jenkins failures passed locally - merging

PR #1819, opened 2021-09-24
metering: update how activity info is parsed

lucasmoura:
## Proposed Commit Message
metering: update how activity info is parsed

The structure of the activity information that we will be using to send
a request to the contract backend will be updates. We are now reflecting
this new infrastructure in the code.

## Test Steps

## Desired commit type::

## Checklist:


features/attached_commands.feature
```diff
@@ -515,8 +515,11 @@ Feature: Command behaviour when attached to an UA subscription
               {
                 "code": 200,
                 "response": {
-                  "activityToken": "test",
-                  "activityID": "ac-id",
+                  "activityInfo": {
+                    "activityToken": "test",
+                    "activityID": "ac-id",
+                    "activityPingInterval": "ping-test"
```
blackboxsw, line 521: We should make this the appropriate type integer
this is what we currently get from their endpoint
```yaml
'activityPingInterval': 3600
```

blackboxsw (commented):
One thing I think we also need to sort in this branch is honoring activityPingInterval and resetting the interval with which we are actually making calls to the machine-activity routes.

Maybe we need either a subclass of `TimedJob.run_interval_seconds` of awareness of activityPingInterval values specifically for the metering job to account for contract server changing this value on us.

features/attached_commands.feature
```diff
@@ -515,8 +515,11 @@ Feature: Command behaviour when attached to an UA subscription
               {
                 "code": 200,
                 "response": {
```
orndorffgrant, line 517: I interpreted the conversation as the new response would only be:
```json
{
  "response": {
    "activityToken": "test",
    "activityID": "ac-id",
    "activityPingInterval": "ping-test"
  }
}
```

Might be worth double checking.

And I'd expect this to maybe cause more changes necessary in contract.py `report_machine_activity`?

features/attached_commands.feature
```diff
@@ -543,6 +546,7 @@ Feature: Command behaviour when attached to an UA subscription
         And I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
         Then I verify that running `grep -q activityToken /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
         And I verify that running `grep -q activityID /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
+        And I verify that running `grep -q activityPingInterval /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
```
orndorffgrant, line 549: Should we be calling the equivalent of `ua config set metering_interval <activityPingInterval>` with this value in `report_machine_activity`?

blackboxsw: I think we decided we don't want to override user-customized values in `/etc/ubuntu-advantage/uaclient.conf` as some customers will choose to completely disable this service or throttle the frequency lower than the contract sets.  But, per my comment above, I think our TimedJob.run_interval_seconds should be aware of this valiue for the metering job somehow as we want to prefer the contract server's recommendation of longer  in this scenario:
```python
if uaclient.conf.metering_timer == 0:
   return uaclient.conf.metering_timer
return max(activityPingInteuaclient.conf.metering_timerrval, uaclient.conf.metering_timer)
```

orndorffgrant: Very good point. I'm wondering how we can avoid limiting ourselves in the future though. 

Because we are including 4 hours in the default conf, this snippet would effectively mean we could never reduce the pingInterval below 4 hours via the contract server. Maybe thats okay though?

renanrodrigo: I think the subclass of TimedJob, overwriting the run_interval_seconds method, is good for this scenario.
I also remember some discussion we had about this situation, where the contract server would put a lower expected interval, and from our side (or for the workflow as a whole) would not be a problem, but this is worth a check with CS I guess.

features/attached_commands.feature
lucasmoura: That was the original idea, but I think me and @jaycee agreed that we should create the `activityInfo` structure on all endpoints. But it is worth to double check that.

@jaycee Is my interpretation correct ?

features/attached_commands.feature
lucasmoura: Yes, I think that's the right approach here

lucasmoura:

uaclient/tests/test_ua_timer.py
```diff
@@ -136,3 +136,29 @@ def test_run_job_ignores_late_next_run(self, FakeConfig):
...
+    @mock.patch("lib.timer.TimedJob.run_interval_seconds")
+    def test_metering_run_interval_seconds(
+        self,
+        m_run_interval_seconds,
+        activity_ping_interval_value,
+        config_value,
+        expected_value,
+    ):
+        m_run_interval_seconds.return_value = config_value
+        m_cfg = mock.MagicMock()
```
renanrodrigo, line 155: You could mock the config here with FakeConfig, and pass a dict with the properties (`activity_ping_interval` and `metering_timer`) when creating the instance

lucasmoura: I think I am missing something here, the `activity_ping_interval` is config property. I don't think we can directly override it with `cfg_overrides`.

renanrodrigo: Indeed... You would need to pass the whole structure (machine_token -> activityInfo -> activity_ping_interval) to cfg_overrides. Feel free to ignore this

uaclient/contract.py
```diff
@@ -187,16 +187,19 @@ def report_machine_activity(self, enabled_services: "List[str]"):
...
+
             machine_token = self.cfg.read_cache("machine-token")
             depth_first_merge_overlay_dict(
                 base_dict=machine_token, overlay_dict=response
             )
             self.cfg.write_cache("machine-token", machine_token)
 
-            machine_id = response.get("machineTokenInfo", {}).get("machineId")
-            if machine_id:
-                self.cfg.write_cache("machine-id", machine_id)
```
orndorffgrant, line 198: Were we using this anywhere else? Do we still need the "machine-id" cache item?

features/staging_commands.feature
```diff
@@ -90,3 +90,60 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+        {
+            "machineTokenInfo": {
+                "contractInfo": {
+                   "id": "testCID"
+                },
+                "machineId": "testMID"
+            }
+        }
+        """
+        And I create the file `/tmp/response-overlay.json` with the following:
```
orndorffgrant, line 110: Is the backend to a point where we can have an integration test that actually talks to the backend for this feature/

orndorffgrant (commented):
This is looking really good! Thanks for ironing out the structure details. Just some minor comments and questions

features/staging_commands.feature
```diff
@@ -90,3 +90,60 @@ Feature: Enable command behaviour when attached to an UA staging subscription
...
+        """
+        And I run `ua config set metering_timer=14400` with sudo
+        And I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
+        Then I verify that running `grep -q activityInfo /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
+        And I verify that running `grep -q activityToken /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
+        And I verify that running `grep -q test-activity-token /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
+        And I verify that running `grep -q activityID /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
+        And I verify that running `grep -q test-activity-id /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
+        And I verify that running `grep -q activityPingInterval /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
+        And I verify that running `grep -q 123456789 /var/lib/ubuntu-advantage/private/machine-token.json` `with sudo` exits `0`
```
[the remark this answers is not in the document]
renanrodrigo: We have a step to check for a format (like json) and the first level keys inside, if useful

uaclient/contract.py
lucasmoura: Originally, the machine activity endpoint would return a new `machine-id` if needed. However this was replaced by the `activityID` token. That's why I dropped it here

features/staging_commands.feature
lucasmoura: It is, the problem is that we would need to have a different test structure to properly test the real backend. We would need to compare if we had a change on the `activityToken` and `activityID`.

Also, we would need to double check with @jaycee to double check if we can use the backend for this purpose, just to double check if they have some metrics that would be affected by it.

Let's talk about this on standup

features/staging_commands.feature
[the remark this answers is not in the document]
lucasmoura: I was trying not to create a longer line for that check, but I agree it is better to have the key and value checked at once.
I will update the tests

uaclient/contract.py
orndorffgrant: Got it that makes sense! So can we drop line 98 of `uaclient/config.py` in this PR as well?

features/staging_commands.feature
orndorffgrant: Aha very good point! I imagine it'd be fine for metrics if we use staging for this test.

Also we could simulate a "clone" by doing the metering request twice in a row with the same activityId and activityToken I think

uaclient/contract.py
lucasmoura: Yep, good catch

uaclient/contract.py
```diff
@@ -187,16 +187,19 @@ def report_machine_activity(self, enabled_services: "List[str]"):
...
+            # structure. However, this structure will be reflected when
+            # we reach the contract for attach and refresh requests.
+            # Because of that, we will manually create this structure here.
+            activity_info = {"activityInfo": response}
+
             machine_token = self.cfg.read_cache("machine-token")
             depth_first_merge_overlay_dict(
-                base_dict=machine_token, overlay_dict=response
+                base_dict=machine_token, overlay_dict=activity_info
             )
```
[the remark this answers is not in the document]
orndorffgrant: We can probably simplify the comment above as well if we go this route.

orndorffgrant (commented):
This is really coming together, just some suggested simplifications and comment changes.

uaclient/contract.py
```diff
@@ -187,16 +187,19 @@ def report_machine_activity(self, enabled_services: "List[str]"):
         # provided by the server. We expect the response to be
```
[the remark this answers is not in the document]
lucasmoura: @orndorffgrant I am already covering that use case on `test_report_machine_activity` . But let me know if you think that test can be improved

features/attached_commands.feature
```diff
@@ -609,3 +541,74 @@ Feature: Command behaviour when attached to an UA subscription
            | xenial  |
            | bionic  |
            | focal   |
+
+    @series.lts
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Run timer script on an attached machine
```
renanrodrigo, line 547: Could we just rename this test to explicitly tell we are testing the activity workflow?

renanrodrigo (approved):
Great approach for the steps btw

PR #1821, opened 2021-09-27
tools: create script to generate lp release branches (SC-448)

orndorffgrant:
## Notes
This turns what is currently an inline script in the RELEASES.md instructions into a separate script that can be run as part of our release process.

## Proposed Commit Message

```
tools: create script to generate lp release branches
```

## Test Steps

Try running the new script.

## Desired commit type::

## Checklist:

### DIFF
--- a/tools/create-lp-release-branches.sh
+++ b/tools/create-lp-release-branches.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/bash
+
+DRY_RUN_USAGE="usage: env DEVEL_RELEASE=impish UA_VERSION=27.3 SRU_BUG=1942929 LP_USER=username bash tools/create-lp-release-branches.sh"
+DO_IT_USAGE="usage: env DO_IT=1 DEVEL_RELEASE=impish UA_VERSION=27.3 SRU_BUG=1942929 LP_USER=username bash tools/create-lp-release-branches.sh"
+
+if [ -z "$DEVEL_RELEASE" ]; then
+  echo "please set DEVEL_RELEASE"
+  echo "$DRY_RUN_USAGE"
+  exit 1
+fi
+if [ -z "$UA_VERSION" ]; then
+  echo "please set UA_VERSION"
+  echo "$DRY_RUN_USAGE"
+  exit 1
+fi
+if [ -z "$SRU_BUG" ]; then
+  echo "please set SRU_BUG"
+  echo "$DRY_RUN_USAGE"
+  exit 1
+fi
+if [ -z "$LP_USER" ]; then
+  echo "please set LP_USER"
+  echo "$DRY_RUN_USAGE"
+  exit 1
+fi
+
+if [ -z "$DO_IT"  ]; then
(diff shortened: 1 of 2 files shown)


tools/create-lp-release-branches.sh
```diff
@@ -0,0 +1,74 @@
...
+  set -x
+  set -e
+fi
+
+for release in xenial bionic focal hirsute
+do
+  echo
+  echo $release
+  if [ -z "$DO_IT" ]; then
+    echo git checkout upload-${UA_VERSION}-${DEVEL_RELEASE} -B upload-${UA_VERSION}-$release
```
lucasmoura, line 40: I think we can store the command in a variable and just reference it on both situations:
```bash
cmd="git checkout upload-${UA_VERSION}-${DEVEL_RELEASE} -B upload-${UA_VERSION}-$release"
if [ -z "$DO_IT" ]; then
  echo $cmd
else:
  $cmd
fi
```

orndorffgrant: Yes good point!

orndorffgrant:
@lucasmoura updated!

tools/create-lp-release-branches.sh
```diff
@@ -0,0 +1,79 @@
...
+    $checkout_cmd
+  fi
+
+  case "${release}" in
+      xenial) version=${UA_VERSION}~16.04.1;;
+      bionic) version=${UA_VERSION}~18.04.1;;
+      focal) version=${UA_VERSION}~20.04.1;;
+      hirsute) version=${UA_VERSION}~21.04.1;;
+  esac
+  dch_cmd=(dch -v ${version} -D ${release} -b  "Backport new upstream release: (LP: #${SRU_BUG}) to $release")
```
lucasmoura, line 53: We are treating this as an array because of the changelog message in the middle of the command ?

renanrodrigo:
Just one question: do we want to keep dry-run as the standard? No strong opinion on this, just asking because usually it's the opposite.

tools/create-lp-release-branches.sh
orndorffgrant: Correct! The quotes get inserted as part of the changelog otherwise.

orndorffgrant:
@renanrodrigo Good point! I did the opposite to naturally guide someone through using it. So if the release process can just say "go run this script".
Then [insert new hire who is unfamiliar or me 6 months from now who has forgotten how this works] can just run the script. They'll see the messages to use the envvars. Then after adding the envvars, they'll see the dry run output and can see what its doing. Then the script tells them to add the do-it var, and then it all works.

I'm totally open to changing it though if we prefer a different approach

lucasmoura (approved):

renanrodrigo (approved):

renanrodrigo:
@orndorffgrant makes sense - good point of view

orndorffgrant:
Just a flaky lxd issue on jenkins. Merging

PR #1822, opened 2021-09-27
Deb build cache for tests (SC-449)

orndorffgrant:
## Notes

Not urgent at all.

This is a build caching idea I've been messing with a bit while testing PRs and has worked pretty well. Feel free to say its a bad idea or point out flaws :). It has the potential to make iterating on an integration test less cumbersome.

Also includes a little build script to just call the build function we have for our integration tests.

## Proposed Commit Message

```
test: automatically cache debs based on git repo state
```

## Test Steps

Try out `./tools/build.sh`
make changes to features and build again
make changes to uaclient code and build again

Ensure integration tests still work

## Desired commit type::

## Checklist:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,17 +1,19 @@
 ## Proposed Commit Message
 <!-- Include a proposed commit message because all PRs can be merged in a variety of ways by the reviewer -->
 
-> summary: no more than 70 characters
->
-> A description of what the change being made is and why it is being
-> made, if the summary line is insufficient.  The blank line above is
-> required. This should be wrapped at 72 characters, but otherwise has
-> no particular length requirements.
->
-> If you need to write multiple paragraphs, feel free.
->
-> LP: #NNNNNNN (replace with the appropriate Launchpad bug reference if applicable)
-> Fixes: #NNNNNNN (replace with the appropriate github issue if applicable)
+```
+summary: no more than 70 characters
+
+A description of what the change being made is and why it is being
+made, if the summary line is insufficient.  The blank line above is
+required. This should be wrapped at 72 characters, but otherwise has
+no particular length requirements.
+
+If you need to write multiple paragraphs, feel free.
+
+LP: #NNNNNNN (replace with the appropriate Launchpad bug reference if applicable)
+Fixes: #NNNNNNN (replace with the appropriate github issue if applicable)
+```
 
 ## Test Steps
 <!-- Please include any steps necessary to verify (and reproduce if
@@ -19,13 +21,13 @@ this is a bug fix) this change on a live deployed system,
 including any necessary configuration files, user-data,
 setup, and teardown. Scripts used may be attached directly to this PR. -->
 
-## Desired commit type::
+## Desired commit type
 <!-- put an `x` in one merge type to allow the reviewer to merge for you -->
  - [ ] Squash and merge: Using the "Proposed Commit Message"
  - [ ] Create a merge commit and use "Proposed Commit Message"
  - [ ] Rebase and merge, no merge commit, rely only on existing commit messages
 
-## Checklist:
+## Checklist
 <!-- Go over all the following points, and put an `x` in all the boxes
 that apply. -->
  - [ ] I have updated or added any unit tests accordingly
--- a/features/environment.py
+++ b/features/environment.py
@@ -5,7 +5,6 @@ import os
 import random
 import re
 import string
-import tempfile
 import textwrap
 from typing import Any, Dict, List, Optional, Tuple, Union  # noqa: F401
 
@@ -653,15 +652,9 @@ def build_debs_from_sbuild(context: Context, series: str) -> List[str]:
         )
 
         with emit_spinner_on_travis("Building debs from local source... "):
-            output_deb_dir = os.path.join(tempfile.gettempdir(), series)
             deb_paths = build_debs(
-                series=series,
-                output_deb_dir=output_deb_dir,
-                cache_source=context.config.cache_source,
+                series=series, cache_source=context.config.cache_source
             )
-            context.config.debs_path = output_deb_dir
-
-        logging.info("--- sbuild has finished building the packages")
 
     if "pro" in context.config.machine_type:
         return deb_paths
--- a/features/util.py
+++ b/features/util.py
@@ -1,3 +1,4 @@
+import hashlib
 import logging
 import multiprocessing
 import os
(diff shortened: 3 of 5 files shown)


features/util.py
```diff
@@ -16,8 +17,11 @@
     "container": {"series": "image.release", "machine_type": "image.type"},
 }
 SLOW_CMDS = ["do-release-upgrade"]  # Commands which will emit dots on travis
-SOURCE_PR_TGZ = os.path.join(tempfile.gettempdir(), "pr_source.tar.gz")
-SOURCE_PR_UNTAR_DIR = os.path.join(tempfile.gettempdir(), "behave-ua")
+UA_TMP_DIR = os.path.join(tempfile.gettempdir(), "uaclient")
+SOURCE_PR_TGZ = os.path.join(UA_TMP_DIR, "pr_source.tar.gz")
+SOURCE_PR_UNTAR_DIR = os.path.join(UA_TMP_DIR, "behave-ua-src")
+SBUILD_DIR = os.path.join(UA_TMP_DIR, "sbuild")
+UA_DEB_BUILD_CACHE = os.path.join(UA_TMP_DIR, "deb-cache")
```
orndorffgrant, line 24: I included some changes to put all of our stuff in one directory under /tmp

features/util.py
```diff
@@ -84,9 +88,19 @@ def lxc_get_property(name: str, property_name: str, image: bool = False):
...
+def repo_state_hash(include_features: bool = False):
+    git_log_cmd = ["git", "log", "-1"]
+    git_diff_cmd = ["git", "diff"]
+    if not include_features:
+        files_and_folders = [f for f in os.listdir(".") if "features" not in f]
+        git_diff_cmd += ["--", *files_and_folders]
+    log_out = subprocess.check_output(git_log_cmd)
+    diff_out = subprocess.check_output(git_diff_cmd)
+    output_to_hash = log_out + diff_out
+    return hashlib.md5(output_to_hash).hexdigest()
```
orndorffgrant, line 100: This is the interesting bit. By hashing the text that includes the current commit and the current diff, we get a string that represents the current state of the local repo, including the dirty bits.

The other code is to exclude the `features` directory from this, so as not to rebuild when you're iterating on an integration test

tools/build.py
```diff
@@ -0,0 +1,13 @@
...
+
+from features.util import build_debs
+
+if __name__ == "__main__":
+    logging.basicConfig()
+    logging.getLogger().setLevel(logging.INFO)
+    if len(sys.argv) < 2 or not sys.argv[1]:
+        print("required arg: series")
+        sys.exit(1)
+    series = sys.argv[1]
```
orndorffgrant, line 12: Didn't seem worth writing an argparser

features/util.py
```diff
@@ -16,8 +17,11 @@
     "container": {"series": "image.release", "machine_type": "image.type"},
 }
 SLOW_CMDS = ["do-release-upgrade"]  # Commands which will emit dots on travis
-SOURCE_PR_TGZ = os.path.join(tempfile.gettempdir(), "pr_source.tar.gz")
-SOURCE_PR_UNTAR_DIR = os.path.join(tempfile.gettempdir(), "behave-ua")
+UA_TMP_DIR = os.path.join(tempfile.gettempdir(), "uaclient")
```
lucasmoura, line 20: I think we should only rename this to indicate that is coming from the integration tests, like `uaclient-behave`

features/util.py
```diff
@@ -98,7 +112,26 @@ def build_debs(
...
+        pro_deb_cache_path
+    ):
+        logging.info(
+            "Using debs in cache: {} and {}".format(
+                tools_deb_cache_path, pro_deb_cache_path
+            )
+        )
+        return [tools_deb_cache_path, pro_deb_cache_path]
+
+    logging.info("Creating: {}".format(SOURCE_PR_TGZ))
```
lucasmoura, line 134: Why are we removing the `---` I think it is useful to differentiate our custom messages from other messages in the tests, like pycloudlib or behave

features/util.py
```diff
@@ -140,9 +173,13 @@ def build_debs(
             check=True,
         )
 
+    if os.path.exists(SBUILD_DIR):
+        shutil.rmtree(SBUILD_DIR)
+    os.makedirs(SBUILD_DIR)
```
lucasmoura, line 178: I think we can turn this into a try..except block:
```python
try:
  shutil.rmtree(SBUILD_DIR)
except FileNotFoundError:
  os.makedirs(SBUILD_DIR)
  ```

File: features/util.py [hunk 7]
```diff
@@ -84,9 +88,19 @@ def lxc_get_property(name: str, property_name: str, image: bool = False):
         return value
 
 
-def build_debs(
-    series: str, output_deb_dir: str, cache_source: bool
-) -> List[str]:
+def repo_state_hash(include_features: bool = False):
+    git_log_cmd = ["git", "log", "-1"]
```
lucasmoura, line 92: I think we can only use the hash of the latest commit here: `git rev-parse HEAD`

features/util.py
lucasmoura: I think this approach will work fine, we always look at the most recent commit and check for any new diffs in the files. I think we should only document this approach for future interaction, but it is a great idea

features/util.py
orndorffgrant: Good idea!

features/util.py
orndorffgrant: No strong reason, I'll add them back

features/util.py
orndorffgrant: I think the current version reads just a little bit better than the `try/except` logic to me imho.

But if I am outvoted, I will change it :)

features/util.py
orndorffgrant: Yes good point! Because all we really need is the commit hash.

features/util.py
orndorffgrant: Good point, I will add a docstring to this function

orndorffgrant:
@lucasmoura updated!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,25 +1,27 @@
...
+```
 
 ## Test Steps
 <!-- Please include any steps necessary to verify (and reproduce if
 this is a bug fix) this change on a live deployed system,
 including any necessary configuration files, user-data,
 setup, and teardown. Scripts used may be attached directly to this PR. -->
 
-## Desired commit type::
+## Desired commit type:
```
renanrodrigo, line 24: I am strongly against removing the double colon, it is a valuable visual enhancement to the title :-1: 


Kidding - but (for OCD reasons) we could have no colon at all (like in the previous two titles) or one colon for each title.

orndorffgrant: I would honestly prefer a [`::<>`](https://turbo.fish) but thought this might be more widely accepted :shrug: 

I left it so that the two sections with checklists have colons and the two sections with written content don't have colons. I can totally change it though. I would lean towards no colons on any titles I think actually

renanrodrigo: Ok brains registered your suggestion as a standard: no strong opinion

orndorffgrant: Ha yes :) Just updated to remove all colons

orndorffgrant:
@renanrodrigo my bad - updated

renanrodrigo (approved):
worksforme

features/environment.py
```diff
@@ -653,13 +652,9 @@ def build_debs_from_sbuild(context: Context, series: str) -> List[str]:
...
-            output_deb_dir = os.path.join(tempfile.gettempdir(), series)
             deb_paths = build_debs(
-                series=series,
-                output_deb_dir=output_deb_dir,
-                cache_source=context.config.cache_source,
+                series=series, cache_source=context.config.cache_source
             )
-            context.config.debs_path = output_deb_dir
 
         logging.info("--- sbuild has finished building the packages")
```
lucasmoura, line 659: I think we should move this alert inside the `build_debs`  function, otherwise even if we reuse the packages we will still emit it

lucasmoura (approved):

features/environment.py
orndorffgrant: Good point! I will just move this and then merge

orndorffgrant: Actually `build_debs` already logs "Succesfully ran sbuild" so I will just remove this line

PR #1825, opened 2021-09-28
cli: detach won't ask to disable dependent service (SC-455)

lucasmoura:
## Proposed Commit Message
cli: detach won't ask to disable dependent service

Currently, when running detach without --assume-yes we will ask to disable dependent services if we have ros enabled services during the operation. We are now sorting the services to be disabled based on the length of the depedent services tuple of each entitlement.

## Test Steps

## Desired commit type::

## Checklist:

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -644,6 +644,15 @@ Feature: Enable command behaviour when attached to an UA subscription
         """
         esm-infra     yes                enabled            UA Infra: Extended Security Maintenance \(ESM\)
         """
+        When I run `ua detach` `with sudo` and stdin `y`
+        Then stdout matches regexp:
+        """
+        Updating package lists
+        Updating package lists
+        Updating package lists
+        Updating package lists
+        This machine is now detached.
+        """
 
         Examples: ubuntu release
            | release | ros-security-source                                    | ros-updates-source                                            |
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -905,6 +905,24 @@ def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
         ent = ent_cls(cfg=cfg, assume_yes=assume_yes)
         if ent.can_disable(silent=True):
             to_disable.append(ent)
+
+    """
+    We will nake sure that services without dependencies are disabled first
+    PS: This will only work because we have only three services with reverse
+    dependencies:
+    * ros: ros-updates
+    * esm-infra: ros, ros-updates
+    * esm-apps: ros, ros-updates
+
+    Therefore, this logic will guarantee that we will always disable ros and
+    ros-updates before diabling the esm services. If that dependency chain
+    change, this logic won't hold anymore and must be properly fixed.
+
+    More details can be seen here:
+    https://github.com/canonical/ubuntu-advantage-client/issues/1831
+    """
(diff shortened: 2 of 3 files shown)


features/attached_enable.feature
```diff
@@ -644,6 +644,15 @@ Feature: Enable command behaviour when attached to an UA subscription
         """
         esm-infra     yes                enabled            UA Infra: Extended Security Maintenance \(ESM\)
         """
+        When I run `ua detach` `with sudo` and stdin `y`
+        Then stdout matches regexp:
+        """
+        Updating package lists
```
blackboxsw, line 650: @lucasmoura,  I get why we emit this, but can we create a JIRA ticket to sort a single apt update call at the end of detach?

lucasmoura: Yes, that's a good point @blackboxsw
Will do

uaclient/cli.py
```diff
@@ -905,6 +905,10 @@ def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
         ent = ent_cls(cfg=cfg, assume_yes=assume_yes)
         if ent.can_disable(silent=True):
             to_disable.append(ent)
+
+    # We will nake sure that services without dependencies are disabled first
+    to_disable.sort(key=lambda ent: len(ent.dependent_services))
```
blackboxsw, line 910: It feels like this will break in the scenario if we introduce any more complex dependency chains. We should discuss if we care, but it might be worth explicit ordering based on inspecting the dependent_service names so we don't break future us if we add more coplexity

- A depends on B
- B depends on C&D
- C depends on nothing
In this example, our sorting would try to disable C first, then A (which would prompt for D) and then B because it has a large dependend_services count.

lucasmoura: Yep, that would break. I thought about implementing a full topological sorting to address this issue, but I thought that it would be an overkill for now. But maybe you are right, maybe we should prepared for this approach.

Let me know if you think we should do that and I can implement the algorithm here, no problem at all

lucasmoura: Another reasonable approach is to ignore the dependent services altogether during detach, since we really don't care about their order anyway, we only care about not prompting the user here

blackboxsw: I think what you have makes sense at the moment we can address complexity if we need to in the future. While the 2nd suggestion you posted is compelling to just ignore disable prompts during detach, I still don't think that gets us out of the responsibility of ordering what we disable to prefer services without dependencies first.

orndorffgrant:
@lucasmoura Finally got to actually look at this. I agree with the discussion so far. I think this is a good solution for now but I suggest we do a couple things before merging this:
1. Create a public Github issue documenting what we're implementing here and the known shortcomings.
2. Add a brief comment above the `.sort` line acknowledging the shortcomings and pointing to the Github issue
3. Create a Jira ticket that points to the Github issue so we have it in our internal backlog as well.

lucasmoura:
@orndorffgrant I have updated the code to add a better explanation of when the solution will fail and created the issue and jira card for ir

renanrodrigo (approved):
Aside from the many  calls to `apt update` (already mentioned), it looks good.

orndorffgrant (approved):
Thank you for the updates @lucasmoura and sorry in the delays getting back to this!

PR #1829, opened 2021-09-30
postinst: make check for gce instance more robust

lucasmoura:
## Proposed Commit Message
postinst: make check for gce instance more robust

Currently, we have a check on postinst to verify if we are on a gce cloud. We only verify if the gce string is in the cloud-id of the instance. Therefore, we are open to matching non-gce clouds that have the gce string on their cloud-id. We are now updating this part of the code to make the match more robust by verifying if the cloud-id starts if the gce string.

## Test Steps
Check if the license check tests are still working

## Desired commit type

## Checklist

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -3,6 +3,8 @@ ubuntu-advantage-tools (27.3~21.10.1) impish; urgency=medium
   * d/tools.postinst:
     - consider cloud to be "none" on any cloud-id error
     - purge old ua-messaging.timer/service files
+    - keep ua-timer.timer disabled if ua-messaging.timer was disabled by
+      the user
   * systemd:
     - remove ua-messaging.timer/service
     - add new ua-timer.timer that runs every 2 hours
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -32,6 +32,7 @@ ESM_APPS_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-esm-apps.list"
 FIPS_APT_SOURCE_FILE="$APT_SRC_DIR/ubuntu-fips.list"
 
 OLD_CLIENT_FIPS_PPA="private-ppa.launchpad.net/ubuntu-advantage/fips/ubuntu"
+UA_TIMER_NAME="ua-timer.timer"
 OLD_MESSAGING_TIMER="ua-messaging.timer"
 OLD_MESSAGING_TIMER_MASKED_LOCATION="/etc/systemd/system/timers.target.wants/$OLD_MESSAGING_TIMER"
 
@@ -318,6 +319,40 @@ enable_periodic_license_check() {
     fi
 }
 
+disable_new_timer_if_old_timer_already_disabled() {
+    # If the user has disabled the ua-messaging
+    # then we will assume that the user would want the
+    # ua-timer to be disabled as well. In that case, we will
+    # disable the ua-timer here.
+    PREVIOUS_PKG_VER=$1
+
+    # We should only perform this check on UA version that have the
+    # ua-messaging.timer: 27.0 until 27.2. This will also guarantee
+    # that on 27.3 and forward, we will not run this logic.
+    if dpkg --compare-versions "$PREVIOUS_PKG_VER" ge-nl "27.0~"; then
+        if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt-nl "27.3~"; then
+            if [ -x "/usr/bin/deb-systemd-helper" ]; then
+                if ! deb-systemd-helper --quiet was-enabled $OLD_MESSAGING_TIMER; then
+                    # We have the following entry on our rules file:
+                    #    dh_systemd_enable -pubuntu-advantage-tools ua-timer.timer
+                    # This rule will append some code at the end of the postinst script
+                    # that checks if the ua-timer.timer was already enabled on the system.
+                    # This means that if we manually disable the service here, that logic will
+                    # still enable it in the end of the postinst script.
+                    # Because of this logic we are now manually enabling ua-timer.timer here
+                    # and then we manually disable it. This will guarantee that the was-enabled
+                    # logic introduced by the rules file will not be triggered and we will not
+                    # re-enable the ua-timer.timer service after calling this function.
+                    echo "$OLD_MESSAGING_TIMER was disabled. Disabling $UA_TIMER_NAME." >&2
+                    deb-systemd-helper enable $UA_TIMER_NAME > /dev/null 2>&1 || true
+                    systemctl stop $UA_TIMER_NAME > /dev/null 2>&1 || true
+                    systemctl disable $UA_TIMER_NAME > /dev/null 2>&1 || true
+                fi
+            fi
+        fi
+    fi
+}
+
 remove_old_systemd_timers() {
     # These are the commands that are run when the package is purged.
     # Since we actually want to remove this service from now on
@@ -404,6 +439,7 @@ case "$1" in
       fi
       mark_reboot_for_fips_pro
       enable_periodic_license_check
+      disable_new_timer_if_old_timer_already_disabled $PREVIOUS_PKG_VER
       remove_old_systemd_timers
       ;;
 esac
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -309,7 +309,7 @@ notify_wrong_fips_metapackage_on_cloud() {
 
 enable_periodic_license_check() {
     cloud_id=$(cloud-id 2>/dev/null) || cloud_id=""
-    if echo "$cloud_id" | grep -q "gce"; then
+    if echo "$cloud_id" | grep -q "^gce"; then
         if check_is_lts "${UBUNTU_CODENAME}"; then
             if [ ! -f $MACHINE_TOKEN_FILE ]; then
                 touch $LICENSE_CHECK_MARKER_FILE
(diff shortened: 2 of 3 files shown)


debian/ubuntu-advantage-tools.postinst
```diff
@@ -318,6 +319,28 @@ enable_periodic_license_check() {
...
+
+    # We should only perform this check on UA version that have the
+    # ua-messaging.timer: 27.0 until 27.2. This will also guarantee
+    # that on 27.3 and forward, we will not run this logic.
+    if dpkg --compare-versions "$PREVIOUS_PKG_VER" ge-nl "27.0~"; then
+        if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt-nl "27.3~"; then
+            if [ -x "/usr/bin/deb-systemd-helper" ]; then
+                if ! deb-systemd-helper --quiet was-enabled $OLD_MESSAGING_TIMER; then
+                    systemctl stop $NEW_TIMER
+                    systemctl mask $NEW_TIMER
```
orndorffgrant, line 337: Can't wait until we add another timer and then we'll have a variable named `$NEW_NEW_TIMER` :upside_down_face: 

Just joking :) This looks good!

debian/changelog
```diff
@@ -3,6 +3,8 @@ ubuntu-advantage-tools (27.3~21.10.1) impish; urgency=medium
   * d/tools.postinst:
     - consider cloud to be "none" on any cloud-id error
     - purge old ua-messaging.timer/service files
+    - keep ua-timer.timer disable if ua-messaging.timer was disabled by
+      the user
```
orndorffgrant, line 7: Smallest of typos
```suggestion
    - keep ua-timer.timer disabled if ua-messaging.timer was disabled by
      the user
```

debian/ubuntu-advantage-tools.postinst
```diff
@@ -318,6 +319,28 @@ enable_periodic_license_check() {
     fi
 }
 
+disable_new_timer_if_old_timer_already_disabled() {
+    # If the user has disable the old timer (ua-messaging)
+    # We will assume that the user would also not want the
+    # new timer to be disabled as well. Therefore, we will be
+    # manually disabling the new timer here.
```
orndorffgrant, line 326: Looks like an accidental double-negative in here 

Also minor grammatical suggestions
```suggestion
    # If the user has disabled the old timer (ua-messaging)
    # then we will assume that the user would want the
    # new timer to be disabled as well. In that case, we will
    # disable the new timer here.
```

debian/ubuntu-advantage-tools.postinst
orndorffgrant: We should consider adding these version if statements on all future postinst migrations too.

orndorffgrant (commented):
Thanks for figuring this out!

Just some small grammar suggestions and I haven't actually tested it yet, doing that next

lucasmoura:
@orndorffgrant I have updated the typos in this PR and I have also changed the variable name here. I know this is not necessary, but you are right we would need to eventually change it in the future so I am doing this now haha

debian/ubuntu-advantage-tools.postinst
```diff
@@ -318,6 +319,28 @@ enable_periodic_license_check() {
...
+
+    # We should only perform this check on UA version that have the
+    # ua-messaging.timer: 27.0 until 27.2. This will also guarantee
+    # that on 27.3 and forward, we will not run this logic.
+    if dpkg --compare-versions "$PREVIOUS_PKG_VER" ge-nl "27.0~"; then
+        if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt-nl "27.3~"; then
+            if [ -x "/usr/bin/deb-systemd-helper" ]; then
+                if ! deb-systemd-helper --quiet was-enabled $OLD_MESSAGING_TIMER; then
+                    systemctl stop $UA_TIMER_NAME
+                    systemctl mask $UA_TIMER_NAME
```
orndorffgrant, line 337: I was playing around with an alternative to masking just now, it looks like this
```
deb-systemd-helper enable 'ua-timer.timer' >/dev/null || true
systemctl stop $NEW_TIMER > /dev/null || true
systemctl disable $NEW_TIMER >/dev/null || true
```
But I haven't tested it thoroughly and its friday afternoon so I won't get back to it til monday :)

The theory behind it is that by calling `deb-systemd-helper enable` and then calling `systemctl disable`, then the `was-enabled` check later in the postinst script will return false.

orndorffgrant: We should also look at what output is being generated from this section during install and make sure its not too confusing (my suggestion here might cause particularly confusing output)

orndorffgrant: @lucasmoura Okay I've arrived at a suggestion. Here it is:
```suggestion
                    echo "$OLD_MESSAGING_TIMER was disabled. Disabling $UA_TIMER_NAME." >&2
                    deb-systemd-helper enable $UA_TIMER_NAME > /dev/null 2>&1 || true
                    systemctl stop $UA_TIMER_NAME > /dev/null 2>&1 || true
                    systemctl disable $UA_TIMER_NAME > /dev/null 2>&1 || true
```

Then when upgrading we get the following output:
```
ua-messaging.timer was disabled. Disabling ua-timer.timer.
Created symlink /etc/systemd/system/multi-user.target.wants/ua-license-check.path → /lib/systemd/system/ua-license-check.path.
ua-timer.timer is a disabled or a static unit, not starting it.
```

The nice thing about this is that the new timer ends up in the same state as the old timer (disabled) instead of a different state (masked).

The not nice thing about this is that the code is unintuitive and could be construed as "tricking the later debhelper snippet into thinking it was disabled". I think we should add a comment explaining what this is doing if we go this route.

Let's talk tradeoffs after standup

sru/release-27.3/ua-messaging-disabled.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+series=$1
+name="test-systemd-$series"
+DEB_PATH=$2
+
+lxc launch ubuntu-daily:$series $name
+sleep 5
+lxc file push $DEB_PATH $name/tmp/ua.deb
+echo "Disabling ua-messaging.timer"
```
orndorffgrant, line 10: Can we add a `list-timers` call before disabling just to easily see it before the disable?
```suggestion
echo "List timers in default state"
echo "--------------------------"
lxc exec $name -- sudo systemctl list-timers --all
echo "--------------------------"
echo "Disabling ua-messaging.timer"
```

debian/ubuntu-advantage-tools.postinst
athos-ribeiro: LGTM. It would be nice to comment the approach, as suggested, since this may not be obvious, as @orndorffgrant stated.

athos-ribeiro (commented):
Hi Lucas,

This LGTM. I added a few comments inline. Also, note that the first commit message seems to have a typo (s/starts if/starts with/ ?) which you may want to fix before merging this :)

lucasmoura:
@athos-ribeiro just to sure, the approach you think it is fine is the one @orndorffgrant is suggesting right ? Not the one that is commited in this PR

debian/ubuntu-advantage-tools.postinst
sergiodj: Sorry, I don't know if I understand @orndorffgrant's comment when he says:

> The theory behind it is that by calling `deb-systemd-helper enable` and then calling `systemctl disable`, then the `was-enabled` check later in the postinst script will return false.

I'm only seeing one call to `deb-systemd-helper was-enabled` in the code.

orndorffgrant: Yes sorry for the poor explanation @sergiodj 

debhelper appends the following at the end of our postinst (among other things)
```bash
# Automatically added by dh_systemd_enable/11.1.6ubuntu2
if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ] ; then
        # This will only remove masks created by d-s-h on package removal.
        deb-systemd-helper unmask 'ua-timer.timer' >/dev/null || true

        # was-enabled defaults to true, so new installations run enable.
        if deb-systemd-helper --quiet was-enabled 'ua-timer.timer'; then
                # Enables the unit on first installation, creates new
                # symlinks on upgrades if the unit file has changed.
                deb-systemd-helper enable 'ua-timer.timer' >/dev/null || true
        else
                # Update the statefile to add new symlinks (if any), which need to be
                # cleaned up on purge. Also remove old symlinks.
                deb-systemd-helper update-state 'ua-timer.timer' >/dev/null || true
        fi
fi
# End automatically added section
```

Because we have this line in our `debian/rules`:
```
dh_systemd_enable -pubuntu-advantage-tools ua-license-check.timer
```

sergiodj: Ah, of course, no need to apologize @orndorffgrant, I just thought you were referring to a call to `deb-systemd-helper` that got removed by mistake.  Indeed, I can see where you're coming from now, thanks.

Yeah, I agree the suggestion is a hack but I cannot think of a better approach right now.  I'm a strong +1 to properly document it on the postinst script, and I think we can proceed with this solution.

orndorffgrant (approved):
Thank you for the iterations on this and ironing out all the edge cases. Also thank you for the descriptive comments - I think they help a lot!

athos-ribeiro (approved):

orndorffgrant:
Jenkins failures are lxd flakiness and passed locally. Merging now

PR #1830, opened 2021-10-01
logs: update logrotate to glob capture all /var/log/ubuntu-advantage*log

blackboxsw:
WIP: Touching integration test failures
## Proposed Commit Message

>  logs: update logrotate to glob capture all /var/log/ubuntu-advantage*log
>  
>  Provide `su root root` to avoid messaging about parent directory perms
>  being insecure and allowing syslog group to write this file.
## Test Steps

```bash
UACLIENT_BEHAVE_CONTRACT_TOKEN=<REDACTED>   UACLIENT_BEHAVE_BUILD_PR=1 UACLIENT_BEHAVE_DEBS_PATH=DEBS tox -e behave-lxd-18.04 features/attached_commands.feature:46
```
## Desired commit type::

## Checklist:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -36,6 +36,9 @@ ubuntu-advantage-tools (27.3~21.10.1) impish; urgency=medium
       + remove unicode-only characters from help file
       + don't print unicode-only characters in ua fix if non-utf8 locale
         (GH: #1463)
+    - logrotate:
+      + add logrotate functionality for ubuntu-advantage-timer.log.
+      + Fix root:root logrotate permissions.
     - ua-timer.timer:
       + introduce a single systemd timer to handle ua recurring jobs
       + timer runs every 2 hours to support most frequent timer job
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -27,19 +27,20 @@ Feature: Command behaviour when attached to an UA subscription
             """
             Successfully refreshed your subscription.
             """
-        When I run `ls /var/log/ubuntu-advantage*` as non-root
-        Then I will see the following on stdout:
-            """
-            /var/log/ubuntu-advantage.log
-            /var/log/ubuntu-advantage-timer.log
-            """
+        When I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
+        And I run `sh -c "ls /var/log/ubuntu-advantage* | sort -d"` as non-root
+        Then stdout matches regexp:
+        """
+        /var/log/ubuntu-advantage.log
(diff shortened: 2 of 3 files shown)


orndorffgrant:
Waiting on jenkins results, but after jenkins is happy, can we also add a changelog entry for this?

features/attached_commands.feature
```diff
@@ -27,6 +27,19 @@ Feature: Command behaviour when attached to an UA subscription
             """
             Successfully refreshed your subscription.
             """
+        When I run `ls /var/log/ubuntu-advantage*` as non-root
+        Then I will see the following on stdout:
+            """
+            /var/log/ubuntu-advantage.log
+            /var/log/ubuntu-advantage-timer.log
+            """
```
orndorffgrant, line 35: @blackboxsw Looks like we're getting some legitimate CI failures. I think its that the `*` here isn't being expanded (we aren't using `shell=True` for our subprocess)

I think we can replace this section and the following with something like this to still use the globbing:
We might also need a `-1` to put each entry on its own line?
```suggestion
        When I run `sh -c "ls -1 /var/log/ubuntu-advantage*"` as non-root
        Then I will see the following on stdout:
            """
            /var/log/ubuntu-advantage.log
            /var/log/ubuntu-advantage-timer.log
            """
```

debian/changelog
```diff
@@ -36,6 +36,8 @@ ubuntu-advantage-tools (27.3~21.10.1) impish; urgency=medium
       + remove unicode-only characters from help file
       + don't print unicode-only characters in ua fix if non-utf8 locale
         (GH: #1463)
+    - logrotate: add logrotate functionality for ubuntu-advantage-timer.log.
+      Fix root:root logrotate permissions.
```
orndorffgrant, line 40:
```suggestion
    - logrotate:
      + add logrotate functionality for ubuntu-advantage-timer.log.
      + fix root:root logrotate permissions.
```

orndorffgrant (approved):
Looks good and integration test passed for me! Approving pending jenkins. And just a small changelog suggestion. In addition, changing the changelog commit message to something like "changelog: list logrotate fixes" would be slightly better but :shrug:

debian/changelog
blackboxsw: +1 will fix the changelog. entry and merge once we see this current CI run succeed.

orndorffgrant:
Jenkins failures are flakes due to focal images not being found by lxd. The new integration test passes for me. Thank you for the updates @blackboxsw! Merging now

PR #1834, opened 2021-10-05
Update logrotate and postrm

lucasmoura:


athos-ribeiro (approved):

sergiodj (approved):

blackboxsw (approved):
Confirmed behavior on upgrade. we're seeing all the logs we care about with the right perms and no errors from apt during install.
```
-rw------- 1 root root 10976 Oct  1 03:39 /var/log/ubuntu-advantage.log.1
-rw------- 1 root root  3628 Oct  5 21:06 /var/log/ubuntu-advantage.log
-rw------- 1 root root     0 Oct  5 21:06 /var/log/ubuntu-advantage-license-check.log
-rw------- 1 root root   157 Oct  5 21:06 /var/log/ubuntu-advantage-timer.log
```

Log rotate functions fine with the su root root inside the scope of the config.

features/attached_commands.feature
```diff
@@ -31,13 +31,16 @@ Feature: Command behaviour when attached to an UA subscription
...
+        /var/log/ubuntu-advantage-license-check.log
         /var/log/ubuntu-advantage.log
         /var/log/ubuntu-advantage-timer.log
         """
         When I run `logrotate --force /etc/logrotate.d/ubuntu-advantage-tools` with sudo
         And I run `sh -c "ls /var/log/ubuntu-advantage* | sort -d"` as non-root
+        # We don't rotate empty log files. That's why the license-check logs are not rotated
         Then stdout matches regexp:
         """
+        /var/log/ubuntu-advantage-license-check.log
```
renanrodrigo, line 43: Should not we have a `.1` here as well?

orndorffgrant: There is a comment above.
```
# We don't rotate empty log files. That's why the license-check logs are not rotated
```

renanrodrigo: :man_facepalming: 
I'm blind or something

PR #1835, opened 2021-10-06
Check systemd existence

lucasmoura:


debian/ubuntu-advantage-tools.postinst
```diff
@@ -345,8 +345,10 @@ disable_new_timer_if_old_timer_already_disabled() {
                     # re-enable the ua-timer.timer service after calling this function.
                     echo "$OLD_MESSAGING_TIMER was disabled. Disabling $UA_TIMER_NAME." >&2
                     deb-systemd-helper enable $UA_TIMER_NAME > /dev/null 2>&1 || true
-                    systemctl stop $UA_TIMER_NAME > /dev/null 2>&1 || true
-                    systemctl disable $UA_TIMER_NAME > /dev/null 2>&1 || true
+                    if [ -d /run/systemd/system ]; then
```
sergiodj, line 348: Can you move this check to the outer scope, please?  If systemd is not present, then we don't want the `echo` command to run either (because the `deb-systemd-helper` invocation will also be a no-op); we want to skip this whole block.

orndorffgrant (approved):
LGTM and manual test still passes. Might as well let jenkins finish before merging at this point. I'll come back to this first thing tomorrow and update our release MP.

orndorffgrant:
The jenkins errors are unrelated to this change. Merging now.

PR #1836, opened 2021-10-07
fix: check reboot-required.pkgs to reboot

lucasmoura:
## Proposed Commit Message
fix: check reboot-required.pkgs to reboot

During a fix operation we may install some package on the system. Those packages may require a reboot to complete their setup. Currently, we only check the existence of the /var/run/reboot-required file to notify the user that the system needs a reboot. However, that marker file could have been generated by a different package install outside of the fix operation. To address that, we are now using the /var/run/reboot-required.pkgs file to check if the packages installed during fix are on that file. Only if we have the packages there and the reboot marker file that we will ask the user to reboot the system.

## Checklist

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -589,12 +589,16 @@ class UAEntitlement(metaclass=abc.ABCMeta):
 
         return True
 
+    def _check_for_reboot(self) -> bool:
+        """Check if system needs to be rebooted."""
+        return util.should_reboot()
+
     def _check_for_reboot_msg(self, operation: str) -> None:
         """Check if user should be alerted that a reboot must be performed.
 
         @param operation: The operation being executed.
         """
-        if util.should_reboot():
+        if self._check_for_reboot():
             print(
                 status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
                     operation=operation
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -52,6 +52,10 @@ class RepoEntitlement(base.UAEntitlement):
 
         return packages
 
+    def _check_for_reboot(self) -> bool:
+        """Check if system needs to be rebooted."""
+        return util.should_reboot(installed_pkgs=set(self.packages))
+
     @property
     @abc.abstractmethod
     def repo_key_file(self) -> str:
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -6,7 +6,7 @@ import socket
 import textwrap
 from collections import defaultdict
 from datetime import datetime
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple
 
 from uaclient import apt, exceptions, serviceclient, status, util
 from uaclient.clouds.identity import (
@@ -30,6 +30,17 @@ API_V1_NOTICE_TMPL = "notices/{notice}.json"
 UBUNTU_STANDARD_UPDATES_POCKET = "Ubuntu standard updates"
 UA_INFRA_POCKET = "UA Infra"
 UA_APPS_POCKET = "UA Apps"
+
+
+ReleasedPackagesInstallResult = NamedTuple(
+    "ReleasedPackagesInstallResult",
+    [
+        ("fix_status", bool),
+        ("unfixed_pkgs", Set[str]),
+        ("installed_pkgs", Set[str]),
+        ("all_already_installed", bool),
+    ],
+)
 
 
 @enum.unique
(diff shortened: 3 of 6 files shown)


uaclient/security.py
```diff
@@ -824,8 +836,15 @@ def _handle_released_package_fixes(
 
             if not upgrade_status:
                 unfixed_pkgs += [src_pkg for src_pkg, _ in pkg_src_group]
+            else:
+                installed_pkgs += binary_pkgs
```
orndorffgrant, line 840: Can this new line go in the if statement above?
```suggestion
                installed_pkgs += binary_pkgs
                
            if not upgrade_status:
                unfixed_pkgs += [src_pkg for src_pkg, _ in pkg_src_group]
```

uaclient/security.py
```diff
@@ -824,8 +836,15 @@ def _handle_released_package_fixes(
...
 
-    return upgrade_status, unfixed_pkgs, all_already_installed
+    return ReleasedPackagesInstallResult(
+        fix_status=upgrade_status,
+        unfixed_pkgs=unfixed_pkgs,
+        installed_pkgs=installed_pkgs,
+        all_already_installed=all_already_installed,
+    )
```
orndorffgrant, line 847: Thanks for adding structure to this!

uaclient/security.py
```diff
@@ -917,29 +936,25 @@ def prompt_for_affected_packages(
...
+    if released_pkgs_install_result.fix_status:
         # fix_status is True if either:
         #  (1) we successfully installed all the packages we needed to
         #  (2) we didn't need to install any packages
         # In case (2), then all_already_installed is also True
-        if all_already_installed:
+        if released_pkgs_install_result.all_already_installed:
             # we didn't install any packages, so we're good
             print(util.handle_unicode_characters(fix_message))
             return (
```
orndorffgrant, line 960: I think we need to change the call below to `util.should_reboot(released_pkgs_install_result.installed_pkgs)`

uaclient/util.py
```diff
@@ -754,9 +756,40 @@ def redact_sensitive_logs(
...
+    except FileNotFoundError:
+        # If the file doesn't exist, we will default to the
+        # reboot  marker file
+        return True
+    else:
+        return (
+            True
+            if installed_pkgs.intersection(reboot_required_pkgs)
+            else False
+        )
```
orndorffgrant, line 792: We can simplify this to just
```suggestion
        return installed_pkgs.intersection(reboot_required_pkgs)
```

lucasmoura: We can't make this change here because the `intersection` method will return another `Set`. Here, we just wan't to verify that the `Set` returned is non-empty.

However, we can update this call to:
```python
return len(installed_pkgs.intersection(reboot_required_pkg) != 0
```

uaclient/security.py
lucasmoura: Yep, great catch

uaclient/util.py
orndorffgrant: You both are totally right, any of those or the original would work, but not my suggestion haha :upside_down_face:

uaclient/entitlements/repo.py
```diff
@@ -52,6 +52,10 @@ def packages(self) -> List[str]:
 
         return packages
 
+    def _check_for_reboot(self) -> bool:
+        """Check if system needs to be rebooted."""
+        return util.should_reboot(installed_pkgs=set(self.packages))
+
```
orndorffgrant, line 58: Cool thanks for adding this!

uaclient/entitlements/base.py
```diff
@@ -589,12 +589,16 @@ def _disable_dependent_services(self):
...
     def _check_for_reboot_msg(self, operation: str) -> None:
         """Check if user should be alerted that a reboot must be performed.
 
         @param operation: The operation being executed.
         """
-        if util.should_reboot():
+        if self._check_for_reboot():
             print(
                 status.MESSAGE_ENABLE_REBOOT_REQUIRED_TMPL.format(
                     operation=operation
```
orndorffgrant, line 604: Do we need this in the base entitlement?
It could suggest that is a reboot is required for livepatch to finish enabling, which I don't think makes sense, right?

uaclient/security.py
orndorffgrant: The preceding if statement is `if upgrade_status:`

So I'm suggesting this new line go in that preceding if statement. Then we could actually turn this `if not upgrade_status` into an `else:`
```python
if upgrade_status:
    # all the stuff above
    # ....
    installed_pkgs.update(binary_pkgs)
else:
    unfixed_pkgs.update([src_pkg for src_pkg, _ in pkg_src_group])
```

uaclient/entitlements/base.py
lucasmoura: No, because on the base entitlement we don't pass the packages that were installed to the `should_reboot` function

lucasmoura: Yes, for non-repo entitlements, that will happen. But this is what is happening today. Do you think we should pass an empty list for those situations ?

orndorffgrant: I think passing an empty list is the same as not passing a list in the current implementation.

But it is probably okay to leave it as is.

orndorffgrant (approved):
Looks great! And I just played around with it manually and it works as expected. Thanks for adding those steps to the integration test!

PR #1838, opened 2021-10-08
postinst: hardcode python binary and PYTHONPATH

orndorffgrant:
## Proposed Commit Message

```
postinst: hardcode python binary and PYTHONPATH

When the user's environment is set up to use a nonstandard python
installation, our postinst would fail. By specifying the exact python
binary to use, we avoid this problem. By also setting PYTHONPATH to the
location where our uaclient library is installed, we increase the
chances that we will succeed even if /usr/bin/python3 is overwritten.

If /usr/bin/python3 is overwritten with a different version of python,
whose standard library is incompatible with the expected version of
python, it is possible postinst could still fail.

Separately, if /usr/bin/python3 is overwritten with a different version
of python, the ua executable _will_ still fail.

LP: #1946034 #1944700 #1930121 #1938290 #1938074 #1927905
```

## Test Steps

1. Run the new integration tests.
2. Forgive me for making Jenkins compile python from source for CI (This is the only "natural" way that I could find where user will end up with a nonstandard python install that causes the symptoms of the linked LP bugs).

## Desired commit type

## Checklist

### DIFF
--- a/features/util.py
+++ b/features/util.py
@@ -8,7 +8,7 @@ import tempfile
 import textwrap
 import time
 from contextlib import contextmanager
-from typing import List
+from typing import Iterable, List
 
 import yaml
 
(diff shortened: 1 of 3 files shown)


debian/ubuntu-advantage-tools.postinst
```diff
@@ -16,6 +16,9 @@ fi
 # https://lintian.debian.org/tags/postinst-does-not-load-confmodule.html
 . /usr/share/debconf/confmodule
 
+PYTHON=/usr/bin/python3
+export PYTHONPATH=/usr/lib/python3/dist-packages
```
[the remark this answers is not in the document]
orndorffgrant: This should only affect commands run as part of this script, but that is a good point that that is unnecessary, we could instead change this to
```
PYTHON=/usr/bin/env PYTHONPATH=/usr/lib/python3/dist-packages /usr/bin/python3
```

and then we wouldn't need the export line. Thoughts?

debian/ubuntu-advantage-tools.postinst
```diff
@@ -16,6 +16,8 @@ fi
 # https://lintian.debian.org/tags/postinst-does-not-load-confmodule.html
 . /usr/share/debconf/confmodule
 
+PYTHON="/usr/bin/env PYTHONPATH=/usr/lib/python3/dist-packages /usr/bin/python3"
```
lucasmoura, line 19: Can we only have `/usr/lib/python3/`  as the python path here ?

features/util.py
```diff
@@ -113,13 +114,24 @@ def repo_state_hash(include_features: bool = False):
...
+
     git_diff_cmd = ["git", "diff"]
     if not include_features:
         files_and_folders = [f for f in os.listdir(".") if "features" not in f]
         git_diff_cmd += ["--", *files_and_folders]
-    rev_out = subprocess.check_output(git_rev_cmd)
     diff_out = subprocess.check_output(git_diff_cmd)
-    output_to_hash = rev_out + diff_out
+
+    git_status_cmd = ["git", "status", "--porcelain"]
```
lucasmoura, line 125: I see why we are doing this, but I don't know if we should. Any test script that we create or any manual test we add, will trigger for a new package build. I know this is the more complete solution, but I don't know if it pays off.

What do you think @renanrodrigo @blackboxsw ?

orndorffgrant: I updated this function to exclude more directories

debian/ubuntu-advantage-tools.postinst
orndorffgrant: I just tested and unfortunately, having only `/usr/lib/python3/` doesn't work here

features/util.py
```diff
@@ -88,38 +88,61 @@ def lxc_get_property(name: str, property_name: str, image: bool = False):
         return value
 
 
-def repo_state_hash(include_features: bool = False):
+def repo_state_hash(
+    exclude_dirs: List[str] = [".github", "demo", "features", "sru", "tools"]
```
lucasmoura, line 92: Even though we are not doing this at the moment, but if we modify the `exclude_dirs` default list on the function, we will modify it on all other calls of `repo_state_hash`.

I know we are not doing that, but it is better to be defensive here. It is better to have that as `None`  by default and set it up with a default list.

features/util.py
```diff
@@ -88,38 +88,61 @@ def lxc_get_property(name: str, property_name: str, image: bool = False):
...
+    git_status_cmd = ["git", "status", "--porcelain"]
+    status_out = subprocess.check_output(git_status_cmd).decode("utf-8")
+    status_lines = status_out.split("\n")
+    new_files = [name[3:] for name in status_lines if name.startswith("?? ")]
+    new_file_content = ""
+    for fname in new_files:
+        exclude = False
+        for exclude_dir in exclude_dirs:
+            if fname.startswith(exclude_dir):
+                exclude = True
```
lucasmoura, line 139: Let's add a `break` here

features/util.py
renanrodrigo: I think it is fine. We can tweak the "watched" directories in the future if needed, but so far, it works

lucasmoura (approved):

orndorffgrant:
LP 1920121 is now the single bug for this issue.

I've updated the commit and fixed the mypy issue that stopped the last jenkins run. Let's let jenkins do a full run now.

PR #1839, opened 2021-10-14
cli: implement update-info endpoint

renanrodrigo:
WIP for the update-info command.

Still needs:
- better unit tests, with a couple more controlled scenarios
- integration tests
- documentation
 
## Desired commit type

## Checklist


uaclient/update_info.py
```diff
@@ -0,0 +1,88 @@
...
+        "upgradable_packages": [],
+    }  # type: Dict[str, Any]
+    cache = Cache()
+    status = cfg.status()
+
+    is_attached = status.get("attached")
+    output["summary"]["attached"] = is_attached
+
+    if is_attached:
+        for service in ("esm-infra", "esm-apps"):
```
lucasmoura, line 59: Let's create a constant variable for this tuple

uaclient/update_info.py
```diff
@@ -0,0 +1,88 @@
...
+            entitled, enabled = get_service_status(status, service)
+            if entitled == "yes":
+                output["summary"]["entitled_services"].append(service)
+            if enabled == "enabled":
+                output["summary"]["enabled_services"].append(service)
+
+    upgradable_packages = [
```
lucasmoura, line 66: Maybe we should create auxiliary functions for the different parts of the code to easily test the individual pieces here

uaclient/cli.py
```diff
@@ -377,6 +387,10 @@ def refresh_parser(parser):
     return parser
 
 
+def action_update_info(args, *, cfg, **kwargs):
+    return json.dumps(update_info.update_info(cfg))
```
lucasmoura, line 391: We are only outputting this in json in the first iteration, right ? We want this to be user readable as well at some point in time, right ?

uaclient/update_info.py
renanrodrigo: I am sure we will need to do it in future iterations. As of now, I'd like to keep it as is, because it represents well which kind of data is being defined for the output and there is a single flow. Feel free to point out to pieces you think that can be isolated.

uaclient/update_info.py
```diff
@@ -0,0 +1,90 @@
+from typing import Any, Dict
```
orndorffgrant, line 1: I like that you moved this subcommand code into its own file +1

Should we also create a "subcommands/" directory and move this file there? Or do you think that is too much right now?

uaclient/update_info.py
```diff
@@ -0,0 +1,90 @@
...
+# All our unit tests are mocking the Cache object, so we are not losing
+# anything here. If by any chance there is no python3-apt in the future, the
+# integration tests should be able to detect it through failure, because we
+# don't expect an empty apt cache there.
+try:
+    from apt import Cache  # type: ignore
+except ImportError:
+
+    def Cache():
+        return []
```
orndorffgrant, line 17: Ack that the suggestion below is basically the same thing, but I think I prefer it because (1) it keeps the test mock in a test file, and (2) it won't require that we add more mock code to this file if we start to rely on additional members of `apt` in the future.

```diff
diff --git a/uaclient/conftest.py b/uaclient/conftest.py
index 5328878d..1fb3fa7d 100644
--- a/uaclient/conftest.py
+++ b/uaclient/conftest.py
@@ -1,5 +1,6 @@
 import io
 import logging
+import sys
 from typing import Any, Dict
 
 import mock
@@ -7,6 +8,10 @@ import pytest
 
 from uaclient.config import UAConfig
 
+# We are doing this because we are sure that python3-apt comes with the distro,
+# but it cannot be installed in a virtual environment to be properly tested.
+sys.modules["apt"] = mock.MagicMock()
+
 
 @pytest.fixture
 def caplog_text(request):
diff --git a/uaclient/update_info.py b/uaclient/update_info.py
index 3d9b9c88..2a981d2a 100644
--- a/uaclient/update_info.py
+++ b/uaclient/update_info.py
@@ -3,18 +3,7 @@ from typing import Any, Dict
 from uaclient import util
 from uaclient.config import UAConfig
 
-# We are doing this because we are sure that python3-apt comes with the distro,
-# but it cannot be installed in a virtual environment to be properly tested.
-# All our unit tests are mocking the Cache object, so we are not losing
-# anything here. If by any chance there is no python3-apt in the future, the
-# integration tests should be able to detect it through failure, because we
-# don't expect an empty apt cache there.
-try:
-    from apt import Cache  # type: ignore
-except ImportError:
-
-    def Cache():
-        return []
+from apt import Cache  # type: ignore
 
 
 ESM_SERVICES = ("esm-infra", "esm-apps")
```

uaclient/update_info.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+ESM_SERVICES = ("esm-infra", "esm-apps")
+
+series = util.get_platform_info()["series"]
```
orndorffgrant, line 22: This is not for this PR but just an observation that I made recently that this reminds me of:

We call `get_platform_info` all over the codebase, and I don't think the result should ever change during one run, but it reads the os release file every single time. For that reason, it is good that you avoid calling this repeatedly (though it makes unit testing harder I think, see below). I _think_ we could just throw a `lru_cache` on this function and then call it repeatedly without remorse - that would clean up our log files as well, which currently log that we're reading that os file all the time. I'm not sure if there might be any unintended consequences to that though, @lucasmoura @blackboxsw thoughts?

uaclient/update_info.py
```diff
@@ -0,0 +1,90 @@
...
+        "{}-infra-updates".format(series),
+        "{}-infra-security".format(series),
+    ):
+        return "esm-infra"
+    if archive in (
+        "{}-apps-updates".format(series),
+        "{}-apps-security".format(series),
+    ):
+        return "esm-apps"
+    return archive
```
orndorffgrant, line 36: This kind of function is perfect for an easy parametrized unit test, though the `series` variable at the module level might complicate things. You could call that down in `update_info` and pass it as a param instead

uaclient/update_info.py
```diff
@@ -0,0 +1,90 @@
...
+    output["summary"]["attached"] = is_attached
+
+    if is_attached:
+        for service in ESM_SERVICES:
+            entitled, enabled = get_service_status(status, service)
+            if entitled == "yes":
+                output["summary"]["entitled_services"].append(service)
+            if enabled == "enabled":
+                output["summary"]["enabled_services"].append(service)
+
```
orndorffgrant, line 67: We already have ways of attaining entitled/enabled status that doesn't require interpreting the results of the current schema/values of `cfg.status()`
```suggestion
            ent = entitlements.ENTITLEMENT_CLASS_BY_NAME[service](cfg)
            if ent.contract_status() == ContractStatus.ENTITLED:
                output["summary"]["entitled_services"].append(service)
            if ent.application_status()[0] == status.ApplicationStatus.ENABLED:
                output["summary"]["enabled_services"].append(service)
```

uaclient/update_info.py
```diff
@@ -0,0 +1,90 @@
...
+    upgradable_packages = [
+        package
+        for package in cache
+        if package.is_installed and max(package.versions) != package.installed
```
orndorffgrant, line 71: We also need to filter out all packages with upgrades that don't come from either esm-infra or esm-apps

ubuntu-advantage.1
```diff
@@ -92,6 +92,11 @@ of:
 whether this service would be available if this machine were attached.
 The possible values are \fIyes\fR or \fIno\fR.
 
+.TP
+.B update-info
+Show updates for packages in the system, including all available UA ESM
+related content.
```
orndorffgrant, line 98:
```suggestion
Show updates for packages in the system that come from UA ESM sources.
```

@renanrodrigo @lucasmoura @blackboxsw Am I correct that we don't want to show _all_ updates here. We only want to show updates that could come from esm-infra or esm-apps, right?

orndorffgrant (changes requested):
This is looking good! Just a few requests - main thing is that I believe we should only be displaying updates that come from either esm-infra or esm-apps.

uaclient/update_info.py
renanrodrigo: Oh this true - I was able to reproduce it working. Thanks for reminding me of this. I will update the code to use this approach.

uaclient/update_info.py
renanrodrigo: nice, will update

features/attached_commands.feature
```diff
@@ -611,3 +611,51 @@ Feature: Command behaviour when attached to an UA subscription
...
+        Then stdout matches regexp:
+        """
+        "entitled_services": \["esm-infra", "esm-apps"\]
+        """
+        When I run `apt-get install -y <test_pkg>=<outdated_version>` with sudo
+        And I run `ua update-info` as non-root
+        Then stdout matches regexp:
+        """
+        "package": "<test_pkg>"
+        """
```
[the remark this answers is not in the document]
lucasmoura: @renanrodrigo Let's move this test into the unattached features

uaclient/cli.py
```diff
@@ -350,6 +350,16 @@ def fix_parser(parser):
...
 
 
+def update_info_parser(parser):
+    """Build or extend an arg parser for update-info subcommand."""
+    parser.prog = "update-info"
+    parser.description = (
+        "Show updates for packages in the system, including all available UA"
+        " ESM related content."
+    )
+    return parser
```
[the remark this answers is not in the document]
renanrodrigo: As discussed, the help message for the missing `--format` is the way to go here.

PR #1843, opened 2021-10-19
chore: cache get_platform_info calls

renanrodrigo:
Per comment in #1839, cache the calls to get_platform_info.
We don't expect it to change often, and we call it here and there.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -268,8 +268,13 @@ class UAContractClient(serviceclient.UAServiceClient):
         if not machine_id:
             machine_id = util.get_machine_id(self.cfg)
         platform = util.get_platform_info()
-        arch = platform.pop("arch")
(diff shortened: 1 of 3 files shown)


lucasmoura (approved):

orndorffgrant (approved):
Thanks for figuring out that error - looks great!

uaclient/contract.py
```diff
@@ -268,8 +268,13 @@ def _get_platform_data(self, machine_id):
         if not machine_id:
             machine_id = util.get_machine_id(self.cfg)
         platform = util.get_platform_info()
-        arch = platform.pop("arch")
-        return {"machineId": machine_id, "architecture": arch, "os": platform}
+        platform_os = platform.copy()
+        arch = platform_os.pop("arch")
```
blackboxsw, line 272: ohh man. nice resolution here! Thanks for getting that.

PR #1845, opened 2021-10-20
proof of concept: build script for standalone distribution of ua

orndorffgrant:
***Do not merge***

This is a proof of concept for how we might build a standalone distribution of ua for use in docker containers.


orndorffgrant:
Closing as it seems unlikely we will need this. We can always find it later if we do need it.

PR #1851, opened 2021-10-27
Security status beta flag and schema version (SC-528)

orndorffgrant:
## ToDo

- [x] update integration tests
- [x] update unit tests

## Proposed Commit Message

```
security-status: add a schema version and required beta flag
```

## Test Steps

Verify that you can't run the command without the beta flag.
Verify that the schema version in the output is sensible. (is a string okay?)

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -625,13 +625,25 @@ Feature: Command behaviour when attached to an UA subscription
            | focal   |
 
     @series.xenial
+    @series.bionic
     @uses.config.machine_type.lxd.container
-    Scenario: Run security-status on an Ubuntu machine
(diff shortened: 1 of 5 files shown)


blackboxsw (approved):
+1 all good here. thanks for the integration coverage. 

Two things I'd like to see later as we iterate:
1. pretty format of json/yaml output by default  to aid in readability.
2.  integration tests might need to grow a new step such as "stdout is formatted as `{output_format}` with `{json_object_path}` value `{value}`. So we can avoid generic regex checks that could match other parts of the output 


Neither of these suggestions are needed for this beta inclusion

blackboxsw:
Unrelated CI timeout error on this.

PR #1853, opened 2021-10-29
Release 27.4

lucasmoura:
I have omitted some docs and tools changes, but let me know if we need to include them 

## TODO
- [] update with new commits as last few PRs land
- [x] add changelog
    - [] include LP references

## To Test/Review:
`git diff main` to make sure the contents match what we have in main

`git diff pkg/ubuntu/devel -- debian/changelog` to check the changelog diff

Review changelog using the checklist in release docs:
- [ ] Is the version correctly updated ? We must ensure that the new version on the changelog is correct and it also targets the latest Ubuntu release at the moment.
- [ ] Is the entry useful for the user ? The changelog entries should be user focused, meaning that we should only add entries that we think users will care about (i.e. we don't need entries when fixing a test, as this doesn't provide meaningful information to the user)
- [ ] Is this entry redundant ? Sometimes we may have changes that affect separate modules of the code. We should have an entry only for the module that was most affected by it
- [ ] Is the changelog entry unique ? We need to verify that the changelog entry is not already reflected in an earlier version of the changelog. If it is, we need not only to remove but double check the process we are using to cherry-pick the commits
- [ ] Is this entry actually reflected on the code ? Sometimes, we can have changelog entries that are not reflected in the code anymore. This can happen during development when we are still unsure about the behavior of a feature or when we fix a bug that removes the code that was added. We must verify each changelog entry that is added to be sure of their presence in the product.


debian/changelog
```diff
@@ -1,3 +1,29 @@
+ubuntu-advantage-tools (27.4~22.04.1) jammy; urgency=medium
+
+  * d/tools.postinst:
+    - hardcode python binary and PYTHONPATH
+    - undo unnecessary log file creation
+  * New upstream release 27.4
```
orndorffgrant, line 6: Should we include an entry like:
```
detach: notify contract server on detach
```

debian/changelog
```diff
@@ -1,3 +1,29 @@
...
+    - cc-eal: remove beta flag
+    - cli:
+      + detach won't ask to disable dependent service
```
orndorffgrant, line 9: Without context, this entry sounds like it is now skipping a needed step. Maybe change to something like this:
```suggestion
      + detach won't ask unnecessary questions
```

debian/changelog
```diff
@@ -1,3 +1,29 @@
...
+      + implement security-status endpoint
```
orndorffgrant, line 10: I think this deserves a bit more detail, something like:
```suggestion
      + new security-status subcommand lists potentially available security and ESM updates (beta)
```

debian/changelog
```diff
@@ -1,3 +1,29 @@
...
+    - contract: save machine id during attach operation
+    - fix:
+      + update return code of the command
```
orndorffgrant, line 13: Return code details can be useful to users IMO
```suggestion
      + exit 0 when fix is successfully applied and completed
      + exit 1 when fix cannot be applied
      + exit 2 when fix requires a reboot to complete
```

debian/changelog
```diff
@@ -1,3 +1,29 @@
...
+      + check reboot-required.pkgs to reboot
```
orndorffgrant, line 14: nit suggestion
```suggestion
      + check reboot-required.pkgs for better reboot suggestions
```

orndorffgrant (changes requested):
> I have omitted some docs and tools changes, but let me know if we need to include them

I think we should include all commits. Otherwise our main branch will drift from what we are actually releasing over time, making future rebases more difficult, and future CI less applicable to our releases.

We also need to update the version in `version.py`.

Other than that just a couple nits on the changelog :) Thanks for preparing this!

orndorffgrant:
I also just noticed that we don't have a test that enables fips-updates after livepatch is enabled. We only have enabling livepatch after fips-updates is enabled in a test. We should probably add livepatch-then-fips-updates before releaseing. I'll do this today

orndorffgrant:
I added a livepatch-then-fips-updates test in https://github.com/canonical/ubuntu-advantage-client/pull/185. Let's merge that and bring it into this release as well

lucasmoura:
@orndorffgrant You are right, I was trying to be concise on what should be included on the release, but I don't think it is worth the price of deviating from the main branch. I will bring those commits for this release

debian/changelog
lucasmoura: This is one I am not sure, because most of the time we are already doing that, only if the user runs an attach and immediately detaches we will be hitting that issue.
I think we should instead add the following entry:

```
attach: save machine-id during operation
```

What do you think ?

orndorffgrant: Good point! Yes I like your suggestion better

lucasmoura:
@orndorffgrant I think this branch can be reviewed again

orndorffgrant (approved):
Thanks @lucasmoura this looks good to me! I think we just need the LP references now

orndorffgrant:
Latest edits look good to me @lucasmoura ! 

I guess we might as well let jenkins finish before merging though

PR #1856, opened 2021-11-01
daemon: add daemon configuration and entrypoint (SC-551)

orndorffgrant:
## Proposed Commit Message

```
daemon: add daemon configuration and entrypoint
```

## Test Steps

Verify the daemon starts on boot and logs correctly.

Review all the systemd settings.

Review all places where I had to add log file configuration - did I miss anything?.

Verify systemctl start|stop|restart work as expected

## Desired commit type

## Checklist

### DIFF
--- a/systemd/ua.service
+++ b/systemd/ua.service
@@ -1,0 +1,21 @@
+# If you are uninterested in the (free for personal use) Ubuntu Advantage
+# services, including security updates after standard EOL and kernel patching
+# without rebooting, then you can safely stop and disable this service:
+# sudo systemctl stop ua.service
+# sudo systemctl disable ua.service
+
+[Unit]
+Description=Ubuntu Advantage Daemon
+Documentation=man:ubuntu-advantage https://ubuntu.com/advantage
+After=network.target network-online.target systemd-networkd.service ua-auto-attach.service
+
+[Service]
+# Xenial and Bionic systemd don't support Type=exec
+Type=notify
+NotifyAccess=main
+ExecStart=/usr/bin/python3 /usr/lib/ubuntu-advantage/daemon.py
+Restart=on-failure
+WorkingDirectory=/var/lib/ubuntu-advantage/
+
+[Install]
+WantedBy=multi-user.target
--- a/README.md
+++ b/README.md
@@ -140,6 +140,13 @@ by a string formatted as:
 
 `<protocol>://[<username>:<password>@]<fqdn>:<port>`
 
+### Daemon
+UA Client installs a daemon that starts on boot via systemd as `ua.service`.
+The service can be safely disabled if you have no interest in Ubuntu Advantage
+services.
+
+TODO fill in this section as we add functionality to the daemon.
+
 ### Timer jobs
 UA client sets up a systemd timer to run jobs that need to be executed recurrently.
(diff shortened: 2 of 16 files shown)


orndorffgrant:
Hi @sergiodj and @athos-ribeiro, If you have time we'd appreciate any feedback you have on how we're setting up our daemon to run: particularly with respect to the systemd configuration: https://github.com/canonical/ubuntu-advantage-client/pull/1856/files#diff-9aeef3fa54dfedf4b02027becff0a619ce9818e1b9717ca4f32e6e8ff90a80bc

We're also going to include you on the future daemon PRs where we slowly replace some our other systemd units and need to handle migrating in our postinst. :)

renanrodrigo (approved):
lgtm

lucasmoura (approved):

systemd/ua.service
```diff
@@ -0,0 +1,20 @@
+# If you are uninterested in the (free for personal use) Ubuntu Advantage
+# services, including security updates after standard EOL and kernel patching
+# without rebooting, then you can safely stop and disable this service:
+# sudo systemctl stop ua.service
+# sudo systemctl disable ua.service
+
+[Unit]
+Description=Ubuntu Advantage Daemon
+Documentation=man:ubuntu-advantage https://github.com/canonical/ubuntu-advantage-client/blob/main/README.md https://ubuntu.com/advantage
```
athos-ribeiro, line 9: Could pointing to the upstream, development version of the README file cause any sort of confusion for users throughout the development cycles?

systemd/ua.service
```diff
@@ -0,0 +1,20 @@
...
+After=network.target network-online.target systemd-networkd.service ua-auto-attach.service
+
+[Service]
+# Xenial and Bionic systemd don't support Type=exec
+Type=simple
```
athos-ribeiro, line 14: I wonder if using `notify` instead would be feasible here. If so, this would allow more control on failures, possible future service dependencies, etc. This would require checking if we could send notifications through all the versions of python-systemd all the way down to xenial though...

athos-ribeiro (approved):
LGTM!

I did add a comment/question on the possibility of using `Type=notify` instead of `simple` though.

systemd/ua.service
orndorffgrant: Good point. Let's leave it out for now. Ideally at some point in the future we would have better documentation, maybe with version specific links. But for now just pointing to the readme on `main` could likely be confusing

systemd/ua.service
orndorffgrant: From my testing it seems doable! I'll push up a version that does this

orndorffgrant:
Hi @athos-ribeiro I just pushed a change so that we're using `notify`. Let me know if this is what you meant or if you'd expect something to happen differently here. This is my first time using a `notify` service so all help is appreciated.

CC @renanrodrigo and @lucasmoura so you can see the changes too

athos-ribeiro:
LGTM

In the future, ensure that the notification is only sent to systemd when whatever the daemon is supposed to "start" is started (the placeholder comment may suggest otherwise).

orndorffgrant:
Good point! I'm keeping that in mind as I'm working on the next daemon-related PR, so I'll go ahead and merge this as-is

PR #1859, opened 2021-11-03
fix: include return code info in help

orndorffgrant:
## Proposed Commit Message

```
fix: include return code info in help
```

## Test Steps

Check if the new help output makes sense.

## Desired commit type

## Checklist

### DIFF
--- a/ubuntu-advantage.1
+++ b/ubuntu-advantage.1
@@ -54,6 +54,18 @@ Activate and configure this machine's access to an Ubuntu Advantage
 service.
 
 .TP
+.BR fix <security_issue>
+Fix a CVE or USN on the system by upgrading the appropriate package(s).
+
+<security_issue> can be any of the following formats: CVE-yyyy-nnnn,
+CVE-yyyy-nnnnnnn, or USN-nnnn-dd.
+
+The exit code can be 0, 1, or 2.
+    0: the fix was successfully applied
+    1: the fix cannot be applied
+    2: the fix was applied but requires a reboot before it takes effect
+
+.TP
 .B refresh
 Refresh contract and service details from Canonical.
 


uaclient/cli.py
```diff
@@ -347,7 +347,10 @@ def fix_parser(parser):
     parser.prog = "fix"
     parser.description = (
         "Inspect and resolve CVEs and USNs (Ubuntu Security Notices) on this"
-        " machine."
+        " machine. This could have one of three exit codes. 0 means that the"
```
lucasmoura, line 350: @orndorffgrant Maybe we could have that on a separate section, like:

```
Inspect and resolve CVEs and USNs (Ubuntu Security Notices) on this machine.

Return codes:
* 0:  means that the fix is successfully applied and completed.
* 1: means that the fix cannot be applied.
* 2:  means that the fix requires a reboot to complete.
```

What do you think ?

renanrodrigo (approved):
lgtm from a non-english speaker perspective

uaclient/cli.py
renanrodrigo: :heavy_plus_sign:

orndorffgrant: @lucasmoura @renanrodrigo 
Yeah I would prefer that too and I actually tried that at first. The problem is that argparser doesn't respect newlines in the description field.

You can set the formatter to RawTextHelpFormatter so that it respects newlines, but that also affects the flag descriptions and causes them not to wrap correctly. Specifically, we currently output:
```
positional arguments:
  security_issue  Security vulnerability ID to inspect and resolve on this
                  system. Format: CVE-yyyy-nnnn, CVE-yyyy-nnnnnnn or USN-nnnn-
                  dd
```

but the raw formatter would cause that to become

```
positional arguments:
  security_issue  Security vulnerability ID to inspect and resolve on this system. Format: CVE-yyyy-nnnn, CVE-yyyy-nnnnnnn or USN-nnnn-dd

```

We could do something hacky [like this person is suggesting](https://stackoverflow.com/questions/3853722/how-to-insert-newlines-on-argparse-help-text#answer-22157136) though: thoughts?

renanrodrigo: The suggestion looks way too hacky for a single use case as we have now, imho.
IMHO the lesser evil is to switch to the Raw Formatter, but just because I already see the `positional_arguments` text in a single line _on my machine_ ™

orndorffgrant: @renanrodrigo The default help formatter is smart about wrapping based on the width of the terminal. If you make your terminal less wide, then it will wrap nicely with an indent

lucasmoura: I think the proposed solution is not that bad. Adding it doesn't look hard and it will allow us to deliver a better user experience here. I am okay with using it

renanrodrigo: I see it now, sorry for the unsmart verification.
In this case, for consistency, I'd just go with the one-line text, putting as a list:
`Inspect and resolve CVEs and USNs (Ubuntu Security Notices) on this machine. The returns codes are: 0 if the fix is successfully applied and completed; 1 if the fix cannot be applied; 2 if the fix requires a reboot to complete.`

lucasmoura (approved):

ubuntu-advantage.1
```diff
@@ -53,6 +53,18 @@ Disable this machine's access to an Ubuntu Advantage service.
 Activate and configure this machine's access to an Ubuntu Advantage
 service.
 
+.TP
+.BR fix <security_issue>
```
blackboxsw, line 57: I 
```suggestion
.BR "fix" " <security_issue>"
```

PR #1861, opened 2021-11-04
fix: update manpage entry

lucasmoura:
Update fix manpage entry to add extra space in the title


orndorffgrant (approved):
Looks good thanks!

PR #1862, opened 2021-11-04
proof of concept: daemon to long poll gcp metadata with asyncio

orndorffgrant:
Demonstrating what it might look like to use asyncio instead of threads. Not as bad as I feared.

### DIFF
--- a/lib/daemon.py
+++ b/lib/daemon.py
@@ -1,20 +1,23 @@
+import asyncio
+import concurrent.futures
 import datetime
 import logging
 import sys
-import threading
 
 from uaclient import util
 from uaclient.cli import setup_logging
 from uaclient.clouds.identity import get_cloud_type
 
+read_url = "this is a hack"
 
-def gcp_auto_attach():
+
+async def gcp_auto_attach():
     # just a for loop in case I mess up, it'll only do whatever mistake 100 times
     for _i in range(100):
         now = datetime.datetime.now()
         logging.info("Requesting metadata at {}".format(now))
         try:
-            result = util.readurl(
+            result = await readurl(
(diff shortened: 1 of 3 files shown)


lib/daemon.py
```diff
@@ -0,0 +1,54 @@
...
+        print("just slept for 5 seconds")
+
+
+async def gcp_auto_attach():
+    # just a for loop in case I mess up, it'll only do whatever mistake 100 times
+    for _i in range(100):
+        now = datetime.datetime.now()
+        logging.info("Requesting metadata at {}".format(now))
+        try:
+            result = await asyncio.get_event_loop().run_in_executor(
```
lucasmoura, line 25: Isn't the `run_in_executor` a way for us to run the courotine through `ThreadPoolExecutor` ?
I need to read again about asyncio to be sure, but I just want to confirm that hypothesis

orndorffgrant: Right! It is for running anything long or compute intensive or blocking on a different thread (via a ThreadPoolExecutor) in a way that it can be used nicely with async code.

So in this instance, we don't want to call `readurl` directly in the coroutine, because it will block the thread (and therefore all other coroutines on the main thread). So we pass it to `run_in_executor` instead. Because we also pass `None` as the ThreadPoolExecutor argument here, the asyncio stdlib will create its own ThreadPoolExecutor and manage it internally. It then runs the `readurl` function on a thread in the threadpool and returns the result asynchronously. This is how we can use existing blocking code in an async function.

lucasmoura: Okay, that's a good point, However, I think only the messages jobs will not have a blocking IO. Therefore, the status, metering and gcp jobs will all need to run through a ThreadPoolExecutor right ?

If that is true, I don't think we gain much from the asyncio implementation here

PR #1863, opened 2021-11-05
fips: update integration tests for focal

lucasmoura:
## Proposed Commit Message
fips: update integration tests for focal

Initially, there was no strongswan-hmac FIPS package for Focal. Therefore, during the enable operation we state that this package could not be installed. Now that the package can be installed on a Focal machine, we are removing that message from the integration tests

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/features/enable_fips_cloud.feature
+++ b/features/enable_fips_cloud.feature
@@ -165,7 +165,6 @@ Feature: FIPS enablement in cloud based machines
             """
             Updating package lists
             Installing <fips-name> packages
-            <fips-name> strongswan-hmac package could not be installed
             <fips-name> enabled
             A reboot is required to complete install
             """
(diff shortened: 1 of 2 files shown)


renanrodrigo (changes requested):
Besides removing the lines, I think we need to check that `strongswan-hmac` is being installed from the FIPS source and that after disabling, the fips version keeps installed (like we do for the Xenial/Bionic tests).

lucasmoura:
@renanrodrigo good catch. I will add that check

lucasmoura:
@renanrodrigo updated

orndorffgrant (approved):
This worked for me, thanks!

renanrodrigo (approved):
LGTM

lucasmoura:

Issue #1860, opened 2021-11-04
Don't log tokens during apt-helper download-file


lucasmoura:
Avoid printing tokens when logs or errors include https://bearer:TOKEN@esm.ubuntu.com (e.g. when apt-helper download-file fails)

PR #1864, opened 2021-11-08
token redaction: redact token log if apt-helper download-file fails

CalvoM:
## Proposed Commit Message

token redaction: redact token log if apt-helper download-file fails.

This redacts the token logged during apt-helper failure e.g. when there the tokens are incorrect. 
It also corrects the path to the *machine-token.json* file in the README. 

Fixes: #1860

## Test Steps
Run the test_util test.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -58,7 +58,7 @@ def assert_valid_apt_credentials(repo_url, username, password):
     except util.ProcessExecutionError as e:
         if e.exit_code == 100:
             stderr = str(e.stderr).lower()
-            if re.search(r"401\s+unauthorized|httperror401", stderr):
+            if re.search(r"401\s+[uU]nauthorized|httperror401", stderr):
                 raise exceptions.UserFacingError(
                     "Invalid APT credentials provided for {}".format(repo_url)
                 )
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -58,7 +58,7 @@ def assert_valid_apt_credentials(repo_url, username, password):
     except util.ProcessExecutionError as e:
         if e.exit_code == 100:
             stderr = str(e.stderr).lower()
-            if re.search(r"401\s+[uU]nauthorized|httperror401", stderr):
+            if re.search(r"401\s+unauthorized|httperror401", stderr):
                 raise exceptions.UserFacingError(
                     "Invalid APT credentials provided for {}".format(repo_url)
                 )
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -667,7 +667,7 @@ def subp(
             if not retry_sleeps:
                 raise
             retry_msg = " Retrying %d more times." % len(retry_sleeps)
-            logging.debug(str(e) + retry_msg)
+            logging.debug(redact_sensitive_logs(str(e) + retry_msg))
             time.sleep(retry_sleeps.pop(0))
     return out, err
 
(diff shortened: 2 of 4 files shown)


uaclient/apt.py
```diff
@@ -58,7 +58,7 @@ def assert_valid_apt_credentials(repo_url, username, password):
     except util.ProcessExecutionError as e:
         if e.exit_code == 100:
             stderr = str(e.stderr).lower()
-            if re.search(r"401\s+unauthorized|httperror401", stderr):
+            if re.search(r"401\s+[uU]nauthorized|httperror401", stderr):
```
orndorffgrant, line 61: I wouldn't expect this to matter because we call `.lower()` on the output on the line above.

Did this change fix something?

orndorffgrant:
Thanks @CalvoM! This is looking really good!

Let's just rebase this to clean up the commit history. Something like this should work:

```
git rebase -i upstream/main
```
then
```
reword 55f5a9a3 correct path to machine-token.json file
reword 5f8232de correct some spelling and grammar errors.
reword ef9fd53f Cater for both U and u in errors.
reword 7d095b90 Redact from the failure from apt-helper.
fixup f9aea89d add test for apt-helper failure token redaction
fixup 82605dbb styling and blackening the code
```

Which will give you the opportunity to rename the commits. We try to include a prefix to commit messages to describe the scope of the change.
So the messages might end up looking like:
```
docs: correct path to machine-token.json file
docs: correct some spelling and grammar errors.
apt: Cater for both U and u in errors.
apt: Redact from the failure from apt-helper.
``` 

And then a `git push --force` after you're sure everything worked :)

(I apologize if you already knew how to do this, I just lean towards overexplaining and don't mean anything by it. :) )

uaclient/util.py
```diff
@@ -759,6 +759,7 @@ def is_config_value_true(config: Dict[str, Any], path_to_value: str):
     r"(\'token\': \')[^\']+",
     r"(\'X-aws-ec2-metadata-token\': \')[^\']+",
     r"(.*\[PUT\] response.*api/token,.*data: ).*",
+    r"(https://bearer:)[\w-]*",
```
blackboxsw, line 762: Let's make this as flexible as possible to avoid accidentally missing a redact.
```suggestion
    r"(https://bearer:)[^\@]+",
```

blackboxsw (changes requested):
Thanks for the PR @CalvoM looks great! One minor comment on the regex you use to make sure it is not too strict a match. Hopefully that sounds like a good suggestion.

~~Minor other git history suggestion:
 If you get a chance, can you use `git rebase -i HEAD~7` or some other means to collapse this set of commits into 2 separate commits, one for docs and one for the redact token.~~ Also please double check that the git config authorship you have on the commits is your canonical address instead of gmail.  [When I started at Canonical I was still accidentally using my @hp.com addresss for my first few branches until I noticed :/]


Once the commits are what you would prefer to land in the codebase, you can check the 
"Desired commit type"
"[ X ] Rebase and merge


Thanks again for putting this together.

OOPS. Didn't see @orndorffgrant's second review until after I clicked submit on my review. +1 on whatever path you would like to take on clean git history.
Thanks again!

uaclient/apt.py
CalvoM: Huh! I did not see the .lower() call, my bad!

orndorffgrant (approved):
Thanks for the updates! This looks great.

The jenkins failures seem like flakes unrelated to this PR - I was unable to reproduce them locally.

blackboxsw (approved):
Looking good thanks @CalvoM !

PR #1866, opened 2021-11-08
Show non-root user updated information in security-status

renanrodrigo:
A non-root user cannot read internals like `UAConfig.is_attached`, for instance.
Using the `.status` dict makes sure the same information surfaced to `ua status` reflects in the `security-status` output.

(Then, since we already have the status dict, we can check the services there as well)

## Test Steps
Run the related integration test

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -733,7 +733,7 @@ Feature: Command behaviour when attached to an UA subscription
         "status": "pending_attach"
         """
         When I attach `contract_token` with sudo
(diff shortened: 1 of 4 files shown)


orndorffgrant (approved):
This looks great thanks for noticing and fixing this.

Lets just let jenkins run

renanrodrigo:
Green Jenkins - must be my lucky day

lucasmoura (approved):

PR #1868, opened 2021-11-09
token redaction: redact token logged when enabling livepatch service (SC-582)

CalvoM:
## Proposed Commit Message

token redaction: redact token logged when enabling livepatch service

## Test Steps

Run the test_util test.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -596,6 +596,7 @@ def _subp(
         env.update(os.environ)
     if rcs is None:
         rcs = [0]
+    redacted_cmd = redact_sensitive_logs(" ".join(args))
     try:
         proc = subprocess.Popen(
             bytes_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=env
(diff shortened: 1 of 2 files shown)


orndorffgrant (approved):
The livepatch redaction is working well for me!

We'll merge #1864 before this one, then we can remove 8cefd93 from this PR and merge

uaclient/util.py
```diff
@@ -619,9 +619,10 @@ def _subp(
             stderr=err.decode("utf-8"),
         )
     if capture:
+        redacted_cmd = redact_sensitive_logs(" ".join(args))
```
blackboxsw, line 622: I think we actually need to pull this up above the [try: proc = subprocess.Popen](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/util.py#L605) so that we can make sure even if there are errors we don't leak that token in [line 613](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/util.py#L613) and  [line 616](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/util.py#L616)

blackboxsw (changes requested):
nice catch, I think we also need to source the redacted log info in the event of errors too.

uaclient/util.py
CalvoM: Oh yes, makes sense.

blackboxsw (approved):
Woo hoo! Looks good to me thanks @CalvoM

PR #1870, opened 2021-11-09
security-status must show all available updates (SC-580)

renanrodrigo:
There can be available security-updated versions which are not the latest.
Thus we check all the versions above the currently installed one.

## Test steps
- Run the modified unittests
- Spin a Xenial container, and:
  - Install `linux-libc-dev=4.4.0-21.37`
  - Run `ua security-status`
  - Verify that version `4.4.0-210.242` is available from the Security pocket, available immediately
  - Verify that version `4.4.0-210.250` is available from the ESM Infra pocket, pending attach

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/security_status.py
+++ b/uaclient/security_status.py
@@ -73,13 +73,11 @@ def filter_security_updates(
     ]
 
     return [
-        package
+        version
         for package in packages
-        if max(package.versions) > package.installed
-        and any(
-            origin.archive in security_repos
-            for origin in max(package.versions).origins
-        )
+        for version in package.versions
+        if version > package.installed
+        and any(origin.archive in security_repos for origin in version.origins)
     ]
 
 
@@ -122,20 +120,18 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
     installed_packages = [package for package in cache if package.is_installed]
     summary["num_installed_packages"] = len(installed_packages)
 
-    security_upgradable_packages = filter_security_updates(installed_packages)
+    security_upgradable_versions = filter_security_updates(installed_packages)
 
     package_count = {"esm-infra": 0, "esm-apps": 0, "standard-security": 0}
 
-    for package in security_upgradable_packages:
-        candidate = max(package.versions)
-        version = candidate.version
+    for candidate in security_upgradable_versions:
         service_name = get_service_name(candidate.origins)
         status = get_update_status(service_name, ua_info)
         package_count[service_name] += 1
         packages.append(
             {
-                "package": package.name,
-                "version": version,
+                "package": candidate.package.name,
+                "version": candidate.version,
                 "service_name": service_name,
                 "status": status,
             }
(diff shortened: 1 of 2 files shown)


uaclient/security_status.py
```diff
@@ -73,13 +73,11 @@ def filter_security_updates(
     ]
 
     return [
-        package
+        version
```
lucasmoura, line 76: I think we need to modify the or unittest to better cover this change, right ?

lucasmoura (commented):
@renanrodrigo do you a package in mind that we can double test to see if this fix solves the issue ?

uaclient/security_status.py
renanrodrigo: Hmm I was thinking about this.
Since we cannot use the (mocked) `apt` classes in unittests, this would add a bigger complexity to the package mocks.
I will check for a real package and add an integration test targeting it.

renanrodrigo: I mean, forget what I said:
the unittest showed as the best option to cover this. Updated the test steps with an example.

uaclient/tests/test_security_status.py
```diff
@@ -154,18 +163,30 @@ def test_finds_updates_for_installed_packages(
...
             ],
             "summary": {
                 "ua": {
                     "attached": False,
                     "enabled_services": [],
                     "entitled_services": [],
                 },
-                "num_installed_packages": 3,
-                "num_esm_infra_updates": 1,
+                "num_installed_packages": 5,
```
orndorffgrant, line 186: Shouldn't this be `4`?

renanrodrigo: Well no, I have 5 packages with an "installed" version in the test (from last revision, I added a package with a couple updates and another one representing a non-security update).

uaclient/tests/test_security_status.py
```diff
@@ -11,9 +11,8 @@
 M_PATH = "uaclient.security_status."
 
 
-def mock_package(
-    name, installed=None, candidate=None, archive=None, origin=None
-):
+# Candidates are a tuple of (version, archive, origin)
+def mock_package(name, installed=None, candidates=[]):
```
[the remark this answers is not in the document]
renanrodrigo: Yeah, I put the comment there exactly not to create too many structures just to mock the tests.
If you think it would be useful, then of course, can be done.

uaclient/tests/test_security_status.py
orndorffgrant: Ah you're right! Looks good then

uaclient/tests/test_security_status.py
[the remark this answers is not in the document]
orndorffgrant: Maybe a named tuple would be a bit much just for this, but how about something like this:
```suggestion
# Candidates are a tuple of (version, archive, origin)
def mock_package(name, installed=None, candidates: Tuple[str, str, str] = ()):
```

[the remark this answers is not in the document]
renanrodrigo: just updated

orndorffgrant (approved):
Thanks for this!

lucasmoura (approved):

Issue #1867, opened 2021-11-09
Attempting to attach when already attached doesn't return error code. (SC-573)


robert-ancell:
Attaching with an invalid token returns 1:
```
$ sudo ua attach x
Invalid token. See https://ubuntu.com/advantage
$ echo $?
1
```

Successfully attaching returns 0:
```
$ sudo ua attach <token>
This machine is now attached to '...'

SERVICE       ENTITLED  STATUS    DESCRIPTION
cis           yes       n/a       Center for Internet Security Audit Tools
esm-infra     yes       n/a       UA Infra: Extended Security Maintenance (ESM)
fips          yes       n/a       NIST-certified core packages
fips-updates  yes       n/a       NIST-certified core packages with priority security updates
livepatch     yes       n/a       Canonical Livepatch service
$ echo $?
0
```

Attempting to attach when already attached prints an error, but returns 0:
```
$ sudo ua attach <token>
This machine is already attached to '...'
To use a different subscription first run: sudo ua detach.
$ echo $?
0
```

I would expect the last case to have returned 1 (i.e. an error).

Because the last case returns 0, we don't currently detect this as an error when running the UA client for GUI applications.

orndorffgrant:
Thanks for the feedback @robert-ancell !

I think this was intentional because (in the human driven UX scenario), the user asked to be attached, and they are attached, so it's not a failure per se.

However, I think it would be good to distinguish that situation from a normal successful attach operation like you suggest.

I also don't want to return the same error code as "true" failure though.

What if we return `2` in the case where the machine is already attached? Would that work for you?
(cc @lucasmoura @renanrodrigo)

renanrodrigo:
I understand there is also a reason for it to be 0 on PRO machines in the past, but looking at it right now it makes sense for me to make it `2`.

robert-ancell:
Returning 2 works fine for me, I'm currently concluding an operation succeeded if the return value is 0. I'm not reading the stdout/stderr text, as that will be better resolved in #1812.

PR #1871, opened 2021-11-09
Return 2 when attaching an already attached machine (SC-573)

renanrodrigo:
No attach for the attached.

Consider attaching a machine that is already attached to an UA subscription as an error, and return `2` as a specific error code. More context in the bug.

Fixes: #1867

## Desired commit type

## Checklist

### DIFF
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -71,6 +71,12 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         Enabling default service esm-infra
         """
+        When I verify that running `ua attach contract_token` `with sudo` exits `2`
+        Then stderr matches regexp:
+        """
+        This machine is already attached to '.+'
+        To use a different subscription first run: sudo ua detach.
+        """
         When I run `ua disable esm-apps --assume-yes` with sudo
         When I append the following on uaclient config:
             """
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -213,7 +213,7 @@ Feature: Command behaviour when attached to an UA subscription
             """
             This command must be run as root \(try using sudo\).
             """
-        When I run `ua auto-attach` with sudo
+        When I verify that running `ua auto-attach` `with sudo` exits `2`
         Then stderr matches regexp:
             """
             This machine is already attached
--- a/features/ubuntu_pro.feature
+++ b/features/ubuntu_pro.feature
@@ -186,6 +186,21 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
             ros           +no  +(-|—) +Security Updates for the Robot Operating System
             ros-updates   +no  +(-|—) +All Updates for the Robot Operating System
             """
+        When I run `systemctl start ua-auto-attach.service` with sudo
+        And I verify that running `systemctl status ua-auto-attach.service` `as non-root` exits `0,3`
+        Then stdout matches regexp:
+        """
+        .*status=0\/SUCCESS.*
+        """
+        And stdout matches regexp:
+        """
+        Skipping attach: Instance '[0-9a-z\-]+' is already attached.
+        """
+        When I run `ua auto-attach` with sudo
+        Then stderr matches regexp:
+        """
+        Skipping attach: Instance '[0-9a-z\-]+' is already attached.
+        """
         When I run `apt-cache policy` with sudo
         Then apt-cache policy for the following url has permission `500`
         """
@@ -279,6 +294,21 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
             ros           +no  +(-|—) +Security Updates for the Robot Operating System
             ros-updates   +no  +(-|—) +All Updates for the Robot Operating System
             """
+        When I run `systemctl start ua-auto-attach.service` with sudo
+        And I verify that running `systemctl status ua-auto-attach.service` `as non-root` exits `0,3`
+        Then stdout matches regexp:
+        """
+        .*status=0\/SUCCESS.*
+        """
+        And stdout matches regexp:
+        """
+        Skipping attach: Instance '[0-9a-z\-]+' is already attached.
+        """
+        When I run `ua auto-attach` with sudo
+        Then stderr matches regexp:
+        """
+        Skipping attach: Instance '[0-9a-z\-]+' is already attached.
+        """
         When I run `apt-cache policy` with sudo
         Then apt-cache policy for the following url has permission `500`
         """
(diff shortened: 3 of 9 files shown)


uaclient/exceptions.py
```diff
@@ -38,7 +38,7 @@ class NonAutoAttachImageError(UserFacingError):
 class AlreadyAttachedError(UserFacingError):
     """An exception to be raised when a command needs an unattached system."""
 
-    exit_code = 0
+    exit_code = 2
```
blackboxsw, line 41: Note that just changing this exit code will result in breaking the status reported by `systemctl status ua-auto-attach.service` which blindly calls `ua auto-attach` on PRO images every boot. this will the leave the overall `systemctl status` in "State: degraded" which won't look so good.

We need to better handle auto-attach scenario in this PR.

blackboxsw (changes requested):
Thanks @renanrodrigo I think we need better logic in uaclient/cli.py maybe to either pre-flight check if we are already attached or handle the raised system.exit(2) and "pass" in this case.
Otherwise PRO images will be unhappy across reboot.

uaclient/exceptions.py
renanrodrigo: I see it, that's what I missed from PRO in the code comments.
I will tackle that, thanks for the pointer

renanrodrigo:
@blackboxsw I added a new commit addressing the valid PRO auto-attach path.

lucasmoura:
I can confirm that the PRO `auto-attach` unit does not fail with this change.

However, we have the following test that is now failing because `auto-attach` is returning 2:
```
Attached auto-attach in a ubuntu machine
```

renanrodrigo:
@lucasmoura updated, thanks for noticing

features/attached_commands.feature
```diff
@@ -203,6 +203,7 @@ Feature: Command behaviour when attached to an UA subscription
            | focal   | yes      | no     | yes | yes  | yes         | no  |
            | xenial  | yes      | yes    | yes | yes  | yes         | yes |
 
+    @wip
```
lucasmoura, line 206: We can remove this

features/attached_commands.feature
```diff
@@ -213,7 +214,7 @@ Feature: Command behaviour when attached to an UA subscription
             """
             This command must be run as root \(try using sudo\).
             """
-        When I run `ua auto-attach` with sudo
+        Then I verify that running `ua auto-attach` `with sudo` exits `2`
```
lucasmoura, line 217: Why did we change the `When` clause for the `Then` clause here ?

features/attached_commands.feature
renanrodrigo: Hmmm... I sent it wipped :|

features/attached_commands.feature
renanrodrigo: :man_facepalming:

renanrodrigo:
@lucasmoura I just sent the proper commit now

lucasmoura (approved):

features/attached_commands.feature
```diff
@@ -213,7 +213,7 @@ Feature: Command behaviour when attached to an UA subscription
             """
             This command must be run as root \(try using sudo\).
             """
-        When I run `ua auto-attach` with sudo
+        When I verify that running `ua auto-attach` `with sudo` exits `2`
         Then stderr matches regexp:
             """
             This machine is already attached
```
blackboxsw, line 219: Shouldn't this now be something like:
When I run `ua auto-attach` with sudo
Then stderr matches regexp:
"""
Skipping attach: Instance 'i-[0-9a-z]+' is already attached.
""""

blackboxsw (commented):
not certain on the integration test. But will let you and lucas sort it.

features/attached_commands.feature
renanrodrigo: In this specific case we are not in a PRO instance, so I think the return code should be `2`

renanrodrigo: @blackboxsw maybe it would be nice to add a new test step in the PRO tests, where we cover the case when it returns `0`. Will put that up.

renanrodrigo:
@blackboxsw @lucasmoura I added integration tests to make sure that we are not breaking any PRO workflow.

features/attached_commands.feature
blackboxsw: Thanks for the PRO coverage on this.

blackboxsw:
Looks like the last part of the PRO integration test is failing on this with non-zero exit code. 
https://jenkins.canonical.com/server-team-jenkins-private/blue/organizations/jenkins/uaclient-ci/detail/PR-1871/6/pipeline#step-123-log-198

renanrodrigo:
I cannot reproduce any errors locally.
Re-triggered Jenkins to see if the behavior persists.

renanrodrigo:
Well, it does persist.
I will figure out why.

renanrodrigo:
@lucasmoura thanks for the help identifying the error there - I sent a new version of the test which should ensure we are not failing.

features/ubuntu_pro.feature
```diff
@@ -279,6 +289,16 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
             ros           +no  +(-|—) +Security Updates for the Robot Operating System
             ros-updates   +no  +(-|—) +All Updates for the Robot Operating System
             """
+        When I run `systemctl start ua-auto-attach.service` with sudo
+        And I verify that running `systemctl status ua-auto-attach.service` `as non-root` exits `0,3`
```
lucasmoura, line 293: Shouldn't this always exit 0 ?

renanrodrigo: No. The service is not always running. If the service ran successfully but now is dead, the return code is 3.

features/attach_validtoken.feature
```diff
@@ -71,6 +71,12 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         Enabling default service esm-infra
         """
+        When I verify that running `ua attach contract_token` `with sudo` exits `2`
+        Then stderr matches regexp:
+        """
+        This machine is already attached to 'UA Client Test'
```
lucasmoura, line 77: I think we should use a regex for the subscription name here, just in case it changes in the future

features/ubuntu_pro.feature
```diff
@@ -186,6 +186,16 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
             ros           +no  +(-|—) +Security Updates for the Robot Operating System
             ros-updates   +no  +(-|—) +All Updates for the Robot Operating System
             """
+        When I run `systemctl start ua-auto-attach.service` with sudo
```
lucasmoura, line 189: I think we also add a seconf `ua auto-attach` step here, just to show the command failing in a PRO machine

renanrodrigo: oki, can do

renanrodrigo:
@lucasmoura all addressed

orndorffgrant:
This looks good to me! And jenkins is even happy :heavy_check_mark: 
I think we just need a quick glance from @blackboxsw to make sure we didn't miss any other corner cases here

lucasmoura (approved):

blackboxsw (approved):
Thanks for the fixes here. Just ran through it on PRO to confirm.

PR #1876, opened 2021-11-11
Fix status job release 27.4

lucasmoura:


blackboxsw (approved):
+1 on this for the moment. Might work on improvements for attached users in the future, but we can talk later on that aspect.

PR #1877, opened 2021-11-12
purge: Remove all gpg files during apt purge (SC-537)

CalvoM:
## Proposed Commit Message

purge: remove all the gpg files during apt purge

Now we can delete the gpg files in **/etc/apt/trusted.gpg.d/** folder belonging to ubuntu-advantage-tools.

## Desired commit type

## Checklist

### DIFF
--- a/debian/ubuntu-advantage-tools.postrm
+++ b/debian/ubuntu-advantage-tools.postrm
@@ -19,9 +19,7 @@ remove_logs(){
 }
 
 remove_gpg_files(){
-    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
-    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-apps.gpg
-    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-fips.gpg
+    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-*.gpg
 }
 
 case "$1" in


debian/ubuntu-advantage-tools.postrm
```diff
@@ -19,9 +19,7 @@ remove_logs(){
 }
 
 remove_gpg_files(){
-    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-infra-trusty.gpg
-    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-esm-apps.gpg
-    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-fips.gpg
+    rm -f /etc/apt/trusted.gpg.d/ubuntu-advantage-*.gpg
```
renanrodrigo, line 22: The only problem here is removing files we should not. Given that, I think that no other APT repository should have `ubuntu-advantage` as a prefix for the keys, and that it is safe and reasonable for us to remove it like this.

renanrodrigo (approved):
Thanks, @CalvoM! LGTM.
Please just remove the unrelated commit from this branch, and we are good.

debian/ubuntu-advantage-tools.postrm
lucasmoura: I agree with that approach, but let's just confirm with @blackboxsw to see if this is really safe

blackboxsw (approved):
I agree here. Our GPG keys are scoped with the "ubuntu-advantage-" prefix and other software shouldn't be using this prefix as they risk being overwritten by our own utilities. +1

renanrodrigo:
CI failure unrelated - landing this

PR #1881, opened 2021-11-16
fips: avoid writing to real file when testing

CalvoM:
## Proposed Commit Message

fips: avoid writing to real file when testing

Tests were failing when some services when enabled and mainly because they, tests, are writing to a real life i.e. **/etc/apt/auth.conf.d/90ubuntu-advantage** resulting to **PermissionError**

## Test Steps

Enable services e.g. esm-infra or cis and run the tests. 
## Desired commit type

## Checklist

### DIFF
--- a/uaclient/entitlements/tests/test_fips.py
+++ b/uaclient/entitlements/tests/test_fips.py
@@ -936,8 +936,9 @@ class TestFipsEntitlementInstallPackages:
         self, m_run_apt, entitlement
     ):
         m_run_apt.side_effect = exceptions.UserFacingError("error")
-        with pytest.raises(exceptions.UserFacingError):
-            entitlement.install_packages()
+        with mock.patch.object(entitlement, "_cleanup"):
+            with pytest.raises(exceptions.UserFacingError):
+                entitlement.install_packages()
 
     @mock.patch(M_PATH + "apt.get_installed_packages")
     @mock.patch(M_PATH + "apt.run_apt_command")


uaclient/entitlements/tests/test_fips.py
```diff
@@ -927,10 +927,12 @@ def test_fips_does_not_show_enabled_when_fips_updates_is(
...
     @mock.patch(M_PATH + "apt.run_apt_command")
+    @mock.patch(M_PATH + "apt.remove_repo_from_apt_auth_file")
     def test_install_packages_fail_if_metapackage_not_installed(
-        self, m_run_apt, entitlement
+        self, m_remove_apt, m_run_apt, entitlement
     ):
         m_run_apt.side_effect = exceptions.UserFacingError("error")
+        m_remove_apt.side_effect = None
         with pytest.raises(exceptions.UserFacingError):
             entitlement.install_packages()
```
orndorffgrant, line 937: Lets mock the `_cleanup` method on the entitlement directly. Instead of the `apt` function which is an implementation detail of the `_cleanup` method. Something like this:
```python
    @mock.patch(M_PATH + "apt.run_apt_command")
    def test_install_packages_fail_if_metapackage_not_installed(
        self, m_run_apt, entitlement
    ):
        m_run_apt.side_effect = exceptions.UserFacingError("error")
        with mock.patch.object(
            entitlement,
            "_cleanup",
        ):
            with pytest.raises(exceptions.UserFacingError):
                entitlement.install_packages()
```

orndorffgrant (changes requested):
Really good find! Just a suggestion on mocking the entitlement method instead of the apt function. Let me know if that doesn't work

uaclient/entitlements/tests/test_fips.py
CalvoM: This seems a better approach. Tested it and works. Updating right now.

renanrodrigo (approved):
lgtm

lucasmoura (approved):

orndorffgrant (approved):
Thanks for the update! Merging now

PR #1885, opened 2021-11-18
status: show only entitlements which exist in the server

renanrodrigo:
If an entitlement does not exist in the Contracts Server, then we don't need to surface it.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -630,18 +630,18 @@ class UAConfig:
         response["version"] = version.get_version(features=self.features)
 
         resources = get_available_resources(self)
-        for resource in sorted(resources, key=lambda x: x["name"]):
-            if resource["available"]:
+        for resource in sorted(resources, key=lambda x: x.get("name", "")):
+            if resource.get("available"):
                 available = status.UserFacingAvailability.AVAILABLE.value
             else:
                 available = status.UserFacingAvailability.UNAVAILABLE.value
-            ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(resource["name"])
+            ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(resource.get("name", ""))
 
             if not ent_cls:
                 LOG.debug(
                     "Ignoring availability of unknown service %s"
                     " from contract server",
-                    resource["name"],
+                    resource.get("name", "without a 'name' key"),
                 )
                 continue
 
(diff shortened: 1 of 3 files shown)


uaclient/config.py
```diff
@@ -729,11 +729,13 @@ def _attached_status(self) -> Dict[str, Any]:
             if not resource["available"]
         }
 
-        for ent_cls in ENTITLEMENT_CLASSES:
-            ent = ent_cls(self)
-            response["services"].append(
-                self._attached_service_status(ent, inapplicable_resources)
-            )
+        for resource in sorted(resources, key=lambda x: x["name"]):
```
lucasmoura, line 732: Why the resources need to be sorted ?

uaclient/tests/test_cli_status.py
```diff
@@ -19,10 +19,23 @@
 RESPONSE_AVAILABLE_SERVICES = [
```
lucasmoura, line 19: Shouldn't we also have a test where we have a service that is not present on the contracts response ?

renanrodrigo: Good point. Will remove one of those

uaclient/config.py
renanrodrigo: We currently rely on it to show them in order when presenting the tabular output

lucasmoura (approved):
Let's just wait on CI to see if everything is fine

uaclient/config.py
lucasmoura: Sorry, just one more thing. Let's use `get` here to access the resource name. It is unlikely that we will receive a resource without that key, but we could be defensive anyway

lucasmoura (changes requested):

uaclient/config.py
renanrodrigo: Hmm, I would like to keep this as-is, for:
a) we do want to raise an error in this case: a resource without name makes no sense, and no default applies if we don't find one
b) mypy complains that we use it, and we need to do some type juggling to avoid those same errors of a)

renanrodrigo:
I'm checking integration tests

uaclient/config.py
lucasmoura: We could just use `get("name", "")` in that way we would not break if the resource doesn't have a name and we would not break either by trying to access the entitlement on the dict

renanrodrigo: oki, done

lucasmoura (approved):
Let's just wait CI on this one before merging it

renanrodrigo:
According to an elder saying from my people, "Green Jenkins is a sign of good luck" :four_leaf_clover:

orndorffgrant (approved):
LGTM thanks!

PR #1886, opened 2021-11-18
tests: add dpkg-query check for version test

lucasmoura:
## Proposed Commit Message
tests: add dpkg-query check for version test

Add dpkg-query check to verify the package version.
We cannot compare if the ua version is equal to the dpkg-query output because of our config overrides on some tests (PRO instances)

However, we now also don't need to provide the package version if the config overrides details on it, which I think is an improvement

## Desired commit type

## Checklist

### DIFF
--- a/features/_version.feature
+++ b/features/_version.feature
@@ -12,10 +12,18 @@ Feature: UA is expected version
     @uses.config.machine_type.gcp.pro
     Scenario Outline: Check ua version
         Given a `<release>` machine with ubuntu-advantage-tools installed
-        When I run `ua version` with sudo
-        Then I will see the following on stdout
+        When I run `dpkg-query --showformat='${Version}' --show ubuntu-advantage-tools` with sudo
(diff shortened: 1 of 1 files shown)


renanrodrigo:
Currently we pass `UACLIENT_BEHAVE_CHECK_VERSION` as `27.4~20.04.1` (for example), but the `dpkg-query` command returns only `27.4` part.
If we pass just the `27.4` to the variable to match the test, aren't we actually removing the ability to test the part after the `~` ?

lucasmoura:
@renanrodrigo Are you seeing only the `27.4` string in the output of `dpkg-query` ? When running the tests I was passing the full version `27.4.1~16.04.1` and it was working.

But let me know if there is an issue here and I can double check this

renanrodrigo (approved):
Found my mistake.
This lgtm

blackboxsw (approved):
LGTM!

Issue #1889, opened 2021-11-20
ua collect-logs should run each logfile through util.redact_sensitive_logs


blackboxsw:
Because old versions of UA client may not have redacted tokens in previous log entries. We should run each log through util.redact_sensitive_logs to ensure the compressed logs do not have any leaked tokens.

supplemental request:
 add /etc/cloud/build.info to the list of logs we grab as it gives info about the cloud image build date if applicable.

PR #1890, opened 2021-11-22
logs: redact older unredacted logs

CalvoM:
## Proposed Commit Message

logs: redact older unredacted logs

Fixes: #1889 

## Test Steps

- Run **ua collect-logs** with current ua tool. Get the compressed logs.
- Build and install this update and then run **ua collect-logs** and collect the logs.
- Compare the logs esp. the ubuntu-advantage.log files for any sensitive data like tokens

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -31,7 +31,11 @@ from uaclient import (
 from uaclient import status as ua_status
 from uaclient import util, version
 from uaclient.clouds import identity
-from uaclient.defaults import CONFIG_FIELD_ENVVAR_ALLOWLIST
+from uaclient.defaults import (
+    CLOUD_BUILD_INFO,
+    CONFIG_FIELD_ENVVAR_ALLOWLIST,
+    DEFAULT_CONFIG_FILE,
+)
 
 NAME = "ua"
 
@@ -1168,12 +1172,13 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
             )
 
         ua_logs = (
-            cfg.cfg_path or "/etc/ubuntu-advantage/uaclient.conf",
+            cfg.cfg_path or DEFAULT_CONFIG_FILE,
             cfg.log_file,
             cfg.timer_log_file,
             cfg.license_check_log_file,
             cfg.daemon_log_file,
             cfg.data_path("jobs-status"),
+            CLOUD_BUILD_INFO,
             *(
                 entitlement.repo_list_file_tmpl.format(name=entitlement.name)
                 for entitlement in entitlements.ENTITLEMENT_CLASSES
@@ -1183,13 +1188,9 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
 
         for log in ua_logs:
             if os.path.isfile(log):
-                print(log)
-                shutil.copy(log, output_dir)
-
-        extra_logs = ("/etc/cloud/build.info",)
-
-        for log in extra_logs:
-            if os.path.isfile(log):
+                log_content = util.load_file(log)
+                log_content = util.redact_sensitive_logs(log_content)
+                util.write_file(log, log_content)
                 shutil.copy(log, output_dir)
 
         with tarfile.open(output_file, "w:gz") as results:
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1183,6 +1183,13 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
 
         for log in ua_logs:
             if os.path.isfile(log):
+                print(log)
+                shutil.copy(log, output_dir)
+
+        extra_logs = ("/etc/cloud/build.info",)
+
+        for log in extra_logs:
+            if os.path.isfile(log):
                 shutil.copy(log, output_dir)
 
         with tarfile.open(output_file, "w:gz") as results:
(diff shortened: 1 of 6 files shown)


uaclient/cli.py
```diff
@@ -1181,8 +1181,17 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
             ),
         )
 
-        for log in ua_logs:
+        extra_logs = ("/etc/cloud/build.info",)
```
lucasmoura, line 1184: Let's add this extra log into the `ua_logs` variable.
Also, let's create a constant for it on `defaults.py`

lucasmoura: Since you are touching this file, let's also replace `/etc/ubuntu-advantage/uaclient.conf` on that variable to `DEFAULT_CONFIG_FILE` that is also present on `defaults.py`

uaclient/cli.py
```diff
@@ -1181,8 +1181,17 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
...
+        logs = ua_logs + extra_logs
+
+        for log in logs:
             if os.path.isfile(log):
+                with open(log, "r+") as f:
```
lucasmoura, line 1189: Let's use here the `load_file` function that is present on the util module

uaclient/cli.py
```diff
@@ -1181,8 +1181,17 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
...
+                    content = f.read()
+                    if len(content) > 0:
+                        f.seek(0)
+                        f.truncate()
+                        f.write(util.redact_sensitive_logs(content))
```
lucasmoura, line 1194: I think if we just run write directly we will already overwrite the content of the file, right ? Or I am missing something ?

uaclient/cli.py
CalvoM: On it, I was not sure to separate the extra logs or not. But having the paths in defaults.py will be better to maintain.

uaclient/cli.py
```diff
@@ -1183,6 +1188,8 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
 
         for log in ua_logs:
             if os.path.isfile(log):
+                log_content = util.load_file(log)
+                util.write_file(log, log_content)
```
lucasmoura, line 1192: I think we need to redact the logs when writing the file, right ?

lucasmoura (commented):
Let's also add this new entry into our integration tests:
https://github.com/canonical/ubuntu-advantage-client/blob/main/features/attached_commands.feature#L594

And here:
https://github.com/canonical/ubuntu-advantage-client/blob/main/features/unattached_commands.feature#L504

uaclient/util.py
```diff
@@ -701,9 +701,9 @@ def write_file(filename: str, content: str, mode: int = 0o644) -> None:
     @param filename: The full path of the file to write.
     @param content: The content to write to the file.
     @param mode: The filesystem mode to set on the file.
-    @param omode: The open mode used when opening the file (w, wb, a, etc.)
     """
     logging.debug("Writing file: %s", filename)
+    content = redact_sensitive_logs(content)
```
lucasmoura, line 706: I think this is a good approach, since we will guarantee that we will always redact sensitive information.
But let's verify with the rest of the team @renanrodrigo @orndorffgrant @blackboxsw

orndorffgrant: This is a good idea, but I think it is too risky.

We often need to write files that _do_ contain secrets (apt auth files and machine-token.json at least), and they may happen to match one of the redaction-regexes. In that case, we could fail to authenticate to an apt repo because we accidentally wrote "<REDACTED>" as the password.

I think to avoid this possibility, we should call `redact_sensitive_logs` over in `action_collect_logs` rather than here for this PR.

uaclient/tests/test_cli_collect_logs.py
```diff
@@ -51,6 +51,7 @@ def test_collect_logs_help(self, _getuid, capsys):
 
     @mock.patch(M_PATH + "tarfile.open")
     # let's pretend all files exist
+    @mock.patch("builtins.open", create=True)
     @mock.patch(M_PATH + "os.path.isfile", return_value=True)
```
orndorffgrant, line 55: let's make sure the comment stays in the right place
```suggestion
    @mock.patch("builtins.open", create=True)
    # let's pretend all files exist
    @mock.patch(M_PATH + "os.path.isfile", return_value=True)
```

orndorffgrant (dismissed):
This looks good. Unfortunately, I don't think we can call `redact_sensitive_logs` in `write_file` though :/

lucasmoura (approved):

uaclient/tests/test_cli_collect_logs.py
```diff
@@ -121,7 +125,7 @@ def test_collect_logs(
             mock.call(["systemctl", "status", "ua.service"], rcs=[0, 3]),
         ]
 
-        assert m_copy.call_count == 14
+        assert m_copy.call_count == 15
```
renanrodrigo, line 128: Could we just add here an assertion that `redact_sensitive_logs` is being called as many times as `copy`?

renanrodrigo (approved):
LGTM
CI errors are unrelated

PR #1894, opened 2021-11-24
refactor: decouple cloud factory from cli

orndorffgrant:
## Notes

- [ ] #1894 (this one)
- [ ] #1895 
- [ ] #1896 
- [ ] #1887

## Proposed Commit Message

```
refactor: decouple cloud factory from cli
```

## Test Steps

Ensure auto-attach still works correctly

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1056,12 +1056,28 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
     """
     try:
         instance = identity.cloud_instance_factory()
-    except exceptions.UserFacingError as e:
+    except exceptions.CloudFactoryError as e:
         if cfg.is_attached:
             # We are attached on non-Pro Image, just report already attached
             raise exceptions.AlreadyAttachedError(cfg)
-        # Unattached on non-Pro return UserFacing error msg details
-        raise e
+        if isinstance(e, exceptions.CloudFactoryNoCloudError):
+            raise exceptions.UserFacingError(
+                ua_status.MESSAGE_UNABLE_TO_DETERMINE_CLOUD_TYPE
+            )
+        if isinstance(e, exceptions.CloudFactoryNonViableCloudError):
+            raise exceptions.UserFacingError(
+                ua_status.MESSAGE_UNSUPPORTED_AUTO_ATTACH
+            )
+        if isinstance(e, exceptions.CloudFactoryUnsupportedCloudError):
+            raise exceptions.NonAutoAttachImageError(
+                ua_status.MESSAGE_UNSUPPORTED_AUTO_ATTACH_CLOUD_TYPE.format(
+                    cloud_type=e.cloud_type
+                )
+            )
+        # we shouldn't get here, but this is a reasonable default just in case
+        raise exceptions.UserFacingError(
+            ua_status.MESSAGE_UNABLE_TO_DETERMINE_CLOUD_TYPE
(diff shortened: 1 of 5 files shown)


uaclient/clouds/identity.py
```diff
@@ -49,7 +49,33 @@ def get_cloud_type() -> Tuple[Optional[str], Optional[NoCloudTypeReason]]:
...
+class CloudFactoryNoCloudError(CloudFactoryError):
+    pass
+
+
+class CloudFactoryUnsupportedCloudError(CloudFactoryError):
+    pass
+
+
+class CloudFactoryNonViableCloudError(CloudFactoryError):
+    pass
```
lucasmoura, line 66: Maybe we should move those into the `exceptions` module ?

uaclient/cli.py
```diff
@@ -1052,12 +1052,28 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
...
+            raise exceptions.UserFacingError(
+                ua_status.MESSAGE_UNSUPPORTED_AUTO_ATTACH
+            )
+        if isinstance(e, identity.CloudFactoryUnsupportedCloudError):
+            raise exceptions.NonAutoAttachImageError(
+                ua_status.MESSAGE_UNSUPPORTED_AUTO_ATTACH_CLOUD_TYPE.format(
+                    cloud_type=e.cloud_type
+                )
+            )
+        # we shuoldn't get here, but this is a reasonable default just in case
```
lucasmoura, line 1073: s/shuoldn't/shouldn't/

uaclient/clouds/identity.py
```diff
@@ -49,7 +49,33 @@ def get_cloud_type() -> Tuple[Optional[str], Optional[NoCloudTypeReason]]:
     return (None, NoCloudTypeReason.NO_CLOUD_DETECTED)
 
 
+class CloudFactoryError(Exception):
```
lucasmoura, line 52: Can't we inherit from `UserFacingError` and bake in the messages into each exception ?
In that way we would avoid the if checks on `_get_contract_token_from_cloud_identity` and we would only need to re-raise the exceptions there

uaclient/clouds/identity.py
orndorffgrant: Good point, will do

uaclient/cli.py
orndorffgrant: Oops! Will fix

uaclient/clouds/identity.py
orndorffgrant: The purpose of this PR was to separate the exceptions thrown by the factory function from the UserFacingErrors. This allows us to interpret the Errors differently based on where we're calling the factory function. It also keeps the CLI-specific details (the messages) in the `cli` module, instead of buried in the `clouds.identity` module.

I don't mind the if checks on the type of CloudFactoryError because it makes it really obvious which situation causes which message.

I'm open to further discussion though :)

orndorffgrant:
@lucasmoura updated!

uaclient/clouds/identity.py
lucasmoura: That's a good point. No problem, we can keep the implementation as is

lucasmoura (approved):

renanrodrigo (approved):
@orndorffgrant LGTM, throw it in!

PR #1895, opened 2021-11-24
refactor: decouple lock from cli

orndorffgrant:
## Notes

- [x] #1894
- [ ] #1895 (this one)
- [ ] #1896 
- [ ] #1887

## Proposed Commit Message

```
refactor: decouple lock from cli
```

## Test Steps

Ensure all locking cli commands still work and correctly conflict with each other.

Ensure trying to run an attach while a pro instance is auto-attaching fails

## Desired commit type

## Checklist

### DIFF
--- a/lib/reboot_cmds.py
+++ b/lib/reboot_cmds.py
@@ -17,16 +17,16 @@ should run at next boot to process any pending/unresovled config operations.
 import logging
 import os
 import sys
-import time
 
-from uaclient import config, contract, entitlements, status
-from uaclient.cli import assert_lock_file, setup_logging
+from uaclient import config, contract, entitlements, lock, status
+from uaclient.cli import setup_logging
 from uaclient.exceptions import LockHeldError, UserFacingError
 from uaclient.util import ProcessExecutionError, UrlError, subp
 
 # Retry sleep backoff algorithm if lock is held.
 # Lock may be held by auto-attach on systems with ubuntu-advantage-pro.
-SLEEP_RETRIES_ON_LOCK_HELD = [1, 1, 5]
+SLEEP_ON_LOCK_HELD = 1
+MAX_RETRIES_ON_LOCK_HELD = 7
 
 
 def run_command(cmd, cfg):
@@ -102,7 +102,6 @@ def process_remaining_deltas(cfg):
     cfg.remove_notice("", status.MESSAGE_LIVEPATCH_LTS_REBOOT_REQUIRED)
 
 
-@assert_lock_file("ua-reboot-cmds")
 def process_reboot_operations(cfg):
 
     reboot_cmd_marker_file = cfg.data_path("marker-reboot-cmds")
@@ -139,21 +138,17 @@ def main(cfg):
     :raises: LockHeldError when lock still held by auto-attach after retries.
              UserFacingError for all other errors
     """
-    while True:
-        try:
+    try:
+        with lock.SpinLock(
+            cfg=cfg,
+            lock_holder="ua-reboot-cmds",
+            sleep_time=SLEEP_ON_LOCK_HELD,
+            max_retries=MAX_RETRIES_ON_LOCK_HELD,
+        ):
             process_reboot_operations(cfg=cfg)
-            break
-        except LockHeldError as e:
-            logging.debug(
-                "Retrying ua-reboot-cmds {} times on held lock".format(
-                    len(SLEEP_RETRIES_ON_LOCK_HELD)
-                )
-            )
-            if SLEEP_RETRIES_ON_LOCK_HELD:
-                time.sleep(SLEEP_RETRIES_ON_LOCK_HELD.pop(0))
-            else:
-                logging.warning("Lock not released. %s", str(e.msg))
-                sys.exit(1)
+    except LockHeldError as e:
+        logging.warning("Lock not released. %s", str(e.msg))
+        sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/uaclient/lock.py
+++ b/uaclient/lock.py
@@ -1,0 +1,106 @@
+import functools
+import logging
+import os
+import time
+
+from uaclient import config, exceptions
+
+LOG = logging.getLogger("ua.lock")
+
+# Set a module-level callable here so we don't have to reinstantiate
+# UAConfig in order to determine dynamic data_path exception handling of
+# main_error_handler
+clear_lock_file = None
+
+
+def clear_lock_file_if_present():
+    global clear_lock_file
+    if clear_lock_file:
+        clear_lock_file()
+
+
+class SingleAttemptLock:
+    """
+    Context manager for gaining exclusive access to the lock file.
+    Create a lock file if absent. The lock file will contain a pid of the
+    running process, and a customer-visible description of the lock holder.
+
+    :param lock_holder: String with the service name or command which is
+        holding the lock. This lock_holder string will be customer visible in
+        status.json.
+    :raises: LockHeldError if lock is held.
+    """
+
+    def __init__(self, *_args, cfg: config.UAConfig, lock_holder: str):
+        self.cfg = cfg
+        self.lock_holder = lock_holder
+
+    def __enter__(self):
+        global clear_lock_file
+        (lock_pid, cur_lock_holder) = self.cfg.check_lock_info()
+        if lock_pid > 0:
+            raise exceptions.LockHeldError(
+                lock_request=self.lock_holder,
+                lock_holder=cur_lock_holder,
+                pid=lock_pid,
+            )
+        self.cfg.write_cache(
+            "lock", "{}:{}".format(os.getpid(), self.lock_holder)
+        )
+        notice_msg = "Operation in progress: {}".format(self.lock_holder)
+        self.cfg.add_notice("", notice_msg)
+        clear_lock_file = functools.partial(self.cfg.delete_cache_key, "lock")
+
+    def __exit__(self, _exc_type, _exc_value, _traceback):
+        global clear_lock_file
+        self.cfg.delete_cache_key("lock")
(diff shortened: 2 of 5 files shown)


uaclient/lock.py
```diff
@@ -0,0 +1,89 @@
...
+    global clear_lock_file
+    if clear_lock_file:
+        clear_lock_file()
+
+
+class SingleAttemptLock:
+    """
+    Context manager for gaining exclusive access to the lock file.
+    Create a lock file if absent. The lock file will contain a pid of the
+        running process, and a customer-visible description of the lock holder.
```
lucasmoura, line 26: I think we can remove this extra space here

uaclient/lock.py
```diff
@@ -0,0 +1,89 @@
...
+    :param lock_holder: String with the service name or command which is
+        holding the lock.
```
lucasmoura, line 28: I think we can remove the new line here

uaclient/lock.py
```diff
@@ -0,0 +1,89 @@
...
+        self.max_retries = max_retries
+
+    def __enter__(self):
+        LOG.debug("spin lock starting for {}".format(self.lock_holder))
+        tries = 0
+        done = False
+        while not done:
+            try:
+                super().__enter__()
+                done = True
```
lucasmoura, line 78: I think we can just use a break here instead

lib/reboot_cmds.py
```diff
@@ -139,21 +138,14 @@ def main(cfg):
     :raises: LockHeldError when lock still held by auto-attach after retries.
              UserFacingError for all other errors
     """
-    while True:
-        try:
+    try:
+        with lock.SpinLock(
```
lucasmoura, line 142: What do you think if we used named parameters for the locks ? Just to make reading the code easier and just to not rely on the parameter order to be correct for every call

uaclient/lock.py
```diff
@@ -0,0 +1,89 @@
...
+        self.cfg.add_notice("", notice_msg)
+        clear_lock_file = functools.partial(self.cfg.delete_cache_key, "lock")
+
+    def __exit__(self, _exc_type, _exc_value, _traceback):
+        global clear_lock_file
+        self.cfg.delete_cache_key("lock")
+        clear_lock_file = None  # Unset due to successful lock delete
+
+
+class SpinLock(SingleAttemptLock):
```
lucasmoura, line 59: Even though I think this is a good name for the class and its intended purpose, I think we should add docstrings for it

uaclient/tests/test_lock.py
```diff
@@ -0,0 +1,219 @@
...
+
+        assert mock.sentinel.success == ret
+        assert [
+            mock.call("lock", "123:some operation")
+        ] == m_write_cache.call_args_list
+        lock_msg = "Operation in progress: some operation"
+        assert [mock.call("", lock_msg)] == m_add_notice.call_args_list
+        assert [mock.call("lock")] == m_delete_cache_key.call_args_list
+
+    def test_creates__and_releases_lock_when_error_occurs(
```
lucasmoura, line 43: I think we can remove the extra `_` between `creates` and `and`

uaclient/tests/test_lock.py
```diff
@@ -0,0 +1,219 @@
...
+            with SingleAttemptLock(cfg, "some operation"):
+                test_function(arg, cfg=cfg)
+
+        assert exc.value.lock_holder == "held"
+        assert exc.value.lock_holder_pid == 10
+        assert exc.value.lock_request == "some operation"
+
+        assert [
+            mock.call("lock", "123:some operation")
+        ] != m_write_cache.call_args_list
```
lucasmoura, line 96: Should we also assert that `m_write_cache` `call_args_list` is empty ?

uaclient/tests/test_lock.py
```diff
@@ -0,0 +1,219 @@
...
+        lock_msg = "Operation in progress: some operation"
+        assert [mock.call("", lock_msg)] != m_add_notice.call_args_list
+        assert [mock.call("lock")] != m_delete_cache_key.call_args_list
+
+
+class TestSpinLock:
```
lucasmoura, line 102: The first two tests of `TestSpinLock`  are pretty similar to the ones one `TestSingleAttemptLock`
What do you think we have a common lock test class that we can parametrize which lock class we are using ?

We can that just use the `TestSpinLock` for the specific functionality of that class

uaclient/lock.py
orndorffgrant: Yep my bad

uaclient/lock.py
orndorffgrant: good point!

lib/reboot_cmds.py
orndorffgrant: ohhh very good idea I'll do that

uaclient/lock.py
orndorffgrant: Yep good catch

uaclient/tests/test_lock.py
orndorffgrant: :facepalm:

uaclient/tests/test_lock.py
orndorffgrant: Yes that would be a better assertion for sure

uaclient/tests/test_lock.py
orndorffgrant: Good idea, I'll try it out. I don't want to complicate the tests too much just to remove a bit of duplication though

orndorffgrant: I had a similar idea but want to run it by you before implementing @lucasmoura.

Would it make sense for the `SpinLock` class tests to only verify that it calls `SingleAttemptLock.__enter__/__exit__` instead of testing that it calls the correct `write_cache/add_notice/delete_cache` functions? That would simplify the SpinLock tests a bit

orndorffgrant (commented):
@lucasmoura updated! Plus one idea in a comment thread I could use your opinion on

uaclient/tests/test_lock.py
lucasmoura: But if we do that, then we would lose a little bit of our tracking of what is happening inside both of these methods right ?

orndorffgrant: Right, actually I think it would just look like deleting `test_creates_lock_and_notice` and `test_creates_lock_and_notice_when_error_occurs`. On second thought I don't think we should do that. I'll do the refactor you suggest

orndorffgrant: @lucasmoura updated! Sorry for the FUD, that was straightforward, and is definitely better

uaclient/exceptions.py
```diff
@@ -70,6 +72,9 @@ class LockHeldError(UserFacingError):
     """
 
     def __init__(self, lock_request: str, lock_holder: str, pid: int):
+        self.lock_request = lock_request
+        self.lock_holder = lock_holder
+        self.lock_holder_pid = pid
```
lucasmoura, line 77: Are we using those variables somewhere else ?

uaclient/tests/test_lock.py
```diff
@@ -0,0 +1,167 @@
...
+    ):
+        cfg = FakeConfig()
+
+        def test_function():
+            pass
+
+        with SpinLock(
+            cfg=cfg, lock_holder="request", sleep_time=1, max_retries=3
+        ):
+            test_function()
```
lucasmoura, line 127: I think we can just call `pass` here, right ?

uaclient/tests/test_lock.py
```diff
@@ -0,0 +1,167 @@
...
+    ):
+        cfg = FakeConfig()
+
+        def test_function():
+            pass
+
+        with pytest.raises(LockHeldError) as exc:
+            with SpinLock(
+                cfg=cfg, lock_holder="request", sleep_time=1, max_retries=2
+            ):
```
lucasmoura, line 156: Same `pass` idea here

lucasmoura (commented):
Just a couple more nits and I am good with the PR

uaclient/exceptions.py
orndorffgrant: Only in the test :/

If you think this is unnecessary I'm open to removing it

orndorffgrant:
Updated with the simpler `pass`es

uaclient/exceptions.py
lucasmoura: Maybe we could directly check the `msg` variable of the exception instead. It will be a proxy for those variables

orndorffgrant: Yeah that's fair, I'll make that change

orndorffgrant: Oops actually I also used these in `SpinLock.__enter__` for the log message there, so I think we should keep these

renanrodrigo (approved):
LGTM.
I tried to come up with a suggestion to remove the global since we now have a module to handle the locks, but as is, my attempts generated only uglier alternatives.

uaclient/exceptions.py
lucasmoura: One thing we can do is rely on the message of the exception and just append the `spinning...` part to it.
But maybe this is unnecessary. Let me know what you think

lucasmoura:
@orndorffgrant let's just double check this test here on xenial to see if everything is alright:
```
Run timer script to valid machine activity endpoint
```

uaclient/exceptions.py
orndorffgrant: Sure we can do that

orndorffgrant: Updated!

orndorffgrant:
@lucasmoura Since I made that change, let's just let jenkins run again and hopefully it will pass this time

lucasmoura (approved):

PR #1896, opened 2021-11-24
refactor: decouple attach from cli

orndorffgrant:
## Notes

- [x] #1894
- [x] #1895
- [ ] #1896 (this one)
- [ ] #1887

## Proposed Commit Message

```
refactor: decouple attach from cli
```

## Test Steps

Ensure attach and auto-attach still function properly

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -15,11 +15,13 @@ import tempfile
 import textwrap
 import time
 from functools import wraps
+from typing import Optional  # noqa: F401
 from typing import List
 
 import yaml
 
 from uaclient import (
+    actions,
     config,
     contract,
     entitlements,
@@ -31,6 +33,7 @@ from uaclient import (
 )
 from uaclient import status as ua_status
 from uaclient import util, version
+from uaclient.clouds import AutoAttachCloudInstance  # noqa: F401
 from uaclient.clouds import identity
 from uaclient.defaults import (
     CLOUD_BUILD_INFO,
@@ -964,27 +967,7 @@ def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
     return 0
 
 
-def _attach_with_token(
-    cfg: config.UAConfig, token: str, allow_enable: bool
-) -> int:
-    """Common functionality to take a token and attach via contract backend"""
-    try:
-        contract.request_updated_contract(
-            cfg, token, allow_enable=allow_enable
-        )
-    except util.UrlError as exc:
-        with util.disable_log_to_console():
-            logging.exception(exc)
-        print(ua_status.MESSAGE_ATTACH_FAILURE)
-        cfg.status()  # Persist updated status in the event of partial attach
-        config.update_ua_messages(cfg)
-        return 1
-    except exceptions.UserFacingError as exc:
-        logging.warning(exc.msg)
-        cfg.status()  # Persist updated status in the event of partial attach
-        config.update_ua_messages(cfg)
-        return 1
-
+def _post_cli_attach(cfg: config.UAConfig) -> None:
     contract_name = cfg.machine_token["machineTokenInfo"]["contractInfo"][
         "name"
     ]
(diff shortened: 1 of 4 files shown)


uaclient/cli.py
```diff
@@ -1068,36 +1034,16 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
...
-    disable_auto_attach = util.is_config_value_true(
-        config=cfg.cfg, path_to_value="features.disable_auto_attach"
-    )
-    if disable_auto_attach:
-        msg = "Skipping auto-attach. Config disable_auto_attach is set."
-        logging.debug(msg)
-        print(msg)
+    try:
+        actions.auto_attach(cfg, instance)
+        _post_cli_attach(cfg)
```
lucasmoura, line 1040: We could call the `_post` function in the else block of the `try..except`
Also, you are create those helper functions just for readability purposes ? If yes, might we could run their logic here directly

orndorffgrant: True I can move it to `else` :+1:

`_post_cli_attach` is also called from  `action_attach` below so I'd rather avoid duplicating the logic

orndorffgrant:
@lucasmoura updated!

uaclient/cli.py
```diff
@@ -1072,36 +1038,17 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
...
-    if disable_auto_attach:
-        msg = "Skipping auto-attach. Config disable_auto_attach is set."
-        logging.debug(msg)
-        print(msg)
+    try:
+        actions.auto_attach(cfg, instance)
+    except util.UrlError:
+        print(ua_status.MESSAGE_ATTACH_FAILURE)
+        return 1
+    except Exception:
```
lucasmoura, line 1047: I think we should inform the user about the `Exception` somehow. 
Even if it is an unknown `Exception`  we could log it and say to the user to check the logs

uaclient/cli.py
```diff
@@ -1046,6 +1021,13 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
         raise exceptions.UserFacingError(
             ua_status.MESSAGE_UNABLE_TO_DETERMINE_CLOUD_TYPE
         )
+
+    if not instance:
+        # we shuoldn't get here, but this is a reasonable default just in case
```
orndorffgrant, line 1026: ugh look at me copy-pasting typos :facepalm:

uaclient/cli.py
orndorffgrant: You're totally right. I think we can just let it bubble up

orndorffgrant: Oh I think I mean to catch `UserFacingError` here to match the old implementation line 982.

orndorffgrant: @lucasmoura updated

uaclient/cli.py
renanrodrigo: yes definitely shuoldn't copypaste those
:P

renanrodrigo (approved):
LGTM

uaclient/cli.py
```diff
@@ -1072,36 +1038,17 @@ def _get_contract_token_from_cloud_identity(cfg: config.UAConfig) -> str:
...
-    if disable_auto_attach:
-        msg = "Skipping auto-attach. Config disable_auto_attach is set."
-        logging.debug(msg)
-        print(msg)
+    try:
+        actions.auto_attach(cfg, instance)
+    except util.UrlError:
+        print(ua_status.MESSAGE_ATTACH_FAILURE)
+        return 1
+    except exceptions.UserFacingError:
```
lucasmoura, line 1047: Sorry @orndorffgrant, I think I am still missing something. Shouldn't we still surface some error message to the user in that situation ?

uaclient/cli.py
```diff
@@ -1112,9 +1059,18 @@ def action_attach(args, *, cfg):
...
-        cfg, token=args.token, allow_enable=args.auto_enable
-    )
+    try:
+        actions.attach_with_token(
+            cfg, token=args.token, allow_enable=args.auto_enable
+        )
+    except util.UrlError:
+        print(ua_status.MESSAGE_ATTACH_FAILURE)
+        return 1
+    except exceptions.UserFacingError:
```
lucasmoura, line 1069: Same question here

uaclient/cli.py
orndorffgrant: @lucasmoura This is just doing the same thing that it was doing before but rearranged. The logging now happens in actions.attach_with_token
It used to look like this in `cli._attach_with_token`
```python
    try:
        contract.request_updated_contract(
            cfg, token, allow_enable=allow_enable
        )
    except util.UrlError as exc:
            logging.exception(exc)
        print(ua_status.MESSAGE_ATTACH_FAILURE)
        cfg.status()  # Persist updated status in the event of partial attach
        config.update_ua_messages(cfg)
        return 1
    except exceptions.UserFacingError as exc:
        # *********************** logged here *************************************
        logging.warning(exc.msg)
        cfg.status()  # Persist updated status in the event of partial attach
        config.update_ua_messages(cfg)
        return 1
```

Now it looks like this.

In `actions.attach_with_token`:
```python
    try:
        contract.request_updated_contract(
            cfg, token, allow_enable=allow_enable
        )
    except util.UrlError as exc:
            LOG.exception(exc)
        # *********************** notice no print statement here *************************************
        cfg.status()  # Persist updated status in the event of partial attach
        config.update_ua_messages(cfg)
        # *********************** raise instead of return *************************************
        raise exc
    except exceptions.UserFacingError as exc:
        # *********************** I left this log here *************************************
        LOG.warning(exc.msg)
        cfg.status()  # Persist updated status in the event of partial attach
        config.update_ua_messages(cfg)
        # *********************** raise instead of return *************************************
        raise exc
```

In `cli.action_attach`:
```python
    try:
        actions.attach_with_token(
            cfg, token=args.token, allow_enable=args.auto_enable
        )
    except util.UrlError:
        # *********************** the print happens at this level now *************************************
        print(ua_status.MESSAGE_ATTACH_FAILURE)
        # *********************** same with return for both cases *************************************
        return 1
    except exceptions.UserFacingError:
        # *********************** same with return for both cases *************************************
        return 1
```

Arguably that UserFacingError should not be logged in `actions.attach_with_token`. I was just trying to keep everything exactly the same for this refactor. But I can add another commit to this PR to let the UserFacingError bubble up instead of handling like it is now.

lucasmoura: Thanks so much for the whole context here. My mistake that I missed that we are actually logging the error.
I think we can keep it as is

lucasmoura (approved):

PR #1897, opened 2021-11-26
logs: redact tokens sent from contract server

CalvoM:
## Proposed Commit Message

logs: redact tokens sent from contract server

## Test Steps
**On pro machines:** 

- run auto-attach to see the logged contractToken
- run auto-attach on gcp and see the logged jwt with machine data and details.

**On both pro and non-pro machines:**

- enable cis/cc-eal and see the logged resourceToken
- change machine token and then try enabling livepatch

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/serviceclient.py
+++ b/uaclient/serviceclient.py
@@ -68,11 +68,15 @@ class UAServiceClient(metaclass=abc.ABCMeta):
                 timeout=self.url_timeout,
             )
         except error.URLError as e:
-            if hasattr(e, "read"):
+            body = None
+            if hasattr(e, "body"):
+                body = e.body
+            elif hasattr(e, "read"):
+                body = e.read().decode("utf-8")
+            if body:
                 try:
                     error_details = json.loads(
-                        e.read().decode("utf-8"),
-                        cls=util.DatetimeAwareJSONDecoder,
+                        body, cls=util.DatetimeAwareJSONDecoder
                     )
                 except ValueError:
                     error_details = None
--- a/uaclient/tests/test_util.py
+++ b/uaclient/tests/test_util.py
@@ -934,6 +934,42 @@ class TestRedactSensitiveLogs:
                 "Contract value for 'resourceToken' changed to S3kR3T",
                 "Contract value for 'resourceToken' changed to <REDACTED>",
             ),
+            (
+                "data: {'contractToken': 'SEKRET', "
+                "'contractTokenInfo':{'expiry'}}",
+                "data: {'contractToken': '<REDACTED>', "
+                "'contractTokenInfo':{'expiry'}}",
+            ),
+            (
+                "data: {'resourceToken': 'SEKRET', "
+                "'entitlement': {'affordances':'blah blah' }}",
+                "data: {'resourceToken': '<REDACTED>', "
+                "'entitlement': {'affordances':'blah blah' }}",
+            ),
+            (
+                "https://contracts.canonical.com/v1/resources/livepatch"
+                "?token=SEKRET: invalid token",
+                "https://contracts.canonical.com/v1/resources/livepatch"
+                "?token=<REDACTED> invalid token",
+            ),
+            (
+                'data: {"identityToken": "SEket.124-_ys"}',
+                'data: {"identityToken": "<REDACTED>"}',
+            ),
+            (
+                "http://metadata/computeMetadata/v1/instance/service-accounts/"
+                "default/identity?audience=contracts.canon, data: none",
+                "http://metadata/computeMetadata/v1/instance/service-accounts/"
+                "default/identity?audience=contracts.canon, data: none",
+            ),
+            (
+                "response: "
+                "http://metadata/computeMetadata/v1/instance/service-accounts/"
+                "default/identity?audience=contracts.canon, data: none",
+                "response: "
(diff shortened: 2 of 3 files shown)


orndorffgrant (changes requested):
This is great!
Just one small addition and then I think we will be succesfully redacting everything we need.

Note that this doesn't completely resolve SC-513 because we aren't logging all failed request responses. To see an example, run `ua auto-attach` on a non-pro instance. The response to `URL [POST]: https://contracts.canonical.com/v1/clouds/gcp/token,` is not logged. For that we'd need another commit that adds something like this:

```diff
diff --git a/uaclient/util.py b/uaclient/util.py
index 6dcbb905..175d2635 100644
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -555,7 +555,13 @@ def readurl(
             )
         )
     )
-    resp = request.urlopen(req, timeout=timeout)
+    try:
+        resp = request.urlopen(req, timeout=timeout)
+        failed = False
+    except error.HTTPError as e:
+        resp = e
+        failed = True
+
     content = resp.read().decode("utf-8")
     if "application/json" in str(resp.headers.get("Content-type", "")):
         content = json.loads(content)
@@ -569,6 +575,8 @@ def readurl(
             )
         )
     )
+    if failed:
+        raise resp
     return content, resp.headers

```

lucasmoura:
I wasn't able to reproduce the `livepatch` scenario. I think it might be related to the problem @orndorffgrant identified.

But if it isn't, can you provide the steps you used to reproduce it @CalvoM ?

uaclient/entitlements/livepatch.py
```diff
@@ -240,7 +240,7 @@ def setup_livepatch_config(
                         break
                 if msg == "Unable to enable Livepatch: ":
                     msg += str(e)
-                print(msg)
+                print(util.redact_sensitive_logs(msg))
```
lucasmoura, line 243: Should we redact this when printing the message too ?
Maybe it is a little unnecessary. What do you think @orndorffgrant

orndorffgrant: Ah I was the one who recommended this. But I think you're right @lucasmoura  we really only need the redaction in the subp logging added below.
Sorry for the churn here @CalvoM - my fault - but we can undo this line change.

CalvoM: Alright, on it.

orndorffgrant:
@CalvoM Sorry it looks like the change to log http errors caused some integration test failures.

I think something like this will fix it:
```diff
diff --git a/uaclient/serviceclient.py b/uaclient/serviceclient.py
index 48a12735..1cfa704d 100644
--- a/uaclient/serviceclient.py
+++ b/uaclient/serviceclient.py
@@ -68,11 +68,15 @@ class UAServiceClient(metaclass=abc.ABCMeta):
                 timeout=self.url_timeout,
             )
         except error.URLError as e:
-            if hasattr(e, "read"):
+            body = None  # type: Optional[str]
+            if hasattr(e, "body"):
+                body = e.body
+            elif hasattr(e, "read"):
+                body = e.read().decode("utf-8")
+            if body:
                 try:
                     error_details = json.loads(
-                        e.read().decode("utf-8"),
-                        cls=util.DatetimeAwareJSONDecoder,
+                        body, cls=util.DatetimeAwareJSONDecoder
                     )
                 except ValueError:
                     error_details = None
diff --git a/uaclient/util.py b/uaclient/util.py
index 3c4b5358..4af214d7 100644
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -562,7 +562,8 @@ def readurl(
     except error.HTTPError as e:
         resp = e
         http_error_found = True
-    content = resp.read().decode("utf-8")
+    setattr(resp, "body", resp.read().decode("utf-8"))
+    content = resp.body
     if "application/json" in str(resp.headers.get("Content-type", "")):
         content = json.loads(content)
     sorted_header_str = ", ".join(

```

uaclient/serviceclient.py
```diff
@@ -68,11 +68,15 @@ def request_url(
                 timeout=self.url_timeout,
             )
         except error.URLError as e:
-            if hasattr(e, "read"):
+            body = None
+            if hasattr(e, "body"):
+                body = e.body
```
lucasmoura, line 73: @orndorffgrant @CalvoM  just to be sure that I understand this change, there are other exception that have their content on the body attribute right ?

'
```diff
@@ -0,0 +1,155 @@
+import abc
```
lucasmoura, line 1: @CalvoM I think this file was accidentally included

uaclient/serviceclient.py
orndorffgrant: @lucasmoura 
We added to `util.readurl` to log the response of the HTTPError and `read` can only be called once on an IO stream (`seek(0)` works on normal files, not sure if it works on an http response). Without this change, this error handling code would fail because there is nothing to `read()`.
So now, when we call `read` in `util.readurl`, we attach the content to the response/error as the new `body` attribute. Then we check for that attribute first here before trying to `read`

lucasmoura (approved):

orndorffgrant (approved):
Thanks for all the iterations! I think this looks really good.

PR #1900, opened 2021-11-30
Add status endpoint to execute a preflight check of a specific token (SC-623)

renanrodrigo:
This adds the `--simulate-with-token` option to the `status` command.
Using it requires a valid UA token, and will output information about services, as: 
- service name and description
- availability of the service for the archuitecture/series/kernel
- entitlement status from the contract perspective
- whether the service is marked to be enabled automatically.

This would give all information needed to execute a preflight check before users attach during installation time, as well as provide token information for any attached/unattached user.

## Desired commit type

## Checklist

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -230,6 +230,20 @@ def when_i_run_command_with_stdin(
     )
 
 
+@when("I do a preflight check for `{contract_token}` {user_spec}")
+def when_i_preflight(context, contract_token, user_spec):
+    token = getattr(context.config, contract_token)
+    command = "ua status --simulate-with-token {}".format(token)
+    if user_spec == "with the all flag":
+        command += " --all"
+    if "formatted as" in user_spec:
+        output_format = user_spec.split()[2]
+        command += " --format {}".format(output_format)
+    when_i_run_command(
+        context=context, command=command, user_spec="as non-root"
+    )
+
+
 @when("I run `{command}` {user_spec}")
 def when_i_run_command(
     context,
--- a/features/unattached_status.feature
+++ b/features/unattached_status.feature
@@ -9,14 +9,14 @@ Feature: Unattached status
             """
             _doc _schema_version account attached config config_path contract effective
             environment_vars execution_details execution_status expires machine_id notices
-            services version
+            services version simulated
             """
         When I run `ua status --format yaml` as non-root
         Then stdout is formatted as `yaml` and has keys:
             """
             _doc _schema_version account attached config config_path contract effective
             environment_vars execution_details execution_status expires machine_id notices
-            services version
+            services version simulated
             """
 
         Examples: ubuntu release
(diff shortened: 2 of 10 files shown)


uaclient/config.py
```diff
@@ -751,7 +751,104 @@ def _attached_status(self) -> Dict[str, Any]:
...
+
+        support = self._get_entitlement_information(entitlements, "support")
+        if support["entitled"]:
+            supportLevel = support["affordances"].get("supportLevel")
+            if supportLevel:
+                response["contract"]["tech_support_level"] = supportLevel
+
+        return response
+
+    def status(
```
lucasmoura, line 849: It seems that the only thing we are using from `status` is the beta service redaction part.
I think we should turn that beta part into a separate function and create function for the status simulation instead

uaclient/config.py
```diff
@@ -751,7 +751,104 @@ def _attached_status(self) -> Dict[str, Any]:
...
+                "products": contract_info.get("products", []),
+            },
+            "account": {
+                "name": account_info.get("name", ""),
+                "id": account_info.get("id"),
+                "created_at": account_info.get("createdAt", ""),
+                "external_account_ids": account_info.get(
+                    "externalAccountIDs", []
+                ),
+            },
```
lucasmoura, line 799: Do we need to surface this information ? It seems we just want to present service information in that command

uaclient/config.py
```diff
@@ -751,7 +751,104 @@ def _attached_status(self) -> Dict[str, Any]:
...
+            "services": [],
+            "execution_status": DEFAULT_STATUS.get("execution_status", ""),
+            "simulated": True,
+        }
+
+        if contract_info.get("effectiveTo"):
+            response["expires"] = contract_info.get("effectiveTo")
+        if contract_info.get("effectiveFrom"):
+            response["effective"] = contract_info.get("effectiveFrom")
```
lucasmoura, line 808: Same question here

uaclient/config.py
```diff
@@ -751,7 +751,104 @@ def _attached_status(self) -> Dict[str, Any]:
...
+
+        resources = get_available_resources(self)
```
lucasmoura, line 810: Isn't the information we want already present in the `contract_information` variable ?

uaclient/config.py
```diff
@@ -751,7 +751,104 @@ def _attached_status(self) -> Dict[str, Any]:
...
+        entitlements = contract_info.get("resourceEntitlements", [])
+
+        inapplicable_resources = {
+            resource["name"]: resource.get("description")
+            for resource in sorted(resources, key=lambda x: x["name"])
+            if not resource["available"]
+        }
+
+        for resource in sorted(resources, key=lambda x: x.get("name", "")):
```
lucasmoura, line 819: Instead of doing this, can't we just loop into the available entitlements and parse their dict into the response we want ?
I understand that this will surface not available service, but maybe in that command, this will be useful, since you want to see the overall information of as many services as you want

features/steps/steps.py
```diff
@@ -251,6 +251,12 @@ def when_i_run_command(
             break
 
     full_cmd = prefix + shlex.split(command)
+
+    if "contract_token" in full_cmd:
```
lucasmoura, line 255: I think it is better for us to create a step for the `simulate` command, just to keep this step clean.
We could do a similar approach to the attach step

uaclient/config.py
renanrodrigo: Indeed.
I just created here a similar flow to the other status calls we have, and there are (and there will be more) small parts that we need to use for them all, where some things can be simply separate.
I think the whole `status` flow needs to be refactored anyway - Taken out of the `UAConfig` class and taken care of.

My proposal here would be: can we merge this functionality first, and then work on a deeper refactoring of the status?
@orndorffgrant @blackboxsw thoughts?

uaclient/config.py
renanrodrigo: I think that basic contract and account information may also be useful for people consuming this endpoint. At installation time, when providing a token, the user can verify that this data is right in the step where there are the services for picking, and also have easy access to this data to report if something goes wrong.
From the complexity side of things, the endpoint in the Contract Server is providing this data directly, so I think it's not bad to include it.

uaclient/config.py
renanrodrigo: same as above

uaclient/config.py
renanrodrigo: Yes, you are right, we should be checking for that in the cache before making the call to the server - will fix.

uaclient/config.py
renanrodrigo: :thinking:

renanrodrigo: As discussed:
The endpoint is not returning all of the entitlements - only the ones you are entitled to. Thus, we still need to loop over the available resources to check for non-entitled services.
As we talk to contracts on how to improve the endpoint, we will have the time to refactor this

uaclient/config.py
renanrodrigo: added a check for the status cache before making this call

uaclient/config.py
renanrodrigo: Oki so
I took a look on the refactoring work and I assume that your suggestion makes sense for this PR. When moving everything to a different module is is expected that those two pieces of functionality are isolated anyway. /me on it.

renanrodrigo:
@lucasmoura comments addressed, ready for a second round of reviews

lucasmoura (approved):

ubuntu-advantage.1
```diff
@@ -114,6 +114,14 @@ of:
 whether this service would be available if this machine were attached.
 The possible values are \fIyes\fR or \fIno\fR.
 
+If the status is simulated with a token, then the output has five columns.
```
orndorffgrant, line 117: Minor opinion
```suggestion
If --simulate-with-token is used, then the output has five columns.
```

uaclient/config.py
orndorffgrant: Aye, I agree and think that the output of this command should be as close the output of a normal status command as possible.

ubuntu-advantage.1
renanrodrigo: agreed updated

features/unattached_status.feature
```diff
@@ -123,3 +123,42 @@ Feature: Unattached status
            | focal   | yes      | no     | yes | yes  | yes         | yes   | no  | yes       |
            | hirsute | no       | no     | no  | no   | no          | no    | no  | no        |
            | impish  | no       | no     | no  | no   | no          | no    | no  | no        |
+
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    @uses.config.contract_token
+    Scenario Outline: Simulate status in a ubuntu machine
```
orndorffgrant, line 131: Should we also have a test for the json output? Maybe to ensure it is the same schema as normal status output?

orndorffgrant (commented):
I agree that especially after this feature, there is a lot of repetition and.this deserves a refactor. In particular, it is unideal that the structure of our status is defined in several places. I would like to move towards a world where our various json schemas are modeled as python objects rather than as dicts so that the structure can be enforced by python types.

features/unattached_status.feature
renanrodrigo: hhhm yes - it will be a subset of the normal status output, but it's worth testing it. will add.

renanrodrigo:
@orndorffgrant test added

features/unattached_status.feature
```diff
@@ -123,3 +123,56 @@ Feature: Unattached status
...
+        fips-updates  <fips>      +yes  +no   +NIST-certified core packages with priority security updates
+        livepatch     <livepatch> +yes  +yes  +Canonical Livepatch service
+        ros           <ros>       +yes  +no   +Security Updates for the Robot Operating System
+        ros-updates   <ros>       +yes  +no   +All Updates for the Robot Operating System
+        """
+        When I do a preflight check for `contract_token` formatted as json
+        Then stdout is formatted as `json` and has keys:
+        """
+        _doc _schema_version version contract account services execution_status
+        simulated expires environment_vars
```
orndorffgrant, line 162: Another refactor I'd like to do is to create json-schemas for all json output we have. Then we can do a deeper integration test than just checking the top-level keys.
Not necessary for this PR though

orndorffgrant (changes requested):
I think the json/yaml output of this should be exactly the same as `ua status --format json` when attached. Anything else and we're introducing subtle (IMO unexpected) differences. I don't want to explain to future consumer: "Well it's still schema_version 0.1 but its different because you're simulating being attached instead of actually being attached".
So we're missing:
* `attached`
* `config`
* `config_path`
* `effective`
* `execution_details`
* `machine_id`
* `notices`

And the normal `ua status --format json` needs to be updated to include:
* `services[].auto_enabled`
* `simulated`

features/unattached_status.feature
renanrodrigo: Agree that we can add the top level keys. Although some information might be empty or not relevant, I see the concern there in the schema version.
I disagree on the `services[].auto_enabled` though, because we have no use for it in the attached status, and we already have (many) differences in the keys of a service in attached vs unattached. I think we can treat those (attached, unattached, sinulated) as different views of the services.

orndorffgrant: Thats a really good point on auto_enabled. Let's leave it out then

renanrodrigo:
@orndorffgrant please take another look and see if it makes sense to you.
The only thing I am not sure is returning `attached: False`, because a user may be attached... but in the end this represents a simulated status, and thus attachment is not important for the context. If you disagree, I can change it as well.

orndorffgrant (approved):
Thanks for the updates!

Regarding `attached` I could be convinced either way... I think it is good as-is

PR #1901, opened 2021-12-01
livepatch: Add full path to all cmds (SC-627)

lucasmoura:
## Proposed Commit Message
livepatch: Add full path to all cmds

When interacting with livepatch, if the user does not have /snap/bin in the PATH, the pure canonical-livepatch command
will fail. We are now specifying the full path for all livepatch commands we are running.

LP: #1951954

## Test Steps
Run any integration test that enable/disables livepatch

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -37,6 +37,11 @@ from uaclient.defaults import (
     DEFAULT_CONFIG_FILE,
 )
 
+# TODO: Better address service commands running on cli
+# It is not ideal for us to import an entitlement directly on the cli module.
+# We need to refactor this to avoid that type of coupling in the code.
+from uaclient.entitlements.livepatch import LIVEPATCH_CMD
+
 NAME = "ua"
 
 USAGE_TMPL = "{name} {command} [flags]"
@@ -1143,7 +1148,7 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
             "ua status --format json", "{}/ua-status.json".format(output_dir)
         )
         _write_command_output_to_file(
-            "canonical-livepatch status",
+            "{} status".format(LIVEPATCH_CMD),
             "{}/livepatch-status.txt".format(output_dir),
         )
         _write_command_output_to_file(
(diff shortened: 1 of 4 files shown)


orndorffgrant (commented):
Thanks for this fix! I'm running some integration tests now. The only other place we could use this is in the collect-logs command where we get the livepatch status.

orndorffgrant:
And my tests passed! Will approve after update to collect-logs :)

lucasmoura:
@orndorffgrant Good catch. I have updated the code

renanrodrigo (approved):
lgtm

orndorffgrant (approved):
Just the smallest of nits. Otherwise LGTM

orndorffgrant:
Oof didn't save my comment, one sec

uaclient/cli.py
```diff
@@ -1143,7 +1143,7 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
             "ua status --format json", "{}/ua-status.json".format(output_dir)
         )
         _write_command_output_to_file(
-            "canonical-livepatch status",
+            "/snap/bin/canonical-livepatch status",
```
orndorffgrant, line 1146: Tiny nit: Should this use the `livepatch.LIVEPATCH_CMD` variable?

lucasmoura: I thought about that, but I did not want to add a direct entitlement import into the cli module.
But maybe this doesn't matter that much.

What do you think ?

orndorffgrant: @lucasmoura Good point, that is a sign that something is a little weird.

I think its a sign that the general livepatch functionality that isn't directly tied to the entitlement should be refactored into its own module maybe? Not urgent though.

Up to you on whether you think its reasonable or not. I think I still lean towards using the constant, but am going to go ahead and approve

orndorffgrant (approved):

uaclient/cli.py
lucasmoura: Yes, currently we have no way to handle service specifc requests anywhere in the codebase.
I agree that we need a refactoring on that part. I will use the constant and add a comment that should touch this in the near future

PR #1902, opened 2021-12-07
tests: add jammy integration tests (SC-663)

lucasmoura:
## Proposed Commit Message
tests: add jammy integration tests

Add some integration tests for 22.04 (jammy)

## Test Steps
Run the new integration tests

## Desired commit type

## Checklist

### DIFF
--- a/features/_version.feature
+++ b/features/_version.feature
@@ -31,7 +31,8 @@ Feature: UA is expected version
             | bionic  |
             | focal   |
             | hirsute |
-            | impish |
+            | impish  |
+            | jammy   |
 
     @series.all
     @uses.config.check_version
@@ -58,4 +59,4 @@ Feature: UA is expected version
             | bionic  |
             | focal   |
             | hirsute |
-            | impish |
+            | impish  |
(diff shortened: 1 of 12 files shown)


renanrodrigo:
Could you include it in `tools/run-integration-tests.py` as well?
Adding it to the constants in the start should do

lucasmoura:
@renanrodrigo updated

renanrodrigo (approved):
lgtm

orndorffgrant (approved):
These changes look good to me!

While we're here though. Should we run any VM tests on jammy? Also do any of the clouds have jammy images yet?

lucasmoura:
@orndorffgrant The problem of adding those tests is that we don't have any services that are applicable on jammy, which limits a lot of what we test on vms and clouds. But if you think there is a scenario we can use for testing, just let me know.

And regarding the jenkins part, I don't know if we should add it just yet. I think once we can enable at least one service there, it would be a good call

orndorffgrant:
@lucasmoura good points! Let's merge as is then

lucasmoura:
CI error is unrelated with this change

PR #1903, opened 2021-12-08
logs: log to stdout/stderr when subp fails

CalvoM:
## Proposed Commit Message

logs: log to stdout/stderr when subp fails

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/tests/test_util.py
+++ b/uaclient/tests/test_util.py
@@ -350,6 +350,29 @@ class TestSubp:
         for log in expected_logs:
             assert log in logs
 
+    @pytest.mark.parametrize("caplog_text", [logging.WARNING], indirect=True)
+    @pytest.mark.parametrize("capture", [True, False])
+    @mock.patch("uaclient.util._subp")
+    def test_log_subp_fails_stdout_stderr_capture_toggle(
+        self, m_subp, capture, caplog_text
+    ):
+        """When subp fails, capture the logs in stdout/stderr"""
+        out = "Tried downloading file"
+        err = "Network error"
+        m_subp.side_effect = util.ProcessExecutionError(
+            "Serious apt error", stdout=out, stderr=err
+        )
+        with pytest.raises(util.ProcessExecutionError):
+            util.subp(["apt", "nothing"], capture=capture)
+
+        logs = caplog_text()
+        expected_logs = ["Stderr: {}".format(err), "Stdout: {}".format(out)]
+        for log in expected_logs:
+            if capture:
+                assert log in logs
+            else:
+                assert log not in logs
+
 
 class TestParseOSRelease:
     @pytest.mark.parametrize("caplog_text", [logging.DEBUG], indirect=True)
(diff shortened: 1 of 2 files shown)


CalvoM:
@lucasmoura will this PR change later after #1905 is approved? i.e. replacing the print messages with event.info

uaclient/util.py
```diff
@@ -676,10 +676,12 @@ def subp(
         except ProcessExecutionError as e:
             if capture:
                 logging.debug(str(e))
+                print(str(e))
             if not retry_sleeps:
                 raise
             retry_msg = " Retrying %d more times." % len(retry_sleeps)
             logging.debug(redact_sensitive_logs(str(e) + retry_msg))
+            print(str(e))
```
lucasmoura, line 684: Should we only print that message if `capture` is `True` ? Otherwise we could be printing this message twice to the user

lucasmoura (commented):
@CalvoM Let's also add an unittest for this change.

And regarding the `event.info` PR, no worries at all. If this PR gets merged first, I can do the change on my side.
And even though I think your PR will be merged first, if that does not happen, then yes, we will need to change this PR, but it will be a simple change

orndorffgrant (changes requested):
Hi @lucasmoura and @CalvoM Sorry I think I miscommunicated the intent of this feature in the issue description.

My request was not to log the current error to stdout.

My request was to take the stderr/stdout of the subprocess and make sure we are always logging that when the subprocess fails. (I've added the mattermost conversation that triggered this issue to the issue description).

Though now I see we have the `capture` argument to `subp`. So I think we should only log stdout/stderr of the subprocess when `capture=True`

Altogether, I think the change would involve adding some code after `util.py:678` that will `log.debug` `e.stdout` and `e.stderr` maybe with some nice string formatting to add labels.

What do you think @lucasmoura @CalvoM ?

lucasmoura:
@orndorffgrant That makes sense to me and I agree that is indeed a better approach

CalvoM:
@orndorffgrant, with the context you have provided, it is surely a better approach.
@lucasmoura  and @orndorffgrant This message from [run_command](https://github.com/canonical/ubuntu-advantage-client/blob/c89d1cab1de5c055167ce23c5193f84f89119d86/lib/reboot_cmds.py#L32) in reboot_cmd is a good message to use. What do you think?

uaclient/tests/test_util.py
```diff
@@ -350,6 +350,43 @@ def test_retry_logs_remaining_retries(self, m_sleep, m_subp, caplog_text):
         for log in expected_logs:
             assert log in logs
 
+    @pytest.mark.parametrize("caplog_text", [logging.WARNING], indirect=True)
+    @mock.patch("uaclient.util._subp")
+    def test_log_subp_fails_stdout_stderr_capture_true(
```
lucasmoura, line 355: Maybe we can combine both of these tests and just parametrize the `capture` parameter

CalvoM: Yes, I was reading more on pytest just about that.

CalvoM: Updated.

lucasmoura (approved):

orndorffgrant (approved):
LGTM! Merging

PR #1904, opened 2021-12-08
cli: add mechanism to ignore print events

lucasmoura:
## Proposed Commit Message
cli: add mechanism to ignore print events

When running some UA commands, we will be required to provide their output in a machine-readable way. When that happens, we don't to print anything to stdout. We are now adding a mechanism that allows us to achieve that

## Test Steps
Run any integration tests that print to stdout and verify that it still works

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -25,6 +25,7 @@ from uaclient import (
     config,
     contract,
     entitlements,
+    event_logger,
     exceptions,
     jobs,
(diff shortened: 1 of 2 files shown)


lucasmoura:
Even though this solution works, we agree that this could be confusing for new users and it could generate maintenance burdens in the future. Closing it in favor of #1905

PR #1905, opened 2021-12-09
cli: add mechanism to ignore print events

lucasmoura:
## Proposed Commit Message
cli: add mechanism to ignore print events

When running some UA commands, we will be required to provide their
output in a machine-readable way. When that happens, we don't to print
anything to stdout. We are now adding a mechanism that allows us to
achieve that

PS: This is an alternative for #1904 

## Test Steps
Run the unittests and any integration test that is printing to stdout

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/event_logger.py
+++ b/uaclient/event_logger.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+
+"""
+This module is responsible for handling all events
+that must be raised to the user somehow. The main idea
+behind this module is to centralize all events that happens
+during the execution of UA commands and allows us to report
+those events in real time or through a machine-readable format.
+"""
+
+import enum
+
+
+@enum.unique
+class EventLoggerMode(enum.Enum):
+    """
+    Defines event logger supported modes.
+    Currently, we only support the cli and machine-readable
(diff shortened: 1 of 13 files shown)


uaclient/event_logger.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
```
orndorffgrant, line 1: Do we need this line?

orndorffgrant (approved):
Looks good! I'll just run some of the tests jenkins didn't like to make sure they're okay

And one tiny comment

uaclient/event_logger.py
lucasmoura: To be honest, I don't think we need it. I just followed the "template" of the `cli` module. But I agree we can remove it

lucasmoura:
@orndorffgrant Thanks for the review on that PR, but I think we can use #1912 altogether. I have made some modifications already on it regarding the `print` part of the event logger. Sorry for that, I should have closed this PR earlier

orndorffgrant:
@lucasmoura All good. I was aware but thought we were going to merge this first as a stepping stone and also to make the diff for #1912 a bit smaller. But this strategy works for me too :)

PR #1908, opened 2021-12-10
Release 27.5

renanrodrigo:
There we go!
Please check the changelog entries and commit set for this release.


debian/changelog
```diff
@@ -1,3 +1,52 @@
...
+  * New upstream release 27.5:
+    - aws: add support for the IPv6 metadata endpoint
+    - cis: update URL for the documentation
+    - cli:
+      + add endpoint to simulate the status using a specific token
+      + fix return code when attaching an already attached machine (GH: #1867)
+      + fix security-status to consider all possible origins to show updates
+      + include cloud build.info in the collect-logs tarball
+      + only show services which exist in the contracts server in ua status
+    - clouds: add pro license check methods
```
lucasmoura, line 23: This entry is related to the daemon work right ? If yes, I think we should exclude it

debian/changelog
```diff
@@ -1,3 +1,52 @@
...
+    - contracts: add query for contract information
```
lucasmoura, line 24: This is related to the simulate token ? If yes, I think we can remove it

debian/changelog
```diff
@@ -1,3 +1,52 @@
...
+    - daemon:
+      + add ua daemon configuration and entrypoint
+      + move the GCP auto-attach job to the daemon
+    - docs:
+      + add the daemon information
+      + fix typos and wrong/outdated information
```
lucasmoura, line 30: I think this entry is unnecessary in the changelog

debian/changelog
```diff
@@ -1,3 +1,52 @@
...
+      + update release process documentation
```
lucasmoura, line 31: Same thing here

debian/changelog
```diff
@@ -1,3 +1,52 @@
...
+    - livepatch: always use the full path in livepatch calls (LP: #1951954)
+    - refactor:
+      + decouple attach, lock and cloud factory logic from the cli
+      + improve inclusive naming best practices
```
lucasmoura, line 35: Same thing here

debian/changelog
```diff
@@ -1,3 +1,52 @@
...
+    - timer: remove timer-based license-check
+    - tools: add impish to the launchpad release script
```
lucasmoura, line 37: And here

debian/changelog
```diff
@@ -1,3 +1,52 @@
...
+    - logs:
+      + improve rules to redact sensitive information from all log files
+      + redact sensitive information from older unredacted log files
+    - tests:
+      + add integration tests for the new scenarios
+      + add manual tests for features not covered by behave
+      + add jammy integration tests
+      + check for the version reported by dpkg when testing
+      + improve mocking for unit tests to avoid leaking and side effects
```
lucasmoura, line 46: And here

lucasmoura (commented):
My rationale for a changelog is that we should include changes that the users care about.
Internal changes that we did on the tool or on our testing infrastructure, do not directly affect the user. That's why I think we should remove them

debian/changelog
renanrodrigo: For those doc entries I see the only exception: I think it is relevant to the users that we keep the docs up to date, and thus I summarized the doc-related commits in a couple lines.

debian/changelog
renanrodrigo: For this one, I don't mind keeping or removing it - it's a matter of will to advertise/register that we are doing the changes. @orndorffgrant, what do you think?

debian/changelog
lucasmoura: I think the daemon information does make sense, but the release process one not.

debian/changelog
renanrodrigo: same docs comment

lucasmoura:
@renanrodrigo You also need to bump the version to 27.5

debian/changelog
orndorffgrant: I think keep this one

debian/changelog
orndorffgrant: And we can drop this one as it kind of counts as "fixing outdated information"

debian/changelog
orndorffgrant: I could go either way, I think its a nice thing but not really relevant to the reader of a changelog since it's only a super tiny refactor. It doesn't really affect end users and the changelog is kind of long already so I guess drop it?

Honestly we could drop the whole refactor section since it doesn't actually change anything

debian/changelog
renanrodrigo: oki

debian/changelog
renanrodrigo: oki

debian/changelog
```diff
@@ -1,3 +1,44 @@
...
+    - logs:
+      + improve rules to redact sensitive information from all log files
+      + redact sensitive information from older unredacted log files
+    - tests:
+      + add integration tests for the new scenarios
+      + add manual tests for features not covered by behave
+      + add jammy integration tests
+      + check for the version reported by dpkg when testing
+
+ -- Renan Rodrigo <renanrodrigo@canonical.com>  Thu, 09 Dec 2021 19:49:30 -0300
```
orndorffgrant, line 40: I think this is supposed to have a dot in it right?
```suggestion
 -- Renan Rodrigo <renan.rodrigo@canonical.com>  Thu, 09 Dec 2021 19:49:30 -0300
```

orndorffgrant (approved):
This is identical to `main` except for the changelog entry and version bump :+1: 

Changelog entry content looks great. Only problem might be the email - I left a comment.

Thanks for pulling this together!

debian/changelog
renanrodrigo: For obscure reasons, I have a dotless email and my signing keys match it.
It's a f̶u̶n̶n̶y̶  long story; I'd keep it as is if you don't mind

orndorffgrant: That's fine with me!

PR #1912, opened 2021-12-14
cli: add json output for enable command (SC-699)

lucasmoura:
## Proposed Commit Message
cli: add json output for enable command

We are now allowing users to modify the output of enable to a machine readable json format. This format will include
all the error and warnings that were generated during the command, while also pointing to which services succeeded and failed during the execution of the command.

## Test Steps
Run the modified integration test.

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -60,6 +60,16 @@ Feature: Enable command behaviour when attached to an UA subscription
             """
             This command must be run as root (try using sudo).
             """
+        And I verify that running `ua enable foobar --format json` `as non-root` exits `1`
+        And stdout is formatted as `json` and has keys:
+            """
+            result processed_services failed_services needs_reboot
+            errors warnings _schema_version
+            """
+        And I will see the following on stdout:
+            """
+            {"_schema_version": 0.1, "errors": [{"message": "This command must be run as root (try using sudo).", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
         And I verify that running `ua enable foobar` `with sudo` exits `1`
         And I will see the following on stdout:
             """
@@ -69,6 +79,11 @@ Feature: Enable command behaviour when attached to an UA subscription
             """
             Cannot enable unknown service 'foobar'.
             Try cc-eal, cis, esm-infra, fips, fips-updates, livepatch.
+            """
+        And I verify that running `ua enable foobar --format json` `with sudo` exits `1`
+        And I will see the following on stdout:
+            """
+            {"_schema_version": 0.1, "errors": [{"message": "Cannot enable unknown service 'foobar'.\nTry cc-eal, cis, esm-infra, fips, fips-updates, livepatch.", "service": null, "type": "system"}], "failed_services": ["foobar"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
             """
         And I verify that running `ua enable ros foobar` `with sudo` exits `1`
         And I will see the following on stdout:
@@ -80,12 +95,22 @@ Feature: Enable command behaviour when attached to an UA subscription
             Cannot enable unknown service 'foobar, ros'.
             Try cc-eal, cis, esm-infra, fips, fips-updates, livepatch.
             """
+        And I verify that running `ua enable ros foobar --format json` `with sudo` exits `1`
+        And I will see the following on stdout:
+        """
+        {"_schema_version": 0.1, "errors": [{"message": "Cannot enable unknown service 'foobar, ros'.\nTry cc-eal, cis, esm-infra, fips, fips-updates, livepatch.", "service": null, "type": "system"}], "failed_services": ["foobar", "ros"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+        """
         And I verify that running `ua enable esm-infra` `with sudo` exits `1`
-        Then I will see the following on stdout:
+        And I will see the following on stdout:
             """
             One moment, checking your subscription first
             UA Infra: ESM is already enabled.
             See: sudo ua status
+            """
+        And I verify that running `ua enable esm-infra --format json` `with sudo` exits `1`
+        Then I will see the following on stdout:
+            """
+            {"_schema_version": 0.1, "errors": [{"message": "UA Infra: ESM is already enabled.\nSee: sudo ua status", "service": "esm-infra", "type": "service"}], "failed_services": ["esm-infra"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
             """
         When I run `apt-cache policy` with sudo
         Then apt-cache policy for the following url has permission `500`
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -51,6 +51,63 @@ Feature: Enable command behaviour when attached to an UA subscription
 
     @series.xenial
     @series.bionic
+    @series.focal
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attached enable of different services using json format
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        Then I verify that running `ua enable foobar --format json` `as non-root` exits `1`
+        And stdout is a json matching the `enable` schema
+        And I will see the following on stdout:
+            """
+            {"_schema_version": "0.1", "errors": [{"message": "This command must be run as root (try using sudo).", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
+        And I verify that running `ua enable foobar --format json` `with sudo` exits `1`
+        And stdout is a json matching the `enable` schema
+        And I will see the following on stdout:
+            """
+            {"_schema_version": "0.1", "errors": [{"message": "Cannot enable unknown service 'foobar'.\nTry <valid_services>", "service": null, "type": "system"}], "failed_services": ["foobar"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
+        And I verify that running `ua enable ros foobar --format json` `with sudo` exits `1`
+        And stdout is a json matching the `enable` schema
+        And I will see the following on stdout:
+        """
+        {"_schema_version": "0.1", "errors": [{"message": "Cannot enable unknown service 'foobar, ros'.\nTry <valid_services>", "service": null, "type": "system"}], "failed_services": ["foobar", "ros"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+        """
+        And I verify that running `ua enable esm-infra --format json` `with sudo` exits `1`
+        Then I will see the following on stdout:
+            """
+            {"_schema_version": "0.1", "errors": [{"message": "UA Infra: ESM is already enabled.\nSee: sudo ua status", "service": "esm-infra", "type": "service"}], "failed_services": ["esm-infra"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
+        When I run `ua disable esm-infra` with sudo
+        And I run `ua enable esm-infra --format json` with sudo
+        Then stdout is a json matching the `enable` schema
+        And I will see the following on stdout:
+        """
+        {"_schema_version": "0.1", "errors": [], "failed_services": [], "needs_reboot": false, "processed_services": ["esm-infra"], "result": "success", "warnings": []}
+        """
+        When I run `ua disable esm-infra` with sudo
+        And I verify that running `ua enable esm-infra foobar --format json` `with sudo` exits `1`
+        Then stdout is a json matching the `enable` schema
+        And I will see the following on stdout:
+        """
+        {"_schema_version": "0.1", "errors": [{"message": "Cannot enable unknown service 'foobar'.\nTry <valid_services>", "service": null, "type": "system"}], "failed_services": ["foobar"], "needs_reboot": false, "processed_services": ["esm-infra"], "result": "failure", "warnings": []}
+        """
+        When I run `ua disable esm-infra esm-apps` with sudo
+        And I run `ua enable esm-infra esm-apps --beta --format json` with sudo
+        Then stdout is a json matching the `enable` schema
+        And I will see the following on stdout:
+        """
+        {"_schema_version": "0.1", "errors": [], "failed_services": [], "needs_reboot": false, "processed_services": ["esm-apps", "esm-infra"], "result": "success", "warnings": []}
+        """
+
+        Examples: ubuntu release
+           | release | valid_services                                         |
+           | xenial  | cc-eal, cis, esm-infra, fips, fips-updates, livepatch. |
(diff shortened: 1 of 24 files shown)


features/attached_enable.feature
```diff
@@ -59,6 +59,16 @@ Feature: Enable command behaviour when attached to an UA subscription
...
             """
+        And I verify that running `ua enable foobar --format json` `as non-root` exits `1`
+        And stdout is formatted as `json` and has keys:
+            """
+            result processed_services failed_services needs_reboot
+            errors warnings
+            """
+        And I will see the following on stdout:
+            """
+            {"errors": [{"message": "This command must be run as root (try using sudo).", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
```
CalvoM, line 70: @lucasmoura just out of curiosity, why are there other fields in addition to message fields, since the issue is running as non-root ?

lucasmoura: @CalvoM It is just for consistency sake. On other situations, those extra fields will be used. We believe it is easier to always return the json with the same structure than to adapt it to each possible failure scenario that we might have when running the command

features/attached_enable.feature
```diff
@@ -59,6 +59,16 @@ Feature: Enable command behaviour when attached to an UA subscription
             """
             This command must be run as root (try using sudo).
             """
+        And I verify that running `ua enable foobar --format json` `as non-root` exits `1`
+        And stdout is formatted as `json` and has keys:
+            """
+            result processed_services failed_services needs_reboot
+            errors warnings
+            """
```
orndorffgrant, line 67: Now might be a good opportunity to start using json schemas in our integration tests instead of checking top-level keys. 
What do you think?

uaclient/cli.py
```diff
@@ -124,9 +127,13 @@ def assert_lock_file(lock_holder=None):
...
-            with lock.SingleAttemptLock(cfg=cfg, lock_holder=lock_holder):
-                retval = f(*args, cfg=cfg, **kwargs)
-            return retval
+            try:
+                with lock.SingleAttemptLock(cfg=cfg, lock_holder=lock_holder):
+                    retval = f(*args, cfg=cfg, **kwargs)
+                return retval
+            except exceptions.LockHeldError as e:
+                event.raise_error_and_finish(exception=e)
+                return 1
```
orndorffgrant, line 136: Can some of the places where this is inserted be removed if we instead call `event.raise_error` in `main_error_handler`?

uaclient/event_logger.py
```diff
@@ -0,0 +1,135 @@
...
+                service=service,
+                event_dict=self._warning_events,
+            )
+
+    def raise_error(self, exception: Exception, service: Optional[str] = None):
+        if self._event_logger_mode == EventLoggerMode.CLI:
+            raise exception
+        else:
+            error_msg = getattr(exception, "msg") or str(exception)
+            self.error(error_msg=error_msg, service=service)
```
orndorffgrant, line 98: Why would we only `raise` if in cli mode? I would think `json` mode shouldn't alter behavior in that way

uaclient/event_logger.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def raise_error_and_finish(
+        self, exception: Exception, service: Optional[str] = None
+    ):
+        self.raise_error(exception=exception, service=service)
+
+        if self._event_logger_mode == EventLoggerMode.MACHINE_READABLE:
+            print(self.process_events())
```
orndorffgrant, line 106: I wouldn't expect this module to be the one calling `print`. in this case, wouldn't this just return a dict or json.dumps string and the caller would call `print`? (This doesn't matter a whole lot, but makes the boundaries clearer for testing IMO).

uaclient/cli.py
```diff
@@ -139,8 +146,26 @@ def assert_root(f):
...
+def set_event_mode(f):
+    """Decorator asserting root user"""
+
+    @wraps(f)
+    def new_f(cmd_args, *args, **kwargs):
+        if cmd_args and cmd_args.format == "json":
+            event.set_event_mode(event_logger.EventLoggerMode.MACHINE_READABLE)
+        ret = f(cmd_args, *args, **kwargs)
+        event.reset()
+        return ret
```
orndorffgrant, line 168: This is nice!

uaclient/event_logger.py
orndorffgrant: Thinking about this more (and assuming `cli` and `json` mode should `raise` in the same situations). I think I'd expect that `raise` shouldn't happen at all in this module. This module can just record the error and the caller can raise if necessary. What do you think?

orndorffgrant (commented):
This is looking really good! Just some comments mostly about what responsibilities this new module should/shouldn't have IMO.

features/attached_enable.feature
lucasmoura: I think that's a great idea, but maybe we should do this in a separate PR, as I think this one is already too long.
My rationale that this will probably require a new dependency for the json schema while also figuring out how we will define the schemas for the json output that we have and how we will address then on the integration test. But maybe it is way simpler than that, so let me know what you think

uaclient/cli.py
lucasmoura: The problem about doing that on `main_error_handler` is that we would need to verify there as well if the `args.format` is a json type to print the json output to the user. But not all commands support that, so it will not gonna that generic

uaclient/event_logger.py
lucasmoura: You mean on the cli function, we would report the error to the event logger and raise it as well ? If yes, my concern is that during the json mode, we could incorrectly print an exception to the user.

Or do you mean that on cli, we should check the format type before raising the exception ?

uaclient/event_logger.py
lucasmoura: Originally, that was my intention, but the problem is that if we do that, we will loose the "real-time" aspect of the cli mode. We would print all the messages at the same time, which can be bad if the user is running a slow command, like enabling fips

features/attached_enable.feature
orndorffgrant: It should be a separate commit, but up to you on whether it can be this PR or a later one. I do think that this set of features (adding json output to many of our commands) is the right time to do it though.

I don't think it will look too complicated.

We can just pull in https://pypi.org/project/jsonschema/ for the integration test dependencies.

Then we write a schema file in `features/schemas/enable.json`

Then we write a new `step` like `And stdout is json matching the "enable" schema`. That looks like (simplified):
```python
validate(
    instance=json.load(context.stdout),
    schema=json.load(open("./features/schemas/{}.json".format(schema_name)).read())
)
```

uaclient/cli.py
orndorffgrant: We default to CLI mode though, and we set the global mode variable to json if the `@set_event_mode` decorator is called and sees the json format option.
By the time we get to `main_error_handler` then the mode will have been set to json or is still at default cli - and I think will be accurate, right? Because for commands that don't support `--format` then it will just default to cli mode.
That means the `print` does have to be _in_ the event module though (contrary to my opinion in a different comment).

uaclient/event_logger.py
orndorffgrant: Right, my comment was more talking about the json output at the end though. But per my comment in the other thread, it is necessary for this print to be in this module as well because it should only happen in machine_readable mode.

uaclient/event_logger.py
orndorffgrant: My main worry is generally that in a given situation, we should either always `raise` or never `raise`. We shouldn't `raise` or not depending on whether the output is going to be cli or json. My reason being: it is hard to tell by reading the code if they actually do the same thing or if, because it doesn't `raise` during json mode, then json mode will actually keep executing further than it would have in cli mode.

I think it should be safe to use the event module in the main_error_handler because at that point the mode should've been set.

uaclient/cli.py
lucasmoura: Yes, you are actually right, I was worried that the `main_error_handler` decorator is called first, but it doesn't matter since the state will still be as MACHINE_READABLE. Good suggestion, I will refactor that bit then

features/attached_enable.feature
lucasmoura: Ohhh fair enough, I though you also meant performing that validation on the code itself.
But you are right, we can start small and just add it to the test. I will do that

orndorffgrant: oh yeah good point I'm not sure if its worth it to do a schema check at runtime. But I think we can gain a lot from adding it to the integration tests.

uaclient/cli.py
```diff
@@ -1477,16 +1518,24 @@ def wrapper(*args, **kwargs):
...
                 logging.error(exc.msg)
-            print("{}".format(exc.msg), file=sys.stderr)
+            event.error(error_msg=exc.msg)
+            event.info(info_msg="{}".format(exc.msg), file_type=sys.stderr)
             if not isinstance(exc, exceptions.LockHeldError):
                 # Only clear the lock if it is ours.
                 lock.clear_lock_file_if_present()
+            event.process_events()
             sys.exit(exc.exit_code)
         except Exception:
```
lucasmoura, line 1540: In case of unexpected exception, should we return then through json or assume that the command should fail just like the cli one ?

lucasmoura:
@orndorffgrant @renanrodrigo marking this one as ready for review now.
Please let me know if you think we need any kind of extra testing here

features/schemas/enable.json
```diff
@@ -0,0 +1,66 @@
...
+            "type": "string"
+           },
+           "service": {
+             "type": ["null", "string"]
+           }
+        },
+        "patternProperties": {
+           "^type$": {
+             "type": "string"
+          }
```
orndorffgrant, line 26: interesting, I haven't seen this before. Is this necessary to have a property with name "type"?

orndorffgrant: We can make this an enum as well

features/attached_enable.feature
```diff
@@ -79,13 +90,23 @@ Feature: Enable command behaviour when attached to an UA subscription
...
             """
             One moment, checking your subscription first
             UA Infra: ESM is already enabled.
             See: sudo ua status
             """
+        And I verify that running `ua enable esm-infra --format json` `with sudo` exits `1`
+        Then I will see the following on stdout:
+            """
+            {"_schema_version": 0.1, "errors": [{"message": "UA Infra: ESM is already enabled.\nSee: sudo ua status", "service": "esm-infra", "type": "service"}], "failed_services": ["esm-infra"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
```
orndorffgrant, line 109: We should probably have some more integration tests. Some scenarios might be:
* successful single service enable
* successful multiple service enable
* multiple service, one success, one failure
* something that results in `needs_reboot: true`

uaclient/cli.py
```diff
@@ -911,14 +944,17 @@ def action_enable(args, *, cfg, **kwargs):
...
         tmpl = ua_status.MESSAGE_INVALID_SERVICE_OP_FAILURE_TMPL
+        event.services_failed(entitlements_not_found)
         raise exceptions.UserFacingError(
             tmpl.format(
                 operation="enable",
                 name=", ".join(entitlements_not_found),
                 service_msg=service_msg,
             )
         )
+        ret = False
```
orndorffgrant, line 955: I'm not sure we'll ever reach this line because of the `raise` above

uaclient/cli.py
orndorffgrant: I think we should always return with the json

uaclient/event_logger.py
```diff
@@ -0,0 +1,152 @@
...
+    CLI = object()
+    MACHINE_READABLE = object()
+
+
+class EventLogger:
+    def __init__(self):
+        self._error_events = []  # type: List[Dict[str, Optional[str]]]
+        self._warning_events = []  # type: List[Dict[str, Optional[str]]]
+        self._processed_services = []  # type: List[str]
+        self._failed_services = set()  # type: Set[str]
```
orndorffgrant, line 47: Why is this one a `set` while the `processed_services` is a list? I would expect them to be the same

uaclient/event_logger.py
```diff
@@ -0,0 +1,152 @@
...
+    def needs_reboot(self, reboot_required: bool):
+        self._needs_reboot = reboot_required
+
+    def _generate_failed_services(self):
+        services_with_error = {
+            error["service"]
+            for error in self._error_events
+            if error["service"]
+        }
+        return list(set.union(self._failed_services, services_with_error))
```
orndorffgrant, line 131: This is a good catchall. I wouldn't have thought of this, nice!

uaclient/exceptions.py
```diff
@@ -72,14 +72,11 @@ class LockHeldError(UserFacingError):
...
-        lock_request = lock_request
-        msg = "Unable to perform: {lock_request}.\n".format(
-            lock_request=lock_request
-        )
-        msg += status.MESSAGE_LOCK_HELD.format(
-            pid=pid, lock_holder=lock_holder
+        super().__init__(
+            status.MESSAGE_LOCK_HELD_ERROR.format(
+                lock_request=lock_request, lock_holder=lock_holder, pid=pid
+            )
```
orndorffgrant, line 78: This is a good refactor, thanks!

orndorffgrant (changes requested):
Theres a ton of great stuff in here! :clap: :clap: 

I left some various small comments.

The main thing I think we need before merging are integration test cases for more scenarios.

features/schemas/enable.json
lucasmoura: Yes, it is a hacky approach I found to allow a field with the name `type`

uaclient/cli.py
lucasmoura: Yes, this is just leftover code. Good catch

uaclient/event_logger.py
lucasmoura: Good point

lucasmoura:
@orndorffgrant code updated

orndorffgrant:
@lucasmoura I think my comments on ` features/schemas/enable.json` could still be addressed, but otherwise looks great!

renanrodrigo (changes requested):
This looks great.
I'm just agreeing with @orndorffgrant 's requests here, and raising a discussion about standardization of the formats.

uaclient/cli.py
```diff
@@ -495,6 +513,13 @@ def enable_parser(parser):
     parser.add_argument(
         "--beta", action="store_true", help="allow beta service to be enabled"
     )
+    parser.add_argument(
+        "--format",
+        action="store",
+        choices=["cli", "json"],
```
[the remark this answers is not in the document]
lucasmoura: I like the idea of changing both `tabular` and `cli` to `plain`.
However, since this change will modify `status` as well, I think we can address this in a separate PR

lucasmoura:
@orndorffgrant @renanrodrigo Sorry for missing the schema comments. I have updated the code now

renanrodrigo (approved):
LGTM!

renanrodrigo:
@lucasmoura there are integration tests failing due to the last change, just noticed them

orndorffgrant (approved):
Looks great! :tada: Let's just let jenkins run

PR #1914, opened 2021-12-15
Change CIS to USG

renanrodrigo:
goodbye `ENTITLEMENT_CLASS_BY_NAME`: Factory-Refactory is a lot of fun (: 
This is another take at the USG implementation, also WIP for pre-reviews.
It is not fully ready yet, still some tests/flows to check/fix.

Achieved so far:
- no changes on the xenial/bionic flows
- both usg and cis valid on focal, but only USG is shown in status
- deprecation note when enabling CIS on focal+
- different packages defined for cis/usg

Not good yet:
- W̶h̶e̶n̶ ̶u̶n̶a̶t̶t̶a̶c̶h̶e̶d̶ ̶o̶n̶ ̶f̶o̶c̶a̶l̶+̶,̶ ̶s̶t̶i̶l̶l̶ ̶s̶h̶o̶w̶i̶n̶g̶ ̶C̶I̶S̶ ̶a̶n̶d̶ ̶h̶e̶l̶p̶ ̶e̶n̶t̶r̶y̶ ̶r̶e̶f̶l̶e̶c̶t̶s̶ ̶o̶n̶l̶y̶ ̶C̶I̶S̶ done
- a̶u̶t̶o̶c̶o̶m̶p̶l̶e̶t̶e̶ ̶s̶u̶f̶f̶e̶r̶s̶ ̶f̶r̶o̶m̶ ̶t̶h̶e̶ ̶s̶a̶m̶e̶ ̶a̶b̶o̶v̶e̶ ̶-̶ ̶C̶I̶S̶ ̶i̶n̶s̶t̶e̶a̶d̶ ̶o̶f̶ ̶U̶S̶G̶  won't do
- U̶S̶G̶ ̶o̶n̶ ̶f̶o̶c̶a̶l̶ ̶s̶t̶i̶l̶l̶ ̶h̶a̶s̶ ̶C̶I̶S̶ ̶t̶e̶x̶t̶ done
- missing unit tests for the new stuff
- m̶i̶s̶s̶i̶n̶g̶ ̶i̶n̶t̶e̶g̶r̶a̶t̶i̶o̶n̶ ̶t̶e̶s̶t̶ ̶f̶i̶x̶e̶s̶

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -59,8 +59,8 @@ MACHINE_TOKEN_FILE="/var/lib/ubuntu-advantage/private/machine-token.json"
 redact_ubuntu_release_from_ua_apt_filenames() {
     DIR=$1
     UA_SERVICES=$(/usr/bin/python3 -c "
-from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
-print(*ENTITLEMENT_CLASS_BY_NAME.keys(), sep=' ')
+from uaclient.entitlements import valid_services
+print(*valid_services(allow_beta=True, all_names=True), sep=' ')
 ")
 
     for file in "$DIR"/*; do
@@ -120,8 +120,8 @@ check_service_is_beta() {
     service_name=$1
     _IS_BETA_SVC=$(/usr/bin/python3 -c "
 from uaclient.config import UAConfig
-from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
-ent_cls = ENTITLEMENT_CLASS_BY_NAME.get('${service_name}')
+from uaclient.entitlements import entitlement_factory
+ent_cls = entitlement_factory('${service_name}')
 if ent_cls:
     cfg = UAConfig()
     allow_beta = cfg.features.get('allow_beta', False)
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -49,7 +49,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         \d+ update(s)? can be applied immediately.
         """
-        When I attach `contract_token` with sudo
+        When I attach `contract_token_staging` with sudo
         Then stdout matches regexp:
         """
         UA Infra: ESM enabled
@@ -62,12 +62,18 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         SERVICE       ENTITLED  STATUS    DESCRIPTION
         cc-eal       +yes      +<cc_status>     +Common Criteria EAL2 Provisioning Packages
-        cis          +yes      +disabled        +Center for Internet Security Audit Tools
+        """
+        And stdout matches regexp:
+        """
         esm-apps     +yes      +enabled  +UA Apps: Extended Security Maintenance \(ESM\)
         esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance \(ESM\)
         fips         +yes      +n/a      +NIST-certified core packages
         fips-updates +yes      +n/a      +NIST-certified core packages with priority security updates
         livepatch    +yes      +n/a      +Canonical Livepatch service
+        """
+        And stdout matches regexp:
+        """
+        <cis_or_usg> +yes      +disabled        +Center for Internet Security Audit Tools
         """
         And stderr matches regexp:
         """
@@ -162,10 +168,10 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
 
         """
         Examples: ubuntu release packages
-           | release | downrev_pkg                 | cc_status |
-           | xenial  | libkrad0=1.13.2+dfsg-5      | disabled  |
-           | bionic  | libkrad0=1.16-2build1       | n/a       |
-           | focal   | hello=2.10-2ubuntu2         | n/a       |
+           | release | downrev_pkg                 | cc_status | cis_or_usg |
+           | xenial  | libkrad0=1.13.2+dfsg-5      | disabled  | cis        |
+           | bionic  | libkrad0=1.16-2build1       | n/a       | cis        |
+           | focal   | hello=2.10-2ubuntu2         | n/a       | usg        |
 
     @series.all
     @uses.config.machine_type.aws.generic
@@ -175,7 +181,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         {
             "machineTokenInfo": {
-                "contractInfo": {
+                "contractInfo":develop {
                     "resourceEntitlements": [
                         {
                             "type": "esm-apps",
@@ -204,11 +210,17 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         SERVICE       ENTITLED  STATUS    DESCRIPTION
         cc-eal       +yes      +<cc_status>     +Common Criteria EAL2 Provisioning Packages
-        cis          +yes      +disabled +Center for Internet Security Audit Tools
+        """
+        And stdout matches regexp:
+        """
         esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance \(ESM\)
         fips         +yes      +<fips_status>      +NIST-certified core packages
         fips-updates +yes      +<fips_status>      +NIST-certified core packages with priority security updates
         livepatch    +yes      +<lp_status>  +<lp_desc>
+        """
+        And stdout matches regexp:
+        """
+        <cis_or_usg>          +yes      +disabled +Center for Internet Security Audit Tools
         """
         And stderr matches regexp:
         """
@@ -216,10 +228,10 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
 
         Examples: ubuntu release livepatch status
-           | release | fips_status |lp_status | lp_desc                       | cc_status |
-           | xenial  | disabled    |enabled   | Canonical Livepatch service   | disabled  |
-           | bionic  | disabled    |enabled   | Canonical Livepatch service   | n/a       |
-           | focal   | n/a         |enabled   | Canonical Livepatch service   | n/a       |
+           | release | fips_status |lp_status | lp_desc                       | cc_status | cis_or_usg |
+           | xenial  | disabled    |enabled   | Canonical Livepatch service   | disabled  | cis        |
+           | bionic  | disabled    |enabled   | Canonical Livepatch service   | n/a       | cis        |
+           | focal   | n/a         |enabled   | Canonical Livepatch service   | n/a       | usg        |
 
     @series.all
     @uses.config.machine_type.azure.generic
@@ -258,11 +270,17 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         SERVICE       ENTITLED  STATUS    DESCRIPTION
         cc-eal       +yes      +<cc_status>     +Common Criteria EAL2 Provisioning Packages
-        cis          +yes      +disabled +Center for Internet Security Audit Tools
+        """
+        And stdout matches regexp:
+        """
         esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance \(ESM\)
         fips         +yes      +<fips_status> +NIST-certified core packages
         fips-updates +yes      +<fips_status> +NIST-certified core packages with priority security updates
         livepatch    +yes      +<lp_status>  +Canonical Livepatch service
+        """
+        And stdout matches regexp:
+        """
+        <cis_or_usg>          +yes      +disabled +Center for Internet Security Audit Tools
         """
         And stderr matches regexp:
         """
@@ -270,10 +288,10 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
 
         Examples: ubuntu release livepatch status
-           | release | lp_status | fips_status | cc_status |
-           | xenial  | n/a       | n/a         | disabled  |
-           | bionic  | n/a       | disabled    | n/a       |
-           | focal   | enabled   | n/a         | n/a       |
+           | release | lp_status | fips_status | cc_status | cis_or_usg |
+           | xenial  | n/a       | n/a         | disabled  | cis        |
+           | bionic  | n/a       | disabled    | n/a       | cis        |
+           | focal   | enabled   | n/a         | n/a       | usg        |
 
     @series.all
     @uses.config.machine_type.gcp.generic
@@ -299,7 +317,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         features:
           machine_token_overlay: "/tmp/machine-token-overlay.json"
         """
-        And I attach `contract_token` with sudo
+        And I attach `contract_token_staging` with sudo
         Then stdout matches regexp:
         """
         UA Infra: ESM enabled
@@ -312,11 +330,17 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         SERVICE       ENTITLED  STATUS    DESCRIPTION
         cc-eal       +yes      +<cc_status>     +Common Criteria EAL2 Provisioning Packages
-        cis          +yes      +disabled +Center for Internet Security Audit Tools
+        """
+        And stdout matches regexp:
+        """
         esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance \(ESM\)
         fips         +yes      +<fips_status> +NIST-certified core packages
         fips-updates +yes      +<fips_status> +NIST-certified core packages with priority security updates
         livepatch    +yes      +<lp_status>  +Canonical Livepatch service
+        """
+        And stdout matches regexp:
+        """
+        <cis_or_usg>          +yes      +disabled +Center for Internet Security Audit Tools
         """
         And stderr matches regexp:
         """
@@ -324,7 +348,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
 
         Examples: ubuntu release livepatch status
-           | release | lp_status | fips_status | cc_status |
-           | xenial  | n/a       | n/a         | disabled  |
-           | bionic  | n/a       | disabled    | n/a       |
-           | focal   | enabled   | n/a         | n/a       |
+           | release | lp_status | fips_status | cc_status | cis_or_usg |
+           | xenial  | n/a       | n/a         | disabled  | cis        |
+           | bionic  | n/a       | disabled    | n/a       | cis        |
+           | focal   | enabled   | n/a         | n/a       | usg        |
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -14,7 +14,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
             """
             SERVICE       ENTITLED  STATUS    DESCRIPTION
             cc-eal        +yes      +n/a      +Common Criteria EAL2 Provisioning Packages
-            cis           +yes      +n/a      +Center for Internet Security Audit Tools
+            cis           +yes      +n/a      +Security compliance and audit tools
             esm-apps      +yes      +n/a      +UA Apps: Extended Security Maintenance \(ESM\)
             esm-infra     +yes      +n/a      +UA Infra: Extended Security Maintenance \(ESM\)
             fips          +yes      +n/a      +NIST-certified core packages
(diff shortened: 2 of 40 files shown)


uaclient/entitlements/__init__.py
```diff
@@ -23,12 +24,14 @@
...
 
 
-ENTITLEMENT_CLASS_BY_NAME = dict(
-    (cast(str, cls.name), cls) for cls in ENTITLEMENT_CLASSES
-)  # type: Dict[str, Type[UAEntitlement]]
+def entitlement_factory(name: str) -> UAEntitlement:
+    for entitlement in ENTITLEMENT_CLASSES:
+        if name in entitlement().valid_names:
+            return entitlement
+    return None
```
blackboxsw, line 31: I think that your creation of a factory function here is good and explicit even though it involved having to add a "valid_names" method on classes and call that method inside `valid_services` function.

An alternative to dropping ENTITLEMENT_CLASS_BY_NAME would be defining aliases keys inside  ENTITLEMENT_CLASS_BY_NAME as such:
```python
{
 'usg': CISEntitlement,
 'cis': CISEntitlement,
} 
```
Having aliases represented in  ENTITLEMENT_CLASS_BY_NAME you'd be able to avoid the `Entitlement.valid_names` class method I think, but you'd add complexity to all call-sites (like `valid_services`) to walk the whole ENTITLEMENT_CLASS_BY_NAME dict and compare values to find all valid_names for a given entitlement. The alternative I suggest seems a bit more implicit a construct than what you have, so I agree with your approach here.

uaclient/entitlements/base.py
```diff
@@ -69,6 +77,16 @@ def description(self) -> str:
...
         pass
 
+    @property
+    def presentation_name(self) -> str:
+        """The user-facing name shown for this entitlement"""
+        return (
+            self.cfg.entitlements.get(self.name, {})
+            .get("entitlement", {})
+            .get("affordances", {})
+            .get("presentAs", self.name)
```
blackboxsw, line 87: good it's clear and falls back to defined original entitlement.name in absence of presentAs.

blackboxsw (commented):
quick drive-by comment on this. This approach looks more sustainable than the other PR. Will look more tomorrow.

uaclient/config.py
```diff
@@ -670,11 +670,13 @@ def _unattached_status(self) -> Dict[str, Any]:
 
             response["services"].append(
                 {
-                    "name": resource["name"],
+                    "name": resource.get("presentAs", resource["name"]),
                     "description": ent_cls.description,
                     "available": available,
                 }
             )
+        response["services"].sort(key=lambda x: x.get("name", ""))
```
lucasmoura, line 678: Is there a reason to perform the sort here instead of the current approach ?

uaclient/config.py
```diff
@@ -752,13 +754,14 @@ def _attached_status(self) -> Dict[str, Any]:
...
-        for resource in sorted(resources, key=lambda x: x.get("name", "")):
-            ent_cls = ENTITLEMENT_CLASS_BY_NAME.get(resource.get("name", ""))
+        for resource in resources:
+            ent_cls = entitlement_factory(resource.get("name", ""))
             if ent_cls:
                 ent = ent_cls(self)
                 response["services"].append(
                     self._attached_service_status(ent, inapplicable_resources)
                 )
+        response["services"].sort(key=lambda x: x.get("name", ""))
```
lucasmoura, line 764: Same question here

uaclient/entitlements/__init__.py
```diff
@@ -23,12 +24,14 @@
...
 
 
-def valid_services(allow_beta: bool = False) -> List[str]:
+def valid_services(allow_beta: bool = False, all_names: bool = False) -> List[str]:
     """Return a list of valid (non-beta) services.
 
     @param allow_beta: if we should allow beta services to be marked as valid
```
lucasmoura, line 37: Missing `all_names` param

uaclient/entitlements/__init__.py
```diff
@@ -38,12 +41,15 @@ def valid_services(allow_beta: bool = False) -> List[str]:
...
-            ent_name
-            for ent_name, ent_cls in ENTITLEMENT_CLASS_BY_NAME.items()
-            if not ent_cls.is_beta
-        ]
-    )
+        entitlements = ENTITLEMENT_CLASSES
+    else:
+        entitlements = [entitlement for entitlement in ENTITLEMENT_CLASSES if not entitlement.is_beta]
+
+    if all_names:
```
lucasmoura, line 48: What situation we would need all names to be returned ?

uaclient/entitlements/__init__.py
```diff
@@ -38,12 +41,15 @@ def valid_services(allow_beta: bool = False) -> List[str]:
...
-    return sorted(
-        [
-            ent_name
-            for ent_name, ent_cls in ENTITLEMENT_CLASS_BY_NAME.items()
-            if not ent_cls.is_beta
-        ]
-    )
+        entitlements = ENTITLEMENT_CLASSES
+    else:
+        entitlements = [entitlement for entitlement in ENTITLEMENT_CLASSES if not entitlement.is_beta]
```
lucasmoura, line 46: I think we can remove the first if check:
```python
entitlements = ENTITLEMENT_CLASSES
if not allow_beta:
    entitlements = [entitlement for entitlement in entitlements if not entitlement.is_beta]
```

uaclient/entitlements/base.py
lucasmoura: What happens if the user is `unattached` ? On Focal, we would return `cis` then ?

uaclient/config.py
renanrodrigo: Yes - CIS and USG appear in different order and we don't know which one will be there before looping.
If we sort first by name, and then the `presentAs` is different, it ends up in the wrong position.

uaclient/config.py
renanrodrigo: same ^

uaclient/entitlements/__init__.py
renanrodrigo: In cases we are interested in any valid service name, not just the presented one - the line on postinst for instance would need visibility to `cis` even in newer series

uaclient/entitlements/__init__.py
blackboxsw: This feels like stylistic changes, and also I think resetting a variable twice is a bit more costly than setting it once in different conditional paths. I get that it looks like less code to read, but timeit seems to think this suggestion is less desirable. Do folks have a strong opinion otherwise about this logical construct or whether one is more pythonic than the other?

```bash
>>> print(orig)
all_names = True
if all_names:
 a = 1
else:
 a = 2

>>> print(new)
all_names = True
a = 1
if all_names:
 a = 2

>>> timeit.timeit(stmt=orig, number=500000000)
11.050870887003839
>>> timeit.timeit(stmt=new, number=500000000)
13.152834953012643
```

lucasmoura: That's really interesting, I always thought that the cost of the conditional was heavier that setting up a variable. I see no problem keeping the version we have with this new evidence

uaclient/entitlements/__init__.py
```diff
@@ -23,12 +24,14 @@
 ]  # type: List[Type[UAEntitlement]]
 
 
-ENTITLEMENT_CLASS_BY_NAME = dict(
-    (cast(str, cls.name), cls) for cls in ENTITLEMENT_CLASSES
-)  # type: Dict[str, Type[UAEntitlement]]
+def entitlement_factory(name: str) -> UAEntitlement:
```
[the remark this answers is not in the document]
orndorffgrant: Now that I submitted I'm leaning towards my second suggestion here anyway, so no change needed for this PR.

uaclient/contract.py
```diff
@@ -383,7 +383,7 @@ def process_entitlement_delta(
                 )
             )
         try:
-            ent_cls = ENTITLEMENT_CLASS_BY_NAME[name]
+            ent_cls = entitlement_factory(name)
         except KeyError:
```
orndorffgrant, line 387: @renanrodrigo Looks like this will need to change so that instead of catching a `KeyError` it checks if `ent_cls` is None

renanrodrigo: already done, hadn't tackled it yet in this WIP version (:

uaclient/entitlements/__init__.py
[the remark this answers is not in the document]
renanrodrigo: Option 2 seems good to me

debian/ubuntu-advantage-tools.postinst
```diff
@@ -59,8 +59,8 @@ MACHINE_TOKEN_FILE="/var/lib/ubuntu-advantage/private/machine-token.json"
 redact_ubuntu_release_from_ua_apt_filenames() {
     DIR=$1
     UA_SERVICES=$(/usr/bin/python3 -c "
-from uaclient.entitlements import ENTITLEMENT_CLASS_BY_NAME
-print(*ENTITLEMENT_CLASS_BY_NAME.keys(), sep=' ')
+from uaclient.entitlements import valid_services
+print(*valid_services(allow_beta=True, all_names=True), sep=' ')
 ")
```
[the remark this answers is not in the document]
renanrodrigo: Right now I agree, let's check if we can change it in future iterations I guess

orndorffgrant:
@renanrodrigo for next year. Please notice I did touch the simulated status code as well as some tests in one of my commits :)

renanrodrigo:
@blackboxsw @lucasmoura @orndorffgrant 
- Fixed the remaining help behavior
- Changed entitlement overall text to be USG flavoured
- Adapted the behave tests - thank you very much again for your help on that.

We can squash+merge commits in a single one when merging, so I am sending a separate commit here.

uaclient/tests/test_update_messaging.py
```diff
@@ -605,18 +616,16 @@ def test_message_based_on_beta_status_and_count_until_active_esm(
             cfg.override_features({"allow_beta": cfg_allow_beta})
 
         m_entitlement_cls = mock.MagicMock()
-        type(m_entitlement_cls).is_beta = is_beta
+        m_entitlement_cls.is_beta = is_beta
```
blackboxsw, line 619: Is there a reason we want to change this from a class type declaration to a direct attribute set?
I **think** the right way to do this is with [mock.PropertyMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock).

I think that would result in:
```suggestion
        type(m_entitlement_cls).is_beta = mock.PropertyMock(return_value=is_beta)
```

features/unattached_status.feature
```diff
@@ -32,16 +32,18 @@ Feature: Unattached status
...
         Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `sed -i 's/contracts.can/contracts.staging.can/' /etc/ubuntu-advantage/uaclient.conf` with sudo
         When I run `ua status` as non-root
         Then stdout matches regexp:
             """
             SERVICE       AVAILABLE  DESCRIPTION
             cc-eal        <cc-eal>    +Common Criteria EAL2 Provisioning Packages
-            cis           <cis>       +Center for Internet Security Audit Tools
-            esm-infra     <infra>     +UA Infra: Extended Security Maintenance \(ESM\)
+            ?<cis>( +<cis-available> +Security compliance and audit tools)?
```
blackboxsw, line 41: Oooh sneaky TIL about full line optional regexes in behave. Nice one

blackboxsw (approved):
I'm happy with this iteration @renanrodrigo. If you have to touch this branch for nits I have a list of comments inline. My consolidated diff for nit/comments is below if useful. Only thing to be wary of us we are not pointing to https://ubuntu.com/security/certifications/docs/usg which still doesn't exist.

Behaved as expected on bionic and looks to work well on focal too:
```bash
root@dev-b:~# ua enable usg
Service cis is recommended by default. Run: sudo ua enable cis
Cannot enable unknown service 'usg'.
Try cc-eal, cis, esm-infra, fips, fips-updates, livepatch.
root@dev-b:~# ua enable cis
Updating package lists
Installing CIS Audit packages
CIS Audit enabled
Visit https://ubuntu.com/security/cis to learn how to use CIS
```

# diff of nit comments
```diff
diff --git a/features/attach_validtoken.feature b/features/attach_validtoken.feature
index b9eeca3f..ec4400b5 100644
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -181,7 +181,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         {
             "machineTokenInfo": {
-                "contractInfo":develop {
+                "contractInfo": {
                     "resourceEntitlements": [
                         {
                             "type": "esm-apps",
diff --git a/help_data.yaml b/help_data.yaml
index d5d6486c..85843ff6 100644
--- a/help_data.yaml
+++ b/help_data.yaml
@@ -7,7 +7,7 @@ cc-eal:
 
 cis:
     help: |
-      Ubuntu Security Guide is a tool for hardening and auditing and allows for
+      Ubuntu Security Guides is a tool for hardening and auditing and allows for
       environment-specific customizations. It enables compliance with profiles
       such as DISA-STIG and the CIS benchmarks. Find out more at
       https://ubuntu.com/security/certifications/docs/usg
diff --git a/uaclient/entitlements/tests/conftest.py b/uaclient/entitlements/tests/conftest.py
index 692dc9f7..03f56aff 100644
--- a/uaclient/entitlements/tests/conftest.py
+++ b/uaclient/entitlements/tests/conftest.py
@@ -123,7 +123,7 @@ def entitlement_factory(tmpdir):
         if services_once_enabled:
             cfg.write_cache("services-once-enabled", services_once_enabled)
 
-        args = {"allow_beta": allow_beta, "called_name": called_name or None}
+        args = {"allow_beta": allow_beta, "called_name": called_name}
         if assume_yes is not None:
             args["assume_yes"] = assume_yes
         return cls(cfg, **args)
diff --git a/uaclient/entitlements/tests/test_entitlements.py b/uaclient/entitlements/tests/test_entitlements.py
index bcd0d3e2..eab56b92 100644
--- a/uaclient/entitlements/tests/test_entitlements.py
+++ b/uaclient/entitlements/tests/test_entitlements.py
@@ -16,15 +16,15 @@ class TestValidServices:
         m_is_config_value.return_value = allow_beta
 
         m_cls_1 = mock.MagicMock()
+        type(m_cls_1).is_beta = mock.PropertyMock(return_value=False)
         m_inst_1 = mock.MagicMock()
-        m_cls_1.is_beta = False
         m_inst_1.presentation_name = "ent1"
         m_inst_1.valid_names = ["ent1", "othername"]
         m_cls_1.return_value = m_inst_1
 
         m_cls_2 = mock.MagicMock()
+        type(m_cls_2).is_beta = mock.PropertyMock(return_value=is_beta)
         m_inst_2 = mock.MagicMock()
-        m_cls_2.is_beta = is_beta
         m_inst_2.presentation_name = "ent2"
         m_inst_2.valid_names = ["ent2"]
         m_cls_2.return_value = m_inst_2

```

features/attach_validtoken.feature
```diff
@@ -175,7 +181,7 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
         """
         {
             "machineTokenInfo": {
-                "contractInfo": {
+                "contractInfo":develop {
```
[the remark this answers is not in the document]
renanrodrigo: ack agree fixed

help_data.yaml
```diff
@@ -7,17 +7,11 @@ cc-eal:
...
-      prevent cyber attacks and malware, auditing privileged operations and
-      restricting administrative privileges. The cis command installs
-      tooling needed to automate audit and hardening according to a desired
-      CIS profile - level 1 or level 2 for server or workstation on
-      Ubuntu 18.04 LTS or 16.04 LTS. The audit tooling uses OpenSCAP libraries
-      to do a scan of the system. The tool provides options to generate a
-      report in XML or a html format. The report shows compliance for all the
-      rules against the profile selected during the scan. You can find out
-      more at https://ubuntu.com/security/certifications#cis
+      Ubuntu Security Guide is a tool for hardening and auditing and allows for
```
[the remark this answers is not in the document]
renanrodrigo: I got this in singular from Nikos, copypasted it - but agree it makes more sense plural, it's plural in other places

uaclient/entitlements/cis.py
```diff
@@ -2,22 +2,47 @@
 
 from uaclient.entitlements import repo
 
-CIS_DOCS_URL = "https://ubuntu.com/security/certifications/docs/cis"
+CIS_DOCS_URL = "https://ubuntu.com/security/cis"
+USG_DOCS_URL = "https://ubuntu.com/security/certifications/docs/usg"
```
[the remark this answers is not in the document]
renanrodrigo: s/we/security - this is a topic for the next release meeting. The URL is enforced from their side.

uaclient/entitlements/tests/conftest.py
```diff
@@ -122,7 +123,7 @@ def factory_func(
         if services_once_enabled:
             cfg.write_cache("services-once-enabled", services_once_enabled)
 
-        args = {"allow_beta": allow_beta}
+        args = {"allow_beta": allow_beta, "called_name": called_name or None}
```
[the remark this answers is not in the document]
renanrodrigo: :man_facepalming:

uaclient/tests/test_update_messaging.py
renanrodrigo: I see it was like this before.
From the mocking manipulation perspective, it will be the same.
From the semantics, having the property mocks makes more sense - the `type(` thing looked ugly to me.

features/unattached_status.feature
renanrodrigo: This is work from the regex magician which helped me last time I was stuck, and then he wrote this. His name is @orndorffgrant I guess not sure if you heard about him (:

renanrodrigo:
@lucasmoura addressed

uaclient/entitlements/__init__.py
orndorffgrant: 19 day old conversation :upside_down_face: but the performance of a branch is also dependent on how well it can be predicted by the cpu. The example scenario branch will always be guessed correctly so it will be fast. However, the performance difference is so minor that I don't think it matters here :smile:

uaclient/config.py
```diff
@@ -342,13 +342,19 @@ def entitlements(self):
...
+
         tokens_by_name = dict(
-            (e["type"], e["token"])
+            (e.get("type"), e.get("token"))
             for e in machine_token.get("resourceTokens", [])
         )
         ent_by_name = dict(
-            (e["type"], e) for e in contractInfo["resourceEntitlements"]
+            (e.get("type"), e)
+            for e in contractInfo.get("resourceEntitlements", [])
```
lucasmoura, line 357: I think the default value here should be a dict, right ?

uaclient/config.py
```diff
@@ -342,13 +342,19 @@ def entitlements(self):
...
+        contractInfo = machine_token.get("machineTokenInfo", {}).get(
+            "contractInfo"
+        )
+        if not contractInfo:
+            return {}
+
         tokens_by_name = dict(
-            (e["type"], e["token"])
+            (e.get("type"), e.get("token"))
             for e in machine_token.get("resourceTokens", [])
```
lucasmoura, line 353: Same thing here for the default value

uaclient/config.py
renanrodrigo: No, `resourceEntitlements` is a list

uaclient/config.py
renanrodrigo: `resourceTokens` is also a list

uaclient/entitlements/__init__.py
renanrodrigo: Premature optimization is evil or something, I don't remember

features/ubuntu_upgrade.feature
```diff
@@ -162,3 +162,44 @@ Feature: Upgrade between releases when uaclient is attached
...
+    @uses.config.machine_type.lxd.container
+    @upgrade
+    Scenario Outline: Attached upgrade with cis enabled across LTS releases
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token_staging` with sudo
+        And I run `ua enable cis` with sudo
+        # update-manager-core requires ua < 28. Our tests that build the package will
+        # generate ua with version 28. We are removing that package here to make sure
+        # do-release-upgrade will be able to run
+        And I run `apt remove update-manager-core -y` with sudo
```
orndorffgrant, line 177: No change needed for this PR, but this discovery is concerning. We may need to sync with the update-manager team to see why this is so, and to talk about versioning/backwards-compat going forward.

renanrodrigo: @orndorffgrant worth a card IMHO

orndorffgrant: https://warthogs.atlassian.net/browse/SC-703

orndorffgrant (approved):
Looks really good, thanks for all the work here @renanrodrigo 
Let's just let Jenkins run on it

renanrodrigo:
@orndorffgrant @lucasmoura I changed the failing test to accept a regex for the URL, and we good to go

PR #1916, opened 2021-12-17
Make sbuild 6x faster with this one simple trick!

orndorffgrant:
## Proposed Commit Message

```
builds: script to seed sbuild chroots with dependencies
```

## Test Steps

```
time ./tools/build.sh xenial # notice how long it takes
rm /tmp/uaclient-behave/deb-cache/xenial-* # clear build cache
env RELEASES="xenial" ARCHS="amd64" OPERATION="create" bash ./tools/setup_sbuild.sh
time ./tools/build.sh xenial # notice how long it takes
```
On my laptop, the second build is about 6x faster :D

## Desired commit type

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,9 +368,9 @@ when you should set UACLIENT_BEHAVE_SNAPSHOT_STRATEGY=1
 | ------------- | ------------------ |
 | lxd.container | num_scenarios > 7  |
 | lxd.vm        | never              |
-| gcp*          | num_scenarios > 5  |
-| azure*        | num_scenarios > 14 |
-| aws*          | num_scenarios > 11 |
+| gcp           | num_scenarios > 5  |
+| azure         | num_scenarios > 14 |
+| aws           | num_scenarios > 11 |
 
 #### Integration testing on EC2
 The following tox environments allow for testing focal on EC2:
@@ -521,6 +521,14 @@ sbuild --dist=<target> ../ubuntu-advantage-tools_*.dsc
 # emulating different architectures in sbuild-launchpad-chroot
 sbuild-launchpad-chroot create --architecture="riscv64" "--name=focal-riscv64" "--series=focal
 ```
+
+> Note: Every so often, it is recommended to update your chroots.
+> ```bash
+> # to update a single chroot
+> sudo sbuild-launchpad-chroot update -n ua-xenial-amd64
+> # this script can be used to update all chroots
+> sudo PATTERN=\* sh /usr/share/doc/sbuild/examples/sbuild-debian-developer-setup-update-all
+> ```
 
 ### Setting up an lxc development container
 ```shell
--- a/README.md
+++ b/README.md
@@ -510,6 +510,10 @@ you're on:
 [configure sbuild](https://wiki.ubuntu.com/SimpleSbuild) and
 use that for the build:
 
+Setup some chroots for sbuild with this script
+```shell
+bash ./tools/setup_sbuild.sh
+```
 
 ```shell
 debuild -S
--- a/features/util.py
+++ b/features/util.py
@@ -8,7 +8,7 @@ import tempfile
 import textwrap
 import time
 from contextlib import contextmanager
-from typing import Iterable, List
+from typing import Iterable, List, Optional
 
 import yaml
 
@@ -153,7 +153,9 @@ def repo_state_hash(
     return hashlib.md5(output_to_hash).hexdigest()
 
 
-def build_debs(series: str, cache_source: bool = False) -> List[str]:
+def build_debs(
+    series: str, cache_source: bool = False, chroot: Optional[str] = None
+) -> List[str]:
     """
     Build the package through sbuild and store the debs into
     output_deb_dir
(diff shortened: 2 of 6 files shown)


tools/setup_sbuild.sh
```diff
@@ -0,0 +1,70 @@
...
+
+set -x
+
+create_sbuild_chroot() {
+    release=$1
+    arch=$2
+    echo "creating $release $arch"
+    name="$release-$arch"
+    sudo sbuild-launchpad-chroot create --architecture="$arch" "--name=$name" "--series=$release"
+    sudo schroot -c source:$name -u root -d / -- sh -c "grep -q updates /etc/apt/sources.list || echo \"deb http://archive.ubuntu.com/ubuntu/ $release-updates main restricted universe multiverse\" >> /etc/apt/sources.list"
```
lucasmoura, line 45: Those commands will fail if for some reason we have another chroot with an alias of `$series-$arch`.
However, I don't know if we can actually address this here

lucasmoura (commented):
That looks great to me. I can confirm the speedup here.
Maybe we could improve how we delete the `chroot`, but it is just a minor nit.

@paride can you also take a look on this one to see if there is anything we are missing here about `chroot` usage

tools/setup_sbuild.sh
```diff
@@ -0,0 +1,70 @@
...
+    sudo schroot -c source:$name -u root -d / -- apt-get update
+    sudo schroot -c source:$name -u root -d / -- apt-get dist-upgrade -y
+    sudo schroot -c source:$name -u root -d / -- apt-get install -y make dpkg-dev git devscripts equivs
+    sudo schroot -c source:$name -u root -d / -- git clone --depth 1 https://github.com/CanonicalLtd/ubuntu-advantage-client /var/tmp/uac
```
orndorffgrant, line 49: Oops this is the wrong github org, will fix

orndorffgrant: Fixed

paride (changes requested):
This is a good idea! I left a couple of inline comments, two more general ones are:

1. We need to remember to update the source chroot from time to time, otherwise it will become outdated over time and thus slow. This can be done via `sbuild-launchpad-chroot update` or `sbuild-update -udcar`. The Jenkins nodes auto-update al the chroots periodically, but I think this should be documented.
2. It's nice if we can make the new script shellcheck clean (`shellcheck setup_sbuild.sh`).

tools/setup_sbuild.sh
orndorffgrant: Yeah these are great points. I think you're right that the correct way to do this would be to expose a `chroot` argument to the build.py script, as well as the integration tests.

Something about the current setup that we should try to keep, though, is that ~97% of sbuilds on our dev machines are test builds and should use the `ua-$release-$arch` chroot. So I think the default sbuild for both `build.sh` and integration tests should be to use `ua-$release-$arch`. Then a `chroot` argument can be used to select the "standard" chroot when needed (i.e. probably for jenkins runs).

So in example form, my suggestion is:
```
./tools/setup_sbuild.sh
```
will create  `ua-$release-$arch`

And
```
./tools/build.sh xenial
tox -e behave-lxd-16.04
```
will both use `ua-$release-$arch`

Then a new argument will be added so that something like
```
./tools/build.sh xenial --chroot=xenial-amd64
env UACLIENT_BEHAVE_SBUILD_CHROOT=xenial-amd64 tox -e behave-lxd-16.04
```
can be used to use the standard chroot

What do you think @paride @lucasmoura ?

orndorffgrant:
@paride Thanks for the feedback!

1. You're absolutely right and I definitely hadn't updated mine in a long time until I started on this PR. What do you suggest we do here. Should we just add this suggestion to the readme?
2. Excellent point - Yes I'll run shellcheck on this.

tools/setup_sbuild.sh
lucasmoura: I think that makes total sense. +1

paride: Makes sense. The other possible approach I see is making `build.py` check if `ua-$release-$arch` exists, in this case pass it to `sbuild` via `--chroot`. If it doesn't exist, no `--chroot` is specified and sbuild will go with its defaults.

This has the advantage of making `build.py` work out of the box without the specia `ua-` chroot.

To check if the chroot exists we can grep the output of `schroot --list` for the right `source:` chroot. I know no better way.

paride:
I thought a bit about how to deal with updating the source chroot. I don't like the idea of updating it every time a build is done:

1. It slows things down, and the whole point here is getting faster builds
2. It requires root permissions, which is better to avoid when possible, especially in automation
3. What happens if we launch more than one build on the same system?

I think it's enough to document in the README the need to upgrade the chroot from time to time. That's done automatically on the Jenkins nodes via a modified version of the `/usr/share/doc/sbuild/examples/sbuild-debian-developer-setup-update-all` script. (I have to check that it will also upgrade chroots with nonstandard names, but that's it.)

orndorffgrant:
@lucasmoura @paride Thanks for the follow ups! I've just pushed an update that I believe addresses everything. Please let me know what you think

features/util.py
```diff
@@ -231,19 +233,35 @@ def build_debs(series: str, cache_source: bool = False) -> List[str]:
...
+    else:
+        # use ua-series-arch chroot if present
+        ua_chroot = "ua-{}-amd64".format(series)
+        proc = subprocess.Popen(
+            ["schroot", "--list"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        out, _ = proc.communicate(timeout=1)
+        if "source:{}".format(ua_chroot) in out.decode("utf-8"):
```
lucasmoura, line 259: Instead of searching for the chroot, we could instead run this:
`schroot -i -c source:ua-xenial-amd64`

orndorffgrant: oh sweet, I'll change it to use that

orndorffgrant: @lucasmoura updated!

features/util.py
```diff
@@ -231,19 +233,34 @@ def build_debs(series: str, cache_source: bool = False) -> List[str]:
...
+        "--no-run-lintian",
+        "--resolve-alternatives",
+        "--no-clean-source",
+        "--arch",
+        "amd64",
+        "-d",
+        series,
+        SOURCE_PR_UNTAR_DIR,
+    ]
+    if chroot is not None:
```
lucasmoura, line 248: I think we can also log which chroot is being used

orndorffgrant: Indeed! I changed the log call below to include the entire sbuild command, which includes the chroot flag if present

README.md
```diff
@@ -518,6 +522,14 @@ sbuild --dist=<target> ../ubuntu-advantage-tools_*.dsc
 sbuild-launchpad-chroot create --architecture="riscv64" "--name=focal-riscv64" "--series=focal
 ```
 
+> Note: Every so often, it is recommended to update your chroots.
+> ```bash
+> # to update a single chroot
+> sudo sbuild-launchpad-chroot update -n ua-xenial-amd64
+> # this script can be used to update all chroots
+> sudo env PATTERN=\* sh /usr/share/doc/sbuild/examples/sbuild-debian-developer-setup-update-all
```
paride, line 530: Nit: `sudo` accepts env variables (`sudo VAR=value <command>`), so we don't need `env`.

paride (approved):
Looks great, thanks! Green light from me on merging if you think it's ready.

lucasmoura:
I have tested the change and everything is working fine

lucasmoura (approved):

README.md
orndorffgrant: TIL! Thanks! Will fix this then merge

PR #1918, opened 2022-01-04
Release 27.5

renanrodrigo:
Few commits + changelog for the release 27.5, without the daemon and with USG.
Not all commits described in the changelog are in the PR, because some of them were merged before the force-push to this target branch.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,31 @@
+ubuntu-advantage-tools (27.5~22.04.1) jammy; urgency=medium
+
+  * d/control:
+    - Update homepage URL
+  * d/tools.postinst:
+    - Refactor to use valid_services
+  * New upstream release 27.5:
+    - aws: add support for the IPv6 metadata endpoint
+    - cis: update URL for the documentation
+    - cli:
+      + add endpoint to simulate the status using a specific contract token
+      + fix return code when attaching an already attached machine (GH: #1867)
+      + fix security-status to consider all possible origins to show updates
+      + include cloud build.info in the collect-logs tarball
+      + only show services which exist in the contracts server in ua status
+    - docs: fix typos and wrong/outdated information
+    - livepatch: always use the full path in livepatch calls (LP: #1951954)
+    - logs:
+      + improve rules to redact sensitive information from all log files
+      + redact sensitive information from older unredacted log files
+      + log errors from external software execution, for debugging purposes
+    - usg:
+      + support the presentedAs affordance from the contract server, showing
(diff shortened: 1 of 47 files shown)


debian/changelog
```diff
@@ -1,3 +1,35 @@
...
+      + redact sensitive information from older unredacted log files
+    - usg:
+      + support the presentedAs affordance from the contract server, showing
+        services in the CLI with the appropriate names
+      + replace the CIS entitlement by USG on Focal and onwards
+    - tests:
+      + add integration tests for the new scenarios
+      + add manual tests for features not covered by behave
+      + add jammy integration tests
+      + check for the version reported by dpkg when testing
```
lucasmoura, line 29: I don't think any of these test entries are useful to the user

debian/changelog
```diff
@@ -1,3 +1,35 @@
...
+      + add endpoint to simulate the status using a specific contract token
+      + fix return code when attaching an already attached machine (GH: #1867)
+      + fix security-status to consider all possible origins to show updates
+      + include cloud build.info in the collect-logs tarball
+      + only show services which exist in the contracts server in ua status
+    - docs: fix typos and wrong/outdated information
+    - livepatch: always use the full path in livepatch calls (LP: #1951954)
+    - logs:
+      + improve rules to redact sensitive information from all log files
+      + redact sensitive information from older unredacted log files
```
lucasmoura, line 20: Aren't we missing the new log commits here ?

debian/changelog
renanrodrigo: Those are exactly the same that made up to the last release.
I agree that for a regular user they would not be that useful, but #1 and #2 are relevant for an SRU reviewer I think. #4 was a request by Robie. In my opinion, we are just making it explicit here, and it is ok.
@orndorffgrant may I call for a third opinion so we decide on this?

debian/changelog
renanrodrigo: just one - added it (:

debian/changelog
lucasmoura: On my opinion, the changelog should only inform about meaningful changes for the end user. Test changes on UA are not that significant on my opinion, that's why I avoid adding many of our change on it. But l agree a third opinion here is a good call

orndorffgrant: I vote delete

renanrodrigo: delete it is

done!

orndorffgrant (approved):

lucasmoura (approved):

PR #1924, opened 2022-01-07
wip: cli: attach config file

orndorffgrant:
## Proposed Commit Message

```
parsing: dataobject classes to facilitate parsing data files
```
```
attach: new attach-config file to customize attach operations
```

## Test Steps

Run `ua attach --attach-config /path/to/file.yaml` with varying configs:
* with invalid config (wrong field name, invalid data type, missing token, etc)
* with varying enable_services lists, including invalid services

## Desired commit type

## Checklist


uaclient/cli.py
```diff
@@ -1073,22 +1083,72 @@ def action_auto_attach(args, *, cfg):
...
+            ua_status.MESSAGE_ATTACH_TOKEN_ARG_XOR_CONFIG
         )
+
+    enable_services_override = None
+    if args.token:
+        token = args.token
+    else:
+        try:
+            attach_config = AttachActionsConfigFile.from_dict(
+                yaml.safe_load(args.attach_config)
```
lucasmoura, line 1101: Should we have an exception in case the file doesn't exist ?

uaclient/cli.py
```diff
@@ -1073,22 +1083,72 @@ def action_auto_attach(args, *, cfg):
...
+            raise exceptions.UserFacingError(
+                textwrap.fill(
+                    "Error while reading {}: {}".format(
+                        args.attach_config.name, e.msg
+                    ),
+                    width=PRINT_WRAP_WIDTH,
+                )
+            )
+        token = attach_config.token
+        if attach_config.enable_services is not None:
```
lucasmoura, line 1113: Why do we need this check here ? can't we just do:
```python
enable_services_override = attach_config.enable_service
```

?

features/attach_validtoken.feature
```diff
@@ -173,6 +173,114 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
+        """
+        esm-apps +yes +enabled
+        """
+        And stdout matches regexp:
+        """
+        esm-infra +yes +disabled
+        """
+        And stdout matches regexp:
+        """
+        cis +yes +enabled
```
lucasmoura, line 227: On focal, this will be `usg`, which will cause the test to fail

uaclient/cli.py
```diff
@@ -1073,22 +1083,72 @@ def action_auto_attach(args, *, cfg):
...
+                        ua_status.MESSAGE_INVALID_SERVICE_OP_FAILURE_TMPL.format(  # noqa: E501
+                            operation="enable", name=name, service_msg=""
+                        )
+                    )
+                    ret = 1
+                    continue
+                entitlement = ent_cls(
+                    cfg, assume_yes=True, allow_beta=True, called_name=name
+                )
+                ent_ret, reason = entitlement.enable()
```
lucasmoura, line 1141: Should we adapt the logic we have on `action_enable` here somehow ?

uaclient/data_types.py
```diff
@@ -0,0 +1,133 @@
...
+    def from_value(val: Any) -> Any:
+        return val
+
+
+class StringDataValue(DataValue):
+    @staticmethod
+    def from_value(val: Any) -> str:
+        if not isinstance(val, str):
+            raise IncorrectTypeError("string", val)
+        return val
```
lucasmoura, line 50: Can't we have that logic on the `DataValue`  class and just determine the type we expect on the subclasses ?

uaclient/data_types.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def data_list(data_cls: Type[DataValue]) -> Type[DataValue]:
+    class _DataList(DataValue):
+        @staticmethod
+        def from_value(val: Any) -> List:
+            if not isinstance(val, list):
+                raise IncorrectTypeError("list", val)
+            ret_list = []
```
lucasmoura, line 67: Why do we need the auxiliary list here ? Can't we just iterate on the original items of the list and perform the validation on then ?

uaclient/data_types.py
```diff
@@ -0,0 +1,133 @@
...
+                    val = field.data_cls.from_value(val)
+                except IncorrectTypeError as e:
+                    raise IncorrectFieldTypeError(e, field.key)
+            kwargs[field.key] = val
+        return cls(**kwargs)
+
+    @classmethod
+    def from_value(cls, val: Any):
+        if not isinstance(val, dict):
+            raise ValueError()
```
lucasmoura, line 121: Why not use the `IncorrectTypeError` exception here ?

uaclient/cli.py
orndorffgrant: You're right!

uaclient/cli.py
orndorffgrant: Oh probably, that might happen during the argparse-ing though. I'll play with this.

uaclient/cli.py
orndorffgrant: It looked a little different, but a small refactor might make it reusable, I'll look into it

uaclient/data_types.py
orndorffgrant: The thing we can't do that way is the return type of the method ` -> str`.
I'd rather keep the extra boilerplate so that we can have the typing.
There might be a fancy way to do this with Generics, but it's nice and simple as it is.

uaclient/data_types.py
orndorffgrant: Ah good catch! I just missed this one when I was adding the error functionality. I'll fix it

orndorffgrant:
I also noticed I should add a `BoolDataValue` as part of this PR even though we don't need it for this feature.

uaclient/cli.py
orndorffgrant: @lucasmoura Right now the following happens
```
root@ua-xenial-e5fbbc69376e12d37d73cd72114de337:~# ua attach --attach-config=/tmp/nonexistent.yaml
usage: ua attach <token> [flags]
argument --attach-config: can't open '/tmp/nonexistent.yaml': [Errno 2] No such file or directory: '/tmp/nonexistent.yaml'
```
What do you think?

lucasmoura: Awesome, I think we can keep as is then

renanrodrigo (commented):
Implementation looks great, data types are a big help.

My only question here would be about a specific use case:
Being `my_file.yaml`:
```
token: mYCOmpleTELyValiDToKEN
enable_services:
  - usg
```
and running `ua attach --attach-config my_file.yaml --no-auto-enable`

I still got USG enabled.
In my opinion, --no-auto-enable should supress any service in the cfg file.
What do you think?

uaclient/cli.py
orndorffgrant: @lucasmoura I did a small refactor to share this code, let me know what you think!

orndorffgrant:
@renanrodrigo Ah good catch, yeah I think that makes sense as well. I'll fix it.

orndorffgrant:
@lucasmoura @renanrodrigo Okay I think I've addressed everything, please take another look

renanrodrigo (approved):
LGTM
If Jenkins is happy I am happy

features/attach_validtoken.feature
```diff
@@ -173,6 +173,115 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
            | bionic  | libkrad0=1.16-2build1       | n/a       | cis        |
            | focal   | hello=2.10-2ubuntu2         | n/a       | usg        |
 
+    @wip
```
lucasmoura, line 176: We can remove that bit

uaclient/cli.py
```diff
@@ -1114,22 +1129,71 @@ def action_auto_attach(args, *, cfg):
...
+        if enable_services_override is not None and args.auto_enable:
+            found, not_found = get_valid_entitlement_names(
+                enable_services_override
+            )
+            for name in not_found:
+                logging.warning(
+                    ua_status.MESSAGE_INVALID_SERVICE_OP_FAILURE_TMPL.format(  # noqa: E501
+                        operation="enable", name=name, service_msg=""
+                    )
+                )
```
lucasmoura, line 1181: Can we just have one log with all the services that are not found ?

uaclient/cli.py
```diff
@@ -1114,22 +1129,71 @@ def action_auto_attach(args, *, cfg):
...
+                    cfg, name, assume_yes=True, allow_beta=True
+                )
+                if not ent_ret:
+                    ret = 1
+                    if (
+                        reason is not None
+                        and isinstance(reason, ua_status.CanEnableFailure)
+                        and reason.message is not None
+                    ):
+                        print(reason.message)
```
lucasmoura, line 1194: We should use the `event.info` bit here

uaclient/tests/test_data_types.py
```diff
@@ -0,0 +1,382 @@
...
+                            IncorrectTypeError("int", "6"), "integer"
+                        ),
+                        0,
+                    ),
+                    "objlist",
+                ),
+            ),
+        ),
+    )
+    def test_error(self, val, error):
```
lucasmoura, line 379: Should we add a case where we don't pass a dict as a val ?

uaclient/testing/fakes.py
```diff
@@ -34,3 +34,20 @@ def request_url(
...
         return response, {"header1": ""}
+
+
+class FakeFile:
+    def __init__(self, content: str, name: str = "fakefile"):
+        self.content = content
+        self.cursor = 0
+        self.name = name
+
+    def read(self, size=None):
```
lucasmoura, line 45: Why can't `read` here just return the full content at once ?

features/attach_validtoken.feature
orndorffgrant: Oy! Thanks good catch

uaclient/cli.py
orndorffgrant: Yep thats a great idea

uaclient/cli.py
orndorffgrant: Right! Sorry I missed that in the rebase

uaclient/tests/test_data_types.py
orndorffgrant: Yes!

uaclient/testing/fakes.py
orndorffgrant: Because `yaml.load` calls `read` with the size parameter in a loop until `read` stops returning content (which I believe is a common pattern).
If we just do `return content`, then `yaml.load` never finishes because it thinks the file keeps returning more content

orndorffgrant:
Thanks @lucasmoura! I think I've addressed your most recent round of comments now

uaclient/cli.py
```diff
@@ -876,6 +886,37 @@ def action_disable(args, *, cfg, **kwargs):
     return 0 if ret else 1
 
 
+def _enable(
```
lucasmoura, line 889: Just a nitpick here, should we move this function to the actions module ?

uaclient/cli.py
```diff
@@ -1114,22 +1136,70 @@ def action_auto_attach(args, *, cfg):
...
+                        reason is not None
+                        and isinstance(reason, ua_status.CanEnableFailure)
+                        and reason.message is not None
+                    ):
+                        event.info(reason.message)
+            if not_found:
+                msg = _create_enable_entitlements_not_found_message(
+                    not_found, allow_beta=True
+                )
+                logging.warning(msg)
```
lucasmoura, line 1199: This will print to stdout as well, right ? If yes, I think we should consider using the event logic here

uaclient/cli.py
orndorffgrant: Hmmm, yes probably, I'll go ahead and do that here

uaclient/cli.py
orndorffgrant: Yes, and I thought of that but I think it will make more sense to leave it as is in this PR. Then, in the PR that adds `--format=json` support to `attach`, we can change this to use the event_logger appropriately

lucasmoura: Fair enough

orndorffgrant:
I also need to go back through and add docstrings everywhere :)

lucasmoura:
@orndorffgrant This test here on a xenial machine is breaking and it does perform an attach operation during its execution. I think it is worth double checking it:
`Fix command on an unattached machine`

orndorffgrant:
@lucasmoura @renanrodrigo sorry for the delay, I think this is ready again. The integration test that failed is fixed, but this latest update includes another small refactor to move the enable function to the actions.py file, so we should definitely wait for jenkins again.

uaclient/entitlements/__init__.py
```diff
@@ -23,16 +23,27 @@
 ]  # type: List[Type[UAEntitlement]]
 
 
-def entitlement_factory(name: str):
+class EntitlementNotFoundError(Exception):
+    pass
+
+
+def entitlement_factory(name: str, not_found_okay: bool = True):
```
lucasmoura, line 30: I think I am missing something here, but why do we need the `not_found_okay` parameter ?

orndorffgrant: @lucasmoura You're not missing anything. It isn't strictly necessary for the functionality in this PR.

However, because we're adding `enable_by_entitlement_name` to `actions.py`, I wanted it to have defined functionality for what happens when the entitlement doesn't exist. I decided an Error should be raised in that situation. And then I thought it makes more sense if the error is raised in this factory function.

I can move this error to the `enable_by_entitlement_name` function though if you prefer

orndorffgrant: Also open to better parameter names for `not_found_okay` :)

lucasmoura: But is there a situation where we don't want to raise that error if the entitlement doesn't exist ? That is what I am missing here

orndorffgrant: Ahhh I understand. I'd need to potentially refactor some of the places we call `entitlement_factory`. I think some expect `None` to be returned if the entitlement doesn't exist.

But now that I think more about it, I think that refactor would be worth it. I'll add it as an extra commit if you agree @lucasmoura

orndorffgrant: @lucasmoura refactor is up. We should let jenkins run again because this touched a lot of places

lucasmoura:
@orndorffgrant just a heads up that there is a conflict now in the PR

orndorffgrant:
Ah another one, will fix

orndorffgrant:
@lucasmoura fixed conflicts!

lucasmoura (approved):

renanrodrigo (approved):
new changes are good.

PR #1930, opened 2022-01-11
daemon: auto-attach when pro license added on gcp

orndorffgrant:
## Notes:
The license-check commit is exactly the same as the last one we merged

The daemon commit is significantly different, but the details of fetching the license are the same.

The biggest thing here (besides making sure the daemon works for auto-upgrade-to-pro) is making sure the daemon doesn't even run on non-gcpgeneric platforms.

TODO:
- [x] make sure manual license-check test scripts still work and put them in the right folder
- [x] fix the couple failing integration tests
- [x] test to confirm ConditionPathExists is sufficient to work with the cloud-id-${cloud-id} cloud-init feature

## Proposed Commit Message

```
daemon: auto-attach when pro license added on gcp
```
```
license-check: remove timer-based license-check
```

## Test Steps

Run new integration tests. Maybe not every combination of series/platform, but run a sample of series and platforms to make sure they pass.
Run the manual test scripts
- the `gcp_` scripts require two arguments: (1) a deb of ua from this PR and (2) a deb of cloud-init that includes the changes in https://github.com/canonical/cloud-init/pull/1244

## Desired commit type

## Checklist


orndorffgrant:
@basak @paride @renanrodrigo @lucasmoura @blackboxsw @CalvoM Large PR warning :grimacing: 

I think this daemon PR is ready for review. Please let me know what you think. It only starts on GCP and the long polling is even turned off for now until GCP supports adding a license while an instance is running. The old 5min license-check timer is removed in this PR since the daemon replaces it's functionality.

features/daemon.feature
```diff
@@ -0,0 +1,228 @@
...
+        """
+        Then stdout matches regexp:
+        """
+        daemon ending
+        """
+        # verify it stays on when configured to do so
+        When I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
+        """
+        ua_config:
+            poll_for_pro_license: true
```
lucasmoura, line 32: Why do we need this extra config here for the daemon to run ? Wouldn't having the cloud-init gce file be enough for it ?

systemd/ubuntu-advantage.service
```diff
@@ -0,0 +1,24 @@
+# If you are uninterested in the (free for personal use) Ubuntu Advantage
+# services, including security updates after standard EOL and kernel patching
+# without rebooting, then you can safely stop and disable this service:
+# sudo systemctl stop ubuntu-advantage.service
+# sudo systemctl disable ubuntu-advantage.service
```
lucasmoura, line 5: Should we add any considerations that this service only runs on GCP instances ?

uaclient/daemon.py
```diff
@@ -0,0 +1,117 @@
...
+        # again
+        pass
+    else:
+        if pro_license_present:
+            attempt_auto_attach(cfg, cloud)
+            return
+
+    if not cfg.poll_for_pro_license:
+        LOG.debug("Configured to not poll for pro license, shutting down")
+        return
```
lucasmoura, line 97: Shouldn't this check happen in the beginning of the function as well ?

systemd/ubuntu-advantage.service
orndorffgrant: Ah yes I'll add that!

uaclient/daemon.py
orndorffgrant: See my answer above. This config option is only to prevent the daemon from entering the polling loop

systemd/ubuntu-advantage.service
orndorffgrant: Updated! Let me know if you'd prefer different wording

renanrodrigo (commented):
The daemon integration test is failing for me on Xenial:
```
When I run `systemctl restart ubuntu-advantage.service` with sudo                            # features/steps/steps.py:248
      Assertion Failed: 
      Expected: <0>
           but: was <5>
(...)
      2022-02-10 18:00:43,950:INFO:pycloudlib.instance:executing: sudo touch /run/cloud-init/cloud-id-gce
      2022-02-10 18:00:44,731:INFO:pycloudlib.instance:executing: sudo truncate -s 0 /var/log/ubuntu-advantage-daemon.log
      2022-02-10 18:00:45,365:INFO:pycloudlib.instance:executing: sudo systemctl restart ubuntu-advantage.service
      2022-02-10 18:00:46,059:ERROR:root:Error executing command: systemctl restart ubuntu-advantage.service
      2022-02-10 18:00:46,060:ERROR:root:stdout: 
      2022-02-10 18:00:46,060:ERROR:root:stderr: Failed to restart ubuntu-advantage.service: Unit ubuntu-advantage.service not found.
```

blackboxsw:
upstream cloud-init landed the PR to surface /run/cloud-init/cloud-id(-*) files. They are already build in add-apt-repository ppa:cloud-init-dev/daily for Jammy (and uploaded to Jammy proper so should be in cloud image shortly).   The ppa:cloud-init-dev/daily recipes should build and publish tomorrow for other series so it should be much easier to test.

orndorffgrant:
@blackboxsw Splendid! I'll adjust the tests to use the daily PPA then

orndorffgrant:
@renanrodrigo I'm having trouble reproducing this failure, let's sync on Monday to figure it out

renanrodrigo:
@orndorffgrant and I saw it passing earlier while going through this, I re-tested it using a clean environment and it passed consistently

lucasmoura (commented):
Tests are passing for me. I think we should only add tests for the new daemon module and the PR should be good to go

uaclient/daemon.py
```diff
@@ -0,0 +1,117 @@
+import logging
```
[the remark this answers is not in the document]
orndorffgrant: Pushed up unit tests for this module. But there is some interference such that when run as part of the entire test suite, some fail. I will look into it tomorrow

uaclient/tests/test_daemon.py
```diff
@@ -0,0 +1,503 @@
...
+
+    @mock.patch(M_PATH + "LOG.warning")
+    def test_stop_warning(self, m_log_warning, m_subp):
+        err = util.ProcessExecutionError("cmd")
+        m_subp.side_effect = err
+        stop()
+        assert [
+            mock.call(["systemctl", "stop", "ubuntu-advantage.service"])
+        ] == m_subp.call_args_list
+        assert [mock.call(err)] == m_log_warning.call_args_list
```
lucasmoura, line 52: I think we can just use `pytest.mark.parametrize` here instead, since the test are really similar

lucasmoura (approved):

uaclient/daemon.py
```diff
@@ -0,0 +1,126 @@
...
+            start = time.time()
+            pro_license_present = cloud.is_pro_license_present(
+                wait_for_change=True
+            )
+            end = time.time()
+        except exceptions.CancelProLicensePolling:
+            LOG.debug("Cancelling polling")
+            return
+        except exceptions.DelayProLicensePolling:
+            time.sleep(600)
```
blackboxsw, line 110: Do we want to make this sleep configurable to allow GCP folks to backoff if the 10 min interval is too often?

blackboxsw (changes requested):
Good SRU manual test scripts in general for typical upgrade cases.


If we don't already have one. In a separate PR, It would be nice to grow an integration testthat runs an upgrade from the released version of ua-tools in <series>-updates to daily ppa:ua-client/daily to assert there are no failures in upgrade path and expected systemd services.Some additional automated test coverage of upgrade path can ensure we don't degrade packaging upgrade path behavior and related systemd services.

I think we generally cover automated tests for the following upgrade test cases:
1. running the suite of integration tests against an image produced by the version of ua-tools that lives in ppa:ua-client/daily
2.  LTS to LTS upgrade test in features/ubuntu_upgrade.feature

But I don't think we have a behave test which validates the package upgrade from the version in xenial-updates to the version in `ppa:ua-client/daily`

I believe that we need to change the After= declarations in the ubuntu-advantage.service, but beyond that the rest of my comments you can take as you see fit. I'll still need tomorrow for reviewing the daemon.py behavior and notify usage before I can fully approve here.

debian/ubuntu-advantage-tools.postinst
```diff
@@ -308,14 +310,9 @@ notify_wrong_fips_metapackage_on_cloud() {
...
-    if echo "$cloud_id" | grep -q "^gce"; then
-        if check_is_lts "${UBUNTU_CODENAME}"; then
-            if [ ! -f $MACHINE_TOKEN_FILE ]; then
-                touch $LICENSE_CHECK_MARKER_FILE
-            fi
-        fi
+rm_old_license_check_marker() {
+    if [ -f $OLD_LICENSE_CHECK_MARKER_FILE ]; then
+        rm -f $OLD_LICENSE_CHECK_MARKER_FILE
     fi
```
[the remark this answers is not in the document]
orndorffgrant: Simpler and better +1

features/daemon.feature
```diff
@@ -0,0 +1,228 @@
...
+    Scenario Outline: daemon should run when appropriate on gcp generic lts
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        # verify its enabled, but stops itself when not configured to poll
+        # TODO-BEGIN: remove this chunk once cloud-init supports this feature
+        When I run `touch /run/cloud-init/cloud-id-gce` with sudo
+        When I run `truncate -s 0 /var/log/ubuntu-advantage-daemon.log` with sudo
+        When I run `systemctl restart ubuntu-advantage.service` with sudo
+        # TODO-END: remove this chunk once cloud-init supports this feature
+        When I wait `1` seconds
+        When I run `cat /var/log/ubuntu-advantage-daemon.log` with sudo
```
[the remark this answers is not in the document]
orndorffgrant: That's a great idea

features/daemon.feature
```diff
@@ -0,0 +1,228 @@
...
+        Then stdout matches regexp:
+        """
+        daemon starting
+        """
+        Then stdout does not match regexp:
+        """
+        daemon ending
+        """
+        # verify attach tops it immediately and doesn't restart after reboot
```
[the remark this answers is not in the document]
orndorffgrant: oops :) good catch

features/daemon.feature
```diff
@@ -0,0 +1,228 @@
...
+        """
+        Then stdout matches regexp:
+        """
+        Configured to not poll for pro license, shutting down
+        """
+        Then stdout matches regexp:
+        """
+        daemon ending
+        """
+        # verify it stays on when configured to do so
```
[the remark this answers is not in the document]
orndorffgrant: Oh interesting, yeah I'll look into adding this

systemd/ubuntu-advantage.service
```diff
@@ -0,0 +1,26 @@
...
+
+[Unit]
+Description=Ubuntu Advantage GCP Auto Attach Daemon
+Documentation=man:ubuntu-advantage https://ubuntu.com/advantage
+After=network.target network-online.target systemd-networkd.service ua-auto-attach.service cloud-config.target
```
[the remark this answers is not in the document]
orndorffgrant: Excellent catch!!!
Yes I agree and I thought that's what this was doing, but after re-reviewing cloud-config.target and cloud-config.service, you are right. We should be `After: cloud-config.service` to guarantee we run after all relevant cloud-init modules.

systemd/ubuntu-advantage.service
```diff
@@ -0,0 +1,26 @@
...
+
+# Only run if not already attached
+ConditionPathExists=!/var/lib/ubuntu-advantage/private/machine-token.json
```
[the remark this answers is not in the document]
orndorffgrant: We do have the section of the integration test labeled with comment 
```
        # verify attach stops it immediately and doesn't restart after reboot
```
The "and doesn't restart after reboot" I think covers this scenario?

features/daemon.feature
```diff
@@ -0,0 +1,228 @@
...
+        Then stdout does not match regexp:
+        """
+        daemon ending
+        """
+        # verify attach tops it immediately and doesn't restart after reboot
+        When I attach `contract_token` with sudo
+        Then I verify that running `systemctl status ubuntu-advantage.service` `with sudo` exits `3`
+        Then stdout matches regexp:
+        """
+        Active: inactive \(dead\)
```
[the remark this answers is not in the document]
orndorffgrant: Ah, yes!

uaclient/daemon.py
orndorffgrant: This should only happen in case of errors.
I imagine if there is a persistent error then the user would just want to turn off the polling completely with the feature flag added elsewhere in this PR.

It wouldn't be too hard to make this configurable too though - I think its probably worth it.

Related: do you think 10 minutes is a good default here? Too short?

systemd/ubuntu-advantage.service
```diff
@@ -0,0 +1,26 @@
...
+# Only run on GCP
+ConditionPathExists=/run/cloud-init/cloud-id-gce
```
RAOF, line 17: SRU review: Am I correct in understanding that as per the [spec document](https://docs.google.com/document/d/1xeOnO1Uh1f4pHx3DlQOGTor0b60zH5gzvx4MZwhwltI) this is pending a new cloud-init feature?

Not blocking, just checking.

RAOF (commented):
SRU pre-review: This seems to implement the spec as-approved, and doesn't appear to introduce unrelated changes. This would be an acceptable SRU.

debian/ubuntu-advantage-tools.postinst
```diff
@@ -355,10 +350,24 @@ remove_old_systemd_timers() {
     # Since we actually want to remove this service from now on
     # we have replicated that behavior here
     if [ -L $OLD_MESSAGING_TIMER_MASKED_LOCATION ]; then
-        if [ -x "/usr/bin/deb-systemd-helper" ]; then
```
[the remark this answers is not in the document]
orndorffgrant: Very good point! Yes this is an unrelated change and really shouldn't be lumped in with this feature. I'll give this change it's own commit and it's own PR to keep things separate.

orndorffgrant:
@blackboxsw I believe the latest couple commits address all your concerns. I'll leave them as separate commits for now so you can more easily remove them

After you review, I'll rebase and squash things together - that is also when I'll take out the unrelated postinst change that RAOF mentioned.

uaclient/config.py
```diff
@@ -218,6 +217,35 @@ def metering_timer(self, value: int):
...
+            self.cfg["ua_config"] = {}
+        self.cfg["ua_config"]["poll_for_pro_license"] = value
+        self.write_cfg()
+
+    @property
+    def polling_error_retry_delay(self) -> int:
+        # TODO: when polling is supported
+        #     1. add this field to UA_CONFIGURABLE_KEYS
+        return self.cfg.get("ua_config", {}).get(
+            "polling_error_retry_delay", 600
```
blackboxsw, line 239: Thanks for providing the additional config option here. Though it may very well be above-and-beyond, making it configurable will give folks an option to backoff if UA is hitting frequent errors related to accessing GCP IMDS.

lib/daemon.py
```diff
@@ -0,0 +1,39 @@
...
+        logging.CRITICAL, logging.ERROR, log_file=cfg.daemon_log_file
+    )
+
+    LOG.debug("daemon starting")
+
+    notify("READY=1")
+
+    daemon.poll_for_pro_license(cfg)
+
+    LOG.debug("daemon ending")
```
blackboxsw, line 34: Not critical, but I wonder if it's worth handling signal.SIGTERM to also report daemon is being shutdown, otherwise we leave open logs in the "daemon starting" without tracking the external termination of that process. Can be separate PR if you think it's worthwhile.

blackboxsw (approved):
Excellent @orndorffgrant. Thank you for the updates. Confirmed upgrade and downgrade behavior is clean with setup and teardown of ua-license-check vs ubuntu-advantage services. 

Though your integration tests cover it well, confirmed that the daemon doesn't start polling automatically due to your config switch set false. 

Something to note is the resident memory that we sit on even when idle:
```
   RES   SHR
 26364  12500 S   0.0  0.7   0:00.29 python3 
```
We may want to look at ways to decrease this mem footprint in subsequent efforts. I don't know if UA's lru_cache is adding much to this footprint, probably not remotely as much as all the module imports we use throughout uaclient. But, it's something we should be wary of as we grow functionality/features in this daemon.

systemd/ubuntu-advantage.service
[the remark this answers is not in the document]
orndorffgrant: Right, thanks for clarifying - and I should've responded again here, this was added on line 88 :)

lib/daemon.py
orndorffgrant: This is a good idea! I've created a separate Jira issue for it here: https://warthogs.atlassian.net/browse/SC-861

orndorffgrant:
Thanks @blackboxsw 

Ack to memory consumption. There is definitely room for optimization there.

Based on some quick testing, we could pretty easily lower that by almost 2Mb by writing a daemon specific logging setup function. Jira issue here: https://warthogs.atlassian.net/browse/SC-862

Then we could potentially lower it by a couple more Mb with a more extensive refactor to UAConfig. We really only need read access to the config for the daemon, and UAConfig currently does much more than that. Jira Issue here: https://warthogs.atlassian.net/browse/SC-863

uaclient/daemon.py
```diff
@@ -0,0 +1,128 @@
...
+            status.NOTICE_DAEMON_AUTO_ATTACH_LOCK_HELD.format(
+                operation=e.lock_holder
+            ),
+        )
+        LOG.debug("Failed to auto attach")
+        return
+    except Exception as e:
+        LOG.exception(e)
+        cfg.add_notice("", status.NOTICE_DAEMON_AUTO_ATTACH_FAILED)
+        lock.clear_lock_file_if_present()
```
lucasmoura, line 46: Isn't the lock cleared by default after we exit the context manager ?

lucasmoura (approved):
Just a minor nit, let me know if it makes sense or not

uaclient/daemon.py
orndorffgrant: It should be. the only case where it wouldn't is if `delete_cache_key` failed for some reason. Then this would try again. I was just following the pattern we have elsewhere where we call this function when we catch an exception. It is a `noop` if the lock was already cleared so it seemed harmless.

I can remove it if you think it is unnecessary though

orndorffgrant:
Just reporting that I re-ran all the new integration tests with the recent changes and they all passed

I will also move the manual test scripts out of the 27.7 folder into a new 27.8 folder and re-run them

orndorffgrant:
> I will also move the manual test scripts out of the 27.7 folder into a new 27.8 folder and re-run them

Done and all manual tests passed

renanrodrigo (approved):

PR #1932, opened 2022-01-13
fips: integration tests for aws pro fips

CalvoM:
## Proposed Commit Message

fips: integration tests for aws pro fips

WIP for the integration tests for fips machine, starting with aws then azure.

## Desired commit type

## Checklist

### DIFF
--- a/features/aws-ids.yaml
+++ b/features/aws-ids.yaml
@@ -1,3 +1,5 @@
-bionic: ami-094e9c95623db463c
-focal: ami-0193aa0a9df84a08b
-xenial: ami-06e94647aeaed1e5c
+bionic: ami-0419d66039473da9d
+bionic-fips: ami-03b75f613f80bcff1
+focal: ami-0489b8bdbbf3a3b32
+xenial: ami-011bcfe2bea365b6a
+xenial-fips: ami-077e4c339a098fc9f
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -218,7 +218,10 @@ class Cloud:
         if "pro" in self.machine_type:
             with open(self.pro_ids_path, "r") as stream:
                 pro_ids = yaml.safe_load(stream.read())
-            image_name = pro_ids[series]
+            if "fips" in self.machine_type:
+                image_name = pro_ids[series + "-fips"]
+            else:
+                image_name = pro_ids[series]
         else:
             image_name = self.api.daily_image(release=series)
 
@@ -539,7 +542,7 @@ class Azure(Cloud):
             List of ports to open for network ingress to the instance
 
         :returns:
-            An AWS cloud provider instance
+            An Azure cloud provider instance
         """
         if not image_name:
             image_name = self.locate_image_name(series)
@@ -654,7 +657,7 @@ class GCP(Cloud):
             List of ports to open for network ingress to the instance
 
         :returns:
-            An AWS cloud provider instance
+            An GCP cloud provider instance
         """
         if not image_name:
             image_name = self.locate_image_name(series)
--- a/features/ubuntu_pro_fips.feature
+++ b/features/ubuntu_pro_fips.feature
@@ -1,0 +1,105 @@
+Feature: Command behaviour when auto-attached in an ubuntu PRO fips image
+
+    @series.lts
+    @uses.config.machine_type.aws.pro.fips
+    Scenario Outline: Check fips is enabled correctly on Ubuntu pro fips AWS machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
+        """
+        contract_url: 'https://contracts.canonical.com'
+        data_dir: /var/lib/ubuntu-advantage
+        log_level: debug
+        log_file: /var/log/ubuntu-advantage.log
+        """
+        And I run `ua auto-attach` with sudo
+        And I run `ua status --wait` as non-root
+        And I run `ua status` as non-root
+        Then stdout matches regexp:
+            """
+            esm-apps      +yes +enabled +UA Apps: Extended Security Maintenance \(ESM\)
+            esm-infra     +yes +enabled +UA Infra: Extended Security Maintenance \(ESM\)
+            fips          +yes +enabled +NIST-certified core packages
+            fips-updates  +yes +disabled +NIST-certified core packages with priority security updates
+            livepatch     +yes +n/a  +Canonical Livepatch service
+            """
+        And I verify that running `apt update` `with sudo` exits `0`
+        And I verify that running `grep Traceback /var/log/ubuntu-advantage.log` `with sudo` exits `1`
+        And I verify that `openssh-server` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-client` is installed from apt source `<fips-apt-source>`
+        And I verify that `strongswan` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-server-hmac` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-client-hmac` is installed from apt source `<fips-apt-source>`
+        And I verify that `strongswan-hmac` is installed from apt source `<fips-apt-source>`
+        When I run `uname -r` as non-root
+        Then stdout matches regexp:
+            """
+            <fips-kernel-version>
+            """
+        When I run `apt-cache policy ubuntu-aws-fips` as non-root
+        Then stdout does not match regexp:
+        """
+        .*Installed: \(none\)
+        """
+        When I run `cat /proc/sys/crypto/fips_enabled` with sudo
+        Then I will see the following on stdout:
+        """
+        1
+        """
+        When I run `systemctl start ua-auto-attach.service` with sudo
+        And I verify that running `systemctl status ua-auto-attach.service` `as non-root` exits `0,3`
+        Then stdout matches regexp:
+        """
+        .*status=0\/SUCCESS.*
+        """
+        And stdout matches regexp:
+        """
+        Skipping attach: Instance '[0-9a-z\-]+' is already attached.
+        """
+        When I run `ua auto-attach` with sudo
+        Then stderr matches regexp:
+        """
+        Skipping attach: Instance '[0-9a-z\-]+' is already attached.
+        """
+        When I run `apt-cache policy` with sudo
+        Then apt-cache policy for the following url has permission `500`
+        """
+        https://esm.ubuntu.com/infra/ubuntu <release>-infra-updates/main amd64 Packages
+        """
+        And apt-cache policy for the following url has permission `500`
+        """
+        https://esm.ubuntu.com/infra/ubuntu <release>-infra-security/main amd64 Packages
+        """
+        And apt-cache policy for the following url has permission `500`
+        """
+        https://esm.ubuntu.com/apps/ubuntu <release>-apps-updates/main amd64 Packages
+        """
+        And apt-cache policy for the following url has permission `500`
+        """
+        https://esm.ubuntu.com/apps/ubuntu <release>-apps-security/main amd64 Packages
+        """
+        And I verify that running `apt update` `with sudo` exits `0`
+        When I run `apt install -y <infra-pkg>/<release>-infra-security` with sudo, retrying exit [100]
+        And I run `apt-cache policy <infra-pkg>` as non-root
+        Then stdout matches regexp:
+        """
+        \s*500 https://esm.ubuntu.com/infra/ubuntu <release>-infra-security/main amd64 Packages
+        \s*500 https://esm.ubuntu.com/infra/ubuntu <release>-infra-updates/main amd64 Packages
+        """
+        And stdout matches regexp:
+        """
+        Installed: .*[~+]esm
+        """
+        When I run `apt install -y <apps-pkg>/<release>-apps-security` with sudo, retrying exit [100]
+        And I run `apt-cache policy <apps-pkg>` as non-root
+        Then stdout matches regexp:
+        """
+        Version table:
+        \s*\*\*\* .* 500
+        \s*500 https://esm.ubuntu.com/apps/ubuntu <release>-apps-security/main amd64 Packages
+        """
+
+        Examples: ubuntu release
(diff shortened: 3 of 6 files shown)


features/aws-ids.yaml
```diff
@@ -1,3 +1,3 @@
-bionic: ami-094e9c95623db463c
+bionic: ami-03b75f613f80bcff1
```
lucasmoura, line 1: I think we should have a separate file for the fips pro images. The rationale for that is because we still want to test the base pro machines

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,505 @@
+Feature: Command behaviour when auto-attached in an ubuntu PRO fips image
+
+    @series.lts
+    @uses.config.machine_type.aws.pro.fips
+    Scenario Outline: Proxy auto-attach in an Ubuntu pro AWS machine
```
lucasmoura, line 5: I don't think we need the proxy tests for the fips pro machines. What do you think @orndorffgrant @renanrodrigo ?

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,505 @@
...
+        .*CONNECT 169.254.169.254.*
+        """
+        Examples: ubuntu release
+           | release | fips-s   | cis-s    | cis_or_usg |
+           | xenial  | enabled  | disabled | cis        |
+           | bionic  | enabled  | disabled | cis        |
+           | focal   | enabled  | disabled | usg        |
+
+    @series.lts
+    @uses.config.machine_type.azure.pro
```
lucasmoura, line 54: Since we are only adding the aws pro fips tests, no need to include azure or gcp tests in this file for now

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,505 @@
...
+        And I run `ua auto-attach` with sudo
+        And I run `ua status --wait` as non-root
+        And I run `ua status` as non-root
+        Then stdout matches regexp:
+            """
+            .*esm-apps      +yes +enabled +UA Apps: Extended Security Maintenance \(ESM\)
+            esm-infra     +yes +enabled +UA Infra: Extended Security Maintenance \(ESM\)
+            fips          +yes +<fips-s> +NIST-certified core packages
+            fips-updates  +yes +disabled +NIST-certified core packages with priority security updates
+            livepatch     +yes +n/a  +Canonical Livepatch service.*
```
lucasmoura, line 183: Let's also verify that `FIPS` is shown as enabled here by default.
Furthermore, let's also check if FIPS was correctly configured on the machine. Take a look on the FIPS tests we have on either `enable_fips_vm` or `enable_fips_cloud` to see the process we use to check that

lucasmoura (commented):
@CalvoM Take a look on `features/cloud.py` module to see how we create the pro images. I think we will need a similar structure to create the `pro fips` images as well

features/aws-ids.yaml
CalvoM: Yeah, I was thinking of adding fips machines in the ids. Lemme work on it.

features/ubuntu_pro_fips.feature
CalvoM: Yeah, lemme make this aws lean.

features/ubuntu_pro_fips.feature
CalvoM: The machines will be fips enabled by default i.e. the fips machines.

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,173 @@
...
+            """
+        Then stdout matches regexp:
+            """
+            <cis_or_usg>           +yes  +<cis-s>  +Security compliance and audit tools
+            """
```
lucasmoura, line 78: We can remove that bit here, since it is not fips related

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,173 @@
...
```
lucasmoura, line 91: I think we can also remove that part here

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,173 @@
...
+        And I verify that `openssh-server` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-client` is installed from apt source `<fips-apt-source>`
+        And I verify that `strongswan` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-server-hmac` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-client-hmac` is installed from apt source `<fips-apt-source>`
+        And I verify that `strongswan-hmac` is installed from apt source `<fips-apt-source>`
+        And  I run `uname -r` as non-root
+        Then stdout matches regexp:
+            """
+            fips
```
lucasmoura, line 103: Since we are on an aws cloud, let's check to see if the `aws-fips`  kernel is the one installed

features/aws-ids.yaml
```diff
@@ -1,3 +1,6 @@
+xenial: ami-06e94647aeaed1e5c
+xenial-fips: ami-077e4c339a098fc9f
```
lucasmoura, line 2: Since you are adding the fips images on the same file, we need to update the `tools/refresh-aws-pro-ids` script to not override the fips ids here

CalvoM: yeah, just updating the releases refreshed alone.

tools/refresh-aws-pro-ids
```diff
@@ -32,40 +33,51 @@ def main():
     os.chdir("products")
     aws_ids = {}
     for aws_listing in glob.glob("listing-aws-pro-*"):
-        m = re.match(
-            r"^listing-aws-pro-(?P<release>\w+).yaml$", aws_listing
-        )
+        m = re.match(r"^listing-aws-pro-(?P<release>\w+).yaml$", aws_listing)
```
lucasmoura, line 36: I don't know if this is feasible, but maybe we should try to update the pro fips images too

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,152 @@
...
+        And I verify that `openssh-server` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-client` is installed from apt source `<fips-apt-source>`
+        And I verify that `strongswan` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-server-hmac` is installed from apt source `<fips-apt-source>`
+        And I verify that `openssh-client-hmac` is installed from apt source `<fips-apt-source>`
+        And I verify that `strongswan-hmac` is installed from apt source `<fips-apt-source>`
+        When I run `uname -r` as non-root
+        Then stdout matches regexp:
+            """
+            aws-fips
```
lucasmoura, line 82: On xenial, It seems like the `ubuntu-fips` metapackage is being installed, so we will not see the `aws-fips` string here, only `fips`

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,152 @@
...
+        Version table:
+        \s*\*\*\* .* 500
+        \s*500 https://esm.ubuntu.com/apps/ubuntu <release>-apps-security/main amd64 Packages
+        """
+
+        Examples: ubuntu release
+           | release | fips-s   | infra-pkg | apps-pkg | fips-apt-source                                |
+           | xenial  | enabled  | libkrad0  | jq       | https://esm.ubuntu.com/fips/ubuntu xenial/main |
+           | bionic  | enabled  | libkrad0  | bundler  | https://esm.ubuntu.com/fips/ubuntu bionic/main |
+           | focal   | enabled  | hello     | ant      | https://esm.ubuntu.com/fips/ubuntu focal/main  |
```
lucasmoura, line 151: Currently, we don't allow enabling fips on focal. This will change after #1939

features/ubuntu_pro_fips.feature
CalvoM: Okay, I noticed I have been testing on bionic mostly. Let me do the tests on xenial as well. Just having issues with updating the python versions

features/ubuntu_pro_fips.feature
CalvoM: Oh, okay. I just saw the blog post on fips on focal and assumed it was available  now.

features/ubuntu_pro_fips.feature
CalvoM: So, technical the string ".*-fips" should work

tools/refresh-aws-pro-ids
CalvoM: Lemme check it out

CalvoM: I will have to remove the focal stuff

features/ubuntu_pro_fips.feature
lucasmoura: Yes, this does work, but I think we should be explicti that `bionic` expected the optimized kernel while xenial does not. We can add a new column on the examples table to say the expected kernel version

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,150 @@
+Feature: Command behaviour when auto-attached in an ubuntu PRO fips image
+
+    @series.lts
+    @uses.config.machine_type.aws.pro.fips
```
lucasmoura, line 4: @orndorffgrant @renanrodrigo do you think we need the proxy test for the fips pro machines ? I think it is not necessary, but I want to hear the team's opinion on that

orndorffgrant: I agree. The only difference is the crypto available, but that is unlikely to interfere with proxy usage. Or at least, if there is a crypto problem, then my guess is that it would probably cause other bigger issues first.

lucasmoura: @CalvoM let's remove that test then

features/ubuntu_pro_fips.feature
orndorffgrant: I agree with @lucasmoura we want to test that the correct specific kernel is installed in each version.

features/cloud.py
```diff
@@ -539,7 +542,7 @@ def _create_instance(
             List of ports to open for network ingress to the instance
 
         :returns:
-            An AWS cloud provider instance
+            An Azure cloud provider instance
```
orndorffgrant, line 545: Oy thanks for noticing these and fixing

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,150 @@
...
+        log_file: /var/log/ubuntu-advantage.log
+        """
+        And I run `ua auto-attach` with sudo
+        And I run `ua status --wait` as non-root
+        And I run `ua status` as non-root
+        Then stdout matches regexp:
+            """
+            .*esm-apps      +yes +enabled +UA Apps: Extended Security Maintenance \(ESM\)
+            esm-infra     +yes +enabled +UA Infra: Extended Security Maintenance \(ESM\)
+            fips          +yes +<fips-s> +NIST-certified core packages
```
orndorffgrant, line 66: I think we can safely hardcode this to say "enabled"
```suggestion
            fips          +yes +enabled +NIST-certified core packages
```

Then we can remove the `fips-s` column from the examples table down below

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,150 @@
...
+        data_dir: /var/lib/ubuntu-advantage
+        log_level: debug
+        log_file: /var/log/ubuntu-advantage.log
+        """
+        And I run `ua auto-attach` with sudo
+        And I run `ua status --wait` as non-root
+        And I run `ua status` as non-root
+        Then stdout matches regexp:
+            """
+            .*esm-apps      +yes +enabled +UA Apps: Extended Security Maintenance \(ESM\)
```
orndorffgrant, line 64: Do we need the `.*` here?
```suggestion
            esm-apps      +yes +enabled +UA Apps: Extended Security Maintenance \(ESM\)
```

CalvoM: Not really, I had made changes when I testing some parts, and forgot to delete, my bad.

features/ubuntu_pro_fips.feature
```diff
@@ -0,0 +1,105 @@
...
+        Then stdout matches regexp:
+        """
+        Version table:
+        \s*\*\*\* .* 500
+        \s*500 https://esm.ubuntu.com/apps/ubuntu <release>-apps-security/main amd64 Packages
+        """
+
+        Examples: ubuntu release
+           | release | infra-pkg | apps-pkg | fips-apt-source                                | fips-kernel-version             |
+           | xenial  | libkrad0  | jq       | https://esm.ubuntu.com/fips/ubuntu xenial/main | \d+\.\d+\.\d+\-\d+\-fips$       |
```
lucasmoura, line 103: Even though this regex looks great, I think we can be even more flexible here, to avoid the test to fail if the kernel changes other parts of kernel name.
Under `fips-kernel-version`, I think bionic should check only `*aws-fips` and xenial, `*fips`

lucasmoura (commented):
This looks really good to me. I just have that minor nit with the regex check, but after that I will approve this PR.

features/ubuntu_pro_fips.feature
CalvoM: So for xenial, the bionic regex check can still pass?

lucasmoura: No, because on xenial we won't have the `aws` part. Also, I have updated my comment, we don't need the `*` bit before the names

CalvoM: Oh understandable. Let me change and run the tests

lucasmoura (commented):
I think you accidentally committed both of these files:
```
 behave-awspro-fips-16.04.txt
  behave-awspro-fips-18.04.txt
  ```
  
  We should remove it from the PR

lucasmoura (approved):
After removing the test result files, I think we can merge this.
Thanks for all the work here @CalvoM

CalvoM:
> After removing the test result files, I think we can merge this. Thanks for all the work here @CalvoM

Oh my bad!, removed.

orndorffgrant:
Looking great! I'll just run the new test locally and then approve

orndorffgrant (approved):
Thanks for getting this working @CalvoM!

Tests worked for me locally. Merging now

PR #1933, opened 2022-01-13
enable: show incompatible services message on json (SC-734)

lucasmoura:
## Proposed Commit Message
enable: show incompatible services message on json

We are now adding the incompatible services message to the json output of the enable command

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -657,11 +657,14 @@ Feature: Enable command behaviour when attached to an UA subscription
         And I will see the following on stdout
             """
             One moment, checking your subscription first
-            """
-        And I will see the following on stderr
-            """
             Cannot enable FIPS when Livepatch is enabled.
             """
+        Then I verify that running `ua enable fips --assume-yes --format json` `with sudo` exits `1`
+        And stdout is a json matching the `enable` schema
+        And I will see the following on stdout:
+        """
+        {"_schema_version": "0.1", "errors": [{"message": "Cannot enable FIPS when Livepatch is enabled.", "service": "fips", "type": "service"}], "failed_services": ["fips"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+        """
 
     @slow
     @series.xenial
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -7,7 +7,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import yaml
 
-from uaclient import config, contract, status, util
+from uaclient import config, contract, event_logger, status, util
 from uaclient.defaults import DEFAULT_HELP_FILE
 from uaclient.status import (
     MESSAGE_DEPENDENT_SERVICE_STOPS_DISABLE,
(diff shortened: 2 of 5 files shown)


features/attached_enable.feature
```diff
@@ -657,11 +657,14 @@ Feature: Enable command behaviour when attached to an UA subscription
         And I will see the following on stdout
             """
             One moment, checking your subscription first
-            """
-        And I will see the following on stderr
```
renanrodrigo, line 661: Is it ok that the error message ends up in `stdout` instead of `stderr`?

renanrodrigo (approved):
LGTM
Just a question

features/attached_enable.feature
lucasmoura: I think for the user it will not matter that much. Or do you think we should keep that behavior somehow ?

features/attached_enable.feature
```diff
@@ -657,11 +657,14 @@ Feature: Enable command behaviour when attached to an UA subscription
...
-            """
             Cannot enable FIPS when Livepatch is enabled.
             """
+        Then I verify that running `ua enable fips --assume-yes --format json` `with sudo` exits `1`
+        And stdout is a json matching the `enable` schema
+        And I will see the following on stdout:
+        """
+        {"_schema_version": "0.1", "errors": [{"message": "Cannot enable FIPS when Livepatch is enabled.", "service": "fips", "type": "service"}], "failed_services": ["fips"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
```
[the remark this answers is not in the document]
renanrodrigo: This falls onto the same discussion in the required-services PR, and how `--assume-yes` works with the JSON output

features/attached_enable.feature
renanrodrigo: I have no strong opinion on this. I think we should keep errors in stderr just for conformance's sake, but I don't think it's mandatory

features/attached_enable.feature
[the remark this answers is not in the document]
lucasmoura: The reason why this fails is that we are using the `block_disable_on_enable` feature override. So that is actually expected in this test

features/attached_enable.feature
lucasmoura: I think this is a good point, but looking at the code, it seems that we have other places where we are printing errors on `stdout` instead of `stderr`.

For example:
https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/cli.py#L1105
https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/cli.py#L1127

And through the `action_enable` function.

Because of that, I think we could tackle this into a separate PR.
What do you all think @orndorffgrant @CalvoM @renanrodrigo ?

orndorffgrant: No strong opinion here. I don't think anyone is relying on this message to be printed to stderr AFAIK

renanrodrigo: Fair points, we can tackle it in the future as a maintenance task, lowest priority

orndorffgrant (approved):
Sorry for the delay, looks great! And the failed jenkins tests pass after a rebase (tested locally). Merging now

PR #1937, opened 2022-01-14
enable: show required services message on json

lucasmoura:
## Proposed Commit Message
enable: show required services message on json

We are now adding the required services message to the json output
of the enable command

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -798,6 +798,12 @@ Feature: Enable command behaviour when attached to an UA subscription
         ROS ESM Security Updates cannot be enabled with UA Apps: ESM disabled.
         Enable UA Apps: ESM and proceed to enable ROS ESM Security Updates\? \(y\/N\) Cannot enable ROS ESM Security Updates when UA Apps: ESM is disabled.
         """
+        When I verify that running `ua enable ros --beta --format json` `with sudo` and stdin `N` exits `1`
+        Then I will see the following on stdout:
+        """
+        ROS ESM Security Updates cannot be enabled with UA Apps: ESM disabled.
+        Enable UA Apps: ESM and proceed to enable ROS ESM Security Updates? (y/N) {"_schema_version": "0.1", "errors": [{"message": "Cannot enable ROS ESM Security Updates when UA Apps: ESM is disabled.\n", "service": "ros", "type": "service"}], "failed_services": ["ros"], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+        """
 
         When I run `ua enable ros --beta` `with sudo` and stdin `y`
         Then stdout matches regexp
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -63,41 +63,53 @@ Feature: Enable command behaviour when attached to an UA subscription
         And stdout is a json matching the `enable` schema
         And I will see the following on stdout:
             """
+            {"_schema_version": "0.1", "errors": [{"message": "json formatted response requires --assume-yes flag.", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
+        Then I verify that running `ua enable foobar --format json` `with sudo` exits `1`
+        And stdout is a json matching the `enable` schema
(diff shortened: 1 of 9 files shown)


features/attached_enable.feature
```diff
@@ -795,6 +795,12 @@ Feature: Enable command behaviour when attached to an UA subscription
         ROS ESM Security Updates cannot be enabled with UA Apps: ESM disabled.
         Enable UA Apps: ESM and proceed to enable ROS ESM Security Updates\? \(y\/N\) Cannot enable ROS ESM Security Updates when UA Apps: ESM is disabled.
         """
+        When I verify that running `ua enable ros --beta --format json` `with sudo` and stdin `N` exits `1`
+        Then I will see the following on stdout:
+        """
+        ROS ESM Security Updates cannot be enabled with UA Apps: ESM disabled.
```
renanrodrigo, line 801: I think when we have `--format json` we should have only JSON returned in the screen. The prompt messages need to be aware of that somehow.
If any script is calling this and trying to parse the JSON response, it will probably error out on the prompt line.

I am just not sure of how to handle this:
- do we require `--assume-yes` when enabling services which depend on others, and error out in the JSON if it is not passed?
- do we `--assume-yes` all the time and forget about the prompt?

orndorffgrant: I think we should require that `--assume-yes` be passed at the cli if `--format=json` is passed.
So `ua enable ros --beta --format=json` would error immediately saying something like:
`{ "error": [{"message": "json formatted response requires --assume-yes flag"}]}`

Then we won't have this problem, and I think it makes sense from a user perspective.

lucasmoura: That's a good question, but I don't know how to better handle. My original idea was to use `--assume-yes` by default when running a command, but we discussed on #1912 and we decided not to use it.

My current opinion is that an user will be able to understand that it needs the `--assume-yes` if the command is running in an automated fashion. 

But maybe we could use the first idea, even though it make the command a little bit more verbose. What do you think @orndorffgrant ?

orndorffgrant: And I just realized my suggestion is exactly @renanrodrigo's first suggestion, so that is what I think we should do  :)

orndorffgrant: Actually, my suggestion is a little stronger. I think we should always require --assume-yes if --format=json (not only if the service has dependent services).
This is simpler to understand and implement, plus it is more consistent for the user - it doesn't require them to understand which services may require a dependent service or not when determining if they need the flag or not

lucasmoura: Okay, I will make that flag required then on `--format json`

lucasmoura:
@orndorffgrant @renanrodrigo @CalvoM PR updated with `--assume-yes` requirement

orndorffgrant (approved):
LGTM!

features/attached_enable.feature
```diff
@@ -61,43 +61,55 @@ Feature: Enable command behaviour when attached to an UA subscription
...
-        And I verify that running `ua enable esm-infra foobar --format json` `with sudo` exits `1`
+        And I verify that running `ua enable esm-infra foobar --format json --assume-yes` `with sudo` exits `1`
         Then stdout is a json matching the `enable` schema
         And I will see the following on stdout:
         """
         {"_schema_version": "0.1", "errors": [{"message": "Cannot enable unknown service 'foobar'.\nTry <valid_services>", "service": null, "type": "system"}], "failed_services": ["foobar"], "needs_reboot": false, "processed_services": ["esm-infra"], "result": "failure", "warnings": []}
         """
         When I run `ua disable esm-infra esm-apps` with sudo
-        And I run `ua enable esm-infra esm-apps --beta --format json` with sudo
+        And I run `ua enable esm-infra esm-apps --beta --format json --assume-yes` with sudo
```
renanrodrigo, line 112: Should we have at least one line running it without `--assume-yes` and showing that it fails?

renanrodrigo (approved):
Single question, otherwise looks good

features/attached_enable.feature
renanrodrigo: :man_facepalming: 
blind as am I

PR #1939, opened 2022-01-17
fips: allow enabling FIPS on focal clouds

lucasmoura:
## Proposed Commit Message
fips: allow enabling FIPS on focal clouds

The focal clouds now possess fips optimized kernel on the focal release. We are now allowing FIPS to be enabled on those
clouds

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/features/enable_fips_cloud.feature
+++ b/features/enable_fips_cloud.feature
@@ -16,8 +16,6 @@ Feature: FIPS enablement in cloud based machines
             | release | release_title | fips_service  |
             | xenial  | Xenial        | fips          |
             | xenial  | Xenial        | fips-updates  |
-            | focal   | Focal         | fips          |
-            | focal   | Focal         | fips-updates  |
 
     @series.xenial
     @uses.config.machine_type.azure.generic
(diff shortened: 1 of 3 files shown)


features/enable_fips_cloud.feature
```diff
@@ -617,3 +648,70 @@ Feature: FIPS enablement in cloud based machines
         Examples: ubuntu release
            | release | fips-name    | fips-service |fips-apt-source                                |
            | bionic  | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips-updates/ubuntu bionic-updates/main |
+
+    @slow
+    @series.focal
+    @uses.config.machine_type.gcp.generic
```
lucasmoura, line 654: For GCP we are also testing for `fips-updates`. Should we add that same test procedure for aws and azure ?
Also, the main difference here between fips and fips-updates is the hold logic we test on FIPS. Do you think we should keep that separation or remove that part of the fips test and unify the fips and fips updates scenarios ?  PS: We would still have the hold test on the VMs.

What do you think @renanrodrigo @orndorffgrant ?

renanrodrigo: I'd say yes, it is important to test fips-updates for Azure/AWS as well, and I think those should be separate tests as they are for GCP. Justification for both is that `fips-updates` is considered a separate service from the UA perspective.

orndorffgrant: I think we should have tests for fips-updates for all the clouds, however I'm wondering if theres a decent way to avoid too much repetition in our test code for this.
Could we move the `apt hold` test steps to their own scenario, and reuse the other fips scenarios for fips and fips-updates.

Maybe something like this for the apt hold test:
```
    @series.bionic
    @series.focal
    @uses.config.machine_type.aws.generic
    @uses.config.machine_type.azure.generic
    @uses.config.machine_type.gcp.generic
    Scenario Outline: FIPS unholds packages
        And I run `DEBIAN_FRONTEND=noninteractive apt-get install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y openssh-client openssh-server strongswan` with sudo
        And I run `apt-mark hold openssh-client openssh-server strongswan` with sudo
        And I run `ua enable fips --assume-yes` with sudo
        When I run `ua disable fips --assume-yes` with sudo
        When I run `apt-mark unhold openssh-client openssh-server strongswan` with sudo
        """
        openssh-client was already not hold.
        openssh-server was already not hold.
        strongswan was already not hold.
        """
        Examples: ubuntu release
           | release |
           | bionic  |
           | focal   |
```

Then the apt hold logic could be removed from the other fips tests and they could be re-used for fips and fips-updates via the "Examples" table

What do you think @lucasmoura @renanrodrigo ?

lucasmoura: I think that is a great idea

renanrodrigo: Just re-looked at what we are testing, and yes, I do think it works

lucasmoura: I will perform that change then

lucasmoura: @CalvoM  @orndorffgrant @renanrodrigo I have updated the tests, but I made some extra modifications there. Let me know what you all think

features/enable_fips_cloud.feature
```diff
@@ -35,93 +33,29 @@ Feature: FIPS enablement in cloud based machines
            | fips          |
```
CalvoM, line 33: Hasn't this changed, i.e. azure xenial has a fips kernel now?

lucasmoura: @CalvoM doesn't look like it. If we take a look at the fips repository here:
https://esm.ubuntu.com/fips/ubuntu/dists/xenial/main/binary-amd64/

We cannot see the `ubuntu-azure-fips` package. But let me know if I am missing something here

renanrodrigo (approved):
lgtm
rewritten tests make sense and are passing

features/enable_fips_cloud.feature
```diff
@@ -130,192 +64,50 @@ Feature: FIPS enablement in cloud based machines
...
-        Then stdout matches regexp:
-            """
-            <fips-service> +yes                disabled
-            """
 
         Examples: ubuntu release
-           | release | fips-name    | fips-service |fips-apt-source                                |
-           | bionic  | FIPS         | fips         |https://esm.ubuntu.com/fips/ubuntu bionic/main |
+           | release | fips-apt-source                                |
+           | bionic  | https://esm.ubuntu.com/fips/ubuntu bionic/main |
```
orndorffgrant, line 70: Can we just add a focal line here instead of having a separate scenario?
```suggestion
           | bionic  | https://esm.ubuntu.com/fips/ubuntu bionic/main |
           | focal  | https://esm.ubuntu.com/fips/ubuntu focal/main |
```

orndorffgrant (commented):
Looks good! Just one small comment, and I'm going to run the tests before I click approve

features/enable_fips_cloud.feature
lucasmoura: The problem is that we don't have the openssh fips packages on focal. That's why it is in another scenario

orndorffgrant: Ahhhh right, thanks I missed that

features/enable_fips_cloud.feature
```diff
@@ -546,14 +301,17 @@ Feature: FIPS enablement in cloud based machines
         Examples: ubuntu release
            | release | fips-name    | fips-service |fips-apt-source                                |
            | bionic  | FIPS         | fips         |https://esm.ubuntu.com/fips/ubuntu bionic/main |
+           | bionic  | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips/ubuntu bionic/main |
+           | focal   | FIPS         | fips         |https://esm.ubuntu.com/fips/ubuntu focal/main  |
+           | focal   | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips/ubuntu focal/main  |
 
     @slow
     @series.bionic
+    @seris.focal
```
orndorffgrant, line 310: Running through the integration tests has proved worth it! - I found a tiny typo :) @lucasmoura 
```suggestion
    @series.focal
```

lucasmoura: Great catch. I have fixed it now

orndorffgrant (approved):
Tests passed - thank you!

features/enable_fips_cloud.feature
CalvoM: Renan and Grant clarified about this during the backlog meeting. So it is okay.

PR #1940, opened 2022-01-17
tests: fix schema version for fips tests

lucasmoura:


renanrodrigo (approved):
Thanks for noticing those!

renanrodrigo:
I think at least the CC-EAL ones were already fixed in 8e0c118

lucasmoura:
@renanrodrigo you are right, I have kept only the fips fix now

renanrodrigo (approved):

PR #1941, opened 2022-01-17
cli: only go for resources on explicit help calls (SC-738)

renanrodrigo:
When checking for `presentedAs` we ended up calling the resources endpoint for every CLI run, on the `get_parser` call, when we only need it on help calls. 
This PR moves the logic to call the resources endpoint to that specific method.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -16,7 +16,7 @@ import textwrap
 import time
 from functools import wraps
 from typing import Optional  # noqa: F401
-from typing import List
+from typing import List, Tuple
 
 import yaml
 
@@ -40,6 +40,7 @@ from uaclient.defaults import (
     CLOUD_BUILD_INFO,
     CONFIG_FIELD_ENVVAR_ALLOWLIST,
     DEFAULT_CONFIG_FILE,
+    PRINT_WRAP_WIDTH,
 )
 
 # TODO: Better address service commands running on cli
@@ -93,8 +94,6 @@ class UAArgumentParser(argparse.ArgumentParser):
         epilog=None,
         formatter_class=argparse.HelpFormatter,
         base_desc: str = None,
-        non_beta_services_desc: List[str] = None,
-        beta_services_desc: List[str] = None,
     ):
         super().__init__(
             prog=prog,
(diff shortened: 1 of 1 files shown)


uaclient/cli.py
```diff
@@ -103,26 +101,62 @@ def __init__(
...
+                self._get_service_descriptions()
+            )
+            services = sorted(non_beta_services_desc)
             if show_all:
-                services = sorted(services + self.beta_services_desc)
+                services = sorted(services + beta_services_desc)
             self.description = "\n".join([self.base_desc] + services)
         super().print_help(file=file)
 
+    @staticmethod
```
lucasmoura, line 120: Why this method need to be static ?

uaclient/cli.py
```diff
@@ -103,26 +101,62 @@ def __init__(
...
-            self.non_beta_services_desc,
-            self.beta_services_desc,
-        ]
-        if any(desc_vars):
-            services = sorted(self.non_beta_services_desc)
+        if self.base_desc:
+            non_beta_services_desc, beta_services_desc = (
+                self._get_service_descriptions()
+            )
+            services = sorted(non_beta_services_desc)
```
lucasmoura, line 114: I think we can change this var name to indicate that this is a list of services descriptions

uaclient/cli.py
```diff
@@ -103,26 +101,62 @@ def __init__(
...
 
     def print_help(self, file=None, show_all=False):
-        desc_vars = [
-            self.base_desc,
-            self.non_beta_services_desc,
-            self.beta_services_desc,
-        ]
-        if any(desc_vars):
-            services = sorted(self.non_beta_services_desc)
+        if self.base_desc:
```
lucasmoura, line 110: Will this value ever be `None` ? It seems we get the `__doc__` attribute of cli module whenever we create the `UAArgumentParser`.

uaclient/cli.py
renanrodrigo: Well it is an utility function inside the class, and it does not need the class itself or any instance to run. I kept it inside the class just because it is relevant only for a method this class particularly.

uaclient/cli.py
renanrodrigo: It will be None when the parser is called implicitly.
I have found that when calling `ua status --help` for instance, it might get `None`'d by argparse somehow.
I kept the same logic @blackboxsw had there before (looking for any of the desc_vars), just adapting it because what we really want to know is if `base_desc` has been set - the services will only be used in that case anyway.

lucasmoura: I think what is happening is that only the main parser will have the `base_desc` value. All subparsers won't. When the `parser.parse_args` is called, we end up calling `print_help` on the subparser, which doesn't have the `base_desc`  value populated.

Which seems to work fine for what we are trying to achieve

renanrodrigo: Makes sense!

uaclient/cli.py
```diff
@@ -103,26 +101,66 @@ def __init__(
...
+                )
+                if len(service_line) <= 80:
+                    service_info = [service_line]
+                else:
+                    wrapped_words = []  # type: List[str]
+                    line = service_line
+                    while len(line) > 80:
+                        [line, wrapped_word] = line.rsplit(" ", 1)
+                        wrapped_words.insert(0, wrapped_word)
+                    service_info = [line + "\n   " + " ".join(wrapped_words)]
```
orndorffgrant, line 155: Since we're touching this logic, we can probably replace this chunk with something like
```python
textwrap.wrap(
  service_line,
  width=defaults.PRINT_WRAP_WIDTH,
  subsequent_indent="   ",
  break_long_words=False,
  break_on_hyphens=False
)
```

orndorffgrant (approved):
Thanks for optimizing this!
Just two optional suggestions

uaclient/cli.py
renanrodrigo: indeed

renanrodrigo:
@orndorffgrant changes applied !

PR #1943, opened 2022-01-19
jobs: disable jobs after attach/auto-attach

lucasmoura:
## Proposed Commit Message
jobs: disable jobs after attach/auto-attach

After a attach/auto-attach opeartion, we are now disabling jobs if necessary, for example, the license check job for GCP

## Test Steps
Run the license check integration tests

## Desired commit type

## Checklist


orndorffgrant (approved):
Thank you for cleaning up my mistakes :)

Tests passed - LGTM

lucasmoura:
@orndorffgrant No worries at all, if we had shipped the daemon, that error would not have happened. So no problem at all

PR #1944, opened 2022-01-20
tests: fix tests for jammy

lucasmoura:
## Proposed Commit Message
tests: fix tests for jammy

On jammy, we are using new versions of the libraries and also Python 3.10. That means that some of our tests are making assumptions with old libraries. We have updated the tests to behave properly on jammy

## Test Steps
Build a jammy package without the fix applied and see that it fails
Apply the test fix and run the build again

## Desired commit type

## Checklist


orndorffgrant (approved):
Thanks for catching this!

PR #1945, opened 2022-01-20
ci: run jammy tests

orndorffgrant:
We should start running these tests on jammy regularly as we approach release.

This will make CI take longer though :/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -141,6 +141,27 @@ pipeline {
                         '''
                     }
                 }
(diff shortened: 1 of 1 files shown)


paride:
Thanks! CI retriggered after setting up Jammy chroots on the Jenkins nodes. That wasn't done automatically because `debootstrap` in `bionic-release` don't know about Jammy yet; I had to install the version from `-proposed`.

paride (approved):
Thanks!

PR #1946, opened 2022-01-20
Release 27.6

lucasmoura:


lucasmoura:
@CalvoM this is the PR we put for a release.
As you can see, we bring the commits from 'main` that we want to include in the release while also creating a new `debian/changelog` entry and bumping the version string

We test this branch by comparing it to main to see if we are not missing anything.
We also build a package from this branch for every series we support to see if everything is working as expected.

Please let me know if you have any other questions regarding it

debian/changelog
```diff
@@ -1,3 +1,17 @@
+ubuntu-advantage-tools (27.6~22.04.1) jammy; urgency=medium
+
+  * d/tools.postinst:
+    - make log files world readable
+  * New upstream release 27.6 (LP: #1958556)
+    - cli: only go for resources on explicit help calls
```
orndorffgrant, line 6: "go for resources" makes sense to me but likely won't for someone outside the project
```suggestion
    - cli: only request available resources from contract server when necessary
```

debian/changelog
```diff
@@ -1,3 +1,17 @@
...
+    - fips:
+      + allow enabling FIPS on focal clouds
+      + update prompt messages
+    - jobs: disable jobs after attach/auto-attach
```
orndorffgrant, line 10: Specifically the license-check job
```suggestion
    - jobs: disable license-check job on GCP after attach
```

orndorffgrant (commented):
Two small changelog suggestions

debian/changelog
lucasmoura: Just modified the message a little bit for size constraints. Let me know what you think

orndorffgrant (approved):
Looks great! Thank you

CalvoM:
Thanks for tagging me on this.

PR #1948, opened 2022-01-21
fips: allow fips on containers

orndorffgrant:
## Proposed Commit Message

```
fips: allow fips on containers
```

## Test Steps

Try out `ua enable fips` on containers and vms to make sure it works like you would expect.

## Desired commit type

## Checklist

### DIFF
--- a/features/enable_fips_container.feature
+++ b/features/enable_fips_container.feature
@@ -1,0 +1,144 @@
+
+@uses.config.contract_token
+Feature: FIPS enablement in lxd containers
+
+    @series.xenial
+    @series.bionic
+    @series.focal
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attached enable of FIPS in an ubuntu lxd container
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `DEBIAN_FRONTEND=noninteractive apt-get install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y openssh-client openssh-server strongswan openssl <libssl> libgcrypt20` with sudo, retrying exit [100]
+        And I run `ua enable fips<updates>` `with sudo` and stdin `y`
+        Then stdout matches regexp:
+            """
+            Warning: Enabling <fips-name> in a container.
+                     This will install the FIPS packages but not the kernel.
+                     This container must run on a host with <fips-name> enabled to be
+                     compliant.
+            Warning: This action can take some time and cannot be undone.
+            """
+        And stdout matches regexp:
+            """
+            Updating package lists
+            Installing <fips-name> packages
+            <fips-name> enabled
+            A reboot is required to complete install.
+            Please run `apt upgrade` to ensure all FIPS packages are updated to the correct
+            version.
+            """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            fips<updates> +yes                enabled
+            """
+        And stdout matches regexp:
+            """
+            FIPS support requires system reboot to complete configuration
+            """
+        And I verify that running `apt update` `with sudo` exits `0`
+        And I verify that `openssh-server` is installed from apt source `https://esm.ubuntu.com/fips<updates>/ubuntu <release><updates>/main`
+        And I verify that `openssh-client` is installed from apt source `https://esm.ubuntu.com/fips<updates>/ubuntu <release><updates>/main`
+        And I verify that `strongswan` is installed from apt source `https://esm.ubuntu.com/fips<updates>/ubuntu <release><updates>/main`
+        And I verify that `strongswan-hmac` is installed from apt source `https://esm.ubuntu.com/fips<updates>/ubuntu <release><updates>/main`
+        And I verify that `openssl` is installed from apt source `https://esm.ubuntu.com/fips<updates>/ubuntu <release><updates>/main`
+        And I verify that `<libssl>` is installed from apt source `https://esm.ubuntu.com/fips<updates>/ubuntu <release><updates>/main`
+        And I verify that `<libssl>-hmac` is installed from apt source `https://esm.ubuntu.com/fips<updates>/ubuntu <release><updates>/main`
+        And I verify that `<additional-fips-packages>` are installed from apt source `https://esm.ubuntu.com/fips<updates>/ubuntu <release><updates>/main`
+        When I reboot the `<release>` machine
+        When I run `ua status --all` with sudo
+        Then stdout does not match regexp:
+            """
+            FIPS support requires system reboot to complete configuration
+            """
+        When I run `ua disable fips<updates>` `with sudo` and stdin `y`
+        Then stdout matches regexp:
+            """
+            This will disable the FIPS entitlement but the FIPS packages will remain installed.
+            """
+        And stdout matches regexp:
+            """
+            Updating package lists
+            """
+        And stdout does not match regexp:
+            """
+            A reboot is required to complete disable operation
+            """
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            fips<updates> +yes                disabled
+            """
+        Then stdout does not match regexp:
+            """
+            Disabling FIPS requires system reboot to complete operation
+            """
+        When I run `apt-cache policy ubuntu-fips` as non-root
+        Then stdout does not match regexp:
+            """
+            .*Installed: \(none\)
+            """
+        Then I verify that `openssh-server` installed version matches regexp `fips`
+        And I verify that `openssh-client` installed version matches regexp `fips`
+        And I verify that `strongswan` installed version matches regexp `fips`
+        And I verify that `strongswan-hmac` installed version matches regexp `fips`
+        And I verify that `openssl` installed version matches regexp `fips`
+        And I verify that `<libssl>` installed version matches regexp `fips`
+        And I verify that `<libssl>-hmac` installed version matches regexp `fips`
+        And I verify that packages `<additional-fips-packages>` installed versions match regexp `fips`
+
+        Examples: ubuntu release
+           | release | fips-name    | updates  | libssl      | additional-fips-packages                                             |
+           | xenial  | FIPS         |          | libssl1.0.0 | openssh-server-hmac openssh-client-hmac                              |
+           | xenial  | FIPS Updates | -updates | libssl1.0.0 | openssh-server-hmac openssh-client-hmac                              |
+           | bionic  | FIPS         |          | libssl1.1   | openssh-server-hmac openssh-client-hmac libgcrypt20 libgcrypt20-hmac |
+           | bionic  | FIPS Updates | -updates | libssl1.1   | openssh-server-hmac openssh-client-hmac libgcrypt20 libgcrypt20-hmac |
+           | focal   | FIPS         |          | libssl1.1   | libgcrypt20 libgcrypt20-hmac                                         |
+           | focal   | FIPS Updates | -updates | libssl1.1   | libgcrypt20 libgcrypt20-hmac                                         |
+
+    @series.xenial
+    @series.bionic
+    @series.focal
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Try to enable FIPS after FIPS Updates in a lxd container
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        When I run `ua status --all` with sudo
+        Then stdout matches regexp:
+            """
+            fips-updates +yes +disabled
+            """
+        And stdout matches regexp:
+            """
+            fips +yes +disabled
+            """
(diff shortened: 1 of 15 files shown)


uaclient/entitlements/cis.py
```diff
@@ -36,9 +36,9 @@ def messaging(self,) -> Dict[str, List[Union[str, Tuple[Callable, Dict]]]]:
         return messages
 
     @property
-    def packages(self) -> List[str]:
+    def packages(self) -> Optional[List[str]]:
         if self._called_name == "usg":
-            return []
+            return None
```
orndorffgrant, line 41: This change was made so that in `fips.py` we can return an empty list for `packages` but `enable` will still call `install_packages`.

Now `install_packages` is called even for len 0 `packages`, but is not called if `packages` is None

orndorffgrant:
Looks like I missed a couple integration tests I'll need to edit

uaclient/entitlements/fips.py
```diff
@@ -12,6 +12,74 @@
 event = event_logger.get_event_logger()
 
 
+FIPS_CONDITIONAL_PACKAGES = {
+    "xenial": [
+        "strongswan",
```
lucasmoura, line 17: I think we should have the list of those package inside auxiliary variables and reference then on the dicts

uaclient/entitlements/fips.py
```diff
@@ -12,6 +12,74 @@
...
+# These lists need to be kept up to date with the
+# Depends of ubuntu-fips.
+FIPS_CONTAINER_CONDITIONAL_PACKAGES = {
+    "xenial": [
+        "strongswan",
+        "strongswan-hmac",
+        "openssh-client",
+        "openssh-client-hmac",
+        "openssh-server",
+        "openssh-server-hmac",
```
lucasmoura, line 51: If we have those variables in auxiliary one, we can just reference it on all the other series add just the specific packages on each one of them

uaclient/entitlements/cis.py
lucasmoura: I think I am missing something, looking at the modification on `install_packages`, wouldn't it work with the empty list as well ?

uaclient/entitlements/fips.py
```diff
@@ -155,7 +211,11 @@ def _allow_fips_on_cloud_instance(
             if series in ("bionic", "focal"):
                 return True
 
-            return bool("ubuntu-gcp-fips" in super().packages)
+            super_packages = super().packages
+            if super_packages is not None:
+                return bool("ubuntu-gcp-fips" in super_packages)
```
lucasmoura, line 216: If `packages` return an empty list, we would not need that logic.
But again, I can still be missing the rationale for that change

uaclient/entitlements/fips.py
```diff
@@ -12,6 +12,74 @@
...
+        "strongswan-hmac",
+        "openssh-client",
+        "openssh-client-hmac",
+        "openssh-server",
+        "openssh-server-hmac",
+        "openssl",
+        "libssl1.1",
+        "libssl1.1-hmac",
+        "libgcrypt20",
+        "libgcrypt20-hmac",
```
lucasmoura, line 67: For all of the extra packages here, I am assuming those are installed by the metapackage.
If that is true, then we would have a problem if the metapackage starts depending on another package, since we would require an SRU to update this list.

We could parse the dependencies of the `ubuntu-fips` package, but it has other dependencies there that we would need to remove, like `grub2-common` for example.

This would guarantee that we would have the updated packages whenever we need them. But it may be an overkill, since I don't expect the metadata dependencies to change. But let me know if you think this is worth it

uaclient/entitlements/fips.py
```diff
@@ -236,13 +290,21 @@ def _replace_metapackage_on_cloud_instance(
...
-        return self._replace_metapackage_on_cloud_instance(packages)
+        if packages is not None:
+            return self._replace_metapackage_on_cloud_instance(packages)
+        return None
 
     def application_status(self) -> Tuple[status.ApplicationStatus, str]:
         super_status, super_msg = super().application_status()
 
+        if util.is_container() and not util.should_reboot():
+            self.cfg.remove_notice("", status.MESSAGE_FIPS_REBOOT_REQUIRED)
```
lucasmoura, line 305: Do we create this notice when enabling FIPS on a container ? Maybe we should instead restrict creating that notice during enable instead of cleaning it up here

orndorffgrant: The problem is that `should_reboot` returns true after enabling fips in a container. I didn't look into it closely, but it made sense to me at the time that upgrading libssl might require services to restart (even if it technically doesn't require a system reboot).

uaclient/entitlements/fips.py
orndorffgrant: Right, these additional packages are from the metapackage.

I'm not sure if parsing the metapackage deps is worth it. I don't think a delay in updating this list via SRU is a major problem because we should recommend anyone who enabled fips to run `apt upgrade` anyway, and that should capture any fips upgrades that are necessary.

But I'm open to other opinions - what do you think?

uaclient/entitlements/fips.py
orndorffgrant: Thats a good idea! I'll make this change.

uaclient/entitlements/repo.py
```diff
@@ -72,7 +74,7 @@ def _perform_enable(self, silent: bool = False) -> bool:
         @raises: UserFacingError on failure to install suggested packages
         """
         self.setup_apt_config(silent=silent)
-        if self.packages:
+        if self.packages is not None:
```
orndorffgrant, line 77: @lucasmoura I should've left my `packages` comment down here.

Before this change, `_perform_enable` only calls `self.install_packages` if self.packages is a list with at least one item.

For the fips on containers use case, we want self.packages to be empty, but we still want to call `self.install_packages` so that `FIPSCommonEntitlement.install_packages` runs to install the condition_packages.

So the way I thought to do that was to distinguish between "No packages to install, don't even bother calling install_packages" (`None`) and the fips on containers scenario which is "No packages to install, but still call install_packages because there are conditional_packages" (`[]`).

I'm open to other ways of accomplishing this though

uaclient/entitlements/cis.py
orndorffgrant: The problem is that `install_packages` won't even get called with an empty list. I left a longer comment on this below

uaclient/entitlements/fips.py
orndorffgrant: Updated!

uaclient/entitlements/repo.py
lucasmoura: Okay, I see the problem now. What do you think if we move that logic into `install_packages` ? We would check inside that function if we actually have any packages to install. In that way, we would need to modify the `packages` property of the FIPS classes.

uaclient/entitlements/tests/test_fips.py
```diff
@@ -39,31 +39,109 @@ def entitlement(fips_entitlement_factory):
...
 class TestFIPSEntitlementDefaults:
-    @pytest.mark.parametrize("series", (("xenial"), ("bionic"), ("focal")))
+    @pytest.mark.parametrize(
+        "series, is_container, expected",
+        (
+            (
+                "xenial",
+                False,
+                [
+                    "strongswan",
```
lucasmoura, line 49: I think we can use the constant package variables here

uaclient/entitlements/repo.py
orndorffgrant: Ahhh thats a good idea! I'll try that out

uaclient/entitlements/tests/test_fips.py
orndorffgrant: Good point!

uaclient/entitlements/fips.py
orndorffgrant: Oh yes we should, I'll add it

orndorffgrant:
@lucasmoura updated! Let me know what you think of the new changes

features/enable_fips_container.feature
```diff
@@ -0,0 +1,97 @@
...
+        When I attach `contract_token` with sudo
+        And I run `DEBIAN_FRONTEND=noninteractive apt-get install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y openssh-client openssh-server strongswan openssl <libssl> libgcrypt20` with sudo, retrying exit [100]
+        And I run `ua enable fips<updates>` `with sudo` and stdin `y`
+        Then stdout matches regexp:
+            """
+            Warning: Enabling <fips-name> in a container.
+                     This will install the FIPS packages but not the kernel.
+                     This container must run on a host with <fips-name> enabled to be
+                     compliant.
+                     Please run `apt upgrade` to ensure all FIPS packages are updated to
```
lucasmoura, line 20: Maybe we should have this message after the operation is finished

uaclient/entitlements/repo.py
```diff
@@ -217,6 +215,16 @@ def install_packages(
...
 
+        if not package_list:
+            package_list = self.packages
+
+        if not package_list:
+            return
+
+        msg_ops = self.messaging.get("pre_install", [])
+        if not util.handle_message_operations(msg_ops):
+            return
```
lucasmoura, line 226: Don't know if this is worth it, but we could change `install_packages` to return a boolean indicating if any packages was actually installed, then we can remove this message logic back into `_perform_enable`. What do you think ?

orndorffgrant: That would work to print the message after installing, but this specifically needs to happen before installing as far as I can tell :/

features/enable_fips_container.feature
orndorffgrant: thats a good idea - I'll add it to post_enable if on a container

features/enable_fips_container.feature
```diff
@@ -0,0 +1,97 @@
...
+                     the correct version.
+            Warning: This action can take some time and cannot be undone.
```
lucasmoura, line 22: When answering no to this prompt I got a an unexpected error prompt.
I saw that on a bionic container after running the `fips-updates` test:

```
Warning: Enabling FIPS in a container.
         This will install the FIPS packages but not the kernel.
         This container must run on a host with FIPS enabled to be
         compliant.
         Please run `apt upgrade` to ensure all FIPS packages are updated to
         the correct version.
Warning: This action can take some time and cannot be undone.
Are you sure? (y/N) N
Unexpected error(s) occurred.
For more details, see the log: /var/log/ubuntu-advantage.log
To file a bug run: ubuntu-bug ubuntu-advantage-tools
```

uaclient/entitlements/repo.py
lucasmoura: My bad here, I incorrectly read this as `post_install` :facepalm:

features/enable_fips_container.feature
orndorffgrant: Oh weird, I thought I tested that but I guess it was early on implementing this and some things have changed. I'll look into it

features/enable_fips_container.feature
```diff
@@ -0,0 +1,97 @@
+
+@uses.config.contract_token
```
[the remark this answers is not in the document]
orndorffgrant: That's fair, we could do that. Do you think that can replace the "fips after fips-updates" vm test? Or should it be in addition?

features/enable_fips_container.feature
orndorffgrant: I don't think it can be a "Notice" because we wouldn't know when to remove it

features/enable_fips_container.feature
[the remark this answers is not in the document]
lucasmoura: I think it should be in addition, since containers tests don't take that long to run

orndorffgrant:
@lucasmoura Ready for another review :) The exception on enable was a little bug that slipped in - fixed in a separate commit

uaclient/util.py
```diff
@@ -839,9 +839,7 @@ def is_installed(package_name: str) -> bool:
         return False
 
 
-def handle_message_operations(
-    msg_ops: List[Union[str, Tuple[Callable, Dict]]],
-) -> bool:
+def handle_message_operations(msg_ops: Optional[MessagingOperations],) -> bool:
```
lucasmoura, line 842: Why are marking the `msg_ops` as `Optional` ?

lucasmoura (approved):
Just one little nitpick, but otherwise LGTM

renanrodrigo (approved):
went late to this review - many useful discussions (:
LGTM

uaclient/util.py
orndorffgrant: This makes the logic over in `FIPSEntitlement.messaging` a bit easier because we can set `"post_enable": None` if not on a container. And it seemed like a reasonable check to make here.

orndorffgrant:
Thanks!
The one jenkins failure was a lxd flake. Merging now

PR #1950, opened 2022-01-21
status: blocked_by services

orndorffgrant:
## Proposed Commit Message

```
status: include blocked_by for each service

blocked_by is a per-service field. If populated for a given service, the
value is an array of other services that are currently enabled and
incompatible with the given service. Disabling all of the services in
"blocked_by" will allow you to enable the blocked service.
```

## Test Steps

Run `ua status --format=json` on a vm with combinations of fips, fips-updates, and livepatch enabled and make sure the "blocked_by" values are appropriate.

Also maybe run the updated fips integration tests (they are marked `@slow`)
## Desired commit type

## Checklist

### DIFF
--- a/features/enable_fips_vm.feature
+++ b/features/enable_fips_vm.feature
@@ -8,7 +8,12 @@ Feature: FIPS enablement in lxd VMs
     Scenario Outline: Attached enable of FIPS in an ubuntu lxd vm
         Given a `<release>` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
-        And I run `ua disable livepatch` with sudo
+        When I run `ua status --format json` with sudo
+        Then stdout contains substring
+        """
+        {"available": "yes", "blocked_by": [{"name": "livepatch", "reason": "Livepatch cannot be enabled while running the official FIPS certified kernel. If you would like a FIPS compliant kernel with additional bug fixes and security updates, you can use the FIPS Updates service with Livepatch.", "reason_code": "livepatch-invalidates-fips"}], "description": "NIST-certified core packages", "description_override": null, "entitled": "yes", "name": "fips", "status": "disabled", "status_details": "FIPS is not configured"}
+        """
+        When I run `ua disable livepatch` with sudo
         And I run `DEBIAN_FRONTEND=noninteractive apt-get install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y openssh-client openssh-server strongswan` with sudo, retrying exit [100]
         And I run `apt-mark hold openssh-client openssh-server strongswan` with sudo
         And I run `ua enable <fips-service>` `with sudo` and stdin `y`
@@ -40,6 +45,12 @@ Feature: FIPS enablement in lxd VMs
         And I verify that `openssh-server-hmac` is installed from apt source `<fips-apt-source>`
         And I verify that `openssh-client-hmac` is installed from apt source `<fips-apt-source>`
         And I verify that `strongswan-hmac` is installed from apt source `<fips-apt-source>`
+        When I run `ua status --format json` with sudo
+        Then stdout contains substring:
+        """
+        {"available": "yes", "blocked_by": [{"name": "fips", "reason": "Livepatch cannot be enabled while running the official FIPS certified kernel. If you would like a FIPS compliant kernel with additional bug fixes and security updates, you can use the FIPS Updates service with Livepatch.", "reason_code": "livepatch-invalidates-fips"}], "description": "Canonical Livepatch service", "description_override": null, "entitled": "yes", "name": "livepatch", "status": "n/a", "status_details": "Cannot enable Livepatch when FIPS is enabled."}
+        """
+
         When I reboot the `<release>` machine
         And  I run `uname -r` as non-root
         Then stdout matches regexp:
@@ -157,6 +168,12 @@ Feature: FIPS enablement in lxd VMs
         And I verify that `openssh-server-hmac` is installed from apt source `<fips-apt-source>`
         And I verify that `openssh-client-hmac` is installed from apt source `<fips-apt-source>`
         And I verify that `strongswan-hmac` is installed from apt source `<fips-apt-source>`
+        When I run `ua status --format json` with sudo
+        Then stdout contains substring:
+        """
+        {"available": "yes", "blocked_by": [{"name": "fips-updates", "reason": "FIPS cannot be enabled if FIPS Updates has ever been enabled because FIPS Updates installs security patches that aren't officially certified.", "reason_code": "fips-updates-invalidates-fips"}], "description": "NIST-certified core packages", "description_override": null, "entitled": "yes", "name": "fips", "status": "n/a", "status_details": "Cannot enable FIPS when FIPS Updates is enabled."}
+        """
+
         When I reboot the `<release>` machine
         And  I run `uname -r` as non-root
         Then stdout matches regexp:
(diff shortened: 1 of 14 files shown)


features/enable_fips_vm.feature
```diff
@@ -8,7 +8,12 @@ Feature: FIPS enablement in lxd VMs
     Scenario Outline: Attached enable of FIPS in an ubuntu lxd vm
         Given a `<release>` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
-        And I run `ua disable livepatch` with sudo
+        When I run `ua status --format json` with sudo
+        Then stdout matches regexp:
```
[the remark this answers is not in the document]
orndorffgrant: We could but then we'd have to have the entire status json output, which is a huge blob of mostly irrelevant data. Neither option seems great to me :/

One thing I just thought of though, is we could add a step that is "Then stdout contains the substring:" which does an `in` comparison (as opposed to `==` or `re.match`) - thoughts?

features/enable_fips_vm.feature
```diff
@@ -8,7 +8,12 @@ Feature: FIPS enablement in lxd VMs
...
+        """
+        \"blocked_by\": \[{\"name\": \"livepatch\", \"reason_code\": \"livepatch\-invalidates\-fips\", \"reason\": \"Livepatch cannot be enabled while running the official FIPS certified kernel. If you would like a FIPS compliant kernel with additional bug fixes and security updates, you can use the FIPS Updates service with Livepatch.\"}\]
```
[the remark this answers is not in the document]
orndorffgrant: This will match the `"blocked_by"` for the `fips` service. So it is saying that `fips` is blocked/cannot be enabled because "livepatch` is enabled.

uaclient/entitlements/fips.py
```diff
@@ -332,6 +332,28 @@ class FIPSEntitlement(FIPSCommonEntitlement):
...
+        from uaclient.entitlements.livepatch import LivepatchEntitlement
+
+        return (
+            IncompatibleService(
+                LivepatchEntitlement,
+                "livepatch-invalidates-fips",
+                "Livepatch cannot be enabled while running the official FIPS"
+                " certified kernel. If you would like a FIPS compliant kernel"
+                " with additional bug fixes and security updates, you can use"
+                " the FIPS Updates service with Livepatch.",
```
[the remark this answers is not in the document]
orndorffgrant: Ah yeah probably, I will do that!

features/enable_fips_vm.feature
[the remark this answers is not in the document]
orndorffgrant: Though it is hard to tell that here, because I didn't include the name of the service that the "blocked_by" belongs to. I should change the integration tests to include that so it is clearer

features/enable_fips_vm.feature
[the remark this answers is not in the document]
lucasmoura: I think that's a good idea

orndorffgrant:
@lucasmoura I think this is looking better now. I also changed slightly the way we're constructing the strings for these messages, that may be somewhat reusable for json output messages

uaclient/status.py
```diff
@@ -616,6 +622,20 @@ def __init__(
     docs_url=DOCUMENTATION_URL
 )
 
+NAMED_MESSAGE_FIPS_UPDATES_INVALIDATES_FIPS = NamedMessage(
+    "fips-updates-invalidates-fips",
+    "FIPS cannot be enabled if FIPS Updates has ever been enabled because"
+    " FIPS Updates installs security patches that aren't officially"
+    " certified.",
+)
+NAMED_MESSAGE_LIVEPATCH_INVALIDATES_FIPS = NamedMessage(
```
lucasmoura, line 631: That looks cool, but just to be sure, the idea would be to extend that to all messages to return the message code on the json response, right ?

uaclient/config.py
```diff
@@ -691,6 +691,15 @@ def _attached_service_status(
...
                 ent_status, details = ent.user_facing_status()
 
+        blocked_by = [
+            {
+                "name": service.entitlement.name,
+                "reason_code": service.named_msg.name,
+                "reason": service.named_msg.msg,
+            }
+            for service in ent.blocking_incompatible_services()
+        ]
```
lucasmoura, line 701: I know this is already being tested on the integration tests, but could we have an unittest that touches this too ?

uaclient/status.py
orndorffgrant: Potentially! I guess it would depend on what the desktop team needs

uaclient/config.py
orndorffgrant: Ah yeah we might as well. I'll add it

lucasmoura:
Also, this xenial test here is failing:
`Attach command with attach config
Maybe it is worth double checking to see if there is any relation with the proposed changes

orndorffgrant:
@lucasmoura looks like a rare timing collision
```
      2022-01-27 21:31:15,796:ERROR:root:Error executing command: ua detach --assume-yes
      2022-01-27 21:31:15,796:ERROR:root:stdout: 
      2022-01-27 21:31:15,796:ERROR:root:stderr: Unable to perform: ua detach.
      Operation in progress: timer metering job (pid:3619)
```

uaclient/status.py
orndorffgrant: I also think we should tackle internationalization with gettext next cycle, so alternatively maybe we could put this off until then

uaclient/config.py
orndorffgrant: Just pushed this test, LMK what you think

lucasmoura (approved):

uaclient/config.py
lucasmoura: LGTM, thanks for that @orndorffgrant

renanrodrigo:
I have a slight impression that the `test_cli_status` unit tests are taking too long to run. Looking into that.

renanrodrigo:
If I got this right, we could mock the call to livepatch in subp
( `[LIVEPATCH_CMD, "status"], retry_sleeps=LIVEPATCH_RETRIES`), 
called by `application_status` for FIPS, when checking for the blocking_incompatible_services. The call to livepatch is what is taking a small second, and many of those calls are generating delays when running the tests.

I only checked the `test_cli_status`, but any test calling status indirectly (and we might have a little leak here and there) will be delayed by this.

orndorffgrant:
Oh wow great find @renanrodrigo thank you! I'll fix this

uaclient/tests/test_gpg.py
```diff
@@ -51,10 +52,11 @@ def test_export_single_key_from_keyring_dir(self, home_dir, tmpdir):
         # Create keyring with both ESM and CC-EAL2 keys
         source_key1.write(data.GPG_KEY1, "wb")
         source_key2.write(data.GPG_KEY2, "wb")
-        gpg.export_gpg_key(
-            source_keyfile=source_key1.strpath,
-            destination_keyfile=destination_keyfile,
-        )
+        with mock.patch("uaclient.util._subp", side_effect=_subp):
+            gpg.export_gpg_key(
```
lucasmoura, line 56: Sorry, just to double check if I understand the logic here. The fixture `_ subp` is yielding the original `_subp` function, therefore if we have a test that does need subp, it can use. Since we are adding the return of `_subp` here as a `side_effect` of the mock, can I assume that all those test that have that mechanism require the real `_subp` call ?

orndorffgrant: @lucasmoura you are exactly right!

Arguably, we should change the code such that these unittests don't require subp, but that would require more significant changes than I thought made sense in this commit (e.g. re-writing this gpg test, re-writing the version_cmp function)

uaclient/tests/test_security.py
```diff
@@ -1479,17 +1485,20 @@ def fake_attach(args, cfg):
...
-            type(m_sys).stdout = m_stdout
-            type(m_stdout).encoding = mock.PropertyMock(return_value="utf-8")
-            prompt_for_affected_packages(
-                cfg=cfg,
-                issue_id="USN-###",
-                affected_pkg_status=affected_pkg_status,
-                installed_packages=installed_packages,
-                usn_released_pkgs=usn_released_pkgs,
-            )
+        with mock.patch("uaclient.util._subp", side_effect=_subp):
```
lucasmoura, line 1488: Just out of curiosity, what is the `subp`  command that we need to test the `prompt_for_affected_packages` ?
Because for that command, I think the test needs to be mocking all of those calls

lucasmoura (commented):
I can be totally wrong here and please correct me if I do, but it seems those new changes are just making explicit places on our tests were we are relying on subp direclty. But we are still having places, like the LIVEPATCH_CMD issue that @renanrodrigo brought it up happening here, right ?

uaclient/tests/test_security.py
orndorffgrant: `version_cmp_le` uses `dpkg --compare-versions` and is important to the functionality of`prompt_for_affected_packages`. There are probably some other `subp`s slipping through here though :/. But the "correct" fix would probably be to re-write `version_cmd_le` in python instead of relying on dpkg and that didn't seem in-scope for this right now.

lucasmoura:
Thanks for all the context here @orndorffgrant and I completely agree with everything you said.

lucasmoura (approved):

renanrodrigo (approved):

orndorffgrant:
The two jenkins failures passed for me locally, merging now

PR #1951, opened 2022-01-24
docs: update information about PRO images

renanrodrigo:
After the release, pinging CPC should be sufficient to update the cloud
images.

## Desired commit type

## Checklist

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -226,9 +226,11 @@ If this is your first time releasing ubuntu-advantage-tools, you'll need to do t
     lxc exec dev-i -- bash /setup_proposed.sh
     ```
 
-    e. Once [ubuntu-advantage-tools shows up in the pending_sru page](https://people.canonical.com/~ubuntu-archive/pending-sru.html), perform the [Ubuntu-advantage-client SRU verification steps](https://wiki.ubuntu.com/UbuntuAdvantageToolsUpdates). This typically involves running all behave targets with `UACLIENT_BEHAVE_ENABLE_PROPOSED=1 UACLIENT_BEHAVE_CHECK_VERSION=<this-version>` and saving the output.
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -256,15 +256,6 @@ If this is your first time releasing ubuntu-advantage-tools, you'll need to do t
...
- 2. Check Xenial, Bionic, Focal, Hirsute, Impish packages
- 3. Select Destination PPA: UA Client Premium [~ua-client/ubuntu/staging]
- 4. Select Destination series: The same series
- 5. Copy options: "Copy existing binaries"
- 6. Click Copy packages
- 7. Notify Pro Image creators about expected Premium PPA version (patviafore/powersj)
- 8. Once new PRO AMIs are publicly available run `./tools/refresh-aws-pro-ids` to update AMIs we test during CI runs
+## Cloud Images Update
+
+After the release process is finished, if there is a need to push a new cloud image with the new version of UA, CPC must be informed. They will be responsible to update the images using the package from the pockets it was released to (whether it is the `stable` PPA or the`-updates` pocket).
```
lucasmoura, line 261: I think we should always ping CPC with a new release. But now I have a another question regarding the release. When should we upload the package to `stable` ? If we do it after the package hits `updates`, then there is no need anymore for using the package on stable and if we upload it to stable after the package moves to `proposed`, errors can be found on the package during the testing phase invalidating that package as `stable`.

So I don't know if we should keep that ppa anymore. What do you think @CalvoM @renanrodrigo  @blackboxsw @orndorffgrant ?

renanrodrigo: Updated according to post-standup discussion

renanrodrigo:
Updated according to post-standup discussion:
- Always ping CPC when the release is done
- Keep the Stable PPA but put the packages there as they hit proposed

lucasmoura (approved):

orndorffgrant (approved):
LGTM

PR #1953, opened 2022-01-27
attach: allow json response format (SC-769)

lucasmoura:
## Proposed Commit Message
attach: allow json response format

This will allow users to have a json response when running `ua attach`

## Test Steps
Run the new integration test

## Desired commit type

## Checklist

### DIFF
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -546,7 +546,7 @@ Feature: Command behaviour when unattached
           """
           {"_schema_version": "0.1", "errors": [{"message": "To use 'esm-infra' you need an Ubuntu Advantage subscription\nPersonal and community subscriptions are available at no charge\nSee https://ubuntu.com/advantage", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
           """
-        
+
         Examples: ubuntu release
           | release |
           | xenial  |
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -1,17 +1,21 @@
 import logging
+import sys
+from typing import Optional  # noqa: F401
 
 from uaclient import (
     clouds,
     config,
     contract,
     entitlements,
+    event_logger,
     exceptions,
-    status,
-    util,
 )
+from uaclient import status as ua_status
+from uaclient import util
 from uaclient.clouds import identity
 
 LOG = logging.getLogger("ua.actions")
+event = event_logger.get_event_logger()
 
 
 def attach_with_token(
@@ -31,13 +35,11 @@ def attach_with_token(
             cfg, token, allow_enable=allow_enable
         )
     except util.UrlError as exc:
-        with util.disable_log_to_console():
-            LOG.exception(exc)
         cfg.status()  # Persist updated status in the event of partial attach
         update_apt_and_motd_messages(cfg)
         raise exc
     except exceptions.UserFacingError as exc:
-        LOG.warning(exc.msg)
+        event.info(exc.msg, file_type=sys.stderr)
         cfg.status()  # Persist updated status in the event of partial attach
         update_apt_and_motd_messages(cfg)
         raise exc
@@ -64,7 +66,7 @@ def auto_attach(
     except contract.ContractAPIError as e:
         if e.code and 400 <= e.code < 500:
             raise exceptions.NonAutoAttachImageError(
-                status.MESSAGE_UNSUPPORTED_AUTO_ATTACH
+                ua_status.MESSAGE_UNSUPPORTED_AUTO_ATTACH
             )
         raise e
     current_iid = identity.get_instance_id()
@@ -94,3 +96,22 @@ def enable_entitlement_by_name(
         cfg, assume_yes=assume_yes, allow_beta=allow_beta, called_name=name
     )
     return entitlement.enable()
+
+
+def status(
+    cfg: config.UAConfig,
+    *,
+    simulate_with_token: Optional[str] = None,
+    show_beta: bool = False
+):
+    """
+    Construct the current UA status dictionary.
+    """
+    if simulate_with_token:
+        status = cfg.simulate_status(
+            token=simulate_with_token, show_beta=show_beta
+        )
+    else:
+        status = cfg.status(show_beta=show_beta)
+
+    return status
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -15,8 +15,7 @@ import tempfile
 import textwrap
 import time
 from functools import wraps
-from typing import Optional  # noqa: F401
-from typing import List, Tuple
+from typing import List, Optional, Tuple  # noqa
 
 import yaml
 
@@ -134,7 +133,7 @@ class UAArgumentParser(argparse.ArgumentParser):
         for resource in resources:
             try:
                 ent_cls = entitlements.entitlement_factory(resource["name"])
-            except entitlements.EntitlementNotFoundError:
+            except exceptions.EntitlementNotFoundError:
                 continue
             # Because we don't know the presentation name if unattached
             presentation_name = resource.get("presentedAs", resource["name"])
@@ -396,6 +395,13 @@ def attach_parser(parser):
             "use the provided attach config file instead of passing the token"
             " on the cli"
         ),
+    )
+    parser.add_argument(
+        "--format",
+        action="store",
+        choices=["cli", "json"],
+        default="cli",
+        help=("output enable in the specified format (default: cli)"),
     )
     return parser
 
@@ -1124,9 +1130,14 @@ def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
 
 
 def _post_cli_attach(cfg: config.UAConfig) -> None:
-    contract_name = cfg.machine_token["machineTokenInfo"]["contractInfo"][
-        "name"
-    ]
+    contract_name = None
+
+    if cfg.machine_token:
+        contract_name = (
+            cfg.machine_token.get("machineTokenInfo", {})
+            .get("contractInfo", {})
+            .get("name")
+        )
 
     if contract_name:
         event.info(
@@ -1138,7 +1149,11 @@ def _post_cli_attach(cfg: config.UAConfig) -> None:
         event.info(ua_status.MESSAGE_ATTACH_SUCCESS_NO_CONTRACT_NAME)
 
     jobs.disable_license_check_if_applicable(cfg)
-    action_status(args=None, cfg=cfg)
+
+    status = actions.status(cfg)
+    output = ua_status.format_tabular(status)
+    event.info(util.handle_unicode_characters(output))
+    event.process_events()
 
 
 @assert_root
@@ -1208,6 +1223,7 @@ def action_auto_attach(args, *, cfg):
         return 0
 
 
+@set_event_mode
 @assert_not_attached
 @assert_root
 @assert_lock_file("ua attach")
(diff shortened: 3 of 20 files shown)


uaclient/cli.py
```diff
@@ -397,6 +397,13 @@ def attach_parser(parser):
             " on the cli"
         ),
     )
+    parser.add_argument(
```
lucasmoura, line 400: I am not creating the `--assume-yes` flag here because we don't prompt the users during attach. But I can add it just for consistency sake

uaclient/entitlements/livepatch.py
```diff
@@ -308,7 +308,7 @@ def process_contract_deltas(
 
         application_status, _ = self.application_status()
         if application_status == status.ApplicationStatus.DISABLED:
-            return True  # only operate on changed directives when ACTIVE
+            return False  # only operate on changed directives when ACTIVE
```
lucasmoura, line 311: @blackboxsw This is the main change I am introducing here, where if the service is marked as disabled, we will mark the `process_contract_deltas` as False. This is how I will be able to detect if a service was successfully enabled during `attach`.

uaclient/tests/test_cli_attach.py
```diff
@@ -202,11 +304,39 @@ def fake_contract_attach(contract_token):
...
         expected_calls = [mock.call(contract_token=token)]
         assert expected_calls == contract_machine_attach.call_args_list
         assert [mock.call(cfg)] == m_update_apt_and_motd_msgs.call_args_list
 
+        # We need to do that since all config objects in this
+        # test will share the same data dir. Since this will
+        # test a successful attach, in the end we write a machine token
+        # file, which will make all other cfg objects here to report
+        # as attached
+        cfg.delete_cache()
```
orndorffgrant, line 318: Oof I understand why this is here. Long term, we should try to eliminate scenarios where our unittests write any files to disk - ideally everything would stay in memory and be reset for each test to avoid this kind of problem IMO.

No action necessary for this PR though

uaclient/tests/test_cli_attach.py
```diff
@@ -276,22 +406,67 @@ def test_attach_config_invalid_config(self, _m_getuid, FakeConfig):
             action_attach(args, cfg=cfg)
         assert "Error while reading fakename: " in e.value.msg
 
+        args = mock.MagicMock(
+            token=None,
+            attach_config=FakeFile(
+                yaml.dump({"token": "something", "enable_services": "cis"}),
+                name="fakename",
+            ),
+        )
```
orndorffgrant, line 415: Do we need to reconstruct this here? or can we reuse the existing `args`?

orndorffgrant: This happens in a couple places in this file

uaclient/tests/test_cli_attach.py
```diff
@@ -311,6 +486,93 @@ def test_attach_config_enable_services(
...
+            "processed_services": ["cis"] if auto_enable else [],
+            "warnings": [],
+        }
+        assert expected == json.loads(fake_stdout.getvalue())
+
+    @mock.patch("uaclient.contract.process_entitlement_delta")
+    @mock.patch("uaclient.util.apply_series_overrides")
+    @mock.patch("uaclient.contract.UAContractClient.request_url")
+    @mock.patch("uaclient.jobs.update_messaging.update_apt_and_motd_messages")
+    def test_attach_when_one_service_fails_to_enable(
```
orndorffgrant, line 517: I'm confused by this test. The name says that one of the services is supposed to fail to enable, but the result doesn't include any failed services

orndorffgrant (commented):
Looking great! Just left a couple scattered comments

uaclient/tests/test_cli_attach.py
lucasmoura: Yep, I totally agree this is necessary refactor

uaclient/actions.py
```diff
@@ -94,3 +95,20 @@ def enable_entitlement_by_name(
         cfg, assume_yes=assume_yes, allow_beta=allow_beta, called_name=name
     )
     return entitlement.enable()
+
+
+def status(
+    cfg: config.UAConfig,
+    *,
+    token: Optional[str] = None,
```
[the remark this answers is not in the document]
lucasmoura: Great point, I will modify it

uaclient/actions.py
```diff
@@ -1,17 +1,20 @@
 import logging
+from typing import Optional  # noqa: F401
```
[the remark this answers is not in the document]
lucasmoura: Actually, I think we are only using `Optional` inside comments here, so we cannot remove that

uaclient/tests/test_cli_attach.py
lucasmoura: Good point, I tried one approach with that test that failed, but I have a better idea now. I have updated the code

lucasmoura:
@orndorffgrant code updated

uaclient/cli.py
renanrodrigo: /meis fine as well

uaclient/actions.py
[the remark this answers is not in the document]
renanrodrigo: Line 103 should have it as a proper type, no?

features/unattached_commands.feature
```diff
@@ -529,6 +529,46 @@ Feature: Command behaviour when unattached
...
+        Then I will see the following on stdout:
+            """
+            {"_schema_version": "0.1", "errors": [{"message": "This command must be run as root (try using sudo).", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
+        When I verify that running `ua attach invalid-token --format json` `with sudo` exits `1`
+        Then I will see the following on stdout:
+            """
+            {"_schema_version": "0.1", "errors": [{"message": "Invalid token. See https://ubuntu.com/advantage", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
+        When I verify that running attach `with sudo` with json response exits `0`
```
renanrodrigo, line 546: If we want attach to work here, I think we could move this whole test to `attach_validtoken.feature.`

features/unattached_commands.feature
```diff
@@ -529,6 +529,46 @@ Feature: Command behaviour when unattached
...
+    @series.all
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Unattached attach with json output
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I verify that running attach `as non-root` with json response exits `1`
+        Then I will see the following on stdout:
+            """
+            {"_schema_version": "0.1", "errors": [{"message": "This command must be run as root (try using sudo).", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            """
+        When I verify that running `ua attach invalid-token --format json` `with sudo` exits `1`
```
renanrodrigo, line 541: imho this specific part of the test should be part of `attach_invalidtoken.feature`

renanrodrigo (approved):
LGTM
Small points on the place for the behave tests, feel free to disagree

features/unattached_commands.feature
lucasmoura: I would avoid splitting the tests like this because we would need to spin a different container just to test that. Or do you mean changing this bit into an existing test on `attach_validtoken` ?

renanrodrigo: Well this one is spinning a container for it, right? I just think it should live somewhere else.

features/unattached_commands.feature
renanrodrigo: This invalid part can be added to the existing invalidtoken test

uaclient/entitlements/livepatch.py
blackboxsw: @lucasmoura +1 on this changeset, though I think your [service_enabled variable name](https://github.com/canonical/ubuntu-advantage-client/pull/1953/files#diff-50c64ebd349b6ae5bfa48dd3603790d30c8b63141732a0d28ae5aa9fab92b6c5R329) is a misnomer because a True returned from process_contract_deltas can also come in the event that there are [no deltas at all via UAEntitlement.process_contract_deltas](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/entitlements/base.py#L824-L825) regardless of whether the service is truly enabled or disabled.


False retval was originally intended to indicate that no config change operations or deltas were processed. But I think we adapted the base class implementation to return True to allow subclasses to short-circuit and avoid any additional processing within the subclass as there is nothing further/more specific to do.

I think your `return False` from the subclass makes sense given our docstrings and original intent. We may need to tease out how the super().process_contract_deltas behaves in the future for the no delta case, but it really shouldn't make too much difference given that your call-site is really just allowing us to track whether we processed all viable deltas for an enabled service. As long as your event handling isn't broken by a case where no contract deltas exist and process_contract_deltas returns True, then I think your approach is safe.

lucasmoura: Good point when there is no deltas to apply. I will see if I can address this somehow.
Thanks for that review @blackboxsw

features/unattached_commands.feature
lucasmoura: True, then we would be spinning two containers if we move it, right ?

uaclient/entitlements/livepatch.py
lucasmoura: @blackboxsw I have updated the code to also check if we have any deltas to apply before marking the service as `processed`

features/unattached_commands.feature
renanrodrigo: No, sorry, I don't understand why.
But if you think it is best to leave this here, it's all right.

features/unattached_commands.feature
```diff
@@ -529,6 +529,46 @@ Feature: Command behaviour when unattached
...
+        Then I will see the following on stdout:
```
renanrodrigo, line 542: Although the JSON output goes to `stdout`, There is still `Invalid token. See https://ubuntu.com/advantage` being printed to `stderr`. Is that a problem? I think we should make sure only JSON is returned to wherever here

lucasmoura: Good catch, I will remove that print to stderr on that situation

features/unattached_commands.feature
lucasmoura: I think I am the one that missed your point. I thought you meant to partially move this test to `attach_validtoken` (that's why I commented that we would then need two containers to test the whole feature), but you mean to move the entire test which makes sense, I will perform this change

lucasmoura:
@renanrodrigo code updated

renanrodrigo (approved):
lgtm

uaclient/actions.py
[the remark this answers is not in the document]
orndorffgrant: @lucasmoura Yep right here https://github.com/canonical/ubuntu-advantage-client/pull/1953/files#diff-e7021664504f26f91b6b5f89c423329b70b2faf10680b2ff22d05c3f3828fa99R104

orndorffgrant:
@lucasmoura I'm seeing some non json formatted output on attach still
```
root@ua-xenial-66074aa13154a4172bf43edf79768ae9:~# ua attach --format json TOKEN
Enabling default service esm-apps
{"_schema_version": "0.1", "errors": [], "failed_services": [], "needs_reboot": false, "processed_services": ["esm-apps", "esm-infra"], "result": "success", "warnings": []}

```

orndorffgrant:
Otherwise it looks good!

lucasmoura:
@orndorffgrant good catch, I have updated the code

orndorffgrant:
@lucasmoura Thanks for the change! It looks like it broke one integration test - was the "Enabling default service.." line changed from stderr to stdout? (Thats fine with me, but that might be what caused the test fail)

orndorffgrant (approved):
Looks good! The two jenkins failures are strange. They passed for me when I ran them locally so approving.

lucasmoura:
Merging

PR #1955, opened 2022-02-01
proxy: delete default route on container tests

lucasmoura:
## Proposed Commit Message
proxy: delete default route on container tests

On the container tests, we are now deleting the default route to guarantee that the ua request are actually going through the proxy

## Test Steps
Run the modified proxy test

## Desired commit type

## Checklist

### DIFF
--- a/features/proxy_config.feature
+++ b/features/proxy_config.feature
@@ -24,11 +24,22 @@ Feature: Proxy configuration
           https_proxy: http://<ci-proxy-ip>:3128
         """
         And I verify `/var/log/squid/access.log` is empty on `proxy` machine
+        # We need this for the route command
+        And I run `apt-get install net-tools` with sudo
+        # We will guarantee that the machine will only use the proxy when
+        # running the ua commands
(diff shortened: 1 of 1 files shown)


renanrodrigo (approved):

orndorffgrant (approved):
This looks good but I'm curious as to why we need this: We are already checking the logs of the proxy to make sure it is being used - is that not enough?

lucasmoura:
@orndorffgrant We recently had one issue regarding the calling `get_available_resources` when processing the cli parser here:
https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/cli.py#L1725

As you can see we do that before we set up the proxies on the system. We introduce that call of `get_available_resources` on the parser during the `usg` release of 27.5. One customer using proxies on that version of UA reported that any cli command was failing, because the machine did not have any default routes and I was only using proxies.  Therefore, that endpoint could not be reached and the command failed. Even though we fixed that on 27.6, we are updating that test to avoid similar issues on the future.

PR #1958, opened 2022-02-03
proxy: limited to UA related repos

CalvoM:
## Proposed Commit Message

proxy: limit to UA repos

LP: #1956764 

Limit the proxy set up via UA to UA-related repos, instead of being global.

## Desired commit type

## Checklist


features/proxy_config.feature
```diff
@@ -61,12 +55,12 @@ Feature: Proxy configuration
...
          \*/
-        Acquire::http::Proxy ".*";
-        Acquire::https::Proxy ".*";
+        Acquire::http::Proxy::esm.ubuntu.com ".*";
+        Acquire::https::Proxy::esm.ubuntu.com ".*";
         """
         When I run `apt update` with sudo
         And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        Then stdout matches regexp:
+        Then stdout does not match regexp:
```
lucasmoura, line 63: Let's also check on the `squid` logs if we have `CONNECT esm.ubuntu.com:443` entries on it

lucasmoura (commented):
The change looks good to me, let's just improve the test to check for esm entries directly on the squid log. After that, I think we can merge it

features/proxy_config.feature
CalvoM: Makes sense, let me add it.

lucasmoura (approved):

features/proxy_config.feature
```diff
@@ -124,208 +117,207 @@ Feature: Proxy configuration
...
-    @slow
-    @series.xenial
-    @series.bionic
-    @uses.config.machine_type.lxd.vm
-    Scenario Outline: Attach command when proxy is configured
-        Given a `<release>` machine with ubuntu-advantage-tools installed
-        When I launch a `focal` `proxy` machine
-        And I run `apt install squid -y` `with sudo` on the `proxy` machine
-        And I add this text on `/etc/squid/squid.conf` on `proxy` above `http_access deny all`:
+        @slow
```
orndorffgrant, line 120: It looks like everything below this point got indented an extra 4 spaces, which makes it inconsistent with the first scenario and makes the diff hard to read :/ so we should unindent this.

features/proxy_config.feature
```diff
@@ -13,41 +13,35 @@ Feature: Proxy configuration
...
+        And I run `ua config set https_proxy=http://<ci-proxy-ip>:3128` with sudo
+        Then stdout matches regexp:
         """
-        contract_url: 'https://contracts.canonical.com'
-        data_dir: /var/lib/ubuntu-advantage
-        log_level: debug
-        log_file: /var/log/ubuntu-advantage.log
-        ua_config:
-          http_proxy: http://<ci-proxy-ip>:3128
-          https_proxy: http://<ci-proxy-ip>:3128
```
orndorffgrant, line 24: I appreciate that this change uses the recommended way of setting the proxy now (`ua config set`), but now there is no integration test that tests the proxies are correctly configured after editing the config file and running `ua refresh`.

We can keep these changes, but we may need a new scenario to test setting the proxy via config file and `ua refresh config` also works so that we don't lose this coverage.

features/proxy_config.feature
```diff
@@ -61,13 +55,17 @@ Feature: Proxy configuration
...
         When I run `apt update` with sudo
         And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
         Then stdout matches regexp:
         """
+        CONNECT esm.ubuntu.com:443
+        """
+        Then stdout does not match regexp:
+        """
         .*GET.*ubuntu.com/ubuntu/dists.*
         """
```
orndorffgrant, line 70: This is the part that tests the new changes are actually working. I'd prefer if we had something a bit more specific here: Both `archive.ubuntu.com` and `security.ubuntu.com` I think should be contacted by apt during a `ua attach`. Can we make sure those urls specifically are not present?

orndorffgrant (changes requested):
Looking good! Just some suggestions to keep the tests thorough

features/proxy_config.feature
CalvoM: Yeah, I think adding a new scenario would be helpful to cover this test.

features/proxy_config.feature
CalvoM: Yeah, updated.

renanrodrigo (dismissed):

features/proxy_config.feature
```diff
@@ -335,8 +344,233 @@ Feature: Proxy configuration
...
+        And I run `ua config set apt_https_proxy=http://<ci-proxy-ip>:3128` with sudo
+        When I run `cat /etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy` with sudo
+        Then stdout matches regexp:
+        """
+        /\*
+         \* Autogenerated by ubuntu-advantage-tools
+         \* Do not edit this file directly
+         \*
+         \* To change what ubuntu-advantage-tools sets, run one of the following:
+         \* Substitute "apt_https_proxy" for "apt_http_proxy" as necessary.
```
renanrodrigo, line 485: in all tests involving regex matching, you need to escape the quotes, otherwise it is failing
```suggestion
         \* Substitute \"apt_https_proxy\" for \"apt_http_proxy\" as necessary.
```

renanrodrigo (approved):
worksforme.
Just needs small adjustments in the tests.
Also, make sure to squash the commits before merging (or on merge)

features/proxy_config.feature
```diff
@@ -42,23 +43,17 @@ Feature: Proxy configuration
         esm-infra     +yes      +enabled      +UA Infra: Extended Security Maintenance \(ESM\)
```
orndorffgrant, line 43: This slipped in in a previous PR but would you mind fixing it here?
This line only passes on xenial.
So we need a change like this:
```diff
         And I run `route del default` with sudo
         And I attach `contract_token` with sudo
+        # Just to verify that the machine is attached
+        Then stdout matches regexp:
+        """
+        This machine is now attached
+        """
-        And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
+        When I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
         Then stdout matches regexp:
         """
        .*CONNECT contracts.canonical.com.*
         """
-        When I run `ua status` with sudo
-        # Just to verify that the machine is attached
-        Then stdout matches regexp:
-        """
-        esm-infra     +yes      +enabled      +UA Infra: Extended Security Maintenance \(ESM\)
-        """
```

orndorffgrant: And we'll need the same change down in the new "manual with config" scenario below

orndorffgrant (commented):
I agree with @renanrodrigo 
And also I noticed one unrelated bug that I think we might as well fix in this PR if that's okay :)

features/proxy_config.feature
orndorffgrant: Actually, this may be indicative of an actual bug unrelated to this PR. So don't make this change, we'll need to investigate this.

features/proxy_config.feature
orndorffgrant: Interesting, this suggestion makes sense, but the existing tests were passing without escaping the quotes (and I thought I saw this new one pass as well, I'll re-run it)

features/proxy_config.feature
CalvoM: Okay, feels like I am listening to your thoughts 😂 😂.

features/proxy_config.feature
CalvoM: Okay, let me update all regex matching tests.

renanrodrigo: It fails consistently here for some reason :|

CalvoM: Yeah, updated it since there is no con to adding the backslash.

orndorffgrant:
Hi @CalvoM, unfortunately I think we need to rethink this change.

I realized this change is breaking backwards compatibility and could confuse/break existing customers who are expecting the current global proxy behavior.

I think we need to add to this PR the following functionality:

- Introduce `global_apt_http_proxy` config option that behaves the same as apt proxy does today
- Introduce `ua_apt_http_proxy` config option that behaves the new way only for esm.ubuntu.com
- `apt_http_proxy` will behave the same way it currently does, but will print a warning suggesting to use `global_apt_http_proxy` instead

Let me know if you have any questions

uaclient/cli.py
```diff
@@ -839,21 +856,105 @@ def action_config_set(args, *, cfg, **kwargs):
...
+        elif proxy_scope == "ua":
+            kwargs = {
+                "http_proxy": cfg.ua_apt_http_proxy,
+                "https_proxy": cfg.ua_apt_https_proxy,
+                "proxy_scope": cfg.apt_proxy_scope,
+            }
+
+        key_groups = set_key.split("_")
+        key = "_".join(key_groups[protocol_type_location:])
+        kwargs[key] = set_value
```
orndorffgrant, line 941: I know this is an adaptation of what was already here, but I think this would be easier to read if we just did this:
```suggestion
        if "https" in set_key:
            kwargs["https_proxy"] = set_value
        else:
            kwargs["http_proxy"] = set_value
```

uaclient/cli.py
```diff
@@ -839,21 +856,105 @@ def action_config_set(args, *, cfg, **kwargs):
...
+                        proxy=prev_val
+                    )
+                )
+        elif proxy_scope == "ua":  # check if global-scoped proxy set
+            unset_current = (
+                True
+                if getattr(cfg, "global_apt_http_proxy")
+                or getattr(cfg, "global_apt_https_proxy")
+                else False
+            )
```
orndorffgrant, line 902: This pattern can be simplified a bit to:
```suggestion
            unset_current = bool(cfg.global_apt_http_proxy or cfg.global_apt_https_proxy)
```

uaclient/cli.py
```diff
@@ -824,6 +833,14 @@ def action_config_set(args, *, cfg, **kwargs):
...
             )
         )
+    apt_proxies_config_options = (
+        "apt_http_proxy",
+        "apt_https_proxy",
+        "ua_apt_http_proxy",
+        "ua_apt_https_proxy",
+        "global_apt_http_proxy",
+        "global_apt_https_proxy",
+    )
```
orndorffgrant, line 843: I think the logic might be easier if we don't treat this as one big group. With everything mixed together it is hard to read
I suggest that we instead treat them as separate groups, something like this:

```python
ua_scoped_apt_proxy_options = ("ua_apt_http_proxy", "ua_apt_https_proxy")
global_scoped_apt_proxy_options = ("global_apt_http_proxy", "global_apt_https_proxy")
deprecated_global_scoped_apt_proxy_options = ("apt_http_proxy", "apt_https_proxy")

...


elif set_key in ua_scoped_apt_proxy_options:
    # ua scoped logic
elif (
    set_key in global_scoped_apt_proxy_options or
    set_key in deprecated_global_scoped_apt_proxy_options
):
    if set_key in deprecated_global_scoped_apt_proxy_options:
        # apt_* warnings
        set_key = "global_" + set_key
    # global scoped logic

```

uaclient/config.py
```diff
@@ -1024,11 +1083,55 @@ def process_config(self):
...
+
+        if getattr(self, "global_apt_http_proxy") or getattr(
+            self, "global_apt_https_proxy"
+        ):
+            if getattr(self, "ua_apt_http_proxy") or getattr(
+                self, "ua_apt_https_proxy"
+            ):
+                # Should we unset the config values?
+                raise exceptions.UserFacingError(
+                    "Warning: your configuration is not correct."
```
orndorffgrant, line 1115: Let's make this message more descriptive as well.
```suggestion
                    "Error: Setting global apt proxy and ua scoped apt proxy at the same time is unsupported. Cancelling config process operation."
```

uaclient/apt.py
```diff
@@ -13,8 +13,14 @@
...
-APT_CONFIG_PROXY_HTTP = """Acquire::http::Proxy "{proxy_url}";\n"""
-APT_CONFIG_PROXY_HTTPS = """Acquire::https::Proxy "{proxy_url}";\n"""
+APT_CONFIG_GLOBAL_PROXY_HTTP = """Acquire::http::Proxy "{proxy_url}";\n"""
+APT_CONFIG_GLOBAL_PROXY_HTTPS = """Acquire::https::Proxy "{proxy_url}";\n"""
+APT_CONFIG_PROXY_HTTP = (
+    """Acquire::http::Proxy::esm.ubuntu.com "{proxy_url}";\n"""
+)
+APT_CONFIG_PROXY_HTTPS = (
+    """Acquire::https::Proxy::esm.ubuntu.com "{proxy_url}";\n"""
+)
```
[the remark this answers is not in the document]
CalvoM: Yeah makes sense.

uaclient/config.py
```diff
@@ -1037,7 +1119,12 @@ def process_config(self):
             "https", self.https_proxy, util.PROXY_VALIDATION_SNAP_HTTPS_URL
         )
 
-        apt.setup_apt_proxy(self.apt_http_proxy, self.apt_https_proxy)
+        apt.setup_apt_proxy(
+            self.global_apt_http_proxy, self.global_apt_https_proxy, "global"
+        )
+        apt.setup_apt_proxy(
+            self.ua_apt_http_proxy, self.ua_apt_https_proxy, "ua"
+        )
```
orndorffgrant, line 1127: The second one of these will overwrite the result of the first. We should only call `setup_apt_proxy` once based on what scope is being used in the config file

uaclient/cli.py
```diff
@@ -839,21 +858,96 @@ def action_config_set(args, *, cfg, **kwargs):
...
+                ua_status.MESSAGE_WARNING_APT_PROXY_OVERWRITE.format(
+                    proxy=prev_val
+                )
+            )
+            kwargs = {
+                "http_proxy": None,
+                "https_proxy": None,
+                "proxy_scope": "global",
+            }
+            setup_apt_proxy(**kwargs)
```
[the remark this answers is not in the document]
CalvoM: Oh, that is true.

uaclient/config.py
```diff
@@ -186,6 +190,50 @@ def apt_https_proxy(self, value: str):
...
+    @property
+    def global_apt_https_proxy(self) -> Optional[str]:
+        return self.cfg.get("ua_config", {}).get("global_apt_https_proxy")
+
+    @global_apt_https_proxy.setter
+    def global_apt_https_proxy(self, value: str):
+        if "ua_config" not in self.cfg:
+            self.cfg["ua_config"] = {}
+        self.cfg["ua_config"]["global_apt_https_proxy"] = value
+        self.write_cfg()
```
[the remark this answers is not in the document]
CalvoM: I understand the first point, where we encourage people to use the `global_*` property by giving them a warning if they still use the older `apt_*` property and not `global_*`.
On the second point, do you mean we have the `apt_*` available on the cli options but the uaconfig internals should not have the property? If that is what you mean, then it is a better approach, I see no issue with it. It will still keep maintain cli api for configuring the apt_* proxy and reduce the properties of UAConfig.

uaclient/entitlements/repo.py
```diff
@@ -259,15 +259,37 @@ def setup_apt_config(self, silent: bool = False) -> None:
...
            configuration
         """
-        http_proxy = util.validate_proxy(
-            "http", self.cfg.apt_http_proxy, util.PROXY_VALIDATION_APT_HTTP_URL
-        )
-        https_proxy = util.validate_proxy(
-            "https",
-            self.cfg.apt_https_proxy,
-            util.PROXY_VALIDATION_APT_HTTPS_URL,
+        http_proxy: Optional[str] = None
```
renanrodrigo, line 262: We can't type variables because we need to support Python 3.5. We can remove the typing, or if you want to make it explicit to mypy:
```suggestion
        http_proxy = None  # type: Optional[str]
```

renanrodrigo (changes requested):

features/proxy_config.feature
```diff
@@ -13,23 +13,18 @@ Feature: Proxy configuration
...
-        # We will guarantee that the machine will only use the proxy when
-        # running the ua commands
-        And I run `route del default` with sudo
-        And I attach `contract_token` with sudo and options `--no-auto-enable`
+        When I run `ua config set http_proxy=http://<ci-proxy-ip>:3128` with sudo
+        Then stdout matches regexp:
+        """
+        Setting snap proxy
+        """
+        # TODO No longer empty, needs researching
```
orndorffgrant, line 26: There was a bug in the "verify `file` is empty" step that was fixed in #1979 
You may need to add a line with this
```
        When I run `truncate -s 0 /var/log/squid/access.log` `with sudo` on the `proxy` machine
```
Early in the test to make sure the log starts out empty

features/proxy_config.feature
```diff
@@ -42,23 +37,20 @@ Feature: Proxy configuration
...
         """
-        contract_url: 'https://contracts.canonical.com'
-        data_dir: /var/lib/ubuntu-advantage
-        log_level: debug
-        log_file: /var/log/ubuntu-advantage.log
-        ua_config:
-          apt_http_proxy: http://<ci-proxy-ip>:3128
-          apt_https_proxy: http://<ci-proxy-ip>:3128
+        Setting APT proxy
+        Setting proxy on ua scope
```
orndorffgrant, line 44: I think it would be more clear if this was one line that said `Setting UA-scoped APT proxy`

features/proxy_config.feature
```diff
@@ -292,10 +571,148 @@ Feature: Proxy configuration
...
+        Then stdout matches regexp:
+        """
+        esm-infra     +yes      +disabled      +UA Infra: Extended Security Maintenance \(ESM\)
+        """
+        When I run `truncate -s 0 /var/log/squid/access.log` `with sudo` on the `proxy` machine
+        And I run `ua config set global_apt_https_proxy=http://<ci-proxy-ip>:3128` with sudo
+        Then stdout matches regexp:
+        """
+        Setting APT proxy
+        Setting proxy on global scope
```
orndorffgrant, line 619: Same suggestion as above, I think this should be `Setting global APT proxy`

orndorffgrant (commented):
The implementation looks great, one minor comment there. Now focusing on the integration tests. Everything looks good, my main comment is that we should check some of the edge cases we've talked about in the integration tests as well.

uaclient/entitlements/repo.py
CalvoM: Oh yeah

uaclient/cli.py
```diff
@@ -839,21 +860,77 @@ def action_config_set(args, *, cfg, **kwargs):
...
+        apt_equivalent = ""
+        if "https" in set_key:
+            kwargs["https_proxy"] = set_value
+            apt_equivalent = "apt_https_proxy"
+        else:
+            kwargs["http_proxy"] = set_value
+            apt_equivalent = "apt_http_proxy"
+
         setup_apt_proxy(**kwargs)
+        setattr(cfg, apt_equivalent, None)  # unset the apt_* equivalent
```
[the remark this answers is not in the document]
CalvoM: I will have to scrape all these `apt_*` attributes.

features/proxy_config.feature
CalvoM: Yeah, It kept failing here, I thought it was my environment setup.

CalvoM: The PR you referenced only shows the inclusion of options --no-auto-enable. Is that needed as well?

uaclient/cli.py
```diff
@@ -839,21 +860,77 @@ def action_config_set(args, *, cfg, **kwargs):
...
+        cfg.global_apt_https_proxy = None
+
+    elif set_key in (
+        deprecated_global_scoped_proxy_options + global_scoped_proxy_options
+    ):
         # setup_apt_proxy is destructive for unprovided values. Source complete
         # current config values from uaclient.conf before applying set_value.
-        protocol_type = set_key.split("_")[1]
+        if set_key in deprecated_global_scoped_proxy_options:
+            event.info(ua_status.MESSAGE_WARNING_APT_PROXY_SETUP)
```
lucasmoura, line 899: Same message consideration

uaclient/cli.py
```diff
@@ -896,6 +973,15 @@ def action_config_unset(args, *, cfg, **kwargs):
...
         )
+    ua_scoped_proxy_options = ("ua_apt_http_proxy", "ua_apt_https_proxy")
+    global_scoped_proxy_options = (
+        "global_apt_http_proxy",
+        "global_apt_https_proxy",
+    )
+    deprecated_global_scoped_proxy_options = (
+        "apt_http_proxy",
+        "apt_https_proxy",
+    )
```
lucasmoura, line 984: We are defining those constant in two different functions, I think we can move them somewhere else, maybe `config` or `defaults`

uaclient/config.py
```diff
@@ -165,25 +169,73 @@ def https_proxy(self, value: str):
...
+
+    @property
+    def global_apt_http_proxy(self) -> Optional[str]:
+        global_val = self.cfg.get("ua_config", {}).get("global_apt_http_proxy")
+        if global_val:
+            return global_val
+
+        old_apt_val = self.cfg.get("ua_config", {}).get("apt_http_proxy")
+        if old_apt_val:
+            event.info(
```
lucasmoura, line 201: Let's move that message into the `message` modules.

Also, I think this is a worth `event.warning` message for the json response case. What do you think @renanrodrigo @orndorffgrant ?

uaclient/config.py
```diff
@@ -165,25 +169,73 @@ def https_proxy(self, value: str):
...
+        )
+        if global_val:
+            return global_val
+
+        old_apt_val = self.cfg.get("ua_config", {}).get("apt_https_proxy")
+        if old_apt_val:
+            event.info(
+                'Using deprecated "apt_https_proxy" config field.'
+                'Please migrate to using "global_apt_https_proxy"'
+            )
```
lucasmoura, line 229: Same message and warning consideration here

uaclient/config.py
```diff
@@ -1024,11 +1076,33 @@ def process_config(self):
...
                 raise exceptions.UserFacingError(error_msg)
 
+        if (self.global_apt_http_proxy or self.global_apt_https_proxy) and (
+            self.ua_apt_http_proxy or self.ua_apt_https_proxy
+        ):
+            # Should we unset the config values?
+            raise exceptions.UserFacingError(
+                "Error: Setting global apt proxy and ua scoped apt proxy at"
+                " the same time is unsupported."
+                " Cancelling config process operation."
```
lucasmoura, line 1086: Let's either create a special Exception here or move the message to the `messages` module

uaclient/status.py
```diff
@@ -620,6 +620,18 @@ def __init__(self, name: str, msg: str):
...
+Warning: Please consider using global_apt_http_proxy/global_apt_https_proxy
+instead of apt_http_proxy/apt_https_proxy
+when configuring a global apt proxy
+
+We will set the global_apt_http_proxy/global_apt_https_proxy.
+"""
+MESSAGE_WARNING_APT_PROXY_OVERWRITE = """\
+Warning: Setting the {current_proxy} proxy will overwrite the {previous_proxy}
+proxy previously set via `ua config`.
+"""
```
lucasmoura, line 634: Let's move those messages to the `message` module

features/proxy_config.feature
orndorffgrant: @CalvoM I shouldve been more specific. This line is what I was referring to that affects the `I verify file is empty` step https://github.com/canonical/ubuntu-advantage-client/pull/1979/files#diff-cf027576a6812e3403bd3d9579d83abfaa74a41607a4f3c986398ef8492a5d3dR196

I'm not sure if you need `--no-auto-enable` here or now. If it works without it, then great :)

uaclient/config.py
orndorffgrant: I agree

features/proxy_config.feature
CalvoM: Oh yeah. However, when setting the proxy via the cli will have some entries in the access.log file since `validate_proxy` makes some requests via the proxy. So, instead of testing if it empty, we could test that the test urls are in the log file.

uaclient/apt.py
```diff
@@ -404,16 +412,34 @@ def setup_apt_proxy(
     :param https_proxy: the url of the https proxy apt should use, or None
     :return: None
     """
+    scope = proxy_scope if proxy_scope in ("global", "ua") else "global"
     if http_proxy or https_proxy:
         event.info(status.MESSAGE_SETTING_SERVICE_PROXY.format(service="APT"))
+        if proxy_scope:
+            event.info(
+                status.MESSAGE_SETTING_SERVICE_PROXY_SCOPE.format(scope=scope)
```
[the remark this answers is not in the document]
CalvoM: Sure, I had tried rebasing and noticed the differences.

uaclient/config.py
renanrodrigo: +1

uaclient/cli.py
```diff
@@ -785,14 +785,25 @@ def action_config_show(args, *, cfg, **kwargs):
...
     row_tmpl = "{key: <" + col_width + "} {value}"
+
     for key in config.UA_CONFIGURABLE_KEYS:
-        print(row_tmpl.format(key=key, value=getattr(cfg, key)))
+        print(row_tmpl.format(key=key, value=getattr(cfg, key, None)))
+
+    if (cfg.global_apt_http_proxy or cfg.global_apt_https_proxy) and (
+        cfg.ua_apt_http_proxy or cfg.ua_apt_https_proxy
+    ):
+        event.info(
```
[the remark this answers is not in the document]
lucasmoura: Yes, we can make the full migration of the other cli commands to use the event logger in a separate PR

uaclient/apt.py
```diff
@@ -404,16 +412,34 @@ def setup_apt_proxy(
     :param https_proxy: the url of the https proxy apt should use, or None
     :return: None
     """
+    scope = proxy_scope if proxy_scope in ("global", "ua") else "global"
```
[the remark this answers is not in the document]
CalvoM: Could I add this as a separate PR, to reduce more additions and changes to the PR.

uaclient/apt.py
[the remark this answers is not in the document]
CalvoM: Done.

features/steps/steps.py
```diff
@@ -523,7 +523,10 @@ def then_stream_does_not_match_regexp(context, stream):
 @then("{stream} matches regexp")
 def then_stream_matches_regexp(context, stream):
     content = getattr(context.process, stream).strip()
-    assert_that(content, matches_regexp(context.text))
+    text = context.text
+    if "<ci-proxy-ip>" in text and "proxy" in context.instances:
```
CalvoM, line 527: @orndorffgrant @renanrodrigo @lucasmoura Is this okay or should I use regex here https://github.com/CalvoM/ubuntu-advantage-client/blob/eda4b48b5f3adf19cf81e1713aeda09cfaf9f18c/features/proxy_config.feature#L990?

uaclient/cli.py
CalvoM: I think `config` is better, I see that `default` is a bit global.

uaclient/apt.py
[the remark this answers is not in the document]
CalvoM: It seems I was able to add this.

uaclient/defaults.py
```diff
@@ -43,3 +43,9 @@
     "ua_log_level",
     "ua_security_url",
 ]
+ua_scoped_proxy_options = ("ua_apt_http_proxy", "ua_apt_https_proxy")
+global_scoped_proxy_options = (
+    "global_apt_http_proxy",
+    "global_apt_https_proxy",
+)
+deprecated_global_scoped_proxy_options = ("apt_http_proxy", "apt_https_proxy")
```
orndorffgrant, line 51: I don't think we need these because they were added to UAConfig

features/steps/steps.py
orndorffgrant: I think this is good as it is

features/proxy_config.feature
```diff
@@ -72,47 +84,54 @@ Feature: Proxy configuration
          \*   sudo ua config set apt_http_proxy=<value>
          \*   sudo ua config unset apt_http_proxy
```
orndorffgrant, line 85: We need to update this message to something like:
```
        /\*
         \* Autogenerated by ubuntu-advantage-tools
         \* Do not edit this file directly
         \*
         \* To change what ubuntu-advantage-tools sets, use the `ua config set` command
         \* command to set either:
         \*     global_apt_http_proxy and global_apt_https_proxy
         \* for a global apt proxy.
         \* or
         \*     ua_apt_http_proxy and ua_apt_https_proxy
         \* for a apt proxy that only applies to UA related repos.
         \*\/
```

uaclient/apt.py
```diff
@@ -455,16 +470,39 @@ def setup_apt_proxy(
     :param https_proxy: the url of the https proxy apt should use, or None
     :return: None
     """
+    scope = proxy_scope if proxy_scope else AptProxyScope.GLOBAL
     if http_proxy or https_proxy:
         event.info(messages.SETTING_SERVICE_PROXY.format(service="APT"))
```
orndorffgrant, line 475: I think we can remove this line since we have the scope specific line.
```suggestion
```

orndorffgrant (commented):
This is very close! Just a couple small things and this should be good to go IMO

features/proxy_config.feature
```diff
@@ -292,10 +571,148 @@ Feature: Proxy configuration
            | focal   |
 
     @slow
-    @series.xenial
-    @series.bionic
-    @uses.config.machine_type.lxd.vm
-    Scenario Outline: Attach command when authenticated proxy is configured
+    @series.lts
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attach command when proxy is configured globally
```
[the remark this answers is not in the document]
lucasmoura: Maybe one scenario we could add is to config any `ua_apt_proxy` and then to go directly into `uaclient.conf` and setting up ANY `global_apt_proxy` or `apt_proxy` will fail when we run `ua refresh` or try to run any other ua command direclty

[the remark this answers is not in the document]
lucasmoura: +1

uaclient/cli.py
```diff
@@ -834,21 +845,74 @@ def action_config_set(args, *, cfg, **kwargs):
...
+            "http_proxy": cfg.ua_apt_http_proxy,
+            "https_proxy": cfg.ua_apt_https_proxy,
+            "proxy_scope": AptProxyScope.UACLIENT,
+        }
+        if "https" in set_key:
+            kwargs["https_proxy"] = set_value
+        else:
+            kwargs["http_proxy"] = set_value
+
+        setup_apt_proxy(**kwargs)
```
lucasmoura, line 874: I think this part of the code looks really similar to the one we have for the `global` scenario. I think we can create a function for it

uaclient/defaults.py
CalvoM: Yeah. On it.

features/proxy_config.feature
CalvoM: I could you this as it is , I will just add the unset command option as well

uaclient/apt.py
CalvoM: I see, yeah, the scoped message is complete in info.

uaclient/apt.py
```diff
@@ -442,7 +455,9 @@ def get_installed_packages() -> List[str]:
 
 
 def setup_apt_proxy(
-    http_proxy: Optional[str] = None, https_proxy: Optional[str] = None
+    http_proxy: Optional[str] = None,
+    https_proxy: Optional[str] = None,
+    proxy_scope: Optional[AptProxyScope] = None,
```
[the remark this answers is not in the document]
CalvoM: Makes sense.

renanrodrigo:
There is a complain from `mypy` on the decorated property...
There is nothing wrong with it, mypy themselves acknowledge it and are fixing it somehow, as seen in https://github.com/python/mypy/issues/1362, so I'd suggest `# type: ignore`ing those so we pass the type checks.

uaclient/config.py
```diff
@@ -100,6 +104,14 @@
 
 event = event_logger.get_event_logger()
 
+# needed for solving mypy errors dealing with _lru_cache_wrapper
+# Found at https://github.com/python/mypy/issues/5858#issuecomment-454144705
+S = TypeVar("S", bound=str)
+
+
+def str_cache(func: Callable[..., S]) -> S:
```
CalvoM, line 112: @lucasmoura @renanrodrigo  I got this workaround from [here](https://github.com/python/mypy/issues/5858). Mypy and lru_cache do not play well.

uaclient/cli.py
```diff
@@ -1579,6 +1631,37 @@ def action_refresh(args, *, cfg: config.UAConfig):
...
+) -> None:
+    """
+    Handles setting part the apt proxies - global and uaclient scoped proxies
+    """
+    # needs to import from within here or else unit tests fail
+    # due to wrong importation path issues i.e
+    # tests will use cli.setup_apt_proxy over apt.setup_apt_proxy
+
+    set_value = args.get("set_value")
+    set_key = args.get("set_key")
```
renanrodrigo, line 1645: Mypy complains that this value cannot be `None`, because you are using the `in` operator below
```suggestion
    set_value = args.get("set_value", "")
    set_key = args.get("set_key", "")
```

uaclient/cli.py
```diff
@@ -1579,6 +1631,37 @@ def action_refresh(args, *, cfg: config.UAConfig):
...
+        kwargs = {
+            "http_proxy": cfg.ua_apt_http_proxy,
+            "https_proxy": cfg.ua_apt_https_proxy,
+            "proxy_scope": scope,
+        }
+    if "https" in set_key:
+        kwargs["https_proxy"] = set_value
+    else:
+        kwargs["http_proxy"] = set_value
+    setup_apt_proxy(**kwargs)
```
renanrodrigo, line 1662: Making the variables explicit instead of passing a `kwargs` object here makes it more readable, and Mypy stops complaining as well 

```suggestion
    if scope == AptProxyScope.GLOBAL:
        http_proxy = cfg.global_apt_http_proxy
        https_proxy = cfg.global_apt_https_proxy
    elif scope == AptProxyScope.UACLIENT:
        http_proxy = cfg.ua_apt_http_proxy
        https_proxy = cfg.ua_apt_https_proxy
    if "https" in set_key:
        https_proxy = set_value
    else:
        http_proxy = set_value
    setup_apt_proxy(http_proxy, https_proxy, scope)
```

renanrodrigo (changes requested):

features/proxy_config.feature
```diff
@@ -311,12 +362,8 @@ Feature: Proxy configuration
         And I run `systemctl restart squid.service` `with sudo` on the `proxy` machine
         And I run `ua config set http_proxy=http://someuser:somepassword@<ci-proxy-ip>:3128` with sudo
         And I run `ua config set https_proxy=http://someuser:somepassword@<ci-proxy-ip>:3128` with sudo
-        And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        Then stdout matches regexp:
-        """
-        .*CONNECT api.snapcraft.io:443.*
-        """
-        When I attach `contract_token` with sudo
+        And I verify `/var/log/squid/access.log` is empty on `proxy` machine
```
renanrodrigo, line 365: This step is failing here... don't we have to truncate the file somewhere?

renanrodrigo (commented):
Integration tests are failing for me

uaclient/cli.py
CalvoM: Yeah, good solution.

features/proxy_config.feature
CalvoM: Yes, It should be before setting the proxies

orndorffgrant (approved):
Excellent work here @CalvoM - thanks for figuring out all the edge cases!

lucasmoura (approved):
Just some minor nits on the new function you added, but otherwise LGTM.
Great work on that task @CalvoM

uaclient/cli.py
```diff
@@ -1571,6 +1623,33 @@ def action_refresh(args, *, cfg: config.UAConfig):
...
+) -> None:
+    """
+    Handles setting part the apt proxies - global and uaclient scoped proxies
+    """
+    # needs to import from within here or else unit tests fail
+    # due to wrong importation path issues i.e
+    # tests will use cli.setup_apt_proxy over apt.setup_apt_proxy
+
+    set_value = args.get("set_value", "")
+    set_key = args.get("set_key", "")
```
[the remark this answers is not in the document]
CalvoM: @lucasmoura updated

renanrodrigo (approved):
LGTM

PR #1968, opened 2022-02-04
status: return errors as JSON/YAML when required (SC-740)

renanrodrigo:
Add the event_logger to the status action, making sure all exceptions are captured as errors.
Added specific errors for the simulated status, when an invalid token is passed.
Added specific w̶a̶r̶n̶i̶n̶g̶s̶ errors for the simulated status, when the token is expired or not valid yet.

## Desired commit type

## Checklist


features/schemas/ua_status.json
```diff
@@ -0,0 +1,246 @@
+{
+  "type": "object",
+  "properties": {
+    "_doc": {
+      "type": "string"
+    },
+    "_schema_version": {
+      "type": "string"
```
orndorffgrant, line 8: I discovered "const" recently so we can do this for the schema version
```suggestion
      "type": "string",
      "const": "0.1"
```

features/attached_status.feature
```diff
@@ -7,18 +7,13 @@ Feature: Attached status
...
-            """
+        Then stdout is a json matching the `ua_status` schema
         When I run `ua status --format yaml` as non-root
         Then stdout is formatted as `yaml` and has keys:
             """
             _doc _schema_version account attached config config_path contract effective
             environment_vars execution_details execution_status expires machine_id notices
-            services version simulated
+            services version simulated result errors warnings
             """
```
orndorffgrant, line 17: Not necessary for this PR, but we could probably write a step function that reads like:
```
Then stdout is yaml matching the `ua_status` schema
```

uaclient/config.py
```diff
@@ -828,10 +836,33 @@ def simulate_status(
             }
         )
 
+        now = datetime.now(timezone.utc)
         if contract_info.get("effectiveTo"):
             response["expires"] = contract_info.get("effectiveTo")
+            expiration_datetime = util.parse_rfc3339_date(response["expires"])
```
orndorffgrant, line 842: TIL we aren't using DatetimeAwareJsonDecoder in util.readurl! We probably should so that we don't need to parse later on like here, but that's probably a separate PR

uaclient/event_logger.py
```diff
@@ -46,6 +49,9 @@ def __init__(self):
         self._processed_services = set()  # type: Set[str]
         self._failed_services = set()  # type: Set[str]
         self._needs_reboot = False
+        self._command = ""
+        self._output_format = "json"
```
orndorffgrant, line 53: Rather than add a new `_output_format` field, I suggest changing the `EventLoggerMode` enum to have different values: i.e. `CLI`, `JSON`, and `YAML`

uaclient/event_logger.py
```diff
@@ -145,6 +178,56 @@ def _generate_failed_services(self):
...
+        if not status.get("simulated"):
+            available_services = [
+                service
+                for service in status.get("services", [])
+                if service.get("available", "yes") == "yes"
+            ]
+            status["services"] = available_services
+
+        # We don't need the origin info in the json output
+        status.pop("origin", "")
```
orndorffgrant, line 212: This seems like it belongs over in the `action.status` function

features/attached_status.feature
renanrodrigo: Can be done in this PR, good call. We can then get rid of this `is formatted as` verification.

features/schemas/ua_status.json
```diff
@@ -0,0 +1,246 @@
+{
```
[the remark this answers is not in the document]
renanrodrigo: (:

uaclient/config.py
renanrodrigo: Let's create a card/issue for this one - must be done with care

uaclient/event_logger.py
renanrodrigo: Missed this after the rebase - will change

uaclient/event_logger.py
renanrodrigo: After playing for it a bit, I would like to retreat from this change. The EventLoggerMode is responsible for determining a change in the behavior, while the format is just an output detail in my point of view. I understand how we have the `--format` option to the commands, but reflecting the same idea to the class led to some confusion imho.

uaclient/event_logger.py
renanrodrigo: Ended up moving it back to `status.py` - the CLI formatting function is also there, and given it is called conditionally based on the event logger mode, it would bloat the action

features/attached_status.feature
renanrodrigo: Created SC-797 for this one

uaclient/event_logger.py
```diff
@@ -145,6 +177,34 @@ def _generate_failed_services(self):
...
+            "processed_services": sorted(self._processed_services),
+            "failed_services": sorted(self._generate_failed_services()),
+            "errors": self._error_events,
+            "warnings": self._warning_events,
+            "needs_reboot": self._needs_reboot,
+        }
+
+        print(json.dumps(response, sort_keys=True))
+
+    def _process_events_status(self):
```
lucasmoura, line 193: Instead of having different methods to handle that, I think we should have classes to support that. This will make `process_events` cleaner

uaclient/event_logger.py
```diff
@@ -145,6 +177,34 @@ def _generate_failed_services(self):
...
+        output = format_machine_readable_output(self._output_content)
```
lucasmoura, line 194: Also, maybe we can add an optional parameter to `process_events` called `output_content` where if it set, we output it directly instead of parsing the response as we do.

uaclient/event_logger.py
renanrodrigo: I see that we will have one less property in the class, but the logic to process events will be practically the same. I personally think it is more readable to have a mode (plain text on CLI x machine-readable output), and then defining the format based on what is in the CLI. But I can change this, no problem.

uaclient/cli.py
```diff
@@ -1508,7 +1509,9 @@ def get_parser():
     return parser
 
 
+@set_event_mode
 def action_status(args, *, cfg):
+    event.set_command("status")
```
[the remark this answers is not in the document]
renanrodrigo: The decorator which sets it to machine-readable mode is applied only when the action function is called, so if any error happens before that it will be printed out as CLI (non structured, plain text) output - this applies to the other commands as well

uaclient/event_logger.py
```diff
@@ -67,15 +75,39 @@ def set_event_mode(self, event_mode: EventLoggerMode):
...
+        """
+        self._output_format = output_format
+
+    def set_output_content(self, output_content: Dict):
+        """Set the event logger output content.
+
+        The command will tell the process_events method which content
+        to use.
+        """
+        self._output_content = output_content
```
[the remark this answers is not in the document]
renanrodrigo: Why should it be a property?

uaclient/cli.py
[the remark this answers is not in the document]
lucasmoura: Yes, but if we have an exception, like when setting up the proxy for example, shouldn't we return this as a json response in that situation ?

uaclient/event_logger.py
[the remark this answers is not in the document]
lucasmoura: No major reason, I think it just more pythonic to define it as  property `setter`

uaclient/event_logger.py
renanrodrigo: As discussed with the team, this will be refactored in the future

uaclient/event_logger.py
renanrodrigo: As discussed with the team, this will be refactored in the future

uaclient/event_logger.py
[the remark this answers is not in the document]
renanrodrigo: subject to refactor as well when we separate the loggers

uaclient/cli.py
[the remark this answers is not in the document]
renanrodrigo: I understand how it can fail there, and although it is pretty much an unexpected case, yes, I agree that we should put out JSON if JSON was requested no matter what.
Then, we need to set the event logger mode on main as well, instead of relying on the decorator to do it. We will already know the command and the requested format.
I will move the logic there to make sure it goes well.

features/attached_status.feature
renanrodrigo: @orndorffgrant I actually found a feasible way to do this! So we don't need another PR - it is in a separate commit to facilitate reviewing/changing/removing as needed.

uaclient/event_logger.py
renanrodrigo: For the simplest way, I ended up implementing the original request. I think it can be like this for now, and the overall logic can be improved when we make a decision on how the overall JSON responses should work.

uaclient/cli.py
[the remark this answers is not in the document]
renanrodrigo: @lucasmoura The earliest that we can do it is just after parsing the args.  The change is in a separate commit to facilitate reviewing/changing/removing as needed.

features/util.py
```diff
@@ -325,3 +325,13 @@ def emit_spinner():
     finally:
         print()
         dot_process.terminate()
+
+
+class SafeLoaderWithoutDatetime(yaml.SafeLoader):
+    pass
+
+
+SafeLoaderWithoutDatetime.yaml_implicit_resolvers = {
```
lucasmoura, line 334: can't we define this as a class property ?

uaclient/event_logger.py
```diff
@@ -145,23 +166,48 @@ def _generate_failed_services(self):
...
+        print(json.dumps(response, sort_keys=True))
+
+    def _process_events_status(self):
+        output = format_machine_readable_output(self._output_content)
+        output["result"] = "success" if not self._error_events else "failure"
+        output["errors"] = self._error_events
+        output["warnings"] = self._warning_events
+
+        if self._event_logger_mode == EventLoggerMode.JSON:
+            from uaclient.util import DatetimeAwareJSONEncoder
```
lucasmoura, line 189: Can't we import this on beginning of the file ?

uaclient/event_logger.py
```diff
@@ -145,23 +166,48 @@ def _generate_failed_services(self):
...
+
+            print(
+                json.dumps(
+                    output, cls=DatetimeAwareJSONEncoder, sort_keys=True
+                )
+            )
+        elif self._event_logger_mode == EventLoggerMode.YAML:
+            import yaml
```
lucasmoura, line 197: Same question here

uaclient/tests/test_cli_status.py
```diff
@@ -818,6 +853,122 @@ def test_unicode_dash_replacement_when_unprintable(
...
+
+        m_get_contract_information.side_effect = exception_to_throw
+
+        cfg = FakeConfig()
+
+        args = mock.MagicMock(
+            format="json", all=False, simulate_with_token="some_token"
+        )
+
+        with pytest.raises(Exception) as exc:
```
lucasmoura, line 897: Should we use `exception_type` here for the `raises` parameter ?

uaclient/tests/test_cli_status.py
```diff
@@ -818,6 +853,122 @@ def test_unicode_dash_replacement_when_unprintable(
...
+        def contract_info_side_effect(cfg, token):
+            response = copy.deepcopy(RESPONSE_CONTRACT_INFO)
+            if token == "expired_token":
+                response["contractInfo"][
+                    "effectiveTo"
+                ] = "2019-12-31T00:00:00Z"
+            if token == "token_not_valid_yet":
+                response["contractInfo"][
+                    "effectiveFrom"
+                ] = "9999-12-31T00:00:00Z"
```
lucasmoura, line 949: I think we can set those variables using `parametrize`, like `contract_field` and `data_value`

uaclient/tests/test_cli_status.py
```diff
@@ -818,6 +853,122 @@ def test_unicode_dash_replacement_when_unprintable(
...
+            event, "_event_logger_mode", event_logger_mode
+        ), mock.patch.object(event, "_command", "status"):
+            assert 0 == action_status(args, cfg=cfg)
+
+        if format_type == "json":
+            output = json.loads(capsys.readouterr()[0])
+        else:
+            output = yaml.safe_load(capsys.readouterr()[0])
+
+        assert output["warnings"][0]["message"] == warning_message
```
lucasmoura, line 970: Can we test the whole output here ? Just to double check that the full json/yaml format is being returned correctly

uaclient/config.py
```diff
@@ -828,10 +836,33 @@ def simulate_status(
...
+            delta = expiration_datetime - now
+            if delta.total_seconds() <= 0:
+                event.warning(
```
lucasmoura, line 845: This will cause the command to fail right ? If yes, I think it should be an error instead

features/util.py
renanrodrigo: Done!

uaclient/tests/test_cli_status.py
renanrodrigo: done

uaclient/tests/test_cli_status.py
renanrodrigo: done

uaclient/event_logger.py
renanrodrigo: Unfortunately no - util has circular imports :/

uaclient/event_logger.py
renanrodrigo: Here the error is even weirder - but no, it does not import correctly :/

uaclient/config.py
renanrodrigo: Actually it does not fail - as this is just getting the token information, it returns everything even if it is expired - (unlike when attaching, when trying an expired token would error out)

uaclient/tests/test_cli_status.py
renanrodrigo: Well the output is covered by other (unit and integration) tests. This one targets this key specifically - the proper way would even be to make sure the event was created instead of checking the output - we might want to change it along with many other tests

uaclient/event_logger.py
lucasmoura: Okay, that one is odd, what is the error that happens ?

renanrodrigo: It fails to import yaml even before running any test. I found it weird because yaml is imported in many other places. I tried to debug this but I think I don't know too deep about the import machinery :(

I am pasting a full log here:
```
==================================================================== log start =====================================================================
Processing ./.tox/.tmp/package/2/ubuntu-advantage-tools-20.3.zip
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /home/renan/work/ubuntu-advantage-client/.tox/py3/bin/python /home/renan/work/ubuntu-advantage-client/.tox/py3/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py get_requires_for_build_wheel /tmp/tmp3jrhjff_
       cwd: /tmp/pip-req-build-zj18i51m
  Complete output (22 lines):
    File "/home/renan/work/ubuntu-advantage-client/.tox/py3/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
      main()
    File "/home/renan/work/ubuntu-advantage-client/.tox/py3/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/renan/work/ubuntu-advantage-client/.tox/py3/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 130, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-9hc7s_9c/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 177, in get_requires_for_build_wheel
      return self._get_build_requires(
    File "/tmp/pip-build-env-9hc7s_9c/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 159, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-9hc7s_9c/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 281, in run_setup
      super(_BuildMetaLegacyBackend,
    File "/tmp/pip-build-env-9hc7s_9c/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 174, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 8, in <module>
      from uaclient import defaults, util, version
    File "/tmp/pip-req-build-zj18i51m/uaclient/util.py", line 29, in <module>
      from uaclient import event_logger, exceptions, status
    File "/tmp/pip-req-build-zj18i51m/uaclient/event_logger.py", line 12, in <module>
      import yaml
  ModuleNotFoundError: No module named 'yaml'
  ----------------------------------------

```

renanrodrigo: Second solution seems way better - I'll check. Thanks for taking a look into this.

renanrodrigo: So, even if we remove the `util` import from `setup.py`, it imports `version`, which in its turn imports `util` (and needs it).

I think that between defining `pyyaml` as a build-time requirement and importing it as it is, I prefer to leave it as is.
Please let me know otherwise.

uaclient/config.py
renanrodrigo: As discussed, we will present this as an error. The overall information about a token will still be there, but there will be an error pointing out that the token out of its valid dates.

renanrodrigo: Done

features/unattached_status.feature
```diff
@@ -128,45 +152,99 @@ Feature: Unattached status
...
+            errors:
+            - message: Invalid token. See https://ubuntu.com/advantage
+              service: null
+              type: system
+            result: failure
+            services: []
+            warnings: []
+            """
+        When I run `sed -i 's/contracts.can/contracts.staging.can/' /etc/ubuntu-advantage/uaclient.conf` with sudo
+        And I do a preflight check for `contract_token_staging_expired` formatted as json
```
lucasmoura, line 201: I checked that the command here is still returning `0`. I think it should return `1` instead. What do you think @orndorffgrant @CalvoM ?

lucasmoura (commented):
@renanrodrigo I think we should still have that update on `setup.py` right ? To allow us to import `yaml` on the beginning of the `event_logger` module

lucasmoura:
@renanrodrigo just saw your comment about the `version` issue on setup. let's leave this as is, but I think we can remove that `git` command call from `get_version`, but I will double check that

features/unattached_status.feature
orndorffgrant: I agree that this should return non-zero

renanrodrigo:
@lucasmoura @orndorffgrant all done

uaclient/config.py
```diff
@@ -792,17 +793,25 @@ def _get_entitlement_information(
 
     def simulate_status(
         self, token: str, show_beta: bool = False
-    ) -> Dict[str, Any]:
+    ) -> Tuple[Dict[str, Any], int]:
         """Return a status dictionary based on a token."""
```
lucasmoura, line 797: Let's just update the docstring for the new return.
Also, instead of returning the tuple directly, we could use a `NamedTuple` here to make the code easier to read on cli.
But that is just an style change, feel free to disagree

uaclient/tests/test_event_logger.py
```diff
@@ -4,14 +4,15 @@
...
 from uaclient.event_logger import JSON_SCHEMA_VERSION, EventLoggerMode
 
 
 class TestEventLogger:
     @pytest.mark.parametrize(
         "event_mode",
-        ((EventLoggerMode.CLI), (EventLoggerMode.MACHINE_READABLE)),
+        ((EventLoggerMode.CLI), (EventLoggerMode.JSON), EventLoggerMode.YAML),
     )
     def test_process_events(self, event_mode, event):
```
lucasmoura, line 17: Should we add a test here were we set the command as `status`  and give an `output` to the `set_output_status` ?

lucasmoura (commented):
Just a couple of nits and it should be good to go for me

orndorffgrant (approved):
This looks great! Thanks for everything that went into this one!

uaclient/tests/test_event_logger.py
renanrodrigo: added

uaclient/config.py
renanrodrigo: I changed the docstring to explain why the int and what it means - then I think it is explanatory enough

renanrodrigo:
@lucasmoura addressed

lucasmoura (approved):

PR #1970, opened 2022-02-08
bump pycloudlib version

lucasmoura:
## Proposed Commit Message
bump pycloudlib version

Updating pycloudlib version used in the integration tests. The main change here is that pycloudlib now supports a credentials
file with information about all clouds that it supports. We are leveraging that now on our integration tests

## Test Steps
Create the `pycloudlib.toml` file with all of the necessary cloud credentials.
The template can be found [here](https://github.com/canonical/pycloudlib/blob/main/pycloudlib.toml.template)
After that, run any of our tests using using the `UACLIENT_BEHAVE_CLOUD_CREDENTIALS_PATH` with the path of the that toml file as a value. This should be the only cloud variable we need now for our integration tests

## Desired commit type

## Checklist


features/environment.py
```diff
@@ -436,16 +399,11 @@ def _should_skip_tags(context: Context, tags: List) -> str:
...
-                                ),
-                                *cloud_manager.format_missing_env_vars(
-                                    cloud_manager.missing_env_vars()
-                                ),
-                            )
+                    cloud_creds_path = context.config.cloud_credentials_path
+                    if cloud_manager and not cloud_creds_path:
+                        return "Skipped: {} machine_type requires:\n{}".format(
+                            machine_type,
+                            "UACLIENT_BEHAVE_CLOUD_CREDENTIALS_PATH",
```
orndorffgrant, line 406: pycloudlib checks a couple places for the pycloudlib.toml file so this envvar isn't totally necessary (see https://github.com/canonical/pycloudlib/blob/main/pycloudlib/config.py#L36-L42).

So I think this should only print a warning instead of preventing the test from running

orndorffgrant:
This is something I just ran into. The awspro tests don't seem to work unless the ec2 region is us-east-2. I guess AMI Ids are different per region? And I guess our `refresh-aws-pro-ids` script was last run againsg us-east-2?

We may want to document that, in this PR since after these changes the code doesn't default to us-east-2 anymore

renanrodrigo:
@orndorffgrant yes - AMIs are different for each region

features/environment.py
lucasmoura: That is true, but the reason I am adding this variable is because we might need a different set of credentials to the ones we currently have on the machine. But I agree that if we don't need that, we can just drop it

features/environment.py
```diff
@@ -273,7 +240,7 @@ def __init__(
         timed_job_tag += str(job_suffix)
         timed_job_tag = timed_job_tag.replace(".", "-")
 
-        # Add 8-digit random suffix to the tag, so it does not conflict when
+        # Add 7-digit random suffix to the tag, so it does not conflict when
```
renanrodrigo, line 243: Isn't this 8? `k=8` below should make it 8

renanrodrigo (approved):
LGTM
Just need to solve the regions problem (or document which ones we use, so the credentials will match)

features/environment.py
orndorffgrant: (Mentioned this earlier but putting it here for the public record): I like the environment variable as an override, I just think that it shouldn't be required. Right now the test won't run if the envvar isn't set. Ideally I could choose between:
1. set the envvar to a specific toml file
2. don't set the envvar so that pycloudlib finds my normal toml file

integration-requirements.txt
```diff
@@ -2,7 +2,8 @@
 behave
 jsonschema
 PyHamcrest
-pycloudlib @ git+https://github.com/canonical/pycloudlib.git@756a2c2de044ca60eaa7cdc76653d23a1339dc0a
+pycloudlib @ git+https://github.com/canonical/pycloudlib.git@d8c7654dbef1a7daf47417337e50515e6842aed6
```
blackboxsw, line 5: This needs to be minimally 44206bb95c49901d994c9eb772eba07f2a1b6661 to get Azure Jammy image support

Jenkinsfile
```diff
@@ -7,18 +7,7 @@ pipeline {
...
-        )
-        UACLIENT_BEHAVE_AZ_CLIENT_ID = credentials('ua-azure-client-id')
-        UACLIENT_BEHAVE_AZ_CLIENT_SECRET = credentials(
-            'ua-azure-client-secret'
-        )
-        UACLIENT_BEHAVE_AZ_TENANT_ID = credentials('ua-azure-tenant')
-        UACLIENT_BEHAVE_AZ_SUBSCRIPTION_ID = credentials(
-            'ua-azure-subscription-id'
-        )
+        UACLIENT_BEHAVE_CLOUD_CREDENTIALS_PATH = credentials('pycloudlib-config')
```
blackboxsw, line 10: Does this need to be established up in UA integration tests? pycloud lib already readys PYCLOUDLIB_CONFIG environment variable. If expected config is absent pycloudlib yells loudly and instructs how to fix it.

features/environment.py
blackboxsw: for what it's worth: cloud-init integration tests work by not requiring you to specify any cloud-specific environment files and just relies on a properly configured pycloudlib.toml or PYCLOUDLIB_CONFIG environment variable as override. I don't know if there is strong need to establish a new environment variable up in UACLIENT_BEHAVE_CLOUD_CREDENTIALS_PATH,  which could ultimately be understood anyway by pycloudlib if UA jenkins jobs/Jenkinsfile provided PYCLOUDLIB_CONFIG environment variable.

blackboxsw (commented):
Just some thoughts about whether we really need a separate env var up in UA tests. We can discuss tomorrow

blackboxsw:
Only real blocker is you need to bump pycloudlib to a higher commit.

features/environment.py
lucasmoura: Yeah, this is was definitely a mistake. I should have not changed that number. Thanks for catching that

Jenkinsfile
lucasmoura: Yeah, I will make that variable optional. However, I think we will need to setup the `PYCLOUDLIB_CONFIG` env var on jenkins, right @paride ?

lucasmoura:
@blackboxsw @orndorffgrant I agree with the point you both raised. Let's make the `UACLIENT_BEHAVE_CLOUD_CREDENTIALS_PATH`  variable optional

Jenkinsfile
paride: @lucasmoura yes but it has to be done here in the Jenkinsfile

    PYCLOUDLIB_CONFIG = credentials('pycloudlib-config')`

For this kind of job that's the only way, we can't use `credentials-binding`.

Jenkinsfile
```diff
@@ -7,22 +7,11 @@ pipeline {
...
         UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING = credentials(
             'ua-contract-token-staging'
         )
-        UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING_EXPIRED = credentials(
+        PYCLOUDLIB_CONFIG = credentials(
             'ua-contract-token-staging-expired'
```
paride, line 15: Wrong credential here?

Jenkinsfile
paride: Here I'm a bit lost: why do we need `UACLIENT_BEHAVE_CLOUD_CREDENTIALS_PATH` at all given that pycloudlib is going to use `PYCLOUDLIB_CONFIG`? Not suggesting dropping it, but I'm failing to get its use case.

Jenkinsfile
paride: @lucasmoura ^^ that should be `pycloudlib-config` I believe, or maybe we can just drop `PYCLOUDLIB_CONFIG` and go with `UACLIENT_BEHAVE_CLOUD_CREDENTIALS_PATH`. If we're keeping `PYCLOUDLIB_CONFIG` then it should be added to `passenv` in `tox.ini` otherwise the test environment will never see the env var.

lucasmoura:
@paride Thanks for the review. I am updating the PR to set up those variables

features/cloud.py
```diff
@@ -552,64 +428,61 @@ def _create_instance(
         )
         inst = self.api.launch(
             image_id=image_name,
+            instance_type="Standard_DC1s_v3",
```
paride, line 431: Let's try with `Standard_A2_v2`. https://docs.microsoft.com/en-us/azure/virtual-machines/av2-series says

> Av2-series VMs have CPU performance and memory configurations best suited for entry level workloads like development and test

and:

> [VM Generation Support](https://docs.microsoft.com/en-us/azure/virtual-machines/generation-2): Generation 1

paride (changes requested):

Jenkinsfile
```diff
@@ -253,6 +242,32 @@ pipeline {
...
+                    steps {
+                        sh '''
+                        set +x
+                        . $TMPDIR/bin/activate
+                        tox --parallel--safe-build -e behave-azurepro-18.04 -- --tags="~slow"
+                        '''
+                    }
+                }
+                stage("gcppro 18.04") {
+                    environment {
```
paride, line 259: Here we need:
```
GOOGLE_APPLICATION_CREDENTIALS = credentials('ua-gce-account')
```
because `pycloudlib.toml` is not enough yet. We also need to `passenv` the variable.

paride (changes requested):

lucasmoura:
Thanks for that @paride. I have updated the code

renanrodrigo:
@lucasmoura I have tested this again, and it works.
Since we are not running Jenkins at the moment, my suggestion is that we separate the pycloudlib bump from the Jenkins work as separate PRs, so we can merge the former while the latter waits for the infrastructure to be ready

renanrodrigo (approved):
Bump works great; some issues with a race condition on Xenial found, but unrelated.
LGTM

renanrodrigo:
Merged for the goods of Pycloudlib - further changes in CI can be done later

PR #1973, opened 2022-02-11
cloud: fix detaching in auto-attach

CalvoM:
## Proposed Commit Message

cloud: fix detaching in auto-attach

Stop the detaching process when calling auto-attach after a manual attach command.

## Desired commit type

## Checklist

### DIFF
--- a/features/detached_auto_attach.feature
+++ b/features/detached_auto_attach.feature
@@ -1,0 +1,32 @@
+@uses.config.contract_token
+Feature: Attached cloud does not detach when auto-attaching after manually attaching
+
+    @series.all
+    @uses.config.machine_type.aws.generic
+    @uses.config.machine_type.azure.generic
+    @uses.config.machine_type.gcp.generic
+    Scenario Outline: No detaching on manually attached machine on all clouds
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `ua refresh` with sudo
+        Then I will see the following on stdout:
+            """
+            Successfully processed your ua configuration.
+            Successfully refreshed your subscription.
+            """
+        When I run `ua auto-attach` with sudo
+        Then stderr matches regexp:
+        """
+        Skipping attach: Instance '[0-9a-z\-]+' is already attached.
+        """
+        When I run `ua status` with sudo
+        Then stdout matches regexp:
+        """
+        esm-infra    +yes      +<esm-service> +UA Infra: Extended Security Maintenance \(ESM\)
+        """
+
+        Examples: ubuntu release
+           | release | esm-service |
+           | bionic  | enabled     |
+           | focal   | enabled     |
+           | xenial  | enabled     |
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1103,7 +1103,7 @@ def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
             to_disable.append(ent)
 
     """
-    We will nake sure that services without dependencies are disabled first
+    We will make sure that services without dependencies are disabled first
     PS: This will only work because we have only three services with reverse
     dependencies:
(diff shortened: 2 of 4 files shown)


lucasmoura:
@CalvoM Yes, I think we should add one using a generic image

features/detached_auto_attach.feature
```diff
@@ -0,0 +1,33 @@
...
+            """
+            Successfully processed your ua configuration.
+            Successfully refreshed your subscription.
+            """
+        When I run `ua auto-attach` with sudo
+        Then stderr matches regexp:
+        """
+        Skipping attach: Instance '[0-9a-z\-]+' is already attached.
+        """
+        Then stdout does not match regexp:
```
lucasmoura, line 22: Let's just add a step here where we check `status` to verify that the machine is still attached. My only concern is that we could change the `This machine is now detached` message in the future and then the test will not be fully testing that scenario anymore

features/detached_auto_attach.feature
```diff
@@ -0,0 +1,34 @@
...
+        esm-infra    +yes      +<esm-service> +UA Infra: Extended Security Maintenance \(ESM\)
+        """
+
+        Examples: ubuntu release
+           | release | esm-service |
+           | bionic  | enabled     |
+           | focal   | enabled     |
+           | xenial  | enabled     |
+           | impish  | disabled    |
+           | jammy   | n/a         |
```
lucasmoura, line 34: I think we can drop these releases for now

lucasmoura:
Tests are passing here, I think after we remove those releases we can merge this PR

uaclient/cli.py
```diff
@@ -1287,6 +1287,9 @@ def action_attach(args, *, cfg):
         event.process_events()
         return 1
     else:
+        current_iid = identity.get_instance_id()
+        if current_iid:
+            cfg.write_cache("instance-id", current_iid)
```
orndorffgrant, line 1292: Instead of adding this here. Lets do something like this in `actions.py`
```diff
diff --git a/uaclient/actions.py b/uaclient/actions.py
index 9af063a6..9b950762 100644
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -44,6 +44,10 @@ def attach_with_token(
         update_apt_and_motd_messages(cfg)
         raise exc
 
+    current_iid = identity.get_instance_id()
+    if current_iid:
+        cfg.write_cache("instance-id", current_iid)
+
     update_apt_and_motd_messages(cfg)
 
 
@@ -69,9 +73,6 @@ def auto_attach(
                 ua_status.MESSAGE_UNSUPPORTED_AUTO_ATTACH
             )
         raise e
-    current_iid = identity.get_instance_id()
-    if current_iid:
-        cfg.write_cache("instance-id", current_iid)
 
     token = tokenResponse["contractToken"]
 

```

features/detached_auto_attach.feature
orndorffgrant: We should drop impish, though maybe we should start including jammy in new integration tests?

uaclient/cli.py
CalvoM: Considering that auto-attach does that in the `action.auto_attach` method, I think this will mean we run the same code snippet twice when running the auto-attach command. Unless, we also remove the same snippet from the auto_attach function and move it into the `attach_with_token` function.

CalvoM: @orndorffgrant Please check the update.

orndorffgrant: @CalvoM Change looks good! Now we just need to check for this write_cache call in test_actions.py:TestAttachWithToken instead of where it is currently failing in test_actions.py:TestAutoAttach

renanrodrigo (approved):
worksforme

lucasmoura (approved):

orndorffgrant (approved):
LGTM! Merging

PR #1974, opened 2022-02-11
Fix tox

lucasmoura:
## Proposed Commit Message
tox: set version of setuptools for some tests

For some older ubuntu releases, we need to set up an older version of pip. That's because the constraints we have for the older releases could not be successfully used with new versions of pip. This pip restriction is now affecting new versions of setuptools as well. Because of that, we are now setting up the versions of setuptools for those older ubuntu version tests

## Test Steps
Delete one of the modified environment and run tox for it again. See that it fails, then apply this fix and run tox again recreating the scenario

## Desired commit type

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ ifneq (,$(findstring trusty,$(TOXENV)))
 endif
 	pip install tox
 	pip install tox-pip-version
+	pip install tox-setuptools-version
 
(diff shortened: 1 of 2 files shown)


orndorffgrant (approved):
Worked great for me as well, thanks! Merging

PR #1975, opened 2022-02-11
Dockerfile tests and docs

orndorffgrant:
## Proposed Commit Message

```
tests: test using ua in docker builds
```

## Test Steps

Try both of the following:
```
env UACLIENT_BEHAVE_BUILD_PR=1 tox -e behave-docker
```
```
tox -e behave-docker
```

## Desired commit type

## Checklist

### DIFF
--- a/docs/howtoguides/enable_ua_in_dockerfile.md
+++ b/docs/howtoguides/enable_ua_in_dockerfile.md
@@ -1,0 +1,127 @@
+# How to Enable Ubuntu Advantage Services in a Dockerfile
+
+Ubuntu Advantage (UA) comes with several services, some of which can be useful in docker. For example, Extended Security Maintenance of packages and FIPS certified packages may be desirable in a docker image. In this how-to-guide, we show how you can use the `ua` tool to take advantage of these services in your Dockerfile.
+
+
+## Step 1: Create a UA Attach Config file
+
+> Warning: the UA Attach Config file will contain your UA Contract token and should be treated as a secret file.
+
+An attach config file for ua is a yaml file that specifies some options when running `ua attach`. The file has two fields, `token` and `enable_services` and looks something like this:
+
+```yaml
+token: TOKEN
+enable_services:
+  - service1
+  - service2
+  - service3
+```
+
+The `token` field is required and must be set to your UA token that you can get from signing into [ubuntu.com/advantage](https://ubuntu.com/advantage).
+
+The `enable_services` field value is a list of UA service names. When it is set, then the services specified will be automatically enabled after attaching with your UA token.
+
+Service names that you may be interested in enabling in your docker builds include:
+- `esm-infra`
+- `esm-apps`
+- `fips`
+- `fips-updates`
+
+You can find out more about these services by running `ua help service-name` on any Ubuntu machine.
+
+
+## Step 2: Create a Dockerfile to use `ua` and your attach config file
+
+Your Dockerfile is going to look something like this.
+
+There are comments inline explaining each line.
+
+```dockerfile
+# Base off of the LTS of your choice
+FROM ubuntu:focal
+
+# We mount a BuildKit secret here to access the attach config file which should
+# be kept separate from the Dockerfile and managed in a secure fashion since it
+# needs to contain your UA token.
+# In the next step, we demonstrate how to pass the file as a secret when
+# running docker build.
+RUN --mount=type=secret,id=ua-attach-config \
+
+    # First we update apt so we install the correct versions of packages in
+    # the next step
+    apt-get update \
+
+    # Here we install `ua` (ubuntu-advantage-tools) as well as ca-certificates,
+    # which is required to talk to the UA authentication server securely.
+    && apt-get install --no-install-recommends -y ubuntu-advantage-tools ca-certificates \
+
+    # With ua installed, we attach using our attach config file from the
+    # previous step
+    && ua attach --attach-config /run/secrets/ua-attach-config \
+
+    ###########################################################################
+    # At this point, the container has access to all UA services specified in
+    # the attach config file.
+    ###########################################################################
+
+    # Always upgrade all packages to the latest available version with the UA services
+    # enabled.
+    && apt-get upgrade -y \
+
+    # Then, you can install any specific packages you need for your docker
+    # container.
+    # Install them here, while UA is enabled, so that you get the appropriate
+    # versions.
+    # Any `apt-get install ...` commands you have in an existing Dockerfile
+    # that you may be migrating to use UA should probably be moved here.
+    && apt-get install -y openssl \
+
+    ###########################################################################
+    # Now that we've upgraded and installed any packages from the UA services,
+    # we can clean up.
+    ###########################################################################
+
+    # This purges ubuntu-advantage-tools, including all UA related secrets from
+    # the system.
+    ###########################################################################
+    # IMPORTANT: As written here, this command assumes your container does not
+    # need ca-certificates so it is purged as well.
+    # If your container needs ca-certificates, then do not purge it from the
+    # system here.
+    ###########################################################################
+    && apt-get purge --auto-remove -y ubuntu-advantage-tools ca-certificates \
+
+    # Finally, we clean up the apt lists which shouldn't be needed anymore
+    # because any `apt-get install`s should've happened above. Cleaning these
+    # lists keeps your image smaller.
+    && rm -rf /var/lib/apt/lists/*
+
+
+# Now, with all of your ubuntu apt packages installed, including all those
+# from UA services, you can continue the rest of your app-specific Dockerfile.
+```
+
+An important point to note about the above Dockerfile is that all of the `apt` and `ua` commands happen inside of one Dockerfile `RUN` instruction. This is critical and must not be changed. Keeping everything as written inside of one `RUN` instruction has two key benefits:
+
+1. Prevents any UA Subscription-related tokens and secrets from being leaked in an image layer
+2. Keeps the image as small as possible by cleaning up extra packages and files before the layer is finished.
+
+> Note: These benefits could also be attained by squashing the image.
+
+## Step 3: Build the Docker image
+
+
+Now, with our attach config file and Dockerfile created, we can build the image with a command like the following
+
+```bash
+DOCKER_BUILDKIT=1 docker build . --secret id=ua-attach-config,src=ua-attach-config.yaml -t ubuntu-focal-ua
+```
+
+There are two important pieces of this command.
+
+1. We enable BuildKit with `DOCKER_BUILDKIT=1`. This is necessary to support the secret mount feature.
+2. We use the secret mount feature of BuildKit with `--secret id=ua-attach-config,src=ua-attach-config.yaml`. This is what passes our attach config file in to be securely used by the `RUN --mount=type=secret,id=ua-attach-config` command in the Dockerfile.
+
+## Success
+
+Congratulations! At this point, you should have a docker image that has been built with UA packages installed from whichever UA service you required.
--- a/features/docker.feature
+++ b/features/docker.feature
@@ -48,9 +48,11 @@ Feature: Build docker images with ua services
         # Bloat is minimal (new size == original size + deb size + test package size)
         Then docker image `ua-test` is not significantly larger than `ubuntu:<container_release>` with `<test_package_name>` installed
 
-        # No secrets present
+        # No secrets or artifacts leftover
         Then `90ubuntu-advantage` is not present in any docker image layer
         Then `machine-token.json` is not present in any docker image layer
+        Then `ubuntu-advantage.log` is not present in any docker image layer
+        Then `uaclient.conf` is not present in any docker image layer
 
         # Service successfully enabled (Correct version of package installed)
         When I run `docker run ua-test dpkg-query --showformat='${Version}' --show <test_package_name>` with sudo
--- a/features/docker.feature
+++ b/features/docker.feature
@@ -1,0 +1,69 @@
+@uses.config.contract_token
+Feature: Build docker images with ua services
+
+    @slow
+    @docker
+    @series.focal
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Build docker images with ua services
+        Given a `focal` machine with ubuntu-advantage-tools installed
+        When I have the `<container_release>` debs under test in `/home/ubuntu`
+        When I run `apt install -y docker.io` with sudo
+        When I create the file `/home/ubuntu/Dockerfile` with the following:
+        """
+        FROM ubuntu:<container_release>
+
+        <copy_local_deb>
+
+        RUN --mount=type=secret,id=ua-attach-config \
+            apt-get update \
+            && apt-get install --no-install-recommends -y ubuntu-advantage-tools ca-certificates \
+
+            && <install_ua_under_test> \
+
+            && ua attach --attach-config /run/secrets/ua-attach-config \
+
+            # Normally an apt upgrade is recommended, but we dont do that here
+            # in order to measure the image size bloat from just the enablement
+            # process
+            # && apt-get upgrade -y \
+
+            && apt-get install -y <test_package_name> \
+
+            # If you need ca-certificates, remove it from this line
+            && apt-get purge --auto-remove -y ubuntu-advantage-tools ca-certificates \
+
+            && rm -rf /var/lib/apt/lists/*
+        """
+        When I replace `<copy_local_deb>` in `/home/ubuntu/Dockerfile` with `COPY ./ubuntu-advantage-tools.deb /ua.deb` if `build_pr` else ` `
+        When I replace `<install_ua_under_test>` in `/home/ubuntu/Dockerfile` with commands to install the `<container_release>` ua version under test
+        When I create the file `/home/ubuntu/ua-attach-config.yaml` with the following:
+        """
+        token: <contract_token>
+        enable_services: <enable_services>
+        """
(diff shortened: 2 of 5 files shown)


features/docker.feature
```diff
@@ -0,0 +1,71 @@
...
+            && apt-get install --no-install-recommends -y ubuntu-advantage-tools ca-certificates \
+
+            && <install_ua_under_test> \
+
+            && ua attach --attach-config /run/secrets/ua-attach-config \
+
+            # Normally an apt upgrade is recommended, but we dont do that here
+            # in order to measure the image size bloat from just the enablement
+            # process
+            # && apt-get upgrade -y \
```
orndorffgrant, line 30: I thought this was true at one point during testing but I'll try adding it back in

features/docker.feature
```diff
@@ -0,0 +1,71 @@
...
+        When I run `apt install -y docker.io` with sudo
+        When I create the file `/home/ubuntu/Dockerfile` with the following:
+        """
+        FROM ubuntu:<container_release>
+
+        <copy_local_deb>
+
+        RUN --mount=type=secret,id=ua-attach-config \
+            apt-get update \
+            && apt-get install --no-install-recommends -y <test_package_name> \
```
orndorffgrant, line 20: This line slipped in at some point but isn't necessary, I'll remove it

features/docker.feature
```diff
@@ -0,0 +1,71 @@
...
+            && apt-get install -y <test_package_name> \
+
+            # If you need ca-certificates, remove it from this line
+            && apt-get purge --auto-remove -y ubuntu-advantage-tools ca-certificates \
+
+            && rm -rf /var/lib/apt/lists/*
+        """
+        When I replace `<copy_local_deb>` in `/home/ubuntu/Dockerfile` with `COPY ./ubuntu-advantage-tools.deb /ua.deb` if `build_pr` else ` `
+        When I replace `<install_ua_under_test>` in `/home/ubuntu/Dockerfile` with commands to install the `<container_release>` ua version under test
+        When I run shell command `cat /home/ubuntu/Dockerfile` with sudo
```
orndorffgrant, line 41: This was left in from debuggint too, I'll remove

features/docker.feature
```diff
@@ -0,0 +1,71 @@
...
+
+        # Build succeeds
+        When I run shell command `DOCKER_BUILDKIT=1 docker build . --secret id=ua-attach-config,src=ua-attach-config.yaml -t ua-test` with sudo
+
+        # Bloat is minimal (new size == original size + deb size + test package size)
+        Then docker image `ua-test` is not significantly larger than `ubuntu:<container_release>` with `<test_package_name>` installed
+
+        # No secrets present
+        #Then `90ubuntu-advantage` is not present in any docker image layer
+        #Then `machine-token.json` is not present in any docker image layer
```
orndorffgrant, line 57: And I need to uncomment these

features/docker.feature
orndorffgrant: The comment here is accurate

features/steps/steps.py
```diff
@@ -441,6 +475,45 @@ def when_i_replace_string_in_file_with_token(
     when_i_replace_string_in_file(context, original, filename, token)
 
 
+@when(
+    "I replace `{original}` in `{filename}` with commands to install the `{series}` ua version under test"  # noqa: E501
+)
+def when_i_replace_string_in_file_with_install_commands(
```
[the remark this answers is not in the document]
orndorffgrant: @lucasmoura This is a good idea! And it is something I looked into briefly but didn't go too far. Maybe you can help though, because my lack of apt knowledge held me back a bit here.

Is there an easy way to `apt download` a `xenial` package when on a `focal` machine?

I'd rather keep the host machine of this test `focal` and just change the container base series version. The main reason is that `xenial`'s `docker` doesn't support mount secrets, which are necessary for this workflow :/

features/docker.feature
```diff
@@ -0,0 +1,69 @@
...
+        When I replace `<contract_token>` in `/home/ubuntu/ua-attach-config.yaml` with token `contract_token`
+
+        # Build succeeds
+        When I run shell command `DOCKER_BUILDKIT=1 docker build . --secret id=ua-attach-config,src=ua-attach-config.yaml -t ua-test` with sudo
+
+        # Bloat is minimal (new size == original size + deb size + test package size)
+        Then docker image `ua-test` is not significantly larger than `ubuntu:<container_release>` with `<test_package_name>` installed
+
+        # No secrets present
+        Then `90ubuntu-advantage` is not present in any docker image layer
```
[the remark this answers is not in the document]
orndorffgrant: We can do that too! I just only included the files with secrets in them because they're the most important.

But I think we might as well check for a couple other files too, I'll add them

features/docker.feature
```diff
@@ -0,0 +1,69 @@
+@uses.config.contract_token
+Feature: Build docker images with ua services
```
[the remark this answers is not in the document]
orndorffgrant: Oh thats a good idea, we should make sure any `ua` error properly breaks the build. And it should be easy to add once this main test is ironed out. I'll add it

features/steps/steps.py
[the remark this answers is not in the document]
lucasmoura: I think we can use `ubuntu-dev-tools` for this. The command `pull-lp-debs ubuntu-advantage-tools xenial` should download the latest version of the package (even if it is on proposed). It also seems that we can specify a ppa to pull the package from, so maybe we can adapt it for the daily case. But it is worth testing it out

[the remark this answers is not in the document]
orndorffgrant: oh very cool! I'll definitely test that out - thanks!

lucasmoura:
@athos-ribeiro could you please take a look at this PR ? I think your docker knowledge will be really useful to see if we are properly testing the docker image we are creating

features/docker.feature
```diff
@@ -0,0 +1,69 @@
...
+            && ua attach --attach-config /run/secrets/ua-attach-config \
+
+            # Normally an apt upgrade is recommended, but we dont do that here
+            # in order to measure the image size bloat from just the enablement
+            # process
+            # && apt-get upgrade -y \
+
+            && apt-get install -y <test_package_name> \
+
+            # If you need ca-certificates, remove it from this line
```
athos-ribeiro, line 33: is this correct?

features/steps/steps.py
```diff
@@ -847,6 +906,120 @@ def stdout_matches_the_json_schema(context, schema):
         )
 
 
+@then("`{file_name}` is not present in any docker image layer")
+def file_is_not_present_in_any_docker_image_layer(context, file_name):
+    when_i_run_command(context, "ls -1R /var/lib/docker/overlay2", "with sudo")
```
athos-ribeiro, line 911: Since we are already running a shell command here, why not using `'find /var/lib/docker/overlay2 -name {}.format(file_name)'` and checking if anything was pushed to stdout (no need for regular expressions)?

features/steps/steps.py
```diff
@@ -847,6 +920,127 @@ def stdout_matches_the_json_schema(context, schema):
...
+    )
+    custom_image_line = ""
+    base_image_line = ""
+    for line in context.process.stdout.strip().split("\n"):
+        if line.startswith("{}:latest".format(name)):
+            custom_image_line = line
+        if line.startswith("{}:latest".format(base_upgraded_image_name)):
+            base_image_line = line
+        if custom_image_line and base_image_line:
+            break
```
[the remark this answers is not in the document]
orndorffgrant: Thats a great idea! Will do

features/steps/steps.py
orndorffgrant: yes that is a much better approach, for some reason I never think to use `find` haha, will do

features/steps/steps.py
```diff
@@ -847,6 +906,120 @@ def stdout_matches_the_json_schema(context, schema):
...
+# This defines "not significantly larger" as "less than 2MB larger"
+@then(
+    "docker image `{name}` is not significantly larger than `ubuntu:{series}` with `{package}` installed"  # noqa: E501
+)
+def docker_image_is_not_larger(context, name, series, package):
+    base_image_name = "ubuntu:{}".format(series)
+    base_upgraded_image_name = "{}-upgraded".format(series)
+
+    # We need to commpare against the base image after apt upgrade
+    # and package install
```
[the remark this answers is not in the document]
orndorffgrant: Ah very good catch! This was there from an earlier iteration where I was running `upgrade` on both the main test image and this internal comparison image. I will re-consider all of this

features/docker.feature
```diff
@@ -0,0 +1,69 @@
...
+            && apt-get purge --auto-remove -y ubuntu-advantage-tools ca-certificates \
```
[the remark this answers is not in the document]
orndorffgrant: No but there's no reason to because it is already in a previous layer, so removing won't affect the size. And we're compensating for the deb file in the size comparison test

I suppose we could use a squashing tool like you suggest above, but I don't want to deviate too much from what we expect a customer would do

features/docker.feature
```diff
@@ -0,0 +1,69 @@
...
+
+            && rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
orndorffgrant: Because this is happening in the same `RUN` as the `apt update` that downloads the lists in the first place, this does keep the image smaller by about 30Mb in the test I just ran.

orndorffgrant:
@athos-ribeiro I think I've addressed all of your comments, please take another look when you have time :)

features/docker.feature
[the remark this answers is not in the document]
athos-ribeiro: Oh, of course! NVM :)
+1

athos-ribeiro (approved):
From a containers perspective, this LGTM.

It would be nice to display any contents that gets included in the 2MB margin that was left for comparing image sizes so we could gradually reduce that in the future though (it should be feasible to account for every byte in those layers).

orndorffgrant:
Thanks @athos-ribeiro!

That's an excellent point about the 2MB margin. Do you know of a good way to display the difference in contents? I found this tool https://github.com/GoogleContainerTools/container-diff which looks interesting. Or we could maybe just do something like `find / -type f | sort | tee image-file-list.txt` for each image and diff them (with some extra steps to get sizes). Any suggestions?

athos-ribeiro:
@orndorffgrant: you could get a list with each of the images layers first (you can get that list with `docker inspect` as well). Then you can discard the layers they have in common (they will have at  the base image layers in common), and proceed to check the layer directories to get that file list you want (note that the directories in the overlayfs directories are named after the layer hashes). Then you could even compare files for sizes (or even contents) when the same files are present in different layers and also report on files that are present in one image and not in the other.  You could also pre-fetch the list of files of any packages that are expected to be in one image and not in the other, and add them to some sort of denylist so they won't be flagged by this process.

orndorffgrant:
@athos-ribeiro Thank you! Those are great tips! I've created a Jira issue for that (SC-837) so as to not block this PR.

docs/howtoguides/enable_ua_in_dockerfile.md
```diff
@@ -0,0 +1,123 @@
+# How to Enable Ubuntu Advantage Services in a Dockerfile
```
orndorffgrant, line 1: Hello @valentincanonical please take a look at the generic how to guide here as well as the fips-centric tutorial below and let us know if you have any feedback :)

orndorffgrant: Thanks @valentincanonical!

Totally agree on all points and will make those changes.

Some specific responses:

> why does ca-certificates have to be explicitly installed instead of being an ubuntu-advantage-tools dependency?
We're generally very conservative about adding explicit dependencies because it adds bloat to all ubuntu-minimal-based machines in existence. `ua` will actually partially function without `ca-certificates` and will tell you to install it if it's needed for a given operation. We may want to consider adding this as an actual dependency though (or perhaps a "Recommended" dep), CC @lucasmoura @renanrodrigo @CalvoM @blackboxsw.

> FROM ubuntu:bionic: does it not work with focal?
It does work with focal. I just arbitrarily chose a release for the example, but you're right that we should default to our latest LTS. I'll change it :)

lucasmoura: Yes, I agree with @renanrodrigo point here

docs/howtoguides/enable_ua_in_dockerfile.md
```diff
@@ -0,0 +1,123 @@
...
+
+Ubuntu Advantage (UA) comes with several services, some of which can be useful in docker. For example, Extended Security Maintenance of packages and FIPS certified packages may be desirable in a docker image. In this how-to-guide, we show how you can use the `ua` tool take advantage of these services in your Dockerfile.
```
lucasmoura, line 3: s/tool take advantage/tool to take advantage

docs/howtoguides/enable_ua_in_dockerfile.md
```diff
@@ -0,0 +1,123 @@
...
+token: TOKEN
+enable_services:
+  - service1
+  - service2
+  - service3
+```
+
+The `token` field is required and must be set to your UA token that you can get from signing into [ubuntu.com/advantage](https://ubuntu.com/advantage).
+
+The `enable_services` field value is a list of UA service names. When it is set, then the services specified will be automatically enabled after attaching (read: authenticating and authorizing) with your UA token.
```
lucasmoura, line 20: Do we need the `(read: authenticating and authorizing)` part ?

docs/howtoguides/enable_ua_in_dockerfile.md
```diff
@@ -0,0 +1,123 @@
...
+
+Service names that you may be interested in enabling in your docker builds include:
```
lucasmoura, line 22: Should we include `cis, usg and cc-eal` or they are too specific that we should not advertise them here ?

docs/howtoguides/enable_ua_in_dockerfile.md
valentincanonical: Right, because it's the same package that is used in non-container ubuntu, correct? (and the "Recommended" won't help, as we're excluding recommends.) Then that should do it!

valentincanonical: @orndorffgrant I tried the tutorial with the daily PPA on focal and bionic today. Works well!
Only thing I would add is: we might want to add a line at the end to advice to squash the layers (as in, they have a ubuntu-focal-fips Dockerfile to build their FIPS-enabled base image, they squash and push it to a **private** registry, and then they can use it FROM whatever else Dockerfile they need)

features/steps/steps.py
```diff
@@ -127,6 +128,53 @@ def cleanup_instance() -> None:
...
+        when_i_run_shell_command(
+            context,
+            "cp ubuntu-advantage-tools*.deb ubuntu-advantage-tools.deb",
+            "with sudo",
+        )
+        when_i_run_shell_command(
+            context,
+            "cp ubuntu-advantage-pro*.deb ubuntu-advantage-pro.deb",
+            "with sudo",
+        )
```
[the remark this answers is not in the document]
orndorffgrant: Sorry I never responded here. Yes this is just in case for future features since it was easy to include, but isn't being used right now.

features/steps/steps.py
```diff
@@ -127,6 +128,53 @@ def cleanup_instance() -> None:
...
+                    )
+                ppa = "ppa:{}".format(match.group(1))
+            ppa_opts = "--distro ppa --ppa {}".format(ppa)
+        download_cmd = "pull-lp-debs {} ubuntu-advantage-tools {}".format(
+            ppa_opts, series
+        )
+        when_i_run_command(
+            context, "apt-get install -y ubuntu-dev-tools", "with sudo"
+        )
+        when_i_run_command(context, download_cmd, "with sudo")
```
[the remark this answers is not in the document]
orndorffgrant: Good idea! I'll add this

docs/howtoguides/enable_ua_in_dockerfile.md
orndorffgrant: No not really, I'll remove it

docs/howtoguides/enable_ua_in_dockerfile.md
orndorffgrant: yeah I'm not sure they make sense in a docker container context :shrug:

docs/howtoguides/enable_ua_in_dockerfile.md
orndorffgrant: @valentincanonical Glad it worked!
Just updated these docs with your suggestions - let me know what you think.

The UA team did discuss more about adding `ca-certificates` as an explicit dependency - we're going to consider adding it as a "Recommends" (which would mean removing the --no-install-recommends from this guide), but it is going to require more discussion and we shouldn't count on that happening before landing this PR.

With regards to squashing: The current tutorials are designed so that you don't need to squash. Because everything happens in one RUN command, everything is already happening in a single layer. So I'm not sure we should recommend it here. I did add a note to the how-to-guide that mentions squashing as an alternative to using one large RUN command though. Let me know if you think this is enough.

features/steps/steps.py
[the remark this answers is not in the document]
orndorffgrant: Added!

orndorffgrant:
@lucasmoura Just updated this addressing your last round of comments :)

lucasmoura (approved):

renanrodrigo (approved):

orndorffgrant:
I'm going to go ahead and merge this. Feel free to review more later @valentincanonical and we can always iterate on this content.

PR #1980, opened 2022-02-22
Add message code to json response

lucasmoura:
## Proposed Commit Message
Add message code to json response

For attach, detach, enable and disable, we are now adding the message_code information on the json response. This will give an idea of what is the idea of the message field when errors or warnings happen

PS: This is a large PR, let me know if need to break it down to ease the review process

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/features/attach_invalidtoken.feature
+++ b/features/attach_invalidtoken.feature
@@ -18,7 +18,7 @@ Feature: Command behaviour when trying to attach a machine to an Ubuntu
         When I verify that running `ua attach invalid-token --format json` `with sudo` exits `1`
         Then I will see the following on stdout:
             """
-            {"_schema_version": "0.1", "errors": [{"message": "Invalid token. See https://ubuntu.com/advantage", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            {"_schema_version": "0.1", "errors": [{"message": "Invalid token. See https://ubuntu.com/advantage", "message_code": "attach-invalid-token", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
             """
 
         Examples: ubuntu release
--- a/uaclient/event_logger.py
+++ b/uaclient/event_logger.py
@@ -103,18 +103,25 @@ class EventLogger:
         msg: str,
         service: Optional[str],
         event_dict: List[Dict[str, Optional[str]]],
+        code: Optional[str] = None,
         event_type: Optional[str] = None,
     ):
         if event_type is None:
             event_type = "service" if service else "system"
 
         event_dict.append(
-            {"type": event_type, "service": service, "message": msg}
+            {
+                "type": event_type,
+                "service": service,
+                "message": msg,
+                "message_code": code,
+            }
         )
 
     def error(
         self,
         error_msg: str,
+        error_code: Optional[str] = None,
         service: Optional[str] = None,
         error_type: Optional[str] = None,
     ):
@@ -129,6 +136,7 @@ class EventLogger:
                 msg=error_msg,
                 service=service,
                 event_dict=self._error_events,
+                code=error_code,
                 event_type=error_type,
             )
 
--- a/uaclient/exceptions.py
+++ b/uaclient/exceptions.py
@@ -1,7 +1,7 @@
 from typing import Dict, Optional
 from urllib import error
 
-from uaclient import status
+from uaclient import messages
 
 
 class UserFacingError(Exception):
@@ -45,9 +45,7 @@ class AlreadyAttachedOnPROError(UserFacingError):
 
     def __init__(self, instance_id: str):
         super().__init__(
-            status.MESSAGE_ALREADY_ATTACHED_ON_PRO.format(
-                instance_id=instance_id
-            )
+            messages.ALREADY_ATTACHED_ON_PRO.format(instance_id=instance_id)
         )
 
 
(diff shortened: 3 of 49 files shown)


features/attach_invalidtoken.feature
```diff
@@ -18,7 +18,7 @@ Feature: Command behaviour when trying to attach a machine to an Ubuntu
         When I verify that running `ua attach invalid-token --format json` `with sudo` exits `1`
         Then I will see the following on stdout:
             """
-            {"_schema_version": "0.1", "errors": [{"message": "Invalid token. See https://ubuntu.com/advantage", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
+            {"_schema_version": "0.1", "errors": [{"message": "Invalid token. See https://ubuntu.com/advantage", "message_code": "attach-invalid-token", "service": null, "type": "system"}], "failed_services": [], "needs_reboot": false, "processed_services": [], "result": "failure", "warnings": []}
```
lucasmoura, line 21: Instead of making the `message_code` numbers, I have decided to turn them into strings, since I think it will be to rely on strings in the Desktop/Subiquity code than on numbers.

If agree with that, one change we can do here is to create constants for the `message_codes` so developers can import then directly from ua

orndorffgrant: I completely agree with message codes as strings.

The idea of exposing the string constants is a good one, and I think we should talk about this more as we discuss making something like a "public" api for uaclient, but probably not include it in this PR.

uaclient/messages.py
```diff
@@ -0,0 +1,674 @@
+from uaclient.defaults import BASE_UA_URL, DOCUMENTATION_URL
+
+
+class NamedMessage:
+    def __init__(self, name: str, msg: str):
+        self.name = name
+        self.msg = msg
+
+
+class FormatedNamedMessage(NamedMessage):
```
orndorffgrant, line 10: Typo
```suggestion
class FormattedNamedMessage(NamedMessage):
```

uaclient/messages.py
```diff
@@ -0,0 +1,674 @@
...
+    def __init__(self, name: str, msg: str):
+        self.name = name
+        self.tmpl_msg = msg
+
+    def format(self, **msg_params):
+        self.msg = self.tmpl_msg.format(**msg_params)
+
+        return self
```
orndorffgrant, line 18: Doesn't matter that much, but a slightly different pattern to consider here (that I think is compatible with how this is used everywhere) might look like this:
```suggestion
class FormattedNamedMessage:
    def __init__(self, name: str, tmpl_msg: str):
        self.name = name
        self.tmpl_msg = tmpl_msg

    def format(self, **msg_params) -> NamedMessage:
        return NamedMessage(self.name, self.tmpl_msg.format(**msg_params))
```

Not worth it if it requires changing all the places where this is used though

orndorffgrant (commented):
Code review done: This is a great PR, with lots of little improvements mixed in :clap: :clap: 

Just a couple small comments mixed in.

We should get jenkins results (after we figure out what's wrong with jenkins) and I'll play with it locally too, but overall this looks good to go!

uaclient/event_logger.py
```diff
@@ -103,18 +103,25 @@ def _record_dict_event(
...
+            }
         )
 
     def error(
         self,
         error_msg: str,
+        error_code: Optional[str] = None,
         service: Optional[str] = None,
         error_type: Optional[str] = None,
     ):
```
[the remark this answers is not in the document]
renanrodrigo: +1
I think the error codes we will find in status will be similar to the attach ones : invalid / exipred token, connectivity error, etc.
We could add the message codes there in this PR as well, or in a separate one if you think it will be better

uaclient/exceptions.py
```diff
@@ -15,8 +17,108 @@ class UserFacingError(Exception):
 
     exit_code = 1
 
-    def __init__(self, msg: str) -> None:
+    def __init__(self, msg: str, msg_code: Optional[str] = None) -> None:
         self.msg = msg
+        self.msg_code = msg_code
```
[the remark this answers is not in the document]
renanrodrigo: Does it make sense to support one _or_ the other?

uaclient/messages.py
renanrodrigo: Does it make sense to just add `format` to `NamedMessage`? It would reduce complexity (and I think it is ok to format a message, semantically speaking) but not sure about how clear you want to be about the need to provide format args

renanrodrigo (commented):
Mostly agree with Grant's points, minor ideas inline

uaclient/event_logger.py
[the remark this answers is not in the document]
lucasmoura: We could do that, but you also need to modify how our exceptions handle `NamedMessage`, We could store the `NamedMessage` directly on the exception and try to access once it is raised. Do you think that is worth it ?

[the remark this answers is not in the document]
lucasmoura: The only problem here is related to exceptions as discussed on the other thread

lucasmoura:
@renanrodrigo Just replying to some comments you made

1) I agree with the status point, but I think it should be another PR, This one is already to big
2) We do need to make the `msg_code' optional here because some `UserFacingExceptions` still only require the error msg, like on the `security` module
3) I think it is better to keep them separate just to ease the differentiation between template messages and normal ones. But I if everybody agrees that it should be a single class, I can make the change, no problem at all

orndorffgrant (approved):
I think this looks good! +1

renanrodrigo (approved):

PR #1981, opened 2022-02-23
detach: remove contract detach call

lucasmoura:
## Proposed Commit Message
detach: remove contract detach call

The contract server is no longer detaching machines on their end. Therefore, we need to remove that call from detach, since it is doing nothing at the moment.

## Test Steps
Just verify that detach is still working as expected on the machine

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1131,10 +1131,6 @@ def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
     for ent in to_disable:
         _perform_disable(ent, cfg, assume_yes=assume_yes, update_status=False)
 
-    contract_client = contract.UAContractClient(cfg)
-    machine_token = cfg.machine_token["machineToken"]
-    contract_id = cfg.machine_token["machineTokenInfo"]["contractInfo"]["id"]
-    contract_client.detach_machine_from_contract(machine_token, contract_id)
     cfg.delete_cache()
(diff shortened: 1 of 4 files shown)


uaclient/cli.py
```diff
@@ -1131,10 +1131,6 @@ def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
     for ent in to_disable:
         _perform_disable(ent, cfg, assume_yes=assume_yes, update_status=False)
 
-    contract_client = contract.UAContractClient(cfg)
-    machine_token = cfg.machine_token["machineToken"]
-    contract_id = cfg.machine_token["machineTokenInfo"]["contractInfo"]["id"]
-    contract_client.detach_machine_from_contract(machine_token, contract_id)
```
renanrodrigo, line 1137: Since we don't need this functionality anymore, I would also remove this method from `UAContractClient` and the related unit tests

alnvdl (approved):

uaclient/cli.py
lucasmoura: Good point, I will do that

lucasmoura: @renanrodrigo updated

orndorffgrant (approved):
LGTM Lets just let Jenkins finish on it but I don't foresee any problems

CalvoM (approved):

PR #1984, opened 2022-03-01
daemon: cloud-id file shim

orndorffgrant:
## Proposed Commit Message

```
daemon: cloud-id file shim

This introduces a new systemd service that will run on startup only if
the version of cloud-init does not create the cloud-id file. The service
creates the same cloud-id file that cloud-init does in newer versions.
This is specifically necessary for xenial, and is a noop if cloud-init
is new enough.

It also runs the same script on postinst to cover the scenario where the
new version of cloud-init is installed but the instance has not
rebooted, and so cloud-init hasn't yet created the cloud-id file.
```

## Test Steps

run the edited integration tests

## Desired commit type

## Checklist

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -57,6 +57,7 @@ override_dh_systemd_enable:
 	dh_systemd_enable -pubuntu-advantage-tools ua-timer.timer
 	dh_systemd_enable -pubuntu-advantage-tools ua-timer.service
 	dh_systemd_enable -pubuntu-advantage-tools ubuntu-advantage.service
+	dh_systemd_enable -pubuntu-advantage-tools ubuntu-advantage-cloud-id-shim.service
 
 override_dh_systemd_start:
 	dh_systemd_start -pubuntu-advantage-tools ua-timer.timer
--- a/debian/rules
+++ b/debian/rules
@@ -57,7 +57,10 @@ override_dh_systemd_enable:
 	dh_systemd_enable -pubuntu-advantage-tools ua-timer.timer
 	dh_systemd_enable -pubuntu-advantage-tools ua-timer.service
 	dh_systemd_enable -pubuntu-advantage-tools ubuntu-advantage.service
+ifeq (${VERSION_ID},"16.04")
+	# Only enable cloud-id-shim on Xenial
 	dh_systemd_enable -pubuntu-advantage-tools ubuntu-advantage-cloud-id-shim.service
+endif
 
 override_dh_systemd_start:
 	dh_systemd_start -pubuntu-advantage-tools ua-timer.timer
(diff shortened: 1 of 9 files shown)


renanrodrigo (approved):
LGTM

CalvoM (approved):
LGTM.

systemd/ubuntu-advantage-cloud-id-shim.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
lucasmoura, line 1: Let's add some docs here on why we need this service

features/daemon.feature
```diff
@@ -1,17 +1,70 @@
...
+        Then I will see the following on stdout
+        """
+        lxd
+        """
+        When I run `cat /run/cloud-init/cloud-id-lxd` with sudo
+        Then I will see the following on stdout
+        """
+        lxd
+        """
+        # verify the shim service runs on boot and creates the cloud-id file
```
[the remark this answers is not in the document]
lucasmoura: +1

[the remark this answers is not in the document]
orndorffgrant: `/run` is a tmpfs so it is already cleared by the reboot :)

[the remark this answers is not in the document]
renanrodrigo: :man_facepalming: you right

orndorffgrant:
@renanrodrigo @lucasmoura @CalvoM @blackboxsw I updated this to reflect that the new version of cloud-init with the cloud-id-x file is now present in new images. This will now only install the shim on xenial.

renanrodrigo (approved):
makes sense, given that people updating u-a-t on existing bionic+ also update cloud-init

debian/rules
```diff
@@ -69,6 +73,11 @@ override_dh_auto_install:
 	# We want to guarantee that we are not shipping any conftest files
 	find $(CURDIR)/debian/ubuntu-advantage-tools -type f -name conftest.py -delete
 
+ifneq (${VERSION_ID},"16.04")
+	# Only install cloud-id-shim on Xenial
+	rm $(CURDIR)/debian/ubuntu-advantage-tools/lib/systemd/system/ubuntu-advantage-cloud-id-shim.service
```
lucasmoura, line 78: Maybe I am missing something here, but the idea would be to remove the `shim` if we upgrade from Xenial right ? If so, shouldn't the command be a `rm -f` in case we are already on a later version of Ubuntu ?

Also, since this is a systemd unit, don't we need to remove it similarly to what we do with the old timer ?

orndorffgrant: My understanding is that this runs during package creation (e.g. during `sbuild`) and this command removes the shim .service file from the package when building for non-xenial.

So then the Xenial package will have the file, so it will install/uninstall it.

Non-xenial packages will not have the file, so they won't even install it.

Note this is similar to how we're preventing `conftest` from getting into our deb packages

lucasmoura: Okay, that makes sense and answer my first question perfectly. But regarding a Xenial machine were the cloud-id-shim service is up and running. If we upgrade it to Bionic, wouldn't just removing the service through `rm` cause a weird state when running `systemctl status` ?

orndorffgrant: I'm also going to move this branch to the canonical repo to get that sweet CI

orndorffgrant: I tested this and couldn't find any errors.

orndorffgrant:
Closing in favor of #2044

PR #1986, opened 2022-03-02
ua security-status round of improvements (SC-433)

renanrodrigo:
This PR addresses feedback received from the beta version of `ua security-status`.
The main changes requested are:
- to include the origin site for the updates shown in the JSON
- to include the amount of packages installed from ESM

Aside from that, this PR also removes the `--beta` flag from the command and sets the schema version to `0.1`.

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -920,7 +920,7 @@ Feature: Command behaviour when attached to an UA subscription
         Then stdout is a json matching the `ua_security_status` schema
         And stdout matches regexp:
         """
-        "_schema_version": "0"
+        "_schema_version": "0.1"
         """
         And stdout matches regexp:
         """
@@ -954,7 +954,7 @@ Feature: Command behaviour when attached to an UA subscription
         And I run `ua security-status --format json` as non-root
         Then stdout matches regexp:
         """
-        "_schema_version": "0"
+        "_schema_version": "0.1"
         """
         And stdout matches regexp:
         """
@@ -976,7 +976,7 @@ Feature: Command behaviour when attached to an UA subscription
         Then stdout is a yaml matching the `ua_security_status` schema
         And stdout matches regexp:
         """
-        _schema_version: '0'
+        _schema_version: '0.1'
         """
         When I verify that running `ua security-status --format unsupported` `as non-root` exits `2`
         Then I will see the following on stderr:
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -944,6 +944,10 @@ Feature: Command behaviour when attached to an UA subscription
         """
         And stdout matches regexp:
         """
+        "origin": "esm.ubuntu.com"
+        """
+        And stdout matches regexp:
+        """
         "status": "pending_attach"
         """
         When I attach `contract_token` with sudo
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -916,7 +916,7 @@ Feature: Command behaviour when attached to an UA subscription
             """
         And I run `dpkg-reconfigure ubuntu-advantage-tools` with sudo
         And I run `apt-get update` with sudo
-        When I run `ua security-status --format json --beta` as non-root
+        When I run `ua security-status --format json` as non-root
         Then stdout is a json matching the `ua_security_status` schema
         And stdout matches regexp:
         """
@@ -947,7 +947,7 @@ Feature: Command behaviour when attached to an UA subscription
         "status": "pending_attach"
         """
         When I attach `contract_token` with sudo
-        And I run `ua security-status --format json --beta` as non-root
+        And I run `ua security-status --format json` as non-root
         Then stdout matches regexp:
         """
         "_schema_version": "0"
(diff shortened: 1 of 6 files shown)


features/attached_commands.feature
```diff
@@ -985,10 +985,14 @@ Feature: Command behaviour when attached to an UA subscription
...
+        And stdout matches regexp:
+        """
         "status": "pending_attach"
         """
         When I attach `contract_token` with sudo
-        And I run `ua security-status --format json --beta` as non-root
+        And I run `ua security-status --format json` as non-root
         Then stdout matches regexp:
         """
         "_schema_version": "0"
```
lucasmoura, line 998: We need to update the `schema_version` on other parts of the test

uaclient/security_status.py
```diff
@@ -26,20 +30,23 @@ class UpdateStatus(Enum):
     UNAVAILABLE = "upgrade_unavailable"
 
 
-def get_service_name(origins: List[apt_package.Origin]) -> str:
+def is_esm_package(package: apt_package.Package, service_name: str):
+    return any(
```
lucasmoura, line 34: Instead of using `any` that will evaluate all the origins, we can have a standard loop that will break if the condition we want is met. In that way, we may not need to go through all the origins

uaclient/security_status.py
```diff
@@ -26,20 +30,23 @@ class UpdateStatus(Enum):
...
-            and origin.origin == ESM_INFRA_ORIGIN
-        ):
-            return "esm-infra"
-        if (
-            "apps-security" in origin.archive
-            and origin.origin == ESM_APPS_ORIGIN
-        ):
-            return "esm-apps"
-    return "standard-security"
+        service = ORIGIN_INFORMATION_TO_SERVICE.get(
```
lucasmoura, line 44: Should we have some preference here ?

I am assuming the following scenario: We have a package that for some reason has the same version delivered on `esm-infra` and `esm-apps`. In that situation, we would choose randomly which package to origin comes first

uaclient/security_status.py
```diff
@@ -67,17 +74,15 @@ def filter_security_updates(
...
-    ]
-
     return [
         version
         for package in packages
         for version in package.versions
         if version > package.installed
-        and any(origin.archive in security_repos for origin in version.origins)
+        and any(
+            (origin.origin, origin.archive) in ORIGIN_INFORMATION_TO_SERVICE
```
lucasmoura, line 83: Why do we need to check both origin and archive here ? Wouldn't checking just one of them enough ?

uaclient/security_status.py
```diff
@@ -120,27 +125,37 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
     installed_packages = [package for package in cache if package.is_installed]
     summary["num_installed_packages"] = len(installed_packages)
 
-    security_upgradable_versions = filter_security_updates(installed_packages)
+    package_count = {"esm-infra": 0, "esm-apps": 0}
```
lucasmoura, line 128: Let's use `defaultdict` here

uaclient/security_status.py
```diff
@@ -120,27 +125,37 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
...
+    update_count = {"esm-infra": 0, "esm-apps": 0, "standard-security": 0}
 
-    package_count = {"esm-infra": 0, "esm-apps": 0, "standard-security": 0}
+    for package in installed_packages:
+        if is_esm_package(package, "esm-apps"):
+            package_count["esm-apps"] += 1
+        if is_esm_package(package, "esm-infra"):
+            package_count["esm-infra"] += 1
```
lucasmoura, line 135: If the package appear on both pockets for some reason, we would be counting it twice, right ?

uaclient/security_status.py
```diff
@@ -120,27 +125,37 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
...
+
+    security_upgradable_versions = filter_security_updates(installed_packages)
```
lucasmoura, line 137: Just to double check, if we have a package with higher version on both `security` and `esm-infra` we will show them both on the json response right ?

uaclient/security_status.py
renanrodrigo: Yes, but it is ok. The information we are providing is: you have |X| packages which are available in ESM-Infra, and |Y| packages which are available in ESM-Apps, but X and Y are not mutually exclusive - the same version can be in both

uaclient/security_status.py
renanrodrigo: Yes, we will - they will be listed as separate updates. The only thing to sort is between ESM though, because it is guaranteed that we will not have the same version in `security` and any other ESM pocket

uaclient/security_status.py
renanrodrigo: Most packages will have 2-3 origins so it is not a big deal to make the list; and `any` short circuits anyway.
Not a strong opinion on this though, can change np

uaclient/security_status.py
renanrodrigo: Indeed. This is a good question. Although I don't foresee any problem showing any of them, this seems like it's worth deeper thought on what we want to prioritize.

uaclient/security_status.py
lucasmoura: Oh okay, TIL that it does short circuit, so no problem at all

uaclient/security_status.py
orndorffgrant: I think this is unlikely because esm-apps is for universe packages and esm-infra is for main packages, right?

lucasmoura: It should be, yes, but I think we do have same packages that collide. I think `krb5-admin-server` is an example that appear on both esm security pockets.

However, maybe this is just a corner case that we don't need to properly address

orndorffgrant (approved):
I didn't find anything that @lucasmoura didn't already point out in the code and it worked for me when trying it out. Looks great!

uaclient/security_status.py
renanrodrigo: Ended up that I rewrote it to only go through origins once, as it costs more than checking the strings in the caller - small improvement (:

uaclient/security_status.py
renanrodrigo: As discussed, we can assume versions are always different across official origins so we don't have to worry here

uaclient/security_status.py
renanrodrigo: Solved with the above comment on versions

lucasmoura (approved):

PR #1988, opened 2022-03-07
Release 27.7

orndorffgrant:
WIP: still need to clean up after the rebase and write the changelog


orndorffgrant:
@lucasmoura @renanrodrigo @CalvoM This is ready for review. Though I still need to create the LP bug and add the number to the changelog

debian/changelog
```diff
@@ -1,3 +1,38 @@
+ubuntu-advantage-tools (27.7~22.04.1) jammy; urgency=medium
+
+  * d/logrotate:
+    - make new logs world readable
+  * d/tools.postinst:
+    - refactor to catch exception from entitlement_factory
```
lucasmoura, line 6: Not entirely sure if this entry should be in the changelog, since it seems like a change that will only matter to users if they have knowledge about the code.

renanrodrigo: Although the details may be, I'd keep it in the changelog because it is a change to postinst, and reviewers will care about it

lucasmoura: I see your point, but I think the changelog should be more user focus then reviewer related. But if we want to keep it, no problem

lucasmoura (approved):

debian/changelog
renanrodrigo: iirc, it is mandatory to have the debian/[files] changes in the changelog - if that does not hold, then we should remove it for sure

renanrodrigo (approved):

debian/changelog
orndorffgrant: I'm on the fence, but I think I'm leaning towards keeping it. If a user runs into an issue trying to install `ua`, they may look at the changelog and see that maybe the refactor was the problem

Issue #1976, opened 2022-02-14
Suggestion: Only ask for a subscription of things that exist


cpaelzer:
TL;DR I mistyped, but wondered about the kind of error message that I got

```
ubuntu@b:~$ sudo ua enable esm-infra--help
To use 'esm-infra--help' you need an Ubuntu Advantage subscription
Personal and community subscriptions are available at no charge
See https://ubuntu.com/advantage
```

I'd have expected something like a friendly version of "esm-infra--help means nothing to me, what are you talking about?"

orndorffgrant:
Tracking in SC-805

PR #1990, opened 2022-03-08
cli: validate services when unattached (SC-805)

renanrodrigo:
When trying to enable/disable, check if the services are valid and
present messages accordingly.

Added a new message name/content for the case when the user provides valid and invalid services to the CLI at the same time.

Fixes: #1976 

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -204,20 +204,20 @@ def verify_json_format_args(f):
     return new_f
 
 
-def assert_attached(unattached_msg_tmpl=None):
+def assert_attached(msg_function=None):
     """Decorator asserting attached config.
-
-    :param unattached_msg_tmpl: Optional msg template to format if raising an
-        UnattachedError
+    :param msg_function: Optional function to generate a custom message
+    if raising an UnattachedError
     """
 
     def wrapper(f):
         @wraps(f)
         def new_f(args, cfg, **kwargs):
             if not cfg.is_attached:
-                if unattached_msg_tmpl:
-                    names = getattr(args, "service", "None")
-                    msg = unattached_msg_tmpl.format(name=", ".join(names))
+                if msg_function:
+                    command = getattr(args, "command", "")
+                    service_names = getattr(args, "service", "")
+                    msg = msg_function(command, service_names)
                     exception = exceptions.UnattachedError(msg)
                 else:
                     exception = exceptions.UnattachedError()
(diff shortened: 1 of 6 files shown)


uaclient/cli.py
```diff
@@ -206,20 +206,38 @@ def new_f(cmd_args, *args, **kwargs):
     return new_f
 
 
-def assert_attached(unattached_msg_tmpl=None):
-    """Decorator asserting attached config.
-
-    :param unattached_msg_tmpl: Optional msg template to format if raising an
-        UnattachedError
-    """
+def assert_attached():
```
CalvoM, line 209: Just a question, why did you move the msg from here?

renanrodrigo: Originally, the message was designed in a way that there is a generic "Unattached Error", but some commands could define their own messages.

I changed it because now there is logic to determine when to use different messages, even for a single command like `enable`. As we need different messages and specific logic, even passing a list of messages would not do, and getting the command from the args is easier than passing information through the decorator imho.

uaclient/cli.py
```diff
@@ -206,20 +206,38 @@ def new_f(cmd_args, *args, **kwargs):
...
+                    elif entitlements_found:
+                        msg = messages.VALID_SERVICE_FAILURE_UNATTACHED.format(
+                            valid_service=", ".join(entitlements_found)
+                        )
+                    else:
+                        msg = messages.INVALID_SERVICE_OP_FAILURE.format(
+                            operation=command,
+                            invalid_service=", ".join(entitlements_not_found),
+                            service_msg="See https://ubuntu.com/advantage",
+                        )
```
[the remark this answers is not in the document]
renanrodrigo: @lucasmoura done - it's setting up a function that takes args to ensure the flexibility for any other command

uaclient/cli.py
```diff
@@ -908,9 +906,40 @@ def action_config_unset(args, *, cfg, **kwargs):
     return 0
 
 
+def _create_enable_disable_unattached_msg(args):
```
lucasmoura, line 909: I think the function should receive the command and services as an argument instead, just so we can better decouple it from the cli args

lucasmoura (approved):
Just a minor nit, but LGTM

uaclient/cli.py
```diff
@@ -204,20 +204,20 @@ def new_f(cmd_args, *args, **kwargs):
...
     def wrapper(f):
         @wraps(f)
         def new_f(args, cfg, **kwargs):
             if not cfg.is_attached:
-                if unattached_msg_tmpl:
-                    names = getattr(args, "service", "None")
-                    msg = unattached_msg_tmpl.format(name=", ".join(names))
+                if msg_function:
+                    command = getattr(args, "command", "None")
+                    service_names = getattr(args, "service", "None")
```
lucasmoura, line 219: Any reason not to use empty strings as default values ?

renanrodrigo: I just followed what was already there hahaha
But no, we don't even expect those to be absent at all.
Will change for consistency sake

renanrodrigo: done

lucasmoura (approved):

PR #1994, opened 2022-03-15
cli: add additional_info to json response errors

lucasmoura:
## Proposed Commit Message
cli: add additional_info to json response errors

If users want to modify the error messages delivered by us, we need to provide additional info to some of those error messages. We are now adding that information only for expired contract errors but we can add more additional info in the future if needed.

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1242,16 +1242,7 @@ def action_attach(args, *, cfg):
     try:
         actions.attach_with_token(cfg, token=token, allow_enable=allow_enable)
     except exceptions.UrlError:
-        msg = messages.ATTACH_FAILURE
-        event.info(msg.msg)
-        event.error(error_msg=msg.msg, error_code=msg.name)
-        event.process_events()
-        return 1
-    except exceptions.UserFacingError as exc:
-        event.info(exc.msg)
-        event.error(error_msg=exc.msg, error_code=exc.msg_code)
-        event.process_events()
-        return 1
+        raise exceptions.AttachError()
     else:
         ret = 0
         if enable_services_override is not None and args.auto_enable:
@@ -1660,7 +1651,12 @@ def main_error_handler(func):
         except exceptions.UserFacingError as exc:
             with util.disable_log_to_console():
                 logging.error(exc.msg)
-            event.error(error_msg=exc.msg, error_code=exc.msg_code)
+
+            event.error(
+                error_msg=exc.msg,
+                error_code=exc.msg_code,
+                additional_info=exc.additional_info,
+            )
             event.info(info_msg="{}".format(exc.msg), file_type=sys.stderr)
             if not isinstance(exc, exceptions.LockHeldError):
                 # Only clear the lock if it is ours.
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -391,14 +391,24 @@ def _create_attach_forbidden_message(
 
         if reason == "no-longer-effective":
             date = info["time"].strftime(ATTACH_FAIL_DATE_FORMAT)
+            additional_info = {
+                "contract_expiry_date": info["time"].strftime("%m-%d-%Y"),
+                "contract_id": contract_id,
+            }
             reason_msg = messages.ATTACH_FORBIDDEN_EXPIRED.format(
                 contract_id=contract_id, date=date
             )
+            reason_msg.additional_info = additional_info
         elif reason == "not-effective-yet":
             date = info["time"].strftime(ATTACH_FAIL_DATE_FORMAT)
+            additional_info = {
+                "contract_effective_date": info["time"].strftime("%m-%d-%Y"),
+                "contract_id": contract_id,
+            }
             reason_msg = messages.ATTACH_FORBIDDEN_NOT_YET.format(
                 contract_id=contract_id, date=date
             )
+            reason_msg.additional_info = additional_info
         elif reason == "never-effective":
             reason_msg = messages.ATTACH_FORBIDDEN_NEVER.format(
                 contract_id=contract_id
@@ -407,6 +417,7 @@ def _create_attach_forbidden_message(
         if reason_msg:
             msg = messages.ATTACH_FORBIDDEN.format(reason=reason_msg.msg)
             msg.name = reason_msg.name
+            msg.additional_info = reason_msg.additional_info
 
     return msg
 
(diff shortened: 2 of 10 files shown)


uaclient/contract.py
```diff
@@ -391,14 +391,24 @@ def _create_attach_forbidden_message(
 
         if reason == "no-longer-effective":
             date = info["time"].strftime(ATTACH_FAIL_DATE_FORMAT)
+            additional_info = {
+                "contract": {"expiry_date": info["time"].strftime("%m-%d-%Y")}
```
orndorffgrant, line 395: Do we need the nesting? I think
```
{"contract_expiry_date": info["time"].strftime("%m-%d-%Y")}
```
would be fine

uaclient/contract.py
```diff
@@ -391,14 +391,24 @@ def _create_attach_forbidden_message(
...
             reason_msg = messages.ATTACH_FORBIDDEN_EXPIRED.format(
                 contract_id=contract_id, date=date
             )
+            reason_msg.additional_info = additional_info
         elif reason == "not-effective-yet":
             date = info["time"].strftime(ATTACH_FAIL_DATE_FORMAT)
+            additional_info = {
+                "contract": {
+                    "effective_date": info["time"].strftime("%m-%d-%Y")
+                }
```
orndorffgrant, line 406: Same question here, is the nesting needed?

uaclient/event_logger.py
```diff
@@ -42,8 +42,8 @@ class EventLoggerMode(enum.Enum):
 
 class EventLogger:
     def __init__(self):
-        self._error_events = []  # type: List[Dict[str, Optional[str]]]
-        self._warning_events = []  # type: List[Dict[str, Optional[str]]]
+        self._error_events = []  # type: List[Dict[str, Any]]
+        self._warning_events = []  # type: List[Dict[str, Any]]
```
orndorffgrant, line 46: Using `Any` in types makes me sad :( but I understand what is happening.

Would it be too much to define a simple class just to handle the different fields here instead of using a dict?

uaclient/cli.py
```diff
@@ -1244,16 +1244,7 @@ def action_attach(args, *, cfg):
...
-        event.info(msg.msg)
-        event.error(error_msg=msg.msg, error_code=msg.name)
-        event.process_events()
-        return 1
-    except exceptions.UserFacingError as exc:
-        event.info(exc.msg)
-        event.error(error_msg=exc.msg, error_code=exc.msg_code)
-        event.process_events()
-        return 1
+        raise exceptions.AttachError
```
orndorffgrant, line 1247: I think this is supposed to be instantiating the class?
```suggestion
        raise exceptions.AttachError()
```

uaclient/event_logger.py
orndorffgrant: Related but not necessarily useful for this case. I'd like to add a `to_dict()` method to `uaclient.data_types.DataObject` to have simple dict->PythonObject->dict translation.

orndorffgrant (commented):
Looking good! Just a couple minor suggestions/questions

uaclient/cli.py
lucasmoura: :facepalm: Yes, my mistake. Thanks for spotting this

uaclient/event_logger.py
lucasmoura: Since we will be using a non-nested approach and also assuming all values will strings, I can turn this back into `List[Dict[str, str]]`. But if that changes in the future, the class idea is a good one

lucasmoura: I have not created a class for it, but a `Union` type, specially since we will not be using a nested dict anymore. Also, I think we can always use string as values for the `additional_info` fields, which will simplify the logic.

What do you think ?

lucasmoura:
@orndorffgrant code updated

uaclient/contract.py
```diff
@@ -391,14 +391,22 @@ def _create_attach_forbidden_message(
 
         if reason == "no-longer-effective":
             date = info["time"].strftime(ATTACH_FAIL_DATE_FORMAT)
+            additional_info = {
+                "contract_expiry_date": info["time"].strftime("%m-%d-%Y")
```
orndorffgrant, line 395: Thanks I think the non-nested is simpler :)

We also have the contract_id in the response here, should we include that as well?

uaclient/event_logger.py
```diff
@@ -9,11 +9,12 @@
 import enum
 import json
 import sys
-from typing import Dict, List, Optional, Set  # noqa: F401
+from typing import Dict, List, Optional, Set, Union  # noqa: F401
 
 from uaclient.status import format_machine_readable_output
 
 JSON_SCHEMA_VERSION = "0.1"
+EventErrorType = Union[Optional[str], Dict[str, str]]
```
orndorffgrant, line 17: This is definitely better than `Any` :)

A small suggestion that is IMO slightly more readable but essentially the same:
```suggestion
EventFieldValueType = Optional[Union[str, Dict[str, str]]]
```

orndorffgrant (approved):
Some small suggestions but overall looks good!

uaclient/event_logger.py
lucasmoura: Just changing the name to `EventFieldErrorType´, since I think the proposed name may imply that all event fields follow that rule. But let me know if you disagree

uaclient/messages.py
```diff
@@ -1,10 +1,21 @@
...
 from uaclient.defaults import BASE_UA_URL, DOCUMENTATION_URL
 
 
 class NamedMessage:
     def __init__(self, name: str, msg: str):
         self.name = name
         self.msg = msg
+        self._additional_info = None  # type: Optional[Dict[str, str]]
+
+    @property
```
renanrodrigo, line 12: As there is no operations/checks when setting/reading `additional_info`, could it be just an attribute instead of a property?

uaclient/tests/test_cli_attach.py
```diff
@@ -273,8 +273,9 @@ def fake_request_updated_contract(cfg, contract_token, allow_enable):
             raise error_class(error_str)
 
         request_updated_contract.side_effect = fake_request_updated_contract
-        ret = action_attach(args, cfg=cfg)
-        assert 1 == ret
+        with pytest.raises(SystemExit):
```
renanrodrigo, line 276: I think it is useful to check the `code` for the `SystemExit` here is 1, as the old test was doing

renanrodrigo (approved):
LGTM
Two minor suggestions, feel free to apply as you wish

uaclient/messages.py
lucasmoura: Yeah, that one I am not entirely sure on how to proceed. I just want to be explicit about the existence of that attribute, so that's why I am using the `property` here. But maybe it is indeed unnecessary

renanrodrigo: I'm fine with any implementation, as this will have no major impact except for the boilerplate for the property.
Maybe a sensible dosctring can point to it explicitly?
Anyway, up to you - feel free to merge it as you think it will be better

PR #1995, opened 2022-03-16
fips: Disable FIPS when enabling FIPS Updates (SC-830)

lucasmoura:
## Proposed Commit Message
fips: Disable FIPS when enabling FIPS Updates

We cannot allow both FIPS and FIPS Updates in the system. FIPS Updates will not be able to hold the certification guarantees that we have for FIPS. Additionally, keeping both of these services alive will not allow us to enable Livepatch, since Livepatch cannot be enabled with FIPS. Because of that we are asking for users enabling FIPS Updates to disable FIPS to make enable process successful

## Test Steps
Run the new integration test

## Desired commit type

## Checklist

### DIFF
--- a/features/enable_fips_vm.feature
+++ b/features/enable_fips_vm.feature
@@ -465,14 +465,11 @@ Feature: FIPS enablement in lxd VMs
            | focal   | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips-updates/ubuntu focal-updates/main |
 
     @slow
-    @series.xenial
-    @series.bionic
+    @series.lts
     @uses.config.machine_type.lxd.vm
     Scenario Outline: Attached enable fips-updates on fips enabled vm
         Given a `<release>` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
-        And I run `ua disable livepatch` with sudo
-        And I run `DEBIAN_FRONTEND=noninteractive apt-get install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y openssh-client openssh-server strongswan` with sudo, retrying exit [100]
         And I run `ua enable fips --assume-yes` with sudo
         Then stdout matches regexp:
             """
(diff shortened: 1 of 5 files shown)


renanrodrigo (approved):
lgtm

features/enable_fips_vm.feature
```diff
@@ -575,3 +575,78 @@ Feature: FIPS enablement in lxd VMs
         Examples: ubuntu release
         | release |
         | focal   |
+
+    @slow
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable FIPS Updates when FIPS already enabled
```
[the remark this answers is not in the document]
lucasmoura: Good point, I have missed that test

lucasmoura:

features/enable_fips_vm.feature
```diff
@@ -487,47 +486,68 @@ Feature: FIPS enablement in lxd VMs
...
         Then stdout matches regexp:
             """
-            Updating package lists
-            Installing FIPS Updates packages
-            FIPS Updates enabled
-            A reboot is required to complete install
+            fips
             """
-        When I run `ua status --all` with sudo
+        When I verify that running `ua enable fips-updates --assume-yes` `with sudo` and stdin `y` exits `0`
```
orndorffgrant, line 494: It is a bit confusing that we still need to answer an interactive prompt even with `--assume-yes`.

Also I think I'd rather see what happens without `--assume-yes` for the integration test IMO

lucasmoura: I have found some bugs related to the `assume-yes` flag. Thanks for spotting that.
However, since we have two prompts (One for fips-updates and one for disable), I couldn't use our default stdin to handle that, so I am keeping the `assume-yes` flag here

renanrodrigo (approved):

features/enable_fips_vm.feature
lucasmoura: I have managed to not use `--assume-yes` on the test. I think we should be ok now

orndorffgrant (approved):

PR #1996, opened 2022-03-16
Cloud overrides core

renanrodrigo:
Support generic overrides from the contract server.
Currently, the supported overrides are `series` and `cloud`, but more can be added in the future.

Further testing, mainly integration testing, can be performed once contracts starts applying those specific overrides with anything different than the regular `series` one that we had before. Right now, no flow has changed.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/tests/test_util.py
+++ b/uaclient/tests/test_util.py
@@ -474,57 +474,244 @@ class TestGetPlatformInfo:
                     assert expected == util.get_platform_info.__wrapped__()
 
 
-class TestApplySeriesOverrides:
+class TestApplyContractOverrides:
+    @pytest.mark.parametrize(
+        "override_selector,expected_weight",
+        (
+            ({"selector1": "valueX", "selector2": "valueZ"}, 0),
+            ({"selector1": "valueA", "selector2": "valueZ"}, 0),
+            ({"selector1": "valueX", "selector2": "valueB"}, 0),
+            ({"selector1": "valueA"}, 1),
+            ({"selector2": "valueB"}, 2),
+            ({"selector1": "valueA", "selector2": "valueB"}, 3),
+        ),
+    )
+    def test_get_override_weight(self, override_selector, expected_weight):
+        selector_values = {"selector1": "valueA", "selector2": "valueB"}
+        selector_weights = {"selector1": 1, "selector2": 2}
+        with mock.patch(
+            "uaclient.util.OVERRIDE_SELECTOR_WEIGHTS", selector_weights
+        ):
+            assert expected_weight == util._get_override_weight(
+                override_selector, selector_values
+            )
+
     def test_error_on_non_entitlement_dict(self):
         """Raise a runtime error when seeing invalid dict type."""
         with pytest.raises(RuntimeError) as exc:
-            util.apply_series_overrides({"some": "dict"})
+            util.apply_contract_overrides({"some": "dict"})
         error = (
             'Expected entitlement access dict. Missing "entitlement" key:'
             " {'some': 'dict'}"
         )
         assert error == str(exc.value)
 
+    @pytest.mark.parametrize("include_overrides", (True, False))
     @mock.patch(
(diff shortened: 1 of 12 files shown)


renanrodrigo:
Yellow flag: while changing this function, I noticed that we'll call it around _4K TIMES_ during attach. For now, we need to be careful with side effects (I needed to cache the `cloud-id` call here) but refactoring this in the future seems sensible.

uaclient/util.py
```diff
@@ -100,7 +102,43 @@ def object_hook(o):
...
+
+
+def _select_overrides(
+    entitlement: Dict[str, Any], series_name: str, cloud_type: str
+) -> Dict[int, Dict[str, Any]]:
+    overrides = {}
+
+    selector_values = {"series": series_name, "cloud": cloud_type}
+
+    series_overrides = entitlement.pop("series", {}).pop(series_name, {})
```
lucasmoura, line 124: Aren't you modifying the global `entitlement` dict here by performing `pop`. I am just worried if might need that information later in the process

uaclient/tests/test_util.py
```diff
@@ -474,57 +474,216 @@ def test_get_platform_info_with_version(
...
+                "obligations": {
+                    "new_obligation": True,
+                    "some_obligation": True,
+                },
+            }
+        }
 
-        assert expected == orig_access
+        util.apply_contract_overrides(orig_access)
+        assert orig_access == expected
```
lucasmoura, line 686: Should we add a test where we replace something with the selector ?

renanrodrigo: This test covers the case where we add new things and replace some others (like the `some_affordance`). Sorry, maybe I missed your point here

lucasmoura: My mistake, I missed that the list will be completely replaced. Just ignore this

uaclient/util.py
orndorffgrant: Right this aligns with the old implementation.

Not for this PR, but I think a worthwhile refactor would be to make this more "functional". So it wouldn't modify the argument and would return a new object with overrides applied

uaclient/util.py
```diff
@@ -117,23 +155,30 @@ def apply_series_overrides(
...
-            # If the key already exists and is a dict, update that dict using
-            # the override
-            current.update(value)
-        else:
-            # Otherwise, replace it wholesale
-            orig_access["entitlement"][key] = value
+
+    overrides = _select_overrides(orig_entitlement, series_name, cloud_type)
+
+    for _weight, overrides_to_apply in sorted(overrides.items()):
```
orndorffgrant, line 172: TIL `sorted` sorts dict_items by key with no extra params

uaclient/util.py
```diff
@@ -100,7 +102,43 @@ def object_hook(o):
         return o
 
 
-def apply_series_overrides(
+def _get_override_weight(
```
orndorffgrant, line 105: Unit tests just for this function might be useful and pretty easy to write

orndorffgrant (commented):
Overall looks great!

I think we're missing a test for this scenario: There is a entitlement with two overrides. Override has just a series selector and override 2 has just a cloud selector. Both match - which one wins? (should be the cloud one from reading the code, but we should have a test)

uaclient/util.py
renanrodrigo: sure will add

renanrodrigo:
@orndorffgrant The test Lucas commented (line 633) is covering the case when those overrides are different: we have series only, cloud only, and series+cloud - in this case, all are applied.
The previous test assures that when the override is the same, the highest weight wins.
Maybe I am missing the point on the scenario you mentioned - could you elaborate?

renanrodrigo:
@lucasmoura I addresses Grant's comments and included an integration test assuring cloud overrides are applied

lucasmoura (approved):

PR #2004, opened 2022-03-21
Remove unnecessary creation of UAConfig instances

lucasmoura:
## Proposed Commit Message
Remove unnecessary creation of UAConfig instances

Creating an UAConfig instance can be an expansive operation that can be avoided if we share the instance we create for every cli call throught the code. We are refactoring the code to allow that reuse of the instance when possible

I have tested this work to see if we would get any speedup when running some UA commands:
`ua status` from around 2.5s to 2.0s
`ua attach` form around 17s to 16.2s

Tests were performed with a bionic LXD container running both commands around 3 times using the latest package of UA and another with the config refactor applied.

## Test Steps
Run the integration steps for attach/enable/disable/deach and verify that the package doesn't break and that it still work as expected

## Desired commit type

## Checklist

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -60,7 +60,9 @@ redact_ubuntu_release_from_ua_apt_filenames() {
     DIR=$1
     UA_SERVICES=$(/usr/bin/python3 -c "
 from uaclient.entitlements import valid_services
-print(*valid_services(allow_beta=True, all_names=True), sep=' ')
+from uaclient.config import UAConfig
+cfg = UAConfig()
+print(*valid_services(cfg=cfg, allow_beta=True, all_names=True), sep=' ')
 ")
 
     for file in "$DIR"/*; do
@@ -122,8 +124,8 @@ check_service_is_beta() {
 from uaclient.config import UAConfig
 from uaclient.entitlements import entitlement_factory
 try:
-    ent_cls = entitlement_factory('${service_name}')
     cfg = UAConfig()
+    ent_cls = entitlement_factory(cfg=cfg, name='${service_name}')
     allow_beta = cfg.features.get('allow_beta', False)
     print(all([ent_cls.is_beta, not allow_beta]))
 except Exception:
--- a/lib/upgrade_lts_contract.py
+++ b/lib/upgrade_lts_contract.py
@@ -47,7 +47,8 @@ current_codename_to_past_codename = {
 
 def process_contract_delta_after_apt_lock() -> None:
     logging.debug("Check whether to upgrade-lts-contract")
-    if not UAConfig().is_attached:
+    cfg = UAConfig()
+    if not cfg.is_attached:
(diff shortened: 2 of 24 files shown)


CalvoM (commented):
Might be needed later, I think we should update the function docstrings as well. We might need another PR for that.

renanrodrigo (approved):
LGTM!
Thanks again for diving into this.
I do agree that we should update the docstrings that this PR has outdated - besides that, it's great. In the future, we may want to improve our docstrings overall.

debian/ubuntu-advantage-tools.postinst
```diff
@@ -60,7 +60,9 @@ redact_ubuntu_release_from_ua_apt_filenames() {
     DIR=$1
     UA_SERVICES=$(/usr/bin/python3 -c "
 from uaclient.entitlements import valid_services
-print(*valid_services(allow_beta=True, all_names=True), sep=' ')
+from uaclient.config import UAConfig
+cfg = UAConfig()
+print(*valid_services(cfg=cfg, allow_beta=True, all_names=True), sep=' ')
 ")
```
orndorffgrant, line 66: This `redact_ubuntu_release_from_ua_apt_filenames` function is only for an old migration of apt filenames. Can we just hardcode this list to the current list of services? (new services will never need to undergo this migration)

orndorffgrant: And sorry, that's not really relevant to this PR, but since we're touching this anyway, it might be a good time :) (but maybe it deserves it's own PR?)

lucasmoura: I completely agree we can change this, but I think we could do this in a separate PR

orndorffgrant:
Thank you for this!!
Before:
```
root@dev-x:~# truncate -s 0 /var/log/ubuntu-advantage.log 
root@dev-x:~# ua attach TOKEN
...
root@dev-x:~# wc -l /var/log/ubuntu-advantage.log
954 /var/log/ubuntu-advantage.log
```
After:
```
root@dev-x:~# truncate -s 0 /var/log/ubuntu-advantage.log 
root@dev-x:~# ua attach TOKEN
...
root@dev-x:~# wc -l /var/log/ubuntu-advantage.log
166 /var/log/ubuntu-advantage.log
```

debian/ubuntu-advantage-tools.postinst
```diff
@@ -122,8 +124,8 @@ check_service_is_beta() {
 from uaclient.config import UAConfig
 from uaclient.entitlements import entitlement_factory
 try:
-    ent_cls = entitlement_factory('${service_name}')
     cfg = UAConfig()
+    ent_cls = entitlement_factory(cfg=cfg, '${service_name}')
```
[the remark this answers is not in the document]
lucasmoura: Good catch, fixing it now

lucasmoura:
@orndorffgrant PR updated

lucasmoura:
Closing this in favor of #2037

PR #2006, opened 2022-03-22
postinst: Adapt script for new messages module

lucasmoura:
postinst: Adapt script for new messages module

## Test Steps
Run any of the PRO FIPS tests we have and assert that no package related issues are raised

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/messages.py
+++ b/uaclient/messages.py
@@ -102,6 +102,7 @@ REBOOT_SCRIPT_FAILED = (
 LIVEPATCH_LTS_REBOOT_REQUIRED = (
     "Livepatch support requires a system reboot across LTS upgrade."
 )
+FIPS_REBOOT_REQUIRED_MSG = "Reboot to FIPS kernel required"
 SNAPD_DOES_NOT_HAVE_WAIT_CMD = (
     "snapd does not have wait command.\n"
     "Enabling Livepatch can fail under this scenario\n"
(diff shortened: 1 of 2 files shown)


uaclient/messages.py
```diff
@@ -102,6 +102,7 @@ class TxtColor:
 LIVEPATCH_LTS_REBOOT_REQUIRED = (
     "Livepatch support requires a system reboot across LTS upgrade."
 )
+FIPS_REBOOT_REQUIRED_MSG = "Reboot to FIPS kernel required"
```
renanrodrigo, line 105: I guess `FIPS_REBOOT_REQUIRED` should be enough

CalvoM:
Looks good to me, The change talks it well. Just a quick question, why do we have some messages in messages.py and some in status.py?

renanrodrigo (approved):
lgtm;
Is there a reason to keep that set of messages (Notices, Prompts, etc) in the status module, instead of moving them to messages? Not for this PR I guess

renanrodrigo:
@CalvoM was 25 seconds faster than me

uaclient/messages.py
lucasmoura: It should, but we have a `NamedMessage` with this name already and I don't want to further expand on this at this moment

lucasmoura:
@CalvoM the rationale is that I believe we should have a dedicated `notices` module in the future. I could move all of the notices into the `messages` module, but that would trigger more refactoring of the code. That's why I am not changing the notices in this PR

uaclient/messages.py
renanrodrigo: makes all sense

renanrodrigo:
So yes, I think prompts and notices are, before anything else, messages - and thus could live together in the messages module, as they were in status for instance, without any downside.
Also agreed that this PR now does not have to cover that.
:shipit:

PR #2014, opened 2022-03-28
fix: pin click version in dev-requirements

renanrodrigo:
Since https://github.com/psf/black/issues/2964 happened, this is needed until the new Black version is out and we start using it.

## Desired commit type

## Checklist


CalvoM (approved):
LGTM, actually had linting fail in some tests.

PR #2017, opened 2022-03-30
Add package counts for different components

renanrodrigo:
Add package counts based on archive components to security-status.
Show how many packages are installed from the different archive components: `main`, `universe`, etc

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/security_status.py
+++ b/uaclient/security_status.py
@@ -32,7 +32,7 @@ def get_origin_for_package(package: apt_package.Package) -> str:
     available_origins = package.installed.origins
     if len(package.installed.origins) == 1:
         if package.installed == package.candidate:
-            return "unavailable"
+            return "unknown"
         available_origins = package.candidate.origins
 
     for origin in available_origins:
@@ -128,7 +128,7 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
     ua_info = get_ua_info(cfg)
 
     summary = {"ua": ua_info}  # type: Dict[str, Any]
-    packages = []
+    updates = []
     cache = Cache()
 
     installed_packages = [package for package in cache if package.is_installed]
@@ -147,7 +147,7 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
         service_name, origin_site = get_service_name(candidate.origins)
         status = get_update_status(service_name, ua_info)
         update_count[service_name] += 1
-        packages.append(
+        updates.append(
             {
                 "package": candidate.package.name,
                 "version": candidate.version,
@@ -164,9 +164,7 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
     summary["num_installed_packages_third_party"] = package_count[
         "third-party"
     ]
-    summary["num_installed_packages_unavailable"] = package_count[
-        "unavailable"
-    ]
+    summary["num_installed_packages_unknown"] = package_count["unknown"]
     summary["num_esm_infra_packages"] = package_count["esm-infra"]
     summary["num_esm_apps_packages"] = package_count["esm-apps"]
     summary["num_esm_infra_updates"] = update_count["esm-infra"]
@@ -175,4 +173,4 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
         "standard-security"
     ]
 
-    return {"_schema_version": "0.1", "summary": summary, "packages": packages}
+    return {"_schema_version": "0.1", "summary": summary, "updates": updates}
--- a/uaclient/security_status.py
+++ b/uaclient/security_status.py
@@ -12,14 +12,11 @@ series = get_platform_info()["series"]
 
 ESM_SERVICES = ("esm-infra", "esm-apps")
 
-SERVICE_TO_ORIGIN_INFORMATION = {
-    "standard-security": ("Ubuntu", "{}-security".format(series)),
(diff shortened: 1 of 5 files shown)


CalvoM (approved):
LGTM and the unit tests passed

uaclient/security_status.py
```diff
@@ -34,17 +31,21 @@ class UpdateStatus(Enum):
...
+            (origin.origin, origin.archive), ""
+        )
+        if service in ESM_SERVICES:
+            esm_services.append(service)
     return esm_services
 
 
+def get_component_for_package(package: apt_package.Package) -> Optional[str]:
+    for origin in package.installed.origins:
+        if origin.origin == "Ubuntu":
```
lucasmoura, line 44: Let's just create a constant for this string

uaclient/security_status.py
```diff
@@ -156,6 +160,10 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
             }
         )
 
+    summary["num_installed_packages_main"] = package_count["main"]
+    summary["num_installed_packages_restricted"] = package_count["restricted"]
+    summary["num_installed_packages_universe"] = package_count["universe"]
+    summary["num_installed_packages_multiverse"] = package_count["multiverse"]
```
lucasmoura, line 166: I think we should be safe here and just change to a `get` call instead, just to guarantee that even on odd scenarios, we will  not error out because of a missing key

uaclient/tests/test_security_status.py
```diff
@@ -139,7 +141,7 @@ def test_finds_updates_for_installed_packages(
             mock_package(name="not_installed"),
             mock_package(
                 name="there_is_no_update",
-                installed=("1.0", "somewhere", "somehow", ""),
+                installed=("1.0", "somewhere", "Ubuntu", ""),
```
lucasmoura, line 144: Can we add a test with a component that is not `Ubuntu`, just to guarantee that the package does not enter the count

CalvoM (commented):
Just noticed that the sum of packages from main, multiverse, restricted and universe is not equal to installed_packages. 
Example I installed on focal lxd container and I got this 
```
{"_schema_version": "0.1", "summary": {"ua": {"attached": true, "enabled_services": ["esm-apps", "esm-infra"], "entitled_services": ["esm-apps", "esm-infra"]}, "num_installed_packages": 572, "num_installed_packages_main": 567, "num_installed_packages_restricted": 0, "num_installed_packages_universe": 0, "num_installed_packages_multiverse": 0, "num_esm_infra_packages": 0, "num_esm_apps_packages": 0, "num_esm_infra_updates": 0, "num_esm_apps_updates": 0, "num_standard_security_updates": 0}, "packages": []}
```
Upon running ubuntu-security-status I got :
```
572 packages installed, of which:
571 receive package updates with LTS until 4/2025
  1 package is no longer available for download

Packages that are not available for download may be left over from a
previous release of Ubuntu, may have been installed directly from a
.deb file, or are from a source which has been disabled.
For more information on the packages, run 'ubuntu-security-status
--unavailable'.
```
I assume the 1 package is the updated uaclient deb file I installed, but 4 packages are not recognised. Is this okay?

uaclient/security_status.py
renanrodrigo: It is a `defaultdict(int)`, so it will return `0` for any missing key. This is covered in the tests, given there are no packages mocked there to be from `universe` or `multiverse` for instance, and those keys are present in the output

uaclient/tests/test_security_status.py
renanrodrigo: There are a couple in this test - there are 5 installed packages, but only 2 of them are expected to be reported from `main` in the output (line 173 has an example)

renanrodrigo:
@CalvoM that's weird, I cannot reproduce it.
I do have a situation where some packages are left out of the count, because they come from sources that are not available anymore but have different versions in different repositories. We can check if that is the case, and act accordingly.

lucasmoura:
@renanrodrigo just noticed a count mismatch as well:
```
summary:
  num_esm_apps_packages: 0
  num_esm_apps_updates: 0
  num_esm_infra_packages: 0
  num_esm_infra_updates: 0
  num_installed_packages: 572
  num_installed_packages_main: 558
  num_installed_packages_multiverse: 0
  num_installed_packages_restricted: 0
  num_installed_packages_universe: 0
  num_standard_security_updates: 9
```

I think it could be caused by what you mentioned, but it is indeed worth double checking

renanrodrigo:
@lucasmoura yes, this one specifically falls to that case I mentioned.
if you run `apt upgrade`, those stale packages will be reinstalled, and the numbers will match again.

renanrodrigo:
@lucasmoura @CalvoM changed it in a way that a) makes sense from a package point of view and b) always sums up to the number of installed packages.
Packages with non-ubuntu sources are not ignored anymore, but counted as third-party
Packages for which there is no way to even guess a source are listed as unavailable

Ready for another round of reviews

lucasmoura:
@renanrodrigo we could document the new entries on the `--help` output of the command

renanrodrigo:
> @renanrodrigo we could document the new entries on the `--help` output of the command

Good idea. Will help to understand the data at least for now.

renanrodrigo:
Closed in favor of #2033 , which comes from the proper repository

PR #2023, opened 2022-04-04
Remove status from config

lucasmoura:
## Proposed Commit Message
Remove status logic from config

Currently, the UAConfig class is responsible for handling the status logic. However, if we want to make UAConfig a module to allow access  uaclient configuration, we need to remove that responsibility from the class.

## Test Steps
Run integration tests related to `ua status` somehow, like the PRO tests and the ones from `attach_validtoken` and `unattached_status`

## Desired commit type

## Checklist


lucasmoura:
Closing this in favor of #2038

PR #2024, opened 2022-04-05
CI: add awspro integration testing (SC-906)

paride:
## Proposed Commit Message

```
CI: add awspro integration testing:
```

## Desired commit type

## Checklist

### DIFF
--- a/.github/workflows/ci-integration.yaml
+++ b/.github/workflows/ci-integration.yaml
@@ -57,6 +57,9 @@ jobs:
       matrix:
         release: ['xenial', 'bionic', 'focal', 'jammy']
         platform: ['lxd']
+        include:
+          - release: bionic
+            platform: awspro
     steps:
       - name: Prepare test tools
         run: |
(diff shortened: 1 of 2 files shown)


paride:
I realized that the `cleanup-ec2` job doesn't check that credentials are available, as we do in the `integration-tests` job via:
```
  integration-tests:
    if: ${{ github.repository == 'canonical/ubuntu-advantage-client' }}
```
We can add the same to `cleanup-ec2`, or directly check if credentials are available. Something like (**untested**):
```
if: secrets.AWS_ACCESS_KEY_ID != ''
```
and also update `integration-tests` to do the same with `PYCLOUDLIB_CONFIG_CONTENTS`. I prefer this approach as:

1. does't reference the official repo directly
2. allows to run full CI on a fork if credentials are set

I implement this in a separate PR (or this one, if you prefer). Wdyt?

renanrodrigo (approved):
Thanks again for all effort here @paride 
I think the cleanup change (plus the potential check change) can be done separately.

LGTM

PR #2025, opened 2022-04-06
status: check if contract is updated when running status

CalvoM:
## Proposed Commit Message

status: check if contract is updated when running status

The user can get a notification of refreshing the contract when they run ua status.

## Desired commit type

## Checklist

### DIFF
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -494,3 +494,65 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
           | xenial  | disabled |
           | bionic  | disabled |
           | focal   | n/a      |
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attach and Check for contract change in status checking
+       Given a `<release>` machine with ubuntu-advantage-tools installed
+       When I attach `contract_token` with sudo
+       Then stdout matches regexp:
+       """
+       UA Infra: ESM enabled
+       """
+       And stdout matches regexp:
+       """
+       This machine is now attached to
+       """
+       And stdout matches regexp:
+       """
+       esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance \(ESM\)
+       """
+       When I create the file `/tmp/machine-token-overlay.json` with the following:
+       """
+       {
+           "machineTokenInfo": {
+               "contractInfo": {
+                   "resourceEntitlements": [
+                       {
+                           "type": "esm-apps",
+                           "entitled": false
+                       }
+                   ]
+               }
+           }
+       }
+       """
+       And I append the following on uaclient config:
+       """
+       features:
+         machine_token_overlay: "/tmp/machine-token-overlay.json"
+       """
+       When I run `ua status` with sudo
+       Then stdout matches regexp:
+       """
+       A change has been detected in your contract.
+       Please run `sudo ua refresh` to update your contract.
+       """
+       When I run `ua refresh contract` with sudo
+       Then stdout matches regexp:
+       """
+       Successfully refreshed your subscription.
+       """
+       When I run `sed -i '/^.*machine_token_overlay:/d' /etc/ubuntu-advantage/uaclient.conf` with sudo
+       And I run `ua status` with sudo
+       Then stdout does not match regexp:
+       """
+       A change has been detected in your contract.
+       Please run `sudo ua refresh` to update your contract.
+       """
+
+        Examples: ubuntu release livepatch status
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1565,7 +1565,17 @@ def action_status(args, *, cfg):
     show_beta = args.all if args else False
     token = args.simulate_with_token if args else None
     active_value = ua_status.UserFacingConfigStatus.ACTIVE.value
-
+    if cfg.is_attached:
+        try:
+            if contract.is_contract_changed(cfg):
+                cfg.add_notice("", messages.NOTICE_REFRESH_CONTRACT_WARNING)
+        except exceptions.UrlError as e:
+            with util.disable_log_to_console():
+                logging.error(
+                    messages.UPDATE_CHECK_CONTRACT_FAILURE.format(
+                        reason=str(e)
+                    )
+                )
     status, ret = actions.status(
         cfg, simulate_with_token=token, show_beta=show_beta
     )
@@ -1627,7 +1637,7 @@ def action_refresh(args, *, cfg: config.UAConfig):
 
     if args.target is None or args.target == "contract":
         _action_refresh_contract(args, cfg)
-
+        cfg.remove_notice("", messages.NOTICE_REFRESH_CONTRACT_WARNING)
     return 0
 
 
@@ -1637,7 +1647,6 @@ def configure_apt_proxy(
     """
     Handles setting part the apt proxies - global and uaclient scoped proxies
     """
-
     if scope == AptProxyScope.GLOBAL:
         http_proxy = cfg.global_apt_http_proxy
         https_proxy = cfg.global_apt_https_proxy
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -177,6 +177,30 @@ class UAContractClient(serviceclient.UAServiceClient):
             machine_token["activityInfo"] = response
             self.cfg.write_cache("machine-token", machine_token)
 
+    def get_updated_contract_info(
+        self,
+        machine_token: str,
+        contract_id: str,
+        machine_id: Optional[str] = None,
+    ) -> Dict[str, Any]:
+        """Get the updated machine token from the contract server.
+
+        @param machine_token: The machine token needed to talk to
+            this contract service endpoint.
+        @param contract_id: Unique contract id provided by contract service
+        @param machine_id: Optional unique system machine id. When absent,
+            contents of /etc/machine-id will be used.
+        """
+        if not machine_id:
+            machine_id = self._get_platform_data(machine_id).get(
+                "machineId", None
+            )
+        return self._get_updated_contract_info(
+            machine_token=machine_token,
+            contract_id=contract_id,
+            machine_id=str(machine_id),
+        )
+
     def _request_machine_token_update(
         self,
         machine_token: str,
@@ -196,9 +220,39 @@ class UAContractClient(serviceclient.UAServiceClient):
 
         @return: Dict of the JSON response containing refreshed machine-token
         """
+        data = self._get_platform_data(machine_id)
+        response = self._get_updated_machine_token_details(
+            data, machine_token, contract_id, detach
+        )
+        if not detach:
+            self.cfg.write_cache("machine-token", response)
+            util.get_machine_id.cache_clear()
+            machine_id = response.get("machineTokenInfo", {}).get(
+                "machineId", data.get("machineId")
+            )
+            self.cfg.write_cache("machine-id", machine_id)
+        return response
+
+    def _get_updated_machine_token_details(
+        self,
+        data: Dict,
+        machine_token: str,
+        contract_id: str,
+        detach: bool = False,
+    ) -> Dict[str, Any]:
+        """Get the updated machine token from the contract server.
+
+        @param data: specific machine platform data
+        @param machine_token: The machine token needed to talk to
+            this contract service endpoint.
+        @param contract_id: Unique contract id provided by contract service.
+        @param detach: Boolean set True if detaching this machine from the
+            active contract. Default is False.
+
+        @return: Dict of the JSON response containing refreshed machine-token
+        """
         headers = self.headers()
         headers.update({"Authorization": "Bearer {}".format(machine_token)})
-        data = self._get_platform_data(machine_id)
         data["activityInfo"] = self._get_activity_info()
         url = API_V1_TMPL_CONTEXT_MACHINE_TOKEN_RESOURCE.format(
             contract=contract_id, machine=data["machineId"]
@@ -212,13 +266,28 @@ class UAContractClient(serviceclient.UAServiceClient):
         response, headers = self.request_url(url, **kwargs)
         if headers.get("expires"):
             response["expires"] = headers["expires"]
-        if not detach:
-            self.cfg.write_cache("machine-token", response)
-            util.get_machine_id.cache_clear()
-            machine_id = response.get("machineTokenInfo", {}).get(
-                "machineId", data.get("machineId")
-            )
-            self.cfg.write_cache("machine-id", machine_id)
+        return response
+
+    def _get_updated_contract_info(
+        self, machine_token: str, contract_id: str, machine_id: str
+    ):
+        """Perform GET request to get the machine details from contract_id
+
+        @param machine_token: The machine token needed to talk to
+            this contract service endpoint.
+        @param contract_id: Unique contract id provided by contract service
+        @param machine_id: Unique system machine id.
+        """
+        headers = self.headers()
+        headers.update({"Authorization": "Bearer {}".format(machine_token)})
+        url = API_V1_TMPL_CONTEXT_MACHINE_TOKEN_RESOURCE.format(
+            contract=contract_id, machine=machine_id
+        )
+        response, headers = self.request_url(
+            url, method="GET", headers=headers
+        )
+        if headers.get("expires"):
+            response["expires"] = headers["expires"]
         return response
 
     def _get_platform_data(self, machine_id):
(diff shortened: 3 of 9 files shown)


uaclient/cli.py
```diff
@@ -1619,6 +1620,64 @@ def action_refresh(args, *, cfg: config.UAConfig):
     return 0
 
 
+def is_contract_changed(cfg: config.UAConfig) -> bool:
+    orig_token = cfg.machine_token
+    orig_entitlements = cfg.entitlements
+    machine_token = orig_token["machineToken"]
+    contract_id = orig_token["machineTokenInfo"]["contractInfo"]["id"]
```
lucasmoura, line 1627: Let's use `get` calls here to access the dict keys, just so we can be safe that status will not break if any of those keys is missing for some reason (even though this should be a corner case situation)

uaclient/cli.py
```diff
@@ -1553,7 +1553,8 @@ def action_status(args, *, cfg):
     show_beta = args.all if args else False
     token = args.simulate_with_token if args else None
     active_value = ua_status.UserFacingConfigStatus.ACTIVE.value
-
+    if is_contract_changed(cfg):
```
lucasmoura, line 1556: Maybe this call here should be performed inside a `try..except` block. My rationale is that if this function fails for some reason, we should still show the status output if we can.

What do you think @renanrodrigo @orndorffgrant ?

uaclient/cli.py
```diff
@@ -1619,6 +1620,64 @@ def action_refresh(args, *, cfg: config.UAConfig):
...
+    orig_entitlements = cfg.entitlements
+    machine_token = orig_token["machineToken"]
+    contract_id = orig_token["machineTokenInfo"]["contractInfo"]["id"]
+    contract_client = contract.UAContractClient(cfg)
+    resp = contract_client.get_updated_contract_info(
+        machine_token, contract_id
+    )
+    try:
+        resp_ = json.dumps(resp, cls=util.DatetimeAwareJSONEncoder)
+        resp = json.loads(resp_, cls=util.DatetimeAwareJSONDecoder)
```
lucasmoura, line 1634: Why are encoding the response to json and immediately decoding it back ? Is it because of the dates ?

uaclient/cli.py
```diff
@@ -1619,6 +1620,64 @@ def action_refresh(args, *, cfg: config.UAConfig):
...
+    except ValueError:
+        return False
+    if (
+        cfg.contract_expiry_datetime
+        != resp["machineTokenInfo"]["contractInfo"]["effectiveTo"]
```
lucasmoura, line 1639: Same get logic here

uaclient/cli.py
```diff
@@ -1619,6 +1620,64 @@ def action_refresh(args, *, cfg: config.UAConfig):
...
+    ):
+        return True
+    curr_entitlements = get_entitlements_from_response(resp)
+    for name, new_entitlement in sorted(curr_entitlements.items()):
+        deltas = util.get_dict_deltas(
+            orig_entitlements.get(name, {}), new_entitlement
+        )
+        if bool(deltas):
```
lucasmoura, line 1647: I think we can just do `if deltas` here

uaclient/cli.py
```diff
@@ -1619,6 +1620,64 @@ def action_refresh(args, *, cfg: config.UAConfig):
...
+        (e.get("type"), e)
+        for e in contractInfo.get("resourceEntitlements", [])
+    )
+    for entitlement_name, ent_value in ent_by_name.items():
+        entitlement_cfg = {"entitlement": ent_value}
+        if entitlement_name in tokens_by_name:
+            entitlement_cfg["resourceToken"] = tokens_by_name[entitlement_name]
+        util.apply_contract_overrides(entitlement_cfg)
+        entitlements[entitlement_name] = entitlement_cfg
+    return entitlements
```
lucasmoura, line 1678: This code is exactly what we do to process entitlements on the `config` module. Maybe we could modify the config module to receive a `machine_token` response. If we can do that, we would not need to recreate this function here, we would just compare the current `cfg.entitlements` with the `new_cfg.entitlements`.

What do you think @orndorffgrant @renanrodrigo ?

uaclient/tests/test_cli_status.py
```diff
@@ -965,6 +978,30 @@ def contract_info_side_effect(cfg, token):
 
         assert output["errors"][0]["message"] == warning_message
 
+    @pytest.mark.parametrize("contract_changed", (False, True))
```
lucasmoura, line 981: Let's also add unittest for the `is_contract_changed` function directly. Just to see if it covers the two scenarios we care about: date and entitlements changes

lucasmoura:
@CalvoM Regarding the status whole output, I see that it is now like this:
```
A change has been detected in your contract.
Please run `sudo ua refresh` to update your contract.

cc-eal        yes       disabled  Common Criteria EAL2 Provisioning Packages
cis           yes       disabled  Security compliance and audit tools
fips          yes       disabled  NIST-certified core packages
fips-updates  yes       disabled  NIST-certified core packages with priority security updates

                Account: UA Client Test
           Subscription: server-team-ua-client-ci-uaa
            Valid until: 9999-12-31 00:00:00+00:00
```

I think instead of putting that message on the top, we could instead create a `notice` for it. Then on every status call we remove that notice and let the code process the contract diffs again to see if it is still needed

uaclient/cli.py
renanrodrigo, line 1634: question: (I'm sorry, maybe I missed the point) what are we checking here? Is it expected that the response will not be a valid JSON? If so, we'd just return `False`?

uaclient/cli.py
```diff
@@ -1619,6 +1620,64 @@ def action_refresh(args, *, cfg: config.UAConfig):
...
+            return True
+    return False
+
+
+def get_entitlements_from_response(data: Dict) -> Dict:
+    machine_token = data
+    if not machine_token:
+        return {}
+
+    entitlements = {}
```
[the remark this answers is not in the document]
renanrodrigo: already pointed out some minutes ago :D

uaclient/cli.py
renanrodrigo: Already asked in the previous review as well

uaclient/cli.py
renanrodrigo: +1

uaclient/cli.py
renanrodrigo: +1 I had the same question

uaclient/cli.py
renanrodrigo: Also had put it in my review - there are indeed a couple ways to reuse the logic, and I'd prefer that. Maybe some optional `from_machine_token` arg would be nice, yes

uaclient/cli.py
CalvoM: Yes, it is because of the date part.

uaclient/cli.py
CalvoM: Yeah, nice catch.

uaclient/cli.py
CalvoM: Yeah, on it.

uaclient/cli.py
CalvoM: Makes sense.

uaclient/cli.py
CalvoM: True, it felt weird repeating the logic. Let me see how the suggestion will work out.

uaclient/cli.py
CalvoM: Should we just print to the log file or also show it to the client?

renanrodrigo: Just log it imho - If we could not detect any contract change for any reason (network? etc) then we just show the status. I'd be worried to inform the user that something happens only in a command like `refresh`, where there is expectation of changes to be applied

uaclient/cli.py
```diff
@@ -1553,7 +1553,14 @@ def action_status(args, *, cfg):
     show_beta = args.all if args else False
     token = args.simulate_with_token if args else None
     active_value = ua_status.UserFacingConfigStatus.ACTIVE.value
-
+    try:
+        if contract.is_contract_changed(cfg):
+            cfg.add_notice("", ua_status.NOTICE_REFRESH_CONTRACT_WARNING)
```
lucasmoura, line 1558: Let's add this notice on the messages module, since we already have a PR(#2023) that does that

uaclient/contract.py
```diff
@@ -492,3 +532,38 @@ def get_contract_information(cfg: UAConfig, token: str) -> Dict[str, Any]:
...
+        .get("contractInfo", {})
+        .get("id", "")
+    )
+    contract_client = UAContractClient(cfg)
+    resp = contract_client.get_updated_contract_info(
+        machine_token, contract_id
+    )
+    try:
+        resp_ = json.dumps(resp, cls=util.DatetimeAwareJSONEncoder)
+        resp = json.loads(resp_, cls=util.DatetimeAwareJSONDecoder)
```
lucasmoura, line 552: I think we can parse the date directly using the function from util `parse_rfc3339_date`. This will avoid this encoding/decoding call

uaclient/tests/test_cli_status.py
lucasmoura: @CalvoM Thanks for the additional tests, but I think we are still missing the `contract.is_contract_changed` tests

lucasmoura:
Another thing we can consider is to make that request to the contracts with a thread, since it is a blocking I/O operation and `ua status` is already taking too long to run in my opinion.

What do you think @orndorffgrant @renanrodrigo ?

uaclient/cli.py
```diff
@@ -1553,7 +1553,14 @@ def action_status(args, *, cfg):
     show_beta = args.all if args else False
     token = args.simulate_with_token if args else None
     active_value = ua_status.UserFacingConfigStatus.ACTIVE.value
-
+    try:
+        if contract.is_contract_changed(cfg):
```
lucasmoura, line 1557: Let's also only check this if the user is attached, otherwise this call will not do anything for the user

uaclient/cli.py
CalvoM: Alright.

orndorffgrant:
@lucasmoura @CalvoM 

RE `ua status` is taking too long. Yeah, can confirm. There's a couple things to do here that come to mind.
1. Rebasing on main will pull in the "remove unnecessary UAConfig creation` which will help a bit
2. This is currently performing a POST to the backend, which is a "write" operation and I don't think is necessary here. We should be doing a GET to this endpoint. See docs here: https://github.com/canonical/ua-contracts/blob/develop/docs/contracts.yaml#L4846= This might shave off some time as well.
3. Not for this PR, but the rest of the time seems to be spent calling "apt-cache policy". It appears to be called more times than is necessary. Also we might be able to speed that up by using `python-apt` instead of `exec`-ing the apt cli.

I'm not sure putting this in a thread would help that much, since we still want to complete it before exiting, and the processing that could be done in parallel isn't taking long. And that would add complexity to the code which i'm not sure is worth it.

I think after we fix (1) and (2), we can reevaluate.

features/attach_validtoken.feature
```diff
@@ -494,3 +494,73 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
+       """
+       UA Infra: ESM enabled
+       """
+       And stdout matches regexp:
+       """
+       This machine is now attached to
+       """
+       And stdout matches regexp:
+       """
+       SERVICE       ENTITLED  STATUS    DESCRIPTION
```
orndorffgrant, line 513: This will need a `+` as soon as realtime is merged
```suggestion
       SERVICE       +ENTITLED  STATUS    DESCRIPTION
```

uaclient/contract.py
```diff
@@ -177,6 +177,22 @@ def report_machine_activity(self):
...
+    def get_updated_contract_info(
+        self,
+        machine_token: str,
+        contract_id: str,
+        machine_id: str = None,
+        detach: bool = False,
+    ) -> Dict[str, Any]:
+        """Get the updated machine token from the contract server."""
+        data = self._get_platform_data(machine_id)
+        return self._get_updated_contract_info(
```
orndorffgrant, line 189: I think instead of reusing this method, we can just GET `API_V1_TMPL_CONTEXT_MACHINE_TOKEN_RESOURCE` right here. That function already serves too many purposes IMO and adjusting it to support a "GET" would make it even more complex.

uaclient/contract.py
```diff
@@ -492,3 +531,48 @@ def get_contract_information(cfg: UAConfig, token: str) -> Dict[str, Any]:
...
+    )
+    contract_client = UAContractClient(cfg)
+    resp = contract_client.get_updated_contract_info(
+        machine_token, contract_id
+    )
+    try:
+        resp_expiry = (
+            resp.get("machineTokenInfo", {})
+            .get("contractInfo", {})
+            .get("effectiveTo", "0001-01-01T00:00:00+00:00")
```
orndorffgrant, line 553: I think this default would make it so we never fall back to `cfg.contract_expiry_datetime`. Maybe we should default to `None` and check for None before trying to parse it?

uaclient/contract.py
```diff
@@ -492,3 +531,48 @@ def get_contract_information(cfg: UAConfig, token: str) -> Dict[str, Any]:
...
+        logging.error(str(e))
+        return False
+    except ValueError as e:
+        logging.error(str(e))
+        return False
+    new_expiry = (
+        resp.get("machineTokenInfo", {})
+        .get("contractInfo", {})
+        .get("effectiveTo", cfg.contract_expiry_datetime)
+    )
```
orndorffgrant, line 568: Can this be simplified to
```suggestion
    resp_expiry = (
        resp.get("machineTokenInfo", {})
        .get("contractInfo", {})
        .get("effectiveTo", None)
    )
    new_expiry = util.parse_rfc3339_date(resp_expiry) if resp_expiry else cfg.contract_expiry_datetime
```

uaclient/contract.py
orndorffgrant: I included this in my suggestion below

features/attach_validtoken.feature
CalvoM: It will only affect ENTITLED and not the other fields after?

uaclient/contract.py
CalvoM: Took me a minute here to understand, but I got it. Yes, makes sense.

orndorffgrant:
We may want to adjust this so that this actually does update the contract expiry date on `ua status`.
For entitlement changes, then we would still do the notice like this PR is currently doing

orndorffgrant (commented):
Just some nitpicks on the implementation :)

features/attach_validtoken.feature
```diff
@@ -494,3 +494,73 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
+       SERVICE       +ENTITLED  STATUS    DESCRIPTION
+       cc-eal       +yes      +<cc_status>     +Common Criteria EAL2 Provisioning Packages
+       """
+       And stdout matches regexp:
+       """
+       esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance \(ESM\)
+       fips         +yes      +<fips_status> +NIST-certified core packages
+       fips-updates +yes      +<fips_status> +NIST-certified core packages with priority security updates
+       livepatch    +yes      +n/a  +Canonical Livepatch service
+       """
```
orndorffgrant, line 522: Lets keep this simpler since this isn't the main intention of this test
```suggestion
       And stdout matches regexp:
       """
       esm-infra    +yes      +enabled  +UA Infra: Extended Security Maintenance \(ESM\)
       """
```

features/attach_validtoken.feature
```diff
@@ -494,3 +494,73 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
+       A change has been detected in your contract.
+       Please run `sudo ua refresh` to update your contract.
+       """
+
+        Examples: ubuntu release livepatch status
+           | release | fips_status | cc_status | cis_or_usg |
+           | xenial  | n/a         | disabled  | cis        |
+           | bionic  | disabled    | disabled  | cis        |
+           | focal   | disabled    | n/a       | usg        |
```
orndorffgrant, line 566: If we accept the above suggestion, then this can be simplified as well
```suggestion
           | release |
           | xenial  |
           | bionic  |
           | focal   |
```

uaclient/contract.py
```diff
@@ -212,13 +260,28 @@ def _request_machine_token_update(
...
+        @param machine_id: Optional unique system machine id. When absent,
+            contents of /etc/machine-id will be used.
+        """
+        headers = self.headers()
+        headers.update({"Authorization": "Bearer {}".format(machine_token)})
+        url = API_V1_TMPL_CONTEXT_MACHINE_TOKEN_RESOURCE.format(
+            contract=contract_id, machine=machine_id
+        )
+        kwargs = {"headers": headers, "method": "GET", "data": None}
+        response, headers = self.request_url(url, **kwargs)
```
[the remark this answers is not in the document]
CalvoM: Oh, yeah.

uaclient/contract.py
```diff
@@ -212,13 +260,28 @@ def _request_machine_token_update(
...
+    def _get_updated_contract_info(
+        self, machine_token: str, contract_id: str, machine_id: str
+    ):
+        """Perform GET request to get the machine details from contract_id
+
+        @param machine: The machine token needed to talk to
+            this contract service endpoint.
+        @param contract_id: Unique contract id provided by contract service
+        @param machine_id: Optional unique system machine id. When absent,
+            contents of /etc/machine-id will be used.
```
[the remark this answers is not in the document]
CalvoM: Oh, I used the description from [this description](https://github.com/canonical/ubuntu-advantage-client/blob/25b895b12d9ae16a49ba0e2d9bd3cc91a89b0d99/uaclient/contract.py#L41l). Maybe we could update all of them.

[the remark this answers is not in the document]
CalvoM: Oh, I see the difference, this is not optional like in the description I have referenced.

features/attach_validtoken.feature
CalvoM: Makes sense. I always thought that is some boilerplate that should always be there 😄😄

uaclient/contract.py
```diff
@@ -497,3 +566,38 @@ def get_contract_information(cfg: UAConfig, token: str) -> Dict[str, Any]:
...
+    resp = contract_client.get_updated_contract_info(
+        machine_token, contract_id
+    )
+    resp_expiry = (
+        resp.get("machineTokenInfo", {})
+        .get("contractInfo", {})
+        .get("effectiveTo", None)
+    )
+    new_expiry = (
+        util.parse_rfc3339_date(resp_expiry)
```
lucasmoura, line 590: I think we should double check if the code will not break if `resp_expiry` is `None`. Looking at `parse_rfc3339_date`, it seems that it will, so it is worth double checking

uaclient/contract.py
lucasmoura: @CalvoM Don't know if I am missing something, but do we still need to address this part here ?

lucasmoura:
@CalvoM Also, let's open this PR from an upstream branch directly, so we can have a full CI run on it

uaclient/contract.py
orndorffgrant: This is part of a one-line if statement python ternary thing (personally not a fan of this python-ism).
So in the next lines you can see that the function ill only be called `if resp_expiry`

lucasmoura: ohhh my mistake here, missed the ternary operation.

CalvoM:
@lucasmoura I already have the upstream branch created, let me create the PR. I did not do that.

CalvoM:
closing in favor of #2065.

PR #2028, opened 2022-04-06
lib: fix upgrade script for unsupported releases

lucasmoura:
## Proposed Commit Message
lib: fix upgrade script for unsupported releases

The upgrade_lts_script fails if it cannot find a codename for the current Ubuntu version it is running on. We are updating the script to properly finish if that situation happens

## Test Steps
Run the new unit test added

## Desired commit type

## Checklist

### DIFF
--- a/lib/upgrade_lts_contract.py
+++ b/lib/upgrade_lts_contract.py
@@ -34,7 +34,8 @@ version_to_codename = {
     "16.04": "xenial",
     "18.04": "bionic",
     "20.04": "focal",
-    "20.10": "groovy",
+    "21.10": "impish",
+    "22.04": "jammy",
 }
 
 current_codename_to_past_codename = {
@@ -42,6 +43,8 @@ current_codename_to_past_codename = {
     "bionic": "xenial",
     "focal": "bionic",
     "groovy": "focal",
+    "impish": "focal",
+    "jammy": "focal",
 }
 
 
@@ -58,7 +61,15 @@ def process_contract_delta_after_apt_lock() -> None:
     logging.debug(msg)
 
     current_version = parse_os_release()["VERSION_ID"]
-    current_release = version_to_codename[current_version]
+    current_release = version_to_codename.get(current_version)
+
+    if current_release is None:
+        msg = "Unable to get release codename for version: {}".format(
+            current_version
+        )
+        print(msg)
+        logging.warning(msg)
+        sys.exit(1)
 
     if current_release == "trusty":
         msg = "Unable to execute upgrade-lts-contract.py on trusty"
--- a/uaclient/tests/test_upgrade_lts_contract.py
+++ b/uaclient/tests/test_upgrade_lts_contract.py
@@ -67,6 +67,37 @@ class TestUpgradeLTSContract:
     )
     @mock.patch("lib.upgrade_lts_contract.parse_os_release")
     @mock.patch("lib.upgrade_lts_contract.subp")
+    def test_upgrade_cancel_when_current_version_not_supported(
+        self, m_subp, m_parse_os, m_is_attached, capsys, caplog_text
+    ):
+        m_parse_os.return_value = {"VERSION_ID": "NOT-SUPPORTED"}
+        m_subp.return_value = ("", "")
+
+        expected_msgs = [
+            "Starting upgrade-lts-contract.",
+            "Unable to get release codename for version: NOT-SUPPORTED",
+        ]
+        expected_logs = ["Check whether to upgrade-lts-contract"]
+        with pytest.raises(SystemExit) as execinfo:
+            process_contract_delta_after_apt_lock()
+
+        assert 1 == execinfo.value.code
+        assert 1 == m_is_attached.call_count
+        assert 1 == m_parse_os.call_count
+        assert 1 == m_subp.call_count
+        out, _err = capsys.readouterr()
+        assert out == "\n".join(expected_msgs) + "\n"
+        debug_logs = caplog_text()
+        for log in expected_msgs + expected_logs:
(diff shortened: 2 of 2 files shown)


renanrodrigo (commented):
How can we make sure this does not happen again? Would it be nice to come up with a guide on "what to change when a new release appears"?
(of course not for this PR)

lib/upgrade_lts_contract.py
```diff
@@ -35,13 +35,19 @@
     "18.04": "bionic",
     "20.04": "focal",
     "20.10": "groovy",
+    "21.04": "hirsute",
```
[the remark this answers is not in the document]
lucasmoura: I am adding it just for consistency sake, since we will not release the new version of the packages on those releases. But I can remove them if needed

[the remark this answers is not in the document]
lucasmoura: Regarding the test, We could start relying on `distro-info` to gather some of that info for us, but I don't think it shows what will be the next release to be installed easily. But we can perform some hacks that allow us to do that

orndorffgrant:
One more thing I think @blackboxsw mentioned that we might want here is a broad `try: except` to catch and log any unexpected errors without breaking things. Or maybe it's good that unexpected errors get sent to errors.ubuntu.com?

paride:
@lucasmoura CI failures because your branch has not been pushed to the project repository.

lucasmoura:
@orndorffgrant I am still unsure if we should do this, because it can actually mask real bugs we do need to care about. But we can better discuss this today

lib/upgrade_lts_contract.py
```diff
@@ -35,13 +35,19 @@
...
 }
 
 current_codename_to_past_codename = {
     "xenial": "trusty",
     "bionic": "xenial",
     "focal": "bionic",
     "groovy": "focal",
+    "hirsute": "groovy",
+    "impish": "hirsute",
+    "jammy": "hirsute",
```
orndorffgrant, line 50: @lucasmoura 

I think this is used to infer the release before the `do-release-upgrade`, and my understanding is that the current/near-future upgrade paths are:
* trusty -> xenial
* xenial -> bionic
* bionic -> focal
* focal -> impish
* impish -> jammy
* focal -> jammy

So I would expect the changes in this PR to look like this:
```suggestion
    "impish": "focal",
    "jammy": "impish",
    "jammy": "focal",
```

except of course we can't have two entries for jammy. Is there a way we can detect what the prior release was instead of inferring based on the current release?

lucasmoura: That's a good point. Since this runs at the end of `do-release-upgrade`, I don't think we have the original release info anymore. However, since the main idea is to apply any contract deltas between the releases, I think we can assume the `focal -> jammy` situation here, since no services are available for impish. What do you think ?

orndorffgrant: Yeah that's probably fine, and I suppose our `behave-upgrade-21.10` test should catch any errors that that assumption might create. So yes I think that makes sense, let's try it

lucasmoura: @orndorffgrant updated. I have also removed from the list the releases we no longer support as mentioned by @renanrodrigo

orndorffgrant:
+1 to "what to change when a new release appears?" document @renanrodrigo 
@lucasmoura can we move this to be in the canonical repo so we get CI?

lib/upgrade_lts_contract.py
```diff
@@ -58,7 +61,15 @@ def process_contract_delta_after_apt_lock() -> None:
     logging.debug(msg)
 
     current_version = parse_os_release()["VERSION_ID"]
-    current_release = version_to_codename[current_version]
+    current_release = version_to_codename.get(current_version)
+
+    if current_release is None:
```
orndorffgrant, line 66: Should we have a similar check down for past_release?

lib/upgrade_lts_contract.py
```diff
@@ -34,14 +34,17 @@
...
-    "20.10": "groovy",
+    "21.10": "impish",
+    "22.04": "jammy",
 }
 
 current_codename_to_past_codename = {
     "xenial": "trusty",
     "bionic": "xenial",
     "focal": "bionic",
     "groovy": "focal",
```
orndorffgrant, line 45: We can probably remove this line

lib/upgrade_lts_contract.py
```diff
@@ -34,14 +34,17 @@
...
+    "impish": "focal",
+    "jammy": "focal",
```
orndorffgrant, line 47: We should leave a comment here explaining that this works even when upgrading from impish -> jammy because there won't be anything enabled on impish anyway.

The output of behave-upgrade-21.10 looks good.

I wonder if we should adjust behave-upgrade-20.04 to upgrade to jammy? and test that too

lib/upgrade_lts_contract.py
```diff
@@ -58,7 +61,15 @@ def process_contract_delta_after_apt_lock() -> None:
     logging.debug(msg)
```
orndorffgrant, line 61: Let's fix the typo just above here "Skiping" -> "Skipping"

lib/upgrade_lts_contract.py
orndorffgrant: > I wonder if we should adjust behave-upgrade-20.04 to upgrade to jammy? and test that too

I just tried and it doesn't look like this is allowed yet AFAICT

lucasmoura: I think I manager to solve this, let me know what you think

lucasmoura:
Closing this in favor of #2049

PR #2030, opened 2022-04-06
Make UA less Trusty

renanrodrigo:
Maintenance work - removing 14.04 (Trusty) specific checks and tests from the codebase.
As we don't release to Trusty anymore, those are not needed.

## Desired commit type

## Checklist


paride:
@renanrodrigo CI failures because your branch has not been pushed to the project repository.

debian/ubuntu-advantage-tools.postinst
```diff
@@ -4,14 +4,6 @@ set -e
```
orndorffgrant, line 4: I think all trusty things in postinst need to stay, because until trusty is well past end of ESM, we could still have people upgrading from trusty to later releases

renanrodrigo: I kept there the data which relates to upgrades - things as `ESM_INFRA_OLD_APT_SOURCE_FILE_TRUSTY` - so the `postinst` can change stuff when you leave Trusty.
Those two checks I removed expect that you are installing on Trusty (e.g. the current `VERSION_ID` is `14.04`). When updating, you will be in a different version anyway

orndorffgrant: Ah excellent point, I'll take a closer look

orndorffgrant (approved):
Looks good, thanks for finding all these places to remove extra trusty logic!

renanrodrigo:
@orndorffgrant I think this is everything to throw away by now :D

renanrodrigo:
Superseded by #2043 let's see CI suffer

PR #2033, opened 2022-04-07
Add package counts for different components

renanrodrigo:
Add package counts based on archive components to security-status.
Show how many packages are installed from the different archive components: `main`, `universe`, etc

## Desired commit type

## Checklist

### DIFF
--- a/features/schemas/ua_security_status.json
+++ b/features/schemas/ua_security_status.json
@@ -31,6 +31,24 @@
         "num_installed_packages": {
           "type": "integer"
         },
+        "num_installed_packages_main": {
+          "type": "integer"
+        },
+        "num_installed_packages_multiverse": {
+          "type": "integer"
+        },
+        "num_installed_packages_restricted": {
+          "type": "integer"
+        },
+        "num_installed_packages_universe": {
+          "type": "integer"
+        },
+        "num_installed_packages_third_party": {
+          "type": "integer"
+        },
+        "num_installed_packages_unavailable": {
+          "type": "integer"
+        },
         "num_esm_infra_packages": {
           "type": "integer"
         },
--- a/features/schemas/ua_security_status.json
+++ b/features/schemas/ua_security_status.json
@@ -46,7 +46,7 @@
         "num_installed_packages_third_party": {
           "type": "integer"
         },
-        "num_installed_packages_unavailable": {
+        "num_installed_packages_unknown": {
           "type": "integer"
         },
         "num_esm_infra_packages": {
@@ -66,7 +66,7 @@
         }
       }
     },
-    "packages": {
+    "updates": {
       "type": "array",
       "items": {
         "type": "object",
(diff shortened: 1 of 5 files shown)


features/schemas/ua_security_status.json
```diff
@@ -48,7 +66,7 @@
         }
       }
     },
-    "packages": {
+    "updates": {
```
orndorffgrant, line 69: This is a major breaking change unfortunately. We would either need to be sure no-one is using this command before we make a change like this, Or we'd need to implement schema versioning for the command

uaclient/security_status.py
```diff
@@ -156,6 +157,14 @@ def security_status(cfg: UAConfig) -> Dict[str, Any]:
             }
         )
 
+    summary["num_installed_packages_main"] = package_count["main"]
```
orndorffgrant, line 160: `num_installed_packages_main` follows a different naming pattern from the currently existing `num_esm_infra_packages` - should we change these to `num_main_packages` and similar for all new fields?

uaclient/security_status.py
```diff
@@ -31,18 +28,23 @@ class UpdateStatus(Enum):
...
-            "esm-apps"
-        ]:
-            esm_services.append("esm-apps")
-    return esm_services
+def get_origin_for_package(package: apt_package.Package) -> str:
+    available_origins = package.installed.origins
+    if len(package.installed.origins) == 1:
+        if package.installed == package.candidate:
+            return "unknown"
+        available_origins = package.candidate.origins
```
orndorffgrant, line 36: These special cases for when the installed origins == 1 aren't obvious to someone unfamiliar with the apt internals. Can you leave a comment explaining the "unknown" case and the "available_origins" special case?

features/schemas/ua_security_status.json
renanrodrigo: There are, right now, more changes requested which will probably break the structure a couple times, until we get the response to a "stable" state...
We do have a `schema_version` field though. Should we just update that?
Or do you think we should by now already be thinking of retrocompatibility?

uaclient/security_status.py
renanrodrigo: I'd go with the opposite, something like `num_installed_packages_esm_infra`

uaclient/security_status.py
renanrodrigo: Sure

features/schemas/ua_security_status.json
orndorffgrant: Because we removed the `--beta` flag in 27.7, I think we need to be backwards compatible from now on.
So any major change in schema like changing key names will require a new separate schema and we'll need to support both. We can default to the old schema and opt-in to the new with a `--version=2` flag, for example.

uaclient/security_status.py
orndorffgrant: We can do that, and it would also be a major schema change, but if we're doing a major schema upgrade, then we might as well make any changes that we want to at once.

renanrodrigo:
@lucasmoura @orndorffgrant @CalvoM only had time to tackle the `--version` flag, up for review... and feel free to add more commits to this branch, no need to wait for me :upside_down_face:

orndorffgrant:
Closed in favor of #2069

PR #2037, opened 2022-04-07
Remove unnecessary creation of UAConfig instances

lucasmoura:
## Proposed Commit Message
Remove unnecessary creation of UAConfig instances

Creating an UAConfig instance can be an expansive operation that can be avoided if we share the instance we create for every cli call throught the code. We are refactoring the code to allow that reuse of the instance when possible

I have tested this work to see if we would get any speedup when running some UA commands:
`ua status` from around 2.5s to 2.0s
`ua attach` form around 17s to 16.2s

Tests were performed with a bionic LXD container running both commands around 3 times using the latest package of UA and another with the config refactor applied.

## Test Steps
Run the integration steps for attach/enable/disable/deach and verify that the package doesn't break and that it still work as expected

## Desired commit type

## Checklist

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -60,7 +60,9 @@ redact_ubuntu_release_from_ua_apt_filenames() {
     DIR=$1
     UA_SERVICES=$(/usr/bin/python3 -c "
 from uaclient.entitlements import valid_services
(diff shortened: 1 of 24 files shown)


orndorffgrant:
There are what look like legitimate CI failures, though I'm not sure how these changes could have caused them - probably worth a closer look

lucasmoura:
@orndorffgrant I have fixed the test, thanks for spotting it

PR #2038, opened 2022-04-07
Remove status from config

lucasmoura:
## Proposed Commit Message
Remove status logic from config

Currently, the UAConfig class is responsible for handling the status logic. However, if we want to make UAConfig a module to allow access  uaclient configuration, we need to remove that responsibility from the class.

## Test Steps
Run integration tests related to `ua status` somehow, like the PRO tests and the ones from `attach_validtoken` and `unattached_status`

## Desired commit type

## Checklist


uaclient/event_logger.py
```diff
@@ -41,6 +42,29 @@ class EventLoggerMode(enum.Enum):
...
+            service
+            for service in status.get("services", [])
+            if service.get("available", "yes") == "yes"
+        ]
+        status["services"] = available_services
+
+    # We don't need the origin info in the json output
+    status.pop("origin", "")
+
+    return status
```
orndorffgrant, line 65: I expected this function to be in the `status` module. IMO it's not so much "formatting" as it is adding extra status information. What do you think?

orndorffgrant (commented):
Just one comment. And the CI failure appears unrelated (related to staging contract changes for jammy)

uaclient/event_logger.py
lucasmoura: We could do that, but right now this will generate a circular dependency between `status` and `event_logger`. Which we could solve by adding the import directly on the event_logger function that requires the function.
Since we will probably do a major refactoring of how the event_logger work in the near future, I would not change this right now, but let me know what you think

orndorffgrant: I see.

I guess I think this function should be called at the end of `uaclient.status.status` and `uaclient.status.simulate_status` instead of in the event_logger.

But I can see how that refactor could be considered out of scope, so not needed to merge this

orndorffgrant (approved):
Needs a rebase to resolve conflicts but looks good!

uaclient/event_logger.py
lucasmoura: Agreed

PR #2042, opened 2022-04-07
readme: add logo and badges

orndorffgrant:
Trying to make our readme look prettier :upside_down_face: 

The preview in the diff doesn't do a good job. See what it looks like here: https://github.com/orndorffgrant/ubuntu-advantage-client/blob/logo-readme/README.md


CalvoM:
On a light note, I thought we would have an A like the avengers, 😆

renanrodrigo (approved):

PR #2043, opened 2022-04-08
Make UA less Trusty

renanrodrigo:
Maintenance work - removing 14.04 (Trusty) specific checks and tests from the codebase.
As we don't release to Trusty anymore, those are not needed.

The demo directory was also removed.

## Desired commit type

## Checklist


debian/ubuntu-advantage-tools.postinst
```diff
@@ -4,14 +4,6 @@ set -e
 
 . /etc/os-release  # For VERSION_ID
```
blackboxsw, line 5: While we are dropping this content, I recognize that we have an expectation that both VERSION_ID and UBUNTU_CODENAME shall always come from /etc/os-release. Maybe we should be defensive about or expectations because we are exposed to two types of errors when UBUNTU_CODENAME is unset:
  1. unintended silent successes from `check_is_lts` due to it's use of grep -q "" 
  2. to an error in `check_is_active_esm` if invalid or empty series is specified
     `ubuntu-distro-info: invalid distribution series ''`

We have the same type of exposure to silent successes if VERSION_ID is unset thoughout the code. I'd prefer if we are explicit and loud in our errors with exit 1 from postinst rather than silently succeeding but not setting up ESM properly. The latter situation could result in an unpatched and insecure system as ESM.

I realize that exit 1 from our postinst is bad and breaks unattended-upgrades etc. So we'd have to weigh whether we want the bug here or not. 
```suggestion
. /etc/os-release  # For VERSION_ID and UBUNTU_CODENAME
if [ -z "${VERSION_ID}" -o -z "${UBUNTU_CODENAME}" ]; then
    echo "Error: missing either VERSION_ID or UBUNTU_CODENAME in /etc/os-release: VERSION_ID=${VERSION_ID}  UBUNTU_CODENAME=${UBUNTU_CODENAME}" 1>&2
    exit 1
elif
```

blackboxsw (commented):
Thanks for this @renanrodrigo just a checkin partway as I've gotten through only d/rules and d/*postinst. I'll wrap up the rest on Monday.

blackboxsw (commented):
While trying to test this branch I re-ran through do-release-upgrade from trusty -> xenial and it seems we have a problem with not re-enabling esm-infra from trusty -> xenial.
https://github.com/canonical/ubuntu-advantage-client/issues/2046

I got a bit distracted (unrelated to this PR) but it is something we still need to sort in Xenial potentially as upgrade_lts_contract.py should only run xenial's version of this script which might be able to remedy this situation.

renanrodrigo:
@blackboxsw There is an extra step needed for this upgrade to work, as per https://github.com/canonical/ubuntu-advantage-client/issues/1590.
Do you think this is a problem worth solving, past the documentation line in discourse?

debian/ubuntu-advantage-tools.postinst
renanrodrigo: It's fine to have no `VERSION_ID`, because it is used throughout the postinst to perform specific things to `xenial` - it is not even a silent fail, as Xenial will always define the `VERSION_ID` - it is success.
For 1. : This is true. I am changing the script to set the  `UBUNTU_CODENAME` to `UNKNOWN` if not found - all our LTSs should define it.
For 2. : This is not a problem if 1 is not a problem, because  `check_is_active_esm` is only executed if `check_is_lts` passes

debian/ubuntu-advantage-tools.postinst
```diff
@@ -2,20 +2,22 @@
...
 
+if [ -z "${VERSION_ID}" ]; then
+    echo "Warning: missing VERSION_ID in /etc/os-release" >&2
+    VERSION_ID="NO-VERSION_ID"
+fi
+if [ -z "${UBUNTU_CODENAME}" ]; then
+    echo "Warning: missing UBUNTU_CODENAME in /etc/os-release" >&2
+    UBUNTU_CODENAME="NO-UBUNTU_CODENAME-$VERSION_ID"
+fi
+
```
orndorffgrant, line 20: @blackboxsw Please take a look. This essentially does the same as what Renan had, but also warns about the situation. I think this is a good middleground, because an early exit might prevent other important postinst tasks from running that don't require these variables. LMK what you think

debian/ubuntu-advantage-tools.postinst
```diff
@@ -58,12 +58,9 @@ MACHINE_TOKEN_FILE="/var/lib/ubuntu-advantage/private/machine-token.json"
...
     DIR=$1
-    UA_SERVICES=$(/usr/bin/python3 -c "
-from uaclient.entitlements import valid_services
-from uaclient.config import UAConfig
-cfg = UAConfig()
-print(*valid_services(cfg=cfg, allow_beta=True, all_names=True), sep=' ')
-")
+    # It is okay if this list is outdated, because this function is only used for an old migration.
+    # Any services that were introduced after this migration was added won't need to be migrated.
+    UA_SERVICES="cc-eal cis esm-infra esm-apps fips fips-updates livepatch ros ros-updates"
```
orndorffgrant, line 63: @blackboxsw one more thing that I added that needs your stamp of approval :) I think hardcoding this list is fine because thi migration will only ever need to run for services that existed prior to the migration. This hardcoded list should be more than enough I think

blackboxsw:
> @blackboxsw There is an extra step needed for this upgrade to work, as per #1590. Do you think this is a problem worth solving, past the documentation line in discourse?

Sorry this is good given that we already have it documented. No need to fix this aspect.

blackboxsw (commented):
LGTM. one minor comment  for the refreesh-keyrings.sh
The other thing I find value in is having a tome of SRU manual scripts available under `sru` folder for ease of reference/grepping in the future as new manual tests get created or someone wants to one-off test something. It has been helpful for cloud-init to host all previous  SRU scripts under [ubuntu-sru/bugs](https://github.com/cloud-init/ubuntu-sru/tree/main/bugs) because we have frequently looked back at some of those examples when explaining how to set up something with KVM/LXD etc or when writing automated tests.

I could see the grooming of the SRU scripts directory as keeping things lean. But some library of previous tests somewhere as a historical reference for test setup/instrumentation may come in handy if in the event that new manual scripts need to be writting for next SRU which don't easily fit into behave tests or pycloudlib use-cases yet.

tools/refresh-keyrings.sh
```diff
@@ -9,10 +9,6 @@
 #
 # N.B. This will rename any existing keyrings with the suffix .old.
 
-# NOTE: If replacing keyrings on services that are intended for trusty, the
```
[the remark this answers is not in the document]
orndorffgrant: Ah excellent catch, yes this message is still applicable, just needs to be rewritten generically. Will do

orndorffgrant:
@blackboxsw I think I agree about keeping the old sru scripts, I'll add them back

orndorffgrant:
@blackboxsw Updated to bring back the SRU tests and adjusted the refresh-keyrings.sh comment. LMK what you think

blackboxsw (approved):
LGTM! +1 on this set of cleanup.

lucasmoura (commented):
LGTM, but I think we can still change more items here:

* `upgrade_lts_contract` and `reboot_cmds`: Our main docstring mentions upgrading from Trusty to Xenial, we can update that.
* `test_cc`: We can drop an specific trusty test there

orndorffgrant:
@lucasmoura Thanks!

For upgrade_lts_contract, I think the comment is still appropriate because upgrading from trusty to xenial is still a situation where that script would run and I think the example is still valid.

For test_cc.py - excellent catch, just updated and pushed

lucasmoura (approved):

orndorffgrant:
That last CI failure passed locally. Merging now

PR #2044, opened 2022-04-08
daemon: cloud-id file shim

orndorffgrant:
## Proposed Commit Message

```
daemon: cloud-id file shim

This introduces a new systemd service that will run on startup only if
the version of cloud-init does not create the cloud-id file. The service
creates the same cloud-id file that cloud-init does in newer versions.
This is specifically necessary for xenial, and is a noop if cloud-init
is new enough.

It also runs the same script on postinst to cover the scenario where the
new version of cloud-init is installed but the instance has not
rebooted, and so cloud-init hasn't yet created the cloud-id file.
```

## Test Steps

run the edited integration tests

## Desired commit type

## Checklist

### DIFF
--- a/lib/cloud-id-shim.sh
+++ b/lib/cloud-id-shim.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+RUN_CLOUD_INIT=/run/cloud-init
+if [ ! -d "$RUN_CLOUD_INIT" ]; then
+  exit 0
+fi
+
+CLOUD_ID_LINK_PATH=$RUN_CLOUD_INIT/cloud-id
+if [ -L "$CLOUD_ID_LINK_PATH" ]; then
+  exit 0
+fi
+
+CLOUD_ID=$(cloud-id 2>/dev/null) || CLOUD_ID=""
+if [ -z "$CLOUD_ID" ]; then
+  exit 0
+fi
+
+CLOUD_ID_FILE_PATH=$RUN_CLOUD_INIT/cloud-id-$CLOUD_ID
+
+echo "$CLOUD_ID" > "$CLOUD_ID_FILE_PATH"
+ln -s "$CLOUD_ID_FILE_PATH" "$CLOUD_ID_LINK_PATH"
+echo "Created symlink $CLOUD_ID_LINK_PATH -> $CLOUD_ID_FILE_PATH." >&2
+
+exit 0
--- a/systemd/ubuntu-advantage.service
+++ b/systemd/ubuntu-advantage.service
@@ -9,7 +9,7 @@
 [Unit]
 Description=Ubuntu Advantage GCP Auto Attach Daemon
 Documentation=man:ubuntu-advantage https://ubuntu.com/advantage
-After=network.target network-online.target systemd-networkd.service ua-auto-attach.service cloud-config.service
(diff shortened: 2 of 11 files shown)


lucasmoura (approved):

systemd/ubuntu-advantage.service
```diff
@@ -9,7 +9,7 @@
 [Unit]
 Description=Ubuntu Advantage GCP Auto Attach Daemon
 Documentation=man:ubuntu-advantage https://ubuntu.com/advantage
-After=network.target network-online.target systemd-networkd.service ua-auto-attach.service cloud-config.service
+After=network.target network-online.target systemd-networkd.service ua-auto-attach.service cloud-config.service ubuntu-advantage-cloud-id-shim.service
```
blackboxsw, line 12: +1 I was thinking there'd be a race if we didn't add After=ubuntu-advantage-cloud-id-shim.service to ubuntu-advantage.service.

lib/cloud-id-shim.sh
```diff
@@ -0,0 +1,23 @@
...
+
+CLOUD_ID=$(cloud-id 2>/dev/null) || CLOUD_ID=""
+if [ -z "$CLOUD_ID" ]; then
+  exit 0
+fi
+
+CLOUD_ID_FILE_PATH=$RUN_CLOUD_INIT/cloud-id-$CLOUD_ID
+
+echo "$CLOUD_ID" > "$CLOUD_ID_FILE_PATH"
+ln -s "$CLOUD_ID_FILE_PATH" "$CLOUD_ID_LINK_PATH"
```
blackboxsw, line 21: Is it worth a message "Creating symlink X->Y" so we have record in `systemctl status ubuntu-advantage-cloud-id-shim.service` that the symlink was actually created? It only makes it easier to confirm that shim actually did something up in systemctl (even though we could just go and ls /run/cloud-init/cloud-id*)

blackboxsw (commented):
LGTM, minor thought/question for future triage.

lib/cloud-id-shim.sh
```diff
@@ -0,0 +1,24 @@
...
+echo "Created symlink $CLOUD_ID_LINK_PATH -> $CLOUD_ID_FILE_PATH." >&2
```
orndorffgrant, line 22: @blackboxsw Just want to make sure that this is what you were expecting before we merge this. Note that this also gets printed during postinst on xenial.

blackboxsw (approved):
LGTM I think the notification
```bash
 ...
Setting up ubuntu-advantage-tools (28.0) ...
Installing new version of config file /etc/ubuntu-advantage/help_data.yaml ...
Created symlink /run/cloud-init/cloud-id -> /run/cloud-init/cloud-id-lxd.
Processing triggers for man-db (2.7.5-1ubuntu0.1~esm1) ...


# across reboot
Apr 15 16:27:01 dev-x systemd[1]: Starting cloud-id shim...
Apr 15 16:27:01 dev-x sh[848]: Created symlink /run/cloud-init/cloud-id -> /run/
Apr 15 16:27:01 dev-x systemd[1]: Started cloud-id shim.

```

orndorffgrant:
After updating the gcp pro ids, CI passed. Merging now

PR #2049, opened 2022-04-12
lib: fix upgrade script for unsupported releases

lucasmoura:
## Proposed Commit Message
lib: fix upgrade script for unsupported releases

The upgrade_lts_script fails if it cannot find a codename for the current Ubuntu version it is running on. We are updating the script to properly finish if that situation happens

## Test Steps
Run the new unit test added

## Desired commit type

## Checklist


lucasmoura:
@orndorffgrant I think you fix is better in the long run, I will drop my commit

orndorffgrant (approved):
Tests passed for me and looks good!

PR #2056, opened 2022-04-19
fix some integration tests caused by status change

lucasmoura:


features/ubuntu_upgrade.feature
```diff
@@ -52,6 +52,10 @@ Feature: Upgrade between releases when uaclient is attached
     Scenario Outline: Attached upgrade across LTS releases
         Given a `<release>` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
+        # update-manager-core requires ua < 28. Our tests that build the package will
+        # generate ua with version 28. We are removing that package here to make sure
+        # do-release-upgrade will be able to run
+        And I run `apt remove update-manager-core -y` with sudo
```
orndorffgrant, line 58: This is unideal, but fine for now. We should meet with the maintainers of this package to see if we can work something out that is better than this.

orndorffgrant (approved):

PR #2057, opened 2022-04-20
cli: add refresh subcommand to update motd message

lucasmoura:
## Proposed Commit Message
cli: add refresh subcommand to update motd message

Right now, users have no direct way to update the MOTD messages UA deliver. For example, if a contract is renewed, users will need to wait for our timer job to run so they can see the expiry message on MOTD disappear. We are now adding a `ua refresh motd` command that updates the MOTD message

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/docs/howtoguides/update_motd_messages.md
+++ b/docs/howtoguides/update_motd_messages.md
@@ -1,0 +1,12 @@
+# How to update MOTD and APT messages
+
+Since ubuntu-advantage-tools is responsible for enabling ESM services, we advertise them on different
+applications thorough the system, such as MOTD and apt commands like upgrade.
+
+To verify that APT and MOTD message is advertising the ESM packages, ensure that we have ESM
+source list files in the system. If that is the case, please run the following command to
+update the state of MOTD and APT messages:
+
+```sh
+ua refresh messages
+```
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -56,7 +56,10 @@ from uaclient.entitlements.entitlement_status import (
 # It is not ideal for us to import an entitlement directly on the cli module.
 # We need to refactor this to avoid that type of coupling in the code.
 from uaclient.entitlements.livepatch import LIVEPATCH_CMD
-from uaclient.jobs.update_messaging import update_apt_and_motd_messages
+from uaclient.jobs.update_messaging import (
+    refresh_motd,
+    update_apt_and_motd_messages,
+)
 
 NAME = "ua"
 
@@ -465,7 +468,7 @@ def refresh_parser(parser):
     parser._optionals.title = "Flags"
     parser.add_argument(
         "target",
-        choices=["contract", "config"],
+        choices=["contract", "config", "messages"],
         nargs="?",
         default=None,
         help=(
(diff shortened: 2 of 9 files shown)


uaclient/cli.py
```diff
@@ -1623,6 +1628,15 @@ def _action_refresh_contract(_args, cfg: config.UAConfig):
     print(messages.REFRESH_CONTRACT_SUCCESS)
 
 
+def _action_refresh_motd(_args, cfg: config.UAConfig):
+    # Not performing any exception handling here since both of these
+    # functions should raise UserFacingError exceptions, which are
+    # covered by the main_error_handler decorator
+    update_apt_and_motd_messages(cfg)
+    refresh_motd()
+    print(messages.REFRESH_MOTD_SUCCESS)
```
lucasmoura, line 1637: I am still unsure if we should have any type of exception handling here

CalvoM: Maybe, the only time I have run to an exception is when I compiled the package in a different distro i.e. jammy and ran it in focal, where glibc version was not correct.

CalvoM:
After testing, looks good. The only thing to talk about is the exception handling that Lucas was talking about.

uaclient/jobs/update_messaging.py
```diff
@@ -352,3 +357,21 @@ def update_apt_and_motd_messages(cfg: config.UAConfig) -> bool:
     # Now that we've setup/cleanedup templates render them with apt-hook
     util.subp(["/usr/lib/ubuntu-advantage/apt-esm-hook", "process-templates"])
     return True
+
+
+def refresh_motd():
```
orndorffgrant, line 362: We might be able to just use `systemctl start motd-news.service`, which is the systemd service that get's run periodically by a timer to update the motd. That might be more reliable than looking for these commands but idk

uaclient/cli.py
orndorffgrant: I don't think we need to worry about jammy packages erroring on focal, that makes sense I think.

@lucasmoura did you mean to modify `update_apt_and_motd_messages` to raise a UserFacingError with the `REFRESH_MOTD_FAILURE` message (a test below mocks that, but I don't see it happening in the actual code)?

We could also catch anything here and then raise a `UserFacingError(REFRESH_MOTD_FAILURE)` and log the exception itself to the log file.

lucasmoura: Yes, I was originally intending to raise that exception on `update_apt_and_motd_messages`, but I abandoned that idea. But I think your idea is a good one, to catch any exception and raise an specific one

uaclient/jobs/update_messaging.py
lucasmoura: I wasn't able to completely remove the `update-notifier` command, since restarting the service doesn't run it apparently. But I replaced the `run-parts` one. Thanks for the suggestion @orndorffgrant

orndorffgrant: Looking closer now - why do we need to run the update-notifier script? It should run after any `apt update` (see /etc/apt/apt.conf.d/99update-notifier), so it shouldn't get out of date like our messages could, right?

orndorffgrant (commented):
Can you also update the MOTD Messages section in CONTRIBUTING.md with the new command? This would also be a good time to move that section to its own "howtoguide" under `contributing-docs/`

lucasmoura:
@orndorffgrant code updated

docs/howtoguides/update_motd_messages.md
```diff
@@ -0,0 +1,12 @@
+# How to update MOTD and APT messages
```
orndorffgrant, line 1: I think it'd be nice to add a link to this in the table of contents in the readme as well - what do you think @lucasmoura ?

lucasmoura:
@orndorffgrant done

orndorffgrant (approved):
LGTM! Just awaiting CI one more time (ignoring expected gcp failure)

PR #2060, opened 2022-04-22
WIP: Jammy integration test updates

orndorffgrant:
Now that esm and livepatch are allowed for jammy, we can update many integration tests to check that.


features/attach_validtoken.feature
```diff
@@ -166,10 +164,11 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Advantage
...
         Examples: ubuntu release packages
-           | release | downrev_pkg                 | cc_status | cis_or_usg |
-           | xenial  | libkrad0=1.13.2+dfsg-5      | disabled  | cis        |
-           | bionic  | libkrad0=1.16-2build1       | disabled  | cis        |
-           | focal   | hello=2.10-2ubuntu2         | n/a       | usg        |
+           | release | downrev_pkg                 | cc_status | cis_or_usg | cis      | fips     | livepatch_desc                |
+           | xenial  | libkrad0=1.13.2+dfsg-5      | disabled  | cis        | disabled | disabled | Canonical Livepatch service   |
+           | bionic  | libkrad0=1.16-2build1       | disabled  | cis        | disabled | disabled | Canonical Livepatch service   |
+           | focal   | hello=2.10-2ubuntu2         | n/a       | usg        | disabled | disabled | Canonical Livepatch service   |
+           | jammy   | hello=2.10-2ubuntu4         | n/a       | cis        | n/a      | n/a      | Available with the HWE kernel |
```
lucasmoura, line 171: I know this is unrelated, but should our livepatch description change for Jammy ? Maybe we can fix that out side somehow

orndorffgrant: This is actually a symptom of running a jammy container on a pre-jammy (and therefore older kernel version) system.

The container reports the old kernel version to the contract server, which responds with this message, and ua uses it in place of the normal description.

This was all put in place for trusty I believe.

And it will go away once we all upgrade our test systems to jammy, so seems non-urgent to address for now :shrug:

lucasmoura (approved):

PR #2068, opened 2022-04-27
CI: build debs in schroots without -proposed enabled

paride:
## Proposed Commit Message

```
CI: build debs in schroots without -proposed enabled

Later testing (integration testing) is done in an environment which
doesn't have -proposed enabled. Building the package with -proposed
enabled can cause dependency issues.
```

## Desired commit type

## Checklist


paride:
I dropped the `full-upgrade` part because runners have weird things like php installed from a ppa, which we don't want to upgrade every time.

orndorffgrant:
Those xenial failures passed locally for me. Merging

orndorffgrant (approved):

PR #2070, opened 2022-04-27
fix: add missing params for some operations

lucasmoura:
## Proposed Commit Message
fix: add missing params for some operations

When running ua fix we might need to enable a service or even renew an expired contract. In that situations, we use functions
from cli. For some of those functions, we were missing some required params. We are fixing that issue.

LP: #1969809

## Test Steps
Run the xenial integration test for fix

## Desired commit type

## Checklist


orndorffgrant (approved):
LGTM, thanks for reorganizing the fix integration tests into their own feature!

PR #2073, opened 2022-04-28
shell-linting: add shellcheck for shell scripts

### LINKED ISSUE #1305: testing/linting: add shell check for all shell scripts (debian/*inst|*rm) to avoid shell errors in packaging

CalvoM:
shell-linting: add shellcheck for shell scripts

Adding linting for shell scripts using shellcheck
Fixes #1305 

## Desired commit type

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,3 +7,8 @@ repos:
     rev: 5.8.0  # Also stored in dev-requirements.txt; update both together!
     hooks:
     - id: isort
+  - repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.8.0.4 # Also stored in dev-requirements.txt; update both together!
+    hooks:
+    - id: shellcheck
+      args: ['--severity=warning']
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -53,7 +53,6 @@ SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/m
 REBOOT_CMD_MARKER_FILE="/var/lib/ubuntu-advantage/marker-reboot-cmds-required"
 OLD_LICENSE_CHECK_MARKER_FILE="/var/lib/ubuntu-advantage/marker-license-check"
 
-MACHINE_TOKEN_FILE="/var/lib/ubuntu-advantage/private/machine-token.json"
 
 # Rename apt config files for ua services removing ubuntu release names
 redact_ubuntu_release_from_ua_apt_filenames() {
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,6 @@
-# The black and isort versions are also in .pre-commit-config.yaml; make sure
-# to update both together
+# The black, isort and shellcheck-py versions are also in .pre-commit-config.yaml; 
+# make sure to update both together
 black==22.3.0
 isort==5.8.0
 pre-commit
+shellcheck-py==0.8.0.4
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py3, flake8, py3-{xenial,bionic}, flake8-{xenial,bionic}, mypy, black, isort
+envlist = py3, flake8, py3-{xenial,bionic}, flake8-{xenial,bionic}, mypy, black, isort, shellcheck
 
 [testenv:flake8-bionic]
 pip_version = 9.0.2
@@ -18,6 +18,7 @@ pip_version = 9.0.2
 setuptools_version = 59.8.0
 
 [testenv]
+allowlist_externals=/usr/bin/bash
 deps =
     -rrequirements.txt
     -rtest-requirements.txt
@@ -28,6 +29,7 @@ deps =
     black: -rdev-requirements.txt
     isort: -rdev-requirements.txt
     behave: -rintegration-requirements.txt
+    shellcheck: -rdev-requirements.txt
 passenv =
     GOOGLE_APPLICATION_CREDENTIALS
     PYCLOUDLIB_CONFIG
@@ -57,6 +59,7 @@ commands =
     mypy-focal: mypy --python-version 3.7 uaclient/ features/ lib/
     black: black --check --diff uaclient/ features/ lib/ setup.py
     isort: isort --check --diff uaclient/ features/ lib/ setup.py
+    shellcheck: bash -O extglob -c "shellcheck -S warning tools/{*.sh,make-release,make-tarball} debian/ubuntu-advantage-tools.{config,postinst,postrm,prerm} lib/*.sh sru/release-!(27|27.3|27.5|27.4|27.4.*|27.7|27.8)/*.sh update-motd.d/**"
 
     behave-lxd-16.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.container" --tags="series.xenial,series.lts,series.all" --tags="~upgrade"
     behave-lxd-18.04: behave -v {posargs} --tags="uses.config.machine_type.lxd.container" --tags="series.bionic,series.lts,series.all" --tags="~upgrade"
--- a/tox.ini
+++ b/tox.ini
@@ -59,7 +59,7 @@ commands =
     mypy-focal: mypy --python-version 3.7 uaclient/ features/ lib/
     black: black --check --diff uaclient/ features/ lib/ setup.py
     isort: isort --check --diff uaclient/ features/ lib/ setup.py
-    shellcheck: bash -O extglob -c "shellcheck -S warning tools/{*.sh,make-release,make-tarball} debian/ubuntu-advantage-tools.{config,postinst,postrm,prerm} lib/*.sh sru/release-!(27|27.3|27.5|27.4|27.4.*|27.7|27.8)/*.sh update-motd.d/**"
+    shellcheck: bash -O extglob -O nullglob -c "shellcheck -S warning tools/{*.sh,make-release,make-tarball} debian/*.{config,postinst,postrm,prerm} lib/*.sh sru/release-!(27|27.3|27.5|27.4|27.4.*|27.7|27.8)/*.sh update-motd.d/*"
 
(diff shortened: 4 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -4,6 +4,11 @@ repos:
     hooks:
     - id: black
   - repo: https://github.com/pycqa/isort
-    rev: 5.8.0  # Also stored in dev-requirements.txt; update both together!
+    rev: 5.10.1  # Also stored in dev-requirements.txt; update both together!
```
orndorffgrant, line 7: I'm okay bumping the version of isort, but it should happen in a separate commit, and per the comment here, the version in dev-requirements.txt needs to be updated at the same time

.pre-commit-config.yaml
```diff
@@ -4,6 +4,11 @@ repos:
...
     hooks:
     - id: isort
+  - repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.8.0.4
```
orndorffgrant, line 11: This needs a comment similar to the others in this file stating that the version needs to be kept in sync with `dev-requirements.txt`
```suggestion
    rev: v0.8.0.4  # Also stored in dev-requirements.txt; update both together!
```

tox.ini
```diff
@@ -57,6 +60,7 @@ commands =
     mypy-focal: mypy --python-version 3.7 uaclient/ features/ lib/
     black: black --check --diff uaclient/ features/ lib/ setup.py
     isort: isort --check --diff uaclient/ features/ lib/ setup.py
+    shellcheck: bash ./shell_lint.sh
```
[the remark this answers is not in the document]
paride: +1 on this, but:
* `{ }` sequences will not be expanded because tox doesn't execute the command in a shell. We need `sh -c` there.
* Do we actually need those `-P` flags? AIUI `-P` is meant to specify paths to search when shellcheck encounters a `source` (or `.`) command.

Also: merging this will cause CI to fail, right? There are several scripts raising shellcheck warnings/errors.

dev-requirements.txt
```diff
@@ -3,3 +3,4 @@
 black==22.3.0
 isort==5.8.0
 pre-commit
+shellcheck-py
```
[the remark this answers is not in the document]
paride: This duplication is a bit ugly. I'd like to suggest a different approach: have a tox environment like:
```
[testenv:pre-commit]
deps = pre-commit
commands = pre-commit run --all-files
```
and then drop the `black`, `isort` and `shellcheck` tox environment, delegating them to pre-commit. This was we avoid duplicating dependencies and we have CI test with the very same tool that is used while developing (pre-commit).

pre-commit used https://github.com/pre-commit/identify to identify file types and normally does a good job, so this also saves us from manually specifying which files to check.

This change may belong to a separate PR, as it's not specific to shellcheck.

tox.ini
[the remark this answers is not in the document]
orndorffgrant: +1 to everything.

Currently ci specifically runs flake8, black, isort, and mypy. So this shouldn't break CI as written AFAICT.

We also generally won't want to fix some things. E.g. it'd be a waste of time to fix old sru test scripts, so we probably shouldn't run shellcheck on those at all.

dev-requirements.txt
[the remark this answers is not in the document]
orndorffgrant: I like this idea, but agree that it should be a separate PR. Created SC-970 to track this

tox.ini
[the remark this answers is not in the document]
CalvoM: Interestingly, `sh -c` does not work and results in the linting failing.However, `bash -c` works fine even without the **allowlist_externals** config.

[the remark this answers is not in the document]
paride: Right, `{ }` sequence expansion is bash only syntax. In Ubuntu we can just assume bash to be there. I do however get:
```
WARNING: test command found but not installed in testenv
  cmd: /usr/bin/sh
  env: /home/paride/ubuntu/git/ubuntu-advantage-client/.tox/shellcheck
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.
```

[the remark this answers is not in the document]
orndorffgrant: Good catch. It looks like you'll need `allowlist_externals = /usr/bin/bash` to get rid of that warning @CalvoM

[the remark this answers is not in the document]
CalvoM: Yes, let me add it.

paride:
I lost a bit what's the plan for this, but I assume that at some point we'll want to run shellcheck in CI. Do we want to merge the new tox env, fix all the things in a separate PR, and then add it to CI? Or is this PR a wip, and it will eventually cover the changes required to add shellcheck to CI? Or something else entirely?

tox.ini
```diff
@@ -57,6 +59,7 @@ commands =
     mypy-focal: mypy --python-version 3.7 uaclient/ features/ lib/
     black: black --check --diff uaclient/ features/ lib/ setup.py
     isort: isort --check --diff uaclient/ features/ lib/ setup.py
+    shellcheck: bash -c "shellcheck -S warning -C always tools/{*.sh,make-release,make-tarball} debian/ubuntu-advantage-tools.{config,postinst,postrm,prerm} lib/*.sh sru/release-27{,.3,.5,.8}/*.sh update-motd.d/**"
```
paride, line 62: Given that none of the scripts under `sru/` is shellcheck clean, should we just exclude it for now? Matching something like, with a couple of other minor changes:
```
tools/{*.sh,make-release,make-tarball} debian/*.{config,postinst,postrm,prerm} lib/*.sh update-motd.d/*
```
Also I'd remove `-C always` which may not play well with logging in automation.

CalvoM: I will update the sru/ folder selection to include only the latest release i.e. `27.8` currently.

CalvoM: `-C always` was just a way of testing the options, so lemme remove it, esp. if it will be a possible issue in the future.

.pre-commit-config.yaml
```diff
@@ -7,3 +7,8 @@ repos:
     rev: 5.8.0  # Also stored in dev-requirements.txt; update both together!
     hooks:
     - id: isort
+  - repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.8.0.4 # Also stored in dev-requirements.txt; update both together!
+    hooks:
+    - id: shellcheck
+      args: ['--severity=warning']
```
paride, line 14: Nit: missing newline

dev-requirements.txt
```diff
@@ -1,5 +1,6 @@
-# The black and isort versions are also in .pre-commit-config.yaml; make sure
+# The black,isort and shellcheck-py versions are also in .pre-commit-config.yaml; make sure
 # to update both together
```
paride, line 2: Nitty nit: space and rewrap.

```suggestion
# The black, isort and shellcheck-py versions are also in .pre-commit-config.yaml;
# make sure to update both together
```

tox.ini
```diff
@@ -57,6 +59,7 @@ commands =
     mypy-focal: mypy --python-version 3.7 uaclient/ features/ lib/
     black: black --check --diff uaclient/ features/ lib/ setup.py
     isort: isort --check --diff uaclient/ features/ lib/ setup.py
+    shellcheck: bash -c "shellcheck -S warning tools/{*.sh,make-release,make-tarball} debian/ubuntu-advantage-tools.{config,postinst,postrm,prerm} lib/*.sh sru/release-27.8/*.sh update-motd.d/**"
```
paride, line 62: This requires us to remember updating the globs on new releases. Maybe it's better to *exclude* what we know we're not going to fix (old sru scripts), something like (note: requires `extglob`);
```
bash -O extglob -c 'shellcheck -S warning tools/{*.sh,make-release,make-tarball} debian/*.{config,postinst,postrm,prerm} lib/*.sh sru/release-!(27|27.3|27.5|27.4|27.4.*|27.7)/*.sh update-motd.d/*'
```
Or we could leave `sru/*` out of shellcheck.

Note that I also made the `debian/` globs more generic and removed the double `*` from `update-motd.d/**` (we're not using `globstar` so I don't think `**` does anything special, but fixme).

paride (changes requested):
I left some minor comments, but it's not clear to me yet if we're going to fix the shellcheck failures as part of this PR or independently of it.

tox.ini
orndorffgrant: We also decided to exclude 27.8 sru scripts

debian/ubuntu-advantage-tools.postinst
```diff
@@ -53,7 +53,6 @@ SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/m
 REBOOT_CMD_MARKER_FILE="/var/lib/ubuntu-advantage/marker-reboot-cmds-required"
 OLD_LICENSE_CHECK_MARKER_FILE="/var/lib/ubuntu-advantage/marker-license-check"
 
-MACHINE_TOKEN_FILE="/var/lib/ubuntu-advantage/private/machine-token.json"
```
CalvoM, line 56: @orndorffgrant I noticed this is not used anyway thus I did not export it.

paride (changes requested):
One last comment, but I think we're there.

tox.ini
```diff
@@ -57,6 +59,7 @@ commands =
     mypy-focal: mypy --python-version 3.7 uaclient/ features/ lib/
     black: black --check --diff uaclient/ features/ lib/ setup.py
     isort: isort --check --diff uaclient/ features/ lib/ setup.py
+    shellcheck: bash -O extglob -c "shellcheck -S warning tools/{*.sh,make-release,make-tarball} debian/*.{config,postinst,postrm,prerm} lib/*.sh sru/release-!(27|27.3|27.5|27.4|27.4.*|27.7|27.8)/*.sh update-motd.d/*"
```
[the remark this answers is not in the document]
CalvoM: Yeah, I noticed. I wanted to add that disclaimer. Thanks for this!

paride (approved):
LGTM!

renanrodrigo:
Suggestion:  put the `Fixes #commithash` in this description to link the PR to the open GH issue

renanrodrigo (approved):
LGTM

PR #2074, opened 2022-04-29
apt: allow caching apt-cache policy commands

lucasmoura:
## Proposed Commit Message

We currently use the output to check if each repo entitlement is enabled in the system. This happens for example when we are
running an ua status operation, where we run an apt-cache policy for each entitled repo service the user has. We can cache that command if no update calls are performed. Therefore, we are speeding up the execution of ua status by caching the result of apt-cache policy calls.

Testing this changeset on a xenial container, `ua status` response average turns from `1.42s` to `0.57`.

## Test Steps
Please, run status with and without this change to confirm the speed up.
Additionally, run attach integration tests where more than one repo service is being enabled, to confirm the status in the end is not incorrect.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -5,6 +5,7 @@ import os
 import re
 import subprocess
 import tempfile
+from functools import lru_cache
 from typing import Dict, List, Optional
 
 from uaclient import event_logger, exceptions, gpg, messages, util
@@ -179,6 +180,16 @@ def run_apt_command(
     return out
 
 
+@lru_cache(maxsize=None)
+def run_apt_cache_policy_command(
+    error_msg: Optional[str] = None,
+    env: Optional[Dict[str, str]] = {},
+) -> str:
+    return run_apt_command(
+        cmd=["apt-cache", "policy"], error_msg=error_msg, env=env
+    )
+
+
 def run_apt_update_command(env: Optional[Dict[str, str]] = {}) -> str:
     try:
         out = run_apt_command(cmd=["apt-get", "update"], env=env)
@@ -191,6 +202,11 @@ def run_apt_update_command(env: Optional[Dict[str, str]] = {}) -> str:
             msg=messages.APT_UPDATE_FAILED.msg + "\n" + e.msg,
             msg_code=messages.APT_UPDATE_FAILED.name,
         )
+    finally:
+        # Whenever we run an apt-get update command, we must invalidate
+        # the existing apt-cache policy cache. Otherwise, we could provide
+        # users with incorrect values.
+        run_apt_cache_policy_command.cache_clear()
 
     return out
 
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -103,8 +103,8 @@ class RepoEntitlement(base.UAEntitlement):
                 messages.NO_APT_URL_FOR_SERVICE.format(title=self.title),
             )
         protocol, repo_path = repo_url.split("://")
-        policy = apt.run_apt_command(
-            ["apt-cache", "policy"], messages.APT_POLICY_FAILED.msg
+        policy = apt.run_apt_cache_policy_command(
+            error_msg=messages.APT_POLICY_FAILED.msg
         )
         match = re.search(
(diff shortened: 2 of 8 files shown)


lucasmoura:
One alternative we could have is instead of shelling out would be to rely on the `python3-apt` module bindings instead.
For example, we could have a function like this:

```python
@lru_cache(maxsize=None)
def run_apt_cache_policy_command(
    error_msg: Optional[str] = None,
    env: Optional[Dict[str, str]] = {},
) -> str:
    cache = apt_pkg.Cache(progress=None)
    depcache = apt_pkg.DepCache(cache)
    policy = depcache.policy
    sl = apt_pkg.SourceList()
    sl.read_main_list()

    output = ""
    for sourcefile in cache.file_list:
        if not sourcefile.site:
            continue

        index = sl.find_index(sourcefile)
        url = index.archive_uri("")
        pin_priority = policy.get_priority(sourcefile)
        source_url = "{} {}/{}".format(
            url,
            sourcefile.archive,
            sourcefile.component,
        )
        output += "{} {}\n".format(pin_priority, source_url)

    return output
```

The problem is that this approach is still 0.2 slower on average than shelling out.
But I am still taking a better look at it.

orndorffgrant:
@lucasmoura would it change anything to more holistically redesign the needs of `RepoEntitlment`'s status check?

In other words, rather than recreating the output of `apt-cache policy` in python, could we write a function like this (also including extra caching of presumably expensive calls):

```python
# not sure how many of these caches would actually be useful

@lru_cache(maxsize=None)
def get_apt_cache():
    return apt_pkg.Cache(progress=None)

@lru_cache(maxsize=None)
def get_apt_dep_cache():
    return apt_pkg.DepCache(get_apt_cache())

@lru_cache(maxsize=None)
def get_apt_source_list():
    sl = apt_pkg.SourceList()
    sl.read_main_list()
    return sl
    
@lru_cache(maxsize=None)
def get_repo_priority(repo_url) -> Optional[int]:
    """
    returns apt priority if found
    returns None if not found
    """
    cache = get_apt_cache()
    policy = get_apt_dep_cache().policy
    sl = get_apt_source_list()

    for sourcefile in cache.file_list:
        if not sourcefile.site:
            continue
        index = sl.find_index(sourcefile)
        url = index.archive_uri("")
        if url == repo_url:
            return policy.get_priority(sourcefile)

    return None
```

And then use `get_repo_priority` in `repo.py` `application_status`?

orndorffgrant:
I'd also understand if what I just suggested is too much effort for this, especially since shelling out and caching is working pretty well. LMK what you think

lucasmoura:
@orndorffgrant according to this profile script, it seems that the `read_main_list` call is the root of the slowdown:
```
from pstats import Stats
from cProfile import Profile

from uaclient.apt import run_apt_cache_policy_command


profiler = Profile()
profiler.runcall(run_apt_cache_policy_command)

stats = Stats(profiler)
stats.strip_dirs()
stats.sort_stats('cumulative')
stats.print_stats()
```

Output:
```
         164 function calls in 0.308 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.251    0.251    0.308    0.308 apt.py:186(run_apt_cache_policy_command)
        1    0.044    0.044    0.044    0.044 {method 'read_main_list' of 'apt_pkg.SourceList' objects}
       32    0.013    0.000    0.013    0.000 {method 'find_index' of 'apt_pkg.SourceList' objects}
        1    0.001    0.001    0.001    0.001 {built-in method apt_pkg.init}
       32    0.000    0.000    0.000    0.000 {method 'archive_uri' of 'apt_pkg.IndexFile' objects}
       64    0.000    0.000    0.000    0.000 {method 'format' of 'str' objects}
       32    0.000    0.000    0.000    0.000 {method 'get_priority' of 'apt_pkg.Policy' objects}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

Currently, we need it to get the full `url` associated with the source.

orndorffgrant (approved):
Looks great to me! Thanks for the thorough investigation into all the options here.

PR #2075, opened 2022-05-02
daemon: lower memory with less imports and start testing memory

orndorffgrant:
## Notes

The tests are mostly so that we can keep an eye on memory usage. They're aren't "hard limits" necessarily.

This lowers memory usage by 2Mb by avoiding importing the cli module and using a smaller daemon-tailored `setup_logging` function.

## Proposed Commit Message

```
daemon: lower memory with less imports and start testing memory
```

## Test Steps

Run the gcp daemon test with new steps on all series

## Desired commit type

## Checklist

### DIFF
--- a/features/daemon.feature
+++ b/features/daemon.feature
@@ -83,19 +83,32 @@ Feature: Pro Upgrade Daemon only runs in environments where necessary
         """
         inactive
         """
+
         # verify it stays on when configured to do so
         When I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
         """
         ua_config:
             poll_for_pro_license: true
         """
+        # Turn on memory accounting
+        When I run `sed -i s/#DefaultMemoryAccounting=no/DefaultMemoryAccounting=yes/ /etc/systemd/system.conf` with sudo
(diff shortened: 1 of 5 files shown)


features/daemon.feature
```diff
@@ -83,19 +83,32 @@ Feature: Pro Upgrade Daemon only runs in environments where necessary
...
         """
+
         # verify it stays on when configured to do so
         When I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
         """
         ua_config:
             poll_for_pro_license: true
         """
+        # Turn on memory accounting
+        When I run `sed -i s/#DefaultMemoryAccounting=no/DefaultMemoryAccounting=yes/ /etc/systemd/system.conf` with sudo
```
CalvoM, line 94: @orndorffgrant Is there a specific reason why gcp generic machines do not default to yes for *DefaultMemoryAccounting*?

orndorffgrant: Focal ones have it on by default, but bionic doesn't. I didn't document the other series because I needed to add this line regardless.
There may be a specific reason for the defaults, but honestly I don't know :)

lucasmoura (approved):

PR #2077, opened 2022-05-04
gcp-pro: handle metadata endpoint error

### LINKED ISSUE #1936: Unhandled exception during auto-attach on GCE
Filing a bug as the log suggests:

```
2022-01-14 17:27:33,439 - util.py:(572) [DEBUG]: URL [GET] response: http://metadata/computeMetadata/v1/instance/service-accounts/default/identity?audience=contracts.canonical.com&format=full&licenses=TRUE, headers: {'Connection': 'Close', 'Content-Length': '94', 'Content-Type': 'text/html', 'Date': 'Fri, 14 Jan 2022 17:27:33 GMT', 'Metadata-Flavor': 'Google', 'Server': 'Metadata Server for VM', 'X-Frame-Options': 'SAMEORIGIN', 'X-XSS-Protection': '0'}, data: <REDACTED>
2022-01-14 17:27:33,439 - util.py:(434) [DEBUG]: Reading file: /var/lib/ubuntu-advantage/notices.json
2022-01-14 17:27:33,440 - util.py:(726) [DEBUG]: Removing file: /var/lib/ubuntu-advantage/notices.json
2022-01-14 17:27:33,440 - cli.py:(1497) [ERROR]: Unhandled exception, please file a bug
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1459, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1541, in main
    return args.action(args, cfg=cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 146, in new_f
    return f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 131, in new_f
    retval = f(*args, cfg=cfg, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1065, in action_auto_attach
    actions.auto_attach(cfg, instance)
  File "/usr/lib/python3/dist-packages/uaclient/actions.py", line 51, in auto_attach
    tokenResponse = contract_client.request_auto_attach_contract_token(
  File "/usr/lib/python3/dist-packages/uaclient/contract.py", line 128, in request_auto_attach_contract_token
    data=instance.identity_doc,
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 251, in decorator
    raise e
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 248, in decorator
    return f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/clouds/gcp.py", line 33, in identity_doc
    url_response, _headers = util.readurl(
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 580, in readurl
    raise resp
  File "/usr/lib/python3/dist-packages/uaclient/util.py", line 560, in readurl
    resp = request.urlopen(req, timeout=timeout)
  File "/usr/lib/python3.8/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.8/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/usr/lib/python3.8/urllib/request.py", line 640, in http_response
    response = self.parent.error(
  File "/usr/lib/python3.8/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found
```
The unit also enters failed state due to this:

```
UNIT                   LOAD   ACTIVE SUB    DESCRIPTION                 
● ua-auto-attach.service loaded failed failed Ubuntu Advantage auto attach
```

The exception needs to be handled to suppress the `Unhandled exception, please file a bug` message, not sure what the unit state supposed to be.

CalvoM:
gcp-pro: handle metadata endpoint error

We are handling for the errors raised when using gcp pro and the service account is not enabled.
Fixes #1936 

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -15,7 +15,7 @@ import tempfile
 import textwrap
 import time
 from functools import wraps
-from typing import Dict, List, Optional, Tuple  # noqa
+from typing import List, Optional, Tuple  # noqa
 
 import yaml
 
@@ -1308,9 +1308,7 @@ def action_auto_attach(args, *, cfg):
     try:
         actions.auto_attach(cfg, instance)
     except exceptions.UrlError:
-        event.info(messages.ATTACH_FAILURE)
-        return 1
-    except exceptions.UserFacingError:
+        event.info(messages.ATTACH_FAILURE.msg)
         return 1
     else:
         _post_cli_attach(cfg)
--- a/uaclient/clouds/gcp.py
+++ b/uaclient/clouds/gcp.py
@@ -5,7 +5,7 @@ import os
 from typing import Any, Dict, List, Optional  # noqa: F401
 from urllib.error import HTTPError
 
-from uaclient import exceptions, util
+from uaclient import exceptions, messages, util
 from uaclient.clouds import AutoAttachCloudInstance
 
 LOG = logging.getLogger("ua.clouds.gcp")
@@ -42,11 +42,33 @@ class UAAutoAttachGCPInstance(AutoAttachCloudInstance):
     # mypy does not handle @property around inner decorators
     # https://github.com/python/mypy/issues/1362
     @property  # type: ignore
-    @util.retry(HTTPError, retry_sleeps=[1, 2, 5])
+    @util.retry(exceptions.GCPProAccountError, retry_sleeps=[1, 2, 5])
     def identity_doc(self) -> Dict[str, Any]:
-        url_response, _headers = util.readurl(
-            TOKEN_URL, headers={"Metadata-Flavor": "Google"}
-        )
+        try:
+            headers = {"Metadata-Flavor": "Google"}
+            url_response, _headers = util.readurl(TOKEN_URL, headers=headers)
+        except HTTPError as e:
+            body = getattr(e, "body", None)
+            error_desc = None
+            if body:
+                try:
+                    error_details = json.loads(
+                        body, cls=util.DatetimeAwareJSONDecoder
+                    )
+                except ValueError:
+                    error_details = None
+                if error_details:
+                    error_desc = error_details.get("error_description", None)
+            msg = error_desc if error_desc else e.reason
+            msg_code = None
+            if error_desc and "service account" in error_desc.lower():
+                msg = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.msg.format(
+                    error_msg=msg
+                )
+                msg_code = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.name
+            raise exceptions.GCPProAccountError(
+                msg=msg, msg_code=msg_code, code=getattr(e, "code", 0)
+            )
         return {"identityToken": url_response}
 
     @property
--- a/uaclient/exceptions.py
+++ b/uaclient/exceptions.py
@@ -258,6 +258,19 @@ class SecurityAPIMetadataError(UserFacingError):
         )
 
 
+class GCPProAccountError(UserFacingError):
+    """An exception raised when GCP Pro service account is not enabled"""
+
+    def __init__(self, msg: str, msg_code: Optional[str], code=None):
+        self.code = code
+        super().__init__(msg, msg_code)
+
+    def __str__(self):
+        return "GCPProServiceAccount Error {code}: {msg}".format(
+            code=self.code, msg=self.msg
+        )
+
+
 class CloudFactoryError(Exception):
     def __init__(self, cloud_type: Optional[str]) -> None:
         self.cloud_type = cloud_type
--- a/uaclient/messages.py
+++ b/uaclient/messages.py
@@ -743,6 +743,13 @@ REALTIME_ERROR_INSTALL_ON_CONTAINER = NamedMessage(
     "Cannot install Real-Time Kernel on a container.",
 )
 
+GCP_SERVICE_ACCT_NOT_ENABLED_ERROR = NamedMessage(
+    "gcp-pro-service-account-not-enabled",
+    "Failed to attach machine\n"
+    "{error_msg}\n"
+    "For more information, "
+    "see https://cloud.google.com/iam/docs/service-accounts",
+)
 
 LOG_CONNECTIVITY_ERROR_TMPL = CONNECTIVITY_ERROR.msg + " {error}"
 LOG_CONNECTIVITY_ERROR_WITH_URL_TMPL = (
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -1,5 +1,4 @@
 import logging
-import sys
 from typing import Optional  # noqa: F401
 
 from uaclient import (
@@ -40,7 +39,6 @@ def attach_with_token(
         update_apt_and_motd_messages(cfg)
         raise exc
     except exceptions.UserFacingError as exc:
-        event.info(exc.msg, file_type=sys.stderr)
         # Persist updated status in the event of partial attach
         ua_status.status(cfg=cfg)
         update_apt_and_motd_messages(cfg)
--- a/uaclient/clouds/tests/test_gcp.py
+++ b/uaclient/clouds/tests/test_gcp.py
@@ -12,6 +12,7 @@ from uaclient.clouds.gcp import (
     WAIT_FOR_CHANGE,
     UAAutoAttachGCPInstance,
 )
+from uaclient.exceptions import GCPProAccountError
 
 M_PATH = "uaclient.clouds.gcp."
 
@@ -57,7 +58,7 @@ class TestUAAutoAttachGCPInstance:
 
         instance = UAAutoAttachGCPInstance()
         if exception:
-            with pytest.raises(HTTPError) as excinfo:
+            with pytest.raises(GCPProAccountError) as excinfo:
                 instance.identity_doc
             assert 704 == excinfo.value.code
         else:
@@ -69,9 +70,12 @@ class TestUAAutoAttachGCPInstance:
         assert expected_sleep_calls == sleep.call_args_list
 
         expected_logs = [
-            "HTTP Error 701: funky error msg Retrying 3 more times.",
-            "HTTP Error 702: funky error msg Retrying 2 more times.",
-            "HTTP Error 703: funky error msg Retrying 1 more times.",
+            "GCPProServiceAccount Error 701: "
+            + "funky error msg Retrying 3 more times.",
+            "GCPProServiceAccount Error 702: "
+            + "funky error msg Retrying 2 more times.",
+            "GCPProServiceAccount Error 703: "
+            + "funky error msg Retrying 1 more times.",
         ]
         logs = caplog_text()
         for log in expected_logs:


uaclient/exceptions.py
```diff
@@ -258,6 +258,20 @@ def __init__(self, msg: str, issue_id: str) -> None:
...
+
+    def __init__(self, msg: str, msg_code: Optional[str], code=None):
+        self.code = code
+        self.error_msg = msg
+        super().__init__(msg, msg_code)
+
+    def __str__(self):
+        return "GCPProServiceAccount Error {code}: {msg}".format(
+            code=self.code, msg=self.error_msg
+        )
```
orndorffgrant, line 272: I think `msg` is already stored in the parent class
```suggestion
    def __init__(self, msg: str, msg_code: Optional[str], code=None):
        self.code = code
        super().__init__(msg, msg_code)

    def __str__(self):
        return "GCPProServiceAccount Error {code}: {msg}".format(
            code=self.code, msg=self.msg
        )
```

uaclient/clouds/gcp.py
```diff
@@ -42,11 +42,40 @@ def __init__(self):
...
+        try:
+            headers = {"Metadata-Flavor": "Google"}
+            url_response, _headers = util.readurl(TOKEN_URL, headers=headers)
+        except URLError as e:
+            body = None
+            if hasattr(e, "body"):
+                body = getattr(e, "body", None)
+            elif hasattr(e, "read"):
+                reader = getattr(e, "read")()
+                body = reader.decode("utf-8")
```
orndorffgrant, line 56: We can simplify this by only catching `HTTPError`s which is what happened in the initial bug report.
```suggestion
        except HTTPError as e:
            body = getattr(e, "body", None)
```

uaclient/clouds/gcp.py
```diff
@@ -42,11 +42,40 @@ def __init__(self):
...
+            if body:
+                try:
+                    error_details = json.loads(
+                        body, cls=util.DatetimeAwareJSONDecoder
+                    )
+                except ValueError:
+                    error_details = None
+                if error_details:
+                    error_desc = error_details.get("error_description", None)
+            msg = error_desc if error_desc else getattr(e, "msg", None)
```
orndorffgrant, line 67: When would `e` have a `msg` property?

uaclient/defaults.py
```diff
@@ -20,6 +20,7 @@
 DOCUMENTATION_URL = (
     "https://discourse.ubuntu.com/t/ubuntu-advantage-client/21788"
 )
+GCP_SERVICE_ACCT_DOC_URL = "https://cloud.google.com/iam/docs/service-accounts"
```
orndorffgrant, line 23: I think this should just be hardcoded in the message string. It's not a default value to any option the cli has.

uaclient/clouds/gcp.py
```diff
@@ -42,11 +42,40 @@ def __init__(self):
...
+            error_code = getattr(e, "code", 0)
+            if error_code == 404:
+                msg += "\n" + messages.GCP_SERVICE_ACCT_URL_MESSAGE
+                msg = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.msg.format(
+                    error_msg=msg
+                )
+                msg_code = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.name
+            raise exceptions.GCPProAccountError(
+                msg=msg, msg_code=msg_code, code=getattr(e, "code", 0)
+            )
```
orndorffgrant, line 78: I think we should only do the special error if we see "service account" in the description
```suggestion
            msg = error_desc if error_desc else getattr(e, "msg", None)
            msg_code = None
            if error_desc is not None and "service account" in error_desc.lower():
                msg = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.msg.format(
                    error_msg=error_desc + "\n" + messages.GCP_SERVICE_ACCT_URL_MESSAGE
                )
                msg_code = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.name
            raise exceptions.GCPProAccountError(
                msg=msg, msg_code=msg_code, code=getattr(e, "code", 0)
            )
```

uaclient/messages.py
```diff
@@ -731,6 +735,14 @@ class TxtColor:
     "Cannot install Real-Time Kernel on a container.",
 )
 
+GCP_SERVICE_ACCT_NOT_ENABLED_ERROR = NamedMessage(
+    "gcp-pro-service-account-not-enabled",
+    "Failed to attach machine\n" + "{error_msg}.",
```
orndorffgrant, line 740: Not sure if we need the `.` and no need for concatenating things
```suggestion
    "Failed to attach machine\n{error_msg}",
```

uaclient/cli.py
```diff
@@ -1291,10 +1291,10 @@ def action_auto_attach(args, *, cfg):
     try:
         actions.auto_attach(cfg, instance)
     except exceptions.UrlError:
-        event.info(messages.ATTACH_FAILURE)
-        return 1
-    except exceptions.UserFacingError:
+        event.info(messages.ATTACH_FAILURE.msg)
         return 1
+    except exceptions.UserFacingError as e:
+        raise e
```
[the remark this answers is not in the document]
CalvoM: Yeah, I noticed that the UserFacingError errors are not handled. @orndorffgrant @lucasmoura Do you foresee any implications of this?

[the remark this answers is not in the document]
CalvoM: i.e. removing this line and leaving the main_error_handler decorator to handle it.

uaclient/defaults.py
CalvoM: Yeah, we will extract it to the defaults when it used in more than one places.

uaclient/clouds/gcp.py
CalvoM: Yeah, hopefully the wording will not change to something like `account of service`

uaclient/clouds/gcp.py
orndorffgrant: Oh interesting. You're right that it uses the `msg` property internally seen here: https://github.com/python/cpython/blob/3.10/Lib/urllib/error.py#L39=

But it isn't documented anywhere here: https://docs.python.org/3/library/urllib.error.html?highlight=httperror#urllib.error.HTTPError

My guess is that they want you to use the `reason` property instead, which just returns `msg` https://github.com/python/cpython/blob/3.10/Lib/urllib/error.py#L61=

So can we use `reason` instead of `msg` here?

orndorffgrant: Oh interesting. You're right that it uses the `msg` property internally seen here: https://github.com/python/cpython/blob/3.10/Lib/urllib/error.py#L39=

But it isn't documented anywhere here: https://docs.python.org/3/library/urllib.error.html?highlight=httperror#urllib.error.HTTPError

My guess is that they want you to use the `reason` property instead, which just returns `msg` https://github.com/python/cpython/blob/3.10/Lib/urllib/error.py#L61=

So can we use `reason` instead of `msg` here?

uaclient/cli.py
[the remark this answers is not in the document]
orndorffgrant: Oh I didn't notice this was for auto-attach. Sorry we can't make this change.

The UserFacingError is already handled in actions.py. So if we don't catch it here, the msg will get printed more than once.
This should be refactored to use the main_error_handler instead of catching in actions.py, but that deserves its own PR.

uaclient/clouds/gcp.py
CalvoM: Oh makes sense, just found the thread proposing inclusion of `reason` property [here](https://bugs.python.org/issue13211)

uaclient/clouds/gcp.py
```diff
@@ -42,11 +42,37 @@ def __init__(self):
...
+                    )
+                except ValueError:
+                    error_details = None
+                if error_details:
+                    error_desc = error_details.get("error_description", None)
+            msg = error_desc if error_desc else getattr(e, "reason", None)
+            msg_code = None
+            if error_desc and "service account" in error_desc.lower():
+                msg += "\n" + messages.GCP_SERVICE_ACCT_URL_MESSAGE
+                msg = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.msg.format(
```
lucasmoura, line 69: @CalvoM I think the message still has an additional newline:
```
Failed to attach machine
Service account not enabled on this instance.

For more information, see https://cloud.google.com/iam/docs/service-accounts
```

CalvoM: When I test I do not get the newline. 

**Failed to attach machine
Service account not enabled on this instance
For more information, see https://cloud.google.com/iam/docs/service-accounts**

CalvoM: @lucasmoura I have tried again and still I am not able to reproduce the newline. Furthermore, considering the msg had the reason msg from httperror, which does not end with newline thus there is no empty line I think.

lucasmoura: Yes, my mistake here. I think I was using an older version of the PR

lucasmoura (approved):

uaclient/cli.py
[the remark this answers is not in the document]
orndorffgrant: @CalvoM we still need to address this. I think if we delete actions.py line 43, then this change will be okay

uaclient/clouds/gcp.py
orndorffgrant: @CalvoM we can still simplify what is there to this suggestion :)

uaclient/clouds/gcp.py
```diff
@@ -42,11 +42,37 @@ def __init__(self):
...
+            if body:
+                try:
+                    error_details = json.loads(
+                        body, cls=util.DatetimeAwareJSONDecoder
+                    )
+                except ValueError:
+                    error_details = None
+                if error_details:
+                    error_desc = error_details.get("error_description", None)
+            msg = error_desc if error_desc else getattr(e, "reason", None)
```
orndorffgrant, line 65: I think we can just use `e.reason`
```suggestion
            msg = error_desc if error_desc else e.reason
```

uaclient/messages.py
```diff
@@ -731,6 +731,13 @@ class TxtColor:
     "Cannot install Real-Time Kernel on a container.",
 )
 
+GCP_SERVICE_ACCT_NOT_ENABLED_ERROR = NamedMessage(
+    "gcp-pro-service-account-not-enabled",
+    "Failed to attach machine\n{error_msg}",
+)
+
+GCP_SERVICE_ACCT_URL_MESSAGE = """\
+For more information, see https://cloud.google.com/iam/docs/service-accounts"""
```
orndorffgrant, line 740: Do these have to be two separate strings?
```suggestion
GCP_SERVICE_ACCT_NOT_ENABLED_ERROR = NamedMessage(
    "gcp-pro-service-account-not-enabled",
    "Failed to attach machine\n"
    "{error_msg}\n"
    "For more information, see https://cloud.google.com/iam/docs/service-accounts",
)

File: uaclient/clouds/gcp.py [hunk 11]
```diff
@@ -42,11 +42,37 @@ def __init__(self):
...
+            msg_code = None
+            if error_desc and "service account" in error_desc.lower():
+                msg += "\n" + messages.GCP_SERVICE_ACCT_URL_MESSAGE
+                msg = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.msg.format(
+                    error_msg=msg
+                )
```
orndorffgrant, line 71: the multiple uses of the `msg` variable are confusing here.
If we combine the message variables like I suggest in my other comment, this can be simplified to
```suggestion
                msg = messages.GCP_SERVICE_ACCT_NOT_ENABLED_ERROR.msg.format(
                    error_msg=error_desc
                )
```

uaclient/cli.py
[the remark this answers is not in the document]
CalvoM: @orndorffgrant I had not seen this comment before. 

Yeah makes sense, lemme delete the extra logging line.

uaclient/messages.py
CalvoM: I thought it would be a line to use later, but I guess we will refactor later when we need it separated.

uaclient/clouds/gcp.py
CalvoM: Done.

orndorffgrant (approved):
Thanks for all the details that went into this one!

PR #2078, opened 2022-05-05
enable: treat empty list affordance differently from null

orndorffgrant:
## Notes

After this, an empty list for the series affordance will mean that no series are supported.
The same will be true for architectures and kernel flavors.

## Proposed Commit Message

```
enable: treat empty list affordance differently from null
```

## Test Steps

Run the new integration test.

Use machine token overlays to test that an empty affordance list behaves as you expect.

## Desired commit type

## Checklist


orndorffgrant:
@alnvdl FYI

lucasmoura (approved):

PR #2079, opened 2022-05-05
Use pycloudlib.toml gce.credentials_path (SC-987)

paride:
## Proposed Commit Message

```
Use pycloudlib.toml gce.credentials_path

Have pycloudlib handle the GCE credentials rather than using
the special GOOGLE_APPLICATION_CREDENTIALS, which is handled
directly by the gcloud library.
```

## Test Steps

Let CI run.

## Desired commit type

## Checklist

### DIFF
--- a/.github/workflows/ci-integration.yaml
+++ b/.github/workflows/ci-integration.yaml
@@ -105,14 +105,14 @@ jobs:
           UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING_EXPIRED: '${{ secrets.UACLIENT_BEHAVE_CONTRACT_TOKEN_STAGING_EXPIRED }}'
         run: |
           PYCLOUDLIB_CONFIG="$(mktemp --tmpdir="${{ runner.temp }}" pycloudlib.toml.XXXXXXXXXX)"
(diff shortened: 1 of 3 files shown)


paride:
This is expected to fail as it requires an update in `pycloudlib.toml`. @blackboxsw can you please update it in the usual way? Then we can retrigger and hopefully GCE integration will pass.

paride:
The awspro failure looks unrelated, I'm a bit surprised by it. I'll try to retrigger the failed jobs (gcppro will fail again - expected for now).

orndorffgrant:
Hi @paride I noticed a small clue that may help here.

When running the test with my personal pycloudlib.toml the test passes.
When running with the shared pycloudlib.toml it times out.

The differences I see are the different credentials, and the shared toml sets the ec2 key-pair settings, which mine does not.

However, the test still times out for me after commenting the ec2 key-pair settings in the shared toml file, which makes me think it could actually be related to the account.

My best guess is that the instances are being launched, but something is preventing us from connecting to them. Is there a new default security group for the account that is blocking ssh maybe (or something like that)?

paride:
Thanks @orndorffgrant, very relevant findings. I'm trying to figure out if anything changed on the keys or security group used in CI.

orndorffgrant (approved):
Thanks for this!

PR #2080, opened 2022-05-05
tests: fix gcpgeneric tests for cis/usg status

lucasmoura:


orndorffgrant (approved):
My bad. Thanks for catching this!

PR #2083, opened 2022-05-09
fips: clean notice for older PRO FIPS images

lucasmoura:
## Proposed Commit Message
fips: clean notice for older PRO FIPS images

We currently detect that an user is on an old PRO FIPS machine on our postinst script and emit a notice for that. However, even
after we fix the issue on the older image, we don't remove the notice. We are now properly cleaning up that notice.

LP: #1972026

## Test Steps
Run the modified unit tests

## Desired commit type

## Checklist

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -1041,7 +1041,7 @@ def docker_image_is_not_larger(context, name, series, package):
 
 
 @then(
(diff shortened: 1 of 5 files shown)


orndorffgrant (approved):

CalvoM (commented):
Why is the features/ubuntu_upgrade file empty?

PR #2091, opened 2022-05-13
chore: fix tox version of python for older tests to 3.8

renanrodrigo:
Jammy runs Python 3.10, and it fails to satisfy the pip and setuptools
pinned versions.
Those tests (for Xenial/Bionic constraints) run fine with Python 3.8.
Python 3.8 (and its distutils) must be in your system for those tests to run properly.

## Test Steps
On a Jammy system, with Python 3.10 as the default, and Python 3.8 installed:
`$ make clean`
`$ tox`
Everything should be executed as usual, and pass.

## Desired commit type

## Checklist

### DIFF
--- a/.github/workflows/ci-base.yaml
+++ b/.github/workflows/ci-base.yaml
@@ -27,7 +27,7 @@ jobs:
         run: tox -e mypy
   unit-tests:
     name: Unit
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - name: Install dependencies
         run: |
@@ -39,5 +39,5 @@ jobs:
         run: |
           python3 -m venv unit-test-venv
           . unit-test-venv/bin/activate
-          pip install tox tox-pip-version
-          tox -e py3 -e py3-bionic -e py3-xenial
+          pip install tox tox-pip-version tox-setuptools-version
+          tox -e py3-test -e py38-test-bionic -e py38-test-xenial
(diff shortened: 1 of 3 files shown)


renanrodrigo:
Well, CI likes it now:
- I added the missing pip-setuptools-version dependency
- I had to change it to run on Focal instead of Bionic, so we get python3.8 easily
- IMHO, as soon as Jammy is available in Actions, we should make sure to cover all releases by:
  - Run the unit tests base (the `py3-` ones) on Jammy.
  - Have a Focal specific constraints file

Not that useful now though

orndorffgrant:
Looks good!

While we're revamping our python-version testing, I think we should go all the way :)

I suggest we adopt [pyenv](https://github.com/pyenv/pyenv) for managing several versions of python on a dev system. This will involve documentation to instruct to use the [installer](https://github.com/pyenv/pyenv-installer). And then a list of python versions to install 3.5, 3.6, 3.8 and 3.10 I think

Then we can use [tox-pyenv](https://github.com/tox-dev/tox-pyenv) to let tox know how to find those installs.

And then we can have granular python version toxenv testing per series. e.g. py35-xenial, py36-bionic, etc.

renanrodrigo:
@orndorffgrant that's a great path to take.
It will complicate the setup of the dev environment a little, but with significant gain of covering exactly the environment we expect for every release.
I'm on it.

renanrodrigo:
Will start a new branch with a better solution

PR #2093, opened 2022-05-17
Rewrite unit test environments for the different releases (SC-1044)

renanrodrigo:
Define `tox` environments to run the tests with different python versions, and add a script to set up `pyenv`.
Make sure all unit tests and flake8/mypy checks are executed with respect to the Python version, setuptools version, and package version constraints for the supported releases.

I tested the pyenv solution on Focal, Impish, Jammy, and it worked.

## Test Steps
`make clean`
~`make testdeps`~ run the `tools/setup_pyenv.sh` script
Then make sure pyenv is working by refreshing your terminal, and then
`tox`
Everything should run and pass.

## Desired commit type

## Checklist


Makefile
```diff
@@ -28,9 +28,7 @@ test:
 	@tox
 
 testdeps:
-	pip install tox
-	pip install tox-pip-version
-	pip install tox-setuptools-version
+	. tools/setup_pyenv.sh
```
renanrodrigo, line 31: I thought it was a good idea to call the script from the Makefile, but I have no idea if this is a good practice.
Should I instead copy the script contents and maintain it here?

tools/setup_pyenv.sh
```diff
@@ -0,0 +1,36 @@
...
+libreadline-dev libsqlite3-dev wget curl llvm libncursesw5-dev xz-utils \
+tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev clang
+
+curl http://pyenv.run | bash
+
+echo 'export PYENV_ROOT="$HOME/.pyenv"
+command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+' >> ~/.bashrc
```
renanrodrigo, line 16: `.bashrc` is bash-centric - I don't know a sensible way of doing this for any generic shell...

renanrodrigo (commented):
Couple questions for the reviewer

renanrodrigo:
@paride as discussed, please take a look at a good way to set up the CI here - thanks much

tools/setup_pyenv.sh
```diff
@@ -0,0 +1,36 @@
...
+
+sudo apt-get update
+sudo apt-get install -y make build-essential libssl-dev zlib1g-dev libbz2-dev \
+libreadline-dev libsqlite3-dev wget curl llvm libncursesw5-dev xz-utils \
+tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev clang
+
+curl http://pyenv.run | bash
+
+echo 'export PYENV_ROOT="$HOME/.pyenv"
+command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"
```
paride, line 13: `$PYENV_ROOT` is undefined at this point

renanrodrigo: even with the export in the previous line?

paride: you're right, I lost count of the quotes

.github/workflows/ci-base.yaml
```diff
@@ -27,17 +27,19 @@ jobs:
         run: tox -e mypy
```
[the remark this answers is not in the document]
paride: I think we can run flake8 on the different releases after all.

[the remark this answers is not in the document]
paride: Done

Makefile
lucasmoura: I think it is okay to keep the script, but I think @paride will not that better than me

lucasmoura (commented):
It is working for me. But I think we can further discuss the points you raised @renanrodrigo

paride:
@renanrodrigo if you rebase this on main hopefully all the jobs will pass

Makefile
paride: Per standup discussion, I think we came to the conclusion to make the script more instructional than auto-configuring, and to document its usage as a devel environment setup step rather than running in as part of the `testdeps` target.

renanrodrigo:
@orndorffgrant @lucasmoura @paride changes addressed - the script is not on the Makefile anymore, it does not change bash or anything, and there is a link in the documentation instructing how to do the changes.

Makefile
renanrodrigo: done

tools/setup_pyenv.sh
renanrodrigo: solved by not configuring it at all, but point to docs which tells how to do that

renanrodrigo:
Had to fix Flake8 (issue that I introduced in a different PR, exactly due to the lack of proper testing :| )

.github/workflows/ci-base.yaml
```diff
@@ -13,31 +13,41 @@ defaults:
...
-      - name: Linting and style
-        run: |
-          python3 -m venv unit-test-venv
-          . unit-test-venv/bin/activate
-          pip install tox tox-pip-version
-          tox -e py3 -e py3-bionic -e py3-xenial
+      - name: Flake8
+        run: tox -e "${{ matrix.testenv.pyver }}-flake8"
+      - name: Unit
+        run: tox -e "${{ matrix.testenv.pyver }}-test"
```
paride, line 53: A single tox call (`tox -e pyX-flake8 -e pyX-test`) is slightly faster, but having envs in separate steps adds clarity to the results IMHO, so I think it's worth keeping it like this, especially given that the bulk of CI time is taken by the integration tests.

image

paride (approved):
LGTM!

lucasmoura (approved):

PR #2095, opened 2022-05-17
fix: add --dry-run option

lucasmoura:
## Proposed Commit Message
fix: add --dry-run option

Add support for --dry-run option for ua fix command. This option allow users to see what will happen in their system once if the fix command is actually used.

## Test Steps
Run the modified integration test added

## Desired commit type

## Checklist

### DIFF
--- a/docs/howtoguides/how_to_run_ua_fix_in_dry_run_mode.md
+++ b/docs/howtoguides/how_to_run_ua_fix_in_dry_run_mode.md
@@ -1,0 +1,50 @@
+# How to run fix command in dry run mode
+
+If you are unsure on what changes will happen on your system after you run `ua fix` to address a
+CVE/USN, you can use the `--dry-run` flag to see that packages will be installed in the system if
+the command was actually run. For example, this is the output of running `ua fix USN-5079-2 --dry-run`:
+
+```
+WARNING: The option --dry-run is being used.
+No packages will be installed when running this command.
+USN-5079-2: curl vulnerabilities
+Found CVEs:
+https://ubuntu.com/security/CVE-2021-22946
+https://ubuntu.com/security/CVE-2021-22947
+1 affected source package is installed: curl
+(1/1) curl:
+A fix is available in UA Infra.
+The machine is not attached to an Ubuntu Advantage (UA) subscription.
+To proceed with the fix, a prompt would ask for a valid UA token.
+{ ua attach TOKEN }
+UA service: esm-infra is not enabled.
+To proceed with the fix, a prompt would ask permission to automatically enable
+this service.
+{ ua enable esm-infra }
+{ apt update && apt install --only-upgrade -y curl libcurl3-gnutls }
+✔ USN-5079-2 is resolved.
+```
+
+You can see that using `--dry-run` will also indicate which actions would need to happen
+to completely address the USN/CVE. Here we can see that the package fix can only be accessed
+through the `esm-infra` service. Therefore, we need a UA subscription, as can be seen on this
+part of the output:
+
+```
+The machine is not attached to an Ubuntu Advantage (UA) subscription.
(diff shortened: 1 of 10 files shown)


lucasmoura:
@orndorffgrant @renanrodrigo @CalvoM we currently have a problem with the `--dry-run` flag for non-root users.
If the user is attached and run an `ua fix --dry-run` that requires a packages that comes from either `esm-infra` or `esm-apps`, we will say to the user that the machine is unattached. That is because of how config verify if we are attached:

https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/config.py#L439

That is because we need to read the `machine_token.json` file to produce the `machine_token` attribute. That is not feasible for non-root users at this moment.

we can:

* Change that check to only verify the existence of the `machine_token.json` file on the system.
* Wait for @CalvoM PR, which I will probably touch on this part of the code as well

Let me know what you all think about this

renanrodrigo:
@lucasmoura as we discussed, I have hit this issue before when dealing with `security-status`. I think the best approach is to wait for @CalvoM 's PR before trying to touch the logic in config, but in the meantime we can use the same workaround: instead of checking it on `Config`, call status directly and verify `status()['attached']`

features/fix.feature
```diff
@@ -101,114 +101,165 @@ Feature: Ua fix command behaviour
...
+        No packages will be installed when running this command..*
+        USN-5079-2: curl vulnerabilities
+        Found CVEs:
+        https://ubuntu.com/security/CVE-2021-22946
+        https://ubuntu.com/security/CVE-2021-22947
+        1 affected source package is installed: curl
+        \(1/1\) curl:
+        A fix is available in UA Infra.
+        The machine is not entitled to an UA subscription.
+        To proceed with fix, a prompt would ask for attaching a token.
```
orndorffgrant, line 177: Maybe put these messages about prompts in bold?

features/fix.feature
```diff
@@ -101,114 +101,165 @@ Feature: Ua fix command behaviour
...
+        UA service: esm-infra is not enabled.
+        To proceed with fix, a prompt would ask for the service to be enabled.
```
orndorffgrant, line 179: It might be nice to see the equivalent commands that would get run for attach and enable, like we normally see during a real fix operation:
```suggestion
        .*\{ ua attach .*\}.*
        UA service: esm-infra is not enabled.
        To proceed with fix, a prompt would ask for the service to be enabled.
        .*\{ ua enable esm-infra .*\}.*
```

lucasmoura:
@orndorffgrant  @renanrodrigo code updated

orndorffgrant (commented):
I tested it out and it looks good! Just some more suggestions on the messaging.

lucasmoura:
@orndorffgrant code updated

docs/howtoguides/how_to_run_ua_fix_in_dry_run_mode.md
```diff
@@ -0,0 +1,48 @@
...
+https://ubuntu.com/security/CVE-2021-22947
+1 affected source package is installed: curl
+(1/1) curl:
+A fix is available in UA Infra.
+The machine is not entitled to an UA subscription.
+To proceed with fix, a prompt would ask for attaching a token.
+{ ua attach TOKEN }
+UA service: esm-infra is not enabled.
+To proceed with fix, a prompt would ask for the service to be enabled.
+{ ua enable esm-infra }
```
orndorffgrant, line 22: I think these messages need to be updated in this example output

orndorffgrant:
Weird that github duplicated almost all my comments last review :/

orndorffgrant (approved):
This works great! Just a tiny update to the how to guide I think and its good to go

renanrodrigo (approved):
worksforme

PR #2100, opened 2022-05-20
docs: badges to advertise that same version is released to all LTS

orndorffgrant:
Only including LTS release badges.

I think this looks like a cool way to advertise the consistency of version across all LTS releases of `ua`


renanrodrigo (approved):
awesome

PR #2101, opened 2022-05-20
Handle old config items

orndorffgrant:
This does a couple things:
1. prevents `parse_config` from being called more than once during a simple cli call (e.g. `ua status` unattached)
2. logs a different warning for old config items vs invalid config items
3. only logs warnings for old/invalid config items to the log file (not stdout/stderr)
4. to accomplish (3) it does a minimal version of setup_logging in the middle of `parse_config`

So a few questions to discuss (on Monday at this point most likely):
* Is (2) reasonable long term? Or should we just treat old config items the same as invalid ones?
* Is (3) okay? or do we want to warn louder?
* Is there a better way to do (4)?
* Should we also try to fix the config for the user in postinst or otherwise?

## Test Steps

* in a jammy container
* edit uaclient.conf to include an unknown key (e.g. `bad: bad`) and the old item (i.e. `license_check_log_file: /var/log/log.log`)
* run `ua status`

Without this PR, you'll see a couple warning messages.

With this PR, you won't see any message on the console. But you will see two messages at the top of /var/log/ubuntu-advantage.log

## Desired commit type

## Checklist


renanrodrigo:
Unrelated, but:
While performing some testing on this branch, I noticed we are really interested in warning people when they have an invalid key in the config. I added a wrong key/value to check behavior, and got:
```
root@jammytest:~# ua version
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
WARNING:root:Ignoring invalid uaclient.conf key: whatever=oh_well
27.8~22.04.1
```

I guess we could take a look on this eventually.

paride (commented):
One very small inline comment, but this LGTM.

A more general question: do you think we should instruct the user in the `uaclient.conf` comment header comment to use `ua config` to edit the configuration? This way the user gets some sanity checks in the new configuration, and I personally find it's less surprising if a package edits its own config file when the file was not supposed to me edited manually.

debian/ubuntu-advantage-tools.postinst
```diff
@@ -355,6 +355,15 @@ remove_old_systemd_timers() {
     fi
 }
 
+remove_old_config_fields() {
+    PREVIOUS_PKG_VER="$1"
+    if dpkg --compare-versions "$PREVIOUS_PKG_VER" le "27.8"; then
+        if grep -q license_check_log_file /etc/ubuntu-advantage/uaclient.conf; then
```
paride, line 361: @orndorffgrant no idea where the inline comment went. TLDR was:
1. Maybe we can make `grep` a bit stricter: `grep -q '^license_check_log_file:'`.
2. I checked what mysql-8.0 does (https://git.launchpad.net/ubuntu/+source/mysql-8.0/tree/debian/mysql-server-8.0.postinst?id=e417d5e8ba04504a15491283ba64d82496f60542#n118) and it's similar, which is reassuring. I don't know why they do the `cmp` + `mv` thing, I think what we have here is fine.

uaclient/config.py
```diff
@@ -109,9 +109,12 @@ def __init__(self, cfg: Dict[str, Any] = None, series: str = None) -> None:
         if cfg:
             self.cfg_path = None
             self.cfg = cfg
+            self.invalid_keys = None
         else:
             self.cfg_path = get_config_path()
-            self.cfg = parse_config(self.cfg_path)
+            parsed_cfg, invalid_keys = parse_config(self.cfg_path)
```
lucasmoura, line 115: I think we can just do:
```python
self.cfg, self,invalid_keys = parse_config(self.cfg_path)
```

renanrodrigo (approved):

orndorffgrant:
@paride  we had another idea to maybe make this a bit better: moving the removal function from postinst to preinst.

This avoids a "would you like to keep your conffile" prompt in 1 case: if you only have the old field and no other changes to uaclient.conf.

Because of that we'd like you to re-review when you're back. If you agree with the change, I'll squash [3a4fb71](https://github.com/canonical/ubuntu-advantage-client/pull/2101/commits/3a4fb71f898b7b7c394413a7e553c92dc0094c45) and [9a7dfad](https://github.com/canonical/ubuntu-advantage-client/pull/2101/commits/9a7dfade372e05ab6cea660fff265cc4ce5916cc) together. If you disagree, I'll drop [9a7dfad](https://github.com/canonical/ubuntu-advantage-client/pull/2101/commits/9a7dfade372e05ab6cea660fff265cc4ce5916cc)

debian/ubuntu-advantage-tools.preinst
```diff
@@ -0,0 +1,22 @@
...
+    PREVIOUS_PKG_VER="$1"
+    if dpkg --compare-versions "$PREVIOUS_PKG_VER" le "27.8"; then
+        if grep -q "^license_check_log_file:" /etc/ubuntu-advantage/uaclient.conf; then
+            sed -i '/^license_check_log_file:.*$/d' /etc/ubuntu-advantage/uaclient.conf || true
+        fi
+    fi
+}
+
+case "$1" in
+    upgrade)
```
paride, line 15: Using preinst is nice, however there is a corner case: the user uninstalled u-a-t <= 27.8 *without purging*, and now wants to install it. Not purging means that the conf files of the package are preserved.

In this case the preinst script is called with as:
```
preinst install old-version
```
which is not handled at the moment. See https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called for doc on this.

Basically we need to also handle the `install)` case, e.g. like here:

https://sources.debian.org/src/unattended-upgrades/2.8/debian/preinst/

paride (changes requested):

debian/ubuntu-advantage-tools.preinst
orndorffgrant: @paride Thanks for noticing this! And that example was really helpful. Just pushed an update to resolve this

paride (approved):
LGTM, +1 modulo CI, thanks!

lucasmoura (approved):

PR #2105, opened 2022-05-25
WIP: java fips telemetry

orndorffgrant:
Drafting - some requirements not totally decided yet

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -2,7 +2,9 @@ import logging
 from typing import Any, Dict, List, Optional, Tuple
 
 from uaclient import (
+    apt,
     clouds,
+    data_types,
     event_logger,
     exceptions,
     messages,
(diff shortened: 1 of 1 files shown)


uaclient/contract.py
```diff
@@ -281,11 +322,27 @@ def _get_activity_info(self, machine_id: Optional[str] = None):
             if ent(self.cfg).user_facing_status()[0] == UserFacingStatus.ACTIVE
         ]
 
-        return {
-            "activityID": activity_id,
-            "activityToken": self.cfg.activity_token,
-            "resources": enabled_services,
-        }
+        activity_info = ActivityInfo(
```
lucasmoura, line 325: I think we can create the class only during the return call, just to make the code a little clearer.
Additionally, shouldn't we use `from_dict`  here to perform the type checks ?

orndorffgrant: Good point!

If we construct it with the constructor, then type checking is handled by mypy.
`from_dict` gives us the type checking when we're creating it from a json or yaml file or other untyped dict source.

uaclient/contract.py
```diff
@@ -281,11 +321,25 @@ def _get_activity_info(self, machine_id: Optional[str] = None):
...
+            java_package_matches = apt.search_installed_packages(
+                r"openjdk-(\d+)-jre-headless"
+            )
+            for java_package_match in java_package_matches:
+                version = java_package_match.groups[0]
+                fips_java_package_installed = apt.is_installed(
+                    "openjdk-{version}-fips".format(version=version)
+                )
+                if fips_java_package_installed:
+                    uses_fips_java = True
```
lucasmoura, line 335: I think we can add a `break` here

uaclient/contract.py
```diff
@@ -281,11 +321,25 @@ def _get_activity_info(self, machine_id: Optional[str] = None):
...
         ]
 
-        return {
-            "activityID": activity_id,
-            "activityToken": self.cfg.activity_token,
-            "resources": enabled_services,
-        }
+        uses_fips_java = False
+        if "fips" in enabled_services or "fips-updates" in enabled_services:
+            java_package_matches = apt.search_installed_packages(
```
lucasmoura, line 326: @orndorffgrant shouldn't we go the other way around ? Identify first which `openjdk-fips` package is installed and then just query for a specific version of `openjdk-jre-headless` package ? Or can we have more than one `openjdk-fips` package in the machine ?

orndorffgrant: I don't know :)

It may be the case that all the -fips packages get installed as part of enable fips, in which case the way it is now is better. Or maybe only one will get installed manually.

After we have more concrete requirements, we can come back to this

uaclient/contract.py
orndorffgrant: good point!

uaclient/apt.py
```diff
@@ -470,6 +470,30 @@ def get_installed_packages() -> List[str]:
     return out.splitlines()
 
 
+class PackageMatch:
```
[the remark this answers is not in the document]
lucasmoura: Fair enough, good point

orndorffgrant:
Closing as it doesn't seem we'll need this after all.

PR #2110, opened 2022-06-01
tests: add jammy pro test support (SC-868)

lucasmoura:
## Proposed Commit Message
tests: add jammy pro test support

## Test Steps
Run new the jammy pro tests

## Desired commit type

## Checklist

### DIFF
--- a/features/ubuntu_pro.feature
+++ b/features/ubuntu_pro.feature
@@ -227,56 +227,40 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
         And I run `ua auto-attach` with sudo
         And I run `ua status --wait` as non-root
         Then stdout matches regexp:
-            """
-            SERVICE       +ENTITLED  STATUS    DESCRIPTION
-            cc-eal        +yes +<cc-eal-s>  +Common Criteria EAL2 Provisioning Packages
-            """
+        """
+        SERVICE       +ENTITLED  STATUS    DESCRIPTION
+        cc-eal        +yes +<cc-eal-s>  +Common Criteria EAL2 Provisioning Packages
+        """
         Then stdout matches regexp:
-            """
-            esm-apps      +yes +enabled +UA Apps: Extended Security Maintenance \(ESM\)
-            esm-infra     +yes +enabled +UA Infra: Extended Security Maintenance \(ESM\)
-            fips          +yes +<fips-s> +NIST-certified core packages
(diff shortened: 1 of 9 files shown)


features/ubuntu_pro.feature
```diff
@@ -336,16 +337,69 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
...
-            """
+        """
+        .*status=0\/SUCCESS.*
+        """
 
         Examples: ubuntu release
            | release | fips-s   | cc-eal-s | cis-s    | infra-pkg | apps-pkg | cis_or_usg |
            | xenial  | disabled | disabled | disabled | libkrad0  | jq       | cis        |
            | bionic  | disabled | disabled | disabled | libkrad0  | bundler  | cis        |
            | focal   | disabled | n/a      | disabled | hello     | ant      | usg        |
```
CalvoM, line 348: Is there no inclusion of jammy in this list?

features/ubuntu_pro.feature
```diff
@@ -604,12 +658,12 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
...
+        """
+        .*status=0\/SUCCESS.*
+        """
 
         Examples: ubuntu release
            | release | fips-s   | cc-eal-s | cis-s    | infra-pkg | apps-pkg | livepatch | cis_or_usg |
            | xenial  | n/a      | disabled | disabled | libkrad0  | jq       | n/a       | cis        |
-           | bionic  | disabled | disabled | disabled | libkrad0  | bundler  | n/a       | cis        |
+           | bionic  | disabled | disabled | disabled | libkrad0  | bundler  | enabled   | cis        |
            | focal   | disabled | n/a      | disabled | hello     | ant      | enabled   | usg        |
```
CalvoM, line 669: same question..

features/ubuntu_pro.feature
orndorffgrant: @lucasmoura  would it be better or worse to break the enabling `usg`/`cis` on pro into it's own test that doesn't include jammy? Then the main pro test could be used for all series without duplicating

lucasmoura: @orndorffgrant just noticed that we also enable the `cis|usg` on the proxy tests. What if we keep it that only on those tests then ?

orndorffgrant: @lucasmoura good point, that works for me!

lucasmoura:
@orndorffgrant updated

PR #2111, opened 2022-06-01
refactor: use remove_file fn everywhere

orndorffgrant:
By using the util.remove_file everywhere we are removing files, we get
consistent logging whenever we remove a file. This will help debug
future issues.

I noticed we weren't logging the removal of some files while debugging a recent issue. I had to check the source to verify my understanding of what was happening. It would've been easier if we just logged it. I also noticed some duplicate functionality that this will simplify.

## Test Steps

All tests should still pass

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 import subprocess
+import sys
 import tempfile
 from functools import lru_cache
 from typing import Dict, List, Optional
@@ -355,7 +356,7 @@ def remove_repo_from_apt_auth_file(repo_url):
             [line for line in apt_auth.splitlines() if auth_prefix not in line]
         )
         if not content:
-            os.unlink(apt_auth_file)
+            util.remove_file(apt_auth_file)
         else:
             util.write_file(apt_auth_file, content, mode=0o600)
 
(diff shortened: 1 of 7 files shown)


uaclient/apt.py
```diff
@@ -455,14 +454,8 @@ def clean_apt_files(*, _entitlements=None):
             continue
         repo_file = ent_cls.repo_list_file_tmpl.format(name=ent_cls.name)
         pref_file = ent_cls.repo_pref_file_tmpl.format(name=ent_cls.name)
-
-        if os.path.exists(repo_file):
-            logging.info("Removing apt source file: %s", repo_file)
-            os.unlink(repo_file)
-
-        if os.path.exists(pref_file):
-            logging.info("Removing apt preferences file: %s", pref_file)
```
CalvoM, line 464: Was there are reason for this specific logging message?

orndorffgrant: Updated!

CalvoM (approved):
LGTM

uaclient/apt.py
```diff
@@ -455,14 +455,18 @@ def clean_apt_files(*, _entitlements=None):
             continue
         repo_file = ent_cls.repo_list_file_tmpl.format(name=ent_cls.name)
         pref_file = ent_cls.repo_pref_file_tmpl.format(name=ent_cls.name)
-
         if os.path.exists(repo_file):
-            logging.info("Removing apt source file: %s", repo_file)
-            os.unlink(repo_file)
-
+            print(
```
lucasmoura, line 459: Even though we are only calling this on function on `prerm`, I think this should be an `event.info` call just for consistency sake

orndorffgrant: @lucasmoura good point, updated!

lucasmoura (approved):

PR #2113, opened 2022-06-02
cli: add initial flow for magic-attach

lucasmoura:
## Proposed Commit Message
cli: add initial flow for magic-attach

Add initial flow for magic-attach in UA.
PS: This is just a draft for the flow, we need unit tests and proper integration tests for this feature yet.

## Test Steps
Verify if the flow is consistent with the proposed spec for this feature

## Desired commit type

## Checklist


uaclient/actions.py
```diff
@@ -51,6 +57,26 @@ def attach_with_token(
...
+
+    while num_min_to_wait > 0:
+        try:
+            response = ua_contract.request_magic_token(magic_id=magic_id)
+            return MagicAttachResult(
+                contract_token=response.get("contract_token"),
+            )
+        except exceptions.MagicAttachIDNotResolved:
+            num_min_to_wait -= 1
+            time.sleep(60)
```
orndorffgrant, line 71: I think the timeout should probably be specified in seconds, and the time between polls should be much lower, like 5 seconds. We don't want a user to have to wait a full minute after clicking the button in their browser for their install to continue.

uaclient/cli.py
```diff
@@ -485,6 +485,58 @@ def refresh_parser(parser):
     return parser
 
 
+def magic_attach_parser(parser):
```
orndorffgrant, line 488: If we get this into an API function, then I think subiquity will use that rather than a CLI subcommand.

We may want a user-facing CLI subcommand, but that probably won't require them to provide the ID. Instead, we'd automatically get that for them

orndorffgrant, line 488: If we get this into an API function, then I think subiquity will use that rather than a CLI subcommand.

We may want a user-facing CLI subcommand, but that probably won't require them to provide the ID. Instead, we'd automatically get that for them

uaclient/contract.py
```diff
@@ -177,6 +179,41 @@ def report_machine_activity(self):
...
+                            magic_id=magic_id
+                        )
+                    elif e.code == 401:
+                        raise exceptions.MagicAttachInvalidID(
+                            magic_id=magic_id
+                        )
+                    elif e.code == 404:
+                        raise exceptions.MagicAttachIDNotResolved(
+                            magic_id=magic_id
+                        )
```
orndorffgrant, line 208: I'm a big fan of the granular exceptions!

orndorffgrant, line 208: I'm a big fan of the granular exceptions!

uaclient/contract.py
```diff
@@ -177,6 +179,41 @@ def report_machine_activity(self):
...
+                raise e
+            with util.disable_log_to_console():
```
orndorffgrant, line 210: unrelated out-of-scope comment: Do we ever want logging to go to the console? Maybe we can remove the console logger entirely and use print/event.info the few times we do want to print to the console.

orndorffgrant, line 210: unrelated out-of-scope comment: Do we ever want logging to go to the console? Maybe we can remove the console logger entirely and use print/event.info the few times we do want to print to the console.

orndorffgrant (commented):
Looking good! I know this is early, but left some comments here.

uaclient/cli.py
orndorffgrant: resolving duplicate

uaclient/contract.py
orndorffgrant: resolving duplicate

uaclient/contract.py
orndorffgrant: resolving duplicate

uaclient/cli.py
lucasmoura: Indeed, once we have the API structure in place, I think we can retire this CLI entry. But for now, it will be useful for testing, like on integration tests

uaclient/actions.py
orndorffgrant: If someone knows what they're doing and is prepared (already has browser open, logged in, knows what subscription they're going to select or will use auto-selected free subscription), I could see it only taking a couple seconds.

I don't think there's a downside to more frequent polling aside from extra requests/traffic to the server, but I think CS is expecting more frequent polling - @alnvdl how often were you expecting `ua` to poll for the magic-attach?

lucasmoura:
@orndorffgrant @renanrodrigo @CalvoM It seems that we will no longer need that work on UA, since we will not be required to fetch the token anymore, as updated in the spec. I will leave this PR open just until we have confirmation that this new flow is final

lucasmoura:
Updating this PR to use the API structure instead

alnvdl-work (commented):
A bunch of comments regarding the API usage.

uaclient/api/u/pro/attach/magic/initiate/v1.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class InitiateSimplifiedAttachResult(DataObject):
+    fields = [
+        Field("_schema", StringDataValue),
+        Field("confirmation_code", StringDataValue),
+        Field("token", StringDataValue),
+        Field("expires", StringDataValue),
+        Field("user_email", StringDataValue),
+    ]
```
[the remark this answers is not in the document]
lucasmoura: Yes, that would be awesome to have

uaclient/api/u/pro/attach/magic/initiate/v1.py
```diff
@@ -0,0 +1,45 @@
...
+from uaclient.data_types import DataObject, Field, StringDataValue
+
+
+class InitiateSimplifiedAttachResult(DataObject):
+    fields = [
+        Field("_schema", StringDataValue),
+        Field("confirmation_code", StringDataValue),
+        Field("token", StringDataValue),
+        Field("expires", StringDataValue),
+        Field("user_email", StringDataValue),
```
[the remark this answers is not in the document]
lucasmoura: True, but I think this endpoint should only be used to create one and at that stage it should not contain those variables, so I think it is okay for us to not include them here

uaclient/cli.py
```diff
@@ -504,6 +504,35 @@ def refresh_parser(parser):
...
+        raise exceptions.UserFacingError(
+            msg=messages.MAGIC_ATTACH_INVALID_WAIT_TIME.msg,
+            msg_code=messages.MAGIC_ATTACH_INVALID_WAIT_TIME.name,
+        )
+
+    event.info("Fetching UA token using magic id: {}".format(args.magic_id))
+    magic_attach_result = actions.magic_attach(
+        cfg=cfg,
+        magic_id=args.magic_id,
+        num_min_to_wait=args.wait_time,
```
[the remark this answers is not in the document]
lucasmoura: Yeah, I was wondering about if that would needed, but fair enough, I will remove it then

uaclient/contract.py
```diff
@@ -171,6 +173,106 @@ def report_machine_activity(self):
             machine_token["activityInfo"] = response
             self.cfg.write_cache("machine-token", machine_token)
 
+    def request_magic_attach_id_info(self, magic_id: str) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
lucasmoura: Good point

uaclient/contract.py
```diff
@@ -171,6 +173,106 @@ def report_machine_activity(self):
...
+            kwargs = {
+                "headers": headers,
+                "method": "DELETE",
+            }
+            self.request_url(url, **kwargs)
+        except exceptions.UrlError as e:
+            if isinstance(e, exceptions.ContractAPIError):
+                if hasattr(e, "code"):
+                    if e.code == 400:
+                        raise exceptions.MagicAttachIDExpiredOrActivated(
```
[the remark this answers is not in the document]
alnvdl-work: So, you should handle it, but it only means the token is already activated.

If it's expired, you will get a 401.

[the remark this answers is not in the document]
lucasmoura: Just to double check, you mean that we should not error out to the user when trying to revoke an activated token, just state that it is already activated, right ?

uaclient/contract.py
```diff
@@ -171,6 +173,90 @@ def report_machine_activity(self):
...
+            kwargs = {
+                "headers": headers,
+                "method": "DELETE",
+            }
+            self.request_url(url, **kwargs)
+        except exceptions.UrlError as e:
+            if isinstance(e, exceptions.ContractAPIError):
+                if hasattr(e, "code"):
+                    if e.code == 400:
+                        raise exceptions.MagicAttachTokenExpiredOrActivated(
```
alnvdl-work, line 246: What I mean in the other comment is: 400 can only mean "this magic attach token has already been activated and cannot be revoked because of that".

So `MagicAttachTokenExpiredOrActivated` should become `MagicAttachTokenAlreadyActivated`.

uaclient/contract.py
```diff
@@ -171,6 +173,90 @@ def report_machine_activity(self):
...
+                            magic_token=magic_token
+                        )
+                    elif e.code == 401:
+                        raise exceptions.MagicAttachTokenNotFound(
```
alnvdl-work, line 250: Same thing as the comment I left in `get_magic_attach_token_info`.

uaclient/contract.py
```diff
@@ -171,6 +173,90 @@ def report_machine_activity(self):
...
+        headers = self.headers()
+        headers.update({"Authorization": "Bearer {}".format(magic_token)})
+        url = API_V1_MAGIC_ATTACH
+
+        try:
+            response, _ = self.request_url(url, headers=headers)
+        except exceptions.UrlError as e:
+            if isinstance(e, exceptions.ContractAPIError):
+                if hasattr(e, "code"):
+                    if e.code == 401:
```
alnvdl-work, line 191: For all endpoints that accept a "Bearer <magic-attach-token>" kind of auth, `401` always means one of: not found, invalid format (e.g., someone said the token is "abc123") or expired already.

The message from ua-contracts will not distinguish between these cases, and it will just say "unauthorized".

I'd suggest: `exceptions.MagicAttachTokenError` with the following message to the user:
> The magic attach token is invalid, has expired or never existed.

uaclient/messages.py
```diff
@@ -466,6 +466,29 @@ class TxtColor:
...
+    "magic-attach-token-expired-or-activated",
+    (
+        "The magic attach token: {magic_token} is already "
+        "activated or expired."
+    ),
+)
+
+MAGIC_ATTACH_EXPIRED_TOKEN = FormattedNamedMessage(
+    "magic-attach-token-expired",
+    "The magic attach token: {magic_token} has expired or never existed.",
```
alnvdl-work, line 479: I'd recommend against printing the token in all these cases. There's nothing the user can do with this information, and it just increases the chances of toking ending up in logs (even though they'd be invalid tokens).

alnvdl-work (commented):
A few more comments.

uaclient/contract.py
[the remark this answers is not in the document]
alnvdl-work: For all endpoints that accept a "Bearer <magic-attach-token>" kind of auth, `401` always means one of: not found, invalid format (e.g., someone said the token is "abc123") or expired already.

The message from ua-contracts will not distinguish between these cases, and it will just say "unauthorized".

I'd suggest: `exceptions.MagicAttachTokenError` with the following message to the user:
> The magic attach token is invalid, has expired or never existed.

uaclient/messages.py
lucasmoura: Good point, I will remove this

lucasmoura:
@alnvdl-work @renanrodrigo @orndorffgrant @CalvoM PR updated and now also following the API structure on #2140

uaclient/api/tests/test_magic_attach_wait.py
```diff
@@ -0,0 +1,48 @@
+import mock
```
orndorffgrant, line 1: Can we rename this file to `test_u_pro_attach_magic_wait_v1.py` or `test.u.pro.attach.magic.wait.v1.py`

uaclient/api/u/pro/attach/magic/initiate/v1.py
```diff
@@ -0,0 +1,59 @@
...
+        confirmation_code=initiate_resp["confirmationCode"],
+        token=initiate_resp["token"],
+        expires=initiate_resp["expires"],
+        user_email=initiate_resp["userEmail"],
+    )
+
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="Magic Attach Initiate",
```
orndorffgrant, line 56: Something we need to decide on. Should these names have a standard format? I was imagining `PascalCase` but I could be convinced of "Title Case" with spaces.

uaclient/api/u/pro/attach/magic/wait/v1.py
```diff
@@ -0,0 +1,85 @@
...
+        self.contract_token = contract_token
+
+
+def wait(
+    options: MagicAttachWaitOptions,
+    cfg: UAConfig,
+) -> MagicAttachWaitResult:
+    contract = UAContractClient(cfg)
+
+    while True:
```
orndorffgrant, line 56: Should we have a `maximum_attempts` or similar? I'm afraid of relying on any external source - no matter how reliable, to break an infinite loop.

uaclient/contract.py
```diff
@@ -171,6 +173,84 @@ def report_machine_activity(self):
...
+
+        When the magic token is registered, it will contain new fields
+        that will allow us to know that the attach process can proceed
+        """
+        headers = self.headers()
+        headers.update({"Authorization": "Bearer {}".format(magic_token)})
+        url = API_V1_MAGIC_ATTACH
+
+        try:
+            response, _ = self.request_url(url, headers=headers)
```
orndorffgrant, line 187: why the extra variable?
```suggestion
        try:
            response, _ = self.request_url(API_V1_MAGIC_ATTACH, headers=headers)
```

uaclient/contract.py
```diff
@@ -171,6 +173,84 @@ def report_machine_activity(self):
...
+        headers = self.headers()
+        url = API_V1_MAGIC_ATTACH
+
+        try:
+            kwargs = {
+                "headers": headers,
+                "method": "POST",
+                "data": {"email": email},
+            }
+            response, _ = self.request_url(url, **kwargs)
```
orndorffgrant, line 214: Why not pass the params directly?
```suggestion
        try:
            response, _ = self.request_url(
                API_V1_MAGIC_ATTACH,
                headers=headers,
                method="POST",
                data={"email": email},
            )
```

uaclient/contract.py
```diff
@@ -171,6 +173,84 @@ def report_machine_activity(self):
...
+        headers = self.headers()
+        headers.update({"Authorization": "Bearer {}".format(magic_token)})
+        url = API_V1_MAGIC_ATTACH
+
+        try:
+            kwargs = {
+                "headers": headers,
+                "method": "DELETE",
+            }
+            self.request_url(url, **kwargs)
```
orndorffgrant, line 239: Same suggestion here of passing params as named parameters instead of as kwargs

uaclient/contract.py
```diff
@@ -171,6 +173,84 @@ def report_machine_activity(self):
...
+        except exceptions.UrlError as e:
+            if isinstance(e, exceptions.ContractAPIError):
```
orndorffgrant, line 241: This pattern is in all three new functions: Would it be easier to `except` both kinds of exception separates?
```suggestion
        except exceptions.ContractAPIError as e:
            ...contract specific stuff
        except exceptions.UrlError as e:
            ...generic stuff
```

Aside/future item: I think ContractAPIError shuold probably extend HTTPError from the stdlib so we can rely on getting http codes without checking

orndorffgrant (commented):
Looking good!

features/schemas/magic_attach.json
```diff
@@ -0,0 +1,88 @@
+{
```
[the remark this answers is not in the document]
lucasmoura: Makes sense, I will update it

uaclient/api/u/pro/attach/magic/wait/v1.py
lucasmoura: That makes sense, we should define the default value for this check. It seems that the token has a lifespan of 10 minutes. With a delay of 10s, we could check `maximum_attempts < 70` just to get an extra minute there. What do you think ?

renanrodrigo (approved):
Looks great, thanks for the work and for the patience (:
I have nothing to add except a suggestion  for a couple more behave tests

features/api.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Client behaviour for the API endpoints
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: API invalid endpoint or args
```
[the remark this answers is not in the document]
lucasmoura: The only problem with the warning test is that I don't call directly any of the endpoints here, except initiate.
If we use initiate for that, we would need to create distinct steps for the warning situation because we should also revoke the token at the end. So I am still unsure if it is worth it, but let me know what you all think

lucasmoura:
@renanrodrigo @orndorffgrant updated

features/api.feature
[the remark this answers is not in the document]
renanrodrigo: Well it is covered by unit tests anyway, we can add it later to behave using a simpler scenario. Leave it on me.

orndorffgrant (approved):
Excellent

PR #2114, opened 2022-06-02
cli: add missing refresh option on help text

lucasmoura:
## Proposed Commit Message
cli: add missing refresh option on help text

## Test Steps
Verify that the `messages` option is now appearing

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -463,24 +463,30 @@ def refresh_parser(parser):
         "Refresh existing Ubuntu Advantage contract and update services."
     )
     parser.usage = USAGE_TMPL.format(
-        name=NAME, command="refresh [contract|config]"
+        name=NAME, command="refresh [contract|config|messages]"
     )
+
     parser._optionals.title = "Flags"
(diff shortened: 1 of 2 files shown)


CalvoM (commented):
Currently, the help message is this
```
Target to refresh. `ua refresh contract` will update contract details from the server and perform any updates necessary. `ua refresh config` will reload /etc/ubuntu-advantage/uaclient.conf and perform any changes necessary. `ua refresh messages` will refresh the APT and MOTD messages
                        associated with UA. `ua refresh` is the equivalent of `ua refresh config && ua refresh contract && ua refresh motd`.
```
I find it not easily readable, what do you think?

lucasmoura:
@CalvoM @orndorffgrant I have updated the whole help content for `refresh`. Let me know what you think.
Also, I am not making that update globally, since this would require for us to manipulate each subparser individually, which I don't think it is worth the refactoring effort now

orndorffgrant (approved):
Looks great!

PR #2117, opened 2022-06-03
collect-logs: collect rotated log files

orndorffgrant:
## Test Steps

rotate your log files
run `ua collect-logs`
ensure the resulting tarfile includes the rotated logs

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -3,6 +3,7 @@
 """Client to manage Ubuntu Advantage services on a machine."""
 
 import argparse
+import glob
 import json
 import logging
 import os
@@ -43,6 +44,7 @@ from uaclient.defaults import (
     CONFIG_FIELD_ENVVAR_ALLOWLIST,
     DEFAULT_CONFIG_FILE,
     DEFAULT_LOG_FORMAT,
+    DEFAULT_LOG_PREFIX,
     PRINT_WRAP_WIDTH,
 )
 from uaclient.entitlements.entitlement_status import (
@@ -1475,7 +1477,8 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
                 return_codes=[0, 3],
             )
 
-        ua_logs = (
+        # include cfg log files here because they could be set to non default
+        state_files = [
             cfg.cfg_path or DEFAULT_CONFIG_FILE,
             cfg.log_file,
             cfg.timer_log_file,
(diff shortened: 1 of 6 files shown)


uaclient/cli.py
```diff
@@ -1481,14 +1483,15 @@ def action_collect_logs(args, *, cfg: config.UAConfig):
...
                 if issubclass(entitlement, entitlements.repo.RepoEntitlement)
             ),
-        )
+        ]
 
-        for log in ua_logs:
-            if os.path.isfile(log):
-                log_content = util.load_file(log)
+        # also get default logrotated log files
+        for f in state_files + glob.glob("/var/log/ubuntu-advantage*"):
```
lucasmoura, line 1489: We can set the path on the logs in `uaclient.conf`, therefore, I think we should update the logic to read that path directly from the `config` module

uaclient/tests/test_cli_collect_logs.py
```diff
@@ -51,6 +51,10 @@ def test_collect_logs_help(self, _m_resources, _getuid, capsys):
         out, _err = capsys.readouterr()
         assert re.match(HELP_OUTPUT, out)
 
+    @mock.patch(
+        M_PATH + "glob.glob",
+        return_value=["ubuntu-advantage.log", "ubuntu-advantage.log.1"],
+    )
```
lucasmoura, line 57: Don't know if this is feasible,  but can't we mock `os.path.isfile` just to check which files are being read by the command ?

uaclient/cli.py
orndorffgrant: @lucasmoura `cfg.log_file`, `cfg.timer_log_file`, and `cfg.daemon_log_file` are all included in `state_files`.

Logrotate is hardcoded to `/var/log/ubuntu-advantage*.log`, so I used a similar glob to what we have hardcoded there. 

Aside, this means non-default log files aren't getting rotated, but I think thats uncommon enough to not worry about(?)

The constant `/var/log/ubuntu-advantage` should probably not be hardcoded here though, and in `defaults.py` instead, I can make that change.

Thoughts?

uaclient/tests/test_cli_collect_logs.py
orndorffgrant: Yeah I can do something like that, good idea!

uaclient/cli.py
lucasmoura: Ohh I missed that logrotate was using that path directly, fair enough.
And I agree that this should not be a huge issue at the moment, since it should be a rare situation.
And I also agree we can create the constant on the `defaults.py` module

CalvoM (approved):
LGTM

orndorffgrant:
@lucasmoura updated

lucasmoura (approved):

PR #2118, opened 2022-06-03
apt-hook: replace golang with cpp for json-hook

orndorffgrant:
golang executable was around 2Mb. cpp version is less than 200Kb.

## Test Steps

Run the new integration test

## Desired commit type

## Checklist


renanrodrigo (approved):
lgtm!
Confirmed reduced installed size from 2.4MB to 686kB on Jammy

test.txt
```diff
@@ -0,0 +1,6 @@
+line one
```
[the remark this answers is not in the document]
orndorffgrant: Hahaha oops I'll get rid of this :)

lucasmoura:
I know we don't have unit test support even for `hook.cc`, but I think it is a good idea for us to add that for our hooks. But this could be a different PR for sure

apt-hook/json-hook.cc
```diff
@@ -0,0 +1,302 @@
...
+        }
+        std::cout << count_msgs[i];
+    }
+    std::cout << std::endl;
+}
+
+int main(int argc, char *argv[])
+{
+    (void) argc;
+    (void) argv;
```
lucasmoura, line 223: Sorry, long time without looking at C++ code, but why do we need that extra definition here ?

apt-hook/json-hook.cc
```diff
@@ -0,0 +1,302 @@
...
+
+    char *fd_c_str = getenv("APT_HOOK_SOCKET");
+    if (fd_c_str == NULL) {
+       return 0;
+    }
+    std::string fd_str(fd_c_str);
+    if (fd_str == "") {
+       return 0;
+    }
```
lucasmoura, line 232: If we are returning 0 here, aren't we masking potential issues related to the hook ?

apt-hook/json-hook.cc
```diff
@@ -0,0 +1,302 @@
...
+    int fd = stoi(fd_str);
+    __gnu_cxx::stdio_filebuf<char> apt_socket_in(fd, std::ios::in);
+    __gnu_cxx::stdio_filebuf<char> apt_socket_out(fd, std::ios::out);
+    std::istream socket_in(&apt_socket_in);
+    std::ostream socket_out(&apt_socket_out);
+
+
```
lucasmoura, line 239: We can remove the extra line here

apt-hook/json-hook.cc
```diff
@@ -0,0 +1,302 @@
...
+    bool success = false;
+
+    // Read hello message, verify version, and get jsonrpc id
+    jsonrpc_request hello_req;
+    success = read_jsonrpc_request(socket_in, hello_req);
+    if (!success) {
+        return 0;
+    }
+    if (hello_req.method != "org.debian.apt.hooks.hello" || hello_req.notification) {
```
lucasmoura, line 248: Why do we need the `OR` clause here ?

apt-hook/json-hook.cc
```diff
@@ -0,0 +1,302 @@
...
+    }
+    if (stats_req.method == "org.debian.apt.hooks.install.statistics") {
+        security_package_counts counts;
+        success = count_security_packages_from_apt_stats_json(stats_req.params, counts);
+        if (success) {
+            print_counts(counts);
+        }
+    }
+    json_object_put(stats_req.root_msg);
+
```
lucasmoura, line 291: We can remove the extra line here

apt-hook/json-hook.cc
```diff
@@ -0,0 +1,302 @@
...
+        return 0;
+    }
+    if (stats_req.method == "org.debian.apt.hooks.install.statistics") {
+        security_package_counts counts;
+        success = count_security_packages_from_apt_stats_json(stats_req.params, counts);
+        if (success) {
+            print_counts(counts);
+        }
+    }
+    json_object_put(stats_req.root_msg);
```
lucasmoura, line 290: Is it possible that `stats_req.root_msg` is NULL pointer during this call ?

apt-hook/json-hook.cc
```diff
@@ -0,0 +1,302 @@
...
+
+        if (origin_origin == from_origin && string_ends_with(origin_archive, archive_ends_with)) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+struct security_package_counts {
```
lucasmoura, line 105: We could use a class here and initialize the count variables

apt-hook/json-hook.cc
orndorffgrant: These prevent a compiler warning about the parameters being unused

apt-hook/json-hook.cc
orndorffgrant: https://www.jsonrpc.org/specification#notification

A JSONRPC notification is a message without an id. The hello method should not be a notification because we need to respond to it with the corresponding id.

apt-hook/json-hook.cc
orndorffgrant: We don't `json_object_put` anything during the processing, so I don't think so.

apt-hook/json-hook.cc
orndorffgrant: I added a couple extra lines to separate chunks of logic inside `main` (including this one). I can remove them if you'd like though

apt-hook/json-hook.cc
orndorffgrant: We could :) do you think it is needed here? I feel that the use-case for this struct is simple enough and we're initializing the variables in the one place it matters so it's okay. But I could be convinced otherwise :)

orndorffgrant:
@lucasmoura @renanrodrigo I added boost tests as well

apt-hook/json-hook.cc
orndorffgrant: WRT to return codes: Our old hook always returns 0 (unless there is a panic), so returning 0 in these cases is keeping the same behavior as the old hook.

lucasmoura: Okay, fair enough, but were those go error message logged somewhere in the apt logs ? If yes, I think we still print them so we can better understand failure scenarios

apt-hook/json-hook.cc
lucasmoura: I would remove it, but no worries if you want to keep it

renanrodrigo (approved):

apt-hook/json-hook.cc
orndorffgrant: Removed!

julian-klode (commented):
Looks ok. I'd really love to see this translatable in a future version so we don't have English popping up in e.g. an otherwise German system, but that was not the case before either, so not an issue.

apt-hook/json-hook.cc
orndorffgrant: Oh that that would be nice! And cool, until then we'll print to stderr.

orndorffgrant:
Thanks for taking a look @julian-klode! Got it - yeah the lack of translation support isn't a regression so I won't hold up this transition on it, but I'll add it to our backlog.

orndorffgrant:
@lucasmoura added warnings printed to stdout

lucasmoura (approved):

PR #2119, opened 2022-06-07
tests: use local PPAs for development behave upgrade tests (SC-703)

renanrodrigo:
Use Local PPAs when testing local debs.
This delivers three small improvements:
- Fixes our Bionic -> Focal upgrade test, which had a dependency incompatibility
- Frees our other upgrade tests from the workaround that uninstalled packages to solve this incompatibility
- Makes sure we test the actual upgrade of the package, including all related scripts, at development time, rendering us less surprises during SRU testing and enabling us to test installation and configuration of the package across releases.

## Test Steps
Run the updated integration tests.

## Desired commit type

## Checklist


paride:
@renanrodrigo sanity check for me: the basic idea here is to always use kinetic to serve the local ppa, no matter what the release under test is, and rely on the fact that support for zstd has been backported to dpkg on all the Ubuntu releases. Is this correct?

features/steps/steps.py
```diff
@@ -1047,6 +1047,63 @@ def systemd_memory_usage_less_than(context, release, mb_limit):
...
+    repo_line = "deb [trusted=yes] http://{}:8000 {} main".format(
+        context.instances["ppa"].ip, release
+    )
+    repo_file = "/etc/apt/sources.list.d/local-ua.list"
+    when_i_run_shell_command(
+        context, "printf '{}\n' > {}".format(repo_line, repo_file), "with sudo"
+    )
+    when_i_run_command_on_machine(
+        context,
+        "sh -c 'nohup python3 -m http.server --directory ~/.aptly/public > /dev/null 2>&1 &'",  # noqa
```
paride, line 1076: TIL that aptly has an `aptly serve` subcommand that should do exactly this. Did you try it instead of using `http.server`? Nothing bad with it, but as we're using aptly already...

orndorffgrant (approved):
Works for me!

PR #2120, opened 2022-06-09
contract: send platform info for resource requests

lucasmoura:
## Proposed Commit Message
contract: send platform info for resource requests

Right now, when attached, we don't send any platform information when requesting machine token resources. This is because the contract side that information stored on their side. To make this process better, can send that request directly to the contract backend.

## Test Steps
Verify that the commands that use the modified endpoint still work as expected

## Desired commit type

## Checklist


pandrey2003 (approved):

alnvdl (approved):
LGTM, thanks for doing this so fast!

renanrodrigo (approved):

orndorffgrant (approved):

PR #2121, opened 2022-06-10
tests: consolidate installation source for u-a-t in behave

renanrodrigo:
Create a single variable to handle where UA is installed from + some code reorg to make it clearer.

## Test Steps
Run the integration tests using `UACLIENT_BEHAVE_INSTALL_FROM`.

## Desired commit type

## Checklist


dev-docs/howtoguides/testing.md
```diff
@@ -38,10 +38,17 @@ By default, integration tests will do the following on a given cloud platform:
...
+- `archive`: install the latest version available in the archive, not adding any PPA
+- `proposed`: install the package from the -proposed pocket - specially useful for SRU testing (see [the release guide](how_to_release_a_new_version_of_ua.md))
+- `custom`: install from a custom provided PPA. If set, then two other variables need to be set: `UACLIENT_BEHAVE_CUSTOM_PPA=<PPA URL>` and `UACLIENT_BEHAVE_CUSTOM_PPA_KEYID=<signing key for the PPA>`.
+- `local`: install from a local copy of the ubuntu-advantage-client source code
+
+`local` is particularly useful, as it runs the suite against the local code, thus including and validating the latest changes made. It is advised to run any related integration tests against local code changes before pushing them to be reviewed.
+
+> Note that, by default, we cache the source even when `UACLIENT_BEHAVE_INSTALL_FROM=local`. This means that if you change the python code locally and want to run the behave tests against your new version, you need to either delete the cache (`rm /tmp/uaclient_behave`) or also set `UACLIENT_BEHAVE_CACHE_SOURCE=0`.
```
lucasmoura, line 51: We should not need to do this right now, right ? We have mechanism that should regenerate the package if there are any changes on the code

features/environment.py
```diff
@@ -326,6 +336,15 @@ def from_environ(cls, config) -> "UAClientBehaveConfig":
...
                     bool_value = False
                 kwargs[key] = bool_value
+
+        if "install_from" in kwargs:
+            install_from = kwargs["install_from"]
+            try:
+                kwargs["install_from"] = InstallationSource(install_from)
+            except ValueError:
+                print("Unknown value for install_from:", install_from)
+                kwargs["install_from"] = InstallationSource.DAILY
```
lucasmoura, line 346: I think we should just raise the `Exception` here and not try to run the tests with a default option

features/environment.py
```diff
@@ -641,58 +658,21 @@ def create_instance_with_uat_installed(
...
-            user_data += USERDATA_APT_SOURCE_PPA.format(
-                ppa_url=ppa, ppa_keyid=ppa_keyid
-            )
     inst = context.config.cloud_manager.launch(
         instance_name=name, series=series, user_data=user_data
     )
     instance_id = context.config.cloud_manager.get_instance_id(inst)
 
+    deb_paths = None
+    if context.config.install_from.value == InstallationSource.LOCAL.value:
```
lucasmoura, line 673: Do can't we just compare here without calling `value` ?

features/environment.py
```diff
@@ -704,6 +684,49 @@ def create_instance_with_uat_installed(
...
+        ppa_keyid = DEFAULT_UA_PPA_KEYID
+    elif config.install_from.value == InstallationSource.STAGING.value:
+        user_data += USERDATA_RUNCMD_PIN_PPA.format("staging")
+        ppa = UA_PPA_TEMPLATE.format("staging")
+        ppa_keyid = DEFAULT_UA_PPA_KEYID
+    elif config.install_from.value == InstallationSource.CUSTOM.value:
+        ppa = config.custom_ppa or ""
+        if ppa.startswith("ppa:"):
+            ppa = ppa.replace("ppa:", "http://ppa.launchpad.net/") + "/ubuntu"
+        ppa_keyid = config.custom_ppa_keyid or ""
```
lucasmoura, line 721: Should we fail loudly here if any of the `ppa` variables are unset ?

features/environment.py
```diff
@@ -704,6 +684,49 @@ def create_instance_with_uat_installed(
...
+
+    user_data += USERDATA_APT_SOURCE_PPA.format(
+        ppa_url=ppa, ppa_keyid=ppa_keyid
+    )
+
+    return user_data
```
orndorffgrant, line 727: Can we `logging.debug` the user_data that ends up getting used?

features/environment.py
```diff
@@ -17,8 +18,18 @@
...
-DAILY_PPA = "http://ppa.launchpad.net/ua-client/daily/ubuntu"
-DAILY_PPA_KEYID = "6E34E7116C0BC933"
+
+class InstallationSource(Enum):
+    ARCHIVE = "archive"
+    LOCAL = "local"
+    DAILY = "daily"
+    STAGING = "staging"
+    PROPOSED = "proposed"
+    CUSTOM = "custom"
```
orndorffgrant, line 28: do we need a stable ppa option?

orndorffgrant: we can always use custom if not

dev-docs/howtoguides/testing.md
renanrodrigo: @lucasmoura The comment is actually accurate. The code expects you to set `cache_source=0` otherwise it will reuse the cached files that are there, even if the hash changes. You have to erase the tarball manually.
IMHO this is a bug, and we should:
- always build the packages based only on their hashes
- get rid of the cache_source variable completely

Let me know if you agree, and I can tackle this.

features/environment.py
renanrodrigo: I didn't add it because of the low (if any) usage of our stable PPA. If you [check the files](https://launchpad.net/~ua-client/+archive/ubuntu/stable), we are still 27.7 there, and 27.8 has been around for a while. We can use custom, yes.

However, it would be straightforward to include it anyway if you think that will be useful.

features/environment.py
renanrodrigo: Now this one is interesting - there is a known caveat for `Enum` comparison, due to how the import machinery works.
I found it weird because in our case there should be no `import`, the `Enum` subclass is defined in the same file where we use it. However, behave does some """trick""" in their runtime which screws this up.

Everything in `environment.py` gets to....... `builtins`. I am no Java specialist, so I don't know if it makes any sense. But take a look on why the direct comparison was failing (with a debugger set right after the config gets instantiated with `from_environ`):
```
(Pdb) InstallationSource.__module__
'builtins'
(Pdb) id(InstallationSource)
94603178161200
(Pdb) id(sys.modules["features.environment"].InstallationSource)
94603179143200
```
Python sees two different Enums with the same name. Now, when we are actually doing the comparison, the config value comes from the `builtins` Enum which behave creates, and the comparison is done against the original one:
```
(Pdb) InstallationSource.__module__
'features.environment'
(Pdb) config.install_from
<InstallationSource.DAILY: 'daily'>
(Pdb) id(config.install_from)
140671518494912
(Pdb) id(InstallationSource("daily"))
140671515232864
```

So the way I fixed it was cheating on how behave screws up the import machinery: I moved the Enum to the `utils` file. Since only the `environment` file "goes to `builtins`", behave is forced to import the Enum from utils when creating the config, and then we do the same when running the comparison. Back to config creation time, this time with the imported Enum:
```
(Pdb) InstallationSource.__module__
'features.util'
(Pdb) id(InstallationSource)
94407529466464
(Pdb) id(sys.modules["features.environment"].InstallationSource)
94407529466464
```

It made me question the structure a little: we could have only the "behave specific" functions (like `before_all` or `after_all` in the `environment.py` and have a separate `config.py` for our things... But to avoid the potential overkill, I just moved it to utils right now (:

dev-docs/howtoguides/testing.md
```diff
@@ -38,10 +38,17 @@ By default, integration tests will do the following on a given cloud platform:
  * Install the appropriate ubuntu-advantage-tools and ubuntu-advantage-pro deb
  * Run the integration tests on that instance.
 
-The testing can be overridden to run using a local copy of the ubuntu-advantage-client source code instead of the daily PPA by providing the following environment variable to the behave test runner:
-```UACLIENT_BEHAVE_BUILD_PR=1```
+The testing can be overridden to run installing ubuntu-advantage-tools from other places instead of the daily PPA by providing `UACLIENT_BEHAVE_INSTALL_FROM` to the behave test runner. The default is `UACLIENT_BEHAVE_INSTALL_FROM=daily`, and the other available options are:
```
lucasmoura, line 41: I think we can change to `The testing can be overridden to install ubuntu-advantage-tools from...`

dev-docs/howtoguides/testing.md
lucasmoura: I agree, we can remove that variable

features/environment.py
```diff
@@ -704,6 +669,57 @@ def create_instance_with_uat_installed(
...
+        return ""
+
+    user_data = "#cloud-config\n"
+
+    if "pro" in config.machine_type:
+        user_data = USERDATA_BLOCK_AUTO_ATTACH_IMG
+
+    if config.userdata_file and os.path.exists(config.userdata_file):
+        with open(config.userdata_file, "r") as stream:
+            user_data = stream.read()
```
lucasmoura, line 688: Do we still need to support that ? I think we are not passing any custom `user_data` in the code are we ? If not, I think we can drop this.

renanrodrigo: If you want, you can pass your own `user_data` to cloud-init using this. I think it can be useful in many ways - for instance, to test any cloud-init specific integration with UA client. I agree it is rare, but still can see where it goes.

lucasmoura: That's fair, but if that happens, we could create a specific step for that instead of allowing any type of user data to be passed to the tests

renanrodrigo: I see. Having a step would make it more explicit that we are passing different user data for sure.
However, the underlying logic will be in this function anyway. So I am marking as a future task to remove the config environment variable and change it to read from a step.

features/environment.py
orndorffgrant: The fact that our stable PPA is out of date isn't a good thing and shouldn't be a reason not to include it here :upside_down_face: 

It was useful recently for the interim 27.1 that was only released to the stable ppa, so I'm leaning towards including the `stable` target if it's easy enough to add.

features/environment.py
orndorffgrant: Wow, yeah I didn't know that about `environment.py`. I agree that in some future PR, we should refcator so that only behave specific functionality that should be treated this way is in that file, and everything else can get factored out.

features/environment.py
renanrodrigo: Sure will do

features/environment.py
```diff
@@ -704,6 +669,57 @@ def create_instance_with_uat_installed(
     return inst
 
 
+def _get_user_data_for_instance(
```
orndorffgrant, line 672: an out-of-scope comment: I think using user-data for these various setups is actually complicating our code. I think it may be more straightforward to start the lxd container, and then run a python function that runs commands and writes files for the given set up.

This would also allow for an easier way to do tests that require setup _before_ upgrading ua to the version under test. (we normally just do these as one-off sru scripts, but ideally they'd be run as integration tests to avoid regressions)

I may explore this at some point

renanrodrigo: That's a great Idea - @blackboxsw is there a strong reason we set the PPAs through cloud-init instead of configuring it after instance creation, right before installing u-a-t?

blackboxsw: @renanrodrigo sorry to have missed this question. No strong reason, cloud-init happens to order apt configuration as being right right before any ubuntu_advantage #cloud-config directives, So we knew apt setup would be completed by cloud-init prior to reacting to any image-creation or test-specific ubuntu_advantage #cloud-config directives.  I don't see a problem with teasing out that aspect of user-data setup as long as you know the PPA is already setup in image creation before installing any test-related packages.

features/environment.py
blackboxsw: +1 your explanation  makes sense @renanrodrigo and you are correct on the original intent `to test any cloud-init specific integration with UA client`. This is the reason we added it in the first place, to give UA-devs or other teams a chance to easily instrument one-off manual tests of UA with cloud-init integration to triage specific bug reports or debug atypical failure paths. That said. If we dropped this 'feature', it is not a hard a lift to manipulate and existing simple behave test case to add a custom user-data step if you go that route. I agree nobody that I'm aware of is using the "feature" to inject userdata_file in testing (mostly because I don't know of any consumers outside of our team driving these behave tests).

dev-docs/howtoguides/testing.md
renanrodrigo: done

features/environment.py
renanrodrigo: done

dev-docs/howtoguides/testing.md
renanrodrigo: done

lucasmoura (approved):

lucasmoura:
The CI error is unrelated to this change

PR #2123, opened 2022-06-13
status: create new file object

CalvoM:
status: create new file object

In the process of creating a non-root status response, we are first creating a new file object to handle machine token file.

## Desired commit type

## Checklist

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -117,7 +117,7 @@ check_service_is_beta() {
 from uaclient.config import UAConfig
 from uaclient.entitlements import entitlement_factory
 try:
-    cfg = UAConfig()
+    cfg = UAConfig(root_mode=True)
     ent_cls = entitlement_factory(cfg=cfg, name='${service_name}')
     allow_beta = cfg.features.get('allow_beta', False)
     print(all([ent_cls.is_beta, not allow_beta]))
@@ -244,7 +244,7 @@ add_notice() {
     /usr/bin/python3 -c "
 from uaclient.config import UAConfig
 from uaclient.messages import ${msg_name}
-cfg = UAConfig()
+cfg = UAConfig(root_mode=True)
 cfg.add_notice(label='', description=${msg_name})
 "
 }
--- a/lib/daemon.py
+++ b/lib/daemon.py
@@ -30,7 +30,7 @@ def setup_logging(console_level, log_level, log_file, logger):
 
 def main() -> int:
 
-    cfg = UAConfig()
+    cfg = UAConfig(root_mode=True)
     setup_logging(
         logging.INFO, logging.DEBUG, log_file=cfg.daemon_log_file, logger=LOG
     )
--- a/lib/reboot_cmds.py
+++ b/lib/reboot_cmds.py
@@ -151,6 +151,6 @@ def main(cfg):
 
 
 if __name__ == "__main__":
-    cfg = config.UAConfig()
+    cfg = config.UAConfig(root_mode=True)
     setup_logging(logging.INFO, logging.DEBUG, log_file=cfg.log_file)
     main(cfg=cfg)
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -89,7 +89,8 @@ class MeteringTimedJob(TimedJob):
         job, we will select the greater value between those two choices.
         """
         return max(
-            cfg.activity_ping_interval or 0, super().run_interval_seconds(cfg)
+            cfg.machine_token_file.activity_ping_interval or 0,
+            super().run_interval_seconds(cfg),
         )
 
 
@@ -127,7 +128,7 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
 
 
 if __name__ == "__main__":
-    cfg = UAConfig()
+    cfg = UAConfig(root_mode=True)
     current_time = datetime.now(timezone.utc)
 
     # The ua-timer logger should log everything to its file
--- a/lib/upgrade_lts_contract.py
+++ b/lib/upgrade_lts_contract.py
@@ -53,7 +53,7 @@ current_codename_to_past_codename = {
 
 def process_contract_delta_after_apt_lock() -> None:
     logging.debug("Check whether to upgrade-lts-contract")
-    cfg = UAConfig()
+    cfg = UAConfig(root_mode=True)
     if not cfg.is_attached:
         logging.debug("Skipping upgrade-lts-contract. Machine is unattached")
         return
@@ -82,8 +82,13 @@ def process_contract_delta_after_apt_lock() -> None:
         logging.warning(msg)
         sys.exit(1)
 
-    past_entitlements = UAConfig(series=past_release).entitlements
-    new_entitlements = UAConfig(series=current_release).entitlements
+    past_entitlements = UAConfig(
+        series=past_release, root_mode=True
+    ).machine_token_file.entitlements
+    new_entitlements = UAConfig(
+        series=current_release,
+        root_mode=True,
+    ).machine_token_file.entitlements
 
     retry_count = 0
     while out:
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -4,13 +4,12 @@ import logging
 import os
 import re
 from collections import namedtuple
-from datetime import datetime
 from functools import lru_cache, wraps
 from typing import Any, Callable, Dict, Optional, Tuple, TypeVar
 
 import yaml
 
-from uaclient import apt, event_logger, exceptions, messages, snap, util
+from uaclient import apt, event_logger, exceptions, files, messages, snap, util
 from uaclient.defaults import (
     BASE_CONTRACT_URL,
     BASE_SECURITY_URL,
@@ -78,7 +77,6 @@ class UAConfig:
     data_paths = {
         "instance-id": DataPath("instance-id", True, False),
         "machine-access-cis": DataPath("machine-access-cis.json", True, False),
-        "machine-token": DataPath("machine-token.json", True, False),
         "lock": DataPath("lock", True, False),
         "status-cache": DataPath("status.json", False, False),
         "notices": DataPath("notices.json", False, False),
@@ -91,9 +89,6 @@ class UAConfig:
         "jobs-status": DataPath("jobs-status.json", False, True),
     }  # type: Dict[str, DataPath]
 
-    _entitlements = None  # caching to avoid repetitive file reads
-    _machine_token = None  # caching to avoid repetitive file reading
-    _contract_expiry_datetime = None
     ua_scoped_proxy_options = ("ua_apt_http_proxy", "ua_apt_https_proxy")
     global_scoped_proxy_options = (
         "global_apt_http_proxy",
@@ -104,7 +99,12 @@ class UAConfig:
         "apt_https_proxy",
     )
 
-    def __init__(self, cfg: Dict[str, Any] = None, series: str = None) -> None:
+    def __init__(
+        self,
+        cfg: Dict[str, Any] = None,
+        series: str = None,
+        root_mode: bool = False,
+    ) -> None:
         """"""
         if cfg:
             self.cfg_path = None
@@ -115,14 +115,18 @@ class UAConfig:
             self.cfg, self.invalid_keys = parse_config(self.cfg_path)
 
         self.series = series
+        self.root_mode = root_mode
+        self._machine_token_file = None
 
     @property
-    def accounts(self):
-        """Return the list of accounts that apply to this authorized user."""
-        if self.is_attached:
-            accountInfo = self.machine_token["machineTokenInfo"]["accountInfo"]
-            return [accountInfo]
-        return []
+    def machine_token_file(self):
+        if not self._machine_token_file:
+            self._machine_token_file = files.MachineTokenFile(
+                self.data_dir,
+                self.root_mode,
+                self.features.get("machine_token_overlay"),
+            )
+        return self._machine_token_file
 
     @property
     def contract_url(self) -> str:
@@ -381,76 +385,9 @@ class UAConfig:
         )
 
     @property
-    def entitlements(self):
-        """Return configured entitlements keyed by entitlement named"""
-        if self._entitlements:
-            return self._entitlements
-        if not self.machine_token:
-            return {}
-        self._entitlements = self.get_entitlements_from_token(
-            self.machine_token
-        )
-        return self._entitlements
-
-    @staticmethod
-    def get_entitlements_from_token(machine_token: Dict):
-        """Return a dictionary of entitlements keyed by entitlement name.
-
-        Return an empty dict if no entitlements are present.
-        """
-        if not machine_token:
-            return {}
-
-        entitlements = {}
-        contractInfo = machine_token.get("machineTokenInfo", {}).get(
-            "contractInfo"
-        )
-        if not contractInfo:
-            return {}
-
-        tokens_by_name = dict(
-            (e.get("type"), e.get("token"))
-            for e in machine_token.get("resourceTokens", [])
-        )
-        ent_by_name = dict(
-            (e.get("type"), e)
-            for e in contractInfo.get("resourceEntitlements", [])
-        )
-        for entitlement_name, ent_value in ent_by_name.items():
-            entitlement_cfg = {"entitlement": ent_value}
-            if entitlement_name in tokens_by_name:
-                entitlement_cfg["resourceToken"] = tokens_by_name[
-                    entitlement_name
-                ]
-            util.apply_contract_overrides(entitlement_cfg)
-            entitlements[entitlement_name] = entitlement_cfg
-        return entitlements
-
-    @property
-    def contract_expiry_datetime(self) -> datetime:
-        """Return a datetime of the attached contract expiration."""
-        if not self._contract_expiry_datetime:
-            self._contract_expiry_datetime = self.machine_token[
-                "machineTokenInfo"
-            ]["contractInfo"]["effectiveTo"]
-
-        return self._contract_expiry_datetime
-
-    @property
     def is_attached(self):
         """Report whether this machine configuration is attached to UA."""
         return bool(self.machine_token)  # machine_token is removed on detach
-
-    @property
-    def contract_remaining_days(self) -> int:
-        """Report num days until contract expiration based on effectiveTo
-
-        :return: A positive int representing the number of days the attached
-            contract remains in effect. Return a negative int for the number
-            of days beyond contract's effectiveTo date.
-        """
-        delta = self.contract_expiry_datetime.date() - datetime.utcnow().date()
-        return delta.days
 
     @property
     def features(self):
@@ -470,80 +407,7 @@ class UAConfig:
     @property
     def machine_token(self):
         """Return the machine-token if cached in the machine token response."""
-        if not self._machine_token:
-            raw_machine_token = self.read_cache("machine-token")
-
-            machine_token_overlay_path = self.features.get(
-                "machine_token_overlay"
-            )
-
-            if raw_machine_token and machine_token_overlay_path:
-                machine_token_overlay = self.parse_machine_token_overlay(
-                    machine_token_overlay_path
-                )
-
-                if machine_token_overlay:
-                    depth_first_merge_overlay_dict(
-                        base_dict=raw_machine_token,
-                        overlay_dict=machine_token_overlay,
-                    )
-
-            self._machine_token = raw_machine_token
-
-        return self._machine_token
-
-    @property
-    def activity_token(self) -> "Optional[str]":
-        if self.machine_token:
-            return self.machine_token.get("activityInfo", {}).get(
-                "activityToken"
-            )
-        return None
-
-    @property
-    def activity_id(self) -> "Optional[str]":
-        if self.machine_token:
-            return self.machine_token.get("activityInfo", {}).get("activityID")
-        return None
-
-    @property
-    def activity_ping_interval(self) -> "Optional[int]":
-        if self.machine_token:
-            return self.machine_token.get("activityInfo", {}).get(
-                "activityPingInterval"
-            )
-        return None
-
-    @property
-    def contract_id(self):
-        if self.machine_token:
-            return (
-                self.machine_token.get("machineTokenInfo", {})
-                .get("contractInfo", {})
-                .get("id")
-            )
-        return None
-
-    def parse_machine_token_overlay(self, machine_token_overlay_path):
-        if not os.path.exists(machine_token_overlay_path):
-            raise exceptions.UserFacingError(
-                messages.INVALID_PATH_FOR_MACHINE_TOKEN_OVERLAY.format(
-                    file_path=machine_token_overlay_path
-                )
-            )
-
-        try:
-            machine_token_overlay_content = util.load_file(
-                machine_token_overlay_path
-            )
-
-            return json.loads(machine_token_overlay_content)
-        except ValueError as e:
-            raise exceptions.UserFacingError(
-                messages.ERROR_JSON_DECODING_IN_FILE.format(
-                    error=str(e), file_path=machine_token_overlay_path
-                )
-            )
+        return self.machine_token_file.machine_token
 
     def data_path(self, key: Optional[str] = None) -> str:
         """Return the file path in the data directory represented by the key"""
@@ -577,9 +441,8 @@ class UAConfig:
             raise RuntimeError(
                 "Invalid or empty key provided to delete_cache_key"
             )
-        if key.startswith("machine-access") or key == "machine-token":
-            self._entitlements = None
-            self._machine_token = None
+        if key.startswith("machine-access"):
+            self._machine_token_file = None
         elif key == "lock":
             self.remove_notice("", "Operation in progress.*")
         cache_path = self.data_path(key)
@@ -615,9 +478,8 @@ class UAConfig:
             os.makedirs(data_dir)
             if os.path.basename(data_dir) == PRIVATE_SUBDIR:
                 os.chmod(data_dir, 0o700)
-        if key.startswith("machine-access") or key == "machine-token":
-            self._machine_token = None
-            self._entitlements = None
+        if key.startswith("machine-access"):
+            self._machine_token_file = None
         elif key == "lock":
             if ":" in content:
                 self.add_notice(
@@ -885,50 +747,3 @@ def apply_config_settings_override(override_key: str):
         return new_f
 
     return wrapper
-
-
-def depth_first_merge_overlay_dict(base_dict, overlay_dict):
-    """Merge the contents of overlay dict into base_dict not only on top-level
-    keys, but on all on the depths of the overlay_dict object. For example,
-    using these values as entries for the function:
-
-    base_dict = {"a": 1, "b": {"c": 2, "d": 3}}
-    overlay_dict = {"b": {"c": 10}}
-
-    Should update base_dict into:
-
-    {"a": 1, "b": {"c": 10, "d": 3}}
-
-    @param base_dict: The dict to be updated
-    @param overlay_dict: The dict with information to be added into base_dict
-    """
-
-    def update_dict_list(base_values, overlay_values, key):
-        values_to_append = []
-        id_key = MERGE_ID_KEY_MAP.get(key)
-        for overlay_value in overlay_values:
-            was_replaced = False
-            for base_value_idx, base_value in enumerate(base_values):
-                if base_value.get(id_key) == overlay_value.get(id_key):
-                    depth_first_merge_overlay_dict(base_value, overlay_value)
-                    was_replaced = True
-
-            if not was_replaced:
-                values_to_append.append(overlay_value)
-
-        base_values.extend(values_to_append)
-
-    for key, value in overlay_dict.items():
-        base_value = base_dict.get(key)
-        if isinstance(base_value, dict) and isinstance(value, dict):
-            depth_first_merge_overlay_dict(base_dict[key], value)
-        elif isinstance(base_value, list) and isinstance(value, list):
-            if len(base_value) and isinstance(base_value[0], dict):
-                update_dict_list(base_dict[key], value, key=key)
-            else:
-                """
-                Most other lists which aren't lists of dicts are lists of
-                strs. Replace that list # with the overlay value."""
-                base_dict[key] = value
-        else:
-            base_dict[key] = value
--- a/uaclient/conftest.py
+++ b/uaclient/conftest.py
@@ -98,9 +98,15 @@ def logging_sandbox():
 @pytest.fixture
 def FakeConfig(tmpdir):
     class _FakeConfig(UAConfig):
-        def __init__(self, cfg_overrides={}, features_override=None) -> None:
-            cfg_overrides.update({"data_dir": tmpdir.strpath})
-            super().__init__(cfg_overrides)
+        def __init__(
+            self,
+            cfg_overrides={},
+            features_override=None,
+            root_mode: bool = True,
+        ) -> None:
+            if not cfg_overrides.get("data_dir"):
+                cfg_overrides.update({"data_dir": tmpdir.strpath})
+            super().__init__(cfg_overrides, root_mode=root_mode)
 
         @classmethod
         def for_attached_machine(
@@ -108,6 +114,7 @@ def FakeConfig(tmpdir):
             account_name: str = "test_account",
             machine_token: Dict[str, Any] = None,
             status_cache: Dict[str, Any] = None,
+            root_mode: bool = True,
         ):
             if not machine_token:
                 machine_token = {
@@ -138,8 +145,8 @@ def FakeConfig(tmpdir):
             if not status_cache:
                 status_cache = {"attached": True}
 
-            config = cls()
-            config.write_cache("machine-token", machine_token)
+            config = cls(root_mode=root_mode)
+            config.machine_token_file.write(machine_token)
             config.write_cache("status-cache", status_cache)
             return config
 
--- a/uaclient/contract_data_types.py
+++ b/uaclient/contract_data_types.py
@@ -1,0 +1,298 @@
+from typing import List, Optional
+
+from uaclient.data_types import (
+    BoolDataValue,
+    DataObject,
+    Field,
+    IntDataValue,
+    StringDataValue,
+    data_list,
+)
+
+
+class ActivityInfo(DataObject):
+    fields = [
+        Field("activityID", StringDataValue, False),
+        Field("activityToken", StringDataValue, False),
+        Field("activityPingInterval", IntDataValue, False),
+        Field("resources", data_list(StringDataValue), False),
+    ]
+
+    def __init__(
+        self,
+        activityID: Optional[str],
+        activityToken: Optional[str],
+        activityPingInterval: Optional[int],
+        resources: Optional[List[str]],
+    ):
+        self.activityID = activityID
+        self.activityPingInterval = activityPingInterval
+        self.activityToken = activityToken
+        self.resources = resources
+
+
+class AvailableResource(DataObject):
+    fields = [
+        Field("available", BoolDataValue, False),
+        Field("name", StringDataValue, False),
+        Field("description", StringDataValue, False),
+        Field("presentedAs", StringDataValue, False),
+    ]
+
+    def __init__(
+        self,
+        available: Optional[bool],
+        name: Optional[str],
+        description: Optional[str],
+        presentedAs: Optional[str],
+    ):
+        self.available = available
+        self.name = name
+        self.description = description
+        self.presentedAs = presentedAs
+
+
+class AccountInfo(DataObject):
+    fields = [
+        Field("name", StringDataValue, False),
+        Field("id", StringDataValue, False),
+        Field("createdAt", StringDataValue, False),
+        Field("type", StringDataValue, False),
+        Field("userRoleOnAccount", StringDataValue, False),
+    ]
+
+    def __init__(
+        self,
+        name: Optional[str],
+        id: Optional[str],
+        createdAt: Optional[str],
+        type: Optional[str],
+        userRoleOnAccount: Optional[str],
+    ):
+        self.name = name
+        self.id = id
+        self.createAt = createdAt
+        self.type = type
+        self.userRoleOnAccount = userRoleOnAccount
+
+
+class Affordances(DataObject):
+    fields = [
+        Field("architectures", data_list(StringDataValue), False),
+        Field("presentedAs", StringDataValue, False),
+        Field("series", data_list(StringDataValue), False),
+        Field("kernelFlavors", data_list(StringDataValue), False),
+        Field("minKernelVersion", StringDataValue, False),
+        Field("tier", StringDataValue, False),
+        Field("supportLevel", StringDataValue, False),
+    ]
+
+    def __init__(
+        self,
+        architectures: Optional[List[str]],
+        presentedAs: Optional[str],
+        series: List[Optional[str]],
+        kernelFlavors: List[Optional[str]],
+        minKernelVersion: Optional[str],
+        tier: Optional[str],
+        supportLevel: Optional[str],
+    ):
+        self.architectures = architectures
+        self.presentedAs = presentedAs
+        self.series = series
+        self.kernelFlavors = kernelFlavors
+        self.minKernelVersion = minKernelVersion
+        self.tier = tier
+        self.supportLevel = supportLevel
+
+
+class Obligations(DataObject):
+    fields = [
+        Field("enableByDefault", BoolDataValue, False),
+        Field("additionalProperties", StringDataValue, False),
+    ]
+
+    def __init__(
+        self,
+        enableByDefault: Optional[bool],
+        additionalProperties: Optional[str],
+    ):
+        self.enableByDefault = enableByDefault
+        self.additionalProperties = additionalProperties
+
+
+class Directives(DataObject):
+    fields = [
+        Field("additionalPackages", data_list(StringDataValue), False),
+        Field("aptURL", StringDataValue, False),
+        Field("suites", data_list(StringDataValue), False),
+        Field("server", StringDataValue, False),
+        Field("remoteServer", StringDataValue, False),
+        Field("caCerts", StringDataValue, False),
+        Field("snapChannel", StringDataValue, False),
+        Field("pypiURL", StringDataValue, False),
+        Field("url", StringDataValue, False),
+    ]
+
+    def __init__(
+        self,
+        additionalPackages: Optional[List[str]],
+        aptURL: Optional[str],
+        suites: Optional[List[str]],
+        server: Optional[str],
+        remoteServer: Optional[str],
+        caCerts: Optional[str],
+        snapChannel: Optional[str],
+        pypiURL: Optional[str],
+        url: Optional[str],
+    ):
+        self.additionalPackages = additionalPackages
+        self.aptURL = aptURL
+        self.suites = suites
+        self.server = server
+        self.remoteServer = remoteServer
+        self.caCerts = caCerts
+        self.snapChannel = snapChannel
+        self.pypiURL = pypiURL
+        self.url = url
+
+
+class OverrideSelector(DataObject):
+    fields = [
+        Field("series", StringDataValue, False),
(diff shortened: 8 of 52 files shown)


uaclient/defaults.py
```diff
@@ -10,6 +10,8 @@
 DEFAULT_MACHINE_TOKEN_PATH = DEFAULT_DATA_DIR + "/private/machine-token.json"
 DEFAULT_CONFIG_FILE = UAC_ETC_PATH + "uaclient.conf"
 DEFAULT_HELP_FILE = UAC_ETC_PATH + "help_data.yaml"
+PRIVATE_MACHINE_TOKEN_FILE = "machine-token.json"
+PUBLIC_MACHINE_TOKEN_FILE = "pub-machine-token.json"
```
orndorffgrant, line 14: Bikeshedding opportunity: Do they need separate names? Or is it sufficient that one is in the `private/` directory?

uaclient/conftest.py
```diff
@@ -99,7 +99,8 @@ def logging_sandbox():
 def FakeConfig(tmpdir):
     class _FakeConfig(UAConfig):
         def __init__(self, cfg_overrides={}, features_override=None) -> None:
-            cfg_overrides.update({"data_dir": tmpdir.strpath})
+            if not cfg_overrides or not cfg_overrides.get("data_dir"):
```
lucasmoura, line 102: I think we can only check for the `cfg_overrides.get("data_dir")` here. This just simplifies for the case where we have an empty dict or a dict without `data_dir` defined

uaclient/entitlements/tests/test_cc.py
```diff
@@ -84,12 +92,12 @@ class TestCommonCriteriaEntitlementCanEnable:
     @mock.patch("uaclient.util.subp", return_value=("", ""))
     @mock.patch("uaclient.util.get_platform_info")
     def test_can_enable_true_on_entitlement_inactive(
-        self, m_platform_info, _m_subp, capsys, tmpdir
+        self, m_platform_info, _m_subp, capsys, tmpdir, FakeConfig
```
lucasmoura, line 95: I think we can also remove `tmpdir` here

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
+import json
+import logging
+import os
+from datetime import datetime
+from typing import Dict, Optional
+
+from uaclient import defaults, event_logger, schemas, util
+
+PRIVATE_SUBDIR = "private"
```
orndorffgrant, line 9: Maybe this value should go in defaults.py?

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+
+event = event_logger.get_event_logger()
+LOG = logging.getLogger(__name__)
+
+
+class UAFile:
+    _root_directory = defaults.DEFAULT_DATA_DIR
+    _file_name = "file"
```
orndorffgrant, line 17: I can't think of a case where this file name would be accurate :) So we probably shouldn't set a default for this

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+    @property
+    def filepath(self):
+        if not self._filepath:
+            self._filepath = os.path.join(
+                self._root_directory, self._file_name
+            )
+        return self._filepath
+
+    @property
+    def is_private_file(self):
```
orndorffgrant, line 41:
```suggestion
    def is_private(self):
```

uaclient/entitlements/tests/test_cis.py
```diff
@@ -12,12 +12,13 @@
 
 @pytest.fixture
 def entitlement(entitlement_factory):
-    return entitlement_factory(
-        CISEntitlement,
-        allow_beta=True,
-        called_name="cis",
-        additional_packages=["pkg1"],
-    )
+    with mock.patch("os.getuid", return_value=0):
```
lucasmoura, line 15: Okay, I think we can make this test a little easier to understand if the `UAConfig` class can accept an `is_root` variable. In that way, we can add a new variable to the `entitlement_factory` and pass it to the `FakeConfig` instance. This will avoid the mocking situation we are seeing on the factory functions

uaclient/files.py
lucasmoura, line 17: We can define this as a property similar to what we do with `name` for the entitlement classes. This will just guarantee that the subclasses will always define a name

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        return self._is_private
+
+    def write(self, content):
+        file_mode = 0o600 if self._is_private else 0o644
+        self.create_file()
+        if not isinstance(content, str):
+            content = json.dumps(content, cls=util.DatetimeAwareJSONEncoder)
+        util.write_file(self.filepath, content, file_mode)
```
orndorffgrant, line 49: I'd rather keep this layer simple and only support string content
```suggestion
    def write(self, content: str):
        file_mode = 0o600 if self._is_private else 0o644
        self.create_file()
        util.write_file(self.filepath, content, file_mode)
```

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class UAFile:
+    _root_directory = defaults.DEFAULT_DATA_DIR
+    _file_name = "file"
+
+    def __init__(
+        self,
+        root_directory: Optional[str] = None,
```
lucasmoura, line 21: Can't the default value here be `defaults.DEFAULT_DATA_DIR` ? Because then we would need need the `if` check later

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        return self._is_private
+
+    def write(self, content):
+        file_mode = 0o600 if self._is_private else 0o644
+        self.create_file()
+        if not isinstance(content, str):
+            content = json.dumps(content, cls=util.DatetimeAwareJSONEncoder)
+        util.write_file(self.filepath, content, file_mode)
+
+    def read(self):
```
orndorffgrant, line 51: I haven't been paying close attention to this, but generally we should give all functions/methods type annotations
```suggestion
    def read(self) -> Optional[str]:
```

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        file_dir = os.path.dirname(self.filepath)
+        if not os.path.exists(file_dir):
+            os.makedirs(file_dir)
+            if self._is_private:
+                os.chmod(file_dir, 0o700)
+
+
+class PublicMachineTokenFile(UAFile):
+    _file_name = defaults.PUBLIC_MACHINE_TOKEN_FILE
+    _is_private = False
```
orndorffgrant, line 73: I don't think these need to be class properties. they can just be instance properties that are set in the constructor

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+                self._root_directory, self._file_name
+            )
+        return self._filepath
+
+    @property
+    def is_private_file(self):
+        return self._is_private
+
+    def write(self, content):
+        file_mode = 0o600 if self._is_private else 0o644
```
lucasmoura, line 45: Let's use the `is_private` property here.
Also, let's add the file modes into constants

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        try:
+            if isinstance(content, str):
+                content = json.loads(content)
+        except Exception:
+            return None
+        self.filter_tokens(schemas.contract_schema, content)
+        content = json.dumps(content, cls=util.DatetimeAwareJSONEncoder)
+        content = content.replace('"', "'")
+        content = util.redact_sensitive_logs(content)
+        content = content.replace("'", '"')
```
orndorffgrant, line 99: This looks delicate :upside_down_face: 
I don't think we should re-use `redact_sensitive_logs` here. Anything sensitive should be handled by the field filtering.

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        self._fn = None
+        if root_mode:  # private file
+            if not root_directory:
+                root_directory = defaults.DEFAULT_DATA_DIR + "/private"
+            file_name = defaults.PRIVATE_MACHINE_TOKEN_FILE
+            self._rd = root_directory
+            self._fn = file_name
+            self.file_handler = UAFile(root_directory, self._fn)
+        else:  # public file
+            self.file_handler = PublicMachineTokenFile(root_directory)
```
orndorffgrant, line 146: If we move the `filter_tokens` logic into this class, then the public machine token file can just be a normal `UAFile` instance

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def write(self, content):
```
orndorffgrant, line 148: If we're in non-root mode, then we probably can't support `write` at all, because the files would get out of sync

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        """Update the machine_token file for both pub/priv files"""
+        self.file_handler.write(content)
+        if self.file_handler.is_private_file:  # write to public file as well
+            pub_file = PublicMachineTokenFile(self._rd)
+            pub_file.write(content)
+        self._machine_token = None
+        self._entitlements = None
+
+    def delete(self):
```
orndorffgrant, line 157: Same here, we can't do this in non-root mode

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        self._machine_token = None
+        self._entitlements = None
+        self.file_handler = None
+
+    @property
+    def machine_token(self):
+        """Return the machine-token if cached in the machine token response."""
+        if not self._machine_token:
+            if not self.file_handler:
+                return None
```
orndorffgrant, line 174: Is this possible?

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+            content = self.file_handler.read()
+            try:
+                content = content.replace("'", '"')
```
orndorffgrant, line 177: If we're writing single quotes to this file somewhere, then that is a separate bug that needs to be fixed. We should not try to support single quoted json because that's not json

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+    _contract_expiry_datetime = None
+
+    def __init__(
+        self, root_directory: Optional[str] = None, root_mode: bool = True
+    ):
+        self._rd = None
+        self._fn = None
+        if root_mode:  # private file
+            if not root_directory:
+                root_directory = defaults.DEFAULT_DATA_DIR + "/private"
```
lucasmoura, line 140: I think the `file_path` property should handle this logic

uaclient/schemas.py
```diff
@@ -0,0 +1,301 @@
+contract_schema = {
```
orndorffgrant, line 1: Let's align the name of this with its purpose
```suggestion
machine_token_file_schema = {
```

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        except Exception:
+            return None
+        self.filter_tokens(schemas.contract_schema, content)
+        content = json.dumps(content, cls=util.DatetimeAwareJSONEncoder)
+        content = content.replace('"', "'")
+        content = util.redact_sensitive_logs(content)
+        content = content.replace("'", '"')
+        super().write(content)
+
+    def filter_tokens(self, schema, data):
```
orndorffgrant, line 102: There might be a way to use DataObjects for this instead of the schema/example dict.

I think you could define the schema with DataObjects, and then just do a `PublicMachineTokenData.from_dict(d).to_dict()` and that would filter out the disallowed fields.

orndorffgrant (commented):
Looking good!

Didn't go in detail through all the tests yet, and I know it's still a draft, but left several small comments :)

uaclient/conftest.py
CalvoM: Makes sense.

uaclient/entitlements/tests/test_cc.py
```diff
@@ -64,16 +64,24 @@ class TestCommonCriteriaEntitlementUserFacingStatus:
...
     def test_inapplicable_on_invalid_affordances(
-        self, m_platform_info, m_getuid, arch, series, version, details, tmpdir
+        self,
+        m_platform_info,
+        m_getuid,
+        arch,
+        series,
+        version,
+        details,
+        tmpdir,
```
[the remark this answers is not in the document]
CalvoM: Yeah, will remove this.

uaclient/files.py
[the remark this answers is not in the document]
CalvoM: Just noticed that I never used them in the class, so I am just deleting them since they are already passed to the parent class constructor.

uaclient/files.py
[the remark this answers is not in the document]
CalvoM: Tried it out and I see it works.

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+            self.file_handler = PublicMachineTokenFile(root_directory)
+
+    def write(self, content):
+        """Update the machine_token file for both pub/priv files"""
+        self.file_handler.write(content)
+        if self.file_handler.is_private_file:  # write to public file as well
+            pub_file = PublicMachineTokenFile(self._rd)
+            pub_file.write(content)
+        self._machine_token = None
+        self._entitlements = None
```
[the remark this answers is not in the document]
CalvoM: Yes we can. This was part of the things I did in the previous PR.

uaclient/defaults.py
CalvoM: The final name was still to be determined, your suggestion is a good one.

uaclient/files.py
CalvoM: Makes sense.

uaclient/files.py
```diff
@@ -0,0 +1,296 @@
...
+        It checks for new keys and filter the data
+        """
+        try:
+            if isinstance(content, str):
+                content = json.loads(content)
+        except Exception:
+            return None
+        self.filter_tokens(schemas.contract_schema, content)
+        content = json.dumps(content, cls=util.DatetimeAwareJSONEncoder)
+        content = content.replace('"', "'")
```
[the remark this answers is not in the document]
CalvoM: I replaced because the `redact_sensitive_logs` uses single apostrophe.

uaclient/files.py
CalvoM: Considering that only two things are redacted in the machine-token, then updated it.

uaclient/files.py
CalvoM: Yes, No writing if not root.

uaclient/files.py
CalvoM: Hahah, that is true. Tested it and no issue found.

uaclient/entitlements/tests/test_cis.py
CalvoM: Yeah, works well and helped solve some problems in a simple way

uaclient/files.py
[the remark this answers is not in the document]
CalvoM: Updated and the replace is not needed.

uaclient/entitlements/tests/test_cc.py
```diff
@@ -64,16 +64,23 @@ class TestCommonCriteriaEntitlementUserFacingStatus:
...
     ):
         """Test invalid affordances result in inapplicable status."""
         unsupported_info = copy.deepcopy(dict(PLATFORM_INFO_SUPPORTED))
         unsupported_info["arch"] = arch
         unsupported_info["series"] = series
         unsupported_info["version"] = version
         m_platform_info.return_value = unsupported_info
-        cfg = config.UAConfig(cfg={"data_dir": tmpdir.strpath})
-        cfg.write_cache("machine-token", CC_MACHINE_TOKEN)
+        cfg = FakeConfig()
```
lucasmoura, line 82: I think we can use `FakeConfig.for_attached_machine(machine_token=CC_MACHINE_TOKEN)` instead

uaclient/entitlements/tests/test_cc.py
```diff
@@ -84,12 +91,12 @@ class TestCommonCriteriaEntitlementCanEnable:
...
     @mock.patch("uaclient.util.get_platform_info")
     def test_can_enable_true_on_entitlement_inactive(
-        self, m_platform_info, _m_subp, capsys, tmpdir
+        self, m_platform_info, _m_subp, capsys, FakeConfig
     ):
         """When entitlement is INACTIVE, can_enable returns True."""
         m_platform_info.return_value = PLATFORM_INFO_SUPPORTED
-        cfg = config.UAConfig(cfg={"data_dir": tmpdir.strpath})
-        cfg.write_cache("machine-token", CC_MACHINE_TOKEN)
+        cfg = FakeConfig()
```
lucasmoura, line 98: Same idea here

uaclient/entitlements/tests/test_cc.py
```diff
@@ -148,8 +158,8 @@ def exists(path):
             return original_exists(path)
 
         m_platform_info.side_effect = fake_platform
-        cfg = config.UAConfig(cfg={"data_dir": tmpdir.strpath})
-        cfg.write_cache("machine-token", CC_MACHINE_TOKEN)
+        cfg = FakeConfig()
+        cfg.machine_token_file.write(CC_MACHINE_TOKEN)
```
lucasmoura, line 162: And here

uaclient/files.py
```diff
@@ -0,0 +1,387 @@
...
+            content = util.load_file(self.path)
+        except FileNotFoundError:
+            if not os.path.exists(self.path):
+                LOG.debug("File does not exist: {}".format(self.path))
+        return content
+
+    def delete(self):
+        util.remove_file(self.path)
+
+    def create_file(self):
```
lucasmoura, line 69: Since we are creating a directory here, I think it is better to name this method as `create_dir`

uaclient/config.py
```diff
@@ -104,7 +102,12 @@ class UAConfig:
         "apt_https_proxy",
     )
 
-    def __init__(self, cfg: Dict[str, Any] = None, series: str = None) -> None:
+    def __init__(
+        self,
+        cfg: Dict[str, Any] = None,
+        series: str = None,
+        root_mode: bool = True,
```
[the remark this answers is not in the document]
CalvoM: Considering that it was always run as root before, we will have many instances where the root_mode is True and few with root_mode as False. So using the default as False, will make us change so many places.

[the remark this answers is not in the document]
lucasmoura: I think we still need to make that change. The rationale is that it is better for us to be explicit here. For example, Imagine in the future that we have the API structure ready. That means users would be able to run ua code directly on their application. On that situation, we need to have correct defaults, because we can assume that non-root users might be running that code.

Regarding the code changes, I think we would only need to add that logic when we create the config on the `main` function. And I don't see a problem with us updating the tests if needed.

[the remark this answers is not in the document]
CalvoM: Okay. Makes sense, esp. on the future use.

uaclient/config.py
```diff
@@ -577,9 +444,10 @@ def delete_cache_key(self, key: str) -> None:
             raise RuntimeError(
                 "Invalid or empty key provided to delete_cache_key"
             )
-        if key.startswith("machine-access") or key == "machine-token":
-            self._entitlements = None
-            self._machine_token = None
+        if key.startswith("machine-access"):
+            # will find a better way for this
+            self.machine_token_file._entitlements = None
+            self.machine_token_file._machine_token = None
```
orndorffgrant, line 450: This can probably be just
```suggestion
            self._machine_token_file = None
```

uaclient/config.py
```diff
@@ -615,9 +483,9 @@ def write_cache(self, key: str, content: Any) -> None:
             os.makedirs(data_dir)
             if os.path.basename(data_dir) == PRIVATE_SUBDIR:
                 os.chmod(data_dir, 0o700)
-        if key.startswith("machine-access") or key == "machine-token":
-            self._machine_token = None
-            self._entitlements = None
+        if key.startswith("machine-access"):
+            self.machine_token_file._entitlements = None
+            self.machine_token_file._machine_token = None
```
orndorffgrant, line 488: Same here

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+        self._is_private = private
+        self._path = None
+
+    @property
+    def path(self) -> str:
+        if not self._path:
+            self._path = os.path.join(
+                self._root_directory, self._file_name
+            )  # type: ignore
+        return self._path  # type: ignore
```
orndorffgrant, line 40: Does this need to be a property? or can it just be
```suggestion
        self.path =  os.path.join(
            self._root_directory, self._file_name
        )  # type: ignore
```

orndorffgrant: Or if we keep it as a property for encapsulation, we can still pre-compute the value
```suggestion
        self._path = os.path.join(
            self._root_directory, self._file_name
        )  # type: ignore

    @property
    def path(self) -> str:
        return self._path
```

orndorffgrant: Also why are we ignoring types here?

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+    def is_private(self) -> bool:
+        return self._is_private
+
+    def write(self, content):
+        file_mode = (
+            defaults.ROOT_READABLE_MODE
+            if self.is_private
+            else defaults.WORLD_READABLE_MODE
+        )
+        self.create_dir()
```
orndorffgrant, line 52: Is create_dir called anywhere else? does it need to be a separate method?

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+        if not isinstance(content, str):
+            content = json.dumps(content, cls=util.DatetimeAwareJSONEncoder)
```
orndorffgrant, line 54: I don't think `UAFile` should assume the format of the file. We should only handle strings at this layer

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+        util.write_file(self.path, content, file_mode)
+
+    def read(self) -> Optional[str]:
+        content = None
+        try:
+            content = util.load_file(self.path)
+        except FileNotFoundError:
+            if not os.path.exists(self.path):
+                LOG.debug("File does not exist: {}".format(self.path))
```
orndorffgrant, line 63: Do we need the extra check here?
```suggestion
            LOG.debug("File does not exist: {}".format(self.path))
```

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+        return content
+
+    def delete(self):
+        util.remove_file(self.path)
+
+    def create_dir(self):
+        file_dir = os.path.dirname(self.path)
+        if not os.path.exists(file_dir):
```
orndorffgrant, line 71: We already know the directory right?
```suggestion
        if not os.path.exists(self._root_directory):
```

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+            os.makedirs(file_dir)
+            if self.is_private:
+                os.chmod(file_dir, 0o700)
```
orndorffgrant, line 74: Interesting, can we assume the directory we're in is also private just because the file is private? I'm not sure if we can.

I think we need to handle the permissions of the directory separately (a `private_dir` arg to the constructor maybe)

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+
+event = event_logger.get_event_logger()
+LOG = logging.getLogger(__name__)
+
+
+class UAFile:
+    def __init__(
+        self,
+        name: str,
+        root_directory: str = defaults.DEFAULT_DATA_DIR,
```
orndorffgrant, line 26: Bikeshedding: "root_directory" has a specific meaning which is `/`.
Can we call this just `directory`?

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+    def delete(self):
+        """Delete both pub and private files"""
+        if self.is_root:
+            self.public_file.delete()
+            self.private_file.delete()
+
+            self._machine_token = None
+            self._entitlements = None
+            self.public_file = None
+            self.private_file = None
```
orndorffgrant, line 116: We may want to do something like
```
machine_token_file.delete()
machine_token_file.write(new_content)
```

so I don't think we can set these fields to None

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+
+    @property
+    def file_handler(self) -> Optional[UAFile]:
+        if self.is_root:
+            return self.private_file
+        return self.public_file
+
+    def read(self) -> Optional[dict]:
+        content = None
+        if self.file_handler:
```
orndorffgrant, line 128: Is `self.file_handler` used anywhere else? If not, then can we just decide which file to use inside this `read` method?

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+        self.public_file = UAFile(file_name, root_directory)
+        self.machine_token_overlay_path = machine_token_overlay_path
+        self._machine_token = None
+        self._entitlements = None
+        self._contract_expiry_datetime = None
+
+    def write(self, content):
+        """Update the machine_token file for both pub/private files"""
+        if self.is_root:
+            self.private_file.write(content)
```
orndorffgrant, line 98: I suggest we call json.dumps here instead of in UAFile

uaclient/files.py
```diff
@@ -0,0 +1,344 @@
...
+        if self.is_root:
+            return self.private_file
+        return self.public_file
+
+    def read(self) -> Optional[dict]:
+        content = None
+        if self.file_handler:
+            content = self.file_handler.read()
+            if not content:
+                return None
```
orndorffgrant, line 131: I don't think this is needed
```suggestion
```

uaclient/schemas.py
```diff
@@ -0,0 +1,301 @@
+machine_token_file_schema = {
```
orndorffgrant, line 1: Do we still need this file?

uaclient/machine_token_data.py
```diff
@@ -0,0 +1,609 @@
...
+        self.activityInfo = activityInfo
+        self.availableResources = availableResources
+        self.machineTokenInfo = machineTokenInfo
+
+
+if __name__ == "__main__":
+    with open("./data.json") as f:
+        data = json.loads(f.read())
+    res = PublicMachineTokenData.from_dict(data).to_dict()
+    print(res)
```
orndorffgrant, line 609: This is useful to test but lets remember to remove before merging

uaclient/machine_token_data.py
```diff
@@ -0,0 +1,609 @@
+import enum
```
orndorffgrant, line 1: Can we rename this file to `contract_data_types.py`

orndorffgrant (commented):
This is looking good! Just some implementation details to work out now I think

uaclient/config.py
CalvoM: Oh yes, makes sense.

uaclient/files.py
CalvoM: I was going ```Incompatible types in assignment (expression has type "str", variable has type "None")``` but with this refactor it solves it.

uaclient/files.py
CalvoM: I was just making the write method cleaner 😅. So that it is easier to know what is happening in the method.

uaclient/files.py
CalvoM: True.

uaclient/files.py
CalvoM: I have seen `config.py` checks if the directory basename is `private`. Is this check okay here as well?

uaclient/files.py
CalvoM: Yeah, `directory` is a better name.

uaclient/files.py
CalvoM: Makes sense, did not see that. Since we only instantiate them in constructor.

uaclient/files.py
CalvoM: Yes.

uaclient/schemas.py
CalvoM: Not at all.

uaclient/files.py
CalvoM: Makes sense.

uaclient/files.py
```diff
@@ -0,0 +1,324 @@
...
+        if self.machine_token:
+            return (
+                self.machine_token.get("machineTokenInfo", {})
+                .get("contractInfo", {})
+                .get("id")
+            )
+        return None
+
+
+def depth_first_merge_overlay_dict(base_dict, overlay_dict):
```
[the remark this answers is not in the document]
CalvoM: Yeah

uaclient/config.py
```diff
@@ -577,9 +441,9 @@ def delete_cache_key(self, key: str) -> None:
             raise RuntimeError(
                 "Invalid or empty key provided to delete_cache_key"
             )
-        if key.startswith("machine-access") or key == "machine-token":
-            self._entitlements = None
-            self._machine_token = None
+        if key.startswith("machine-access"):
+            # will find a better way for this
```
orndorffgrant, line 445: is this comment still valid?

orndorffgrant: if so, can you expand on what is wrong about the current way?

uaclient/data_types.py
```diff
@@ -114,12 +114,13 @@ def from_value(val: Any) -> List:
...
+    val: List[Union["DataObject", list, str, int, bool]],
+    keep_none: bool = True,
 ) -> list:
     new_val = []  # type: list
     for item in val:
         if isinstance(item, DataObject):
-            new_val.append(item.to_dict())
+            new_val.append(item.to_dict(keep_none))
         elif isinstance(item, list):
             new_val.append(data_list_to_list(item))
```
orndorffgrant, line 125: keep_none propagation is needed here too
```suggestion
            new_val.append(data_list_to_list(item, keep_none))
```

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,609 @@
...
+        months: Optional[int],
+    ):
+        self.value = value
+        self.currency = currency
+        self.months = months
+
+
+class RenewalItem(DataObject):
+    fields = [
+        Field("allowance", AllowanceInfo, False),
```
orndorffgrant, line 169: Interesting, how does the enum/str interact with everything? does it work appropriately with from_dict and to_dict? We should add unit tests with this kind of usage before relying on it

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,609 @@
...
+        self.activityInfo = activityInfo
+        self.availableResources = availableResources
+        self.machineTokenInfo = machineTokenInfo
+
+
+if __name__ == "__main__":
+    with open("./data.json") as f:
+        data = json.loads(f.read())
+    res = PublicMachineTokenData.from_dict(data).to_dict()
+    print(res)
```
orndorffgrant, line 609: just a reminder to remove this :)

uaclient/files.py
orndorffgrant: `create_dir` isn't that complex and I think could be inlined here, but not a strong opinion :upside_down_face:

uaclient/data_types.py
CalvoM: Thanks for that catch.

uaclient/contract_data_types.py
CalvoM: Yeah, almost thought I had removed it.

uaclient/config.py
CalvoM: It is not valid currently. I put this comment before the new file object.

uaclient/contract_data_types.py
CalvoM: @orndorffgrant I have added some code on this and tests, please check it out.

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,602 @@
...
+        self.contractInfo = contractInfo
+        self.expires = expires
+
+
+class PublicMachineTokenData(DataObject):
+    fields = [
+        Field("activityInfo", ActivityInfo, False),
+        Field("machineTokenInfo", MachineTokenInfo, False),
+        Field("availableResources", data_list(AvailableResource), False),
+        Field("resourceTokens", data_list(ResourceToken), False),
```
orndorffgrant, line 589: @renanrodrigo @lucasmoura @CalvoM what if we just drop `resourceTokens` from the public file?

lucasmoura: +1

uaclient/data_types.py
```diff
@@ -48,6 +68,23 @@ def from_value(val: Any) -> Any:
...
+    @classmethod
+    def from_value(cls, val: Any) -> str:
+        expected_values = tuple(i.value for i in cls)  # type: ignore
+        if not isinstance(val, type(expected_values[0])):
+            raise IncorrectEnumValueTypeError(str(val))
+        if val not in expected_values:
+            raise IncorrectEnumValueError(str(val), cls)
+        return val
+
+
```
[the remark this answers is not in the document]
CalvoM: Yes I can

uaclient/data_types.py
```diff
@@ -11,6 +11,12 @@
 INCORRECT_FIELD_TYPE_ERROR_MESSAGE = (
     'Got value with incorrect type for field "{key}": {nested_msg}'
 )
+INCORRECT_ENUM_VALUE_TYPE_ERROR_MESSAGE = (
```
lucasmoura, line 14: I think we can remove this since it is already on #2147

uaclient/files.py
```diff
@@ -0,0 +1,270 @@
...
+
+                if machine_token_overlay:
+                    util.depth_first_merge_overlay_dict(
+                        base_dict=content,
+                        overlay_dict=machine_token_overlay,
+                    )
+            self._machine_token = content
+        return self._machine_token
+
+    @staticmethod
```
lucasmoura, line 137: I don't think this needs to be a `staticmethod` anymore

uaclient/data_types.py
CalvoM: Oh yes, they slipped through after deleting everything else.

debian/ubuntu-advantage-tools.postinst
```diff
@@ -114,10 +114,12 @@ check_is_active_esm() {
 check_service_is_beta() {
     service_name=$1
     _IS_BETA_SVC=$(/usr/bin/python3 -c "
+import os
 from uaclient.config import UAConfig
 from uaclient.entitlements import entitlement_factory
 try:
-    cfg = UAConfig()
+    root_mode=os.getuid==0
+    cfg = UAConfig(root_mode=root_mode)
```
orndorffgrant, line 122: This file will only ever get run as root
```suggestion
    cfg = UAConfig(root_mode=True)
```

debian/ubuntu-advantage-tools.postinst
```diff
@@ -242,9 +244,11 @@ mark_reboot_for_fips_pro() {
 add_notice() {
     msg_name=$1
     /usr/bin/python3 -c "
+import os
 from uaclient.config import UAConfig
 from uaclient.messages import ${msg_name}
-cfg = UAConfig()
+root_mode=os.getuid==0
+cfg = UAConfig(root_mode=root_mode)
```
orndorffgrant, line 251: same here

lib/daemon.py
```diff
@@ -30,7 +31,8 @@ def setup_logging(console_level, log_level, log_file, logger):
 
 def main() -> int:
 
-    cfg = UAConfig()
+    root_mode = os.getuid() == 0
+    cfg = UAConfig(root_mode=root_mode)
```
orndorffgrant, line 35: same here
```suggestion
    cfg = UAConfig(root_mode=True)
```

lib/reboot_cmds.py
```diff
@@ -151,6 +151,7 @@ def main(cfg):
 
 
 if __name__ == "__main__":
-    cfg = config.UAConfig()
+    root_mode = os.getuid() == 0
+    cfg = config.UAConfig(root_mode=root_mode)
```
orndorffgrant, line 155: Same here :)
```suggestion
    cfg = config.UAConfig(root_mode=True)
```

lib/timer.py
```diff
@@ -127,7 +129,8 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
 
 
 if __name__ == "__main__":
-    cfg = UAConfig()
+    root_mode = os.getuid() == 0
+    cfg = UAConfig(root_mode=root_mode)
```
orndorffgrant, line 133: Same here :)
```suggestion
    cfg = UAConfig(root_mode=True)
```

lib/upgrade_lts_contract.py
```diff
@@ -53,7 +54,8 @@
 
 def process_contract_delta_after_apt_lock() -> None:
     logging.debug("Check whether to upgrade-lts-contract")
-    cfg = UAConfig()
+    root_mode = os.getuid() == 0
+    cfg = UAConfig(root_mode=root_mode)
```
orndorffgrant, line 58: Same here :)
```suggestion
    cfg = UAConfig(root_mode=True)
```

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,572 @@
...
+
+
+class Entitlement(DataObject):
+    fields = [
+        Field("entitled", BoolDataValue, False),
+        Field("type", StringDataValue, False),
+        Field("affordances", Affordances, False),
+        Field("obligations", Obligations, False),
+        Field("directives", Directives, False),
+        Field("series", SeriesEntitlements, False),
```
orndorffgrant, line 462: We can drop the `series` field as it is old.

We'll need the `overrides` field though https://github.com/canonical/ua-contracts/blob/develop/docs/contracts.yaml#L1312=

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,572 @@
...
+        Field("value", IntDataValue, False),
+        Field("reason", StringDataValue, False),
+        Field("externalIDs", ExternalID, False),
+        Field("purchaseID", StringDataValue, False),
+        Field("productListingID", StringDataValue, False),
+        Field("disabled", BoolDataValue, False),
+        Field("disabledExplanation", StringDataValue, False),
+        Field("subscriptionID", StringDataValue, False),
+        Field("period", StringDataValue, False),
+        Field("renewal", Renewal, False),
```
orndorffgrant, line 287: I think should can drop `renewal` and all its nested content - lots of it is bordering on sensitive, and we don't use any of it anyway

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,572 @@
...
+    ):
+        self.name = name
+        self.id = id
+        self.createAt = createdAt
+        self.type = type
+        self.userRoleOnAccount = userRoleOnAccount
+        self.externalAccountIDs = externalAccountIDs
+
+
+class ActiveMachine(DataObject):
```
orndorffgrant, line 93: A user may not want this information to be non-root readable either. And we don't use it. Let's not include it

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,572 @@
+from typing import List, Optional
```
orndorffgrant, line 1: @lucasmoura @renanrodrigo please double check my opinions commented in this file

uaclient/files.py
```diff
@@ -0,0 +1,270 @@
...
+        return content
+
+    def delete(self):
+        util.remove_file(self.path)
+
+
+class MachineTokenFile:
+    def __init__(
+        self,
+        directory: str = defaults.DEFAULT_DATA_DIR,
```
orndorffgrant, line 61: Do we need the `directory` param here?

orndorffgrant (commented):
This is really close now! Most of my suggestions here are just around minimizing the content of the public file a bit

uaclient/conftest.py
```diff
@@ -98,16 +98,23 @@ def logging_sandbox():
...
 def FakeConfig(tmpdir):
     class _FakeConfig(UAConfig):
-        def __init__(self, cfg_overrides={}, features_override=None) -> None:
-            cfg_overrides.update({"data_dir": tmpdir.strpath})
-            super().__init__(cfg_overrides)
+        def __init__(
+            self,
+            cfg_overrides={},
+            features_override=None,
+            root_mode: bool = False,
```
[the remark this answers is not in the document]
orndorffgrant: I don't mind either way :upside_down_face:

uaclient/contract_data_types.py
CalvoM: Thanks for catching `overrides`.

uaclient/contract_data_types.py
CalvoM: Okay. Makes sense.

uaclient/contract_data_types.py
CalvoM: Okay. Makes sense.

uaclient/files.py
CalvoM: yeah, used during testing.

orndorffgrant: ah got it cool!

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,401 @@
...
+        value: Optional[int],
+        currency: Optional[str],
+        months: Optional[int],
+    ):
+        self.value = value
+        self.currency = currency
+        self.months = months
+
+
+class ContractItem(DataObject):
```
orndorffgrant, line 122: This class can be removed

orndorffgrant (commented):
A couple more things to remove and then I think we're done! :)

orndorffgrant:
I triggered a re-run of the failed integration test just in case it is a flake

renanrodrigo (approved):
I have a couple opinions about how the machine token files could be handled in a different way (like separating the private and public files, decoupling implementation details from config), but I think future iterations could cover that.
I just pointed a couple things to be thought about, but those are non-blocking for this PR.
Thanks again for all the effort.

uaclient/contract_data_types.py
```diff
@@ -0,0 +1,309 @@
...
+
+class ExternalID(DataObject):
+    fields = [
+        Field("origin", StringDataValue, False),
+        Field("IDs", data_list(StringDataValue), False),
+    ]
+
+    def __init__(self, origin: Optional[str], IDs: Optional[List[str]]):
+        self.origin = origin
+        self.IDs = IDs
```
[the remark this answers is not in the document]
CalvoM: Thanks.

uaclient/files.py
```diff
@@ -0,0 +1,270 @@
...
+        self._file_name = name
+        self._is_private = private
+        self._path = os.path.join(self._directory, self._file_name)
+
+    @property
+    def path(self) -> str:
+        return self._path
+
+    @property
+    def is_private(self) -> bool:
```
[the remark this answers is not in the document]
CalvoM: I made it a property so that we can avoid accidentally changing the privacy of the file.

uaclient/util.py
```diff
@@ -979,3 +979,54 @@ def handle_unicode_characters(message: str) -> str:
...
+
+    {"a": 1, "b": {"c": 10, "d": 3}}
+
+    @param base_dict: The dict to be updated
+    @param overlay_dict: The dict with information to be added into base_dict
+    """
+
+    def update_dict_list(base_values, overlay_values, key):
+        merge_id_key_map = {
+            "availableResources": "name",
```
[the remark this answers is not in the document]
CalvoM: Good point. We can either make it generic if we need it for other parts of the code or we just move it to the machine token file. Do we need it for other parts of the code?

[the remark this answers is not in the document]
renanrodrigo: I see it was just moved here per previous reviews. No need to change it now, it's ok

uaclient/files.py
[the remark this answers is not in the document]
renanrodrigo: fair

lucasmoura (approved):

lib/upgrade_lts_contract.py
```diff
@@ -82,8 +82,12 @@ def process_contract_delta_after_apt_lock() -> None:
         logging.warning(msg)
         sys.exit(1)
 
-    past_entitlements = UAConfig(series=past_release).entitlements
-    new_entitlements = UAConfig(series=current_release).entitlements
+    past_entitlements = UAConfig(
```
lucasmoura, line 85: I think we need to pass `root_mode` here as well

lucasmoura (commented):
Just one tiny nit and after fixing it, we can merge this PR

lucasmoura (approved):

renanrodrigo:
lgtm, just waiting on CI

PR #2124, opened 2022-06-13
postinst: cleanup cloud-id-shim on xenial->bionic

orndorffgrant:
## Test Steps

Run the new manual test script

## Desired commit type

## Checklist

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -37,6 +37,7 @@ OLD_MESSAGING_TIMER="ua-messaging.timer"
 OLD_MESSAGING_TIMER_MASKED_LOCATION="/etc/systemd/system/timers.target.wants/$OLD_MESSAGING_TIMER"
 OLD_LICENSE_CHECK_PATH="ua-license-check.path"
 OLD_LICENSE_CHECK_PATH_MASKED_LOCATION="/etc/systemd/system/multi-user.target.wants/$OLD_LICENSE_CHECK_PATH"
+XENIAL_CLOUD_ID_SHIM_UNIT_LOCATION="/etc/systemd/system/multi-user.target.wants/ubuntu-advantage-cloud-id-shim.service"
 
 ESM_APT_PREF_FILE_TRUSTY="$APT_PREFERENCES_DIR/ubuntu-esm-trusty"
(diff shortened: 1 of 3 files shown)


orndorffgrant:
@renanrodrigo we don't have the nicer ubuntu upgrade integration test in `release-27` that would allow us to test this path.

We could write the manual script to use do-release-upgrade with a temporary ppa as an argument if we think that is worth it

renanrodrigo:
I see - feel free to evaluate that. It is good for me as is (:

orndorffgrant:
The failed tests are test bugs that have been fixed in `main` (because configurations on the backend have changed). We'll need to use those test fixes when we verify the release, but they don't require any application code changes/updates for 27.9.

sru/release-27.9/cleanup_cloud_id_shim.sh
```diff
@@ -0,0 +1,62 @@
+x_deb=$1
+b_deb=$2
+series=bionic
```
lucasmoura, line 3: I think this should be xenial right ?

lucasmoura (approved):

orndorffgrant:
FYI I am writing a manual test to do the full upgrade

orndorffgrant:
Added the full upgrade test. Try it with:
```
bash sru/release-27.9/cleanup_cloud_id_shim_ppa.sh ppa:orndorffgrant/test-cleanup-shim
```

lucasmoura (approved):
New test script also worked here

debian/ubuntu-advantage-tools.postinst
```diff
@@ -354,6 +356,17 @@ remove_old_systemd_timers() {
         deb-systemd-helper purge ua-license-check.path > /dev/null || true
         deb-systemd-helper unmask ua-license-check.path > /dev/null || true
     fi
+
+    # If we're do-release-upgrad-ing to bionic, then clean up the xenial-only
+    # cloud-id-shim unit
+    if [ "$VERSION_ID" = "18.04" ]; then
+        if echo "$PREVIOUS_PKG_VER" | grep -q "16.04"; then
```
basak, line 363: This match against 16.04 seems fragile to me, but you can leave it as-is if you prefer.

basak (approved):
Looks good. One minor (not blocking) inline comment.

PR #2125, opened 2022-06-14
Handle reverse dependencies

lucasmoura:
## Proposed Commit Message
Handle reverse dependencies

Properly sort service disable/enable order based either on required or dependent services necessities.

PS: This PR is also updating both required and dependent services to explicit define which `UAEntitlment` classes that rely upon.

## Test Steps
Run the modified unit tests and verify that the `ros` integration tests are still working as expected

## Desired commit type

## Checklist


uaclient/entitlements/base.py
```diff
@@ -132,7 +132,7 @@ def incompatible_services(self) -> Tuple[IncompatibleService, ...]:
         return self._incompatible_services
 
     @property
-    def required_services(self) -> Tuple[str, ...]:
+    def required_services(self) -> Tuple[Type["UAEntitlement"], ...]:
```
CalvoM, line 135: Might be a nit-pick, is it possible to have no speech marks on the type to be consistent with line 62?

uaclient/entitlements/base.py
```diff
@@ -142,7 +142,7 @@ def required_services(self) -> Tuple[str, ...]:
         return self._required_services
 
     @property
-    def dependent_services(self) -> Tuple[str, ...]:
+    def dependent_services(self) -> Tuple[Type["UAEntitlement"], ...]:
```
CalvoM, line 145: Same possible nit-pick.

uaclient/contract.py
```diff
@@ -309,9 +309,29 @@ def process_entitlements_delta(
...
+        except ValueError:
+            ent_pos = len(ent_order)
+
+        return ent_pos
+
+    sorted_entitlements = sorted(
+        new_entitlements.items(), key=lambda ent: sort_entitlements(ent[0])
+    )
+
+    for name, new_entitlement in sorted_entitlements:
```
orndorffgrant, line 334: Would it work to do something like this?
```suggestion
    for name in entitlements_enable_order(cfg):
        try:
            new_entitlement = new_entitlements[name]
        except KeyError:
            continue
```

uaclient/contract.py
```diff
@@ -309,9 +309,29 @@ def process_entitlements_delta(
...
     delta_error = False
     unexpected_error = False
-    for name, new_entitlement in sorted(new_entitlements.items()):
+
+    # We need to sort our entitlements because some of them
+    # depend on other service to be enable first. Additionally,
+    # we must be careful because no all entitlements that are
+    # provided by the contract are supported by UA.
+    def sort_entitlements(key_name):
+        ent_order = entitlements_enable_order(cfg)
```
orndorffgrant, line 322: If we don't do the refactor I suggest below/above, then we should only call `entitlements_enable_order` once by moving it outside of the `sort_entitlements` function

uaclient/cli.py
```diff
@@ -1222,22 +1223,9 @@ def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
...
-    Therefore, this logic will guarantee that we will always disable ros and
-    ros-updates before diabling the esm services. If that dependency chain
-    change, this logic won't hold anymore and must be properly fixed.
-
-    More details can be seen here:
-    https://github.com/canonical/ubuntu-advantage-client/issues/1831
-    """
-    to_disable.sort(key=lambda ent: len(ent.dependent_services))
+    to_disable.sort(
+        key=lambda ent: entitlements_disable_order(cfg).index(ent.name)
```
orndorffgrant, line 1227: We should only call `entitlements_disable_order` once by moving it outside of the lambda.

E.g.
```
    disable_order = entitlements_disable_order(cfg)
    to_disable.sort(
        key=lambda ent: disable_order.index(ent.name)
    )
```

orndorffgrant: Alternatively we might be able to replace the above for loop that creates the `to_disable` list to avoid needing a sort later.

Something like:
```diff
     to_disable = []
-    for ent_cls in entitlements.ENTITLEMENT_CLASSES:
+    for name in entitlements_disable_order(cfg):
+        try:
+            ent_cls = entitlement_factory(cfg, name)
+        except EntitlementNotFoundError
+            continue
         ent = ent_cls(cfg=cfg, assume_yes=assume_yes)
         # For detach, we should not consider that a service
         # cannot be disabled because of dependent services,
         # since we are going to disable all of them anyway
         ret, _ = ent.can_disable(ignore_dependent_services=True)
         if ret:
             to_disable.append(ent)

-    to_disable.sort(
-        key=lambda ent: entitlements_disable_order(cfg).index(ent.name)
-    )
```

orndorffgrant (commented):
Nice this is a much more robust solution that we had before!

uaclient/contract.py
orndorffgrant: Actually, I guess this isn't a big deal since its only recursing over a list <10 items long. Same for below similar suggestion

uaclient/entitlements/base.py
lucasmoura: The reason for that is that we are using a "type" of the same class that is being created. That's why we need to delay the type registration using ```"UAEntitlement"```

uaclient/entitlements/base.py
```diff
@@ -390,14 +378,11 @@ def check_required_services_active(self):
...
-                ent_cls = entitlement_factory(
-                    cfg=self.cfg, name=required_service
-                )
-                ent_status, _ = ent_cls(self.cfg).application_status()
+                ent_status, _ = required_service_cls(
+                    self.cfg
+                ).application_status()
                 if ent_status != ApplicationStatus.ENABLED:
                     return False
             except exceptions.EntitlementNotFoundError:
```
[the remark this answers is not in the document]
lucasmoura: No, you are right, I will remove that part

uaclient/contract.py
lucasmoura: I like that refacor, even though the recursion is not an issue as you mentioned, but I think this is easier to read than my solution.

lucasmoura:
@orndorffgrant Thanks for the suggestions. I have updated the code

renanrodrigo (approved):
worksforme

uaclient/entitlements/__init__.py
```diff
@@ -25,6 +26,10 @@
     ROSUpdatesEntitlement,
 ]  # type: List[Type[UAEntitlement]]
 
+ENTITLEMENT_CLS_BY_NAME = {
```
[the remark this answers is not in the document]
orndorffgrant: Bump, looks like it can be removed

orndorffgrant (approved):
Thanks for the updates - this looks good and worked for me! Just one minor thing brought up by Renan to be addressed

lucasmoura:

orndorffgrant:
Go ahead and merge as soon as CI passes

PR #2134, opened 2022-06-21
status: bring features from ua version to ua status

renanrodrigo:
Advertise the enabled features on status instead of changing the version string

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -343,7 +343,7 @@ Feature: Command behaviour when attached to an UA subscription
 
     @series.all
     @uses.config.machine_type.lxd.container
-    Scenario Outline: Unattached status in a ubuntu machine with feature overrides
+    Scenario Outline: Attached status in a ubuntu machine with feature overrides
         Given a `<release>` machine with ubuntu-advantage-tools installed
         When I create the file `/tmp/machine-token-overlay.json` with the following:
         """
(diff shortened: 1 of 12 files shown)


features/attached_commands.feature
```diff
@@ -370,14 +370,30 @@ Feature: Command behaviour when attached to an UA subscription
...
-        Then I will see the uaclient version on stdout with features ` +disable_auto_attach +machine_token_overlay -other`
-        When I run `ua version` as non-root
-        Then I will see the uaclient version on stdout with features ` +disable_auto_attach +machine_token_overlay -other`
+        """
+        SERVICE       +ENTITLED  STATUS    DESCRIPTION
+        cc-eal        +no
+        """
+        And stdout matches regexp:
+        """
+        FEATURES
```
blackboxsw, line 379: This looks really good and easy to consume/understand. I like that format=json works shows the specific value types and that custom or typo'd keys are represented to give us hints of misconfiguration possible.

lucasmoura (commented):
Now that we are introducing the `FEATURES` field, it may be worth for us to have a doc explaining that field and the `NOTICES` one. What do you think ?

renanrodrigo:
Good idea - will add

features/attached_commands.feature
```diff
@@ -370,14 +370,30 @@ Feature: Command behaviour when attached to an UA subscription
...
+        \+disable_auto_attach
+        \+machine_token_overlay
+        -other
```
orndorffgrant, line 395: Since this no longer has to fit on one line in the version, I think we can be more verbose.
```suggestion
        disable_auto_attach: True
        machine_token_overlay: /path/to/overlay
        other: other
```

orndorffgrant (approved):
Looks good! My suggestion is not blocking

features/attached_commands.feature
renanrodrigo: ack agree
done

PR #2136, opened 2022-06-22
fix: better output structure

lucasmoura:
## Proposed Commit Message
fix: better output structure

Separate blocks of content generated by ua fix using newlines.

PS: Note that this doesn't fully solve the Launchpad [issue](https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1926182), because when are actually installing package, we might prompt users and I am unsure if we should adapt the prompts too

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/features/fix.feature
+++ b/features/fix.feature
@@ -64,30 +64,36 @@ Feature: Ua fix command behaviour
             """
             USN-4539-1: AWL vulnerability
             Found CVEs:
-            https://ubuntu.com/security/CVE-2020-11728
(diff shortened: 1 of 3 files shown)


orndorffgrant:
@lucasmoura I think if we make sure there is a blank line before and after each prompt, then we can call the LP issue done

orndorffgrant (approved):

renanrodrigo (approved):
failing test is flaky and unrelated

PR #2137, opened 2022-06-23
man: update proxy config

CalvoM:
## Proposed Commit Message

man: update proxy config

We are updating the man pages to include the information about the global and scoped proxy configuration.

## Desired commit type

## Checklist

### DIFF
--- a/ubuntu-advantage.1
+++ b/ubuntu-advantage.1
@@ -220,13 +220,37 @@ If set, ua will use the specified https proxy when making any https requests
 .TP
 .B
 \fBapt_http_proxy\fP
-If set, ua will configure apt to use the specified http proxy by writing a apt
-config file to /etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy
+\fB[DEPRECATED]\fP If set, ua will configure apt to use the specified http proxy by writing a apt
+config file to /etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy. (Please use \fBglobal_apt_http_proxy\fP)
 .TP
 .B
 \fBapt_https_proxy\fP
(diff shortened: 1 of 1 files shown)


orndorffgrant (commented):
Looks great! Just some extra spaces slipped in that should be removed - see comments below.

In addition to those comments, I'm seeing extra spaces in the global_apt_https_proxy and ua_apt_https_proxy paragraphs. I'll paste what I'm seeing here.
```
       global_apt_https_proxy
              If set, ua will configure apt to use the specified https proxy by writing a apt config  file  to  /etc/apt/apt.conf.d/90ubuntu-advantage-
              aptproxy. Set this if you prefer a global proxy for all resources, not just the ones from esm.ubuntu.com

...

       ua_apt_https_proxy
              If set, ua will configure apt to use the specified https proxy by writing a apt config  file  to  /etc/apt/apt.conf.d/90ubuntu-advantage-
              aptproxy. This proxy is limited to accessing resources from esm.ubuntu.com
```

Notice the extra spaces in
```
config  file  to  /etc/...
```
But I'm not sure where they're coming from because I don't see extra spaces in the file :thinking: 
Not a big deal I guess

Edit: interesting github is collapsing extra spaces in inline code blocks...
`config  file  to  /etc/...`

```
config  file  to  /etc/...
```
those two strings are the same, but rendered differently. I don't think this is related to the problem I'm describing above, just interesting...

CalvoM:
@orndorffgrant Concerning the extra spaces, I found this SO question - https://askubuntu.com/questions/1277166/uneven-spacing-in-man-pages.

orndorffgrant (approved):
Thanks for this!

PR #2142, opened 2022-06-27
wip: security status v2

orndorffgrant:
## Proposed Commit Message

```
summary: no more than 70 characters

A description of what the change being made is and why it is being
made, if the summary line is insufficient.  The blank line above is
required. This should be wrapped at 72 characters, but otherwise has
no particular length requirements.

If you need to write multiple paragraphs, feel free.

LP: #NNNNNNN (replace with the appropriate Launchpad bug reference if applicable)
Fixes: #NNNNNNN (replace with the appropriate github issue if applicable)
```

## Desired commit type

## Checklist


orndorffgrant:
@renanrodrigo @lucasmoura @CalvoM I'm starting to think that security_status should not be a single api call, but its parts should be split into several api endpoints.

Maybe:

1. `u.pro.is_attached` for attached bool
2. `u.pro.services.status` for services entitled/enabled
3. `u.packages.count_installed_package_sources` for all the `num_*_packages` fields
4. `u.packages.security.count_updates` for all the `num_*_updates` fields
5. `u.packages.list_available_security_updates` for what is currently the `packages` list

lucasmoura:
@orndorffgrant I think this is a good idea and will help us in the future if we need to surface only some content of `security-status`

renanrodrigo:
Hmmm yes I think it makes sense. Do you think we would then still need the command to aggregate all this information?
Trying to improve the question: do you think the current model for `security-status` should stay besides those, or be replaced by those?

orndorffgrant:
Closing as this will be replaced by the separate api functions

PR #2146, opened 2022-06-30
security-status: include download_size of package updates

orndorffgrant:
## Desired commit type

## Checklist

### DIFF
--- a/uaclient/tests/test_security_status.py
+++ b/uaclient/tests/test_security_status.py
@@ -28,12 +28,15 @@ def mock_origin(
 
 
 def mock_version(
-    version: str, origin_list: List[mock.MagicMock] = []
+    version: str,
+    origin_list: List[mock.MagicMock] = [],
+    size: int = 1,
 ) -> mock.MagicMock:
     mock_version = mock.MagicMock()
     mock_version.__gt__ = lambda self, other: self.version > other.version
     mock_version.version = version
     mock_version.origins = origin_list
+    mock_version.size = size
     return mock_version
 
 
@@ -326,7 +329,7 @@ class TestSecurityStatus:
     ):
         """Make sure the output format matches the expected JSON"""
         cfg = FakeConfig()
-        m_version = mock_version("1.0")
+        m_version = mock_version("1.0", size=123456)
         m_package = mock_package("example_package", m_version)
 
         m_cache.return_value = [m_package] * 10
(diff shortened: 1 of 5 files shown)


uaclient/tests/test_security_status.py
```diff
@@ -28,11 +28,14 @@ def mock_origin(
 
 
 def mock_version(
-    version: str, origin_list: List[mock.MagicMock] = []
+    version: str,
+    size: int,
```
renanrodrigo, line 32: Just as a quick suggestion, you might set a default of `1` for this and only change where needed (:
Just as a quick comment, oh my god those mocks :face_exhaling:

orndorffgrant: good idea! I'll do that

orndorffgrant: updated!

uaclient/tests/test_security_status.py
```diff
@@ -189,7 +192,8 @@ def test_get_ua_info(self, m_status, is_attached, FakeConfig):
             ),
             (
                 mock_version(
-                    "1.0", [MOCK_ORIGINS["archive_main"], MOCK_ORIGINS["now"]]
+                    "1.0",
+                    [MOCK_ORIGINS["archive_main"], MOCK_ORIGINS["now"]],
```
orndorffgrant, line 196: `black` is insisting on these changes - I guess we haven't run `black` on this file since we bumped the version?

renanrodrigo: Nah we run `black` on everything :black_heart: 
This is one of the few things `black` is not strongly opinionated on: the trailing comma. If you remove the comma (as it was before) it can stay as is, but if there is a comma, then you need to split lines.

renanrodrigo: now I see there are many instances of it - which version of `black` is doing this?

renanrodrigo (approved):
Anyway, worksforme! Thanks for the quick PR.

Try to remove the trailing commas if you don't like the format changes, but even if you don't, I don't mind and neither does `black` (:

uaclient/tests/test_security_status.py
orndorffgrant: oh trailing commas!! Okay I'll try removing them. I dislike the unnecessary diff

PR #2147, opened 2022-07-01
data object: add enum data type

CalvoM:
## Proposed Commit Message

data object: add enum data type.

Adding enum type support in data object. We have enum support in the contract server response.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/data_types.py
+++ b/uaclient/data_types.py
@@ -1,4 +1,5 @@
 import json
+from enum import Enum
 from typing import Any, List, Optional, Type, TypeVar, Union
 
 from uaclient import exceptions
@@ -11,6 +12,9 @@ INCORRECT_LIST_ELEMENT_TYPE_ERROR_MESSAGE = (
 )
 INCORRECT_FIELD_TYPE_ERROR_MESSAGE = (
     'Got value with incorrect type for field "{key}": {nested_msg}'
+)
+INCORRECT_ENUM_VALUE_ERROR_MESSAGE = (
+    "Value provided was not found in {enum_class}'s allowed: value: {values}"
 )
 
 
@@ -38,6 +42,13 @@ class IncorrectFieldTypeError(IncorrectTypeError):
         self.key = key
 
 
+class IncorrectEnumValueError(IncorrectTypeError):
+    def __init__(self, values: List[str], enum_class: Any):
+        self.msg = INCORRECT_ENUM_VALUE_ERROR_MESSAGE.format(
+            values=values, enum_class=repr(enum_class)
+        )
+
+
 class DataValue:
     """
     Generic data value to be extended by more specific typed data values.
@@ -48,6 +59,25 @@ class DataValue:
     @staticmethod
     def from_value(val: Any) -> Any:
         return val
+
+
+E = TypeVar("E", bound="EnumDataValue")
+
+
+class EnumDataValue(DataValue, Enum):
+    """
+    To be used for parsing enum values
+    from_value raises an error if the value is not in the enum class values
+    and returns the value if found.
+    """
+
+    @classmethod
+    def from_value(cls: Type[E], val: Any) -> E:
+        try:
+            return cls(val)
+        except ValueError:
+            values = [i.value for i in cls]
+            raise IncorrectEnumValueError(values, cls)
 
 
 class StringDataValue(DataValue):
@@ -116,7 +146,7 @@ def data_list(data_cls: Type[DataValue]) -> Type[DataValue]:
 
 
 def data_list_to_list(
-    val: List[Union["DataObject", list, str, int, bool]],
+    val: List[Union["DataObject", list, str, int, bool, Enum]],
     keep_none: bool = True,
 ) -> list:
     new_val = []  # type: list
@@ -125,6 +155,8 @@ def data_list_to_list(
             new_val.append(item.to_dict(keep_none))
         elif isinstance(item, list):
             new_val.append(data_list_to_list(item, keep_none))
+        elif isinstance(item, Enum):
+            new_val.append(item.value)
         else:
             new_val.append(item)
     return new_val
@@ -179,6 +211,8 @@ class DataObject(DataValue):
                 new_val = val.to_dict(keep_none)
             elif isinstance(val, list):
                 new_val = data_list_to_list(val, keep_none)
+            elif isinstance(val, Enum):
+                new_val = val.value
             else:
                 # simple type, just copy
                 new_val = val
--- a/uaclient/tests/test_data_types.py
+++ b/uaclient/tests/test_data_types.py
@@ -6,7 +6,9 @@ from uaclient.data_types import (
     BoolDataValue,
     DataObject,
     DataValue,
+    EnumDataValue,
     Field,
+    IncorrectEnumValueError,
     IncorrectFieldTypeError,
     IncorrectListElementTypeError,
     IncorrectTypeError,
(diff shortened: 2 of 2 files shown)


uaclient/data_types.py
```diff
@@ -48,6 +61,25 @@ def from_value(val: Any) -> Any:
...
+
+    @classmethod
+    def from_value(cls, val: Any) -> str:
+        expected_values = tuple(i.value for i in cls)  # type: ignore
+        if not isinstance(val, type(expected_values[0])):
+            raise IncorrectTypeError(
+                type(val).__name__, type(expected_values[0]).__name__
+            )
+        if val not in expected_values:
+            raise IncorrectEnumValueError(str(val), cls)
```
orndorffgrant, line 79: I think we want to return an actual enum value here, not the raw int/string.

Also, turns out enums have a lot of what we need already! This can be simplified to something like:
```suggestion
        try:
            return = cls(val)
        except ValueError:
            raise IncorrectEnumValueError([i.value for i in cls])
```

uaclient/data_types.py
```diff
@@ -11,6 +11,12 @@
 INCORRECT_FIELD_TYPE_ERROR_MESSAGE = (
     'Got value with incorrect type for field "{key}": {nested_msg}'
 )
+INCORRECT_ENUM_VALUE_TYPE_ERROR_MESSAGE = (
+    "Got value {value} of type {value_type} which is not supported in enum"
+)
```
orndorffgrant, line 16: I don't think we need this

uaclient/data_types.py
```diff
@@ -11,6 +11,12 @@
...
+INCORRECT_ENUM_VALUE_ERROR_MESSAGE = (
+    "Got value {value} which is not found in enum {enum_class}"
+)
```
orndorffgrant, line 19: Let's avoid printing actual values. But we can print the list of all allowed values. Maybe something like:
```
"Value was not found in TestEnum allowed values: [val1, val2, val3]"
```

uaclient/data_types.py
```diff
@@ -36,6 +42,13 @@ def __init__(self, err: IncorrectTypeError, key: str):
         )
 
 
+class IncorrectEnumValueError(exceptions.UserFacingError):
```
orndorffgrant, line 45: I think this needs to extend IncorrectTypeError to be caught properly by the recursive parsing functions

orndorffgrant (commented):
Thanks for this PR! I think enumms will be very valuable. Left some implementation comments here.

Also we'll need to support `to_dict` for enums as well. So I think we'll need another case in `DataObject.to_dict` as well as `data_list_to_list`

uaclient/data_types.py
```diff
@@ -50,6 +61,22 @@ def from_value(val: Any) -> Any:
         return val
 
 
+class EnumDataValue(DataValue):
+    """
+    To be used for parsing enum values
+    from_value raises an error if the value is not in the enum class values
+    and returns the value if found.
+    """
```
orndorffgrant, line 69: Let's include something here about how to correctly create an EnumDataValue enum. Importantly, we should specify to inherit from both `EnumDataValue` and `Enum`

Something like:
```
To define a new EnumDataValue, create a new class that inherits from both `EnumDataValue` and `Enum`. E.g.
    class MyEnum(EnumDataValue, Enum):
        ONE = 1
        TWO = 2
```

uaclient/data_types.py
```diff
@@ -50,6 +61,22 @@ def from_value(val: Any) -> Any:
...
+
+    @classmethod
+    def from_value(cls, val: Any) -> str:
```
orndorffgrant, line 72: We'll need to do something similar to DataObject for the typing here. Something like:
```python
E = TypeVar("E", bound="EnumDataValue")
class EnumDataValue(DataValue, Enum):
    @classmethod
    def from_value(cls: Type[E], val: Any) -> E:
        try:
            return cls(val)
        except ValueError:
            values = [i.value for i in cls]
            raise IncorrectEnumValueError(values, cls)
```

uaclient/data_types.py
```diff
@@ -126,7 +153,10 @@ def data_list_to_list(
         elif isinstance(item, list):
             new_val.append(data_list_to_list(item, keep_none))
         else:
-            new_val.append(item)
+            if isinstance(item, Enum):
+                new_val.append(item.value)
+            else:
+                new_val.append(item)
```
orndorffgrant, line 159: Let's just continue the existing `elif`s
```python
        elif isinstance(item, Enum):
            new_val.append(item.value)
        else:
            new_val.append(item)
```

uaclient/data_types.py
```diff
@@ -180,8 +210,11 @@ def to_dict(self, keep_none: bool = True) -> dict:
             elif isinstance(val, list):
                 new_val = data_list_to_list(val, keep_none)
             else:
-                # simple type, just copy
-                new_val = val
+                if isinstance(val, Enum):
+                    new_val = val.value
+                else:
+                    # simple type, just copy
+                    new_val = val
```
orndorffgrant, line 217: Let's just continue the existing `elif`s
```python
        elif isinstance(val, Enum):
            new_val = val.value
        else:
            new_val = val
```

uaclient/tests/test_data_types.py
```diff
@@ -170,6 +249,8 @@ class ExampleDataObject(DataObject):
         Field("integerlist_opt", data_list(IntDataValue), required=False),
         Field("objlist", data_list(ExampleNestedObject)),
         Field("objlist_opt", data_list(ExampleNestedObject), required=False),
+        Field("enum_opt_list", data_list(ExampleEnum), required=False),
+        Field("enum_opt", ExampleEnum, required=False),
```
orndorffgrant, line 253: The intent of this example data object is to exercise every possible type. So we need some non optional enums and lists as well. Let's also continue the existing naming patterns for consistency.
```suggestion
        Field("enum", ExampleEnum),
        Field("enum_opt", ExampleEnum, required=False),
        Field("enumlist", data_list(ExampleEnum)),
        Field("enumlist_opt", data_list(ExampleEnum), required=False),

```

uaclient/tests/test_data_types.py
```diff
@@ -229,6 +314,8 @@ def __init__(
         {"string": "nestedstring3", "integer": 7},
     ],
     "objlist_opt": None,
+    "enum_opt_list": None,
+    "enum_opt": None,
 }
 example_data_object_dict_with_optionals = {
```
orndorffgrant, line 320: This object should include all optionals, including `enum_opt`

uaclient/data_types.py
CalvoM: I think adding `Enum` to the class is a better approach. So that usage is only inheriting from EnumDataValue.

uaclient/tests/test_data_types.py
```diff
@@ -249,6 +366,22 @@ def __init__(
         {"string": "nestedstring2_opt", "integer": 66},
         {"string": "nestedstring3_opt", "integer": 77},
     ],
+    "enum_list": ["two", "three"],
+    "enum": "one",
+    "enum_opt_list": ["one", "four"],
```
orndorffgrant, line 371: `example_data_object_dict_with_optionals` should include a value for all of the optional fields. So we need a value for `enum_opt` here.

CalvoM: Yes, now the earlier comment makes sense.

uaclient/tests/test_data_types.py
CalvoM: I thought you were talking about line 317 and not 320

orndorffgrant: Ah yeah my bad: github is annoying in not making it clear which line I clicked on to comment on :(

PR #2151, opened 2022-07-07
prevent unneeded daemon activation and mitigate stalled daemon bug

orndorffgrant:
~~Note that this does not fix the root cause of the stalled daemon reported in https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1980865~~

~~Debugging that may take some more time and I think we should include this mitigation in the upcoming release.~~

The daemon was stalling because it is configured "After: ua-auto-attach.service" but was started during that service. This results in a deadlock. So simply not starting the daemon during the auto attach service is the full fix here.

## Test Steps

build the deb and store at `./ua.deb`. Then run the test script here using this branch of pycloudlib: https://github.com/canonical/pycloudlib/pull/204

It should successfully boot the snapshotted image.

If you comment out the lines that install the deb, then the snapshotted image would never finish booting

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1216,7 +1216,10 @@ def action_detach(args, *, cfg) -> int:
 
     @return: 0 on success, 1 otherwise
     """
-    return _detach(cfg, assume_yes=args.assume_yes)
+    ret = _detach(cfg, assume_yes=args.assume_yes)
+    daemon.start()
+    event.process_events()
+    return ret
 
 
 def _detach(cfg: config.UAConfig, assume_yes: bool) -> int:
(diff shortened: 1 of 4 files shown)


uaclient/cli.py
```diff
@@ -1216,7 +1216,10 @@ def action_detach(args, *, cfg) -> int:
 
     @return: 0 on success, 1 otherwise
     """
-    return _detach(cfg, assume_yes=args.assume_yes)
+    ret = _detach(cfg, assume_yes=args.assume_yes)
```
lucasmoura, line 1219: I think we should only start the daemon if `ret` is `0` right ?

orndorffgrant: ah good catch

PR #2152, opened 2022-07-07
api: change magic-attach key names

lucasmoura:
## Proposed Commit Message
api: change magic-attach key names

The contract server has modified some key names in the response of some magic-attach endpoints. We are now reflecting those changes here as well.

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/tests/test_util.py
+++ b/uaclient/tests/test_util.py
@@ -1169,6 +1169,18 @@ class TestRedactSensitiveLogs:
                 "http://metadata/computeMetadata/v1/instance/service-accounts/"
                 "default/identity?audience=contracts.canon, data: <REDACTED>",
             ),
+            (
+                "'token': 'SEKRET'",
+                "'token': '<REDACTED>'",
+            ),
+            (
+                "'userCode': 'SEKRET'",
+                "'userCode': '<REDACTED>'",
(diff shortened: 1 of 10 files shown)


orndorffgrant:
> I understand the field with underscore (e.g., `user_code`) are the adaptation you make for the installer, right?

Yep, our api functions return snake_case, so we just want to be consistent with that

uaclient/tests/test_util.py
```diff
@@ -1169,6 +1169,14 @@ class TestRedactSensitiveLogs:
...
                 "default/identity?audience=contracts.canon, data: <REDACTED>",
             ),
+            (
+                "'token': 'SEKRET'",
+                "'token': '<REDACTED>'",
+            ),
+            (
+                "'userCode': 'SEKRET'",
+                "'userCode': '<REDACTED>'",
+            ),
```
orndorffgrant, line 1179: @lucasmoura another place to redact is
```
Executed with sys.argv: ['/usr/bin/ua', 'api', 'u.pro.attach.magic.wait.v1', '--args', 'magic_token=TOKEN`]
```

lucasmoura: Good catch

PR #2153, opened 2022-07-07
api: include related environment vars in the JSON responses

renanrodrigo:
Add the `environment_vars` field to the JSON API responses.

## Desired commit type

## Checklist


orndorffgrant (approved):
LGTM after the name change to `get_pro_environment` as discussed

orndorffgrant (approved):

features/api.feature
```diff
@@ -7,12 +7,12 @@ Feature: Client behaviour for the API endpoints
     When I verify that running `pro api invalid.endpoint` `with sudo` exits `1`
     Then stdout matches regexp:
     """
-    {\"_schema_version\": \"v1\", \"data\": {}, \"errors\": \[{\"code\": \"api\-invalid\-endpoint", \"meta\": {}, \"title\": \"'invalid\.endpoint' is not a valid endpoint\"}], \"result\": \"failure\", \"version\": \".*\", \"warnings\": \[]}
```
lucasmoura, line 10: Can we add an integration test where we setup an env var ?

uaclient/api/data_types.py
```diff
@@ -31,10 +30,16 @@ class APIData(DataObject):
         Field("meta", DataObject),
     ]
 
-    def __init__(self, *, type: str, attributes: DataObject, meta: DataObject):
+    def __init__(self, *, type: str, attributes: DataObject, meta: dict):
```
lucasmoura, line 33: We have two `meta` keys in our response. The top level one and the one inside `data`. Since this is a generic information, I think we should only include it on the top level `meta` field. This will also make it consistent with errors generated by the API

features/api.feature
renanrodrigo: sure

uaclient/api/data_types.py
renanrodrigo: Actually there is no top-level `meta` - there is one `meta` for the Data, and one for each Error/Warning.
Do you think we should unify those as a single top-level key meta?

features/api.feature
renanrodrigo: done

lucasmoura (approved):

PR #2154, opened 2022-07-11
magic-attach: Don't fail on first connection error

lucasmoura:
## Proposed Commit Message
magic-attach: Don't fail on first connection error

When running the wait endpoint, we will now try at least 3 times if we receive a connectivity error. The rationale for that is
to avoid flake connectivity errors to stop the execution of the command

## Test Steps
Run the new unit test and check that the integration test for magic attach is still working as expected

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/api/u/pro/attach/magic/wait/v1.py
+++ b/uaclient/api/u/pro/attach/magic/wait/v1.py
@@ -61,6 +61,7 @@ def wait(
     contract = UAContractClient(cfg)
 
     num_attempts = 0
+    num_connection_errors = 0
 
     while num_attempts < MAXIMUM_ATTEMPTS:
         try:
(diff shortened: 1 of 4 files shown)


renanrodrigo (approved):
lgtm

ogayot (commented):
Thanks for doing this! I appreciate it.

This looks better now. May I suggest to reset `num_connection_errors` to zero after a successful connection though?

<details>
<summary>
Also, if you're interested in fixing typos in tests
</summary>
<p>

* s/succeds/succeeds/
* s/connectiviry/connectivity/

</p>
</details>

uaclient/api/u/pro/attach/magic/wait/v1.py
```diff
@@ -69,8 +70,17 @@ def wait(
...
             break
-
-        if wait_resp.get("contractToken") is not None:
+        # If we have a flaky connectivity error, this part of the code
+        # will make sure that we try at least three more times before
+        # raising a ConnectivityError.
+        except exceptions.ConnectivityError as e:
+            if num_connection_errors < 3:
+                num_connection_errors += 1
+                wait_resp = None
```
ogayot, line 79: I guess you could also use `continue` here, but it's up to you of course.

lucasmoura: My only concern with `continue` here is that we would not call sleep and increment the number of attempts. That's why I am using that approach

ogayot: Oh you're right. I read the code wrong. My bad!

PR #2158, opened 2022-07-12
Release 27.10

lucasmoura:


docs/howtoguides/configure_proxies.md
```diff
@@ -1,17 +1,100 @@
...
+> **Note**
+> Any configuration file that comes later in the apt.conf.d
+> directory could override the proxy configured by the UA client.
+
+To remove the APT proxy configuration, run the following:
+
+$ sudo ua config unset apt\_http\_proxy
+$ sudo ua config unset apt\_https\_proxy
+
+> **Note**
```
CalvoM, line 61: Considering this is 27.10, could we add that `apt_http(s)_proxy` will be deprecated.

lucasmoura: Yes, I think we can improve on that, but let's use a separate PR for this. The main goal of this PR is just to align on what we will release next and not to provide any extra functionality

renanrodrigo (approved):
LGTM
ok to wait a little if we want to include the other doc PR about the deprecation of the proxy vars

lucasmoura:
@renanrodrigo I don't think this is necessary. We don't ship the docs in the package, so we can do this in parallel

docs/howtoguides/configure_proxies.md
CalvoM: Okay. No problem.

PR #2160, opened 2022-07-14
Update pycloudlib

lucasmoura:
## Proposed Commit Message
bump pycloudlib version
    
With the latest version of pycloudlib, we can now fetch the Pro and Pro FIPS images directly from the clouds. Because of that, we are now removing the logic to launch those types of images from our integration tests

## Test Steps
Run the Pro and Pro FIPS test and see that they are working as expected

## Desired commit type

## Checklist


renanrodrigo (approved):
worksforme

PR #2161, opened 2022-07-15
status: correct data in non-root

CalvoM:
## Proposed Commit Message

status: correct data in non-root

Currently, when `ua status` as non-root it gets its status from status cache file.
Incase of any changes on the contract server concerning the machine's account,
we do not get up to date info.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -307,7 +307,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         super_status, super_msg = super().application_status()
 
         if system.is_container() and not system.should_reboot():
-            self.cfg.notice_file.remove(
+            self.cfg.notice_file.try_remove(
                 "", messages.FIPS_SYSTEM_REBOOT_REQUIRED.msg
             )
             return super_status, super_msg
@@ -317,21 +317,23 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
             # We are now only removing the notice if there is no reboot
             # required information regarding the fips metapackage we install.
             if not system.should_reboot(set(self.packages)):
-                self.cfg.notice_file.remove(
+                self.cfg.notice_file.try_remove(
                     "", messages.FIPS_SYSTEM_REBOOT_REQUIRED.msg
                 )
 
-            self.cfg.notice_file.remove("", messages.FIPS_REBOOT_REQUIRED_MSG)
+            self.cfg.notice_file.try_remove(
+                "", messages.FIPS_REBOOT_REQUIRED_MSG
+            )
             if system.load_file(self.FIPS_PROC_FILE).strip() == "1":
-                self.cfg.notice_file.remove(
+                self.cfg.notice_file.try_remove(
                     "", messages.NOTICE_FIPS_MANUAL_DISABLE_URL
                 )
                 return super_status, super_msg
             else:
-                self.cfg.notice_file.remove(
+                self.cfg.notice_file.try_remove(
                     "", messages.FIPS_DISABLE_REBOOT_REQUIRED
                 )
-                self.cfg.notice_file.add(
+                self.cfg.notice_file.try_add(
                     "", messages.NOTICE_FIPS_MANUAL_DISABLE_URL
                 )
                 return (
@@ -341,7 +343,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
                     ),
                 )
         else:
-            self.cfg.notice_file.remove(
+            self.cfg.notice_file.try_remove(
                 "", messages.FIPS_DISABLE_REBOOT_REQUIRED
             )
 
@@ -379,10 +381,12 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
 
     def _perform_enable(self, silent: bool = False) -> bool:
         if super()._perform_enable(silent=silent):
-            self.cfg.notice_file.remove(
+            self.cfg.notice_file.try_remove(
                 "", messages.NOTICE_WRONG_FIPS_METAPACKAGE_ON_CLOUD
             )
-            self.cfg.notice_file.remove("", messages.FIPS_REBOOT_REQUIRED_MSG)
+            self.cfg.notice_file.try_remove(
+                "", messages.FIPS_REBOOT_REQUIRED_MSG
+            )
             return True
 
         return False
@@ -507,7 +511,9 @@ class FIPSEntitlement(FIPSCommonEntitlement):
                 "defaulting to generic FIPS package."
             )
         if super()._perform_enable(silent=silent):
-            self.cfg.notice_file.remove("", messages.FIPS_INSTALL_OUT_OF_DATE)
+            self.cfg.notice_file.try_remove(
+                "", messages.FIPS_INSTALL_OUT_OF_DATE
+            )
             return True
 
         return False
@@ -568,7 +574,7 @@ class FIPSUpdatesEntitlement(FIPSCommonEntitlement):
 
     def _perform_enable(self, silent: bool = False) -> bool:
         if super()._perform_enable(silent=silent):
-            self.cfg.notice_file.remove(
+            self.cfg.notice_file.try_remove(
                 "", messages.FIPS_DISABLE_REBOOT_REQUIRED
             )
             services_once_enabled = (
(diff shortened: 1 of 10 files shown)


uaclient/tests/test_status.py
```diff
@@ -574,9 +575,19 @@ def test_root_followed_by_nonroot(
 
         # Run as regular user and confirm that we see the result from
         # last time we called .status()
+        # Since machine is attached then diff is found
         m_getuid.return_value = 1000
```
lucasmoura, line 579: Do we still need this mock here ?

uaclient/tests/test_status.py
```diff
@@ -574,9 +575,19 @@ def test_root_followed_by_nonroot(
...
+        cfg = FakeConfig(root_mode=False)
         after = status.status(cfg=cfg)
```
lucasmoura, line 581: I think we also need a check to see if `root` and `non-root` status are the same

uaclient/tests/test_status.py
```diff
@@ -795,6 +806,7 @@ def test_expires_handled_appropriately(
         _m_remove_notice,
         all_resources_available,
         FakeConfig,
+        tmpdir,
```
lucasmoura, line 809: Why do we need this fixture here ?

uaclient/tests/test_status.py
```diff
@@ -855,7 +875,8 @@ def test_nonroot_user_uses_cache_and_updates_if_available(
             }
         )
 
-        assert expected_status == status.status(cfg=cfg)
+        assert expected_status != status.status(cfg=cfg)
```
lucasmoura, line 878: Shouldn't they be the same here ?

uaclient/tests/test_status.py
CalvoM: No we do not actually. Removing.

uaclient/tests/test_status.py
CalvoM: I was testing something, forgot to cleanup.

uaclient/tests/test_status.py
CalvoM: No, because the `expected_status` is written to status-cache which non-root does not use.

uaclient/tests/test_status.py
CalvoM: Makes sense.

lucasmoura (commented):
Maybe I missing this test, but couldn't spot one where attached non-root and root report the same values. We could use this test to also assert than even on attached scenarios, we don't use the `status-cache`

uaclient/tests/test_status.py
```diff
@@ -574,16 +564,37 @@ def test_root_followed_by_nonroot(
...
-        m_getuid.return_value = 1000
+        # Since machine is attached then diff is found
+        cfg = FakeConfig(root_mode=False)
         after = status.status(cfg=cfg)
 
+        assert before != after
+
+        # delete machine_token file first to make unattached
+        cfg = FakeConfig()
+        cfg.machine_token_file.delete()
```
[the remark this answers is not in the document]
CalvoM: Actually, I found out that it is the same `tmpdir` object.

renanrodrigo (approved):
lgtm, pending only Lucas's review points (test the `try`s)

uaclient/entitlements/fips.py
```diff
@@ -301,49 +301,53 @@ def packages(self) -> List[str]:
...
     def application_status(
         self,
     ) -> Tuple[ApplicationStatus, Optional[messages.NamedMessage]]:
         super_status, super_msg = super().application_status()
 
         if system.is_container() and not system.should_reboot():
-            self.cfg.notice_file.remove(
-                "", messages.FIPS_SYSTEM_REBOOT_REQUIRED.msg
-            )
+            self._notice_remove("", messages.FIPS_SYSTEM_REBOOT_REQUIRED.msg)
```
lucasmoura, line 322: Let's add a test here that shows that a non-root call where we should delete the notices, doesn't fail the `application_status` call

lucasmoura (commented):
@CalvoM just noticed that the `machine_token.json` file is now being stored outside of the `private` folder under `/var/lib/ubuntu-advantage/`. This looks like a bug on the `MachineTokenFile` class.

Additionally, I performed an extra test where I deleted the status cache on an attached machine and ran `pro status` as non-root. In that situation, the output is showing the system as unattached. This looks odd, since I believe we shouldn't be relying on the status cache for status anymore

uaclient/entitlements/fips.py
```diff
@@ -301,49 +301,53 @@ def packages(self) -> List[str]:
         packages = super().packages
         return self._replace_metapackage_on_cloud_instance(packages)
 
+    def _notice_remove(self, label: str, description: str):
+        try:
+            self.cfg.notice_file.remove(label, description)
+        except exceptions.NonRootUserError:
+            pass
```
orndorffgrant, line 308: I'd rather see this function on the notice_file itself, so it can be used in other places if needed.

It could look like this:
```python
notice_file.remove() # raises NonRootUserError when not root
notice_file.try_remove() # logs warning but doesn't raise error when not root
```

orndorffgrant: for example, I think we need to catch the NonRootUserError in cli.py line 1557

uaclient/tests/test_status.py
```diff
@@ -536,20 +530,17 @@ def test_nonroot_unattached_is_same_as_unattached_root(
...
 
-        m_getuid.return_value = 0
+        cfg = FakeConfig(root_mode=True)
         root_unattached_status = status.status(cfg=cfg)
 
         assert root_unattached_status == nonroot_status
 
     @mock.patch("uaclient.status.get_available_resources")
-    @mock.patch("uaclient.status.os.getuid")
     def test_root_followed_by_nonroot(
```
orndorffgrant, line 542: I think this whole test can just be removed, because its purpose was to test that the nonroot used the cache. Right? @lucasmoura @CalvoM

uaclient/tests/test_status.py
```diff
@@ -574,16 +564,37 @@ def test_root_followed_by_nonroot(
...
+        m_get_available_resources,
+        _m_should_reboot,
+        _m_remove_notice,
+        FakeConfig,
+    ):
+        root_cfg = FakeConfig()
+        root_status = status.status(cfg=root_cfg)
+        normal_cfg = FakeConfig(root_mode=False)
+        normal_status = status.status(cfg=normal_cfg)
+        assert normal_status == root_status
```
orndorffgrant, line 593: We should have a version of this test for an attached config.

uaclient/tests/test_status.py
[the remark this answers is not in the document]
orndorffgrant: the tmpdir is created when creating the fixture. So each unit test gets its own tmpdir, but within a function, each instance of `FakeConfig` will re-use the same tmpdir

uaclient/tests/test_status.py
```diff
@@ -811,33 +818,36 @@ def test_expires_handled_appropriately(
                 },
             },
         }
-        cfg = FakeConfig.for_attached_machine(
+        cfg = FakeConfig().for_attached_machine(
```
orndorffgrant, line 821: `for_attached_machine` is a classmethod
```suggestion
        cfg = FakeConfig.for_attached_machine(
```

uaclient/tests/test_status.py
orndorffgrant: (Accidentally clicked the resolve/unresolve button three times haha)

If we don't expect the status to be `expected_status` then `expected_status` isn't really the expected status, is it?

Let's rename `expected_status` to `status_cache_value`

renanrodrigo (approved):
lgtm

uaclient/tests/test_status.py
orndorffgrant: @CalvoM tagging to make sure you see this comment ^^

orndorffgrant (approved):
Looks good! Please just see my suggestion about renaming the non-expected `expected_status` test variable

uaclient/tests/test_status.py
CalvoM: I deleted the test as we do not use `status_cache` for non-root user. @orndorffgrant

orndorffgrant:
Thanks for the small fix @CalvoM LGTM

@lucasmoura did you want to look at this again before we merge? (since you've left several comments on it so far)

PR #2163, opened 2022-07-19
Detect pro api

lucasmoura:
## Proposed Commit Message
api: add detect is_pro endpoint

Add endpoint that detect if we are running on a Pro machine or not

## Test Steps
Run the updated integration tests

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/api/api.py
+++ b/uaclient/api/api.py
@@ -18,6 +18,7 @@ VALID_ENDPOINTS = [
     "u.pro.attach.magic.initiate.v1",
     "u.pro.attach.magic.wait.v1",
     "u.pro.attach.magic.revoke.v1",
+    "u.pro.attach.auto.should_auto_attach.v1",
 ]
 
 
--- a/features/api.feature
+++ b/features/api.feature
@@ -25,7 +25,7 @@ Feature: Client behaviour for the API endpoints
 
     @series.all
     @uses.config.machine_type.lxd.container
-    Scenario Outline: Version v1 endpoint
+    Scenario Outline: Basic endpoints
     Given a `<release>` machine with ubuntu-advantage-tools installed
     When I run `pro api u.pro.version.v1` with sudo
     Then stdout matches regexp:
@@ -37,6 +37,12 @@ Feature: Client behaviour for the API endpoints
     """
     {\"_schema_version\": \"v1\", \"data\": {\"attributes\": {\"installed_version\": \".*\"}, \"meta\": {\"environment_vars\": \[{\"name\": \"UA_LOG_FILE\", \"value\": \"\/tmp\/some_file\"}]}, \"type\": \"Version\"}, \"errors\": \[], \"result\": \"success\", \"version\": \".*\", \"warnings\": \[]}
     """
+    When I run `ua api u.pro.attach.auto.should_auto_attach.v1` with sudo
+    Then stdout matches regexp:
+    """
+    {"_schema_version": "v1", "data": {"attributes": {"should_auto_attach": false}, "meta": {"environment_vars": \[\]}, "type": "ShouldAutoAttach"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+    """
+
     Examples: ubuntu release
            | release |
            | bionic  |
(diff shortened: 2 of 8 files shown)


uaclient/api/u/pro/detect/is_pro/v1.py
```diff
@@ -0,0 +1,50 @@
...
+    BoolDataValue,
+    DataObject,
+    Field,
+    StringDataValue,
+)
+
+
+class DetectIsProResult(DataObject, AdditionalInfo):
+    fields = [
+        Field("_schema", StringDataValue),
```
orndorffgrant, line 17: Why is there a _schema field?

uaclient/api/api.py
```diff
@@ -18,6 +18,7 @@
     "u.pro.attach.magic.initiate.v1",
     "u.pro.attach.magic.wait.v1",
     "u.pro.attach.magic.revoke.v1",
+    "u.pro.detect.is_pro.v1",
```
orndorffgrant, line 21: Naming bikeshed: the instance isn't pro yet, but it we're detecting if it _should_ be pro.

How about: "u.pro.detect.is_pro_eligible.v1"

of even: "u.pro.detect.should_auto_attach.v1"

thoughts?

uaclient/api/u/pro/detect/is_pro/v1.py
lucasmoura: :facepalm: removing it

uaclient/api/api.py
lucasmoura: I don't know, the first one is a bit misleading because it might sound that machine is not pro yet, but it could be. However, our is to detect the running machine is already Pro.

Because of that, I think the second one is better. I will use it instead

uaclient/api/u/pro/detect/should_auto_attach/v1.py
```diff
@@ -0,0 +1,41 @@
...
+class ShouldAutoAttachResult(DataObject, AdditionalInfo):
+    fields = [
+        Field("should_auto_attach", BoolDataValue),
+    ]
+
+    def __init__(self, should_auto_attach: bool):
+        self.should_auto_attach = should_auto_attach
+
+
+def should_auto_attach_on_machine(cfg: UAConfig):
```
orndorffgrant, line 19: I expected this to match the api endpoint name: `should_auto_attach`. Do we need the "_on_machine"?

uaclient/api/u/pro/detect/should_auto_attach/v1.py
```diff
@@ -0,0 +1,41 @@
...
+    try:
+        cloud_instance_factory()
+    except (
+        exceptions.CloudFactoryNoCloudError,
+        exceptions.CloudFactoryUnsupportedCloudError,
+        exceptions.CloudFactoryNonViableCloudError,
```
orndorffgrant, line 25: These all extend `CloudFactoryError`
```suggestion
        exceptions.CloudFactoryError,
```

uaclient/api/u/pro/detect/should_auto_attach/v1.py
lucasmoura: Yeah, this is just so we avoid naming the function and the Result object attribute the same, but I guess it is better to change the attribute name instead

uaclient/api/u/pro/detect/should_auto_attach/v1.py
lucasmoura: Good point

uaclient/api/u/pro/detect/should_auto_attach/v1.py
orndorffgrant: @lucasmoura I wouldn't mind if the function and the attribute have the same name :upside_down_face:

lucasmoura: Yeah, good point. I think this is not a concern at all. Changing it

uaclient/api/api.py
```diff
@@ -18,6 +18,7 @@
     "u.pro.attach.magic.initiate.v1",
     "u.pro.attach.magic.wait.v1",
     "u.pro.attach.magic.revoke.v1",
+    "u.pro.detect.should_auto_attach.v1",
```
orndorffgrant, line 21: Sorry to drag the naming discussion along further, but keeping in mind the conversation here: https://github.com/canonical/ubuntu-advantage-client/pull/2171#discussion_r930243892 

I think this would fit nicely with my proposal there if it was named: `u.pro.attach.auto.should_auto_attach.v1`

lucasmoura:
@orndorffgrant updated

orndorffgrant (approved):
Thanks! LGTM

PR #2166, opened 2022-07-21
auto-attach: no-op when ua info on cloud-init data

lucasmoura:
## Proposed Commit Message
auto-attach: no-op when ua info on cloud-init data

When we detect the ubuntu_advantage key on cloud-init userdata, we should no-op the ua-auto-attach service. That's because if we detect that key, cloud-init will be responsible for orchestrating the auto-attach operation on their side.

## Test Steps
Run the new integration tests

## Desired commit type

## Checklist

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -648,7 +648,8 @@ def create_instance_with_uat_installed(
     user_data = _get_user_data_for_instance(context.config, series)
 
     if custom_user_data:
-        user_data += "\n{}".format(custom_user_data)
+        prefix = "" if user_data else "#cloud-config"
+        user_data += "{}\n{}".format(prefix, custom_user_data)
 
     logging.info(
         "--- Launching VM to create a base image with ubuntu-advantage"
--- a/features/environment.py
+++ b/features/environment.py
@@ -614,7 +614,10 @@ def build_debs_from_sbuild(context: Context, series: str) -> List[str]:
 
 
 def create_instance_with_uat_installed(
-    context: Context, series: str, name: str
+    context: Context,
+    series: str,
+    name: str,
+    custom_user_data: Optional[str] = None,
 ) -> pycloudlib.instance.BaseInstance:
     """Create a given series lxd image with ubuntu-advantage-tools installed
 
@@ -627,6 +630,10 @@ def create_instance_with_uat_installed(
         it.
     :param series:
        A string representing the series name to create
+    :param name:
+       A string representing the instance name
+    :param custom_user_data:
+       A string representing custom userdata to be added to the instance
 
     :return: A pycloudlib Instance
     """
@@ -639,6 +646,9 @@ def create_instance_with_uat_installed(
         return
 
     user_data = _get_user_data_for_instance(context.config, series)
+
+    if custom_user_data:
+        user_data += "\n{}".format(custom_user_data)
 
     logging.info(
         "--- Launching VM to create a base image with ubuntu-advantage"
@@ -714,7 +724,7 @@ def _get_user_data_for_instance(
         ppa_url=ppa, ppa_keyid=ppa_keyid
     )
 
-    logging.debug("User data:\n" + user_data)
+    logging.info("-- user data:\n" + user_data)
     return user_data
 
 
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -59,7 +59,7 @@ def add_test_name_suffix(context, series, prefix):
 
 
 @given("a `{series}` machine with ubuntu-advantage-tools installed")
-def given_a_machine(context, series):
+def given_a_machine(context, series, custom_user_data=None):
     if series in context.reuse_container:
         context.instances = {}
         context.container_name = context.reuse_container[series]
@@ -103,7 +103,7 @@ def given_a_machine(context, series):
         )
     else:
         inst = create_instance_with_uat_installed(
-            context, series, instance_name
+            context, series, instance_name, custom_user_data
         )
 
     context.series = series
@@ -141,6 +141,14 @@ def when_i_take_a_snapshot(context):
             )
 
     context.add_cleanup(cleanup_image)
+
+
+@given(
+    "a `{series}` machine with ubuntu-advantage-tools installed adding this cloud-init user_data"  # noqa
+)
+def given_a_machine_with_user_data(context, series):
+    custom_user_data = context.text
+    given_a_machine(context, series, custom_user_data)
 
 
 @when("I have the `{series}` debs under test in `{dest}`")
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -74,13 +74,13 @@ def given_a_machine(context, series, custom_user_data=None):
 
     instance_name = add_test_name_suffix(context, series, CONTAINER_PREFIX)
 
-    if context.config.snapshot_strategy:
+    if context.config.snapshot_strategy and not custom_user_data:
         if series not in context.series_image_name:
             build_container_name = add_test_name_suffix(
                 context, series, IMAGE_BUILD_PREFIX
             )
             image_inst = create_instance_with_uat_installed(
-                context, series, build_container_name
+                context, series, build_container_name, custom_user_data
             )
 
             image_name = add_test_name_suffix(context, series, IMAGE_PREFIX)
@@ -92,6 +92,7 @@ def given_a_machine(context, series, custom_user_data=None):
                 image_name=image_name,
                 cloud_api=context.config.cloud_api,
             )
+
             context.series_image_name[series] = image_id
             image_inst.delete(wait=False)
 
--- a/features/ubuntu_pro.feature
+++ b/features/ubuntu_pro.feature
@@ -580,3 +580,83 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
            | bionic  | disabled | disabled | disabled | libkrad0  | bundler  | enabled   | cis        |
            | focal   | disabled | n/a      | disabled | hello     | ant      | enabled   | usg        |
            | jammy   | n/a      | n/a      | n/a      | hello     | hello    | enabled   | usg        |
+
+    @series.lts
+    @uses.config.machine_type.gcp.pro
+    @uses.config.machine_type.aws.pro
+    @uses.config.machine_type.azure.pro
+    Scenario Outline: Auto-attach service works on Pro Machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `systemctl start ua-auto-attach.service` with sudo
+        And I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
+        """
+        contract_url: 'https://contracts.canonical.com'
+        data_dir: /var/lib/ubuntu-advantage
+        log_level: debug
+        log_file: /var/log/ubuntu-advantage.log
+        """
+        And I reboot the `<release>` machine
+        And I run `pro status --wait` with sudo
+        And I run `pro security-status --format json` with sudo
+        Then stdout matches regexp:
+        """
+        "attached": true
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
+           | jammy   |
+
+    @series.lts
+    @uses.config.machine_type.gcp.pro
+    @uses.config.machine_type.aws.pro
+    @uses.config.machine_type.azure.pro
+    Scenario Outline: Auto-attach no-op when cloud-init has ubuntu_advantage on userdata
+        Given a `<release>` machine with ubuntu-advantage-tools installed adding this cloud-init user_data:
+        # This user_data should not do anything, just guarantee that the ua-auto-attach service
+        # # does nothing
+        """
+        ubuntu_advantage:
+        """
+        When I run `cloud-init query userdata` with sudo
+        Then stdout matches regexp:
+        """
+        ubuntu_advantage:
+        """
+        # On GCP, this service will auto-attach the machine automatically after we override
+        # the uaclient.conf file. To guarantee that we are not auto-attaching on reboot
+        # through the ua-auto-attach.service, we are masking it
+        When I run `systemctl mask ubuntu-advantage.service` with sudo
+        And I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
+        """
+        contract_url: 'https://contracts.canonical.com'
+        data_dir: /var/lib/ubuntu-advantage
+        log_level: debug
+        log_file: /var/log/ubuntu-advantage.log
+        """
+        And I reboot the `<release>` machine
+        And I run `pro status --wait` with sudo
+        And I run `pro security-status --format json` with sudo
+        Then stdout matches regexp:
+        """
+        "attached": false
+        """
+        When I run `cat /var/log/ubuntu-advantage.log` with sudo
+        Then stdout matches regexp:
+        """
+        cloud-init userdata has ubuntu-advantage key.
+        """
+        And stdout matches regexp:
+        """
+        Skipping auto-attach and deferring to cloud-init to setup and configure auto-attach
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
+           | jammy   |
--- a/lib/auto_attach.py
+++ b/lib/auto_attach.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+
+"""
+Perform auto-attach operation
+
+On Ubuntu Pro machines, we try to perform an auto-attach operation
+on first boot. This happens through a systemd unit that triggers
+that executes this script on every boot. However, if we detect
+that cloud-init has user data related to ua, we cannot run
+auto-attach here, since cloud-init will drive this operation on
+their side.
+"""
+import logging
+
+import yaml
+
+from uaclient.cli import action_auto_attach, setup_logging
+from uaclient.config import UAConfig
+from uaclient.exceptions import (
+    AlreadyAttachedOnPROError,
+    ProcessExecutionError,
+)
+from uaclient.util import subp, which
+
+
+def check_cloudinit_userdata_for_ua_info():
+    if not which("cloud-init"):
+        return False
+
+    try:
+        userdata, _ = subp(["cloud-init", "query", "userdata"])
+    except ProcessExecutionError:
+        # if we cannot query the userdata, we should not block auto-attach
+        return False
+
+    try:
+        userdata_dict = yaml.safe_load(userdata)
+    except Exception:
+        # if there is any error parsing the userdata, we should not block
+        # auto-attach
+        return False
+
+    if userdata_dict and "ubuntu_advantage" in userdata_dict.keys():
+        return True
+
+    return False
+
+
+def main(cfg: UAConfig):
+    if not check_cloudinit_userdata_for_ua_info():
+        # Once we have the api functions ready, we should
+        # update this part of the code to not call the cli
+        # function directly
+        try:
+            action_auto_attach(args=None, cfg=cfg)
+        except AlreadyAttachedOnPROError as e:
+            logging.info(e.msg)
+    else:
+        auto_attach_msg = (
+            "Skipping auto-attach and deferring to cloud-init "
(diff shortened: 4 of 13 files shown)


lib/auto_attach.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if userdata_dict and userdata_dict.get("ubuntu_advantage"):
+        return True
+
+    return False
+
+
+def main(cfg: UAConfig):
+    if not check_cloudinit_userdata_for_ua_info():
+        # Once we have the api functions ready, we shoukd
```
aciba90, line 51:
```suggestion
        # Once we have the api functions ready, we should
```

lib/auto_attach.py
```diff
@@ -0,0 +1,71 @@
...
+        # update this part of the code to not call the cli
+        # function directly
+        try:
+            action_auto_attach(args=None, cfg=cfg)
+        except AlreadyAttachedOnPROError as e:
+            print(e.msg)
```
aciba90, line 57: Should we log the error instead of print it?

```suggestion
            logging.error(e.msg)
```

uaclient/tests/test_auto_attach.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_check_cloudinit_data(self, m_which, m_subp, expected, userdata):
+        m_which.return_value = True
+        m_subp.return_value = (userdata, "")
+
+        assert expected is check_cloudinit_userdata_for_ua_info()
+
+
+class TestMain:
+    @pytest.mark.parametrize(
+        "ubuntu_advantage_in_userdata",
```
aciba90, line 82: Nit: Not sure if you use this layout to make it extensible in the future. But, this could be simplified to:

```py
@pytest.mark.parametrize("ubuntu_advantage_in_userdata", [True, False])
```

aciba90 (commented):
LGTM! Thanks for implementing this. I left some minor comments.

features/environment.py
```diff
@@ -640,7 +640,8 @@ def create_instance_with_uat_installed(
         )
         return
 
-    user_data = _get_user_data_for_instance(context.config, series)
+    if not user_data:
+        user_data = _get_user_data_for_instance(context.config, series)
```
orndorffgrant, line 644: could this interfere with how we're installing `pro` from -proposed or the daily ppa?

lib/auto_attach.py
```diff
@@ -0,0 +1,71 @@
...
+)
+from uaclient.util import subp, which
+
+
+def check_cloudinit_userdata_for_ua_info():
+    if not which("cloud-init"):
+        return False
+
+    try:
+        userdata, _ = subp(["cloud-init", "query", "userdata"])
```
orndorffgrant, line 31: I'm not that familiar with the cloud-init query feature. But I thought we'd be able to do something like:
```suggestion
        userdata, _ = subp(["cloud-init", "query", "userdata.ubuntu_advantage"])
```
Where that only returns data if there is anything set in the ubuntu_advantage module in userdata. This would let us skip parsing the yaml ourselves.

Is something like that possible @blackboxsw ?

uaclient/tests/test_auto_attach.py
```diff
@@ -0,0 +1,132 @@
+import contextlib
```
orndorffgrant, line 1: nitpick, can we name the file `test_lib_auto_attach.py` to make it more clear that this is testing the file in the lib directory.

lib/auto_attach.py
lucasmoura: The rationale for this is that the old systemd unit was running the cli command directly, which prints that message when already attached. But I think having a `logging.info` will be better here actually

features/environment.py
lucasmoura: I don't think so, because this `user_data` in only being passed by the step that creates the machine. Only on the new integration test, this variable will contain something that is not `None`. But let me know if I am missing something here

orndorffgrant: Right, but will we be able to run that new integration test against -proposed if the user_data is overwritten?

lib/auto_attach.py
blackboxsw: @orndorffgrant currently `cloud-init query userdata` only returns the raw userdata which could be `#cloud-config` YAML,  but it could also be a `#include-url https://someurl-which-has-cloud-config-stuff` or it could even be a binary blob that cloud-init doesn't react to. So, the expectation of userdata being a structured object that we can dive into can be a problem in some cases.

What I had originally thought was that we could use `cloud-init query merged_cfg.ubuntu_advantage` because the merged config is guaranteed to be rendered based on merged sources. But, I neglected to recall that `merged_cfg` is only merged config from 'system config' which is the aggregated files in /etc/cloud/cloud.cfg and /etc/cloud/cloud.cfg.d/*.
So, merged_config only gives you part of the puzzle. You'll need to query both userdata and merged_cfg from the filesystem somehow.

Long-term I think we want cloud-init to make `merged_cfg` reflect not only /etc/cloud/cloud.cfg* but also the merged userdata and vendordata which affects the configuration of the VM. But, this feature may require a bit of work outside of this cycle because userdata and vendordata in some cases are processed and rendered in the "init network" boot stage instead of init-local when the datasource is initially detected.

At the moment, the best solution to query both userdata and merged_cfg from disk is one of 3 options:
 1. Grab it the merged config direct from the fully realized init config which will merge userdata/vendordata into filesystem config for you
 ```python
from cloudinit.stages import Init
init = Init()
if init.cfg and "ubuntu_advantage" in init.cfg:
    skip_autoattach()
```
 2. Shell out to all instance data and YAML load to check for UA directives:
 ```python
instancedata_yaml, _ = subp(["cloud-init", "query", "--all"])
instance_data = yaml.safe_load(instancedata_yaml)
merged_cfg = instance_data["merged_cfg"]
userdata = yaml.safe_load(instance_data["userdata"])
if "ubuntu_advantage" not in merged_cfg and "ubuntu_advantage" not in userdata:
   skip_autoattach()
```
3. Shell out to all instance data and 'grep' for ubuntu_advantage directives:
````python
instancedata_yaml, _ = subp(["cloud-init", "query", "--all"])
if re.search("ubuntu_advantage:") in instancedata_yaml:
    skip_autoattach()
```

lucasmoura: Thanks for all the input here @blackboxsw. I like the first idea more to avoid unnecessary shell out. However, `cloud-init` is not one our dependencies at the moment. If we manage to directly import it in the code, should we add it as a dependency ? And I am saying that only for the pro package

orndorffgrant: There is a general desire to move away from having the -pro package at all so I'd rather not add more to it, if we can avoid it.

We can catch an ImportError if the module doesn't exist.

Is there any chance the interface of `cloudinit.stages.Init().cfg` will change in the future though? If it does, it would then break our integration if we rely on it directly.

renanrodrigo (commented):
My main suggestion is, given that we now have the step to define *additional* user data, instead of completely overriding it, we get rid of the `userdata_file` config variable - I think it can be safely removed here

lib/auto_attach.py
blackboxsw: @orndorffgrant `cloudinit.stages.Init().cfg` shouldn't change significantly, and hasn' in the last 6 years. 

cloud-init will keep Init().cfg stable until we can provide UA and others with a succinct representation of all merged_cfg data direct from `/run/cloud-init/instance-data-sensitive.json`. We want the flat files to be our primary API in case we need to change init logic in the future.


Ideally we'd like to steer consumers to read static `/run/cloud-init/instance-data-sensitive.json` `merged_cfg` key instead of re-running the logic that Init().cfg does to remerge all instance data into `.cfg`. We are already [shifting subquity live installer](https://github.com/canonical/subiquity/pull/1352) away from Init().cfg for that reason.

That said, currently /run/cloud-init/instance-data-sensitive.json merged_cfg doesn't have user-data present in merged_cfg and cloud-init will sort publishing user-data/vendor-data in `merged_cfg` before we steer UA away from `Init().cfg`.


@lucasmoura: as discussed separately. There is a potential race in how ua-auto-attach is triggered based on /run/cloud-init/instance-data.json. This file is written before user-data is processed so a race exists where init-network detected datasources (Xenial GCP) will not have rendered the user-data yet and `Init().cfg` may not see it yet. To resolve this to ensure ua-auto-attach.service can't start until user-data is completely visible, we should add an `After=cloud-config.target` to ua-auto-attach.service. I think we can then drop the `ua-auto-attach.path` directives as this will ensure the any applicable instance data to support both `cloud-id` calls and user-data queries via `Init().cfg` is present.

lib/auto_attach.py
```diff
@@ -0,0 +1,69 @@
...
+import logging
+import sys
+
+from uaclient.cli import action_auto_attach, setup_logging
+from uaclient.config import UAConfig
+from uaclient.exceptions import AlreadyAttachedOnPROError
+
+try:
+    import cloudinit.stages as ci_stages  # type: ignore
+except ImportError:
```
blackboxsw, line 22: Is the intent here to support images which may not contain cloud-init? maybe some container builds or maybe desktop images? `cloudinit.stages` should always exist on most server images everywhere, and any desktop live install images. But there is the case for desktop not-live-installer images that maybe this import handling is for. If that's the case, can we add a comment here about what image types this import handling is being used for.

blackboxsw: I guess additionally do we have a use case for desktop images or containers running auto-attach where we'd be concerned about even executing this code on those image types?

lib/auto_attach.py
lucasmoura: Awesome, I have updated the code to address this race condition as discussed. Thanks for all of the investigation you performed for this PR @blackboxsw

lib/auto_attach.py
lucasmoura: Yes, the general idea here is to avoid potential issue for any environment that lacks cloud-init and have the pro package installed, even by accident. However, I don't know a real use case where that happens

lucasmoura:
@blackboxsw @renanrodrigo @orndorffgrant @CalvoM @aciba90 I have performed the following benchmark checks for all the clouds we support to see the average time we take to make a request to get the attach token:

AWS (instance: t3.micro)
--------------
```python
import time

from uaclient.util import readurl
from uaclient.clouds.aws import IMDS_V2_TOKEN_URL, IMDS_IPV4_ADDRESS, AWS_TOKEN_REQ_HEADER, AWS_TOKEN_TTL_SECONDS

exec_time = []
for i in range(0, 10):
    start_time = time.time()
    content = readurl(
        IMDS_V2_TOKEN_URL.format(IMDS_IPV4_ADDRESS),
        method="PUT",
        headers={AWS_TOKEN_REQ_HEADER: AWS_TOKEN_TTL_SECONDS}
    )
    final_time = time.time() - start_time
    exec_time.append(final_time)

# Mean: 0.0013877391815185548
print("Mean: {}".format(sum(exec_time) / len(exec_time)))
```

Azure (instance: Standard_DS1_v2)
--------------
```python
import time

from uaclient.util import readurl
from uaclient.clouds.azure import IMDS_URLS

exec_time = []
for i in range(0, 10):
    start_time = time.time()
    content = readurl(
        IMDS_URLS["pkcs7"],
        headers={"Metadata": "true"},
    )
    final_time = time.time() - start_time
    exec_time.append(final_time)

# Mean: 0.053901076316833496 
print("Mean: {}".format(sum(exec_time) / len(exec_time)))
```

GCP (instance: n1-standard-1)
--------------
```python
import time

from uaclient.util import readurl
from uaclient.clouds.gcp import LICENSES_URL

exec_time = []
for i in range(0, 10):
    start_time = time.time()
    content = readurl(
        LICENSES_URL,
        headers={"Metadata-Flavor": "Google"}, 
    )
    final_time = time.time() - start_time
    exec_time.append(final_time)

# Mean: 0.0030882596969604493
print("Mean: {}".format(sum(exec_time) / len(exec_time)))
```

Here is the average time I have obtained: 

* aws: 0.0013877391815185548
* azure: 0.053901076316833496 
* gcp: 0.0030882596969604493

Note that only `Azure` takes a little more time. Based on that, I have modified the timeouts we have for our clouds during auto-attach. Let me know if you think I can reduce them even more

lib/auto_attach.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python3
+
+"""
+Perform auto-attach operation
+
+On Ubuntu Pro machines, we try to perform an auto-attach operation
+on first boot. This happens through a systemd unit that triggers
+that executes this script on every boot. However, if we detect
+that cloud-init has user data related to ua, we cannot run
```
orndorffgrant, line 9: grammar nits
```suggestion
on first boot. This happens through a systemd unit
that executes this script on every boot. However, if we detect
that cloud-init has user data related to ua, we don't run
```

lib/auto_attach.py
```diff
@@ -0,0 +1,69 @@
...
+    import cloudinit.stages as ci_stages  # type: ignore
+except ImportError:
+    pass
+
+
+def get_cloudinit_init_stage():
+    if "cloudinit.stages" in sys.modules:
+        return ci_stages.Init()
+
+    return None
```
orndorffgrant, line 30: instead of the check in sys.modules, would it be better or worse to do a runtime import?
```suggestion
def get_cloudinit_init_stage():
    try:
        import cloudinit.stages as ci_stages  # type: ignore
        return ci_stages.Init()
    except ImportError:
        return None
```

systemd/ua-auto-attach.service
```diff
@@ -1,10 +1,10 @@
 [Unit]
 Description=Ubuntu Advantage auto attach
-Before=cloud-config.service
+After=cloud-config.service
```
orndorffgrant, line 3: I think we want
```suggestion
After=cloud-config.target
Before=cloud-config.service
```

See https://github.com/canonical/cloud-init/blob/main/systemd/cloud-config.target

orndorffgrant (commented):
Looking good! I see that you've modified the retry sleep times for the metadata endpoint calls, but azure and gcp are missing specific request timeouts in their `util.readurl` calls

lib/auto_attach.py
lucasmoura: Yeah, that was my original idea. The problem is that if we import cloudinit like this we end up messing our logging, because this cloudinit module seems to override our setup. But maybe I missed another thing we could here to not mess up the logs

systemd/ua-auto-attach.service
lucasmoura: Indeed, good catch

lib/auto_attach.py
orndorffgrant: Oh the import is modifying root logger settings? Then the current implementation makes sense

lucasmoura:
@orndorffgrant sorry, about the `util.readurl` timeout, you mean the ones on the `cloud` specific modules ?

orndorffgrant:
> @orndorffgrant sorry, about the `util.readurl` timeout, you mean the ones on the `cloud` specific modules ?

@lucasmoura right, for example I was expecting a `timeout=1` in the `readurl` call at line 30 of azure.py - unless I'm missing somewhere else where the timeout is specified?

lucasmoura:
@orndorffgrant Good catch, I have added the missing timeouts

lib/auto_attach.py
orndorffgrant: I don't think we currently have a use-case for non-cloud images. However, cloud-init is not (and cannot be) an explicit dependency, and so I think we should be defensive when relying on it.

systemd/ua-auto-attach.service
```diff
@@ -1,10 +1,10 @@
 [Unit]
 Description=Ubuntu Advantage auto attach
-Before=cloud-config.service
+After=cloud-config.target
```
orndorffgrant, line 3: We need to keep the before line to ensure cloud-init waits for this service to finish before it runs its `ubuntu_advantage` module or `package_upgrade` module etc
```suggestion
After=cloud-config.target
Before=cloud-config.service
```

features/ubuntu_pro.feature
```diff
@@ -580,3 +580,84 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
...
+        # through the ua-auto-attach.service, we are masking it
+        When I run `systemctl mask ubuntu-advantage.service` with sudo
+        When I run `systemctl start ua-auto-attach.service` with sudo
+        And I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
+        """
+        contract_url: 'https://contracts.canonical.com'
+        data_dir: /var/lib/ubuntu-advantage
+        log_level: debug
+        log_file: /var/log/ubuntu-advantage.log
+        """
```
orndorffgrant, line 640: Shouldn't we need to fix the config file before running the auto-attach service? Right now, `disable_auto_attach` is set when the service is started
```suggestion
        And I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
        """
        contract_url: 'https://contracts.canonical.com'
        data_dir: /var/lib/ubuntu-advantage
        log_level: debug
        log_file: /var/log/ubuntu-advantage.log
        """
        When I run `systemctl start ua-auto-attach.service` with sudo
```

features/ubuntu_pro.feature
```diff
@@ -580,3 +580,84 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
...
+        And I reboot the `<release>` machine
+        And I run `pro status --wait` with sudo
+        And I run `pro security-status --format json` with sudo
+        Then stdout matches regexp:
+        """
+        "attached": false
+        """
```
orndorffgrant, line 647: I think we can just use the status output to verify attached is false
```suggestion
        And I run `pro status --wait --format yaml` with sudo
        Then stdout matches regexp:
        """
        attached: false
        """
```

features/ubuntu_pro.feature
lucasmoura: My idea was to test only the script only during reboot, so when we start the service again the feature flag will block it and only after reboot it should kick off correctly and perform the attach operation

features/ubuntu_pro.feature
lucasmoura: We could, but I would like us to avoid using status for that as we discussed for the status refactoring of the code. Right now, the status command here is just to ensure that there is no background service running. But let me know what you think

features/ubuntu_pro.feature
orndorffgrant: Ah okay that makes sense! But then I don't think we need the `When I run `systemctl start ua-auto-attach.service` with sudo` line right? I don't think it should do anything

features/ubuntu_pro.feature
orndorffgrant: I'm not sure using `pro security-status --format json` is much better than using `pro status --wait --format yaml` but not a strong opinion :) this is good

features/ubuntu_pro.feature
lucasmoura: I have added because we stop that service using our custom cloud-init userdata. This is just to guarantee that the reboot will behave as we intend

features/steps/steps.py
```diff
@@ -74,13 +74,13 @@ def given_a_machine(context, series):
 
     instance_name = add_test_name_suffix(context, series, CONTAINER_PREFIX)
 
-    if context.config.snapshot_strategy:
-        if series not in context.series_image_name:
+    if context.config.snapshot_strategy and not custom_user_data:
+        if series not in context.series_image_name or custom_user_data:
```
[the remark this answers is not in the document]
lucasmoura: indeed, good catch. I will remove the second check

lucasmoura:
@renanrodrigo done

features/ubuntu_pro.feature
orndorffgrant: I see and understand now!

I think the `stop` won't affect the next boot though. I'd expect a `disable` or a `mask` to affect next boot, but I'd be surprised if a `stop` did. Did you try the test without that `start` line? (Sorry if you already tested this and I'm just making noise :upside_down_face:)

lucasmoura: Good point @orndorffgrant, I have added just for safety sake, but you are right, we don't need it.
I have removed that line from the test

orndorffgrant (approved):
LGTM thank you!

renanrodrigo (approved):

Issue #913, opened 2019-11-09
ua status doesn't update availability after booting into/out of supported kernel


blackboxsw:
Once #912 lands, we need to close a gap in functionality.

On a trusty vm with stock kernel:
```
sudo ua attach
sudo apt-get install --install-recommends linux-generic-lts-xenial  
sudo reboot             # boot into HWE kernel
;   # machine-token not yet refreshed, so any ua status will still report:
;  #  livepatch     yes       n/a       Available with the HWE kernel
sudo ua status
```

orndorffgrant:
This will be addressed as part of SC-1008

PR #2167, opened 2022-07-21
display livepatch support

orndorffgrant:
## New Notes

Fixes: #913

TODO
- [x] move cache file definition to state_files.py

## Notes

While starting this I've noticed some features to break out into separate PRs and merge beforehand. They are:

- [x] #2168 move system query functions from `util.py` to a new `system.py` because I'll need to add more here and util is already getting large. This split will help organize things
- [x] #2184 refactor how we get and parse the kernel version. livepatch uses `/proc/version_signature` and needs more granular version information (specifically, the hwe revision). We need to parse the version in the same way to properly query for kernel support.
- [x] #2189 add a new function to get the arch from `lscpu`. This is the value that livepatch uses and could be slightly different from what dpkg reports.
- [x] #2169 `DateDataValue` for parsing date strings into DataObjects
- [x] #2172 `DataObjectFile`. For the cache, it will be useful to have an file object that automatically serializes/parses content to/from a DataObject. I'll make this reuse-able.

From there I think I'll have the important building blocks for this feature and the rest will probably be in this PR.

## Test Steps
- run the new integration test
- start a vm and try out the UX for an unsupported vs supported kernel when attached/unattached/enabled/disabled
- check the logs to verify the api/cache/cli are used for support information appropriately (search for "livepatch.py")

## Checklist


uaclient/entitlements/livepatch.py
```diff
@@ -26,9 +38,235 @@
...
+            if cli_supported == "unsupported":
+                return False
+
+            cli_state = cli_status.get("State", None)
+            if cli_state == "applied":
+                logging.debug(
+                    "kernel must be supported because livepatches "
+                    " have been applied"
+                )
+                return True
```
orndorffgrant, line 155: We actually may not be able to assume this means the kernel is supported. For example, if a kernel is supported for a while but then we end support, then we need to warn the user that their support has ended even if the kernel has patches currently applied.

renanrodrigo (approved):
this lgtm

orndorffgrant:
The cloud test failures indicate there may be a gap in the support database

renanrodrigo:
@orndorffgrant if we confirm that the cloud kernels are now right in the database, our tests are looking good and I guess we can click the magic button here

orndorffgrant:
Cloud kernel support is updated in the database and just rebased and resolved conflicts. Let's let CI run

renanrodrigo:
MERGE, MERGE EVERYTHING

PR #2168, opened 2022-07-22
refactor: move system query functions to their own module

orndorffgrant:
## Test Steps

Everything should still work. run any integration tests you think are relevant in addition to those in CI

## Desired commit type

## Checklist


uaclient/system.py
```diff
@@ -0,0 +1,213 @@
...
+DBUS_MACHINE_ID = "/var/lib/dbus/machine-id"
+
+# N.B. this relies on the version normalisation we perform in get_platform_info
+REGEX_OS_RELEASE_VERSION = r"(?P<release>\d+\.\d+) (LTS )?\((?P<series>\w+).*"
+
+
+@lru_cache(maxsize=None)
+def get_machine_id(cfg) -> str:
+    """Get system's unique machine-id or create our own in data_dir."""
+    # Generate, cache our own uuid if not present in config or on the system
```
lucasmoura, line 21: I think the phrasing of the second sentence is confusing. I prefer the one we already have in the docstring.  We can improve the docstring by just adding that we can retrieve the `machine_id` from the `machine_token` file.

lucasmoura (commented):
I like the idea of the separation, specially removing the contract override definitions from `util`. However, I think most of the leftover functions could be still related to `system` needs, like `is_exe` and `write_file`. Maybe we could do the other way, rename `util` to `system` and extract the functions that we don't think apply there any longer. For example, the message related functions like prompts could be moved to the `message` module

orndorffgrant:
Thanks @lucasmoura! I brought in the subp, is_exe, and file reading/writing functions into the new system module.

I don't think the rest of the util module fits into system, and while some of the pieces there might make more sense in other modules, some other pieces in util don't really fit anywhere else (e.g. `parse_rfc339_date`). So I'd rather not increase the scope of this PR to reorganize all of util, and just stay focused on the separation of the "system" functions.

I only moved the `apply_contract_overrides` function to `contract.py` to avoid a circular dependency.

uaclient/system.py
orndorffgrant: Good point! I only copy pasted this, but I added another commit here to make the adjustments you suggest.

renanrodrigo (approved):
This works for me.
I have the impression that the less "utils" we get the better, and that specific functions (like the contracts one) can and should live in their respective modules. +1 to the messages related ones, as Lucas said. Let's make sure to do this in the next PRs

lucasmoura (approved):

PR #2169, opened 2022-07-22
data_types: datetime data value

orndorffgrant:
## Test Steps

Run new unit tests

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/data_types.py
+++ b/uaclient/data_types.py
@@ -1,8 +1,9 @@
+import datetime
 import json
 from enum import Enum
 from typing import Any, List, Optional, Type, TypeVar, Union
 
-from uaclient import exceptions
+from uaclient import exceptions, util
 
 INCORRECT_TYPE_ERROR_MESSAGE = (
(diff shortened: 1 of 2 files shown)


uaclient/data_types.py
```diff
@@ -122,6 +123,21 @@ def from_value(val: Any) -> bool:
         return val
 
 
+class DatetimeDataValue(DataValue):
+    """
+    This expects that value is a datetime.
+    Use util.DatetimeAwareJSONDecoder when parsing json
```
lucasmoura, line 129: Sorry @orndorffgrant, I think I am missing the point of this part of the docstring.

Should we also update the `DataObject` `to_json` method to use the `DatetimeAwareJSONDecoder` class ?

orndorffgrant: Thanks @lucasmoura. I added the DatetimeAwareJSONEncoder to the `to_json` method and corresponding tests. And I removed the confusing line in the docstring - it didn't really add anything anyway

renanrodrigo (approved):
lgtm

PR #2170, opened 2022-07-22
api: adjustments to not require config as an arg

orndorffgrant:
## Test Steps

make sure the python script experience of using our api doesn't requiring touching any ua-internals like the UAConfig

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/api/u/pro/attach/magic/initiate/v1.py
+++ b/uaclient/api/u/pro/attach/magic/initiate/v1.py
@@ -12,7 +12,6 @@ from uaclient.data_types import (
 
 class MagicAttachInitiateResult(DataObject, AdditionalInfo):
     fields = [
-        Field("_schema", StringDataValue),
         Field("user_code", StringDataValue),
         Field("token", StringDataValue),
         Field("expires", StringDataValue),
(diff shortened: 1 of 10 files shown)


orndorffgrant:
If we agree on this general approach, then I'll fix the various test failures

uaclient/api/u/pro/attach/magic/initiate/v1.py
```diff
@@ -21,25 +22,29 @@ class MagicAttachInitiateResult(DataObject, AdditionalInfo):
...
 
 
-def initiate(cfg: UAConfig) -> MagicAttachInitiateResult:
+def initiate() -> MagicAttachInitiateResult:
+    return _initiate()
+
+
+def _initiate(*, cfg: Optional[UAConfig] = None) -> MagicAttachInitiateResult:
+    if cfg is None:
+        cfg = UAConfig()
```
lucasmoura, line 42: I like this approach that covers both sides of using the API. I have no problems going forward with this refactoring

renanrodrigo (approved):
Nice!

CalvoM (approved):
LGTM, check the failing integration test.

orndorffgrant:
Yep I made a silly mistake while editing that test. Should be fixed now but let's wait for CI to go green

PR #2171, opened 2022-07-25
auto_attach: add short retry loop api endpoint

CalvoM:
## Proposed Commit Message

auto_attach: add short retry loop api endpoint

We are adding an api endpoint to perform auto-attach with short retries incase of failures;
this is before the process of longer retries.

## Desired commit type

## Checklist

### DIFF
--- a/features/api_full_auto_attach.feature
+++ b/features/api_full_auto_attach.feature
@@ -1,0 +1,39 @@
+Feature: Full Auto-Attach Endpoint
+
+    @series.lts
+    @uses.config.machine_type.aws.pro
+    @uses.config.machine_type.azure.pro
+    @uses.config.machine_type.gcp.pro
+    Scenario Outline: Run auto-attach on cloud instance.
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
+        """
+        contract_url: 'https://contracts.canonical.com'
+        data_dir: /var/lib/ubuntu-advantage
+        log_level: debug
+        log_file: /var/log/ubuntu-advantage.log
+        """
+        When I create the file `/tmp/full_auto_attach.py` with the following:
+        """
+        from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import full_auto_attach, FullAutoAttachOptions
+
+        full_auto_attach(FullAutoAttachOptions(enable=["esm-infra", "esm-apps"]))
+        """
+        And I run `python3 /tmp/full_auto_attach.py` with sudo
+        And I run `pro status --all` with sudo
+        Then stdout matches regexp:
+        """
+        esm-apps      +yes +enabled +Extended Security Maintenance for Applications
+        esm-infra     +yes +enabled +Extended Security Maintenance for Infrastructure
+        """
+        Then stdout matches regexp:
+        """
+        livepatch     +yes +disabled  +Canonical Livepatch service
+        """
+        Examples:
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
+           | jammy   |
+           
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -14,6 +14,7 @@ from uaclient import (
 )
 from uaclient import status as ua_status
 from uaclient import system, util
+from uaclient.clouds import AutoAttachCloudInstance  # noqa: F401
 from uaclient.clouds import identity
 from uaclient.defaults import (
     CLOUD_BUILD_INFO,
@@ -71,7 +72,9 @@ def attach_with_token(
 
 
 def auto_attach(
-    cfg: config.UAConfig, cloud: clouds.AutoAttachCloudInstance
+    cfg: config.UAConfig,
+    cloud: clouds.AutoAttachCloudInstance,
+    allow_enable=True,
 ) -> None:
     """
     :raise UrlError: On unexpected connectivity issues to contract
@@ -95,7 +98,7 @@ def auto_attach(
 
     token = tokenResponse["contractToken"]
 
-    attach_with_token(cfg, token=token, allow_enable=True)
+    attach_with_token(cfg, token=token, allow_enable=allow_enable)
 
 
 def enable_entitlement_by_name(
@@ -220,3 +223,52 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
             system.write_file(
                 os.path.join(output_dir, os.path.basename(f)), content
             )
+
+
+def should_disable_auto_attach(cfg: config.UAConfig) -> bool:
+    disable_auto_attach = util.is_config_value_true(
+        config=cfg.cfg, path_to_value="features.disable_auto_attach"
+    )
+    if disable_auto_attach:
+        msg = "Skipping auto-attach. Config disable_auto_attach is set."
+        logging.debug(msg)
+        print(msg)
+
+    return disable_auto_attach
+
+
+def get_cloud_instance(
+    cfg: config.UAConfig,
+) -> AutoAttachCloudInstance:
+    instance = None  # type: Optional[AutoAttachCloudInstance]
+    try:
+        instance = identity.cloud_instance_factory()
+    except exceptions.CloudFactoryError as e:
+        if cfg.is_attached:
+            # We are attached on non-Pro Image, just report already attached
+            raise exceptions.AlreadyAttachedError(cfg)
+        if isinstance(e, exceptions.CloudFactoryNoCloudError):
+            raise exceptions.UserFacingError(
+                messages.UNABLE_TO_DETERMINE_CLOUD_TYPE,
+                msg_code="auto-attach-cloud-type-error",
+            )
+        if isinstance(e, exceptions.CloudFactoryNonViableCloudError):
+            raise exceptions.UserFacingError(messages.UNSUPPORTED_AUTO_ATTACH)
+        if isinstance(e, exceptions.CloudFactoryUnsupportedCloudError):
+            raise exceptions.NonAutoAttachImageError(
+                messages.UNSUPPORTED_AUTO_ATTACH_CLOUD_TYPE.format(
+                    cloud_type=e.cloud_type
+                ),
+                msg_code="auto-attach-unsupported-cloud-type-error",
+            )
+        # we shouldn't get here, but this is a reasonable default just in case
+        raise exceptions.UserFacingError(
+            messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
+        )
+
+    if not instance:
+        # we shouldn't get here, but this is a reasonable default just in case
+        raise exceptions.UserFacingError(
+            messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
+        )
+    return instance
--- a/uaclient/api/api.py
+++ b/uaclient/api/api.py
@@ -21,6 +21,7 @@ VALID_ENDPOINTS = [
     "u.pro.attach.magic.wait.v1",
     "u.pro.attach.magic.revoke.v1",
     "u.pro.attach.auto.should_auto_attach.v1",
+    "u.pro.attach.auto.full_auto_attach.v1",
 ]
 
 
--- a/uaclient/api/exceptions.py
+++ b/uaclient/api/exceptions.py
@@ -1,0 +1,40 @@
+from uaclient import messages
+from uaclient.exceptions import (
+    AlreadyAttachedError,
+    BetaServiceError,
+    ContractAPIError,
+    EntitlementNotFoundError,
+    NonAutoAttachImageError,
+    UrlError,
+    UserFacingError,
+)
+
+__all__ = [
+    "AlreadyAttachedError",
+    "BetaServiceError",
+    "ContractAPIError",
+    "EntitlementNotFoundError",
+    "NonAutoAttachImageError",
+    "UrlError",
+    "UserFacingError",
+]
+
+
+class EntitlementNotEnabledError(UserFacingError):
+    """An exception raised when enabling of an entitlement fails"""
+
+    pass
+
+
+class FullAutoAttachFailureError(UserFacingError):
+    """An exception raised when auto attach at boot fails"""
+
+    def __init__(self):
+        super().__init__(
+            msg=messages.FULL_AUTO_ATTACH_ERROR.msg,
+            msg_code=messages.FULL_AUTO_ATTACH_ERROR.name,
+        )
+
+
+class IncompatibleEntitlementsDetected(UserFacingError):
+    pass
--- a/uaclient/api/tests/test_u_pro_attach_auto_full_auto_attach_v1.py
+++ b/uaclient/api/tests/test_u_pro_attach_auto_full_auto_attach_v1.py
@@ -1,0 +1,161 @@
+import mock
+import pytest
+
+from uaclient.api import exceptions
+from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
+    FullAutoAttachOptions,
+    _full_auto_attach,
+    _is_incompatible_services_present,
+)
+from uaclient.entitlements.entitlement_status import (
+    CanEnableFailure,
+    CanEnableFailureReason,
+)
+from uaclient.messages import NamedMessage
+
+M_API = "uaclient.api.u.pro."
+
+
+class TestFullAutoAttachV1:
+    @mock.patch(
+        "uaclient.actions.enable_entitlement_by_name",
+        return_value=(True, None),
+    )
+    def test_error_when_beta_in_enable_list(
+        self,
+        enable_ent_by_name,
+        FakeConfig,
+    ):
+        cfg = FakeConfig(root_mode=True)
+        options = FullAutoAttachOptions(
+            enable=["esm-infra", "realtime-kernel"]
+        )
+        with pytest.raises(exceptions.BetaServiceError):
+            _full_auto_attach(options, cfg)
+
+        assert 0 == enable_ent_by_name.call_count
+
+    @mock.patch(
+        "uaclient.actions.enable_entitlement_by_name",
+        return_value=(True, None),
+    )
+    @mock.patch("uaclient.actions.get_cloud_instance")
+    @mock.patch("uaclient.actions.auto_attach")
+    def test_error_invalid_ent_names(
+        self,
+        _auto_attach,
+        _get_cloud_instance,
+        enable_ent_by_name,
+        FakeConfig,
+    ):
+        cfg = FakeConfig(root_mode=True)
+        options = FullAutoAttachOptions(
+            enable=["esm-infra", "esm-apps", "cis"],
+            enable_beta=["realtime-kernel", "test", "wrong"],
+        )
+        with pytest.raises(exceptions.EntitlementNotFoundError):
+            _full_auto_attach(options, cfg)
+
+        assert 0 == enable_ent_by_name.call_count
+
+    @mock.patch(
+        "uaclient.actions.enable_entitlement_by_name",
+        return_value=(
+            False,
+            CanEnableFailure(
+                CanEnableFailureReason.ALREADY_ENABLED,
+                NamedMessage("test", "test"),
+            ),
+        ),
+    )
+    @mock.patch("uaclient.actions.get_cloud_instance")
+    @mock.patch("uaclient.actions.auto_attach")
+    def test_error_ent_not_enabled(
+        self,
+        _auto_attach,
+        _get_cloud_instance,
+        enable_ent_by_name,
+        FakeConfig,
+    ):
+        cfg = FakeConfig(root_mode=True)
+        options = FullAutoAttachOptions(
+            enable=["esm-infra", "esm-apps", "cis"],
+            enable_beta=["realtime-kernel"],
+        )
+        with pytest.raises(exceptions.EntitlementNotEnabledError):
+            _full_auto_attach(options, cfg)
+
+        assert 1 == enable_ent_by_name.call_count
+
+    @mock.patch(
+        "uaclient.actions.enable_entitlement_by_name",
+        return_value=(False, None),
+    )
+    @mock.patch("uaclient.actions.get_cloud_instance")
+    @mock.patch("uaclient.actions.auto_attach")
+    def test_error_full_auto_attach_fail(
+        self,
+        _auto_attach,
+        _get_cloud_instance,
+        enable_ent_by_name,
+        FakeConfig,
+    ):
+        cfg = FakeConfig(root_mode=True)
+        options = FullAutoAttachOptions(
+            enable=["esm-infra", "esm-apps", "fips"],
+            enable_beta=["ros"],
+        )
+        with pytest.raises(exceptions.EntitlementNotEnabledError):
+            _full_auto_attach(options, cfg)
+
+        assert 1 == enable_ent_by_name.call_count
+
+    def test_error_incompatible_services(
+        self,
+        FakeConfig,
+    ):
+        cfg = FakeConfig(root_mode=True)
+        options = FullAutoAttachOptions(
+            enable=["esm-infra", "esm-apps", "fips"],
+            enable_beta=["livepatch"],
+        )
+        with pytest.raises(exceptions.IncompatibleEntitlementsDetected) as e:
+            _full_auto_attach(options, cfg)
+
+        expected_info = {
+            "service": "fips",
+            "incompatible_services": "livepatch,fips-updates,realtime-kernel",
+        }
+
+        assert expected_info == e.value.additional_info
+
+    @pytest.mark.parametrize(
+        "ent_list,service,incompatible_services,detected",
+        (
+            (
+                ["fips", "esm-infra", "fips-updates", "esm-apps"],
+                "fips",
+                ["livepatch", "fips-updates", "realtime-kernel"],
+                True,
+            ),
+            (
+                ["fips-updates", "esm-infra", "realtime-kernel", "esm-apps"],
+                "fips-updates",
+                ["fips", "realtime-kernel"],
+                True,
+            ),
+            (
+                ["livepatch", "esm-infra", "esm-apps"],
+                "",
+                [],
+                False,
+            ),
+        ),
+    )
+    def test_incompatible_services(
+        self, ent_list, service, incompatible_services, detected, FakeConfig
+    ):
+        res = _is_incompatible_services_present(FakeConfig(), ent_list)
+        assert detected == res[0]
+        assert service == res[1]
+        assert incompatible_services == res[2]
--- a/uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
+++ b/uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
@@ -1,0 +1,149 @@
+from typing import List, Optional, Tuple, Type  # noqa: F401
+
+from uaclient import actions, entitlements, event_logger
+from uaclient.api import exceptions
+from uaclient.api.api import APIEndpoint
+from uaclient.api.data_types import AdditionalInfo
+from uaclient.clouds import AutoAttachCloudInstance  # noqa: F401
+from uaclient.config import UAConfig
+from uaclient.data_types import DataObject, Field, StringDataValue, data_list
+from uaclient.entitlements.base import UAEntitlement  # noqa: F401
+from uaclient.entitlements.entitlement_status import CanEnableFailure
+
+event = event_logger.get_event_logger()
+
+
+class FullAutoAttachOptions(DataObject):
+    fields = [
+        Field("enable", data_list(StringDataValue), False),
+        Field("enable_beta", data_list(StringDataValue), False),
+    ]
+
+    def __init__(
+        self,
+        enable: Optional[List[str]] = None,
+        enable_beta: Optional[List[str]] = None,
+    ):
+        self.enable = enable
+        self.enable_beta = enable_beta
+
+
+class FullAutoAttachResult(DataObject, AdditionalInfo):
+    pass
+
+
+def _is_any_beta(cfg: UAConfig, ents: List[str]) -> Tuple[bool, str]:
+    for name in ents:
+        try:
+            ent_cls = entitlements.entitlement_factory(cfg, name)
+            if ent_cls.is_beta:
+                return True, name
+        except exceptions.EntitlementNotFoundError:
+            continue
+    return False, ""
+
+
+def _is_incompatible_services_present(
+    cfg, ent_list: List[str]
+) -> Tuple[bool, str, List[str]]:
+    ent_cls = list()  # type: List[Type[UAEntitlement]]
+    for e in ent_list:
+        ent_cls.append(entitlements.entitlement_factory(cfg, e))
+
+    for ent in ent_cls:
+        ent_inst = ent(cfg)
+        if ent_inst.incompatible_services:
+            incompat = [
+                service.entitlement
+                for service in ent_inst.incompatible_services
+            ]
+            if any(e in incompat for e in ent_cls):
+                return True, ent_inst.name, [e(cfg).name for e in incompat]
+    return False, "", []
+
+
+def full_auto_attach(options: FullAutoAttachOptions) -> FullAutoAttachResult:
+    return _full_auto_attach(options, UAConfig(root_mode=True))
+
+
+def _full_auto_attach(options: FullAutoAttachOptions, cfg: UAConfig):
+    event.set_event_mode(event_logger.EventLoggerMode.JSON)
+
+    services = set()
+    if options.enable:
+        is_beta_found, service = _is_any_beta(cfg, options.enable)
+        if is_beta_found:
+            raise exceptions.BetaServiceError(
+                msg="beta service found in the enable list",
+                msg_code="beta-service-found",
+                additional_info={"beta_service": service},
+            )
+        services.update(options.enable)
+    if options.enable_beta:
+        services.update(options.enable_beta)
+
+    service_list = list(services)
+
+    found, not_found = entitlements.get_valid_entitlement_names(
+        service_list, cfg
+    )
+    if not_found:
+        msg = entitlements.create_enable_entitlements_not_found_message(
+            not_found, cfg=cfg, allow_beta=True
+        )
+        raise exceptions.EntitlementNotFoundError(msg.msg, not_found)
+
+    incompat_detected, ent, incompat_ents = _is_incompatible_services_present(
+        cfg, sorted(found)
+    )  # sort for easy testing exceptions raised
+    if incompat_detected:
+        err_msg = "{ent} is incompatible with any of these services {incompat}"
+        raise exceptions.IncompatibleEntitlementsDetected(
+            msg=err_msg.format(ent=ent, incompat=incompat_ents),
+            msg_code="incompatible-services-detected",
+            additional_info={
+                "service": ent,
+                "incompatible_services": ",".join(incompat_ents),
+            },
+        )
+
+    instance = actions.get_cloud_instance(cfg)
+    enable_default_services = (
+        options.enable is None and options.enable_beta is None
+    )
+    actions.auto_attach(cfg, instance, enable_default_services)
+
+    if enable_default_services:
+        return FullAutoAttachResult()
+
+    for name in found:
+        ent_ret, reason = actions.enable_entitlement_by_name(
+            cfg, name, assume_yes=True, allow_beta=True
+        )
+        if not ent_ret:
+            if (
+                reason is not None
+                and isinstance(reason, CanEnableFailure)
+                and reason.message is not None
+            ):
+                raise exceptions.EntitlementNotEnabledError(
+                    msg=reason.message.msg,
+                    msg_code=reason.message.name,
+                    additional_info={"service": name},
+                )
+            else:
+                raise exceptions.EntitlementNotEnabledError(
+                    msg="Failed to enable service: {}".format(name),
+                    msg_code="entitlement-not-enabled",
+                    additional_info={"service": name},
+                )
+
+    return FullAutoAttachResult()
+
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="FullAutoAttach",
+    fn=_full_auto_attach,
+    options_cls=FullAutoAttachOptions,
+)
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -12,7 +12,7 @@ import tempfile
 import textwrap
 import time
 from functools import wraps
-from typing import List, Optional, Tuple  # noqa
+from typing import List, Tuple  # noqa
 
 import yaml
 
@@ -34,11 +34,13 @@ from uaclient import status as ua_status
 from uaclient import util, version
 from uaclient.api.api import call_api
 from uaclient.apt import AptProxyScope, setup_apt_proxy
-from uaclient.clouds import AutoAttachCloudInstance  # noqa: F401
-from uaclient.clouds import identity
 from uaclient.data_types import AttachActionsConfigFile, IncorrectTypeError
 from uaclient.defaults import DEFAULT_LOG_FORMAT, PRINT_WRAP_WIDTH
-from uaclient.entitlements import entitlements_disable_order
+from uaclient.entitlements import (
+    create_enable_entitlements_not_found_message,
+    entitlements_disable_order,
+    get_valid_entitlement_names,
+)
 from uaclient.entitlements.entitlement_status import (
     ApplicationStatus,
     CanDisableFailure,
@@ -795,25 +797,6 @@ def _perform_disable(entitlement, cfg, *, assume_yes, update_status=True):
     return ret
 
 
-def get_valid_entitlement_names(names: List[str], cfg: config.UAConfig):
-    """Return a list of valid entitlement names.
-
-    :param names: List of entitlements to validate
-    :return: a tuple of List containing the valid and invalid entitlements
-    """
-    entitlements_found = []
-
-    for ent_name in names:
-        if ent_name in entitlements.valid_services(
-            cfg=cfg, allow_beta=True, all_names=True
-        ):
-            entitlements_found.append(ent_name)
-
-    entitlements_not_found = sorted(set(names) - set(entitlements_found))
-
-    return entitlements_found, entitlements_not_found
-
-
 def action_config(args, *, cfg, **kwargs):
     """Perform the config action.
 
@@ -1112,33 +1095,6 @@ def action_disable(args, *, cfg, **kwargs):
     return 0 if ret else 1
 
 
-def _create_enable_entitlements_not_found_message(
-    entitlements_not_found, cfg: config.UAConfig, *, allow_beta: bool
-) -> messages.NamedMessage:
-    """
-    Constructs the MESSAGE_INVALID_SERVICE_OP_FAILURE message
-    based on the attempted services and valid services.
-    """
-    valid_services_names = entitlements.valid_services(
-        cfg=cfg, allow_beta=allow_beta
-    )
-    valid_names = ", ".join(valid_services_names)
-    service_msg = "\n".join(
-        textwrap.wrap(
-            "Try " + valid_names + ".",
-            width=80,
-            break_long_words=False,
-            break_on_hyphens=False,
-        )
-    )
-
-    return messages.INVALID_SERVICE_OP_FAILURE.format(
-        operation="enable",
-        invalid_service=", ".join(entitlements_not_found),
-        service_msg=service_msg,
-    )
-
-
 @verify_json_format_args
 @assert_root
 @assert_attached(_create_enable_disable_unattached_msg)
@@ -1198,7 +1154,7 @@ def action_enable(args, *, cfg, **kwargs):
             ret = False
 
     if entitlements_not_found:
-        msg = _create_enable_entitlements_not_found_message(
+        msg = create_enable_entitlements_not_found_message(
             entitlements_not_found, cfg=cfg, allow_beta=args.beta
         )
         event.services_failed(entitlements_not_found)
@@ -1305,44 +1261,11 @@ def action_auto_attach(args, *, cfg):
     if cfg.is_attached:
         raise exceptions.AlreadyAttachedOnPROError()
 
-    disable_auto_attach = util.is_config_value_true(
-        config=cfg.cfg, path_to_value="features.disable_auto_attach"
-    )
-    if disable_auto_attach:
-        msg = "Skipping auto-attach. Config disable_auto_attach is set."
-        logging.debug(msg)
-        print(msg)
+    skip_auto_attach = actions.should_disable_auto_attach(cfg)
+    if skip_auto_attach:
         return 0
 
-    instance = None  # type: Optional[AutoAttachCloudInstance]
-    try:
-        instance = identity.cloud_instance_factory()
-    except exceptions.CloudFactoryError as e:
-        if cfg.is_attached:
-            # We are attached on non-Pro Image, just report already attached
-            raise exceptions.AlreadyAttachedError(cfg)
-        if isinstance(e, exceptions.CloudFactoryNoCloudError):
-            raise exceptions.UserFacingError(
-                messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
-            )
-        if isinstance(e, exceptions.CloudFactoryNonViableCloudError):
-            raise exceptions.UserFacingError(messages.UNSUPPORTED_AUTO_ATTACH)
-        if isinstance(e, exceptions.CloudFactoryUnsupportedCloudError):
-            raise exceptions.NonAutoAttachImageError(
-                messages.UNSUPPORTED_AUTO_ATTACH_CLOUD_TYPE.format(
-                    cloud_type=e.cloud_type
-                )
-            )
-        # we shouldn't get here, but this is a reasonable default just in case
-        raise exceptions.UserFacingError(
-            messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
-        )
-
-    if not instance:
-        # we shouldn't get here, but this is a reasonable default just in case
-        raise exceptions.UserFacingError(
-            messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
-        )
+    instance = actions.get_cloud_instance(cfg)
 
     try:
         actions.auto_attach(cfg, instance)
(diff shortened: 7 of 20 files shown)


uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
```diff
@@ -0,0 +1,243 @@
...
+                raise AutoAttachError(
+                    msg=str(e.strerror), msg_code="unable-to-get-token"
+                )
+            except exceptions.UserFacingError as e:
+                raise AutoAttachError(
+                    msg=str(e.msg), msg_code="unable-to-get-token"
+                )
+            already_attached = True
+
+        services = list(set(options.enable + options.enable_beta))
```
aciba90, line 190: We agreed on [US006](https://docs.google.com/document/d/1tc3ZJC4wD6iyMWCFXC_rxWzbjrl0mN4J_YbF2iFoiRY) to:

1. fail if some beta service is given under `enable`
2. allow non-beta services under `enable_beta` to ease backward compatibility in cloud-init configs.

We do not check 1, is that difficult to check?

aciba90 (commented):
Thanks, @CalvoM, for the PR. I left you a comment and I have also updated my [draft PR](https://github.com/canonical/cloud-init/pull/1583) to show-case the integration with this one.

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
```diff
@@ -0,0 +1,243 @@
...
+        self.enable_beta = enable_beta
+
+
+class AutoAttachWithShortRetryResult(DataObject, AdditionalInfo):
+    fields = [
+        Field("successful", BoolDataValue),
+    ]
+
+    def __init__(self, *, successful: bool):
+        self.successful = successful
```
lucasmoura, line 72: I think we can return an empty result like we do on the magic attach revoke api

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def get_cloud_instance(cfg: UAConfig) -> Optional[AutoAttachCloudInstance]:
```
lucasmoura, line 75: This seems like the same code we use on `action_auto_attach` on the `cli` module. I think it is better to capture that behavior on the `actions` module instead. In that way, both `cli` and the `api` module can reuse the same code.

Additionally, @orndorffgrant  do you think on this refactoring we should already consider making the `action` function return a `Result` object ?

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
```diff
@@ -0,0 +1,243 @@
...
+
+    if not instance:
+        # we shouldn't get here, but this is a reasonable default just in case
+        raise exceptions.UserFacingError(
+            messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
+        )
+    return instance
+
+
+def detach_before_auto_attach(cfg: UAConfig):
```
lucasmoura, line 117: Same consideration for all the other functions

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
```diff
@@ -0,0 +1,243 @@
...
+                    and isinstance(reason, CanEnableFailure)
+                    and reason.message is not None
+                ):
+                    event.info(reason.message.msg)
+                    event.error(
+                        error_msg=reason.message.msg,
+                        error_code=reason.message.name,
+                        service=name,
+                    )
+                    # Should we raise an Error?
```
lucasmoura, line 210: We should not have `event` calls in the api module. Everything that goes wrong here need to be raised using an Exception

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
```diff
@@ -0,0 +1,243 @@
...
+            try:
+                instance = get_cloud_instance(cfg)
+            except exceptions.UserFacingError as e:
+                raise CloudError(
+                    msg=str(e.msg), msg_code="unable-to-complete-cloud-checks"
+                )
+            if not instance:
+                raise AutoAttachDisabledError(
+                    msg="UA Config disable_auto_attach is set."
+                )
```
lucasmoura, line 163: We don't need to convert the exceptions into `APIErrors` here. The `api` module already covers that for us. This apply to all other functions we are manually calling here

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
```diff
@@ -0,0 +1,243 @@
...
+            return AutoAttachWithShortRetryResult(successful=True)
+        i += 1
+        time.sleep(2)
+
+    # create this file to show retry is needed
+    retry_file = UAFile(
+        "auto-attach-background-retry-in-progress",
+        "/run/ubuntu-advantage/flags",
+    )
+    retry_file.write("1")
```
lucasmoura, line 231: Do we need content in that file ?

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
```diff
@@ -0,0 +1,243 @@
...
+                raise AutoAttachError(
+                    msg=str(e.msg), msg_code="unable-to-get-token"
+                )
+            already_attached = True
+
+        services = list(set(options.enable + options.enable_beta))
+        found, not_found = cli.get_valid_entitlement_names(services, cfg)
+        enabled_services = 0
+
+        for name in found:
```
lucasmoura, line 194: This logic is also the one found on the cli. We should refactor it to avoid code duplication

lucasmoura (commented):
I also think the endpoint should be something like:
`u.pro.auto_attach.v1` - for the short endoint
`u.pro.auto_attach.long.v1` - for the long call endopint

Even though the `long` here is just a placeholder name, I think we can make the default endpoint be the short version of `auto-attach`. What do you think @orndorffgrant @CalvoM ?

uaclient/api/api.py
```diff
@@ -18,6 +18,7 @@
     "u.pro.attach.magic.initiate.v1",
     "u.pro.attach.magic.wait.v1",
     "u.pro.attach.magic.revoke.v1",
+    "u.pro.auto_attach.with_short_retry.v1",
```
orndorffgrant, line 21: Moving the endpoint naming discussion to an inline-comment so that it can have it's own comment thread. @lucasmoura @CalvoM 

We already have an endpoint namespace for `attach` so we should reuse that.
I also think we will likely have the individual steps of auto-attach available as endpoints in the future, so we should be clear this endpoint is doing the whole auto-attach process.
The number of retries should probably be an optional parameter so I don't think "retry" should be in the name.

With all of that in mind, I propose: `u.pro.attach.auto.full_auto_attach.v1`

What do you think?

lucasmoura: +1 on the endpoint name

CalvoM: Makes sense. I saw the `attach` as a possible namespace to use, you have convinced me to change to it.

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
CalvoM: @aciba90 Updated.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,168 @@
...
+class BetaEntitlementFoundError(APIError):
+    pass
+
+
+class EntitlementNotEnabledError(APIError):
+    pass
+
+
+class FullAutoAttachFailureError(APIError):
+    pass
```
lucasmoura, line 29: As we discussed today, let us not create specific `APIError` exceptions here, instead just rely on the `message_code` we already have for the existing exceptions. However, let's double check to see if the exception being raise are lacking `message_codes`. If they are, we need to add them to the exceptions

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,168 @@
...
+            instance = None  # type: Optional[AutoAttachCloudInstance]
+            token = None  # Optional[str]
+            try:
+                instance = actions.get_cloud_instance(cfg)
+            except Exception as e:
+                raise e
+            try:
+                actions.detach_before_auto_attach(cfg)
+            except Exception as e:
+                raise e
```
lucasmoura, line 99: I don't think we need this `try..except` block for the functions.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,168 @@
...
+                msg.msg
+            )  # should we detach?
+        if len(services) == enabled_services:
+            return FullAutoAttachResult()
+        i += 1
+        time.sleep(2)
+
+    raise FullAutoAttachFailureError(
+        msg="full_auto_attach was not successful",
+        msg_code="full_auto_attach-error",
```
lucasmoura, line 159: Let's make this `full-auto-attach-error`

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
orndorffgrant: Sorry I missed this. I think the actions module may go away eventually. Its purpose was to be the one true implementation of a given feature, but that role is now filled by the api functions.
So once all of our cli is moved to using the api, then the `actions` module won't need to exist anymore.

The `Result` object is an api specific thing, so I think it might be a bit odd for the `actions` function to return it, but I don't feel strongly about it. I do think it is important to keep our PRs smaller in general, so if we want to refactor `actions` to return API `Results` then I think that should be a separate PR.

uaclient/api/exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+    BetaServiceError,
+    ContractAPIError,
+    EntitlementNotEnabledError,
+    EntitlementNotFoundError,
+    FullAutoAttachFailureError,
+    NonAutoAttachImageError,
+    UrlError,
+    UserFacingError,
+)
+
```
CalvoM, line 12: @aciba90 We are importing exceptions here to use for the api calls. So instead of importing from `uaclient.exceptions`, you could use `uaclient.api.exceptions`.

uaclient/actions.py
```diff
@@ -117,3 +119,62 @@ def status(
...
+    instance = None  # type: Optional[AutoAttachCloudInstance]
+    try:
+        instance = identity.cloud_instance_factory()
+    except exceptions.CloudFactoryError as e:
+        if cfg.is_attached:
+            # We are attached on non-Pro Image, just report already attached
+            raise exceptions.AlreadyAttachedError(cfg)
+        if isinstance(e, exceptions.CloudFactoryNoCloudError):
+            raise exceptions.UserFacingError(
+                messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
```
lucasmoura, line 145: Let's add `msg_code` to all of those exceptions

uaclient/actions.py
```diff
@@ -117,3 +119,62 @@ def status(
...
+    if not instance:
+        # we shouldn't get here, but this is a reasonable default just in case
+        raise exceptions.UserFacingError(
+            messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
+        )
+    return instance
+
+
+def detach_before_auto_attach(cfg: config.UAConfig):
+    from uaclient.cli import _detach
```
lucasmoura, line 169: This could be problematic, if we are using the `cli` module, we might be leaking `event.info`calls for the api.
Additionally, we could also prompt users here, which is not ideal.

I think we should refactor this to allow us to be decouple from any `cli` function

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,146 @@
...
+        Field("enable", data_list(StringDataValue), False),
+        Field("enable_beta", data_list(StringDataValue), False),
+        Field("retries", IntDataValue, False),
+    ]
+
+    def __init__(
+        self,
+        enable: Optional[List[str]],
+        enable_beta: Optional[List[str]],
+        retries: Optional[int],
```
lucasmoura, line 31: I think we should add the `None` default value here just to explicit

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,146 @@
...
+            allow_enable = not any([options.enable, options.enable_beta])
+            token = get_auto_attach_token(cfg, instance)  # type: ignore
+            actions.attach_with_token(
+                cfg, token=token, allow_enable=allow_enable
+            )
+            if allow_enable:
+                return FullAutoAttachResult()
+            already_attached = True
+
+        services = list()
```
lucasmoura, line 90: I think we can create a `set` here directly and just use the `update` method to create the final set result

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,146 @@
...
+                raise exceptions.BetaServiceError(
+                    msg="beta service found in the enable list",
+                    msg_code="beta-service-found",
+                )
+            services += options.enable
+        if options.enable_beta:
+            services += options.enable_beta
+
+        services = list(set(services))
+        found, not_found = cli.get_valid_entitlement_names(services, cfg)
```
lucasmoura, line 102: Let's avoid using any `cli` function here. We should refactor the code to not make that import directly from there

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,146 @@
...
+        retries_limit = options.retries
+    already_attached = False  # avoid AlreadyAttachedOnProError after 1st time
+    while 0 < retries_limit:
+        if not already_attached:
+            instance = None  # type: Optional[AutoAttachCloudInstance]
+            token = None  # Optional[str]
+            instance = actions.get_cloud_instance(cfg)
+            actions.detach_before_auto_attach(cfg)
+            allow_enable = not any([options.enable, options.enable_beta])
+            token = get_auto_attach_token(cfg, instance)  # type: ignore
```
lucasmoura, line 82: Maybe I am missing something here, but I believe the retry should only be executed for this function in particular. Additionally, this function is responsible for making two separate requests:

* The cloud metadata
* The contracts server with that metadata

For the cloud metadata, we already have a timeout mechanism in place. Maybe we should do something similar to the contract call directly, instead of looping here in the API code. What do you all think ? @renanrodrigo @CalvoM @orndorffgrant

uaclient/files.py
```diff
@@ -54,6 +54,10 @@ def read(self) -> Optional[str]:
     def delete(self):
         system.remove_file(self.path)
 
+    def create(self):
+        f = open(self.path, "x")
+        f.close()
```
lucasmoura, line 59: I think we can `Pathlib` touch here instead. If I am not mistaken, `Pathlib` should be available in python 3.5
But I think I also missing where in the code we are using this

uaclient/exceptions.py
```diff
@@ -306,6 +306,18 @@ def __str__(self):
         )
 
 
+class EntitlementNotEnabledError(UserFacingError):
+    """An exception raised when enabling of an entitlement fails"""
+
+    pass
+
+
+class FullAutoAttachFailureError(UserFacingError):
```
lucasmoura, line 315: We could already create the exception with the message we want here directly, like we do on `APTInstallError` for example. In that way, we can move the message into the `message` module and just use it here

uaclient/api/u/pro/auto_attach/with_short_retry/v1.py
lucasmoura: Agreed, let's not touch on the `actions` module for now

uaclient/actions.py
CalvoM: Alright.

uaclient/files.py
CalvoM: I noticed that with the retry file, I could not create the file only without having to call the `write` method. I thought having this method might be of help later, like that instance.

uaclient/actions.py
orndorffgrant: @CalvoM I just merged https://github.com/canonical/ubuntu-advantage-client/pull/2176 so you no longer need the `detach_before_auto_attach` function :tada:

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
lucasmoura: @CalvoM this abstraction is still retrying on any exception that is raised by those calls. I think instead of having the retry loop here, we could use the `util.retry` decorator directly on the functions we are calling. The only downside is that the retry limits will not be a parameter that can be passed to the api, but I think it is still worth it

uaclient/api/tests/test_u_pro_attach_auto_full_auto_attach_v1.py
```diff
@@ -0,0 +1,58 @@
...
+from uaclient.api import exceptions
+from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
+    FullAutoAttachOptions,
+    full_auto_attach,
+)
+
+M_API = "uaclient.api.u.pro."
+
+
+class TestFullAutoAttachV1:
```
lucasmoura, line 13: Let's add tests for all the exceptions that we can raise in the API, just to be sure everything is working as expected

lucasmoura (commented):
@CalvoM I think we can already add integration tests for the API

uaclient/files.py
lucasmoura: Fair enough, but I missing where it is actually being called in the code. Are we actually using it ?

CalvoM: Not currently. I noticed I do not need to create the retry file, but the caller of the api endpoint will create it.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: I thought of this approach, and had the same concern. If it is worth it then it is okay.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+            if ent_cls.is_beta:
+                return True
+        except exceptions.EntitlementNotFoundError:
+            continue
+    return False
+
+
+def full_auto_attach(options: FullAutoAttachOptions, cfg: UAConfig):
+    instance = None  # type: Optional[AutoAttachCloudInstance]
+    token = None  # Optional[str]
```
orndorffgrant, line 63: Why do we need to initialize these to None?

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+    instance = actions.get_cloud_instance(cfg)
+    allow_enable = not any([options.enable, options.enable_beta])
```
orndorffgrant, line 65: I had to stare at this for a moment to understand it. I'd find something like this more readable (but maybe this is just me)
```suggestion
    allow_enable_by_default = options.enable is None and options.enable_beta is None
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+        if is_any_beta(cfg, options.enable):
+            raise exceptions.BetaServiceError(
+                msg="beta service found in the enable list",
+                msg_code="beta-service-found",
+            )
+        services.update(options.enable)
+    if options.enable_beta:
+        services.update(options.enable_beta)
+
+    services_ = list(set(services))
```
orndorffgrant, line 82: services is already a `set` here, right?
And a trailing underscore is easy to miss while reading. Let's make the name noticeably different
```suggestion
    services_list = list(services)
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if not_found:
+        msg = cli._create_enable_entitlements_not_found_message(
+            not_found, cfg=cfg, allow_beta=True
+        )
+        raise exceptions.EntitlementNotFoundError(msg.msg)
+    if len(services_) == enabled_services:
+        return FullAutoAttachResult()
+
+    raise exceptions.FullAutoAttachFailureError()
```
orndorffgrant, line 112: We have the data to raise a better error here. Let's include which services failed to enable in the error as additional_info

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+                    additional_info={"service": name},
+                )
+        else:
+            enabled_services += 1
+
+    if not_found:
+        msg = cli._create_enable_entitlements_not_found_message(
+            not_found, cfg=cfg, allow_beta=True
+        )
+        raise exceptions.EntitlementNotFoundError(msg.msg)
```
orndorffgrant, line 108: Can we include the not found services as additional_info

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+    if allow_enable:
+        return FullAutoAttachResult()
+
+    services = set()
+    if options.enable:
+        if is_any_beta(cfg, options.enable):
+            raise exceptions.BetaServiceError(
+                msg="beta service found in the enable list",
+                msg_code="beta-service-found",
+            )
```
orndorffgrant, line 77: Lets include which service was beta in the additional_info

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+        try:
+            ent_cls = entitlements.entitlement_factory(cfg, name)
+            if ent_cls.is_beta:
+                return True
+        except exceptions.EntitlementNotFoundError:
+            continue
+    return False
+
+
+def full_auto_attach(options: FullAutoAttachOptions, cfg: UAConfig):
```
orndorffgrant, line 61: To align with #2170, lets make this small change
```suggestion
def full_auto_attach(options: FullAutoAttachOptions) -> FullAutoAttachResult:
    return _full_auto_attach(option)


def _full_auto_attach(options: FullAutoAttachOptions, cfg: UAConfig) -> FullAutoAttachResult:
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+    if len(services_) == enabled_services:
+        return FullAutoAttachResult()
+
+    raise exceptions.FullAutoAttachFailureError()
+
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="AutoAttachWithShortRetry",
+    fn=full_auto_attach,
```
orndorffgrant, line 118: To align with #2170 
```suggestion
    fn=_full_auto_attach,
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+        raise exceptions.EntitlementNotFoundError(msg.msg)
+    if len(services_) == enabled_services:
+        return FullAutoAttachResult()
+
+    raise exceptions.FullAutoAttachFailureError()
+
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="AutoAttachWithShortRetry",
```
orndorffgrant, line 117:
```suggestion
    name="FullAutoAttach",
```

uaclient/contract.py
```diff
@@ -81,6 +82,17 @@ def request_contract_information(
...
+    def request_url(
+        self,
+        path,
+        **kwargs,
+    ):
+        return super().request_url(
+            path,
+            **kwargs,
+        )
+
```
orndorffgrant, line 95: @lucasmoura @CalvoM do we want these retries on _every_ contract server request?

I suggest we only add this to the ones where we want it for this PR.

I also suggest we change the sleeps to something shorter, maybe [1, 2, 2]

uaclient/exceptions.py
```diff
@@ -306,6 +306,22 @@ def __str__(self):
...
+    """An exception raised when auto attach at boot fails"""
+
+    def __init__(self):
+        super().__init__(
+            msg=messages.FULL_AUTO_ATTACH_ERROR.msg,
+            msg_code=messages.FULL_AUTO_ATTACH_ERROR.name,
+        )
```
orndorffgrant, line 322: If these were added specifically to be raised by the api function, then I think we should go ahead and define them in `uaclient.api.exceptions` - thoughts @lucasmoura @CalvoM @renanrodrigo ?

uaclient/contract.py
lucasmoura: +1

uaclient/exceptions.py
lucasmoura: +1

uaclient/contract.py
```diff
@@ -81,6 +82,17 @@ def request_contract_information(
...
+    @util.retry(socket.timeout, retry_sleeps=[1, 3, 5])
+    def request_url(
+        self,
+        path,
+        **kwargs,
+    ):
+        return super().request_url(
+            path,
+            **kwargs,
+        )
```
renanrodrigo, line 94: Python 3.5 is complaining about the syntax here... But we can oneline it and then Python sees the truth.
```suggestion
    def request_url(self, path, **kwargs):
        return super().request_url(path, **kwargs)
```

uaclient/exceptions.py
renanrodrigo: +1

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
+from typing import List, Optional
+
+from uaclient import actions, cli, contract, entitlements, messages
```
renanrodrigo, line 3: I'd like to suggest that none of our API functionality imports from `uaclient.cli`. In the future, we want to do the opposite: CLI will leverage the API functions. So I'd suggest not importing this, and moving the imported function out of the CLI module - maybe to Entitlements?

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,120 @@
...
+from uaclient.api import exceptions
+from uaclient.api.api import APIEndpoint
+from uaclient.api.data_types import AdditionalInfo
+from uaclient.clouds import AutoAttachCloudInstance  # noqa: F401
+from uaclient.config import UAConfig
+from uaclient.data_types import DataObject, Field, StringDataValue, data_list
+from uaclient.entitlements import get_valid_entitlement_names
```
renanrodrigo, line 10: For consistency, there is no need to import this if you have `entitlements` imported above. Or, just import the Entitlement Factory here as well, and remove that.

uaclient/actions.py
```diff
@@ -218,3 +219,49 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
             system.write_file(
                 os.path.join(output_dir, os.path.basename(f)), content
             )
+
+
+def get_cloud_instance(
```
renanrodrigo, line 224: This makes all sense for me, but then, to avoid duplication, let's make the `action_auto_attach` function to call this, instead of having the same logic there.

uaclient/contract.py
orndorffgrant: In [this comment](https://github.com/canonical/ubuntu-advantage-client/pull/2171#discussion_r939159521), @lucasmoura and I agreed that the retry should not be applied to every `request_url` so I think this code will get removed anyway.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: Oh forgot to remove this after updating the code.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: Your suggestion is much more readable.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: We raise an error when to fail to enable an entitlement i.e. `EntitlementNotEnableError`. Infact, I do not think we will ever reach the `FullAutoAttachFailureError` since we would have failed before incase of an error.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: Nice suggestion. On it.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,125 @@
...
+    ):
+        self.enable = enable
+        self.enable_beta = enable_beta
+
+
+class FullAutoAttachResult(DataObject, AdditionalInfo):
+    pass
+
+
+def get_auto_attach_token(cfg: UAConfig, cloud: AutoAttachCloudInstance):
```
orndorffgrant, line 32: Since this and `is_any_beta` are helper functions that are not part of the API, let's give them underscores in front of their name to follow the convention to indicate that they're "private".
```suggestion
def _get_auto_attach_token(cfg: UAConfig, cloud: AutoAttachCloudInstance):
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,125 @@
...
+            raise exceptions.NonAutoAttachImageError(
+                messages.UNSUPPORTED_AUTO_ATTACH
+            )
+        raise e
+
+    token = tokenResponse["contractToken"]
+    return token
+
+
+def is_any_beta(cfg: UAConfig, ents: List[str]) -> Tuple[bool, str]:
```
orndorffgrant, line 49: See above comment explanation for suggestion
```suggestion
def _is_any_beta(cfg: UAConfig, ents: List[str]) -> Tuple[bool, str]:
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,125 @@
...
+            if (
+                reason is not None
+                and isinstance(reason, CanEnableFailure)
+                and reason.message is not None
+            ):
+                raise exceptions.EntitlementNotEnabledError(
+                    msg=reason.message.msg,
+                    msg_code=reason.message.name,
+                    additional_info={"service": name},
+                )
```
orndorffgrant, line 105: We should always raise something if the entitlement fails to enable.
```suggestion
        if not ent_ret:
            if (
                reason is not None
                and isinstance(reason, CanEnableFailure)
                and reason.message is not None
            ):
                raise exceptions.EntitlementNotEnabledError(
                    msg=reason.message.msg,
                    msg_code=reason.message.name,
                    additional_info={"service": name},
                )
            else:
                raise exceptions.EntitlementNotEnabledError(
                    msg="Failed to enable service: {}".format(name),
                    msg_code="entitlement-not-enabled",
                    additional_info={"service": name},
                )
```

With this, then I don't think we need the `enabled_services` count and associated check at the end of the function

features/api_full_auto_attach.feature
```diff
@@ -0,0 +1,30 @@
...
+        When I create the file `/tmp/full_auto_attach.py` with the following:
+        """
+        from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import full_auto_attach, FullAutoAttachOptions
+        from uaclient.config import UAConfig
+
+        cfg = UAConfig(root_mode=True)
+        opts = {'enable': ['esm-infra', 'esm-apps'], 'enable_beta': []}
+        options = FullAutoAttachOptions.from_dict(opts)
+        full_auto_attach(options)
+        """
```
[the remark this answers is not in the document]
CalvoM: Oh yeah, it is still WIP. I had put it there to start the integration test.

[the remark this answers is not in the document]
orndorffgrant: oh cool sorry :)

features/api_full_auto_attach.feature
```diff
@@ -0,0 +1,54 @@
...
+        And I run `python3 /tmp/full_auto_attach.py` with sudo
+        Then stdout matches regexp:
+        """
+        Ubuntu Pro: ESM Apps enabled
+        """
+        And stdout matches regexp:
+        """
+        Ubuntu Pro: ESM Infra enabled
+        """
```
orndorffgrant, line 30: That script shouldn't print anything :upside_down_face: 

For now, this may have to be done by setting the event_logger to "json" mode, so that `event.info` calls don't print to stdout.

Ideally in the future, we'll break up all the individual pieces of auto attach so that none of them print in the actual action. Printing will only happen in the cli module. But that is the future state -  not for this PR

orndorffgrant: CC @lucasmoura

features/api_full_auto_attach.feature
```diff
@@ -0,0 +1,54 @@
...
+        SERVICE       +ENTITLED  STATUS    DESCRIPTION
+        cc-eal        +yes +<cc-eal-s>  +Common Criteria EAL2 Provisioning Packages
+        """
+        Then stdout matches regexp:
+        """
+        esm-apps      +yes +enabled +Extended Security Maintenance for Applications
+        esm-infra     +yes +enabled +Extended Security Maintenance for Infrastructure
+        fips          +yes +<fips-s> +NIST-certified core packages
+        fips-updates  +yes +<fips-s> +NIST-certified core packages with priority security updates
+        livepatch     +yes +enabled  +Canonical Livepatch service
```
orndorffgrant, line 43: livepatch wasn't passed in the `enable` parameter, so it shouldn't get enabled.
```suggestion
        livepatch     +yes +disabled  +Canonical Livepatch service
```

features/api_full_auto_attach.feature
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Then stdout matches regexp:
+        """
+        <cis_or_usg>           +yes  +<cis-s>  +Security compliance and audit tools
+        """
```
orndorffgrant, line 48: We only really need to test that
1. the machine is attached
2. the services that got passed in the `enable` param got enabled
3. the services that normally get enabled by default but aren't in the `enable` array did not get enabled
```suggestion
        When I run `pro status --all` with sudo
        Then stdout matches regexp:
        """
        esm-apps      +yes +enabled +Extended Security Maintenance for Applications
        esm-infra     +yes +enabled +Extended Security Maintenance for Infrastructure
        """
        Then stdout matches regexp:
        """
        livepatch     +yes +disabled  +Canonical Livepatch service
        """
```

uaclient/actions.py
```diff
@@ -218,3 +219,52 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
...
+
+
+def should_disable_auto_attach(cfg: config.UAConfig) -> bool:
+    disable_auto_attach = util.is_config_value_true(
+        config=cfg.cfg, path_to_value="features.disable_auto_attach"
+    )
+    if disable_auto_attach:
+        msg = "Skipping auto-attach. Config disable_auto_attach is set."
+        logging.debug(msg)
+        print(msg)
```
orndorffgrant, line 231: This printing shouldn't happen in the api function. This should only happen in the cli module. (sorry I didn't notice this last time I commented on this bit of code :frowning_face:)

uaclient/api/tests/test_u_pro_attach_auto_full_auto_attach_v1.py
```diff
@@ -0,0 +1,125 @@
...
+        cfg = FakeConfig(root_mode=True)
+        opts = {
+            "enable": ["esm-infra", "esm-apps", "fips"],
+            "enable_beta": ["realtime-kernel"],
+        }
+        options = FullAutoAttachOptions.from_dict(opts)
+        with pytest.raises(exceptions.UserFacingError):
+            _full_auto_attach(options, cfg)
+
+        assert 1 >= enable_ent_by_name.call_count
```
orndorffgrant, line 125: lets make this stricter
```suggestion
        assert 1 == enable_ent_by_name.call_count
```

uaclient/api/tests/test_u_pro_attach_auto_full_auto_attach_v1.py
```diff
@@ -0,0 +1,125 @@
...
+        enable_ent_by_name,
+        FakeConfig,
+    ):
+        cfg = FakeConfig(root_mode=True)
+        opts = {
+            "enable": ["esm-infra", "esm-apps", "fips"],
+            "enable_beta": ["realtime-kernel"],
+        }
+        options = FullAutoAttachOptions.from_dict(opts)
+        with pytest.raises(exceptions.UserFacingError):
```
orndorffgrant, line 122: This scenario has a more specific expected error right?
```suggestion
        with pytest.raises(exceptions.EntitlementNotEnabledError):
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,126 @@
...
+def full_auto_attach(options: FullAutoAttachOptions) -> FullAutoAttachResult:
+    return _full_auto_attach(options, UAConfig(root_mode=True))
+
+
+def _full_auto_attach(options: FullAutoAttachOptions, cfg: UAConfig):
+    instance = actions.get_cloud_instance(cfg)
+    allow_enable = options.enable is None and options.enable_beta is None
+    token = _get_auto_attach_token(cfg, instance)  # type: ignore
+    actions.attach_with_token(cfg, token=token, allow_enable=allow_enable)
+    if allow_enable:
```
orndorffgrant, line 69: Can we change the name of this variable to
```suggestion
    if enable_default_services:
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,126 @@
...
+            raise exceptions.BetaServiceError(
+                msg="beta service found in the enable list",
+                msg_code="beta-service-found",
+                additional_info={"beta_service": service},
+            )
+        services.update(options.enable)
+    if options.enable_beta:
+        services.update(options.enable_beta)
+
+    service_list = list(set(services))
```
orndorffgrant, line 85: services is already a set, right?
```suggestion
    service_list = list(services)
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,126 @@
...
+                    msg="Failed to enable service: {}".format(name),
+                    msg_code="entitlement-not-enabled",
+                    additional_info={"service": name},
+                )
+
+    if not_found:
+        msg = entitlements.create_enable_entitlements_not_found_message(
+            not_found, cfg=cfg, allow_beta=True
+        )
+        raise exceptions.EntitlementNotFoundError(msg.msg, not_found)
```
orndorffgrant, line 116: An idea to discuss with the team. Some of these checks we can do up front. even before auto-attaching. Should we do a full verification of the requested services and raise an error before doing anything? @CalvoM @lucasmoura @renanrodrigo

orndorffgrant: Or alternatively, should we move all checks as late as possible? Like we could move the beta service error to the end

uaclient/actions.py
```diff
@@ -218,3 +219,52 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
...
+
+    return disable_auto_attach
+
+
+def get_cloud_instance(
+    cfg: config.UAConfig,
+) -> Optional[AutoAttachCloudInstance]:
```
orndorffgrant, line 238: We raise an error in this function if the instance is None
```suggestion
) -> AutoAttachCloudInstance:
```

uaclient/cli.py
```diff
@@ -1282,44 +1238,13 @@ def action_auto_attach(args, *, cfg):
...
-            messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
-        )
-
+    instance = actions.get_cloud_instance(cfg)
     if not instance:
-        # we shouldn't get here, but this is a reasonable default just in case
-        raise exceptions.UserFacingError(
-            messages.UNABLE_TO_DETERMINE_CLOUD_TYPE
-        )
+        return 0
```
orndorffgrant, line 1247: I don't think we need this

orndorffgrant: To clarify, we don't need the `if not instance:` check because it can't be None

uaclient/actions.py
CalvoM: Yes, that is why I separated it to another function, `shoudl_disable_auto_attach` which is only called by the cli module. We could move this code to the cli module as well, since it is not called by another module.

orndorffgrant: Ah I see - sorry I misunderstood! This looks good, no need to change

features/api_full_auto_attach.feature
CalvoM: Oh yes,

uaclient/cli.py
CalvoM: True, we will just raise an exception.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: yes.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+    instance = actions.get_cloud_instance(cfg)
+    enable_default_services = (
+        options.enable is None and options.enable_beta is None
+    )
+    token = _get_auto_attach_token(cfg, instance)  # type: ignore
+    actions.attach_with_token(
+        cfg, token=token, allow_enable=enable_default_services
+    )
```
orndorffgrant, line 127: Can this be replaced with
```suggestion
    actions.auto_attach(
        cfg, instance, allow_enable=enable_default_services
    )
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
orndorffgrant: @CalvoM this validation is still being treated differently from the beta and incompatible checks. Should we move it up before any attaching happens?

orndorffgrant (commented):
Just two minor things, this is really close!

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
lucasmoura: Additionally, we should have a better way to check for incompatible services in the list as @renanrodrigo mentioned, like `["fips", "livepatch"]`. But maybe it should be a follow up PR

CalvoM: @lucasmoura I added the `_is_compatible_services_present` function, you can check it out.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: Oh, yeah. I am trying to remember why I created `_get_auto_attach_token`, it seems I was trying to solve an issue that currently is not there.

CalvoM: Oh I have just remembered, it is because in `actions.auto_attach` the `allow_enable` parameter for `attach_with_token` function defaults to true. If I change it, i.e. add `allow_enable` parameter to `actions.auto_attach`, I will have to change more places. Could I create a PR for this change?

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,166 @@
...
+
+def _is_compatible_services_present(
+    cfg, ent_list: List[str]
+) -> Tuple[bool, str, List[str]]:
+    ent_cls = list()  # type: List[Type[UAEntitlement]]
+    for e in ent_list:
+        ent_cls.append(entitlements.entitlement_factory(cfg, e))
+
+    for ent in ent_cls:
+        if ent().incompatible_services:
```
lucasmoura, line 71: Let's instantiate the entitlement only once in the code. Additionally, let's pass the `cfg` object to it, otherwise we will unnecessarily create another cfg object when creating the object.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,166 @@
...
+            incompat = [
+                service.entitlement for service in ent().incompatible_services
+            ]
+            if any(e in incompat for e in ent_cls):
```
lucasmoura, line 75: Let's add a test that verify this check, because I am not entirely sure if that class comparison will work as expected.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,166 @@
...
+        try:
+            ent_cls = entitlements.entitlement_factory(cfg, name)
+            if ent_cls.is_beta:
+                return True, name
+        except exceptions.EntitlementNotFoundError:
+            continue
+    return False, ""
+
+
+def _is_compatible_services_present(
```
lucasmoura, line 63: Should we rename it to `_ is_incompatible_services_present` ?

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: Updated

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: Updated

uaclient/api/tests/test_u_pro_attach_auto_full_auto_attach_v1.py
```diff
@@ -0,0 +1,176 @@
...
+        return_value=(True, None),
+    )
+    def test_error_when_beta_in_enable_list(
+        self,
+        enable_ent_by_name,
+        FakeConfig,
+    ):
+        cfg = FakeConfig(root_mode=True)
+        opts = {"enable": ["esm-infra", "realtime-kernel"]}
+        options = FullAutoAttachOptions.from_dict(opts)
```
orndorffgrant, line 31: Sorry I didn't look closely at these tests before. There's no need to create all these Options objects from dicts. And there is a reason to create them directly: mypy will help us pass the correct args.

I think all of the tests in this file should construct the Options object directly like so
```suggestion
        options = FullAutoAttachOptions(enable=["esm-infra", "realtime-kernel"])
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -0,0 +1,168 @@
...
+        )
+        raise exceptions.EntitlementNotFoundError(msg.msg, not_found)
+
+    incompat_detected, ent, incompat_ents = _is_incompatible_services_present(
+        cfg, sorted(found)
+    )  # sort for easy testing exceptions raised
+    if incompat_detected:
+        msg_ = "{ent} is incompatible with any of these services {incompat}"
+        raise exceptions.IncompatibleEntitlementsDetected(
+            msg=msg_.format(ent=ent, incompat=incompat_ents),
```
orndorffgrant, line 119: why is there an extra underscore in this variable name?
```suggestion
        msg = "{ent} is incompatible with any of these services {incompat}"
        raise exceptions.IncompatibleEntitlementsDetected(
            msg=msg.format(ent=ent, incompat=incompat_ents),
```

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
orndorffgrant: @CalvoM 
I think the change is just this?
```diff
diff --git a/uaclient/actions.py b/uaclient/actions.py
index def703f1..aa70786e 100644
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -71,7 +71,7 @@ def attach_with_token(
 
 
 def auto_attach(
-    cfg: config.UAConfig, cloud: clouds.AutoAttachCloudInstance
+    cfg: config.UAConfig,
+    cloud: clouds.AutoAttachCloudInstance,
+    allow_enable: bool = True
 ) -> None:
     """
     :raise UrlError: On unexpected connectivity issues to contract
@@ -95,7 +95,7 @@ def auto_attach(
 
     token = tokenResponse["contractToken"]
 
-    attach_with_token(cfg, token=token, allow_enable=True)
+    attach_with_token(cfg, token=token, allow_enable=allow_enable)
 
 
 def enable_entitlement_by_name(

```

and all tests should still pass with that change.


But if you'd like to do it in a follow up PR that's okay too :) (maybe I'm missing something)

orndorffgrant (commented):
Just some last minute nits :)

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
CalvoM: mypy has errors. It thinks `msg` is of type `NamedMessage` because it is first created 
```
entitlements.create_enable_entitlements_not_found_message(
            not_found, cfg=cfg, allow_beta=True
        )
```

features/api_full_auto_attach.feature
```diff
@@ -0,0 +1,38 @@
...
+        Then stdout matches regexp:
+        """
+        livepatch     +yes +disabled  +Canonical Livepatch service
+        """
+        Examples:
+           | release |
+           | xenial  |
+           | bionic  |
+           | focal   |
+           | jammy   |
```
renanrodrigo, line 38: Evil missing newline

CalvoM: Oh yes.

CalvoM: Updated.

renanrodrigo (approved):

orndorffgrant (approved):
LGTM! Thanks for this! Let's just let CI go green first

PR #2172, opened 2022-07-25
data object file

orndorffgrant:
## Test Steps

Edit a file to create a DataObjectFile.

Write code to `write` an invalid object or wrong type. Verify that mypy catches it

Write code to use the return value of `read` improperly (e.g. dont check for None, or look at wrong field name). Verify mypy catches it

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/files.py
+++ b/uaclient/files.py
@@ -3,9 +3,20 @@ import logging
 import os
 import re
 from datetime import datetime
-from typing import Any, Dict, Optional
+from enum import Enum
+from typing import Any, Dict, Generic, Optional, Type, TypeVar
 
-from uaclient import defaults, event_logger, exceptions, messages, system, util
+import yaml
+
+from uaclient import (
+    data_types,
+    defaults,
+    event_logger,
+    exceptions,
+    messages,
+    system,
+    util,
+)
 from uaclient.contract_data_types import PublicMachineTokenData
 
 event = event_logger.get_event_logger()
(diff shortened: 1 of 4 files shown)


uaclient/files.py
```diff
@@ -55,6 +66,52 @@ def delete(self):
...
+    def read(self) -> Optional[DOFType]:
+        raw_data = self.ua_file.read()
+        if raw_data is None:
+            return None
+
+        parsed_data = None
+        if self.file_format == DataObjectFileFormat.JSON:
+            parsed_data = json.loads(raw_data)
+        elif self.file_format == DataObjectFileFormat.YAML:
+            parsed_data = yaml.safe_load(raw_data)
```
lucasmoura, line 97: Should we perform any exception handling for invalid json or yaml content ?

lucasmoura (commented):
LGTM, but we might have a valid unit test failure

uaclient/files.py
orndorffgrant: Yeah we probably should - good catch! We should probably just catch either json or yaml decode error and re-raise it as a "InvalidFormatError" so that the caller can just catch that without know what the underlying format is.

orndorffgrant:
@lucasmoura updated! I wasn't sorting the json output by key before which is what led to the CI failure. Hopefully CI is happy this time

lucasmoura (commented):
@orndorffgrant can we use this abstraction for the machine token files classes somehow ?

orndorffgrant:
@lucasmoura Yes! As well as notices.json and uaclient.conf and jobs-status.json and maybe more!

I wanted to keep this PR small with just this implementation, specifically to server the new cache file that will be necessary for the livepatch support functionality, but I hope we will over time refactor data files to use this abstraction.

lucasmoura (approved):

PR #2173, opened 2022-07-26
notices: limit to root_users only via UAfile

CalvoM:
## Proposed Commit Message

notices: limit to root_users only via UAfile

This limits writing the writing of the notice file to the root users only. Before, we had a user could write to it 
before changing the file mode.

## Desired commit type

## Checklist


lucasmoura (commented):
@CalvoM we also need to check the debian scripts, like `postinst` to make the changes there as well

orndorffgrant:
This looks good to me as well. @lucasmoura do you mean that we should add a couple lines to `postinst` to do a `chown root root` if the file exists - just to make sure all existing notices files are owned by root?

debian/ubuntu-advantage-tools.postinst
```diff
@@ -246,7 +246,7 @@ add_notice() {
 from uaclient.config import UAConfig
 from uaclient.messages import ${msg_name}
 cfg = UAConfig(root_mode=True)
-cfg.add_notice(label='', description=${msg_name})
+cfg.notice_file.add(label='', description=${msg_name})
```
CalvoM, line 249: @orndorffgrant I think @lucasmoura meant I should add this.

orndorffgrant: Ah okay! and those changes were already there when I was reviewing - I didn't see a response to his comment so was confused.

Looks good!

lucasmoura (approved):

PR #2175, opened 2022-07-27
tests: use prod env for magic-attach tests

lucasmoura:
## Proposed Commit Message
tests: use prod env for magic-attach tests

The magic-attach changes are now released on prod. We are updating the tests to use it instead of staging

## Test Steps
Check the magic-attach tests are working as expected on the prod env

## Desired commit type

## Checklist


features/api_magic_attach.feature
```diff
@@ -1,11 +1,9 @@
 Feature: Magic Attach endpoints
 
-    @series.lts
```
renanrodrigo, line 3: I think we should keep the series, right?

lucasmoura: Yes, my mistake

renanrodrigo (approved):
tests passing!

PR #2176, opened 2022-07-28
dont detach on autoattach

orndorffgrant:
## Test Steps

Run edited integration test.

## Desired commit type

## Checklist

### DIFF
--- a/docs/source/tutorials/create_a_fips_updates_pro_cloud_image.md
+++ b/docs/source/tutorials/create_a_fips_updates_pro_cloud_image.md
@@ -30,6 +30,11 @@ And verify that `fips-updates` is enabled in the output of `ua status`
 sudo ua status
 ```
 
+Also remove the machine-id so that it is regenerated for each instance launch from the snapshot.
+```bash
+sudo rm /etc/machine-id
+```
+
 ## Step 3: Snapshot the instance as a cloud image
 
 Cloud-specific instructions are here:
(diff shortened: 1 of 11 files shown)


docs/source/tutorials/create_a_fips_updates_pro_cloud_image.md
```diff
@@ -0,0 +1,55 @@
...
+
+You will need to re-enable fips-updates on each instance that is launched from the custom image.
+
+This won't require a reboot and is only necessary to ensure the instance gets updates to fips packages when they become available.
+
+```bash
+sudo ua enable fips-updates --assume-yes
+```
+
+You can easily script this using [cloud-init](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#runcmd)
```
blackboxsw, line 52: Might as well just provide that for folks:
```suggestion
You can easily script with the user-data at launch time using [cloud-init](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#runcmd)
```
#cloud-config
runcmd:
 - 'ua status --wait'
 - 'ua enable fips-updates --assume-yes'
power_state:
  mode: reboot
```

```

uaclient/cli.py
```diff
@@ -1340,17 +1343,6 @@ def action_auto_attach(args, *, cfg):
...
-    if cfg.is_attached:
-        prev_iid = cfg.read_cache("instance-id")
-        if str(current_iid) == str(prev_iid):
-            raise exceptions.AlreadyAttachedOnPROError(str(current_iid))
-        print("Re-attaching Ubuntu Pro subscription on new instance")
-        if _detach(cfg, assume_yes=True) != 0:
-            raise exceptions.UserFacingError(
-                messages.DETACH_AUTOMATION_FAILURE
-            )
-
```
blackboxsw, line 1353: I think this is reasonable given the change in the contract server hanlding which disambiguates multiple attached machines under the same contract. +1

blackboxsw (approved):
This behavior, documentation and integration testss look good to me. Note as well that cloud-init is adding a `cloud-init clean --machine-id` parameter to help folks with image cloning in general adhere to best practices to ensure new boots of cloned images create unique /etc/machine-id on next boot. I realize this isn't quite what you are getting at with your doc example here. But wanted to FYI on that option that landed in tip of main for cloud-init.

docs/source/tutorials/create_a_fips_updates_pro_cloud_image.md
orndorffgrant: The reboot is specifically not required, but point taken! I'll add the runcmd config to the doc.

orndorffgrant:
Thanks @blackboxsw for the review! Do you think we should suggest cleaning the machine-id in our docs? If that is a generally recommended practice, we should probably include it.

renanrodrigo (approved):
lgtm
:shipit:

blackboxsw:
> 

I'd say we want to suggest `rm /etc/machine-id` if creating a cloned image to ensure new machine-ids are created and cloned images don't have to rely on the contract backend disambiguation. What you'd lose by suggesting that step is a simple way in contract backend to track how many clones of a PRO image are out there. So that might be a thing the talk about with Product as to whether that is a "feature" that is wanted.

orndorffgrant:
That makes sense @blackboxsw! I added the `rm /etc/machine-id` step to our docs

merging now

PR #2182, opened 2022-07-31
Added livepatch check in fix_security_issue_id

### LINKED ISSUE #2181: ua fix should check if livepatch has fixed the provided cve
and provide a descriptive message if it has

dixonmyles:
## Proposed Commit Message

Adds a check of livepatch output during ua fix

Adds a subprocess call of canonical-livepatch with verbose flag
within the fix_security_issue_id method, parses the output as json,
and checks for the presence of the cve issue id in the output's fixes
property, if present.
Resolves #2181 

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/messages.py
+++ b/uaclient/messages.py
@@ -74,7 +74,8 @@ SECURITY_AFFECTED_PKGS = (
 USN_FIXED = "{issue} is addressed."
 CVE_FIXED = "{issue} is resolved."
 CVE_FIXED_BY_LIVEPATCH = (
-    "{issue} is resolved by livepatch patch version: {version}."
+    OKGREEN_CHECK
+    + " {issue} is resolved by livepatch patch version: {version}."
 )
 SECURITY_URL = "{issue}: {title}\nhttps://ubuntu.com/security/{url_path}"
 SECURITY_DRY_RUN_UA_SERVICE_NOT_ENABLED = """\
--- a/uaclient/messages.py
+++ b/uaclient/messages.py
@@ -73,6 +73,9 @@ SECURITY_AFFECTED_PKGS = (
 )
 USN_FIXED = "{issue} is addressed."
 CVE_FIXED = "{issue} is resolved."
+CVE_FIXED_BY_LIVEPATCH = (
+    "{issue} is resolved by livepatch patch version: {version}."
+)
 SECURITY_URL = "{issue}: {title}\nhttps://ubuntu.com/security/{url_path}"
 SECURITY_DRY_RUN_UA_SERVICE_NOT_ENABLED = """\
 {bold}Ubuntu Pro service: {{service}} is not enabled.
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -564,6 +564,7 @@ def fix_security_issue_id(
 
     if "CVE" in issue_id:
         # Check livepatch status for CVE in fixes before checking CVE api
+        status_stdout = None
         try:
             status_stdout, _ = system.subp(
                 [
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -3,7 +3,6 @@ import enum
 import json
 import os
 import socket
-import subprocess
 import textwrap
 from collections import defaultdict
 from datetime import datetime
@@ -566,16 +565,14 @@ def fix_security_issue_id(
     if "CVE" in issue_id:
         # Check livepatch status for CVE in fixes before checking CVE api
         try:
-            status_stdout = subprocess.run(
+            status_stdout, _ = system.subp(
                 [
                     "canonical-livepatch",
                     "status",
                     "--verbose",
                     "--format=json",
-                ],
-                capture_output=True,
-                encoding="utf-8",
-            ).stdout
+                ]
+            )
         except exceptions.ProcessExecutionError:
             pass
         if status_stdout:
@@ -583,21 +580,22 @@ def fix_security_issue_id(
                 parsed_patch = json.loads(status_stdout)["Status"][0][
                     "Livepatch"
                 ]
-            except (KeyError, IndexError):
-                pass
-            if parsed_patch:
-                fixes = parsed_patch.get("Fixes", [])
-                if any(
-                    fix.Name is issue_id.lower() and fix.Patched
-                    for fix in fixes
-                ):
-                    print(
-                        messages.CVE_FIXED_BY_LIVEPATCH.format(
-                            issue=issue_id,
-                            version=parsed_patch.get("Version", "N/A"),
+
+                if parsed_patch:
+                    fixes = parsed_patch.get("Fixes", [])
+                    if any(
+                        fix.Name is issue_id.lower() and fix.Patched
+                        for fix in fixes
+                    ):
+                        print(
+                            messages.CVE_FIXED_BY_LIVEPATCH.format(
+                                issue=issue_id,
+                                version=parsed_patch.get("Version", "N/A"),
+                            )
                         )
-                    )
-                    return FixStatus.SYSTEM_NON_VULNERABLE
+                        return FixStatus.SYSTEM_NON_VULNERABLE
+            except (ValueError, KeyError, IndexError):
+                pass
 
         try:
             cve = client.get_cve(cve_id=issue_id)
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -1,7 +1,9 @@
 import copy
 import enum
+import json
 import os
 import socket
+import subprocess
 import textwrap
 from collections import defaultdict
 from datetime import datetime
@@ -572,6 +574,23 @@ def fix_security_issue_id(
                     issue_id=issue_id
                 )
             raise exceptions.UserFacingError(msg)
+
+        # Check livepatch status for CVE in fixes
+        status = subprocess.run(
+            ["canonical-livepatch", "status", "--verbose", "--format=json"],
+            capture_output=True,
+            encoding="utf-8",
+        ).stdout
+        if status:
+            parsed_patch = json.loads(status)["Status"][0]["Livepatch"]
+            if cve.id.lower() in parsed_patch.get("Fixes", ""):
+                print(
+                    messages.CVE_FIXED_BY_LIVEPATCH.format(
+                        issue=cve.id,
+                        version=parsed_patch.get("Version", "N/A"),
+                    )
+                )
+
         affected_pkg_status = get_cve_affected_source_packages_status(
             cve=cve, installed_packages=installed_packages
         )
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -584,7 +584,7 @@ def fix_security_issue_id(
                 if parsed_patch:
                     fixes = parsed_patch.get("Fixes", [])
                     if any(
-                        fix.Name is issue_id.lower() and fix.Patched
+                        fix["Name"] is issue_id.lower() and fix["Patched"]
                         for fix in fixes
                     ):
                         print(
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -564,6 +564,41 @@ def fix_security_issue_id(
     }
 
     if "CVE" in issue_id:
+        # Check livepatch status for CVE in fixes before checking CVE api
+        try:
+            status_stdout = subprocess.run(
+                [
+                    "canonical-livepatch",
+                    "status",
+                    "--verbose",
+                    "--format=json",
+                ],
+                capture_output=True,
+                encoding="utf-8",
+            ).stdout
+        except exceptions.ProcessExecutionError:
+            pass
+        if status_stdout:
+            try:
+                parsed_patch = json.loads(status_stdout)["Status"][0][
+                    "Livepatch"
+                ]
+            except (KeyError, IndexError):
+                pass
+            if parsed_patch:
+                fixes = parsed_patch.get("Fixes", [])
+                if any(
+                    fix.Name is issue_id.lower() and fix.Patched
+                    for fix in fixes
+                ):
+                    print(
+                        messages.CVE_FIXED_BY_LIVEPATCH.format(
+                            issue=issue_id,
+                            version=parsed_patch.get("Version", "N/A"),
+                        )
+                    )
+                    return FixStatus.SYSTEM_NON_VULNERABLE
+
         try:
             cve = client.get_cve(cve_id=issue_id)
             usns = client.get_notices(details=issue_id)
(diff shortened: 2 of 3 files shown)


uaclient/security.py
```diff
@@ -572,6 +574,23 @@ def fix_security_issue_id(
                     issue_id=issue_id
                 )
             raise exceptions.UserFacingError(msg)
+
+        # Check livepatch status for CVE in fixes
```
orndorffgrant, line 578: We can do this check before we even look up the CVE in the API. I suggest moving this block of code up to line 567.

uaclient/security.py
```diff
@@ -572,6 +574,23 @@ def fix_security_issue_id(
...
+        status = subprocess.run(
+            ["canonical-livepatch", "status", "--verbose", "--format=json"],
+            capture_output=True,
+            encoding="utf-8",
+        ).stdout
```
orndorffgrant, line 583: We have a subprocess helper that includes logging - let's use that here, and we can just move on if there's an error. I think that should catch the case where canonical-livepatch isn't installed.
```suggestion
        try:
            status_stdout, _ = util.subp(
                ["canonical-livepatch", "status", "--verbose", "--format=json"],
            )
        except exceptions.ProcessExecutionError:
            pass
```

uaclient/security.py
```diff
@@ -572,6 +574,23 @@ def fix_security_issue_id(
...
+        if status:
+            parsed_patch = json.loads(status)["Status"][0]["Livepatch"]
+            if cve.id.lower() in parsed_patch.get("Fixes", ""):
+                print(
+                    messages.CVE_FIXED_BY_LIVEPATCH.format(
+                        issue=cve.id,
+                        version=parsed_patch.get("Version", "N/A"),
+                    )
+                )
```
orndorffgrant, line 592: If we get here, that means the CVE is fixed, so we can end early with a 
```python
return FixStatus.SYSTEM_NON_VULNERABLE
```

orndorffgrant:
@dixonmyles also there's a small logistical step to do before we can merge this. Please create a [Launchpad](https://launchpad.net/) account if you don't already have one and fill out the [Canonical contributor agreement form](https://ubuntu.com/legal/contributors/agreement). You can enter my name (Grant Orndorff) in the "Canonical Project Manager or contact" field.

uaclient/security.py
```diff
@@ -572,6 +574,23 @@ def fix_security_issue_id(
...
+
+        # Check livepatch status for CVE in fixes
+        status = subprocess.run(
+            ["canonical-livepatch", "status", "--verbose", "--format=json"],
+            capture_output=True,
+            encoding="utf-8",
+        ).stdout
+        if status:
+            parsed_patch = json.loads(status)["Status"][0]["Livepatch"]
+            if cve.id.lower() in parsed_patch.get("Fixes", ""):
```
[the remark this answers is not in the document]
dixonmyles: Thanks! Was having difficulty finding documentation on the output of the livepatch command itself as my personal Linux box isn't missing any patches

uaclient/security.py
dixonmyles: I tried implementing `util.subp` as you have shown here, but got an error that there was no `subp` attribute of the `uaclient.util` module. Perhaps there's a different canonical repo where that is available? I searched through the rest of the `ubuntu-advantage-client` repo and didn't see any reference to a `util.subp` call, while I do see `subprocess.run()` calls.

As such, I left this as a call to `subprocess.run()` since there isn't a `util.subp` to use. Happy to pull a different branch or create this util method if desired! I did go ahead and rename the variable to `status_stdout` as you suggested here to better represent what the variable was capturing, and included a try/except block that still raises the same exception.

dixonmyles:
I believe I've addressed all the feedback you provided, happy to continue iterating! I did set up that account and filled out the form, sorry about missing that step.

uaclient/security.py
dixonmyles: Did some more digging and found the `system.subp` helper which I believe is what you were referring to before, and implemented it instead of `subprocess.run`

uaclient/security.py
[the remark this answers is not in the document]
orndorffgrant: Yep the documentation is a bit lacking there. For future reference, I got that example by creating a 16.04/xenial vm with [multipass](https://multipass.run/) (`multipass launch xenial -n dev-x && multipass shell dev-x`) and enabling livepatch on it. Since the xenial image has an old kernel, livepatch immediately patched it.

uaclient/security.py
orndorffgrant: Ah my bad :facepalm: we just moved that to the `system` module. Good job finding it even though I pointed you in the wrong direction!

We use `subprocess.run` in a couple places in our integration testing code, but we stick to `system.subp` in the actual cli code.

orndorffgrant:
Thanks for signing the CLA! I just confirmed that your name is there.

No worries on missing that at first - we need to properly document how to do that because it's not clear from the readme or contributing docs or the PR template. We're probably going to add [this github action](https://github.com/canonical/has-signed-canonical-cla) to make it a little easier.

uaclient/security.py
```diff
@@ -562,6 +563,40 @@ def fix_security_issue_id(
...
+        if status_stdout:
+            try:
+                parsed_patch = json.loads(status_stdout)["Status"][0][
+                    "Livepatch"
+                ]
+
+                if parsed_patch:
+                    fixes = parsed_patch.get("Fixes", [])
+                    if any(
+                        fix.Name is issue_id.lower() and fix.Patched
```
orndorffgrant, line 587: `fix` is a dict so we need to access the values with `["key"]`. Also for string comparison we need to use `==` because `is` will only return true if the strings are the same object in memory (and therefore may return false even if the contents are the same)
```suggestion
                        fix["Name"] == issue_id.lower() and fix["Patched"]
```

uaclient/messages.py
```diff
@@ -73,6 +73,9 @@ class TxtColor:
 )
 USN_FIXED = "{issue} is addressed."
 CVE_FIXED = "{issue} is resolved."
+CVE_FIXED_BY_LIVEPATCH = (
+    "{issue} is resolved by livepatch patch version: {version}."
```
orndorffgrant, line 77: Let's add the green check in front of this. (The above strings USN_FIXED and CVE_FIXED are misleading in not having the check)
```suggestion
    OKGREEN_CHECK + " {issue} is resolved by livepatch patch version: {version}."
```

orndorffgrant (commented):
Thanks for iterating! Just some small fixes left and the implementation looks good.

Then the only other thing is that we should add a unit test for this new path. For now I think a "success case" test is enough. Can you add a test after line 2280 of `test_security.py` that asserts `fix_security_issue_id` returns `FixStatus.SYSTEM_NON_VULNERABLE` when `system.subp` is mocked to return a json string with the given cve in the Fixes list?

I can help and/or provide more specific suggestions if you need - just ask. Our tests are a bit messy :upside_down_face:

dixonmyles:
Hey @orndorffgrant, definitely running into some confusion with writing the test case for this. I _think_ I understand how the testing library works, and how it expects mock values to be provided, but I'm running into trouble with the subprocess portion of the test. I tried looking at some of the other test cases that also have `system.subp` calls for reference but alas, no dice. Below is the test case I have written so far, perhaps you'll notice something obvious I'm missing?

```python
class TestFixSecurityIssueId:
    @pytest.mark.parametrize(
        "issue_id,livepatch_status,exp_ret",
        (
            (
                "cve-2013-1798",
                {
                    "Status": [
                        {
                            "Kernel": "4.4.0-210.242-generic",
                            "Running": True,
                            "Livepatch": {
                                "CheckState": "checked",
                                "State": "applied",
                                "Version": "87.1",
                                "Fixes": [
                                    {
                                        "Name": "cve-2013-1798",
                                        "Description": "Andrew Honig reported a flaw in the way KVM (Kernel-based Virtual \nMachine) emulated the IOAPIC. A privileged guest user could exploit \nthis flaw to read host memory or cause a denial of service (crash the \nhost).",
                                        "Bug": "",
                                        "Patched": True,
                                    }
                                ],
                            },
                        }
                    ]
                },
                FixStatus.SYSTEM_NON_VULNERABLE,
            ),
        ),
    )
    @mock.patch("uaclient.system.subp")
    def test_patched_msg_when_issue_id_fixed_by_livepatch(
        self,
        subp,
        issue_id,
        livepatch_status,
        exp_ret,
        FakeConfig,
    ):
        """fix_security_id returns system not vulnerable when issue_id fixed by livepatch"""
        subp.return_value = livepatch_status, ""
        assert exp_ret == fix_security_issue_id(FakeConfig(), issue_id)
...
```

If needed, I can provide the test failure output I'm seeing as well, it's long so I didn't include it here.

dixonmyles:
Disregard my last comment, I realized I needed to add a patch decorator within my actual test function to get the `query_installed_source_pkg_versions` function available to the `fix_security_issue_id` call

dixonmyles:
@orndorffgrant was hoping you could provide some insight or suggested next step on this workflow item that keeps failing. I ran `tox` on my local and all the test cases passed. If I understand the output correctly it seems like this failing case is when running the `py36` on Ubuntu 18.04?

uaclient/security.py
```diff
@@ -562,6 +563,40 @@ def fix_security_issue_id(
...
+        if status_stdout:
+            try:
+                parsed_patch = json.loads(status_stdout)["Status"][0][
+                    "Livepatch"
+                ]
+
+                if parsed_patch:
+                    fixes = parsed_patch.get("Fixes", [])
+                    if any(
+                        fix["Name"] is issue_id.lower() and fix["Patched"]
```
renanrodrigo, line 587: Hello, @dixonmyles 

I think I saw something related to your test failure:
This comparison will never work, because `is` will go for the obj reference, and it will never be the same. So, you'd need to change it to:

```suggestion
                        fix["Name"] == issue_id.lower() and fix["Patched"]
```

Because this check fails, it will not have the print+return behavior expected in the block, and will continue processing. 

Then, on lines 601-602 there is a network leak - the code actually goes for the CVE/USN provided, and since you provided a real one as an example, it will return proper information about USNs and packages related.
The USNs related to this CVE won't have released packages on Jammy (my system, for instance), but on Bionic they do, and that's why you can't see the error locally but can on the CI: Our CI is actually running the test on a Bionic system.

Applying the aforementioned change fixes the test, as it returns early as it should with the given mocked information.

renanrodrigo: @lucasmoura @orndorffgrant there you have the LSN without a source link that Grant mentioned yesterday :D
Besides that, I wonder what do we lose for not being able to run the Xenial tests on Xenial... (And glad we changed CI to cover that, if running all from same base we'd never see this - aha)

dixonmyles: D’oh! Grant actually mentioned the same thing about is vs == and I completely spaced it! Given that the test cases all pass locally for me, does that mean 22.04 compares it fine but not 18.04? Or was it just coincidence that the object references were the same locally?

Either way, thanks for the nudge! Two pairs of eyes are better than one, and this has been a great exercise! I’ll get that fix to the test case committed.

orndorffgrant: Nice! I think what happened is that the `is` check did fail on 22.04 so it didn't return early and the rest of the function executed. But because `cve-2013-1798` doesn't affect the kernel on 22.04 the function still eventually returned `FixStatus.SYSTEM_NOT_VULNERABLE`. So the test passed, but not because of the code added in this PR :upside_down_face:

dixonmyles: The joys of software 😂

uaclient/security.py
```diff
@@ -562,6 +563,40 @@ def fix_security_issue_id(
...
+                [
+                    "canonical-livepatch",
+                    "status",
+                    "--verbose",
+                    "--format=json",
+                ]
+            )
+        except exceptions.ProcessExecutionError:
+            pass
+        if status_stdout:
```
lucasmoura, line 578: Maybe I am missing something, but if we have an exception when calling the `canonical-livepatch` command, wouldn't the `status_stdout` variable be undefined here ?

We can confirm that by adding a test where raise an exception for that command and see how it proceeds. To make that test easier, we could create a function for that new livepatch check instead and then just test the function directly for the exception scenarios and not the whole `fix_security_issue_id`

dixonmyles: Hey, @lucasmoura thanks for the feedback! The direction I'd received from @orndorffgrant was that just proceeding with the usual flow for `fix_security_issue_id` if an exception is raised was fine. That's why I had the `except exceptions.ProcessExecutionError: pass` and `if status_stdout:` checks.

If the group thinks a separate function with separate tests is the better approach and within the scope of the issue, happy to pursue that line of thought!

orndorffgrant:
Thanks @dixonmyles I'm just going to run some of our integration tests to check for regressions and if all is good I think we can merge!

uaclient/security.py
```diff
@@ -562,6 +563,40 @@ def fix_security_issue_id(
     }
 
     if "CVE" in issue_id:
+        # Check livepatch status for CVE in fixes before checking CVE api
+        try:
```
renanrodrigo, line 567: @lucasmoura is right. If livepatch fails, the variable `status_stdout` will not exist.
You need to create it before the `try` block`.

```suggestion
        status_stdout = None
        try:
```

Easily reproducible:
- Launch an LXD container, don't enable livepatch
- install the client from this branch
- run `ua fix cve-2013-1798`

I'd suggest a test covering this case.

uaclient/security.py
orndorffgrant: Ah sorry I missed this @dixonmyles. I think we can skip the refactor to move this into its own function for now, but @lucasmoura is right about the `status_stdout` variable. When the `canonical-livepatch` command fails, then we get this error:
```
UnboundLocalError: local variable 'status_stdout' referenced before assignment
```
The simplest fix to this is to define the variable before the `try` block. I suggest adding 
```python
status_stdout = None
```
right before the `try` on line 567

uaclient/security.py
orndorffgrant: @renanrodrigo ha! we commented the same thing at the same time :upside_down_face: 

@dixonmyles the extra unit test for this case is optional, since this is already caught by our integration tests.

uaclient/security.py
dixonmyles: @orndorffgrant @lucasmoura thank you both! I think given the back and forth, and the comment from @renanrodrigo, I'll look into adding a test case for this path as well

uaclient/security.py
dixonmyles: Thanks, Renan! @orndorffgrant should I get that additional test case created before this gets approved?

dixonmyles: Great, in that case I’ll make that change to add the variable declaration outside of the try block then. Unfortunately I kinda nuked my Linux install when expanding my root partition (to get enough disk space for multipass to work 🙃), so in the interest of not making you all wait on me fixing that I’ll leave the new test case out

orndorffgrant (approved):
Thank you for this feature @dixonmyles! I think it will be useful to many. :smile: 

Integration tests passed for me. Sorry for the uncoordinated reviews today.

Congratulations on your first contribution here! :tada: :tada: 

Squashing and merging now

dixonmyles:
Thanks, @orndorffgrant! Glad to help, and thanks to @renanrodrigo and @lucasmoura for the feedback as well! This was a fun experience and I'll keep an eye out for other open issues 👀

Issue #2159, opened 2022-07-12
'ua status' output inconsistent between formats


seb128:
On Ubuntu 22.04

$ ua version
27.9~22.04.1

$ ua status | grep fips
fips             yes                n/a                NIST-certified core packages
fips-updates     yes                n/a                NIST-certified core packages with priority security updates

$ ua status --format=yaml | grep fips
$

Issue #2156, opened 2022-07-12
'ua status' includes unavailable services in the default output


seb128:
On Ubuntu 22.04

$ ua version
27.9~22.04.1

$ ua status -h 
...
If the --all flag is set, beta and unavailable services are also
listed in the output.

^ suggests that without the option the unavailable services shouldn't be included

but 

$ ua status | grep usg
usg              yes                n/a                Security compliance and audit tools

$ sudo ua enable usg
One moment, checking your subscription first
Ubuntu Security Guide is not available for Ubuntu 22.04 LTS (Jammy Jellyfish).

Since 'usg' isn't available it shouldn't be including according the help description

PR #2183, opened 2022-08-01
status: don't show unavailable services by default

lucasmoura:
## Proposed Commit Message
status: don't show unavailable services by default

We are not only showing unavailable services if the user provides
the --all flag to status.

Fixes: #2156
Fixes: #2159

## Test Steps

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_status.feature
+++ b/features/attached_status.feature
@@ -18,3 +18,128 @@ Feature: Attached status
            | xenial  |
            | jammy   |
            | kinetic |
+
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attached status in a ubuntu machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I verify root and non-root `pro status` calls have the same output
+        And I run `pro status` as non-root
+        Then stdout matches regexp:
+        """
+        SERVICE         +ENTITLED +STATUS   +DESCRIPTION
+        cc-eal          +yes      +disabled +Common Criteria EAL2 Provisioning Packages
+        cis             +yes      +disabled +Security compliance and audit tools
+        esm-apps        +yes      +enabled  +Extended Security Maintenance for Applications
+        esm-infra       +yes      +enabled  +Extended Security Maintenance for Infrastructure
+        fips            +yes      +disabled +NIST-certified core packages
+        fips-updates    +yes      +disabled +NIST-certified core packages with priority security updates
+        ros             +yes      +disabled +Security Updates for the Robot Operating System
+        ros-updates     +yes      +disabled +All Updates for the Robot Operating System
+
+        Enable services with: pro enable <service>
+        """
+        When I verify root and non-root `pro status --all` calls have the same output
+        And I run `pro status --all` as non-root
+        Then stdout matches regexp:
+        """
+        SERVICE         +ENTITLED +STATUS   +DESCRIPTION
+        cc-eal          +yes      +disabled +Common Criteria EAL2 Provisioning Packages
+        cis             +yes      +disabled +Security compliance and audit tools
+        esm-apps        +yes      +enabled  +Extended Security Maintenance for Applications
+        esm-infra       +yes      +enabled  +Extended Security Maintenance for Infrastructure
+        fips            +yes      +disabled +NIST-certified core packages
+        fips-updates    +yes      +disabled +NIST-certified core packages with priority security updates
+        livepatch       +yes      +n/a      +Canonical Livepatch service
+        realtime-kernel +yes      +n/a      +Beta-version Ubuntu Kernel with PREEMPT_RT patches
+        ros             +yes      +disabled +Security Updates for the Robot Operating System
+        ros-updates     +yes      +disabled +All Updates for the Robot Operating System
+
+        Enable services with: pro enable <service>
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
+
+    @series.focal
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attached status in a ubuntu machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I verify root and non-root `pro status` calls have the same output
+        And I run `pro status` as non-root
+        Then stdout matches regexp:
+        """
+        SERVICE         +ENTITLED +STATUS   +DESCRIPTION
+        esm-apps        +yes      +enabled  +Extended Security Maintenance for Applications
+        esm-infra       +yes      +enabled  +Extended Security Maintenance for Infrastructure
+        fips            +yes      +disabled +NIST-certified core packages
+        fips-updates    +yes      +disabled +NIST-certified core packages with priority security updates
+        usg             +yes      +disabled +Security compliance and audit tools
+
+        Enable services with: pro enable <service>
+        """
+        When I verify root and non-root `pro status --all` calls have the same output
+        And I run `pro status --all` as non-root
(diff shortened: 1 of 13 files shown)


features/attached_status.feature
```diff
@@ -18,3 +18,123 @@ Feature: Attached status
            | xenial  |
            | impish  |
            | jammy   |
+
+    @series.xenial
+    @series.bionic
+    @series.focal
```
orndorffgrant, line 24:
```suggestion
```

features/attached_status.feature
```diff
@@ -18,3 +18,123 @@ Feature: Attached status
...
+        usg             +yes      +disabled +Security compliance and audit tools
+
+        Enable services with: pro enable <service>
+        """
+
+        Examples: ubuntu release
+           | release |
+           | focal   |
+
+    @series.all
```
orndorffgrant, line 106:
```suggestion
    @series.jammy
```

features/attached_status.feature
```diff
@@ -18,3 +18,123 @@ Feature: Attached status
            | xenial  |
            | impish  |
            | jammy   |
+
+    @series.xenial
+    @series.bionic
+    @series.focal
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attached status in a ubuntu machine
```
orndorffgrant, line 26: I really like the idea of consolidating our status checking output into tests like these! I might expand on this further in a future PR.

orndorffgrant (approved):
Looks great! Just some nits on the new integration tests

orndorffgrant:
Ah there are some CI failures though

lucasmoura:
@orndorffgrant done, I have fixed the tests and addressed the points you made

features/unattached_status.feature
```diff
@@ -40,155 +40,389 @@ Feature: Unattached status
...
+
+        Examples: ubuntu release
+           | release |
+           | focal   |
+
+    @series.jammy
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Unattached status in a ubuntu machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `pro status` as non-root
```
orndorffgrant, line 248: I appreciate the thoroughness of these additions! And like we talked about earlier, I think we should extend this idea to all machine_types x all series, but thats for a future PR.

I think we can simplify it in one way though. Which is that non-root vs root status should always be the same when unattached right? (and after #2085 they will always be the same when attached)

So I think there should maybe be one test that verifies they are the same, but I don't think we need both for every series - what do you think?

features/unattached_status.feature
```diff
@@ -216,15 +450,69 @@ Feature: Unattached status
...
-           | focal   | yes      | no     |     | yes           | yes  | yes       | no  | yes       | usg | no              |
-           | impish  | no       | no     | cis | no            | no   | no        | no  | no        |     | no              |
-           | jammy   | yes      | no     |     | no            | no   | yes       | no  | yes       | usg | yes             |
+           | release |
+           | focal   |
 
+    @series.jammy
+    @uses.config.machine_type.lxd.container
+    @uses.config.contract_token
+    Scenario Outline: Simulate status in a ubuntu machine
```
orndorffgrant, line 459: I'm not convinced we need to test full simulate status output for each series. Isn't the availability logic the same as normal status?

lucasmoura: It should yes, but do you think should only test a subset of the series ?

features/unattached_status.feature
lucasmoura: We could create a step that checks if they are equal and use only `non-root` calls for the tests. What do you think ?

renanrodrigo: We can add it to the current run step, extending `as non-root` and `with sudo` with another option, who would run the command as both and make sure it is the same I guess.

features/unattached_status.feature
renanrodrigo: IMHO testing all series covers all use cases without us having to make any previous assumption. I think it is beneficial for the completeness of the integration tests.

uaclient/cli.py
```diff
@@ -1549,6 +1549,7 @@ def action_status(args, *, cfg):
     if not cfg:
         cfg = config.UAConfig()
     show_beta = args.all if args else False
+    show_all = args.all if args else False
```
renanrodrigo, line 1552: With this changeset, `show_beta` and `show_all` are coming from the same flag (`--all`) and there is no difference between them. I think we should either:
1. Create a new flag, `--beta`, and let it take care only of beta services, and then `--all` would mean unavailable+beta services, or
2. Remove the `show_beta` variable, as it is only depending on show_all here anyway when running from the CLI. Then, the only way to show beta without showing the unavailable services would be the `allow_beta` feature.

I would be inclined to 2 for simplicity purposes, but this is not a strong opinion.

features/unattached_status.feature
orndorffgrant: Okay! mine wasn't a strong opinion - let's keep as is

features/unattached_status.feature
orndorffgrant: I like the idea of a new step, maybe something like:
```
When I run ua status `--flags --go --here`
```
That runs the command as both root and non-root, asserts they have equal stdout,stderr, and then sets the appropriate stdout/stderr variables on the behave context

uaclient/cli.py
orndorffgrant: I agree with (2). No need to arg-drill two separate variables equal to the same value.

When we get to the function that handles `allow_beta` from the config, we can give that the value of `show_all` for `show_beta` (I don't remember the exact place where this happens, so that might not totally make sense)

renanrodrigo: It does - `_handle_beta_services` is receving `show_all or show_beta` in the changeset, so yeah it works

features/unattached_status.feature
lucasmoura: I think I would prefer something like `When I verify root and non-root status `flags` calls have the same output`. The idea is just to be as explicit as possible. But let me know what you all think

uaclient/cli.py
lucasmoura: That makes sense, I will change that

features/unattached_status.feature
orndorffgrant: That works for me!

lucasmoura:
@orndorffgrant @renanrodrigo done

features/steps/steps.py
```diff
@@ -1157,6 +1157,20 @@ def when_i_create_local_ppas(context, release, next_release):
     )
 
 
+@when("I verify root and non-root `{cmd}` calls have the same output")
+def status_root_vs_nonroot_comparison(context, cmd):
```
renanrodrigo, line 1161: I'd just rename the function (and variables) as the functionality isn`t only related to status - can be used with any command actually.

uaclient/tests/test_cli_status.py
```diff
@@ -607,26 +678,7 @@ def test_attached_formats(
...
-            for service in RESPONSE_AVAILABLE_SERVICES
-            if not service["available"]
-        ]
-
-        filtered_services = [
-            service
-            for service in services
-            if service["name"] not in inapplicable_services
-        ]
+        services = SERVICES_JSON_ALL if use_all else SERVICES_JSON
```
renanrodrigo, line 681: Testing locally, I have livepatch enabled, and this test fails because I get the message that Livepatch blocks FIPS in the json/yaml output... so it is actually calling status, which imho is a leak. Is there a sensible way to mock the status call to make sure this test is not dependent on the actual system?

renanrodrigo (approved):

orndorffgrant (approved):
I think this just needs a conflict resolved and then is good to go!

PR #2184, opened 2022-08-01
wip: refactor: get kernel from /proc/version_signature

orndorffgrant:
This is to become more compatible with how livepatch parses the kernel
version

## Test Steps

Everything should still work

## Desired commit type

## Checklist


uaclient/system.py
```diff
@@ -17,6 +17,123 @@
...
+@lru_cache(maxsize=None)
+def get_kernel_info() -> KernelInfo:
+    uname_release = os.uname().release
+
+    proc_version_signature_full = ""
+    proc_version_signature_version = ""
+    try:
+        proc_version_signature_full = load_file("/proc/version_signature")
+        proc_version_signature_version = proc_version_signature_full.split(
+            " "
```
lucasmoura, line 78: I think this is default parameter for split, we can omit it if we want

uaclient/system.py
```diff
@@ -17,6 +17,123 @@
...
+
+    if proc_version_signature_version != "":
+        match = re.match(
+            RE_KERNEL_PROC_VERSION_SIGNATURE, proc_version_signature_version
+        )
+        if match is None:
+            raise exceptions.UserFacingError(
+                "Failed to parse kernel: {}".format(
+                    proc_version_signature_version
+                )
```
lucasmoura, line 94: Since this error could be surfaced when using a json format for a cli command, I think we should add a `msg_code` to it. Additionally, we can also move it to the `messages` module

uaclient/system.py
```diff
@@ -17,6 +17,123 @@
...
+        abi = match.group("abi")
+        subrev = match.group("subrev")
+        hwerev = match.group("hwerev") or ""
+        flavor = match.group("flavor")
+    else:
+        match = re.match(RE_KERNEL_UNAME, uname_release)
+        if match is None:
+            raise exceptions.UserFacingError(
+                "Failed to parse kernel: {}".format(uname_release)
+            )
```
lucasmoura, line 106: Same message consideration here

uaclient/system.py
```diff
@@ -17,6 +17,123 @@
...
+        abi = match.group("abi")
+        subrev = ""
+        hwerev = ""
+        flavor = match.group("flavor")
+
+    version_split_match = re.match(RE_KERNEL_VERSION_SPLIT, version)
+    if version_split_match is None:
+        raise exceptions.UserFacingError(
+            "Failed to split kernel version: {}".format(version)
+        )
```
lucasmoura, line 117: Same message consideration here

uaclient/system.py
```diff
@@ -17,6 +17,123 @@
...
+RE_KERNEL_PROC_VERSION_SIGNATURE = (
+    r"^"
+    r"(?P<version>[\d]+[.-][\d]+[.-][\d]+)"
+    r"-"
+    r"(?P<abi>[\d]+)"
+    r"[.-]"
+    r"(?P<subrev>[\d]+)"
+    r"(~(?P<hwerev>[\d.]+))?"
+    r"-"
+    r"(?P<flavor>[A-Za-z0-9_-]+)"
```
lucasmoura, line 38: Just double checking, the kernel will always have the `abi` and `flavor` information on the version signature ?

uaclient/entitlements/base.py
```diff
@@ -590,10 +583,8 @@ def applicability_status(
                 )
                 return (ApplicabilityStatus.INAPPLICABLE, invalid_msg)
 
-            if not match:
-                return ApplicabilityStatus.INAPPLICABLE, invalid_msg
-            kernel_major = int(match.group("major"))
-            kernel_minor = int(match.group("minor"))
+            kernel_major = int(kernel_info.major)
```
lucasmoura, line 586: Can't we make that type casting directly when creating the `KernelInfo` object ?

uaclient/system.py
orndorffgrant: Looks like with no arg it splits on "runs of consecutive whitespace" but that works too and maybe is even better if the format of that file slightly changes ever. I'll change it

uaclient/system.py
orndorffgrant: Excellent point

uaclient/system.py
orndorffgrant: Yes I believe so. This regex was taken from here: https://github.com/canonical/livepatch-server/blob/a33d53646acb739de4f4bcd8c5e7cf3f9bd8ec6f/params/kernel.go

uaclient/entitlements/base.py
orndorffgrant: Yep good idea, we can do this for major minor and patch at least

orndorffgrant:
@lucasmoura updated!

lucasmoura (approved):

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -361,90 +408,210 @@ def test_can_enable_true_on_entitlement_inactive(
...
+                True,
+            ),
         ),
     )
+    @mock.patch("uaclient.system.get_kernel_info")
+    @mock.patch(
+        "uaclient.system.get_platform_info",
+        return_value=PLATFORM_INFO_SUPPORTED,
+    )
     def test_can_enable_false_on_unsupported_min_kernel_version(
```
renanrodrigo, line 588: Can't `test_can_enable_false_on_unsupported_kernel_min_version` be merged with this test? I don't see a difference between those

renanrodrigo (approved):
LGTM just a small test suggestion

uaclient/entitlements/tests/test_livepatch.py
orndorffgrant: Nice catch! I'll combine the cases.

PR #2186, opened 2022-08-02
Warn about new versions (SC-552)

renanrodrigo:
Warn users about a new version of the client being available.
If executed from the CLI, print a warning to stderr.
If it's an API call, include a warning in the JSON.

## Desired commit type

## Checklist

### DIFF
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -218,3 +218,61 @@ Feature: Command behaviour when unattached
           | kinetic | disable  |
           | jammy   | enable   |
           | jammy   | disable  |
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Check for newer versions of the client in an ubuntu machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        #  Make sure we have a fresh, just rebooted, environment
+        When I reboot the machine
+        Then I verify that no files exist matching `/tmp/ubuntu-advantage/candidate-version`
+        When I run `pro status` as non-root
+        Then I will see the following on stderr
+        """
+        """
+        And I verify that files exist matching `/tmp/ubuntu-advantage/candidate-version`
+        # We forge a candidate to see results
+        When I delete the file `/tmp/ubuntu-advantage/candidate-version`
+        And I create the file `/tmp/ubuntu-advantage/candidate-version` with the following
+        """
+        99.9.9
+        """
+        And I run `chmod 777 /tmp/ubuntu-advantage/candidate-version` with sudo
+        And I run `pro status` as non-root
+        Then stderr matches regexp:
+        """
+        .*\[info\].* A new version is available: 99.9.9
+        Please run:
+            sudo apt-get install ubuntu-advantage-tools
+        to get the latest version with new features and bug fixes.
+        """
+        When I run `pro status --format json` as non-root
+        Then I will see the following on stderr
+        """
+        """
+        When I run `pro config show` as non-root
+        Then stderr matches regexp:
+        """
+        .*\[info\].* A new version is available: 99.9.9
+        Please run:
+            sudo apt-get install ubuntu-advantage-tools
+        to get the latest version with new features and bug fixes.
+        """
+        When I run `pro api u.pro.version.v1` as non-root
+        Then I will see the following on stderr
+        """
+        """
+        When I run `apt-get update` with sudo
+        # apt-get update will bring a new candidate, which is the current installed version
+        And I run `pro status` as non-root
+        Then I will see the following on stderr
+        """
+        """
+
+        Examples: ubuntu release
+          | release |
+          | xenial  |
+          | bionic  |
+          | focal   |
+          | jammy   |
+          | kinetic |
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -259,6 +259,16 @@ Feature: Command behaviour when unattached
         to get the latest version with new features and bug fixes.
         """
         When I run `pro api u.pro.version.v1` as non-root
+        Then stdout matches regexp
+        """
+        \"code\": \"new-version-available\"
+        """
+        When I verify that running `pro api u.pro.version.inexistent` `as non-root` exits `1`
+        Then stdout matches regexp
+        """
+        \"code\": \"new-version-available\"
+        """
+        When I run `pro api u.pro.version.v1` as non-root
         Then I will see the following on stderr
         """
         """
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -570,3 +570,14 @@ def setup_unauthenticated_repo(
         apt_origin,
         "never",
     )
+
+
+def compare_versions(version1: str, version2: str, relation: str) -> bool:
+    """Return True comparing version1 to version2 with the given relation."""
+    try:
+        system.subp(
+            ["dpkg", "--compare-versions", version1, relation, version2]
+        )
+        return True
+    except exceptions.ProcessExecutionError:
+        return False
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -30,6 +30,9 @@ APT_METHOD_HTTPS_FILE = "/usr/lib/apt/methods/https"
 CA_CERTIFICATES_FILE = "/usr/sbin/update-ca-certificates"
 APT_PROXY_CONF_FILE = "/etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy"
 
+APT_UPDATE_SUCCESS_STAMP_PATH = "/var/lib/apt/periodic/update-success-stamp"
+APT_LISTS_PATH = "/var/lib/apt/lists"
+
 # Since we generally have a person at the command line prompt. Don't loop
 # for 5 minutes like charmhelpers because we expect the human to notice and
 # resolve to apt conflict or try again.
@@ -581,3 +584,12 @@ def compare_versions(version1: str, version2: str, relation: str) -> bool:
         return True
     except exceptions.ProcessExecutionError:
         return False
+
+
+def get_apt_cache_time() -> Optional[float]:
+    cache_time = None
+    if os.path.exists(APT_UPDATE_SUCCESS_STAMP_PATH):
+        cache_time = os.stat(APT_UPDATE_SUCCESS_STAMP_PATH).st_mtime
+    elif os.path.exists(APT_LISTS_PATH):
+        cache_time = os.stat(APT_LISTS_PATH).st_mtime
+    return cache_time
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -182,12 +182,22 @@ def run_apt_command(
 
 
 @lru_cache(maxsize=None)
-def run_apt_cache_policy_command(
+def get_apt_cache_policy(
     error_msg: Optional[str] = None,
     env: Optional[Dict[str, str]] = {},
 ) -> str:
     return run_apt_command(
         cmd=["apt-cache", "policy"], error_msg=error_msg, env=env
+    )
+
+
+def get_apt_cache_policy_for_package(
+    package: str,
+    error_msg: Optional[str] = None,
+    env: Optional[Dict[str, str]] = {},
+) -> str:
+    return run_apt_command(
+        cmd=["apt-cache", "policy", package], error_msg=error_msg, env=env
     )
 
 
@@ -207,7 +217,7 @@ def run_apt_update_command(env: Optional[Dict[str, str]] = {}) -> str:
         # Whenever we run an apt-get update command, we must invalidate
         # the existing apt-cache policy cache. Otherwise, we could provide
         # users with incorrect values.
-        run_apt_cache_policy_command.cache_clear()
+        get_apt_cache_policy.cache_clear()
 
     return out
 
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -69,6 +69,16 @@ STATUS_FORMATS = ["tabular", "json", "yaml"]
 
 UA_COLLECT_LOGS_FILE = "ua_logs.tar.gz"
 
+NEW_VERSION_NOTICE = (
+    "\n"
+    + messages.BLUE_INFO
+    + """\
+ A new version is available: {version}
+Please run:
+    sudo apt-get install ubuntu-advantage-tools
+to get the latest version with new features and bug fixes."""
+)
+
 event = event_logger.get_event_logger()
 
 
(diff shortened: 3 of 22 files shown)


uaclient/cli.py
```diff
@@ -1772,6 +1779,17 @@ def action_help(args, *, cfg):
     return 0
 
 
+def _warn_about_new_version(cmd_args=None) -> None:
+    # If no args, then it was called from the main error handler.
+    # We don't want to show this text for the "api" CLI output.
```
orndorffgrant, line 1784: We also don't want to show this for commands with `--format=json` right?

uaclient/defaults.py
```diff
@@ -12,6 +12,8 @@
 DEFAULT_PRIVATE_MACHINE_TOKEN_PATH = (
     DEFAULT_DATA_DIR + PRIVATE_SUBDIR + "/" + MACHINE_TOKEN_FILE
 )
+CANDIDATE_REGEX = r"Candidate: (?P<candidate>.*?)\n"
+CANDIDATE_CACHE_PATH = "/tmp/ua-candidate"
```
orndorffgrant, line 16: I suggest we give ourselves a directory in `/run` because we'll likely want more temporary files in the future and we should avoid polluting the top level of /tmp or /run
```suggestion
CANDIDATE_CACHE_PATH = "/run/ubuntu-advantage/candidate-version"
```

features/unattached_commands.feature
```diff
@@ -218,3 +218,55 @@ Feature: Command behaviour when unattached
...
+        When I delete the file `/tmp/ua-candidate`
+        And I create the file `/tmp/ua-candidate` with the following
+        """
+        99.9.9
+        """
+        When I run `pro status` as non-root
+        Then I will see the following on stderr
+        """
+        *Warning*: There is a new version of the client available (99.9.9
+        ).
```
orndorffgrant, line 242: What's causing this to go onto another line?

uaclient/defaults.py
```diff
@@ -12,6 +12,8 @@
 DEFAULT_PRIVATE_MACHINE_TOKEN_PATH = (
     DEFAULT_DATA_DIR + PRIVATE_SUBDIR + "/" + MACHINE_TOKEN_FILE
 )
+CANDIDATE_REGEX = r"Candidate: (?P<candidate>.*?)\n"
```
orndorffgrant, line 15: I think the regex probably belongs in `version.py`

orndorffgrant (commented):
Looking good!

uaclient/cli.py
renanrodrigo: good point - will hide it

uaclient/defaults.py
renanrodrigo: The problem is that non-root cannot create things in `/run`

features/unattached_commands.feature
renanrodrigo: yeah took some time for me to see: an unintended \n... Will fix

uaclient/defaults.py
orndorffgrant: Ah you're right okay. How about we just create our own directory in tmp then? `/tmp/ubuntu-advantage/candidate-version`?

renanrodrigo: should we go for `ubuntu-advantage` (package name) or `pro-client`?

uaclient/cli.py
renanrodrigo: done

uaclient/defaults.py
renanrodrigo: went for `ubuntu-advantage`, just as the `/etc` one

renanrodrigo:
I will still add a commit here trying to stop this function from leaking in (many) other tests which call the CLI. I will try to create a fixture for it so we don't need to mock it everywhere, and just "unmock" it where it is actually tested.

features/unattached_commands.feature
```diff
@@ -218,3 +218,69 @@ Feature: Command behaviour when unattached
...
+        99.9.9
+        """
+        And I run `chmod 777 /tmp/ubuntu-advantage/candidate-version` with sudo
+        And I run `pro status` as non-root
+        Then I will see the following on stderr
+        """
+        *Warning*: There is a new version of the client available:
+        99.9.9. Make sure to run
+        sudo apt install ubuntu-advantage-tools
+        to upgrade to the latest version and get all new features and bug fixes.
```
orndorffgrant, line 245: Message bikeshedding thread:

I'd like to see the initial warning and version all on one line.
I'm not sure the `*` are necessary, but could be convinced either way
"Make sure to" can be replaced by "Please" IMO
Minor change ideas at the end, but they're very subjective and I don't have a strong opinion there
```suggestion
        Warning: A new version is available: 99.9.9.
        Please run:
            sudo apt install ubuntu-advantage-tools
        to get the latest version with new features and bug fixes.
```

Take or leave as much/little as you'd like from this suggestion

uaclient/apt.py
```diff
@@ -560,3 +566,23 @@ def setup_unauthenticated_repo(
...
+        system.subp(
+            ["dpkg", "--compare-versions", version1, relation, version2]
+        )
+        return True
+    except exceptions.ProcessExecutionError:
+        return False
+
+
+def get_apt_cache_time():
+    cache_time = inf
```
orndorffgrant, line 583: This probably deserves a docstring explaining why `inf` is the default return value

Alternatively, this function could not assume what the caller needs, and just return `None` if it can't find the time. Then the caller can determine if that should be treated as "inf" or 0 or an error

uaclient/apt.py
```diff
@@ -560,3 +566,23 @@ def setup_unauthenticated_repo(
...
+    if os.path.exists(APT_UPDATE_SUCCESS_STAMP_PATH):
+        cache_time = os.stat(APT_UPDATE_SUCCESS_STAMP_PATH).st_mtime
+    elif os.path.exists(APT_LISTS_PATH):
+        cache_time = os.stat(APT_LISTS_PATH).st_mtime
```
orndorffgrant, line 587: Curious: in what cases would one of these files be present but not the other? (I know nothing about these)

uaclient/apt.py
```diff
@@ -183,12 +187,14 @@ def run_apt_command(
...
     error_msg: Optional[str] = None,
     env: Optional[Dict[str, str]] = {},
 ) -> str:
-    return run_apt_command(
-        cmd=["apt-cache", "policy"], error_msg=error_msg, env=env
-    )
+    cmd = ["apt-cache", "policy"]
+    if package:
+        cmd.append(package)
+    return run_apt_command(cmd, error_msg=error_msg, env=env)
```
orndorffgrant, line 197: `apt-cache policy` has drastically different output with/without a package argument. Should we model this as two separate functions in our code?

I could be convinced either way

uaclient/version.py
```diff
@@ -29,8 +39,43 @@ def get_version():
...
+        except Exception:
+            pass
+
+    try:
+        with open(CANDIDATE_CACHE_PATH, "r") as f:
+            return f.read()
+    except Exception:
+        pass
+
+    return ""
```
orndorffgrant, line 74: IMO we should either return `Optional[str]` (so `None` here), or document that it returns `""` if there is no candidate

features/unattached_commands.feature
renanrodrigo: I'll take everything! Except for the warning `*`s: @lucasmoura suggested those on the spec to "highlight" the word and catch the user's attention, and I think it makes sense.

uaclient/apt.py
renanrodrigo: Well some versions of APT didn't have the stamp file - and it is just an empty file that APT will touch after it changes the cache. When the file is not there, we can as a fallback see the APT lists - the raw data that APT actually modifies.

I kinda think that in the systems we care about (Ubuntu, Xenial+) the stamp file will always be there... But I preferred to keep the fallback option when I bald-facedly copied this script from the [Ansible codebase](https://github.com/ansible/ansible/blob/12734fa21c08a0ce8c84e533abdc560db2eb1955/lib/ansible/modules/apt.py#L1012-L1022)

uaclient/apt.py
renanrodrigo: IMHO the idea of the function is just to wrap the call to policy, and mainly cache it so we don't need to run it more than once per execution.
I think we should keep this one function, for simplicity's sake: the return value is a string in both cases, and I don't think we gain much for adding the granularity here.

features/unattached_commands.feature
renanrodrigo: updated - let me know if I should change anything else

lucasmoura: We could also highlight the warning as we do with the commands on `fix`. My idea was just to have a visual cue for the user to not ignore that message. But I don't have a strong opinion here

uaclient/apt.py
lucasmoura: I don't know, I tend to agree with @orndorffgrant here. Because the output is so distinct, I think it would be better for us to have a new function just to be explicit. Additionally, even though this is a minor thing, but I don't think caching will be that beneficial for the `apt-cache policy package` command

features/unattached_commands.feature
```diff
@@ -218,3 +218,70 @@ Feature: Command behaviour when unattached
           | kinetic | disable  |
           | jammy   | enable   |
           | jammy   | disable  |
+
+    @wip
```
lucasmoura, line 222: We can remove this bit

features/unattached_commands.feature
```diff
@@ -218,3 +218,70 @@ Feature: Command behaviour when unattached
...
+        When I delete the file `/tmp/ubuntu-advantage/candidate-version`
+        And I create the file `/tmp/ubuntu-advantage/candidate-version` with the following
+        """
+        99.9.9
+        """
+        And I run `chmod 777 /tmp/ubuntu-advantage/candidate-version` with sudo
+        And I run `pro status` as non-root
+        Then I will see the following on stderr
+        """
+        *Warning*: A new version is available: 99.9.9.
```
lucasmoura, line 243: @orndorffgrant @CalvoM @renanrodrigo should we move this message to the beginning of the output instead ?

uaclient/cli.py
```diff
@@ -1676,6 +1683,23 @@ def action_help(args, *, cfg):
...
 
+def _warn_about_new_version(cmd_args=None) -> None:
+    # If no args, then it was called from the main error handler.
+    # We don't want to show this text for the "api" CLI output,
+    # or for --format json|yaml
+    if (
+        cmd_args
+        and cmd_args.command == "api"
+        or hasattr(cmd_args, "format")
+        and cmd_args.format in ("json", "yaml")
```
lucasmoura, line 1694: We could use `getattr` here with a default value

features/unattached_commands.feature
renanrodrigo: I think it is less invasive to have it at the bottom, but can be convinced otherwise.

uaclient/apt.py
renanrodrigo: oki can do

uaclient/apt.py
orndorffgrant: I see I see. My mind is turning on a way that we could trigger messaging updates based on watching this stamp file rather than on a 12 hour timer... but that's for another PR.

features/unattached_commands.feature
orndorffgrant: I think it is less jarring and more likely to be seen at the bottom.

I've been thinking of pip and npm as we develop this, which both do their update notice at the bottom IIRC.

I just verified with pip and accidentally found out that there actually is a pip package named "nonsense" :shrug: 
```
> pip install nonsense
Collecting nonsense
  Downloading nonsense-0.1.0.tar.gz (1.4 kB)
  Preparing metadata (setup.py) ... done
Building wheels for collected packages: nonsense
  Building wheel for nonsense (setup.py) ... done
  Created wheel for nonsense: filename=nonsense-0.1.0-py3-none-any.whl size=2026 sha256=ad5fddb17be4cfe4459df962013703421b6bc7fddd0080c113654ac35a36982f
  Stored in directory: /home/grant/.cache/pip/wheels/8a/f3/b9/2853d7947af0ac3f214067ffd60ba431349f292eabbada2562
Successfully built nonsense
Installing collected packages: nonsense
Successfully installed nonsense-0.1.0

[notice] A new release of pip available: 22.1.2 -> 22.2.2
[notice] To update, run: pip install --upgrade pip
```

features/unattached_commands.feature
orndorffgrant: I also don't have a strong opinion. We could highlight it by making it a color.

Now that I was playing with pip for the other comment, theirs is a "notice", not a warning, and they have "notice" in blue.
```
[notice] A new release of pip available: 22.1.2 -> 22.2.2
[notice] To update, run: pip install --upgrade pip
```

npm also uses the word "notice" and they color it teal (in my terminal at least, not sure what terminal color it is - maybe "bright green"?)
```
changed 1 package, and audited 2 packages in 2s

found 0 vulnerabilities
npm notice 
npm notice New minor version of npm available! 8.5.1 -> 8.17.0
npm notice Changelog: https://github.com/npm/cli/releases/tag/v8.17.0
npm notice Run npm install -g npm@8.17.0 to update!
npm notice 
```

So maybe we shouldn't even call it a warning? We could say "Info" in blue (since we already have a specific meaning for "Notice" in status output).

Sorry I'm adding fuel to the flaming bike shed. :fire: :bike: :house: :upside_down_face:

renanrodrigo: Blue [info] is there!

uaclient/cli.py
```diff
@@ -69,6 +69,17 @@
...
 
+NEW_VERSION_NOTICE = (
+    "\n"
+    + messages.BLUE_INFO
+    + """\
+ A new version is available: {version}
+Please run:
+    sudo apt-get install ubuntu-advantage-tools
+to get the latest version with new features and bug fixes.
+"""
```
orndorffgrant, line 80: I don't think we need the extra newline
```suggestion
to get the latest version with new features and bug fixes."""
```

uaclient/version.py
```diff
@@ -29,8 +41,44 @@ def get_version():
...
+                os.makedirs(UAC_TMP_PATH, exist_ok=True)
+                with open(CANDIDATE_CACHE_PATH, "w") as f:
+                    f.write(candidate_version)
+                os.chmod(CANDIDATE_CACHE_PATH, 0o777)
+        except Exception:
+            pass
+
+    try:
+        with open(CANDIDATE_CACHE_PATH, "r") as f:
+            return f.read()
```
orndorffgrant, line 73: Let's just strip the output in case a newline or whitespace slips in
```suggestion
            return f.read().strip()
```

orndorffgrant (approved):
LGTM just two minor nits

orndorffgrant (approved):

orndorffgrant:
There are some failing integration tests though

renanrodrigo:
Some of them were flaky, and the ones that are not flaky are also failing on `main` :/

PR #2189, opened 2022-08-05
get arch from lscpu output for livepatch

orndorffgrant:
## Test Steps

On your system or a test system:
- `env PYTHONPATH=1 python3`
- `from uaclient import system`
- `system.get_lscpu_arch()`

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/messages.py
+++ b/uaclient/messages.py
@@ -926,3 +926,8 @@ KERNEL_VERSION_SPLIT_ERROR = FormattedNamedMessage(
     name="kernel-version-split-error",
     msg="Failed to split kernel version: {version}",
 )
+
+LSCPU_ARCH_PARSE_ERROR = NamedMessage(
+    name="lscpu-arch-parse-error",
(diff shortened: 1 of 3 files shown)


renanrodrigo (approved):
I mean the changeset looks good to me but there is an abstract question that bugged me when reading context.* **

Why would livepatch pick the architecture from `lscpu`? As far as I understand, Livepatch uses this information to check compatibility with the Kernel. `lscpu` is reporting the processor architecture, while `uname` is reporting the architecture the kernel was compiled for...

Shouldn't livepatch use `uname` to determine the arch?

* Maybe I'm just dumb and this does not make sense
** Of course this does not block us from landing this and continuing the integration in any way

orndorffgrant:
These are good points! And I don't know the answer to your questions. I double checked that livepatch is using `lscpu` so we'll continue with this (as discussed out-of-band). If that changes then we will of course follow suit.

PR #2197, opened 2022-08-10
enable: access-only flag for usecases where auto-install is undesired

orndorffgrant:
## Test Steps

run new integration tests

## Desired commit type

## Checklist

### DIFF
--- a/docs/source/howtoguides/enable_realtime_kernel.md
+++ b/docs/source/howtoguides/enable_realtime_kernel.md
@@ -1,0 +1,62 @@
+# How to enable Real-Time Kernel
+
+> **Warning**:
+> Real-Time Kernel is currently in beta.
+
+Real-Time Kernel is supported only on 22.04. For more information on it,
+please see https://ubuntu.com/realtime-kernel
+
+## Enable and auto-install
+
+To enable it through UA, please run:
+
+```console
+$ sudo ua enable realtime-kernel --beta
+```
+
+You'll need to acknowledge a warning and then you should see output like the following, indicating that the Real-Time Kernel package has been installed.
+
+```
+One moment, checking your subscription first
+The real-time kernel is a beta version of the 22.04 Ubuntu kernel with the
+PREEMPT_RT patchset integrated for x86_64 and ARM64.
+
+This will change your kernel. You will need to manually configure grub to
+revert back to your original kernel after enabling real-time.
+
+Do you want to continue? [ default = Yes ]: (Y/n) yes
+Updating package lists
+Installing Real-Time Kernel packages
+Real-Time Kernel enabled
+A reboot is required to complete install.
+```
+
+After rebooting you'll be running the Real-Time Kernel!
+
+## Enable and manually install
+
+> **Note**
+> The --access-only flag is introduced in version 27.11
+
+If you would like to enable access to the Real-Time Kernel apt repository but not install the kernel right away, use the `--access-only` flag while enabling.
+
+```console
+$ sudo ua enable realtime-kernel --beta --access-only
+```
+
+With that extra flag you'll see output like the following:
+
+```
+One moment, checking your subscription first
+Updating package lists
+Skipping installing packages: ubuntu-realtime
+Real-Time Kernel access enabled
+```
+
+To install the kernel you can then run:
+
+```console
+$ sudo apt install ubuntu-realtime
+```
+
+You'll need to reboot after installing to boot into the Real-Time Kernel.
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -28,6 +28,26 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
             | xenial  |
             | bionic  |
 
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Enable cc-eal with --access-only
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        When I run `pro enable cc-eal --access-only` with sudo
+        Then I will see the following on stdout:
+        """
+        One moment, checking your subscription first
+        Updating package lists
+        Skipping installing packages: ubuntu-commoncriteria
+        CC EAL2 access enabled
+        """
+        Then I verify that running `apt-get install ubuntu-commoncriteria` `with sudo` exits `0`
+        Examples: ubuntu release
+            | release |
+            | xenial  |
+            | bionic  |
+
     @series.focal
     @series.jammy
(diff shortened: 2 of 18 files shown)


lucasmoura (commented):
@orndorffgrant we might have a problem with this flag for some services. For example, either `fips` or `livepatch` is now show as enabled after we run the enable command with the access flag. Maybe we could restrict which services are allowed to have that of enablement

orndorffgrant:
@lucasmoura I think that's a good idea, and we can include the list of services that support that in the help text.

Which services should support it? I'm thinking
- cc-eal
- cis/usg
- fips
- fips-updates
- realtime-kernel

fips(-updates) should probably have different messaging when the access-only flag is passed, and fips status on containers should probably check the versions of the mandatory packages.

So maybe fips support for --access-only can be a low prio follow up PR?

lucasmoura:
@orndorffgrant I just re-read my last comment and it was terribly written :facepalm: Sorry for that.
But I think I am missing your point for `fips` and `livepatch`. The current issue that I see is that on `fips`  we will not install the metapackage and we will never have the proc file we check to see if the service is enabled. I agree that this will not be a problem on a container, but it will for the other instances types we support.

Livepatch has the same issue with the `canonical-livepatch` missing, since we will not be able to run the status command for livepatch.

So in my opinion, we should exclude those services from that feature.

orndorffgrant:
@lucasmoura I think I understood your message, but I didn't properly explain my message :upside_down_face:.

To respond directly to your points:
> The current issue that I see is that on fips we will not install the metapackage and we will never have the proc file we check to see if the service is enabled.

Yep, so fips on a vm will show as "disabled" if --access-only is passed - which I think would be correct, but is confusing. This is a good argument to not allow `pro enable fips --access-only`. (or at least we would need to do more work to make the UX understandable and consistent if we want to support that).

> this will not be a problem on a container

I was trying to say that I think this is also a problem for fips on containers, because `pro enable fips --access-only` on a container will result in a fips status of `enabled`, even though none of the fips packages are installed - which is misleading. This is another good argument to not allow `pro enable fips --access-only`. (As an aside: I think fips status on containers should be smarter and actually check that fips versions of the core fips packages are installed. we may even want to check the kernel proc file on containers too)

> Livepatch has the same issue with the canonical-livepatch missing, since we will not be able to run the status command for livepatch.

The current implementation doesn't affect livepatch at all. It is implemented in `repo.py`. I can't think of a way to enable "access only" for livepatch, so I don't think this option should be allowed for livepatch either.


So --access-only should not work for fips, fips-updates, or livepatch. It can return an error.

So if we're going to allowlist the services that support `--access-only`, which services go in that list?

Some services don't and likely never will auto-install packages (esm and ros), so they probably shouldn't go in the list.
Per above, livepatch doesn't really make sense here.
Per above, fips/fips-updates have some UX issues. I would like to support `pro enable fips --access-only` but I think that's not necessary for now.

So that only leaves:
- cc-eal
- cis/usg
- realtime-kernel

does that sounds like a good list of services to support --access-only @lucasmoura ?

lucasmoura:
@orndorffgrant Yes, totally agree with the selected list and the rationale you presented. I also agree we can improve on how we check the status of `fips` on containers, good call there

uaclient/actions.py
```diff
@@ -113,8 +114,20 @@ def enable_entitlement_by_name(
...
+        access_only=access_only,
     )
+    if not entitlement.supports_access_only and access_only:
+        msg = messages.ENABLE_ACCESS_ONLY_NOT_SUPPORTED.format(
+            service=entitlement.title
+        )
+        raise exceptions.UserFacingError(
+            msg=msg.msg,
+            msg_code=msg.name,
+        )
```
lucasmoura, line 130: This is just an idea, but could make this check inside the `can_enable` method. I think this would be better since we might not use the `actions` module in the future

orndorffgrant: thats a good idea, I'll move it there!

orndorffgrant: @lucasmoura updated!

uaclient/entitlements/base.py
```diff
@@ -346,6 +352,17 @@ def can_enable(self) -> Tuple[bool, Optional[CanEnableFailure]]:
                     ),
                 )
 
+        if not self.supports_access_only and self.access_only:
```
lucasmoura, line 355: I think we can also add an unittest for this scenario

lucasmoura (commented):
@orndorffgrant I think we should just add a how to guide for this feature and I think it should be good to go

orndorffgrant:
@lucasmoura I integrated this feature into the how to guide for enabling cc-eal and I added a how to guide for enabling realtime-kernel, including the use of this flag.

I didn't include it in the cis/usg guide because the flag only makes sense for cis and not usg. So to add this flag to that document, it would require breaking up that document into two separate howtoguides for cis and usg. That felt unnecessary right now, because I think the most important documentation is that its use with realtime-kernel is explained.

features/attached_enable.feature
```diff
@@ -28,6 +28,25 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
...
+    Scenario Outline: Enable cc-eal with --access-only
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        When I run `pro enable cc-eal --access-only` with sudo
+        Then I will see the following on stdout:
+        """
+        One moment, checking your subscription first
+        Updating package lists
+        Skipping installing packages: ubuntu-commoncriteria
+        CC EAL2 access enabled
```
renanrodrigo, line 43: do you think it would be beneficial to include in the tests another command making sure that installing the package works after enabling with --access-only?

uaclient/entitlements/realtime.py
```diff
@@ -1,9 +1,13 @@
-from typing import Tuple
+from typing import Optional, Tuple  # noqa: F40`
```
renanrodrigo, line 1: `F401`?

renanrodrigo (approved):
looks great!
just a couple small things

uaclient/entitlements/realtime.py
orndorffgrant: oops! will fix

features/attached_enable.feature
orndorffgrant: good idea!

docs/source/howtoguides/enable_realtime_kernel.md
```diff
@@ -0,0 +1,62 @@
+# How to enable CC EAL
```
orndorffgrant, line 1: oops

orndorffgrant:
updated! we'll let CI pass before merging

PR #2198, opened 2022-08-12
messaging: rewrite apt and motd messages

orndorffgrant:
## Test Steps

Use new integration tests

## Desired commit type

## Checklist


lucasmoura (commented):
@orndorffgrant I think we should also update the `apt_messages` doc here

features/apt_messages.feature
```diff
@@ -0,0 +1,328 @@
...
+        Building dependency tree...
+        Reading state information...
+        Calculating upgrade...
+        The following security updates require Ubuntu Pro with 'esm-infra' enabled:
+          .*
+        Learn more about Ubuntu Pro for 16\.04 at https:\/\/ubuntu\.com\/16-04
+        0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded\.
+        """
+        When I attach `contract_token` with sudo
+        When I run shell command `apt-get upgrade || true` with sudo
```
lucasmoura, line 122: Why do we need the `|| true` here ?

features/staging_commands.feature
```diff
@@ -30,53 +30,6 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro staging subscri
         \s*\*\*\* .* 500
         \s*500 https://esm.staging.ubuntu.com/apps/ubuntu <release>-apps-security/main amd64 Packages
         """
-        When I run `mkdir -p /var/lib/ubuntu-advantage/messages` with sudo
```
lucasmoura, line 33: I think I am missing something, but why are we removing that part of the test ?

features/apt_messages.feature
orndorffgrant: ah this was a quick hack to run apt-get upgrade without providing stdin and without actually upgrading.

I'll replace it by just using the --dry-run flag

features/staging_commands.feature
orndorffgrant: The new tests test the full process of creating the messages and the apt hook filling in these variables. So I didn't think we needed these anymore, which only test how the apt hook fills in variables and concatenates files.

But I can add them back if we think they add value!

lucasmoura: Fair enough, just wanted to double check. No problem removing that

lucasmoura (approved):

features/apt_messages.feature
```diff
@@ -0,0 +1,328 @@
...
+        Then stdout matches regexp:
+        """
+        Reading package lists...
+        Building dependency tree...
+        Reading state information...
+        Calculating upgrade...
+        The following security updates require Ubuntu Pro with 'esm-infra' enabled:
+          .*
+        Learn more about Ubuntu Pro for 16\.04 at https:\/\/ubuntu\.com\/16-04
+        0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded\.
```
orndorffgrant, line 119:
```suggestion
        Reading package lists...
        Building dependency tree...
        Reading state information...
        Calculating upgrade...
        The following package was installed but can be removed:
          package1 package2...
        run autoremove
        The following security updates require Ubuntu Pro with 'esm-infra' enabled:
          package1 package2 package3 package4
        Learn more about Ubuntu Pro for 16\.04 at https:\/\/ubuntu\.com\/16-04
        The following packages will get upgraded:
          package1 package2
        0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded\.
```
@lechsandecki

PR #2201, opened 2022-08-16
security-status: list packages in oci manifest format

CalvoM:
## Proposed Commit Message

security-status: list packages in oci manifest format

Add an endpoint where we can get a list of packages installed with their versions.
The output is in the OCI manifest file format so that cloud providers can use it with 
OVAL data.

## Desired commit type

## Checklist

### DIFF
--- a/features/api_security_package_manifest.feature
+++ b/features/api_security_package_manifest.feature
@@ -1,0 +1,54 @@
+Feature: API security status package list in manifest format
+
+    @series.lts
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Call package manifest endpoint for machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        And I run `pro status` as non-root
+        Then stdout matches regexp:
+        """
+        esm-infra       +yes      +enabled  +Expanded Security Maintenance for Infrastructure
+        """
+        When I run `apt update` with sudo
+        And I run `apt upgrade -y` with sudo
+        And I run `apt install jq bzip2 libopenscap8 -y` with sudo
+        And I run shell command `pro api u.security.package_manifest.v1 | jq -r '.data.attributes.manifest_data' > manifest` as non-root
+        And I run shell command `wget https://security-metadata.canonical.com/oval/oci.com.ubuntu.<release>.usn.oval.xml.bz2` as non-root
+        And I run `bunzip2 oci.com.ubuntu.<release>.usn.oval.xml.bz2` as non-root
+        And I run shell command `oscap oval eval --report report.html oci.com.ubuntu.<release>.usn.oval.xml` as non-root
+        Then stdout matches regexp:
+        """
+        oval:com.ubuntu.<release>:def:<CVE_ID>:\s+false
+        """
+        # Trigger CVE https://ubuntu.com/security/CVE-2018-10846 with ID 39991000000 in OVAL data (<release> == Xenial $ Bionic)
+        # Trigger CVE https://ubuntu.com/security/CVE-2022-2509 with ID 55501000000 in OVAL data (<release> > Xenial)
+        When I run shell command `sed -i -E 's/libgnutls30:amd64\s+.*/libgnutls30:amd64 <base_version>/' manifest` as non-root
+        And I run shell command `oscap oval eval --report report.html oci.com.ubuntu.<release>.usn.oval.xml` as non-root
+        Then stdout matches regexp:
+        """
+        oval:com.ubuntu.<release>:def:<CVE_ID>:\s+true
+        """
+        # Update the manifest
+        When I run shell command `pro api u.security.package_manifest.v1 | jq -r '.data.attributes.manifest_data' > manifest` as non-root
+        And I run shell command `oscap oval eval --report report.html oci.com.ubuntu.<release>.usn.oval.xml` as non-root
+        Then stdout matches regexp:
+        """
+        oval:com.ubuntu.<release>:def:<CVE_ID>:\s+false
+        """
+        # Downgrade the package 
+        When I run shell command `apt install libgnutls30=<base_version> -y --allow-downgrades` with sudo
+        And I run shell command `pro api u.security.package_manifest.v1 | jq -r '.data.attributes.manifest_data' > manifest` as non-root
+        And I run shell command `oscap oval eval --report report.html oci.com.ubuntu.<release>.usn.oval.xml` as non-root
+        Then stdout matches regexp:
+        """
+        oval:com.ubuntu.<release>:def:<CVE_ID>:\s+true
+        """
+        
+
+        Examples: ubuntu release
+            | release | base_version    | CVE_ID      |
+            | xenial  | 3.4.10-4ubuntu1 | 39991000000 |
+            | bionic  | 3.5.18-1ubuntu1 | 55501000000 |
+            | focal   | 3.6.13-2ubuntu1 | 55501000000 |
+            | jammy   | 3.7.3-4ubuntu1  | 55501000000 |
--- a/uaclient/api/api.py
+++ b/uaclient/api/api.py
@@ -22,6 +22,7 @@ VALID_ENDPOINTS = [
     "u.pro.attach.magic.revoke.v1",
     "u.pro.attach.auto.should_auto_attach.v1",
     "u.pro.attach.auto.full_auto_attach.v1",
+    "u.security.package_manifest.v1",
 ]
 
 
--- a/uaclient/api/u/security/package_manifest/v1.py
+++ b/uaclient/api/u/security/package_manifest/v1.py
@@ -1,0 +1,47 @@
+from uaclient import apt, snap
+from uaclient.api.api import APIEndpoint
+from uaclient.api.data_types import AdditionalInfo
+from uaclient.config import UAConfig
+from uaclient.data_types import DataObject, Field, StringDataValue
+
+
+class PackageManifestResults(DataObject, AdditionalInfo):
+    fields = [
+        Field("manifest_data", StringDataValue),
+    ]
+
+    def __init__(self, manifest_data: str):
+        self.manifest_data = manifest_data
+
+
+def package_manifest() -> PackageManifestResults:
+    return _package_manifest(UAConfig())
+
+
+def _package_manifest(cfg: UAConfig) -> PackageManifestResults:
+    """Returns the status of installed packages (apt and snap packages)
+    Returns a string in manifest format i.e. package_name\tversion
+    """
+    manifest = ""
+    apt_pkgs = apt.get_installed_packages()
+    for apt_pkg in apt_pkgs:
+        arch = "" if apt_pkg.arch == "all" else ":" + apt_pkg.arch
+        manifest += "{}{}\t{}\n".format(apt_pkg.name, arch, apt_pkg.version)
+
+    pkgs = snap.get_installed_snaps()
+    for pkg in pkgs:
+        manifest += "snap:{name}\t{tracking}\t{rev}\n".format(
+            name=pkg.name,
+            tracking=pkg.tracking,
+            rev=pkg.rev,
+        )
+
+    return PackageManifestResults(manifest_data=manifest)
+
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="Packages",
+    fn=_package_manifest,
+    options_cls=None,
+)
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -7,7 +7,7 @@ import subprocess
 import sys
 import tempfile
 from functools import lru_cache
-from typing import Dict, List, Optional
+from typing import Dict, List, NamedTuple, Optional
 
 from uaclient import event_logger, exceptions, gpg, messages, system
 
@@ -46,6 +46,11 @@ event = event_logger.get_event_logger()
 class AptProxyScope(enum.Enum):
     GLOBAL = object()
     UACLIENT = object()
+
+
+InstalledAptPackages = NamedTuple(
+    "InstalledAptPackages", [("name", str), ("version", str), ("arch", str)]
+)
 
 
 def assert_valid_apt_credentials(repo_url, username, password):
@@ -487,13 +492,26 @@ def clean_apt_files(*, _entitlements=None):
 
 
 def is_installed(pkg: str) -> bool:
-    return pkg in get_installed_packages()
+    return pkg in get_installed_packages_names()
 
 
-def get_installed_packages() -> List[str]:
+def get_installed_packages() -> List[InstalledAptPackages]:
     out, _ = system.subp(["apt", "list", "--installed"])
     package_list = out.splitlines()[1:]
-    return [entry.split("/")[0] for entry in package_list]
+    return [
+        InstalledAptPackages(
+            name=entry.split("/")[0],
+            version=entry.split(" ")[1],
+            arch=entry.split(" ")[2],
+        )
+        for entry in package_list
+    ]
+
+
+def get_installed_packages_names(include_versions: bool = False) -> List[str]:
+    package_list = get_installed_packages()
+    pkg_names = [pkg.name for pkg in package_list]
+    return pkg_names
 
 
 def setup_apt_proxy(
--- a/uaclient/snap.py
+++ b/uaclient/snap.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from typing import List, Optional
 
 from uaclient import apt, event_logger, exceptions, messages, system
@@ -7,13 +8,14 @@ SNAP_CMD = "/usr/bin/snap"
 SNAP_INSTALL_RETRIES = [0.5, 1.0, 5.0]
 HTTP_PROXY_OPTION = "proxy.http"
 HTTPS_PROXY_OPTION = "proxy.https"
+SNAP_CHANNEL_SHORTEN_VALUE = "…"
 
 event = event_logger.get_event_logger()
 
 
 def is_installed() -> bool:
     """Returns whether or not snap is installed"""
-    return "snapd" in apt.get_installed_packages()
+    return "snapd" in apt.get_installed_packages_names()
 
 
 def configure_snap_proxy(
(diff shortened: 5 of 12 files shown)


uaclient/api/api.py
```diff
@@ -21,6 +21,7 @@
     "u.pro.attach.magic.wait.v1",
     "u.pro.attach.magic.revoke.v1",
     "u.pro.attach.auto.should_auto_attach.v1",
+    "u.pro.security_status.packages.v1",
```
orndorffgrant, line 24: api endpoint name bikeshedding thread:

While related to ESM (which is why we're doing this at all), it isn't returning information directly related to pro features. So I don't think it should be in the "pro" namespace.
It should probably be in some namespace, maybe a generic "security".
The method name should be specific for this format, something like "get_package_manifest"

So I suggest: `u.security.get_package_manifest`

uaclient/api/u/pro/security_status/packages/v1.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    cache = Cache()
+    manifest = ""
+    # get snap packages
+    out, _ = system.subp(["snap", "list"])
+    apps = out.splitlines()
+    apps = apps[1:]
+    for line in apps:
+        pkg = line.split()
+        manifest += "{app}\t{version}\n".format(app=pkg[0], version=pkg[1])
```
orndorffgrant, line 35: snaps should appear at the end of the file in the format 
```
snap:core20	stable	1587
snap:snapd	stable	16292
snap:lxd	4.0/stable/ubuntu-20.04	22753
```

(example taken from https://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64-root.manifest)

features/api_security_status_package_manifest.feature
```diff
@@ -0,0 +1,40 @@
+Feature: API security status package list in manifest format
+
+    @series.lts
+    @uses.config.machine_type.aws.pro
+    @uses.config.machine_type.azure.pro
+    @uses.config.machine_type.gcp.pro
```
orndorffgrant, line 6: I don't think this test needs to run on pro cloud instances, probably just lxd containers.
But otherwise this is a great test!

uaclient/api/api.py
CalvoM: makes sense, even making more sense why the integration tests should just be run on containers

features/api_security_status_package_manifest.feature
```diff
@@ -0,0 +1,73 @@
...
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Call package manifest endpoint
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
+        """
+        contract_url: 'https://contracts.canonical.com'
+        data_dir: /var/lib/ubuntu-advantage
+        log_level: debug
+        log_file: /var/log/ubuntu-advantage.log
+        """
```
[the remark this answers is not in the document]
CalvoM: Oh yes,

uaclient/api/u/security_status/get_package_manifest/v1.py
```diff
@@ -0,0 +1,57 @@
...
+    for package in cache:
+        if package.is_installed:
+            version = [
+                version.version
+                for version in package.versions
+                if version.is_installed
+            ][0]
+            manifest += "{app}\t{version}\n".format(
+                app=package.name, version=version
+            )
```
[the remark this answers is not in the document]
CalvoM: I see it only printing the packages without the versions. Unless I add a function to print the version as well.

uaclient/api/api.py
renanrodrigo: I would even remove the `get`, leaving it `u.security.package_manifest.v1` but it's just an opinion.

uaclient/snap.py
```diff
@@ -93,3 +93,33 @@ def get_config_option_value(key: str) -> Optional[str]:
...
+        self.tracking = tracking
+        self.publisher = publisher
+        self.notes = notes
+
+
+def get_installed_packages() -> List[SnapPackage]:
+    out, _ = system.subp(["snap", "list"])
+    apps = out.splitlines()
+    apps = apps[1:]
+    snaps = list()  # type: List[SnapPackage]
```
renanrodrigo, line 112: I'd use 
```suggestion
    snaps = []
```
but feel free to ignore it.

uaclient/api/u/security_status/get_package_manifest/v1.py
```diff
@@ -0,0 +1,56 @@
...
```
renanrodrigo, line 39: There is always a single installed version, no need to loop through.
```suggestion
            manifest += "{app}\t{version}\n".format(
                app=package.name, version=package.installed.version
            )
```

And I think the implementation with `Cache` is ok, unless you want the optimized shortest delay here. 
I'd say we don't need it, but `dpkg-query -W` will return names and versions just in case.

lucasmoura: @CalvoM I think we should either:

* modify the apt module function to use cache and control if the output will be with or without the versions
* modify the apt module function dpkg call to include the package version if required.

I would vote for the `dpkg` change since I believe is a faster approach that we are already using

uaclient/snap.py
CalvoM: I was trying to avoid some mypy errors, which, apparently, never existed.

uaclient/api/u/security_status/get_package_manifest/v1.py
CalvoM: I see, then the second option is okay, since it already being used. I will check if I need versions as well.

uaclient/snap.py
renanrodrigo: I see
So
when you create an empty list and just append things to it, the first append tells mypy which kind of list it is. Appending something different later would give you an error, but here you're fine

CalvoM:
We will only close this after the oval xml files are updated by security.

renanrodrigo (approved):
+1 just waiting on the fix on the security side

features/api_security_package_manifest.feature
```diff
@@ -0,0 +1,67 @@
...
+
+        Examples: ubuntu release
+            | release | pkg_version       |
+            | bionic  | 3.5.18-1ubuntu1.5 |
+            | focal   | 3.6.13-2ubuntu1.6 |
+            | jammy   | 3.7.3-4ubuntu1.0  |
+
+    @series.xenial
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Call package manifest endpoint for machine on ESM
```
orndorffgrant, line 37: Why does this need to be a separate scenario? I think we can just attach for every release

orndorffgrant (changes requested):
Looking good! A couple in line comments. Also we need to address the different format for snaps.

And super minor: our manifest has an extra newline at the end compared to the official image manifests

uaclient/api/u/pro/security_status/packages/v1.py
orndorffgrant: @CalvoM this comment is still valid

features/api_security_package_manifest.feature
CalvoM: I was thinking of having a separate scenario where we test it on an unattached machine. What if I just add the other releases to the attached scenario and rename it to `call package manifest endpoint on an attached machine`

features/api_security_package_manifest.feature
```diff
@@ -0,0 +1,100 @@
...
+        Then stdout matches regexp:
+        """
+        0
+        """
+        # Trigger CVE https://ubuntu.com/security/CVE-2022-2509
+        When I run shell command `sed -i -E 's/libgnutls30:amd64\s+.*/libgnutls30:amd64 <base_version>/' manifest` as non-root
+        And I run shell command `oscap oval eval --report report.html oci.com.ubuntu.<release>.usn.oval.xml | grep true | wc -l` as non-root
+        Then stdout does not match regexp:
+        """
+        0
```
orndorffgrant, line 25: Is there an expected number here like `1`? I think it is more intuitive if we assert the content matches something rather than not matching something
```suggestion
        Then stdout matches regexp:
        """
        1
```

features/api_security_package_manifest.feature
orndorffgrant: I don't think we need a scenario for both attached and unattached, because this functionality doesn't change based on whether the machine is attached or not

features/api_security_package_manifest.feature
CalvoM: It is not possible to correctly say the number since we do not how many updates have been done after the base package version.

CalvoM: Oh, found a way to do just this,by getting the cve ID in the OVAL data  and checking if it true.

uaclient/api/u/security/package_manifest/v1.py
```diff
@@ -0,0 +1,44 @@
...
+    manifest = ""
+    apt_pkgs = apt.get_installed_packages(include_versions=True)
+    for apt_pkg in apt_pkgs:
+        manifest += apt_pkg + "\n"
+
+    pkgs = snap.get_installed_packages()
+    for pkg in pkgs:
+        manifest += "{app}\t{version}\n".format(
+            app=pkg.name, version=pkg.version
+        )
```
orndorffgrant, line 34: Snaps should be in the format:
```
snap:core20	stable	1587
snap:snapd	stable	16292
snap:lxd	4.0/stable/ubuntu-20.04	22753
```

which I think is `name`, `tracking`, `rev`

(example taken from https://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64-root.manifest)

And unfortunately it looks like `snap list` cuts its output sometimes. on a fresh focal container notice the tracking channel of lxd is cut-off:
```
root@dev-f:~# snap list
Name    Version        Rev    Tracking       Publisher   Notes
core20  20220805       1611   latest/stable  canonical✓  base
lxd     4.0.9-8e2046b  22753  4.0/stable/…   canonical✓  -
snapd   2.56.2         16292  latest/stable  canonical✓  snapd
```
So we may need to run `snap info $package` for each and grab the `tracking` field from there

It is also curious that the manifest doesn't include "latest/" for the core20 or snapd channels :/

orndorffgrant: We should probably include "latest/" instead of trying to edit what we get from snap. Oval should be able to handle both via a regex anyway

features/api_security_package_manifest.feature
```diff
@@ -0,0 +1,54 @@
...
+        """
+        # Trigger CVE https://ubuntu.com/security/CVE-2018-10846 with ID 39991000000 in OVAL data (<release> == Xenial $ Bionic)
+        # Trigger CVE https://ubuntu.com/security/CVE-2022-2509 with ID 55501000000 in OVAL data (<release> > Xenial)
+        When I run shell command `sed -i -E 's/libgnutls30:amd64\s+.*/libgnutls30:amd64 <base_version>/' manifest` as non-root
+        And I run shell command `oscap oval eval --report report.html oci.com.ubuntu.<release>.usn.oval.xml | grep -E 'oval:com.ubuntu.<release>:def:<CVE_ID>:\s+true' | wc -l` as non-root
+        Then stdout matches regexp:
+        """
+        1
+        """
```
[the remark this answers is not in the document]
CalvoM: Oh yes, yes that is easier actually.

uaclient/snap.py
```diff
@@ -93,3 +95,44 @@ def get_config_option_value(key: str) -> Optional[str]:
         return out.strip()
     except exceptions.ProcessExecutionError:
         return None
+
+
+def get_snap_package_info_tracking(package: str) -> Optional[str]:
```
lucasmoura, line 100: Let's add unittest for the new snap functions

uaclient/snap.py
```diff
@@ -93,3 +95,44 @@ def get_config_option_value(key: str) -> Optional[str]:
...
+        self.rev = rev
+        self.tracking = tracking
+        self.publisher = publisher
+        self.notes = notes
+
+
+def get_installed_packages() -> List[SnapPackage]:
+    out, _ = system.subp(["snap", "list"])
+    apps = out.splitlines()
+    apps = apps[1:]
```
lucasmoura, line 121: Why are we removing the first line ?

uaclient/snap.py
```diff
@@ -93,3 +95,44 @@ def get_config_option_value(key: str) -> Optional[str]:
...
+    def __init__(self, name, version, rev, tracking, publisher, notes):
+        self.name = name
+        self.version = version
+        self.rev = rev
+        self.tracking = tracking
+        self.publisher = publisher
+        self.notes = notes
+
+
+def get_installed_packages() -> List[SnapPackage]:
```
lucasmoura, line 118: I think we can rename this function to `get_installed_snaps`

uaclient/snap.py
```diff
@@ -93,3 +95,44 @@ def get_config_option_value(key: str) -> Optional[str]:
...
+    out, _ = system.subp(["snap", "list"])
```
lucasmoura, line 119: The output of that call could have `unicode`  and `color`values. I think it is safe for us to use `--color never --unicode never` here

uaclient/snap.py
```diff
@@ -93,3 +95,44 @@ def get_config_option_value(key: str) -> Optional[str]:
...
+    apps = apps[1:]
+    snaps = []
+    for line in apps:
+        pkg = line.split()
+        snap = SnapPackage(
+            name=pkg[0],
+            version=pkg[1],
+            rev=pkg[2],
+            tracking=pkg[3],
+            publisher=pkg[4],
```
lucasmoura, line 130: We might have some special character for `publisher`, like double `**`. Should we remove those ?

uaclient/snap.py
```diff
@@ -93,3 +95,44 @@ def get_config_option_value(key: str) -> Optional[str]:
         return out.strip()
     except exceptions.ProcessExecutionError:
         return None
+
+
+def get_snap_package_info_tracking(package: str) -> Optional[str]:
+    out, _ = system.subp(["snap", "info", package])
```
lucasmoura, line 101: I think we can also add the `--unicode never --color never` options here

uaclient/snap.py
```diff
@@ -93,3 +95,44 @@ def get_config_option_value(key: str) -> Optional[str]:
...
+
+
+def get_installed_packages() -> List[SnapPackage]:
+    out, _ = system.subp(["snap", "list"])
+    apps = out.splitlines()
+    apps = apps[1:]
+    snaps = []
+    for line in apps:
+        pkg = line.split()
+        snap = SnapPackage(
```
lucasmoura, line 125: I think we can create the class like this:

`snap = SnapPackage(*pkg)`

uaclient/snap.py
CalvoM: It has the column names `Name Version Rev Tracking Publisher Notes`

uaclient/snap.py
CalvoM: I do not see any harm, considering they indicate a verified publisher

uaclient/snap.py
lucasmoura: :facepalm: my mistake there.

uaclient/apt.py
```diff
@@ -496,6 +496,16 @@ def get_installed_packages() -> List[str]:
...
 
 
+def get_installed_packages_with_version(
+    include_versions: bool = False,
+) -> List[str]:
+    filters = ["--showformat=${Package}\\n"]
+    if include_versions:
+        filters = []
+    out, _ = system.subp(["dpkg-query", "-W", *filters])
+    return out.splitlines()
```
orndorffgrant, line 506: Renan discovered last week that `dpkg-query -W` can return packages that are not installed. He changed the `get_installed_packages` function to this https://github.com/canonical/ubuntu-advantage-client/pull/2238/files#diff-e0ee470947c8a3ab51ba84f5750d9e55511cc77d91556b1f54d07f076a690d41R493

We should change this function to use `apt list --installed` as well

orndorffgrant (commented):
This looks great! just one change left because of the bug @renanrodrigo found in out `get_installed_packages` function last week

uaclient/apt.py
renanrodrigo: I think the way to go is:
- add a flag parameter like `show_version` to the `get_installed_packages` function
- change the way to parse the output based on this flag

But I think this was on this PR somehow before the new function showed up, feel free to disagree and disregard this idea.

features/api_security_package_manifest.feature
```diff
@@ -0,0 +1,54 @@
...
+        oval:com.ubuntu.<release>:def:<CVE_ID>:\s+false
+        """
+        # Downgrade the package 
+        When I run shell command `apt install libgnutls30=<base_version> -y --allow-downgrades` with sudo
+        And I run shell command `pro api u.security.package_manifest.v1 | jq -r '.data.attributes.manifest_data' > manifest` as non-root
+        And I run shell command `oscap oval eval --report report.html oci.com.ubuntu.<release>.usn.oval.xml` as non-root
+        Then stdout matches regexp:
+        """
+        oval:com.ubuntu.<release>:def:<CVE_ID>:\s+true
+        """
```
lucasmoura, line 46: @orndorffgrant @renanrodrigo @CalvoM should we add a scenario where a snap has CVE associated with ?

orndorffgrant: Our OVAL doesn't cover snaps and oscap doesn't look at snaps. So I don't think we can create a realistic test case around this :(

uaclient/apt.py
```diff
@@ -490,10 +490,20 @@ def is_installed(pkg: str) -> bool:
...
+    if include_versions:
+        pkg_versions = [entry.split(" ")[1:] for entry in package_list]
+        pkgs = []
+        for pkg in zip(pkg_names, pkg_versions):
+            arch = pkg[1][1]
+            arch = "" if arch == "all" else ":" + arch
+            version = pkg[1][0]
+            pkgs.append("{}{}\t{}".format(pkg[0], arch, version))
+        return pkgs
+    return pkg_names
```
orndorffgrant, line 506: Sorry for disagreeing here, but I think this should be two different functions. Generally I don't think the value of a parameter should meaningfully change the type of the return value.

I suggest:
```python
PackageVersion = NamedTuple("PackageVersion", [("name", str), ("version", str)])
def get_installed_packages_with_versions() -> List[PackageVersion]:
    out, _ = system.subp(["apt", "list", "--installed"])
    package_list = out.splitlines()[1:]
    return [
        PackageVersion(
            name=entry.split("/")[0],
            version=entry.split(" ")[1]
        ) for entry in package_list
    ]

def get_installed_packages() -> List[str]:
    return [package.name for package in get_installed_packages_with_versions()]
```

CalvoM: I was thinking about this part, thanks for this different approach
I thought my approach was abit brittle.

uaclient/api/u/security/package_manifest/v1.py
```diff
@@ -0,0 +1,46 @@
...
+class InstalledPackagesResults(DataObject, AdditionalInfo):
+    fields = [
+        Field("manifest_data", StringDataValue),
+    ]
+
+    def __init__(self, manifest_data: str):
+        self.manifest_data = manifest_data
+
+
+def get_installed_packages() -> InstalledPackagesResults:
```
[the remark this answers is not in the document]
CalvoM: Oh yes, I do not why I forgot to to just that when I changed the api endpoint path.

uaclient/apt.py
```diff
@@ -490,10 +495,33 @@ def is_installed(pkg: str) -> bool:
     return pkg in get_installed_packages()
 
 
-def get_installed_packages() -> List[str]:
+def generic_installed_packages() -> List[InstalledAptPackages]:
```
CalvoM, line 498: @orndorffgrant I mean a function like this which will have all the package details from `apt list --installed` and thus there is no need for a separate function for versions or architecture(we might need it later, who knows e.g in the manifest file).

orndorffgrant (approved):
Changes look great - thank you!

renanrodrigo (approved):

PR #2202, opened 2022-08-17
tests: fix docker test for new ua dependency

lucasmoura:


features/docker.feature
```diff
@@ -19,7 +19,9 @@ Feature: Build docker images with ua services
             apt-get update \
             && apt-get install --no-install-recommends -y ubuntu-advantage-tools ca-certificates \
 
-            && dpkg -i /ua.deb \
+            && dpkg -i /ua.deb | true \
```
orndorffgrant, line 22:
```suggestion
            && dpkg -i /ua.deb || true \
```

orndorffgrant (approved):
looks good! just a minor nit

features/docker.feature
orndorffgrant: actually, will using `||` mess with the overall command flow? since the whole thing is `&&`'d together?

Maybe keeping `|` is best - up to you

lucasmoura: Fair enough, I will change it and merge the PR

PR #2203, opened 2022-08-18
tests: update jammy livepatch support

orndorffgrant:


lucasmoura (approved):

lucasmoura:
@orndorffgrant there still seems to have a Livepatch related failure on one Jammy test. I think we can also address it here

lucasmoura (approved):

PR #2206, opened 2022-08-19
enable: fix apt auth line replacement

orndorffgrant:
The regex was failing to grab the entire url and would only match
"esm.ubuntu.com" for each line. Because every service is a more specific
path of esm.ubuntu.com and there was no check to prevent multiple
additions of a new line, then a new line would be inserted before every
existing line each time.

This changes to a simpler split-based approach to get each line's url
and compare. We also add an integration test to insure the behavior is
not inflating the size of the file.

LP: #1985863

## Test Steps

Run the new integration test

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -1065,3 +1065,34 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         """
         Stderr: E: Unable to locate package some-package-aws
         """
+
+    @series.xenial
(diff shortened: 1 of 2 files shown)


renanrodrigo (approved):

orndorffgrant:
checking the failed test

orndorffgrant:
looks like a flake. I triggered a re-run to make sure

PR #2212, opened 2022-08-25
security-status: add CVEs fixed by Livepatch to the JSON output (SC-1233)

renanrodrigo:
Add a list with the CVEs fixed by livepatch to the output of `security-status`.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/security_status.py
+++ b/uaclient/security_status.py
@@ -1,3 +1,5 @@
+import json
+import logging
 from collections import defaultdict
 from enum import Enum
 from typing import Any, DefaultDict, Dict, List, Tuple  # noqa: F401
@@ -6,8 +8,10 @@ from apt import Cache  # type: ignore
 from apt import package as apt_package
 
 from uaclient.config import UAConfig
+from uaclient.entitlements.livepatch import LIVEPATCH_CMD
+from uaclient.exceptions import ProcessExecutionError, UserFacingError
 from uaclient.status import status
-from uaclient.system import get_platform_info
+from uaclient.system import get_kernel_info, get_platform_info, subp
 
 series = get_platform_info()["series"]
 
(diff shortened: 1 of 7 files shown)


uaclient/security_status.py
```diff
@@ -132,6 +134,35 @@ def get_ua_info(cfg: UAConfig) -> Dict[str, Any]:
     return ua_info
 
 
+def get_livepatch_fixed_cves() -> List[Dict[str, str]]:
+    if not which(LIVEPATCH_CMD):
+        return []
+
+    out, _err = subp([LIVEPATCH_CMD, "status", "--format", "json"])
+    livepatch_output = json.loads(out)
+    status_list = livepatch_output.get("Status", [])
```
lucasmoura, line 143: Since we are checking to see if the `status_list` is not None in the next line, I don't think we need an empty list as a default value

uaclient/security_status.py
```diff
@@ -132,6 +136,42 @@ def get_ua_info(cfg: UAConfig) -> Dict[str, Any]:
...
+    try:
+        livepatch_output = json.loads(out)
+    except json.JSONDecodeError:
+        msg = "Could not parse Livepatch Status JSON: {}".format(out)
+        logging.debug(msg)
+        return []
+
+    status_list = livepatch_output.get("Status")
+    if status_list:
+        kernel_info = get_kernel_info()
```
lucasmoura, line 154: `get_kernel_info` can also fail, although it should be rare, I think we should still handle it somehow.
We could try to run `uname -r` directly if it fails or just fail loudly if we have more than on dict on the status list

renanrodrigo: Well the function calls `uname` as part of the execution through the Python code. 
I´d say if we get to the way weird scenario of not being able to parse the the Kernel information, logging it and returning empty list should be enough for this functionality (as there would be more to care about...)
Would that be enough?

lucasmoura: Yeah, I think that also works

renanrodrigo: Well it seems (per MM conversation in ~livepatch) that the kernel check is useless anyway - I just removed it

lucasmoura (approved):

renanrodrigo:
@orndorffgrant lowercase keys are there
@lucasmoura kernel check is there

this is ready now

orndorffgrant (approved):
LGTM Ready to merge after CI is happy

lucasmoura (approved):

PR #2215, opened 2022-08-29
status: reduce calls to read machine token file when unattached SC1063

CalvoM:
## Proposed Commit Message

status: reduce calls to read machine token file when unattached

We had almost 231 machine token access when running `pro status` as root. This PR reduces this to about 5.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -90,12 +90,15 @@ class UAEntitlement(metaclass=abc.ABCMeta):
     @property
     def presentation_name(self) -> str:
         """The user-facing name shown for this entitlement"""
-        return (
-            self.cfg.machine_token_file.entitlements.get(self.name, {})
-            .get("entitlement", {})
-            .get("affordances", {})
-            .get("presentedAs", self.name)
-        )
+        if self.cfg.machine_token_file.is_present:
+            return (
+                self.cfg.machine_token_file.entitlements.get(self.name, {})
+                .get("entitlement", {})
+                .get("affordances", {})
+                .get("presentedAs", self.name)
+            )
+        else:
+            return self.name
 
     @property
     def help_info(self) -> str:
--- a/uaclient/files.py
+++ b/uaclient/files.py
@@ -43,6 +43,10 @@ class UAFile:
     def is_private(self) -> bool:
         return self._is_private
 
+    @property
+    def is_present(self):
+        return os.path.exists(self.path)
+
     def write(self, content: str):
         file_mode = (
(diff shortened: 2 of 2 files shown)


uaclient/entitlements/base.py
```diff
@@ -71,8 +71,9 @@ def name(self) -> str:
     def valid_names(self) -> List[str]:
         """The list of names this entitlement may be called."""
         valid_names = [self.name]
-        if self.presentation_name != self.name:
-            valid_names.append(self.presentation_name)
+        if self.cfg.machine_token_file.is_present:
+            if self.presentation_name != self.name:
+                valid_names.append(self.presentation_name)
```
orndorffgrant, line 76: This makes sense! I only think this check should go down in `presentation_name` since that is the function that directly looks at the machine_token_file.

So maybe something like this:
```python
    @property
    def presentation_name(self) -> str:
        """The user-facing name shown for this entitlement"""
        if self.cfg.machine_token_file.is_present:
            return (
                self.cfg.machine_token_file.entitlements.get(self.name, {})
                .get("entitlement", {})
                .get("affordances", {})
                .get("presentedAs", self.name)
            )
        else:
            return self.name
```

orndorffgrant (commented):
Thanks for figuring out what was causing the extra logs! Just some organization suggestions

uaclient/status.py
```diff
@@ -239,11 +239,12 @@ def _unattached_status(cfg: UAConfig) -> Dict[str, Any]:
...
             )
         except exceptions.EntitlementNotFoundError:
-            LOG.debug(
-                messages.AVAILABILITY_FROM_UNKNOWN_SERVICE.format(
-                    service=resource.get("name", "without a 'name' key")
+            if cfg.machine_token_file.is_present:
+                LOG.debug(
+                    messages.AVAILABILITY_FROM_UNKNOWN_SERVICE.format(
+                        service=resource.get("name", "without a 'name' key")
+                    )
```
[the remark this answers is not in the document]
CalvoM: Just something I was trying to see. Updating.

uaclient/files.py
```diff
@@ -66,6 +66,9 @@ def read(self) -> Optional[str]:
     def delete(self):
         system.remove_file(self.path)
 
+    def is_present(self):
```
orndorffgrant, line 69: This needs `@property` if we want to be able to access it like `file.is_present`

CalvoM: Oh, my bad 🤦‍♂️🤦‍♂️

PR #2216, opened 2022-08-29
ci: build packages on jammy

orndorffgrant:
This should fix the kinetic package build errors we are seeing


renanrodrigo (approved):
unbreak my CI
say I'm running on Jammyyy....

PR #2217, opened 2022-08-31
system: update regex for get_kernel_info

lucasmoura:
## Proposed Commit Message
system: update regex for get_kernel_info

Update the regex so it can also detect fips kernels

## Test Steps
Run the new unittest and a cloud fips tests

## Desired commit type

## Checklist


uaclient/system.py
```diff
@@ -34,7 +34,7 @@
     r"[.-]"
     r"(?P<subrev>[\d]+)"
     r"(~(?P<hwerev>[\d.]+))?"
-    r"-"
+    r"[-+]"
```
renanrodrigo, line 37: I think you need to escape the "+" right? Won't this catch "one or more dashes"?

renanrodrigo: TIL it does not because of the brackets

orndorffgrant:
@lucasmoura this looks good! thanks for fixing these
Should we add a pro-fips test to CI? so we can spot this more easily in the future. just a single pro-fips scenario on one release. Maybe `ubuntu_pro_fips.feature:181` by adding a special tag to that one?

renanrodrigo:
Could we add ubuntu-pro-fips on bionic to the included integration tests, and run the full suite?

lucasmoura:
Yes, I think that is a good idea. Let me update the actions file

lucasmoura:
@orndorffgrant @renanrodrigo just adding another commit to also remove unavailable services when calling status on the Pro FIPS tests

renanrodrigo (approved):
As soon as CI is happy I'm also happy

orndorffgrant (approved):
ci failure is unrelated

uaclient/tests/test_system.py
```diff
@@ -97,6 +97,23 @@ class TestGetKernelInfo:
...
+                    proc_version_signature_full="Ubuntu 5.4.0-1021.21+fips2-aws-fips 5.4.44",  # noqa: E501
+                    proc_version_signature_version="5.4.0-1021.21+fips2-aws-fips",  # noqa: E501
+                    version="5.4.0",
+                    major=5,
+                    minor=4,
+                    patch=0,
+                    abi="1021",
+                    subrev="21",
+                    hwerev="",
+                    flavor="fips2-aws-fips",
```
orndorffgrant, line 114: looking at this again @lucasmoura wouldn't we expect the flavor to just be `aws-fips`? I think `fips2` might be a different property of the kernel version that we need to add to our kernelinfo object

orndorffgrant (changes requested):
Sorry it looks like this is failing to match `Ubuntu 4.4.0-1017.22~recert1-fips 4.4.185`
I'm investigating more

orndorffgrant:
So I think overall it should become
```
^(?P<version>[\d]+[.-][\d]+[.-][\d]+)-(?P<abi>[\d]+)[.-](?P<subrev>[\d]+)(~(?P<hwerev>[\w\d.]+))?(\+(?P<tag>[\w\d.]+))?[-](?P<flavor>[A-Za-z0-9_-]+)$
```
The difference is the addition of `\w` to the hwerev group, and separating out the +fips2 into its own group.
But let's ask kernel team about it

lucasmoura:
@orndorffgrant Good catch, I have updated the code with the suggestions

uaclient/system.py
```diff
@@ -33,8 +33,9 @@
     r"(?P<abi>[\d]+)"
     r"[.-]"
     r"(?P<subrev>[\d]+)"
-    r"(~(?P<hwerev>[\d.]+))?"
-    r"-"
+    r"(~(?P<hwerev>[\w\d.]+))?"
+    r"(\+(?P<tag>[\w\d.]+))?"
+    r"[-]"
```
orndorffgrant, line 38:
```suggestion
    r"-"
```

orndorffgrant (approved):
One tiny nit but LGTM!

renanrodrigo (approved):
lgtm

uaclient/system.py
lucasmoura: Thanks @orndorffgrant, updated

PR #2218, opened 2022-08-31
tests: add integration test for airgapped setup

lucasmoura:
## Proposed Commit Message
tests: add integration test for airgapped setup

Add integration test for airgapped setup together with UA

## Test Steps
Run the new integration test

## Desired commit type

## Checklist

### DIFF
--- a/features/airgapped.feature
+++ b/features/airgapped.feature
@@ -1,0 +1,45 @@
+@uses.config.contract_token
+Feature: Performing attach using ua-airgapped
+
+    @series.jammy
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attached enable Common Criteria service in an ubuntu lxd container
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I launch a `jammy` `mirror` machine
+        And I run `add-apt-repository ppa:yellow/ua-airgapped -y` `with sudo` on the `mirror` machine
+        And I run `add-apt-repository ppa:yellow/ua-airgapped -y` with sudo
+        And I run `apt-get update` `with sudo` on the `mirror` machine
+        And I run `apt-get update` with sudo
+        And I run `apt-get install apt-mirror get-resource-tokens -yq` `with sudo` on the `mirror` machine
+        And I run `apt-get install contracts-airgapped ua-airgapped -yq` with sudo
+        And I download the service credentials on the `mirror` machine
+        And I extract the `esm-infra` credentials from the `mirror` machine
+        And I extract the `esm-apps` credentials from the `mirror` machine
+        And I set the apt-mirror file for `<release>` with the `esm-infra,esm-apps` credentials on the `mirror` machine
+        And I run `apt-mirror` `with sudo` on the `mirror` machine
+        And I serve the `esm-infra` mirror using port `8000` on the `mirror` machine
+        And I serve the `esm-apps` mirror using port `9000` on the `mirror` machine
+        And I run `apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4067E40313CB4B13` with sudo
+        And I create the contract config overrides file for `esm-infra,esm-apps` using the `mirror` machine
+        And I configure the ua-airgapped service
+        And I start the ua-airgapped service
+        And I change contract key `contract_url` to use value `http:\/\/127.0.0.1:8484`
+        And I attach `contract_token` with sudo
+        Then stdout matches regexp:
+        """
+        esm-apps     +yes      +enabled .*
+        esm-infra    +yes      +enabled .*
+        """
+        When I run `apt-cache policy hello` with sudo
+        Then stdout matches regexp:
+        """
+        500 .*:9000/ubuntu jammy-apps-security/main
+        """
+        And stdout matches regexp:
+        """
+        500 .*:8000/ubuntu jammy-infra-security/main
+        """
+
+        Examples: ubuntu release
+            | release |
+            | jammy   |
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -488,6 +488,17 @@ def change_contract_endpoint_to_staging(context, user_spec):
     )
 
 
+@when("I change contract key `{key}` to use value `{value}`")
+def change_contract_key_to_use_value(context, key, value):
+    when_i_run_command(
+        context,
+        "sed -i 's/{}: .*/{}: {}/g' {}".format(
+            key, key, value, DEFAULT_CONFIG_FILE
+        ),
+        "with sudo",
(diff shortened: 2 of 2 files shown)


pandrey2003 (approved):
LGTM, thanks for quickly iterating on this :)

renanrodrigo (approved):
small nit about a name, otherwise lgtm

orndorffgrant (commented):
This is great! I just have some opinions that we can consider.

* There are (at least) 3 roles here (airgapped client, airgapped contract server, non-airgapped apt-mirror) so I think we should give each role it's own machine in our test.
* (Optional) Ideally we'd cut off internet access on the airgapped machines in the test, but this might be hard to do without causing test problems
* I think `ua-airgapped` should go on the apt-mirror machine, not the same machine as `contracts-airgapped`
* In my head the overall flow will look something like:
  1. launch machine A for apt-mirror (this one has internet access)
  2. install `get-resource-tokens ua-airgapped apt-mirror` on machine A
  3. run `get-resource-tokens` on machine A
  4. use the tokens to set up apt-mirror on machine A
  5. on machine A, use IP:port of machine A in input to run `ua-airgapped`
  6. launch machine B for airgapped contract server (no internet, but can reach other test machines)
  7. install `contracts-airgapped` on machine B
  8. use output of `ua-airgapped` from machine A to start `contracts-airgapped` on machine B
  9. launch machine C for client  (no internet, but can reach other test machines)
  10. configure `ua` on machine C to use machine B as contract server
  11. `ua attach` on machine C
  * so I think the test should follow that ordering more closely. Right now it launches all the machines at the top, then interweaves installing packages on all of the machines. The rest follows what I outlined well I think. 

Also, the ppa seems to only have packages for jammy. We'll probably at least want packages for focal as well.

lucasmoura:
@orndorffgrant Good points. I have created a separate machine for the `ua-airgapped` setup, but I am still unsure about blocking the connection on the main machine. But we can further discuss this

lucasmoura:
@orndorffgrant I have updated the tests to drop the internet connection as well. However, I am still unsure about the `ua-airgapped` suggestion. Do you think we should do that just to better replicate the field engineering process of sending the file to the customer with the proxies already applied ?

features/airgapped.feature
```diff
@@ -0,0 +1,49 @@
...
+        500 .*:9000/ubuntu jammy-apps-security/main
+        """
+        And stdout matches regexp:
+        """
+        500 .*:8000/ubuntu jammy-infra-security/main
+        """
+
+        Examples: ubuntu release
+            | release |
+            | jammy   |
```
orndorffgrant, line 49:
```suggestion
    @series.jammy
    @uses.config.machine_type.lxd.container
    Scenario Outline: Attached enable Common Criteria service in an ubuntu lxd container
        # set up the apt mirror and airgapped contract configuration
        When I launch a `jammy` `mirror` machine
        And I run `add-apt-repository ppa:yellow/ua-airgapped -y` `with sudo` on the `mirror` machine
        And I run `apt-get update` `with sudo` on the `mirror` machine
        And I run `apt-get install apt-mirror get-resource-tokens ua-airgapped -yq` `with sudo` on the `mirror` machine
        And I download the service credentials on the `mirror` machine
        And I extract the `esm-infra` credentials from the `mirror` machine
        And I extract the `esm-apps` credentials from the `mirror` machine
        And I set the apt-mirror file for `<release>` with the `esm-infra,esm-apps` credentials on the `mirror` machine
        And I run `apt-mirror` `with sudo` on the `mirror` machine
        And I serve the `esm-infra` mirror using port `8000` on the `mirror` machine
        And I serve the `esm-apps` mirror using port `9000` on the `mirror` machine
        And I create the contract config overrides file for `esm-infra,esm-apps` on the `mirror` machine
        And I generate the contracts-airgapped configuration on the `mirror` machine
        
        # set up an airgapped contract server
        When I launch a `jammy` `contracts` machine
        And I run `add-apt-repository ppa:yellow/ua-airgapped -y` `with sudo` on the `contracts` machine
        And I run `apt-get update` `with sudo` on the `contracts` machine
        And I run `apt-get install contracts-airgapped -yq` `with sudo` on the `contracts` machine
        And I run `apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4067E40313CB4B13` `with sudo` on the `contracts` machine
        And I disable any internet connection on the `contracts` machine
        And I send the contracts-airgapped config from the `mirror` machine to the `contracts` machine
        And I start the contracts-airgapped service on the `contracts` machine
        
        # attach an airgapped machine to the airgapped contract server
        Given a `<release>` machine with ubuntu-advantage-tools installed
        And I disable any internet connection on the machine
        And I change config key `contract_url` to use value `contracts:ip-address:8484`
        And I attach `contract_token` with sudo
        Then stdout matches regexp:
        """
        esm-apps     +yes      +enabled .*
        esm-infra    +yes      +enabled .*
        """
        When I run `apt-cache policy hello` with sudo
        Then stdout matches regexp:
        """
        500 .*:9000/ubuntu jammy-apps-security/main
        """
        And stdout matches regexp:
        """
        500 .*:8000/ubuntu jammy-infra-security/main
        """

        Examples: ubuntu release
            | release |
            | jammy   |
```

orndorffgrant (approved):
LGTM!

My suggestion belowlooks huge but I'm really only suggesting moving the "launch a machine" steps to precisely right above where their respective machine is actually set up. Plus some small descriptive comments for the different chunks of what is happening

features/airgapped.feature
orndorffgrant: Emphasis is on these parts:

```
    Scenario Outline: Attached enable Common Criteria service in an ubuntu lxd container
        # set up the apt mirror and airgapped contract configuration
        When I launch a `jammy` `mirror` machine
```

```
        # set up an airgapped contract server
        When I launch a `jammy` `contracts` machine
```


```
        # attach an airgapped machine to the airgapped contract server
```

lucasmoura: I have updated the code to better isolate the machines, however the `given` step is a little trickier since we set up the `context.instance` attribute there and we need it on the other steps. We could move that logic to the `launch` step but I still think it is odd that we have the `given` state in the middle of the tests. But let me know if you disagree and agree with moving the `context.instance` logic to another place

renanrodrigo: imho the comments explain well the setup of the machines, and the `given` step in the start makes sense for me.

orndorffgrant: Yep this is good! No need to refactor the `context.instance` set up - forgot about that

orndorffgrant (approved):

lucasmoura:

PR #2219, opened 2022-08-31
wip: retry auto attach (SC-1011)

### LINKED ISSUE #951: auto-attach should retry on contract server issues
This is part of the spec that was not implemented due to time constraints. Without this, auto-attaching machines that launch during contract server downtime will never automatically attach.

orndorffgrant:
Fixes #951 

I'll break a number of things into separate PRs in preparation for this feature:
- [ ] ~~change notices to use ids, so that we can add/remove based on id rather than content~~ I'll work around it for now, to avoid wasted effort since we'll change notices in SC-1235
- [x] #2220 use full_auto_attach api in ua-auto-attach.service and use service setup_logging for all systemd services and grab lock for full_auto_attach api endpoint
- [ ] maybe more

## Test Steps

TODO

## Desired commit type

## Checklist


uaclient/files.py
```diff
@@ -116,6 +116,16 @@ def read(self) -> Optional[DOFType]:
 
         return self.data_object_cls.from_dict(parsed_data)
 
+    def try_read(self) -> Optional[DOFType]:
```
lucasmoura, line 119: Do we need this method ? It seems that the `read` method already returns None if the file is not found. We could add the other additional exceptions on that method as well in my opinion

uaclient/api/u/pro/attach/auto/start_retry_service/v1.py
```diff
@@ -0,0 +1,42 @@
...
+from uaclient.services import retry_auto_attach
+
+StartRetryServiceOptions = FullAutoAttachOptions
+
+
+class StartRetryServiceResult(DataObject, AdditionalInfo):
+    pass
+
+
+def start_retry_service(
```
lucasmoura, line 19: Maybe I am missing something here, but why are we adding an API endpoint for the retry ?

orndorffgrant:
thanks for taking a look here. I'm now experimenting with integrating this directly into the daemon we already have instead of adding a new service. It's looking a bit cleaner and avoids series-dependent systemd behavior.

features/ubuntu_pro.feature
```diff
@@ -677,3 +677,202 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
...
+
+
+    # really wip
+    @wip
+    @series.lts
+    @uses.config.machine_type.gcp.pro
+    Scenario Outline: auto-attach retries stop if manual auto-attach succeeds
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I change contract to staging with sudo
+        When I create the file `/tmp/response-overlay.json` with the following:
```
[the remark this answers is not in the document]
orndorffgrant: That was a mistake, this integration test was still wip at the time of review - my bad

uaclient/files.py
orndorffgrant: Good point. In the replacement PR, I instead refactored `read` to just return None if any exceptions occur

uaclient/api/u/pro/attach/auto/start_retry_service/v1.py
orndorffgrant: This will be useful for cloud-init to start the service, but maybe we don't need it in this feature. We can discuss further in the replacement PR

orndorffgrant:
replaced by #2250

PR #2222, opened 2022-09-01
Update README.md

Hooloovoo:
Add slight tweaks to the introductory paragraphs to make it clear that the tool is also useful for Ubuntu systems management.

## Proposed Commit Message
Add slight tweaks to the introductory paragraphs to make it clear that the tool is also useful for Ubuntu systems management.


lucasmoura:
Even though we already provide some system management functionalities, we are currently not the official tool for that. Because of that, we are closing this PR

PR #2228, opened 2022-09-02
security-status: remove module level logic

orndorffgrant:
With the current bug being fixed in #2217 ua fails to run on a kernel that doesn't match the broken regex.
It fails so early that it isn't even able to log anything or print why it failed. It fails that early because of this use of `get_platform_info` that happens at module import time.

The fix is to just only make this function call when it is needed and cache the result, rather than compute it at module import time

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/security_status.py
+++ b/uaclient/security_status.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 from enum import Enum
+from functools import lru_cache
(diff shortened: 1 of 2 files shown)


renanrodrigo (approved):
Hmm I wonder who wrote this horrible piece of code you're refactoring :thinking: 

The version in this changeset is definitely better

PR #2229, opened 2022-09-02
reboot-required: add api endpoint for reboot info (SC-1266)

lucasmoura:
## Proposed Commit Message
reboot-required: add api endpoint for reboot info

Add endpoint to verify the reboot state of a machine taking into
consideration livepatch as well. We have also added this information
on the output of running security-status

## Test Steps
Run the new integration and unittests

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -706,6 +706,61 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
            | focal   |
            | jammy   |
 
+    @series.xenial
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable livepatch
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        Then stdout matches regexp:
+        """
+        Installing canonical-livepatch snap
+        Canonical livepatch enabled
+        """
+        When I run `pro status` with sudo
+        Then stdout matches regexp:
+        """
+        livepatch +yes +enabled
+        """
+        When I run `pro api u.pro.security.status.reboot.v1` with sudo
+        Then stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"reboot_required": "no"}, "meta": {"environment_vars": \[\]}, "type": "RebootStatus"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+        """
+        When I create the file `/var/run/reboot-required` with the following:
+        """
+        """
+        And I run `pro api u.pro.security.status.reboot.v1` with sudo
+        Then stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"reboot_required": "yes"}, "meta": {"environment_vars": \[\]}, "type": "RebootStatus"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+        """
+        When I create the file `/var/run/reboot-required.pkgs` with the following:
+        """
+        linux-base
+        linux-image-5.4.0-1074
+        test
+        test123
+        """
+        And I run `pro api u.pro.security.status.reboot.v1` with sudo
+        Then stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"reboot_required": "yes"}, "meta": {"environment_vars": \[\]}, "type": "RebootStatus"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+        """
+        When I create the file `/var/run/reboot-required.pkgs` with the following:
+        """
+        linux-base
+        linux-image-5.4.0-1074
+        """
+        And I run `pro api u.pro.security.status.reboot.v1` with sudo
+        Then stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"reboot_required": "yes-kernel-livepatches-applied"}, "meta": {"environment_vars": \[\]}, "type": "RebootStatus"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+
     @slow
     @series.bionic
     @uses.config.machine_type.lxd.vm
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -721,40 +721,54 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         """
         livepatch +yes +enabled
         """
-        When I run `pro api u.pro.security.status.reboot.v1` with sudo
+        When I run `pro api u.pro.security.status.reboot_required.v1` with sudo
         Then stdout matches regexp:
         """
-        {"_schema_version": "v1", "data": {"attributes": {"reboot_required": "no"}, "meta": {"environment_vars": \[\]}, "type": "RebootStatus"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+        {"_schema_version": "v1", "data": {"attributes": {"reboot_required": "no"}, "meta": {"environment_vars": \[\]}, "type": "RebootRequired"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
         """
-        When I create the file `/var/run/reboot-required` with the following:
+        When I run `pro system reboot-required` as non-root
+        Then I will see the following on stdout:
         """
+        no
         """
-        And I run `pro api u.pro.security.status.reboot.v1` with sudo
+        When I run `apt-get install libc6 -y` with sudo
+        And I run `pro api u.pro.security.status.reboot_required.v1` as non-root
         Then stdout matches regexp:
         """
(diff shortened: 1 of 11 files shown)


uaclient/security_status.py
```diff
@@ -6,8 +8,17 @@
 from apt import package as apt_package
 
 from uaclient.config import UAConfig
+from uaclient.exceptions import UserFacingError
```
CalvoM, line 11: I think we could the exceptions found in `uaclient.api.exceptions`

uaclient/security_status.py
```diff
@@ -132,6 +149,70 @@ def get_ua_info(cfg: UAConfig) -> Dict[str, Any]:
...
+    for patch_status in patch_statuses:
+        if patch_status.get("Kernel") == kernel_name:
+            patch_state = patch_status.get("Livepatch", {}).get("State", "")
+
+    if patch_state == "applied":
+        return RebootStatus.REBOOT_REQUIRED_LIVEPATCH_APPLIED
+
+    # Any other Livepatch status will not be considered here to modify the
+    # reboot state
+    return RebootStatus.REBOOT_REQUIRED
```
CalvoM, line 213: If the `State` is not `applied` e.g. like on my machine I have `nothing-to-apply` why do we still have to reboot? Or is it that the fact that there are packages on the `REBOOT_PKGS_FILE_PATH` already means we need to reboot?

lucasmoura: Yes, is the fact that we have the `/var/run/reboot-required` files that state that we need a reboot even if livepatch did nothing on the system

lucasmoura:
@renanrodrigo @orndorffgrant @CalvoM I have added the CLI command for reboot-required

uaclient/cli.py
```diff
@@ -737,6 +737,30 @@ def disable_parser(parser, cfg: config.UAConfig):
...
+        This command will output one of the three following states
+        for the machine regarding reboot:
+
+        * no: The machine doesn't require a reboot
+        * yes: The machine requires a reboot
+        * yes-kernel-livepatches-applied: There are only kernel related
+          packages that require a reboot, but Livepatch has already provided
+          patches for the current running kernel. The machine still needs a
+          reboot, but you can assess if the reboot can be performed in the
+          nearest maintenance window.
```
orndorffgrant, line 759: I think the cli command should return different codes depending on the state as well

uaclient/cli.py
```diff
@@ -1580,6 +1610,11 @@ def action_status(args, *, cfg):
     return ret
 
 
+def action_reboot_required(args, cfg: config.UAConfig):
+    event.info(security_status.get_reboot_status().value)
```
orndorffgrant, line 1614: I think this should call the api instead of the internal implementation

uaclient/api/tests/test_api_u_pro_security_status_reboot_v1.py
```diff
@@ -0,0 +1,25 @@
+import mock
+import pytest
+
+from uaclient.api.u.pro.security.status.reboot.v1 import reboot_status
```
orndorffgrant, line 4: I think generally the python function name should match the api endpoint/module name. How about `u.pro.security.status.reboot_required.v1` and `reboot_required`?

features/attached_enable.feature
```diff
@@ -706,6 +706,81 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
            | focal   |
            | jammy   |
 
+    @series.xenial
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attached enable livepatch
```
orndorffgrant, line 711: Instead of manually manipulating the contents of the reboot-required files, can we get the system into "real" states that result in each reboot-required status?

I think this should be possible on a xenial vm with esm. I think there should be patches available to the kernel in apt in esm-infra.

So the system starts out as "no"
then enable esm-infra and livepatch and upgrade the kernel
now status should be "yes-kernel-livepatches-applied"
then install something else from esm-infra that requires a reboot - openssh-server?
now status should be "yes"

features/attached_enable.feature
```diff
@@ -706,6 +706,81 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
...
+        Then stdout matches regexp:
+        """
+        livepatch +yes +enabled
+        """
+        When I run `pro api u.pro.security.status.reboot.v1` with sudo
+        Then stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"reboot_required": "no"}, "meta": {"environment_vars": \[\]}, "type": "RebootStatus"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+        """
+        When I run `ua reboot-required` as non-root
```
orndorffgrant, line 729: I'm a little worried about how this will fit in with the _next_ similar request.

If we implement similar things to needsrestart, then we'll need to have another cli command to restart certain services, for example.

Maybe we should nest these system status/management functions under something else? like `pro system reboot-required` and `pro system restart-services`... Just wondering how we can start integrating thesse things without adding a new toplevel command for everything

uaclient/cli.py
lucasmoura: My only concern regarding this is that we non-zero return codes might be considered an error, which none of those states represent

lucasmoura:
@orndorffgrant PR updated

uaclient/cli.py
orndorffgrant: yeah good point

uaclient/api/u/pro/security/status/reboot_required/v1.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def reboot_required(cfg: UAConfig) -> RebootStatusResult:
+    status = get_reboot_status()
+    return RebootStatusResult(reboot_required=status.value)
+
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="RebootStatus",
```
orndorffgrant, line 27: Just for consistency with the function name, should the name here be "RebootRequired"?

uaclient/cli.py
```diff
@@ -1581,6 +1638,27 @@ def action_status(args, *, cfg):
...
+    :return: 0 on success, 1 otherwise
+    """
+    _print_help_for_subcommand(
+        cfg, cmd_name="system", subcmd_name=args.command
+    )
+    return 0
+
+
+def action_system_reboot_required(args, *, cfg: config.UAConfig):
+    result = call_api("u.pro.security.status.reboot_required.v1", [], cfg=cfg)
```
orndorffgrant, line 1653: let's just call the python function directly instead
```suggestion
    from uaclient.api.u.pro.security.status.reboot_required.v1 import reboot_required
    result = reboot_required()
```

uaclient/cli.py
```diff
@@ -1581,6 +1638,27 @@ def action_status(args, *, cfg):
...
+    event.info(
+        result.to_dict()
+        .get("data", {})
+        .get("attributes", {})["reboot_required"]
+    )
```
orndorffgrant, line 1658: if we call the function directly, we can use the typed python object instead of converting to a dict
```suggestion
    event.info(result.reboot_required)
```

uaclient/tests/test_cli_reboot_required.py
```diff
@@ -0,0 +1,37 @@
+import mock
+import pytest
+
+from uaclient.cli import main
+
+HELP_OUTPUT = """\
+usage: pro system <command> [flags] reboot-required [-h]
```
orndorffgrant, line 7: I'm not sure how to fix this but the `<command>` is misleading here. Preferably it would look like
```suggestion
usage: pro system reboot-required [-h]
```

orndorffgrant (commented):
Looks great! Just some nits and suggestion to use the python api function directly

uaclient/cli.py
lucasmoura: The only issue here is that we would be using the "private" function to avoid creating an extra unnecessary `UAConfig` instance, but this is a minor issue

uaclient/tests/test_cli_reboot_required.py
```diff
@@ -0,0 +1,37 @@
+import mock
+import pytest
+
+from uaclient.cli import main
+
+HELP_OUTPUT = """\
+usage: pro system reboot-required [flags]
```
renanrodrigo, line 7: nit: `[flags]` here should be just `[-h]`, as there are no real flags to pass

uaclient/api/api.py
```diff
@@ -22,6 +22,7 @@
     "u.pro.attach.magic.revoke.v1",
     "u.pro.attach.auto.should_auto_attach.v1",
     "u.pro.attach.auto.full_auto_attach.v1",
+    "u.pro.security.status.reboot_required.v1",
```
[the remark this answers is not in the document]
renanrodrigo: +1 keep status

renanrodrigo (approved):
this lgtm

uaclient/api/api.py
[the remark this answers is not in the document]
orndorffgrant: sounds good lets keep status

orndorffgrant (approved):
LGTM! just needs the conflict resolved

the test failure looks like it was due to the launchpad outage - let's just let CI pass after the rebase before merging

lucasmoura:

PR #2232, opened 2022-09-06
tests: add proxy test for a system without internet

renanrodrigo:
## Desired commit type

## Checklist

### DIFF
--- a/features/proxy_config.feature
+++ b/features/proxy_config.feature
@@ -1183,3 +1183,34 @@ Feature: Proxy configuration
            | bionic  |
            | focal   |
            | jammy   |
+
+    @slow
+    @series.jammy
+    @uses.config.machine_type.lxd.vm
+    Scenario: Enable realtime kernel through proxy on a machine with no internet
+        Given a `jammy` machine with ubuntu-advantage-tools installed
+        When I disable any internet connection on the machine
+        And I launch a `focal` `proxy` machine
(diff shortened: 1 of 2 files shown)


features/proxy_config.feature
```diff
@@ -1183,3 +1183,34 @@ Feature: Proxy configuration
            | bionic  |
            | focal   |
            | jammy   |
+
+    @slow
+    @series.jammy
+    @uses.config.machine_type.lxd.vm
+    Scenario: Enable realtime kernel through proxy on a machine with no internet
+        Given a `jammy` machine with ubuntu-advantage-tools installed
+        When I disable any internet connection on the machine
```
lucasmoura, line 1192: That's really great :rocket: 
But I think we can add that step to the other releases as well right ?

renanrodrigo: I used this one here to test the realtime-kernel enablement (which implies Jammy).
If you think we can have other tests using those features, then yes, we can add more
Just let me know which ones should be in this PR

orndorffgrant: I think this is good for now, we can expand later if needed.

lucasmoura: I think you are right, we could add other similar tests for the older releases, but maybe it is not required

lucasmoura (approved):

orndorffgrant (approved):

PR #2234, opened 2022-09-07
undo parsing /proc/version_signature

### LINKED ISSUE #2233: Regression: ubuntu-advantage-tools no longer installs on Ubuntu Xenial with an AWS kernel
We noticed that `ubuntu-advantage-tools` no longer installs on Ubuntu Xenial with the following error. This error seems to have been caused by a new version of `ubuntu-advantage-tools`. It looks like https://github.com/canonical/ubuntu-advantage-client/commit/bfc150e9ae937a4dcfb245c8b901400d4107a772 might be the cause.

```
[2022-09-07T19:24:26.631Z] #12 23.58 Setting up ubuntu-advantage-tools (28.0-1806~g1edfc5cb~ubuntu16.04.1) ...
[2022-09-07T19:24:26.886Z] #12 23.66 debconf: unable to initialize frontend: Dialog
[2022-09-07T19:24:26.886Z] #12 23.66 debconf: (TERM is not set, so the dialog frontend is not usable.)
[2022-09-07T19:24:26.886Z] #12 23.66 debconf: falling back to frontend: Readline
[2022-09-07T19:24:26.886Z] #12 23.66 debconf: unable to initialize frontend: Readline
[2022-09-07T19:24:26.886Z] #12 23.66 debconf: (Can't locate Term/ReadLine.pm in @INC (you may need to install the Term::ReadLine module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.22.1 /usr/local/share/perl/5.22.1 /usr/lib/x86_64-linux-gnu/perl5/5.22 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.22 /usr/share/perl/5.22 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base .) at /usr/share/perl5/Debconf/FrontEnd/Readline.pm line 7.)
[2022-09-07T19:24:26.886Z] #12 23.66 debconf: falling back to frontend: Teletype
[2022-09-07T19:24:28.248Z] #12 25.06 Processing triggers for libc-bin (2.23-0ubuntu11.3) ...
[2022-09-07T19:24:28.248Z] #12 25.26 failed to process /proc/version_signature. using uname for all kernel info
[2022-09-07T19:24:28.503Z] #12 25.26 failed to process /proc/version_signature. using uname for all kernel info
[2022-09-07T19:24:29.064Z] #12 25.98 Failed to parse kernel: 5.4.196-108.356.amzn2.x86_64
```

orndorffgrant:
We don't actually need all the granular info from
/proc/version_signature so we're no longer using a regex to parse it.

Also, raising exceptions during this function was causing unexpected
failures elsewhere, so here we just return Nones if any parsing failures
occur.

Fixes #2233 

## Test Steps

Existing integration tests should pass

## Desired commit type

## Checklist


lucasmoura (commented):
@orndorffgrant on `security_status` line 165 we are making a comparison of `proc_version_signature_version` and the kernel information provided by Livepatch. In a really unlikely scenario where both Livepatch doesn't provide kernel info and we cannot parse the `proc` file, both values will be `None` and the check will pass. I think we can add a `is not None` check on that `if` condition just to be safe

uaclient/system.py
```diff
@@ -59,96 +41,51 @@
...
-    proc_version_signature_version = ""
+    proc_version_signature_version = None
     try:
         proc_version_signature_full = load_file("/proc/version_signature")
         proc_version_signature_version = proc_version_signature_full.split()[1]
     except Exception:
+        logging.warning("failed to process /proc/version_signature.")
+
+    uname_release = os.uname().release.strip()
+    match = re.match(RE_KERNEL_UNAME, uname_release)
```
lucasmoura, line 64: And just a nitpick,  that was probably already here in the code, but let's just rename that variable, since `match` seems to be a reserved name in python

renanrodrigo (commented):
There is a call to `get_kernel_info` in `security_status.py:181` which is `except`ing something that will not happen anymore - I think that check can be simplified there

edit: oh my @lucasmoura was 5min faster haha
But yeah we speaking about the same chunk of code

orndorffgrant:

lucasmoura (approved):

orndorffgrant:
oops I guess I missed something, fixing...

orndorffgrant:
hopefully CI will be happy now

PR #2236, opened 2022-09-08
status: make lock world readable so non-root status --wait works

orndorffgrant:
## Test Steps

New integration test should pass

## Desired commit type

## Checklist

### DIFF
--- a/features/attached_status.feature
+++ b/features/attached_status.feature
@@ -20,6 +20,46 @@ Feature: Attached status
            | kinetic |
 
     @series.xenial
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Non-root status can see in-progress operations
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token` with sudo
+        When I run shell command `sudo pro enable cis >/dev/null & pro status` as non-root
+        Then stdout matches regexp:
+        """
+        NOTICES
+        Operation in progress: pro enable
+        """
+        When I run `pro status --wait` as non-root
+        When I run `pro disable cis --assume-yes` with sudo
+        When I run shell command `sudo pro enable cis & pro status --wait` as non-root
+        Then stdout matches regexp:
+        """
+        One moment, checking your subscription first
+        Updating package lists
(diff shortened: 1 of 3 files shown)


uaclient/config.py
```diff
@@ -86,7 +86,7 @@ class UAConfig:
     data_paths = {
         "instance-id": DataPath("instance-id", True, False),
         "machine-access-cis": DataPath("machine-access-cis.json", True, False),
-        "lock": DataPath("lock", True, False),
+        "lock": DataPath("lock", False, False),
```
blackboxsw, line 89: Confirmed behavior on system with boh auto-attach in progress and/or manual ua attach/detach.

This visibility of the lock file also ensures `ua status --wait` blocks for non-root user until operation is complete.

features/attached_status.feature
```diff
@@ -19,6 +19,39 @@ Feature: Attached status
...
+        When I attach `contract_token` with sudo
+        When I run shell command `sudo pro enable cis >/dev/null & pro status` as non-root
+        Then stdout matches regexp:
+        """
+        NOTICES
+        Operation in progress: pro enable
+        """
+        When I run `pro status --wait` as non-root
+        When I run `pro disable cis --assume-yes` with sudo
+        When I run shell command `sudo pro enable cis & pro status --wait` as non-root
```
blackboxsw, line 35: Nice and confims that we block an emit "...." as non-root.
Also we could verify just before this something like:
`sudo pro enable esm-apps --assume-yes  & pro status --format json | jq -r .execution_status`
matches "active" instead of "inactive"

blackboxsw (approved):
LGTM! could tweak the integration test if desired for an explicit test on json output containing the execution_status or execution_description if desired.

features/attached_status.feature
orndorffgrant: Good idea! I'll add it

orndorffgrant: Added!

renanrodrigo (approved):

lucasmoura (approved):

PR #2237, opened 2022-09-08
create public machine-token for attached machines

lucasmoura:
## Proposed Commit Message
create public machine-token for attached machines

On already attached machines, we won't have the public machine-token file. We need that file for non-root status calls. Because of that, we are creating that file during postinst to guarantee the correct behavior of non-root status

## Test Steps
Run the new integration test

## Desired commit type

## Checklist

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -54,6 +54,9 @@ SYSTEMD_HELPER_ENABLED_WANTS_LINK="/var/lib/systemd/deb-systemd-helper-enabled/m
 REBOOT_CMD_MARKER_FILE="/var/lib/ubuntu-advantage/marker-reboot-cmds-required"
 OLD_LICENSE_CHECK_MARKER_FILE="/var/lib/ubuntu-advantage/marker-license-check"
 
+MACHINE_TOKEN_FILE="/var/lib/ubuntu-advantage/private/machine-token.json"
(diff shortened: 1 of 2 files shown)


debian/ubuntu-advantage-tools.postinst
```diff
@@ -368,6 +371,15 @@ remove_old_systemd_units() {
     fi
 }
 
+create_public_machine_token_file() {
+    /usr/bin/python3 -c "
```
orndorffgrant, line 375: maybe add a comment here explaining that the MachineTokenFile class handles redacting secrets from the private file and writing the redacted version to the public file

orndorffgrant (approved):
LGTM, just pending CI

renanrodrigo (approved):
lgtm

debian/ubuntu-advantage-tools.postinst
```diff
@@ -368,6 +371,19 @@ remove_old_systemd_units() {
     fi
 }
 
+create_public_machine_token_file() {
+    # When we perform the write operation through
+    # MachineTokenFile we already write the public
+    # version of the machine-token file if all the
```
renanrodrigo, line 377:
```suggestion
    # version of the machine-token file with all the
```

CalvoM (approved):
LGTM

debian/ubuntu-advantage-tools.postinst
lucasmoura: Good catch. I have just fixed it

PR #2239, opened 2022-09-09
tests: organize behave steps

orndorffgrant:
## Note
This is only intended to organize our existing steps, not to change any of them.
After they're organized (after this is merged), we will be able to consolidate or remove unneeded steps or do other refactors more easily

## Test Steps

Integration tests should still pass.
Check my organization and tell me if it doesn't make sense or if something is mis-categorized or anything like that.

## Desired commit type

## Checklist


features/steps/attach.py
```diff
@@ -0,0 +1,119 @@
...
+        ),
+        user_spec="as non-root",
+    )
+
+
+@when("I wait for the magic attach token to be activated")
+def when_i_wait_for_magic_attach_token(context):
+    when_i_run_command(
+        context=context,
+        command="pro api u.pro.attach.magic.wait.v1 --args magic_token={}".format(  # noqa
```
lucasmoura, line 47: Can we also isolate the magic-attach steps in a separate file ?

lucasmoura (commented):
Great refactoring @orndorffgrant
Just a minor nit to also isolate the magic-attach logic

features/steps/attach.py
orndorffgrant: Will do!

renanrodrigo (approved):
this is great

PR #2247, opened 2022-09-15
contract-expiry: update expiry to updated motd

CalvoM:
## Proposed Commit Message

contract-expiry: update expiry to update motd

Updating the contract expiry date when the expiry date is approaching or lapsed.

## Desired commit type

## Checklist

### DIFF
--- a/features/apt_messages.feature
+++ b/features/apt_messages.feature
@@ -128,7 +128,21 @@ Feature: APT Messages
         Calculating upgrade...
         The following packages will be upgraded:
         """
-        When I update contract to use `effectiveTo` as `days=+2`
+        When I create the file `/tmp/machine-token-overlay.json` with the following:
+        """
+        {
+            "machineTokenInfo": {
+                "contractInfo": {
+                    "effectiveTo": <now+2>
+                }
+            }
+        }
+        """
+        And I append the following on uaclient config:
+        """
+        features:
+          machine_token_overlay: "/tmp/machine-token-overlay.json"
+        """
         When I run `pro refresh messages` with sudo
         When I run `apt-get upgrade --dry-run` with sudo
         Then stdout matches regexp:
@@ -144,7 +158,16 @@ Feature: APT Messages
 
         The following packages will be upgraded:
         """
-        When I update contract to use `effectiveTo` as `days=-3`
+        When I create the file `/tmp/machine-token-overlay.json` with the following:
+        """
+        {
+            "machineTokenInfo": {
+                "contractInfo": {
+                    "effectiveTo": <now-3>
+                }
+            }
+        }
+        """
         When I run `pro refresh messages` with sudo
         When I run `apt-get upgrade --dry-run` with sudo
         Then stdout matches regexp:
@@ -161,7 +184,16 @@ Feature: APT Messages
 
         The following packages will be upgraded:
         """
-        When I update contract to use `effectiveTo` as `days=-20`
+        When I create the file `/tmp/machine-token-overlay.json` with the following:
+        """
+        {
+            "machineTokenInfo": {
+                "contractInfo": {
+                    "effectiveTo": <now-20>
+                }
+            }
+        }
+        """
         When I run `pro refresh messages` with sudo
         When I run `apt-get upgrade --dry-run` with sudo
         Then stdout matches regexp:
@@ -240,23 +272,37 @@ Feature: APT Messages
         The following packages will be upgraded:
           hello
         """
-#        When I update contract to use `effectiveTo` as `days=-20`
-#        When I run `pro refresh messages` with sudo
-#        When I run `apt-get upgrade --dry-run` with sudo
-#        Then stdout matches regexp:
-#        """
-#        Reading package lists...
-#        Building dependency tree...
-#        Reading state information...
-#        Calculating upgrade...
-#
-#        \*Your Ubuntu Pro subscription has EXPIRED\*
-#        The following security updates require Ubuntu Pro with 'esm-apps' enabled:
-#          hello
-#        Renew your service at https:\/\/ubuntu.com\/pro
-#
-#        The following packages will be upgraded:
-#        """
+        # When I create the file `/tmp/machine-token-overlay.json` with the following:
+        # """
+        # {
+        #     "machineTokenInfo": {
+        #         "contractInfo": {
+        #             "effectiveTo": <now-20>
+        #         }
+        #     }
+        # }
+        # """
+        # And I append the following on uaclient config:
+        # """
+        # features:
+        #   machine_token_overlay: "/tmp/machine-token-overlay.json"
+        # """
+        # When I run `pro refresh messages` with sudo
+        # When I run `apt-get upgrade --dry-run` with sudo
+        # Then stdout matches regexp:
+        # """
+        # Reading package lists...
+        # Building dependency tree...
+        # Reading state information...
+        # Calculating upgrade...
+
+        # \*Your Ubuntu Pro subscription has EXPIRED\*
+        # The following security updates require Ubuntu Pro with 'esm-apps' enabled:
+        #   hello
+        # Renew your service at https:\/\/ubuntu.com\/pro
+
+        # The following packages will be upgraded:
+        # """
         When I run `apt-get upgrade -y` with sudo
         When I run `pro detach --assume-yes` with sudo
         When I run `pro refresh messages` with sudo
--- a/features/steps/files.py
+++ b/features/steps/files.py
@@ -1,5 +1,7 @@
+import datetime
 import json
 import os
+import re
 import tempfile
 
 from behave import then, when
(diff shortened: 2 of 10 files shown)


uaclient/jobs/update_messaging.py
```diff
@@ -390,3 +397,25 @@ def refresh_motd():
             logging.exception(exc)
 
     system.subp(["sudo", "systemctl", "restart", "motd-news.service"])
+
+
+def update_contract_expiry(cfg: config.UAConfig):
+    orig_token = cfg.machine_token
+    machine_token = orig_token["machineToken"]
+    contract_id = orig_token["machineTokenInfo"]["contractInfo"]["id"]
+    contract_client = contract.UAContractClient(cfg)
```
lucasmoura, line 406: Just to be safe, let's add an exception handling here for the case where we don't have those keys in the dict

uaclient/jobs/update_messaging.py
```diff
@@ -390,3 +397,25 @@ def refresh_motd():
             logging.exception(exc)
 
     system.subp(["sudo", "systemctl", "restart", "motd-news.service"])
+
+
+def update_contract_expiry(cfg: config.UAConfig):
```
lucasmoura, line 402: Let's also add unittests for this function

features/apt_messages.feature
```diff
@@ -128,6 +128,21 @@ Feature: APT Messages
...
+                }
+            }
+        }
+        """
+        And I append the following on uaclient config:
+        """
+        features:
+          machine_token_overlay: "/tmp/machine-token-overlay.json"
+        """
         When I update contract to use `effectiveTo` as `days=+2`
```
orndorffgrant, line 146: Since we're using the token overlay, do we need these "When I update contract to use `effectiveTo` as `days=+2`" lines anymore?

uaclient/jobs/update_messaging.py
```diff
@@ -390,3 +397,25 @@ def refresh_motd():
...
+    orig_token = cfg.machine_token
+    machine_token = orig_token["machineToken"]
+    contract_id = orig_token["machineTokenInfo"]["contractInfo"]["id"]
+    contract_client = contract.UAContractClient(cfg)
+    resp = contract_client.request_machine_token_update(
+        machine_token=machine_token,
+        contract_id=contract_id,
+        machine_id=None,
+        update_write=False,
+    )
```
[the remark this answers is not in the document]
CalvoM: Oh, thanks for this function. It solves it well. @lucasmoura do you think we should still refactor the `request_machine_token_request` even though I will replace its use in the `update_contract_expiry` with `contract.get_updated_contract_info`?

features/apt_messages.feature
CalvoM: Yes we need, since we will first read the `machine-token.json` file then we know that the contract is approaching expiry for us to contact the contract server and get the overlay response.

uaclient/jobs/update_messaging.py
CalvoM: I have updated and used `.get` method to cover any exceptions from `KeyError` exceptions

uaclient/contract.py
```diff
@@ -333,12 +339,13 @@ def _request_machine_token_update(
         if headers.get("expires"):
             response["expires"] = headers["expires"]
         if not detach:
-            self.cfg.machine_token_file.write(response)
-            system.get_machine_id.cache_clear()
-            machine_id = response.get("machineTokenInfo", {}).get(
-                "machineId", data.get("machineId")
-            )
-            self.cfg.write_cache("machine-id", machine_id)
+            if update_write:
```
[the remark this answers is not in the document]
CalvoM: @lucasmoura  it seem [here](https://github.com/canonical/ubuntu-advantage-client/blob/2a9515f5e6436d955e439bcdd306e9eff0ce3958/uaclient/contract.py#L135) was where was last used. So we can remove the `detach` parameter.

features/apt_messages.feature
orndorffgrant: the machine token overlay is applied at machine token read time, not at contract server request time. So I'm still not sure this line is necessary :upside_down_face:

renanrodrigo: @CalvoM keep in mind we have 2 different test overlays, one for the machine token, another one for the contract server response

CalvoM: Oh, yes I had mixed up the two overlays, thanks guys!

uaclient/tests/test_contract.py
```diff
@@ -49,8 +49,8 @@ class TestUAContractClient:
         "machine_id_response", (("contract-machine-id"), None)
     )
     @pytest.mark.parametrize(
-        "detach,expected_http_method",
-        ((None, "POST"), (False, "POST"), (True, "DELETE")),
+        "expected_http_method",
+        (("POST"), ("POST")),
```
orndorffgrant, line 53: the value no longer changes so I don't think we need this `@pytest.mark.parametrize` anymore
we can just hardcode "POST" in the body of the test

CalvoM: Yes, on it.

orndorffgrant (approved):
Looks good! Just needs conflicts resolved

features/steps/files.py
```diff
@@ -38,11 +40,25 @@ def when_i_create_file_with_content(context, file_path, machine="uaclient"):
...
         text = text.replace("<ci-proxy-ip>", context.instances["proxy"].ip)
+
+    date_match = re.search(r"<now(?P<offset>.*)>", text)
+    if date_match:
+        day_offset = date_match.group("offset")
+        offset = 0 if day_offset == "" else int(day_offset)
+        now = datetime.datetime.utcnow()
+        contract_expiry = now + datetime.timedelta(days=offset)
+        new_value = (
+            '\\"' + contract_expiry.strftime("%Y-%m-%dT00:00:00Z") + '\\"'
```
orndorffgrant, line 51:
```suggestion
            '"' + contract_expiry.strftime("%Y-%m-%dT00:00:00Z") + '"'
```
I think that will fix the tests

CalvoM: Oh thanks, I was thinking the same, but wondering why it only affects xenial and focal lxd alone.
Thanks.

uaclient/jobs/update_messaging.py
```diff
@@ -417,3 +424,32 @@ def refresh_motd():
...
+def update_contract_expiry(cfg: config.UAConfig):
+    orig_token = cfg.machine_token
+    machine_token = orig_token.get("machineToken", "")
+    contract_id = (
+        orig_token.get("machineTokenInfo", {})
+        .get("contractInfo", {})
+        .get("id", "")
+    )
+    contract_client = contract.UAContractClient(cfg)
+    resp = contract_client.get_updated_contract_info(
```
lucasmoura, line 438: If we can't get the `contract id` here, I don't think we should make the request, because it will only fail here and raise an Exception to the job. I think we should only proceed if the id is valid

CalvoM: Makes sense. I think we should also update `is_contract_changed` as well.

uaclient/jobs/update_messaging.py
```diff
@@ -417,3 +424,33 @@ def refresh_motd():
...
+
+def update_contract_expiry(cfg: config.UAConfig):
+    orig_token = cfg.machine_token
+    machine_token = orig_token.get("machineToken", "")
+    contract_id = (
+        orig_token.get("machineTokenInfo", {})
+        .get("contractInfo", {})
+        .get("id", "")
+    )
+    if len(contract_id) > 0:
```
orndorffgrant, line 437: I see the comment chain of why this was added. I just have a little nit :upside_down_face: 

Checking the length of the string implies that we care about the length. But we don't care about the length, we only care if it is empty or not. So I suggest changing this to:
```suggestion
    if contract_id != "":
```

orndorffgrant: The same applies to the other place where this was added

renanrodrigo (approved):
LGTM  - only need to address Grant's comments

uaclient/jobs/update_messaging.py
CalvoM: Makes sense.

uaclient/contract.py
```diff
@@ -636,31 +639,32 @@ def is_contract_changed(cfg: UAConfig) -> bool:
...
-    )
-    if cfg.machine_token_file.contract_expiry_datetime != new_expiry:
-        return True
-    curr_entitlements = cfg.machine_token_file.get_entitlements_from_token(
-        resp
-    )
-    for name, new_entitlement in sorted(curr_entitlements.items()):
-        deltas = util.get_dict_deltas(
-            orig_entitlements.get(name, {}), new_entitlement
+    if contract_id != "":
```
lucasmoura, line 642: I think it is simpler to just return False here if the `contract_id` is None. We can check for None if we modify get to not have an empty string as a default value when we don't have the `id` key

orndorffgrant (approved):

orndorffgrant:
Oh there is what appears to be a legitimate CI failure actually

orndorffgrant:
@CalvoM I think the test is failing because MachineTokenFile caches the expiry date in files.py:229

I think we can just clear that cache variable on write:
```diff
diff --git a/uaclient/files/files.py b/uaclient/files/files.py
index b237e245..3c279984 100644
--- a/uaclient/files/files.py
+++ b/uaclient/files/files.py
@@ -94,6 +94,7 @@ class MachineTokenFile:
 
             self._machine_token = None
             self._entitlements = None
+            self._contract_expiry_datetime = None
         else:
             raise exceptions.NonRootUserError()
 
```

orndorffgrant:
@CalvoM just a reminder that I think the above diff will fix the failing test here.

uaclient/files.py
```diff
@@ -160,6 +160,7 @@ def write(self, content: dict):
 
             self._machine_token = None
             self._entitlements = None
+            self._contract_expiry_datetime = None
```
CalvoM, line 163: @orndorffgrant so apparently I had done it in the wrong file

Tyrael:
for the record this change is preventing ua status from working on an xenial machine with expired subscription (procurement hell):

> 2023-01-19 06:19:52,411 - cli.py:(1902) [ERROR]: Unhandled exception, please file a bug
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1846, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1959, in main
    return_value = args.action(args, cfg=cfg)
  File "/usr/lib/python3/dist-packages/uaclient/cli.py", line 1624, in action_status
    if contract.is_contract_changed(cfg):
  File "/usr/lib/python3/dist-packages/uaclient/contract.py", line 657, in is_contract_changed
    else cfg.machine_token_file.contract_expiry_datetime
  File "/usr/lib/python3/dist-packages/uaclient/files/files.py", line 236, in contract_expiry_datetime
    ]["contractInfo"]["effectiveTo"]
KeyError: 'effectiveTo'

based on the payload dump ommitted there is no effectiveTo field present under contractInfo

orndorffgrant:
Hi @Tyrael sorry for the delayed response. We have fixed this issue in https://github.com/canonical/ubuntu-advantage-client/pull/2421/files and the fix will be included in the next version which should be released in a couple days. I hope this helps.

PR #2250, opened 2022-09-16
wip: retry auto attach 2

orndorffgrant:
## Proposed Commit Message

```
summary: no more than 70 characters

A description of what the change being made is and why it is being
made, if the summary line is insufficient.  The blank line above is
required. This should be wrapped at 72 characters, but otherwise has
no particular length requirements.

If you need to write multiple paragraphs, feel free.

LP: #NNNNNNN (replace with the appropriate Launchpad bug reference if applicable)
Fixes: #NNNNNNN (replace with the appropriate github issue if applicable)
```

## Desired commit type

## Checklist


features/steps/files.py
```diff
@@ -31,13 +32,23 @@ def when_i_verify_file_is_empty_on_machine(context, file_name, instance_name):
 
 
 @when("I create the file `{file_path}` with the following")
-def when_i_create_file_with_content(context, file_path, machine="uaclient"):
-    text = context.text.replace('"', '\\"')
+def when_i_create_file_with_content(context, file_path):
```
orndorffgrant, line 35: lost the machine bit on a rebase

orndorffgrant:
some staging expired tests need updating for the new token - I'll fix that in a separate PR

features/steps/files.py
```diff
@@ -32,12 +36,23 @@ def when_i_verify_file_is_empty_on_machine(context, file_name, instance_name):
...
+        tmpf_path = os.path.join(tmpd, "tmpfile")
+        with open(tmpf_path, mode="w") as tmpf:
+            tmpf.write(text)
+        context.instances[machine].push_file(tmpf_path, "/tmp/behave_tmpfile")
+    when_i_run_command_on_machine(
+        context,
+        "cp /tmp/behave_tmpfile {}".format(file_path),
+        "with sudo",
+        machine,
+    )
```
lucasmoura, line 55: That's a great improvement on creating the files

lib/auto_attach.py
```diff
@@ -50,21 +54,41 @@ def check_cloudinit_userdata_for_ua_info():
...
+    )
+    try:
+        full_auto_attach(FullAutoAttachOptions())
+    except AlreadyAttachedError as e:
+        logging.info(e.msg)
+    except Exception as e:
+        logging.error(e)
+        system.remove_file(AUTO_ATTACH_STATUS_MOTD_FILE)
+        logging.info("creating flag file to trigger retries")
+        system.create_file(retry_auto_attach.FLAG_FILE_PATH)
```
lucasmoura, line 79: I am wondering if we should only retry on the following exceptions:

* `ConnectivityError`
* `ContractAPIError`
* `LockHeldError`

The other seem that will not benefit much by triggering the retry job

lib/daemon.py
```diff
@@ -31,7 +35,17 @@ def main() -> int:
 
     notify("READY=1")
 
-    daemon.poll_for_pro_license(cfg)
+    if os.path.exists("/run/cloud-init/cloud-id-gce") and not os.path.exists(
```
lucasmoura, line 38: Isn't the `should_poll_for_pro_license` already handle that case ? I mean it should only return `True` for `gcp` instances right ?

uaclient/daemon/retry_auto_attach.py
```diff
@@ -0,0 +1,228 @@
...
+    RetryOptions,
+    files.UAFile(
+        "retry-auto-attach-options.json",
+        private=False,
+    ),
+    files.DataObjectFileFormat.JSON,
+)
+STATE_FILE = files.DataObjectFile(
+    RetryState,
+    files.UAFile("retry-auto-attach-state.json", private=False),
```
lucasmoura, line 73: I wonder if this file should be private. My only concern is just that it is possible to manipulate the failure message here

uaclient/daemon/retry_auto_attach.py
```diff
@@ -0,0 +1,228 @@
...
+            api_exceptions.EntitlementNotFoundError,
+            api_exceptions.IncompatibleEntitlementsDetected,
+        ) as e:
+            LOG.warning(
+                "auto attach failed because of args. error message: {}".format(
+                    e.msg
+                )
+            )
+            if persisted_options is not None:
+                LOG.info("trying again with defaults")
```
lucasmoura, line 132: Sorry, I may have missed or forgotten about a discussion around this, but isn't it risk if we `auto-attach` without the user options ? We could enable unwanted services on this setup right ?

uaclient/daemon/retry_auto_attach.py
```diff
@@ -0,0 +1,228 @@
...
+        # skip intervals we've already waited
+        offset = persisted_state.interval_index
+        intervals = RETRY_INTERVALS[offset:]
+        failure_reason = persisted_state.failure_reason
+    else:
+        offset = 0
+        intervals = RETRY_INTERVALS
+        failure_reason = None
+
+    print(intervals)
```
lucasmoura, line 162: We can remove this bit

uaclient/daemon/retry_auto_attach.py
```diff
@@ -0,0 +1,228 @@
...
+            pass
+        auto_attach_status_msg = messages.AUTO_ATTACH_RETRY_NOTICE.format(
+            num_attempts=offset + index + 1,
+            reason=msg_reason,
+            next_run_datestring=next_attempt.isoformat(),
+        )
+        system.write_file(
+            AUTO_ATTACH_STATUS_MOTD_FILE, auto_attach_status_msg + "\n\n"
+        )
+        cfg.notice_file.remove("", messages.AUTO_ATTACH_RETRY_NOTICE_PREFIX)
```
lucasmoura, line 188: Should we consider acquiring a lock to write the notices file ? I know we are doing this on other service we have, but as we are adding more and more parallel jobs, we could start to see odd states in the future

features/api_configure_retry_service.feature
```diff
@@ -0,0 +1,59 @@
+Feature: api.u.pro.attach.auto.configure_retry_service
+
+    @series.lts
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: v1 successfully triggers retry service when run during startup
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I change contract to staging with sudo
+        When I create the file `/lib/systemd/system/apitest.service` with the following
```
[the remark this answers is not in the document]
orndorffgrant: No lock conflict should be happening.

This test is for running the `configure_retry_service`, which deserves a better explanation and documentation. I'll include that as soon as the api docs PR is in `main`.

Right now, the `configure_retry_service` function is written such that if it is called before `ubuntu-advantage.service` runs, then `ubuntu-advantage.service` will go into "retry" mode. `configure_retry_service` does not do anything that actually causes `ubuntu-advantage.service` to start (no `systemctl start ubuntu-advantage.service` and no path file or anything like that). 

That is because the intended use case is for cloud-init to call during boot before the `ubuntu-advantage.service` ConditionPathExists are checked. That is also why I named it `configure_retry_service` instead of `start_retry_service`.

features/environment.py
```diff
@@ -331,6 +337,8 @@ def from_environ(cls, config) -> "UAClientBehaveConfig":
 def before_all(context: Context) -> None:
     """behave will invoke this before anything else happens."""
     context.config.setup_logging()
+    logging.getLogger("botocore").setLevel(logging.INFO)
```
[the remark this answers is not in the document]
orndorffgrant: Sorry, the intent here is to actually lower the verbosity. I often run the tests with `--log-level=debug` which triggers an inordinate amount of `boto` logging for aws tests. By setting `boto` to "info" then it stays quiet even when the root logger is at "debug".

I'll remove this and give it it's own PR - it might deserve something smarter like, "only set to info if the root level is debug"

features/retry_auto_attach.feature
```diff
@@ -0,0 +1,366 @@
+Feature: auto-attach retries periodically on failures
+
+    @series.lts
+    @uses.config.machine_type.aws.generic
+    @uses.config.machine_type.azure.generic
+    @uses.config.machine_type.gcp.generic
+    Scenario Outline: auto-attach retries for a month and updates status
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I change contract to staging with sudo
```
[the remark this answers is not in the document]
orndorffgrant: We're causing errors on the contract server here and I don't want to pollute production with our test errors

lib/daemon.py
orndorffgrant: Yes, I just want to be able to look at this file and at the `ubuntu-advantage.service` ConditionPathExists rules, and be able to easily audit that each rule triggers a different background activity.

Without this check here, a reviewer would need deeper knowledge of the `should_poll_for_pro_license` function (like you have :smile: ). I'm also worried about future refactors changing that check and us accidentally ending up in the wrong mode.

uaclient/daemon/retry_auto_attach.py
orndorffgrant: Doesn't hurt to make it private and can avoid things like that. I'll change it

uaclient/daemon/retry_auto_attach.py
orndorffgrant: Let's meet and discuss

uaclient/daemon/retry_auto_attach.py
orndorffgrant: oops yep, thanks for catching that

uaclient/daemon/retry_auto_attach.py
orndorffgrant: Yeah, doesn't hurt to grab it with a spin lock, and we can just `pass` on a LockHeldError

lib/auto_attach.py
orndorffgrant: I think I'll do the NonAutoAttachImageError refactor in a different PR, to be merged before this one

features/environment.py
```diff
@@ -484,6 +494,16 @@ def before_scenario(context: Context, scenario: Scenario):
         "status",
         "ua-reboot-cmds.service",
     ],
+    "systemctl-status-pro-auto-attach-retry": [
+        "systemctl",
+        "status",
+        "pro-auto-attach-retry.service",
+    ],
```
orndorffgrant, line 501: I need to remove this

uaclient/api/tests/test_api_u_pro_version.py
```diff
@@ -14,13 +14,11 @@ def test_version(self, _m_get_version, FakeConfig):
...
-            "something wrong", "fn-specific"
-        )
+        m_get_version.side_effect = UserFacingError("something wrong")
 
         with pytest.raises(VersionError) as excinfo:
             _version(FakeConfig())
 
         assert excinfo.errisinstance(VersionError)
-        assert excinfo.value.msg == "('something wrong', 'fn-specific')"
+        assert excinfo.value.msg == "something wrong"
```
orndorffgrant, line 23: This wasn't really necessary, I'll undo it

uaclient/daemon/retry_auto_attach.py
orndorffgrant: We decided that the full_auto_attach api function should actually do a best-effort attempt to attach and enable services instead of failing early if it detects a bad argument. 
That change will make this fallback to auto-attach without user options unnecessary

lib/auto_attach.py
lucasmoura: I think now after all the refactoring, we might reduce the list of exceptions to just:

* ConnectivityError
* ContractAPIError
* UrlError
* LockHeldError
* InvalidProImage


I still don't know if we should try on `UserFacingError`. We could be masking a real issue for a month and a cannot think of a case where a retry might solve it. But I know retrying on that should be useful for solving those unpredictable situations, I am just still not sure if it can do more harm than good here

uaclient/daemon/retry_auto_attach.py
```diff
@@ -0,0 +1,183 @@
...
+            LOG.info("already attached, ending retry service")
+            break
+        except api_exceptions.EntitlementsNotEnabledError as e:
+            LOG.warning(e.msg)
+            break
+        except Exception as e:
+            failure_reason = full_auto_attach_exception_to_failure_reason(e)
+            LOG.error(e)
+
+    cleanup(cfg)
```
lucasmoura, line 162: I could be missing something here, but if we cleanup the `retry_auto_attach_options_file` here, wouldn't we have a problem reading that file again during the retry ?

uaclient/daemon/retry_auto_attach.py
```diff
@@ -0,0 +1,183 @@
...
+            msg_reason = "an unknown error"
+        auto_attach_status_msg = (
+            messages.AUTO_ATTACH_RETRY_TOTAL_FAILURE_NOTICE.format(
+                num_attempts=len(RETRY_INTERVALS) + 1, reason=msg_reason
+            )
+        )
+        system.write_file(
+            AUTO_ATTACH_STATUS_MOTD_FILE, auto_attach_status_msg + "\n\n"
+        )
+        cfg.notice_file.add("", auto_attach_status_msg)
```
lucasmoura, line 183: Just to double check, we are actually removing that notice in the `cleanup` function ? It seems to be a custom message which is not covered by `AUTO_ATTACH_RETRY_NOTICE_PREFIX`

features/api_configure_retry_service.feature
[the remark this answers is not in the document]
lucasmoura: Got it and totally agree that we can a doc explanation for it

uaclient/daemon/retry_auto_attach.py
```diff
@@ -0,0 +1,183 @@
...
+        time.sleep(interval)
+
+        if cfg.is_attached:
+            # We attached while sleeping - hooray!
+            break
+
+        try:
+            persisted_options = (
+                state_files.retry_auto_attach_options_file.read()
+            )
```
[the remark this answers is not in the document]
orndorffgrant: My thought was that if the options were causing problems then someone might edit them in between retrys. But maybe that's not likely enough to take into account. What do you think?

uaclient/daemon/retry_auto_attach.py
orndorffgrant: This is after the retry loop so it should only happen after all retries have occurred, in which case we don't need the options or state file anymore.

features/api_configure_retry_service.feature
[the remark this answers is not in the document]
orndorffgrant: Added!

uaclient/daemon/retry_auto_attach.py
```diff
@@ -0,0 +1,183 @@
...
+                )
+        else:
+            failure_reason = "a network error"
+        failure_reason += ': "{}"'.format(str(e))
+        return failure_reason
+    elif isinstance(e, api_exceptions.UserFacingError):
+        return '"{}"'.format(e.msg)
+    else:
+        LOG.error("Unexpected exception: {}".format(e))
+        return str(e) or "an unknown error"
```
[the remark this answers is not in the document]
orndorffgrant: Updated!

uaclient/daemon/retry_auto_attach.py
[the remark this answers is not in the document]
renanrodrigo: My 5c: I don't think we lose anything by reading this and it covers the scenario where options are changed for any reason. I wouldn't mind keeping this here.

renanrodrigo (approved):
Arrived late to the party here, but imho all looks good

uaclient/daemon/retry_auto_attach.py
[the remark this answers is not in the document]
lucasmoura: Fair enough, those are all good points

PR #2253, opened 2022-09-20
add API documentation (SC-1251)

renanrodrigo:
Document the current endpoints for the API, and set a basic structure for documenting the next ones.
Also, fix a couple details to make the HTML rendering look nice.

## Test steps
`tox -e docs`, open the results and behold!

## Desired commit type

## Checklist
 - [x] I have updated or added any documentation accordingly :thinking:

### DIFF
--- a/docs/references/api.md
+++ b/docs/references/api.md
@@ -1,0 +1,304 @@
+# Pro API Reference Guide
+
+
+The Pro Client has a Python-based API to be consumed by users who want to integrate the client's functionality to their software.
+The functions and objects are available through the `uaclient.api` module, and all of the available endpoints return an object with specific data for the calls.
+
+Besides importing the Python code directly, consumers who are not writing Python may use the CLI to call the same functionality, using the `pro api` command. This command will always return a JSON with a standard structure, as can be seen below:
+
+```json
+{
+    "_schema_version": "v1",
+    "data": {
+        "attributes": {
+            // <endpoint specific attributes>
+        },
+        "meta": {
+            "environment_vars": []
+        },
+        "type": "<response_type>"
+    },
+    "errors": [], 
+    "result": "<success|failure>",
+    "version": "<client version>", 
+    "warnings": []
+}
+```
+
+The currently available endpoints are:
+- [u.pro.version.v1](#uproversionv1)
+- [u.pro.attach.magic.initiate.v1](#uproattachmagicinitiatev1)
+- [u.pro.attach.magic.wait.v1](#uproattachmagicwaitv1)
+- [u.pro.attach.magic.revoke.v1](#uproattachmagicrevokev1)
+- [u.pro.attach.auto.should_auto_attach.v1](#uproattachautoshould_auto_attachv1)
+- [u.pro.attach.auto.full_auto_attach.v1](#uproattachautofull_auto_attachv1)
+
+## u.pro.version.v1
+Shows the installed Client version.
+
+### Args
+This endpoint takes no arguments.
+
+### Python API interaction
+#### Calling from Python code
+```python
+from uaclient.api.u.pro.version.v1 import version
+
+result = version()
+```
+
+#### Expected return object:
+`uaclient.api.u.pro.version.v1.VersionResult`
+
+|Field Name|Type|Description|
+|-|-|-|
+|installed_version|str|The current installed version|
+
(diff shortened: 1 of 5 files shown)


docs/references/api.md
```diff
@@ -0,0 +1,277 @@
...
+
+result = version()
+```
+
+#### Expected return object:
+`uaclient.api.u.pro.version.v1.VersionResult`
+
+|Field Name|Type|Description|Required|
+|-|-|-|-|
+|installed_version|str|The current installed version|Yes|
```
orndorffgrant, line 48: Not sure which is better, but we could embed the "required" information into the type. So this example would like:
```suggestion
|Field Name|Type|Description|
|-|-|-|-|
|installed_version|str|The current installed version|
```

and something else not required might look like

```
|Field Name|Type|Description|
|-|-|-|-|
|non_required|Optional[str]|The current installed version|
```

docs/references/api.md
```diff
@@ -0,0 +1,277 @@
...
+```bash
+pro api u.pro.version.v1
+```
+
+#### Expected JSON output:
+```json
+"attributes": {
+    "installed_version":"<version>"
+}
+```
```
orndorffgrant, line 65: Maybe just tweak this header to make it clear these attributes are part of the bigger structure at the top
```suggestion
#### Expected attributes in JSON structure
```json
{
    "installed_version": "<version>"
}
```
```

File: docs/references/api.md (hunk 1 above)
Comment (renanrodrigo) [reply docs/references/api.md]: Less is more

File: docs/references/api.md (hunk 2 above)
Comment (renanrodrigo) [reply docs/references/api.md]: makes sense

File: docs/references/api.md [hunk 3]
```diff
@@ -0,0 +1,297 @@
...
+The functions and objects are available through the `uaclient.api` module, and all of the available endpoints return an object with specific data for the calls.
+
+Besides importing the Python code directly, consumers who are not writing Python may use the CLI to call the same functionality, using the `pro api` command. This command will always return a JSON with a standard structure, as can be seen below:
+
+```json
+{
+    "_schema_version": "v1",
+    "data": {
+        "attributes": {
+            <endpoint specific attributes>
```
lucasmoura, line 13: I think we can add "_"  instead of white spaces just so we don't have that highlight on the page

lucasmoura (approved):
LGTM, my only suggestion is that maybe we can add an index on the beginning of the file to make navigation easier

CalvoM (approved):
LGTM

docs/references/api.md
renanrodrigo: Actually made it a command, so it does not break syntax

renanrodrigo:
@lucasmoura addressed!

renanrodrigo:
Lemme justify some commits. The story is:
We were not rendering the bullet points in unordered lists and spacing was horrible - you can see it if building the docs from `main`.
This seems to be a problem happening only with `sphinx_rtd_theme`, but the package to blame is `docutils` (as taken from [here](https://github.com/readthedocs/sphinx_rtd_theme/issues/1115))
tl;dr: From version `0.17`, they switched to `HTML5` rendering, and included semantic tags - `sphinx_rtd_theme` adapted to use the same from `1.0.0`, but the bullet points are still missing...
Since the latest comments point that updating everything to latest does not solve the problem, we pin `sphinx_rtd_theme` up and `docutils` down, as in [this fix](https://github.com/ing-bank/popmon/commit/63044c17151a8a0ae1fb93f13f187daf90b55714), so everything renders fine.

But AHA! We have also updated `myst` in the process, and due to a [(breaking) improvement](https://myst-parser.readthedocs.io/en/latest/develop/_changelog.html#markdown-link-resolution-improvements) it does not gererate the anchors anymore, so we need to do it [explicitly](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html?highlight=anchor#auto-generated-header-anchors).

And that's the best we have so far.

renanrodrigo:
btw @lucasmoura the JSON wrong line renders fine in the final HTML doc, so I'd like to keep it, so the highlighting is consistent with the other JSON blocks

orndorffgrant (approved):
Looks great!

Failing test is unrelated

lucasmoura (approved):

PR #2255, opened 2022-09-21
files: extract services-once-enabled from config (SC-1267)

lucasmoura:
## Proposed Commit Message
files: extract services-once-enabled from config

Extract the services-once-enabled state file from the UAConfig class and handle it on a dedicated module.

This PR also proposed a refactoring on the structure of how we manage files on UA.

## Test Steps
Run the integration scenario`Attached enable of FIPS-updates in an ubuntu lxd vm` on the `enable_fips_vm`  feature

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -9,6 +9,10 @@ from uaclient.clouds.identity import NoCloudTypeReason, get_cloud_type
 from uaclient.entitlements import repo
 from uaclient.entitlements.base import IncompatibleService
 from uaclient.entitlements.entitlement_status import ApplicationStatus
+from uaclient.files.state_files import (
+    ServicesOnceEnabledData,
+    services_once_enabled_file,
+)
 from uaclient.types import (  # noqa: F401
     MessagingOperations,
     MessagingOperationsDict,
@@ -441,30 +445,30 @@ class FIPSEntitlement(FIPSCommonEntitlement):
     def static_affordances(self) -> Tuple[StaticAffordance, ...]:
         static_affordances = super().static_affordances
 
-        fips_update = FIPSUpdatesEntitlement(self.cfg)
+        fips_updates = FIPSUpdatesEntitlement(self.cfg)
         enabled_status = ApplicationStatus.ENABLED
-        is_fips_update_enabled = bool(
-            fips_update.application_status()[0] == enabled_status
+        is_fips_updates_enabled = bool(
+            fips_updates.application_status()[0] == enabled_status
         )
 
-        services_once_enabled = (
-            self.cfg.read_cache("services-once-enabled") or {}
-        )
-        fips_updates_once_enabled = services_once_enabled.get(
-            fips_update.name, False
+        services_once_enabled_obj = services_once_enabled_file.read()
+        fips_updates_once_enabled = (
+            services_once_enabled_obj.fips_updates
+            if services_once_enabled_obj
+            else False
         )
 
         return static_affordances + (
             (
                 messages.FIPS_ERROR_WHEN_FIPS_UPDATES_ENABLED.format(
-                    fips=self.title, fips_updates=fips_update.title
+                    fips=self.title, fips_updates=fips_updates.title
                 ),
-                lambda: is_fips_update_enabled,
+                lambda: is_fips_updates_enabled,
                 False,
             ),
             (
                 messages.FIPS_ERROR_WHEN_FIPS_UPDATES_ONCE_ENABLED.format(
-                    fips=self.title, fips_updates=fips_update.title
+                    fips=self.title, fips_updates=fips_updates.title
                 ),
                 lambda: fips_updates_once_enabled,
                 False,
@@ -577,14 +581,10 @@ class FIPSUpdatesEntitlement(FIPSCommonEntitlement):
             self.cfg.notice_file.try_remove(
                 "", messages.FIPS_DISABLE_REBOOT_REQUIRED
             )
-            services_once_enabled = (
-                self.cfg.read_cache("services-once-enabled") or {}
-            )
-            services_once_enabled.update({self.name: True})
-            self.cfg.write_cache(
-                key="services-once-enabled", content=services_once_enabled
-            )
 
+            services_once_enabled_file.write(
+                ServicesOnceEnabledData(fips_updates=True)
+            )
             return True
 
         return False
--- a/uaclient/files/data_types.py
+++ b/uaclient/files/data_types.py
@@ -1,0 +1,69 @@
+import json
+from enum import Enum
+from typing import Callable, Dict, Generic, Optional, Type, TypeVar
+
+import yaml
+
+from uaclient import exceptions
+from uaclient.data_types import DataObject
+from uaclient.files.files import UAFile
+
+
+class DataObjectFileFormat(Enum):
+    JSON = "json"
+    YAML = "yaml"
+
+
+DOFType = TypeVar("DOFType", bound=DataObject)
+
+
+class DataObjectFile(Generic[DOFType]):
+    def __init__(
+        self,
+        data_object_cls: Type[DOFType],
+        ua_file: UAFile,
+        file_format: DataObjectFileFormat = DataObjectFileFormat.JSON,
+        preprocess_data: Optional[Callable[[Dict], Dict]] = None,
+    ):
+        self.data_object_cls = data_object_cls
+        self.ua_file = ua_file
+        self.file_format = file_format
+        self.preprocess_data = preprocess_data
+
+    def read(self) -> Optional[DOFType]:
+        raw_data = self.ua_file.read()
+        if raw_data is None:
+            return None
+
+        parsed_data = None
+        if self.file_format == DataObjectFileFormat.JSON:
+            try:
+                parsed_data = json.loads(raw_data)
+            except json.JSONDecodeError:
+                raise exceptions.InvalidFileFormatError(
+                    self.ua_file.path, "json"
+                )
+        elif self.file_format == DataObjectFileFormat.YAML:
+            try:
+                parsed_data = yaml.safe_load(raw_data)
+            except yaml.parser.ParserError:
+                raise exceptions.InvalidFileFormatError(
+                    self.ua_file.path, "yaml"
+                )
+
+        if parsed_data is None:
+            return None
+
+        if self.preprocess_data:
+            parsed_data = self.preprocess_data(parsed_data)
+
+        return self.data_object_cls.from_dict(parsed_data)
+
+    def write(self, content: DOFType):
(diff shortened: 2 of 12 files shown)


uaclient/files/state_files.py
```diff
@@ -0,0 +1,50 @@
...
+from uaclient.exceptions import InvalidStateFile
+from uaclient.files.data_types import DataObjectFile
+from uaclient.files.files import UAFile
+
+SERVICES_ONCE_ENABLED = "services-once-enabled"
+
+
+class ServicesOnceEnabledData(DataObject):
+    fields = [
+        Field("services_once_enabled", data_list(StringDataValue), False)
```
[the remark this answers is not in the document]
orndorffgrant: I also suggest making this `required`

uaclient/files/state_files.py
```diff
@@ -0,0 +1,50 @@
...
+        self.write(
+            ServicesOnceEnabledData(
+                services_once_enabled=list(
+                    set(
+                        file_content.get("services_once_enabled", [])
+                        + [service]
+                    )
+                )
+            )
+        )
```
orndorffgrant, line 30: I'm not sure if the specialized class/method is necessary.

Without the specialized class/method, use would look like this (using the field name "service_names" and assuming it is required as suggested above):
```python
services_once_enabled_file = DataObjectFile(
    data_object_cls=ServicesOnceEnabledData,
    ua_file=UAFile(
        name=SERVICES_ONCE_ENABLED,
        private=False,
    ),
)
# all of this is checked by mypy
services_once_enabled_do = services_once_enabled_file.read() or ServicesOnceEnabledData(service_names=[])
services_once_enabled_do.service_names = list(set(services_once_enabled_do.service_names + ["fips-updates"]))
services_once_enabled_file.write(services_once_enabled_do)
```

With the specialized class/method, it needs to be adjusted to specify the generic type of DataObjectFile so that mypy understands what is going on and can help us. Something like this:
```suggestion
class ServiceOnceEnableDataObjectFile(DataObjectFile[ServicesOnceEnabledData]):
    def add_service(self, service: str):
        file_content = self.read() or ServicesOnceEnabledData(service_names=[])
        self.write(
            ServicesOnceEnabledData(
                service_names=list(
                    set(
                        file_content.service_names
                        + [service]
                    )
                )
            )
        )
```

uaclient/files/state_files.py
```diff
@@ -0,0 +1,50 @@
...
+
+def _get_status_file(file_name: str) -> DataObjectFile:
+    if file_name == SERVICES_ONCE_ENABLED:
+        return _services_once_enabled_file
+
+    raise InvalidStateFile(file_name)
+
+
+def get_services_once_enabled_file() -> DataObjectFile:
+    return _get_status_file(file_name=SERVICES_ONCE_ENABLED)
```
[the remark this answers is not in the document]
orndorffgrant: Going further (maybe too far?) do we even need a getter function for the file object? or should we just access it directly from the module? What benefits do we get from having a getter function here?

orndorffgrant (commented):
I'm excited about this refactor!

This changes the format of the file. I think this makes sense, but we need to include some type of migration from the old format to the new format somwhere. This could happen on first read maybe? or in postinst.

Or we can just keep the old format. That would require a new feature of DataObjects for the python field name to be different from the serialized field name - to support the field named "fips-updates", which isn't a valid python variable name.

Either way could work

uaclient/files/state_files.py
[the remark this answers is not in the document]
lucasmoura: Good point

uaclient/files/data_types.py
```diff
@@ -0,0 +1,68 @@
...
+                parsed_data = yaml.safe_load(raw_data)
+            except yaml.parser.ParserError:
+                raise exceptions.InvalidFileFormatError(
+                    self.ua_file.path, "yaml"
+                )
+
+        if parsed_data is None:
+            return None
+
+        parsed_data = self.preprocess_data(parsed_data)
```
orndorffgrant, line 58: This preprocess hook will be useful for any other migrations we may want to do in the future. +1

uaclient/files/state_files.py
```diff
@@ -0,0 +1,47 @@
...
+    def preprocess_data(self, data: Dict[str, Any]) -> Dict[str, Any]:
+        # Since we are using now returning DataObject instances from read, we
+        # cannot have variables with "-" in them. We need to explictly convert
+        # them before creating the object
+        updated_data = {}
+        for key in data.keys():
+            if "-" in key:
+                updated_data[key.replace("-", "_")] = True
+
+        return updated_data
```
orndorffgrant, line 38: If we're only supporting `fips_updates` as a field in this file then I'm not sure we need these to be so generic, but also I could see potential for needing a generic version in the future.

If we limit the functionality to what is strictly needed today we could do this:
```suggestion
    # def add_service(self, service: str):
    # not needed because we only have one service, so FipsUpdatesEntitlement can just do
    # services_once_enabled.fips_updates = True

    def preprocess_data(self, data: Dict[str, Any]) -> Dict[str, Any]:
        data["fips_updates"] = data.pop("fips-updates", None)
        return updated_data
```

But I could be convinced either way that the generic version is needed

uaclient/files/state_files.py
```diff
@@ -0,0 +1,47 @@
...
+class _ServicesOnceEnabledData(DataObject):
+    fields = [
+        Field("fips_updates", BoolDataValue, False),
+    ]
+
+    def __init__(self, fips_updates: bool):
+        self.fips_updates = fips_updates
+
+
+class _ServiceOnceEnableDataObjectFile(DataObjectFile):
```
orndorffgrant, line 19: I think this still needs to specify the TypeVar for DataObjectFile
```suggestion
class _ServiceOnceEnableDataObjectFile(DataObjectFile[_ServicesOnceEnabledData]):
```

orndorffgrant (commented):
I think this simpler migration makes sense! a couple more thoughts on typing and simplification for now

uaclient/files/state_files.py
lucasmoura: My logic here is exactly if we want to add more services in the future. I think it will not hurt if we keep it generic, but I can simplify it if needed. What do you think @renanrodrigo @CalvoM ?

renanrodrigo: I'd say the `add_service` is also calling `write`, which would be needed direcly by the caller as well - so I like having it there.
About being generic, I think we can go either way. And since any service with two or more words in the name has the dash, this is good I guess.

renanrodrigo (approved):
I think treating the files like this is a better approach, and the changeset is solid - lgtm overall.

uaclient/files/state_files.py
CalvoM: I am always for generic functions/methods.

CalvoM (commented):
This is a good refactor.

uaclient/files/data_types.py
```diff
@@ -0,0 +1,68 @@
...
+        data_object_cls: Type[DOFType],
+        ua_file: UAFile,
+        file_format: DataObjectFileFormat = DataObjectFileFormat.JSON,
+    ):
+        self.data_object_cls = data_object_cls
+        self.ua_file = ua_file
+        self.file_format = file_format
+
+    def preprocess_data(self, data: Dict[str, Any]) -> Dict[str, Any]:
+        return data
```
orndorffgrant, line 32: what do you think of making `preprocess_data` an argument instead of a method?
```suggestion
    def __init__(
        self,
        data_object_cls: Type[DOFType],
        ua_file: UAFile,
        file_format: DataObjectFileFormat = DataObjectFileFormat.JSON,
        preprocess_data: Optional[Callable[[Dict], Dict]] = None,
    ):
        self.data_object_cls = data_object_cls
        self.ua_file = ua_file
        self.file_format = file_format
        self.preprocess_data = preprocess_data
```

uaclient/entitlements/fips.py
```diff
@@ -441,30 +442,34 @@ def incompatible_services(self) -> Tuple[IncompatibleService, ...]:
...
-            self.cfg.read_cache("services-once-enabled") or {}
+            {}
+            if not services_once_enabled_obj
+            else services_once_enabled_obj.to_json()
         )
-        fips_updates_once_enabled = services_once_enabled.get(
-            fips_update.name, False
+
+        fips_updates_once_enabled = (
+            fips_updates.name.replace("-", "_") in services_once_enabled
```
orndorffgrant, line 459: I think this can all be replaced with
```python
services_once_enabled_obj = services_once_enabled_file.read()
fips_updates_once_enabled = services_once_enabled_obj.fips_updates if services_once_enabled else False
```

uaclient/files/data_types.py
lucasmoura: Good point, I will make the change

uaclient/files/state_files.py
```diff
@@ -0,0 +1,49 @@
...
+    DataObjectFile[_ServicesOnceEnabledData]
+):
+    def add_service(self, service: str):
+        file_content = self.read()
+        services_dict = {} if not file_content else file_content.to_dict()
+        service = service.replace("-", "_")
+
+        if service not in services_dict:
+            services_dict[service] = True
+            self.write(_ServicesOnceEnabledData(**services_dict))
```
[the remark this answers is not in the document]
lucasmoura: Sorry, I think I am missing the point here. Why should we only support `fips-updates` here ? Even though it is the only service using that logic, we should still make it generic if we need it for other services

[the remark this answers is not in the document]
orndorffgrant: The function as currently written only supports fips-updates and will already raise an exception if any other service name is passed. That is because `_ServicesOnceEnabledData` only has a field for `fips_updates`, so `_ServicesOnceEnabledData(**services_dict)` will raise an exception if any other service is passed.

We could change it to `_ServicesOnceEnabledData.from_dict(services_dict)`, but then it still wouldn't work for other services. it would just silently swallow the other services and not work, which we might not catch in the future.

So my suggestion is that we do one of the following:
* Option A: support all services by adding the necessary fields to `_ServicesOnceEnabledData`
* Option B: make it more explicit that this only supports fips-updates currently (that is what my previous comment in this thread is trying to do)
* Option C: don't add the `add_service` method, and instead directly use write and read wherever needed

(Option C is what I was advocating for in a different thread but really any of them are okay IMO)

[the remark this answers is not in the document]
lucasmoura: Ohhh great catch. Fair enough, thinking better about this method I think you are right regarding option C. If in the future we want to support more services we can add back that method

orndorffgrant (approved):
Thanks! This is great!

renanrodrigo (approved):
lgtm

renanrodrigo:
test fail is unrelated

PR #2260, opened 2022-09-26
several small refactors in prep for retry-auto-attach

orndorffgrant:
These are all the small refactor bits that I hit along the road while writing #2250 
They are each broken into separate commits for easier review-ability


uaclient/messages.py
```diff
@@ -1029,3 +1021,8 @@ class TxtColor:
 with '{{service}}' enabled""".format(
     bold=TxtColor.BOLD, end_bold=TxtColor.ENDC
 )
+
+ENTITLEMENT_NOT_FOUND = FormattedNamedMessage(
```
lucasmoura, line 1025: I am wondering if we should use this exception on the `action_enable` function on `cli`

uaclient/files.py
```diff
@@ -100,21 +100,41 @@ def read(self) -> Optional[DOFType]:
             try:
                 parsed_data = json.loads(raw_data)
             except json.JSONDecodeError:
-                raise exceptions.InvalidFileFormatError(
-                    self.ua_file.path, "json"
+                LOG.warning(
```
lucasmoura, line 103: Just to double check, this is to allow potential `cloud-init` invalid configs to not break the code right ? Because I don't think we gain much by not failing early here

uaclient/messages.py
orndorffgrant: `action_enable` currently has a more descriptive error message for a human at the terminal in the case where it is passed an unknown entitlement. Maybe we should revisit that message and consolidate, but I'd like to consider it out of scope for this PR.

uaclient/files.py
orndorffgrant: This change was made because of the comment here https://github.com/canonical/ubuntu-advantage-client/pull/2219#discussion_r971244096

If we want these error messages then I can put them back in, and I will re-add the `try_read` for scenarios where we don't need to handle them. I don't mind either way :upside_down_face:

lucasmoura: I totally forgot about that hahah
But maybe we could have something in the middle. What if we add an parameter to read to not raise Exceptions on errors ? In that way, we won't modify the existing use cases where failing early would be preferred

lucasmoura: Okay, another option we could consider is to catch those exception directly on the `cloud-init` endpoint that needs it. We then not need to surface the `try_read` method or modify the existing `read` method. But I don't have a strong opinion here if you think bringing back the `try`read` method is better

orndorffgrant: I think this will be used in more places than cloud-init.

Any future state file that uses DataObjectFile will need to explicitly handle these errors. I think many of those cases (but maybe not all) will want to just ignore the errors and treat them as `None`.

For example, currently #2255 should probably handle these errors. If we add either `try_read` or `read(return_none_on_errors=True)` then #2255 doesn't have to explicitly handle the errors; it can just accept that any invalid format will be treated as `None`.

I can try implementing it as `read(return_none_on_errors=True)` to see what it looks like

orndorffgrant: We discussed out of band and decided to remove this commit. If a particular use case wants to ignore these errors, then it can `except` them

lucasmoura (approved):

PR #2261, opened 2022-09-27
Fix esm across release upgrade

renanrodrigo:
From the associated LP bug:
```
With u-a-t 27.11, currently on -proposed, upgrading from Xenial to Bionic fails when unattached.

The reason is: Bionic has some updates available in esm-apps, and the repositories are present due to the postinst script seeing that the service is not beta anymore.

However, the preferences file only pins esm for Xenial, and do-release-upgrade:
1. Sees the packages in the Bionic ESM repository
2. Ignores the pinning to 'never'
3. Fails to get those packages - all of them are 401ed
4. Fail to calculate the diffs and exits 1.

Although found and reproducible when upgrading from Xenial to Bionic, this bug may affect any upgrade across releases where the system is unattached, and has any given package installed which is present in the ESM repository of the next release.
```

This changeset removes the release information from the preferences file, pinning the repositories to "never" based only on having ESM as an origin.

The immediate side effect is another bug, where the preferences file is not removed when enabling esm-infra on any LTS which is not EOSS yet. As esm-infra can be enabled in any release (even if it has only the dummy package) this does not make sense. This changeset tackles that by returning the "never" pin value for any LTS, making sure the file is removed when the service is enabled (although it is only created again in EOSS releases, which makes sense for advertisement purposes).

LP: #1990907

## Test Steps
Run `features/ubuntu_upgrade_unattached.feature` for Xenial using version `27.11`, and see it fails.
Run the same including the contents of this changeset, and see it succeeds.

## Desired commit type

## Checklist
 - [x] I have slightly updated or added any documentation accordingly

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -217,7 +217,7 @@ EOF
         cat > "${apt_pref_file}" <<EOF
 # Written by ubuntu-advantage-tools
 Package: *
-Pin: release o=${apt_origin}, n=${release}
+Pin: release o=${apt_origin}
 Pin-Priority: never
 EOF
     fi
--- a/uaclient/entitlements/esm.py
+++ b/uaclient/entitlements/esm.py
@@ -95,10 +95,10 @@ class ESMInfraEntitlement(ESMBaseEntitlement):
 
     @property
     def repo_pin_priority(self) -> Optional[str]:
-        """Once a release goes into EOL it is entitled to ESM Infra."""
-        if system.is_active_esm(system.get_platform_info()["series"]):
+        """All LTS are entitled to ESM Infra."""
+        if system.is_lts(system.get_platform_info()["series"]):
             return "never"
-        return None  # No pinning on non-ESM releases
+        return None  # No pinning on non-LTS releases
 
     @property
     def disable_apt_auth_only(self) -> bool:
--- a/uaclient/tests/test_apt.py
+++ b/uaclient/tests/test_apt.py
@@ -69,7 +69,7 @@ class TestAddPPAPinning:
         expected_pref = dedent(
             """\
             Package: *
-            Pin: release o=MYORIG, n=xenial
+            Pin: release o=MYORIG
             Pin-Priority: 1003\n"""
         )
         assert expected_pref == system.load_file(pref_file)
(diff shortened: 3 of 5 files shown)


lucasmoura (approved):

debian/ubuntu-advantage-tools.postinst
```diff
@@ -217,7 +217,7 @@ EOF
         cat > "${apt_pref_file}" <<EOF
 # Written by ubuntu-advantage-tools
 Package: *
-Pin: release o=${apt_origin}, n=${release}
+Pin: release o=${apt_origin}
```
blackboxsw, line 220: Ack. Undertstood that we want this config to be release-agnostic, because do-release-upgrade when calculating package diffs to upgrade will incorrectly preserve this preferences.d file without updating the n=${release} to the target upgraded release name which then leaves this upgrade path exposed to trying to install any new pkg from Bionic.

Dropping the `n=${release}` filter will allow the never pin to be honored across release upgrade paths

ESM_INFRA_APT_PREF_FILE is a release-agnostic filename `/etc/apt/preferences.d/ubuntu-esm-apps` so dropping the release-specific filter makes sense to me. Our postinst will also double check upon bionic package install if no configuration APT source yet exists for esm-apps

uaclient/entitlements/esm.py
```diff
@@ -95,10 +95,10 @@ class ESMInfraEntitlement(ESMBaseEntitlement):
 
     @property
     def repo_pin_priority(self) -> Optional[str]:
-        """Once a release goes into EOL it is entitled to ESM Infra."""
-        if system.is_active_esm(system.get_platform_info()["series"]):
+        """All LTS are entitled to ESM Infra."""
+        if system.is_lts(system.get_platform_info()["series"]):
```
blackboxsw, line 99: Yes an I think this represents a product change in behavior that happened a couple years ago that wasn't accurately reflected in the APT preferences setup for esm-infra

uaclient/tests/test_apt.py
```diff
@@ -69,7 +69,7 @@ def test_write_apt_pin_file_to_apt_preferences(self, m_platform, tmpdir):
         expected_pref = dedent(
             """\
             Package: *
-            Pin: release o=MYORIG, n=xenial
```
blackboxsw, line 72: I bet we can drop the [mocked get_platform_info](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/tests/test_apt.py#L58) this related test now.

blackboxsw (commented):
+! thanks for the discussion and behavioral cleanup here.

Do we need a followup PR to ensure [ua-t.postinst:configure_esm](https://github.com/canonical/ubuntu-advantage-client/blob/main/debian/ubuntu-advantage-tools.postinst#L228) also configures APT repo/keys for esm-infra if check_is_lts instead of check_is_active_esm?

PR #2269, opened 2022-09-29
Remove apps information from security-status

renanrodrigo:
This will only be available when it is GA, and it is according to what was discussed with Lech.

Aside from the message change, it is the same @orndorffgrant had sent before. Thanks much for that.

## Test Steps
Integration tests have a good coverage of the scenarios

## Desired commit type

## Checklist


lucasmoura (approved):
Let's just wait for CI here

PR #2277, opened 2022-10-04
test: Extend auto-attach no-op scenario (SC-1270) 

aciba90:
## Proposed Commit Message

```
test: Extend auto-attach no-op scenario

The test has been divided into two stages / boots. The first
one to prepare the instance environment and the second one
to execute the test.

In the first boot:

- Install cloud-init from ppa:cloud-init-dev/daily to ensure
  we are getting latest feature behavior.
- Remove explicit user-data, as features/enviroment.py already
  prepares user-data to stop auto-attach, and instead, assert
  that the instance is not auto attached at the end of the first
  boot.
- Create snapshot and boot with a valid user-data to auto attach
  with cloud-init.

Second boot:

- Assert UA nooped
- Assert cloud-init status shows success
- Assert the instance is auto attached
```

## Test Steps

Execute the modified test scenario.

## Desired commit type

## Checklist

### DIFF
--- a/features/steps/machines.py
+++ b/features/steps/machines.py
@@ -32,7 +32,9 @@ def add_test_name_suffix(context, series, prefix):
 
 
 @given("a `{series}` machine with ubuntu-advantage-tools installed")
-def given_a_machine(context, series, custom_user_data=None):
+def given_a_machine(
+    context, series, custom_user_data=None, cloud_init_ppa=None
+):
     if series in context.reuse_container:
         context.instances = {}
         context.container_name = context.reuse_container[series]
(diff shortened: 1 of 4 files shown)


features/steps/machines.py
```diff
@@ -154,20 +180,38 @@ def launch_machine(context, series, instance_name, ports=None):
...
+        instance_name
+    ] = context.config.cloud_manager.launch(
         context.series,
         instance_name=name,
         image_name=context.instance_snapshot,
+        user_data=user_data,
+    )
+    instance_id = context.config.cloud_manager.get_instance_id(inst)
+    logging.info(
+        "--- Instance launched rom snapshot: instance_id=%s, instance_name=%s",
```
lucasmoura, line 198: s/rom/from

features/ubuntu_pro.feature
```diff
@@ -613,18 +613,11 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
...
-        When I run `cloud-init query userdata` with sudo
-        Then stdout matches regexp:
-        """
-        ubuntu_advantage:
-        """
+    Scenario Outline: Auto-attach no-op when cloud-init has ubuntu_advantage on userdata and cloud-init auto attaches
+        # In the first boot we install uat and daily cloud-init and prepare/check preconditions
+        # In the second boot we execute the test
+
+        Given a `<release>` machine with ubuntu-advantage-tools installed and cloud-init upgraded to the latest daily version
```
lucasmoura, line 620: I understand that right now we need to use the daily version of cloud-init to test that, but once it is released, I think we should keep using the one in the archive instead, otherwise this test could be flaky, since we will be keep using the devel version of cloud-init.

I think we can add this test as is, but we need to update it after the release of cloud-init that possess that ua attach feature

aciba90:
Per the discussion within this PR and daily standup, we came to the conclusion that it's better to not install daily cloud-init and wait for the next cloud-init release, presumably 22.3.4 (https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1987318), which includes the custom auto attach functionality. We will extend this test without needing to install the aforementioned daily cloud-init version, avoiding possible test noise.

PR #2281, opened 2022-10-04
pro docs

orndorffgrant:


lucasmoura (approved):

lucasmoura (approved):

renanrodrigo (approved):
Looks good as is, throwing some suggestions

- Should the `ua-auto-attach.service` "Description" field (line 2 in the file) say Ubuntu Pro auto attach? Same for other services;
- There is a reference to `ubuntu.com/advantage` in the `ubuntu-advantage.service` file, line 11
- I see references in code/docstrings but agree this is out of scope

PR #2285, opened 2022-10-06
collect-logs: Do not fail if we can't read a file

lucasmoura:
## Proposed Commit Message
collect-logs: Do not fail if we can't read a file

Currently, if we fail to read a file, we will error out during the execution of the command. We are updating that so that we will just warn the user that a file could not be read and proceed with the command.

LP: #1991858

## Test Steps
Run the new unittest

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -153,6 +153,23 @@ def _write_command_output_to_file(
         system.write_file(filename, out)
 
 
+def _get_state_files(cfg: config.UAConfig):
+    # include cfg log files here because they could be set to non default
+    return [
+        cfg.cfg_path or DEFAULT_CONFIG_FILE,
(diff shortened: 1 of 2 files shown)


uaclient/actions.py
```diff
@@ -193,27 +209,19 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
         )
 
     # include cfg log files here because they could be set to non default
```
renanrodrigo, line 211: this comment would move up to inside of the function imho

renanrodrigo (approved):

orndorffgrant (approved):
I agree with renan's comment, looks good otherwise

lucasmoura:
@renanrodrigo @orndorffgrant done

PR #2292, opened 2022-10-11
notices: refactor to have separate files

CalvoM:
notices: refactor to have separate files

This is part of refactoring the notices structure we have. 
We are separating the notices to different files instead of all notices being in one file.

LP: #1987738

## Desired commit type

## Checklist


uaclient/files.py
```diff
@@ -350,6 +352,147 @@ def contract_id(self):
         return None
 
 
+class NoticesLabelID(NoticeFileDetails, Enum):
```
orndorffgrant, line 355: I like the idea of the notices as values of an enum. I think we might as well go all the way and include the notice message content in the tuple as well.

And if we do that we can rename the Enum to just `Notices`

uaclient/files.py
```diff
@@ -350,6 +352,147 @@ def contract_id(self):
...
+    )
+    DAEMON_AUTO_ATTACH_FAILED = NoticeFileDetails(
+        label="daemon_auto_attach_failed", order_id="51"
+    )
+    OPERATION_IN_PROGRESS = NoticeFileDetails(
+        label="operation_in_progress", order_id="60"
+    )
+
+
+class Notices:
```
orndorffgrant, line 401: I think instead of a class, this can just be a module. I think the use I'd like to see is something like:

```python
from uaclient.files import notices

notices.add(Notices.FIPS_SYSTEM_REBOOT_REQUIRED)
notices.remove(Notices.FIPS_SYSTEM_REBOOT_REQUIRED)
notices.list()  # returns list of all Notices tuples on disk currently
for notice in notices.list():
    print(notice.message)
```

uaclient/files.py
```diff
@@ -350,6 +352,147 @@ def contract_id(self):
...
+    def add(self, notice_label_id: NoticeFileDetails, description: str):
+        label = notice_label_id.label
+        file_order_id = notice_label_id.order_id
+        if self.is_root:
+            notice_files = self.get_all_notice_files()
+            labels_present = [file.split("-")[-1] for file in notice_files]
+            if label in labels_present:
+                return
+            filename = "{}-{}".format(file_order_id, label)
+            file = UAFile(filename, self.directory)
```
orndorffgrant, line 428: I don't think the `UAFile` abstraction is helping much for this particular feature (as evidenced below where in `remove` we just directly call `system.remove_file`). I think we can just use `system.write_file` directly in the `notices` module.

uaclient/files.py
```diff
@@ -350,6 +352,147 @@ def contract_id(self):
...
+            file.write(description)
+        else:
+            raise exceptions.NonRootUserError
+
+    def try_add(self, notice_label_id: NoticeFileDetails, description: str):
```
orndorffgrant, line 433: I think the normal `notices.add` should have the behavior of what is currently `try_add`. I can't think of a reason we'd want to raise an error when we use the wrong method - it should just log and do the right thing (not create the file if nonroot)

orndorffgrant (commented):
I know it's still a draft but just did a high level review of the structures with some suggestions

uaclient/files.py
CalvoM: Some message content uses formatting tmpl which makes it  impossible to edit the enum since it is readonly.

uaclient/files/notices.py
```diff
@@ -0,0 +1,153 @@
...
+    def __init__(
+        self,
+        is_permanent: bool = True,
+        root_mode: bool = True,
+        override_dir: Optional[str] = None,
+    ):
+        if not override_dir:
+            self.directory = "/var/lib/ubuntu-advantage/notices/"
+            if not is_permanent:
+                self.directory = "/var/run/notices/"
```
lucasmoura, line 69: Let's move those paths to the defaults module

lucasmoura: Also, instead of having to pass the `is_permanent` variable to dictate that the notice should not be on `/var/run`, we could instead add a parameter `NoticeFileDetails` for that, since we know that all reboot notice should have the same behavior, we can control that directly through the `NoticeFileDetails` instance

uaclient/files/notices.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def try_remove(self, notice_label_id: NoticeFileDetails):
+        try:
+            self.remove(notice_label_id)
+        except exceptions.NonRootUserError:
+            event.warning("Trying to add notice as non-root user")
+
+    def get_all_notice_files(self):
+        if not os.path.exists(self.directory):
+            os.makedirs(self.directory)
```
lucasmoura, line 118: Shouldn't we just return an empty list here instead of creating the folder ? I know this is just to make the logic in the following lines work, but I don't think the `get` method should be responsible for creating the folder

uaclient/files/notices.py
```diff
@@ -0,0 +1,153 @@
...
+
+        notices = [
+            file
+            for file in os.listdir(self.directory)
+            if os.path.isfile(os.path.join(self.directory, file))
+        ]
+        notices.sort()
```
lucasmoura, line 125: I think we should sort based on the priority of the notice here right ?  I know this will require us to match the file title with existing notices, but maybe it is worth it.

uaclient/files/notices.py
```diff
@@ -0,0 +1,153 @@
...
+        return notices
+
+    def read(self):
+        if not self.override:
+            notice_directories = (
+                "/var/lib/ubuntu-advantage/notices/",
+                "/var/run/notices/",
```
lucasmoura, line 132: Let's move this into defaults module

uaclient/files/notices.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            notice_directories = (self.directory,)
+        notices = []
+        for notice_directory in notice_directories:
+            notices = [
+                file
+                for file in os.listdir(notice_directory)
+                if os.path.isfile(os.path.join(notice_directory, file))
+            ]
+            notices.sort()
```
lucasmoura, line 143: Can we use `get_all_notice_files` somehow here ? We would need to pass the `notice_directory` as a parameter, but maybe it is worth it

uaclient/files/notices.py
CalvoM: You are right that the `get` method should not have to create anything. Should we create the folder where we call this method after this returns an empty list?

uaclient/files/notices.py
CalvoM: Ah, it could work.

uaclient/files/notices.py
lucasmoura: Yes, I think we should only create the folder on the `add` method.

uaclient/files/notices.py
```diff
@@ -0,0 +1,206 @@
...
+
+class Singleton(type):
+    _instances = {}  # type: Dict[Any, Any]
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super(Singleton, cls).__call__(
+                *args, **kwargs
+            )
+        return cls._instances[cls]
```
orndorffgrant, line 23: While very cool, I think the metaclass is a bit overkill here to enforce the singleton.
We can just only create one.
```python
_notices = None
def get_notices():
    if _notices is None:
        _notices = Notices()
    return _notices


def add():
    notices = get_notices()
    notices.add()
```

uaclient/files/notices.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class NoticesDetails(NoticeFileDetails, Enum):
+    REBOOT_REQUIRED = NoticeFileDetails(
+        label="reboot_required", order_id="10", is_permanent=False
+    )
```
orndorffgrant, line 29: Can we re-investigate including the notice message itself here? I think `.format` doesn't modify the value, it just returns a new string that is formatted. So I think it'd be okay to include in a Tuple,

Then we could format notices like:
```python
def add(notice, **kwargs):
    msg = notice.value.msg.format(**kwargs)

notices.add(NoticeDetails.REBOOT_REQUIRED, format_arg=format_val)
```

uaclient/files/notices.py
```diff
@@ -0,0 +1,206 @@
...
+        self,
+        root_mode: bool = True,
+        override_dir: Optional[str] = None,
+    ):
+        self.directory = override_dir
+        if not override_dir:
+            self.override = False
+        else:
+            self.override = True
+        self.is_root = root_mode
```
orndorffgrant, line 87: Ideally I'd like to find a way to test this without needing an override_dir. In theory we can just mock all read/write calls so the dir shouldn't matter and shouldn't need to be overriden.

uaclient/files/notices.py
```diff
@@ -0,0 +1,206 @@
...
+    )
+    OPERATION_IN_PROGRESS = NoticeFileDetails(
+        label="operation_in_progress", order_id="60", is_permanent=True
+    )
+
+
+class Notices(metaclass=Singleton):
+    def __init__(
+        self,
+        root_mode: bool = True,
```
orndorffgrant, line 79: This is going against the conclusion of a past conversation, but I think we should discuss the `root_mode` pattern again.
I think it will be easier to just check for root close to the place where it matters (e.g. in the add and remove functions here)

uaclient/files/notices.py
```diff
@@ -0,0 +1,206 @@
...
+                label_file = [
+                    file for file in notice_files if file == filename
+                ]
+                if len(label_file) == 1:
+                    system.remove_file(
+                        os.path.join(self.directory, label_file[0])
+                    )
+        else:
+            event.warning("Trying to remove notice as non-root user")
+            raise exceptions.NonRootUserError
```
orndorffgrant, line 151: I'd like to reconsider this as well. We can probably just not do anything here rather than raise an error. Another thing we can discuss

uaclient/files/notices.py
```diff
@@ -0,0 +1,206 @@
...
+
+        notices = [
+            file
+            for file in os.listdir(notice_dir)
+            if os.path.isfile(os.path.join(notice_dir, file))  # type: ignore
+        ]
+        notices.sort()
+        return notices
+
+    def read(self):
```
orndorffgrant, line 168: I'd like this to have a typed return value. And I'd like the return type to include the id and ordering number of each notice in addition to the notice message contents

orndorffgrant (commented):
some quick drive by comments :upside_down_face:

uaclient/files/notices.py
renanrodrigo: +1 I think messages should definitely be in the Enum, and the `add` method can be responsible for the formatting. For messages where the field depends on the caller, Grant's suggestion should work fine.

uaclient/files/notices.py
orndorffgrant: I could be convinced either way :upside_down_face: maybe we should convert other places to use the Singleton metaclass

uaclient/files/notices.py
CalvoM: The file names are lexicographically made, the restart notices are highest in the priority. We can discuss the other notices priority.

uaclient/files/notices.py
lucasmoura: I would rather use the simpler approach, but I don't have a strong opinion here

lib/reboot_cmds.py
```diff
@@ -82,13 +84,16 @@ def fix_pro_pkg_holds(cfg):
                         )
                     )
                     sys.exit(1)
-                cfg.notice_file.remove(
-                    "", messages.FIPS_SYSTEM_REBOOT_REQUIRED.msg
+                notices.remove(
```
lucasmoura, line 87: Those notices should be on the `/var/run/` folder. Should we still manually remove them ?

uaclient/defaults.py
```diff
@@ -58,3 +58,5 @@
 
 ROOT_READABLE_MODE = 0o600
 WORLD_READABLE_MODE = 0o644
+NOTICES_PERMANENT_DIRECTORY = DEFAULT_DATA_DIR + "/notices/"
+NOTICES_TEMPORARY_DIRECTORY = "/var/run/notices/"
```
lucasmoura, line 62: Let's add `pro` prefix here to the folder, just to make it clearer to users that this is related to our tool

uaclient/entitlements/fips.py
```diff
@@ -535,8 +547,16 @@ def messaging(self) -> MessagingOperationsDict:
 
     def _perform_enable(self, silent: bool = False) -> bool:
         if super()._perform_enable(silent=silent):
-            self.cfg.notice_file.try_remove(
-                "", messages.FIPS_DISABLE_REBOOT_REQUIRED
+            notices.remove(
```
lucasmoura, line 550: Same remove consideration

lucasmoura (commented):
@CalvoM I know that you are still going to test the fips cases to avoid removing the notices manually. Therefore, my comments are just a remainder to remove those cases after you conclude the tests

uaclient/entitlements/fips.py
CalvoM: Makes sense.

uaclient/entitlements/fips.py
```diff
@@ -278,24 +283,28 @@ def application_status(
...
+                notices.remove(
+                    self.cfg.root_mode,
+                    NoticesDetails.FIPS_SYSTEM_REBOOT_REQUIRED.value,
                 )
 
-            self.cfg.notice_file.try_remove(
-                "", messages.FIPS_REBOOT_REQUIRED_MSG
+            notices.remove(
+                self.cfg.root_mode, NoticesDetails.FIPS_REBOOT_REQUIRED.value
             )
```
orndorffgrant, line 293: I think we can remove all of this

uaclient/entitlements/fips.py
```diff
@@ -304,8 +313,9 @@ def application_status(
                     ),
                 )
         else:
-            self.cfg.notice_file.try_remove(
-                "", messages.FIPS_DISABLE_REBOOT_REQUIRED
+            notices.remove(
+                self.cfg.root_mode,
+                NoticesDetails.FIPS_DISABLE_REBOOT_REQUIRED.value,
```
orndorffgrant, line 318: I think we can remove this one too

uaclient/entitlements/fips.py
```diff
@@ -342,11 +352,12 @@ def remove_packages(self) -> None:
...
-            self.cfg.notice_file.try_remove(
-                "", messages.NOTICE_WRONG_FIPS_METAPACKAGE_ON_CLOUD
+            notices.remove(
+                self.cfg.root_mode,
+                NoticesDetails.WRONG_FIPS_METAPACKAGE_ON_CLOUD.value,
             )
-            self.cfg.notice_file.try_remove(
-                "", messages.FIPS_REBOOT_REQUIRED_MSG
+            notices.remove(
+                self.cfg.root_mode, NoticesDetails.FIPS_REBOOT_REQUIRED.value
```
orndorffgrant, line 360: I'm confused by this one - @lucasmoura do you remember why this notice gets removed on enable?

uaclient/entitlements/fips.py
orndorffgrant: I think this comment still applies and this remove call can be removed @CalvoM

uaclient/entitlements/fips.py
```diff
@@ -535,8 +547,16 @@ def messaging(self) -> MessagingOperationsDict:
...
+            notices.remove(
+                self.cfg.root_mode,
+                NoticesDetails.FIPS_DISABLE_REBOOT_REQUIRED.value,
+            )
+            services_once_enabled = (
+                self.cfg.read_cache("services-once-enabled") or {}
+            )
+            services_once_enabled.update({self.name: True})
+            self.cfg.write_cache(
+                key="services-once-enabled", content=services_once_enabled
```
orndorffgrant, line 559: I think this must've accidentally slipped in during a rebase - it can be removed

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+from uaclient import defaults, event_logger, messages, system
+from uaclient.files import UAFile
+
+event = event_logger.get_event_logger()
+NoticeFileDetails = namedtuple(
+    "NoticeFileDetails", ["order_id", "label", "is_permanent", "message"]
+)
+
+
+class NoticesDetails(NoticeFileDetails, Enum):
```
orndorffgrant, line 15: minor bikeshed suggestion. I think the code reads slightly better with this rename
```suggestion
class Notice(NoticeFileDetails, Enum):
```

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+        is_permanent=True,
+        message=messages.AUTO_ATTACH_RETRY_TOTAL_FAILURE_NOTICE,
+    )
+
+
+class Notices:
+    def __init__(
+        self,
+    ):
+        pass
```
orndorffgrant, line 112: if this is a `pass` do we need it?
```suggestion
```

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+    ):
+        """Adds a notice file.
+
+        :param notice_details: Holds details concerning the notice file.
+        :param description: The content to be written to the notice file.
+        """
+        if root_mode:
+            label = notice_details.label
+            file_order_id = notice_details.order_id
+            is_notice_permanent = notice_details.is_permanent
```
orndorffgrant, line 128: I don't think we need to assign these to local variables, just use them as "notice_details.label" later on
```suggestion
```

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+            self.directory = (
```
orndorffgrant, line 129: This doesn't need to be an instance variable
```suggestion
            directory = (
```

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+                defaults.NOTICES_PERMANENT_DIRECTORY
+                if is_notice_permanent
+                else defaults.NOTICES_TEMPORARY_DIRECTORY
+            )
+            notice_files = self.get_all_notice_files()
+            if notice_files is None:
+                os.makedirs(self.directory)  # type: ignore
+                notice_files = []
```
orndorffgrant, line 137: instead of using `get_all_notice_files`, I suggest we test directly here
```suggestion
            if not os.path.exists(directory):
                os.makedirs(directory)  # type: ignore
```

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+            labels_present = [file.split("-")[-1] for file in notice_files]
+            if label in labels_present:
+                return
```
orndorffgrant, line 140: I think we can just overwrite a notice if `add` is called for the same notice twice
```suggestion
```
thoughts @lucasmoura @renanrodrigo @CalvoM ?

uaclient/files/notices.py
orndorffgrant: Right, because we name the files with the priority number first, we can just sort them by file name :+1:

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+
+        notices = [
+            file
+            for file in os.listdir(notice_dir)
+            if os.path.isfile(os.path.join(notice_dir, file))  # type: ignore
+        ]
+        notices.sort()
+        return notices
+
+    def read(self) -> List[List[Optional[str]]]:
```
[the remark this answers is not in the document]
orndorffgrant: Also - in light of not showing labels for the notices, we can simplify the return type
```suggestion
    def list(self) -> List[str]:
```

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+
+def get_notice():
+    global _notice_cls
+    if _notice_cls is None:
+        _notice_cls = Notices()
+
+    return _notice_cls
+
+
+def add(root_mode: bool, notice_details: NoticeFileDetails, **kwargs):
```
orndorffgrant, line 231: I think if the arg here is the enum class `NoticesDetails` (or `Notice` if you accept my rename suggestion) then we don't need to do `.value` everywhere this is called

same for `remove`

uaclient/files/notices.py
```diff
@@ -0,0 +1,244 @@
...
+    notice = get_notice()
+    description = notice_details.message.format(**kwargs)
+    notice.add(root_mode, notice_details, description)
```
orndorffgrant, line 234: I think the `format` call should probably happen in the `Notices.add` function instead of here

uaclient/status.py
```diff
@@ -349,11 +356,8 @@ def status(cfg: UAConfig, show_all: bool = False) -> Dict[str, Any]:
 
         # Try to remove fix reboot notices if not applicable
         if not system.should_reboot():
-            cfg.notice_file.remove(
-                "",
-                messages.ENABLE_REBOOT_REQUIRED_TMPL.format(
-                    operation="fix operation"
-                ),
+            notices.remove(
+                cfg.root_mode, NoticesDetails.ENABLE_REBOOT_REQUIRED.value
```
orndorffgrant, line 360: this notice is temporary now so I don't think this is needed

orndorffgrant (commented):
Looking good!

We also need a migration from the existing json file to these new files in `postinst`. This might be as simple as removing the json file if it exists and if we think its okay to lose any existing notices. Thoughts @lucasmoura @renanrodrigo @CalvoM ?

Also there is `type: ignore` in several places - we should only use that in extreme circumstances. I think it's being used here to pass Path objects instead of strings - lets just convert them to strings where needed.

uaclient/daemon/retry_auto_attach.py
```diff
@@ -125,14 +132,30 @@ def retry_auto_attach(cfg: UAConfig) -> None:
...
             with lock.SpinLock(
                 cfg=cfg,
                 lock_holder="pro.daemon.retry_auto_attach.notice_updates",
             ):
-                cfg.notice_file.remove(
-                    "", messages.AUTO_ATTACH_RETRY_NOTICE_PREFIX
+                notices.remove(
+                    cfg.root_mode,
+                    note,
+                )
```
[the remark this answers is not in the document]
orndorffgrant: with my suggestion below, the `add` will overwrite, so if we accept that, then we don't need this

uaclient/files/notices.py
orndorffgrant: FFR: we talked out of band and decided that for now we would stick with the approach we use in other places, and there will be a future task to migrate us to use the singleton metaclass everywhere

uaclient/files/notices.py
lucasmoura: +1

uaclient/files/notices.py
```diff
@@ -0,0 +1,211 @@
...
+    CONTRACT_REFRESH_WARNING = NoticeFileDetails(
+        label="contract_refresh_warning",
+        order_id="40",
+        is_permanent=True,
+        message=messages.NOTICE_REFRESH_CONTRACT_WARNING,
+    )
+    OPERATION_IN_PROGRESS = NoticeFileDetails(
+        label="operation_in_progress",
+        order_id="60",
+        is_permanent=True,
```
orndorffgrant, line 90: this can be false (just to be defensive in case of errors - there is no case where this file would be valid across a reboot)

uaclient/files/notices.py
```diff
@@ -0,0 +1,211 @@
...
+        message="Operation in progress: {operation}",
+    )
+    AUTO_ATTACH_RETRY_FULL_NOTICE = NoticeFileDetails(
+        label="auto_attach_retry_full_notice",
+        order_id="70",
+        is_permanent=True,
```
orndorffgrant, line 96: same reasoning here, this can be false

uaclient/files/notices.py
```diff
@@ -0,0 +1,211 @@
...
+
+from uaclient import defaults, event_logger, messages, system
+
+event = event_logger.get_event_logger()
+NoticeFileDetails = namedtuple(
+    "NoticeFileDetails", ["order_id", "label", "is_permanent", "message"]
+)
+
+
+class Notice(NoticeFileDetails, Enum):
```
orndorffgrant, line 14: I left comments on `is_permanent` values that i think can be changed - the rest look good IMO!

debian/ubuntu-advantage-tools.postinst
```diff
@@ -245,27 +245,26 @@ configure_esm() {
...
-    msg_name=$1
+    notice_type=$1
     /usr/bin/python3 -c "
-from uaclient.config import UAConfig
-from uaclient.messages import ${msg_name}
-cfg = UAConfig(root_mode=True)
-cfg.notice_file.add(label='', description=${msg_name})
+from uaclient.files import notices
+from uaclient.files.notices import NoticesDetails
+notices.add(True, NoticesDetails.$notice_type.value)
```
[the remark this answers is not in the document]
orndorffgrant: created https://warthogs.atlassian.net/browse/SC-1358

debian/ubuntu-advantage-tools.postinst
```diff
@@ -407,6 +406,51 @@ migrate_ubuntu_pro_beta_banner() {
...
+        filename=/run/ubuntu-advantage/notices/21-fips_system_reboot_required
+        printf "%s" "$fips_system_reboot_msg" > $filename
+    fi
+    if grep -q "$livepatch_reboot_msg" $notices_file; then
+        filename=/run/ubuntu-advantage/notices/30-lp_lts_reboot_required
+        printf "%s" "$livepatch_reboot_msg" > $filename
+    fi
+    #TODO: should we delete the file
+    rm /var/lib/ubuntu-advantage/notices.json
+    # TODO: Should we check for the notices ENABLE_REBOOT_REQUIRED
```
orndorffgrant, line 451: I think we only want to do the simplest possible migration here, since the probability that any of this is needed is already tiny.

Also we should stick with one method of creating notices in the postinst. For now that is the `add_notice` function. https://warthogs.atlassian.net/browse/SC-1358 will change it.

Also yes we should delete the old file.

So in all I think it will look something like:
```suggestion
    # This migration will happen for pro client versions 27.12 or older that still use notices.json.
    # Notices are generally short lived, so the chances of an upgrade happening while a
    # notice is in place is very small.
    # Despite that we do a simple migration here of the most important notices: reboot required notices.
    # All notices with "reboot" in them can be safely transformed into a generic reboot required message.
    # The new message won't include the exact reason for the reboot, but the recommended action is the same.
    if grep -q -i "reboot" $notices_file; then
        add_notice REBOOT_REQUIRED
    fi

    rm $notices_file
```

orndorffgrant: @renanrodrigo @lucasmoura @CalvoM please confirm that this is what we agreed to the other day

debian/ubuntu-advantage-tools.postinst
```diff
@@ -494,6 +512,7 @@ case "$1" in
       fi
       migrate_ubuntu_pro_beta_banner $PREVIOUS_PKG_VER
       cleanup_candidate_version_stamp_permissions
+      migrate_old_notices
```
orndorffgrant, line 515: @CalvoM Per the discussion the other day and Lucas' comment above, lets just guard this with a version check:
```suggestion
      if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "27.13~"; then
          migrate_old_notices
      fi
```

lucasmoura: @CalvoM this is just a nit, so feel free to disagree, but I think we can add indentation here on the if block

features/attached_status.feature
```diff
@@ -28,7 +28,7 @@ Feature: Attached status
         Then stdout matches regexp:
         """
         NOTICES
-        Operation in progress: pro enable
+        operation_in_progress: Operation in progress: pro enable
```
lucasmoura, line 31: Should we have `operation_in_progress`  twice here ?

features/api_configure_retry_service.feature
```diff
@@ -46,7 +46,7 @@ Feature: api.u.pro.attach.auto.configure_retry_service
         Then stdout matches regexp:
         """
         NOTICES
-        Failed to automatically attach to Ubuntu Pro services 1 time\(s\).
+        auto_attach_retry_full_notice: Failed to automatically attach to Ubuntu Pro services 1 time\(s\).
```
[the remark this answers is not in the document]
lucasmoura: @CalvoM I think this is still affecting this PR. Once this is sorted out, I think we can merge it

orndorffgrant (commented):
Thanks! After the January release is cut, let's go ahead and merge this

renanrodrigo:
As sad as I can be typing this but merging the stuff from the release created conflicts :|

renanrodrigo:
Took the liberty to fix the conflicts - CI happy means merge

renanrodrigo (approved):

PR #2295, opened 2022-10-17
version: use /run for candidate-version cache

orndorffgrant:
## Test Steps

make sure version upgrade notification still works

## Desired commit type

## Checklist

### DIFF
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -231,19 +231,18 @@ Feature: Command behaviour when unattached
         Given a `<release>` machine with ubuntu-advantage-tools installed
         #  Make sure we have a fresh, just rebooted, environment
         When I reboot the machine
-        Then I verify that no files exist matching `/tmp/ubuntu-advantage/candidate-version`
-        When I run `pro status` as non-root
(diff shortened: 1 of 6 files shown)


features/unattached_commands.feature
```diff
@@ -231,19 +231,18 @@ Feature: Command behaviour when unattached
         Given a `<release>` machine with ubuntu-advantage-tools installed
         #  Make sure we have a fresh, just rebooted, environment
         When I reboot the machine
-        Then I verify that no files exist matching `/tmp/ubuntu-advantage/candidate-version`
-        When I run `pro status` as non-root
+        Then I verify that no files exist matching `/run/ubuntu-advantage/candidate-version`
+        When I run `pro status` with sudo
```
lucasmoura, line 235: Why are we running just this status as sudo ?

orndorffgrant: the new file can only be written by root.
non-root calls will be able to read the cache , but they won't be able to update the cache.

This test is testing the cache file specifically, so it requires a root call to create the cache file

lucasmoura (approved):

PR #2297, opened 2022-10-18
auto-attach: use api in cli entry-point (SC-1274)

aciba90:
## Proposed Commit Message

```
auto-attach: Refactor cli entry-point to use the api

In order to simplify the code base, make use of
full_auto_attach api endpoint in cli `pro auto-attach` command.

Behavior change:

Previously, in the case where the user tried to auto attach using
the cli `pro auto-attach` on instances that were already attached
or have disable_auto_attach configured as true in the
uaclient.conf file, the cli exited with a 0 exit_code.

Now, the cli returns non 0 exit codes in both cases. The cli returned
0s because we would call the cli command on the
ua-auto-attach.service systemd unit. If we detected that the user 
was already attached or hand explicit config disabling that service,
we didn't want to show that this service failed.

After previous and this refactoring, both entry-points make use of the
new external api to auto-attach (full_auto_attach) and handle the error
cases differently.

We didn't advertise the use of `pro auto-attach` for programmatic
consumption.

Therefore we believe it is safe to change the exit codes in these
scenarios.
```

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
+++ b/uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
@@ -67,14 +67,17 @@ def full_auto_attach(options: FullAutoAttachOptions) -> FullAutoAttachResult:
 
 
 def _full_auto_attach(
-    options: FullAutoAttachOptions, cfg: UAConfig
+    options: FullAutoAttachOptions,
+    cfg: UAConfig,
+    *,
+    mode: event_logger.EventLoggerMode = event_logger.EventLoggerMode.JSON
 ) -> FullAutoAttachResult:
     try:
         with lock.SpinLock(
             cfg=cfg,
             lock_holder="pro.api.u.pro.attach.auto.full_auto_attach.v1",
         ):
-            ret = _full_auto_attach_in_lock(options, cfg)
+            ret = _full_auto_attach_in_lock(options, cfg, mode=mode)
     except Exception as e:
         lock.clear_lock_file_if_present()
         raise e
@@ -82,9 +85,11 @@ def _full_auto_attach(
 
 
 def _full_auto_attach_in_lock(
-    options: FullAutoAttachOptions, cfg: UAConfig
+    options: FullAutoAttachOptions,
+    cfg: UAConfig,
+    mode: event_logger.EventLoggerMode,
 ) -> FullAutoAttachResult:
-    event.set_event_mode(event_logger.EventLoggerMode.JSON)
+    event.set_event_mode(mode)
 
     if cfg.is_attached:
         raise exceptions.AlreadyAttachedError(
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -1390,8 +1390,6 @@ def action_auto_attach(args, *, cfg: config.UAConfig) -> int:
         return 1
     except api_exceptions.AutoAttachDisabledError:
         return 0
-    except exceptions.AlreadyAttachedError as e:
-        raise exceptions.AlreadyAttachedOnPROError() from e
     else:
         _post_cli_attach(cfg)
         return 0
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -33,6 +33,7 @@ from uaclient import (
 )
 from uaclient import status as ua_status
 from uaclient import util, version
+from uaclient.api import exceptions as api_exceptions
 from uaclient.api.api import call_api
 from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
     FullAutoAttachOptions,
@@ -1378,13 +1379,6 @@ def action_api(args, *, cfg):
 
 @assert_root
 def action_auto_attach(args, *, cfg: config.UAConfig) -> int:
-    if cfg.is_attached:
-        raise exceptions.AlreadyAttachedOnPROError()
-
-    skip_auto_attach = actions.should_disable_auto_attach(cfg)
-    if skip_auto_attach:
-        return 0
-
     try:
         _full_auto_attach(
             FullAutoAttachOptions(),
@@ -1394,6 +1388,10 @@ def action_auto_attach(args, *, cfg: config.UAConfig) -> int:
     except exceptions.UrlError:
         event.info(messages.ATTACH_FAILURE.msg)
         return 1
+    except api_exceptions.AutoAttachDisabledError:
+        return 0
+    except exceptions.AlreadyAttachedError as e:
+        raise exceptions.AlreadyAttachedOnPROError() from e
     except exceptions.UserFacingError as e:
         event.info(e.msg)
         return 1
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -33,7 +33,6 @@ from uaclient import (
 )
 from uaclient import status as ua_status
 from uaclient import util, version
-from uaclient.api import exceptions as api_exceptions
 from uaclient.api.api import call_api
 from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
     FullAutoAttachOptions,
@@ -1388,8 +1387,6 @@ def action_auto_attach(args, *, cfg: config.UAConfig) -> int:
     except exceptions.UrlError:
         event.info(messages.ATTACH_FAILURE.msg)
         return 1
-    except api_exceptions.AutoAttachDisabledError:
-        return 0
     else:
         _post_cli_attach(cfg)
         return 0
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -34,6 +34,10 @@ from uaclient import (
 from uaclient import status as ua_status
 from uaclient import util, version
 from uaclient.api.api import call_api
+from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
+    FullAutoAttachOptions,
+    _full_auto_attach,
+)
 from uaclient.api.u.pro.security.status.reboot_required.v1 import (
     _reboot_required,
 )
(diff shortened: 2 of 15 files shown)


uaclient/cli.py
```diff
@@ -1372,22 +1376,26 @@ def action_api(args, *, cfg):
 
 
 @assert_root
-@assert_lock_file("pro auto-attach")
-def action_auto_attach(args, *, cfg):
+def action_auto_attach(args, *, cfg: config.UAConfig) -> int:
     if cfg.is_attached:
```
aciba90, line 1380: The reason to maintain this guard is backwards compatibility. I realize `_full_auto_attach` checks against this condition too but, it does not conditionally auto attach depending on `skip_auto_attach`. So, in order to continue informing the user that the reason to not auto attach is that 'the instance is already attached' with more preference than 'skip_auto_attach' we need to maintain this check guard here.

If the previous change is an acceptable change of behavior, we could simplify this by removing the `is_attached` check.

lucasmoura: This is a good point. I think we should also include checking for the `disable_auto_attach` feature on the API too. This will make that endpoint more strict and simplify the cli method. What do you think @orndorffgrant @renanrodrigo @CalvoM ?

orndorffgrant: Yeah I think that makes sense, I can include that in the PR I'll put up touching the api full auto attach or we can do it in this PR

orndorffgrant: draft pr that includes this here https://github.com/canonical/ubuntu-advantage-client/pull/2299

aciba90: Could there exists a situation where callers of the API want to auto attach but `UAConfig`'s `disable_auto_attach` is true?

As the public `full_auto_attach` doesn't expose the ability to pass in a different `UAConfig`, there would be no clean way to override that, but tightly-coupling clients to `UAConfig`/the actual file.

I am thinking about the `cloud-init` use-case, but this would apply to other clients.

aciba90: Per internal discussions, we think that this situation from `cloud-init`'s perspective is a corner case. Furthermore, if the consumer would really want to force the auto attach, then the config module `cc_write-files` can be used to modify the UA config file, prior `cc_ubuntu_advantage` execution.

uaclient/cli.py
```diff
@@ -1372,22 +1377,23 @@ def action_api(args, *, cfg):
 
 
 @assert_root
-@assert_lock_file("pro auto-attach")
-def action_auto_attach(args, *, cfg):
-    if cfg.is_attached:
-        raise exceptions.AlreadyAttachedOnPROError()
-
-    skip_auto_attach = actions.should_disable_auto_attach(cfg)
```
orndorffgrant, line 1380: This is arguably out of scope so feel free to say no.

There is only one other caller of `actions.should_disable_auto_attach(cfg)` in lib/auto_attach.py and it can be replaced by catching `AutoAttachDisabledError`. If we do that as well then we can remove the `actions.should_disable_auto_attach` function. Then we can move the log statement defined in that function here.

uaclient/cli.py
```diff
@@ -1372,22 +1377,23 @@ def action_api(args, *, cfg):
...
+            cfg=cfg,
+            mode=event_logger.EventLoggerMode.CLI,
+        )
     except exceptions.UrlError:
         event.info(messages.ATTACH_FAILURE.msg)
         return 1
+    except api_exceptions.AutoAttachDisabledError:
+        return 0
+    except exceptions.AlreadyAttachedError as e:
+        raise exceptions.AlreadyAttachedOnPROError() from e
```
orndorffgrant, line 1393: `AlreadyAttachedOnPRO` only existed to change the return code to 0. And we onnly changed the return code to 0 in that case because we used to call this cli command in the on-boot ua-auto-attach.service.

Now that this is really only a CLI function to be called by a human, I think we should return 1 for both "already attached" and for "auto attach disabled". What do you think @lucasmoura @renanrodrigo @CalvoM 

If we follow this proposal then we don't need any special handling for these exceptions, and we can delete `AlreadyAttachedOnPRO`

uaclient/cli.py
```diff
@@ -1372,22 +1377,23 @@ def action_api(args, *, cfg):
...
+    except exceptions.UserFacingError as e:
+        event.info(e.msg)
+        return 1
```
orndorffgrant, line 1396: No need for this, it will be caught in the main error handler.
```suggestion
```

orndorffgrant (commented):
Excited for this one! After this then everything auto attaching will be calling the same function :tada:

Just some nits and ideas to simplify things a bit

uaclient/cli.py
aciba90: Thanks for the context here. My main concern was actually the message change, as the exceptions contain different error messages [1] that are later shown to the user. If that change is acceptable, we could simplify this and probably drop `AlreadyAttachedOnPRO` and its associated message, as we only use it here.

[1] https://github.com/canonical/ubuntu-advantage-client/blob/a5f3f0eb06a4ae6732d17fda4de5b11f298ed72b/uaclient/messages.py#L590-L602/blob/a5f3f0eb06a4ae6732d17fda4de5b11f298ed72b/uaclient/messages.py#L590-L602

uaclient/cli.py
aciba90: Let do that!

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
```diff
@@ -69,24 +69,29 @@ def full_auto_attach(options: FullAutoAttachOptions) -> FullAutoAttachResult:
...
 
 
 def _full_auto_attach_in_lock(
-    options: FullAutoAttachOptions, cfg: UAConfig
+    options: FullAutoAttachOptions,
+    cfg: UAConfig,
+    mode: event_logger.EventLoggerMode,
 ) -> FullAutoAttachResult:
-    event.set_event_mode(event_logger.EventLoggerMode.JSON)
+    event.set_event_mode(mode)
```
orndorffgrant, line 94: I just want to acknowledge this change. I don't like that we have to do this, but I think we're stuck doing this until more functions (i.e. enable and attach) are migrated to the API.

uaclient/cli.py
```diff
@@ -1372,22 +1377,18 @@ def action_api(args, *, cfg):
...
+        _full_auto_attach(
+            FullAutoAttachOptions(),
+            cfg=cfg,
+            mode=event_logger.EventLoggerMode.CLI,
+        )
     except exceptions.UrlError:
         event.info(messages.ATTACH_FAILURE.msg)
         return 1
+    except api_exceptions.AutoAttachDisabledError:
+        return 0
```
orndorffgrant, line 1391: I think we can return 1 for this error as well, so I think we can just let the main_error_handler handle this as well - which means we don't need these lines :D
```suggestion
```

uaclient/testing/helpers.py
```diff
@@ -1,12 +1,17 @@
-from contextlib import contextmanager
+try:  # Drop try-except after xenial EOL
+    from contextlib import AbstractContextManager
+except ImportError:
+    AbstractContextManager = object
```
orndorffgrant, line 4: I'm okay with these changes but I'm not sure if they're necessary. `tox -e py35-test` is already passing without these changes. Were you running into failures?

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
lucasmoura, line 94: I know why you are setting the `event_logger` here, but I wonder if we actually need to set this module here. Do we need this setup anymore @orndorffgrant @renanrodrigo @CalvoM ? It seems we are not even using it on the `actions` module anymore

orndorffgrant: I think `event.info` is used during `enable` which is why we still need this.

So I think we'll need to migrate enable to the api structure before we can remove this

uaclient/testing/helpers.py
aciba90: Yes, I run into failures in py-35 because [AbstractContextManager](https://docs.python.org/3/library/contextlib.html#contextlib.AbstractContextManager) was introduced in py-36.

uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py
aciba90: Yes, the reason to add this as a configurable parameter is to be able to set the `event_logger` in CLI mode when we use it from `uaclient.cli` and print out events to stdout and stderr, as we do without this refactoring.

aciba90: Being discussed here: https://github.com/canonical/ubuntu-advantage-client/pull/2297#discussion_r1010438664

uaclient/cli.py
aciba90: Done.

aciba90:
> LGTM, I think we only need the context for the return code changes on the latest commit

Thanks, @lucasmoura and @orndorffgrant, for the review. I have tried to explain the context of the behavior change. Please feel free to improve it if you think something is not correctly described.

lucasmoura:
@aciba90 That's a good explanation, I would just change this paragraph here:
```
Now, the cli returns non 0 exit codes in both cases. The cli returned
0s because the same auto-attach funcionality was used for the cli and
the auto-attach daemon (`lib/auto_attach.py`) and we wanted the deamon
to succeed in those cases.
```

to
```
Now, the cli returns non 0 exit codes in both cases. The cli returned
0s because we would call the cli command on the ua-auto-attach.service systemd unit.
If we detected that the user was already attached or hand explicit config disabling that service,
we didn't want to show that this service failed.
```

Additionally, I would suggest that we squash all the existing commits and put that explanation directly on the commit message, just to make the SRU review easier if we forget to bring that rationale on the SRU bug we create

aciba90:
Message improved and `Squash and merge` selected.

orndorffgrant (approved):
Thanks! I'll squash and merge with the given message

PR #2305, opened 2022-11-02
tests: clean ssh keys if instance is destroyed

lucasmoura:
## Proposed Commit Message
tests: clean ssh keys if instance is destroyed

If we choose to delete the instance after an integration test is complete, we should also remove the ssh key associated with it

## Test Steps
Run an integration test with `UACLIENT_BEHAVE_DESTROY_INSTANCES=1` and verify if the ssh keys are removed.
Run an integration test with `UACLIENT_BEHAVE_DESTROY_INSTANCES=0` and verify if the ssh keys are kept.

## Desired commit type

## Checklist

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -575,6 +575,16 @@ def after_all(context):
             else:
                 context.config.cloud_api.delete_image(image)
 
+    if context.config.destroy_instances:
+        try:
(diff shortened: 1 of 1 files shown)


orndorffgrant (approved):
:tada:

CalvoM (approved):
This will be a good addition.

Issue #917, opened 2019-11-13
Collapse same arches into a single name


panlinux:
With https://github.com/CanonicalLtd/ua-contracts/issues/539, which is live in the staging contract server at the moment, the server has joined the architectures list of `dpkg --print-architecture` and `uname -m`. That means the ua client will get, for example, `i386` and `i686` for x86 32bits, and `amd64` and `x86_64` for x86 64 bits, and similar aliases for other arches.
When trying to enable a service on an unsupported architecture, the client will say it's not possible because the current architecture X is not in the list of supported architectures, and will then present this list, which will contain all the aliases.

For the esm-infra case on ppc64el, for example, we get:
```
root@trusty-ppc64el:~# ua enable esm-infra
One moment, checking your subscription first
ESM Infra is not available for platform ppc64le.
Supported platforms are: amd64, i386, i686, x86_64
```
We should probably collapse that list and only use one architecture name. In this case, it should be `amd64` and `i386`.
Note this server change is only live on staging, and things are still a bit in flux, but we thought best to file this bug as is looks like we are going ahead with that staging change.

orndorffgrant:
Tracking in SC-1125

PR #2306, opened 2022-11-02
enable: show deduplicated list of supported arches

orndorffgrant:
Fixes: #917

## Test Steps

Run `pro enable` in a way that causes the arch affordance check to fail. Ensure that a de-duplicated list of arches is printed.

## Desired commit type

## Checklist

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -554,12 +554,13 @@ class UAEntitlement(metaclass=abc.ABCMeta):
             affordance_arches is not None
             and platform["arch"] not in affordance_arches
         ):
+            deduplicated_arches = util.deduplicate_arches(affordance_arches)
             return (
                 ApplicabilityStatus.INAPPLICABLE,
                 messages.INAPPLICABLE_ARCH.format(
                     title=self.title,
                     arch=platform["arch"],
-                    supported_arches=", ".join(affordance_arches),
+                    supported_arches=", ".join(deduplicated_arches),
                 ),
             )
         affordance_series = affordances.get("series", None)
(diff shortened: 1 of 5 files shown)


CalvoM:
Just a quick question @orndorffgrant, why do we have only two esm-supported arches in [postinst](https://github.com/canonical/ubuntu-advantage-client/blob/225499706b7f73d262dfb1609ecff5ee4094c2b6/debian/ubuntu-advantage-tools.postinst#L48) while in the [docs](https://github.com/canonical/ubuntu-advantage-client/blob/main/docs/references/support_matrix.md) we have more?

orndorffgrant:
Good question @CalvoM! The variable in postinst you're referencing is the list of architectures supported by ESM. The docs you're referencing are the architectures supported by the Pro Client. The Pro Client supports more architectures than ESM because it also delivers other products that support more architectures than ESM does.

lucasmoura (commented):
@orndorffgrant this would work for sure, but couldn't we change the contract definitions to only return the expected arches, the ones returned by `dpkg --print-architecture` for the `affordances` instead ?

orndorffgrant:
@lucasmoura we could! But I think it is correct for the server to understand the aliases for different architectures. IMO showing a de-duplicated set is a UI concern that is best handled by the client.

I believe the decision was originally made because of older clients that were still sending the other names. I'm not sure if that specifically applies anymore, but there is at least one other client to the API: the website for /pro. And there could be more clients to the API in the future. Those clients may need to use "x86_64" for some reason, so I think it is good that the API supports it.

lucasmoura:
@orndorffgrant Fair enough, good point

lucasmoura (approved):

PR #2310, opened 2022-11-08
mypy: use explicit-package-bases

orndorffgrant:
This should fix the recent CI mypy failures.

Adding this flag also uncovered some things mypy has been missing until now - so those are fixed as well.

We also get a note recommending the --check-untyped-defs flag. I think we should add that too, but it it reveals a lot more errors and I don't have time to fix them right now. And that isn't necessary to unblock CI checks. So adding --check-untyped-defs can be a future PR

## Desired commit type

## Checklist


lucasmoura:
Great finding @orndorffgrant

lucasmoura (approved):

PR #2311, opened 2022-11-09
tools: correct importation issues

CalvoM:
## Proposed Commit Message

tools: delete refresh scripts.

The refresh functionality is already covered in pycloudlib.

## Desired commit type

## Checklist


lucasmoura:
@CalvoM I think we can actually remove the refresh scripts, since we have moved that logic to pycloudlib

lucasmoura:
@CalvoM let's use change the commit message now that we are deleting the files

PR #2312, opened 2022-11-09
tests: modify pro status output on non-LTS release

lucasmoura:
## Proposed Commit Message
tests: modify pro status output on non-LTS release

On non-LTS releases, we don't provide any support to the exising services delivered by the pro client. Because of that, we are modifying the pro status output to reflect that

LP: 1994923

## Test Steps
Verify if the new output for `pro status` on non-LTS releases make sense

## Desired commit type

## Checklist

### DIFF
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -2,30 +2,6 @@
 Feature: Command behaviour when attaching a machine to an Ubuntu Pro
         subscription using a valid token
 
-    @series.kinetic
-    @series.lunar
-    @uses.config.machine_type.lxd.container
-    Scenario Outline: Attached command in a non-lts ubuntu machine
-        Given a `<release>` machine with ubuntu-advantage-tools installed
-        When I attach `contract_token` with sudo
-        And I run `pro status --all` as non-root
-        Then stdout matches regexp:
-            """
-            SERVICE       +ENTITLED  STATUS    DESCRIPTION
-            cc-eal        +yes      +n/a      +Common Criteria EAL2 Provisioning Packages
-            cis           +yes      +n/a      +Security compliance and audit tools
-            esm-apps      +yes      +n/a      +Expanded Security Maintenance for Applications
-            esm-infra     +yes      +n/a      +Expanded Security Maintenance for Infrastructure
-            fips          +yes      +n/a      +NIST-certified core packages
-            fips-updates  +yes      +n/a      +NIST-certified core packages with priority security updates
-            livepatch     +yes      +n/a      +Canonical Livepatch service
-            """
-
-        Examples: ubuntu release
-            | release |
-            | kinetic |
-            | lunar   |
-
     @series.lts
     @uses.config.machine_type.lxd.container
     Scenario Outline: Attach command in a ubuntu lxd container
--- a/features/attached_status.feature
+++ b/features/attached_status.feature
@@ -20,6 +20,7 @@ Feature: Attached status
            | kinetic |
            | lunar   |
 
+
     @series.xenial
     @uses.config.machine_type.lxd.container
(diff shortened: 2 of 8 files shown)


lechsandecki:
how about `You are running Ubuntu 22.10 `kinetic kudu` interim release. Interim Ubuntu releases are supported for 9 months and do not offer Ubuntu Pro features, such as the full 10 years security maintenance. If you wish to use Ubuntu Pro, please upgrade to an LTS release. Learn more about Ubuntu Pro at ubuntu.com/pro`

wdyt @elioqoshi ?

cpaelzer:
On Fri, Dec 9, 2022 at 9:44 PM lechsandecki ***@***.***>
wrote:

> ... You are running Ubuntu 22.10 kinetic kudu interim release ... please
> upgrade to an LTS release ...
>

I generally like that as first approach, but in several conditions there is
no "upgrade" to an LTS.
Like here, the latest LTS is 22.04 - and the next they could upgrade to is
24.04 which isn't released yet.
I think we need a different text if there is a follow on LTS and another
one if they are so new that they'd need to "go back".

lechsandecki:
good point. perhaps something like "switch" instead of an "upgrade"?

lucasmoura:
@cpaelzer @lechsandecki @elioqoshi can we agree to the message:
```
You are running Ubuntu 22.10 kinetic kudu interim release. Interim Ubuntu releases are supported for 9 months and do not offer Ubuntu Pro features, such as the full 10 years security maintenance. If you wish to use Ubuntu Pro, please switch to an LTS release. Learn more about Ubuntu Pro at ubuntu.com/pro
```

Additionally, should we also output that same message if the user is `attached` to a Pro subscription ?

lechsandecki:
Perhaps I am missing something but I think that the user can't be successfully attached to a Pro on an interim release, no? and there is no need to display that message if an LTS release is in use

lucasmoura:
@lechsandecki currently, an user can attach to a non-LTS release. The only benefit I can think of an user attaching in a non-LTS release, is to have Pro ready when he upgrades to a LTS release.

However, if you think that is not worth it, we can block attach on `non-LTS`  releases.

lechsandecki:
i am not sure if it's worth letting users attach pro subscriptions to non-lts releases. Let's discuss it with the broader team, pls

lucasmoura:
@lechsandecki I have discussed with the team and we are okay blocking the attach functionality for interim releases. However, before we do that we need clarifications on two questions:

1) Are we sure that we will not have any support for any service for the non-LTS releases in the near future ?
2) What should happen if an attached LTS user upgrades to a non-LTS release ? For example, if I am attached on Jammy and upgrade to Kinetic, should we end up detached on Kinetic ? Should we alert the user that the machine will no longer be Pro supported after the upgrade ?

lechsandecki:
picking that topic again. Frankly, either way would work - blocking from attaching to interim release is most straightforward, but if you believe that upgrades between LTS and non-LTS are happening that often, then perhaps we should simply display the message that Ubuntu Pro is not available whenever an interim release is deployed.

lucasmoura:
@lechsandecki it seems we don't have a way to tell the user before the `do-release-upgrade` operation that the Pro subscription is no longer useful when upgrading from a LTS to non-LTS. I can maybe add that at the end of the upgrade operation instead, but I don't know if that is still valuable though. What do you think ?

aburrage-canonical:
Given that a user running an interim release will always have the option to downgrade to an LTS, I think it is fine to allow the user to take the requested options without us interfering with messages, but blocking them from attaching a pro-token, since that will have no value.

lucasmoura:
Closing this PR as we have decided to change the message on interim releases

PR #2313, opened 2022-11-10
Context based security-status API endpoints (SC-1319)

renanrodrigo:
Create context-based endpoints for the output of Security Status.
Also, include all updates instead of just security ones when fetching through the API.

## Desired commit type

## Checklist
 - [ ] I have updated or added any documentation accordingly - This will be a new commit/PR, will send it asap


uaclient/api/api.py
```diff
@@ -16,15 +16,18 @@
 from uaclient.version import check_for_new_version
 
 VALID_ENDPOINTS = [
```
orndorffgrant, line 18: thanks for sorting these :D

uaclient/security_status.py
```diff
@@ -171,7 +183,8 @@ def get_ua_info(cfg: UAConfig) -> Dict[str, Any]:
     return ua_info
 
 
-def get_livepatch_fixed_cves() -> List[Dict[str, str]]:
+# Yeah Any is bad, but so is python<3.8 without TypedDict
```
orndorffgrant, line 186: Very true :cry: 
This is fine IMO.

uaclient/security_status.py
```diff
@@ -129,23 +129,35 @@ def get_update_status(service_name: str, ua_info: Dict[str, Any]) -> str:
...
+                            break  # No need to loop through all the origins
+                    # Also no need to report backports at least for now...
+                    expected_origin = version.origins[0]
+                    if (
+                        not is_security
+                        and "backports" not in expected_origin.archive
+                    ):
+                        result["standard-updates"].append(
+                            (version, version.origins[0].site)
+                        )
```
orndorffgrant, line 160: Could this report the same _version_ twice? once for -security and once -updates (I don't think we want that but maybe I'm misremembering)

orndorffgrant (commented):
Looks really good! One implementation question

And I'll do some manual testing now too

uaclient/api/api.py
renanrodrigo: OCD triggers

uaclient/security_status.py
renanrodrigo: It won't. Any version will be only listed once. Then... well:
- if it is in `-security` AND in `-updates` (which is a pretty common case) we report security - and that makes sense as it is indeed a security update.
- if it is in esm-infra AND esm-apps, then OH WELL OH WELL the first origin gets picked based on apt configuration, but this is a corner case for a very limited set of packages, and when we implemented this the first time we agreed that was good. Let me know if there is any major concern now about this decision.

orndorffgrant: Ah! You're right - I misread the code. This all sounds good.

uaclient/api/u/pro/packages/updates/v1.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def _updates(cfg: UAConfig) -> PackageUpdatesResult:
+    ua_info = get_ua_info(cfg)
+    packages = get_installed_packages_by_origin()
+    upgradable_versions = filter_security_updates(packages["all"])
+    update_list = create_updates_list(upgradable_versions, ua_info)
+
+    num_esm_apps_updates = len(upgradable_versions["esm-apps"])
+    num_esm_infra_updates = len(upgradable_versions["esm-infra"])
```
lucasmoura, line 94: Can't we use the list of `esm-apps` and `esm-infra` in the return of  `get_installed_packages_by_origin` here ?

uaclient/security_status.py
```diff
@@ -129,23 +129,35 @@ def get_update_status(service_name: str, ua_info: Dict[str, Any]) -> str:
...
+                    for origin in version.origins:
+                        service = get_origin_information_to_service_map().get(
+                            (origin.origin, origin.archive)
+                        )
+                        if service:
+                            result[service].append((version, origin.site))
+                            is_security = True
+                            break  # No need to loop through all the origins
+                    # Also no need to report backports at least for now...
+                    expected_origin = version.origins[0]
```
lucasmoura, line 153: Why are using the first origin here ?

uaclient/security_status.py
```diff
@@ -129,23 +129,35 @@ def get_update_status(service_name: str, ua_info: Dict[str, Any]) -> str:
...
+                    if (
+                        not is_security
+                        and "backports" not in expected_origin.archive
+                    ):
+                        result["standard-updates"].append(
+                            (version, version.origins[0].site)
```
lucasmoura, line 159: We can replace `version.origins[0]` for `expected_origin`

orndorffgrant (approved):
LGTM

uaclient/security_status.py
renanrodrigo: Because at this point, we know it is a regular update, which will be either on main/universe/multiverse/restricted but not in more than one of them. So, the first origin is where APT is going to download it from, and given it is not security, we're fine with it.

uaclient/api/u/pro/packages/updates/v1.py
renanrodrigo: Nope, unfortunately - a package from `main` (returned in  `get_installed_packages_by_origin`) may have an update in `esm-infra` - so we need to filter the upgradable versions

lucasmoura (approved):

PR #2315, opened 2022-11-11
Release 27.11.4

lucasmoura:


debian/changelog
```diff
@@ -1,3 +1,27 @@
+ubuntu-advantage-tools (27.11.4~22.10.1) kinetic; urgency=medium
+
+  * New upstream release 27.11.4:
+    - auto-attach:
+      + use auto-attach api in cli entry-point
+      + retry auto-attach for up to one month on Ubuntu Pro cloud instances
+      + best_effort auto_attach api instead of fail fast
```
renanrodrigo, line 7: suggestion: `make a best effort to auto-attach when using the API` ?

debian/changelog
```diff
@@ -1,3 +1,27 @@
...
+    - enable: show deduplicated list of supported arches (GH: #917)
+    - fips: remove cloud package override logic
```
renanrodrigo, line 9: hhm, I think we should make explicit we are removing it *from the client side*

debian/changelog
```diff
@@ -1,3 +1,27 @@
...
+    - messaging: verify contract expiration date on contract server before
+      outputting expired message on MOTD
+    - realtime-kernel: make service non-beta
+    - reboot-required:
+      + add api support to show if the system requires a reboot
+      + add cli command for the functionality
+    - security-status:
+      + add api support to report standard updates
+      + add api support to show livepatch cves
```
renanrodrigo, line 18: s/livepatch cves/CVEs patched by Livepatch

debian/changelog
```diff
@@ -1,3 +1,27 @@
...
+      + add api support to show packages summary information
+      + list packages in oci manifest format
+    - systemd: add machine token condition for auto-attach
```
renanrodrigo, line 21: suggestion: `do not attempt to auto-attach if a machine-token is present`

renanrodrigo (commented):
One question, a few suggestions

renanrodrigo:
And, shouldn't this be `27.12` instead of `27.11.4` ?

lucasmoura:
Yep, it should :facepalm: 
Good catch @renanrodrigo

renanrodrigo (approved):
LGTM
:shipit:

debian/changelog
```diff
@@ -1,3 +1,27 @@
+ubuntu-advantage-tools (27.12~23.04.1) lunar; urgency=medium
+
+  * New upstream release 27.12 (LP: #1996424):
+    - auto-attach:
+      + use auto-attach api in cli entry-point
```
orndorffgrant, line 5: this line is an implementation detail, not sure if necessary
```suggestion
```

debian/changelog
```diff
@@ -1,3 +1,27 @@
...
+      outputting expired message on MOTD
+    - realtime-kernel: make service non-beta
+    - reboot-required:
+      + add API support to show if the system requires a reboot
+      + add cli command for the functionality
+    - security-status:
+      + add API support to report standard updates
+      + add API support to show CVEs patched by Livepatch
+      + add API support to show packages summary information
+      + list packages in oci manifest format
```
orndorffgrant, line 20: should we include the api endpoint names and new cli commands?

orndorffgrant (approved):
LGTM! comments are optional

debian/changelog
orndorffgrant: I think that looks good!

orndorffgrant (approved):

PR #2320, opened 2022-11-15
docs: add github issue feedback link to each page

orndorffgrant:
The goal here is to add a link to each page to easily open a GitHub
issue related to that page's contents. This can be an easy way to ask a
question or report a problem with that piece of documentation.

Instead of forking the furo theme, this uses some custom CSS/JS to add
an extra element to the theme after the page loads. We will likely
switch to a more Ubuntu-styled theme in the future, so this is just a
hack for now to prove the concept. This hack can be adapted to a
future theme, or, ideally, this feature would get built into the
future theme.

## Test Steps

`tox -e dev-docs` and open the docs in your browser. Make sure the link at the top of each page makes sense and links to an appropriate github issue form.

## Desired commit type

## Checklist

### DIFF
--- a/docs/_static/css/github_issue_links.css
+++ b/docs/_static/css/github_issue_links.css
@@ -1,0 +1,7 @@
+.github-issue-link-container {
+    padding-right: 0.5rem;
+}
+.github-issue-link {
+    font-size: var(--font-size--small);
+    font-weight: bold;
+}
--- a/docs/_static/js/github_issue_links.js
+++ b/docs/_static/js/github_issue_links.js
@@ -1,0 +1,25 @@
+window.onload = function() {
+    const link = document.createElement("a");
+    link.classList.add("muted-link");
+    link.classList.add("github-issue-link");
+    link.text = "Have a question?";
+    link.href = (
+        "https://github.com/canonical/ubuntu-advantage-client/issues/new?"
+        + "title=docs%3A+TYPE+YOUR+QUESTION+HERE"
+        + "&body=*Please describe the question or issue you're facing with "
+        + `"${document.title}"`
+        + ".*"
+        + "%0A%0A%0A%0A%0A"
+        + "---"
+        + "%0A"
+        + `*Reported+from%3A+${location.href}*`
+    );
+    link.target = "_blank";
+
+    const div = document.createElement("div");
+    div.classList.add("github-issue-link-container");
+    div.append(link)
+
+    const container = document.querySelector(".article-container > .content-icon-container");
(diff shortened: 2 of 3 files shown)


docs/_static/css/github_issue_links.css
```diff
@@ -0,0 +1,6 @@
+.github-issue-link-container {
+    padding-right: 0.5rem;
+}
+.github-issue-link {
+    font-size: 0.8125rem;
```
renanrodrigo, line 5: I wonder where those numbers come from :D

docs/_static/css/github_issue_links.css
```diff
@@ -0,0 +1,6 @@
+.github-issue-link-container {
+    padding-right: 0.5rem;
+}
+.github-issue-link {
```
renanrodrigo, line 4: As a suggestion, I think it calls a little more attention if we make it
`font-weight: bold`
Feel free to ignore if disagree though

docs/_static/js/github_issue_links.js
```diff
@@ -0,0 +1,25 @@
...
+    link.classList.add("muted-link");
+    link.classList.add("github-issue-link");
+    link.text = "Have a Question?";
+    link.href = (
+        "https://github.com/canonical/ubuntu-advantage-client/issues/new?"
+        + "title=docs%3A+TYPE+YOUR+QUESTION+HERE"
+        + "&body=*Please describe the question or issue you're facing with "
+        + `"${document.title}"`
+        + ".*"
+        + "%0A%0A%0A%0A%0A"
```
renanrodrigo, line 12: oaoaoaoa love this

renanrodrigo (approved):
As a final suggestion - do we want to hide the small "edit this page" pencil right next to this new link?

docs/_static/js/github_issue_links.js
```diff
@@ -0,0 +1,25 @@
+window.onload = function() {
+    const link = document.createElement("a");
+    link.classList.add("muted-link");
+    link.classList.add("github-issue-link");
+    link.text = "Have a Question?";
```
s-makin, line 5: "Have a question?"

s-makin (approved):

docs/_static/css/github_issue_links.css
orndorffgrant: I saw it used elsewhere in the theme, but now I looked again and found `var(--font-size-small)` in the theme which is about the same but uses a variable which is cooler so I switched to that

PR #2321, opened 2022-11-16
docs: migrate to furo theme

orndorffgrant:
While at it, adds some features that are recommended and explained in the furo docs. These include the logo, admonitions, and copy button for code blocks.

And also fixes some warnings we had in our sphinx build around syntax highlighting.

## Test Steps

Try out `tox -e docs-dev`
make sure everything looks right

## Desired commit type

## Checklist

 - [ ] Yes
 - [x] No


docs/howtoguides/enable_in_dockerfile.md
```diff
@@ -49,41 +50,41 @@ FROM ubuntu:focal
...
     # Install them here, while Ubuntu Pro is enabled, so that you get the appropriate
     # versions.
     # Any `apt-get install ...` commands you have in an existing Dockerfile
     # that you may be migrating to use Ubuntu Pro should probably be moved here.
     && apt-get install -y openssl \
-
+    #
     ###########################################################################
-    # Now that we've upgraded and installed any packages from the Ubuntu Pro
+    # Now that we have upgraded and installed any packages from the Ubuntu Pro
```
orndorffgrant, line 84: These grammatical changes are to avoid `'`s in the comments, which were causing the syntax highlighting to fail

docs-requirements.txt
```diff
@@ -1,5 +1,6 @@
-sphinx==5.1.1
+furo
 m2r2
 myst-parser
-sphinx_rtd_theme>=1.0.0
-docutils<0.17
+sphinx==5.1.1
```
renanrodrigo, line 4: WOW Unordered lists without stupid version workarounds I love this

dev-docs/explanations/how_auto_attach_works.md
```diff
@@ -17,8 +17,9 @@ The `pro auto-attach` command follows a specific flow on every **Public Cloud Ub
    * **Azure**: http://169.254.169.254/metadata/attested/document?api-version=2020-09-01
    * **GCP**: http://metadata/computeMetadata/v1/instance/service-accounts/default/identity
 
-> **Note**
-> On some instances, like AWS, we can also use the IPv6 address to fetch the metadata
+```{note}
+On some instances, like AWS, we can also use the IPv6 address to fetch the metadata
```
s-makin, line 21: Is an instance a 'thing' that some other thing sits on? (in which case, this is fine and please disregard my comment). 

Or did you mean it in the sense of "in some cases"? If so, this should be "In some instances".

s-makin (approved):
I had a look and everything looks fine. No weird errors or things displaying incorrectly that I could spot.

dev-docs/explanations/how_auto_attach_works.md
renanrodrigo: `Instance` in this case refers to a running Virtual Machine - so I think it makes sense as is.
However, we could rephrase it as
```suggestion
Some instances, in AWS for example, are able to also use the IPv6 address to fetch the metadata
```
or something alike

orndorffgrant: Just double checked and this is only a feature on AWS (for now), so I think I'll just say that outright
```suggestion
On AWS, the client will also try the IPv6 address (`[fd00:ec2::254]`) to fetch the metadata if the IPv4 address doesn't work.
```

lucasmoura (approved):

PR #2322, opened 2022-11-17
Add missing docs for the API endpoints

renanrodrigo:
Docs should now cover all API functionality.
Adjusted small pieces in the code for safety / compliance with the standards.

## Desired commit type

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/api/u/security/package_manifest/v1.py
+++ b/uaclient/api/u/security/package_manifest/v1.py
@@ -5,7 +5,7 @@ from uaclient.config import UAConfig
 from uaclient.data_types import DataObject, Field, StringDataValue
 
 
-class PackageManifestResults(DataObject, AdditionalInfo):
+class PackageManifestResult(DataObject, AdditionalInfo):
     fields = [
         Field("manifest_data", StringDataValue),
     ]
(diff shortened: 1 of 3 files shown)


uaclient/api/u/security/package_manifest/v1.py
```diff
@@ -5,7 +5,7 @@
 from uaclient.data_types import DataObject, Field, StringDataValue
 
 
-class PackageManifestResults(DataObject, AdditionalInfo):
```
orndorffgrant, line 8: Technically this is a breaking change, but I think we're confident no one is relying on importing this class so I think we can make it.

orndorffgrant (approved):

lucasmoura:
@s-makin feel free to review the new addition to the api docs if you have some time

s-makin (approved):

uaclient/api/u/security/package_manifest/v1.py
renanrodrigo: I kept the backwards compatibility using Alberto's suggestion

PR #2329, opened 2022-11-23
apt: move all message output to json-hook

orndorffgrant:
This means that all messages including esm update messages will only show up when using the apt binary (not apt-get).

This also simplifies some of our code because granularity of the json-hook config and method names replaces all of the command line parsing that was previously done to make sure that messages were printed for the right commands.

## Test Steps

Test various apt commands and scenarios. Make sure all the existing messages (esm-apps in particular) are still surfaced accurately, but only for the `apt` binary.

## Desired commit type

## Checklist

 - [ ] Yes
 - [x] No
This makes the apt hook integration simpler and less "loud" but doesn't substantially change behavior, so I don't think we need extra review.


apt-hook/esm-templates.cc
```diff
@@ -0,0 +1,273 @@
+#include <apt-pkg/cachefile.h>
```
orndorffgrant, line 1: I split `hook.cc` into three files:
* esm-templates.cc (this one) as most of the pre-existing code
* esm-templates.hh
* hook.cc is just a main function now

This was done so that we could use the `process_all_templates` function from the json-hook binary.

apt-hook/esm-templates.cc
```diff
@@ -0,0 +1,273 @@
...
+      space_separated_esm_i_packages.append(res.esm_i_packages[res.esm_i_packages.size() - 1]);
+   }
+   std::string space_separated_esm_a_packages = "";
+   if (res.esm_a_packages.size() > 0) {
+      for (uint i = 0; i < res.esm_a_packages.size() - 1; i++) {
+         space_separated_esm_a_packages.append(res.esm_a_packages[i]);
+         space_separated_esm_a_packages.append(" ");
+      }
+      space_separated_esm_a_packages.append(res.esm_a_packages[res.esm_a_packages.size() - 1]);
+   }
```
orndorffgrant, line 178: all of this used to be in `main` but I moved it into this function because it is all necessary whether the old hook is executed or if this function is called by the json hook

features/apt_messages.feature
```diff
@@ -113,14 +113,14 @@ Feature: APT Messages
         Building dependency tree...
         Reading state information...
         Calculating upgrade...
+        The following security updates require Ubuntu Pro with 'esm-infra' enabled:
+          .*
+        Learn more about Ubuntu Pro for 16\.04 at https:\/\/ubuntu\.com\/16-04
```
orndorffgrant, line 118: I moved this to the top so that it would be next to the similar messaging around packages that need an `apt autoremove`

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,13 +1,5 @@
 APT::Update::Post-Invoke-Stats {
-	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook post-invoke-stats || true";
-};
-
-APT::Install::Post-Invoke-Success {
-	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook post-invoke-success || true";
-};
```
orndorffgrant, line 7: we probably haven't actually needed this one for a long time

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,13 +1,5 @@
 APT::Update::Post-Invoke-Stats {
-	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook post-invoke-stats || true";
-};
-
-APT::Install::Post-Invoke-Success {
-	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook post-invoke-success || true";
-}; 
-
-APT::Install::Pre-Invoke {
-	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook pre-invoke || true";
```
orndorffgrant, line 10: no longer needed because this is now happening in the json hook configured below

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,13 +1,5 @@
...
+	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-hook ] || /usr/lib/ubuntu-advantage/apt-esm-hook || true";
```
orndorffgrant, line 2: there is now only one use-case for the hook, which is rendering the message templates with esm package counts/names, so no cli args are required anymore

features/apt_messages.feature
renanrodrigo: Should we have another call in this(these maybe) test(s) to ensure that the message is not shown in `apt-get`?

renanrodrigo (approved):
lgtm

features/apt_messages.feature
orndorffgrant: Good idea! I'll add one in an extra commit

orndorffgrant: updated!

Issue #2280, opened 2022-10-04
cli: auto-completion does not work for pro


aciba90:
auto-completion does work for `ua` but not for `pro`

## Steps to reproduce

```bash
$ ua <TAB>
ua 
--debug          --version        attach           collect-logs     detach           enable           help             refresh          status           version          
--help           api              auto-attach      config           disable          fix              list             security-status  system 

$ pro <TAB>
.bash_history  .bashrc        .profile ...

$ ua --version
27.11~22.10.1
```

orndorffgrant:
Thanks for pointing this out!
When fixing this we should also add a test for regressions in auto-complete :)

Issue #2279, opened 2022-10-04
cli: auto-completion shows a system command that does not exist


aciba90:
```bash
lxc launch ubuntu-daily:kinetic dev-k && lxc shell dev-k

$ ua TAB
--debug          --version        attach           collect-logs     detach           enable           help             refresh          status           version          
--help           api              auto-attach      config           disable          fix              list             security-status  system

$ ua system
usage: pro <command> [flags]
argument : invalid choice: 'system' (choose from 'attach', 'api', 'auto-attach', 'collect-logs', 'config', 'detach', 'disable', 'enable', 'fix', 'security-status', 'help', 'refresh', 'status', 'version')

$ ua --version
27.11~22.10.1
```

orndorffgrant:
This is fixed in 27.12 :)

renanrodrigo:
Re-opening, as we realized this is unintended behavior

PR #2330, opened 2022-11-24
Autocomplete fixes and tests

renanrodrigo:
Make sure we test autocomplete, don't show nonexistent sub-commands, don't set unexpected environment variables and make it work for `pro`.

Fixes: #2279 
Fixes: #2280 

## Desired commit type

## Checklist

 - [ ] Yes
 - [x] No


aciba90 (approved):
LGTM, thanks @renanrodrigo. I really like the way it is tested!

renanrodrigo:
hah thanks!
Well I will have to review that, as in Xenial it seems the order of the services is not right... but it's really close

CalvoM (approved):
LGTM, can zsh users have autocompletion as well... asking for a friend...😁😁

renanrodrigo:
@CalvoM hmmm probably yes.  Googling around I haz https://linuxhint.com/setup-configure-autocomplete-zsh/ - should work maybe?
Or, I wonder if there is something to be done in u-a-t side but I am not sure

orndorffgrant:
created https://warthogs.atlassian.net/browse/SC-1359 to track completion for non-bash shells

features/steps/misc.py
```diff
@@ -4,10 +4,25 @@
...
+
+set command [lindex $argv 0];
+set send_human {.1 .3 1 .05 2}
+set timeout 10
+
+spawn /bin/bash
+expect -exact "$"
+send -h "$command \t\t"
+expect -exact "$ $command"
+"""
```
orndorffgrant, line 24: very cool!

features/steps/misc.py
```diff
@@ -38,3 +53,24 @@ def verify_timer_interval_for_job(context, job, interval):
...
+    )
+
+
+@when("I press tab twice to autocomplete the `{command}` command")
+def when_i_autocomplete_command(context, command):
+    when_i_run_command(
+        context,
+        "/tmp/test_autocomplete.exp '{}'".format(command),
+        "as non-root",
+    )
```
orndorffgrant, line 76: If we want to include xenial again and maybe make this a little less fragile as we add more subcommands. We could do something like this
```suggestion
    )
    context.autocomplete_suggestions = sorted(context.process.stdout.split())
    
@then("the autocomplete suggestions are")
def then_the_autocomplete_suggestions_are(context):
    assert sorted(context.text.split()) == context.autocomplete_suggestions
```

Then in the feature file, something like:
```gherkin
        And I press tab twice to autocomplete the `ua` command
        Then the autocomplete suggestions are
        """
        --debug
        --help
        --version
        api
        attach
        auto-attach
        collect-logs
        config
        detach
        disable
        enable
        fix
        help
        refresh
        security-status
        status
        system
        version
        """
```

thoughts? Is that too complicated for this? or does it help?

tools/ua.bash
```diff
@@ -1,19 +1,14 @@
...
 {
-    local cur_word prev_word
+    local cur_word prev_word services subcmds base_params
     cur_word="${COMP_WORDS[COMP_CWORD]}"
     prev_word="${COMP_WORDS[COMP_CWORD-1]}"
 
-    subcmds=$(pro --help | awk '/^\s*$|Available|Use/ {next;} /Flags:/{flag=1;next}/Use ubuntu-avantage/{flag=0}flag{ if ( $1 ~ /,/ ) { print $2} else print $1}')
+    services=$(python3 -c "from uaclient.entitlements import valid_services; from uaclient.config import UAConfig; print(*valid_services(cfg=UAConfig()), sep=' ')
+")
+    subcmds=$(pro --help | awk '/^\s*$|^\s{5,}|Available|Use/ {next;} /Flags:/{check=1;next}/Use ubuntu-avantage/{check=0}check{ if ( $1 ~ /,/ ) { print $2} else print $1}')
```
orndorffgrant, line 11: one day I hope can we introduce something like `pro autocomplete` that does all this instead of an ad-hoc inline python script and awk-ing the help output

until then I think this change makes sense

features/steps/misc.py
renanrodrigo: Thank Christian for this one

features/steps/misc.py
renanrodrigo: 1. Not including Xenial is not making us fragile, as we know the output is always the same and ordering is the only difference
2. There is more manipulation that needs to be done in the output, as it would not be just exactly the autocomplete options, but rather the whole shell output

I vote not worth it

tools/ua.bash
renanrodrigo: Is your intention to say that a `awk` against the help output is some kind of ugly undesired workaround?
You're totally right.
`pro autocomplete` can be worked on in the future for sure.

orndorffgrant: yes that is what I was trying to say :upside_down_face:

features/steps/misc.py
orndorffgrant: Sorry I meant "include xenial again" and "make this less fragile" as two separate concerns.

I agree that including xenial isn't that important.

My "fragile" concern was that when we add a new subcommand or flag, then the multi-column suggestion output will be annoying to update - because an addition in the first column will shift all the other values.


> There is more manipulation that needs to be done in the output, as it would not be just exactly the autocomplete options, but rather the whole shell output

I didn't realize that, and I agree that makes this not worth it

orndorffgrant (approved):

PR #2336, opened 2022-11-29
tests: change the packages for the security status test (SC-1353)

renanrodrigo:
To test the full output of security-status, we install packages to count them as third-party and unknown.
We install `code` to be the third party option, and `gh` to be the unknown package. The reason for that is that the `code` deb sets up a parts file with a source entry (thus being recognized as third-party) and `gh ` does not.

However, `gh` just landed on `universe` from `kinetic` on. So, when testing for the release (when u-a-t is installed from `-proposed`), there is no unknown package anymore :|

[EDITED] 
~The """"fix"""" here is to change the `gh` package to any other which is not in the archive. The chosen one is `surfshark-release` just because I could not find a `.deb` in the wild which wouldn't setup a repository. This one does, but we remove the parts file to make sure it is counted as unknown.~

The better and actual fix suggested by @orndorffgrant here is to install dummy packages that we created instead.
Example packages were created and hosted appropriately in a S3 bucket, and the approach works like a charm.

F̶e̶e̶l̶ ̶f̶r̶e̶e̶ ̶t̶o̶ ̶h̶a̶t̶e̶ ̶t̶h̶i̶s̶ ̶P̶R̶.̶ 
It is mandatory to love this PR now.

## Test Steps
Run the Kinetic security-status test, installing u-a-t from any source not local/prebuilt, and see it fail. Apply this change and see it passing.

## Desired commit type

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/steps/packages.py
+++ b/features/steps/packages.py
@@ -3,8 +3,15 @@ import re
 from behave import then, when
 from hamcrest import assert_that, contains_string, matches_regexp
 
+from features.steps.files import when_i_create_file_with_content
 from features.steps.shell import when_i_run_command
 from features.util import SUT
+
+APT_REPO_ENTRY = (
+    "deb [trusted=yes] "
+    "http://proclientdummyrepo.s3-website.us-east-2.amazonaws.com/"
+    "actual_ppa/ unstable main"
+)
 
 
 @when("I apt install `{package_name}`")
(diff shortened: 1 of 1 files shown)


orndorffgrant (approved):
Approving with the request that we change how this test works in the future.
Ideally we wouldn't be relying on these randomly picked third party debs.
We should build our own minimal debs specifically for this test.

renanrodrigo:
@orndorffgrant the only thing now is that I am hosting those packages using my company credentials - which is fine for now, but in the future we should migrate those to some team-owned bucket.
If you think this is acceptable for now, then it is done. If you think it is harmful for me to host those, we can discuss and migrate, and then update this PR to point to the appropriate URLs before merging.

orndorffgrant (approved):
Created https://warthogs.atlassian.net/browse/SC-1362 to move them to team owned infra

Feel free to merge after resolving conflicts

PR #2337, opened 2022-11-29
Remove jobs-status from config

lucasmoura:
## Proposed Commit Message
remove jobs-status from config

Remove jobs-status from config and move it to the `state_file` module.
Additionally, this PR is also fixing an issue on the `metering` job where we don't honor disabling the service if we have
`activity_ping_interval` on `machine-token.json`

## Test Steps
Verify that the changes in the integration tests are passing

## Desired commit type

## Checklist

 - [ ] Yes
 - [ ] No


lib/timer.py
```diff
@@ -112,20 +124,30 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
...
+        jobs_status_obj = AllTimerJobsState(
+            metering=None, update_messaging=None
+        )
+
     for job in UACLIENT_JOBS:
-        if job.name in jobs_status:
-            next_run = jobs_status[job.name]["next_run"]
-            if next_run > current_time:
+        if getattr(jobs_status_obj, job.name):
+            next_run = getattr(jobs_status_obj, job.name).next_run
```
orndorffgrant, line 138: by using `getattr` here (and `setattr` later), mypy can't check most of this for errors. But also I understand why you're doing it here.

If you can think of a way to avoid `getattr` then let's try it, but if not then that's okay too.

orndorffgrant (approved):
Great! Piece by piece we will dismantle the all-knowing `UAConfig` :smiling_imp: 

I think `apport/source_ubuntu-advantage-tools.py` needs updating as well.

lib/timer.py
lucasmoura: @orndorffgrant I have modified the PR to make it more type friendly. Let me know what you think

PR #2338, opened 2022-12-01
apt news

orndorffgrant:
- [x] make it work
- [x] integration tests
- [x] add unit tests for all new functions as part of each respective commit
- [ ] make write_file atomic using `mv`-equivalent #2356 

## Test Steps

TODO

## Desired commit type

## Checklist

 - [x] Yes
 - [ ] No


apt-hook/20apt-esm-hook.conf
```diff
@@ -1,3 +1,7 @@
+APT::Update::Pre-Invoke {
+	"[ ! -f /usr/lib/ubuntu-advantage/apt_update_nonblocking_hook.py ] || flock -n -E 0 /var/lib/ubuntu-advantage/apt-update-lock /usr/lib/ubuntu-advantage/apt_update_nonblocking_hook.py &";
```
orndorffgrant, line 2: flock may require a dependency?

orndorffgrant: it's essential yes, so no dep required to use it:
```
> dpkg -S /usr/bin/flock 
util-linux: /usr/bin/flock

> apt info util-linux
Package: util-linux
Version: 2.38-4ubuntu1
Priority: required
Essential: yes
...
```

orndorffgrant: and that is true back in xenial as well

features/cloud.py
```diff
@@ -151,19 +151,8 @@ def launch(
             inbound_ports=inbound_ports,
         )
         logging.info(
-            "--- {} instance launched: {}. Waiting for ssh access".format(
-                self.name, inst.name
-            )
+            "--- {} instance launched: {}".format(self.name, inst.name)
         )
-        time.sleep(15)
-        for sleep in (5, 10, 15):
```
lucasmoura, line 159: Why are we removing this retry mechanism ?

lib/apt_update_nonblocking_hook.py
```diff
@@ -0,0 +1,37 @@
...
+    last_update = apt.get_apt_cache_datetime()
+    one_day_ago = datetime.datetime.now(
+        datetime.timezone.utc
+    ) - datetime.timedelta(days=1)
+    if last_update is not None and last_update > one_day_ago:
+        return
+    fetch_and_process_apt_news(cfg)
+
+
+def main(cfg: UAConfig):
```
lucasmoura, line 22: Do we need this function ? I think we can perform this check directly on `__main__`

uaclient/apt_news.py
```diff
@@ -0,0 +1,153 @@
...
+        self.lines = lines
+
+
+def is_control_char(c: str) -> bool:
+    return unicodedata.category(c)[0] == "C"
+
+
+def clean_line(line: str) -> str:
+    line_without_control_chars = "".join(
+        [c for c in line if not is_control_char(c)]
```
lucasmoura, line 56: This is just a safe mechanism right ? Because we shouldn't allow control characters on the apt news messages in the first place

uaclient/apt_news.py
```diff
@@ -0,0 +1,153 @@
...
+            # All control characters will be removed.
+            # Newlines are particularly important to formatting expectations.
+            # Since we can't honor newlines, and removing them could
+            # substantially change meaning, we consider messages with them
+            # to be invalid.
+            return False
+        if len(line) > 77:
+            return False
+
+    if all(len(clean_line(line)) == 0 for line in msg.lines):
```
lucasmoura, line 114: What happens if one of the lines after cleaning has a zero length size ? Can it potentially change the meaning of the message ?

I know we are already checking for the existence of newlines and tabs, but I am just wondering if there are other special character that could compromise the message if excluded

features/cloud.py
orndorffgrant: From conversation: I'll make a separate PR for this and we'll see if it breaks anything :upside_down_face:

lib/apt_update_nonblocking_hook.py
orndorffgrant: You're right, that extra function isn't helping us here - I'll simplify it

uaclient/apt_news.py
orndorffgrant: From conversation: We'll just disallow control characters entirely instead of trying to remove them. I'll make that change

uaclient/apt_news.py
orndorffgrant: Right, but we can't trust the server to be strict (the url is configurable) so we do need these kinds of checks in the client as well.

features/cloud.py
orndorffgrant: https://github.com/canonical/ubuntu-advantage-client/pull/2347

lib/apt_update_nonblocking_hook.py
orndorffgrant: done!

uaclient/apt_news.py
orndorffgrant: done!

lib/apt_update_nonblocking_hook.py
```diff
@@ -0,0 +1,28 @@
...
+from uaclient.config import UAConfig
+from uaclient.daemon import setup_logging
+
+
+def main(cfg: UAConfig):
+    if cfg.apt_news:
+        last_update = apt.get_apt_cache_datetime()
+        one_day_ago = datetime.now(timezone.utc) - timedelta(days=1)
+        if last_update is None or last_update < one_day_ago:
+            fetch_and_process_apt_news(cfg)
```
orndorffgrant, line 17: @basak this is where the additional checks happen to avoid unnecessary apt_news requests.

`if cfg.apt_news` is what respects the `pro config set apt_news=false` setting

the date logic compares "now" to the timestamp of the apt-update stamp file. If the apt-update stamp is within the past 24 hours, we don't do anything here.

LMK what you think

uaclient/apt_news.py
```diff
@@ -0,0 +1,181 @@
...
+    acq = apt_pkg.Acquire()
+    apt_news_file = apt_pkg.AcquireFile(
+        acq, cfg.apt_news_url, destdir=defaults.UAC_RUN_PATH
+    )
+    acq.run()
+    apt_news_contents = system.load_file(apt_news_file.destfile)
+    return json.loads(
+        apt_news_contents,
+        cls=util.DatetimeAwareJSONDecoder,
+    )
```
orndorffgrant, line 166: @julian-klode Here is where we use `apt_pkg.Acquire` to fetch the aptnews.json file.

Please LMK if I'm doing anything wrong with the API here.

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,3 +1,7 @@
...
+};
+
```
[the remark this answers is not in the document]
julian-klode: Well I guess that works, but it might be nicer to just use `fnctl.flock()` or `fcntl.lockf()` [that's the type of lock apt and dpkg use] in the script itself?

Might be worth instead of forking a background process here to `systemd-run` it `if [ -e /run/systemd/system ]` or stuff it into a .service you launch (with a timeout even).

That seems a bit more resilient and easier to investigate.

uaclient/apt_news.py
julian-klode: Looks ok if nothing else calls apt_pkg in the same process; otherwise you have to make sure you don't call init() if something else has written config (usually it's not called inside a function but at program entry).

apt-hook/20apt-esm-hook.conf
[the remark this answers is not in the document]
orndorffgrant: Thanks for the quick feedback! That all makes sense. I'll move it to a .service file and just call `systemctl start apt-news.service` here.

And yep good point It'll be easier to follow if the lock check is in the script itself. I'll make that change as well

uaclient/apt_news.py
orndorffgrant: Cool, and good catch, I'll move the init call up to the entry point function for this service.

apt-hook/20apt-esm-hook.conf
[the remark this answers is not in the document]
julian-klode: If you don't care about news inside docker and chroots (systems without systemd), you don't need to worry about locking at all if it's a service.

But I don't know if you do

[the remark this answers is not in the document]
orndorffgrant: Ah right! I think it's fair to not worry about this feature in non-systemd environments

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,3 +1,7 @@
+APT::Update::Pre-Invoke {
+	"[ ! -e /run/systemd/system ] || systemctl start apt-news.service || true";
+};
```
orndorffgrant, line 3: Updated based on @julian-klode's suggestions. The service file is here at [systemd/apt-news.service](https://github.com/canonical/ubuntu-advantage-client/pull/2338/files#diff-6ac1fa2387a109e054cc5986bba22c26e5d561717959597f4058bdd64af48b2c)

apt-hook/20apt-esm-hook.conf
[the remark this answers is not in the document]
orndorffgrant: Updated so moving the conversation [here](https://github.com/canonical/ubuntu-advantage-client/pull/2338#discussion_r1044786676) for any follow ups. Resolving this one.

features/apt_messages.feature
```diff
@@ -374,26 +331,295 @@ Feature: APT Messages
...
+        When I run `pro refresh messages` with sudo
+        When I run shell command `rm -f /var/lib/ubuntu-advantage/messages/apt-pre*` with sudo
+        When I run `apt upgrade` with sudo
+        Then I will see the following on stdout
+        """
+        Reading package lists...
+        Building dependency tree...
+        Reading state information...
+        Calculating upgrade...
+        0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```
lucasmoura, line 615: @orndorffgrant Sorry if I am missing something, but can we add a test for the case where we have the contract expired message as well

uaclient/apt_news.py
```diff
@@ -0,0 +1,180 @@
...
+    if selectors is None:
+        return True
+
+    if selectors.codenames is not None:
+        if system.get_platform_info()["series"] not in selectors.codenames:
+            return False
+
+    if selectors.clouds is not None:
+        cloud_id, fail = get_cloud_type()
+        if fail is not None:
```
lucasmoura, line 84: Should we log here that `cloud-id` failed somehow ?

features/apt_messages.feature
lucasmoura: Nevermind, you already have a PR for [that](https://github.com/canonical/ubuntu-advantage-client/pull/2339)

uaclient/apt_news.py
orndorffgrant: We could do that, though I suggest we put it inside the get_cloud_type function like this:

```diff
 @lru_cache(maxsize=None)
 @apply_config_settings_override("cloud_type")
 def get_cloud_type() -> Tuple[Optional[str], Optional[NoCloudTypeReason]]:
     if system.which("cloud-id"):
         # Present in cloud-init on >= Xenial
         try:
             out, _err = system.subp(["cloud-id"])
             return (out.strip(), None)
-        except exceptions.ProcessExecutionError:
+        except exceptions.ProcessExecutionError as e:
+            logging.debug("error running cloud-id: %s", str(e))
             return (None, NoCloudTypeReason.CLOUD_ID_ERROR)
     # If no cloud-id command, assume not on cloud
     return (None, NoCloudTypeReason.NO_CLOUD_DETECTED)
```

I can add that in an extra commit if you agree

lucasmoura: Fair enough, that's a better change for sure

apt-hook/20apt-esm-hook.conf
orndorffgrant: @basak does this mechanism look okay to you as well?

panlinux: Works with the `Type=exec` service and no timeout in the unit file. Maybe add a timeout to the python script itself, but consider if you want to exit non-zero if the timeout is hit or not, it will affect the systemd service unit state.

julian-klode: I remember needing to use --no-block otherwise it waits for one shot units to finish or something, but I don't know how it's solved here.

panlinux: It was changed to `Type=exec`. `oneshot` does indeed block.

lucasmoura (approved):

lucasmoura:
The failed integration test looks like a flake, but worth a double check just in case

apt-hook/20apt-esm-hook.conf
orndorffgrant: @panlinux So I noticed that the version of systemd on xenial doesn't recognize `Type=exec`. I think it falls back to "simple", but I decided to change the type back to "oneshot" (which is recognized back on xenial). I think "oneshot" most accurately describes the use of the service.

To avoid blocking on the oneshot service, I've added `--no-block` to the` systemctl start` command. It's working well in my tests.

uaclient/apt_news.py
orndorffgrant: added!

PR #2343, opened 2022-12-05
Do not import version on setup.py

renanrodrigo:
We rely on `version.py` (mainly on `get_version` to determine the version for the client for pretty much all situations, and this included `setup.py` of course.
The thing is: our package is installed on Ubuntu using apt, and all of the machinery is in place to properly version the package when building (whether locally or on PPAs) and report it appropriately.

The only tooling that relies on installing through setup.py is `tox`, for the tests, and we could not care less of what version it sees - currently it thinks it is running on 20.3 because of git tagging and runs fine.

The problem with `setup.py` importing from `version` brings the fact that `version` itself imports other modules, and:
a) There is no guarantee that those modules will not import version again, generating a circular dependency
b) There are guaranteed dependencies from the package perspective (yaml, apt) which break when importing in setup.py because no requirement installation was done yet, if any. 

This PR tells tox to go with `28.0.dev1` and do its thing, and removes the import from `setup.py`.

The visible side effects are: the ghost `20.3` version is not in the logs anymore, and we can finally import yaml in event_logger without being bothered by our own tooling.

## Test Steps
Tox and behave, as usual - and make sure building and testing it in containers still reports the correct version nonetheless.

## Desired commit type

## Checklist
 - [x] I have updated or added any "documentation" accordingly

 - [ ] Yes
 - [x] No - unless you think SRUers will be sad

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import glob
 
 import setuptools
 
-from uaclient import defaults, version
+from uaclient import defaults
 
 NAME = "ubuntu-advantage-tools"
 
@@ -30,14 +30,6 @@ def split_link_deps(reqs_filename):
 TEST_REQUIRES, TEST_LINKS = split_link_deps("test-requirements.txt")
 
 
-def _get_version():
-    parts = version.get_version().split("-")
-    if len(parts) == 1:
-        return parts[0]
-    major_minor, _subrev, _commitish = parts
-    return major_minor
-
-
 def _get_data_files():
     return [
         ("/etc/ubuntu-advantage", ["uaclient.conf", "help_data.yaml"]),
@@ -59,7 +51,9 @@ def _get_data_files():
 
 setuptools.setup(
     name=NAME,
-    version=_get_version(),
+    # This version does not matter, it is not used anywhere but in unit tests
+    # AND IT IS OVER 8000
+    version="8001",
     packages=setuptools.find_packages(
         exclude=[
             "*.testing",
--- a/uaclient/version.py
+++ b/uaclient/version.py
@@ -1,8 +1,5 @@
 """
-Version determination functions
-
-These are in their own file so they can be imported by setup.py before we have
-any of our dependencies installed.
+Client version related functions
 """
 import os.path
 import re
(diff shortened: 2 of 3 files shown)


uaclient/version.py
```diff
@@ -33,7 +30,8 @@ def get_version() -> str:
          XX.Y so return the --long version to allow daily build recipes
          to count commit offset from upstream's XX.Y signed tag.
       b. If run in a git-ubuntu pkg repo, upstream tags aren't visible,
-         parse the debian/changelog in that case
+         believe __VERSION__ is correct - there is and MUST always be a
```
CalvoM, line 33: I do not have much context for this, but would these tests be? How will they be implemented?

renanrodrigo: It already exists, the docstring there is out of date.
There is a CI step in the base workflow: 
https://github.com/canonical/ubuntu-advantage-client/blob/508db0f80664cb9dca78d7ea63e5736c004dc67f/.github/workflows/ci-base.yaml#L28-L29

which calls the script to check for this

setup.py
```diff
@@ -59,7 +51,7 @@ def _get_data_files():
 
 setuptools.setup(
     name=NAME,
-    version=_get_version(),
+    version="28.0.dev1",
```
lucasmoura, line 54: Should we consider using a dummy version here ? My concern is that if in the future we bump the version past 28, this might get misleading again as we have today with the `20.3`  version we see on the logs

orndorffgrant: +1 to more obviously fake version. Maybe something over 9000? `9001`? :upside_down_face:

renanrodrigo: The American translation is wrong - in Portuguese, as well as the original in Japanese, it is over 8000
I will set it to 8001

renanrodrigo:
@orndorffgrant @lucasmoura updated

setup.py
orndorffgrant: TIL!

setup.py
```diff
@@ -59,7 +51,8 @@ def _get_data_files():
 
 setuptools.setup(
     name=NAME,
-    version=_get_version(),
+    # There is no way that can be right
```
orndorffgrant, line 54: lol, this might seem a bit cryptic in the future - and might be intepreted as a "FIXME" comment. Maybe make it more clear that this is intentional

renanrodrigo: Sure, but then I lose the joke - it's Nappa's response in the scene.
Will add relevant information of course.

lucasmoura (approved):

PR #2346, opened 2022-12-07
cli: add magic-attach flow

lucasmoura:
## Proposed Commit Message
cli: add magic-attach flow on CLI

Add support for the magic-attach flow through the `pro attach` CLI. This flow allows users to attach to Ubuntu Pro subscription without a token. When running `pro attach` without providing a token, the Pro client will output a user code on the screen and the user will then login into the Pro portal url and input that token there. Once that is done, our code will detect that and fetch the token directly from the Contract Server. Once we have the token, we can than attach as usual.

PS: We are still waiting on some functionality to be delivered to make that flow useful

## Desired commit type

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/features/magic_attach.feature
+++ b/features/magic_attach.feature
@@ -1,0 +1,61 @@
+Feature: Magic attach flow related tests
+
+    @series.lts
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: Attach using the magic attach flow
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I change contract to staging with sudo
+        And I create the file `/tmp/response-overlay.json` with the following:
+        """
+        {
+            "https://contracts.staging.canonical.com/v1/magic-attach": [
+            {
+              "code": 200,
+              "response": {
+                "userCode": "123",
+                "token": "testToken",
+                "expires": "expire-date",
+                "expiresIn": 2000
+              }
+            },
+            {
+                "code": 200,
+                "response": {
+                    "userCode": "123",
+                    "token": "testToken",
+                    "expires": "expire-date",
+                    "expiresIn": 2000,
+                    "contractID": "test-contract-id",
+                    "contractToken": "$behave_var{contract_token_staging}"
+                }
+            }]
+        }
+        """
+        And I append the following on uaclient config:
+        """
+        features:
+          serviceclient_url_responses: "/tmp/response-overlay.json"
+        """
+        And I run `pro attach` with sudo
+        Then stdout matches regexp:
+        """
+        Initiating attach operation...
+
+        Please sign in to your Ubuntu Pro account at this link:
+        https://ubuntu.com/pro/attach
+        And provide the following code: .*123.*
+
+        Attaching the machine...
+        """
+        When I run `pro status --format yaml` with sudo
+        Then stdout matches regexp:
+        """
+        attached: true
+        """
+
+        Examples: ubuntu release
+            | release |
+            | xenial  |
+            | bionic  |
+            | focal   |
+            | jammy   |
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -39,6 +39,15 @@ from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
     FullAutoAttachOptions,
     _full_auto_attach,
 )
+from uaclient.api.u.pro.attach.magic.initiate.v1 import _initiate
+from uaclient.api.u.pro.attach.magic.revoke.v1 import (
+    MagicAttachRevokeOptions,
+    _revoke,
+)
+from uaclient.api.u.pro.attach.magic.wait.v1 import (
+    MagicAttachWaitOptions,
+    _wait,
+)
 from uaclient.api.u.pro.security.status.reboot_required.v1 import (
     _reboot_required,
 )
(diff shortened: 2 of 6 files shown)


orndorffgrant (commented):
I just remembered: I think we might want to avoid having a special sub command for this. Also I don't think the name "magic attach" is intended to be the public facing name of the function.
This is really a special method of attaching and so I think it shuold be integrated with the normal `pro attach` command. Could we default to this method if `pro attach` with no token is run? Or we could add a flag like `pro attach --with-browser`?

lucasmoura:
Good point @orndorffgrant. I think we can add it as a default flow for `pro attach` when the user doesn't provide a token.
We can also discuss if we should add support for `--format json` here. My opinion is that we shouldn't do that, as users that want json responses can use the API directly, but we can discuss this for sure

uaclient/cli.py
```diff
@@ -399,10 +408,20 @@ def config_parser(parser):
...
-        " from {}".format(defaults.BASE_UA_URL)
+        " from:\n{}".format(defaults.BASE_UA_URL)
+    )
+    parser.description = (
+        base_desc
+        + "\n\n"
+        + (
+            "When running this command without a token, we will use the "
+            "magic-attach\nflow to attach this machine into a Ubuntu Pro "
+            "subsctription without a token\nLearn more at: magic-attach-url"
```
orndorffgrant, line 423: I don't think "magic attach" should appear in any user-facing text. Here's a suggestion:
```suggestion
            "When running this command without a token, it will generate a short "
            "code and prompt you to attach the machine to your Ubuntu Pro account "
            "using a web browser."
```

uaclient/cli.py
```diff
@@ -1400,22 +1419,48 @@ def action_auto_attach(args, *, cfg: config.UAConfig) -> int:
         return 0
 
 
+def _magic_attach(args, *, cfg, **kwargs):
+    event.info("Initiating magic-attach flow...")
+    initiate_resp = _initiate(cfg=cfg)
+
+    event.info("Please sign in to advantage portal:")
+    event.info("https://ubuntu.com/pro")
```
orndorffgrant, line 1427: I think it might be "https://ubuntu.com/pro/attach"?

uaclient/cli.py
```diff
@@ -1400,22 +1419,48 @@ def action_auto_attach(args, *, cfg: config.UAConfig) -> int:
...
     if args.token and args.attach_config:
         raise exceptions.UserFacingError(
             msg=messages.ATTACH_TOKEN_ARG_XOR_CONFIG.msg,
             msg_code=messages.ATTACH_TOKEN_ARG_XOR_CONFIG.name,
         )
-
-    if args.token:
+    elif not args.token and not args.attach_config:
+        token = _magic_attach(args, cfg=cfg)
+        enable_services_override = None
```
orndorffgrant, line 1462: I think we should have a check here to disallow `--format json` for this flow

lucasmoura:
@orndorffgrant updated

uaclient/tests/test_cli_attach.py
```diff
@@ -13,17 +14,45 @@
...
+HELP_OUTPUT = textwrap.dedent(
+    """\
+usage: pro attach <token> [flags]
+
+Attach this machine to Ubuntu Pro with a token obtained from:
+https://ubuntu.com/pro
+
+When running this command without a token, it will generate a short code
+and prompt you to attach the machine to your Ubuntu Pro account using
+a web browser.
```
lucasmoura, line 40: @lechsandecki what do you think about this help text for the magic-attach flow ?

features/magic_attach.feature
```diff
@@ -0,0 +1,61 @@
...
+        And I run `pro attach` with sudo
+        Then stdout matches regexp:
+        """
+        Initiating attach operation...
+
+        Please sign in to advantage portal:
+        https://ubuntu.com/pro/attach
+        And provide the following user code: 123
+
+        Attaching the machine...
```
lucasmoura, line 48: @lechsandecki this will be the output of running `pro attach` without a token. Let us know if you think we can improve it

orndorffgrant: I just tested the flow on staging and it works great! After my experience though there was a significant delay between seeing the "Successfully Attached Machine" message in my browser and the cli actually attaching.

A delay is inevitable, but maybe we should just add an extra line to this output, something like:

```
        Initiating attach operation...

        Please sign in to advantage portal:
        https://ubuntu.com/pro/attach
        And provide the following user code: 123

        Waiting for you to enter the code on your account... There may be a delay of several seconds
```

And maybe the 3 lines requesting action ("Please sign in ..... the following user code: 123") should be in bold?

Just ideas to consider. None of this should block the feature IMO

lechsandecki:
looking great. were you able to test it end-to-end?
i think that Grant is right- the code should be bold

orndorffgrant:
@lechsandecki Yes I tested it end-to-end using the staging servers.

Do you mean that only the code itself (e.g. ABC123) should be bold? (and not the rest of the message)

lechsandecki:
thanks
yes, exactly. Do you feel that the whole message should be bold?

orndorffgrant:
@lechsandecki I don't have a strong opinion. I was just making sure I understood yours :)

I'll make the code bold and then we'll land this feature. It will be in the daily ppa by monday for you to test.

lechsandecki:
Sounds good, thanks!

orndorffgrant:
test fail appears to be a flake. merging.

orndorffgrant (approved):

PR #2348, opened 2022-12-07
more datetime json decoding

orndorffgrant:
machine-token: avoid extra json parsing step by recognizing datetime fields
* uses DatetimeDataValue in PublicMachineToken data object to avoid an extra json.loads (DatetimeDataValue didn't exist at the time that was written)

data_types: fix error classes to be proper UserFacingErrors
* while fixing test cases from these changes, I found out that IncorrectFieldType errors fail during our main_error_handler function because they don't have msg_code or additional_info defined.
* This commit is the simplest fix to this - maybe not the best. I want a common IncorrectTypeError that all the more specific errors inherit from, and can't find a better way to make the constructors call super nicely. LMK what you think

util: always parse datetimes in readurl
* uses JsonDatetimeAwareDecoder in util.readurl so we can always be sure datetimes are decoded later on.
* This removes the need for a few other instances of `parse_rfc3339_date`

## Test Steps

Everything should still work

## Desired commit type

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -268,7 +268,7 @@ create_public_machine_token_file() {
 from uaclient.files import MachineTokenFile
 machine_token_file = MachineTokenFile()
 content = machine_token_file.read()
-machine_token_file.write(content=content)
+machine_token_file.write(content)
 "
 }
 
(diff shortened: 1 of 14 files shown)


orndorffgrant:
fixing unit tests...

orndorffgrant:
This is less urgent than I thought it was. I'm setting this aside for now and will come back to it later.

lucasmoura:
@orndorffgrant I think we just need to make that change on postinst as well, since we have a `write` call there as well

orndorffgrant:
@lucasmoura thanks for pointing that out! just fixed and pushed, hopefully that will make CI happy

renanrodrigo (approved):

PR #2350, opened 2022-12-08
LHS nav bar now collapsed by default

s-makin:
## Proposed Commit Message
```
The navigation menu is now consistent with those docs
rendered via Discourse, where the default state is for the
Diataxis sections to be collapsed by default.

The spelling inconsistencies were my own fault when I 
created the landing pages, so I have taken the opportunity
to fix them. 

Finally, I added the CC EAL acronym to the front page list
of services, to be consistent with the other services that
already had their acronyms shown.
```

### DIFF
--- a/docs/explanations.rst
+++ b/docs/explanations.rst
@@ -1,15 +1,15 @@
-Ubuntu Pro Client explanations
-==============================
+Ubuntu Pro Client explanation
+*****************************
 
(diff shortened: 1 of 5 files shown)


orndorffgrant (approved):
Big fan of the collapsible sections in the sidebar +1

PR #2356, opened 2022-12-13
util: all file writing is now done atomically by renaming a temp file

orndorffgrant:
## Test Steps

All CI should still pass

## Desired commit type

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/system.py
+++ b/uaclient/system.py
@@ -4,6 +4,7 @@ import os
 import pathlib
 import re
 import subprocess
+import tempfile
 import time
 import uuid
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


uaclient/system.py
```diff
@@ -370,12 +371,22 @@ def write_file(filename: str, content: str, mode: int = 0o644) -> None:
...
+        logging.debug(
+            "Writing file %s atomically via tempfile %s", filename, tmpf.name
+        )
+        tmpf.write(content.encode("utf-8"))
+        tmpf.flush()
+        tmpf.close()
+        os.chmod(tmpf.name, mode)
+        os.makedirs(os.path.dirname(filename), exist_ok=True)
+        os.rename(tmpf.name, filename)
+    except Exception as e:
```
lucasmoura, line 386: Maybe this is an overkill, but should we add an unit test for this code ? Just for the case where the write part fails and we call the `flush` method.

But let me know what you think

lucasmoura (approved):

uaclient/system.py
orndorffgrant: Good idea, I'll add one

lucasmoura: Yes, I think that covers it perfectly

lucasmoura:
Merging it once CI is green

renanrodrigo (approved):
love it

PR #2357, opened 2022-12-14
apt_news: treat contract expiration notices as apt news overrides

orndorffgrant:
## Test Steps

Make sure the new contract expiration notices match what you expect.

## Desired commit type

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -32,7 +32,6 @@ CA_CERTIFICATES_FILE = "/usr/sbin/update-ca-certificates"
 APT_PROXY_CONF_FILE = "/etc/apt/apt.conf.d/90ubuntu-advantage-aptproxy"
 
 APT_UPDATE_SUCCESS_STAMP_PATH = "/var/lib/apt/periodic/update-success-stamp"
-APT_LISTS_PATH = "/var/lib/apt/lists"
 
 # Since we generally have a person at the command line prompt. Don't loop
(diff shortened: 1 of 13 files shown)


renanrodrigo (approved):
LGTM

lucasmoura (approved):

uaclient/apt.py
```diff
@@ -611,8 +610,6 @@ def get_apt_cache_time() -> Optional[float]:
     cache_time = None
     if os.path.exists(APT_UPDATE_SUCCESS_STAMP_PATH):
         cache_time = os.stat(APT_UPDATE_SUCCESS_STAMP_PATH).st_mtime
-    elif os.path.exists(APT_LISTS_PATH):
-        cache_time = os.stat(APT_LISTS_PATH).st_mtime
```
orndorffgrant, line 615: @renanrodrigo I had to make these changes because the apt list path gets updated before the apt-news script runs. So if for any reason we use it's timestamp, then apt-news will think that apt update was called <24h ago and end early.

LMK if that makes sense and if you think this change is okay.

renanrodrigo: I think this change is fine.
Lists were used here only because that's what ansible does (I stole the code from there), but ansible needs more general purposes things.
The stamp file should be enough because it is present in every Ubuntu > 16.04, which is what we need anyway.

renanrodrigo (approved):

PR #2358, opened 2022-12-15
wip: cpp hook using esm apt cache

orndorffgrant:
Note: this currently assumes #2357 will get merged first and we'll rebase on top of that

In theory once we rebase such that esm-apps is nonbeta and contract notices are in apt-news, the integration tests should pass without further changes.

## Desired commit type

## Checklist

 - [x] Yes
 - [ ] No


apt-hook/json-hook.cc
```diff
@@ -275,11 +361,15 @@ int run()
...
-            esm_message_file.close();
+        ESMUpdates esm_updates;
+        bool success = get_potential_esm_updates(esm_updates);
+        if (success) {
+            ESMContext esm_context = get_esm_context();
+            if (!esm_updates.infra_packages.empty()) {
+                print_esm_packages("esm-infra", esm_updates.infra_packages, esm_context);
+            } else if (!esm_updates.apps_packages.empty()) {
+                print_esm_packages("esm-apps", esm_updates.infra_packages, esm_context);
+            }
```
orndorffgrant, line 372: we might need a smarter "are we EOL" check here instead
we can parse distro-info to get that I think

orndorffgrant: from out of band conversation: a smarter eol check isn't needed here because the infra source is only added to the separate apt cache on eol machines

orndorffgrant:
closing this one because it has been brought into #2342

PR #2366, opened 2023-01-04
system: make tempfile on the dir as target file

lucasmoura:
## Proposed Commit Message
system: make tempfile on the dir as target file

When calling the write_file function, we write the content of the file first on a temp file moving it to target file. We are now creating the temp file on the same folder of the target file to avoid errors due to the files being on different filesystems

## Test Steps
Check if we can use the function to write a file on `/run` (This folder was the one that raised the issue being fixed)

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/system.py
+++ b/uaclient/system.py
@@ -373,7 +373,10 @@ def write_file(filename: str, content: str, mode: int = 0o644) -> None:
     """
     tmpf = None
     try:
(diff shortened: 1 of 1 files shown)


orndorffgrant (approved):
This worked for me! Let's just let CI pass

renanrodrigo (approved):

CalvoM (approved):

PR #2370, opened 2023-01-09
attach: support attaching without being able to install snapd (SC-1356)

renanrodrigo:
The side effect is that Livepatch will not be installed/enabled, but esm will work.
LP: #1997514

## Test Steps
Run the new integration test.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -672,6 +672,45 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
            | xenial  |
            | bionic  |
 
+    @series.xenial
+    @series.bionic
+    @uses.config.machine_type.lxd.vm
+    Scenario Outline: Attach works when snapd cannot be installed
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `apt-get remove -y snapd` with sudo
+        And I create the file `/etc/apt/preferences.d/no-snapd` with the following
+        """
+        Package: snapd
+        Pin: release o=*
+        Pin-Priority: -10
+        """
+        And I run `apt-get update` with sudo
+        When I attempt to attach `contract_token` with sudo
+        Then I will see the following on stderr:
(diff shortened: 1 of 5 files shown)


features/attached_enable.feature
```diff
@@ -672,6 +672,45 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
...
+        """
+        One moment, checking your subscription first
+        Installing snapd
+        Updating package lists
+        Failed to install snapd on the system
+        """
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
```
orndorffgrant, line 712: should we also run this on focal or jammy?

orndorffgrant (approved):
LGTM! Minor comments about test release coverage

features/attached_enable.feature
renanrodrigo: I ddin't add focal/jammy because livepatch is not supported for the kvm kernel there, and without your new changes it wouldn't work. Other livepatch tests in this file also rely only on Xenial/Bionic, and it covers well the bugfix imho. Do you think all LTS should be included?

features/attached_enable.feature
```diff
@@ -703,8 +742,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
            | release |
            | xenial  |
            | bionic  |
-           | focal   |
-           | jammy   |
```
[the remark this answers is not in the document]
renanrodrigo: Well Well they are not being executed anyway: the header says 
``` 
   @series.bionic
   @series.xenial
```

And then we fall to the same response I wrote above: should we adapt the test to cover all LTS?

features/attached_enable.feature
orndorffgrant: Good points. I think it is good as it is then!

PR #2371, opened 2023-01-09
tests: add first Lunar integration tests

renanrodrigo:
Also build the Lunar package on CI runs.
This way we can keep up with anything happening in builds, and have the option to test new things in the devel release.

## Test Steps
Wait until the moon is lit and shining on the dark sky, grab a lobster, and then run the added/modified integration tests.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/.github/workflows/ci-integration.yaml
+++ b/.github/workflows/ci-integration.yaml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        release: ['xenial', 'bionic', 'focal', 'jammy', 'kinetic']
+        release: ['xenial', 'bionic', 'focal', 'jammy', 'kinetic', 'lunar']
     steps:
       - name: Prepare build tools
         env:
--- a/.github/workflows/ci-integration.yaml
+++ b/.github/workflows/ci-integration.yaml
@@ -65,7 +65,7 @@ jobs:
       # as much information as possible from them.
       fail-fast: false
       matrix:
-        release: ['bionic', 'focal', 'jammy']
+        release: ['bionic', 'focal', 'jammy', 'kinetic']
         platform: ['lxd']
         host_os: ['ubuntu-22.04']
         include:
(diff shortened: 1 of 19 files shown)


.github/workflows/ci-integration.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        release: ['xenial', 'bionic', 'focal', 'jammy', 'kinetic']
+        release: ['xenial', 'bionic', 'focal', 'jammy', 'kinetic', 'lunar']
```
orndorffgrant, line 30: This builds the package for lunar as part of CI. Should we also run the lunar integration tests as CI?

orndorffgrant (approved):
Looks good! Just one question

.github/workflows/ci-integration.yaml
renanrodrigo: Hm, I would say no, and the reason is: we do not run kinetic (:
I will extend your question instead of answering: should we run any non-lts as CI?

orndorffgrant: Ahhh interesting!

I think we should, as we don't currently have any automated assurances that we aren't breaking non-lts during dev. Maybe let's only run the latest non-lts though?

Proposal: CI should cover lxd tests for:
* all supported LTS
* devel
  * if devel is LTS, then also the latest non-lts

renanrodrigo: one point I have against it is that devel is going to change ALOT

renanrodrigo: thinking twice after posting it:
- it will change a lot in LTSs, as services may change, not be ready yet, etc;

orndorffgrant: excellent points

Once we have daily/weekly jenkins runs, maybe a devel run would work better there.

And here for CI we do:
* all supported LTS
* latest supported non-LTS

does that match what you're thinking?

renanrodrigo: definitely. Will add kinetic integration here, and then we rotate it as new interim releases are, well, released.

renanrodrigo:
As discussed in a different channel, I just commented the Kinetic tests. If we have nothing on Kinetic anymore I will merge this.

PR #2372, opened 2023-01-10
chore: add script to test local changes in multipass

renanrodrigo:
Inspired by a change that @orndorffgrant has done in #2167, to run local changes in VMs, and moved by my struggle to make Xenial/Bionic VMs run.
Also, multipass ships the generic kernel as default so testing livepatch there may be important on some scenarios.

I didn't bother to change our script to accept more parameters and act accordingly, but instead copypastaed it...
If you think it is better to change `test-in-lxd.sh`  to accept multipass as a parameter, just close this PR and I'll make that instead.

### DIFF
--- a/tools/test-in-multipass.sh
+++ b/tools/test-in-multipass.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/bash
+series=$1
+
+set -x
+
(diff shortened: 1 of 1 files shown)


orndorffgrant (approved):
LGTM!

PR #2379, opened 2023-01-16
docs: Part 2 styleguide polishing

s-makin:
```
- Explanation/Reference landing pages have been revised/expanded.
- All Explanation/Reference pages have now been "polished".
- Headings in the nav bar have been shortened slightly.
- All pages specifically ordered (grouped by subject).
- Applied same standardisation as in Part 1 (#2375).
- Bullet lists with bolded initial text changed to subheadings to make better use of the RHS nav bar (in the Furo theme).

Continuing the changes made in Part 1 - Public-facing docs will now conform to
the Ubuntu Style Guide. I have also continued to make small adjustments (nits,
really) to language and grammar. The landing pages have had the biggest changes
since they now have introductions to the explanation/reference pages. 

Please let me know if these introductions are suitable -- as always, 
any problems please lmk. 
```


lucasmoura (approved):

lucasmoura:
@s-makin just a suggestion, but maybe we could move the new styleguide docs into `dev-docs` instead, since it is aimed for our team use. But let me know what you think

s-makin:
> @s-makin just a suggestion, but maybe we could move the new styleguide docs into `dev-docs` instead, since it is aimed for our team use. But let me know what you think

I put the file in the docs section while I was testing it (making sure the auto-header-referencing worked) so I could see the html preview. It's possible I maaaay have forgotten to move it afterwards. I'll fix it now!

docs/explanations.rst
```diff
@@ -5,11 +5,49 @@ Our explanatory and conceptual guides are written to provide a better
 understanding of how the Ubuntu Pro Client (``pro``) works. They enable you
 to expand your knowledge and become better at using and configuring ``pro``.
 
-Explanation
-===========
+Messaging
+=========
+
+Here you'll find details about APT and MOTD messages -- what they are, when
```
renanrodrigo, line 11: We could specify here that the section has information about the *pro client related* APT and MOTD messages - as those are wider concepts and there are messages there that we do not control

renanrodrigo (approved):
small suggestion, lgtm overall
thanks!

docs/explanations.rst
s-makin: That's a great point - changed it to reflect your suggestion!

PR #2380, opened 2023-01-16
Release 27.13

renanrodrigo:
New upstream release!

Carrying the GA for ESM Apps, but also fixes and fun and more.


debian/changelog
```diff
@@ -1,3 +1,53 @@
...
+    - apt-news:
+      + fetch and display APT News in apt upgrade
+      + show contract expiration notices in the apt news output
+    - attach: support attaching without being able to install snapd
+      (LP: #1997514)
+    - cli:
+      + do not show invalid subcommands in autocomplete (GH: #2279)
+      + add commands to support the magic-attach flow
+    - config: add apt_news_url option
+    - docs: reorganize documentation and correct information
```
lucasmoura, line 25: I think we can remove this changelog entry

debian/changelog
```diff
@@ -1,3 +1,53 @@
...
+        the lists for esm-infra and esm-apps
+    - apt-hook: update the cpp hook to use the local esm apt cache
+    - apt-news:
+      + fetch and display APT News in apt upgrade
+      + show contract expiration notices in the apt news output
+    - attach: support attaching without being able to install snapd
+      (LP: #1997514)
+    - cli:
+      + do not show invalid subcommands in autocomplete (GH: #2279)
+      + add commands to support the magic-attach flow
```
lucasmoura, line 23: It is not a new command per se, since it is part of `pro attach`. I think we can change it to `add support for attaching without a token`

lucasmoura (approved):

debian/changelog
renanrodrigo: I kept it there for the same reason we kept it before: it is a significant amount of commits docs-related, and Sally's work is focused here, so I'd like to report it

PR #2381, opened 2023-01-17
contract: send machine_type when getting resources

lucasmoura:
## Proposed Commit Message
contract: send machine_type when getting resources

When unattached, we send platform information to the contract server that replies with which services are available. One gap that we have is that some services are not available on all machine types. For example, Livepatch cannot be enabled on a container. We are now sending that information to the contract server so that they can take that information in consideration when informing which services are available.

## Test Steps
Verify if the integration tests for unattached scenarios are still working as expected

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/system.py
+++ b/uaclient/system.py
@@ -116,6 +116,15 @@ def get_dpkg_arch() -> str:
 
 
 @lru_cache(maxsize=None)
+def get_virt_type() -> str:
+    try:
+        out, _ = subp(["systemd-detect-virt"])
+        return out.strip()
+    except exceptions.ProcessExecutionError:
+        return ""
+
+
+@lru_cache(maxsize=None)
 def get_machine_id(cfg) -> str:
     """
     Get system's unique machine-id or create our own in data_dir.
@@ -181,8 +190,10 @@ def get_platform_info() -> Dict[str, str]:
             "series": series.lower(),
             "kernel": get_kernel_info().uname_release,
             "arch": get_dpkg_arch(),
+            "virt": get_virt_type(),
         }
     )
+
     return platform_info
 
 
--- a/docs/explanations/status_columns.md
+++ b/docs/explanations/status_columns.md
@@ -19,10 +19,10 @@ livepatch        yes        Canonical Livepatch service
 
 Where:
 
-* **SERVICE**: Is the name of service being offered.
-* **AVAILABLE**: Shows if that service is available on the machine. To verify
-  if a service is available, we check the machine kernel version, architecture
-  and the Ubuntu release it is running.
+* **SERVICE**: Is the name of service being offered
+* **AVAILABLE**: Shows if that service is available on that machine. To verify
+  if a service is available, we check the machine kernel version, architecture,
+  Ubuntu release being used and the machine type (i.e lxd for LXD containers)
 * **DESCRIPTION**: A short description of the service.
 
 ## With Pro subscription attached
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -358,6 +358,7 @@ class UAContractClient(serviceclient.UAServiceClient):
             "architecture": platform["arch"],
             "series": platform["series"],
             "kernel": platform["kernel"],
+            "virt": platform["virt"],
         }
 
     def _get_activity_info(self, machine_id: Optional[str] = None):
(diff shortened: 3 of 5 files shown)


uaclient/system.py
```diff
@@ -113,6 +113,12 @@ def get_dpkg_arch() -> str:
     return out.strip()
 
 
+@lru_cache(maxsize=None)
+def get_machine_type() -> str:
+    out, _ = subp(["systemd-detect-virt"])
+    return out.strip()
```
orndorffgrant, line 119: can we be sure this is always installed?
we might need something like
```suggestion
def get_machine_type() -> Optional[str]:
    try:
        out, _ = subp(["systemd-detect-virt"])
        return out.strip()
    except ProcessExecutionError:
        return None
```

lucasmoura: This is delivered by systemd, which should be around on all releases we support. But you are right, better be safe than sorry

lucasmoura: @orndorffgrant updated. I have used an empty string as a return to avoid changing the type check for the platform dict. But let me know if you disagree

orndorffgrant: Looks good! I'll just tag allan for his opinion on the default value in a new comment thread

uaclient/system.py
```diff
@@ -113,6 +113,15 @@ def get_dpkg_arch() -> str:
     return out.strip()
 
 
+@lru_cache(maxsize=None)
+def get_machine_type() -> str:
+    try:
+        out, _ = subp(["systemd-detect-virt"])
+        return out.strip()
+    except exceptions.ProcessExecutionError:
+        return ""
```
orndorffgrant, line 122: @alnvdl-work this is what we're planning on sending to the GetAvailableResources endpoint as `machine_type` - is an empty string a reasonable default value?

alnvdl-work: I think it's fine.
@TeodorPt Please state your opinion here :)

alnvdl-work: BTW, I didn't know we were going to do this change for the available resources calculation as well (I thought it was just the machine activity work).

I have a few concerns about how exactly we will categorize machine types, I think just virtualization type as returned by `systemd-detect-virt` is not enough for the long term. @TeodorPt will schedule a call for us to discuss this issue.

orndorffgrant: Excellent point - this is a separate but very similar issue to the machine activity and container detection work we talked about recently. We should meet and discuss if they can be consolidated or use the same/similar mechanisms/field names and such.

renanrodrigo (approved):

uaclient/system.py
```diff
@@ -181,8 +190,10 @@ def get_platform_info() -> Dict[str, str]:
             "series": series.lower(),
             "kernel": get_kernel_info().uname_release,
             "arch": get_dpkg_arch(),
+            "host_type": get_host_type(),
```
alnvdl-work, line 193: I'm having second thoughts. Perhaps we should call this `virt` or something like that? It would be weird to have "hostType: none".

alnvdl-work (approved):
LGTM, with one suggestion.

This will require some work in ua-contracts. When are we supposed to release that? (Sorry, I haven't been closely following this activity).

uaclient/system.py
orndorffgrant: `virt` makes sense to me

orndorffgrant:
@alnvdl-work is the work on ua-contracts needed _before_ we release this on the client (i.e. will it give an error on an expected field)? If not I think it's okay from our perspective to start sending this and ua-contracts can just ignore it for now.

alnvdl-work:
> @alnvdl-work is the work on ua-contracts needed _before_ we release this on the client (i.e. will it give an error on an expected field)? If not I think it's okay from our perspective to start sending this and ua-contracts can just ignore it for now.

@orndorffgrant Fields that we don't recognize are ignored. We can implement this later in ua-contracts.

uaclient/system.py
lucasmoura: Changing it to `virt` then

lucasmoura:
@orndorffgrant @alnvdl-work done

alnvdl-work (approved):
LGTM!

TeodorPt (approved):
LGTM!

orndorffgrant (approved):

PR #2382, opened 2023-01-19
dev-doc: add guide to permanently detach Pro instances (SC-1214)

aciba90:
## Proposed Commit Message

```
dev-doc: add guide to permanently detach Pro instances

Fixes: #2177
```

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/dev-docs/howtoguides/detach_pro_instances.md
+++ b/dev-docs/howtoguides/detach_pro_instances.md
@@ -1,0 +1,25 @@
+# How to permanently detach Pro instances
+
+## TL;DR
+
+1. Modify the client configuration file, normally `/etc/ubuntu-advantage/uaclient.conf`,
+to contain:
+
+    ```yaml
+    features:
+      disable_auto_attach: true
+    ```
+
+2. Perform a `sudo pro detach --assume-yes`.
+
+## Explanation
+
+On Pro instances, a `pro detach` won't permanently detach them as,
+the instance will be reauto-attached on the next boot (on non GCE instances)
+or immediately (on GCE instances due to the daemon).
+
+The config in step 1 will prevent the [daemon](../../systemd/ubuntu-advantage.service)
+and the a [service](../../systemd/ua-auto-attach.service) at next boot to auto-reattach.
+
+If you want to allow the instance to reauto attach by itself, then remove or set to false
+`disable_auto_attach` in the configuration file.


lucasmoura:
I am just wondering if this should on our public docs instead. What do you think @orndorffgrant @renanrodrigo @CalvoM ?

renanrodrigo:
I think it should be dev-docs, because it is a workaround for internal uses, but I can be convinced either way

s-makin:
> I am just wondering if this should on our public docs instead. What do you think @orndorffgrant @renanrodrigo @CalvoM ?

We don't currently have anything in the public docs about users even having the option to permanently detach (iirc), so if we want to make it public-facing it'll need a bit more info/context. What we could do is put this in the devdocs as is for now, and then if we start getting queries about how to permanently detach, we could move it to somewhere public-facing? 

My main question (apologies if it's a stupid one): is this something we expect users will want to do? If so, should it be implemented as a feature rather than a workaround (and then we document the feature)?

aciba90:
> > I am just wondering if this should on our public docs instead. What do you think @orndorffgrant @renanrodrigo @CalvoM ?
> 
> We don't currently have anything in the public docs about users even having the option to permanently detach (iirc), so if we want to make it public-facing it'll need a bit more info/context. What we could do is put this in the devdocs as is for now, and then if we start getting queries about how to permanently detach, we could move it to somewhere public-facing?

I like that approach.

> My main question (apologies if it's a stupid one): is this something we expect users will want to do? If so, should it be implemented as a feature rather than a workaround (and then we document the feature)?

To detach Ubuntu Pro instances is discouraged, and we did not find any valid use case for apart from testing. That's why we have decided to only document it, and if a use case arises then we will implement it as a feature.

s-makin:
> To detach Ubuntu Pro instances is discouraged, and we did not find any valid use case for apart from testing. That's why we have decided to only document it, and if a use case arises then we will implement it as a feature.

That makes sense - given that context, I think if we put this workaround in the public-facing docs, we'd need to think carefully about what message that sends (i.e. that we somehow endorse/encourage doing permanent detaches). If we *did* put it in the public docs, I think it should come with an advisory note to warn against doing it (and what the consequences are if you *do* do it, so that anyone who tries does so from an informed position).

s-makin (approved):

PR #2383, opened 2023-01-19
fix: add support for attach without a token

lucasmoura:
## Proposed Commit Message
fix: add support for attach without a token

If during the fix operation we identify that we need a Pro service to install the package fix, we can now allow user to attach without a token.

## Test Steps
On an unattached xenial machine, try running:

$ pro fix USN-5079-2

And type the `s` option when prompted for attach.
After that, make sure that the attach flow work as expected.

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -9,6 +9,15 @@ from datetime import datetime
 from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple
 
 from uaclient import apt, exceptions, messages, serviceclient, system, util
+from uaclient.api.u.pro.attach.magic.initiate.v1 import _initiate
+from uaclient.api.u.pro.attach.magic.revoke.v1 import (
+    MagicAttachRevokeOptions,
+    _revoke,
+)
+from uaclient.api.u.pro.attach.magic.wait.v1 import (
+    MagicAttachWaitOptions,
+    _wait,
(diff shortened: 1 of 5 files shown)


uaclient/security.py
```diff
@@ -1138,15 +1174,13 @@ def _prompt_for_attach(cfg: UAConfig) -> bool:
...
         valid_choices=["s", "a", "c"],
     )
     if choice == "c":
         return False
     if choice == "s":
-        print(messages.PROMPT_UA_SUBSCRIPTION_URL)
-        # TODO(GH: #1413: magic subscription attach)
-        input("Hit [Enter] when subscription is complete.")
+        return _perform_magic_attach(cfg)
     if choice in ("a", "s"):
```
renanrodrigo, line 1184: I guess "s" can be removed here since now it returns early - we can make it explicit.

renanrodrigo (approved):
I believe it's not worth it writing an integration test for this right - then LGTM

uaclient/security.py
lucasmoura: @renanrodrigo done

orndorffgrant (approved):
Works great!

orndorffgrant:
CI failures appear unrelated - merging

PR #2389, opened 2023-01-20
fix: make sure the local apt cache is isolated on update

renanrodrigo:
This isolates the local apt esm cache in the function where it is updated.
There is no need to change it in the other places it is instantiated (security-status, stats hook) because there we only read the package names.
"Acquire" related vars are not cleared, so we can use the same download machinery as the system apt does.

This clears the concerns we had about our internal Cache writing/triggering things it should not.
- logs are dependent on the 'Dir' config entry, so we never log to `/var/log/apt` at all
- Hooks are removed when the config is clear and reinitialized, so none of them are triggered when our update runs
- Acquire related configuration defines how apt downloads things, and we need them preserved to make sure we can get our files the same way as the system is configured to do. Looking at the configuration defined under `Acquire::*`, there is no clear sign that any of that will interfere with the system apt mechanisms.
- Do we really need everything under Acquire? Probably not. But having it all is consistent and it is a fair change compared to inspecting each key in each level individually.

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -603,6 +603,7 @@ def update_esm_caches(cfg) -> None:
         return
 
     import apt  # type: ignore
+    import apt_pkg  # type: ignore
 
     from uaclient.entitlements.entitlement_status import ApplicationStatus
     from uaclient.entitlements.esm import (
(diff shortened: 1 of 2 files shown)


julian-klode:
This looks good to me

julian-klode (approved):

basak (approved):
I had assumed in the original design that the apt environment used to acquire ESM statistics when the user has not explicitly opted in would be isolated from the system apt configuration. My understanding is that this is what this PR achieves in the case that the cache is being updated, and this avoids the double hook calls and logging to "system apt", so +1.

The caveat is that I think that this kind of thing should take place in all places this non-system apt cache is used, but I think we agreed that doing it only in the read/write case would do for now, and it'd be tech debt to cover it in the read-only places later. Either way, that means I think this is acceptable as-is for now.

This is just a review in principle - I leave it to Julian to confirm that the actual code changes do as we expect.

lucasmoura (approved):
I have tested both `apt update` and `pro security-status` with `execsnoop-bpfcc` and I have found:

* `apt update`: It seems that is not calling the update command twice, as I can now only see one call to create the `update-success-stamp` file.
* `pro security-status`: I can confirm that is not spawning any unwanted process or accessing any unwanted file or accessing apt files twice

PR #2396, opened 2023-01-27
Add check to not prompt user on apt_news changes

lucasmoura:
## Proposed Commit Message
Add check to not prompt user on apt_news changes

In the previous release we have changed the uaclient.conf file to capitalized a word. If the user has changed apt_news through pro config, the user will now get a debconf prompt because of it. This is not ideal, since we are the ones that instruct the user to run that command if they don't want apt advertisement on their system. This PR is creating a mechanism to avoid that prompt if the has just changed the apt_news config

## Test Steps
Launch a lxd machine with an older version of the pro client (27.12) and runs either `pro config set apt_news=false` or `pro config set apt_news=true` and verify that upgrading to a package with those fixes applied do not generate a debconf prompt and keep the `apt_news` setting for the user

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -382,6 +382,23 @@ case "$1" in
       if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "27.13~"; then
           cleanup_apt_news_flag_file
       fi
+
+      
+      # LP: #2003977
+      # The version gate is open ended, and should be closed when the apt_news
+      # configuration is moved away from a conffile.
+      if dpkg --compare-versions "$2" ge 27.11.3~; then
+          if [ -f /var/lib/ubuntu-advantage/preinst-detected-apt-news-disabled ]; then
+              # The preinst has left us a note to tell us that apt-news was
+              # previously disabled and should be re-disabled, so disable it
+              # and discard the note.
+              pro config set apt_news=false
+              rm /var/lib/ubuntu-advantage/preinst-detected-apt-news-disabled
+          fi
+          # Remove the conffile backup if it exists now that an error unwind is
+          # no longer possible
+          rm -f /etc/ubuntu-advantage/uaclient.conf.preinst-remove
+      fi
       ;;
 esac
 
--- a/debian/ubuntu-advantage-tools.postrm
+++ b/debian/ubuntu-advantage-tools.postrm
@@ -30,6 +30,15 @@ case "$1" in
         remove_logs
         remove_gpg_files
         ;;
+    abort-install|abort-upgrade)
+        # LP: #2003977
+        # The version gate is open ended, and should be closed when the
+        # apt_news configuration is moved away from a conffile.
+        if dpkg --compare-versions "$2" ge 27.11.3~; then
+            rm -f /var/lib/ubuntu-advantage/preinst-detected-apt-news-disabled
+            [ -f /etc/ubuntu-advantage/uaclient.conf.preinst-remove ] && mv /etc/ubuntu-advantage/uaclient.conf.preinst-remove /etc/ubuntu-advantage/uaclient.conf
+        fi
+        ;;
 esac
 
 #DEBHELPER#
--- a/debian/ubuntu-advantage-tools.preinst
+++ b/debian/ubuntu-advantage-tools.preinst
@@ -11,11 +11,62 @@ remove_old_config_fields() {
     fi
 }
 
+restore_previous_conffile() {
+    # Back up existing conffile in case of an error unwind
+    cp -a /etc/ubuntu-advantage/uaclient.conf /etc/ubuntu-advantage/uaclient.conf.preinst-remove
+    # Restore the default conffile that shipped with 27.11.3 through 27.12
+    cat > /etc/ubuntu-advantage/uaclient.conf <<EOT
+# Ubuntu Pro Client config file.
+# If you modify this file, run "pro refresh config" to ensure changes are
+# picked up by Ubuntu Pro Client.
+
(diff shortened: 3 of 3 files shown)


debian/ubuntu-advantage-tools.preinst
```diff
@@ -11,13 +11,64 @@ remove_old_config_fields() {
...
+                touch /var/lib/ubuntu-advantage/preinst-detected-apt-news-disabled
+                restore_previous_conffile
+            elif [ "$conffile_hash" = "3b01d7406cbb4ba628a9ffa57485d324" -o "$conffile_hash" = "d9971401a6409032b1c9069236040dc4" ]; then
+                # User had run "pro config set apt_news=true" with no other
+                # conffile changes
+                restore_previous_conffile
+            fi
+        fi
+    fi
+    ;;
```
orndorffgrant, line 71: @lucasmoura I think this should be indented more

debian/ubuntu-advantage-tools.postrm
```diff
@@ -30,6 +30,15 @@ case "$1" in
...
         ;;
+    abort-install|abort-upgrade)
+        # LP: #2003977
+        # The version gate is open ended, and should be closed when the
+        # apt_news configuration is moved away from a conffile.
+        if dpkg --compare-versions "$2" ge 27.11.3; then
+            rm -f /var/lib/ubuntu-advantage/preinst-detected-apt-news-disabled
+            [ -f /etc/ubuntu-advantage/uaclient.conf.preinst-remove ] && mv /etc/ubuntu-advantage/uaclient.conf.preinst-remove /etc/ubuntu-advantage/uaclient.conf
+        fi
+    ;;
```
orndorffgrant, line 41:
```suggestion
        ;;
```

lucasmoura:
@orndorffgrant done

debian/ubuntu-advantage-tools.postinst
```diff
@@ -382,6 +382,23 @@ case "$1" in
       if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "27.13~"; then
           cleanup_apt_news_flag_file
       fi
+
+      
+      # LP: #2003977
+      # The version gate is open ended, and should be closed when the apt_news
+      # configuration is moved away from a conffile.
+      if dpkg --compare-versions "$2" ge 27.11.3; then
```
panlinux, line 390: I think this needs to be compared to `27.11.3~`, since all u-a-t 27.11.3 releases are of the form `27.11.3~YY.MM.N`. Let's see if others agree.

lucasmoura: @panlinux fixed

lucasmoura:
CI errors are unrelated to this change and are being fixed here:
https://github.com/canonical/ubuntu-advantage-client/pull/2398

debian/ubuntu-advantage-tools.preinst
```diff
@@ -11,12 +11,63 @@ remove_old_config_fields() {
...
             remove_old_config_fields "$PREVIOUS_PKG_VER"
         fi
+    
+        # LP: #2003977
+        # If the user used "pro config set apt_news=false|true" previously,
+        # then we don't want a conffile prompt if they haven't otherwise
+        # changed the conffile. In these two cases, restore the conffile back
+        # to how it shipped, and then fix up in postinst if required. The
+        # version gate is open ended, and should be closed when the apt_news
+        # configuration is moved away from a conffile.
```
vorlonofportland, line 54: Has this been tested for the case when the user has already upgraded to the version of the package currently in -updates, and rejected the conffile update?  In that case, the "expected" conffile hash will not match that from 27.11.3, and resetting the file to the previous state will not clear the prompt the *next* time the conffile changes; you would want to reset the file to the pristine contents of the 27.13.1 version.

debian/ubuntu-advantage-tools.preinst
```diff
@@ -11,12 +11,63 @@ remove_old_config_fields() {
...
+        if dpkg --compare-versions "$2" ge 27.11.3~; then
+            if [ -f /etc/ubuntu-advantage/uaclient.conf ]; then
+                conffile_hash=$(md5sum /etc/ubuntu-advantage/uaclient.conf|awk '{print $1}')
+                if [ "$conffile_hash" = "038902993a843cac6cbe3efa4d1fcb92" -o "$conffile_hash" = "664dff27e212a77aef514e4b64" ]; then
```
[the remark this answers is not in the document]
basak: +1; @lucasmoura feel free to adjust this. I think the existing test plan will cover changes in this area.

debian/ubuntu-advantage-tools.preinst
lucasmoura: Yes, we understand this issue and we will address this in the next release to check the md5sum for the one we deliver for 27.13.1 to prevent exactly that. But we understand it should not be a blocker for this release. But let us know if that assumption is wrong

basak: I'm aware that `restore_previous_conffile` will need adjusting in the subsequent upload after this one to restore the more recent conffile. Is this what you mean? I didn't implement this initially as it's tricky to test, but have tried to make it clear in surrounding docs (in my https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/2003977/comments/5) that will be a hard requirement to adjust `preinst` to restore the correct conffile based on `$2` in the next upload after this.

debian/ubuntu-advantage-tools.preinst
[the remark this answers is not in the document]
orndorffgrant: updated

vorlonofportland (approved):

PR #2397, opened 2023-01-27
fix: only run the services in the apt hook when root

renanrodrigo:
There is a corner case happening in other people's test suites when updating the apt cache as non-root presents errors because systemd can't trigger the services we inserted in our hook.

This guarantees we only try to run stuff when we are root, which is the only case we are interested in.

## Test Steps
Added a script to the `sru/` folder. 
Before merging, use the script as a base for a manual test.
For staging/proposed it is a matter of adapting the script when running.

### DIFF
--- a/sru/release-27.13.2/test-apt-hook-nonroot.sh
+++ b/sru/release-27.13.2/test-apt-hook-nonroot.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+set -e
+
+series=$1
+name=$series-dev
+
+
+function cleanup {
+  lxc delete $name --force
+}
+
+function on_err {
+  echo -e "Test Failed"
+  cleanup
+  exit 1
+}
+
+trap on_err ERR
+
+lxc launch ubuntu-daily:$series $name
+sleep 5
+
+# Check ubuntu-advantage-tools version
+lxc exec $name -- apt-get update > /dev/null
+lxc exec $name -- apt-get install ubuntu-advantage-tools > /dev/null
+echo -e "\n* UA is still the old one"
+echo "###########################################"
+lxc exec $name -- apt-cache policy ubuntu-advantage-tools
+echo -e "###########################################\n"
+
+# Move python snippet to SUT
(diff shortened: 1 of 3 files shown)


sru/release-27.13.2/test-apt-hook-nonroot.sh
```diff
@@ -0,0 +1,61 @@
...
+lxc exec $name -- apt-get install ubuntu-advantage-tools > /dev/null
+echo -e "\n* Upgrading UA to new version"
+echo "###########################################"
+lxc exec $name -- apt-cache policy ubuntu-advantage-tools
+echo -e "###########################################\n"
+
+# Try to run it, see it passes
+echo -e "\n* Success when trying to run the python snippet"
+echo "###########################################"
+lxc exec $name --user 1000 -- python3 /home/ubuntu/python-apt-snippet.py
```
lucasmoura, line 58: Should we also run that command on the old client to show that the error is indeed happening ?

julian-klode (approved):
I don't think these are blockers, but here are some shellcheck complaints:

```
In - line 1:
[ ! -e /run/systemd/system -o $(id -u) -ne 0 ] || systemctl start --no-block apt-news.service esm-cache.service || true
^-- SC2148 (error): Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.
                           ^-- SC2166 (warning): Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.
                              ^------^ SC2046 (warning): Quote this to prevent word splitting.

```

renanrodrigo:
sure, /me is looking

orndorffgrant (approved):
I think its probably worth fixing the shellcheck complaints, but I've tested this and it's working well on my end

renanrodrigo:
So, I fixed the `SC2166` one.
`SC2148` was already there before so I'll keep it for now and we'll see.
`SC2046` wasn't fixed even throwing the backslash fiesta, and it will not bite us because `id -u` should be a number, word splitting doesn't apply.

renanrodrigo:
The outage kills our integration tests :D

PR #2401, opened 2023-01-30
migrate user config

orndorffgrant:
This is the PR that moves apt_news and similar config items (those that are intended to be managed via the `pro config` cli) out of the conffile.

This doesn't go so far as to rm_conffile uaclient.conf as some users may still expect it to be there.

There are several little changes in separate commits that are prereqs for the main migration commit.

See the comment in preinst and the migrate_user_config.py script for some more details.

Then there are separate commits with a test script to be used at SRU time of this migration, and of docs that will contain what to do if one of the unlikely errors occurs.

Also notice that we maintain backwards compatibility so even if migrations fail or if a user manually edits uaclient.conf later, we'll still respect `ua_config` settings in uaclient.conf. This is required because at least one other tool relies on this (cloud-init :eyes:)

All feedback welcome.

LP #2004280

To Do:
- [x] support old ua_config at runtime to support https://cloudinit.readthedocs.io/en/latest/reference/modules.html#ubuntu-advantage
    - medium term: change how this works in cloud-init (and optionally remove our support?)
    - medium term: add integration tests for all cloud-init pro options - SC-1427
- [x] fix remaining CI failures
- [x] search for `ua_config` and fix everywhere we use it to the new solution
- [x] add LP reference to main commit
- [x] add explanation doc with the contents of possible error messages and what to do about them
- [x] add additional unit tests for any of the small PRs?
- [x] additional testing of failure cases
- [x] add test script for SRU verification
- [x] integration test for still supporting ua_config in uaclient.conf
- [x] update version guards to 27.14~

## Test Steps

Try all kinds of upgrade scenarios you can think of: try various edits to uaclient.conf and then upgrade to this version. Does the UX make sense? Do conffile prompts only happen in reasonable scenarios?

Maybe also hack preinst to add failures in the python script and see what happens - is it good enough?

## Checklist

 - [x] Yes
 - [ ] No


lib/migrate_user_config.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+"""
+This will run in preinst when upgrading from any version prior to 27.13.3.
+It removes the "ua_config" sub-document from uaclient.conf and moves it to it's own json file in /var/lib.
+The fields under "ua_config" are intended to be set/unset by the user, so we don't want them coverred as a "conffile".
```
lucasmoura, line 5: s/coverred/covered

lib/migrate_user_config.py
```diff
@@ -0,0 +1,82 @@
...
+        ),
+        "security_url": old_uaclient_conf.get(
+            "security_url", defaults.BASE_SECURITY_URL
+        ),
+        "log_level": old_uaclient_conf.get("log_level", "debug"),
+    }
+    # the rest only if they're different from defaults
+    for field in ("data_dir", "log_file", "timer_log_file", "daemon_log_file"):
+        old_val = old_uaclient_conf.get(field)
+        if old_val != defaults.CONFIG_DEFAULTS.get(field):
```
lucasmoura, line 67: Why do we want to only add those fields if they have been changed ? If they don't have that distinction, we could just loop on the `old_uaclient_conf` keys (besides `ua_config`) and do a get where the default value comes from `defaults.CONFIG_DEFAULTS`.

lucasmoura: Okay, just notice that we are removing those fields from the config. But wouldn't this also hide to the user that those fields can be changed in the first place ? Maybe that is what we want, but I just want to double check

uaclient/config.py
```diff
@@ -121,6 +122,14 @@ def __init__(
             self.cfg_path = get_config_path()
             self.cfg, self.invalid_keys = parse_config(self.cfg_path)
 
+        if user_config:
+            self.user_config = user_config
+        else:
+            self.user_config = (
+                state_files.user_config_file.read()
+                or state_files.UserConfigData()
```
lucasmoura, line 130: Should we support this scenario where we the config json file doesn't exist ? We will create user config with all `None` values which can cause errors in down the line or even worse, cause an unexpected result  (i.e. the user relying on a proxy for example)

lib/migrate_user_config.py
```diff
@@ -0,0 +1,82 @@
...
+        old_uaclient_conf = yaml.safe_load(uaclient_conf_file)
+
+    new_uaclient_conf = {
+        "contract_url": old_uaclient_conf.get(
+            "contract_url", defaults.BASE_CONTRACT_URL
+        ),
+        "security_url": old_uaclient_conf.get(
+            "security_url", defaults.BASE_SECURITY_URL
+        ),
+        "log_level": old_uaclient_conf.get("log_level", "debug"),
```
CalvoM, line 62: Quick  question, considering we transform the log_level to upper in [config](https://github.com/canonical/ubuntu-advantage-client/blob/12c12d0c19bf55cd5f0f48148a7cdd8b1dfb8a3b/uaclient/config.py#L363), shouldn't we not then default to capitalized version of `debug` as well?

lib/migrate_user_config.py
orndorffgrant: Right, my suggestion is that we reduce the set of fields in the default uaclient.conf because we don't want to encourage users to change them. They're mostly for testing anyway.

We should talk as a team about this though. We should also consider the rm_conffile approach as well.

uaclient/config.py
orndorffgrant: Most of the getters on UAConfig for these fields return `Optional` already so the type signature won't change and in theory mypy has checked that we don't assume they're not None anywhere.
For the rest, we'll implement default values in UAConfig.
Do you think this is sufficient?

lib/migrate_user_config.py
orndorffgrant: Alternatively: because we transform to upper in config, why do we need to capitalize the value in the file?

IMO: it doesn't matter, but the current default is `debug` and lowercase is easier to type :upside_down_face: so I'd prefer to keep it lowercase here.

uaclient/config.py
lucasmoura: My only concern here is the corner case where the user had a valid config that was corrupted somehow and know we will be running commands with default values. For example, if the user has a proxy configured in the old configuration.

I know this should be really rare, but I think we should alert the user somehow that this is happening

debian/ubuntu-advantage-tools.preinst
```diff
@@ -2,100 +2,143 @@
...
+        # This removes the "ua_config" sub-document from uaclient.conf and
+        # moves it to it's own json file in /var/lib/ubuntu-advantage/user-config.json.
+        # The fields under "ua_config" are intended to be set/unset by the
+        # user via the `pro config set` command, so we don't want them covered as a "conffile".
+        # It writes back a uaclient.conf that will match the new default
+        # uaclient.conf as long as no defaults have changed - avoiding
+        # an unnecessary conffile prompt.
+        # We print warning messages if errors occur with instructions to help the user resolve them.
+        /usr/bin/python3 -c "
+import json
```
renanrodrigo, line 29: I think this fits better in a separate file, and then we just call it here?

orndorffgrant: I would love that!
AFAICT it's not possible since we have to do this in preinst - but I would really like to be wrong here.

lucasmoura: @orndorffgrant This need to be in preinst to avoid the debconf prompt right ? If yes, what if on preinst we just create a backup of the conf file and bring back the default value. Then, on postinst we can perform the transition and move that logic into a separate module

uaclient/config.py
```diff
@@ -121,6 +123,26 @@ def __init__(
             self.cfg_path = get_config_path()
             self.cfg, self.invalid_keys = parse_config(self.cfg_path)
 
+        if user_config:
+            self.user_config = user_config
+        else:
+            try:
+                self.user_config = (
+                    state_files.user_config_file.read()
+                    or state_files.UserConfigData()
```
lucasmoura, line 132: I think we can remove the `or` clause here

debian/ubuntu-advantage-tools.preinst
orndorffgrant: That's a great idea! Yes I'll do that

uaclient/config.py
orndorffgrant: `read` could return `None`, so this `or` would give us the default data object instead of None. This makes it so we don't need to check for None in all of the property functions below

orndorffgrant:
I'll rebase this on `main`

uaclient/config.py
lucasmoura: Fair enough, the `None` case is when the file doesn't exist. My mistake

lib/migrate_user_config.py
```diff
@@ -0,0 +1,166 @@
...
+        "ua_apt_http_proxy": None,
+        "ua_apt_https_proxy": None,
+        "global_apt_http_proxy": None,
+        "global_apt_https_proxy": None,
+        "update_messaging_timer": 21600,
+        "metering_timer": 14400,
+        "apt_news": True,
+        "apt_news_url": "https://motd.ubuntu.com/aptnews.json",
+    }
+    for field in (
```
lucasmoura, line 61: I think we can loop on `USER_CONFIG_DEFAULT.keys()` instead

lib/migrate_user_config.py
```diff
@@ -0,0 +1,166 @@
...
+        "daemon_log_file",
+    ):
+        old_val = old_uaclient_conf.get(field)
+        if old_val is not None and old_val != defaults.CONFIG_DEFAULTS.get(
+            field
+        ):
+            new_uaclient_conf[field] = old_val
+
+    try:
+        print(messages.USER_CONFIG_MIGRATION_MIGRATING, file=sys.stderr)
```
lucasmoura, line 132: Should this message be on `stderr` ?

sru/release-27.14/test-migrate-user-config.sh
```diff
@@ -0,0 +1,206 @@
...
+    lxc launch -q ubuntu-daily:$series $name
+    sleep 5
+    lxc exec $name -- apt-get update > /dev/null
+    lxc exec $name -- apt-get install debsums -y > /dev/null
+    echo "-------------------------------------------"
+
+    install_old_version $name $series $old_version
+
+    echo -e "\n-------------------------------------------"
+    echo "** make changes to uaclient.conf root"
```
lucasmoura, line 156: Should we have more changes to `uaclient.conf` here ? For example, adding the `features` and `settings_overrides` here ?

lib/migrate_user_config.py
```diff
@@ -0,0 +1,166 @@
...
+        "apt_news",
+        "apt_news_url",
+    ):
+        old_val = old_user_config.get(field)
+        if old_val is not None and old_val != USER_CONFIG_DEFAULTS.get(field):
+            new_user_config[field] = old_val
+
+    try:
+        with open(
+            "/var/lib/ubuntu-advantage/user-config.json", "w"
```
lucasmoura, line 81: I think we should change this path definition to the `defaults` module

sru/release-27.14/test-migrate-user-config.sh
```diff
@@ -0,0 +1,206 @@
...
+    lxc exec $name -- pro config show
+    echo "-------------------------------------------"
+
+    lxc delete --force $name
+    echo "###########################################"
+}
+
+
+# xenial
+test_normal_upgrade         xenial 27.11.3~16.04.1
```
lucasmoura, line 190: On each test, should we also show the output of  `cat /var/lib/ubuntu-advantage/user-config.json`  ? II know we are already doing that through `pro config show`, but is just to make it explicit that the file doesn't contain any extra information we don't need.

Additionally, should we check that we have removed `/etc/ubuntu-advantage/uaclient.conf.preinst-backup` in the tests that modify `uaclient.conf` ?

sru/release-27.14/test-migrate-user-config.sh
```diff
@@ -0,0 +1,206 @@
...
+test_apt_news_false_upgrade  xenial 27.11.3~16.04.1
+test_apt_news_false_upgrade  xenial 27.12~16.04.1
+test_apt_news_false_upgrade  xenial 27.13.1~16.04.1
+test_apt_news_false_upgrade  xenial 27.13.2~16.04.1
+test_apt_news_false_upgrade  xenial 27.13.3~16.04.1
+test_uaclient_conf_changes_upgrade  xenial 27.11.3~16.04.1
+test_uaclient_conf_changes_upgrade  xenial 27.12~16.04.1
+test_uaclient_conf_changes_upgrade  xenial 27.13.1~16.04.1
+test_uaclient_conf_changes_upgrade  xenial 27.13.2~16.04.1
+test_uaclient_conf_changes_upgrade  xenial 27.13.3~16.04.1
```
lucasmoura, line 204: Can we add a test where we mess up the file permissions of `uaclient.conf` just to see the error message that will be printed during postinst ?

lib/migrate_user_config.py
orndorffgrant: I think so, that's where we print some other informational messages in postinst, but I'm not 100% sure

sru/release-27.14/test-migrate-user-config.sh
orndorffgrant: yep that's a good idea!

sru/release-27.14/test-migrate-user-config.sh
orndorffgrant: Yep, both are great ideas, I'll add them

sru/release-27.14/test-migrate-user-config.sh
orndorffgrant: Since the migration is run as root, even 000 permissions won't cause the error.

I was able to write a test that causes the error though: by making the uaclient.conf file invalid yaml.

lib/migrate_user_config.py
lucasmoura: I was just wondering because this is not an error message per se, just a status update

debian/ubuntu-advantage-tools.postinst
```diff
@@ -326,6 +326,16 @@ cleanup_old_motd_files() {
     rm -f $UA_MESSAGES_DIR/motd*
 }
 
+migrate_user_config_post() {
+    # LP: #2004280
+    if [ -f /etc/ubuntu-advantage/uaclient.conf.preinst-backup ]; then
+        # This script modifies the preinst-backup version of the file in-place
+        /usr/bin/python3 /usr/lib/ubuntu-advantage/migrate_user_config.py
```
lucasmoura, line 333: I know this is a corner case scenario, but is it possible for this script to fail and leave `uaclient.conf.preinst-backup` in a poor state ? Because if yes, then we could be creating an invalid config for the user.

Maybe we should have an extra file for the migration and only performs the move command if the script was able to run without issues

uaclient/config.py
```diff
@@ -598,40 +611,26 @@ def process_config(self):
...
-        content += yaml.dump(cfg_dict, default_flow_style=False)
-        system.write_file(config_path, content)
-
     def warn_about_invalid_keys(self):
         if self.invalid_keys is not None:
             for invalid_key in sorted(self.invalid_keys):
                 logging.warning(
                     "Ignoring invalid uaclient.conf key: %s", invalid_key
                 )
+        if "ua_config" in self.cfg:
```
lucasmoura, line 620: Should we have an integration test for this ? I think we will show this to the user when running a CLI command because we have change the log level to debug, which should overwrite `disable_log_to_console`. But it is worth having a test confirming that

lib/migrate_user_config.py
```diff
@@ -0,0 +1,156 @@
...
+
+import json
+import sys
+
+import yaml
+
+from uaclient import defaults, messages
+
+UACLIENT_CONF_BACKUP_PATH = (
+    "/etc/ubuntu-advantage/uaclient.conf.preinst-backup"
```
lucasmoura, line 21: I think we can move this into the defaults module

sru/release-27.14/test-migrate-user-config.sh
```diff
@@ -0,0 +1,273 @@
...
+
+# xenial
+test_normal_upgrade                 xenial 27.11.3~16.04.1
+test_normal_upgrade                 xenial 27.13.1~16.04.1
+test_apt_news_false_upgrade         xenial 27.11.3~16.04.1
+test_apt_news_false_upgrade         xenial 27.13.1~16.04.1
+test_uaclient_conf_changes_upgrade  xenial 27.11.3~16.04.1
+test_uaclient_conf_changes_upgrade  xenial 27.13.1~16.04.1
+test_migration_failure              xenial 27.11.3~16.04.1
+test_migration_failure              xenial 27.13.1~16.04.1
```
lucasmoura, line 271: @orndorffgrant maybe I am missing something, but I think we are still missing the check for the `uaclient.conf.preinst-backup` file

lib/migrate_user_config.py
orndorffgrant: That's a good point, but is it weird for us to print to stdout in the middle of an upgrade? I don't think we do that anywhere else (but maybe I'm missing something).
@panlinux do you know what is the correct way to display an informational message during an apt upgrade? (OR should we not print anything at all unless something goes wrong?)

debian/ubuntu-advantage-tools.postinst
orndorffgrant: We have to always `mv` it back because even in failure scenarios it should contain the backed up config.

The code already mitigates the risk that the file is corrupted by doing everything possible, including yaml serialization, before `open`ing and `write`ing the file. So the only possibility of a corrupted file is if the process is killed in  the middle of the `write`. That is possible though.

(Maybe this is what you're suggesting): We could go one step farther and write the new contents to a separate file and `os.rename` it as the final step in the python script - I think that would remove that last possibility of a partial write.

What do you think?

sru/release-27.14/test-migrate-user-config.sh
orndorffgrant: Ah oops yep I was focused on everything else and just forgot - I'll add it

lib/migrate_user_config.py
orndorffgrant: This is only used in this script and only ever will be used in this script, so I'm not sure `defaults`, which is used more broadly, is the right place for it.

uaclient/config.py
orndorffgrant: The console log level is hard-coded to INFO unless you call the cli with --debug, so I think this should behave the same as the existing warning.

It's still probably a good idea to have an integration test specifically for supporting the old `ua_config` and asserting what the output will be. I'll add it

sru/release-27.14/test-migrate-user-config.sh
orndorffgrant: Oh wait sorry I did add it. There are now sections called "Backup file is gone after successful migration" where we `ls -al /etc/ubuntu-advantage`

Or did you want something more?

uaclient/config.py
orndorffgrant: Added!

lib/migrate_user_config.py
lucasmoura: Good point

debian/ubuntu-advantage-tools.postinst
lucasmoura: Yes, I think that is a safer approach. Even though this failure should be rare as you mentioned it, I think it is better to be safe than sorry here

orndorffgrant: That makes sense! I'll do this then

orndorffgrant: Updated!

features/config.feature
```diff
@@ -0,0 +1,55 @@
...
+        global_apt_http_proxy   None
+        global_apt_https_proxy  None
+        update_messaging_timer  21600
+        metering_timer          14400
+        apt_news                False
+        apt_news_url            https://motd.ubuntu.com/aptnews.json
+        """
+        Then I will see the following on stderr:
+        """
+        """
```
lucasmoura, line 50: This is just a suggestion, can we run a cli command here just to guarantee that we will warn the user about the `ua_config` key we have added ?

lucasmoura (approved):

renanrodrigo (approved):
LGTM overall, just one question

uaclient/entitlements/tests/test_esm.py
```diff
@@ -17,153 +16,6 @@ def entitlement(request, entitlement_factory):
     return entitlement_factory(request.param, suites=["xenial"])
 
 
-@mock.patch("uaclient.system.is_lts", return_value=True)
-@mock.patch("uaclient.util.validate_proxy", side_effect=lambda x, y, z: y)
-@mock.patch("uaclient.jobs.update_messaging.update_motd_messages")
-@mock.patch("uaclient.apt.setup_apt_proxy")
-class TestESMEntitlementEnable:
```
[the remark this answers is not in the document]
orndorffgrant: `test_enable_calls_adds_apt_repo_and_calls_apt_update` in test_repo.py already asserts:
* add_apt_repository is called appropriately
* apt-get update is called
* apt-get install is called appropriately
* setup_apt_proxy is called

This test might have slightly different details in some assertions, but even if it covers a small aspect that the above test doesn't, I don't think it is worth maintaining this test for that coverage. If we notice a gap, we can add a new unit test to cover that gap.

lib/migrate_user_config.py
```diff
@@ -0,0 +1,163 @@
...
+    new_user_config = {}
+    # only keep a setting if it was changed
+    for field in USER_CONFIG_DEFAULTS.keys():
+        old_val = old_user_config.get(field)
+        if old_val is not None and old_val != USER_CONFIG_DEFAULTS.get(field):
+            new_user_config[field] = old_val
+
+    try:
+        with open(
+            defaults.DEFAULT_DATA_DIR + "/user-config.json", "w"
```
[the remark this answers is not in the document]
orndorffgrant: Updated!

orndorffgrant:
@lucasmoura @renanrodrigo this one is updated

lucasmoura (approved):

orndorffgrant:
tests passed locally

PR #2403, opened 2023-01-30
preinst: add more md5sum checks

lucasmoura:
## Proposed Commit Message
preinst: fix md5sum check

The md5sum for for the situation where the user set apt_news to false on 27.13.1 and later is not entirely correct, as it is missing some values in the beginning of the hash. We are adding those missing characters

## Test Steps
Guarantee that there is no prompt if the user modifies `apt_news` on either 27.13.1 or 27.13.2 and upgrades to this version of pro client

## Checklist

 - [ ] Yes
 - [ ] No


renanrodrigo:
Still need to ensure all versions: 27.12 and 27.11.3 too imo

orndorffgrant (approved):
LGTM!

PR #2406, opened 2023-01-30
preinst: restore correct uaconfig

lucasmoura:
## Proposed Commit Message
preinst: restore correct uaconfig

If we need to restore a default uaconfig, we need to check which package version the user is running on, so that we can select the appropriate default values for the config

## Test Steps
Launch a container and install pro version 27.13.1
Run `pro config set apt_news=true`
Upgrade to the version that contains this change
Verify that uaclient.conf is the default version for 27.13.1 (missing `update_status_timer` key)

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/debian/ubuntu-advantage-tools.preinst
+++ b/debian/ubuntu-advantage-tools.preinst
@@ -12,10 +12,34 @@ remove_old_config_fields() {
 }
 
 restore_previous_conffile() {
+    previous_pkg_version=$1
     # Back up existing conffile in case of an error unwind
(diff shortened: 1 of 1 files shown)


orndorffgrant (approved):
Tested with the following steps:

1. install u-a-t 27.13.1
2. `pro config set apt_news=true`
3. upgrade to new version built from source
4. `debsums -e ubuntu-advantage-tools`

before this change I get:
`/etc/ubuntu-advantage/uaclient.conf                                       FAILED`

After this change I get:
`/etc/ubuntu-advantage/uaclient.conf                                           OK`

Approving and merging

PR #2413, opened 2023-02-01
apt: change esm security updates message

lucasmoura:
## Proposed Commit Message
apt: change esm security updates message

Change apt esm message to make it clearer that any esm service is providing more security updates, not restricting the user to only get updates if they have the service enabled

## Test Steps

## Checklist

 - [ ] Yes
 - [x] No


docs/explanations/apt_messages.md
```diff
@@ -53,7 +53,7 @@ where we deliver those messages are:
 
   ```
   *Your Ubuntu Pro subscription has EXPIRED*
-  The following security updates require Ubuntu Pro with 'esm-infra' enabled:
+  Get more security updates through Ubuntu Pro with 'esm-infra' enabled
```
renanrodrigo, line 56: The problem I see is that it only apply for apps. For infra, the `require` message is more accurate

lucasmoura: That is true, what about if we only changed that for `esm-apps` then ?

renanrodrigo: Yep, that's what I think we should do

lucasmoura: @renanrodrigo updated

apt-hook/json-hook.cc
```diff
@@ -266,7 +266,12 @@ ESMContext get_esm_context() {
...
 void print_esm_packages(std::string esm_name, std::vector<std::string> package_names, ESMContext &esm_context) {
-    std::cout << "The following security updates require Ubuntu Pro with '";
+
+    if (esm_name == "esm-apps")
+        std::cout << "Get more security updates through Ubuntu Pro with '";
+    else
+        std::cout << "The following security updates require Ubuntu Pro with '";
+
     std::cout << esm_name;
     std::cout << "' enabled:" << std::endl;
```
orndorffgrant, line 277: I know it's not technically necessary but I suggest we always use brackets.

Also, since we're doing different messages based on this anyway, we can just put the whole message in the if statement.
```suggestion
    if (esm_name == "esm-apps") {
        std::cout << "Get more security updates through Ubuntu Pro with 'esm-apps` enabled:";
    } else {
        std::cout << "The following security updates require Ubuntu Pro with 'esm-infra' enabled";
    }
```

Also, we could change the parameter of this function from `std::string esm_name` to an enum like: `enum ESMType {APPS, INFRA};`

orndorffgrant (approved):
Just stylistic suggestions

apt-hook/json-hook.cc
lucasmoura: @orndorffgrant Good suggestions. I have updated the code

renanrodrigo (approved):

lucasmoura:
CI errors are unrelated

PR #2414, opened 2023-02-01
Allow to use UAC with downstream distributions

tintou:
## Proposed Commit Message
Please use rebase-merge type as the two commits are not really related.

## Test Steps
The elementary os-release file is at https://github.com/elementary/os-patches/blob/base-files-jammy-patched/etc/os-release#L18
It can be used to test this branch but I've included it in the tests too.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/system.py
+++ b/uaclient/system.py
@@ -160,6 +160,12 @@ def get_platform_info() -> Dict[str, str]:
     platform_info["version"] = version
 
     match = re.match(REGEX_OS_RELEASE_VERSION, version)
+    if not match and "UBUNTU_VERSION" in os_release:
+        version = os_release["UBUNTU_VERSION"]
+        version = re.sub(r"\.\d LTS", " LTS", version)
+        platform_info["version"] = version
+
+    match = re.match(REGEX_OS_RELEASE_VERSION, version)
     if not match:
         raise RuntimeError(
             "Could not parse /usr/lib/os-release VERSION: {} (modified to"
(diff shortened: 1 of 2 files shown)


orndorffgrant:
Hi @tintou - thanks for the PR! I just want to let you know that we've seen this and, on brief inspection, are not opposed to the changes. The team is busy preparing a bugfix release now, but we'll come back and do a more thorough review of this soon :)

lucasmoura:
Hi @tintou, sorry for the huge delay here. As @orndorffgrant mentioned, we are not opposed to the changes, but we just want to double-check your expectations regarding it. If that PR is supposed to avoid showing a degraded state on systemd, we have hopefully fixed that already on version 27.13.5 (#2420). Therefore, please let us know if that is the case.

If that is not the case and if the intention here is to add support for downstream distributions, we have no issue on accepting the PR, but we are currently not explicitly supporting any downstream releases. Additionally, since we don't have a test infrastructure for any downstream releases as well, we understand that we can unintentionally "break" support for them in future. Please let us know if that is acceptable.

Finally, maybe this PR covers another issue that we have not identified. If that is true, no worries at all, you can point us to the issue we missed and we can work to land this PR for the next release of the project.

cpaelzer:
Hey @tintou 
did you have any chance to ponder about the open questions?
Due to time passing this gets out of date, but without the answers we are unsure what exactly you want to achieve.

Actually your first commit `system: Default to usr/lib/os-release` is rather clear and without a debate.
We will have a look at cherry-picking that into the next release already.

I'm marking this PR closed for a lack of interactivity, but we'd be happy to reconsider once you have a chance to discuss with us.

PR #2417, opened 2023-02-02
docs: add Discourse link to front page

s-makin:
```
Now that the docs are live, they need to be linked up
with the forums. I have therefore added a link to the
beta invitation post made by Lech on Discourse. 

I have also reformatted the "get help" section slightly.
Please let me know if you think I need to tweak the
wording, or if you can think of any other links we should
add while I'm at it! 
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -55,12 +55,23 @@ for you.
 Getting help
 ============
 
-Having trouble? We would like to help! For help on a specific page in this
-documentation, click on the "Have a question?" link at the top of that page.
-You can also...
+Ubuntu Pro is a new product, and we're keen to know about your experience of
+using it!
 
-- Ask a question in the ``#ubuntu-server`` `IRC channel on Libera`_
-- `Report bugs on Launchpad`_
+- **Having trouble?**
+  We would like to help! To get help on a specific page in this documentation,
+  simply click on the "Have a question?" link at the top of that page. This
+  will open up an issue in GitHub where you can tell us more about the problem
+  you're having and we will do our best to resolve it for you.
+
+- **Found a bug?**
+  You can `Report bugs on Launchpad`_!
+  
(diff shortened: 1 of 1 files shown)


renanrodrigo (approved):

docs/index.rst
```diff
@@ -108,3 +119,4 @@ constructive feedback.
 .. _Code of conduct: https://ubuntu.com/community/code-of-conduct
 .. _Contribute: https://github.com/canonical/ubuntu-advantage-client/blob/main/CONTRIBUTING.md
 .. _IRC channel on Libera: https://kiwiirc.com/nextclient/irc.libera.chat/ubuntu-server
+.. _our Discourse forum: https://discourse.ubuntu.com/t/ubuntu-pro-beta-invitation/30957
```
orndorffgrant, line 122: Alternatively, we could link to the "Ubuntu Pro" topic on discourse https://discourse.ubuntu.com/c/ubuntu-pro/116
Then people can create new posts or browse the existing posts.
What do you think?

s-makin: I did have a scan through all the posts on there, but it looks a bit...empty...so I went for the one that was best constructed and most helpful to new users.

That said, I can change it to the topic if you think that's better, or we could create a new post specifically to welcome peoples' feedback/questions/etc on the pro client.

renanrodrigo: hmmmm Grant is right, this one is for the beta. Pointing to the topic makes more sense? I would expect it to start empty as people didn't have the time to fill it yet (:

s-makin: Should we start a "welcome" post, in that case?

s-makin: On reflection, I think Grant was right in the first place - I've updated the link as he suggested to the one for the Pro Discourse main channel.

PR #2419, opened 2023-02-02
apt: make sure all calls to the ESM cache are isolated

renanrodrigo:
system configuration, except Acquire configuration, is cleared everywhere we call the cache

This PR guarantees a promise we made in the SRU of 27.13.1, where the calls to apt.Cache would guarantee isolation from the system environment except for the needed configuration.

## Test Steps
Green CI, everything still working as normal.

## Checklist

 - [x] Yes - Robie or Andreas, maybe Julian?
 - [ ] No

### DIFF
--- a/apt-hook/esm-counts.cc
+++ b/apt-hook/esm-counts.cc
@@ -52,11 +52,24 @@ bool get_potential_esm_updates(ESMUpdates &updates) {
    }
 
    // set up esm cache
+   // make sure the configuration is isolated, apart from Acquire
+   for (const Configuration::Item *ConfigItem = _config->Tree(0); ConfigItem != NULL; ConfigItem = ConfigItem->Next) {
+      if (ConfigItem->FullTag(0) != "Acquire") {
+         _config->Clear(ConfigItem->FullTag(0));
+      }
+   }
+
    _config->Set("Dir", "/var/lib/ubuntu-advantage/apt-esm/");
    _config->Set("Dir::State::status", "/var/lib/ubuntu-advantage/apt-esm/var/lib/dpkg/status");
+
+   if (!pkgInitConfig(*_config)) {
+      return false;
+   }
+
    if (!pkgInitSystem(*_config, _system)) {
       return false;
    }
+
    pkgCacheFile esm_cachefile;
    pkgCache *esm_cache = esm_cachefile.GetPkgCache();
    if (esm_cache == NULL) {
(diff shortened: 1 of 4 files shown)


uaclient/apt.py
```diff
@@ -598,12 +598,33 @@ def get_apt_cache_datetime() -> Optional[datetime.datetime]:
...
+    import apt_pkg  # type: ignore
+
+    try:
+        # Take care to initialize the cache with only the
+        # Acquire configuration preserved
+        for key in apt_pkg.config.keys():
+            if "Acquire" not in key:
+                apt_pkg.config.clear(key)
+                apt_pkg.config.set("Dir", ESM_APT_ROOTDIR)
+                apt_pkg.init_config()
```
orndorffgrant, line 613: I think these are too indented
```suggestion
        apt_pkg.config.set("Dir", ESM_APT_ROOTDIR)
        apt_pkg.init_config()
```

renanrodrigo: they are they are

apt-hook/esm-counts.cc
```diff
@@ -52,11 +52,26 @@ bool get_potential_esm_updates(ESMUpdates &updates) {
...
 
    // set up esm cache
+   // make sure the configuration is isolated, apart from Acquire
+   const Configuration::Item *ConfigItem = _config->Tree(0);
+   while (ConfigItem) {
+      if (ConfigItem->FullTag(0) != "Acquire") {
+         _config->Clear(ConfigItem->FullTag(0));
+      }
+      ConfigItem = ConfigItem->Next;
+   }
```
orndorffgrant, line 62: Optional: I think we could turn this into a for loop to be maybe slightly more readable. Something like this:
```suggestion
   for (Configuration::Item *ConfigItem = _config->Tree(0); ConfigItem != NULL; ConfigItem = ConfigItem->Next) {
      if (ConfigItem->FullTag(0) != "Acquire") {
         _config->Clear(ConfigItem->FullTag(0));
      }
   }
```

orndorffgrant (approved):
LGTM! Optional stylistic suggestion

apt-hook/esm-counts.cc
renanrodrigo: sure why not

renanrodrigo:
Since this didn't land in the release branch in time, I am changing it in favor of #2440, which targets main

PR #2420, opened 2023-02-06
system: use VERSION_CODENAME if needed

lucasmoura:
## Proposed Commit Message
system: use VERSION_CODENAME if needed

When parsing the VERSION field from `/etc/os-release`, we might not have the series in the `VERSION` string. If we detect that, we now fallback to `VERSION_CODENAME` for that info

## Test Steps
Guarantee that no integration test fails due to this change or that we don't incorrectly fallback to `VERSION_CODENAME`

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/uaclient/system.py
+++ b/uaclient/system.py
@@ -20,7 +20,9 @@ DBUS_MACHINE_ID = "/var/lib/dbus/machine-id"
 DISTRO_INFO_CSV = "/usr/share/distro-info/ubuntu.csv"
 
 # N.B. this relies on the version normalisation we perform in get_platform_info
-REGEX_OS_RELEASE_VERSION = r"(?P<release>\d+\.\d+) (LTS )?\((?P<series>\w+).*"
+REGEX_OS_RELEASE_VERSION = (
+    r"(?P<release>\d+\.\d+) (LTS\s*)?(\((?P<series>\w+))?.*"
+)
 
 RE_KERNEL_UNAME = (
     r"^"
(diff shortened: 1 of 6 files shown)


uaclient/system.py
```diff
@@ -154,28 +156,33 @@ def get_platform_info() -> Dict[str, str]:
         "type": "Linux",
     }
 
-    version = os_release["VERSION"]
-    # Strip off an LTS point release (20.04.1 LTS -> 20.04 LTS)
-    version = re.sub(r"\.\d LTS", " LTS", version)
+    version = re.sub(r"\.\d LTS", " LTS", os_release["VERSION"])
```
orndorffgrant, line 159: while we're here lets be more defensive on this one too
```diff
-version = re.sub(r"\.\d LTS", " LTS", os_release["VERSION"])
+version = re.sub(r"\.\d LTS", " LTS", os_release.get("VERSION", ""))
```

Same down below where we also use `os_release["VERSION"]`

uaclient/system.py
```diff
@@ -154,28 +156,33 @@ def get_platform_info() -> Dict[str, str]:
...
     platform_info["version"] = version
+    series = os_release.get("VERSION_CODENAME")
+    release = os_release.get("VERSION_ID", "")
```
orndorffgrant, line 162: Let's be consistent with our fallback values - I'm okay with either None or ""
```diff
-series = os_release.get("VERSION_CODENAME")
+series = os_release.get("VERSION_CODENAME", "")
```

Same down below where we use `.get("series")`

orndorffgrant (approved):
LGTM! Just some nits

lucasmoura:
@orndorffgrant updated

orndorffgrant (approved):
:tada: Let's let CI pass

uaclient/tests/test_system.py
```diff
@@ -679,6 +690,43 @@ def test_get_platform_info_error_no_version(self, m_parse):
...
+            (
+                {
+                    "NAME": "Ubuntu",
+                    "VERSION": "22.04 LTS",
+                    "VERSION_CODENAME": "Jammy",
+                },
+                "amd64",
+                "kernel-ver4",
+                {
+                    "arch": "amd64",
```
cpaelzer, line 702: Since the real errors came from derivatives like mint/popos and such - should we add test data based on those as well here?

lucasmoura: We can, but our goal is not to prevent those releases from failing here, but failing gracefully if we can't extract that information. The refactor in this part of the code is just so we can better identify in the logs the cause of failures. By that, I mean that we don't want to support those releases with this PR, just allow the code to fail gracefully. This is achieved by the new `try..except` block on the `esm_cache` module that is now running on those releases. But let me know if you disagree

cpaelzer: I agree with the solution, I just wanted to see a full content block as it would appear in one of the derivatives in the test data.

uaclient/system.py
cpaelzer: solved via code

PR #2422, opened 2023-02-06
fix: recycle invalid jobs-status.json file

renanrodrigo:
If the file is corrupted/invalid, erase it, run the jobs and create a new one.

LP: #2006261

## Test Steps
new unit test added
test for SRU: create the jobs-status file with random stuff, run it, see it works

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -8,6 +8,7 @@ from typing import Callable, Optional
 
 from uaclient.cli import setup_logging
 from uaclient.config import UAConfig
+from uaclient.exceptions import InvalidFileFormatError
 from uaclient.files.state_files import (
     AllTimerJobsState,
     TimerJobState,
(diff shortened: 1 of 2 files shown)


lib/timer.py
```diff
@@ -139,7 +140,13 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
...
     state introspection for jobs which have not yet run.
     """
-    jobs_status_obj = timer_jobs_state_file.read()
+    jobs_status_obj = None
+    # If the file format is wrong we remove it, and after the jobs are
+    # executed it will be recreated with the proper data.
+    try:
+        jobs_status_obj = timer_jobs_state_file.read()
+    except InvalidFileFormatError:
+        timer_jobs_state_file.delete()
```
cpaelzer, line 149: I can see scenarios where even deletion fails.
We can't do much in those cases ...
But - while we are at trying to change crashes into proper errors - should we change it to also gracefully die if it fails to remove it?

I see that down in the stack `ensure_file_absent` will prevent `FileNotFoundError` from ever happening, but `OSError` and `PermissionError` might still occur.

renanrodrigo: Good catch - I will update this to fail gracefully in that case.
But then, what should we do? Always run the jobs when the timer triggers, because we don't know when they ran last time? or don't run the jobs at all?

orndorffgrant (approved):

lib/timer.py
```diff
@@ -139,7 +140,19 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
...
-    jobs_status_obj = timer_jobs_state_file.read()
+    jobs_status_obj = None
+    # If the file format is wrong we remove it, and after the jobs are
+    # executed it will be recreated with the proper data.
+    try:
+        jobs_status_obj = timer_jobs_state_file.read()
+    except InvalidFileFormatError:
+        try:
+            timer_jobs_state_file.delete()
+        except (OSError, PermissionError):
```
lucasmoura, line 151: I think we can just catch the generic Exception here and also log the exception just so we can better debug this if needed

renanrodrigo: I think yes, logging the exception is good. But catching generic, well, I don't know. I prefer to be specific, people will complain about the "bad practice" again...

PR #2426, opened 2023-02-07
Release 27.13.4

lucasmoura:


debian/changelog
```diff
@@ -1,3 +1,30 @@
+ubuntu-advantage-tools (27.13.4~23.04.1) lunar; urgency=medium
+
```
renanrodrigo, line 2: I think there is an extra newline here

renanrodrigo (approved):
LGTM, small correction

debian/changelog
lucasmoura: Fixed

PR #2431, opened 2023-02-09
Fix wrong changelog entry

renanrodrigo:
d/postinst is wrong it seems


orndorffgrant (approved):
You can just push this kind of thing directly to `release` IMO

renanrodrigo:
Well @lucasmoura have fixed it directly to the MP branches, so we're good here anyway.
I think we need to define how hard we follow the process as a whole.

PR #2433, opened 2023-02-10
postinst cleanups + remove root_mode

orndorffgrant:
## Test Steps

LMK if you disagree with the postinst changes.
LMK if you disagree with the new util function for is_root (it's intended to be a bit more readable)
Integration tests should still pass.
Manually run the Pro fips tests.
Manually run the new notice code in postinst to make sure it works as expected.

## Checklist

 - [x] I have updated or added any documentation accordingly (these are all "under-the-hood" changes - no docs necessary)

 - [ ] Yes
 - [x] No (? unless you think we need extra reviews because of the changes to postinst)

### DIFF
--- a/features/ubuntu_pro_fips.feature
+++ b/features/ubuntu_pro_fips.feature
@@ -118,7 +118,6 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO fips image
         And stdout matches regexp:
         """
         NOTICES
-        Disabling FIPS requires system reboot to complete operation.
         FIPS support requires system reboot to complete configuration.
         """
         When I reboot the machine
@@ -336,7 +335,6 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO fips image
         And stdout matches regexp:
         """
         NOTICES
-        Disabling FIPS requires system reboot to complete operation.
         FIPS support requires system reboot to complete configuration.
         """
         When I reboot the machine
@@ -607,7 +605,6 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO fips image
         And stdout matches regexp:
         """
         NOTICES
-        Disabling FIPS requires system reboot to complete operation.
         FIPS support requires system reboot to complete configuration.
         """
         When I reboot the machine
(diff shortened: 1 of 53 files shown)


uaclient/entitlements/tests/test_repo.py
```diff
@@ -433,7 +433,7 @@ def test_enable_can_exit_on_pre_or_post_disable_messaging_hooks(
     @pytest.mark.parametrize("should_reboot", (False, True))
     @pytest.mark.parametrize("with_pre_install_msg", (False, True))
     @pytest.mark.parametrize("packages", (["a"], [], None))
-    @mock.patch("os.getuid", return_value=0)
+    @mock.patch("uaclient.util.we_are_currently_root", return_value=True)
```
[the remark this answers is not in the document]
orndorffgrant: Ah yes this is something I wanted to bring up. I agree, but I wanted to see if anyone disagreed with the autouse mock before going down that path. Let's just sync to confirm after standup tomorrow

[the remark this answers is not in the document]
orndorffgrant: Done!

uaclient/files/notices.py
```diff
@@ -160,27 +160,48 @@ def list(self) -> List[Tuple[str, str]]:
...
+                        notice = next(
+                            iter(
+                                [
+                                    n
+                                    for n in Notice
+                                    if n.order_id == order_id
+                                    and n.label == label
+                                ]
+                            ),
+                        )
```
lucasmoura, line 193: This is just a style suggestion, so feel free to disagree here, but we can't we just loop the notices and raise an exception if we can find any that matches the empty notice name ?

features/ubuntu_pro_fips.feature
```diff
@@ -527,6 +527,7 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO fips image
         """
         1
         """
+        When I run `systemctl daemon-reload` with sudo
```
lucasmoura, line 530: I know this is unrelated, but those pro-fips tests are failing due to the extra check for the `esm-infra-security` when running apt-cache policy on the tested package. Maybe it is worth an extra commit just to address that and fix those tests as well

lucasmoura (approved):

uaclient/files/notices.py
orndorffgrant: I agree that will be more readable, I'll add it before merging

features/ubuntu_pro_fips.feature
orndorffgrant: Ah darn that wasn't happening when I was editing this. I'll see if I can fix it and tack on an extra commit here before merging

PR #2409, opened 2023-01-31
status: refactor update contract info to be a background job (SC-1411)

aciba90:
## Proposed Commit Message

```
status: refactor update contract info to be a background job

At the moment, the backend service call to check if the
contract info has to be updated can be slow under load,
and this was executed during `pro status`, blocking it
during some seconds.

Move the update contract info check as a background job
to improve user experience.
```

## Checklist

 - [x] I have updated or added any unit tests accordingly
 - [x] I have updated or added any integration tests accordingly
 - [ ] I have updated or added any documentation accordingly

## Does this PR require extra reviews?

 - [ ] Yes
 - [x] No

PR #2446, opened 2023-02-16
tests: fix unit test to assert correct call count

orndorffgrant:


uaclient/tests/test_ua_timer.py
```diff
@@ -183,7 +183,7 @@ def test_run_jobs_tries_to_create_new_file_when_invalid(
             assert [
                 mock.call(m_jobs_state())
             ] == fake_file.write.call_args_list
-            assert 2 == m_run_job.call_count
+            assert 3 == m_run_job.call_count
```
lucasmoura, line 186: But don't we have only two jobs ? Why are we running it three times ?

orndorffgrant: we have 3 now :) from https://github.com/canonical/ubuntu-advantage-client/pull/2409

PR #2448, opened 2023-02-17
Force importing yaml from the system dist-packages

renanrodrigo:
When users install a new pyyaml using pip, it will end up in pythonpath with a higher priority.
This PR introduces some logic to import yaml only directly from the system dist-packages.
This fixes two bugs:
LP: #2007234 : If yaml is not present, then a UserFacingError is thrown - this can only happen if the `python3-yaml` deb is not present or was directly modified, and as it is a dependency, it does not make sense to support anything like this.
LP: #2007241 : The system yaml won't have a version which is incompatible with the python version, so the safe_load bug is prevented.

## Test Steps
Added unit tests and integration tests to verify the fix.

## Checklist

 - [ ] Yes
 - [ ] No
 - [x] I hope not

### DIFF
--- a/uaclient/messages.py
+++ b/uaclient/messages.py
@@ -1171,3 +1171,10 @@ from the system by running:
 
 $ sudo rm {lock_file_path}""",
 )
+
+MISSING_YAML_MODULE = NamedMessage(
+    "missing-yaml-module",
+    """\
+Couldn't import the YAML module from /usr/lib/python3/dist-packages.
+Make sure the 'python3-yaml' package is installed correctly.""",
+)
--- a/uaclient/yaml.py
+++ b/uaclient/yaml.py
@@ -1,0 +1,42 @@
+import sys
+from importlib import machinery as importlib_machinery
+from importlib import util as importlib_util
+
+from uaclient.exceptions import UserFacingError
+from uaclient.messages import MISSING_YAML_MODULE
+
+
+def get_imported_yaml_module():
+    # Maybe we already imported it?
+    if "yaml" in sys.modules:
+        return sys.modules["yaml"]
+
+    # Try to get the system yaml module
+    pyyaml_spec = importlib_machinery.PathFinder.find_spec(
+        "yaml", path=["/usr/lib/python3/dist-packages"]
+    )
+    # If there is no system 'yaml' module (which is weird enough),
+    # then raise an exception asking for it to be there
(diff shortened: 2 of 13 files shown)


uaclient/messages.py
```diff
@@ -1169,3 +1169,10 @@ class TxtColor:
 
 $ sudo rm {lock_file_path}""",
 )
+
+MISSING_YAML_MODULE = NamedMessage(
+    "missing-yaml-module",
+    """\
+Couldn't import the YAML module from /usr/lib/python3/dist-packages.
+Make sure the 'python3-yaml' package is installed correctly.""",
```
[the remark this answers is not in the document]
renanrodrigo: Yes unecessary - it is a dependency of ubuntu-advantage-tools

orndorffgrant (commented):
Wow thanks for figuring all of this out!
Looks good, but I can't get the new integration test to fail with e.g. install_from=archive - ideally the test would fail on the current version and pass on this PR though

renanrodrigo:
@orndorffgrant hmmm well I'll investigate and see why it does not break... Thanks for the heads up

renanrodrigo:
@orndorffgrant surprisingly, Python itself ensures retro compatibility and it does not fail until 3.10!
SO the bugs are vast on Jammy/Kinetic because there is where it matters.
I edited the test to run on Jammy+, where you can see the failure and fix flow.

orndorffgrant (approved):
Thanks for the update! LGTM

uaclient/yaml.py
```diff
@@ -0,0 +1,43 @@
...
+from uaclient.messages import MISSING_YAML_MODULE
+
+
+def get_imported_yaml_module():
+    # Maybe we already imported it?
+    if "yaml" in sys.modules:
+        return sys.modules["yaml"]
+
+    # Try to get the system yaml module
+    finder = importlib_machinery.PathFinder()
```
lucasmoura, line 15: Do we need to instantiate the `PathFinder` object ? it seems that `find_spec` is a class method

lucasmoura (approved):

uaclient/yaml.py
renanrodrigo: You are right, will change accordingly

renanrodrigo:
CI failures unrelated

PR #2449, opened 2023-02-17
logging: Added Redaction Filter for logs

CalvoM:
## Proposed Commit Message

logging: Added Redaction Filter for logs

This is adds the redaction filter for the logging mechanism. 
It is the first of many PRs to come, from the same branch.
- [x] Add Redaction Filter.
- [ ] Add Json Array Formatter.
- [ ] Provide a way for non-root user to log.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/log.py
+++ b/uaclient/log.py
@@ -1,0 +1,11 @@
+import logging
+
+from uaclient import util
+
+
+class RedactionFilter(logging.Filter):
+    """A logging filter to redact confidential info"""
+
+    def filter(self, record: logging.LogRecord):
+        record.msg = util.redact_sensitive_logs(str(record.msg))
+        return True
--- a/uaclient/util.py
+++ b/uaclient/util.py
@@ -182,7 +182,7 @@ def get_dict_deltas(
                 + str(new_value)
                 + "'"
             )
-            logging.debug(redact_sensitive_logs(log))
+            logging.debug(log)
             deltas[key] = new_value
     for key, value in new_dict.items():
         if key not in orig_dict:
@@ -301,7 +301,7 @@ def readurl(
     headers: Dict[str, str] = {},
     method: Optional[str] = None,
     timeout: Optional[int] = None,
-    potentially_sensitive: bool = True,
+    log_response_body: bool = True,
 ) -> Tuple[Any, Union[HTTPMessage, Mapping[str, str]]]:
     if data and not method:
         method = "POST"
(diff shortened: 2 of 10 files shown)


orndorffgrant:
@CalvoM Thanks! Can we make this PR actually use the redaction filter and remove the old redaction method? That would make it a nice self-contained feature that we can test and release independently of the rest of the logging refactor if we need to.

uaclient/log.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def redact_log(self, content: str) -> str:
+        if not isinstance(content, str):
+            content = str(content)
+        redacted_log = util.redact_sensitive_logs(content)
+        return redacted_log
+
+    def filter(self, record: logging.LogRecord):
+        record.msg = self.redact_log(record.msg)
+        return True
```
orndorffgrant, line 17: Can these be combined into one function?

```suggestion
    def filter(self, record: logging.LogRecord):
        record.msg = util.redact_sensitive_logs(str(record.msg))
        return True
```

uaclient/util.py
```diff
@@ -336,9 +334,9 @@ def readurl(
     debug_msg = "URL [{}] response: {}, headers: {{{}}}, data: {}".format(
         method or "GET", url, sorted_header_str, content
     )
-    if potentially_sensitive:
-        # For large responses, this is very slow (several minutes)
-        debug_msg = redact_sensitive_logs(debug_msg)
+    # if potentially_sensitive:
+    #     # For large responses, this is very slow (several minutes)
+    #     debug_msg = redact_sensitive_logs(debug_msg)
```
orndorffgrant, line 339: I think we need a solution to this in this PR.

I believe `potentially_sensitive` is only set to `False` in one place - it defaults to true.

We can change the semantics of this to get a similar result though. The only place where this is set to `False` doesn't really need the full response to be logged anyway.

So maybe remove the `potentially_sensitive` argument, and add a new `log_response_body` argument that defaults to `True`. Then here we'd have something like:
```python
    debug_msg = "URL [{}] response: {}, headers: {{{}}}".format(
        method or "GET", url, sorted_header_str
    )
    if log_response_body:
        # For large responses, this is very slow (several minutes)
        debug_msg += ", data: {}".format(content)
    logging.debug(debug_msg)
```

orndorffgrant (commented):
Looking good! Just two things I think need addressing. Also I haven't done any manual checks yet to verify that logs are still properly redacted.

uaclient/log.py
CalvoM: Oh yes, Ithought of this approach but then I noticed it fails when we pass the exception object to it. Let me confirm that it does not fail or if I can find a better way of logging it.

CalvoM: Oh yes, thanks.

orndorffgrant (approved):
LGTM thanks!

lucasmoura (approved):

PR #2453, opened 2023-02-22
ci: github is removing bionic support, stop running unit-tests against the correct version of python (aside from package builds)

orndorffgrant:
## Test Steps

ci-base should pass

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/.github/workflows/ci-base.yaml
+++ b/.github/workflows/ci-base.yaml
@@ -23,33 +23,21 @@ jobs:
         uses: actions/checkout@v3
       - name: Formatting
         run: tox -e black -e isort
+      - name: Style
+        run: tox -e flake8
       - name: Mypy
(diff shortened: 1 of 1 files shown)


orndorffgrant:
Welp, that didn't work :upside_down_face: - this will require some more effort and investigation into solutions

renanrodrigo:
#maybemaybemaybe we should just not run unit tests in the CI -  we already run them when packaging
what we lose: nothing - those tests are running in different releases with deadsnakes anyway
what we gain: less jobs, more fun

orndorffgrant:
@renanrodrigo @lucasmoura this is updated with our idea to rely on package builds for unit tests

lucasmoura (approved):

renanrodrigo (approved):

PR #2455, opened 2023-02-24
api: add support for unattended-upgrades status

lucasmoura:
## Proposed Commit Message
api: add support for unattended-upgrades status

We are now addding an endpoint to return unattended-upgrades status for users.

## Test Steps
Run the new integration test for the api endpoint

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/features/api_unattended_upgrades.feature
+++ b/features/api_unattended_upgrades.feature
@@ -1,0 +1,207 @@
+Feature: api.u.unattended_upgrades.status.v1
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: v1 unattended upgrades status
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `pro api u.unattended_upgrades.status.v1` as non-root
+        Then stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"apt_periodic_job_enabled": true, "package_lists_refresh_frequency_days": 1, "systemd_apt_timer_enabled": true, "unattended_upgrades_allowed_origins": \["\$\{distro_id\}:\$\{distro_codename\}", "\$\{distro_id\}:\$\{distro_codename\}\-security", "\$\{distro_id\}ESMApps:\$\{distro_codename\}\-apps-security", "\$\{distro_id\}ESM:\$\{distro_codename\}\-infra-security"\], "unattended_upgrades_disabled_reason": null, "unattended_upgrades_frequency_days": 1, "unattended_upgrades_last_run": null, "unattended_upgrades_running": true}, "meta": {"environment_vars": \[\], "raw_config": {"APT::Periodic::Enable": "1", "APT::Periodic::Unattended-Upgrade": "1", "APT::Periodic::Update-Package-Lists": "1", "Unattended-Upgrade::Allowed-Origins": \["\$\{distro_id\}:\$\{distro_codename\}", "\$\{distro_id\}:\$\{distro_codename\}\-security", "\$\{distro_id\}ESMApps:\$\{distro_codename\}\-apps-security", "\$\{distro_id\}ESM:\$\{distro_codename\}\-infra-security"\][,]?\s*<extra_field>}}, "type": "UnattendedUpgradesStatus"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+        """
+        When I create the file `/etc/apt/apt.conf.d/99test` with the following:
+        """
+        APT::Periodic::Enable "0";
+        """
+        And I run `apt-get update` with sudo
+        And I run `apt-get install jq -y` with sudo
+        And I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.apt_periodic_job_enabled` as non-root
+        Then I will see the following on stdout:
+        """
+        false
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_running` as non-root
+        Then I will see the following on stdout:
+        """
+        false
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_disabled_reason.msg` as non-root
+        Then I will see the following on stdout:
+        """
+        "APT::Periodic::Enable is turned off"
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_disabled_reason.code` as non-root
+        Then I will see the following on stdout:
+        """
+        "unattended-upgrades-cfg-value-turned-off"
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq '.data.meta.raw_config.\"APT::Periodic::Enable\"'` as non-root
+        Then I will see the following on stdout:
+        """
+        "0"
+        """
+        When I create the file `/etc/apt/apt.conf.d/99test` with the following:
+        """
+        APT::Periodic::Update-Package-Lists "0";
+        """
+        And I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.apt_periodic_job_enabled` as non-root
+        Then I will see the following on stdout:
+        """
+        true
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.package_lists_refresh_frequency_days` as non-root
+        Then I will see the following on stdout:
+        """
+        0
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_running` as non-root
+        Then I will see the following on stdout:
+        """
+        false
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_disabled_reason.msg` as non-root
+        Then I will see the following on stdout:
+        """
+        "APT::Periodic::Update-Package-Lists is turned off"
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_disabled_reason.code` as non-root
+        Then I will see the following on stdout:
+        """
+        "unattended-upgrades-cfg-value-turned-off"
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq '.data.meta.raw_config.\"APT::Periodic::Update-Package-Lists\"'` as non-root
+        Then I will see the following on stdout:
+        """
+        "0"
+        """
+        When I create the file `/etc/apt/apt.conf.d/99test` with the following:
+        """
+        APT::Periodic::Unattended-Upgrade "0";
+        """
+        And I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_frequency_days` as non-root
+        Then I will see the following on stdout:
+        """
+        0
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.package_lists_refresh_frequency_days` as non-root
+        Then I will see the following on stdout:
+        """
+        1
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_running` as non-root
+        Then I will see the following on stdout:
+        """
+        false
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_disabled_reason.msg` as non-root
+        Then I will see the following on stdout:
+        """
+        "APT::Periodic::Unattended-Upgrade is turned off"
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_disabled_reason.code` as non-root
+        Then I will see the following on stdout:
+        """
+        "unattended-upgrades-cfg-value-turned-off"
+        """
+        When I run shell command `pro api u.unattended_upgrades.status.v1 | jq '.data.meta.raw_config.\"APT::Periodic::Unattended-Upgrade\"'` as non-root
+        Then I will see the following on stdout:
+        """
+        "0"
+        """
+        When I run `systemctl stop apt-daily.timer` with sudo
+        And I run `rm /etc/apt/apt.conf.d/99test` with sudo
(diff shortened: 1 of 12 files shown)


orndorffgrant (commented):
Sorry I don't have time for a complete review - I'll review more later.

uaclient/api/api.py
```diff
@@ -28,6 +28,7 @@
     "u.pro.security.status.reboot_required.v1",
     "u.pro.version.v1",
     "u.security.package_manifest.v1",
+    "u.pro.unattended_upgrades.status.v1",
```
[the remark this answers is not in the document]
lucasmoura: That's a good question, maybe not, because this API is not for package related information per se.
But I can understand the rationale for the change.

What do you think @renanrodrigo ?

uaclient/api/u/pro/unattended_upgrades/status/v1.py
```diff
@@ -0,0 +1,190 @@
...
+        *,
+        systemd_apt_timer_enabled: bool,
+        apt_periodic_job_enabled: bool,
+        package_lists_refresh_frequency: str,
+        unattended_upgrades_frequency: str,
+        unattended_upgrades_allowed_origins: List[str],
+        unattended_upgrades_running: bool,
+        unattended_upgrades_disabled_reason: str,
+        unattended_upgrades_last_run: Optional[datetime.datetime],
+        meta: Dict[str, Union[str, List[str]]]
```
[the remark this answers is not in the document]
lucasmoura: Good point, I think that is a better pattern indeed

features/api_unattended_upgrades.feature
```diff
@@ -0,0 +1,81 @@
...
+        Then stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"apt_periodic_job_enabled": true, "package_lists_refresh_frequency": "1", "systemd_apt_timer_enabled": true, "unatteded_upgrades_running": true, "unattended_upgrades_allowed_origins": \["\$\{distro_id\}:\$\{distro_codename\}", "\$\{distro_id\}:\$\{distro_codename\}\-security", "\$\{distro_id\}ESMApps:\$\{distro_codename\}\-apps-security", "\$\{distro_id\}ESM:\$\{distro_codename\}\-infra-security"\], "unattended_upgrades_disabled_reason": "", "unattended_upgrades_frequency": "1", "unattended_upgrades_last_run": null}, "meta": {"APT::Periodic::Enable": "1", "APT::Periodic::Unattended-Upgrade": "1", "APT::Periodic::Update-Package-Lists": "1", "Unattended-Upgrade::Allowed-Origins": \["\$\{distro_id\}:\$\{distro_codename\}", "\$\{distro_id\}:\$\{distro_codename\}\-security", "\$\{distro_id\}ESMApps:\$\{distro_codename\}\-apps-security", "\$\{distro_id\}ESM:\$\{distro_codename\}\-infra-security"\], "environment_vars": \[\]}, "type": "UnattendedUpgradesStatus"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
```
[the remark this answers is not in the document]
lucasmoura: Good idea, I have implemented it in the tests now

lucasmoura:
@orndorffgrant updated

uaclient/api/api.py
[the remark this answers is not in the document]
renanrodrigo: Agreed not directly related to packages, voting for keep-as-is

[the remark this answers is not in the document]
orndorffgrant: That makes sense - resolving

features/api_unattended_upgrades.feature
```diff
@@ -0,0 +1,182 @@
...
+        """
+        When I run shell command `pro api u.pro.unattended_upgrades.status.v1 | jq .data.attributes.systemd_apt_timer_enabled` as non-root
+        Then I will see the following on stdout:
+        """
+        true
+        """
+        When I run shell command `pro api u.pro.unattended_upgrades.status.v1 | jq .data.attributes.unattended_upgrades_allowed_origins` as non-root
+        Then I will see the following on stdout:
+        """
+        []
```
orndorffgrant, line 131: I'm surprised this is an empty list. Am I missing something in the test that sets this?

features/util.py
```diff
@@ -139,6 +139,8 @@ def repo_state_hash(
                 break
         if exclude:
             continue
+        if os.path.isdir(fname):
+            continue
```
orndorffgrant, line 143: Good addition :100:

uaclient/api/u/pro/unattended_upgrades/status/v1.py
```diff
@@ -0,0 +1,195 @@
...
+    "APT::Periodic::Update-Package-Lists",
+    "APT::Periodic::Unattended-Upgrade",
+    "Unattended-Upgrade::Allowed-Origins",
+]
+
+UNATTENDED_UPGRADES_STAMP_PATH = "/var/lib/apt/periodic/upgrade-stamp"
+
+
+class UnattendedUpgradesError(APIError):
+    pass
```
orndorffgrant, line 32: I think this should go in `uaclient.pro.api.exceptions`

uaclient/apt.py
```diff
@@ -653,3 +653,42 @@ def update_esm_caches(cfg) -> None:
...
         logging.warning("Failed to fetch the ESM Apt Cache")
+
+
+def _get_apt_config():
+    import apt_pkg
+
+    # We need to clear the config values in case another module
+    # has already initiated it
+    for key in apt_pkg.config.keys():
+        apt_pkg.config.clear(key)
```
orndorffgrant, line 664: This scares me a bit - maybe because I don't fully understand.

Would it be better or worse if we did something like:
```python
@lru_cache(maxsize=None)
def get_init_apt_pkg():
    import apt_pkg
    apt_pkg.init()
    return apt_pkg
```

and then use that function in the places we need to import apt_pkg and init it?


(and maybe more of a stretch, is there any way to use python3-apt in a non-global-state-ish way?)

orndorffgrant: And I'm not sure how my suggestion would interact with our other usage of `import apt; apt.Cache()` :/

and I'm not sure how it would interact with other programs that use apt_pkg and are using our api :(

orndorffgrant: Reading the docs here: https://apt-team.pages.debian.net/python-apt/library/apt_pkg.html#configuration-and-command-line-parsing

It _might_ be possible to do something like:
```python
def our_apt_related_fn():
    backup_apt_pkg_config = apt_pkg.config
    apt_pkg.config = apt_pkg.Configuration()
    apt_pkg.init_config()
    # do whatever we need to do with our special config
    # but before we return, restore the backup config
    apt_pkg.config = backup_apt_pkg_config
```

CC @renanrodrigo for your thoughts here as well

uaclient/system.py
```diff
@@ -518,3 +518,25 @@ def ensure_folder_absent(folder_path: str) -> None:
...
+    """
+    try:
+        out, _ = subp(["systemctl", "is-active", job_name])
+    except exceptions.ProcessExecutionError as e:
+        out = e.stdout
+
+    if not out:
+        return False
+
+    return True if out.strip() == "active" else False
```
orndorffgrant, line 542: This can be simplified a bit if you want
```suggestion
    return out.strip() == "active"
```

uaclient/api/u/pro/unattended_upgrades/status/v1.py
```diff
@@ -0,0 +1,195 @@
...
+def _get_unattended_upgrades_last_run() -> Optional[datetime.datetime]:
+    try:
+        creation_epoch = os.path.getctime(UNATTENDED_UPGRADES_STAMP_PATH)
+    except FileNotFoundError:
+        return None
+
+    return datetime.datetime.fromtimestamp(creation_epoch)
+
+
+def unattended_upgrades_status() -> UnattendedUpgradesResult:
```
orndorffgrant, line 132: Our pattern so far has been that the function name should be the last word in the endpoint name.
So since this endpoint is `u.pro.unattended_upgrades.status`, then this function name should be `status`

uaclient/api/u/pro/unattended_upgrades/status/v1.py
```diff
@@ -0,0 +1,195 @@
...
+]
+
+UNATTENDED_UPGRADES_STAMP_PATH = "/var/lib/apt/periodic/upgrade-stamp"
+
+
+class UnattendedUpgradesError(APIError):
+    pass
+
+
+class UnattendedUpgradesResult(DataObject, AdditionalInfo):
```
orndorffgrant, line 35: I think this should include "Status" in the name
```suggestion
class UnattendedUpgradesStatusResult(DataObject, AdditionalInfo):
```

uaclient/apt.py
```diff
@@ -653,3 +653,42 @@ def update_esm_caches(cfg) -> None:
...
+
+    apt_pkg.init_config()
+    return apt_pkg.config
+
+
+def get_apt_config_keys(key_base):
+    apt_cfg = _get_apt_config()
+    return apt_cfg.list(key_base)
```
orndorffgrant, line 672: I think this will only return one level of configuration past the key_base - Does this need to search recursively?

uaclient/apt.py
renanrodrigo: Clearing all the config and `init`ing again is fine. Let's keep it as is.
IN THE FUTURE we will have a CtxMgr for our own cache, and then we can just believe apt_pkg without calling `.init` again.

features/api_unattended_upgrades.feature
lucasmoura: Yes, this is empty because we modify `/etc/apt/apt.conf.d/50unattended-upgrades` which is the file that properly configures that variable

uaclient/api/u/pro/unattended_upgrades/status/v1.py
lucasmoura: Indeed, let me fix that

uaclient/apt.py
lucasmoura: Maybe I am missing something, but I think we just want to explore the first level of configuration here. For example, we want all the keys for `Unattended-Upgrades::`. But maybe I am missing your point about search recursively

renanrodrigo:
So far the changeset looks ok to me
Besides the differences in treating the apt_pkg stuff, mentioned some times, the config interaction works well.

uaclient/apt.py
orndorffgrant: I guess my question is: are there only keys like
* `Unattended-Upgrades::OptionA`
* `Unattended-Upgrades::OptionB`

Or are there also nested keys like:
* `Unattended-Upgrades::Nested::OptionC`
* `Unattended-Upgrades::Nested::OptionD`

and if there are nested keys, does this function need to be recursive in some way to find them?

But if there is only one level of keys that we care about then I think this is okay

features/api_unattended_upgrades.feature
orndorffgrant: Ahhhh I see now thanks!
I also just learned that the file in our repo `apt.conf.d/51ubuntu-advantage-esm` is not used I guess

uaclient/api/u/pro/unattended_upgrades/status/v1.py
```diff
@@ -0,0 +1,188 @@
...
+        ),
+        unattended_upgrades_allowed_origins=list(
+            unattended_upgrades_cfg.get("Unattended-Upgrade::Allowed-Origins")
+            or []
+        ),
+        unattended_upgrades_running=unattended_upgrades_running,
+        unattended_upgrades_disabled_reason=disabled_reason,
+        unattended_upgrades_last_run=unattended_upgrades_last_run,
+    )
+    unattended_upgrades_result.meta = unattended_upgrades_cfg
```
orndorffgrant, line 178: my earlier suggestion probably led you astray, but in the spec, we said this would be under the `raw_config` key inside meta.

uaclient/api/u/pro/unattended_upgrades/status/v1.py
```diff
@@ -0,0 +1,188 @@
...
+
+class UnattendedUpgradesStatusResult(DataObject, AdditionalInfo):
+    fields = [
+        Field("systemd_apt_timer_enabled", BoolDataValue),
+        Field("apt_periodic_job_enabled", BoolDataValue),
+        Field("package_lists_refresh_frequency_days", IntDataValue),
+        Field("unattended_upgrades_frequency_days", IntDataValue),
+        Field("unattended_upgrades_allowed_origins", StringDataValue),
+        Field("unattended_upgrades_running", BoolDataValue),
+        Field("unattended_upgrades_disabled_reason", StringDataValue),
```
orndorffgrant, line 39: I think this may need to be a "named string" (e.g. `{ "code": "message-code", "message": "human readable" }`) so that the consumer can choose or translate the actual message as they see fit

orndorffgrant, line 39: I think this may need to be a "named string" (e.g. `{ "code": "message-code", "message": "human readable" }`) so that the consumer can choose or translate the actual message as they see fit

uaclient/apt.py
lucasmoura: Ohhh now I see your point. Yes, the function will indeed just return the first level of keys. 

I have checked the [documentation](https://github.com/mvo5/unattended-upgrades/blob/master/README.md#supported-options-reference) for unattended-upgrades to see if we have a nested variable but it doesn't seem like it.

However, since this document doesn't show all of the unattended-upgrade options, I have also checked the [module](https://github.com/mvo5/unattended-upgrades/blob/8c62c05444ed505c2ca96a8092ee6be027033af3/unattended-upgrade) directly and I couldn't find any as well.

So it seems safe for us to just get the top level keys here

uaclient/api/u/pro/unattended_upgrades/status/v1.py
lucasmoura: That's a good point, let me update this

uaclient/apt.py
orndorffgrant: Sounds good to me!

lucasmoura:
@orndorffgrant updated

features/api_unattended_upgrades.feature
lucasmoura: Good catch. I will create a follow up PR removing that file

orndorffgrant (approved):
LGTM!

renanrodrigo:
This PR is still missing the API documentation for the new endpoint

lucasmoura:
I have updated the endpoint address as we have discussed.
Regarding the documentation, I will create a follow up PR for it, as there are certain characteristics of this endpoint that we will need a discussion beforehand

PR #2456, opened 2023-02-24
apt-cache: only setup the local cache if the service is available

renanrodrigo:
still need to address what to do on systems where the cache is already present, we will probably have to remove it...

LP: #2004018

## Test Steps
I added unit tests, but the verification on a real system was manual on a system that does not actually support ESM.

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -670,23 +670,47 @@ def update_esm_caches(cfg) -> None:
 
     import apt  # type: ignore
 
+    from uaclient.actions import status
     from uaclient.entitlements.entitlement_status import ApplicationStatus
     from uaclient.entitlements.esm import (
         ESMAppsEntitlement,
         ESMInfraEntitlement,
     )
 
+    apps_available = False
+    infra_available = False
+
+    current_status = cfg.read_cache("status-cache")
+    if current_status is None:
+        current_status = status(cfg)[0]
+
+    for service in current_status.get("services", []):
+        if service.get("name", "") == "esm-apps":
+            apps_available = service.get("available", "no") == "yes"
+        if service.get("name", "") == "esm-infra":
+            infra_available = service.get("available", "no") == "yes"
+
     apps = ESMAppsEntitlement(cfg)
 
     # Always setup ESM-Apps
-    if apps.application_status()[0] == ApplicationStatus.DISABLED:
+    if (
+        apps_available
+        and apps.application_status()[0] == ApplicationStatus.DISABLED
+    ):
         apps.setup_local_esm_repo()
+    else:
+        apps.disable_local_esm_repo()
 
     # Only setup ESM-Infra for EOSS systems
     if system.is_current_series_active_esm():
         infra = ESMInfraEntitlement(cfg)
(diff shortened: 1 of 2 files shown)


uaclient/apt.py
```diff
@@ -626,20 +626,31 @@ def update_esm_caches(cfg) -> None:
...
         ESMAppsEntitlement,
         ESMInfraEntitlement,
     )
 
+    current_status = status(cfg)
+    for service in current_status.get("services", []):
+        if service.get("name", "") == "esm-apps":
+            apps_available = service.get("available", "no")
+        if service.get("name", "") == "esm-infra":
+            infra_available = service.get("available", "no")
```
orndorffgrant, line 641: Small optional stylistic suggestion: make these bools here rather than later
```suggestion
            apps_available = service.get("available", "no") == "yes"
        if service.get("name", "") == "esm-infra":
            infra_available = service.get("available", "no") == "yes"
```

orndorffgrant (commented):
I agree that we should probably remove the configuration if the service isn't available.

Is that sufficient in this situation?
1. old version of pro - gets esm-apps cache populated even though its not available
2. install this version - removes esm-apps cache config and runs apt-update
3. Does that clear the existing cache? Or will the old set of esm-apps updates stick around in our cache?

Do you think that removal + solving above problem should go in this PR or be separate?

renanrodrigo:
Thanks @orndorffgrant 
I think removal must be in this very same PR for completeness. But I'm not sure on running apt update; removing the data should be enough - as it is "non invasive" as people have discussed before...

orndorffgrant:
@renanrodrigo Right! Yep that makes sense, sorry for the confusion

uaclient/apt.py
```diff
@@ -626,20 +626,31 @@ def update_esm_caches(cfg) -> None:
...
     import apt  # type: ignore
 
+    from uaclient.actions import status
     from uaclient.entitlements.entitlement_status import ApplicationStatus
     from uaclient.entitlements.esm import (
         ESMAppsEntitlement,
         ESMInfraEntitlement,
     )
 
+    current_status = status(cfg)
```
lucasmoura, line 636: We only create the `status` cache if we run a command as root, but should we try reusing that cache here ? Or do you think this should be dynamic to always get the latest information available ?

renanrodrigo: I added the call to the status cache first, only falling back to status (and creating the cache, because we will be root) in the case there is no cache in the first place.
I will create a subsequent PR to make sure this cache gets stale at some point.

renanrodrigo:
@orndorffgrant I implemented the removal of the repo files as an else in the call itself, as removing it is removing files and the "extra" task there for machines which don't have it available will be two `os.path.exists` checks on apt update.
Besides that, this way we don't need to touch postinst.

orndorffgrant (approved):

orndorffgrant:
@renanrodrigo sorry the other merge caused this to require a rebase

lucasmoura (approved):

renanrodrigo:
rebased

PR #2460, opened 2023-02-28
logging: Added JsonArrayFormatter

CalvoM:
## Proposed Commit Message

logging: Added Redaction Filter for logs

This is adds the redaction filter for the logging mechanism. 

- [x] Add Redaction Filter.
- [x] Add Json Array Formatter.
- [ ] Provide a way for non-root user to log.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/logs.feature
+++ b/features/logs.feature
@@ -1,0 +1,30 @@
+Feature: Logs in Json Array Formatter
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: The log file can be successfully parsed as json array
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `apt update` with sudo
+        And I run `apt install jq -y` with sudo
+        And I verify that running `pro status` `with sudo` exits `0`
+        And I verify that running `pro enable test_entitlement` `with sudo` exits `1`
+        And I run shell command `cat /var/log/ubuntu-advantage.log | jq -r .` as non-root
+        Then I will see the following on stderr
+        """
+        """
+        When I attach `contract_token` with sudo
+        And I verify that running `pro refresh` `with sudo` exits `0`
+        And I verify that running `pro status` `with sudo` exits `0`
+        And I verify that running `pro enable test_entitlement` `with sudo` exits `1`
+        And I run shell command `cat /var/log/ubuntu-advantage.log | jq -r .` as non-root
+        Then I will see the following on stderr
+        """
+        """
+        Examples: ubuntu release
+          | release |
+          | xenial  |
+          | bionic  |
+          | focal   |
+          | kinetic |
+          | jammy   |
+          | lunar   |
--- a/uaclient/log.py
+++ b/uaclient/log.py
@@ -1,4 +1,7 @@
+import json
 import logging
+from collections import OrderedDict
+from typing import Any, Dict  # noqa: F401
 
 from uaclient import util
 
@@ -9,3 +12,49 @@ class RedactionFilter(logging.Filter):
     def filter(self, record: logging.LogRecord):
         record.msg = util.redact_sensitive_logs(str(record.msg))
         return True
+
+
+class JsonArrayFormatter(logging.Formatter):
+    """Json Array Formatter for our logging mechanism
+    Custom made for Pro logging needs
+    """
+
+    default_time_format = "%Y-%m-%dT%H:%M:%S"
+    default_msec_format = "%s.%03d"
+    required_fields = (
+        "asctime",
+        "levelname",
+        "name",
+        "funcName",
+        "lineno",
+        "message",
+    )
+
+    def format(self, record: logging.LogRecord) -> str:
+        record.message = record.getMessage()
+        record.asctime = self.formatTime(record)
+
+        extra_message_dict = {}  # type: Dict[str, Any]
+        if record.exc_info:
+            extra_message_dict["exc_info"] = self.formatException(
+                record.exc_info
+            )
+        if not extra_message_dict.get("exc_info") and record.exc_text:
+            extra_message_dict["exc_info"] = record.exc_text
+        if record.stack_info:
+            extra_message_dict["stack_info"] = self.formatStack(
+                record.stack_info
+            )
+        extra = record.__dict__.get("extra")
+        if extra and isinstance(extra, dict):
+            extra_message_dict.update(extra)
+
+        # is ordered to maintain order of fields in log output
+        local_log_record = OrderedDict()  # type: Dict[str, Any]
+        # update the required fields in the order stated
+        for field in self.required_fields:
+            value = record.__dict__.get(field)
+            local_log_record[field] = value
+
+        local_log_record["extra"] = extra_message_dict
+        return json.dumps(list(local_log_record.values()))
--- a/uaclient/cli.py
+++ b/uaclient/cli.py
@@ -49,7 +49,7 @@ from uaclient.api.u.pro.security.status.reboot_required.v1 import (
 )
 from uaclient.apt import AptProxyScope, setup_apt_proxy
 from uaclient.data_types import AttachActionsConfigFile, IncorrectTypeError
-from uaclient.defaults import DEFAULT_LOG_FORMAT, PRINT_WRAP_WIDTH
+from uaclient.defaults import PRINT_WRAP_WIDTH
 from uaclient.entitlements import (
     create_enable_entitlements_not_found_message,
     entitlements_disable_order,
@@ -64,6 +64,7 @@ from uaclient.entitlements.entitlement_status import (
 from uaclient.files import notices, state_files
 from uaclient.files.notices import Notice
 from uaclient.jobs.update_messaging import refresh_motd, update_motd_messages
+from uaclient.log import JsonArrayFormatter
 
 NAME = "pro"
 
@@ -1835,7 +1836,6 @@ def setup_logging(console_level, log_level, log_file=None, logger=None):
         cfg = config.UAConfig()
         log_file = cfg.log_file
     console_formatter = util.LogFormatter()
-    log_formatter = logging.Formatter(DEFAULT_LOG_FORMAT)
     if logger is None:
         # Then we configure the root logger
         logger = logging.getLogger()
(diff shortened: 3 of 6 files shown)


uaclient/log.py
```diff
@@ -0,0 +1,59 @@
+import json
+import logging
+import re
+from collections import OrderedDict
+from typing import Any, Dict, List
+
+
+class JsonArraryFormatter(logging.Formatter):
```
orndorffgrant, line 8:
```suggestion
class JsonArrayFormatter(logging.Formatter):
```

uaclient/log.py
```diff
@@ -0,0 +1,59 @@
...
+        # We only support the % style formatting...currently
+        if isinstance(self._style, logging.PercentStyle):
+            formatter_style_pattern = re.compile(r"%\((.+?)\)", re.IGNORECASE)
+        else:
+            raise ValueError("Invalid format: %s" % self._fmt)
+
+        if self._fmt:
+            return formatter_style_pattern.findall(self._fmt)
+        else:
+            return []
```
orndorffgrant, line 27: This would be useful if we were writing this as a library, but we only need it to work for our specific use case.

I suggest we just hard-code our list of fields somewhere in this file rather than parsing them from the _fmt string.

uaclient/log.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Any, Dict, List
+
+
+class JsonArraryFormatter(logging.Formatter):
+    default_time_format = "%Y-%m-%dT%H:%M:%S%z"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if not self.datefmt:
+            self.datefmt = self.default_time_format
```
orndorffgrant, line 14: Does this need to happen in the constructor?
We can just include this logic in-line in the `format` function.
Or we can just hardcode the date format since we don't have to support a configuration for it

uaclient/log.py
```diff
@@ -0,0 +1,59 @@
...
+            formatter_style_pattern = re.compile(r"%\((.+?)\)", re.IGNORECASE)
+        else:
+            raise ValueError("Invalid format: %s" % self._fmt)
+
+        if self._fmt:
+            return formatter_style_pattern.findall(self._fmt)
+        else:
+            return []
+
+    def update_local_log_records(
```
orndorffgrant, line 29: I don't think this needs to be a separate function. Let's just do it down in `format`

uaclient/log.py
```diff
@@ -0,0 +1,59 @@
...
+        record: logging.LogRecord,
+        local_log_record: Dict[str, Any],
+        message_dict: Dict[str, Any],
+    ):
+        # update the required fields in the order stated
+        for field in self._required_fields:
+            local_log_record[field] = record.__dict__.get(field)
+
+        # update the exc and stack info
+        local_log_record.update(message_dict)
```
orndorffgrant, line 40: The message_dict contains message-specific fields that don't occur in all messages right?
In that case, I think it should be it's own field and formatted as a json object.

For example:
```json
["EXCEPTION", "datetime",  "here is the actual log message", {"exc_info": "stuff here", "stack_info": "lots\nof\nlines\nhere"}]
```
Note that my example doesn't include all the fields that we should include - it only illustrates how I think the message_dict shuold get included - see the last item in the example.

orndorffgrant (commented):
Just like we did for the redaction filter, let's make this a self-contained PR. Can you modify the setup_logging functions to use the JsonFormatter?

uaclient/log.py
CalvoM: Okay, I had this question...should it be generic or custom made to fit our use case...thanks for answering.

uaclient/log.py
CalvoM: You are answering all questions I had :-) :-)

uaclient/log.py
CalvoM: Okay, so that it is in line with how we handle the `extra` field.

CalvoM: I think I will move the `extra` field to the message_dict, as well since it is not always there.

CalvoM: Here is the difference with the optional extra vs having extra always https://paste.ubuntu.com/p/34tSJmb79C/

uaclient/log.py
```diff
@@ -0,0 +1,57 @@
...
+        "lineno",
+        "levelname",
+        "message",
+        "extra",
+    )
+
+    def format(self, record: logging.LogRecord) -> str:
+        # format the time if is  present
+        record.message = record.getMessage()
+        if "asctime" in self.required_fields:
```
lucasmoura, line 26: I think we can drop this check, as we have discussed that all field should be present on an array entry

lucasmoura (commented):
Maybe we can also add an integration test that tries parsing the log file just to see that everything is working as expected.

That integration test can create scenarios where we have errors while running the command too, just to certify that the logs are correct.

What do you think @orndorffgrant @renanrodrigo ?

uaclient/tests/test_cli_api.py
```diff
@@ -24,7 +24,8 @@
 
 
 class TestActionAPI:
-    def test_api_help(self, capsys):
+    @mock.patch("uaclient.cli.entitlements.valid_services", return_value=[])
```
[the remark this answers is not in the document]
renanrodrigo: If we think this is a leak, this should stay; I closed the other PR that was bringing only this change

renanrodrigo:
+ on at least one integration test, with at least one failure and a regular operation log parsing

features/logs.feature
```diff
@@ -0,0 +1,23 @@
+Feature: Logs in Json Array Formatter
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    Scenario Outline: The log file can be successfully parsed as json array
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `apt update` with sudo
+        And I run `apt upgrade -y` with sudo
```
lucasmoura, line 8: I think we can remove that step here, just to make the test faster

features/logs.feature
```diff
@@ -0,0 +1,23 @@
...
+        And I run `apt install jq -y` with sudo
+        And I verify that running `pro status` `with sudo` exits `0`
+        And I verify that running `pro enable test_entitlement` `with sudo` exits `1`
+        And I run shell command `cat /var/log/ubuntu-advantage.log | jq -r .` as non-root
```
lucasmoura, line 12: Let's also attach the machine here, just to make the test more complete

uaclient/log.py
```diff
@@ -9,3 +12,53 @@ class RedactionFilter(logging.Filter):
...
+        "name",
+        "funcName",
+        "lineno",
+        "levelname",
+        "message",
+        "extra",
+    )
+
+    def format(self, record: logging.LogRecord) -> str:
+        # format the time if is  present
```
orndorffgrant, line 34: I think this comment can be removed?
```suggestion
```

uaclient/log.py
```diff
@@ -9,3 +12,53 @@ class RedactionFilter(logging.Filter):
...
+
+    default_time_format = "%Y-%m-%dT%H:%M:%S%z"
+    required_fields = (
+        "asctime",
+        "name",
+        "funcName",
+        "lineno",
+        "levelname",
+        "message",
+        "extra",
```
orndorffgrant, line 30: removing "extra" and treating it specially would make the loop down below simpler
```suggestion
```

uaclient/log.py
```diff
@@ -9,3 +12,53 @@ class RedactionFilter(logging.Filter):
...
+    """
+
+    default_time_format = "%Y-%m-%dT%H:%M:%S%z"
+    required_fields = (
+        "asctime",
+        "name",
+        "funcName",
+        "lineno",
+        "levelname",
+        "message",
```
orndorffgrant, line 29: I think levelname should go earlier to make ERROR and WARNING more visible
```suggestion
        "asctime",
        "levelname",
        "name",
        "funcName",
        "lineno",
        "message",
```

uaclient/log.py
```diff
@@ -9,3 +12,53 @@ class RedactionFilter(logging.Filter):
...
+            value = record.__dict__.get(field)
+            if field != "extra":
+                local_log_record[field] = value
+            else:
+                # Save extra, non-exception messages
+                # in another object
+                if value:
+                    extra_message_dict[field] = value
+
+        local_log_record.update({"extra_info": extra_message_dict})
```
orndorffgrant, line 63: Since this is adding a single field, we don't need to use `update`
```suggestion
        local_log_record["extra"] = extra_message_dict
```

uaclient/log.py
```diff
@@ -9,3 +12,53 @@ class RedactionFilter(logging.Filter):
...
         record.msg = util.redact_sensitive_logs(str(record.msg))
         return True
+
+
+class JsonArrayFormatter(logging.Formatter):
+    """Json Array Formatter for our logging mechanism
+    Custom made for Pro logging needs
+    """
+
+    default_time_format = "%Y-%m-%dT%H:%M:%S%z"
```
orndorffgrant, line 22: Let's add microseconds here
```suggestion
    default_time_format = "%Y-%m-%dT%H:%M:%S.%f%z"
```

orndorffgrant (approved):
Just nitpicks at this point so going ahead and approving

orndorffgrant:
We could bring the test leaks fix here or give it it's own PR - either way

uaclient/log.py
CalvoM: `%f` does not work, but there is another way of using `record.msecs` with `record.asctime` which I have used.

CalvoM:
@renanrodrigo @orndorffgrant @lucasmoura I will create a separate PR. Adding it here, will make this PR, go longer.

uaclient/log.py
```diff
@@ -9,3 +12,49 @@ class RedactionFilter(logging.Filter):
...
+            )
+        if not extra_message_dict.get("exc_info") and record.exc_text:
+            extra_message_dict["exc_info"] = record.exc_text
+        if record.stack_info:
+            extra_message_dict["stack_info"] = self.formatStack(
+                record.stack_info
+            )
+        extra = record.__dict__.get("extra")
+        if extra:
+            extra_message_dict["extra"] = extra
```
orndorffgrant, line 50: One small thing:
I don't think we want the extra object to look like `{"extra": { "some_field": "some_val"}}` - I think we want it to look like `{"some_field": "some_val"}`

which would make this part look like:
```suggestion
        extra = record.__dict__.get("extra")
        if extra and isinstance(extra, dict):
            extra_message_dict.update(extra)
```

orndorffgrant (commented):
latest update introduced a small bug, and a clarification to my earlier nit :upside_down_face:

uaclient/log.py
CalvoM: Makes sense.

orndorffgrant (approved):
Thanks for the final tweaks! merging

PR #2466, opened 2023-03-01
fix: verify if package can be installed

lucasmoura:
## Proposed Commit Message
fix: verify if package can be installed

When running a pro fix command, we check if the fixed package version is higher than the one currently installed. If it is, we try to update the package, but we don't check if the package can indeed be upgrade. For example, if the version that fix the error is higher than the package candidate version, we cannot install that fix. To address that, We are now checking the target package candidate version to address those potential scenarios

LP: #2006705

## Test Steps
Run the new integration test for this scenario

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/security.py
+++ b/uaclient/security.py
@@ -58,6 +58,16 @@ ReleasedPackagesInstallResult = NamedTuple(
 )
 
 
+BinaryPackageFix = NamedTuple(
+    "BinaryPackageFix",
+    [
+        ("source_pkg", str),
+        ("binary_pkg", str),
+        ("fixed_version", str),
+    ],
+)
+
+
 @enum.unique
 class FixStatus(enum.Enum):
     """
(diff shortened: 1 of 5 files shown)


uaclient/security.py
```diff
@@ -1379,17 +1422,23 @@ def upgrade_packages_and_attach(
...
-        )
+        try:
+            apt.run_apt_update_command()
+            apt.run_apt_command(
+                cmd=["apt-get", "install", "--only-upgrade", "-y"]
+                + upgrade_pkgs,
+                env={"DEBIAN_FRONTEND": "noninteractive"},
+                capture=False,
+            )
+        except Exception as e:
```
renanrodrigo, line 1439: except Exception could be improved here, if in `run_apt*_command` we capture the subp error and raise a more specific one

renanrodrigo (approved):
LGTM overall.
We should have a better way of getting the apt/apt_pkg/esm caches, but that's subject to another PR

orndorffgrant (approved):
LGTM I think it just needs a rebase to resolve conflicts

uaclient/security.py
lucasmoura: I think the idea here is to be more resilient to any possible exception. For example an error that is not caught by the `ProcessExecutionError`

orndorffgrant (approved):

PR #2467, opened 2023-03-01
realtime-kernel: remove `ubuntu-realtime` package on disable

orndorffgrant:
## Test Steps

Run the existing fips_vm tests
Run the edited realtime test
Test the UX of enabling and disabling realtime in a VM

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -319,17 +319,9 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         )
         remove_packages = fips_metapackage.intersection(installed_packages)
         if remove_packages:
-            env = {"DEBIAN_FRONTEND": "noninteractive"}
-            apt_options = [
-                '-o Dpkg::Options::="--force-confdef"',
-                '-o Dpkg::Options::="--force-confold"',
-            ]
-            apt.run_apt_command(
-                ["apt-get", "remove", "--assume-yes"]
-                + apt_options
-                + list(remove_packages),
+            apt.remove_packages(
+                list(fips_metapackage),
                 messages.DISABLE_FAILED_TMPL.format(title=self.title),
-                env=env,
             )
 
     def _perform_enable(self, silent: bool = False) -> bool:
--- a/uaclient/messages.py
+++ b/uaclient/messages.py
@@ -823,8 +823,18 @@ Do you want to continue? [ default = Yes ]: (Y/n) """.format(
     bold=TxtColor.BOLD, end_bold=TxtColor.ENDC
 )
 REALTIME_PRE_DISABLE_PROMPT = """\
-This will disable Ubuntu Pro updates to the Real-time kernel on this machine.
-The Real-time kernel will remain installed.\
+This will remove the boot order preference for the Real-time kernel and
+disable updates to the Real-time kernel.
+
+This will NOT fully remove the kernel from your system.
+
+After this operation is complete you must:
+  - Ensure a different kernel is installed and configured to boot
+  - Reboot into that kernel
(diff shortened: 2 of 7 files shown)


uaclient/messages.py
```diff
@@ -823,8 +833,8 @@ class TxtColor:
     bold=TxtColor.BOLD, end_bold=TxtColor.ENDC
 )
 REALTIME_PRE_DISABLE_PROMPT = """\
-This will disable Ubuntu Pro updates to the Real-time kernel on this machine.
-The Real-time kernel will remain installed.\
+This will uninstall the Real-time kernel package.
+You are responsible for configuring a kernel to boot after this action.
```
lucasmoura, line 837: Should we better alert that the realtime-kernel will not indeed be uninstalled ? Just to highlight to users that they really to manually change the kernel configuration

renanrodrigo: Agreed; imho this message could be either more explanatory or "scarier".

renanrodrigo (commented):
3 maybe stupid questions:
- maybe not related to this PR, as this is part of enabling RT instead of disabling: If `livepatch` is enabled and we enable `realtime-kernel`, then all calls to `pro status` are showing `canonical-livepatch returned error when checking kernel support`; should we show this?
- after disabling `realtime-kernel`, the `ubuntu-realtime` package is uninstalled, but the dependencies persist; a call to `sudo apt autoremove` removed `linux-headers-realtime` and `linux-realtime`. Is it the expected behavior?
- After disabling and uninstalling the packages, if we enable livepatch again, the status shows `server check-in: failed: livepatch check failed: unauthorized`. Is this also expected?

uaclient/entitlements/fips.py
```diff
@@ -313,23 +313,15 @@ def remove_packages(self) -> None:
...
-                '-o Dpkg::Options::="--force-confold"',
-            ]
-            apt.run_apt_command(
-                ["apt-get", "remove", "--assume-yes"]
-                + apt_options
-                + list(remove_packages),
-                messages.DISABLE_FAILED_TMPL.format(title=self.title),
-                env=env,
+        try:
+            apt.remove_packages(list(fips_metapackage))
```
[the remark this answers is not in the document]
lucasmoura: Indeed, that is not a cause of concern

orndorffgrant:
>     * maybe not related to this PR, as this is part of enabling RT instead of disabling: If `livepatch` is enabled and we enable `realtime-kernel`, then all calls to `pro status` are showing `canonical-livepatch returned error when checking kernel support`; should we show this?

Arg that's a bug and I just confirmed it on my end. I'll make sure console logging is disabled for that.

>     * after disabling `realtime-kernel`, the `ubuntu-realtime` package is uninstalled, but the dependencies persist; a call to `sudo apt autoremove` removed `linux-headers-realtime` and `linux-realtime`. Is it the expected behavior?

I think that is what we want? Let's see what the behavior is after we move this back to a `apt-get remove` subp call.

>     * After disabling and uninstalling the packages, if we enable livepatch again, the status shows `server check-in: failed: livepatch check failed: unauthorized`. Is this also expected?

I see that every now and then and don't think it is related to us - but I could be wrong

orndorffgrant:
@lucasmoura @renanrodrigo I fixed the livepatch status warning to not print to the console and made the message more explanatory - LMK what you think

lucasmoura (approved):
@orndorffgrant even though this is unrelated to that PR, I have noticed that we are not emitting a reboot notice when enabling the RT kernel. We can of course address this in a follow up PR

renanrodrigo (approved):

PR #2470, opened 2023-03-03
docs: cosmetic improvements

s-makin:
```
A series of cosmetic docs changes that will make them look more like the
documentation produced on Ubuntu.com. It's not *exactly* the same, but
close enough to make it look recognisable as Canonical docs.

I made the "Have a question" link (top of the page) more prominent and
easier to spot. Hacky or not, it's a great feature that most docs miss
and I want people to be able to find it!

Although I tried to test this on a variety of screen sizes, if you spot
anything that doesn't render right or looks weird, please let me know. The
default font sizes were problematic from an accessibility pov so I made
them larger. They are still variable so they should render right on smaller
screens but it's possible I missed some.

Code literals are now the same size as regular font for the same reason,
but made slightly bolded to help pick them out from text (which should
help e.g. when commands are included as inline text and not a code block.

For easier navigation, I've given the sidebar headings a higher weight, and
as an added bonus, when scrolling through a long page the right-hand side bar
also shows what part you're on.

For accessibility, I changed the weight and spacing between topics in the
side bars to help separate them and make it easier to tell where one link
ends and the next begins.

As always, let me know if there's any other improvements I can make or any
things I've done that don't work!  
```

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -1,0 +1,240 @@
+/** Fix the font weight (300 for normal, 400 for slightly bold) **/
+/** Should be 100 for all headers, 400 for normal text  **/
+
+h1, h2, h3, h4, h5, h6, .sidebar-tree .current-page>.reference, button, input, optgroup, select, textarea, th.head {
+    font-weight: 200;
+}
+
+.toc-title {
+    font-weight: 400;
+}
+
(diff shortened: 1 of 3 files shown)


renanrodrigo (approved):
I think there are a couple ways to restructure the selectors but this works fine; lgtm

PR #2471, opened 2023-03-03
Temporary DGX release

orndorffgrant:
**DO NOT MERGE**

Opening as a draft PR so we can easily review the diff for this special release.

We'll want to cherry-pick these changes over to `main` after they're finalized.

Test builds are hosted in [`ppa:orndorffgrant/proclient-nvidia-test`](https://launchpad.net/~orndorffgrant/+archive/ubuntu/proclient-nvidia-test)

Final build will go in TODO

### DIFF
--- a/features/motd_messages.feature
+++ b/features/motd_messages.feature
@@ -284,6 +284,26 @@ Feature: MOTD Messages
 
         [\w\d.]+
         """
+
+        # detach and pretend to be DGX OS to check the special DGX message
+        When I create the file `/tmp/machine-token-overlay.json` with the following:
+        """
+        {}
+        """
+        When I run `pro detach --assume-yes` with sudo
+        When I run `touch /etc/dgx-release` with sudo
+        When I wait `1` seconds
+        When I run `pro refresh messages` with sudo
+        And I run `run-parts /etc/update-motd.d/` with sudo
+        Then stdout matches regexp:
+        """
+        [\w\d.]+
+
+        Your DGX contract entitles you to ESM updates.
+        Please contact your NVIDIA account manager to get your Pro subscription.
+
+        [\w\d.]+
+        """
         Examples: ubuntu release
            | release | service   |
            | xenial  | esm-infra |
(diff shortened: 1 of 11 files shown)


uaclient/system.py
```diff
@@ -518,3 +518,34 @@ def ensure_folder_absent(folder_path: str) -> None:
...
+
+
+def is_dgx_os() -> bool:
+    return os.path.exists("/etc/dgx-release")
+
+
+def is_dgx_smbios() -> bool:
+    try:
+        baseboard_man_out, _ = subp(
+            ["dmidecode", "-s", "baseboard-manufacturer"]
```
dannf, line 530: We probably need to add a runtime dependency on the `dmidecode` package.

cpaelzer: We need to keep dependencies small as this is also in ubuntu-minimal and many things - like dmidecode are not.

It is in "ubuntu-standard" so we can assume most Ubuntu and derivatives have it.
A better fallback would be "if no dmidecode is around, check sysfs".

We lack a DGX system atm, @dannf could you check if there is anything reasonable for that in `/sys/class/dmi/id/*` ?

dannf: Good call @cpaelzer  - looks like the same values are available in `/sys/class/dmi/id/board_name` and `/sys/class/dmi/id/board_vendor`:

```
ubuntu@exotic-skunk:~$ cat /sys/class/dmi/id/board_name
DGX-1 with V100
ubuntu@exotic-skunk:~$ cat /sys/class/dmi/id/board_vendor
NVIDIA
```

Verified with bionic on a DGX-1 and jammy on a DGX A100 (oldest system/oldest supported kernel + newest system/newest supported kernel). So +1 on using that instead of `dmidecode`.

features/motd_messages.feature
```diff
@@ -282,6 +282,26 @@ Feature: MOTD Messages
...
+        When I wait `1` seconds
+        When I run `pro refresh messages` with sudo
+        And I run `run-parts /etc/update-motd.d/` with sudo
+        Then stdout matches regexp:
+        """
+        [\w\d.]+
+
+        Your DGX contract entitles you to ESM updates.
+        Please contact your NVIDIA account manager to get your Pro subscription.
+
```
lucasmoura, line 304: @orndorffgrant should we also check that we remove both message after the user attaches to a subscription ?

orndorffgrant:
Doesn't seem like we'll need this so closing. We can always bring it back if needed.

PR #2476, opened 2023-03-08
livepatch: using support status in reboot-required (SC-1307)

orndorffgrant:
## Test Steps

Everything should still pass. Including existing livepatch.feature test

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -221,17 +221,11 @@ class LivepatchEntitlement(UAEntitlement):
         if not livepatch.is_livepatch_installed():
             return (ApplicationStatus.DISABLED, messages.LIVEPATCH_NOT_ENABLED)
 
-        try:
-            system.subp(
-                [livepatch.LIVEPATCH_CMD, "status"],
-                retry_sleeps=LIVEPATCH_RETRIES,
-            )
-        except exceptions.ProcessExecutionError as e:
+        if livepatch.status() is None:
             # TODO(May want to parse INACTIVE/failure assessment)
-            logging.debug("Livepatch not enabled. %s", str(e))
             return (
                 ApplicationStatus.DISABLED,
-                messages.NamedMessage(name="", msg=str(e)),
+                messages.LIVEPATCH_APPLICATION_STATUS_CLIENT_FAILURE,
             )
         return status
 
--- a/uaclient/security_status.py
+++ b/uaclient/security_status.py
@@ -266,6 +266,7 @@ def get_reboot_status():
         and our_kernel_version == lp_status.kernel
         and lp_status.livepatch is not None
         and lp_status.livepatch.state == "applied"
+        and lp_status.supported == "supported"
     ):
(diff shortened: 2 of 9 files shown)


orndorffgrant:
not for the march release

uaclient/entitlements/livepatch.py
```diff
@@ -221,17 +221,14 @@ def application_status(
...
-        except exceptions.ProcessExecutionError as e:
+        if livepatch.status() is None:
             # TODO(May want to parse INACTIVE/failure assessment)
-            logging.debug("Livepatch not enabled. %s", str(e))
             return (
                 ApplicationStatus.DISABLED,
-                messages.NamedMessage(name="", msg=str(e)),
+                messages.NamedMessage(
+                    name="livepatch-client-failure",
+                    msg="canonical-livepatch status didn't finish successfully",
```
lucasmoura: This should in the `messages` module right ? Since we might show that message on the JSON output of  `enable` or `disable`

lucasmoura (approved):

uaclient/entitlements/livepatch.py
orndorffgrant: You're right, and I did that but accidentally put that change in the wrong commit :facepalm: I'll fix it as I rebase to fix conflicts presently

orndorffgrant: Updated!

uaclient/security_status.py
```diff
@@ -266,6 +266,7 @@ def get_reboot_status():
         and our_kernel_version == lp_status.kernel
         and lp_status.livepatch is not None
         and lp_status.livepatch.state == "applied"
+        and lp_status.supported == "supported"
```
orndorffgrant, line 269: @kian99 Can you confirm that if these two conditions are true, then we can give our "in-between" reboot status (indicating livepatch has the kernel covered for high/critical CVEs)?

kian99: Yes, according to the original UA [spec](https://docs.google.com/document/d/1Y9qs8JQAOFleB9qebAgD_J1YJ_UDl67zlvTFbiLuTYw/edit#) you only want to give that "in-between" status when patches have been applied and the kernel is still within support so this is the right way to check for that.

I didn't perhaps have the insight to mention this previously but a value of `lp_status.livepatch.state == "nothing-to-apply"` alongside `lp_status.supported == "supported"` would indicate that livepatches will come through for high/critical CVEs even though none have come through yet. Just mentioning this as it might be something you want down the road (a fair portion of the time, machines won't actually have Livepatches applied because there are no high/critical CVEs but the system will still receive patches).

orndorffgrant: Thank you @kian99!

That's an excellent point about the `nothing-to-apply` state. That seems to be the same situation and I maybe we should allow that state to trigger the "in-between" reboot status as well. @lucasmoura Thoughts? (might be worth a quick email and/or sync to make sure all stake holders are okay with that)

orndorffgrant:
The CI failures are unrelated

PR #2478, opened 2023-03-09
livepatch: display tailored warnings for new support statuses (SC-1538)

orndorffgrant:
## Test Steps

The new integration test should pass.
Notice that the older integration test -- the first one in livepatch.feature -- is failing with the new implementation. This is exposing a bug in livepatch where unsupported kernels are being reported as "unknown" support by the livepatch-client. So it is "correct" to be failing :/

## Checklist

 - [x] Yes - livepatch team - done
 - [ ] No

### DIFF
--- a/features/livepatch.feature
+++ b/features/livepatch.feature
@@ -76,6 +76,56 @@ Feature: Livepatch
             | release |
             | focal   |
 
+    @series.focal
+    @uses.config.machine_type.any
+    @uses.config.machine_type.gcp.generic
+    Scenario Outline: Attached livepatch status shows upgrade required when on an old kernel
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token_staging` with sudo
+        When I run `apt-get install linux-headers-<old_kernel_version> linux-image-<old_kernel_version> -y` with sudo
+        When I run `DEBIAN_FRONTEND=noninteractive apt-get remove linux-image*-gcp -y` with sudo
+        When I run `update-grub` with sudo
+        When I reboot the machine
+        When I run `uname -r` with sudo
+        Then stdout contains substring:
+        """
+        <old_kernel_version>
+        """
+        When I run `pro status` with sudo
+        Then stdout matches regexp:
+        """
+        livepatch +yes +warning +Canonical Livepatch service
+        """
+        Then stdout contains substring:
+        """
+        NOTICES
+        The running kernel has reached the end of its active livepatch window.
+        Please upgrade the kernel with apt and reboot for continued livepatch support.
+
+        """
+        When I run `apt-get install linux-headers-generic linux-image-generic -y` with sudo
+        When I reboot the machine
+        When I run `uname -r` with sudo
+        Then stdout does not contain substring:
+        """
+        <old_kernel_version>
+        """
+        When I run `pro status` with sudo
+        Then stdout matches regexp:
+        """
+        livepatch +yes +enabled +Canonical Livepatch service
+        """
+        Then stdout does not contain substring:
+        """
+        NOTICES
+        The running kernel has reached the end of its active livepatch window.
+        Please upgrade the kernel with apt and reboot for continued livepatch support.
+
+        """
+        Examples: ubuntu release
+            | release | machine_type | old_kernel_version |
+            | focal   | gcp.generic  | 5.4.0-28-generic   |
+
     @series.kinetic
     @series.lunar
     @uses.config.machine_type.lxd-vm
--- a/features/livepatch.feature
+++ b/features/livepatch.feature
@@ -2,9 +2,10 @@
 Feature: Livepatch
 
     @series.focal
+    @uses.config.machine_type.any
     @uses.config.machine_type.lxd-vm
     Scenario Outline: Attached livepatch status shows warning when on unsupported kernel
-        Given a `<release>` machine with ubuntu-advantage-tools installed
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
         Then I verify that no files exist matching `/home/ubuntu/.cache/ubuntu-pro/livepatch-kernel-support-cache.json`
         When I run `pro status` as non-root
         Then I verify that files exist matching `/home/ubuntu/.cache/ubuntu-pro/livepatch-kernel-support-cache.json`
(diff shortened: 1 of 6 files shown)


orndorffgrant:
@kian99 I preemptively started working on integrating with the new livepatch support statuses in this PR, but I probably won't continue until I can start testing end-to-end. LMK when the new support statuses are available in the livepatch-client (even if it's in an edge channel or similar).

kian99:
> @kian99 I preemptively started working on integrating with the new livepatch support statuses in this PR, but I probably won't continue until I can start testing end-to-end. LMK when the new support statuses are available in the livepatch-client (even if it's in an edge channel or similar).

@orndorffgrant Will do! A beta version with the new support statuses is currently available in `latest/edge` though we still have one bug. When the `uname -a` call shows a truncated kernel build date bc of this [bug](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2008084), we try to grab the build date by "stat"ing the kernel deb package changelog file, and it seems the snap doesn't have permission to access that folder. So with that ironed out hopefully a final release of the client will be done this week.

orndorffgrant:
@kian99 Sorry I haven't checked in in a while, is the livepatch-client support for the kernel-upgrade-required and other more granular support statuses released yet?
If so, is there an easy way to get a VM into a state where I can consistently get the kernel-upgrade-required support status?

kian99:
@orndorffgrant Heya, yes the client has been released with the granular support statuses, it's available from `latest/stable`. One unfortunate note is that our production server hasn't had a new release in a while as we're blocked due to a Juju bug so on production we're not sending the new granular support statuses down to clients. But in any case, staging is updated and available. So the easiest way to trigger the new states would be to launch a VM, attach to staging and switch to an old kernel.

There is a script to do that [here](https://github.com/kian99/livepatch-server/blob/manual-test-script/scripts/ci/manual-patchable-client.sh) (this is on my fork). It'll switch you to an old kernel for Focal where support has ended and that should get you what you need. Just a final note, we haven't yet updated the command `canonical-livepatch kernel-upgrade-required` so the most accurate thing to do is still to continue to parse `canonical-livepatch status --format json`

orndorffgrant:
@kian99 That is perfect! Thank you for the update - that totally unblocks me on this PR.

uaclient/messages.py
```diff
@@ -1202,6 +1202,19 @@ class TxtColor:
     "Value provided was not found in {enum_class}'s allowed: value: {values}",
 )
 
+LIVEPATCH_KERNEL_UPGRADE_REQUIRED = NamedMessage(
+    name="livepatch-kernel-upgrade-required",
+    msg="""\
+The running kernel has reached the end of its active livepatch window.
+Please upgrade the kernel with apt and reboot for continued livepatch service.""",  # noqa: E501
```
lucasmoura, line 1209: just a suggestion, we can change the message to `continued livepatch support` ?

features/livepatch.feature
```diff
@@ -73,14 +74,46 @@ Feature: Livepatch
...
+        Then stdout contains substring:
+        """
+        <old_kernel_version>
+        """
+        When I run `pro status` with sudo
+        Then stdout matches regexp:
+        """
+        livepatch +yes +warning +Canonical Livepatch service
+        """
+        Then stdout contains substring:
```
lucasmoura, line 100: Should we add the upgrade step just to guarantee that we are getting rid of the notice ?

features/livepatch.feature
```diff
@@ -73,14 +74,46 @@ Feature: Livepatch
...
-            | focal   |
+            | release | machine_type |
+            | focal   | lxd-vm       |
+
+    @series.focal
+    @uses.config.machine_type.any
+    @uses.config.machine_type.gcp.generic
+    Scenario Outline: Attached livepatch status shows upgrade required when on an old kernel
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+        When I attach `contract_token_staging` with sudo
```
lucasmoura, line 85: Do we need the staging token ?

features/livepatch.feature
orndorffgrant: yep good idea!

features/livepatch.feature
orndorffgrant: Yes, for now, because the production livepatch server doesn't support granular statuses yet

orndorffgrant:
@lucasmoura updated

lucasmoura (approved):

uaclient/livepatch.py
```diff
@@ -203,14 +213,20 @@ def is_kernel_supported(
...
+        return LivepatchSupport.SUPPORTED
+    if lp_status.supported == "kernel-upgrade-required":
+        return LivepatchSupport.KERNEL_UPGRADE_REQUIRED
+    if lp_status.supported == "kernel-end-of-life":
+        return LivepatchSupport.KERNEL_EOL
     if lp_status.supported == "unsupported":
-        return False
+        return LivepatchSupport.UNSUPPORTED
+    if lp_status.supported == "unknown":
+        return LivepatchSupport.UNKNOWN
```
orndorffgrant, line 229: @kian99 Can you confirm that these are the expected string values of the kernel support status that we could get from `canonical-livepatch status --format=json`?

uaclient/messages.py
```diff
@@ -1202,6 +1202,19 @@ class TxtColor:
...
+The running kernel has reached the end of its active livepatch window.
+Please upgrade the kernel with apt and reboot for continued livepatch support.""",  # noqa: E501
+)
+LIVEPATCH_KERNEL_EOL = FormattedNamedMessage(
+    name="livepatch-kernel-eol",
+    msg="""\
+The current kernel ({version}, {arch}) has reached the end of its livepatch support.
+Supported kernels are listed here: https://ubuntu.com/security/livepatch/docs/kernels
+Either switch to a supported kernel or `pro disable livepatch` to dismiss this warning.""",  # noqa: E501
+)
```
orndorffgrant, line 1217: @kian99 Can you confirm that these messages make sense for the "kernel-upgrade-required" and "kernel-end-of-life" support statuses, respectively?

orndorffgrant (commented):
Hi @kian99 I think the messaging for more granular support statuses is ready in this PR, can you just double check the parts I highlighted here for you to make sure they're accurate?

uaclient/messages.py
kian99: These look good to me.

uaclient/livepatch.py
kian99: Ah gotcha, sounds good.

renanrodrigo (approved):

PR #2479, opened 2023-03-10
readme changes - link to docs instead of a separate TOC

orndorffgrant:
## Test Steps

Our Readme TOC of docs was getting out of date, and now that we have rtd, we don't really need it.
I also added some new badges and made our existing badges link to more interesting things than the badge itself.

LMK if this does or doesn't make sense.

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,16 @@
 </h1>
 
 ###### Clean and Consistent CLI for your Ubuntu Pro Systems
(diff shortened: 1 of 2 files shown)


renanrodrigo (approved):
I think this is a good idea.
@s-makin to confirm or deny

s-makin:
> I think this is a good idea. @s-makin to confirm or deny

I think it's an excellent idea! Thanks for taking the initiative on that, Grant :)

s-makin (approved):

PR #2480, opened 2023-03-10
reboot_cmds refactor (SC-1304)

orndorffgrant:
no-lp no-gh

Part of my technical debt ticket was to simplify our systemd services. I think we still need this service and I think it should remain a separate service. But the it could use some updating to line up with our more recent patterns of file and status access.

## Test Steps

Make sure the reboot-cmds behavior hasn't changed. The new integration test _should_ cover this, but maybe do your own experiments too.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/lib/reboot_cmds.py
+++ b/lib/reboot_cmds.py
@@ -18,12 +18,18 @@ import logging
 import os
 import sys
 
-from uaclient import config, contract, defaults, exceptions, lock, messages
+from uaclient import (
+    config,
+    contract,
+    defaults,
+    exceptions,
+    lock,
+    messages,
+    system,
+)
 from uaclient.cli import setup_logging
 from uaclient.entitlements.fips import FIPSEntitlement
 from uaclient.files import notices
-from uaclient.files.notices import Notice
-from uaclient.system import subp
 
 # Retry sleep backoff algorithm if lock is held.
 # Lock may be held by auto-attach on systems with ubuntu-advantage-pro.
@@ -33,7 +39,7 @@ MAX_RETRIES_ON_LOCK_HELD = 7
 
 def run_command(cmd, cfg: config.UAConfig):
     try:
-        out, _ = subp(cmd.split(), capture=True)
+        out, _ = system.subp(cmd.split(), capture=True)
         logging.debug("Successfully executed cmd: {}".format(cmd))
     except exceptions.ProcessExecutionError as exec_error:
         msg = (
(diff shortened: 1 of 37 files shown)


lib/reboot_cmds.py
```diff
@@ -58,38 +59,27 @@ def run_command(cmd, cfg: config.UAConfig):
 
 
 def fix_pro_pkg_holds(cfg: config.UAConfig):
-    status_cache = cfg.read_cache("status-cache")
-    if not status_cache:
+    fips = FIPSEntitlement(cfg)
+    fips_status, _ = fips.application_status()
```
lucasmoura, line 37: One concern here is that we would be evaluating the fips status for the current setup. 
If the machine has all the fips package holds in place, but don't have the metapackage for some reason, we might not have the `/proc/sys/crypto/fips_enabled` file and show the service as disabled.

uaclient/upgrade_lts_contract.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
+
+"""
```
lucasmoura, line 3: Just double checking, what is the rationale for moving this into the uaclient module ?

orndorffgrant: Previously, the reboot_cmds script called `python3 /usr/lib/ubuntu-advantage/upgrade_lts_contract.py`. In the new version, this function is shared in uaclient, and so `reboot_cmds.py` imports the function and calls it instead of starting a subprocess. (If you're going one commit at a time, this happens in a later commit)

lib/reboot_cmds.py
orndorffgrant: I guess that's what I thought this was supposed to be doing. The previous version checked the status cache, which is the status as of the last time `pro status` got executed as root, which feels less accurate.

Or are you saying we want to always remove fips holds even if fips isn't enabled currently?

Maybe, I'm misunderstanding the intent of this section :upside_down_face:

uaclient/upgrade_lts_contract.py
lucasmoura: ohhh indeed. Good point, no blocks on that at all

lib/reboot_cmds.py
orndorffgrant: @lucasmoura updated!

lucasmoura:
@orndorffgrant CI tests are also failing locally. I think it is better to double check them

orndorffgrant:
Will do! Thanks for checking

orndorffgrant:
They didn't fail locally so I re-ran them and they passed - I'll say it was a flake this time and merge away - but let's continue to monitor

PR #2482, opened 2023-03-13
logging_revamp: fix unittests logging leak SC-1530

CalvoM:
## Proposed Commit Message

logging_revamp: fix unittests logging leak

We are having some issues with logging leaks in unittests.

no-lp no-gh no-sc

## Test Steps

Run the all the tests `tox -e py310-test`.
Run the tests in test_util.py `tox -e py310-test uaclient/tests/test_util.py`.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/conftest.py
+++ b/uaclient/conftest.py
@@ -154,6 +154,10 @@ def FakeConfig(tmpdir):
         ) -> None:
             if not cfg_overrides.get("data_dir"):
                 cfg_overrides.update({"data_dir": tmpdir.strpath})
+            if not cfg_overrides.get("log_file"):
+                cfg_overrides.update(
+                    {"log_file": tmpdir.join("log_file.log").strpath}
+                )
             super().__init__(
                 cfg_overrides,
                 user_config=UserConfigData(),
--- a/uaclient/conftest.py
+++ b/uaclient/conftest.py
@@ -85,7 +85,13 @@ def caplog_text(request):
     (It returns a function so that the requester can decide when to examine the
     logs; if it returned caplog.text directly, that would always be empty.)
     """
-    log_level = getattr(request, "param", logging.INFO)
+    cap_params = getattr(request, "param", logging.INFO)
+    log_filter = None
+    if isinstance(cap_params, tuple):
+        log_level = cap_params[0]
+        log_filter = cap_params[1]
+    else:
+        log_level = cap_params
     try:
         try:
             caplog = request.getfixturevalue("caplog")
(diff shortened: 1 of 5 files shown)


CalvoM (commented):
@lucasmoura @renanrodrigo I was able to solve the setup_logging mocking problem by mocking  `cfg.log_file` since it was the bone of contention.

uaclient/conftest.py
```diff
@@ -85,6 +85,13 @@ def caplog_text(request):
     logs; if it returned caplog.text directly, that would always be empty.)
     """
     log_level = getattr(request, "param", logging.INFO)
```
lucasmoura, line 87: It seems we don't need to create that variable anymore

lucasmoura (commented):
Just to confirm, we call the `setup_logging` function on other scripts, i.e `upgrade_lts_contract.py`, where we don't pass `cfg.log_file` as a parameter. Would we have the isolation issue when testing those scripts ?

CalvoM:
@lucasmoura We would have to mock the `UAConfig` object with the `FakeConfig` object, thus we will have no isolation issue. Even though it does not pass the `cfg.log_file` we default to it in `setup_logging` [here](https://github.com/canonical/ubuntu-pro-client/blob/43d519f413818073385039898246355ffac793f2/uaclient/cli.py#L1838)

orndorffgrant:
@CalvoM just want to make sure you saw my latest questions :)

CalvoM:
Oh! I updated the description but I forgot to update my comments. @orndorffgrant. For question 2, I have broken the commits into 2.

uaclient/conftest.py
```diff
@@ -144,6 +154,10 @@ def __init__(
         ) -> None:
             if not cfg_overrides.get("data_dir"):
                 cfg_overrides.update({"data_dir": tmpdir.strpath})
+            if not cfg_overrides.get("log_file"):
+                cfg_overrides.update(
+                    {"log_file": tmpdir.join("log_file.log").strpath}
+                )
```
orndorffgrant, line 160: I think this part is supposed to be in the second commit

orndorffgrant (approved):
Thanks for the explanation @CalvoM! I've verified that this works.

And thank you for splitting the commits! I just noticed one part of the diff that is currently in the first commit but should be in the second.

Pre-emptively approving but that should be fixed before merging.

uaclient/conftest.py
CalvoM: Updated @orndorffgrant .

orndorffgrant: Thanks!

PR #2483, opened 2023-03-13
dx: move all mypy config to pyproject.toml

aciba90:
## Proposed Commit Message

```
dx: move all mypy config to pyproject.toml

Improve developer experience by unifying mypy config to pyproject.toml
and extracting mypy config from tox.ini.

This change makes possible to run mypy as a standlone tool with the
same config as tox -e mypy has.
```

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (approved):
bless you

PR #2485, opened 2023-03-13
Release 27.14

orndorffgrant:
* Is the version correctly updated? We must ensure that the new version in the changelog is
  correct and it also targets the latest Ubuntu release at the moment.
* Is the entry useful for the user? The changelog entries should be user focused, meaning
  that we should only add entries that we think users will care about (i.e. we don't need
  entries when fixing a test, as this doesn't provide meaningful information to the user).
* Is this entry redundant? Sometimes we may have changes that affect separate modules of the
  code. We should have an entry only for the module that was most affected by it.
* Is the changelog entry unique? We need to verify that the changelog entry is not already
  reflected in an earlier version of the changelog. If it is, we need not only to remove but double
  check the process we are using to cherry-pick the commits.
* Is this entry actually reflected in the code? Sometimes, we can have changelog entries
  that are not reflected in the code anymore. This can happen during development when we are
  still unsure about the behaviour of a feature or when we fix a bug that removes the code
  that was added. We must verify each changelog entry that is added to be sure of their
  presence in the product.


debian/changelog
```diff
@@ -1,3 +1,38 @@
+ubuntu-advantage-tools (27.14~23.04.1) lunar; urgency=medium
```
orndorffgrant, line 1: @s-makin Please take at this draft of the changelog entry for our upcoming release.

debian/changelog
```diff
@@ -1,3 +1,38 @@
...
+
+  * d/ubuntu-advantage-tools.{postinst,postrm,preinst}:
+    - migrate certain settings out of uaclient.conf to a new file managed by
+      the pro config subcommand
+  * d/ubuntu-advantage-tools.postinst:
+    - refactor PREVIOUS_PKG_VER as a global variable
+    - simplify how we add notices
+  * api: new u.unattended_upgrades.status.v1 endpoint for querying status of
```
lucasmoura, line 9: I know this is a draft, but I think we should separate the debian changes from the features by using the:
`New upstream release 27.14`

orndorffgrant: Yep my bad, I'll update it

orndorffgrant: updated!

debian/changelog
s-makin: Will do!

s-makin (approved):

renanrodrigo (approved):

PR #2486, opened 2023-03-14
logging: setup logging with defaults before uacfg is read (SC-1428)

aciba90:
Early draft PR to gather feedback about the implementation.

## Proposed Commit Message

```
logging: setup logging with defaults before uacfg is read 

TODO
```

## Test Steps

```
$ ./tools/test-in-lxd.sh
$ pro status

# The first log entries come from loading the UAConfig and proxy staff from cli.py.
$ head /var/log/ubuntu-advantage.log
["2023-03-14T10:50:12.797", "DEBUG", "uaclient.config", "parse_config", 670, "Using client configuration file at /etc/ubuntu-advantage/uaclient.conf", {}]
["2023-03-14T10:50:12.797", "DEBUG", "root", "load_file", 373, "Reading file: /etc/ubuntu-advantage/uaclient.conf", {}]
["2023-03-14T10:50:12.804", "DEBUG", "root", "configure_web_proxy", 289, "Setting no_proxy: 169.254.169.254,[fd00:ec2::254],metadata", {}]
["2023-03-14T10:50:12.805", "DEBUG", "root", "main", 1950, "Executed with sys.argv: ['/usr/bin/pro', 'status']", {}]
...
```

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/uaclient/log.py
+++ b/uaclient/log.py
@@ -1,9 +1,14 @@
+import enum
 import json
 import logging
+import pathlib
+import sys
 from collections import OrderedDict
+from typing import Optional  # noqa
 from typing import Any, Dict  # noqa: F401
 
-from uaclient import util
+from uaclient import config, util
+from uaclient.defaults import CONFIG_DEFAULTS
 
 
 class RedactionFilter(logging.Filter):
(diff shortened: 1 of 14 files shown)


uaclient/log.py
```diff
@@ -58,3 +63,87 @@ def format(self, record: logging.LogRecord) -> str:
...
+            return CONFIG_DEFAULTS["log_file"]
+        raise NotImplementedError("Implement me for {}".format(self))
+
+    def extract_log_file(self, ua_cfg: config.UAConfig) -> str:
+        if self == self.MAIN:
+            return ua_cfg.log_file
+        raise NotImplementedError("Implement me for {}".format(self))
+
+
+def setup_logging(console_level, log_level, log_file=None, logger=None):
```
aciba90, line 84: Being this used on more than the cli entry-point, I thought it was a good idea to move it here.

uaclient/log.py
```diff
@@ -58,3 +63,87 @@ def format(self, record: logging.LogRecord) -> str:
...
+            log_file_path.chmod(0o644)
+
+        file_handler = logging.FileHandler(log_file)
+        file_handler.setFormatter(JsonArrayFormatter())
+        file_handler.setLevel(log_level)
+        file_handler.set_name("ua-file")
+        logger.addHandler(file_handler)
+
+
+class EarlyLoggingSetup:
```
aciba90, line 122: The reasons to implement this as a context manager are:

* Perform the initial log config in the `__enter__`.
* During the lifetime of the CM, client code could potentially need the uacfg. If so, they can hand over it by setting `self.ua_cfg` and avoid multiple reads of the config.
* On __exit__ time, the CM will set up the final logging config.
* If, in the future, more state needs to be tracked between those two events, a CM is easier to maintain than two calls to free functions.

An example of usage: https://github.com/canonical/ubuntu-pro-client/pull/2486/files#diff-d558bdbcb7ee1bef85fd138367f1e64708ea8271c09cf8e4c19941c24c8883d6R1923

orndorffgrant: I like the idea, but I'm worried it's a premature design. IOW, I agree that "If more state needs to be tracked..." then this would be nice, but right now we don't have any extra state. So the simplicity of directly calling setup_logging twice with different parameters is appealing.
Maybe I'm missing other advantages though - what do you think?

aciba90: Yeah, I do agree, probably overengineering here. I am going to close this PR and I will re-open it with a simplified version, thanks!

PR #2488, opened 2023-03-14
security_status: use cache mechanism for apt.Cache

lucasmoura:
## Proposed Commit Message
security_status: use cache mechanism for apt.Cache

There is problem on the apt library when we import and use both apt_pkg and apt in the same module. For example, once we create the esm cache, that affects the apt.Cache creation. To avoid that, we are now better isolating the apt.Cache creation to handle potential configuration changes caused by the esm cache

## Test Steps
Launch a xenial container and install a package that contains this fix.
After running `apt-get update` on the container, run the following script and check that all results should be consistent:

```python
from uaclient.api.u.pro.packages.updates.v1 import updates
from uaclient.system import subp

result = updates()
print(result.summary)
result = updates()
print(result.summary)
subp(["apt-get", "upgrade", "-y"])
result = updates()
print(result.summary)

```

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -212,14 +212,14 @@ def get_apt_cache_policy(
 
 
 @lru_cache(maxsize=None)
-def _get_apt_cache():
+def get_apt_cache():
     import apt  # type: ignore
 
     return apt.Cache()
 
 
 def get_pkg_candidate_version(pkg: str) -> Optional[str]:
-    cache = _get_apt_cache()
+    cache = get_apt_cache()
 
     try:
         package = cache[pkg]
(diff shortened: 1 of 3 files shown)


orndorffgrant (commented):
This doesn't look like it will solve the case of running the function once, then installing a package, then running the function again for updated results - which I think was the reporter's original issue.

orndorffgrant (commented):
@lucasmoura this looks like it could work (though I don't understand why the shared config was causing the problem). I think it's worth adding an integration test for this scenario though

renanrodrigo:
#2495 was merged as an alternative to this

PR #2490, opened 2023-03-16
tests: fixups discovered while testing lunar

orndorffgrant:
removing ca-certificates doesn't actually remove the file anymore

the log file may have old non-json lines at the beginning from test setup


aciba90 (approved):
LGTM!

PR #2494, opened 2023-03-17
logging: setup logging with defaults before the config is read (SC-1428)

aciba90:
Continuation of #2486.

## Proposed Commit Message

```
logging: setup logging with defaults before the config is read

Previously, we didn’t call setup_logging until after reading
the config (because the config can affect how we setup logging).

Set up the logging with defaults before reading the config to
capture logs from the start of the cli and lib entry-points.

Capture those default log files in apport.

data_types: warn on wrong data types.
```

## Test Steps

```
$ ./tools/test-in-lxd.sh
$ pro status

# The first log entries come from loading the UAConfig and proxy staff from cli.py.
$ head /var/log/ubuntu-advantage.log
["2023-03-14T10:50:12.797", "DEBUG", "uaclient.config", "parse_config", 670, "Using client configuration file at /etc/ubuntu-advantage/uaclient.conf", {}]
["2023-03-14T10:50:12.797", "DEBUG", "root", "load_file", 373, "Reading file: /etc/ubuntu-advantage/uaclient.conf", {}]
["2023-03-14T10:50:12.804", "DEBUG", "root", "configure_web_proxy", 289, "Setting no_proxy: 169.254.169.254,[fd00:ec2::254],metadata", {}]
["2023-03-14T10:50:12.805", "DEBUG", "root", "main", 1950, "Executed with sys.argv: ['/usr/bin/pro', 'status']", {}]
...
```

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/apport/source_ubuntu-advantage-tools.py
+++ b/apport/source_ubuntu-advantage-tools.py
@@ -2,6 +2,7 @@ import os
 import tempfile
 
 from apport.hookutils import attach_file_if_exists
+from uaclient import defaults
 from uaclient.actions import collect_logs
 from uaclient.config import UAConfig
 
@@ -12,7 +13,7 @@ def add_info(report, ui=None):
     cfg = UAConfig()
     with tempfile.TemporaryDirectory() as output_dir:
         collect_logs(cfg, output_dir)
-        auto_include_log_files = [
+        auto_include_log_files = {
             "cloud-id.txt",
             "cloud-id.txt-error",
             "ua-status.json",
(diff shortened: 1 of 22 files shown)


apport/source_ubuntu-advantage-tools.py
aciba90, line 1: Notice that the default logs were already being collected in `collect_logs`.

lib/esm_cache.py
```diff
@@ -18,6 +19,12 @@ def main(cfg: UAConfig) -> None:
 
 
 if __name__ == "__main__":
+    setup_logging(
+        logging.INFO,
+        logging.DEBUG,
+        defaults.CONFIG_DEFAULTS["log_file"],
+        logger=logging.getLogger(),
```
aciba90, line 26: I am mimicking https://github.com/canonical/ubuntu-pro-client/pull/2494/files#diff-6fea9b378e472552634bc16317f26aec387dd558c1754cd545806364476782d0R33 and setting the root logger, but this module uses a specific logger (line 10).

Is there a specific reason to do this, or should we set up the specific logger only?

orndorffgrant: This is right for now. One of the high priority logging revamp tasks @CalvoM has is to make all of our loggers use the `uaclient` logger and then all of these setup_logging calls will be adjusted to configure that instead of the root logger.

But this is doing the right thing until that fix happens.

orndorffgrant (approved):
Just tested it out and it works great!

PR #2496, opened 2023-03-17
experiment: custom pr check actions

orndorffgrant:
no-jira no-lp no-gh

Take a look at the logic and lmk if you think this would be useful.

We can expand on this action (or create new actions based on this template) to do more custom checks for things in the future like e.g. "if the diff is big, at least one commit message should contain a message > 3 lines" or something like that.

### DIFF
--- a/.github/actions/bug-refs/index.js
+++ b/.github/actions/bug-refs/index.js
@@ -1,0 +1,61 @@
+const core = require('@actions/core');
+const github = require('@actions/github');
+
+async function run() {
+    const context = github.context;
+    if (context.eventName !== "pull_request") {
+      console.log(
+        'The event that triggered this action was not a pull request, skipping.'
+      );
+      return;
+    }
+
+    const body = context.payload.pull_request.body.toLocaleUpperCase();
+    const ignoreJira = body.includes("NO-JIRA");
+    const ignoreLaunchpad = body.includes("NO-LP");
+    const ignoreGithub = body.includes("NO-GH");
+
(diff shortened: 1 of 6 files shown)


.github/actions/bug-refs/index.js
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    const launchpadPresent = false;
+    const githubPresent = false;
+    commits.data.forEach(commit => {
+        const message = commit.commit.message;
+        if (message.toLocaleUpperCase().includes("LP: #")) {
+            launchpadPresent = true;
+        }
+        if (message.toLocaleUpperCase().includes("FIXES: #")) {
```
aciba90, line 41: Per the release documentation, I think we use `log2dch`, which collects GitHub's issues referenced as `GH: #NNNN`: https://github.com/canonical/uss-tableflip/blob/3c02cd907c2c5bb799b6c88926e041b2e91f4069/scripts/log2dch#L218

Shouldn't we use and check for that references instead of `FIXES: #`?

aciba90: Thanks for the info. My bad there, I thought `log2dch` look for `GH`. Taking into account that this PR would work as-is in that regard, I leave it open to you to support `Closes` or not.

orndorffgrant: I went ahead and added it!

aciba90 (approved):
LGTM, thanks!

lucasmoura (approved):

PR #2501, opened 2023-03-21
apt: use context manager for handling apt config

lucasmoura:
## Proposed Commit Message
apt: use context manager for handling apt config

Some apt functions require us to change the global apt config. Since we might be changing behavior on other parts of the code that require apt functionality, we are now using a context manager to guarantee that the original system config is kept after we call an apt function

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -1,3 +1,4 @@
+import copy
 import datetime
 import enum
 import glob
@@ -214,6 +215,29 @@ def get_apt_cache_policy(
     )
 
 
+class PreserveAptCfg:
+    def __init__(self, apt_func):
+        self.apt_func = apt_func
+        self.current_apt_cfg = {}  # Dict[str, Any]
+
+    def __enter__(self):
+        cfg = apt_pkg.config
+        self.current_apt_cfg = {
+            key: copy.deepcopy(cfg.get(key)) for key in cfg.keys()
+        }
+
+        return self.apt_func()
+
+    def __exit__(self, type, value, traceback):
+        cfg = apt_pkg.config
+        # We need to restore the apt cache configuration after creating our
+        # cache, otherwise we may break people interacting with the
+        # library after importing our modules.
+        for key in self.current_apt_cfg.keys():
+            cfg.set(key, self.current_apt_cfg[key])
+        apt_pkg.init_system()
+
+
 def get_apt_cache():
     for key in apt_pkg.config.keys():
         apt_pkg.config.clear(key)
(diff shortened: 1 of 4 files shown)


uaclient/apt.py
```diff
@@ -214,6 +215,29 @@ def get_apt_cache_policy(
...
+
+        return self.apt_func()
+
+    def __exit__(self, type, value, traceback):
+        cfg = apt_pkg.config
+        # We need to restore the apt cache configuration after creating our
+        # cache, otherwise we may break people interacting with the
+        # library after importing our modules.
+        for key in self.current_apt_cfg.keys():
+            cfg.set(key, self.current_apt_cfg[key])
```
renanrodrigo, line 238: If I am not mistaken we can simplify this by just assigning the config back to apt_pkg.
```suggestion
        apt_pkg.config = self.current_apt_cfg
```

uaclient/apt.py
```diff
@@ -754,8 +771,10 @@ def _get_apt_config():
 
 
 def get_apt_config_keys(base_key):
-    apt_cfg = _get_apt_config()
-    return apt_cfg.list(base_key)
+    with PreserveAptCfg(_get_apt_config) as apt_cfg:
```
renanrodrigo, line 774: I don't understand why we need the ctx change here if we want the current apt config... but, if it's needed here, wouldn't it be needed in `get_apt_config_values` too?

uaclient/apt.py
```diff
@@ -214,6 +215,29 @@ def get_apt_cache_policy(
     )
 
 
+class PreserveAptCfg:
+    def __init__(self, apt_func):
+        self.apt_func = apt_func
+        self.current_apt_cfg = {}  # Dict[str, Any]
+
+    def __enter__(self):
+        apt_pkg.init_config()
```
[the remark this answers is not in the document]
lucasmoura: You are right, I have removed that part of the code

uaclient/apt.py
```diff
@@ -214,6 +215,28 @@ def get_apt_cache_policy(
...
+
+        return self.apt_func()
+
+    def __exit__(self, type, value, traceback):
+        cfg = apt_pkg.config
+        # We need to restore the apt cache configuration after creating our
+        # cache, otherwise we may break people interacting with the
+        # library after importing our modules.
+        for key in self.current_apt_cfg.keys():
+            cfg.set(key, self.current_apt_cfg[key])
```
renanrodrigo, line 237: @lucasmoura should we maybe call `init_system` here?

renanrodrigo (approved):

PR #2503, opened 2023-03-21
logging_revamp: enable non-root user logging SC-1551

CalvoM:
## Proposed Commit Message

logging_revamp: enable non-root user logging
Now, we can capture logs when the pro client is run by a non-root user.

no-lp no-gh
- [x] Add Redaction Filter.
- [x] Add Json Array Formatter.
- [x] Provide a way for non-root user to log.

## Checklist

 - [ ] Yes
 - [x] No


uaclient/cli.py
```diff
@@ -1853,20 +1853,24 @@ def setup_logging(console_level, log_level, log_file=None, logger=None):
...
 
+    try:
         file_handler = logging.FileHandler(log_file)
         file_handler.setFormatter(JsonArrayFormatter())
         file_handler.setLevel(log_level)
         file_handler.set_name("ua-file")
         logger.addHandler(file_handler)
+    except PermissionError as e:
+        logging.error(e)
+        return 1
```
orndorffgrant, line 1873: `setup_logging`  doesn't have a return type and we don't use a return value where it's called AFAIK

When do we expect this exception to occur? a `logging.warning` might be sufficient

uaclient/conftest.py
```diff
@@ -57,15 +57,18 @@ def _warn_about_new_version():
...
-    from uaclient.util import we_are_currently_root
+    from uaclient import util
 
-    original = we_are_currently_root
-    with mock.patch("uaclient.util.we_are_currently_root", return_value=True):
+    is_root = getattr(request, "param", True)
+    with mock.patch(
+        "uaclient.util.we_are_currently_root", return_value=is_root
+    ):
+        original = util.we_are_currently_root
```
orndorffgrant, line 71: This isn't needed. You can mock.patch('uaclient.util.we_are_currently_root`) to return a different value as normal for a unit test if you need it to return False.

uaclient/defaults.py
```diff
@@ -69,3 +69,4 @@
 WORLD_READABLE_MODE = 0o644
 NOTICES_PERMANENT_DIRECTORY = DEFAULT_DATA_DIR + "/notices/"
 NOTICES_TEMPORARY_DIRECTORY = UAC_RUN_PATH + "notices/"
+USER_LOG_FILE_PATH = "ubuntu-pro/ubuntu-pro"
```
orndorffgrant, line 72: I think this should be
```suggestion
USER_LOG_FILE_PATH = "ubuntu-pro/ubuntu-pro.log"
```
and we don't need to add the user name to the file name.

During collect-logs, we should add anonymous tags to the user log files: e.g. "user1.log" "user2.log", etc

uaclient/log.py
```diff
@@ -58,3 +62,35 @@ def format(self, record: logging.LogRecord) -> str:
...
+    user, _ = system.subp(["whoami"])
+    user = user.strip()
+    user_cache = os.environ.get("XDG_CACHE_HOME")
+    log_file = "{}-{}.log".format(defaults.USER_LOG_FILE_PATH, user)
+    if user_cache:
+        return (pathlib.Path(user_cache) / log_file).as_posix()
+    return (pathlib.Path.home() / ".cache" / log_file).as_posix()
+
+
+def get_users_log_file() -> List[str]:
```
orndorffgrant, line 82: recommended name change
```suggestion
def get_all_user_log_files() -> List[str]:
```

uaclient/log.py
```diff
@@ -58,3 +62,35 @@ def format(self, record: logging.LogRecord) -> str:
...
+
+
+def get_user_log_file() -> str:
+    """Gets the correct user log_file storage location
+
+    returns XDG_CACHE_HOME env variable, if it is set,
+    If not set, we return the default path
+    """
+    user, _ = system.subp(["whoami"])
+    user = user.strip()
```
orndorffgrant, line 74: I think we can use `os.getlogin()`

uaclient/log.py
```diff
@@ -58,3 +62,35 @@ def format(self, record: logging.LogRecord) -> str:
...
+    return (pathlib.Path.home() / ".cache" / log_file).as_posix()
+
+
+def get_users_log_file() -> List[str]:
+    """Gets all the log files for the users in the system
+
+    Returns a list of all user log files in their home directories.
+    """
+    out, _ = system.subp(["ls", "/home/"])
+    user_directories = out.splitlines()[:USER_LOG_COLLECTED_LIMIT]
```
orndorffgrant, line 88: we can use `os.listdir()` here

aside: generally speaking, we shouldn't have to use subprocesses for almost anything. subp is a last resort for if there is no proper API to what we're trying to do. This is just a general rule and there are nuances/exceptions.

uaclient/log.py
```diff
@@ -58,3 +62,35 @@ def format(self, record: logging.LogRecord) -> str:
...
+    returns XDG_CACHE_HOME env variable, if it is set,
+    If not set, we return the default path
+    """
+    user, _ = system.subp(["whoami"])
+    user = user.strip()
+    user_cache = os.environ.get("XDG_CACHE_HOME")
+    log_file = "{}-{}.log".format(defaults.USER_LOG_FILE_PATH, user)
+    if user_cache:
+        return (pathlib.Path(user_cache) / log_file).as_posix()
+    return (pathlib.Path.home() / ".cache" / log_file).as_posix()
```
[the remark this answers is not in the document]
orndorffgrant: I admit I have a bias against operator overloading though :upside_down_face: and this is probably considered more "pythonic" by some. @lucasmoura @renanrodrigo how do you feel about it? If the team prefers this style, then we can start moving towards it

orndorffgrant (commented):
Looking good!
Can we split this into two commits within this PR? The first can be logging when non-root. The second can be collecting non-root logs in `pro collect-logs`

uaclient/log.py
CalvoM: This is after testing if I change users in the terminal using `su -l <user>` and noticed `os.getlogin()` returned the my user name instead of the other user I logged in as. i.e. if my user name is `user1` and I change the user in the terminal to `user2` 
`os.getlogin()` returns `user1`.

orndorffgrant: Out of band conversation we decided to use getpass.getuser()

uaclient/log.py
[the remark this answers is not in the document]
renanrodrigo: consistency > pythoniceties
I would go with the same standards we have been following

uaclient/cli.py
CalvoM: Actually we were able to solve the problem , when the user log file is in /var/ folder, so no reason for the try...except clause.

uaclient/conftest.py
CalvoM: I wanted it to return both True and False in a parametrized way,  instead of having the test duplicated with different return values.

uaclient/defaults.py
CalvoM: This might be me overthinking, but I imagine that the root/admin might need to know which user had run which command, thus I thought adding the user will make it easier for them.

uaclient/conftest.py
orndorffgrant: Yep! Check [this](https://github.com/canonical/ubuntu-pro-client/blob/main/uaclient/tests/test_security.py#L2267-L2274) out for an example of this

uaclient/log.py
```diff
@@ -58,3 +63,33 @@ def format(self, record: logging.LogRecord) -> str:
...
+    """
+    user = getpass.getuser()
+    user_cache = os.environ.get("XDG_CACHE_HOME")
+    log_file = "{}-{}.log".format(defaults.USER_LOG_FILE_PATH, user)
+    if user_cache:
+        return user_cache + "/" + log_file
+    return pathlib.Path.home().as_posix() + "/.cache/" + log_file
+
+
+def get_all_user_log_files() -> List[str]:
```
orndorffgrant, line 82: Right now this will return files that don't exist, and the 10 log limit is in this function.

This function should probably return _all_ user log files that exist.  Then the 10-file limit can go in the `collect_logs` function.

uaclient/cli.py
```diff
@@ -1855,20 +1855,20 @@ def setup_logging(console_level, log_level, log_file=None, logger=None):
...
-        log_file_path = pathlib.Path(log_file)
+    log_file_path = pathlib.Path(log_file)
 
-        if not log_file_path.exists():
-            log_file_path.touch()
+    if not log_file_path.exists():
+        log_file_path.parent.mkdir(parents=True, exist_ok=True)
+        log_file_path.touch()
+        if util.we_are_currently_root():
             log_file_path.chmod(0o644)
```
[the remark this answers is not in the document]
orndorffgrant: Actually maybe we want to make them more restrictive. The default of `touch` is `666`. Maybe user log files should only be `640` though? What do you think @lucasmoura @renanrodrigo

uaclient/conftest.py
CalvoM: Thanks for the link.

uaclient/cli.py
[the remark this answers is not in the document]
lucasmoura: I think that makes sense. +1 on that @orndorffgrant

lucasmoura (commented):
@CalvoM This is looking good, but I think we should also add integration test for this

uaclient/log.py
CalvoM: Oh makes, sense I thought the the check [here](https://github.com/canonical/ubuntu-pro-client/blob/7da5a991c6143f0650d9e927f3668e56f6df810b/uaclient/actions.py#L215) was enough but we may run having chosen user log files that are not present over present ones. Thanks

uaclient/log.py
```diff
@@ -58,3 +60,35 @@ def format(self, record: logging.LogRecord) -> str:
...
+    """Gets all the log files for the users in the system
+
+    Returns a list of all user log files in their home directories.
+    """
+    user_directories = os.listdir("/home")
+    dirlist = []
```
orndorffgrant, line 84: This isn't a list of dirs, so let's not call it a dirlist
```suggestion
    log_files = []
```

uaclient/system.py
```diff
@@ -370,14 +370,26 @@ def create_file(filename: str, mode: int = 0o644) -> None:
...
+    is_file_present = os.path.isfile(filename)
+    if is_file_present:
+        file_stat = pathlib.Path(filename).stat()
+        f_mode = util.get_file_perm_mode(filename)
+        if not mode:
+            mode = f_mode
+
+    else:
+        if not mode:
+            mode = 0o644
```
orndorffgrant, line 392: Please double check that this is equivalent, but I think we can do this instead of `stat`-ing the file twice. Then we don't need the util function

```suggestion
    is_file_present = os.path.isfile(filename)
    if is_file_present:
        file_stat = pathlib.Path(filename).stat()
        f_mode = stat.S_IMODE(file_stat.st_mode)
        if mode is None:
            mode = f_mode
    elif mode is None:
        mode = 0o644
```

uaclient/cli.py
```diff
@@ -1969,10 +1967,13 @@ def wrapper(*args, **kwargs):
 
 @main_error_handler
 def main(sys_argv=None):
+    log_file = (
+        defaults.CONFIG_DEFAULTS["log_file"]
+        if util.we_are_currently_root()
+        else pro_log.get_user_log_file()
+    )
```
orndorffgrant, line 1974: I think this logic can happen inside of `setup_logging` instead of doing it twice at this level

uaclient/actions.py
```diff
@@ -225,6 +236,13 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
             if util.we_are_currently_root():
                 # if root, overwrite the original with redacted content
                 system.write_file(f, content)
+
+            if os.path.basename(f) == os.path.basename(
+                defaults.USER_LOG_FILE_PATH
+            ):
+                log_dir = os.path.dirname(f)
+                f = log_dir + "/ubuntu-pro" + str(log_file_count) + ".log"
+                log_file_count += 1
```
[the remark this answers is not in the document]
CalvoM: I thought of this earlier, however, I saw we will need to read the contents of the files before renaming the files in some output temp folder. This seemed to be a longer process than we have currently. What do you think?

features/steps/files.py
```diff
@@ -46,6 +46,13 @@ def when_i_verify_file_is_empty_on_machine(context, file_name, machine_name):
     assert_that(context.process.stdout.strip(), matches_regexp("0"))
 
 
+@when("I verify `{file_name}` is empty")
+def when_i_verify_file_is_empty(context, file_name):
+    command = 'sh -c "cat {} | wc -l"'.format(file_name)
+    when_i_run_command(context, command, user_spec="with sudo")
+    assert_that(context.process.stdout.strip(), matches_regexp("0"))
```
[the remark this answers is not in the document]
CalvoM: Thanks for this.

uaclient/actions.py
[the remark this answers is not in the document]
orndorffgrant: I'm not sure I understand - to clarify what my suggestion is: I don't think the user log files should be processed in this loop at all - they shouldn't go in the list that they're currently in. I think there should be separate for-loop after this one that grabs and renames the user log files. Sorry if that doesn't help - let's talk about this tomorrow to iron it out.

uaclient/cli.py
```diff
@@ -1992,7 +1992,7 @@ def main(sys_argv=None):
 
     log_level = cfg.log_level
     console_level = logging.DEBUG if args.debug else logging.INFO
-    setup_logging(console_level, log_level, cfg.log_file)
+    setup_logging(console_level, log_level)
```
orndorffgrant, line 1995: Removing this will result in `setup_logging` instantiating a new UAConfig object unnecessarily. Let's leave it as it is for now
```suggestion
    setup_logging(console_level, log_level, cfg.log_file)
```

uaclient/system.py
```diff
@@ -370,14 +371,25 @@ def create_file(filename: str, mode: int = 0o644) -> None:
     os.chmod(filename, mode)
 
 
-def write_file(filename: str, content: str, mode: int = 0o644) -> None:
+def write_file(
+    filename: str, content: str, mode: Optional[int] = None
+) -> None:
     """Write content to the provided filename encoding it if necessary.
```
orndorffgrant, line 377: Can we add some content to this docstring describing the new behavior - that it preserves ownership and permissions of an existing file unless `mode` is not-None

uaclient/util.py
```diff
@@ -631,3 +631,7 @@ def deduplicate_arches(arches: List[str]) -> List[str]:
 
 def we_are_currently_root() -> bool:
     return os.getuid() == 0
+
+
+def get_file_perm_mode(path: str) -> int:
+    return stat.S_IMODE(os.lstat(path).st_mode)
```
orndorffgrant, line 637: I don't think we need this anymore

uaclient/cli.py
CalvoM: Oh yes, it is like saying we default to the cfg.log_file unless we are non-root, the condition in `setup_logging`.

orndorffgrant (commented):
Thank you this is super close and I think we can include it in 27.15! I just left some final nits on the collect-logs part

lucasmoura (approved):
Just a really tiny nit, but we don't need that change to land this PR

uaclient/tests/test_cli_collect_logs.py
```diff
@@ -57,22 +59,47 @@ def test_collect_logs_help(
...
     ):
+        util_we_are_currently_root.return_value = is_root
+        m_get_user.return_value = tmpdir.join("user-log").strpath
+        m_get_users.return_value = [
+            tmpdir.join("user1-log").strpath,
+            tmpdir.join("user2-log").strpath,
+        ]
+        is_file_calls = 17
+        user_log_files = [mock.call(m_get_user())]
+        if util_we_are_currently_root():
```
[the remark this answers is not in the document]
CalvoM: Yes, makes sense. I think either is okay.

renanrodrigo (approved):
lgtm

PR #2514, opened 2023-03-29
tests: adding i8n test support (SC-1476)

CalvoM:
## Proposed Commit Message

tests: adding i8n test support

Adding integration tests to check if all commands run successfully 
in systems with different locales/languages.

no-lp no-gh

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/i8n.feature
+++ b/features/i8n.feature
@@ -1,0 +1,128 @@
+Feature: Pro supports multiple languages
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    @uses.config.contract_token
+    Scenario Outline: Pro client's commands run successfully
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        ## Change the locale
+        When I run `apt install language-pack-fr -y` with sudo
+        And I run `update-locale LANG=fr_FR.UTF-8` with sudo
+        And I reboot the machine
+        And I run `cat /etc/default/locale` as non-root
+        Then stdout matches regexp:
+        """
+        LANG=fr_FR.UTF-8
+        """
+        #Attach invalid token
+        When I verify that running `pro attach INVALID_TOKEN` `with sudo` exits `1`
+        Then stderr matches regexp:
+        """
+        Invalid token. See https://ubuntu.com/pro
+        """
+        When I run `lscpu` as non-root
+        Then stdout does not match regexp:
+        """
+        Architecture:
+        """
+        When I run `apt update` with sudo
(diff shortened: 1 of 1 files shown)


features/i8n.feature
```diff
@@ -0,0 +1,118 @@
+Feature: Pro supports multiple languages
+
+    @series.all
+    @uses.config.machine_type.lxd.container
+    @uses.config.contract_token
+    Scenario Outline: Pro client's commands run successfully
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        ## Change the locale
+        When I run `locale-gen uk_UA.UTF-8` with sudo
+        And I run `update-locale LANG=uk_UA.UTF-8` with sudo
```
orndorffgrant, line 10: I'm not familiar with changing locales but I was expecting to see the installation of a language-pack somewhere like `apt install language-pack-uk`?

I'm also confused about the locale-gen command since the manpage and help text don't indicate that it accepts a positional argument like it's being used here.

But again I'm not familiar with switching locales so IDK :upside_down_face:

features/i8n.feature
```diff
@@ -0,0 +1,118 @@
...
+        Then stdout matches regexp:
+        """
+        LANG=uk_UA.UTF-8
+        """
+        #Attach invalid token
+        When I verify that running `pro attach INVALID_TOKEN` `with sudo` exits `1`
+        Then stderr matches regexp:
+            """
+            Invalid token. See https://ubuntu.com/pro
+            """
```
orndorffgrant, line 22: I think we're standardizing on not indenting these text sections, so this applies to all of these
```suggestion
        """
        Invalid token. See https://ubuntu.com/pro
        """
```

features/i8n.feature
```diff
@@ -0,0 +1,118 @@
...
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        ## Change the locale
+        When I run `locale-gen uk_UA.UTF-8` with sudo
+        And I run `update-locale LANG=uk_UA.UTF-8` with sudo
+        And I reboot the machine
+        And I run `cat /etc/default/locale` as non-root
+        Then stdout matches regexp:
+        """
+        LANG=uk_UA.UTF-8
+        """
```
orndorffgrant, line 16: We should also run a couple commands to make sure their output is non-english. `lscpu` and `apt update` are good candidates

renanrodrigo (approved):
we can fix alignment on all tests later
thanks for this

PR #2519, opened 2023-03-30
kernel_info: use uname.machine for kernel arch

orndorffgrant:
no-jira no-lp

Also brings in the removal of lscpu parsing from `release`

#914 is related but that problem no longer applies. The contract server now contains all aliases for architectures. And we still use dpkg arch for package-based services.

## Test Steps

Ensure livepatch.feature still passes

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (approved):
LGTM

PR #2520, opened 2023-03-30
config: no longer load uaclient.conf from cwd

orndorffgrant:
no-jira no-lp no-gh

## Test Steps

Everything except this feature should still work

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (commented):
I love this PR and I would say this could even enter 27.14.3 if there is no barrier, but it seems you just found a loophole in our tests (:

renanrodrigo (approved):

PR #2521, opened 2023-03-31
livepatch: check series affordances again

orndorffgrant:
livepatch treats Ubuntu Release version support differently from how it treats kernel version/flavor/arch support. The Ubuntu Release is checked against hardcoded list and causes the canonical-livepatch command to raise errors. Because of this, we need to use the Contract Server-defined series affordance to prevent even attempting to install and run canonical-livepatch on unsupported releases (interim releases).

## Test Steps

`pro attach` on an interim release vm. ensure we don't try (and fail) to enable livepatch

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+ubuntu-advantage-tools (27.14.3) lunar; urgency=medium
+
+  * livepatch: prevent livepatch from auto-enabling and subsequently failing
+    on interim releases (LP: #2013409)
+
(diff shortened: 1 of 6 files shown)


renanrodrigo (approved):
I know this is secondary, but given tox.ini got upgraded, could you please include the new VM scenarios to `tools/run-integration-tests.py`? (fine if the answer is no for SRU reasons, can do later)

orndorffgrant:
done!

orndorffgrant:
This now also includes the commit to update the version and add the new changelog entry

renanrodrigo (approved):

PR #2523, opened 2023-03-31
postinst: more specific regex for ua_config warning

orndorffgrant:
no-jira no-lp no-gh

The old warning would be shown even if ua_config: was commented out. By ensuring we only warn when the line begins with ua_config: we don't warn in this situation where it is unnecessary.

## Test Steps

On the previous version of pro-client edit uaclient.conf to say
```
#ua_config:
{{ <- this causes the migration to fail
```
Then upgrade to this version. Ensure there is no warning about "ua_config" being present in the config

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -443,7 +443,7 @@ case "$1" in
           migrate_user_config_post
       fi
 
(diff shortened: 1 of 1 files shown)


renanrodrigo (approved):

orndorffgrant:
this won't go in the hotfix, removing label
I'll rebase this onto `main`

PR #2525, opened 2023-03-31
config: no longer load uaclient.conf from cwd - again

orndorffgrant:
no-jira no-lp no-gh

I'm removing this from release, so putting it back here as a PR into main

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/clouds/tests/test_identity.py
+++ b/uaclient/clouds/tests/test_identity.py
@@ -71,11 +71,12 @@ class TestGetCloudType:
             ),
         ),
     )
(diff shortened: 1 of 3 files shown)


renanrodrigo (approved):
this nice
given this was something caught in-review, should we open a LP bug about it and reference it here? if no, then good to land

orndorffgrant:
I don't think we need to create an LP bug since this will just be included in the next major release. Open to counter args though

lucasmoura (approved):
I agree with Grant, we don't need a LP bug for this

PR #2539, opened 2023-04-06
Add variant support to realtime-kernel (SC-1465)

lucasmoura:
no-lp no-gh

## Proposed Commit Message
Add support for realtime-kernel variants

This PR is adding the whole variant support to the Pro client, while also defining specific variants for the realtime-kernel service. I still need to add the affordances check for the nvidia variant, but this should not block the review for this PR.

## Test Steps
Run the modified realtime-kernel integration test

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/realtime_kernel.feature
+++ b/features/realtime_kernel.feature
@@ -109,6 +109,47 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
             """
             Installed: (none)
             """
+        When I set the machine token overlay to the following yaml
+        """
+        machineTokenInfo:
+          contractInfo:
+            resourceEntitlements:
+              - type: realtime-kernel
+                overrides:
+                  - directives:
+                      additionalPackages:
+                        - nvidia-prime
+                    selector:
+                      platform: nvidia-tegra
+                  - directives:
+                      additionalPackages:
+                        - intel-pkg
+                    selector:
+                      platform: intel-iotg
+        """
+        And I run `pro enable realtime-kernel --variant nvidia-tegra` with sudo
+        Then I will see the following on stdout:
+        """
+        One moment, checking your subscription first
+        Updating package lists
+        Installing Real-time Nvidia Tegra Kernel packages
+        Real-time Nvidia Tegra Kernel enabled
+        """
+        When I run `pro status` as non-root
+        Then stdout matches regexp:
+        """
+        realtime-kernel\* yes +enabled +Ubuntu kernel with PREEMPT_RT patches integrated
+
+         \* This service has options, use pro status --all to see more details.
+        """
+         When I run `pro status --all` as non-root
+         Then stdout matches regexp:
+         """
+         realtime-kernel  yes +enabled   +Ubuntu kernel with PREEMPT_RT patches integrated
+         ├ generic        yes +disabled  +Generic version of the RT kernel \(default\)
+         ├ nvidia-tegra   yes +enabled   +RT kernel optimized for NVidia Tegra platforms
+         └ intel-iotg     yes +disabled  +RT kernel optimized for Intel IOTG platform
+         """
 
         Examples: ubuntu release
             | release |
--- a/features/realtime_kernel.feature
+++ b/features/realtime_kernel.feature
@@ -109,6 +109,11 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
             """
             Installed: (none)
             """
+        When I verify that running `pro enable realtime-kernel --access-only --variant nvidia-tegra` `with sudo` exits `1`
+        Then I will see the following on stderr:
+        """
+        Error: Cannot use --access-only together with --variant.
+        """
         When I set the machine token overlay to the following yaml
         """
         machineTokenInfo:
--- a/features/realtime_kernel.feature
+++ b/features/realtime_kernel.feature
@@ -50,71 +50,91 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         When I attach `contract_token` with sudo and options `--no-auto-enable`
         Then I verify that running `pro enable realtime-kernel` `as non-root` exits `1`
         And I will see the following on stderr:
-            """
-            This command must be run as root (try using sudo).
-            """
+        """
+        This command must be run as root (try using sudo).
+        """
         When I run `pro enable realtime-kernel` `with sudo` and stdin `y`
         Then stdout matches regexp:
-            """
-            One moment, checking your subscription first
-            The Real-time kernel is an Ubuntu kernel with PREEMPT_RT patches integrated.
+        """
+        One moment, checking your subscription first
+        The Real-time kernel is an Ubuntu kernel with PREEMPT_RT patches integrated.
 
-            .*This will change your kernel. To revert to your original kernel, you will need
-            to make the change manually..*
+        .*This will change your kernel. To revert to your original kernel, you will need
+        to make the change manually..*
 
-            Do you want to continue\? \[ default = Yes \]: \(Y/n\) Updating package lists
-            Installing Real-time kernel packages
-            Real-time kernel enabled
-            A reboot is required to complete install.
-            """
+        Do you want to continue\? \[ default = Yes \]: \(Y/n\) Updating package lists
+        Installing Real-time kernel packages
+        Real-time kernel enabled
+        A reboot is required to complete install.
+        """
         When I run `apt-cache policy ubuntu-realtime` as non-root
         Then stdout does not match regexp:
-            """
-            .*Installed: \(none\)
-            """
+        """
+        .*Installed: \(none\)
+        """
         And stdout matches regexp:
-            """
-            \s* 500 https://esm.ubuntu.com/realtime/ubuntu <release>/main amd64 Packages
-            """
+        """
+        \s* 500 https://esm.ubuntu.com/realtime/ubuntu <release>/main amd64 Packages
+        """
         When I verify that running `pro enable realtime-kernel` `with sudo` exits `1`
         Then stdout matches regexp
-            """
-            One moment, checking your subscription first
-            Real-time kernel is already enabled.
-            See: sudo pro status
-            """
+        """
+        One moment, checking your subscription first
+        Real-time kernel is already enabled.
+        See: sudo pro status
+        """
         When I reboot the machine
         When I run `uname -r` as non-root
         Then stdout matches regexp:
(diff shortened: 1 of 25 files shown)


orndorffgrant:
This looks really good!

I have a couple questions/ideas from a high-level review:
1. I think the selector should be `variant` instead of `platform` since this will likely be re-used for variants that are not platforms in the future. I don't think we have anything in the pipeline yet, but I can foresee this being useful for something like e.g. `pro enable landscape --variant=saas` vs `pro enable landscape --variant=self-hosted` or `pro enable fips --variant=140-2` vs `pro enable fips --variant=140-3`. @alnvdl-work 
2. Similarly, I'd expect the obligation to use a variant to be called `use_variant` rather than `use_selector`. (also I'm sorry - I probably could've brought up both 1 and 2 at spec time but I missed these details at the time :frowning_face:)
3. Can the variant selector override code be integrated into the existing `apply_contract_overrides` function?
4. Can the new affordance checking go in the base entitlement?

alnvdl-work:
> This looks really good!
> 
> I have a couple questions/ideas from a high-level review:
> 
> 1. I think the selector should be `variant` instead of `platform` since this will likely be re-used for variants that are not platforms in the future. I don't think we have anything in the pipeline yet, but I can foresee this being useful for something like e.g. `pro enable landscape --variant=saas` vs `pro enable landscape --variant=self-hosted` or `pro enable fips --variant=140-2` vs `pro enable fips --variant=140-3`. @alnvdl-work
> 2. Similarly, I'd expect the obligation to use a variant to be called `use_variant` rather than `use_selector`. (also I'm sorry - I probably could've brought up both 1 and 2 at spec time but I missed these details at the time frowning_face)
> 3. Can the variant selector override code be integrated into the existing `apply_contract_overrides` function?
> 4. Can the new affordance checking go in the base entitlement?

I was the one who originally suggested `platform` instead of `variant`, because I thought `variant` was a bit too generic. If we have a clear idea of how this will be used in other places in the future, perhaps going with a generic name is better indeed, so no objections here.

lucasmoura:
@orndorffgrant I agree with points 1 and 2. I think we can rename them on Contract Server @alnvdl-work. I will adapt the code in parallel.

Regarding points 3 and 4:

3) Great idea, I have updated the code use that instead
4) Because nvidia and intel will have different ways to obtain the vendor information, I have modified the code to only specify the specific vendor information on the variants

uaclient/entitlements/realtime.py
```diff
@@ -32,6 +32,13 @@ def _check_for_reboot(self) -> bool:
         event.needs_reboot(reboot_required)
         return reboot_required
 
+    def _get_variants(self) -> Dict[str, Type[UAEntitlement]]:
+        return {
+            GenericRealtime.variant_name: GenericRealtime,
+            NvidiaTegraRealtime.variant_name: NvidiaTegraRealtime,
+            IntelIotgRealtime.variant_name: IntelIotgRealtime,
+        }
```
orndorffgrant, line 40: How do we indicate that the generic variant is the default?

`pro enable realtime-kernel` and `pro enable realtime-kernel --variant generic` should do the exact same thing. It looks like we'll get different entitlement objects for them though.

uaclient/entitlements/realtime.py
```diff
@@ -101,3 +108,93 @@ def remove_packages(self) -> None:
...
+
+class GenericRealtime(RealtimeKernelEntitlement):
+    variant_name = "generic"
+    title = "Real-time kernel"
+    description = "Generic version of the RT kernel (default)"
+    is_variant = True
+    check_packages_are_installed = True
+
+    @property
+    def incompatible_services(self) -> Tuple[IncompatibleService, ...]:
```
orndorffgrant, line 121: For all the variants that are overriding properties of the RealtimeKernelEntitlement like this, I think we'll need to call `super().incompatible_services` and include that in the return value

features/realtime_kernel.feature
```diff
@@ -109,6 +109,134 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
...
+        Updating package lists
+        Installing Real-time Nvidia Tegra Kernel packages
+        Real-time Nvidia Tegra Kernel enabled
+        """
+        When I run `pro status` as non-root
+        Then stdout matches regexp:
+        """
+        realtime-kernel\* yes +enabled +Ubuntu kernel with PREEMPT_RT patches integrated
+
+         \* This service has options, use pro status --all to see more details.
```
orndorffgrant, line 148:
```suggestion
         \* This service has variants, use pro status --all to see more details.
```
idea to be slightly more consistent in language

uaclient/entitlements/realtime.py
orndorffgrant: Separately, I'd like the visibility of variants to be driven by the contract server. In two ways preferably (though maybe we can add one or both of these as follow up PRs)
1. variants defined in pro-client but not defined in the entitlement **should not** get displayed
2. variants defined in the entitlement but not defined in the pro-client **should** get displayed (with fallback generic descriptions)

I think (1) can be easily implemented in this function by not including variants in the return value unless they're found in the entitlement.

(2) is a bit trickier, but can probably be an anonymous subclass or something that is returned here.

orndorffgrant: Part of my goal is that we should be able to merge and release this before the variants are finalized and released on the backend. Right now if you run this without overriding the machine_token then you get some weird behavior like all variants saying they're enabled at once.

orndorffgrant (commented):
I started playing with it and it's looking really good! I have some comments and ideas in comments. As-is I don't think we can merge until all the variants are finalized on the backend. But I don't think we even have a target date on these being finalized

alnvdl-work:
@lucasmoura Please let me know when I should rename `platform` to `variant` it in ua-contracts, and if you think it will affect your testing.

uaclient/entitlements/realtime.py
orndorffgrant: @lucasmoura this comment still applies

features/realtime_kernel.feature
```diff
@@ -50,65 +50,288 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
...
+        """
+        When I run `pro enable realtime-kernel --variant intel-iotg` `with sudo` and stdin `y\ny\n`
+        When I run `pro status --all` as non-root
+        Then stdout matches regexp:
+        """
+        realtime-kernel  yes +enabled   +Ubuntu kernel with PREEMPT_RT patches integrated
+        ├ generic        yes +disabled   +Generic version of the RT kernel \(default\)
+        └ intel-iotg     yes +enabled  +RT kernel optimized for Intel IOTG platform
+        """
+        When I verify that running `pro enable realtime-kernel` `with sudo` exits `1`
```
CalvoM, line 189: Question in line with @orndorffgrant's suggestion and observation, should this command without `--variant` option install the generic one?... Or since it is already enabled, running the `enable` command will not do anything even though the variant is not generic?

renanrodrigo (commented):
Found a bug and a concern when testing (keeping in mind I think the refactor of classes and such may come later). Will add those to next comments

renanrodrigo:
Bug: when you try to enable a variant that does not exist, the CLI tells you that the entitlement itself could not be found.
```
# pro enable realtime-kernel --variant whatever
could not find entitlement named "realtime-kernel"
```

renanrodrigo:
My concern is the status JSON - we may not be able to come up with a better solution if this lands as is; as I foresee a better solution would rearrange the status JSON to better represent the variants instead of shaping them as services themselves.
A couple current implemented side effects of that decision are:
```
- available: 'yes'
  blocked_by: []
  description: Ubuntu kernel with PREEMPT_RT patches integrated
  description_override: null
  entitled: 'yes'
  name: realtime-kernel
  status: enabled
  status_details: Real-time kernel is active
  variants:
    generic:
      available: 'yes'
      blocked_by:
      - name: realtime-kernel
        reason: Real-time kernel cannot be enable together with Real-time Intel IOTG
          Kernel
        reason_code: realtime-variant-incompatible
      description: Generic version of the RT kernel (default)
      description_override: null
      entitled: 'yes'
      name: generic
      status: disabled
      status_details: 'The realtime-kernel service is not enable because the ubuntu-realtime
        package is

        not installed.'
      variants: {}
      warning: null
    intel-iotg:
      available: 'yes'
      blocked_by: []
      description: RT kernel optimized for Intel IOTG platform
      description_override: null
      entitled: 'yes'
      name: intel-iotg
      status: enabled
      status_details: Real-time Intel IOTG Kernel is active
      variants: {}
      warning: null
(need to check if that blank line is not just terminal breaking lines, but still)
```
- the generic variant is shown blocked by `realtime-kernel` itself, instead of the other variant, which is what makes more sense
- variants have a `variant: {}` key, which I understand is no-op-able and don't represent anything, but this very assumption justifies that it does not make sense to have those

uaclient/messages.py
```diff
@@ -1296,3 +1307,24 @@ class TxtColor:
...
 )
+STATUS_SERVICE_HAS_VARIANTS = " * Service has variants"
+
+STATUS_ALL_HINT_WITH_VARIANTS = """\
+For a list of all Ubuntu Pro services and variants, run 'pro status --all'"""
+
+SERVICE_DISABLED_MISSING_PACKAGE = FormattedNamedMessage(
+    "service-disabled-missing-package",
+    """\
+The {service} service is not enable because the {package} package is
```
renanrodrigo, line 1318:
```suggestion
The {service} service is not enabled because the {package} package is
```

uaclient/messages.py
```diff
@@ -840,6 +847,10 @@ class TxtColor:
     "realtime-livepatch-incompatible",
     "Livepatch is not currently supported for the Real-time kernel.",
 )
+REALTIME_VARIANT_INCOMPATIBLE = FormattedNamedMessage(
+    "realtime-variant-incompatible",
+    "{service} cannot be enable together with {variant}",
```
renanrodrigo, line 852:
```suggestion
    "{service} cannot be enabled together with {variant}",
```

features/realtime_kernel.feature
lucasmoura: Yes, by default we are installing the generic variant when no `--variant` is used. And yes, the reason way it fails it is because it is already enabled

renanrodrigo (approved):
thanks for all the work, and for all iteration on top of this - I think we are in a good direction

PR #2540, opened 2023-04-07
timer: only run when attached (SC-1304)

orndorffgrant:
no-lp no-gh
The final PR of my 23.04 tech-debt ticket.

After our recent improvements, we only need an intermittent job to run when attached. So let's not waste everyone's cpu cycles by leaving the timer on when unattached.

## Test Steps

Verify the timer is off when upgrading to this version while unattached.
Verify the timer stays on when upgrading to this version while attached.
Verify that attaching starts the timer.
Verify that detaching stops the timer.

## Checklist

 - [x] Yes
 - [ ] No


uaclient/timer/__init__.py
```diff
@@ -0,0 +1,20 @@
+import logging
+from subprocess import TimeoutExpired
+
+from uaclient import exceptions, system
+
+LOG = logging.getLogger("uaclient.timer")
```
aciba90, line 6: Portability nit:

```suggestion
LOG = logging.getLogger(__package__)
```

aciba90 (approved):
LGTM, thanks!

One question. What is the reason to not start the timer right after a successful auto-attach? I understand it will eventually get activated at some point per `systemd/ua-timer.timer`, but I do not see the reason why in the case of attach with token a timer.start() is needed but not ofr the auto-attach case.

uaclient/timer/__init__.py
orndorffgrant: Good point! Is there a reason to prefer `__package__` over `__name__`?

orndorffgrant:
@aciba90 Thanks for taking a look! auto-attach includes calling `attach_with_token`, so this should start the timer right after an auto-attach as well

renanrodrigo:
All succeeded all approved clicking the magic button

PR #2543, opened 2023-04-11
`pro status` improvements (SC-1360)

renanrodrigo:
Instead of showing an empty services table on pro status, just put a message saying no service is available. Plus, add a hint for the `-all` flag if one wants to see everything Ubuntu Pro has to offer.

## Test Steps
Integration tests properly added/modified.

## Checklist

 - [ ] Yes
 - [x] No

no-lp no-gh

### DIFF
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -604,7 +604,7 @@ def format_expires(expires: Optional[datetime]) -> str:
     return expires.strftime("%c %Z")
 
 
-def format_tabular(status: Dict[str, Any]) -> str:
+def format_tabular(status: Dict[str, Any], show_all_hint: bool = False) -> str:
     """Format status dict for tabular output."""
     if not status.get("attached"):
         if status.get("simulated"):
@@ -659,6 +659,9 @@ def format_tabular(status: Dict[str, Any]) -> str:
             for key, value in sorted(status["features"].items()):
                 content.append("{}: {}".format(key, value))
 
+        if show_all_hint:
+            content.extend(["", messages.STATUS_ALL_HINT])
+
         content.extend(["", messages.UNATTACHED.msg])
         if livepatch.on_supported_kernel() is False:
             content.extend(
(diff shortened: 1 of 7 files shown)


uaclient/status.py
```diff
@@ -621,25 +624,30 @@ def format_tabular(status: Dict[str, Any]) -> str:
...
-            )
-        ]
-        for service in status["services"]:
-            descr_override = service.get("description_override")
-            description = (
-                descr_override if descr_override else service["description"]
-            )
-            content.append(
+        if not status["services"]:
+            content = [messages.STATUS_NO_SERVICES_AVAILABLE]
```
lucasmoura, line 628: I think can just return the `content` here directly

uaclient/status.py
```diff
@@ -659,26 +670,30 @@ def format_tabular(status: Dict[str, Any]) -> str:
...
-        }
-        warning = service_status.get("warning", None)
-        if warning is not None:
-            warning_message = warning.get("message", None)
-            if warning_message is not None:
-                service_warnings.append(warning_message)
-        content.append(STATUS_TMPL.format(**fmt_args))
-    tech_support_level = status["contract"]["tech_support_level"]
+    if not status["services"]:
+        content = [messages.STATUS_NO_SERVICES_AVAILABLE]
```
lucasmoura, line 674: Same idea here

uaclient/status.py
renanrodrigo: Well, there may be notices, for instance, or features to report - so the idea is that this piece of code replaces only the services table.
Is there a specific reason not to show anything else than the message?

lucasmoura: Yeah, my rationale was that if we don't have any services available, there is no other information we can provide. But you are right, maybe some notices will be useful (even though it would be odd for them to appear, but better be safe than sorry)

uaclient/status.py
```diff
@@ -707,6 +725,7 @@ def format_tabular(status: Dict[str, Any]) -> str:
 
     if status["origin"] != "free":
         pairs.append(("Valid until", format_expires(status["expires"])))
+        tech_support_level = status["contract"]["tech_support_level"]
```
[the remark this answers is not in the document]
renanrodrigo: well I just copypastaed this here from above - I believe those filelds are mandatory when attached, we were accessing those directly already. If we want to `.get` the status dict data, there are more entries being directly accessed, and I think it would be better to address that separately/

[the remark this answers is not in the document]
lucasmoura: I agree we are doing that on other parts of the code (which we also need to change), but since we are touching that part already, I think we should already make the change. 

And I agree those fields are mandatory, but thinks can change and weird scenarios can happen, so I think we should be defensive here just in case

orndorffgrant (approved):
LGTM and I added on one more commit to address Lucas' comment

Issue #2178, opened 2022-07-29
docs request: explanation of ua status --format json output


orndorffgrant:
Similar to https://github.com/canonical/ubuntu-advantage-client/blob/main/docs/source/explanations/how_to_interpret_the_security_status_command.md but for `ua status --format json`.

orndorffgrant:
This should help people answer questions like “Is this machine covered by an active subscription with Canonical” or "Do all my packages get security updates today"

orndorffgrant:
Tracking in SC-1213

PR #2550, opened 2023-04-13
docs: explain json output of status

orndorffgrant:
no-lp no-jira

Fixes: #2178

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/docs/explanations/status_columns.md
+++ b/docs/explanations/status_columns.md
@@ -101,3 +101,105 @@ allow_beta: True
 It's important to keep in mind that any feature defined like this will be
 listed, even if it is invalid or typed the wrong way. Those appear in `status`
 output for informational and debugging purposes.
+
+## Machine-readable output
+
+The `pro status` command supports a `--format` flag with options including `json` and `yaml`. These result in a machine-readable form of the information presented by the `pro status` command.
+
+```{note}
+`pro status` should return the same results whether using `sudo` or not, but earlier versions did not always do this. We recommend using `sudo` whenever possible.
+```
+
+For example, running `sudo pro status --format=json` on an attached machine may give you something like this:
+```javascript
+{
+  "_doc": "Content provided in json response is currently considered Experimental and may change",
+  "_schema_version": "0.1",
+  "account": {
+    "created_at": "2000-01-02T03:04:05+06:00",
+    "id": "account_id",
+    "name": "Test"
+  },
+  "attached": true,
+  "config": { ...effectiveConfiguration },
+  "config_path": "/etc/ubuntu-advantage/uaclient.conf",
+  "contract": {
+    "created_at": "2000-01-02T03:04:05+06:00",
+    "id": "contract_id",
+    "name": "contract_name",
+    "products": [ "uaa-essential" ],
+    "tech_support_level": "essential"
+  },
+  "effective": null,
+  "environment_vars": [...proClientEnvironmentVariables],
+  "errors": [],
+  "execution_details": "No Ubuntu Pro operations are running",
+  "execution_status": "inactive",
+  "expires": "9999-12-31T00:00:00+00:00",
+  "features": {},
+  "machine_id": "machine_id",
+  "notices": [],
+  "result": "success",
+  "services": [
+    {
+      "available": "yes",
+      "blocked_by": [],
+      "description": "Expanded Security Maintenance for Applications",
+      "description_override": null,
+      "entitled": "yes",
+      "name": "esm-apps",
+      "status": "enabled",
+      "status_details": "Ubuntu Pro: ESM Apps is active",
+      "warning": null
+    },
+    {
+      "available": "yes",
+      "blocked_by": [],
+      "description": "Expanded Security Maintenance for Infrastructure",
+      "description_override": null,
+      "entitled": "yes",
+      "name": "esm-infra",
+      "status": "enabled",
+      "status_details": "Ubuntu Pro: ESM Infra is active",
+      "warning": null
+    },
+    {
+      "available": "yes",
+      "blocked_by": [],
+      "description": "Canonical Livepatch service",
+      "description_override": null,
+      "entitled": "yes",
+      "name": "livepatch",
+      "status": "enabled",
+      "status_details": "",
+      "warning": null
+    },
+    ...otherServiceStatusObjects
+  ],
+  "simulated": false,
+  "version": "27.13.6~18.04.1",
+  "warnings": []
+}
+```
+
+Some particularly important attributes in the output include:
+* `attached`: This boolean value indicates whether this machine is attached to an Ubuntu Pro account. This does not tell you if any particular service (e.g. `esm-infra`) is enabled. You must check the individual service item in the `services` list for that status (described below).
+* `expires`: This is the date that the Ubuntu Pro subscription is valid until (in RFC3339 format). After this date has passed the machine should be treated as if not attached and no services are enabled. `attached` may still say `true` and services may still say they are `entitled` and `enabled`, but if the `expires` date has passed, you should assume the services are not functioning.
+* `services`: This is a list of Ubuntu Pro services. Each item has its own attributes. Widely applicable services include those with `name` equal to `esm-infra`, `esm-apps`, and `livepatch`. Some important fields in each service object are:
+  * `name`: The name of the service.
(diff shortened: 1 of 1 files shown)


docs/explanations/status_columns.md
```diff
@@ -101,3 +101,98 @@ allow_beta: True
 It's important to keep in mind that any feature defined like this will be
 listed, even if it is invalid or typed the wrong way. Those appear in `status`
 output for informational and debugging purposes.
+
+## Machine Readable Output
```
s-makin, line 105:
```suggestion
## Machine-readable output
```

docs/explanations/status_columns.md
```diff
@@ -101,3 +101,98 @@ allow_beta: True
...
+
+The `pro status` command supports a `--format` flag with options including `json` and `yaml`. These result in a machine-readable form of the information presented in the `pro status` command.
```
s-makin, line 107:
```suggestion
The `pro status` command supports a `--format` flag with options including `json` and `yaml`. These result in a machine-readable form of the information presented by the `pro status` command.
```

docs/explanations/status_columns.md
```diff
@@ -101,3 +101,98 @@ allow_beta: True
...
+
+```{note}
+`pro status` should return the same results with and without sudo/root, but earlier versions did not always and we would still recommend using sudo if possible.
```
s-makin, line 110:
```suggestion
`pro status` should return the same results whether using `sudo` or not, but earlier versions did not always do this. We recommend using `sudo` whenever possible.
```

docs/explanations/status_columns.md
```diff
@@ -101,3 +101,98 @@ allow_beta: True
...
+    ...otherServiceStatusObjects
+  ],
+  "simulated": false,
+  "version": "27.13.6~18.04.1",
+  "warnings": []
+}
+```
+
+Some particularly important attributes in the output include:
+* `attached`: This boolean value indicates whether this machine is attached to an Ubuntu Pro account or not. Note that this does not imply that any particular service (e.g. esm-infra) is enabled. You must check the individual service object for that status.
```
s-makin, line 183:
```suggestion
* `attached`: This boolean value indicates whether or not this machine is attached to an Ubuntu Pro account. This does not tell you if any particular service (e.g. `esm-infra`) is enabled. You must check the individual service object for that status.
```

s-makin: As a quick note, I'm not sure about the wording for the final sentence - I've not heard of a "service object" before so I'm not sure what the sentence means. Is there a clearer way we can express this?

docs/explanations/status_columns.md
```diff
@@ -101,3 +101,98 @@ allow_beta: True
...
+* `expires`: This is the date that the Ubuntu Pro subscription is valid until.
+* `services`: This is a list of Ubuntu Pro services. Each item has its own attributes. Widely applicable services include those with `name` equal to `esm-infra`, `esm-apps`, and `livepatch`. Some important fields in each service object are:
+  * `name`: The name of the service.
+  * `entitled`: A boolean indicating whether the attached Ubuntu Pro account is allowed to enable this service.
+  * `status`: An enum value indicating the services current status on the machine. Possible values are `enabled`, `disabled`, `n/a`, and `warning`. Any value other than `enabled` should be treated as if the service is not enabled and not working properly on the machine.
```
s-makin, line 188:
```suggestion
  * `status`: An enum value indicating the service's current status on the machine. Possible values are `enabled`, `disabled`, `n/a`, and `warning`. Any value other than `enabled` should be treated as if the service is not enabled and not working properly on the machine.
```

docs/explanations/status_columns.md
```diff
@@ -101,3 +101,98 @@ allow_beta: True
...
+
+For example, if you want to programatically determine if esm-infra is enabled and working on a particular machine, you can use the following command:
```
s-makin, line 190:
```suggestion
For example, if you want to programatically determine if `esm-infra` is enabled and working on a particular machine, you can use the following command:
```

s-makin, line 190:
```suggestion
For example, if you want to determine if esm-infra is enabled and working on a particular machine, you can use the following command:
```

docs/explanations/status_columns.md
```diff
@@ -101,3 +101,98 @@ allow_beta: True
...
+```shell
+sudo pro status --format=json | jq '.services[] | select(.name == "esm-infra").status == "enabled"'
+```
+That command will print `true` if esm-infra is enabled, or `false` if it is not.
+
```
s-makin, line 195: Side question - what should they do if it's disabled and they expected it to be enabled and working?

docs/explanations/status_columns.md
orndorffgrant: I reworded it - let me know if it makes more sense now.

docs/explanations/status_columns.md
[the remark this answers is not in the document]
orndorffgrant: It's jargon for "there's a finite list of enumerated values that could be here". I changed it to be simpler

docs/explanations/status_columns.md
orndorffgrant: It depends on what the value is.
If it's "disabled", then run `pro enable`.
If it's `n/a` then you don't have the option to enable it.
If it's `warning` then something specific is wrong and you need to go read the warning detailed message to figure out what to do.
I'm not sure if all of that belongs here or not - I could be convinced either way - what do you think?

docs/explanations/status_columns.md
```diff
@@ -101,3 +101,98 @@ allow_beta: True
 It's important to keep in mind that any feature defined like this will be
 listed, even if it is invalid or typed the wrong way. Those appear in `status`
 output for informational and debugging purposes.
+
+## Machine-readable Output
```
s-makin, line 105:
```suggestion
## Machine-readable output
```

docs/explanations/status_columns.md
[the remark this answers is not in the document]
s-makin: Much clearer!

docs/explanations/status_columns.md
s-makin: I might have misunderstood, but the command reads to me like you're only checking if it's enabled (or not) - does the output actually tell you more than that whether it's available-but-disabled, n/a, or broken? So my initial question was more along the lines of "what would they need to follow up with if they get "false" as the result?"

docs/explanations/status_columns.md
orndorffgrant: fixed for real this time :)

docs/explanations/status_columns.md
orndorffgrant: I think I see your concern now. I've modified the example and the preceding text a bit. LMK what you think

docs/explanations/status_columns.md
```diff
@@ -101,3 +101,105 @@ allow_beta: True
...
+  ],
+  "simulated": false,
+  "version": "27.13.6~18.04.1",
+  "warnings": []
+}
+```
+
+Some particularly important attributes in the output include:
+* `attached`: This boolean value indicates whether this machine is attached to an Ubuntu Pro account. This does not tell you if any particular service (e.g. `esm-infra`) is enabled. You must check the individual service item in the `services` list for that status (described below).
+* `expires`: This is the date that the Ubuntu Pro subscription is valid until (in RFC3339 format). After this date has passed the machine should be treated as if not attached and no services are enabled. `attached` may still say `true` and services may still say they are `entitled` and `enabled`, but if the `expires` date has passed, you should assume the services are not functioning.
```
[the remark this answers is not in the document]
s-makin: That's a strange behaviour -- does it not warn the user that the services are inactive when the expiry date is exceeded? Or at least prompt them to resubscribe?

(The amended wording is fine btw, I'm just trying to understand why it does that)

docs/explanations/status_columns.md
s-makin: Makes much more sense! Thanks for changing it :)

s-makin (approved):
LGTM. My final question isn't about the contents of the PR, we can talk about that separately.

docs/explanations/status_columns.md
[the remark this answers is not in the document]
orndorffgrant: Elsewhere (in APT and MOTD) yes, but not in the output of this command. This is a known issue but I don't think we had something to track it before so just created #2553

Issue #2527, opened 2023-03-31
apt-get drops system proxy variables


leungtimothy:
# Issue
`pro` commands fail in our environment due to `apt-get` commands being run without the proxy environment variables. 

The provided proxy config settings do not work for us. We have repos that use transport methods that require proxy vars.  ([apt-transport-s3](https://github.com/MayaraCloud/apt-transport-s3)). These transport methods rely on the `http_proxy`, `https_proxy` and `no_proxy` variables.

# Code
The issue is the call to system.subp. The env argument is either `{}` or contains `DEBIAN_FRONTEND=noninteractive` when calling `run_apt_command`.

Either scenario drops the system environment variables. Per the python docs:
> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment. It is passed directly to [Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen). This mapping can be str to str on any platform or bytes to bytes on POSIX platforms much like [os.environ](https://docs.python.org/3/library/os.html#os.environ) or [os.environb](https://docs.python.org/3/library/os.html#os.environb).

https://github.com/canonical/ubuntu-pro-client/blob/main/uaclient/apt.py#L186
```
def run_apt_command(
    cmd: List[str],
    error_msg: Optional[str] = None,
    env: Optional[Dict[str, str]] = {},
) -> str:
    """Run an apt command, retrying upon failure APT_RETRIES times.
    :param cmd: List containing the apt command to run, passed to subp.
    :param error_msg: The string to raise as UserFacingError when all retries
       are exhausted in failure.
    :param env: Optional dictionary of environment variables to pass to subp.
    :return: stdout from successful run of the apt command.
    :raise UserFacingError: on issues running apt-cache policy.
    """
    try:
        out, _err = system.subp(
            cmd, capture=True, retry_sleeps=APT_RETRIES, env=env
        )
```

# Proposed Solution
The call to `subp` should preserve the proxy variables at the very least. We have patched our own client by doing the following:

```
        merged_env = {**env, **os.environ}
        out, _err = system.subp(
            cmd, capture=True, retry_sleeps=APT_RETRIES, env=merged_env
        )
```

orndorffgrant:
Tracking in SC-1529

PR #2551, opened 2023-04-13
subp: use system environment vars by default (SC-1529)

lucasmoura:
no-lp

## Proposed Commit Message
subp: use system environment vars by default

We are now always using the default system environment variable when running a subprocess command. This will allow user setting to be propagated to other tools we run during some Pro operations.

Fixes: #2527

## Test Steps
Run the new unit test

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/uaclient/system.py
+++ b/uaclient/system.py
@@ -429,8 +429,13 @@ def _subp(
     bytes_args = [
         x if isinstance(x, bytes) else x.encode("utf-8") for x in args
     ]
-    if env:
+
+    # If env is None, subprocess.Popen will use the process environment
+    # variables by default, as stated here:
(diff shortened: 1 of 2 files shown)


uaclient/system.py
```diff
@@ -431,8 +431,11 @@ def _subp(
     bytes_args = [
         x if isinstance(x, bytes) else x.encode("utf-8") for x in args
     ]
-    if env:
-        env.update(os.environ)
```
orndorffgrant, line 435: I think the bug is that `{}` is falsey, so it currently won't get the os env.

So an alternative fix would be to replace this line with:
```python
if env is not None:
```

orndorffgrant (approved):
This looks good, but I'm also proposing a slightly simpler (IMO) fix.

And good idea on adding a unit-test for this!

PR #2555, opened 2023-04-17
Add XML sitemap to docs

s-makin:
```
Add XML sitemap for Google Search Console

GSC has not been able to correctly find more than 12 pages in
our docs because the default RTD sitemap is pointing to a URL
Google doesn't expect.

Adding a new sitemap specifically for GSC so that the docs
can be crawled and then reliably listed in search results.
```

## Checklist

 - [ ] Yes
 - [x] No


CalvoM (approved):

orndorffgrant:
In order to get included in the build, we need to add this to `html_extra_path` in `conf.py`.

Also, this will end up at `https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/sitemap-index.xml` - is that what we want?
I imagine we'd need to configure GSC to look there somehow. But if we can do that then couldn't we configure them to look at `https://canonical-ubuntu-pro-client.readthedocs-hosted.com/sitemap.xml` where this is currently hosted?

I guess I'm just confused again as to what this solves and how. Let's sync on it tomorrow.

cpaelzer:
> In order to get included in the build, we need to add this to
> html_extra_path in conf.py.
>
> Good catch

> Also, this will end up at
> https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/sitemap-index.xml
> - is that what we want?
> I imagine we'd need to configure GSC to look there somehow. But if we can
> do that then couldn't we configure them to look at
> https://canonical-ubuntu-pro-client.readthedocs-hosted.com/sitemap.xml
> where this is currently hosted?
>
AFAIU it is vice versa.
It already is (by default) at
https://canonical-ubuntu-pro-client.readthedocs-hosted.com/sitemap.xml but
needs to be at
https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/sitemap-index.xml
to work as this is the entry level URL into our docs.

I guess we should document (remember our talk in the commit message) "why"
it needs to be on that path.

> I guess I'm just confused again as to what this solves and how. Let's sync
> on it tomorrow.
>
Agreed - worth discussing before landing.
I won't be around, but I'm sure you and Sally will sort  it out

PR #2558, opened 2023-04-19
logging_revamp replace module loggers SC-1539

CalvoM:
## Proposed Commit Message

Logging revamp replace module loggers

NO-LP NO-GH

Add logger in modules instead of the root logger.

## Checklist

 - [x] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any integration tests accordingly
 - [ ] I have updated or added any documentation accordingly

## Does this PR require extra reviews?

 - [ ] Yes
 - [x] No

PR #2563, opened 2023-04-21
logging_revamp: refactor logging tests SC-1541

CalvoM:
## Proposed Commit Message

logging_revamp: refactor logging tests

NO-LP NO-GH

Refactor testing scenarios for logging.

## Checklist

 - [ ] Yes
 - [x] No


CalvoM:
I would recommend that this is passed after [non-root logging PR](https://github.com/canonical/ubuntu-pro-client/pull/2503). I wrote this with the changes in that PR in mind.

uaclient/tests/test_cli.py
```diff
@@ -779,179 +800,81 @@ def test_argparse_errors_well_formatted(
...
+        log_level = logging.DEBUG
+        logger = logging.getLogger("logger_a")
+        with mock.patch(
+            "uaclient.cli.config.UAConfig", return_value=FakeConfig()
+        ):
+            setup_logging(console_level, log_level, logger=logger)
+        assert len(logger.handlers) == 2
+        assert logger.handlers == get_logger_handler(
+            logger, console_level, log_level
+        )
```
orndorffgrant, line 832: Instead of these assertions, could we do something more like this?
```
assert repr(logger.handlers) == repr([logging.StreamHandler(sys.stderr), logging.FileHandler(log_file)])
```
(IDK if that will work, but something like that would be nice)

orndorffgrant:
@CalvoM Since we've landed the non-root logging PR, should we finish up this one before proceeding with #2558 ?

uaclient/tests/test_cli.py
CalvoM: Quick question, is there a specific reason for this specific test?

orndorffgrant: Looking closer, this particular test is probably not needed because the code it is testing is so simple IMO

uaclient/tests/test_cli.py
```diff
@@ -828,208 +849,76 @@ def test_status_human_readable_warning(
...
-    def test_console_log_configured_if_not_present(
-        self,
-        m_get_user,
-        m_we_are_currently_root,
-        level,
-        capsys,
-        logging_sandbox,
-        FakeConfig,
-        tmpdir,
+    def test_console_log_outputs_to_stderr(
```
CalvoM, line 852: remove this as well.

uaclient/tests/test_cli.py
CalvoM: After discussion, we saw it is better to check the types of the handlers and the number of the handlers.

CalvoM: We could combine with the `test_previous_handlers_removed_in_logger` method.

uaclient/tests/test_cli.py
```diff
@@ -828,208 +849,76 @@ def test_status_human_readable_warning(
...
+            setup_logging(console_level, log_level, logger=logger)
 
-        setup_logging(logging.INFO, logging.INFO)
-        logging.info("after setup")
+        assert len(logger.handlers) == 2
+        assert logger.handlers == get_logger_handler(
+            logger, console_level, log_level
+        )
 
+    def test_log_file_created_if_not_present(self, logging_sandbox, tmpdir):
```
CalvoM, line 902: mock the file creation methods and combine with `test_log_file_has_correct_permissions`

uaclient/tests/test_cli.py
```diff
@@ -828,208 +829,54 @@ def test_status_human_readable_warning(
...
-        assert "ERROR: after setup" in err
+            setup_logging(console_level, log_level, logger=logger)
+        assert len(logger.handlers) == 2
+        for idx, handler in enumerate(logger.handlers):
+            if idx == 0:
+                assert handler.name == "ua-console"
+                assert handler.level == console_level
+            elif idx == 1:
+                assert handler.name == "ua-file"
+                assert handler.level == log_level
```
orndorffgrant, line 857: IMO this is a bit easier to read
```suggestion
        assert logger.handlers[0].name == "ua-console"
        assert logger.handlers[0].level == console_level
        assert logger.handlers[1].name == "ua-file"
        assert logger.handlers[1].level == log_level
```

uaclient/tests/test_cli.py
```diff
@@ -828,208 +829,54 @@ def test_status_human_readable_warning(
...
+        logger = logging.getLogger("logger_b")
+        log_file = tmpdir.join("log_file").strpath
+        setup_logging(
+            logging.DEBUG,
+            logging.INFO,
+            log_file=log_file,
+            logger=logger,
+        )
+        assert m_path_touch.call_count == 1
+        assert m_path_touch.call_args_list == [mock.call(mode=0o640)]
```
orndorffgrant, line 870: asserting the call_args_list is what it is means that it was only called once, so no need for the call_count assertion
```suggestion
        assert m_path_touch.call_args_list == [mock.call(mode=0o640)]
```

uaclient/tests/test_cli.py
CalvoM: I had implemented this way, and though it was a bit manual.

orndorffgrant: It does feel more manual, but there is no need for anything more complex. I don't think that feeling is a bad thing.

IMO the goal of tests is to minimize abstraction and do things directly as possible. I think the overhead of reading a for loop and figuring out that that it asserts something differently for two different indexes makes the test harder to read. (aside: more generally, I consider any loop or if-statement to be a code-smell in a test)

In both versions of this code, there is an assumption that there are only two items in the list (as asserted on the line above). I think this assumption is more obvious in the manual version.

uaclient/tests/test_config.py
```diff
@@ -1155,14 +1153,21 @@ class TestFeatures:
...
+        warnings,
+        FakeConfig,
     ):
         user_cfg = {"features": cfg_features}
         cfg = FakeConfig(cfg_overrides=user_cfg)
         assert expected == cfg.features
         if warnings:
-            assert warnings in caplog_text()
+            expected = [mock.call(warnings, cfg_features)]
+            assert expected == m_log_warning.call_args_list
```
orndorffgrant, line 1170: It doesn't have any adverse affects here, but let's keep good practices. So don't overwrite the `expected` argument. Also no need to have the intermediate variable anyway
```suggestion
            assert [mock.call(warnings, cfg_features)] == m_log_warning.call_args_list
```

uaclient/tests/test_cli.py
```diff
@@ -828,208 +829,50 @@ def test_status_human_readable_warning(
...
-    @mock.patch("uaclient.log.get_user_log_file")
-    @mock.patch("uaclient.cli.util.we_are_currently_root", return_value=False)
-    def test_user_file_log_configured_if_not_root(
-        self,
-        m_we_are_currently_root,
-        m_log_get_user_log_file,
-        tmpdir,
-        logging_sandbox,
-    ):
+    def test_log_file_has_correct_permissions(self, tmpdir, logging_sandbox):
```
orndorffgrant, line 868: I think we discussed removing this one since it's been combined with the above test.

orndorffgrant (approved):
Two minor nits, but LGTM

uaclient/tests/test_cli.py
CalvoM: Oh, I forgot this.

uaclient/tests/test_config.py
CalvoM: Updated

renanrodrigo (approved):

Issue #2567, opened 2023-04-25
`/tmp/ubuntu-advantage/livepatch-kernel-support-cache.json` cannot be overwritten by non-root users once created


orndorffgrant:
Title is both a problem and an assertion.

If the ubuntu-advantage tmp dir and file are created by root, then non-root `pro status` will fail because it cannot update the cache file.

We cannot set the file to be world-writeable because that is considered bad practice and will cause alerts in security scanners.

The medium-term solution is probably that every user should get their own tmp cache directory and files.

The short-term solution is probably that only root will cache this data. If we see we're non-root we won't even try to update the cache.

PR #2568, opened 2023-04-26
cache: use user owned cache for livepatch support

orndorffgrant:
Fixes: #2567

no-jira no-lp

Designed so that it should integrate nicely with #2503 
## Test Steps

`time pro status`
see /home/ubuntu/.cache/ubuntu-pro/livepatch-kernel-support-cache.json
`time pro status` - much shorter because cache was used now

change the kernel and reboot
`time pro status` - longer again because cache was invalid
see /home/ubuntu/.cache/ubuntu-pro/livepatch-kernel-support-cache.json was updated
`time pro status` - much shorter because cache was used now

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/features/livepatch.feature
+++ b/features/livepatch.feature
@@ -5,6 +5,10 @@ Feature: Livepatch
     @uses.config.machine_type.lxd.vm
     Scenario Outline: Attached livepatch status shows warning when on unsupported kernel
         Given a `<release>` machine with ubuntu-advantage-tools installed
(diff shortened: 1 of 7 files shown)


lucasmoura (approved):
I think this change makes sense. We can also remove the `tmp` cache file just to cleanup the system, but maybe this is not needed

renanrodrigo (approved):
Once the changes in #2503 land, there will be small adjustments to the defaults file, and some functionality may overlap (get_user_cache_dir?)

this one good

PR #2570, opened 2023-04-27
Release 27.15

orndorffgrant:
Waiting on:
- https://github.com/canonical/ubuntu-pro-client/pull/2539
- https://github.com/canonical/ubuntu-pro-client/pull/2568

Combing through LP for bugs not referenced but fixed:
- ~~1994923~~ - added
- possibly 2017350 after #2568 - this bug is still incomplete because we don't have enough information, so we won't list it as fixed by this release

## Test Steps

Ensure this matches `main` except for the changelog addition

* Is the version correctly updated? We must ensure that the new version in the changelog is
  correct and it also targets the latest Ubuntu release at the moment.
* Is the entry useful for the user? The changelog entries should be user focused, meaning
  that we should only add entries that we think users will care about (i.e. we don't need
  entries when fixing a test, as this doesn't provide meaningful information to the user).
* Is this entry redundant? Sometimes we may have changes that affect separate modules of the
  code. We should have an entry only for the module that was most affected by it.
* Is the changelog entry unique? We need to verify that the changelog entry is not already
  reflected in an earlier version of the changelog. If it is, we need not only to remove but double
  check the process we are using to cherry-pick the commits.
* Is this entry actually reflected in the code? Sometimes, we can have changelog entries
  that are not reflected in the code anymore. This can happen during development when we are
  still unsure about the behaviour of a feature or when we fix a bug that removes the code
  that was added. We must verify each changelog entry that is added to be sure of their
  presence in the product.


renanrodrigo (commented):
Interesting how we have almost no launchpad bug in this release changelog
I remember though some GH issues were closed - shouldn't those be mentioned in the changelog? they were before.

renanrodrigo (approved):

orndorffgrant:
> Interesting how we have almost no launchpad bug in this release changelog I remember though some GH issues were closed - shouldn't those be mentioned in the changelog? they were before.

My bad! I've added the GH references

debian/changelog
```diff
@@ -1,3 +1,41 @@
...
+    - apt messaging: add bionic-specific urls
+    - config: no longer load uaclient.conf from current working directory
+    - fix:
+      + add support for --no-related flag
+      + separate target USN from related USNs
+    - general:
+      + logs to user cache directory when run as non-root
+      + fix bug where non-root commands failed with file permission error
+        accessing /tmp/ubuntu-advantage (GH: #2567)
+      + use system environment vars by default in sub processes
```
lucasmoura, line 16: I think we missing the github issue for this entry:
https://github.com/canonical/ubuntu-pro-client/issues/2527

lucasmoura (approved):

debian/changelog
orndorffgrant: Thanks! Just fixed

PR #2578, opened 2023-05-05
ci: change bug-refs action to post comment

orndorffgrant:
## Why is this needed?
This PR is exploring adding a comment to each PR explaining the results of the bug-refs (future "does doc PR also exist") action.

We no-longer need to write no-jira no-lp no-gh. Instead we'll just ack the comment, and part of review is making sure the comment is acked and correct.


orndorffgrant:
Test comment

renanrodrigo (approved):

PR #2583, opened 2023-05-15
Add new endpoint (SC-1514)

lucasmoura:
no-lp no-gh

## Why is this needed?
We are adding two distinct endpoint that deliver information that can only be seen today through the machine readable output of `pro status`. They are:

* **is_attached**: Verify if the machine is attached to a pro subscription
* **enabled_services**: Display which services are enabled in the machine

## Test Steps

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/docs/references/api.md
+++ b/docs/references/api.md
@@ -110,6 +110,8 @@ The currently available endpoints are:
 - [u.pro.security.status.reboot_required.v1](#uprosecuritystatusreboot_requiredv1)
 - [u.pro.packages.summary.v1](#upropackagessummaryv1)
 - [u.pro.packages.updates.v1](#upropackagesupdatesv1)
+- [u.pro.status.is_attached.v1](#uprostatusis_attachedv1)
+- [u.pro.status.enabled_services.v1](#uprostatusenabled_servicesv1)
 - [u.security.package_manifest.v1](#usecuritypackage_manifestv1)
 - [u.unattended_upgrades.status.v1](#uunattended_upgradesstatusv1)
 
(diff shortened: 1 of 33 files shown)


docs/references/api.md
```diff
@@ -776,6 +778,70 @@ pro api u.pro.packages.updates.v1
 }
 ```
 
+## u.pro.status.is_attached.v1
+
+Introduced in Ubuntu Pro Client Version: `27.15~`
```
orndorffgrant, line 783: This will be in version 28 I think

docs/references/api.md
```diff
@@ -776,6 +778,70 @@ pro api u.pro.packages.updates.v1
...
+
+Shows if the machine is attached to a Pro subscription.
+
+### Args
+
+This endpoint takes no arguments.
+
+### Python API interaction
```
orndorffgrant, line 791: I think we need "CLI interaction" sections as well

docs/references/api.md
```diff
@@ -776,6 +778,70 @@ pro api u.pro.packages.updates.v1
...
+
+#### Calling from Python code
+
+```python
+from uaclient.api.u.pro.status.enabled_services.v1 import is_attached
```
orndorffgrant, line 824:
```suggestion
from uaclient.api.u.pro.status.enabled_services.v1 import enabled_services
```

docs/references/api.md
```diff
@@ -776,6 +778,70 @@ pro api u.pro.packages.updates.v1
...
+|-|-|-|
+|`enabled_services`|*List[EnabledService]*|A list of EnabledServices objects|
+
+`uaclient.api.u.pro.status.enabled_services.v1.EnabledService`
+
+|Field Name|Type|Description|
+|-|-|-|
+|`name`           |*str* |name of the service                         |
+|`variant_enabled`|*bool*|If a variant of the service is enabled      |
+|`variant_name`   |*str* |name of the variant, if a variant is enabled|
```
orndorffgrant, line 843: We could make this one Optional field instead
```suggestion
|`variant_name`   |*Optional[str]* |name of the variant if a variant is enabled, or None if no variant is enabled|
```

features/api.feature
```diff
@@ -43,6 +43,16 @@ Feature: Client behaviour for the API endpoints
...
     """
+    When I run `ua api u.pro.status.is_attached.v1` with sudo
+    Then stdout matches regexp:
+    """
+    {"_schema_version": "v1", "data": {"attributes": {"is_attached": false}, "meta": {"environment_vars": \[\]}, "type": "IsAttached"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
+    """
+    When I run `ua api u.pro.status.enabled_services.v1` with sudo
+    Then stdout matches regexp:
+    """
+    {"_schema_version": "v1", "data": {"attributes": {"enabled_services": null}, "meta": {"environment_vars": \[\]}, "type": "EnabledServices"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
```
orndorffgrant, line 54: I think this should be an empty list in this case rather than null.

features/cloud_pro_clone.feature
```diff
@@ -13,10 +13,10 @@ Feature: Creating golden images based on Cloud Ubuntu Pro instances
         log_file: /var/log/ubuntu-advantage.log
         """
         When I run `pro auto-attach` with sudo
-        And I run `pro status --format yaml` with sudo
+        And I run `pro api u.pro.status.is_attached.v1` as non-root
         Then stdout matches regexp:
         """
-        attached: true
+        \"is_attached\": true
```
orndorffgrant, line 19: This comment is very nitpicky so feel free to ignore it :upside_down_face: 

I think it would be just slightly nicer to use a substring check for this instead of a regexp:
```
        Then stdout contains substring:
        """
        "is_attached": true
```

Or if we want to be more ambitious, we could write a step: "Then the machine is attached" that does the API call under-the-hood and actually parses the json and checks the value is true.

uaclient/api/u/pro/status/is_attached/v1.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, *, is_attached: bool):
+        self.is_attached = is_attached
+
+
+def is_attached() -> IsAttachedResult:
+    return _is_attached(UAConfig())
+
+
+def _is_attached(cfg: UAConfig) -> IsAttachedResult:
+    return IsAttachedResult(is_attached=cfg.is_attached)
```
orndorffgrant, line 21: Maybe this is too much for this PR, but this is also an opportunity to decouple `is_attached` from UAConfig.

We could directly check for the existence of the machine token file here and remove the property from UAConfig entirely - thoughts?

orndorffgrant (commented):
Looking good! I'm excited to continue to move this kind of functionality into API functions. Who do we know will start using these? We should get them to review this PR as well

features/cloud_pro_clone.feature
lucasmoura: I like the idea of the new step, as it will be easier to reuse. I will implement it

uaclient/api/u/pro/status/is_attached/v1.py
lucasmoura: Good idea, I think that is doable

lucasmoura:
@orndorffgrant updated

renanrodrigo (approved):
Nice step forward in the API, I like it

PR #2584, opened 2023-05-17
ignore any unicode char if we can't print it

lucasmoura:
no-gh no-jira

## Why is this needed?
If we cannot handle unicode characters in the output we remove any of those characters from the message.
This is to guarantee that users that are not using UTF-8 by default can still use the tool

## Test Steps
Run the new unittest added

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (approved):

renanrodrigo:
Idea for the future: make our print functions always use the handler. There may be more unicode here and there, potentially many more with translations.

PR #2588, opened 2023-05-19
security-status: don't call status

renanrodrigo:
## Why is this needed?

This PR solves all of our problems because we have bugs open about security status, or more precisely the updates API endpoint, not working without an internet connection.
That's because we call `status` while unattached, and try to hit the unauthenticated endpoint.
Now we don't anymore.

Fixes: #2536
LP: #2015286

## Test Steps
- cut your access to `contracts.canonical.com`
- run `pro security status` and see now it runs fine
- run `pro api u.pro.packages.updates.v1` and see now it runs fine

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/security_status.py
+++ b/uaclient/security_status.py
@@ -10,6 +10,7 @@ from typing import Any, DefaultDict, Dict, List, Tuple, Union  # noqa: F401
 import apt  # type: ignore
 
 from uaclient import livepatch, messages
+from uaclient.api.u.pro.status.is_attached.v1 import _is_attached
 from uaclient.apt import (
     PreserveAptCfg,
     get_apt_cache,
@@ -21,8 +22,8 @@ from uaclient.entitlements import ESMAppsEntitlement, ESMInfraEntitlement
 from uaclient.entitlements.entitlement_status import (
     ApplicabilityStatus,
     ApplicationStatus,
+    ContractStatus,
 )
-from uaclient.status import status
 from uaclient.system import (
     REBOOT_PKGS_FILE_PATH,
     get_distro_info,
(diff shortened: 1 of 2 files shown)


lucasmoura (approved):

lucasmoura:
CI errors are unrelated to this fix

uaclient/security_status.py
```diff
@@ -192,15 +193,15 @@ def filter_security_updates(
...
+        "attached": is_attached,
         "enabled_services": [],
         "entitled_services": [],
     }  # type: Dict[str, Any]
 
-    status_dict = status(cfg=cfg, show_all=True)
-    if status_dict["attached"]:
-        ua_info["attached"] = True
+    if is_attached:
+        status_dict = status(cfg=cfg, show_all=True)
```
orndorffgrant, line 204: This is already a great improvement and I think it would be good to merge this.
I think we could go a step further and directly use `ESMAppsEntitlement` and `ESMAppsEntitlement` here instead of `status`. One downside of `status` is that it gets the status of all services, which can include, for example, talking to the livepatch api (though the response is cached).
But if you think that's too much for this PR, then we can do it at a later time.

orndorffgrant (approved):

uaclient/security_status.py
renanrodrigo: not too much - I think it makes sense to be done here right away

renanrodrigo:
@orndorffgrant addressed

orndorffgrant:
@renanrodrigo looks like a small test fix is needed:
```diff
And stdout matches regexp
      """
-      "enabled_services": \["esm-apps", "esm-infra"\]
+      "enabled_services": \["esm-infra", "esm-apps"\]
```

After that I vote to go ahead and merge :rocket:

renanrodrigo:
/me cries
I will change the append order instead, alphabetical ftw

PR #2589, opened 2023-05-22
livepatch: support new kernels api (SC-1538)

orndorffgrant:
## Why is this needed?

The behavior of the livepatch API endpoint for querying kernel support
is changing without bumping the version. The new version is currently on
staging and the old version is blocked from going to production by a
bug affecting infrastructure. Because we don't know exactly when the new
version will reach production, but we have an impending scheduled
release, this commit supports both the old and new version at the same
time.

## Test Steps

```
./tools/test-in-multipass.sh xenial
python3
>>> import os
>>> from uaclient import system
>>> system._get_kernel_build_date(os.uname())
>>> system._get_kernel_changelog_timestamp(os.uname())
```
```
tox -e behave-any -- -n "Unattached livepatch status"
```

## Checklist

 - [x] Yes - livepatch team - done
 - [ ] No


uaclient/system.py
```diff
@@ -78,6 +79,59 @@
...
+        dt = datetime.datetime.strptime(date_str, "%a %b %d %H:%M:%S %Z %Y")
+    except ValueError:
+        with util.disable_log_to_console():
+            logging.warning("Unable to parse build date from uname version")
+        return _get_kernel_changelog_timestamp(uname)
+    if dt.tzinfo is None:
+        # Give it a default timezone if it didn't get one from strptime
+        # The Livepatch API requires a timezone
+        dt = dt.replace(tzinfo=datetime.timezone.utc)
+    return dt
```
orndorffgrant, line 132: @kian99 Can you skim this and confirm that this is the same way that livepatch-client is getting the build-date of the kernel from a machine?

uaclient/livepatch.py
```diff
@@ -210,24 +229,21 @@ def is_kernel_supported(
...
 
-        return bool(result.get("Supported", False))
+        api_supported_val = result.get("Supported")
+        if api_supported_val is None or isinstance(api_supported_val, bool):
+            # old version, True means supported, None means unsupported
+            if api_supported_val:
+                return LivepatchSupport.SUPPORTED
+            return LivepatchSupport.UNSUPPORTED
+        # new version, value is a string
+        return _convert_str_to_livepatch_support_status(api_supported_val)
```
orndorffgrant, line 239: @kian99 For reference, as discussed, this is how we're going to handle both versions of the API. Once the new version hits production, we can remove the compatibility for the old version

uaclient/tests/test_system.py
```diff
@@ -155,23 +193,139 @@ class TestGetKernelInfo:
...
+                ],
+                datetime.datetime(
+                    2023, 4, 6, 6, 35, 51, tzinfo=datetime.timezone.utc
+                ),
+            ),
+        ],
+    )
+    @mock.patch("os.stat")
+    @mock.patch("uaclient.system.is_container")
+    def test_get_kernel_changelog_timestamp(
```
lucasmoura, line 314: Should we add a test that confirm we are failling back to `_get_kernel_changelog_timestamp` if we fail to get the date from the uname result

orndorffgrant: I'm testing this using mock sentinels in the test above this one. See lines: 239-256

uaclient/system.py
kian99: Yip this looks good. The regex is basically identical and the the call to `stat` is grabbing the `Modify` time.

uaclient/livepatch.py
kian99: Gotcha, appreciate Pro client handling this oversight from our side.

uaclient/tests/test_system.py
lucasmoura: Indeed, I missed that strategy

lucasmoura (approved):

renanrodrigo (approved):

PR #2593, opened 2023-05-23
cli: warn users not to use human readable output in scripts (SC-1553)

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because we will change human readable outputs and, if people are scripting it, we want to warn them that there is machine readable output instead.

## Test Steps
Run the new integration test

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -192,8 +192,12 @@ Feature: Command behaviour when unattached
         When I reboot the machine
         Then I verify that no files exist matching `/run/ubuntu-advantage/candidate-version`
         When I run `pro status` with sudo
-        Then I will see the following on stderr
+        Then stderr does not match regexp:
         """
+        .*\[info\].* A new version is available: 99.9.9
+        Please run:
+            sudo apt-get install ubuntu-advantage-tools
+        to get the latest version with new features and bug fixes.
         """
         And I verify that files exist matching `/run/ubuntu-advantage/candidate-version`
         # We forge a candidate to see results
@@ -211,8 +215,12 @@ Feature: Command behaviour when unattached
         to get the latest version with new features and bug fixes.
         """
         When I run `pro status --format json` as non-root
-        Then I will see the following on stderr
+        Then stderr does not match regexp:
         """
+        .*\[info\].* A new version is available: 99.9.9
+        Please run:
+            sudo apt-get install ubuntu-advantage-tools
+        to get the latest version with new features and bug fixes.
         """
         When I run `pro config show` as non-root
         Then stderr matches regexp:
@@ -233,14 +241,22 @@ Feature: Command behaviour when unattached
         \"code\": \"new-version-available\"
         """
         When I run `pro api u.pro.version.v1` as non-root
-        Then I will see the following on stderr
+        Then stderr does not match regexp:
         """
+        .*\[info\].* A new version is available: 99.9.9
+        Please run:
+            sudo apt-get install ubuntu-advantage-tools
+        to get the latest version with new features and bug fixes.
         """
         When I run `apt-get update` with sudo
         # apt-get update will bring a new candidate, which is the current installed version
         And I run `pro status` as non-root
-        Then I will see the following on stderr
+        Then stderr does not match regexp:
         """
+        .*\[info\].* A new version is available: 99.9.9
+        Please run:
+            sudo apt-get install ubuntu-advantage-tools
+        to get the latest version with new features and bug fixes.
         """
 
         Examples: ubuntu release
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -424,3 +424,98 @@ Feature: Command behaviour when unattached
           | kinetic | python3.10     |                         |
           # Lunar has a BIG error message explaining why this is a clear user error...
           | lunar   | python3.11     | --break-system-packages |
+
+
+    @series.all
+    @uses.config.machine_type.lxd-container
+    Scenario Outline: Warn users not to redirect/pipe human readable output
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run shell command `pro version | cat` as non-root
+        Then I will see the following on stderr
+        """
+        """
+        When I run shell command `pro version > version_out` as non-root
+        Then I will see the following on stderr
+        """
+        """
+        When I run shell command `pro status | cat` as non-root
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readable data from the `pro api` command,
+        or use `pro status --format json`.
+        """
+        When I run shell command `pro status | cat` with sudo
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readable data from the `pro api` command,
+        or use `pro status --format json`.
+        """
+        When I run shell command `pro status > status_out` as non-root
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readable data from the `pro api` command,
+        or use `pro status --format json`.
+        """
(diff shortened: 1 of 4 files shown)


renanrodrigo:
What a surprise... Looking at the test failures, it seems all `When I run` commands have their outputs redirected...
I'll check that.

features/unattached_commands.feature
```diff
@@ -424,3 +424,98 @@ Feature: Command behaviour when unattached
...
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readble data from the `pro api` command,
+        or use `pro status --format json`.
+        """
+        When I run shell command `pro status | cat` with sudo
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readble data from the `pro api` command,
```
s-makin, line 452:
```suggestion
        In scripts, prefer using machine readable data from the `pro api` command,
```

features/unattached_commands.feature
```diff
@@ -424,3 +424,98 @@ Feature: Command behaviour when unattached
...
+        or use `pro status --format json`.
+        """
+        When I run shell command `pro status > status_out` as non-root
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readble data from the `pro api` command,
```
s-makin, line 459:
```suggestion
        In scripts, prefer using machine readable data from the `pro api` command,
```

features/unattached_commands.feature
```diff
@@ -424,3 +424,98 @@ Feature: Command behaviour when unattached
...
+        or use `pro status --format json`.
+        """
+        When I run shell command `pro status > status_out` with sudo
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readble data from the `pro api` command,
```
s-makin, line 466:
```suggestion
        In scripts, prefer using machine readable data from the `pro api` command,
```

features/unattached_commands.feature
```diff
@@ -424,3 +424,98 @@ Feature: Command behaviour when unattached
...
+        or use `pro status --format json`.
+        """
+        When I run shell command `pro status --format tabular > status_out` as non-root
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readble data from the `pro api` command,
```
s-makin, line 480:
```suggestion
        In scripts, prefer using machine readable data from the `pro api` command,
```

features/unattached_commands.feature
```diff
@@ -424,3 +424,98 @@ Feature: Command behaviour when unattached
...
+        """
+        When I run shell command `pro status --format json > status_out` as non-root
+        Then I will see the following on stderr
+        """
+        """
+        When I run shell command `pro security-status | cat` as non-root
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readble data from the `pro api` command,
```
s-makin, line 495:
```suggestion
        In scripts, prefer using machine readable data from the `pro api` command,
```

features/unattached_commands.feature
```diff
@@ -424,3 +424,98 @@ Feature: Command behaviour when unattached
...
+        or use `pro security-status --format json`.
+        """
+        When I run shell command `pro security-status > status_out` as non-root
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readble data from the `pro api` command,
```
s-makin, line 502:
```suggestion
        In scripts, prefer using machine readable data from the `pro api` command,
```

uaclient/messages.py
```diff
@@ -1332,3 +1332,9 @@ class TxtColor:
     "pro-help-service-info",
     "Use pro help <service> to get more details about each service",
 )
+
+WARNING_HUMAN_READABLE_OUTPUT = """\
+WARNING: this output is intended to be human readable, and subject to change.
+In scripts, prefer using machine readble data from the `pro api` command,
```
s-makin, line 1338:
```suggestion
In scripts, prefer using machine readable data from the `pro api` command,
```

lucasmoura: s/readble/readable

renanrodrigo:
@s-makin same typo everywhere, copypasta drawbacks
fixed

uaclient/cli.py
```diff
@@ -1865,6 +1865,21 @@ def _warn_about_new_version(cmd_args=None) -> None:
         logging.warning(NEW_VERSION_NOTICE.format(version=new_version))
 
 
+def _warn_about_output_redirection(cmd_args) -> None:
```
lucasmoura, line 1868: I am wondering if we should turn this into a decorator, since right now we are running it for every cli command, but it is aimed only at two commands

uaclient/cli.py
```diff
@@ -1865,6 +1865,21 @@ def _warn_about_new_version(cmd_args=None) -> None:
...
+    """Warn users that the user readable output may change."""
+    if (
+        cmd_args.command in ("status", "security-status")
+        and not sys.stdout.isatty()
```
lucasmoura, line 1872: Sorry, what is the idea behind the extra check ?

uaclient/cli.py
renanrodrigo: well, the command is the first check - it is a no-op for any other commands. So making the decorator is more of an implementation detail imho; we can add/remove new checks or commands in any case.
If you see some advantage in turning it into a decorator, can be done.

uaclient/cli.py
renanrodrigo: We are only printing the warning to stderr if the stdout is not a tty - if the caller is piping it for instance. If just called in the terminal, the commands should not print any warning.

renanrodrigo: This mimics what apt does btw
`apt update` -> no warning
`apt update > output_text` -> warning
`apt update | cat` -> warning

uaclient/cli.py
lucasmoura: I think that it makes it explicit which commands do indeed need this warning. But I agree that it bottles down to an implementation detail. So feel free to choose what you prefer

features/unattached_commands.feature
```diff
@@ -424,3 +440,98 @@ Feature: Command behaviour when unattached
...
+        Then I will see the following on stderr
+        """
+        """
+        When I run shell command `pro status | cat` as non-root
+        Then I will see the following on stderr
+        """
+        WARNING: this output is intended to be human readable, and subject to change.
+        In scripts, prefer using machine readable data from the `pro api` command,
+        or use `pro status --format json`.
+        """
```
orndorffgrant, line 463: Is there any way we can write a test like this?
```
        When I run shell command `pro status as non-root
        Then I will not see the following on stderr
        """
        WARNING: this output is intended to be human readable, and subject to change.
        In scripts, prefer using machine readable data from the `pro api` command,
        or use `pro status --format json`.
        """
```

orndorffgrant (approved):
This looks great! Just an idea for one more test if it's feasible - which it might not be.

features/unattached_commands.feature
renanrodrigo: Here is the catch: we cannot :/
This thing will always be on stderr on tests because our step involves redirecting the output :D
Even this test that you highlighted wouldn't need the `| cat`  to work :D :D 
If you have any suggestion on how to improve those tests, please post - appreciate it :D :D :D

orndorffgrant: I was just wondering if there was a straightforward way to mimic a tty :upside_down_face:, it looks like it's possible but we'd need support in pycloudlib first so I created https://github.com/canonical/pycloudlib/issues/307

But this PR is good in the mean time :D

orndorffgrant:
CI failure unrelated

PR #2596, opened 2023-05-25
auto-attach: add new return code

lucasmoura:
## Why is this needed?
If we can attach the machine, but we fail to enable any service during the process, we will exit with a return code of 3

## Test Steps
Verify that the new unit test is working as expected

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -64,6 +64,7 @@ def attach_with_token(
         # Persist updated status in the event of partial attach
         ua_status.status(cfg=cfg)
         update_motd_messages(cfg)
+        # raise this exception in case we cannot enable all services
         raise exc
 
     current_iid = identity.get_instance_id()
--- a/uaclient/api/exceptions.py
+++ b/uaclient/api/exceptions.py
@@ -1,5 +1,3 @@
-from typing import List, Tuple
-
 from uaclient import messages
 from uaclient.api.errors import APIError
 from uaclient.exceptions import (
@@ -7,6 +5,7 @@ from uaclient.exceptions import (
     ConnectivityError,
     ContractAPIError,
     EntitlementNotFoundError,
+    EntitlementsNotEnabledError,
     InvalidProImage,
     LockHeldError,
     NonAutoAttachImageError,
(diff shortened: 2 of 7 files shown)


renanrodrigo (commented):
question: for this case, wouldn't it be enough to set `exit_code = 3` to the `EntitlementsNotEnabledError`?

orndorffgrant:
@cpaelzer also mentioned using powers of 2 for exit codes as a standard practice so that we can combine them in the future if needed

lucasmoura:

uaclient/cli.py
```diff
@@ -1452,6 +1452,9 @@ def action_auto_attach(args, *, cfg: config.UAConfig) -> int:
     except exceptions.UrlError:
         event.info(messages.ATTACH_FAILURE.msg)
         return 1
+    except exceptions.EntitlementsNotEnabledError as e:
```
renanrodrigo, line 1455: sorry I could have explained my thought better
I think this part of the code is not needed. Defining the `exit_code` for `UserFacingError` is enough for the main error handler to exit correctly ( like happens today with the `AlreadyAttachedError`, for instance)

lucasmoura: Fair enough, good point. I will update the code

lucasmoura: @renanrodrigo fixed

renanrodrigo (approved):
lgtm

orndorffgrant:
Bad news - this exception is only raised if you set the `enable` or `enable_beta` list args to the full_auto_attach endpoint. It won't get raised from the automatically enabled services if they fail while getting enabled.

So I think we need to modify the auto-enable code path so that it can raise the same exception.

orndorffgrant:
I think that would involve changing the exception raised here: https://github.com/canonical/ubuntu-pro-client/blob/main/uaclient/contract.py#L467-L470

renanrodrigo (approved):
lgtm again, thanks for all effort

orndorffgrant (approved):

Issue #2559, opened 2023-04-20
docs: support diagramming without image files


orndorffgrant:
We could use https://github.com/mgaitan/sphinxcontrib-mermaid or something similar to more easily add and edit diagrams in our documentation.

Not urgent but WDYT @s-makin?

s-makin:
If it makes creating and maintaining diagrams easier then it could definitely be worth adding yeah. I just want to do a bit of investigating on how it is for accessibility, but I'll report back here with my findings next week.

s-makin:
So, after doing some digging, it looks like [they've already been working](https://github.com/mermaid-js/mermaid/issues/2732) hard on the accessibility problem. Accessibility is always going to be tricky with content of this type, because it's hard to write good alt text or captions that really describes what the content is conveying (in a brief way). However, that problem exists for any diagram/flow chart. 

Although they're not fully there yet, the fact that they are aware of the issues and are actively working on various ways of improving their accessibility means that this is automatically better than the standard ways of creating these images. This proposal gets a thumbs up from me - thanks for bringing it to my attention!

s-makin:
It occurs to me that this needs an action - over the next cycle I'll add some documentation around this to our devdocs and link the PR here when it's up

PR #2606, opened 2023-05-31
docs: add mermaid plugin and replace image

orndorffgrant:
## Why is this needed?

Fixes: #2559

## Test Steps

Make sure the diagram is rendered correctly and is an appropriate replacement.
Make sure it works for each of `tox -e docs`, `tox -e docs-dev` and on readthedocs

## Checklist

 - [ ] Yes
 - [x] No


s-makin (changes requested):
Nice, thanks for doing this! I just have two suggestions. 

1) The diagram looks really nice when rendered on the light theme (which is the one I use by default). However, when you switch to dark mode it's really hard to see the arrows and where they point. I don't know if there's a way to specify "if light mode then this theme, if dark mode then that", but mermaid does allow different themes to be used https://mermaid.js.org/config/theming.html - probably worth doing, if it's possible.

2) It's probably a good idea for us to include at least a link to how to create these diagrams in the devdocs. It's not the easiest or most intuitive system to use, but the docs at least are really good. If we'd like people to use mermaid to reduce the maintenance burden of images in the future, let's make it real easy for them to find the info (to avoid explaining it ourselves). I'd suggest maybe this one about the basic syntax - it's easy to find more specific diagram info from here in their navigation https://mermaid.js.org/syntax/flowchart.html?id=flowcharts-basic-syntax

orndorffgrant:
@s-makin excellent points! I just updated this PR to include dark mode support (by using the furo theme css variables) and to add a short dev-doc with some helpful links to getting started with mermaid. LMK what you think!

renanrodrigo (approved):
LGTM

PR #2607, opened 2023-06-02
lintian: override invalid error where lintian doesn't evaluate second debhelper dep

orndorffgrant:
## Why is this needed?

This overrides an invalid lintian error. We'd like to fix this for version 28.

## Test Steps

Build and run lintian. Make sure there are no errors on jammy++

## Checklist

 - [x] Yes
 - [ ] No

### DIFF
--- a/debian/source/lintian-overrides
+++ b/debian/source/lintian-overrides
@@ -1,6 +1,8 @@
 # Lintian doesn't see dh-systemd alternative when building on xenial
 ubuntu-advantage-tools: missing-build-dependency-for-dh_-command dh_systemd_start => dh-systemd
-ubuntu-advantage-tools: missing-build-dependency-for-dh-addon systemd => dh-systemd
+
+# Lintian can't handle the multiline debhelper/dh-systemd dependencies explained in debian/control
+ubuntu-advantage-tools: missing-build-dependency-for-dh-addon *systemd*
 
 # Lintian doesn't like mentioning riscv64 for older go package
 ubuntu-advantage-tools: invalid-arch-string-in-source-relation riscv64 [build-depends: golang-1.10-go [!powerpc !riscv64]]


debian/source/lintian-overrides
```diff
@@ -2,5 +2,8 @@
 ubuntu-advantage-tools: missing-build-dependency-for-dh_-command dh_systemd_start => dh-systemd
 ubuntu-advantage-tools: missing-build-dependency-for-dh-addon systemd => dh-systemd
 
+# Lintian doesn't see the second debhelper dependency with >13.3 on jammy++
+ubuntu-advantage-tools: missing-build-dependency-for-dh-addon systemd *debhelper* (>= 9.20160709~)*
+
 # Lintian doesn't like mentioning riscv64 for older go package
```
renanrodrigo, line 8: Question: do we need this override? we do not go anymore do we?

orndorffgrant: I've created a follow up issue to look at all warnings and also remove that line here #2608

orndorffgrant: This PR is just a minimal diff to override the error for release 28

renanrodrigo: aha nice good good

debian/source/lintian-overrides
```diff
@@ -2,5 +2,8 @@
 ubuntu-advantage-tools: missing-build-dependency-for-dh_-command dh_systemd_start => dh-systemd
 ubuntu-advantage-tools: missing-build-dependency-for-dh-addon systemd => dh-systemd
 
+# Lintian doesn't see the second debhelper dependency with >13.3 on jammy++
+ubuntu-advantage-tools: missing-build-dependency-for-dh-addon systemd *debhelper* (>= 9.20160709~)*
```
lucaskanashiro, line 6: This is correct and overrides the lintian error. There is an alternative that would avoid the almost duplicate with line 3:

```
ubuntu-advantage-tools: missing-build-dependency-for-dh-addon  *systemd* 
```

With that, line three could be removed and this new override would cover both cases. I understand you want to minimize the changes here to release version 28, but I thought I should at least raise this here :)

Feel free to land this as you proposed and fix it later as part of #2608 .

lucaskanashiro (approved):

debian/source/lintian-overrides
orndorffgrant: This is closely related enough I think we can include it now! I'll test this locally and update the PR with it

orndorffgrant: @lucaskanashiro updated!

lucaskanashiro: Perfect, thanks @orndorffgrant !

PR #2612, opened 2023-06-06
pro enable anbox (SC-1507)

lucasmoura:
## Why is this needed?
We are now adding support to the `anbox-cloud` service. This service requires the configuration of apt sources while also the installation of 3 distinct snaps.

To support that, we have implemented a new directive called `additionalSnaps` which is currently being handled directly on the `base` module

This PR is marked as draft because we still need more input from the anbox team (i.e. defining the text for anbox-cloud on the output of  `pro status`)

## Test Steps
Run the new anbox integration test

## Checklist

 - [ ] Yes
 - [x] No


uaclient/entitlements/base.py
```diff
@@ -445,6 +454,13 @@ def enable(
...
+        # TODO: Move all logic from RepoEntitlement that
+        # handles the additionalPackages and APT directives
+        # to the base class. The additionalSnaps handling
+        # is a step on that direction
+        if not self.handle_additional_snaps():
+            return False, None
+
         ret = self._perform_enable(silent=silent)
         if not ret:
             return False, None
```
orndorffgrant: That we're installing snaps before enabling is interesting!

`additionalPackages` are installed after enabling, so it's slightly weird that they are handled differently.

But canonical-livepatch is similar, it is a snap installed before enabling, so this fits with the livepatch service.

I think this is good then - just an observation. And maybe: should we change the name from `additionalSnaps` to `requiredSnaps` or otherwise to make it clear that it is handled differently? (I think that may have been the original name :upside_down_face:)

uaclient/snap.py
```diff
@@ -1,18 +1,27 @@
...
 SNAP_CHANNEL_SHORTEN_VALUE = "…"
+SNAPD_SOCKET_PATH = "/run/snapd.socket"
+SNAPD_SNAPS_API = "/v2/snaps/{}"
 
 event = event_logger.get_event_logger()
 
 
+SnapInfo = namedtuple("SnapInfo", ("name", "revision", "channel"))
+
+
```
orndorffgrant, line 24: We already have `SnapPackage` defined below - we should probably use that, but add types (and convert to a `NamedTuple` if you prefer)

uaclient/snap.py
```diff
@@ -133,3 +142,94 @@ def get_installed_snaps() -> List[SnapPackage]:
...
+        capture=True,
+        retry_sleeps=SNAP_INSTALL_RETRIES,
+    )
+
+
+def get_snap_info(snap: str) -> SnapInfo:
+    snap_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    snap_sock.connect(SNAPD_SOCKET_PATH)
+
+    conn = http.client.HTTPConnection("localhost")
```
orndorffgrant, line 192: This usage of the actual API is great! I think should replace the usage of the `get_snap_package_info_tracking` function above

uaclient/entitlements/base.py
```diff
@@ -455,6 +471,59 @@ def enable(
...
+            .get("directives", {})
+            .get("additionalSnaps")
+        )
+
+        # If we don't have the directive, there is nothing
+        # to process here
+        if additional_snaps is None:
+            return True
+
+        if not system.which(snap.SNAP_CMD):
```
orndorffgrant, line 488: Why not `if not snap.is_installed()`?

uaclient/files/state_files.py
```diff
@@ -232,3 +232,45 @@ def __init__(
...
+
+anbox_cloud_credentials_file = DataObjectFile(
+    AnboxCloudData,
+    UAFile(
+        "anbox-cloud-credentials",
+        defaults.DEFAULT_PRIVATE_DATA_DIR,
+        private=True,
+    ),
+    DataObjectFileFormat.JSON,
+    optional_type_errors_become_null=True,
```
orndorffgrant, line 275: This only needs to be set to True if we expect other people or software to be modifying the file (which could result in a value of the wrong type getting written)

uaclient/entitlements/base.py
lucasmoura: Yes, as we have discussed, let's change this to use `requiredSnaps` instead

uaclient/snap.py
lucasmoura: Done

uaclient/entitlements/base.py
lucasmoura: Right now, we have two checks regarding snapd:

```python
if not system.which(snap.SNAP_CMD):
    pass
elif not snap.is_snapd_installed():
    pass
```

I think what this check is covering is the scenario where we have a `snap` command, but for some reason we don't have the `snapd` package installed. This is the logic we use today on Livepatch, and honestly, I don't know if we need it.

Maybe we can directly check if `snapd` is installed and if not, try to install it. What do you think ?

orndorffgrant: Yeah I'm not sure why we need to check both - maybe @blackboxsw might remember something?

uaclient/entitlements/anbox.py
```diff
@@ -0,0 +1,80 @@
...
+            return ret
+
+        directives = self.entitlement_cfg.get("entitlement", {}).get(
+            "directives", {}
+        )
+        anbox_images_directives = (
+            self.cfg.machine_token_file.entitlements.get("anbox-images", {})
+            .get("entitlement", {})
+            .get("directives", {})
+        )
```
orndorffgrant, line 57: I think this won't process overrides - is that a problem?

uaclient/entitlements/anbox.py
```diff
@@ -0,0 +1,80 @@
...
+        anbox_images_resource_token = ""
+        for resource_token in self.cfg.machine_token_file.resource_tokens:
+            if resource_token.get("type") == "anbox-images":
+                anbox_images_resource_token = resource_token.get("token", "")
+                break
```
orndorffgrant, line 62: This only works because anbox-images has `enableByDefault: true`, but I think we'd like to change that. Instead this can call `contract_client.get_resource_machine_access` to fetch the resource token from the server

uaclient/entitlements/anbox.py
```diff
@@ -0,0 +1,80 @@
...
+    AnboxCloudCredentials,
+    AnboxCloudData,
+    anbox_cloud_credentials_file,
+)
+from uaclient.types import MessagingOperationsDict, StaticAffordance
+
+event = event_logger.get_event_logger()
+
+
+class AnboxEntitlement(RepoEntitlement):
```
orndorffgrant, line 15: I think we need something to disable `enableByDefault` since it is currently `true` and we may need to leave it as true for a while until `pro enable anbox-cloud` is the only way being used to enable anbox-cloud

orndorffgrant: We may want an integration test for that too

uaclient/files/state_files.py
```diff
@@ -232,3 +232,44 @@ def __init__(
...
+        self.anbox_cloud_apt_url = anbox_cloud_apt_url
+
+
+class AnboxCloudData(DataObject):
+    fields = [
+        Field("anbox_attributes", AnboxCloudCredentials),
+    ]
+
+    def __init__(self, anbox_attributes: AnboxCloudCredentials):
+        self.anbox_attributes = anbox_attributes
```
orndorffgrant, line 264: Do we expect anything else to need to go at this layer? If not, we could flatten the file and have the attributes at the top level

uaclient/messages.py
```diff
@@ -785,9 +785,9 @@ class TxtColor:
     "livepatch-not-enabled", "canonical-livepatch snap is not installed."
 )
 
-LIVEPATCH_ERROR_INSTALL_ON_CONTAINER = NamedMessage(
+SERVICE_ERROR_INSTALL_ON_CONTAINER = FormattedNamedMessage(
     "livepatch-error-install-on-container",
-    "Cannot install Livepatch on a container.",
+    "Cannot install {title} on a container.",
 )
```
orndorffgrant, line 791: The name is still `"livepatch-error-install-on-container"` - should that change too?

uaclient/files/state_files.py
lucasmoura: Yes, right now there isn't. My concern is just that this might happen in the future, but maybe it is not that problematic

uaclient/entitlements/anbox.py
lucasmoura: That's fair because of the resourceTokens. But maybe there is another way to do that. I will double check

uaclient/entitlements/anbox.py
lucasmoura: I don't think so, because we are only fetching the "url" and "resourceToken" which shouldn't be modified by a selector

uaclient/files/state_files.py
renanrodrigo: +1

uaclient/entitlements/base.py
lucasmoura: @blackboxsw just double checking if you can verify this part of the code

lucasmoura:
@renanrodrigo @orndorffgrant I have updated the code to ignore `enableByDefault` for now

uaclient/entitlements/base.py
```diff
@@ -455,6 +473,58 @@ def enable(
 
         return True, None
 
+    def handle_required_snaps(self) -> bool:
+        """ "install snaps necessary to enable a service."""
+        additional_snaps = (
```
orndorffgrant, line 478: A really minor thing, but maybe we should rename the variable to match the directive name too
```suggestion
        required_snaps = (
```

orndorffgrant (approved):
I think the mechanism for turning off enableByDefault for now looks good +1

uaclient/entitlements/base.py
blackboxsw: I don't recall why we check both here, my original comment on an earlier iteration of the same PR that introduced this "def is_installed" was basically [do we need both](https://github.com/canonical/ubuntu-pro-client/pull/1678#discussion_r660014208)   .... that said, there was a related PR that changed our subp(["snap"...]) to subp(["/usr/bin/snap",...])  in https://github.com/canonical/ubuntu-pro-client/commit/698b3ca5299dc3973f3bdea4fd58cca884effeac. so I don't think we even need the which anymore

lucasmoura:
CI error is unrelated to this change

Issue #1539, opened 2021-04-09
d/postinst: rename keyrings/ubuntu-advantage-esm-infra-trusty.gpg to keyrings/ubuntu-advantage-esm-infra.gpg


blackboxsw:
minor postinst nit. Since keyrings/ubuntu-advantage-esm-infra.gpg is actually a keyring valid for all Ubuntu releases, let's rename that file in postinst dropping the "-trusty" suffix.

Ensure upgrade from previously attached Trusty Xenial systems migrates the old artifact if it exists to the new filename. Adjust esm key name definitions in the uaclient/entitlements/esm.py

orndorffgrant:
Tracking in SC-758

renanrodrigo:
actually renaming the key to `ubuntu-pro-esm-infra.gpg` to standardize them all

PR #2614, opened 2023-06-06
chore: rename repo GPG keys from -advantage to -pro  (SC-758)

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because we will have no more keys called 'trusty', and our keys are the same for all releases - `esm-infra` was the first and last one.
Besides that, advantage goes away and pro is there now.

oh and it
Fixes: #1539

## Test Steps
- attach on older version
- update
- see keys changed names
- see everything works

## Checklist

 - [x] Yes - touching debian/ scripts
 - [ ] No


debian/ubuntu-advantage-tools.postinst
```diff
@@ -16,7 +16,9 @@ fi
 
 APT_TRUSTED_KEY_DIR="/etc/apt/trusted.gpg.d"
 
-ESM_INFRA_KEY_TRUSTY="ubuntu-advantage-esm-infra-trusty.gpg"
+# Old key renamed in version 29
+OLD_ESM_INFRA_KEY_TRUSTY="ubuntu-advantage-esm-infra-trusty.gpg"
+ESM_INFRA_KEY="ubuntu-advantage-esm-infra.gpg"
```
lucasmoura, line 21: Should we rename it instead to "ubuntu-pro-esm-infra.gpg" ?

lucasmoura (commented):
@renanrodrigo I think we should have a manual test for this

debian/ubuntu-advantage-tools.postinst
renanrodrigo: didn't add a manual test at first because there is no launchpad bug, but it's trivial to write one - will do.

About renaming it to `ubuntu-pro` - it will break consistency; in this case, should we rename all `ubuntu-advantage` keys to `ubuntu-pro`?

orndorffgrant: I think we should keep consistency for now. Next cycle we should have a roadmap item for "Complete the rename of advantage to pro for all the technical bits" and we can look at renaming these at that point IMO

renanrodrigo:
test added

sru/release-28/test-esm-infra-key-rename.sh
```diff
@@ -0,0 +1,72 @@
...
+  lxc exec $name -- apt-get install ubuntu-advantage-tools -y > /dev/null
+else
+  lxc file push $install_from $name/new-ua.deb
+  lxc exec $name -- dpkg -i /new-ua.deb > /dev/null
+fi
+# ----------------------------------------------------------------
+echo -e "\n* u-a-t now has the change"
+echo "###########################################"
+lxc exec $name -- apt-cache policy ubuntu-advantage-tools
+echo -e "###########################################\n"
```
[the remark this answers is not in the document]
renanrodrigo: It would be even if we delete the key. What we can do is run `apt update` and make sure it does not fail (as the signatures would be there) - but that would be testing that we didn't corrupt the key while renaming. I mean, we could rename the key to anything really, and it would still work fine, as keys are identified by their IDs.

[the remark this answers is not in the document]
lucasmoura: True, I think we should indeed run the apt update command. And I agree that this should work as is, but I am just thinking that the test will be covered during the SRU and we should show them all of those details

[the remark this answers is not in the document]
renanrodrigo: call added

renanrodrigo:
@lucasmoura @orndorffgrant well well before merging this, I've been thinking
We adding logic to postinst to rename keys right;
Key names have a very low importance as they IDs are the important thing right; 
When renaming again (`advantage` to `pro`) we will have to modify postinst again

So I am starting to agree with Lucas that maybe that can be done right now, at once?

renanrodrigo:
renamed all keys to -pro

debian/ubuntu-advantage-tools.postinst
```diff
@@ -359,6 +364,37 @@ migrate_user_config_post() {
...
+    fi
+    # cc-eal
+    if [ -f "$APT_TRUSTED_KEY_DIR/ubuntu-advantage-cc-eal.gpg" ]; then
+        mv "$APT_TRUSTED_KEY_DIR/ubuntu-advantage-cc-eal.gpg" "$APT_TRUSTED_KEY_DIR/ubuntu-pro-cc-eal.gpg"
+    fi
+    # cis (and usg)
+    if [ -f "$APT_TRUSTED_KEY_DIR/ubuntu-advantage-cis.gpg" ]; then
+        mv "$APT_TRUSTED_KEY_DIR/ubuntu-advantage-cis.gpg" "$APT_TRUSTED_KEY_DIR/ubuntu-pro-cis.gpg"
+    fi
+    # fips (+upodates)
```
orndorffgrant, line 384:
```suggestion
    # fips (+updates)
```
:upside_down_face: 
and copypastad below

orndorffgrant (approved):

debian/ubuntu-advantage-tools.postinst
renanrodrigo: didn't you know about this? we decided it will be called `upodates` from next version.
I will keep you upodated.

Issue #2595, opened 2023-05-24
Docs: fips incompatibility matrix


s-makin:
Enabling `fips` (and being FIPS certified) means you can't activate certain other services or you'll no longer be certified. We should document which services are compatible/incompatible with `fips`, or if there are other services with similar compatibility issues, have a matrix showing which can be enabled at the same time.

@lucasmoura

PR #2615, opened 2023-06-07
docs: add incompatibility matrix for Pro services

lucasmoura:
## Why is this needed?

Define an incompatibility matrix for the services and reference it on each how-to-guide that shows how to enable a service that is incompatible with another service. We need that to better inform the user that some service cannot be enabled together

Fixes: #2595

## Test Steps
Double checks if the docs are rendering and if the table format is appropriate

## Checklist

 - [ ] Yes
 - [x] No


docs/howtoguides/enable_fips.md
```diff
@@ -24,6 +24,12 @@ this operation makes changes to underlying SSL-related libraries and requires a
 reboot into the FIPS-certified kernel.
 ```
 
+```{important}
+Once you enable FIPS, enabling some Pro services will not be possible.
```
lucasmoura, line 28: For both `fips` and `realtime-kernel`, I avoided mentioning that the services can be disabled, as we still leave the kernel around after disable which still conflicts with some services. But let me know what you think

docs/references/incompatibility_matrix.md
```diff
@@ -0,0 +1,11 @@
+# Incompatibility matrix for Ubuntu Pro services
+
+Some Ubuntu Pro services are incompatible with each other. The following
+matrix display the services that are incompatible:
+
+|                 | fips    | fips-updates   | livepatch   | realtime-kernel   |
```
lucasmoura, line 6: @s-makin I am not that happy with this table style, let me know if there another one that complies with the Ubuntu style we can use.

docs/howtoguides/enable_fips.md
s-makin: That makes sense. If they can't be cleanly disabled, it's probably safer to not bring the ability to disable them to users' attention or we might cause them problems. Since these services are aimed at enterprise users, if they want to disable it they can presumably contact support to be walked through that process?

docs/references/incompatibility_matrix.md
s-makin: I think it only looks a little odd because all of these services are incompatible with each other - this we could just as easily say as "you can't enable any of these services at the same time", which doesn't need a table. 

Looking at it now, I'm thinking it might make more sense if we change it to a "compatibility" (rather than incompatibility) matrix, and then also include the other services to show the full table of what you can and can't enable together? wdyt?

docs/howtoguides/enable_fips.md
lucasmoura: Yes, there is ways to disable those services, it is just that it is not that straightforward as you mentioned.

docs/references/incompatibility_matrix.md
lucasmoura: I think that makes sense, let's see how the table will look like

lucasmoura: @s-makin I have made the compatibility matrix with all the services we support today. Let me know what you think

s-makin (commented):
I think it looks a lot better now, after your changes. Just a couple of small tweaks would (I think) make it a lot better. Hopefully it's not too much extra faffing!

First one is; I think it would be better to have a dash or N/A in the boxes with e.g. cis vs. cis and fips vs. fips. 

Second one is - alphabetical seems to mean putting the least used service first, which is a bit weird. Should we order them by most popular/most used service, or group them according to their category (general/everyone, security hardening etc)?

lucasmoura:
@s-makin I guess we can change that order, but I think if it would be better if we group by category:

base-services:  esm-apps, esm-infra, livepatch
security hardening: cc-eal, usg, cis
security  certification: fips, fips-updates
kernel-related: realtime-kernel
robotic: ros, ros-updates

Do you think this category and order makes sense ?

s-makin:
> @s-makin I guess we can change that order, but I think if it would be better if we group by category:
> 
> **base-services**: esm-apps, esm-infra, livepatch **security hardening**: cc-eal, usg, cis **security certification**: fips, fips-updates **kernel-related**: realtime-kernel **robotic**: ros, ros-updates
> 
> Do you think this category and order makes sense ?

Yeah I think so :) putting the base services at the top would be more useful than having cc-eal first at least

docs/references/compatibility_matrix.md
```diff
@@ -0,0 +1,18 @@
+# compatibility matrix for Ubuntu Pro services
+
+Some Ubuntu Pro services are compatible with each other. The following
+matrix display the services that are compatible:
```
s-makin, line 4:
```suggestion
Many Ubuntu Pro services are compatible with each other, but some are not.
The following matrix displays the services that are compatible:
```

renanrodrigo (approved):

PR #2617, opened 2023-06-09
[docs] Add sync tabs to API reference

s-makin:
- Convert original file from .md to .rst
- Add sync tabs for Python vs. CLI content
- General style changes

## Why is this needed?
The API reference document showed both Python and
CLI code to the user at the same time. Now through
the use of tabs, the user can choose whether to view
Python or the CLI version. Tabs are synced so their
choice is applied to all entries. The API page is
now made cleaner and more functional.


CalvoM:
The separation and syncing of the tabs is a really cool and better addition. It is a better UI.

orndorffgrant (commented):
This is really great! The downside is that there is a lot of jank when switching tabs partway down the page because the height of the different sections are so different.

I've prototyped a solution to this - try it and let me know what you think:

Add a new file `docs/_static/js/synced_tab_links.js` with this content
```js
// Override the onLabelClick provided by sphinx_design
window.sdOnLabelClick = window.onLabelClick;
window.onLabelClick = function() {
    sdOnLabelClick.bind(this)();
    // turn off smooth scrolling to avoid jank during this transition
    const html = document.getElementsByTagName("html")[0];
    html.style["scroll-behavior"] = "auto";
    location.hash = "#" + this.getAttribute("for");
    html.style["scroll-behavior"] = "smooth";
}
```

Then set this in docs/conf.py to include it
```
html_js_files = [
    "js/github_issue_links.js",
    "js/synced_tab_links.js",
]
```

s-makin:
> This is really great! The downside is that there is a lot of jank when switching tabs partway down the page because the height of the different sections are so different.
> 
> I've prototyped a solution to this - try it and let me know what you think:
> 
> Add a new file `docs/_static/js/synced_tab_links.js` with this content
> 
> ```js
> // Override the onLabelClick provided by sphinx_design
> window.sdOnLabelClick = window.onLabelClick;
> window.onLabelClick = function() {
>     sdOnLabelClick.bind(this)();
>     // turn off smooth scrolling to avoid jank during this transition
>     const html = document.getElementsByTagName("html")[0];
>     html.style["scroll-behavior"] = "auto";
>     location.hash = "#" + this.getAttribute("for");
>     html.style["scroll-behavior"] = "smooth";
> }
> ```
> 
> Then set this in docs/conf.py to include it
> 
> ```
> html_js_files = [
>     "js/github_issue_links.js",
>     "js/synced_tab_links.js",
> ]
> ```

Thanks Grant - I've pushed that change, but I'm not sure if it's an improvement or not. When I tried it out it put the active tab to the top of the page (which is fine) but if I then switched to the other tab it switches to the section header then scrolls down again to put the tabs at the top of the page. Is there a way for us to keep the section header the tabs belong in at the top of the page to prevent the page from whizzing around?

orndorffgrant:
Okay @s-makin I have a new snippet that works better by keeping the clicked tab in the same location on the screen. And I've tested it in firefox and chrome. LMK if this works for you:
```js
// Override the onLabelClick provided by sphinx_design
window.sdOnLabelClick = window.onLabelClick;
window.onLabelClick = function() {
    // Get the position of the clicked element within the viewport before the document height changes
    const elementTopBefore = this.getBoundingClientRect().top;

    // Call the original, this changes all of the synced tabs and changes the height of the document
    sdOnLabelClick.bind(this)();

    // Get the new position of the clicked element
    const elementTopAfter = this.getBoundingClientRect().top;

    // Scroll the difference so that it is back in the same place where it was before the change
    window.scroll({
        top: document.documentElement.scrollTop + (elementTopAfter - elementTopBefore),
        left: document.documentElement.scrollLeft,
        behavior: "instant",
    });
```

orndorffgrant (approved):
Thank you! This is great

PR #2618, opened 2023-06-12
refactor: implementation-agnostic interface to readurl

orndorffgrant:
## Why is this needed?

This PR moves all http-related code into its own module and defines a simple interface to the `readurl` function.

The previous interface to `readurl` had different return types depending on the request, and it propogated exceptions that were raised from the urllib library. Those urllib exceptions were then caught and used in various places throughout the codebase.

The tls-in-tls implementation will require swapping out the urllib request function for a pycurl-based one, which will not have any of the same return types or exceptions as urllib, so we need to stop using urllib-specific things everywhere.

To achieve this, I've defined a simple HTTPResponse tuple that will always be returned by the `readurl` function - this should be easily implemented by any underlying http implementation. Notice that the function always returns this tuple, even if there is a http code above 300. Previously we expected/excepted errors on 4xx/5xx responses, so I've changed a lot of the calling code to check the return code rather than except errors. The only error that is raised is now our `exceptions.UrlError` and I've redefined it to only be raised if something about the connection fails. I think that overall this interface is simpler and will be easier to implement than raising errors for 4xx/5xx http responses. LMK what you think

TODO:
- [ ] add above explanation to commit message
- [x] let CI run to catch integration test errors
- [ ] run several cloud pro tests to make sure auto-attach still works

## Test Steps

Run a variety of cloud/pro tests to make sure all the metadata requests are still working

## Checklist

 - [ ] Yes
 - [x] No


uaclient/http/__init__.py
```diff
@@ -99,17 +108,42 @@ def configure_web_proxy(
...
+
+    json_body = None
+    if "application/json" in resp.headers.get("Content-type", ""):
+        json_body = json.loads(body, cls=util.DatetimeAwareJSONDecoder)
+
+    return HTTPResponse(
+        code=resp.code,
+        headers=resp.headers,
+        body=body,
+        json=json_body,
```
lucasmoura, line 131: I am wonder if we should create a mechanism here to enforce that we expect a valid json response. If we add such mechanism, we would not need of the `isinstance` checks throught the code

uaclient/daemon/retry_auto_attach.py
```diff
@@ -48,27 +48,13 @@ def full_auto_attach_exception_to_failure_reason(e: Exception) -> str:
...
-            else:
-                failure_reason = (
-                    messages.RETRY_ERROR_DETAIL_URL_ERROR_URL.format(url=e.url)
-                )
-        else:
-            failure_reason = messages.RETRY_ERROR_DETAIL_URL_ERROR_GENERIC
-        failure_reason += ': "{}"'.format(str(e))
-        return failure_reason
+        return messages.RETRY_ERROR_DETAIL_URL_ERROR_URL.format(
+            url=e.url
```
lucasmoura, line 56: Just double checking, are we sure we will always have a `url` here ?

lucasmoura (commented):
I think that's a great refactoring @orndorffgrant
I just have an idea regarding how we are handling the json content of the requests. Let me know if that makes sense

uaclient/daemon/retry_auto_attach.py
orndorffgrant: Yep - I've redefined UrlError to require a url

uaclient/exceptions.py
```diff
@@ -419,59 +412,28 @@ def __init__(
...
-            + "]"
-            + ", "
-            + self.api_error.get("message", "")
+class ExternalAPIError(UserFacingError):
+    def __init__(self, url: str, code: int, body: str):
+        self.url = url
+        self.code = code
+        self.body = body
+        super().__init__(
+            "Error connecting to {}: {} {}".format(
```
lucasmoura, line 421: Should this message sit on the `messages` module ?

uaclient/http/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+        method or "GET", url, sorted_header_str
+    )
+    if log_response_body:
+        # Due to implicit logging redaction, large responses might take longer
+        body_to_log = resp.body  # type: Any
+        if json_dict:
+            body_to_log = json_dict
+        elif json_list:
+            body_to_log = json_list
+        debug_msg += ", data: {}".format(body_to_log)
```
lucasmoura, line 190: I know this is a weird scenario and I am not sure if we want to cover it, but should we log both items if they exist ?

lucasmoura (approved):
That looks great, just a couple of nits

uaclient/exceptions.py
orndorffgrant: Yeah, I'll move it

uaclient/http/__init__.py
orndorffgrant: I'd rather log both of them in the "extra" fields of the log, but that would be a change to the log format and would require changes to how we redact things from these logs. So I thought that this PR should keep the exact same log format as before, which I think is what this does. What do you think?

orndorffgrant:
@lucasmoura updated!

uaclient/http/__init__.py
lucasmoura: That's fair. We can change that on a following PR if we think it would be useful in the future

lucasmoura (approved):

renanrodrigo (approved):

PR #2619, opened 2023-06-12
tests: do not delete user ssh keys on cleanup

lucasmoura:
## Why is this needed?

If we have not set the ssh keys when running integration tests, we might accidentally delete the ssh keys stored in the user .ssh folder. We are checking to see if the keys are stored in that folder before deleting it

## Test Steps
Run any integration test requesting the `wip` tag to be present. Certify that no test will run and the `.ssh` keys are not deleted.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -524,9 +524,11 @@ def after_all(context):
 
     if context.pro_config.destroy_instances:
         try:
-            key_pair = context.pro_config.default_cloud.api.key_pair
-            os.remove(key_pair.private_key_path)
-            os.remove(key_pair.public_key_path)
+            if context.pro_config.default_cloud._ssh_key_managed:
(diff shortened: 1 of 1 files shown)


orndorffgrant:
@lucasmoura I think this will work, but I think we can take advantage of `Cloud` instance state to catch this scenario if we add a new method to the `Cloud` class that looks something like this:
```python
def cleanup_ssh_keys(self):
    if not self._ssh_key_managed:
        return # this is the case where we would've deleted the users keys I think
    else:
        # proceed to delete them
```
I think this organization makes more sense to me if it works. But if you disagree then this PR is good as is :+1:

lucasmoura:
Yes, I think that works as well and uses on our abstraction to solve it. Let me update the code to use it

orndorffgrant (approved):
Thanks!

PR #2622, opened 2023-06-16
consolidating machineInfo and activityInfo + reporting docker more accurately (SC-1503)

orndorffgrant:
## Why is this needed?

This PR implements the following two specs:
- https://docs.google.com/document/d/1ut5IW1ZLRXmNw-pSHoO8QqmseavNxdbq5A-EkQftzB0/edit#
- https://docs.google.com/document/d/1CJXfQu-Oq3CYXcLg0yFnKlTK2c8T3USdYIw3RUh8oh8/edit#

## Test Steps
To test the docker detection changes, use the following Dockerfile
```dockerfile
from ubuntu:focal

run systemd-detect-virt || echo "systemd-detect-virt not available"

run apt update

copy ./ua.deb /ua.deb
run apt install /ua.deb -y

run python3 -c "from uaclient import system; print(system.get_virt_type() or 'failed')" | tee /result.txt
```

Run the following against a deb of pro-client from before and after these changes.
```
# docker old
docker build .

# docker new
env DOCKER_BUILDKIT=1 docker build . -t test
docker run -it test cat /result.txt

# podman
podman build .
```

debs built before these changes will display "failed" in the last step (or in result.txt). After these changes, it should accurately detect docker or podman, respectively.

## Checklist

 - [x] Yes - csteam
 - [ ] No


uaclient/testing/fakes.py
```diff
@@ -11,7 +11,7 @@ class FakeContractClient(UAContractClient):
         contract="cid", machine="mid"
     )
 
-    def __init___(self, cfg, responses=None):
```
renanrodrigo, line 14: lol

uaclient/contract.py
```diff
@@ -65,9 +70,14 @@ def add_contract_machine(self, contract_token, machine_id=None):
...
         @return: Dict of the JSON response containing the machine-token.
         """
+        if not machine_id:
+            machine_id = system.get_machine_id(self.cfg)
+
         headers = self.headers()
         headers.update({"Authorization": "Bearer {}".format(contract_token)})
-        data = self._get_platform_data(machine_id)
+        activity_info = self._get_activity_info()
+        activity_info["lastAttachment"] = attachment_dt.isoformat()
```
renanrodrigo, line 79: Won't this blow up when `attachment_dt` is `None`?

renanrodrigo (approved):
overall lgtm
Question: is this new docker detection step, and the information we send to the Contracts Server, documented somewhere we need to update?

features/airgapped.feature
```diff
@@ -47,6 +47,7 @@ Feature: Performing attach using ua-airgapped
         """
         500 .*:8000/ubuntu jammy-infra-security/main
         """
+        When I run `pro refresh` with sudo
```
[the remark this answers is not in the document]
orndorffgrant: Yep I'll change it to use that because that is the only purpose

uaclient/contract.py
orndorffgrant: yes, but I don't think attachment_dt is ever None (?) I can probably add types to make us more confident of that

orndorffgrant: Oh I totally put a default arg value of `None` - that probably shouldn't be there

uaclient/system.py
```diff
@@ -334,6 +334,20 @@ def is_container(run_path: str = "/run") -> bool:
     return False
 
 
+@lru_cache(maxsize=None)
+def is_desktop() -> bool:
+    """Returns True if any package installed has "ubuntu-desktop" in the name.
+
+    This includes ubuntu-desktop, ubuntu-desktop-minimal, kubuntu-desktop, etc.
+    """
+    from uaclient import apt
```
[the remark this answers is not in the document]
renanrodrigo: not at all. the consumption of the apt functionality is good, it's the the circular imports that break my heart, not your code (:

uaclient/actions.py
```diff
@@ -51,26 +56,49 @@ def attach_with_token(
     """
     from uaclient.timer.update_messaging import update_motd_messages
 
+    contract_client = contract.UAContractClient(cfg)
+    attached_at = datetime.datetime.now(tz=datetime.timezone.utc)
+
     try:
-        contract.request_updated_contract(
```
lucasmoura, line 55: Just double checking, can't we update `request_updated_contract` with this extra logic ?

uaclient/contract.py
```diff
@@ -368,80 +369,87 @@ def _update_contract_machine(
...
-        )
-        self.cfg.write_cache("machine-id", machine_id)
-
-    def _get_platform_data(self, machine_id):
-        """Return a dict of platform-related data for contract requests"""
-        if not machine_id:
-            machine_id = system.get_machine_id(self.cfg)
-        return {
-            "machineId": machine_id,
+    def _get_activity_info(self):
```
lucasmoura, line 396: Why can't this function already returns the compatible structure we need to send to the Contract server ?

uaclient/system.py
```diff
@@ -188,7 +188,19 @@ def get_virt_type() -> str:
...
+        # container that doesn't have it installed. So we look for hints
+        # of that situation to report it accurately.
+        try:
+            proc_1_cgroup = load_file("/proc/1/cgroup")
+            if "docker" in proc_1_cgroup or "buildkit" in proc_1_cgroup:
+                return "docker"
+            elif "buildah" in proc_1_cgroup:
+                return "podman"
+            else:
+                return ""
```
lucasmoura, line 201: Those these names appear in specific patterns in the string (i.e. beginning or end of the line) ? If yes, should we have a more restrictive regex for them or do you think name clashing is really unlikely here ?

orndorffgrant: here is what the file looks like in each situation, using the following dockerfile:
```
from ubuntu:focal
run cat /proc/1/cgroup | tee result.txt
```

docker build .
```
13:perf_event:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
12:blkio:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
11:rdma:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
10:pids:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
9:cpuset:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
8:net_cls,net_prio:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
7:devices:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
6:freezer:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
5:cpu,cpuacct:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
4:hugetlb:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
3:memory:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
2:misc:/
1:name=systemd:/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
0::/docker/43820adb8df766d314faece4ddf62332a0eee1fc1c5154ffd4b883fb88cb1232
```

env DOCKER_BUILDKIT=1 docker build . -t test
docker run -it test cat /result.txt
```
13:perf_event:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
12:blkio:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
11:rdma:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
10:pids:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
9:cpuset:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
8:net_cls,net_prio:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
7:devices:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
6:freezer:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
5:cpu,cpuacct:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
4:hugetlb:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
3:memory:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
2:misc:/
1:name=systemd:/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
0::/docker/buildkit/mc0szzrnjv8hx2hqwzvmzriq5
```

podman build .
```
13:perf_event:/
12:blkio:/user.slice
11:rdma:/
10:pids:/user.slice/user-1000.slice/user@1000.service
9:cpuset:/
8:net_cls,net_prio:/
7:devices:/user.slice
6:freezer:/
5:cpu,cpuacct:/user.slice
4:hugetlb:/
3:memory:/user.slice/user-1000.slice/user@1000.service
2:misc:/
1:name=systemd:/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.kde.konsole-1b54bbb6634a43a78eb7033cd3b3ac17.scope/buildah-buildah2868387759
0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.kde.konsole-1b54bbb6634a43a78eb7033cd3b3ac17.scope
```

The least obvious one is the podman one where "buildah" only shows up in the "name" field.

So, we could try to make things more strict, but my thought was that it is really unlikely for a clash here. It would only happen in an environment where `systemd-detect-virt` wasn't installed (so we fall back to this function), then the user would have to be doing something tricky/non-standard for one of the magic words to show up there.

And since we're only doing this to get a sense of usage, not for strict tracking, I think its okay.

uaclient/actions.py
orndorffgrant: Yes, and that could solve the specific need of this PR, but the refactor stemmed from two observations:
1. The exception cases below with comments claiming to run after a partial attach are wrong - they run after failed attaches as well. That is due to the attachment and enablement being combined in the one request_updated_contract function and that function doesn't separate the exceptions raised in attachment vs during enablement (my first idea of refactoring was do exactly that).
2. The request_updated_contract function was kind-of two-functions-in-one. Most of it's uses were as "refresh", and one of its uses was as "attach". Since there was only one usage of the "attach" variant of request_updated_contract, I thought i made sense to bring the attach logic up into our "attach_with_token" function. That leaves only the refresh code behind, which is what every other caller was using.

The other thing is that I think `attach_with_token` should be the place where "attach"y things happen, and it mostly is, so it would be confusing to put some actions related to "attach" (saving attachment date, and sending machine activity ping) down in request_updated_contract.

uaclient/contract.py
```diff
@@ -368,80 +369,87 @@ def _update_contract_machine(
...
+            "clientVersion": version.get_version(),
         }
 
-    def _get_platform_basic_info(self):
-        """Return a dict of platform basic info for some contract requests"""
+        if _is_attached(self.cfg).is_attached:
+            enabled_services = _enabled_services(self.cfg).enabled_services
+            attachment_data = attachment_data_file.read()
+            activity_info = {
+                "activityID": self.cfg.machine_token_file.activity_id,
```
orndorffgrant, line 412: I think this needs a 
```suggestion
                "activityID": self.cfg.machine_token_file.activity_id or system.get_machine_id(cfg)
```

uaclient/contract.py
orndorffgrant: It returns exactly this structure: https://github.com/canonical/ua-contracts/blob/develop/docs/contracts.yaml#L275

That is used directly as the body for `update_activity_token` and used directly as the value of "activityInfo" for the body of `update_contract_machine`.

Everything else is a special case that requires the same values but only a subset of them, or in add_contract_machine needs `lastAttachment` even though we're not attached yet.

orndorffgrant:
@lucasmoura @renanrodrigo I think this is good to go now.
@renanrodrigo I don't think we have any documentation on what info is sent in attach/refresh/activity API calls, so I created #2655 to add something.

lucasmoura:
@orndorffgrant I think we just need to fix a test for xenial and we should be good here

renanrodrigo (approved):
LGTM
I think this can be merged with the required csteam review

lucasmoura (approved):

TeodorPt:
@orndorffgrant we are currently at a sprint, it would be great if this could wait until next week s.t. I can pay proper attention to it

orndorffgrant:
@TeodorPt Yep it can wait til next week - all good!

uaclient/contract.py
```diff
@@ -212,7 +208,7 @@ def update_activity_token(self):
         machine_token = self.cfg.machine_token.get("machineToken")
         machine_id = system.get_machine_id(self.cfg)
 
-        request_data = self._get_activity_info(machine_id)
+        request_data = self._get_activity_info()
```
TeodorPt, line 211: Not necessarily for this PR, but for the overall approach:

I think in this method, lower below, you are saving the response from the UPDATE_ACTIVITY_TOKEN request in a file. As far as I can tell, in the current implementation the endpoint will return the entire payload (with the new activity ID, token and all the new information you send as is). Is this fine? If not, should we change the behavior s.t. we only return the activity identifiers and not the metadata?

TeodorPt (commented):
Sorry for the delays. The new payload per discussions in US039 LGTM. I have some questions but they shouldn't influence the solution here.

uaclient/contract.py
```diff
@@ -371,80 +372,88 @@ def _update_contract_machine(
...
+    servers deployed that we need to support.
+    This function is used for attach and refresh calls.
+    """
+    activity_info = request_body.get("activityInfo", {})
+
+    return {
+        "machineId": request_body.get("machineId"),
+        "activityInfo": activity_info,
+        "architecture": activity_info.get("architecture"),
+        "os": {
```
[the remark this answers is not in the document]
pandrey2003: No, seems like the airgapped neither checks nor uses this field. However, thanks for re-checking!

As a matter of fact, all the provided properties here are correct 👍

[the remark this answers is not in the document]
orndorffgrant: Thanks for checking! I referenced the airgapped implementation when writing this function which is why I didn't include `version`.

uaclient/cli.py
```diff
@@ -1558,6 +1564,10 @@ def action_attach(args, *, cfg):
                 event.info(msg.msg, file_type=sys.stderr)
                 event.error(error_msg=msg.msg, error_code=msg.name)
                 ret = 1
+
+        contract_client = contract.UAContractClient(cfg)
```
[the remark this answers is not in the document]
orndorffgrant: We added this because the attach request itself can't include any services that got enabled. At spec time we decided we wanted data of what services got enabled immediately after attach (this could include `enableByDefault: true` services, or services that the user configured the pro-client to immediately enable). To get that data we decided to immediately send an activity ping after the post-attach-service-enables to get that info recorded.

uaclient/contract.py
orndorffgrant: That's an excellent point! And something nobody else noticed.

I think this behavior is fine since the extra data is minimal and non-sensitive and its presence doesn't hurt the client - we treat that state file as a json blob, so extra fields are ignored.

In the future we hope to improve/organize our on-disk state and at that time we may think about discarding the extra metadata on the client side, but even then I don't think it hurts for the server to send it back.

PR #2625, opened 2023-06-21
Rename more ubuntu-advantage occurrences to ubuntu-pro

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because we are not advantage anymore, we are pro now.
Plus copyright is updated.


lucasmoura (approved):

lucasmoura:
CI errors seem unrelated to this change

PR #2628, opened 2023-06-26
Fix simulate status

lucasmoura:
## Why is this needed?
On the --simulate-with-token status flow, we are not addressing the proper status cache format when checking for inapplicable services. Because of that, some services might appear as available. This change address the situation where we are evaluating service availability from the status cache and the contract response

## Test Steps
Run status as root and run the simulate command in a bionic/focal/jammy container and verify that the services listed there are correct

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/status.py
+++ b/uaclient/status.py
@@ -518,7 +518,7 @@ def simulate_status(
     inapplicable_resources = [
         resource["name"]
         for resource in sorted(resources, key=lambda x: x["name"])
-        if not resource["available"]
+        if not resource["available"] or resource["available"] == "no"
     ]
 
     for resource in resources:


uaclient/status.py
```diff
@@ -518,7 +518,7 @@ def simulate_status(
     inapplicable_resources = [
         resource["name"]
         for resource in sorted(resources, key=lambda x: x["name"])
-        if not resource["available"]
+        if not resource["available"] or resource["available"] == "no"
```
renanrodrigo, line 521: same change needs to be done for attached status maybe, on line `235` in this same file

renanrodrigo (commented):
As discussed, besides making this change, we don't need the status cache in simulated status at all. So I think we can modify the above lines to always get the resources from the contracts server

orndorffgrant:
I included this in #2630

PR #2631, opened 2023-06-28
attach: minimize apt update calls (sc-186)

aciba90:
## Why is this needed?

Minimize the number of apt-update calls to 1 during auto-attach on cloud instances.

## Additional context
[SC-186](https://warthogs.atlassian.net/browse/SC-186)
[SF: #00313994](https://bootstack.canonical.com/cases/00313994)

## Checklist

 - [ ] Yes
 - [x] No

[SC-186]: https://warthogs.atlassian.net/browse/SC-186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -4,10 +4,12 @@ import os
 from typing import List, Optional  # noqa: F401
 
 from uaclient import (
+    apt,
     clouds,
     config,
     contract,
     entitlements,
+    event_logger,
     exceptions,
     livepatch,
 )
@@ -25,6 +27,7 @@ from uaclient.defaults import (
 from uaclient.files.state_files import timer_jobs_state_file
 
 LOG = logging.getLogger("pro.actions")
+event = event_logger.get_event_logger()
 
 
 UA_SERVICES = (
@@ -123,7 +126,14 @@ def enable_entitlement_by_name(
         called_name=name,
         access_only=access_only,
     )
-    return entitlement.enable()
+    ent_ret, reason, apt_update = entitlement.enable()
+    if not ent_ret:
+        return ent_ret, reason
+    if apt_update:
+        event.info(messages.APT_UPDATING_LISTS)
+        apt.run_apt_update_command()
+    entitlement.post_enable()
+    return ent_ret, reason
 
 
 def status(
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -1,7 +1,7 @@
 import glob
 import logging
 import os
-from typing import List, Optional  # noqa: F401
+from typing import List, NamedTuple, Optional, Tuple, Union  # noqa: F401
 
 from uaclient import (
     apt,
@@ -24,6 +24,8 @@ from uaclient.defaults import (
     DEFAULT_CONFIG_FILE,
     DEFAULT_LOG_PREFIX,
 )
+from uaclient.entitlements.base import UAEntitlement
+from uaclient.entitlements.entitlement_status import CanEnableFailure
 from uaclient.files.state_files import timer_jobs_state_file
 
 LOG = logging.getLogger("pro.actions")
@@ -101,9 +103,23 @@ def auto_attach(
     attach_with_token(cfg, token=token, allow_enable=allow_enable)
 
 
-def enable_entitlement_by_name(
+_EnableInfo = Tuple[bool, Union[None, CanEnableFailure], Optional[Exception]]
+
+_ServiceInfo = NamedTuple(
+    "_ServiceInfo",
+    [
+        ("name", str),
+        ("service", UAEntitlement),
+        ("enable", _EnableInfo),
+        ("apt_update", bool),
+        ("post_enable", Optional[bool]),
+    ],
+)
+
+
+def enable_entitlements_by_name(
     cfg: config.UAConfig,
-    name: str,
+    names: List[str],
     *,
     assume_yes: bool = False,
     allow_beta: bool = False,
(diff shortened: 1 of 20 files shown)


aciba90:
[features/apt_messages.feature:218](https://github.com/canonical/ubuntu-pro-client/blob/7ce4928f210e90fb9b2b742f4ccd6603be61acd5/features/apt_messages.feature#L218C16-L218C16) fails in the step [apt upgrade --dry-run](https://github.com/canonical/ubuntu-pro-client/blob/7ce4928f210e90fb9b2b742f4ccd6603be61acd5/features/apt_messages.feature#L192C1-L201) because of the following extra output:

```
Get more security updates through Ubuntu Pro with 'esm-apps' enabled:
  hello
Learn more about Ubuntu Pro at https://ubuntu.com/pro
The following packages will be upgraded:
  hello
```

And this does happen because the last performed _apt update_ was before post-enabling esm-* services, so the apt-hook did not fully realize that esm-* are enabled.

I guess this UX change is something we want to avoid. My question is, do you have any insight on how to refresh the apt-hooks without performing a second _apt update_?

uaclient/actions.py
```diff
@@ -123,7 +126,14 @@ def enable_entitlement_by_name(
         called_name=name,
         access_only=access_only,
     )
-    return entitlement.enable()
+    ent_ret, reason, apt_update = entitlement.enable()
```
lucasmoura, line 129: Maybe we don't want to solve that issue, but if we run the CLI enable command with multiple services, we will still perform more than one `apt update` on this setup

uaclient/contract.py
```diff
@@ -504,11 +517,69 @@ def process_entitlements_delta(
                     "Unexpected error processing contract delta for {name}:"
                     " {delta}".format(name=name, delta=new_entitlement)
                 )
+        else:
+            if service_enabled:
```
lucasmoura, line 521: This logic can be problematic if the service transitions to unentitled as can be seen [here](https://github.com/canonical/ubuntu-pro-client/blob/main/uaclient/entitlements/base.py#L1054)

lucasmoura (commented):
The only problem I have with this solution is that:

1) An entitlement cannot be enabled on its own anymore. For example, calling `entitlement.enable` is no longer sufficient.
2) The logic that actually enables an entitlement is now on the `process_contract_deltas` function that sits on the `contract` module, which is not intuitive

I was wondering if we could maybe create a class to orchestrate enabling any service. Then both `actions` and `contract` module could reuse it. However, this is just a shallow idea at the moment, I will try to better think about it

orndorffgrant:
Another idea to consider, though I'm not sure it would solve everything: `process_entitlements_delta` could be modified to return a list of entitlements to enable, and then enabling would happen separately.
Maybe `enable_entitlement_by_name` could start to accept a list, and that would be the one-true function that knows how to enable multiple entitlements at the same time.

aciba90:
I have extended `enable_entitlement_by_name` to accept a list of services to enable and use this function for attach, enable and auto-attach.

`process_entitlements_delta` updates deltas as it did, for already enabled services, via `process_entitlements_delta` and `UAEntitlement.process_contract_deltas`.

The difference is that `UAEntitlement.process_contract_deltas` return a list of services to enable if the service has to be enabled, and then `process_entitlements_delta` uses `enable_entitlements_by_name` to enable them.

Could you, @lucasmoura and @orndorffgrant, please take a look? If you agree, I will continue with this approach.

aciba90:
After a live PR review with @orndorffgrant, we agreed to only return a boolean indicating if the services as to be enabled in `UAEntitlement.process_contract_deltas`, because the _required services_ use case is out of the scope of this issue.

And a +1 on the general idea of the last commit.

uaclient/entitlements/base.py
```diff
@@ -536,16 +534,41 @@ def handle_required_snaps(self) -> bool:
...
+        """
+        ret = self._perform_post_enable(silent=silent)
+        if not ret:
+            return False
+        msg_ops = self.messaging.get("post_enable", [])
+        if not util.handle_message_operations(msg_ops):
+            return False
+        return True
+
+    def _perform_post_enable(self, silent: bool = False) -> bool:
```
lucasmoura, line 562: Do we still need that ?

uaclient/actions.py
```diff
@@ -113,17 +132,57 @@ def enable_entitlement_by_name(
...
+            return ent_ret, reason
+        do_apt_update |= apt_update
+
+    apt_exception = None
+    if do_apt_update:
+        event.info(messages.APT_UPDATING_LISTS)
+        try:
+            apt.run_apt_update_command()
+        except Exception as e:
+            apt_exception = e
```
lucasmoura, line 168: Shouldn't we disable the services if any exception happen here ?

uaclient/cli.py
```diff
@@ -1297,16 +1297,17 @@ def action_enable(args, *, cfg, **kwargs):
...
         try:
-            ent_ret, reason = actions.enable_entitlement_by_name(
-                cfg,
-                ent_name,
-                assume_yes=args.assume_yes,
-                allow_beta=args.beta,
-                access_only=access_only,
-                variant=variant,
-            )
+            if e is not None:
```
lucasmoura, line 1309: Is there a reason why we shouldn't raise the exception directly on the `enable_entitlements_by_name` function ?

lucasmoura (commented):
I think this solution is better than the other one. However, another thing we can consider is to not perform the full `apt update` every time, but instead just update the a specific APT source. I think this is achievable using the `python3-apt` library. But we would still need to figure out if doing that make it faster or not

aciba90:
> I think this solution is better than the other one. However, another thing we can consider is to not perform the full `apt update` every time, but instead just update the a specific APT source. I think this is achievable using the `python3-apt` library. But we would still need to figure out if doing that make it faster or not

Thanks, @lucasmoura, for the review and suggestions.

https://github.com/canonical/ubuntu-pro-client/pull/2732 includes a POC using python3-apt and only updating required source lists. About performance, I did a comparison between tip of main and attaching and enabling esm-infra and esm-apps and there was an improvement of 10s on average, I will attach those results to the other PR.

aciba90:
Closing in favor of https://github.com/canonical/ubuntu-pro-client/pull/2732.

PR #2635, opened 2023-06-30
docs: enabled services and machine readable status updates

orndorffgrant:
aside: I wish there were configurable PR templates per base branch :(

This adds references to the new status related API functions to the machine readable status section.
Also adds list of possible enabled services.


docs/explanations/status_columns.md
```diff
@@ -104,6 +104,21 @@ output for informational and debugging purposes.
...
 ## Machine-readable output
 
+### Status API functions
+
+Some status information can be obtained from API functions. These include:
+
+- Is the machine currently attached to an Ubuntu Pro subscription? See [u.pro.status.is_attached.v1](references/api:u.pro.status.is_attached.v1)
+- Which Ubuntu Pro services are currently enabled on the machine? See [u.pro.status.enabled_services.v1](references/api:u.pro.status.enabled_services.v1)
+
+If you need status related information for which there is not an API function, then `pro status --format=json` (described below) is the recommended machine-readable interface.
```
s-makin, line 114:
```suggestion
If you need status-related information for which there is no API function, then `pro status --format=json` (described below) is the recommended machine-readable interface.
```

s-makin (approved):
Just one tiny nit, LGTM :)

Hooloovoo:
I think what you have written is good. I would also add a note to each of the relevant bullet points in the JSON section, e.g. on the attached bullet say something like "For pro client versions more recent than version 28, we recommend using the [u.pro.status.is_attached.v1](https://chat.canonical.com/canonical/messages/references/api:u.pro.status.is_attached.v1) endpoint."

So that anyone going back to that list has their attention drawn to the new endpoints and it is also clear which pieces of information do and do not have equivalents in the API.

Hooloovoo (commented):
I think what you have written is good. I would also add a note to each of the relevant bullet points in the JSON section, e.g. on the attached bullet say something like "For pro client versions more recent than version 28, we recommend using the [u.pro.status.is_attached.v1](https://chat.canonical.com/canonical/messages/references/api:u.pro.status.is_attached.v1) endpoint."

So that anyone going back to that list has their attention drawn to the new endpoints and it is also clear which pieces of information do and do not have equivalents in the API

docs/references/api.rst
```diff
@@ -1087,7 +1087,8 @@ This endpoint shows the Pro services that are enabled on the machine.
                - Description
              * - ``name``
                - *str*
-               - Name of the service
+               - | Name of the service.
+                 | Possible values are: ``cc-eal``, ``cis``, ``esm-apps``, ``esm-infra``, ``fips``, ``fips-updates``, ``livepatch``, ``realtime-kernel``, ``ros``, ``ros-updates``.
```
renanrodrigo, line 1091: I am missing `usg` here. Should both `usg` and `cis` make it to the list? Maybe some `*` with an explanation or reference to the explanation?

renanrodrigo (approved):
lgtm, small change needed

docs/references/api.rst
s-makin: Good catch. Since there's one release that has both, I'd say both (but like ``usg``/``cis``). It is probably worth including a short note to say which service is available on which version, but an explanation of the two is not for here imo.

orndorffgrant: `usg` won't appear here. I'll add something short to say that "cis" is listed when "usg" is enabled.

orndorffgrant:
I think I've addressed all comments

docs/explanations/status_columns.md
```diff
@@ -188,9 +203,9 @@ For example, running `sudo pro status --format=json` on an attached machine may
 ```
 
 Some particularly important attributes in the output include:
-* `attached`: This boolean value indicates whether this machine is attached to an Ubuntu Pro account. This does not tell you if any particular service (e.g. `esm-infra`) is enabled. You must check the individual service item in the `services` list for that status (described below).
+* `attached`: This boolean value indicates whether this machine is attached to an Ubuntu Pro account (Starting with version 28.1, we recommend using the [u.pro.status.is_attached.v1](references/api:u.pro.status.is_attached.v1) API instead). This does not tell you if any particular service (e.g. `esm-infra`) is enabled. You must check the individual service item in the `services` list for that status (described below).
```
s-makin, line 206: Since it's not really an aside, I think this is better off as a new sentence rather than being in brackets (same with the one just below at line 208).

s-makin (approved):
Just one small suggestion, but would also be fine to accept as-is

docs/explanations/status_columns.md
renanrodrigo: Addressed

renanrodrigo:
Addressed @s-makin 's comment as Grant is off, I'll merge and we can iterate later if neeeded

PR #2638, opened 2023-07-05
api: u.apt_news.current_news.v1

orndorffgrant:
## Why is this needed?

Desktop update-manager needs to integrate apt-news, so we need to provide an api function for them to get the current apt news. This way they won't need to re-implement any of the aptnews.json parsing logic.

## Test Steps

Run the new version of the apt-news test

## Checklist

 - [x] Changes here need to be documented, and this was done in: #2642

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/api/u/apt_news/current_news/v1.py
+++ b/uaclient/api/u/apt_news/current_news/v1.py
@@ -1,0 +1,32 @@
+from typing import Optional
+
+from uaclient.api.api import APIEndpoint
+from uaclient.api.data_types import AdditionalInfo
+from uaclient.config import UAConfig
+from uaclient.data_types import DataObject, Field, StringDataValue
+from uaclient.files.state_files import apt_news_raw_file
+
+
+class CurrentNewsResult(DataObject, AdditionalInfo):
+    fields = [
+        Field("current_news", StringDataValue, required=False),
+    ]
+
+    def __init__(self, *, current_news: Optional[str]):
+        self.current_news = current_news
+
+
+def current_news() -> CurrentNewsResult:
+    return _current_news(UAConfig())
+
+
+def _current_news(cfg: UAConfig) -> CurrentNewsResult:
+    return CurrentNewsResult(current_news=apt_news_raw_file.read())
+
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="CurrentNews",
+    fn=_current_news,
+    options_cls=None,
+)
--- a/uaclient/apt_news.py
+++ b/uaclient/apt_news.py
@@ -67,14 +67,6 @@ class AptNewsMessage(DataObject):
         self.lines = lines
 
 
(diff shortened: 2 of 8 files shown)


uaclient/apt_news.py
```diff
@@ -210,24 +193,38 @@ def local_apt_news(cfg: UAConfig) -> bool:
...
 
 def update_apt_news(cfg: UAConfig):
-    local_news_written = local_apt_news(cfg)
-    if not local_news_written:
-        apt_pkg.init()
-        fetch_and_process_apt_news(cfg)
+    try:
+        news = local_apt_news(cfg)
+        if not news:
+            apt_pkg.init()
```
lucasmoura, line 218: Maybe this is an overkill, but should we wrap this with `PreserveAptCfg` ? What do you think @renanrodrigo ?

uaclient/api/u/apt_news/current_news/v1.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self, *, current_news: Optional[str]):
+        self.current_news = current_news
+
+
+def current_news() -> CurrentNewsResult:
+    return _current_news(UAConfig())
+
+
+def _current_news(cfg: UAConfig) -> CurrentNewsResult:
+    return CurrentNewsResult(current_news=apt_news_raw_file.read())
```
lucasmoura, line 24: Shouldn't the api try to fetch/update apt new when it is called ? My concern is that could be some conditions here we would output outdated messages to the user

uaclient/apt_news.py
orndorffgrant: Ohhh, now that this could be called from an existing long-running process that might also use apt (e.g. update-manager, yeah we need to be more careful about this now)

uaclient/api/u/apt_news/current_news/v1.py
orndorffgrant: The message gets updated asynchronously every `apt update` at most once per day - and I don't think we want to change that cadence. The purpose of this API is for displaying the news in update-manager, similar to how it is displayed in `apt upgrade`, so I think it should always display the same message as `apt upgrade`.

I did consider adding a separate endpoint like "force_update_apt_news" to force a fetch/update outside of the normal `apt update` but since no-one is asking for it I figured it wasn't worth the effort (yet - until someone does ask for it :upside_down_face: )

uaclient/apt_news.py
orndorffgrant: oh wait, this section of code still only gets called by `apt-news.service` so we don't need to worry about this

lucasmoura (approved):

uaclient/apt_news.py
renanrodrigo: yep +1 no need to worry

PR #2640, opened 2023-07-06
tests: require contract token in the correct tests

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because now we have the tests which require a token to run well defined.
If the whole feature requires attachment, require it on feature level. If the feature does not imply attachment, require it on scenario level.

## Test Steps
Run the CI suite without a token, see that no test fails because you don't have the token, and that those which require the token were skipped.

## Checklist
 - [x] I have updated or added any integration tests accordingly - aha

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -945,7 +945,6 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
 
     @series.xenial
     @series.bionic
-    @uses.config.contract_token
     @uses.config.machine_type.lxd-container
     Scenario Outline: Attached enable ros on a machine
         Given a `<release>` machine with ubuntu-advantage-tools installed
@@ -1160,7 +1159,6 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         """
 
     @series.xenial
-    @uses.config.contract_token
     @uses.config.machine_type.lxd-container
     Scenario Outline: APT auth file is edited correctly on enable
         Given a `<release>` machine with ubuntu-advantage-tools installed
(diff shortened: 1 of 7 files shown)


features/attached_enable.feature
```diff
@@ -945,7 +945,6 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
 
     @series.xenial
     @series.bionic
-    @uses.config.contract_token
```
lucasmoura, line 948: Should we remove that line on this test ?

features/attached_enable.feature
```diff
@@ -1160,7 +1159,6 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         """
 
     @series.xenial
-    @uses.config.contract_token
```
lucasmoura, line 1163: Same question here, as we have an attach operation as one of the steps

features/ubuntu_upgrade_unattached.feature
```diff
@@ -1,10 +1,10 @@
-@uses.config.contract_token
 Feature: Upgrade between releases when uaclient is unattached
 
     @slow
     @series.all
     @uses.config.machine_type.lxd-container
     @upgrade
+    @uses.config.contract_token
```
lucasmoura, line 7: I think we can remove this line, as this is an unattached test

features/attached_enable.feature
renanrodrigo: This line is useless here, in the `Scenario`, because the `Feature` has the decorator. (And here it makes sense , because "attached_enable" implies that all scenarios will always need the token)

features/attached_enable.feature
renanrodrigo: same answer

orndorffgrant (approved):

lucasmoura (approved):

PR #2641, opened 2023-07-06
tests: fail if installing from 'prebuilt' but no debs are found

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because if we set the install_from and debs_path variables correctly, but the debs_path has no debs for the specific series, we silently run the test with whatever version if u-a-t is in the image. This leads to unexpected errors.

## Test Steps
Create some empty folder, run CI with debs_path pointing there and set install_from to prebuilt.
See it crash.

## Checklist
 - [x] I have updated or added any integration tests accordingly I guess

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/steps/machines.py
+++ b/features/steps/machines.py
@@ -1,12 +1,18 @@
 import datetime
 import logging
+import sys
 from typing import Dict, NamedTuple
 
 from behave import given, when
 from pycloudlib.instance import BaseInstance  # type: ignore
 
 from features.steps.ubuntu_advantage_tools import when_i_install_uat
-from features.util import SUT, InstallationSource, build_debs
+from features.util import (
+    SUT,
+    InstallationSource,
+    build_debs,
+    get_debs_for_series,
+)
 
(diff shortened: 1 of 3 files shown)


lucasmoura (commented):
On `UAClientBehaveConfig` we already have a check for the `prebuilt` option like [this](https://github.com/canonical/ubuntu-pro-client/blob/main/features/environment.py#L172)

I think it is better to move that logic there, because then we would need even create an instance before failing

renanrodrigo:
@lucasmoura yes I know - the problem is that the way we look for the correct debs is filtering for the `series` string in their names. When creating UAClientBehaveConfig we don't know the series.

renanrodrigo:
@lucasmoura I have moved the verification to the `given_a_sut_machine` step instead - this way it fails before creating the machine

renanrodrigo:
CI errors unrelated

lucasmoura (approved):

PR #2643, opened 2023-07-06
system: try/except logic to remove files and folders

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because this avoids a possible inconsistent state between checking if a file exists and trying to remove it.

LP: #2025731

## Test Steps
Unsure on how to replicate this without mocks...

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/system.py
+++ b/uaclient/system.py
@@ -507,9 +507,11 @@ def write_file(
 
 def ensure_file_absent(file_path: str) -> None:
     """Remove a file if it exists, logging a message about removal."""
-    if os.path.exists(file_path):
+    try:
         logging.debug("Removing file: %s", file_path)
         os.unlink(file_path)
+    except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


renanrodrigo:
Questions:
a) Should we also catch `PermissionError` here? I didn't, because we don't currently and I think the correct thing is to error out when trying to remove something we don't have the permission to.
b) Should we write unit tests for those functions? I think there is nothing there to test really.

lucasmoura (approved):

CalvoM:
I do not think we need to catch PermissionError. We could catch it at layers higher since it is called in other methods where we can check if we are root.

orndorffgrant (approved):
I agree PermissionError should bubble up and cause a real error

I agree that we don't need unit tests for these

renanrodrigo:
CI failures unrelated

PR #2644, opened 2023-07-07
Add more data to reboot_required endpoint

lucasmoura:
## Why is this needed?
The reboot required enpoint just inform the users if a reboot is required or not. However, it does not provide the data we used to generate that decision. We are updating the endpoint to provide all of this data, allowing our users to better understand the decision making process or even adapt it based on their own needs.

## Test Steps
Run the new integration test added

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/api/u/pro/security/status/reboot_required/v1.py
+++ b/uaclient/api/u/pro/security/status/reboot_required/v1.py
@@ -1,5 +1,7 @@
+from enum import Enum
 from typing import List, Optional
 
+from uaclient import livepatch
 from uaclient.api.api import APIEndpoint
 from uaclient.api.data_types import AdditionalInfo
 from uaclient.config import UAConfig
@@ -10,10 +12,10 @@ from uaclient.data_types import (
     StringDataValue,
     data_list,
 )
-from uaclient.livepatch import status
-from uaclient.security_status import (
+from uaclient.system import (
+    get_kernel_info,
     get_reboot_required_pkgs,
-    get_reboot_status,
+    should_reboot,
 )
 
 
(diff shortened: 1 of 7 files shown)


uaclient/api/u/pro/security/status/reboot_required/v1.py
```diff
@@ -1,29 +1,111 @@
...
-from uaclient.data_types import DataObject, Field, StringDataValue
-from uaclient.security_status import get_reboot_status
+from uaclient.data_types import (
+    BoolDataValue,
+    DataObject,
+    Field,
+    StringDataValue,
+    data_list,
+)
+from uaclient.livepatch import status
```
orndorffgrant, line 13: I suggest 
```suggestion
from uaclient import livepatch
```

because then later it would be `livepatch.status()` which is clear. When it's just `status` it looks like it could be the normal overall pro status function.

uaclient/security_status.py
```diff
@@ -248,30 +246,22 @@ def get_reboot_status():
...
-        # kernel related packages that require a reboot. Therefore, we
-        # will just say that a reboot is required
-        return RebootStatus.REBOOT_REQUIRED
-
-    pkg_list = pkg_list_str.split()
-    num_pkgs = len(pkg_list)
-
-    kernel_regex = "^(linux-image|linux-base).*"
-    num_kernel_pkgs = 0
+    reboot_required_pkgs = get_reboot_required_pkgs()
```
orndorffgrant, line 249: Should we instead move all this logic into the API function and replace calls to `get_reboot_status` with calls to the API?

uaclient/security_status.py
```diff
@@ -248,30 +244,22 @@ def get_reboot_status():
...
-        pkg_list_str = load_file(REBOOT_PKGS_FILE_PATH)
-    except FileNotFoundError:
-        # If we cannot load that file, we cannot evaluate if we have
-        # kernel related packages that require a reboot. Therefore, we
-        # will just say that a reboot is required
-        return RebootStatus.REBOOT_REQUIRED
-
-    pkg_list = pkg_list_str.split()
-    num_pkgs = len(pkg_list)
+    reboot_required_pkgs = get_reboot_required_pkgs()
```
renanrodrigo, line 247: the `get_reboot_status` function is not used in any other place, and could be removed from here now that the logic lives in the API file

lucasmoura:
@renanrodrigo updated

renanrodrigo:
CI error is related to a new related USN to the one we use in the bionic tests, and can be fixed in a follow up PR

PR #2645, opened 2023-07-07
Few fixes and tweaks to tests

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it addresses specific test failures we saw on Jenkins.
In the end, it adds functionality by letting us choose architectures when running our build scripts.

## Test Steps
- Run CI and see the updated tests pass
- create some other-arch-chroot for ua, and build the package there (:
- Jenkins will be able to build other arches using InstallationSource.LOCAL

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/util.py
+++ b/features/util.py
@@ -12,6 +12,8 @@ from typing import Callable, Iterable, List, Optional
 
 import yaml
 
+from uaclient.system import get_dpkg_arch
+
 SUT = "system-under-test"
 LXC_PROPERTY_MAP = {
     "image": {"series": "properties.release", "machine_type": "Type"},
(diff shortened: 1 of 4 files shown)


features/util.py
```diff
@@ -151,6 +153,7 @@ def repo_state_hash(
 
 def build_debs(
     series: str,
+    architecture: str = get_dpkg_arch(),
```
orndorffgrant, line 156: Currently, passing `architecture=None` like we do down in build.py will overwrite this default value and then we'll get an error trying to create the sbuild command. We can do something like the following to allow the build.py to work:
```suggestion
    architecture: Optional[str] = None,
```
and then
```python
if architecture is None:
    architecture = get_dpkg_arch()
```

orndorffgrant (commented):
Small thing, otherwise looks good to me!

features/util.py
renanrodrigo: Overlooked that - fixed

lucasmoura (approved):

renanrodrigo:
I'll add a fix to the apt_news test, to make sure we don't expect specific output from apt

PR #2654, opened 2023-07-17
apt_news: allow current_news api to be called as non-root

orndorffgrant:
## Why is this needed?

update-manager will be the consumer of this api and is usually not running as root

## Test Steps

```
env ./tools/test-in-lxd.sh xenial
pro api u.apt_news.current_news.v1
```

## Checklist

 - [ ] Yes
 - [x] No


lucasmoura (approved):

lucasmoura:
I have tested this locally and it is working as expected

PR #2657, opened 2023-07-17
Add fix plan API endpoint

lucasmoura:
## Why is this needed?
We are adding support to two new endpoints:

```
u.pro.security.fix.cve.plan.v1
u.pro.security.fix.usn.plan.v1
```

Those endpoints detail all of the necessary steps we need to actually fix a CVE/USN without executing any of those steps.
This PR is the first step required for us to implement the execute part of the API while also modifying the CLI command to use the endpoints as well

## Test Steps
Run all of the new integration tests and certify that the API endpoint is consistent with the result we see for the CLI output

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (approved):
lgtm minor nits
CI error unrelated

uaclient/messages.py
```diff
@@ -1384,3 +1383,26 @@ class TxtColor:
...
+API_JSON_DATA_FORMAT_ERROR = FormattedNamedMessage(
+    "api-json-data-format-error",
+    "Error parsing API json data parameter:\n{data}",
+)
+
+INVALID_SECURITY_ISSUE = FormattedNamedMessage(
+    "invalid-security-issue",
+    """\
+Error: issue "{issue_id}" is not recognized.\n
+CVE should follow the pattern CVE-yyyy-nnn.\n
```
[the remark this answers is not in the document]
lucasmoura: Going with the first suggestion

uaclient/api/api.py
```diff
@@ -35,8 +39,76 @@
...
+                ErrorWarningObject(
+                    title=API_UNKNOWN_ARG.format(arg=k).msg,
+                    code=API_UNKNOWN_ARG.name,
+                    meta={},
+                )
+            )
+
+        kwargs[k] = v
+
+    return kwargs, warnings
```
orndorffgrant, line 107: Some of this can be replaced by `endpoint.options_cls.from_dict(json_data)` and that would give us type checking as well

uaclient/api/u/pro/security/fix/cve/plan/v1.py
```diff
@@ -0,0 +1,21 @@
+# from uaclient.api.api import APIEndpoint
```
orndorffgrant: This looks like it got left in this commit by mistake

uaclient/fix.py
```diff
@@ -0,0 +1,720 @@
...
+        operation: FixStepType,
+        data: Dict[str, Any],
+    ):
+        # just to make mypy happy
+        fix_step = None  # type: Optional[FixPlanStep]
+
+        if operation == FixStepType.ATTACH:
+            fix_step = FixPlanAttachStep(
+                order=self.order, data=AttachData.from_dict(data)
+            )
```
orndorffgrant, line 315: I don't remember the `order` field or why we need it.
It looks like keeping track of `order` complicates the `register_step` (and `register_warning` etc) functions because without it I think registering a step could look more like
```
plan.register_step(FixPlanNoOpStep(data=NoOpData(status=FixPlanNoOpStatus.NOT_AFFECTED)))
```

But if we do need `order`, then this looks good

features/schemas/cve_fix_plan.json
```diff
@@ -0,0 +1,91 @@
...
+                      "patternProperties": {
+                        "^.*$": {
+                          "anyOf": [
+                            {"type": "string"},
+                            {"type": "array", "items": {"type": "string"}}
+                          ]
+                        }
+                      },
+                      "additionalProperties": false
+                    },
```
orndorffgrant, line 41: We could make this more complete by specifying every possible data object using `anyOf`, but not sure if it's worth it. This is good for now.

orndorffgrant (approved):
Just a couple comments, but overall looks good!

features/api_fix_plan.feature
```diff
@@ -0,0 +1,239 @@
...
+    @series.lts
+    @uses.config.machine_type.lxd-container
+    Scenario Outline: Fix command on an unattached machine
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `pro api u.pro.security.fix.cve.plan.v1 --data '{"cves": ["CVE-1800-123456"]}'` as non-root
+        Then stdout is a json matching the `api_response` schema
+        And the json API response data matches the `cve_fix_plan` schema
+        And stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"cves_data": {"cves": \[{"error": {"code": "security-fix-not-found-issue", "msg": "Error: CVE-1800-123456 not found."}, "expected_status": "error", "plan": \[\], "title": "CVE-1800-123456", "warnings": \[\]}\], "expected_status": "error"}}, "meta": {"environment_vars": \[\]}, "type": "CVEFixPlan"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
```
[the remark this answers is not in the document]
lucasmoura: The main problem is the version field we have on the APIs. We could try a substring, but I don't think that would be that much better. But I don't have a strong opinion here

uaclient/fix.py
lucasmoura: The order is just to for the user to easily combine what we have in warning with the plan and get the exact order of execution of fix. Additinally, we can use that when we use this endpoint in the CLI, allowing us to keep the same message order we currently have

features/api_fix_plan.feature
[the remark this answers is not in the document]
orndorffgrant: ohhh right - I forgot and skimmed past it here.
We could use a `$behave_var` for it but that can be a future improvement.

uaclient/api/api.py
lucasmoura: We already perform that conversion just after we call that function. Although I do agree that maybe we just need to load the json data, we would then loose the warnings about extra keys. I don't have a strong opinion here, if we decide to simplify this, I don't see a problem

features/schemas/cve_fix_plan.json
lucasmoura: I am okay with either approach. What do you think @renanrodrigo ?

renanrodrigo: Maybe a future iteration could cover the specification of the objects?

orndorffgrant: Works for me +1

uaclient/fix.py
```diff
@@ -0,0 +1,720 @@
+import enum
```
orndorffgrant, line 1: Since the code in this file is only used by the api, and the data structures defined here _are_ the API (and can't be changed without breaking compat), should this code live in the `api.u.pro.security.fix` module?

lucasmoura: That is fair, I will move the entire module to this path

orndorffgrant:
CI fails unrelated

PR #2660, opened 2023-07-19
Add Ubuntu-themed header

s-makin:
```
This change will further align us with Ubuntu branding.

I haven't changed much past the default in the starter pack
so I can tailor it to our needs. Let me know what other links
we should add to the "more resources" menu, or any other changes
you'd like me to make. When I uprate the devdocs, they will be
added to this header (so they align with ubuntu.com docs)
```

 - [ ] Yes
 - [X] No

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -6,3 +6,4 @@ sphinx_autobuild
 sphinx_copybutton
 sphinx_design
 sphinxcontrib-mermaid
+sphinxcontrib-jquery
--- a/docs/_static/css/header.css
+++ b/docs/_static/css/header.css
@@ -1,0 +1,167 @@
+.p-navigation {
+  border-bottom: 1px solid var(--color-sidebar-background-border);
+}
+
+.p-navigation__nav {
+  background: #333333;
+  display: flex;
+}
+
+.p-logo {
+  display: flex !important;
+  padding-top: 0 !important;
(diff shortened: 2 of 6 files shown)


s-makin:
> @s-makin just a question, do we need to keep the link around in the header ? Because if we click on the logo it already redirects you to the right place right ?
> 
> My position would be to remove it or make sure that it is highlighted somewhere else in the landing page

No, I don't think we need to - actually my intention would be to replace it with something more useful to avoid the repetition. I'm just not sure what to replace it with!

lucasmoura:
I think having nothing there would be okay right now

orndorffgrant:
I think we could put a link to the discourse somewhere in the header too.
Another idea is to have a "Services" dropdown with links to ubuntu.com pages for esm, livepatch, fips etc.

Also there is an extra list item in the "More resources" dropdown above "Github" which is empty

s-makin:
> I think we could put a link to the discourse somewhere in the header too. Another idea is to have a "Services" dropdown with links to ubuntu.com pages for esm, livepatch, fips etc.
> 
> Also there is an extra list item in the "More resources" dropdown above "Github" which is empty

Great suggestion - thanks Grant! I've added the list of services as suggested. I wasn't sure whether the GitHub link really belonged there after that so I've removed it.

orndorffgrant (approved):

PR #2662, opened 2023-07-20
Look for the snapd package instead of looking for the snap command

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because we don't need two ways to check if snapd is there. If the package is installed, then the command exists. Otherwise, we install it and it creates the binary. Weird combinations (overriding the binary??) is not supported, so we don't need to check for that.

## Test Steps
It is refactor-ish, so CI should take care of this.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -486,13 +486,9 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         if required_snaps is None:
             return True
 
-        if not system.which(snap.SNAP_CMD):
+        if not snap.is_snapd_installed():
             event.info("Installing snapd")
             snap.install_snapd()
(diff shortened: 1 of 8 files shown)


CalvoM (approved):
LGTM

orndorffgrant (approved):
LGTM, it is weird that the kinetic lxd test is failing now - maybe just double check it but I don't think this PR caused it

renanrodrigo:
@orndorffgrant kinetic went EOL yesterday! I will send another PR today removing this old  legacy outdated integration test from our suite (:

orndorffgrant:
Oh right! Yep that test breaking makes sense then :upside_down_face:

PR #2668, opened 2023-07-24
jammy fips preview 1: support enabling fips on jammy

orndorffgrant:
**Do not merge**

This is a special preview release to go in a PPA. The purpose of this is to allow the certification lab to test `pro enable fips` on jammy before we officially support the feature.


orndorffgrant:
jammy fips support will be in v30 to be released next week so we don't need to keep this PR around anymore.

PR #2670, opened 2023-07-26
logging_revamp: add journald logging for daemon and timer (SC-1595)

CalvoM:
## Why is this needed?

We are replacing the 2 extra log files i.e. daemon and timer log files with systemd 
logging since they are called via systemd services.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/apport/source_ubuntu-advantage-tools.py
+++ b/apport/source_ubuntu-advantage-tools.py
@@ -20,14 +20,11 @@ def add_info(report, ui=None):
             "ua-status.json-error",
             "livepatch-status.txt",
             "livepatch-status.txt-error",
+            "pro-journal.txt",
             os.path.basename(cfg.cfg_path),
             os.path.basename(cfg.log_file),
-            os.path.basename(cfg.timer_log_file),
-            os.path.basename(cfg.daemon_log_file),
             os.path.basename(cfg.data_path("jobs-status")),
             os.path.basename(defaults.CONFIG_DEFAULTS["log_file"]),
-            os.path.basename(defaults.CONFIG_DEFAULTS["timer_log_file"]),
-            os.path.basename(defaults.CONFIG_DEFAULTS["daemon_log_file"]),
         }
         for f in auto_include_log_files:
             attach_file_if_exists(report, os.path.join(output_dir, f), key=f)
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -38,7 +38,6 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
         Then stdout matches regexp:
         """
         /var/log/ubuntu-advantage.log
-        /var/log/ubuntu-advantage-timer.log
         """
         When I run `logrotate --force /etc/logrotate.d/ubuntu-advantage-tools` with sudo
         And I run `sh -c "ls /var/log/ubuntu-advantage* | sort -d"` as non-root
@@ -46,8 +45,6 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
         """
         /var/log/ubuntu-advantage.log
         /var/log/ubuntu-advantage.log.1
-        /var/log/ubuntu-advantage-timer.log
-        /var/log/ubuntu-advantage-timer.log.1
         """
 
         Examples: ubuntu release
@@ -744,8 +741,8 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
         """
         { "metering_timer": "notanumber", "update_messaging_timer": -10 }
         """
-        And I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
-        Then I verify that running `grep "Invalid value for update_messaging interval found in config." /var/log/ubuntu-advantage-timer.log` `with sudo` exits `0`
+        And I run `systemctl start ua-timer.service` with sudo
+        Then I verify that running `sh -c 'journalctl -u ua-timer.service | grep "Invalid value for update_messaging interval found in config."'` `with sudo` exits `0`
         And I verify that the timer interval for `update_messaging` is `21600`
         And I verify that the timer interval for `metering` is `14400`
         When I create the file `/var/lib/ubuntu-advantage/jobs-status.json` with the following:
--- a/features/daemon.feature
+++ b/features/daemon.feature
@@ -61,16 +61,16 @@ Feature: Pro Upgrade Daemon only runs in environments where necessary
     Scenario Outline: daemon should run when appropriate on gcp generic lts
         Given a `<release>` machine with ubuntu-advantage-tools installed
         # verify its enabled, but stops itself when not configured to poll
-        When I run `cat /var/log/ubuntu-advantage-daemon.log` with sudo
-        Then stdout matches regexp:
+        When I run `journalctl -o cat -u ubuntu-advantage.service` with sudo
+        Then stdout contains substring:
         """
         daemon starting
         """
-        Then stdout matches regexp:
+        Then stdout contains substring:
         """
         Configured to not poll for pro license, shutting down
         """
-        Then stdout matches regexp:
+        Then stdout contains substring:
         """
         daemon ending
         """
@@ -94,7 +94,10 @@ Feature: Pro Upgrade Daemon only runs in environments where necessary
         When I run `sed -i s/#DefaultMemoryAccounting=no/DefaultMemoryAccounting=yes/ /etc/systemd/system.conf` with sudo
         When I run `systemctl daemon-reexec` with sudo
 
-        When I run `truncate -s 0 /var/log/ubuntu-advantage-daemon.log` with sudo
+        # on bionic, systemd version=237; which does not allow for log rotation + vacuum in same line e.g.
+        # journalctl --flush --rotate --vacuum-time=1s
+        When I run `journalctl --flush --rotate` with sudo
+        When I run `journalctl --vacuum-time=1s` with sudo
         When I run `systemctl restart ubuntu-advantage.service` with sudo
 
         # wait to get memory after it has settled/after startup checks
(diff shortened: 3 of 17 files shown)


orndorffgrant (commented):
Bad news: python3-systemd is not in ubuntu-minimal so we should avoid adding it at all possible (there is a tangential problem here of what to do about our auto-attach service currently depending on it without it being in Depends, but we should solve that separately - I'll create an issue for it).

Good news: we don't need python3-systemd for this because we only want to log to journald for our systemd services, and for those if we just log to stdout or stderr, systemd will push the logs into journald for us automatically.

apport/source_ubuntu-advantage-tools.py
```diff
@@ -22,12 +22,8 @@ def add_info(report, ui=None):
             "livepatch-status.txt-error",
             os.path.basename(cfg.cfg_path),
             os.path.basename(cfg.log_file),
-            os.path.basename(cfg.timer_log_file),
-            os.path.basename(cfg.daemon_log_file),
```
orndorffgrant, line 26: We should include the new equivalents of these

features/daemon.feature
```diff
@@ -61,16 +61,16 @@ Feature: Pro Upgrade Daemon only runs in environments where necessary
     Scenario Outline: daemon should run when appropriate on gcp generic lts
         Given a `<release>` machine with ubuntu-advantage-tools installed
         # verify its enabled, but stops itself when not configured to poll
-        When I run `cat /var/log/ubuntu-advantage-daemon.log` with sudo
-        Then stdout matches regexp:
+        When I run `journalctl -o cat -u ubuntu-advantage.service` with sudo
```
orndorffgrant, line 64: Why do we need the `cat`?
```suggestion
        When I run `journalctl -u ubuntu-advantage.service` with sudo
```

features/schemas/ua_status.json
```diff
@@ -209,9 +209,6 @@
         "log_file": {
           "type": "string"
         },
-        "timer_log_file": {
-          "type": "string"
-        },
         "license_check_log_file": {
           "type": "string"
         },
```
orndorffgrant, line 214: this looks old and can be removed while we're here

lib/daemon.py
```diff
@@ -4,48 +4,22 @@
...
 
 
 def main() -> int:
-    setup_logging(
-        logging.INFO,
-        logging.DEBUG,
-        defaults.CONFIG_DEFAULTS["daemon_log_file"],
-        logger=LOG,
-    )
+    LOG.setLevel(logging.DEBUG)
```
orndorffgrant, line 15: We still need our formatter and redaction filter - I think we still need a daemon.setup_logging or similar function

lib/daemon.py
```diff
@@ -4,48 +4,22 @@
...
     # by propagating to the root logger
     LOG.propagate = False
-    daemon_logger.propagate = False
-    # The root logger should only log errors to the daemon log file
-    setup_logging(
-        logging.CRITICAL,
-        logging.ERROR,
-        cfg.daemon_log_file,
-    )
-    http.configure_web_proxy(cfg.http_proxy, cfg.https_proxy)
```
orndorffgrant, line 48: please don't delete this line

lib/daemon.py
```diff
@@ -4,48 +4,22 @@
...
-from uaclient.daemon import (
-    poll_for_pro_license,
-    retry_auto_attach,
-    setup_logging,
-)
+from uaclient.daemon import poll_for_pro_license, retry_auto_attach
 
-LOG = logging.getLogger("ubuntupro.lib.daemon")
+LOG = logging.getLogger("ubuntupro.daemon")
+root_logger = logging.getLogger("ubuntupro")
```
orndorffgrant, line 11: This isn't the root logger. Also, why is it defined as a module level variable?

uaclient/actions.py
```diff
@@ -195,8 +195,6 @@ def _get_state_files(cfg: config.UAConfig):
     return [
         cfg.cfg_path or DEFAULT_CONFIG_FILE,
         cfg.log_file,
-        cfg.timer_log_file,
```
orndorffgrant, line 198: We still want to collect the logs of these, so we need to update collect_logs

uaclient/conftest.py
```diff
@@ -13,8 +13,13 @@
 # Those need to be mocked here, before importing our modules, so the pytest
 # virtualenv doesn't cry because it can't find the modules
 m_apt_pkg = mock.MagicMock()
+m_systemd_pkg = mock.MagicMock()
+m_systemd_pkg.journal = mock.MagicMock()
+m_systemd_pkg.journal.JournalHandler = mock.MagicMock()
+m_systemd_pkg.journal.JournalHandler.return_value.level = logging.DEBUG
 sys.modules["apt"] = mock.MagicMock()
 sys.modules["apt_pkg"] = m_apt_pkg
+sys.modules["systemd"] = m_systemd_pkg
```
orndorffgrant, line 22: do we still need these changes?

orndorffgrant (commented):
Just did a quick review pass and left some initial comments

uaclient/conftest.py
CalvoM: Oh no, I had removed this. I guess I had not pushed the changes.

features/daemon.feature
renanrodrigo: :black_cat:

CalvoM: Honestly, I just copied it from [unattached_commands.feature](https://github.com/canonical/ubuntu-pro-client/blob/ef9fe5d042abb52435a2147fa285e3b655150887/features/unattached_commands.feature#L406)

orndorffgrant: I just looked up the option in `man journalctl` and it seems reasonable to use

features/attached_commands.feature
```diff
@@ -751,8 +748,8 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
         { "metering_timer": "notanumber", "update_messaging_timer": -10 }
         """
         And I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
-        Then I verify that running `grep "Invalid value for update_messaging interval found in config." /var/log/ubuntu-advantage-timer.log` `with sudo` exits `0`
-        And I verify that the timer interval for `update_messaging` is `21600`
+        Then I verify that running `sh -c journalctl -t python3 | grep "Invalid value for update_messaging interval found in config."` `with sudo` exits `0`
```
orndorffgrant, line 751: If you want to select the timer logs I suggest using the `-u` option for journalctl
```suggestion
        Then I verify that running `sh -c journalctl -u ua-timer.service | grep "Invalid value for update_messaging interval found in config."` `with sudo` exits `0`
```

orndorffgrant (commented):
This is looking good! We'll also want this for our other systemd services as well (apt-news.service, esm-cache.service, ua-auto-attach.service, ua-reboot-cmds.service) - would it be easy to include them here? If that would bloat this PR too much, then we can do that later

lib/daemon.py
```diff
@@ -4,47 +4,22 @@
...
-    setup_logging(
-        logging.CRITICAL,
-        logging.ERROR,
-        cfg.daemon_log_file,
-    )
+    setup_journald_logging(logging.ERROR, logging.getLogger("ubuntupro"))
+
+    cfg = UAConfig()
+
+    # Make sure the ubuntupro.daemon logger does not generate double logging
```
orndorffgrant, line 22: This comment should be above the LOG.propagate line

lib/timer.py
```diff
@@ -179,24 +179,14 @@ def run_jobs(cfg: UAConfig, current_time: datetime):
...
-    setup_logging(
-        logging.DEBUG,
-        log_file=cfg.timer_log_file,
-        logger=LOG,
-    )
-    # Make sure the ua-timer logger does not generate double logging
-    LOG.propagate = False
-    # The root logger should log any error to the timer log file
-    setup_logging(logging.ERROR, log_file=cfg.timer_log_file)
+    # Make sure the ubuntupro.timer logger does not generate double logging
```
orndorffgrant, line 189: This comment should be above the LOG.propagate line

uaclient/log.py
```diff
@@ -1,11 +1,14 @@
...
 import logging
 import os
+import tempfile
 from collections import OrderedDict
 from typing import Any, Dict, List  # noqa: F401
 
 from uaclient import defaults, system, util
 
+PRO_TMP_DIR = os.path.join(tempfile.gettempdir(), "proclient")
+
```
orndorffgrant, line 11: This can be removed

uaclient/tests/test_lib_migrate_user_config.py
```diff
@@ -195,8 +195,6 @@ def test_file_access_error(
                     "contract_url": "https://contracts.canonical.com",
                     "log_level": "debug",
                     "log_file": "/var/log/ubuntu-advantage.log",
-                    "daemon_log_file": "/var/log/ubuntu-advantage-daemon.log",
-                    "timer_log_file": "/var/log/ubuntu-advantage-timer.log",
```
orndorffgrant, line 199: This dict is supposed to represent old conf file contents, which could contain these fields. Please add it back.

orndorffgrant (commented):
Just some final polish things, then I think this will be good!

lib/daemon.py
CalvoM: Ah, forgot this.

features/attached_commands.feature
```diff
@@ -751,8 +748,8 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
         { "metering_timer": "notanumber", "update_messaging_timer": -10 }
         """
         And I run `python3 /usr/lib/ubuntu-advantage/timer.py` with sudo
-        Then I verify that running `grep "Invalid value for update_messaging interval found in config." /var/log/ubuntu-advantage-timer.log` `with sudo` exits `0`
-        And I verify that the timer interval for `update_messaging` is `21600`
+        Then I verify that running `sh -c journalctl -u ua-timer.service | grep "Invalid value for update_messaging interval found in config."` `with sudo` exits `0`
+        Then I verify that the timer interval for `update_messaging` is `21600`
```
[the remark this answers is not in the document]
CalvoM: Hahahha... I was wondering why this test passed, I expected it to fail. It has been on my mind all this time. Thanks for pointing it out.

orndorffgrant (approved):
Looks great thank you!

lucasmoura (approved):

PR #2671, opened 2023-07-27
deps: fix yaml version after cython 3

CalvoM:
## Why is this needed?

`tox -e py310-test` fails since [cython upgraded version to 3](https://github.com/yaml/pyyaml/issues/724).
It affects pyyaml `5.4.1`. Downgrading to `5.3.1` helped.

## Test Steps

`tox -e py310-test`

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/tests/constraints/constraints-jammy.txt
+++ b/uaclient/tests/constraints/constraints-jammy.txt
@@ -5,4 +5,4 @@ pycodestyle==2.8.0
 pyflakes==2.4.0
 pytest==6.2.5
 pytest-cov==3.0.0
-pyyaml==5.4.1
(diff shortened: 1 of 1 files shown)


renanrodrigo:
I could not reproduce this error, but maybe the reason is that I am running Jammy, and the packaged version of `pyyaml` does not carry the same problem. (And pip does not try to install it again if it is installed in the system I guess)

Besides that, we cannot change this dependency because our `py310` tests are targetting Jammy, and `5.3.1` is not available on Jammy.

CalvoM:
@renanrodrigo I agree, I meant to remove this PR for your second reason.

PR #2656, opened 2023-07-17
pro enable landscape (SC-1508)

orndorffgrant:
TODO:
- [x] break up commits
  - tests: bool behave userdata
  - tests: support arbitrary config var replacement
  - tests: machine-name template var
  - tests: better errors in output comparison steps
  - tests: long stdin step
  - system: is service active helper
  - cli: pass everything after "--" as extra_args to action
  - pro enable landscape
- [x] unit tests
- [x] change integration test to use staging server
- [x] update resource on staging contract server
- [x] fix all integration tests
- [x] finalize messages/help text

## Why is this needed?

This is `pro enable landscape` as spec-ed with the landscape team.

## Test Steps

The new integration tests require some new secrets
```
tox -e behave-any -- features/landscape.feature
```

Also try it out manually
```
./tools/test-in-lxd.sh mantic
pro attach $token
pro enable landscape
```

## Checklist

 - [x] I have updated or added any unit tests accordingly
 - [x] I have updated or added any integration tests accordingly
 - [x] Changes here need to be documented, and this was done in: #2674 to be completed after merging this to get it in for the impending release

## Does this PR require extra reviews?

 - [x] Yes - help text and messaging should be reviewed by landscape pm - done
 - [ ] No

PR #2674, opened 2023-07-27
pro enable landscape docs

orndorffgrant:
Docs for #2656

### DIFF
--- a/docs/howtoguides/enable_landscape.rst
+++ b/docs/howtoguides/enable_landscape.rst
@@ -1,0 +1,64 @@
+How to enable Landscape
+***********************
+
+You can register a machine with Landscape via the ``pro enable landscape``
+command. You can register interactively for convenience, or non-interactively
+which is useful for hands-off automation.
+
+To register a machine, you'll need, at minimum, your Landscape Account Name
+and a name for the machine you are registering. If you're not using Landscape
+SaaS (https://landscape.canonical.com), then you'll also need the URL of your
+hosted Landscape server.
+
+Enable interactively
+====================
+
+To register your machine by interactively providing your Landscape account
+details at the CLI, run:
+
+.. code:: bash
+
+   sudo pro enable landscape
+
+This command will install ``landscape-client`` and start up an interactive
+wizard to complete the Landscape registration for the machine.
+
+Enable non-interactively
+========================
+
+If you know the details of your Landscape setup then you can register a
+machine without using the wizard. Under the hood, `pro` installs and executes
+``landscape-config``, so you can pass any `parameters supported by`_
+``landscape-config`` to ``pro enable landscape`` after a ``--``. You should
+also use the ``--assume-yes`` flag to automatically accept the defaults for
+any un-provided parameters.
+
+The command to enable Landscape takes the following format:
+
+.. code:: bash
+
+   sudo pro enable landscape <pro enable parameters> -- <landscape-config parameters>
+
+Which, when the parameters are added, should look something like this:
+
+.. code:: bash
+
+   sudo pro enable landscape --assume-yes -- --account-name <my-account> --computer-title <my-computer>
+
+That command will install ``landscape-client`` and pass the provided parameters
+after ``--`` to the ``landscape-config`` tool to automatically register the
+machine.
+
+What next?
+==========
+
+After successfully running ``pro enable landscape``, either interactively
+or non-interactively, an administrator of your Landscape account will need to go
+to the "Pending Computers" page in Landscape to accept the machine you just
+registered.
+
+And that's it! The machine should now appear in the Landscape dashboard for
+management.
+
+.. LINKS:
+.. _parameters supported by: https://manpages.ubuntu.com/landscape-config
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -65,6 +65,18 @@ that stream specifically on those machines.
            the `Livepatch documentation`_, or learn
            :doc:`how to enable it<howtoguides/enable_livepatch>` with ``pro``.
 
+       * **Landscape**
+
+         `Landscape`_ is the leading management and administration tool for
+         Ubuntu. It can manage up to 40,000 Ubuntu machines with a single
+         interface, and can automate security patching, as well as hardening
+         and compliance. Landscape can be `self-hosted`_ or you can use Canonical's
+         `Landscape SaaS`_ offering.
+
+         * Find out more about
+           :doc:`registering a machine with Landscape<howtoguides/enable_landscape>`
+           using ``pro``.
+
        Want more information?
 
     .. tab-item:: Security certification
@@ -280,6 +292,9 @@ constructive feedback.
 .. _Livepatch: https://ubuntu.com/security/livepatch
 .. _Livepatch documentation: https://ubuntu.com/security/livepatch/docs
 .. _list of supported kernels: https://ubuntu.com/security/livepatch/docs/livepatch/reference/kernels
+.. _Landscape: https://ubuntu.com/landscape
+.. _self-hosted: https://ubuntu.com/landscape/pricing
+.. _Landscape SaaS: https://ubuntu.com/landscape/pricing
 .. _real-time kernel: https://ubuntu.com/realtime-kernel
 .. _more information about real-time Ubuntu: https://ubuntu.com/kernel/real-time/contact-us
 .. _Landscape: https://ubuntu.com/landscape
--- a/docs/howtoguides/enable_index.rst
+++ b/docs/howtoguides/enable_index.rst
@@ -9,5 +9,6 @@ How to enable...
    CIS <enable_cis.md>
    ESM-Infra and ESM-Apps <enable_esm_infra.md>
    FIPS <enable_fips.md>
+   Landscape <enable_landscape.rst>
    Livepatch <enable_livepatch.md>
    Real-Time Kernel <enable_realtime_kernel.md>


docs/howtoguides/enable_landscape.rst
```diff
@@ -0,0 +1,47 @@
+How to enable Landscape
+***********************
+
+``pro`` offers a convenient way to register a machine with Landscape via the
+``pro enable landscape`` command.
```
s-makin, line 5: `pro` offers convenient ways of accessing all the Pro services ;) can we remove that bit and just leave "you can register a machine with Landscape via the ``pro enable landscape`` command. Then, let's add a bit about the two different options (interactive/non-interactive) for enabling so that we can add a couple of the keywords that might lead people to find the page if they're searching for it (like hands off/automatic, that sort of thing).

s-makin: Probably also an idea to state up front what info they'll need to complete the registration (which I assume is the same info no matter which method they choose?)

docs/howtoguides/enable_landscape.rst
```diff
@@ -0,0 +1,47 @@
...
+   sudo pro enable landscape
+
+That command will install ``landscape-client`` and drop you into an interactive
+wizard to complete the Landscape registration for the machine.
+
+Enable non-interactively
+========================
+
+If you know the details of your Landscape setup and would like to register a
+machine without any interaction necessary, you can pass ``landscape-config``
```
s-makin, line 24:
```suggestion
machine without using the wizard, you can pass ``landscape-config``
```

docs/howtoguides/enable_landscape.rst
```diff
@@ -0,0 +1,47 @@
...
+parameters to ``pro enable landscape`` after a ``--``. You should also use the
+``--assume-yes`` flag to ensure there are no prompts.
```
s-makin, line 26: One question I have (and is probably not really a problem for us but still might occur to a reader) is - if we can and should avoid the prompts by using --assume-yes, why are the prompts there by default? Surely it should have been that yes is assumed by default and that you can turn that off by doing --assume-no. Can you --assume-no? What is assume-yes actually doing? (I assume it's not just suppressing prompts, but is rather doing something else - I think it's better to be clear here to avoid raising questions). (also sorry, that's way more than one question)

docs/howtoguides/enable_landscape.rst
s-makin: Also, a question a reader might naturally have is "what's landscape-config and do I need to install it?". It's probably worth mentioning that it's part of the pro enable landscape command.

docs/howtoguides/enable_landscape.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Altogether, it should look something like this:
+
+.. code:: bash
+
+   sudo pro enable landscape --assume-yes -- --account-name my-account --computer-title my-computer
```
s-makin, line 32:
```suggestion
The command to enable Landscape takes the following format:

.. code:: bash

   sudo pro enable landscape -- <landscape-config parameters>

Which, when the parameters are added, should look something like this:

.. code:: bash

   sudo pro enable landscape --assume-yes -- --account-name <my-account> --computer-title <my-computer>
```

s-makin: Also, do we need to list all the (currently) possible parameters?

docs/howtoguides/enable_landscape.rst
```diff
@@ -0,0 +1,47 @@
...
+That command will install ``landscape-client`` and pass the provided parameters
+after ``--`` to the ``landscape-config`` tool to automatically register the
+machine.
+
+What next?
+==========
+
+After successfully running ``pro enable landscape``, either interactively
+or non-interactively, an administrator of your Landscape account will need to go
+to the Pending Computers page in Landscape to accept the machine you just
```
s-makin, line 43:
```suggestion
to the "Pending Computers" page in Landscape to accept the machine you just
```

docs/howtoguides/enable_landscape.rst
```diff
@@ -0,0 +1,47 @@
...
+registered.
+
+And that's it! The machine should now show up in the Landscape dashboard for
```
s-makin, line 46:
```suggestion
And that's it! The machine should now appear in the Landscape dashboard for
```

docs/index.rst
```diff
@@ -65,6 +65,19 @@ that stream specifically on those machines.
            the `Livepatch documentation`_, or learn
            :doc:`how to enable it<howtoguides/enable_livepatch>` with ``pro``.
 
+       * **Landscape**
+
+         `Landscape`_ is the leading management and administration tool for
+         Ubuntu. It can manage up to 40,000 Ubuntu machines with a single
+         interface and can automate security patching as well as hardening
```
s-makin, line 72:
```suggestion
         interface, and can automate security patching, as well as hardening
```

docs/index.rst
```diff
@@ -65,6 +65,19 @@ that stream specifically on those machines.
...
+         and compliance. Landscape can be `self-hosted`_ or you can use Canonical's
+         `Landscape SaaS`_ offering. You can register a machine with Landscape
+         using ``pro enable landscape``.
```
s-makin, line 75: Do we need to include the command to enable here? It sort of negates the need to include the guide immediately after it

docs/howtoguides/enable_landscape.rst
s-makin: Another question that users might have (which occurred to me after I saw the 40k machines thing added to the index) - how do I register more than a few machines at once? Doing this one by one for an entire fleet would be quite inefficient so is there some method we recommend for doing this at larger scale?

s-makin (commented):
Generally looks really good, thanks! Not many nits, but I do have a few comments and questions

docs/howtoguides/enable_landscape.rst
```diff
@@ -0,0 +1,47 @@
...
+====================
+
+To register your machine by interactively providing your Landscape account
+details at the CLI, run:
+
+.. code:: bash
+
+   sudo pro enable landscape
+
+That command will install ``landscape-client`` and drop you into an interactive
```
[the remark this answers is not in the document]
orndorffgrant: Yep I'll change it to that wording

docs/howtoguides/enable_landscape.rst
orndorffgrant: > `pro` offers convenient ways of accessing all the Pro services ;) can we remove that bit and just leave "you can register a machine with Landscape via the `pro enable landscape` command. Then, let's add a bit about the two different options (interactive/non-interactive) for enabling so that we can add a couple of the keywords that might lead people to find the page if they're searching for it (like hands off/automatic, that sort of thing).

Sounds good, I'll remove/add these.

> Probably also an idea to state up front what info they'll need to complete the registration (which I assume is the same info no matter which method they choose?)

Good idea, I'll add it.

> how do I register more than a few machines at once?

This is out of scope for pro-client. The correct answer is dependent on the customer's infrastructure setup. We can add a line about the non-interactive case being useful for automation, but it is up to the customer to run the command on many machines.

docs/howtoguides/enable_landscape.rst
orndorffgrant: I'll make the suggested change.

`landscape-config` isn't part of `pro enable landscape` per se. It is a command installed as part of the `landscape-client` package and owned by the landscape team. It could completely change underneath of us without us knowing, so we shouldn't try to document it too much. The only web-accessible documentation that I can find (which I would've linked before but didn't find it until now), is the manpage on line here: https://manpages.ubuntu.com/landscape-config

So I'll add that link and an explanation that under-the-hood, pro enable landscape installs landscape-config and executes it, which should answer the "do I need to install it" part

docs/howtoguides/enable_landscape.rst
orndorffgrant: I'll make that change, and I posted above a link that I'll include which describes the possible parameters

docs/index.rst
orndorffgrant: Right, I'll just remove this last sentence and keep the link to the guide after

docs/howtoguides/enable_landscape.rst
orndorffgrant: We only _should_ avoid prompts if you want a completely non-interactive enablement. The prompts are there in the `landscape-config` command that we don't own/control. 

Prompts can be useful - if you provide some but not all required parameters to landscape-config, it will prompt you for the other required fields. If you miss a required parameter with `--assume-yes` then it will just fail. The "interactive but with some parameters supplied" is a valid invocation of the command, just not something I think is worth recommending - so I didn't document it here.

The "problem" is that `landscape-config` currently prompts for non-required parameters, and that is why we _have_ to recommend `--assume-yes` here. That is arguably a UX problem with `landscape-config` but that is out of our control (and the landscape team is aware of it).

What `--assume-yes` is really doing is accepting defaults for those non-required parameters - that should probably be the description instead of "ensure there are no prompts".

orndorffgrant:
@s-makin I think I addressed everything - please take another look

docs/howtoguides/enable_landscape.rst
s-makin: Understood, thanks for the context there. I think that given this situation, the edit you made is the best one that could be made. It neatly sidesteps the entire thorny issue - good job :)

s-makin (approved):
Really nice work, thanks Grant :)

Issue #2580, opened 2023-05-09
Enabling ESM app and ESM infra does not set a pin priority for the ESM archives


philroche:
I discussed this with @lucasmoura @orndorffgrant @renanrodrigo @dodys in person last week at the Canonical sprint in Prague.

The issue is that when ESM apps or ESM infra are enabled the ESM repos are added, but there is never a pin priority created, to ensure that any packages that do exist in ESM are installed from the ESM repos.

As ESM is all about security I believe this is a flaw in how these pro features are enabled and could potentially leave users with vulnerable systems they believed to be secure.

This is best described by walking through the following scenario with Ubuntu 20.04.

Day 1: package chkrootkit exist in universe with version 0.55-4
Day 2: pro is attached on the  Ubuntu 20.04 and ESM apps and infra is enabled
Day 3: A CVE is found in chkrootkit and is patched in ESM with version 0.55-4.1. 
Day 4: Ubuntu 20.04 instance receives update from ESM and  chkrootkit is upgraded to 0.55-4.1and CVEs are mitigated
Day 4 - Day 26: Ubuntu 20.04 instance is secure and chkrootkit is not vulnerable
Day 27: An update to chkrootkit is uploaded to universe with  version 0.55-5
Day 28: Ubuntu 20.04 instance receives update from universe and  chkrootkit is upgraded to 0.55-5. 
Day 28 - Day 32: Ubuntu 20.04 instance is no longer secure and is vulnerable as the 0.55-4.1 patches were not pushed upstream
Day 32: Canonical/Ubuntu security team rebase chkrootkit from universe and re-apply the patches to mitigate the CVE. This is uploaded as 0.55-5.1
Day 33: Ubuntu 20.04 instance receives update from ESM and  chkrootkit is upgraded to 0.55-5.1and CVEs are mitigated once again.

In the above scenario you can see that from Day 28 to Day 32 the system is vulnerable despite having ESM enabled and the package being present in the ESM repos.

If we set pin priority for the ESM repos then this would not happen .... but it would mean that feature updates or bug updates from universe could be delayed until the security team perform their rebase. I feel, given the whole selling point or ESM that this is a price worth paying.

This issue is to track the request to have ESM repos pinned by default.

alexmurray:
The versioning used above is not correct (although the outcome is similar) - ie. if  0.55-4 is patched in ESM, this gets given the version number  0.55-4+esm1 - so Day 3 above should use this version.

[...]

Issue #2680, opened 2023-07-31
Move fips/fips-updates pinning to static files


renanrodrigo:
Instead of creating the pin files dinamically, we ship those as static files (as was decided for ESM) so we a) standardize how we deal with it and b) leverage all apt/dpkg mechanisms to deal with changes in configuration files.

orndorffgrant:
This idea has gotten high level SRU approval, but needs product/security review to make sure it doesn't have unexpected consequences.

PR #2681, opened 2023-07-31
Pin esm but only esm

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it 
Fixes: #2580
context is provided in the issue and the internal spec

This is a version of #2664 which does not change the existing FIPS logic - thus, we have both ways to pin repositories (static and dynamic) at the same time. #2680 was created to reflect the next steps of this work.

## Test Steps
Run CI and the new `sru/` test.

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo:
CI errors unrelated

orndorffgrant (approved):

PR #2682, opened 2023-07-31
contract: change default value of lastAttachment so server doesn't try to parse empty string

orndorffgrant:
## Why is this needed?

Machines attached with a pro version <=28.1 don't have a lastAttachment time. The fallback value of "" was trying to be parsed by the contract server, which resulted in errors. The new fallback value of `None` is treated as nil by the backend and isn't parsed.

## Test Steps

```
env SHELL_BEFORE=1 ./tools/test-in-lxd.sh xenial
apt update
apt install ubuntu-advantage-tools
pro attach $token
exit # new version gets installed after exit and lxc shell is re-started
sudo pro disable esm-infra
# Before this PR: you'd get an error here
# With this PR: this should succeed
```

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -417,7 +417,7 @@ class UAContractClient(serviceclient.UAServiceClient):
                 "resources": [service.name for service in enabled_services],
                 "lastAttachment": attachment_data.attached_at.isoformat()
                 if attachment_data
-                else "",
+                else None,
             }
         else:
             activity_info = {}


orndorffgrant:
@TeodorPt FYI

renanrodrigo (approved):
dis good

orndorffgrant:
CI fails unrelated

PR #2683, opened 2023-07-31
logging: no more cli console logging

orndorffgrant:
## Why is this needed?

This removes some old code around conditionally logging messages to the console. In most cases, that is what we do, and in some of the few cases where we didn't disable the console logger, that was a bug.

We should treat the console as our cli UX and treat the logs as logs. There should be no instance where a log message is also part of our UX. In the rare case where we want to print the same message to the User and our logs, then that can be two separate statements.

The exception is the `--debug` flag which is useful to include log statements in the middle of the console output - that functionality is kept and enhanced by removing the formatter that stripped a lot of useful information from those logs.

## Test Steps

CI should pass.

Please review all instances of logging.info, logging.warn, logging.error, and logging.exception. Please double check that, where necessary, the messages are also printed/event.infoed properly. This includes statements that are not a part of the diff.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/yaml.py
+++ b/uaclient/yaml.py
@@ -10,6 +10,7 @@ try:
     import yaml
 except ImportError:
     LOG.error(MISSING_YAML_MODULE.msg)
+    print(MISSING_YAML_MODULE.msg, file=sys.stderr)
     sys.exit(1)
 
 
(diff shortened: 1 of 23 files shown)


uaclient/yaml.py
```diff
@@ -10,22 +10,27 @@
     import yaml
 except ImportError:
     LOG.error(MISSING_YAML_MODULE.msg)
+    print(MISSING_YAML_MODULE.msg, file=sys.stderr)
```
lucasmoura, line 13: I think it should be fine to use `print`, but I just want to double check if we should change to `event.info` as well

orndorffgrant:
@lucasmoura updated!

lucasmoura (approved):

PR #2691, opened 2023-08-03
docs: add entry about pinning esm-infra and esm-apps

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it exposes the pinning of the esm services, and adds a pointer to the relevant configuration documentation.

### DIFF
--- a/docs/explanations/about_esm.md
+++ b/docs/explanations/about_esm.md
@@ -61,3 +61,30 @@ You can manage `esm-infra` and `esm-apps` using `pro` on the command line. To
 find out how, read our guide on
 [enabling and disabling these services](../howtoguides/enable_esm_infra.md)
 on your machine.
+
+## Are `esm-infra` and `esm-apps` packages preferred over regular updates?
+
+Yes. The Pro Client will deliver the following configuration files to apt to
+ensure the `esm` updates are preferred when running `apt upgrade`:
+
+- `/etc/apt/preferences.d/ubuntu-pro-esm-infra`
+- `/etc/apt/preferences.d/ubuntu-pro-esm-apps`
+
+These files are *pinning* the packages coming from `esm` to give them a higher
+priority than the standard defined in apt. This means that the version to be
+installed by apt (in `install` or `upgrade`) or by unattended-upgrades will
+always be the highest `esm` version available for a given package, even if a
+higher version is present on a non-esm source.
+
+This behavior guarantees that if you have `esm-infra` or `esm-apps` enabled,
+your system will always have the `esm` patches installed for any package
+available in the ESM repositories.
+
+Although the preference files listed above will always be delivered by the
+package, they will only take effect when the packages from the sources they
+reference are available, i.e. when the services are enabled. Otherwise, it
+is safe to keep those files around.
+
+Check the
+[APT Configuration article in the Debian Wiki](https://wiki.debian.org/AptConfiguration#apt_preferences_.28APT_pinning.29)
+to learn more about pinning and priorities.


docs/explanations/about_esm.md
```diff
@@ -61,3 +61,20 @@ You can manage `esm-infra` and `esm-apps` using `pro` on the command line. To
...
+## Are `esm-infra` and `esm-apps` packages preferred over regular updates?
+
+Yes. The Pro Client will deliver the following configuration files to apt to
+ensure the `esm` updates are preferred when running `apt upgrade`:
+
+- `/etc/apt/preferences.d/ubuntu-pro-esm-infra`
+- `/etc/apt/preferences.d/ubuntu-pro-esm-apps`
+
+Those files are *pinning* the respective `esm` origins to 510 - a value which
+is higher than the standard (500) defined in apt. Although the files will
```
s-makin, line 74: Do we need to include the values? Because without already knowing what pinning is or how it works, the values could be really any values and hold no meaning for the reader. I think that if you want to include the values, we will need to explain what pinning is (however briefly). I'm also not 100% sure what you mean by "esm origins". It seems to imply there's more than one place to get esm updates from. What packages are supplied at the standard 500? (i.e., what is pinning esm to 510 giving preference over?)

renanrodrigo: I removed specificities (as origins or values) and tried to better explain what is the impact about the pinning.
Please let me know if there is any more improvement needed.

docs/explanations/about_esm.md
```diff
@@ -61,3 +61,29 @@ You can manage `esm-infra` and `esm-apps` using `pro` on the command line. To
...
+
+## Are `esm-infra` and `esm-apps` packages preferred over regular updates?
+
+Yes. The Pro Client will deliver the following configuration files to apt to
+ensure the `esm` updates are preferred when running `apt upgrade`:
+
+- `/etc/apt/preferences.d/ubuntu-pro-esm-infra`
+- `/etc/apt/preferences.d/ubuntu-pro-esm-apps`
+
+Those files are *pinning* the packages coming from `esm` to a higher value than
```
s-makin, line 73:
```suggestion
These files are *pinning* the packages coming from `esm` to give them a higher priority than
```

docs/explanations/about_esm.md
```diff
@@ -61,3 +61,29 @@ You can manage `esm-infra` and `esm-apps` using `pro` on the command line. To
...
+the standard defined in apt. This means that the version to be installed by apt
+(in `install` or `upgrade`) or by unattended-upgrades will always be the
+highest esm version available for a given package, even if a higher version is
```
s-makin, line 76:
```suggestion
highest `esm` version available for a given package, even if a higher version is
```

docs/explanations/about_esm.md
```diff
@@ -61,3 +61,29 @@ You can manage `esm-infra` and `esm-apps` using `pro` on the command line. To
...
+present on a non-esm source.
+
+This behavior guarantees that if you have esm-infra or esm-apps enabled, your
```
s-makin, line 79:
```suggestion
This behavior guarantees that if you have `esm-infra` or `esm-apps` enabled, your
```

docs/explanations/about_esm.md
```diff
@@ -61,3 +61,29 @@ You can manage `esm-infra` and `esm-apps` using `pro` on the command line. To
...
+system will always have the esm patches installed for any package available in
```
s-makin, line 80:
```suggestion
system will always have the `esm` patches installed for any package available in
```

docs/explanations/about_esm.md
```diff
@@ -61,3 +61,29 @@ You can manage `esm-infra` and `esm-apps` using `pro` on the command line. To
...
+the esm repositories.
```
s-makin, line 81:
```suggestion
the ESM repositories.
```

docs/explanations/about_esm.md
```diff
@@ -61,3 +61,29 @@ You can manage `esm-infra` and `esm-apps` using `pro` on the command line. To
...
+
+Although the files will always be delivered by the package, they will only take
+effect when the packages from those origins are available, i.e. when the
+services are enabled.
```
s-makin, line 85: I found this a bit confusing because of the terminology switch. Which files are we referring to here? Do we mean that patches will be downloaded but not installed if the services are disabled?

s-makin (commented):
I think I meant to apply these comments in a review but now they're in a batch of suggestions. I think we're almost there!

docs/explanations/about_esm.md
renanrodrigo: Well it means that, if the services are disabled, there is no harm in keeping those files around

renanrodrigo: and the files are the preferences files... I tried to rewrite it.

s-makin (approved):
Thanks for your changes, I think it's much clearer now exactly what's happening.

PR #2694, opened 2023-08-04
Release 29 review

lucasmoura:


debian/ubuntu-advantage-tools.postinst
```diff
@@ -256,11 +256,11 @@ remove_old_systemd_units() {
...
             # it is there so that the system doesn't say it is degraded.
             # If the old timer was not running, then this is a noop.
-            systemctl --system daemon-reload > /dev/null || true
-            systemctl reset-failed ua-license-check.timer > /dev/null 2>&1 || true
+            deb-systemd-invoke --system daemon-reload > /dev/null || true
+            deb-systemd-invoke reset-failed ua-license-check.timer > /dev/null 2>&1 || true
             # In rare race-condition scenarios, the service can also get into
             # the same failed state.
-            systemctl reset-failed ua-license-check.service > /dev/null 2>&1 || true
+            deb-systemd-invoke reset-failed ua-license-check.service > /dev/null 2>&1 || true
```
orndorffgrant, line 263: `deb-systemd-invoke` only supports start stop and restart which is why we have to use systemctl here
See https://salsa.debian.org/debian/init-system-helpers/-/blob/master/script/deb-systemd-invoke

debian/source/lintian-overrides
```diff
@@ -1,8 +1,2 @@
-# Lintian doesn't see dh-systemd alternative when building on xenial
-ubuntu-advantage-tools: missing-build-dependency-for-dh_-command dh_systemd_start => dh-systemd
-
```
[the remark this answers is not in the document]
sergiodj: Good question. I believe it's OK leaving the override, then.  Although I don't see any of us building the package manually on Xenial, it's still a good idea to document a behaviour that happens there because we still support the distro.

sergiodj:
Thanks, guys.  I left two comments which I hope can clarify questions raised by @renanrodrigo and @orndorffgrant.  Otherwise, this PR LGTM.

debian/ubuntu-advantage-tools.postinst
orndorffgrant: Aha! yep my grepping failed me there. Cool +1 to this change then; I think we need a changelog entry to go with it though?

sergiodj: Good idea, +1 to having a changelog entry (I'm all for having more changelog entries!).

lucasmoura:
@sergiodj @orndorffgrant @renanrodrigo updated

sergiodj (approved):

renanrodrigo (approved):
We got rant in the past because changes to the changelog should be in separate commits than the changes themselves. But if @sergiodj is fine as is, so am I!

PR #2695, opened 2023-08-04
docs: add anbox documentation

lucasmoura:
## Why is this needed?
We are adding information about the Anbox Cloud service that will be supported through the Pro Client in the next release


docs/references/compatibility_matrix.md
```diff
@@ -3,16 +3,17 @@
...
+| fips            | &#x2705;  |  &#x2705;  | &#10060;   | &#x2705;| &#x2705;| &#x2705; | N/A      | &#10060;      | &#10060;         | &#x2705; | &#x2705;    | &#x2705;    |
+| fips-updates    | &#x2705;  |  &#x2705;  | &#10060;   | &#x2705;| &#x2705;| &#x2705; | &#10060; | N/A           | &#10060;         | &#x2705; | &#x2705;    | &#x2705;    |
+| realtime-kernel | &#x2705;  |  &#x2705;  | &#10060;   | &#x2705;| &#x2705;| &#x2705; | &#10060; | &#10060;      | N/A              | &#x2705; | &#x2705;    | &#x2705;    |
+| ros             | &#x2705;  |  &#x2705;  | &#x2705;   | &#x2705;| &#x2705;| &#x2705; | &#x2705; | &#x2705;      | &#x2705;         | N/A      | &#x2705;    | &#x2705;    |
+| ros-updates     | &#x2705;  |  &#x2705;  | &#x2705;   | &#x2705;| &#x2705;| &#x2705; | &#x2705; | &#x2705;      | &#x2705;         | &#x2705; | N/A         | &#x2705;    |
+| anbox-cloud     | &#x2705;  |  &#x2705;  | &#x2705;   | &#x2705;| &#x2705;| &#x2705; | &#x2705; | &#x2705;      | &#x2705;         | &#x2705; | &#x2705;    | &#x2705;    |
```
lucasmoura, line 19: @s-makin I don't like how the table is presented after this new change. Let me know if you have a better idea on how to address that

s-makin: We had the same concerns last time and now it's come back to bite us :/ I've been digging, and I think the only way we'll be able to display this table sensibly (e.g., with vertical top headers) is by using html and css. We could use tabs, to only show one column at a time, but that would be weird. Although if we followed the grouping on the front page, that might work? The columns could then just be the columns applicable to a specific service need, and it wouldn't matter if we showed all rows for each. We might need to have a chat about it later, let me know what you think!

lucasmoura: Yes, I agree we should chat about this for sure

lucasmoura:
@s-makin I have updated the header text to be vertically align on the compatibility matrix table

docs/_static/css/custom.css
```diff
@@ -1,6 +1,20 @@
 /** Fix the font weight (300 for normal, 400 for slightly bold) **/
 /** Should be 100 for all headers, 400 for normal text  **/
 
+/** We are creating both of these styles only to have the
+ *  headers of the compatibility matrix to be vertically
+ *  aligned. This should not impact the other tables
```
renanrodrigo, line 6: I would say it slightly different:
The impact for other tables is that headers will be forcedly aligned to the bottom line, but that does not change the visualization

Or some better phrase of course

lucasmoura: That's a great suggestion. I have updated the PR to use it

s-makin:
> @lucasmoura I think this PR needs to be rebased on `docs-devel` without the extra commits that are only in `docs` and not in `docs-devel` yet.

Yeah, the table looks really nice now but I was confused about all the extra stuff that suddenly appeared

orndorffgrant:
@lucasmoura The livepatch row, fips-updates column needs to be updated to a green check.

Also I'm experimenting with a less repetitive way of displaying the table, hopefully posting another comment soon

orndorffgrant:
@s-makin @lucasmoura I came up with this but I'm not actually convinced its easier to read so feel free to ignore.

image

docs/howtoguides/enable_anbox.md
```diff
@@ -0,0 +1,45 @@
+# How to enable Anbox Cloud
+
+```{important}
+[Anbox Cloud is supported on 20.04 and 22.04 releases](https://anbox-cloud.io/).
+```
+
+To use Anbox, you will need to enable it directly throught the Ubuntu Pro Client,
+which will install all of the necessary snaps and setup the APT sources needed for
+the service.
+
```
s-makin, line 10: In our other guides, we explain how to update Pro Client. Should we do that here as well, or does it not matter what version they use for Anbox?

docs/howtoguides/enable_anbox.md
```diff
@@ -0,0 +1,45 @@
...
+```{important}
```
s-makin, line 11: Can we combine this admonition with the one at the top of the page? It really confused me to see two of them so close together right at the start of the guide.

docs/howtoguides/enable_anbox.md
```diff
@@ -0,0 +1,45 @@
...
+Anbox Cloud enabled
+To finish setting up the Anbox Cloud Appliance, run:
+
+$ sudo anbox-cloud-appliance init
+
+You can accept the default answers if you do not have any specific
+configuration changes.
+For more information, see https://anbox-cloud.io/docs/tut/installing-appliance
+```
+
```
s-makin, line 40: Do we need to include anything about disabling/removing it?

docs/howtoguides/enable_anbox.md
lucasmoura: That is a good idea, I will add that section

docs/howtoguides/enable_anbox.md
lucasmoura: I moved it to another part of the doc. Let me know what you think

docs/howtoguides/enable_anbox.md
lucasmoura: I  have added that section

lucasmoura:
@s-makin Do you think Grant's table suggestion is better than what we currently have ?

docs/howtoguides/enable_anbox.md
```diff
@@ -0,0 +1,76 @@
...
+## Disable the service
+
+If you wish to disable Anbox, you can use the following command to
+disable it:
+
+```bash
+sudo pro disable anbox-cloud
+```
+
+Note that this command will only remove the APT sources, but not uninstall the snaps.
```
s-makin, line 76: This seems like for a useability pov that it should also remove the snaps. Is that something we have control over, or is it for the Anbox team?

docs/howtoguides/enable_anbox.md
```diff
@@ -0,0 +1,76 @@
+# How to enable Anbox Cloud
+
+```{important}
+[Anbox Cloud is supported on 20.04 and 22.04 releases](https://anbox-cloud.io/).
+```
+
+To use Anbox, you will need to enable it directly throught the Ubuntu Pro Client,
+which will install all of the necessary snaps and setup the APT sources needed for
```
s-makin, line 8:
```suggestion
which will install all of the necessary snaps and set up the APT sources needed for
```

docs/howtoguides/enable_anbox.md
```diff
@@ -0,0 +1,76 @@
...
+## Enable Anbox
+
+To enable Anbox Cloud, run:
+
+```console
+$ sudo pro enable anbox-cloud
+```
+
+```{important}
+The Anbox Cloud service can only be installed on containers using the --access-only flag.
```
s-makin, line 30:
```suggestion
The Anbox Cloud service can only be installed on containers using the `--access-only` flag.
```

s-makin: (to make sure the `--` characters show correctly)

docs/howtoguides/enable_anbox.md
```diff
@@ -0,0 +1,76 @@
...
+This option will only setup the APT sources for anbox, but not install any of the snaps.
```
s-makin, line 31:
```suggestion
This option will only set up the APT sources for Anbox, but not install any of the snaps.
```

docs/howtoguides/enable_anbox.md
```diff
@@ -0,0 +1,76 @@
...
+For more information, see https://anbox-cloud.io/docs/tut/installing-appliance
+```
+
+```{important}
+Please note that the output states an additional step is required to
+complete the Anbox Cloud setup
+```
+
+You can also confirm that the service is enabled by running the `pro status` command.
+It should contain the following line for `anbox-cloud`
```
s-makin, line 60:
```suggestion
It should contain the following line for `anbox-cloud`:
```

docs/howtoguides/enable_anbox.md
lucasmoura: Usually, we don't remove any packages/snaps that we install when enabling a service. Anbox is just following that principle

lucasmoura:
@s-makin updated

PR #2696, opened 2023-08-04
docs: add reference for fix plan API endpoints

lucasmoura:
## Why is this needed?
Add references for the fix plan API endpoints that will be released in the next Pro release


docs/references/api.rst
```diff
@@ -127,6 +127,21 @@ will be populated like this if we have a connectivity issue when running a
...
+
+* ``--args``: Use this to individually provide arguments to the CLI endpoint.
+
+  For example: `pro api u.pro.attach.magic.revoke.v1 --args magic_token=TOKEN`
+
+* ``data``: Use this to provide a JSON object containing all the data:
+
+  For example: `pro api u.pro.security.fix.cve.plan.v1 --data '{"cves": ["CVE-1234-1235"]}'`
+
+  Notice that only the *plan* endpoints support the `--data` parameter.
```
orndorffgrant, line 143: > Notice that only the *plan* endpoints support the `--data` parameter.

Is this true? Do we want this to be true?

orndorffgrant: data should be usable by full_auto_attach as well I think

docs/references/api.rst
```diff
@@ -613,6 +630,386 @@ like ``systemctl start ubuntu-advantage.service``.
...
           version is available.
 
+u.pro.security.fix.cve.plan.v1
+===============================
+
+This endpoint shows the necessary steps required to fix CVEs in the system without
+executing any of those steps.
+
+- Introduced in Ubuntu Pro Client Version: ``29~``
+- This endpoint requires the `--data` argument
```
orndorffgrant, line 640: > - This endpoint requires the `--data` argument

I think this is already captured in the CLI usage section

docs/references/api.rst
```diff
@@ -613,6 +630,386 @@ like ``systemctl start ubuntu-advantage.service``.
...
+               - Type
+               - Description
+             * - ``expected_status``
+               - *str*
+               - The expected status of fixing the CVEs
+             * - ``cves``
+               - *List[FixPlanResult]*
+               - A list of FixPlanResult objects
+
+        - ``uaclient.api.u.pro.security.fix import FixPlanResult``
```
orndorffgrant, line 688: It wasn't until seeing it laid out in our docs like this that I realized: This is an unversioned return type :/

This is what we're releasing and it's okay, but we should consider changing it next release. I think all return types should be tied directly to the version of the endpoint they're used in. So I think ideally this might get defined in something like `api.u.pro.security.fix._common.v1.FixPlanResult`, but we wouldn't tell users about that - instead we'd import and re-export as `api.u.pro.security.fix.cve.plan.v1.FixPlanResult` so then the complete nested return type is all defined in the module of the endpoint itself.

orndorffgrant (commented):
I think defining the complete nested return type for each endpoint makes sense. I think trying to refer to a common definition would be harder to follow

docs/references/api.rst
lucasmoura: That's an excellent point and I will make this change for the next release

lucasmoura:
@orndorffgrant updated

docs/references/api.rst
```diff
@@ -613,6 +629,544 @@ like ``systemctl start ubuntu-advantage.service``.
...
+        - No exceptions raised by this endpoint.   
+
+   .. tab-item:: CLI interaction
+      :sync: CLI
+
+      - Calling from the CLI:
+
+        .. code-block:: bash
+
+           pro api u.pro.security.fix.usn.plan.v1 --data '{"usns": [...]}'
```
orndorffgrant, line 1137: Same comment as above
```suggestion
           pro api u.pro.security.fix.usn.plan.v1 --data '{"usns": ["USN-1234-1", "USN-1235-1"]}'
```

orndorffgrant (commented):
Just some nits

docs/references/api.rst
```diff
@@ -613,6 +629,544 @@ like ``systemctl start ubuntu-advantage.service``.
...
+u.pro.security.fix.usn.plan.v1
+===============================
+
+This endpoint shows the necessary steps required to fix USNs in the system without
+executing any of those steps.
+
+- Introduced in Ubuntu Pro Client Version: ``29~``
+- Args:
+
+  - ``usns``: A list of USNs (i.e CVE-6188-1) titles
```
s-makin, line 901: I'm not sure if this is intended or not, but a couple of times in this section about the usn endpoint there are references to CVE (like here). Might be worth a double-check to make sure that any references to CVE for this endpoint are intended (I'm not sure!)

s-makin (commented):
Just two quick nits, otherwise I think this should be good to go

docs/references/api.rst
lucasmoura: Good catch @s-makin, this is indeed incorrect. I will fix that

lucasmoura:
@orndorffgrant @s-makin updated

s-makin (approved):
LGTM :) thanks for all your work on this

orndorffgrant (approved):

PR #2697, opened 2023-08-07
tests: use prod environment for magic-attach test

orndorffgrant:
The staging esm environment is presenting some problems when the apt-helper command tries to verify the credential for some services. We are now switching for the production environment to mitigate that issue

already in `release`, just bringing to `main` so our CI stops being so red


renanrodrigo:
CI is still missing `landscape` on some status tests, so it will be red for a little while :D

renanrodrigo:
I made my CI less red by changing to colorblind mode. It is yellow now. Feels less bad.

renanrodrigo (approved):

orndorffgrant:
Ah we just updated landscape on prod yesterday, so that messed up some tests. I'll fix those today

PR #2702, opened 2023-08-10
tests: mask iscsi service to avoid intermittent errors on github runners

orndorffgrant:
## Why is this needed?
iscsid.service has been getting intermittent resource errors when restarted as part of a full apt upgrade to ESM on xenial containers. This only seems to happen on our github runners and I haven't been able to reproduce it locally. Since this service isn't necessary for testing pro-client, we can mask it to let the tests continue.

## Test Steps
All I want is for CI to be green


renanrodrigo (approved):
godlike

PR #2703, opened 2023-08-11
`uaclient` to `ubuntupro`

orndorffgrant:
## Why is this needed?

This is a proof-of-concept for renaming our python module name from uaclient to ubuntupro while maintaining compatibility for releases prior to 24.04

Before moving forward, we need:
- [ ] approval of new module name
- [ ] approval of having both module names on x,b,f,j,l,m
- [ ] pre-SRU review
- [ ] integration tests using old module name on old releases
- [ ] anything else?

## Test Steps
```
./tools/test-in-lxd.sh xenial
pro help
python3 -c "from ubuntupro import config, status; print(status.status(config.UAConfig()))"
python3 -c "from uaclient import config, status; print(status.status(config.UAConfig()))"
```

## Checklist

 - [x] Yes
 - [ ] No


debian/ubuntu-advantage-tools.postinst
```diff
@@ -493,6 +493,17 @@ case "$1" in
...
       fi
 
+      if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "30~"; then
+          if [ "$VERSION_ID" = "16.04" ] \
+              || [ "$VERSION_ID" = "18.04" ] \
+              || [ "$VERSION_ID" = "20.04" ] \
+              || [ "$VERSION_ID" = "22.04" ] \
+              || [ "$VERSION_ID" = "23.04" ] \
+              || [ "$VERSION_ID" = "23.10" ]; then
+              ln -s ubuntupro /usr/lib/python3/dist-packages/uaclient
```
orndorffgrant, line 503: we should do this in a pythonic way isntead like import *

orndorffgrant: we shuold test reverse deps agains this shim

orndorffgrant:
I did get an informal "yeah that should be okay as long as you test it well against reverse deps" SRU opinion.
We should eventually do this, but this PR is now way out of date and should be closed. It is a good example of what would need to be done.
The main thing we'd want to do differently is use python imports to support the old `uaclient` name (e.g. actually have an __init__.py that imports everything from `ubuntupro`) as opposed to the linking method used here. There may be odd side effects with object references with the linking approach.
So I'll close this for now and leave a "Breaks" so that we don't forget about it.

orndorffgrant:
Breaks: rename `uaclient` python package name to `ubuntupro`

orndorffgrant:
I forgot: "Breaks" only works for merged PRs. Creating issue manually.

PR #2705, opened 2023-08-14
Add airgapped/offline documentation

s-makin:
## Why is this needed?
This PR solves all of our problems because...we have a lot of questions
about airgapping/offline.

```
Adding documentation about implementing Pro in an offline/airgapped
environment. 

```

### DIFF
--- a/docs/explanations/using_pro_offline.rst
+++ b/docs/explanations/using_pro_offline.rst
@@ -1,0 +1,68 @@
+Using Ubuntu Pro in an offline environment
+******************************************
+
+Airgapping (or air gapping) is a security measure that involves isolating a
+machine or network from external/unsecured networks to protect sensitive data.
+Airgapped systems usually operate “internetless”, or with limited network
+connectivity. 
+
+Ubuntu Pro can still be used in an offline environment, via an airgapped
+contract server and local mirrors. This allows Ubuntu Pro tokens to be
+distributed across the isolated network, and provides the security support and
+other benefits of Pro even without internet access.
+
+How do I sync Pro packages in an airgapped system?
+==================================================
+
+This is an advanced use-case, and our `Customer Support teams`_ can help you
+implement this in the best way for your specific environment. Note that this
+requires a paid subscription to Ubuntu Pro.
+
+What tools do I need?
+=====================
+
+Updating Ubuntu systems in an offline environment requires mirroring Ubuntu
+repositories. Since Ubuntu packages can be consumed via deb packages and/or
+snaps, the tool needed to mirror Ubuntu repositories depends on where you
+source your packages from.
+
+Mirror deb repositories with Landscape
+--------------------------------------
+
+Landscape is a management tool that can host a mirror of Ubuntu deb
+repositories. 
+
+In future release cycles, Landscape will also include features specific to
+airgapped use-cases. You can keep up to date with these developments in the
+`Landscape Beta Discourse`_.
+
+Mirror snaps with Snap store proxy
+----------------------------------
+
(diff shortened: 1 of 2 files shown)


orndorffgrant (commented):
You happened to stumble upon a pet peeve of mine :smile: 

I consider the difference between airgapped and  "airgapped"  to be important and probably worth two separate pages. Here airgapped == offline/internetless and "airgapped" == firewalled/limited network access + desire to move most things on-prem.

I think separating is worthwhile especially because the tooling support is different for both scenarios. For example, I know that Livepatch on-prem works for firewalled environments, but I'm not sure if it works for true airgapped environments. We'd have to check the other tooling recommendations as well, and we may want to link to different sections of their documentation for the different use-cases.

s-makin:
> I consider the difference between airgapped and "airgapped" to be important and probably worth two separate pages. Here airgapped == offline/internetless and "airgapped" == firewalled/limited network access + desire to move most things on-prem.

That's a really interesting point, thanks! I thought these were sub-aspects of the same thing, rather than two separate use-cases. That's something I can and should reflect on this page. However, I don't think having two separate pages is a good idea. 

First reason is because we're not actually documenting how to achieve airgapping - this page exists only to tell people "yes, you *can* use Pro airgapped, but you have to talk to the Support team to do it" (it's basically a signpost and redirect). I don't want us to set a precedent of spending our docs "real estate" on documenting things that are handled by other teams. 

Second reason is because we're getting more searches for "airgapped" than either offline or on-prem, and on-prem could itself refer to two different use-cases. I think I'd rather just rework this page to represent the fact that these are two different use-cases, to make it easier for people searching for "air gapped" to find the information they're looking for by directing everyone to one page. I don't want to confuse people unnecessarily, or require them to use specific terminology to find their content. As I've recently discovered with cloud-init, the search bar doesn't use Google, and so searching doesn't behave the way we think it does.

I'd be happy to have a quick sesh later (or whenever) to discuss -- if we can make this page into a "catchall" for any queries about airgapped, offline, and possibly on-prem that would be ideal. Hopefully without further peeving :D

s-makin:
I've made the changes we discussed the other day. Please let me know if there are any bits of wording that still need to be tweaked for accuracy :)

docs/explanations/using_pro_offline.rst
```diff
@@ -0,0 +1,68 @@
...
+that after your device has been registered with the proxy, all communications
+with the Snap store will be sent via the proxy.
+ 
+Security patching
+=================
+
+Livepatch is the tool that provides patching for critical and high security
+vulnerabilities in the Ubuntu kernel. It can also be used in an airgapped
+setup using Livepatch on-prem, which our `Customer Support teams`_ can help you
+to set up.
```
orndorffgrant, line 55: I'm not sure if this part is true - I think livepatch-on-prem requires a network connection to the Canonical livepatch server. We should double check with livepatch team before promising this.

s-makin: I said airgapped when I think I should have said "firewalled", which is the language they use in the livepatch on-prem docs.

docs/explanations/using_pro_offline.rst
```diff
@@ -0,0 +1,64 @@
...
+----------------------------------
+
+The `Snap-store-proxy`_ is the equivalent tool for hosting the snap packages that
+your organisation is consuming.
+ 
+Security patching
+=================
+
+Livepatch is the tool that provides patching for critical and high security
+vulnerabilities in the Ubuntu kernel. It can also be used in an airgapped
```
s-makin, line 49:
```suggestion
vulnerabilities in the Ubuntu kernel. It can also be used in a firewalled
```

PR #2712, opened 2023-08-18
i18n prep: logging: standardize logging callsite conventions (SC-1602)

orndorffgrant:
## Why is this needed?
This commit standardizes our logging calls in a couple ways:
 - log messages should be inline strings at the callsite
 - log string formatting should use lazy % style
 - logs shouldn't share strings with printed messages

In the places where log and print calls shared a string, I separated
them and attempted to make the log string more "log-like" and the
printed string more user-friendly. I also moved the printed string
to the messages module.

## Test Steps
Please review every log call and make sure it follows these conventions now.
Please review the changes in messages.

CI should pass

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/entitlements/livepatch.py
+++ b/uaclient/entitlements/livepatch.py
@@ -139,9 +139,8 @@ service at https://ubuntu.com/security/livepatch"""
             try:
                 process_config_directives(entitlement_cfg)
             except exceptions.ProcessExecutionError as e:
-                msg = "Unable to configure Livepatch: " + str(e)
-                event.info(msg)
-                LOG.error(msg)
+                LOG.error(str(e), exc_info=e)
+                event.info("Unable to configure livepatch: %s", str(e))
                 return False
         if process_token:
             livepatch_token = entitlement_cfg.get("resourceToken")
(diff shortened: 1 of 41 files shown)


renanrodrigo (approved):
phew that is a true chore
lgtm

uaclient/entitlements/livepatch.py
```diff
@@ -288,9 +288,13 @@ def process_contract_deltas(
...
         if any([process_directives, process_token]):
-            msg = "Updating '{}' on changed directives.".format(self.name)
-            LOG.info(msg)
-            event.info(msg)
+            LOG.info(
+                "New livepatch directives or token. running "
+                "setup_livepatch_config"
+            )
+            event.info(
+                "Updating '{}' on changed directives.".format(self.name)
```
lucasmoura, line 296: Shouldn't this be on the `messages` module ?

lucasmoura: Or is this going to be addressed in another PR ?

uaclient/yaml.py
```diff
@@ -8,29 +8,31 @@
...
 
 def safe_load(stream):
     try:
         return yaml.safe_load(stream)
-    except AttributeError:
-        msg = BROKEN_YAML_MODULE.format(path=yaml.__path__).msg
-        LOG.error(msg)
-        print(msg, file=sys.stderr)
+    except AttributeError as e:
+        LOG.exception(e)
```
lucasmoura, line 21: Should we log here also a similar message to what we are printing ?

uaclient/livepatch.py
```diff
@@ -136,20 +136,21 @@ def status() -> Optional[LivepatchStatusStatus]:
...
         )
         return None
 
     try:
         status_root = LivepatchStatus.from_dict(status_json)
     except IncorrectTypeError:
         LOG.warning(
-            "canonical-livepatch status returned unexpected "
-            "structure: {}".format(out)
+            "canonical-livepatch status returned unexpected " "structure: %s",
```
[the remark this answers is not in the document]
orndorffgrant: whoops, yep I'll fix it

uaclient/yaml.py
orndorffgrant: We can if you want, but `LOG.exception` will include the error and stacktrace pointing to this line, so I think we'll be able to see what happened in the logs

lucasmoura (approved):

PR #2713, opened 2023-08-21
i18n prep: messages: move all messages to messages module (SC-1605)

orndorffgrant:
## Why is this needed?
This moves all messages that are intended to be internationalized into the messages module. While not strictly necessary, this helps keep us organized and will hopefully allow easier automated checks in the future, e.g. spell check, or did messages change without updating pot/po files

## Test Steps
Everything should work the same as before.

## Checklist

 - [ ] Yes
 - [x] No


lucasmoura (approved):

renanrodrigo (approved):
all here makes sense, if some wandering string is left behind we chase it later

PR #2715, opened 2023-08-25
Fix enabled services circular import

orndorffgrant:
## Why is this needed?
This moves an import down into the function where it is used to ~~fix~~ workaround a circular import problem

## Test Steps

The new test should pass

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/api.feature
+++ b/features/api.feature
@@ -1,6 +1,39 @@
 Feature: Client behaviour for the API endpoints
 
     @series.all
+    @uses.config.machine_type.any
+    @uses.config.machine_type.lxd-container
+    Scenario Outline: all API endpoints can be imported individually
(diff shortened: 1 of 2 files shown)


features/api.feature
```diff
@@ -1,5 +1,38 @@
 Feature: Client behaviour for the API endpoints
 
+    @series.all
+    @uses.config.machine_type.any
+    @uses.config.machine_type.lxd-container
+    Scenario Outline: all API endpoints can be imported individually
```
renanrodrigo, line 6: This is really great

renanrodrigo (approved):

features/api.feature
lucasmoura: +1

lucasmoura (approved):

PR #2716, opened 2023-08-25
Release 29.3

lucasmoura:


debian/changelog
```diff
@@ -1,3 +1,10 @@
+ubuntu-advantage-tools (29.3) mantic; urgency=medium
+
+  * api: fix circular import that prevents enabled_services
+    endpoint to be imported
```
orndorffgrant, line 4:
```suggestion
  * api: fix circular import that prevents enabled_services
    endpoint from being imported
```

orndorffgrant (approved):
nit on wording but LGTM

debian/changelog
lucasmoura: Updated

Issue #2692, opened 2023-08-03
lib/daemon.py depends on python3-systemd but u-a-t does not


orndorffgrant:
lib/daemon.py seems to work because in the only places where it is used (on AWS/GCP/Azure cloud images) `python3-systemd` is installed by default anyway (pulled in as a `Recommends` of apport).

We should do one of:
- Add `python3-systemd` to ubuntu-advantage-tools `Depends`/`Recommends` (which would bring it into `ubuntu-minimal`)
- Remove our dependency on `python3-systemd` (probably by switching ubuntu-advantage.service to Type=exec so we don't need to notify)

philroche:
The CPC team at Canonical are working on aggressive minimization of the mantic minimal cloud images. This involves moving to the https://ubuntu-archive-team.ubuntu.com/seeds/ubuntu.mantic/cloud-minimal seed and also no longer installing any recommends. One of the fallouts of this is the ubuntu-advantage service can no longer start.

```
Aug 21 14:16:25 daily-ubuntu-minimal-2310-mantic-amd64-v20230821 python3[792]: Traceback (most recent call last):
Aug 21 14:16:25 daily-ubuntu-minimal-2310-mantic-amd64-v20230821 python3[792]:   File "/usr/lib/ubuntu-advantage/daemon.py", line 5, in <module>
Aug 21 14:16:25 daily-ubuntu-minimal-2310-mantic-amd64-v20230821 python3[792]:     from systemd.daemon import notify  # type: ignore
Aug 21 14:16:25 daily-ubuntu-minimal-2310-mantic-amd64-v20230821 python3[792]:     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Aug 21 14:16:25 daily-ubuntu-minimal-2310-mantic-amd64-v20230821 python3[792]: ModuleNotFoundError: No module named 'systemd'
```

This appears to be due to the now missing `python3-systemd` package from the new mantic minimal builds. 

The reason it was present previously is because it was a recommends of `apport`

```
$ aptitude why python3-systemd
i   apport Recommends python3-systemd
```

As the`ubuntu-advantage` service won't even start without `python3-systemd` I suggest that it be added as a dependency. 

I have confirmed that installing python3-systemd and restarting the service resolves the problem.

To add some urgency to this bug - this will block publication of the new mantic minimized images until resolved.

I would urge a depends over a recommends due to the failed service start details above without python3-systemd.

PR #2717, opened 2023-08-25
systemd: change u-a.service type from notify to exec

renanrodrigo:
## Why is this needed?

This PR solves all of our problems because we have a dependency on python3-systemd to use the 'notify' function, which cannot be satisfied as the package is not part of ubuntu-minimal. This dependency was never declared in the debian/control file anyway, and we are removing it.

We don't lose much in the change anyway, as we are only configuring logging before notifying.

Fixes: #2692

## Test Steps
Execute all behave features which involve the `ubuntu-advantage.service` and see them all passing

## Checklist

 - [x] Yes - basak or panlinux
 - [ ] No

### DIFF
--- a/lib/daemon.py
+++ b/lib/daemon.py
@@ -2,8 +2,6 @@ import logging
 import os
 import sys
 
-from systemd.daemon import notify  # type: ignore
-
 from uaclient import http
 from uaclient.config import UAConfig
 from uaclient.daemon import poll_for_pro_license, retry_auto_attach
@@ -23,8 +21,6 @@ def main() -> int:
     http.configure_web_proxy(cfg.http_proxy, cfg.https_proxy)
 
     LOG.debug("daemon starting")
-
-    notify("READY=1")
 
     is_correct_cloud = any(
         os.path.exists("/run/cloud-init/cloud-id-{}".format(cloud))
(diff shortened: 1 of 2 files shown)


panlinux (approved):
Looks ok. The previous `notify` is somewhat similar to `exec`, except for the notification of course. Now the service will be considered started after the `exec` in the `fork+exec` duo, *if* the `exec` worked. Just please at some point soon run a test on the clouds where this daemon is needed :)

renanrodrigo:
[pr-2717.tar.gz](https://github.com/canonical/ubuntu-pro-client/files/12721314/pr-2717.tar.gz)
As @panlinux suggested, here are the results for all tests involving `ubuntu-advantage.service` from our CI - including the Pro ones not executed in PRs - so we can make sure all works as expected.

basak (approved):
I agree that to the best of my knowledge this shouldn't really change any behaviour since you were not doing anything of consequence before notifying "ready" anyway. However I'm not an expert on systemd.

Note though that it does affect timing, so might change the outcome of a race condition if there is a latent bug elsewhere. It might be worth considering how this service interacts with others to get some confidence in that area.

orndorffgrant:
We've tested all the expected scenarios for this service in our integration tests so we're confident that it works within our package.

We've talked about the timing change in the team. This service doesn't provide anything that could be depended on by other software. IOW there is no reason that we can think of that someone would want to time their service based on the startup of this one. So we consider the chances of this affecting another package very small.

Issue #2719, opened 2023-08-28
'permission 510'


setharnold:
https://github.com/canonical/ubuntu-pro-client/blob/279114c76cf6132e01d8e9393f7463678b96697d/features/attached_enable.feature#L236

Hello, I noticed a string like this repeated multiple times:

```
        Then apt-cache policy for the following url has permission `510`
```

The context suggests that this is a pin priority and not read+execute permission for the owner and execute permission for the group.

While this doesn't look like it actually influences anything, it can complicate looking for permissions problems or pinning problems in the future.

Thanks

cpaelzer:
You are right, that should be '_permission_' -> '_priority_' in all those cases.

I always wanted to mess with those .feature files and see how bad I would break them...

PR #2720, opened 2023-08-29
tests: fix typo permission -> priority

cpaelzer:
This is an accidental wrong word, it is only used inside the tests and not bleeding to a user or any such. Hence it can be fixed, but has no urgency.

Fixes: #2719

## Why is this needed?

It resolves an accidental typo

## Test Steps

Run the tests and watch the output no more speaking of "url permission" but instead of "priority".
(This isn't in the code active on a Ubuntu system, only in the self tests)

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (approved):
Even though CI didn't run, checked locally, all good

Issue #2722, opened 2023-08-29
chmod should not be unconditional in postinst


cpaelzer:
In https://github.com/canonical/ubuntu-pro-client/blob/main/debian/ubuntu-advantage-tools.postinst#L447 I've seen that we do chmod this on every install/upgrade.

```
      private_dir="/var/lib/ubuntu-advantage/private"
      if [ -d "$private_dir" ]; then
          chmod 0700 "$private_dir"
      fi
```

While closing LP: #2024204 I thought this should be wrong as well then.
Can we reduce that to only when we create it.
If we need this to be on a particular upgrade make it version bound and only happen on upgrade from the last version it was not done well.

cpaelzer:
I'd appreciate if one could - once this plus lp2024204 are both resolved a test that verifies the permissions "on install" as well as "on re-creation".

orndorffgrant:
Yes this is valid. We should version gate this so it only runs when upgrading from versions <= 19.6 (decided based on looking at the repo state on when this piece of code was committed).

PR #2724, opened 2023-08-30
Fix usage of PRIVATE_DIR

cpaelzer:
## Why is this needed?

This has two motivations
1. Careful review by Seth (thanks) revealed a potential race where an attacker could exchange the element we chmod to gain benefit. In this case to potentially gain execute rights. So we reduce the race by setting the permissions correctly in the mkdir() call right away. (LP: #2024204)
2. We found that there are conflicting definitions and usage of PRIVATE_DIR that cause misunderstandings and issues (like recreation of /private not having the right [too open] permissions)

## Test Steps

Usually the path is created by the postinst, which IMHO it should not do (one place is fine) see #2722 - To test the actual python code (and surprise, it was broken without a test) one needs to remove the dir created by the postinst as the python code only creates it if missing.

```
# Set up a system to be ready to be pro attached
# ensure there is no old private dir so we can check its (re-)creation
$ rm -rf /var/lib/ubuntu-advantage/private
$ strace -o verify-access.strace -rTf pro attach <redacted>
# old entries had in the log then
06:56:17.120305 mkdir("/var/lib/ubuntu-advantage/private", 0777) = 0 <0.000108>
```

The resulting file should no more have first a `mkdir` with permission `777` followed by a `chmod` to `0700`. Instead it should be created with `0700` right away.

## Checklist
 - [ ] Changes here need to be documented

 - [x] Yes (Sarnold)
 - [ ] No


uaclient/files/files.py
```diff
@@ -42,9 +42,10 @@ def write(self, content: str):
             else defaults.WORLD_READABLE_MODE
         )
         if not os.path.exists(self._directory):
-            os.makedirs(self._directory)
             if os.path.basename(self._directory) == defaults.PRIVATE_SUBDIR:
-                os.chmod(self._directory, 0o700)
+                os.makedirs(self._directory)
```
renanrodrigo, line 46: I'd say this is inverted, right?
We only want the mode to be applied if it is the private subdir

cpaelzer: Good catch, pushing a fix in a few seconds. Since this is no new logical commit I'll squash that fixup with the commit I had and force push.

renanrodrigo (approved):

PR #2728, opened 2023-09-01
Make fix CLI command rely on the fix plan API endpoint

lucasmoura:
## Why is this needed?
Since we have created the fix plan API endpoint, we can now rely on it for the CLI command as well. This PR is making the necessary changes on the API endpoint itself and while also modifying the CLI command to use the API directly

## Test Steps
Run the fix integration tests and check that the output has not changed after the CLI update

## Checklist

 - [ ] Yes
 - [x] No


uaclient/cli/__init__.py
```diff
@@ -69,9 +70,18 @@
 from uaclient.timer.update_messaging import refresh_motd, update_motd_messages
 from uaclient.yaml import safe_dump, safe_load
 
-NAME = "pro"
+EPILOG_TMPL = (
```
renanrodrigo, line 73: I think this should somehow end up in the `messages` module, as it can be translated. IMHO keeping it here will make the i18n effort harder I guess

renanrodrigo: same for other visible messages in this module

uaclient/api/u/pro/security/fix/__init__.py
```diff
@@ -318,9 +425,18 @@ def register_step(
...
+                    order=self.order, data=NoOpLivepatchFixData.from_dict(data)
+                )
+            elif "source_packages" in data:
+                fix_step = FixPlanNoOpAlreadyFixedStep(
+                    order=self.order, data=NoOpAlreadyFixedData.from_dict(data)
+                )
+            else:
+                fix_step = FixPlanNoOpStep(
+                    order=self.order, data=NoOpData.from_dict(data)
+                )
```
renanrodrigo, line 439: I like how the inheritance works here, and I don't think we need a version bump to accommodate this change. The other changes being just additions, I'd say we good overall.

uaclient/cli/fix.py
```diff
@@ -52,6 +104,747 @@ def fix_parser(parser):
...
+
+def fix_usn(
+    security_issue: str, dry_run: bool, no_related: bool, cfg: UAConfig
+):
+    fix_plan = usn_plan(
+        options=USNFixPlanOptions(usns=[security_issue]), cfg=cfg
+    )
+    error = fix_plan.usns_data.usns[0].target_usn_plan.error
+    if error:
+        raise exceptions.UserFacingError(error.msg)
```
orndorffgrant, line 163: Shouldn't `usn_plan` raise exceptions itself? Though I guess it's too late to change now.

I'm forgetting - what class of errors get returned instead of raised from usn_plan?

uaclient/cli/fix.py
```diff
@@ -52,6 +104,747 @@ def fix_parser(parser):
...
+
+    print(colorize_commands([["pro", "attach", token]]))
+    try:
+        ret_code = cli.action_attach(
+            argparse.Namespace(
+                token=token, auto_enable=True, format="cli", attach_config=None
+            ),
+            cfg,
+        )
+        return ret_code == 0
```
orndorffgrant, line 315: Now might be a good time to change this to use `actions.attach_with_token` instead

uaclient/cli/fix.py
```diff
@@ -52,6 +104,747 @@ def fix_parser(parser):
...
+
+    :return: True if enable performed.
+    """
+    import argparse
+
+    from uaclient import cli
+
+    print(messages.SECURITY_SERVICE_DISABLED.format(service=service))
+    choice = util.prompt_choices(
+        "Choose: [E]nable {} [C]ancel".format(service),
```
orndorffgrant, line 456: This string is already defined in the messages module so let's use that

uaclient/cli/fix.py
```diff
@@ -52,6 +104,747 @@ def fix_parser(parser):
...
+                argparse.Namespace(
+                    service=[service],
+                    assume_yes=True,
+                    beta=False,
+                    format="cli",
+                    access_only=False,
+                ),
+                cfg,
+            )
+        )
```
orndorffgrant, line 474: This might be a good opportunity to change this to use `actions.enable_entitlement_by_name`

uaclient/cli/fix.py
```diff
@@ -0,0 +1,866 @@
+import re
+import textwrap
+from typing import Dict, List, NamedTuple, Optional, Set, Tuple, Union  # noqa
```
renanrodrigo, line 3: Could we specify the `noqa`? I guess it is unused import, but it's better to make it explicit;
Same for the one below

uaclient/cli/fix.py
```diff
@@ -52,6 +104,747 @@ def fix_parser(parser):
...
+            textwrap.fill(
+                msg,
+                width=PRINT_WRAP_WIDTH,
+                subsequent_indent="    ",
+                replace_whitespace=False,
+            )
+        )
+
+    for step in sorted(full_plan, key=lambda x: x.order):
+        if isinstance(step, FixPlanWarningPackageCannotBeInstalled):
```
orndorffgrant, line 599: An idea: Do you think it would be any easier to read if there was a function for each step type? So this loop would look more like this:
```python
fix_context = FixContext(all the shared state that need to be updated by each step)
for step in steps:
    if isinstance(step, Klass):
        execute_klass_step(fix_context, step)
    elif isinstance(step, Klass2):
        execute_klass2_step(fix_context, step)
    ....
```

orndorffgrant (commented):
I need to do a more through review of the big rewrite of cli fix, but overall it's looking good, and I have some initial comments here

renanrodrigo (commented):
I was preparing initial comments but @orndorffgrant outtimed me again

uaclient/cli/fix.py
lucasmoura: The only reason why we don't raise the exception directly from the API is because the user can provide multiple USNs/CVEs to be evaluated at the same time. Therefore, we try to create plans for as many as we can.

For the errors, we only raise errors if the USN name is either invalid or don't exist

lucasmoura:
@renanrodrigo @orndorffgrant updated

uaclient/cli/fix.py
```diff
@@ -0,0 +1,964 @@
+import re
+import textwrap
+from typing import (  # noqa: F401
+    Dict,
+    List,
+    NamedTuple,
```
renanrodrigo, line 6: this one is actually not needed

renanrodrigo (approved):
This looks good, and it brings some other great improvements in testing and organizing the repository with the new `cli` structure.
Thanks much for all work here!

orndorffgrant (commented):
Just some follow ups on pocket descriptions and exceptions - otherwise LGTM!

uaclient/api/u/pro/security/fix/__init__.py
```diff
@@ -647,19 +833,27 @@ def _generate_fix_plan(
...
-    all_already_installed = True
     for pocket in [
         UBUNTU_STANDARD_UPDATES_POCKET,
         UA_INFRA_POCKET,
         UA_APPS_POCKET,
     ]:
         pkg_src_group = src_pocket_pkgs[pocket]
         binary_pkgs = binary_pocket_pkgs[pocket]
+        source_pkgs = [src_pkg for src_pkg, _ in pkg_src_group]
+        pocket_name = get_pocket_short_name(pocket)
```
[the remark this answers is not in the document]
lucasmoura: Yes, this is totally feasible, but I think it is better to make that change on the PR that will retire the security module. My concern is that this PR is already too big as is and even though I believe this will be a straightforward refactor, I may be wrong and the complexity of this PR will just grow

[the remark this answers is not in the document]
orndorffgrant: That sounds good to me +1

uaclient/cli/fix.py
```diff
@@ -0,0 +1,972 @@
...
+        options=CVEFixPlanOptions(cves=[security_issue]), cfg=cfg
+    )
+
+    error = fix_plan.cves_data.cves[0].error
+    if error and error.msg:
+        exc = exceptions.UbuntuProError()
+        exc.named_msg = messages.NamedMessage(
+            error.code or "",
+            error.msg,
+        )
```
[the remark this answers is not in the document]
lucasmoura: That's a great idea, let me use it instead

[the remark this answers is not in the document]
orndorffgrant: I don't think its that big of a problem, but also not idea. Thinking about it more, I think you're original idea of essentially re-raising the error is more accurate, and the extra message I suggested isn't really helpful to the user.

Can you add a more official way to raise a custom error with a NamedMessage though?
Maybe something like this in exceptions.py
```python
class AnonymousUbuntuProError(UbuntuProErro):
    """For cases where we have an NamedMessage of an Error already but it isn't in exception form, and we need to raise it as an exception. This generally should only be used in cli-only code."""
    def __init__(self, named_msg: NamedMessage, **kwargs):
        super().__init__(**kwargs)
        self.named_msg = named_msg
```

Then use it like:
```python
raise exceptions.AnonymousUbuntuProError(NamedMessage(error.code, error.msg))
```

lucasmoura:
@orndorffgrant updated

orndorffgrant (approved):
Looks great! Thanks for the final fixups

lucasmoura:
CI errors are unrelated to this feature

PR #2731, opened 2023-09-05
tests: add scenario where cloud-init is disabled

lucasmoura:
## Why is this needed?
Add a scenario where cloud-init is disabled after boot. In that situation there is no cloud-id command anymore. To test that the Pro client doesn't break, we will enable FIPS on a cloud instance to show that service is indeed enabled even though the cloud metapackage is not installed

This is also needed because there is a Launchpad [bug](https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1938208) that address the lack of this test

## Test Steps
Run the new integration test

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/enable_fips_cloud.feature
+++ b/features/enable_fips_cloud.feature
@@ -389,3 +389,53 @@ Feature: FIPS enablement in cloud based machines
            | bionic  | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips/ubuntu bionic/main |
            | focal   | FIPS         | fips         |https://esm.ubuntu.com/fips/ubuntu focal/main  |
            | focal   | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips/ubuntu focal/main  |
+
+    @slow
+    @series.lts
+    @uses.config.machine_type.any
+    @uses.config.machine_type.aws.generic
+    Scenario Outline: Attached enable of FIPS in an ubuntu image with cloud-init disabled
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `touch /etc/cloud/cloud-init.disabled` with sudo
+        And I reboot the machine
+        And I verify that running `cloud-id` `with sudo` exits `1`
+        Then stderr matches regexp:
+        """
+        File not found '/run/cloud-init/instance-data.json'. Provide a path to instance data json file using --instance-data
+        """
+        When I attach `contract_token` with sudo
+        And I run `pro enable fips --assume-yes` with sudo
+        Then stdout matches regexp:
+        """
+        Updating package lists
+        Installing FIPS packages
+        FIPS enabled
+        A reboot is required to complete install
+        """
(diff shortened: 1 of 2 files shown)


features/enable_fips_cloud.feature
```diff
@@ -389,3 +389,52 @@ Feature: FIPS enablement in cloud based machines
...
+        When I run `cat /proc/sys/crypto/fips_enabled` with sudo
+        Then I will see the following on stdout:
+        """
+        1
+        """
+
+        Examples: ubuntu release
+           | release |
+           | xenial  |
+           | bionic  |
```
CalvoM, line 438: I ran into an isssue that I cannot get `aws.generic` for bionic. I think this is okay?

renanrodrigo: this is 100% not ok! But it was fixed in the release branch - will be pulled up to main once we actually release
see https://github.com/canonical/ubuntu-pro-client/pull/2727/commits/d3bd975f045ca592fe1b12fce4763467a5602fa8

renanrodrigo: Bionic daily images are not in aws anymore because it is EOSS

features/enable_fips_cloud.feature
```diff
@@ -389,3 +389,52 @@ Feature: FIPS enablement in cloud based machines
            | bionic  | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips/ubuntu bionic/main |
            | focal   | FIPS         | fips         |https://esm.ubuntu.com/fips/ubuntu focal/main  |
            | focal   | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips/ubuntu focal/main  |
+
+    @series.lts
+    @uses.config.machine_type.any
+    @uses.config.machine_type.aws.generic
```
renanrodrigo, line 395: any reason we should run this only on AWS?

features/enable_fips_cloud.feature
```diff
@@ -389,3 +389,52 @@ Feature: FIPS enablement in cloud based machines
            | bionic  | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips/ubuntu bionic/main |
            | focal   | FIPS         | fips         |https://esm.ubuntu.com/fips/ubuntu focal/main  |
            | focal   | FIPS Updates | fips-updates |https://esm.ubuntu.com/fips/ubuntu focal/main  |
+
+    @series.lts
```
renanrodrigo, line 393: maybe mark this as a `@slow` test, as it enables fips

features/enable_fips_cloud.feature
lucasmoura: It is just that this scenario is not cloud related and I just chose one to test it on. But if you prefer, I can test it on all clouds we support

renanrodrigo: I'm completely fine with that, thanks

renanrodrigo (approved):

PR #2732, opened 2023-09-06
feat(entitlements): update specific sources list on service enable (SC-186)

aciba90:

```
feat(entitlements): update specific sources list on service enable

Repo services will apt-update while enabling only its associated ppa source list,
and `/etc/apt/sources.list` if require packages are needed to be installed.

The Snap service only apt-updates `/etc/apt/sources.list`.

This change improves boot time on Ubuntu Pro instances by 10s-15s depending
on the series.

Fixes: #1482
Fixes: #1311
```

## Additional context

[SC-186](https://warthogs.atlassian.net/browse/SC-186)
[SF: #00313994](https://bootstack.canonical.com/cases/00313994)

## Performance

Attach time comparing tip-main vs this PR with a token with esm-infra and esm-apps enabled by default is improved from 10s to 15s depending on the Ubuntu series:

attach

## Checklist

 - [ ] Yes
 - [x] No

[SC-186]: https://warthogs.atlassian.net/browse/SC-186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -115,7 +115,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         Then stdout matches regexp:
         """
         One moment, checking your subscription first
-        Updating Ubuntu Pro: ESM package list
+        Updating Ubuntu Pro: ESM Infra package list
         Ubuntu Pro: ESM Infra enabled
         """
         Examples: ubuntu release
@@ -412,6 +412,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
             replaced by 'pro enable usg'. See more information at:
             https://ubuntu.com/security/certifications/docs/usg
             Updating CIS Audit package list
+            Updating main package list
             Installing CIS Audit packages
             CIS Audit enabled
             Visit https://ubuntu.com/security/cis to learn how to use CIS
--- a/features/attached_enable.feature
+++ b/features/attached_enable.feature
@@ -17,7 +17,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         Then I will see the following on stdout:
             """
             One moment, checking your subscription first
-            Updating package lists
+            Updating CC EAL2 package list
             (This will download more than 500MB of packages, so may take some time.)
             Installing CC EAL2 packages
             CC EAL2 enabled
@@ -38,7 +38,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         Then I will see the following on stdout:
         """
         One moment, checking your subscription first
-        Updating package lists
+        Updating CC EAL2 package list
         Skipping installing packages: ubuntu-commoncriteria
         CC EAL2 access enabled
         """
@@ -115,7 +115,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         Then stdout matches regexp:
         """
         One moment, checking your subscription first
-        Updating package lists
+        Updating Ubuntu Pro: ESM package list
         Ubuntu Pro: ESM Infra enabled
         """
         Examples: ubuntu release
@@ -324,7 +324,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         Then I will see the following on stdout:
         """
         One moment, checking your subscription first
-        Updating package lists
+        Updating CIS Audit package list
         Skipping installing packages: usg-cisbenchmark usg-common
         CIS Audit access enabled
         Visit https://ubuntu.com/security/cis to learn how to use CIS
@@ -334,7 +334,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         Then I will see the following on stdout:
         """
         One moment, checking your subscription first
-        Updating package lists
+        Updating CIS Audit package list
         Installing CIS Audit packages
         CIS Audit enabled
         Visit https://ubuntu.com/security/cis to learn how to use CIS
@@ -411,7 +411,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
             From Ubuntu 20.04 and onwards 'pro enable cis' has been
             replaced by 'pro enable usg'. See more information at:
             https://ubuntu.com/security/certifications/docs/usg
-            Updating package lists
+            Updating CIS Audit package list
             Installing CIS Audit packages
             CIS Audit enabled
             Visit https://ubuntu.com/security/cis to learn how to use CIS
@@ -509,7 +509,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         Then I will see the following on stdout:
             """
             One moment, checking your subscription first
-            Updating package lists
+            Updating Ubuntu Security Guide package list
             Ubuntu Security Guide enabled
             Visit https://ubuntu.com/security/certifications/docs/usg for the next steps
             """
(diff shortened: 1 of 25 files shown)


aciba90:
The failing test in https://github.com/canonical/ubuntu-pro-client/actions/runs/6156502630/job/16705466921?pr=2732 is independent of this PR and is being addressed in https://github.com/canonical/ubuntu-pro-client/pull/2735.

uaclient/entitlements/repo.py
```diff
@@ -256,6 +256,13 @@ def install_packages(
         if not util.handle_message_operations(msg_ops):
             return
 
+        try:
+            self._update_main_repo()
```
aciba90, line 260: This update is required because the following:

The installed packages might depend on packages from the main archives, as for example, on a bionic instance that was not `apt-update`ed yet if one run:

```
pro attach <token> 
pro enable cis
```

then the pro client will install `usg-common` that depends on `libopenscap8` which comes from `-updates` and might not be yet on the apt cache.

orndorffgrant (approved):
This turned out to be pretty straightforward, and the performance win is fantastic!
I'm approving, but @renanrodrigo at least should also take a look in case he's aware of some `apt` oddity with this feature.

features/attached_enable.feature
```diff
@@ -634,7 +637,7 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         """
         One moment, checking your subscription first
         Installing snapd
-        Updating package lists
+        Updating Livepatch package lists
```
lucasmoura, line 640: Looking at the snap module, should this be `Updating main package lists` ?

lucasmoura (approved):
This a great refactor @aciba90. I just have one nit regarding Livepatch, but otherwise LGTM.

features/attached_enable.feature
aciba90: Let me have a look and fix it.

aciba90: Solved.

aciba90:
Thanks, @lucasmoura and @orndorffgrant, for the reviews and comments. I have rebased to fix the merge conflicts and fixed the integration test.

uaclient/snap.py
```diff
@@ -121,12 +121,12 @@ def get_installed_snaps() -> List[SnapPackage]:
 
 
 def install_snapd():
-    event.info(messages.APT_UPDATING_LISTS)
+    event.info(messages.APT_UPDATING_LIST.format("main"))
     try:
-        apt.run_apt_update_command()
+        apt.update_sources_list("/etc/apt/sources.list")
```
renanrodrigo, line 126: This could go wrong if someone has a snapd candidate from another repository (or if they don't really maintain the ubuntu repos in sources.list but I think that is too odd to support). However, if they have a different candidate, it would be even weirder that it would be not installed. So I think although we can't guarantee this update will help, it's fine for  99.9% of the cases.

uaclient/entitlements/base.py
```diff
@@ -296,6 +296,7 @@ def __init__(
             self.extra_args = []
         self._called_name = called_name
         self._valid_service = None  # type: Optional[bool]
+        self._is_main_repo_updated = False
```
renanrodrigo, line 299: `main` has a different meaning when speaking on repositories, and updating the whole sources.list will give you more than `main`. A nit, I know, but I'd call this something like `_is_sources_list_updated`  to better reflect what it means.

uaclient/apt.py
```diff
@@ -340,6 +341,27 @@ def run_apt_update_command(
     return out
 
 
+def update_sources_list(sources_list: str):
+    retry_sleeps = APT_RETRIES.copy()
+    try:
+        while True:
```
renanrodrigo, line 347: Could we use the `@util.retry` decorator to simplify the logic here?

uaclient/apt.py
```diff
@@ -340,6 +341,27 @@ def run_apt_update_command(
...
+            try:
+                apt.cache.Cache().update(sources_list=sources_list)
```
renanrodrigo, line 349: I'd suggest getting the cache from `uaclient.apt.get_apt_cache` for consistency and safety - that'll help on the tests I am working on right now

uaclient/entitlements/base.py
```diff
@@ -540,8 +541,7 @@ def handle_required_packages(self) -> bool:
         if not required_packages:
             return True
 
-        event.info(messages.APT_UPDATING_LISTS)
-        apt.run_apt_update_command()
+        self._update_main_repo()
```
renanrodrigo, line 544: same `main` renaming consideration

features/attached_status.feature
```diff
@@ -58,7 +58,8 @@ Feature: Attached status
         Then stdout matches regexp:
         """
         One moment, checking your subscription first
-        Updating package lists
+        Updating CIS Audit package list
+        Updating main package list
```
renanrodrigo, line 62: As I said in another comment, I would not use `main` here as it indicates we are only updating the `main` pocket - I'd say something different, like `standard Ubuntu package lists` or something like this. In this case it will be more than one package list btw

renanrodrigo (commented):
Overall ~will bring me trouble~ looks great and works fine - thanks!
I have a couple considerations:

uaclient/entitlements/base.py
aciba90: Updated.

uaclient/apt.py
aciba90: Thanks for pointing this out, refactored.

uaclient/apt.py
aciba90: Done.

uaclient/entitlements/base.py
aciba90: Renamed.

features/attached_status.feature
aciba90: Renamed.

uaclient/snap.py
aciba90: I lean more towards keeping the 99% safeness and favor boot time. But, if you think it is more appropriate to be 100% and do a full update while enabling this service, please let me know and I will revert it.

What do @lucasmoura and @orndorffgrant think?

aciba90:
Thanks, @renanrodrigo, for the review and suggestions! I have applied them and left open for discussion the one related to full update or not while enabling Livepatch / installing snapd.

uaclient/snap.py
renanrodrigo: nah let's keep this as-is: this was more of a point-of-attention comment than a request for change. This is good.

uaclient/apt.py
```diff
@@ -340,6 +340,21 @@ def run_apt_update_command(
...
 
 
+def update_sources_list(sources_list: str):
+    cache = get_apt_cache()
+    try:
+        util.retry(
+            (apt.cache.LockFailedException, apt.cache.FetchFailedException),
+            APT_RETRIES,
+        )(cache.update)(sources_list=sources_list)
+    except apt.cache.LockFailedException:
```
renanrodrigo, line 350: For readability's sake, I suggest that we can wrap the outer function instead.
Feel free to disagree though, works well as is.

```suggestion
@util.retry(
    (exceptions.APTProcessConflictError, exceptions.APTUpdateFailed),
    APT_RETRIES,
)
def update_sources_list(sources_list: str):
    cache = get_apt_cache()
    try:
        cache.update(sources_list=sources_list)
    except apt.cache.LockFailedException:
```
`(...)`

renanrodrigo (approved):
LGTM 
last nit inline, but I think we good

renanrodrigo:
(CI errors completely unrelated btw)

uaclient/apt.py
aciba90: Addressed, thanks.

aciba90:
Nit addressed and CI green. Thanks @renanrodrigo!

renanrodrigo:
@aciba90 we all thank you for all the effort here (:

PR #2734, opened 2023-09-07
wip: organize messages

orndorffgrant:
Early wip - based on #2723 

## Why is this needed?

## Checklist

 - [ ] Yes
 - [ ] No


lucasmoura (approved):
That's a great refactor, maybe we can create a different module for all named messages, but this is just a suggestion and shouldn't block this PR at all

renanrodrigo (approved):
LGTM

+1 on creating separate modules for each of those sections in the future, for the sake of organization

orndorffgrant:
CI fail is unrelated.

Good idea on actually breaking the messages into separate modules. We'll merge this PR as-is for now to unblock others and we can tackle that more formal structuring in the future.

PR #2741, opened 2023-09-18
Change airgapping descriptions

s-makin:
## Why is this needed?

These descriptions were written/updated back when Grant reviewed PR #2705 based on the
suggestions he made. However, it seems I never actually pushed them at the time.

### DIFF
--- a/docs/explanations/using_pro_offline.rst
+++ b/docs/explanations/using_pro_offline.rst
@@ -1,61 +1,77 @@
-Using Ubuntu Pro in an offline environment
-******************************************
+Using Ubuntu Pro airgapped
+**************************
 
 Airgapping (or air gapping) is a security measure that involves isolating a
 machine or network from external/unsecured networks to protect sensitive data.
-Airgapped systems usually operate “internetless”, or with limited network
-connectivity. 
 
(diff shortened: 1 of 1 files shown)


docs/explanations/using_pro_offline.rst
```diff
@@ -1,61 +1,77 @@
...
 ----------------------------------
 
-The `Snap-store-proxy`_ is the equivalent tool for hosting the snap packages that
+The Snap store proxy is the equivalent tool for hosting the snap packages that
 your organisation is consuming.
+
+`Snap-store-proxy`_ is a snap that provides an edge proxy to the Snap store, so
+that after your device has been registered with the proxy, all communications
+with the Snap store will be sent via the proxy. This is part of the "firewalled"
+deployment.
```
orndorffgrant, line 62: Snap store proxy does have a true airgapped mode that we can link to https://docs.ubuntu.com/snap-store-proxy/en/airgap

orndorffgrant (approved):
LGTM! just a small suggestion

docs/explanations/using_pro_offline.rst
s-makin: Nice find! Added a reference to it

PR #2742, opened 2023-09-19
Api fix execute

lucasmoura:
## Why is this needed?
Now that the fix plan endpoint is ready, we can now create the fix execute endpoint that indeed applies the fixes on the user machine. This PR add support for both CVEs and USNs  execute API endpoints

## Test Steps
Run the new integration test for the API

## Checklist

 - [ ] Yes
 - [x] No


uaclient/api/u/pro/security/fix/common/__init__.py
```diff
@@ -0,0 +1,47 @@
+from typing import Optional
```
orndorffgrant, line 1: A nit but I suggest changing the folder name from `common` to `_common` if that's allowed, to signal that it is for internal use only

features/api_fix_execute.feature
```diff
@@ -0,0 +1,273 @@
...
+    @series.lts
+    @uses.config.machine_type.lxd-container
+    Scenario Outline: Fix execute command on invalid CVEs/USNs
+        Given a `<release>` machine with ubuntu-advantage-tools installed
+        When I run `pro api u.pro.security.fix.cve.execute.v1 --data '{"cves": ["CVE-1800-123456"]}'` as non-root
+        Then stdout is a json matching the `api_response` schema
+        And the json API response data matches the `cve_fix_execute` schema
+        And stdout matches regexp:
+        """
+        {"_schema_version": "v1", "data": {"attributes": {"cves_data": {"cves": \[{"description": null, "errors": \[{"error_type": "security-fix-not-found-issue", "failed_upgrades": null, "reason": "Error: CVE-1800-123456 not found."}\], "status": "error", "title": "CVE-1800-123456", "upgraded_packages": \[\]}\], "status": "error"}}, "meta": {"environment_vars": \[\]}, "type": "CVEFixExecute"}, "errors": \[\], "result": "success", "version": ".*", "warnings": \[\]}
```
orndorffgrant, line 12: Not for this PR, but I think we need to come up with a better way of testing what we care about in json responses. These tests are hard to read and when something goes wrong, it's hard to tell which part went wrong.

orndorffgrant (approved):
This looks good to me! Can we delete `uaclient/security.py` yet or is it still used somewhere?

uaclient/api/u/pro/security/fix/common/__init__.py
lucasmoura: I though about this, but once we move the plan logic there as well, we will import some `DataObjects` on the cli module. So it not something that will be only used internally on the API. But if you think the rename is still worth it, I can do it. No problem

features/api_fix_execute.feature
lucasmoura: Totally agree

lucasmoura:
@orndorffgrant we cannot delete the security module right now, because we still need to refactor the unit tests tied to it. It is a larger change so I am working on this on a separate PR

renanrodrigo (approved):
very small nits but overall LGTM
Thanks!

lucasmoura:

PR #2745, opened 2023-09-21
[draft] feat: warn/confirm with user if enabling fips downgrades the kernel

catmsred:
This change adds a new message prompting the user to confirm if enabling fips downgrades the kernel.

## Why is this needed?
There have been reports of people enabling FIPS (not realizing it will downgrade their kernel), and having their hardware fail as it needed the new kernel.  This change adds an additional warning/prompt if a FIPS enablement is determined to downgrade the kernel.

Jira card: https://warthogs.atlassian.net/browse/SC-1534

## Test Steps
TODO: create integration tests

## Checklist

 - [ ] Yes
 - [x] No


catmsred:
Will re-open from an upstream branch when this is good to go

PR #2749, opened 2023-09-22
i18n (SC-1517)

orndorffgrant:
## Why is this needed?

This PR adds translation support and delivers translations in a new separate package that is "Recommend"ed by ubuntu-advantage-tools.

## Test Steps
Set up a build chroot with the version of pkgbinarymangler that has ubuntu-pro-client-l10n in striptranslations.blacklist (there might be a better way to do this?)
```
sudo sbuild-launchpad-chroot create -s xenial -a amd64 -n pro-test-xenial-amd64

sudo schroot -c source:pro-test-xenial-amd64 -d / -u root -- sh -c 'echo "deb http://ppa.launchpad.net/orndorffgrant/pkgbinarymangler-ubuntu-pro-client-l10n/ubuntu xenial main" >> /etc/apt/sources.list.d/testppa.list'
sudo schroot -c source:pro-test-xenial-amd64 -d / -u root -- sh -c 'apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 74E59E9E'

sudo schroot -c source:pro-test-xenial-amd64 -d / -u root -- sh -c 'apt update'
sudo schroot -c source:pro-test-xenial-amd64 -d / -u root -- sh -c 'apt install pkgbinarymangler'

./tools/build.sh --chroot pro-test-xenial-amd64 xenial
./tools/test-in-lxd.sh xenial
LANGUAGE=pt_BR.UTF-8 pro security-status
```

You should see one line that is translated.

## Checklist

 - [ ] Yes
 - [x] No


lib/upgrade_lts_contract.py
```diff
@@ -13,6 +14,8 @@
 
 if __name__ == "__main__":
     setup_logging(logging.DEBUG)
+    gettext.bindtextdomain("ubuntu-pro", "/usr/share/locale")
```
lucasmoura, line 17: Why do we need that call ? Aren't we using the default path location to store the translations ?

orndorffgrant: The path is the default, but all examples of python apps in ubuntu that I've looked at pass this value to this call anyway, so I'm following their example.

(Note this has changed to a slightly different call at the top of `messages/__init__.py` but still hardcodes this path)

renanrodrigo (approved):
This works fine - thanks again @orndorffgrant for the work here and in all previous preparation PRs

PR #2750, opened 2023-09-22
Bring changes from release 29.4

lucasmoura:


features/anbox.feature
```diff
@@ -6,16 +6,46 @@ Feature: Enable anbox on Ubuntu
...
+        Skipping installing packages
+        Anbox Cloud access enabled
+        """
+        When I run `pro status` as non-root
+        Then stdout matches regexp:
+        """
+        anbox-cloud +yes +enabled
+        """
+        When I run `apt-cache policy` with sudo
+        Then apt-cache policy for the following url has permission `500`
```
renanrodrigo, line 40: This needs to be updated at some point in time, as the step in main calls it `priority`, as it should.
Should this be done in this PR or right after it lands?

features/attached_enable.feature
```diff
@@ -233,20 +233,16 @@ Feature: Enable command behaviour when attached to an Ubuntu Pro subscription
         See: sudo pro status
         """
         When I run `apt-cache policy` with sudo
-        Then apt-cache policy for the following url has priority `510`
```
renanrodrigo, line 236: same here, priority is being set back to permission but the step stays correct

lucasmoura:
@renanrodrigo updated

orndorffgrant (approved):

renanrodrigo (approved):

PR #2763, opened 2023-09-29
Purge option for RepoEntitlements

renanrodrigo:
This is intended to land after #2744 - it contains the commits from there and will be rebased on whatever lands on that side.

still missing: integration tests

## Why is this needed?
This PR solves all of our problems because it implements the purge option when disabling services involving apt repositories. This removes/downgrades all the packages from that specific source.

## Test Steps
Integration tests still to come - but generally run disable for all services with `--purge` and see what happens :D

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -453,6 +453,24 @@ def get_installed_packages_by_origin(origin: str) -> List[apt_pkg.Package]:
     return list(result)
 
 
+def get_alternative_versions_for_package(
+    package: apt_pkg.Package, exclude_origin: Optional[str] = None
+) -> List[apt_pkg.Version]:
+    alternative_versions = [
+        version
+        for version in package.version_list
+        if version != package.current_ver
+    ]
+    if exclude_origin:
+        for version in alternative_versions:
+            for file, _ in version.file_list:
+                if file.origin == exclude_origin:
+                    alternative_versions.remove(version)
+                    break
+
+    return alternative_versions
+
+
 def add_auth_apt_repo(
     repo_filename: str,
     repo_url: str,
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -438,6 +438,21 @@ def run_apt_install_command(
     return out
 
 
+def get_installed_packages_by_origin(origin: str) -> List[apt_pkg.Package]:
+    # Avoiding duplicate entries, which may happen due to version being in
+    # multiple pockets or supporting multiple architectures.
+    result = set()
+    with PreserveAptCfg(get_apt_pkg_cache) as cache:
+        for package in cache.packages:
+            installed_version = package.current_ver
+            if installed_version:
+                for file, _ in installed_version.file_list:
+                    if file.origin == origin:
+                        result.add(package)
+
+    return list(result)
+
+
 def add_auth_apt_repo(
     repo_filename: str,
     repo_url: str,
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -281,6 +281,7 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         allow_beta: bool = False,
         called_name: str = "",
         access_only: bool = False,
+        purge: bool = False,
         extra_args: Optional[List[str]] = None,
     ) -> None:
         """Setup UAEntitlement instance
@@ -293,6 +294,7 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         self.assume_yes = assume_yes
         self.allow_beta = allow_beta
         self.access_only = access_only
+        self.purge = purge
         if extra_args is not None:
             self.extra_args = extra_args
         else:
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -95,10 +95,90 @@ class RepoEntitlement(base.UAEntitlement):
         return True
 
     def _perform_disable(self, silent=False):
+        if self.purge and self.origin:
+            repo_origin_packages = apt.get_installed_packages_by_origin(
+                self.origin
+            )
+            packages_to_reinstall = []
+            packages_to_remove = []
+            for package in repo_origin_packages:
+                alternatives = apt.get_alternative_versions_for_package(
+                    package, exclude_origin=self.origin
+                )
+                if alternatives:
+                    # We can call max(List[Version]) but mypy doesn't know.
+                    # Or doesn't like it.
+                    packages_to_reinstall.append(
+                        (package, max(alternatives))  # type: ignore
+                    )
+                else:
+                    packages_to_remove.append(package)
+
+            if not self.prompt_for_purge(
+                packages_to_remove, packages_to_reinstall
+            ):
+                return False
+
         if hasattr(self, "remove_packages"):
             self.remove_packages()
         self.remove_apt_config(silent=silent)
+
+        if self.purge and self.origin:
+            self.execute_purge(packages_to_remove, packages_to_reinstall)
         return True
+
+    def prompt_for_purge(self, packages_to_remove, packages_to_reinstall):
+        prompt = False
+        if packages_to_remove:
+            print(messages.WARN_PACKAGES_REMOVAL)
+            util.print_package_list(
+                [package.name for package in packages_to_remove]
+            )
+            prompt = True
+
+        if packages_to_reinstall:
+            print(messages.WARN_PACKAGES_REINSTALL)
+            util.print_package_list(
+                [package.name for (package, _) in packages_to_reinstall]
+            )
+            prompt = True
+
+        if prompt:
+            return util.prompt_for_confirmation(messages.PROCEED_YES_NO)
+        return True
+
+    def execute_purge(self, packages_to_remove, packages_to_reinstall):
+        # We need to check for package.current_ver again, because there is an
+        # intermediate step between listing the packages and acting on them.
+        # Packages may be removed between those operations.
+        to_remove = [
+            package.name
+            for package in packages_to_remove
+            if package.current_ver
+        ]
+        if to_remove:
+            apt.remove_packages(
+                to_remove,
+                messages.UNINSTALLING_PACKAGES_FAILED.format(
+                    packages=to_remove
+                ),
+            )
+
+        to_reinstall = [
+            "{}={}".format(package.name, version.ver_str)
+            for (package, version) in packages_to_reinstall
+            if package.current_ver
+        ]
+        if to_reinstall:
+            apt.run_apt_install_command(
+                to_reinstall,
+                apt_options=[
+                    "--allow-downgrades",
+                    '-o Dpkg::Options::="--force-confdef"',
+                    '-o Dpkg::Options::="--force-confold"',
+                ],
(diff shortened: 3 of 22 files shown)


uaclient/entitlements/base.py
```diff
@@ -281,6 +281,7 @@ def __init__(
         allow_beta: bool = False,
         called_name: str = "",
         access_only: bool = False,
+        purge: bool = False,
```
lucasmoura, line 284: Can't this be an attribute we use when calling the `disable` method ?

uaclient/entitlements/repo.py
```diff
@@ -95,11 +95,90 @@ def _perform_enable(self, silent: bool = False) -> bool:
         return True
 
     def _perform_disable(self, silent=False):
+        if self.purge:
```
lucasmoura, line 120: maybe we can add `self.purge and self.origin` instead ?

lucasmoura (commented):
I still need to review this more carefully, but these is the initial look

uaclient/entitlements/repo.py
```diff
@@ -95,11 +95,90 @@ def _perform_enable(self, silent: bool = False) -> bool:
...
+
+        if prompt:
+            return util.prompt_for_confirmation(messages.PROCEED_YES_NO)
+        return True
+
+    def execute_purge(self, packages_to_remove, packages_to_reinstall):
+        to_remove = [
+            package.name
+            for package in packages_to_remove
+            if package.current_ver
```
lucasmoura, line 178: Can't we guarantee the the `package.current_ver` is always set when creating the `packages_to_remove` list ?

uaclient/entitlements/repo.py
```diff
@@ -95,11 +95,90 @@ def _perform_enable(self, silent: bool = False) -> bool:
...
+        ]
+        if to_remove:
+            apt.remove_packages(
+                [package.name for package in packages_to_remove],
```
lucasmoura, line 182: I think we can use `to_remove`  here

uaclient/entitlements/repo.py
```diff
@@ -95,11 +95,90 @@ def _perform_enable(self, silent: bool = False) -> bool:
...
+                messages.UNINSTALLING_PACKAGES_FAILED.format(
+                    packages=to_remove
+                ),
+            )
+
+        to_reinstall = [
+            "{}={}".format(package.name, version.ver_str)
+            for (package, version) in packages_to_reinstall
+            if package.current_ver
```
lucasmoura, line 191: Same `current_ver` consideration here

uaclient/util.py
```diff
@@ -458,3 +459,20 @@ def deduplicate_arches(arches: List[str]) -> List[str]:
 
 def we_are_currently_root() -> bool:
     return os.getuid() == 0
+
+
+def print_package_list(
+    package_list: List[str],
+):
+    package_names = " ".join(package_list)
```
lucasmoura, line 467: Just a nit, but I don't think we need this variable here, we can the operation directly on the print call

uaclient/apt.py
```diff
@@ -453,6 +453,24 @@ def get_installed_packages_by_origin(origin: str) -> List[apt_pkg.Package]:
...
+        version
+        for version in package.version_list
+        if version != package.current_ver
+    ]
+    if exclude_origin:
+        for version in alternative_versions:
+            for file, _ in version.file_list:
+                if file.origin == exclude_origin:
+                    alternative_versions.remove(version)
+                    break
```
lucasmoura, line 469: Can't we already perform this logic when creating the `alternative_versions` variable ?

lucasmoura (commented):
Also, how should we treat reboot here ? We will always say to the user that we need a reboot after running the purge operation or should we check after each step to see if a package needs a reboot to list to the user which ones are causing the reboot operation

uaclient/entitlements/base.py
renanrodrigo: I guess we could - I am setting it in the class to mimic what was done with `assume_yes` and `access_only`, so for consistency I defined it similarly

uaclient/entitlements/repo.py
renanrodrigo: We do the can_disable check, but yes - it doesn't hurt - can do.

uaclient/entitlements/repo.py
renanrodrigo: yes we can, because we only pass installed packages to the function which returns those. However, some entitlements have this `packages_to_remove` directive/variable, and then some of those packages can be removed in between the purge-driven selection of packages and the actual purge - `ubuntu-fips` is a good example when purging FIPS.

uaclient/apt.py
renanrodrigo: Well, we could, but sincerely I don't see how that would simplify the logic. Do you have a suggestion?

uaclient/entitlements/base.py
orndorffgrant: The others were added as instance fields because they were needed in other methods where we couldn't easily pass them from the enable function otherwise.
I don't think this option is needed in any calls outside `disable`, but also I don't think it hurts to follow the patterns of the others, and we may need it somewhere else in the future

uaclient/entitlements/repo.py
orndorffgrant: I think that nuance is worth adding here as a comment, otherwise it may be tempting to some future developer to remove the check

uaclient/messages/__init__.py
```diff
@@ -1772,6 +1772,13 @@ def __repr__(self):
     + ENABLE_FAILED,
 )
 
+E_REPO_PURGE_FAIL_NO_ORIGIN = FormattedNamedMessage(
+    "repo-purge-fail-no-origin",
+    "Cannot disable {entitlement_name} with purge: no origin value defined"
+    + "\n"
+    + DISABLE_FAILED_TMPL,
+)
```
orndorffgrant, line 1799: This isn't an error per se since it isn't used in an exception. Other CanDisable messages are in the "Named Messages" section above the error section

uaclient/entitlements/repo.py
```diff
@@ -72,6 +76,28 @@ def _check_for_reboot(self) -> bool:
...
+        if not self.origin and self.purge:
+            return (
+                False,
+                CanDisableFailure(
+                    CanDisableFailureReason.NO_PURGE_WITHOUT_ORIGIN,
+                    messages.E_REPO_PURGE_FAIL_NO_ORIGIN.format(
+                        entitlement_name=self.title, title=self.title
+                    ),
+                ),
+            )
```
orndorffgrant, line 99: This is an error message for a bug if we, the developers, introduce a repo-entitlement without an origin. I wish we could structure the code in such a way that this is so unlikely as to not deserve a user-facing error message. But this is probably the right thing to do, I'm just complaining

uaclient/messages/__init__.py
```diff
@@ -272,6 +272,24 @@ class TxtColor:
 APT_REMOVING_SOURCE_FILE = "Removing apt source file: {filename}"
 APT_REMOVING_PREFERENCES_FILE = "Removing apt preferences file: {filename}"
 
+# Kernel checks for Purge
+PURGE_KERNEL_REMOVAL = (
+    "Purging the fips packages would uninstall the following kernel(s):"
```
orndorffgrant, line 277: I think this message could happen for disabling realtime-kernel with --purge as well

uaclient/system.py
```diff
@@ -178,6 +179,38 @@ def get_kernel_info() -> KernelInfo:
         )
 
 
+# This only works if we are root (because of permissions to 'file' the kernels
+# in /boot), but can't assert_root here due to circular imports.
```
orndorffgrant, line 183: I understand we can't use the decorator, but we could do
```
if not util.we_are_currently_root():
    raise Exception()
```

uaclient/entitlements/base.py
renanrodrigo: we need this to avoid the pre-disablement message on FIPS at least - so I think following the pattern is the way to go

uaclient/system.py
renanrodrigo: well well we don't reeeeally need the exception - when executed as non-root this returns an empty list. Do you think it is worth failing anyway?

uaclient/messages/__init__.py
renanrodrigo: true will make the service a variable

uaclient/messages/__init__.py
renanrodrigo: will move

uaclient/messages/__init__.py
```diff
@@ -119,6 +120,11 @@ class TxtColor:
     "Cannot install package {package} version {version}"
 )
 
+WARN_PACKAGES_REMOVAL = "The following package(s) will be REMOVED:"
+WARN_PACKAGES_REINSTALL = (
+    "The following packages will be reinstalled from the archive:"
+)
```
orndorffgrant, line 126: Either both of these should use "(s)" or they should be pluralized

renanrodrigo: (s) for the easy go

uaclient/entitlements/repo.py
```diff
@@ -95,11 +123,131 @@ def _perform_enable(self, silent: bool = False) -> bool:
...
+            print(messages.PURGE_KERNEL_REMOVAL.format(service=self.title))
+            print(" ".join(linux_image_versions))
+
+            current_kernel = system.get_kernel_info().uname_release
+            if current_kernel in linux_image_versions:
+                print(
+                    messages.PURGE_CURRENT_KERNEL.format(
+                        kernel_version=current_kernel
+                    )
+                )
```
orndorffgrant, line 178: Let's always print this because the user will always want to know
```suggestion
            print(
                messages.PURGE_CURRENT_KERNEL.format(
                    kernel_version=current_kernel
                )
            )
```

uaclient/system.py
renanrodrigo: decided this should be a `RuntimeError`

renanrodrigo:
Also we can add a warning telling this feature is experimental. We have many warnings, but this one should appear everywhere regardless.

lucasmoura (approved):
I have tested this feature and it is working as expected. Just have a tiny nit, but overall LGTM

uaclient/entitlements/repo.py
```diff
@@ -124,6 +124,9 @@ def _perform_enable(self, silent: bool = False) -> bool:
 
     def _perform_disable(self, silent=False):
         if self.purge and self.origin:
+            print(messages.PURGE_EXPERIMENTAL)
```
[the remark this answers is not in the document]
renanrodrigo: I think this is not relevant because purge should never be in APIs (unless it becomes so safe in the future right)
So self.purge will always be false on API calls.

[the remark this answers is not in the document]
lucasmoura: Fair enough, this path should never be reached through the API

orndorffgrant (approved):

renanrodrigo:
Rebased.
The failing tests are related to fips-preview on focal mostly - those can be sorted out before release verification, as the behavior will change based on contract server changes.

PR #2769, opened 2023-10-03
fix: update message for cloud advertisement

lucasmoura:
## Why is this needed?
The attribute of a message for Ubuntu Pro cloud advertisement need to be update, since it is no longer matching the definition on the messages module

## Test Steps
Guarantee that the message attributes are being correctly used

## Checklist

 - [ ] Yes
 - [x] No


orndorffgrant (approved):
@lucasmoura just a formatting error to fix but then go ahead and merge

PR #2773, opened 2023-10-04
chore: archive release-29 sru scripts

renanrodrigo:
also create a folder for release-30 and let the esm pinning test rest there


lucasmoura:
@renanrodrigo I think we should move the esm pin test to the release-30 folder instead

lucasmoura (approved):

PR #2779, opened 2023-10-06
logging: use journald logging for all systemd services

orndorffgrant:
## Why is this needed?

This PR brings journald logging to all systemd services, plus a few small related refactors.

## Test Steps

```
./tools/test-in-lxd.sh xenial
apt update
# ensure logs of esm_cache end up in journald
```

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/unattached_commands.feature
+++ b/features/unattached_commands.feature
@@ -340,19 +340,18 @@ Feature: Command behaviour when unattached
       esm-cache\.service
       """
     When I run `journalctl -o cat -u esm-cache.service` with sudo
-    Then stdout does not match regexp:
+    Then stdout does not contain substring:
       """
-      raise FetchFailedException\(\)
+      raise FetchFailedException()
       """
-    When I run `ls /var/crash/` with sudo
-    Then stdout does not match regexp:
-      """
-      _usr_lib_ubuntu-advantage_esm_cache
-      """
-    When I run `cat /var/log/ubuntu-advantage.log` with sudo
     Then stdout matches regexp:
       """
-      Failed to fetch ESM Apt Cache item:
+      "WARNING", "ubuntupro.apt", "fail", \d+, "Failed to fetch ESM Apt Cache item: https://esm.ubuntu.com/apps/ubuntu/dists/<release>-apps-security/InRelease", {}]
+      """
+    When I run `ls /var/crash/` with sudo
+    Then stdout does not contain substring:
+      """
+      _usr_lib_ubuntu-advantage_esm_cache
       """
 
     Examples: ubuntu release
--- a/uaclient/cli/__init__.py
+++ b/uaclient/cli/__init__.py
@@ -3,7 +3,6 @@
 import argparse
 import json
 import logging
-import pathlib
 import sys
 import tarfile
 import tempfile
@@ -24,9 +23,10 @@ from uaclient import (
     exceptions,
     http,
     lock,
+    log,
+    messages,
+    security_status,
 )
-from uaclient import log as pro_log
-from uaclient import messages, security_status
 from uaclient import status as ua_status
 from uaclient import timer, util, version
 from uaclient.api.api import call_api
@@ -63,7 +63,7 @@ from uaclient.entitlements.entitlement_status import (
 )
 from uaclient.files import notices, state_files
 from uaclient.files.notices import Notice
-from uaclient.log import JsonArrayFormatter, get_user_or_root_log_file_path
+from uaclient.log import get_user_or_root_log_file_path
 from uaclient.timer.update_messaging import refresh_motd, update_motd_messages
 from uaclient.yaml import safe_dump, safe_load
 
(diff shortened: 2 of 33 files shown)


aciba90 (approved):
LGTM, thanks for the refactors.

lucasmoura:
@orndorffgrant should we use our JSON format when logging to `journalctl` ? I am just wondering if we need to be consistent with the other logs there. But maybe this is not needed

orndorffgrant:
@lucasmoura I think it should use the json logging to be consistent. The `setup_journald_logging` function does use the JsonArrayFormatter and I just updated the integration test to actually check the json output. Does that align with what you were thinking?

lucasmoura:
My rationale is that it shouldn't to be honest, just  to align with the other ` journalctl` entries. But I don't think is an actual problem if we keep the JSON output

orndorffgrant:
@sergiodj @athos-ribeiro @lucasmoura After our discussion in Riga, I've updated this to remove the json formatting in journald logs.

We still want to include multiple fields with structure, though, so I've introduced a simpler (but more fragile) `|`-delimited format.

Taking the example from my last comment here, it would now look like this:
```
...
"_CMDLINE":"/usr/bin/python3 /usr/lib/ubuntu-advantage/esm_cache.py",
"__CURSOR":"s=e4ea46b1825d4b5d833f9f8a48dccd3c;i=862;b=28db989000224dc39694c489bad64176;m=40a4ff859d;t=608138598a742;x=35915d735a50dc59",
"MESSAGE":"WARNING|ubuntupro.apt|fail|872|Failed to fetch ESM Apt Cache item: https://esm.ubuntu.com/apps/ubuntu/dists/xenial-apps-security/InRelease|{}",
"SYSLOG_IDENTIFIER":"python3"
...
```

Do you think this is better or worse than the json formatting?

athos-ribeiro:
> @sergiodj @athos-ribeiro @lucasmoura After our discussion in Riga, I've updated this to remove the json formatting in journald logs.
> 
> We still want to include multiple fields with structure, though, so I've introduced a simpler (but more fragile) `|`-delimited format.
> 
> Taking the example from my last comment here, it would now look like this:
> 
> ```
> ...
> "_CMDLINE":"/usr/bin/python3 /usr/lib/ubuntu-advantage/esm_cache.py",
> "__CURSOR":"s=e4ea46b1825d4b5d833f9f8a48dccd3c;i=862;b=28db989000224dc39694c489bad64176;m=40a4ff859d;t=608138598a742;x=35915d735a50dc59",
> "MESSAGE":"WARNING|ubuntupro.apt|fail|872|Failed to fetch ESM Apt Cache item: https://esm.ubuntu.com/apps/ubuntu/dists/xenial-apps-security/InRelease|{}",
> "SYSLOG_IDENTIFIER":"python3"
> ...
> ```
> 
> Do you think this is better or worse than the json formatting?

Thanks, Grant!

Would you mind showing what an old (current) log entry would look like compared to the new format?
Do you think there are people out there parsing those logs somehow and would this break their workflows?

orndorffgrant:
Thanks for looking @athos-ribeiro!

Prior to this PR, most services logged to /var/log/ubuntu-advantage.log in the JSONArray format. This was not done in a multi-process-safe way. Any messages that ended up in journald were accidental and did not contain any structure, e.g. it would just be the message string like "Failed to fetch ESM Apt Cache item: https://esm.ubuntu.com/apps/ubuntu/dists/xenial-apps-security/InRelease".

Two services already logged to journald in the JSONArray format.

So there are several changes here from the perspective of someone interested in our logs:
1. Some logs that used to appear in ubuntu-advantage.log will no longer appear
2. More logs will show up in journald for their respective services
3. logs in journald that used to be JSONArray formatted will no longer be JSONArray formatted
4. logs in journald that used to be just a string will now have some structure and extra information 

We haven't considered backwards compatibility of our logs very seriously in the past. The biggest breaking change, though, would be moving away from JSONArray formatting for the services that were already using it. Anything else would've been handled as a string, so adding structure wouldn't break it significantly IMO.

panlinux:
> Thanks for looking @athos-ribeiro!
> 
> Prior to this PR, most services logged to /var/log/ubuntu-advantage.log in the JSONArray format. This was not done in a multi-process-safe way. Any messages that ended up in journald were accidental and did not contain any structure, e.g. it would just be the message string like "Failed to fetch ESM Apt Cache item: https://esm.ubuntu.com/apps/ubuntu/dists/xenial-apps-security/InRelease".
> 
> Two services already logged to journald in the JSONArray format.
> 
> So there are several changes here from the perspective of someone interested in our logs:
> 
> 1. Some logs that used to appear in ubuntu-advantage.log will no longer appear
> 2. More logs will show up in journald for their respective services
> 3. logs in journald that used to be JSONArray formatted will no longer be JSONArray formatted
> 4. logs in journald that used to be just a string will now have some structure and extra information
> 
> We haven't considered backwards compatibility of our logs very seriously in the past. The biggest breaking change, though, would be moving away from JSONArray formatting for the services that were already using it. Anything else would've been handled as a string, so adding structure wouldn't break it significantly IMO.

This is quite the breaking change. If I, as a user, would be parsing the logs, I would be very annoyed with such a change in an LTS upgrade.
I Have you considered a config option to toggle between these logging formats? On upgrade, the logging format wouldn't change, but on a fresh install, it would be what this PR is proposing?

panlinux:
Other points:
1. You should document the logging. How it's done, in which format, where.
2. Are there more logging changes coming? When will this be done?

features/unattached_commands.feature
```diff
@@ -383,19 +383,18 @@ Feature: Command behaviour when unattached
...
         """
-        raise FetchFailedException\(\)
+        raise FetchFailedException()
         """
-        When I run `ls /var/crash/` with sudo
-        Then stdout does not match regexp:
+        Then stdout matches regexp:
         """
-        _usr_lib_ubuntu-advantage_esm_cache
+        WARNING|ubuntupro.apt|fail|\d+|Failed to fetch ESM Apt Cache item: https://esm.ubuntu.com/apps/ubuntu/dists/<release>-apps-security/InRelease|{}
```
lucasmoura, line 392: Just to double check, why is the function name `fail` here ?

orndorffgrant: Inside `update_esm_caches`, an `EsmAcquireProgress` class is defined with a `fail` method that is called by the `apt` library when an error occurs. That method is where we log this line

orndorffgrant:
@panlinux @lucasmoura I've updated this PR so it only does the following:

- Removes not-multiprocess-safe logging to same file from multiple processes
- Each systemd service that previously did not log to journald now does
  - These services log in JSON format

There are no breaking changes.

Is this okay?

renanrodrigo (approved):
imho this is a good changeset after all the discussion

uaclient/cli/__init__.py
```diff
@@ -886,6 +888,7 @@ def action_disable(args, *, cfg, **kwargs):
     return 0 if ret else 1
 
 
+<<<<<<< HEAD
```
renanrodrigo, line 891: oh no

orndorffgrant: ugh

orndorffgrant: fixed

orndorffgrant:
CI fail unrelated

orndorffgrant:
I'm going ahead and merging the least controversial version of this PR (new logging to journald using JSON format).
We can revisit at SRU review time if necessary.

PR #2780, opened 2023-10-06
Remove security module

lucasmoura:
## Why is this needed?
Now that we have both fix plan and execute API endpoints, we no longer need the security module. This PR moves the functions in that module to the related fix API module while also adapting the test for the security module to the new API use case

## Test Steps
Guarantee that all fix related integration tests are still working as expected

## Checklist

 - [ ] Yes
 - [ ] No


orndorffgrant (commented):
I'm excited for this one! I think it needs some tweaks though. Currently, this PR results in the following 4 related modules:
- `uaclient.api.u.pro.security.fix`
- `uaclient.api.u.pro.security.fix._common`
- `uaclient.api.u.pro.security.fix._common.execute.v1`
- `uaclient.api.u.pro.security.fix._common.plan.v1`

My main comment is that `uaclient.api.u.pro.security.fix` and `uaclient.api.u.pro.security.fix._common` seem redundant - why both?

I think ideally we'd have only
- `uaclient.api.u.pro.security.fix._common.v1`
  - for code that is shared between all fix endpoints
  - if this includes anything we might expect a caller to import, then it should get a `.v1`, if not then it's good as it is IMO
- `uaclient.api.u.pro.security.fix._common.execute.v1`
  - for code shared only between the fix execute endpoints
- `uaclient.api.u.pro.security.fix._common.plan.v1`
  - for code shared only between the fix plan endpoints

Then our documentation should instruct the caller to import everything from e.g. `uaclient.api.u.pro.security.fix.cve.plan.v1` including all the things that are actually defined in one of the `_common` modules. (So this PR needs an accompanying docs PR I think)

We already published docs telling people to import from `uaclient.api.u.pro.security.fix` so that module still needs to exist for backwards compatibility, but it should only look like:
```python
from uaclient.api.u.pro.security.fix._common.plan.v1 import FixStepType
...

__all__ = [
    "FixStepType",
    ...
]
```

WDYT?

lucasmoura:
@orndorffgrant I agree with most of the suggestion and I have already updated the code based on them. However, I have just not created a `v1` for the `_common` module as I don't believe there is any code there that should be tied to a version. But please let me know if you disagree

uaclient/api/u/pro/security/fix/cve/plan/v1.py
```diff
@@ -2,8 +2,20 @@
 
 from uaclient.api.api import APIEndpoint
 from uaclient.api.data_types import AdditionalInfo
-from uaclient.api.u.pro.security.fix import FixPlanResult, fix_plan_cve
 from uaclient.api.u.pro.security.fix._common import get_expected_overall_status
+from uaclient.api.u.pro.security.fix._common.plan.v1 import (  # noqa: F401
```
orndorffgrant, line 6: I think we need a comment here saying why these are imported

orndorffgrant: and similar comments in the other places where we do this

orndorffgrant (approved):
Just a minor nit otherwise LGTM

lucasmoura:
@orndorffgrant updated

renanrodrigo:
@lucasmoura besides not marked in the checklist, I agree with Grant that this PR needs documentation for people to know where to import things from (plan, execute) - maybe a new explanation, but at least updating the API reference (which still misses the execute part)

renanrodrigo (approved):

renanrodrigo:
breaks: missing documentation for fix execute endpoints in the API reference

PR #2781, opened 2023-10-09
docs: update fix plan references

lucasmoura:
We have added new fields for the fix plan API endpoints. We are now reflecting those new fields on the docs


docs/explanations/how_to_interpret_output_of_fix_plan_api.md
```diff
@@ -73,6 +75,8 @@ If we take a look at a CVE object, we will see the following structure:
 * **warnings**: Warnings that can happen when generating the plan. For example, if the CVE doesn't
   have a fix available for a given package, this will be reflected here. We will detail the possible
   warnings in a following section.
+* **additional_data**: Additional data for the CVE being fixed. Currently, this field is only
```
s-makin, line 78: I'm a bit confused about this bit. The "additional data" field doesn't appear in the USN section at the bottom of the page, so if it's only found under the CVEs enpoint but only used by USNs then I'm not sure how that should work? Surely it would never come up? Why aren't Launchpad bugs also reported for the CVEs?

docs/explanations/how_to_interpret_output_of_fix_plan_api.md
```diff
@@ -157,6 +165,44 @@ The JSON representation for this step is:
 The `data` object will state why no operation is needed. This can either be because the CVE doesn't
 affect the system, or because the CVE is already fixed in the machine.
 
+### NoOpAlreadyFixed
+
+This is a subtype of a NoOp step that indicates that the CVE is already
```
s-makin, line 170:
```suggestion
This is a sub-type of a NoOp step that indicates that the CVE is already
```

docs/explanations/how_to_interpret_output_of_fix_plan_api.md
```diff
@@ -157,6 +165,44 @@ The JSON representation for this step is:
...
+  "order": 1
+}
+```
+
+The `data` object will state which packages are already fixed and which
+pocket they come from.
+
+### NoOpLivepatchFixData
+
+This is a subtype of a NoOp step that indicates that the CVE is already
```
s-makin, line 190:
```suggestion
This is a sub-type of a NoOp step that indicates that the CVE is already
```

docs/explanations/how_to_interpret_output_of_fix_plan_api.md
```diff
@@ -157,6 +165,44 @@ The JSON representation for this step is:
...
+fixed by a Livepatch patch. The JSON representation for this step is:
```
s-makin, line 191: I would suggest "fixed by a patch from Livepatch" - "Livepatch patch" sounds really odd

docs/references/api.rst
```diff
@@ -815,7 +828,40 @@ executing any of those steps.
...
+
+             * - Field Name
+               - Type
+               - Description
+             * - ``status``
+               - *str*
+               - The status of the CVE when no operation can be performed
+             * - ``patch_version``
+               - *str*
+               - version of the Livepatch patch that fixed the CVE
```
s-makin, line 862: I'd avoid use of "Livepatch patch" here as well

s-makin, line 862:
```suggestion
               - Version of the Livepatch patch that fixed the CVE
```

s-makin (commented):
Mostly looks good to me (although I think I would include this one in the overall discussion about where the explanation should properly go). Just a couple of nits :)

orndorffgrant:
IMO this one deserves an explanation doc because its result is a set of instructions that need to be interpreted in a specific way to reach a specific goal. The other open PR (about fix execute) returns only the final status which is easier to explain inside the endpoint entry in api.rst.

docs/explanations/how_to_interpret_output_of_fix_plan_api.md
lucasmoura: I agree that is better

lucasmoura:
I agree with @orndorffgrant that we need a dedicated section for this endpoint, as it is not that straightforward as the execute one

lucasmoura:
@s-makin updated

s-makin (approved):
LGTM! Sorry for the delay

PR #2782, opened 2023-10-09
Fix api execute

lucasmoura:
Add docs for the fix execute API endpoints


docs/explanations.rst
```diff
@@ -28,6 +28,7 @@ selection of some of the commands -- what they do, and how they work.
 
     explanations/how_to_interpret_the_security_status_command.md
     explanations/how_to_interpret_output_of_unattended_upgrades.md
+    explanations/how_to_interpret_output_of_fix_execute_api.md
```
s-makin, line 31: I have some thoughts about how this. We now have three of these "interpret the output" pages, which is now enough for a category of its own. However, that approach would mean that eventually we might have an explanation page for almost every API endpoint and I'm not sure that's an efficient use of real estate.

Alternatively, I'm sort of wondering why we don't have this content in the API doc itself? The main reason being that if we duplicate the contents of the output in two places manually, there's big potential for them to drift out of sync with each other over time. I'm all in favour of making things easier for you guys to maintain (i.e., with one place to update instead of multiple).

docs/explanations/how_to_interpret_output_of_fix_execute_api.md
```diff
@@ -0,0 +1,115 @@
+# How to interpret the output of the fix execute API endpoint
+
+In the Pro Client version 29, we introduce two distinct endpoints:
+
+* `u.pro.security.fix.cve.execute.v1`
+* `u.pro.security.fix.usn.execute.v1`
+
+These endpoints can be used to fix a list of [CVEs/USNs](cves_and_usns_explained.md).
+When using the CVE endpoint, the expected output is as follows:
```
s-makin, line 9: Let's put a heading between lines 8 and 9 with something like "Using the CVE endpoint"

docs/explanations.rst
s-makin: Another alternative is that we could consolidate the "fix" content onto a single page (fix plan, fix execute).

docs/explanations/how_to_interpret_output_of_fix_execute_api.md
```diff
@@ -0,0 +1,115 @@
...
+  ]
+}
+```
+
+We can see that the representation has the following fields:
+
+* **error_type**: The error type
+* **reason**: The explanation of why the error happened
+* **failed_upgrade**: A list of objects that always contain the name of the package
+  that was not upgraded and the pocket where the upgraded would come from.
```
s-makin, line 95:
```suggestion
  that was not upgraded and the pocket where the upgrade would have come from.
```

docs/explanations/how_to_interpret_output_of_fix_execute_api.md
```diff
@@ -0,0 +1,115 @@
...
+  is considered **not-affected** if the CVE doesn't affect the system at all.
+* **upgraded_packages**: A list of UpgradedPackage objects referencing each package
+  that was upgraded during the fix operation. The UpgradePackage object always contain
+  the **name** of the package, the **version** it was upgraded to and the **pocket** where
+  the package upgrade came from.
+
+## What errors can be generated?
+
+There some errors that can happen when executing this endpoint. For example, the system
+might require for the user to attach to a Pro subscription to install the upgrades,
```
s-makin, line 72:
```suggestion
might require the user to attach to a Pro subscription to install the upgrades,
```

docs/references/api.rst
```diff
@@ -629,6 +631,345 @@ like ``systemctl start ubuntu-advantage.service``.
...
+                        "error": null
+                    }
+                  ]
+              }
+           }
+
+u.pro.security.fix.usn.execute.v1
+===================================
+
+This endpoint fix the specified USNs on the machine.
```
s-makin, line 797:
```suggestion
This endpoint fixes the specified USNs on the machine.
```

s-makin (commented):
Overall looks good, just a few nits. However, we'll want to decide on what the approach will be about these pages before we add many more of them and create lots of work further down the line. Creating the additional explanation pages seems to be making a lot of extra work for you, so let's see if there's a way to present this sort of content without the additional burden.

docs/explanations.rst
orndorffgrant: I agree that at least for this explanation, all of the content of the explanation can be moved into the field descriptions within api.rst

lucasmoura: I agree, I will move it to the `api.rst`

docs/explanations/how_to_interpret_output_of_fix_execute_api.md
```diff
@@ -0,0 +1,115 @@
...
+    },
+    "type": "CVEFixExecute"
+  },
+  "errors": [],
+  "result": "success",
+  "version": "30",
+  "warnings": []
+}
+```
+
```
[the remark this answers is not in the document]
lucasmoura: Do you think we really need that distinction here ? Isn't it more straightforward to have the explanation just below the JSON representation ?

lucasmoura:
@s-makin @orndorffgrant I have update the docs and moved the explanation into `api.rst`. Let me know what you think

docs/explanations/how_to_interpret_output_of_fix_execute_api.md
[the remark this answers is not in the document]
s-makin: I think my original question is probably redundant now :)

docs/references/api.rst
```diff
@@ -629,6 +631,524 @@ like ``systemctl start ubuntu-advantage.service``.
...
+                    This means that if one USN cannot be fixed, this field will reflect that.
+                    Note that related USNs don't interfere with this attribute, meaning
+                    that a related USN can fail to be fixed without modyfing the **status**
+                    value.
+
+      Each **usn** object contains a reference for both **target_usn** and **related_usns**.
+      The target is the USN requested to be fixed by the user, while related USNs are USNs
+      that are related to the main USN and an attempt to fix them will be performed by the
+      endpoint too. To better understand that distinction, please refer to 
+      [our explanation of CVEs and USNs](cves_and_usns_explained.md).
```
s-makin, line 1104: This link doesn't seem to be working. I think probably because the cves page is now in a different directory.

docs/explanations/how_to_interpret_output_of_fix_execute_api.md
```diff
@@ -0,0 +1,115 @@
...
+We can see that the representation has the following fields:
+
+* **error_type**: The error type
+* **reason**: The explanation of why the error happened
+* **failed_upgrade**: A list of objects that always contain the name of the package
+  that was not upgraded and the pocket where the upgraded would come from.
+
+## What about the USN endpoint?
```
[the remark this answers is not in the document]
s-makin: No longer applicable

s-makin (approved):
Just one broken link, then good to go :) thanks for all your work on this, and nice use of the tabs on the API page!

lucasmoura:
@s-makin fixed

PR #2783, opened 2023-10-10
refactor: simplify prerm to not import python

orndorffgrant:
LP: #2021988

## Why is this needed?

Importing our python module in prerm exposes the packaing script to all the complexity of our python and the risk that a seemingly unrelated change in the python could affect the prerm script.

This happened when we started relying on python3-apt and resulted in LP: #2021988

This PR removes the python import entirely and instead hardcodes the files that need to be removed.

## Test Steps
```
./tools/test-in-lxd.sh xenial
pro attach
apt remove ubuntu-advantage-tools
ls /etc/apt/sources.list.d
# pro sources files should be gone
```

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/debian/ubuntu-advantage-tools.prerm
+++ b/debian/ubuntu-advantage-tools.prerm
@@ -3,12 +3,14 @@
 set -e
 
 remove_apt_files() {
-    /usr/bin/python3 -c '
-from uaclient.apt import clean_apt_files
-
-clean_apt_files()
-'
-
+    # This list should be kept up to date with the list of available apt-repo-based services
+    for service in anbox cc-eal cis esm-apps esm-infra fips fips-preview fips-updates realtime-kernel ros ros-updates; do
+        rm -f /etc/apt/sources.list.d/ubuntu-${service}.list
+    done
+    # preferences are only dynamically created for fips services
+    for fips_service in fips fips-preview fips-updates; do
+        rm -f /etc/apt/preferences.d/ubuntu-${fips_service}
+    done
 }
 
 case "$1" in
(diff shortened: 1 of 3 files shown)


debian/ubuntu-advantage-tools.prerm
```diff
@@ -3,12 +3,14 @@
...
 
 remove_apt_files() {
-    /usr/bin/python3 -c '
-from uaclient.apt import clean_apt_files
-
-clean_apt_files()
-'
-
+    # This list should be kept up to date with the list of available apt-repo-based services
+    for service in anbox cc-eal cis esm-apps esm-infra fips fips-preview fips-updates realtime-kernel ros ros-updates; do
```
lucasmoura, line 7: I think we also need `usg` here

lucasmoura (approved):
LGTM, I just think there is one service missing

debian/ubuntu-advantage-tools.prerm
orndorffgrant: It won't hurt, but is the apt source file named with the `presented_as` name?

renanrodrigo: nope, source-file and key will be `ubuntu-cis.list` unfortunately :clown_face: :upset_cry:

uaclient/apt.py
```diff
@@ -661,43 +660,6 @@ def remove_apt_list_files(repo_url, series):
         system.ensure_file_absent(path)
 
 
-def clean_apt_files(*, _entitlements=None):
```
renanrodrigo, line 664: removing code and keeping functionality? good good

renanrodrigo (approved):
/methinks we can merge this

lucasmoura:
CI errors are unrelated

PR #2785, opened 2023-10-11
tests: fix missing mock in attach unit test

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it fixes a unit test leak in `_post_cli_attach`.

## Test Steps
Unit tests runs

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/cli/tests/test_cli_attach.py
+++ b/uaclient/cli/tests/test_cli_attach.py
@@ -9,6 +9,7 @@ import pytest
 
 from uaclient import event_logger, http, messages, util
 from uaclient.cli import (
+    _post_cli_attach,
     action_attach,
     attach_parser,
     get_parser,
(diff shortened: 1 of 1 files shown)


uaclient/cli/tests/test_cli_attach.py
```diff
@@ -241,8 +241,10 @@ def test_lock_file_exists(
     @mock.patch(M_PATH + "contract.UAContractClient.add_contract_machine")
     @mock.patch("uaclient.actions.status", return_value=("", 0))
     @mock.patch("uaclient.status.format_tabular")
+    @mock.patch(M_PATH + "daemon")
```
lucasmoura, line 244: Would it be better to mock `_post_cli_attach` like we do on the other tests ? The concern is that for some reason we add new functionality to that method, we might forget to mock it again here

renanrodrigo: yes, but in this case another test needs to be added to cover `_post_cli_attach`, as it is being tested as part of this test.
I added this as a quick path, but I can add the extra test if that suits better

lucasmoura: I think that would be better tbh, since we it will be more explicit that we have a dedicated test just for that function

renanrodrigo:
@lucasmoura done there

uaclient/cli/tests/test_cli_attach.py
```diff
@@ -270,46 +288,13 @@ def fake_contract_attach(contract_token, attachment_dt):
...
-        ):
-            with mock.patch.object(
-                cfg, "check_lock_info"
-            ) as m_check_lock_info:
-                m_check_lock_info.return_value = (0, "lock_holder")
-                fake_stdout = io.StringIO()
-                with contextlib.redirect_stdout(fake_stdout):
-                    main_error_handler(action_attach)(args, cfg)
-
-        expected = {
```
lucasmoura, line 303: With this new setup we are no longer testing the `event_logger` part of the command. Should we add an explicit test on for this as well ?

orndorffgrant: We have integration tests that cover json output for attach, so I think this is okay to remove.

orndorffgrant (approved):

lucasmoura (approved):

lucasmoura:

PR #2787, opened 2023-10-12
Add docs explaining disable --purge

renanrodrigo:
This is the documentation for `pro disable <service> --purge`.

### DIFF
--- a/docs/explanations/purging_services.md
+++ b/docs/explanations/purging_services.md
@@ -1,0 +1,80 @@
+# Disabling and purging services
+
+All services enabled using the Pro Client can be disabled using the command line:
+
+```bash
+$ sudo pro disable <service-name>
+```
+
+When you disable a service, the Client will remove all access to that specific service's packages and sources in the system. However, only the access is removed - the packages (debs, snaps) installed when the service was enabled will remain in your system.
+
+For example, the command
+```bash
+$ sudo pro disable esm-apps
+```
+will remove the `sources.list` entry for esm-apps, and also the authentication information that granted access to the packages, but not the debs installed from the esm-apps repository.
+
+On the same line, the command
+```bash
+$ sudo pro disable livepatch
+```
+will disable the Livepatch service, but the `canonical-livepatch` snap will still be installed on the system (although inactive).
+
+If there are other services which depend on the service being disabled, those need to be disabled too. A prompt in the CLI will list those and and ask to disable them. The `--assume-yes` flag can be used to automatically accept this prompt.
+
+## What does `purge` mean in this context?
+
+Disabling and **purging** a service will have a similar effect to what is done by [`ppa-purge`](https://launchpad.net/ppa-purge): it will remove access to the sources from where the service packages can be obtained, and then make a best-effort attempt to downgrade those packages back to their Ubuntu versions (the ones published in the Ubuntu Archive).
+
+If a package is specific to the given service (`ubuntu-fips` for the FIPS related services, for example) then it will be removed during the purge process, when possible.
+
+```{warning}
+The operation of reverting/uninstalling packages may leave the system in an undesired state, due to packages not meeting dependencies (partial or broken system caches) or due to the removal of kernels when alternative kernels are not working properly. Therefore, the `purge` feature is aimed at users/administrators who understand the impacts.
+```
+
+## Using the CLI to disable and purge a service
+
+First of all it is important to note that this is an ***experimental feature***, so although it is available it should be used with care and discretion.
+
+To purge a service, use the CLI command:
+```bash
+$ sudo pro disable <service-name> --purge
+```
+
+We recommend only using purge when you are able to monitor the screen. Since the purge feature has the potential to do serious accidental damage to a system if used unattended (for example, in a script), the --assume-yes flag was made incompatible with the --purge flag.
+
+Because packages are being uninstalled/reinstalled, the execution of the `disable` command with `--purge` may take a while to complete.
+
+## What happens when I purge:
+
+### - livepatch, realtime-kernel, landscape
+
+These services do not currently support the `--purge` operation.
+
+### - anbox, cc-eal, cis/usg, esm-apps, esm-infra*, ros (and ros-updates)
+
+When these services are disabled with `--purge`, the sources and authentication will be removed first. Then, packages that are available only in the service-specific APT repository will be removed from the system.
+
+The origin is detected based on the `origin` metadata defined in `apt` for all packages in this repository.
+
+Then, for the packages that are also present in the Ubuntu archives, there will be a downgrade to the highest possible version in the archive. Downgrade operations are the most common because the packages for specific services usually have higher version strings than the counterparts in the archive.
+
+Note that purging a service often results in the installation of newer versions of packages than were originally present in the system (i.e., before the given service was enabled). It is important to use the latest versions available in the archive to guarantee that dependency chains are resolved - in other words, that there are no broken dependencies between packages.
+
+```{note}
+\* Disabling `esm-infra` with `--purge` may involve removing a kernel, see below for more information.
+```
+
+### - FIPS (and fips-updates/fips-preview)
+
+In the case of FIPS-related services (and in some cases, `esm-infra`), there is an extra consideration when purging the packages: there may be Linux kernel packages among the ones to be removed/downgraded.
+
+In this case, the Pro Client will look for at least one more kernel installed in the system. This check is performed looking at the `apt` packages installed, and matching the version strings to `vmlinu[z|x]` files in `/boot`.
+
+If no other kernel is found in the system, then the current kernel cannot be removed. The Client will warn the user and abort the operation.
+
+Kernels which are manually compiled and installed, or that are not shipped in Ubuntu as APT packages will *not* be considered and validated.
+
+If another Ubuntu kernel is found in the system, `--purge` will proceed to remove and downgrade packages normally. In the process, the user will be warned that a kernel is being removed, and that it is their responsibility to make sure the alternative kernels can be booted and are working.
+
+A reboot is always needed when kernel packages are changed when purging a service.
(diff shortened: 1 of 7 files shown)


docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+## Using the CLI to disable and purge a service
+
+First of all it is important to note that this is an ***experimental feature***, so it is available but it is advisable to use at one's own discretion.
+
+To purge a service, use the CLI command:
+```bash
+$ sudo pro disable <service-name> --purge
+```
+
+Because of the risks involved, the expectation is that someone is looking at the screen while purging a service. To make it harder to use this feature on scripts, the `--assume-yes` flag is incompatible with the `--purge` flag.
```
s-makin, line 43: I'm not sure I like the way this is worded - it makes it sounds like you guys implemented it that way to be awkward/annoying when actually it was a safeguard to prevent accidental damage to users systems, so I'd suggest something more like the following:

We recommend only using `purge` when you are able to monitor the screen. Since the `purge` feature has the potential to do serious accidental damage to a system if used unattended (for example, in a script), the `--assume-yes` flag was made incompatible with the `--purge` flag.

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+Because packages are being uninstalled/reinstalled, the execution of the `disable` command with `--purge` may take a while to complete.
+
+## What happens when I purge:
+
+### - livepatch, realtime-kernel, landscape
+
+Those services do not support the `--purge` operation as of now.
```
s-makin, line 51:
```suggestion
These services do not currently support the `--purge` operation.
```

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+### - anbox, cc-eal, cis/usg, esm-apps, esm-infra*, ros (and ros-updates)
+
+When those services are disabled with `--purge`, the sources and authentication will be removed first. Then, packages that are available only in the service repository will be removed from the system.
```
s-makin, line 55:
```suggestion
When these services are disabled with `--purge`, the sources and authentication will be removed first. Then, packages that are available only in the service repository will be removed from the system.
```

s-makin: I'm not sure what you mean by "service repository"

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+The origin is detected based on the `origin` metadata defined on `apt` for all packages in this repository.
```
s-makin, line 57:
```suggestion
The origin is detected based on the `origin` metadata defined in `apt` for all packages in this repository.
```

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+Then, for the packages that are also present in the Ubuntu archives, there will be a downgrade to the highest possible version in the archive. Downgrade operations are the most common because the packages for specific services usually have higher version strings than the counterparts in the archive.
+
+Note that this does not mean that the installed versions after purging a service would be the same versions that were in the system before the given service was enabled. It is important to use the latest versions available in the archive to guarantee that dependency chains are resolved - in other words, that there are no broken dependencies between packages.
```
s-makin, line 61:
```suggestion
Note that purging a service often results in the installation of newer versions of packages than were originally present in the system (i.e., before the service was enabled). It is important to use the latest versions available in the archive to guarantee that dependency chains are resolved - in other words, that there are no broken dependencies between packages.
```

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+```{note}
+\* Disabling esm-infra with --purge may involve removing a kernel, see below for more information
```
s-makin, line 64:
```suggestion
\* Disabling `esm-infra` with `--purge` may involve removing a kernel, see below for more information.
```

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+### - FIPS (and fips-updates/fips-preview)
+
+In the case of FIPS related services (and in some cases, esm-infra), there is an extra care when purging the packages: there may be Linux Kernel packages among the ones to be removed/downgraded.
```
s-makin, line 69:
```suggestion
In the case of FIPS-related services (and in some cases, `esm-infra`), there is an extra consideration when purging the packages: there may be Linux Kernel packages among the ones to be removed/downgraded.
```

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+In this case, the Pro Client will look for at least one more kernel installed in the system. This check is performed looking at the `apt` packages installed, and matching the version strings to `vmlinu[z|x]` files in `/boot`.
```
s-makin, line 71: should kernel be capitalised or no? I've seen both and I'm not sure which way is right

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+If no other kernel is found in the system, then the current kernel cannot be removed. The Client will warn the user and abort the operation.
+
+Kernels which are manually compiled and installed, or that are not shipped in Ubuntu as apt packages will *not* be considered and validated.
```
s-makin, line 75:
```suggestion
Kernels which are manually compiled and installed, or that are not shipped in Ubuntu as APT packages will *not* be considered and validated.
```

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+If another Ubuntu kernel is found in the system, `--purge` will proceed to remove and downgrade packages normally. In the process, the user will be warned that a kernel is being removed, and that it is a user responsibility to make sure the alternative kernels can be booted and are working.
```
s-makin, line 77:
```suggestion
If another Ubuntu kernel is found in the system, `--purge` will proceed to remove and downgrade packages normally. In the process, the user will be warned that a kernel is being removed, and that it is their responsibility to make sure the alternative kernels can be booted and are working.
```

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
+# Disabling and purging services
+
+All services enabled using the Pro Client can be disabled using the command line:
+
+```bash
+$ sudo pro disable <service-name>
```
s-makin, line 6: Can we safely remove the $ symbols, or do they need to be here for clarity?

s-makin (commented):
Mostly just nits :) I do have a couple of questions that need clarifying though (I told you I would have!)

s-makin:
> Mostly just nits :) I do have a couple of questions that need clarifying though (I told you I would have!)

One suggestion I have, which I think falls within the scope of this PR, is that we should probably include a link to this page in the "enable" how-to guides (except for the ones where --purge isn't available). This isn't a page that users will know to look for (or realise is an option available to them) so to help with discovery we should point them to it.

docs/explanations/purging_services.md
renanrodrigo: Symbols are there for clarity.
`$` means you are running as a non-root user, and we want to show it and the `sudo` command there instead of `#` or nothing, to make it explicit that this is a command which requires privileges

docs/explanations/purging_services.md
renanrodrigo: not capitalized - fixed

docs/explanations/purging_services.md
renanrodrigo: the service-specific apt repository. added.

docs/explanations/purging_services.md
renanrodrigo: accepted

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+If there are other services which depend on the service being disabled, those need to be disabled too. A prompt in the CLI will list those and and ask to disable them. The `--assume-yes` flag can be used to automatically accept this prompt.
+
+## What does `purge` mean in this context?
+
+Disabling and *purging* a service will have a similar effect of what is done by [`ppa-purge`](https://launchpad.net/ppa-purge) - it will remove access to the sources from where the service packages can be obtained, and then make best-effort to downgrade those packages back to their Ubuntu versions - the ones published in the Ubuntu Archive.
+
+If a package is specific to the given service (`ubuntu-fips` for the FIPS related services, for example) then it will be removed during the purge process, when possible.
+
+The operation of reverting/uninstalling packages may leave the system in an undesired state, due to packages not meeting dependencies (partial or broken system caches) or to removal of kernels when alternative kernels are not working properly. Therefore, the feature is aimed at users/administrators who understand the impacts.
```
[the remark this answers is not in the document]
renanrodrigo: I had to google what 'admonition' means but yes I added it to a box

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+$ sudo pro disable <service-name>
+```
+
+When you disable a service, the Client will remove all access to that specific service's packages and sources in the system. Note that only the access is removed, not the packages (debs, snaps) installed using the service.
+
+For example, the command
+```bash
+$ sudo pro disable esm-apps
+```
+will remove the sources.list entry for esm-apps, and also the authentication information that granted access to the packages. Any deb installed from esm-apps will remain istalled on the system.
```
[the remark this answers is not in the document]
renanrodrigo: hmmm I changed the voice in this one, but esm-apps does not install any package. Tried to adapt

docs/explanations/purging_services.md
```diff
@@ -0,0 +1,79 @@
...
+
+On the same line, the command
+```bash
+$ sudo pro disable livepatch
+```
+will disable the livepatch service, but the `canonical-livepatch` snap will still be installed in the system, although inactive.
+
+If there are other services which depend on the service being disabled, those need to be disabled too. A prompt in the CLI will list those and and ask to disable them. The `--assume-yes` flag can be used to automatically accept this prompt.
```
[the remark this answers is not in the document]
renanrodrigo: When a service `A` depends on service `B` to be enabled, but you are disabling `B`.
AFAICT the only case we have now in the client is `ros`, which depends on `esm-apps` to work.

renanrodrigo:
@s-makin I added a new commit with the reference in the `enable` howtos - PTAL (:

s-makin (approved):
All LGTM, thanks for adding the extra links!

PR #2789, opened 2023-10-13
livepatch: better inform if command fails

lucasmoura:
## Why is this needed?
If the canonical-livepatch status command fails, we can mistakenly tell the users that Livepatch is disabled. This is an incorrect behavior as we should alert the users about the error. We are now reflecting that went wrong with the command on the output of pro status.

LP: #2019997

## Test Steps
1. Launch a Jammy LXD VM
2. Attach it to a Pro subscription
3. Verify that that there is a warning for Livepatch when running `pro status` (warning due to kernel incompatibility)
4. Switch to ubuntu user
5. Run `pro status` again and verify that we output an error running `canonical-livepatch status`

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/livepatch.py
+++ b/uaclient/livepatch.py
@@ -130,9 +130,12 @@ def status() -> Optional[LivepatchStatusStatus]:
         out, _ = system.subp(
             [LIVEPATCH_CMD, "status", "--verbose", "--format", "json"]
         )
-    except exceptions.ProcessExecutionError:
-        LOG.warning("canonical-livepatch returned error when checking status")
-        return None
+    except exceptions.ProcessExecutionError as e:
+        LOG.warning(
+            "canonical-livepatch returned error when checking status:\n%s",
+            exc_info=e,
+        )
+        raise e
 
     try:
         status_json = json.loads(out)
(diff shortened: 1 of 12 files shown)


lucasmoura:
@orndorffgrant updated

uaclient/livepatch.py
```diff
@@ -130,9 +130,9 @@ def status() -> Optional[LivepatchStatusStatus]:
         out, _ = system.subp(
             [LIVEPATCH_CMD, "status", "--verbose", "--format", "json"]
         )
-    except exceptions.ProcessExecutionError:
+    except exceptions.ProcessExecutionError as e:
         LOG.warning("canonical-livepatch returned error when checking status")
-        return None
+        raise e
```
orndorffgrant, line 135: While you're improving this, can you add the error message to this logging statement too, so we get a record in bug reports

orndorffgrant (commented):
LGTM, but now that `livepatch.status()` can raise that exception, I think it should be caught at all call-sites. We call this in 4 other places - they should all come up with a search for `livepatch.status()` - and I think we should `try-except` all of them

lucasmoura:
Good catch @orndorffgrant. I have updated the code

orndorffgrant (approved):
LGTM Thank you!

Issue #2601, opened 2023-05-30
Non-root user getting errors collecting jobs-status.json


renanrodrigo:
`/var/lib/ubuntu-advantage/jobs-status.json` is not (for any reason?) user readable, so it fails when non-root tries to collect it.

## How to reproduce:
- (as root)  `python3 /usr/lib/ubuntu-advantage/timer.py`
- (as non-root) `pro collect-logs`

PR #2796, opened 2023-10-16
timer: make jobs-status file world readable

lucasmoura:
## Why is this needed?
When running collect-logs as non-root, we might fail trying to collect the jobs-status information. Since there is no private date in this file, we are making it world readable now.

Fixes: #2601

## Test Steps
Run the modified integration test and the manual test in this PR

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/collect_logs.feature
+++ b/features/collect_logs.feature
@@ -8,9 +8,9 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
         # simulate logrotate
         When I run `touch /var/log/ubuntu-advantage.log.1` with sudo
         When I run `touch /var/log/ubuntu-advantage.log.2.gz` with sudo
-        When I run `pro collect-logs` with sudo
+        When I run `pro collect-logs` as non-root
         Then I verify that files exist matching `ua_logs.tar.gz`
-        When I run `tar zxf ua_logs.tar.gz` as non-root
+        When I run `tar zxf ua_logs.tar.gz` with sudo
         Then I verify that files exist matching `logs/`
         When I run `sh -c "ls -1 logs/ | sort -d"` as non-root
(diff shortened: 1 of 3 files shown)


debian/ubuntu-advantage-tools.postinst
```diff
@@ -424,6 +424,12 @@ case "$1" in
       fi
       chown root:root /var/log/ubuntu-advantage.log
 
+      if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "31~"; then
+        if [ -f /var/lib/ubuntu-advantage/jobs-status.json ]; then
+            chmod 0644 /var/lib/ubuntu-advantage/jobs-status.json
+        fi
+      fi
+
```
orndorffgrant, line 432: Does this change deserve a postinst transition? I think the next time the file is written it will get the new permissions anyway.

features/collect_logs.feature
```diff
@@ -1,5 +1,6 @@
 Feature: Command behaviour when attached to an Ubuntu Pro subscription
 
+    @wip
```
orndorffgrant, line 3: some `@wip`s left behind :upside_down_face:

debian/ubuntu-advantage-tools.postinst
lucasmoura: Yeah, I think you are right. My concern was that if the user disabled the timer jobs, we would not update the file, but this is not the case. I will remove this bit together with the manual test

lucasmoura:
@orndorffgrant updated

orndorffgrant (approved):
LGTM thank you!

CalvoM (approved):
LGTM

renanrodrigo (approved):

PR #2797, opened 2023-10-17
Update contributing docs

s-makin:
The contributing docs needed a bit of additional info on docs
contributions specifically.

```
Updated the README in this branch to point people to the contributions
guide so they can find build instructions.

Split out "how to contribute to docs" from the style guide, then expanded
both with more detail.

Updated the docs index to point to the docs-branch version of
CONTRIBUTING.md
```
## Checklist

 - [ ] Yes
 - [x] No


aciba90 (approved):
LGTM, thanks!

Not sure if this worth doing, as we do not have many external contributions. But, do you think it is worth referencing `dev-docs/documentation/contribute_docs.md` in https://github.com/canonical/ubuntu-pro-client/blob/main/.github/PULL_REQUEST_TEMPLATE.md? In that way we could drive external / newbie contributors to how to contribute to the docs.

s-makin:
> LGTM, thanks!
> 
> Not sure if this worth doing, as we do not have many external contributions. But, do you think it is worth referencing `dev-docs/documentation/contribute_docs.md` in https://github.com/canonical/ubuntu-pro-client/blob/main/.github/PULL_REQUEST_TEMPLATE.md? In that way we could drive external / newbie contributors to how to contribute to the docs.

That's actually a really good idea, since we request docs be done before a feature can land. Good to point people to where they should look for that. I'll add it!

aciba90:
For the record, the PR template has been removed because only the one from the `default` branch is used:  https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates#pull-request-templates

orndorffgrant (approved):
LGTM thank you!

PR #2800, opened 2023-10-17
Release 30 review

renanrodrigo:
This commit addresses review comments made in Launchpad regarding release 30

### DIFF
--- a/uaclient/entitlements/repo.py
+++ b/uaclient/entitlements/repo.py
@@ -25,6 +25,8 @@ from uaclient.entitlements.entitlement_status import (
 event = event_logger.get_event_logger()
 LOG = logging.getLogger(util.replace_top_level_logger_name(__name__))
 
+# See 'What does a specific Ubuntu kernel version number mean?' in
+# https://wiki.ubuntu.com/Kernel/FAQ
 RE_KERNEL_PKG = r"^linux-image-[\d]+[.-][\d]+[.-][\d]+-[\d]+-[A-Za-z0-9_-]+$"
 
 
@@ -166,6 +168,20 @@ class RepoEntitlement(base.UAEntitlement):
         return True
 
     def purge_kernel_check(self, package_list):
+        """
+        Checks if the purge operation involves a kernel.
+        
+        When package called 'linux-image-*' is in the package list, warn the
+        user that a kernel is being removed. Then, show the user what the
+        current kernel is.
+
+        If the current kernel is to be removed, and there are no other valid
+        Ubuntu Kernels installed in the system, return False to abort the
+        operation.
+
+        If there is another Ubuntu kernel - besides the one installed - then
+        prompt the user for confirmation before proceeding.
+        """
         linux_image_versions = [
             package.name[len("linux-image-") :]
             for package in package_list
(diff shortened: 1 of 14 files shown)


renanrodrigo:
Of course we using some pytest fixture that is not present on Xenial, right?
Will fix.

uaclient/entitlements/repo.py
```diff
@@ -183,9 +183,10 @@ def purge_kernel_check(self, package_list):
         prompt the user for confirmation before proceeding.
         """
         linux_image_versions = [
-            package.name[len("linux-image-") :]
+            m.group(1)
             for package in package_list
-            if re.match(RE_KERNEL_PKG, package.name)
+            for m in [re.search(RE_KERNEL_PKG, package.name)]
+            if m
         ]
```
orndorffgrant, line 191: I'm not a fan of multiple loops in a list comprehension, and in this case it's kind of a hack to iterate over a list of size one just to get the match variable and check if it's None.

I recommend something like:
```python
        linux_image_versions = []
        for package in package_list:
            m = re.search(RE_KERNEL_PKG, package.name)
            if m:
                linux_image_versions.append(m.group(1))
```

uaclient/messages/__init__.py
```diff
@@ -1991,12 +1991,12 @@ def __repr__(self):
...
 E_NOT_SETTING_PROXY_NOT_WORKING = FormattedNamedMessage(
     "proxy-not-working",
-    't.gettext("{proxy}") is not working. Not setting as proxy.',
+    t.gettext('"{proxy}" is not working. Not setting as proxy.'),
 )
 
 E_NOT_SETTING_PROXY_INVALID_URL = FormattedNamedMessage(
     "proxy-invalid-url",
-    't.gettext("{proxy}") is not a valid url. Not setting as proxy.',
+    t.gettext('"{proxy}" is not a valid url. Not setting as proxy.'),
```
orndorffgrant, line 1999: Whoops :upside_down_face: thanks

uaclient/files/files.py
```diff
@@ -41,11 +41,15 @@ def write(self, content: str):
             if self.is_private
             else defaults.WORLD_READABLE_MODE
         )
-        if not os.path.exists(self._directory):
+        # tyy/except-ing here avoids race conditions the best
```
orndorffgrant, line 44:
```suggestion
        # try/except-ing here avoids race conditions the best
```

orndorffgrant (approved):
Some minor nits, otherwise looks good - thank you!

orndorffgrant:
But oh no the tmp_path pytest thing isn't working on xenial :(

orndorffgrant:
It looks like other unit tests use the `tmpdir` fixture instead

uaclient/entitlements/repo.py
renanrodrigo: agreed readability++

uaclient/messages/__init__.py
renanrodrigo: anytime (:

uaclient/files/files.py
renanrodrigo: I can´t  type sorry

lucasmoura (approved):

bryceharrington (approved):
This all looks great.  The updated docs and comments read well.  Thanks for adding the thorough ESM cache and kernel test cases with such fast turnaround.  I'm not going to insist on adding the apt cache tests although would be nice in the future to strengthen the tests at that layer since it's an important interface to the underlying system, but you're right that this is already well exercised already elsewhere.

PR #2803, opened 2023-10-20
Fix brazilian portuguese translations

lucasmoura:
## Why is this needed?
Some brazilian portuguese translations are incorrect. This PR fix those issues

## Test Steps
Check the modified translation matches the english string

## Checklist

 - [ ] Yes
 - [x] No


panlinux:
Consider adding `--check-format` to the `msgfmt` call in `debian/po/Makefile`, as it seems to catch the class of errors where format strings or python `{}` vars in the original string are missing from the translated one:
```console
$ make
mkdir -p usr/share/locale/pt_BR/LC_MESSAGES
msgfmt -f --check-format pt_BR.po -o usr/share/locale/pt_BR/LC_MESSAGES/ubuntu-pro.mo
pt_BR.po:204: number of format specifications in 'msgid' and 'msgstr' does not match
pt_BR.po:214: number of format specifications in 'msgid' and 'msgstr' does not match
pt_BR.po:1319: a format specification for argument 'date' doesn't exist in 'msgstr'
msgfmt: found 3 fatal errors
make: *** [Makefile:8: pt_BR.mo] Error 1
```

lucasmoura:
@panlinux I have applied the suggestion to the Makefile command

debian/po/pt_BR.po
```diff
@@ -465,12 +463,11 @@ msgid "The following package(s) will be REMOVED:"
 msgstr ""
 
 #: ../../uaclient/messages/__init__.py:148
-#, fuzzy
 msgid "The following package(s) will be reinstalled from the archive:"
-msgstr "Você não tem pacotes instalados de terceitos."
+msgstr "O(s) pacote(s) à seguir serão reinstalados:"
```
panlinux, line 468: This translation is incorrect. Before removing a `fuzzy` translation, check that it is correct.

debian/po/pt_BR.po
```diff
@@ -493,7 +490,7 @@ msgstr[1] ""
 "Renove sua assinatura em {url}"
```
panlinux, line 490: `ubuntu Pro` is lowercase in the translation, but `Ubuntu Pro` in the original text. Also, wasn't this translation meant to support singular and plural texts? Just wondering because both in the original string and in the translated one, it's plural (or at least using `(s)`).

panlinux: Also, sorry, just saw it now, but the translated string is using `PRO` instead of `Pro`.

debian/po/pt_BR.po
```diff
@@ -2350,9 +2347,8 @@ msgid "FIPS Updates"
 msgstr ""
 
 #: ../../uaclient/messages/__init__.py:1320
-#, fuzzy
 msgid "FIPS compliant crypto packages with stable security updates"
-msgstr "Pacotes instalados com atualizações disponíveis por {service}:"
+msgstr ""
```
panlinux, line 2352: You removed the incorrect translation, but didn't add a fixed one?

lucasmoura: The reason for that is because we have not translated any service-related messages for this release. So for consistency sake, I left that one untranslated too

debian/po/pt_BR.po
lucasmoura: Good catch, I have solved the issue

debian/po/pt_BR.po
```diff
@@ -2349,6 +2349,8 @@ msgstr ""
 #: ../../uaclient/messages/__init__.py:1320
 msgid "FIPS compliant crypto packages with stable security updates"
 msgstr ""
+"Pacotes de criptografia compaíveis com FIPS com atualizações de segurança "
```
lucasmoura, line 2352: s/compaíveis/compatíveis

debian/po/pt_BR.po
```diff
@@ -2554,7 +2557,7 @@ msgstr ""
 
 #: ../../uaclient/messages/__init__.py:1463
 msgid "All Updates for the Robot Operating System"
-msgstr ""
+msgstr "Todas as Atualizações para o Robot Operating System"
```
lucasmoura, line 2560: s/Atualizações/atualizações

debian/po/pt_BR.po
renanrodrigo: :|

debian/po/pt_BR.po
renanrodrigo: I kept the uppercase to match the uppercase in All Updates.
I think we must have a policy for this, because it is not consistent across messages

PR #2804, opened 2023-10-20
config: make user-config world readable

lucasmoura:
## Why is this needed?
To allow non-root users to see user-config changes performed by root users, we need to make the user-config file world-readable

## Test Steps
Verify that the modified integration test and the manual test are passing

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -429,6 +429,12 @@ case "$1" in
           chmod 0700 "$private_dir"
       fi
 
+      if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "30~"; then
+        if [ -f /var/lib/ubuntu-advantage/user-config.json ]; then
+            chmod 0644 /var/lib/ubuntu-advantage/user-config.json
+        fi
+      fi
+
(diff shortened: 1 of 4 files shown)


debian/ubuntu-advantage-tools.postinst
```diff
@@ -429,6 +429,12 @@ case "$1" in
           chmod 0700 "$private_dir"
       fi
 
+      if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "30~"; then
+        if [ -f /var/lib/ubuntu-advantage/user-config.json ]; then
+            chmod 0644 /var/lib/ubuntu-advantage/user-config.json
+        fi
+      fi
```
orndorffgrant, line 436: I'd like to continue the pattern of keeping the version migrations in order, so can you move this down after the `if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "29~";` block?

orndorffgrant:
Discussion decision: we shouldn't make this file world-readable because it may include proxy secrets. We want to fix the underlying bug properly and not rush it into this release. I'll create an issue then close this.

orndorffgrant:
Issue is here https://github.com/canonical/ubuntu-pro-client/issues/2809

PR #2808, opened 2023-10-24
plan: fix error when variable is not set

lucasmoura:
## Why is this needed?
There is a corner case scenario where we end up with a src_pocket_pkgs variable unset. We are now fixing this issue by using a default value for this variable

## Test Steps
Run the new integration test

## Checklist

 - [ ] Yes
 - [x] No


orndorffgrant (approved):
Thanks this worked in my test!

PR #2812, opened 2023-10-26
Change order of operations when purging services

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because now purge works on generic kernels.

## Test Steps
See fixed bug - it does not happen here anymore.

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo:
release-checks will fail because we don't want to list GH bugs that originated from the review in the changelog

orndorffgrant (approved):
LGTM and tests passed

PR #2816, opened 2023-10-27
Fix status test

lucasmoura:
## Why is this needed?
Now that fips-preview is on prod, we need to update status tests to acknowledge that new service on the output

Additionally, on GCP, we block enabling FIPS if we detect that there is no FIPS cloud optimized kernel. However, since fips-preview is intended for early access and the packages there are not yet certified, we believe we should not block
enabling the service on clouds. Additionally, fips-preview is already using default fips metapackages for other clouds we support

## Test Steps
Run the modified integration tests and check that they are all passing now

## Checklist

 - [ ] Yes
 - [ ] No


orndorffgrant (approved):
LGTM

renanrodrigo (approved):
lgtm

Issue #2647, opened 2023-07-07
Contract server error message should be more specific


rpocase:
When users hit various errors, even those not related to the contract server, they may receive this message:

```
Failed to connect to authentication server\n
check your internet connection and try again
```

This leads to questions like:
1. Whats a authentication server? 
2. What endpoint can I not reach?
3. Is this problem on my end or the ubuntu pro infra side?

These error messages should be more descriptive of the exact error being hit, and should in no cases be used for errors not related to the contract server.

PR #2817, opened 2023-10-30
update connectivity error message

lucasmoura:
## Why is this needed?
We want to make connectivity errors clearer and easier to debug for the user. To achieve that, we are now displaying the url that cause the error together with the cause behind the error.

Fixes: #2647

## Test Steps
1. Launch a lxd container
2. Disable network on container with: `lxc config device add NAME eth0 none`
3. Run `pro fix USN-4539-1`

Expect the following message:
```
Failed to connect to https://ubuntu.com/security/notices/USN-4539-1.json
[Errno -3] Temporary failure in name resolution 
```

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/unattached_status.feature
+++ b/features/unattached_status.feature
@@ -9,26 +9,28 @@ Feature: Unattached status
         When I run `sed -i 's/contracts.can/invalidurl.notcan/' /etc/ubuntu-advantage/uaclient.conf` with sudo
         And I verify that running `pro status --format json` `as non-root` exits `1`
         Then stdout is a json matching the `ua_status` schema
-        And I will see the following on stdout:
-            """
-            {"environment_vars": [], "errors": [{"message": "Failed to connect to authentication server\nCheck your Internet connection and try again.", "message_code": "connectivity-error", "service": null, "type": "system"}], "result": "failure", "services": [], "warnings": []}
-            """
+        And stdout matches regexp:
+        """
+        {"environment_vars": \[\], "errors": \[{"message": "Failed to connect to .*\\n\[Errno -2\] Name or service not known\\n", "message_code": "connectivity-error", "service": null, "type": "system"}\], "result": "failure", "services": \[\], "warnings": \[\]}
+        """
         And I verify that running `pro status --format yaml` `as non-root` exits `1`
         Then stdout is a yaml matching the `ua_status` schema
-        And I will see the following on stdout:
-            """
-            environment_vars: []
-            errors:
-            - message: 'Failed to connect to authentication server
+        And stdout matches regexp:
+        """
+        environment_vars: \[\]
+        errors:
+        - message: 'Failed to connect to https://invalidurl.notcanonical.com/v1/resources(.*)
 
-                Check your Internet connection and try again.'
-              message_code: connectivity-error
-              service: null
-              type: system
-            result: failure
-            services: []
-            warnings: []
-            """
+            \[Errno -2\] Name or service not known
+
+            '
+          message_code: connectivity-error
+          service: null
+          type: system
+        result: failure
+        services: \[\]
+        warnings: \[\]
+        """
 
         Examples: ubuntu release
            | release | machine_type  |
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -50,7 +50,7 @@ def attach_with_token(
 ) -> None:
     """
     Common functionality to take a token and attach via contract backend
-    :raise UrlError: On unexpected connectivity issues to contract
+    :raise ConnectivityError: On unexpected connectivity issues to contract
         server or inability to access identity doc from metadata service.
     :raise ContractAPIError: On unexpected errors when talking to the contract
         server.
@@ -59,14 +59,9 @@ def attach_with_token(
 
     contract_client = contract.UAContractClient(cfg)
     attached_at = datetime.datetime.now(tz=datetime.timezone.utc)
-
-    try:
-        new_machine_token = contract_client.add_contract_machine(
-            contract_token=token, attachment_dt=attached_at
-        )
-    except exceptions.UrlError as e:
-        LOG.exception(str(e))
-        raise exceptions.ConnectivityError()
+    new_machine_token = contract_client.add_contract_machine(
+        contract_token=token, attachment_dt=attached_at
+    )
 
     cfg.machine_token_file.write(new_machine_token)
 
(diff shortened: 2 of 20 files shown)


lucasmoura:
@orndorffgrant @renanrodrigo I am wondering if we should add the HTTP error code if available or if stating the cause is enough for debugging. Let me know what you think

orndorffgrant:
I think the HTTP code doesn't hurt and could help

I'm tempted to suggest we combine `UrlError` and `ConnectivityError` into one exception class - but it'd be a little tricky to combine them. In order to maintain backwards compatibility, the new class would need to doubly inherit from UbuntuProError and IOError. And we'd need to keep both names pointing to the same class. What do you think?

lucasmoura:
@orndorffgrant I think that is a good idea, since we only raise `ConnectivyError` exceptions when an `UrlError` has already happened. We would need to handle the MRO for the `__init__` call, which I don't think is a huge problem. The only issue I see is that it will now take more parameters to instantiate a `ConnectivityError` exception, due to the `IOError`requirements, which might affect backwards compatibility. But maybe I am overthinking this.

lucasmoura:
We have decided not to add the HTTP response code on the message, since we only raise a  `ConnectivityError`  when the issue is not tied to an HTTP error

uaclient/exceptions.py
```diff
@@ -236,6 +218,31 @@ def __init__(self, url, **kwargs) -> None:
...
+    ):
+        if getattr(cause, "reason", None):
+            cause_error = str(getattr(cause, "reason"))
+        else:
+            cause_error = str(cause)
+        IOError.__init__(self, cause_error)
+        UbuntuProError.__init__(self, cause_error=cause_error, url=url)
+
+        # Even though we already set those variable through UbuntuProError
+        # we need to set them again to avoid mype warnings
```
renanrodrigo, line 237:
```suggestion
        # we need to set them again to avoid mypy warnings
```

renanrodrigo (approved):

uaclient/exceptions.py
lucasmoura: Fixed

uaclient/contract.py
```diff
@@ -264,9 +264,9 @@ def get_magic_attach_token_info(self, magic_token: str) -> Dict[str, Any]:
             response = self.request_url(
                 API_V1_GET_MAGIC_ATTACH_TOKEN_INFO, headers=headers
             )
-        except exceptions.UrlError as e:
-            LOG.exception(e)
-            raise exceptions.ConnectivityError()
+        except exceptions.ConnectivityError as e:
+            LOG.exception(e.cause_error)
+            raise e
```
orndorffgrant, line 269: There are a couple places where we catch ConnectivityError, log the message, and then re-raise it. Maybe we should log the error when raising it originally in http/__init__.py and then remove all the catch-log-reraise blocks - what do you think?

lucasmoura: That makes sense, I will update the PR

lucasmoura:
@orndorffgrant updated

features/unattached_status.feature
```diff
@@ -9,26 +9,28 @@ Feature: Unattached status
...
-        And I will see the following on stdout:
-            """
-            environment_vars: []
-            errors:
-            - message: 'Failed to connect to authentication server
+        And stdout matches regexp:
+        """
+        environment_vars: \[\]
+        errors:
+        - message: 'Failed to connect to .*
```
orndorffgrant, line 22: (same comment applies to json version above)

I agree it's easiest to make this a regex but I think we can assert a bit more since we know mostly what the URL will be. I think this should work
```suggestion
        - message: 'Failed to connect to https://invalidurl.notcanonical.com/v1/resources?architecture=amd64&kernel=(.*)&series=<release>&virt=lxc
```

uaclient/actions.py
```diff
@@ -64,9 +64,8 @@ def attach_with_token(
         new_machine_token = contract_client.add_contract_machine(
             contract_token=token, attachment_dt=attached_at
         )
-    except exceptions.UrlError as e:
-        LOG.exception(str(e))
-        raise exceptions.ConnectivityError()
+    except exceptions.ConnectivityError as e:
+        raise e
```
orndorffgrant, line 68: We can delete these now since they're empty :D
```suggestion
```

uaclient/contract.py
```diff
@@ -264,9 +264,8 @@ def get_magic_attach_token_info(self, magic_token: str) -> Dict[str, Any]:
             response = self.request_url(
                 API_V1_GET_MAGIC_ATTACH_TOKEN_INFO, headers=headers
             )
-        except exceptions.UrlError as e:
-            LOG.exception(e)
-            raise exceptions.ConnectivityError()
+        except exceptions.ConnectivityError as e:
+            raise e
```
orndorffgrant, line 268:
```suggestion
```

uaclient/contract.py
```diff
@@ -290,9 +289,8 @@ def new_magic_attach_token(self) -> Dict[str, Any]:
                 headers=headers,
                 method="POST",
             )
-        except exceptions.UrlError as e:
-            LOG.exception(e)
-            raise exceptions.ConnectivityError()
+        except exceptions.ConnectivityError as e:
+            raise e
```
orndorffgrant, line 293:
```suggestion
```

uaclient/contract.py
```diff
@@ -315,9 +313,8 @@ def revoke_magic_attach_token(self, magic_token: str):
                 headers=headers,
                 method="DELETE",
             )
-        except exceptions.UrlError as e:
-            LOG.exception(e)
-            raise exceptions.ConnectivityError()
+        except exceptions.ConnectivityError as e:
+            raise e
```
orndorffgrant, line 317:
```suggestion
```

orndorffgrant (approved):
The tiniest of nits so pre-approving, no need for another review from me

features/unattached_status.feature
```diff
@@ -9,26 +9,28 @@ Feature: Unattached status
...
-        And I will see the following on stdout:
-            """
-            environment_vars: []
-            errors:
-            - message: 'Failed to connect to authentication server
+        And stdout matches regexp:
+        """
+        environment_vars: \[\]
+        errors:
+        - message: 'Failed to connect to https://invalidurl.notcanonical.com/v1/resources\?architecture=amd64&kernel=(.*)&series=<release>&virt=lxc
```
orndorffgrant, line 22: I gave a bad suggestion sorry. I think it's good to check at least the domain name, but it turns out we can't rely on the order of the parameters on xenial
```suggestion
        - message: 'Failed to connect to https://invalidurl.notcanonical.com/v1/resources(.*)
```

orndorffgrant (commented):
There are several CI failures. I think they're mostly unrelated but probably worth checking

lucasmoura:
@orndorffgrant I have fixed the `unattached_status` test and I for the other ones failing on CI, I wasn't able to reproduce them locally, so I am inclined to say that they are all flaky. But let's see what a second run of CI will tell

orndorffgrant:
The bionic fix fail will be fixed by #2897 

Merging

PR #2818, opened 2023-10-31
Fix #1579

### LINKED ISSUE #1579: apt-hook: parameterize singular/plural in messaging based on whether package counts are 1 or > 1
Currently a number of static messages contain '(s)' to account for singular or plural ESM packages.
It would be better to make the messaging templates more dynamic and allow apt-hook to inject the "s" if needed or "is" vs "are" where applicable

orndorffgrant:
## Why is this needed?

Fixes: #1579 
Everything else in the apt-hook is already pluralized properly after the i18n refactors

## Test Steps

Edited test should pass

## Checklist

 - [ ] Yes
 - [x] No


lucasmoura (approved):

lucasmoura:
@renanrodrigo just added the portuguese fix for the updated message

renanrodrigo (approved):

PR #2821, opened 2023-10-31
test: add fips-updates tests for Jammy

lucasmoura:
## Why is this needed?
Test fips-updates support on bionic.

PS: Note that we need to update the tests once the product definition goes to production

## Test Steps
Run the new integration tests

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/features/enable_fips_cloud.feature
+++ b/features/enable_fips_cloud.feature
@@ -466,3 +466,22 @@ Feature: FIPS enablement in cloud based machines
            | release | machine_type |
            | bionic  | aws.generic  |
            | focal   | aws.generic  |
+
+    Scenario Outline: Attached enable of FIPS in an ubuntu GCP vm
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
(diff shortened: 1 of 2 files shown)


orndorffgrant (approved):

features/enable_fips_cloud.feature
```diff
@@ -466,3 +466,22 @@ Feature: FIPS enablement in cloud based machines
            | release | machine_type |
            | bionic  | aws.generic  |
            | focal   | aws.generic  |
+
+    Scenario Outline: Attached enable of FIPS in an ubuntu GCP vm
```
renanrodrigo, line 470: I'm not worrying that much now as we never did in the past, but at some point in time we need to verify the Scenario names... even the ones that are correct could be more descriptive

renanrodrigo (approved):

features/enable_fips_cloud.feature
lucasmoura: Indeed

PR #2823, opened 2023-11-02
cloud_override option in u.pro.attach.auto.full_auto_attach.v1 to support auto-attached pro docker builds

orndorffgrant:
## Why is this needed?

This adds integration test coverage for our settings_overrides.cloud_type config option. It also adds a more ergonomic way of overriding the cloud_type for only a single auto-attach operation by adding a new optional parameter to the full_auto_attach.v1 API. This is useful in situations where cloud-id cannot identify the cloud. In particular, this is useful for making pro docker containers on auto-attached instances.

## Test Steps

Run the new integration tests

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/docker.feature
+++ b/features/docker.feature
@@ -119,3 +119,47 @@ Feature: Build docker images with pro services
            | jammy   | aws.pro      | aws            | xenial            | [ "esm-infra" ] | curl              | esm                  | --network=host   |
            | jammy   | azure.pro    | azure          | bionic            | [ "fips" ]      | openssl           | fips                 |                  |
            | jammy   | gcp.pro      | gce            | focal             | [ "esm-apps" ]  | hello             | esm                  |                  |
+
+    Scenario Outline: Build pro docker images auto-attached instances - API arg method
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+        When I have the `<container_release>` debs under test in `/home/ubuntu`
+        When I run `apt-get update` with sudo
+        When I apt install `docker.io docker-buildx`
+        When I create the file `/home/ubuntu/Dockerfile` with the following:
+        """
+        FROM ubuntu:<container_release>
+        ARG PRO_CLOUD_OVERRIDE=
+
+        COPY ./ubuntu-advantage-tools.deb /ua.deb
+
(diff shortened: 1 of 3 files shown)


orndorffgrant:
@renanrodrigo @lucasmoura let me know what you think of the approach in the second commit on this PR. If you think it makes sense, I'll write up some docs to go with it (and fix the first commit in this PR which currently isn't working)

uaclient/clouds/identity.py
```diff
@@ -75,7 +77,11 @@ def cloud_instance_factory() -> clouds.AutoAttachCloudInstance:
         "gce": gcp.UAAutoAttachGCPInstance,
     }  # type: Dict[str, Type[clouds.AutoAttachCloudInstance]]
 
-    cloud_type, _ = get_cloud_type()
+    if cloud_override is not None:
```
lucasmoura, line 80: Should we check if the `cloud_override` is valid ?

features/docker.feature
```diff
@@ -74,3 +74,92 @@ Feature: Build docker images with pro services
            | mantic  | lxd-vm       | xenial            | [ esm-infra ]  | curl              | esm                  |
            | mantic  | lxd-vm       | bionic            | [ fips ]       | openssl           | fips                 |
            | mantic  | lxd-vm       | focal             | [ esm-apps ]   | hello             | esm                  |
+
```
lucasmoura, line 77: I think we need to drop the `@uses.config.contract_token` from the Feature definition, otherwise we will skip the following tests

uaclient/clouds/identity.py
orndorffgrant: That happens in the following lines in the same way that the output of cloud-id is checked. We look up the value in the dictionary of cloud-name -> cloud auto-attach class. If it isn't there, then we say we raise an error saying we don't support this cloud. Let me know if you think that is not sufficient

features/docker.feature
orndorffgrant: Good catch! I'll add it to only the scenarios that need it within this feature.

orndorffgrant: @lucasmoura updated!

lucasmoura (approved):

PR #2824, opened 2023-11-03
cli: small improvments to bash autocomplete

orndorffgrant:
## Why is this needed?

The main thing here is adding api endpoint autocomplete - the rest are very small refactors that I think are slight improvements

## Test Steps

Test some autocomplete scenarios in bash
- `pro api `-tab
- `pro `-tab
- `pro enable `-tab
- `pro enable ros-updates r`-tab

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/tools/ua.bash
+++ b/tools/ua.bash
@@ -4,20 +4,18 @@
 
 API_ENDPOINTS=$(/usr/bin/python3 -c 'from uaclient.api.api import VALID_ENDPOINTS; print(" ".join(VALID_ENDPOINTS))')
 SERVICES="anbox-cloud cc-eal cis esm-apps esm-infra fips fips-updates landscape livepatch realtime-kernel ros ros-updates usg"
+SUBCMDS="--debug --help --version api attach auto-attach collect-logs config detach disable enable fix help refresh security-status status system version"
 
 _ua_complete()
 {
-    local cur_word prev_word subcmds base_params
+    local cur_word prev_word
     cur_word="${COMP_WORDS[COMP_CWORD]}"
     prev_word="${COMP_WORDS[COMP_CWORD-1]}"
-
-    subcmds="--debug --help --version api attach auto-attach collect-logs config detach disable enable fix help refresh security-status status system version"
-    base_params=""
 
     case ${COMP_CWORD} in
         1)
             # shellcheck disable=SC2207
-            COMPREPLY=($(compgen -W "$base_params $subcmds" -- $cur_word))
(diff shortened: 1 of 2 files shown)


tools/ua.bash
```diff
@@ -2,40 +2,54 @@
...
+SUBCMDS="--debug --help --version api attach auto-attach collect-logs config detach disable enable fix help refresh security-status status system version"
+
 _ua_complete()
 {
-    local cur_word prev_word services subcmds base_params
+    local cur_word prev_word
     cur_word="${COMP_WORDS[COMP_CWORD]}"
     prev_word="${COMP_WORDS[COMP_CWORD-1]}"
 
-    if [ "$VERSION_ID" = "16.04" ] || [ "$VERSION_ID" == "18.04" ]; then
```
renanrodrigo, line 11: I think this part is still relevant - `usg` may be misleading to Xenial/Bionic users

orndorffgrant: Is it more misleading than anbox-cloud, and realtime-kernel?

IMO tab completion isn't promising that the command will work when auto-completed.

I'd like to avoid the per-release service availability here if we can justify it.

renanrodrigo: eeehn it still bugs me because realtime-kernel for instance is recognized on the older releases, it is just not available:
```
root@fresh-adder:~# pro enable usg
Cannot enable unknown service 'usg'.
Try anbox-cloud, cc-eal, cis, esm-apps, esm-infra, fips, fips-updates,
landscape, livepatch, realtime-kernel, ros, ros-updates.

root@fresh-adder:~# pro enable anbox-cloud
Anbox Cloud is not available for Ubuntu 16.04 LTS (Xenial Xerus).
```

renanrodrigo: On a second thought, users affected would be the ones interested in disa-stig on those right now, no?
If you are interested in `cis`, you can definitely see it in the list. hm....

orndorffgrant: Yes this would only affect users who are autocompleting commands without knowing what they're doing in the first place. None of our documentation claims that usg will work on xenial or bionic. Anyone interested in disa-stig will need to be on focal onward anyway.

Further, though the difference in UX you presented is true today, we have plans to bring `usg` into its own proper separate service, which will fix that issue. Since we are planning on addressing that separately, and this is only about autocompletion of service names, I think the difference in UX for wrong commands is acceptable for the interim until `usg` is fixed.

So I'd like to keep it as-is if that's okay.

renanrodrigo: sold

renanrodrigo (approved):

PR #2827, opened 2023-11-03
Fix #1522

### LINKED ISSUE #1522: fix: "Found CVEs" for a USN misleading
When running `ua fix` on a USN number, we print all of the CVEs that our security API associates with that USN under the header: "Found CVEs:".

To a new user, this might be reasonably interpreted as "ua found these CVEs present/unpatched on your system" - which is not the case.

A different header to indicate that these are simply the known CVEs that fall under the given USN might be "Associated CVEs:"

orndorffgrant:
## Why is this needed?

Fixes: #1522

Needs updated pr_BR translation

## Checklist

 - [x] Changes here need to be documented, and this was done in: #2842 

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/po/pt_BR.po
+++ b/debian/po/pt_BR.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-10-31 12:04+0200\n"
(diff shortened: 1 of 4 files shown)


orndorffgrant:
This is a clarification of an existing message and requires an updated translation

lucasmoura (approved):
I have just added the portuguese translation for the message.
Additionally, I think we also need to update the fix docs to reflect the new message

renanrodrigo (approved):
@orndorffgrant rebase, then merge!

PR #2828, opened 2023-11-06
automated-doc-checks

orndorffgrant:
I'm looking through the docs starter pack, which has evolved since we started, and bringing in some pieces that have been added there.


orndorffgrant:
Not sure why rtd isn't building the PR now :thinking:

orndorffgrant:
Otherwise I think this has what we talked about now @s-makin - please check if I'm forgetting something

s-makin:
> Not sure why rtd isn't building the PR now 🤔

I have a suspicion it might be related to the email I got from RTD about webhook integrations now needing a "secret". Our docs were on the list of projects affected - we can have a quick catchup about it post-standup. It doesn't seem to affect the regular docs, which still render just fine for me, so I think it's probably only PR builds that are affected.

s-makin (approved):
All LGTM. Thanks for sorting this out, and sorry for the delayed review! 

The non-building RTD is a separate issue on RTD itself (nothing to do with this PR). Definitely not something to hold up this PR over.

renanrodrigo (approved):

PR #2834, opened 2023-11-09
add dummy message for the lightning talk example

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because I can prove we create issues when merging PRs.

## Test Steps
Merge the PR and see the magic happen

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo:
Breaks: Example Feature implemented here needs to be refactored for better performance

PR #2837, opened 2023-11-14
mypy: remove workaround for mypy bug

orndorffgrant:
## Why is this needed?

Now that the mypy bug is fixed, mypy is failing us. This removes the workaround and mypy passes again

## Test Steps

```
make clean
tox -e mypy
```

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (approved):
Thanks! This looks so better now (:

PR #2843, opened 2023-11-15
auto-label docs PRs

orndorffgrant:
This should auto-label any PRs against the docs branch, so that they are easily identified in the list of open Pull Requests.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,3 @@
+# Add "docs" label to any PR against this branch
+docs:
+- '**'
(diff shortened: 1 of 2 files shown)


renanrodrigo (approved):
/melikes this, same for `docs-devel` maybe?

orndorffgrant:
I was going to just wait for the sync since it is coming soon, unless we want a different `docs-devel` label?

renanrodrigo:
we good. Merging this so we can sync it together.

PR #2844, opened 2023-11-15
cli: retry grabbing the lock

orndorffgrant:
## Why is this needed?

This helps when the timer job is running in the background and a cli command happens to collide with it. The timer job is usually over quickly, so retrying several times with a small sleep time should be sufficient.

## Test Steps

Everything should still work. Maybe try running more than one command at once and see one wait until the other is finished.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/cli/__init__.py
+++ b/uaclient/cli/__init__.py
@@ -180,7 +180,7 @@ def assert_lock_file(lock_holder=None):
     def wrapper(f):
         @wraps(f)
         def new_f(*args, cfg, **kwargs):
-            with lock.SingleAttemptLock(cfg=cfg, lock_holder=lock_holder):
+            with lock.SpinLock(cfg=cfg, lock_holder=lock_holder, sleep_time=1):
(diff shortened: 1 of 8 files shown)


orndorffgrant:
I could be convinced either way - the separation is kind of nice for readability IMO, and we may need the single attempt in the future? (or realize using the spinlock is actually a bad idea in some case?)
There are certainly some other things in there that could use refactoring though, like the use of methods on UAConfig

lucasmoura:
I think it is fine if we remove it. We can remove it on a separate commit so that we can revert it easily in the future if we need the single attempt lock, specially because this module doesn't change that often

orndorffgrant:
Sounds good, I'll add a commit on top of this one combining the lock classes

orndorffgrant:

renanrodrigo (approved):
lgtm thanks!

renanrodrigo:
CI errors completely unrelated

PR #2856, opened 2023-11-27
Docs for release 30

renanrodrigo:
This brings the documentation from docs-devel so we publish docs for release 30


orndorffgrant:
Ah unfortunately the changes in docs-devel were added before the new CI. We should fix the link and spellcheck errors before merging. If you want I can push a commit on top of this branch?

renanrodrigo:
@orndorffgrant nah I can handle that - will be good exercise too

Issue #2860, opened 2023-11-29
`collect-logs` is missing some services


panlinux:
**Description of the bug**

`pro collect-logs` attempts to collect journal logs from the UA services. These services are defined in https://github.com/canonical/ubuntu-pro-client/blob/483d7154c621b2a4a3ef308cda35f58192bf8024/uaclient/actions.py#L34

Currently at least two services are missing from the list: `apt-news.service` and `esm-cache.service`.

**Expected behavior**

When I run `pro collect-logs`, I would also like to have logs for `apt-news.service` and `esm-cache.service` in `pro-journal.txt`.

**Current behavior**
The `pro-journal.txt` file generated by `pro collect-logs` does not contain the journal logs for `apt-news.service` and `esm-cache.service`.

**To Reproduce**
Run `pro collect-logs`, expand the tarball, and verify that `pro-journal.txt` does not contain journal entries for `apt-news.service` nor `esm-cache.service`.

**System information:**

- Pro Client version: current main

PR #2861, opened 2023-11-29
collect-logs: include esm-cache and apt-news services

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because... it
Fixes: #2860

## Test Steps
Run collect logs with this patch and see the services there.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/collect_logs.feature
+++ b/features/collect_logs.feature
@@ -15,9 +15,11 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
         # So the -error suffix does not appear there.
         Then stdout matches regexp:
         """
+        apt-news.service.txt
         build.info
         cloud-id.txt
         cloud-init-journal.txt
+        esm-cache.service.txt
         jobs-status.json
         livepatch-status.txt-error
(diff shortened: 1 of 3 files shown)


features/collect_logs.feature
```diff
@@ -15,9 +15,11 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
...
         Then stdout matches regexp:
         """
+        apt-news.service.txt
         build.info
         cloud-id.txt
         cloud-init-journal.txt
+        esm-cache.service.txt
         jobs-status.json
         livepatch-status.txt-error
         pro-journal.txt
```
orndorffgrant, line 25: I don't think this is necessary because the extra units should go into pro-journal.txt, I'll run this test to confirm

orndorffgrant: oh nvm this is for the status output, not the logs

orndorffgrant (approved):

lucasmoura (approved):

Issue #2788, opened 2023-10-12
Bug: attach/enable language messages is inconsistent


orndorffgrant:
**Description of the bug**

Not all messages use the attach/detach/enable/disable verbage and associated object consistently.

**Expected behavior**

"attach" and "detach" are verbs that act on a "subscription". "enable" and "disable" are verbs that act on a "service". All messages should use these consistently.

**Current behavior**

Some messages use "subscription" and "service(s)" interchangeably or with a verb that isn't quite right.
Examples:
```
Your Ubuntu Pro subscription has EXPIRED
Renew your service at {url}
```
The above should use "subscription" instead of "service".
```
Failed to automatically attach to Ubuntu Pro services {num_attempts}
```
The above is referring to the "attach" operation, so should use "subscription" instead of "services".

**To Reproduce**

Look through uaclient/messages/__init__.py

**System information:**

- Ubuntu release: x, b, f, f, l , m
- Pro Client version: 30

mm-cr:
Hi, I'm Marlon. I would like to work on this issue. Could you assign it to me? Thanks!
@orndorffgrant

orndorffgrant:
Hello @mm-cr - sure! Let me know if you have any questions

PR #2868, opened 2023-12-03
Refactor messages wording

mm-cr:
## Why is this needed?
This change adjusts messages for consistency, using the 'attach/detach/enable/disable' correct usage for the 'subscription' and 'service' terms.

Fixes: #2788

## Checklist

 - [ ] Yes
 - [x] No @orndorffgrant could you please take a look?

### DIFF
--- a/debian/po/pt_BR.po
+++ b/debian/po/pt_BR.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-11-16 16:10-0300\n"
+"POT-Creation-Date: 2023-12-04 21:27-0600\n"
 "PO-Revision-Date: 2023-09-25 12:29-0400\n"
 "Last-Translator: Lucas Moura <lucas.moura@canonical.com>\n"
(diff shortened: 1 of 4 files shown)


orndorffgrant (commented):
These changes look great @mm-cr! Thanks for finding all of these instances of that issue.

Just a couple small things to get this PR ready to merge.
1. We have release-tooling that will grab fixed issues from commit messages, but it will only work if they're formatted in a particular way. Please change the line in your commit from `fixes issue #2788` to `Fixes: #2788`
2. This updates translatable messages, so please run `./tools/update-pos.sh` which will update the translation files.

Then I'll click the button to let CI run. Some of them may fail since they expect a branch rather than a fork, but I expect the important ones to pass.

After that, since some of these messages are already translated to pt_BR, we'll ask @lucasmoura or @renanrodrigo to update those translations.

mm-cr:
Ready! Please @orndorffgrant let me know if anything else is needed.

orndorffgrant (approved):
Thanks for the contribution @mm-cr! I'll go ahead and merge this.

@lucasmoura and @renanrodrigo how do you want to handle updating translations? We could do it right away in a separate PR, or we could wait until release time and then update all `fuzzy` translations before releasing WDYT?

orndorffgrant:
Created #2877

PR #2869, opened 2023-12-04
bring docs to docs-devel

renanrodrigo:


orndorffgrant:
I'm not sure why the spelling error wasn't caught in the past, but lets fix it in `docs` before merging docs into docs-devel here. I put up a PR to fix it #2870

Issue #2871, opened 2023-12-04
Bug: integration tests missing for mantic, noble


panlinux:
**Description of the bug**

The integration tests should also run on mantic and noble. Now they only run up to lunar:

https://github.com/canonical/ubuntu-pro-client/blob/main/.github/workflows/ci-integration.yaml#L68

**Expected behavior**

The integration tests should run on all curently supported Ubuntu releases, and the devel release (which is going to be the next LTS).

**Current behavior**

The integration tests do not run on mantic nor noble.

**To Reproduce**

Run integration tests on github.

**System information:**

- Pro Client version: main

**Additional context**

orndorffgrant:
@renanrodrigo and @lucasmoura reminded me that we purposefully don't run integration tests on the dev release on every PR. That is because the dev release changes frequently and frequently causes errors unrelated to pro-client. IOW, the signal-to-noise ratio is too low. Because of that, we won't add noble to the integration test runs on every PR until after Feature freeze, when noble is stable enough to be a reliable testing base.

We will go ahead and add the package build CI for noble though

PR #2872, opened 2023-12-04
release-update: drop lunar

orndorffgrant:
## Why is this needed?
We can stop testing on lunar since our next release will go out after
lunar EOL.

Also configures CI to build package for noble, and run integration tests
for on mantic.

Fixes: #2871 

## Test Steps
Check that CI builds for noble and runs tests on mantic. Those tests should pass.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/.github/workflows/ci-integration.yaml
+++ b/.github/workflows/ci-integration.yaml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        release: ['xenial', 'bionic', 'focal', 'jammy', 'lunar', 'mantic']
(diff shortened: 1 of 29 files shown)


renanrodrigo:
I am slightly, but very slightly, against removing the lunar tests from `main` before lunar gets actually EOD.
I know it is purely bureaucratic, as we are unlikely to break things in lunar at this point in time, and given we won't release there anymore...

orndorffgrant:
I think this is updated how we discussed @lucasmoura @renanrodrigo  - let's let CI run on it and see how it goes

renanrodrigo (approved):

lucasmoura (approved):

PR #2875, opened 2023-12-05
Fix/test fips messages

catmsred:
This changes updates the integration tests to include new messages about checking subscription, disabling livepatch, etc.

Also adds the new period to reboot message in integration tests.

## Why is this needed?
When testing new FIPS messages, I ran into failing integration tests that seemed unrelated to my changes.  This change brings upstream in line with the current message set specifically for FIPS messages (and a reboot period), but is not a comprehensive review of all message/integration test mismatches.  I tested against GCP.

## Checklist

 - [ ] I have updated or added any unit tests accordingly [N/A]
 - [ ] Changes here need to be documented, and this was done in:  [N/A]

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -266,7 +266,7 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
         When I verify that running `pro auto-attach` `with sudo` exits `2`
         Then stderr matches regexp:
             """
-            This machine is already attached to 'UA Client Test'
+            This machine is already attached to '.+'
             To use a different subscription first run: sudo pro detach.
             """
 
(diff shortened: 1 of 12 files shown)


renanrodrigo (approved):
LGTM
only `features/enable_fips_vm.feature:631` is missing the same kind of change - if we could change it there as well I'm good to go here

catmsred:
> LGTM only `features/enable_fips_vm.feature:631` is missing the same kind of change - if we could change it there as well I'm good to go here

Fixed

renanrodrigo:
CI errors are a) sad and b) completely unrelated to this change.

Issue #2600, opened 2023-05-30
Improve the default error message when  uncaught exceptions happen


renanrodrigo:
user logs should be in the local `.cache` dir instead - we should reflect that in the messaging.

## How to reproduce:
force an error for a command
run it as non-root
```
$ pro security-status
Unexpected error(s) occurred.
For more details, see the log: /var/log/ubuntu-advantage.log
To file a bug run: ubuntu-bug ubuntu-advantage-tools
```

a-dubs:
For clarification, should the error message be adaptable so that when the message is detected as being run as non-root, it changes the log location in the message accordingly?
@orndorffgrant

orndorffgrant:
@a-dubs yes that would be great!

PR #2883, opened 2023-12-06
Better unexpected exception msgs

a-dubs:
## Why is this needed?

The unexpected error message does not reflect accurate log location when run by non root, and also the error message obfuscates the original error message so the end user has no context for what actually went wrong. This PR fixes that. 

Changes `messages.UnexpectedError` from a `NamedMessage` to a `FormattedNamedMessage` so that it must be formatted with a string representing an exception/error message to give better context on what errors actually occurred.   

cli: improved cli/log message for unexpected errors

Fixes: #2600

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/log.py
+++ b/uaclient/log.py
@@ -70,7 +70,7 @@ def get_user_or_root_log_file_path() -> str:
     if util.we_are_currently_root():
         return UAConfig().log_file
     else:
-        return "~/.cache/" + defaults.USER_CACHE_SUBDIR + "/ubuntu-pro.log"
+        return get_user_log_file()
 
 
 def get_user_log_file() -> str:
--- a/uaclient/log.py
+++ b/uaclient/log.py
@@ -5,6 +5,7 @@ from collections import OrderedDict
 from typing import Any, Dict, List  # noqa: F401
 
 from uaclient import defaults, system, util
+from uaclient.config import UAConfig
 
 
 class RedactionFilter(logging.Filter):
@@ -59,6 +60,17 @@ class JsonArrayFormatter(logging.Formatter):
 
         local_log_record["extra"] = extra_message_dict
         return json.dumps(list(local_log_record.values()))
+
+
+def get_user_or_root_log_file_path() -> str:
+    """
+    Gets the correct log_file path,
+    adjusting for whether the user is root or not.
+    """
+    if util.we_are_currently_root():
+        return UAConfig().log_file
+    else:
+        return "~/.cache/" + defaults.USER_CACHE_SUBDIR + "/ubuntu-pro.log"
 
 
 def get_user_log_file() -> str:
--- a/uaclient/messages/__init__.py
+++ b/uaclient/messages/__init__.py
@@ -1553,13 +1553,14 @@ It is only possible to enable Anbox Cloud on a container using
 the --access-only flag.""",
 )
 
-UNEXPECTED_ERROR = NamedMessage(
+UNEXPECTED_ERROR = FormattedNamedMessage(
     "unexpected-error",
     t.gettext(
         """\
-Unexpected error(s) occurred.
-For more details, see the log: /var/log/ubuntu-advantage.log
-To file a bug run: ubuntu-bug ubuntu-advantage-tools"""
+Unexpected error(s) occurred: {error_msg}
+For more details, see the log: ~/.cache/ubuntu-pro/ubuntu-pro.log
+First, try searching online for solutions to this error.
+Otherwise, file a bug using the command: ubuntu-bug ubuntu-advantage-tools"""
     ),
 )
 
--- a/uaclient/messages/__init__.py
+++ b/uaclient/messages/__init__.py
@@ -1557,10 +1557,9 @@ UNEXPECTED_ERROR = FormattedNamedMessage(
     "unexpected-error",
     t.gettext(
         """\
-Unexpected error(s) occurred: {error_msg}
-For more details, see the log: ~/.cache/ubuntu-pro/ubuntu-pro.log
-First, try searching online for solutions to this error.
-Otherwise, file a bug using the command: ubuntu-bug ubuntu-advantage-tools"""
+An unexpected error occurred: {error_msg}
+For more details, see the log: {log_path}
+If you think this is a bug, please run: ubuntu-bug ubuntu-advantage-tools"""
     ),
 )
 
--- a/uaclient/tests/test_log.py
+++ b/uaclient/tests/test_log.py
@@ -162,39 +162,46 @@ class TestLoggerFormatter:
 
 
 class TestLogHelpers:
-    def test_get_user_or_root_log_file_path(self):
+    @pytest.mark.parametrize(
+        [
+            "we_are_currently_root",
+            "expected",
+        ],
+        [
+            (True, "cfg_log_file"),
+            (False, "user_log_file"),
+        ],
+    )
+    @mock.patch(
+        "uaclient.log.get_user_log_file",
+        return_value="user_log_file",
+    )
+    @mock.patch(
+        "uaclient.config.UAConfig.log_file",
+        new_callable=mock.PropertyMock,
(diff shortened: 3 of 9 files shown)


uaclient/util.py
```diff
@@ -477,3 +477,10 @@ def print_package_list(
         )
     )
     print("")
+
+
+def get_user_or_root_log_path() -> str:
+    if we_are_currently_root():
+        return "/var/log/ubuntu-advantage.log"
```
lucasmoura, line 484: We can extract that information directly through the `UAConfig` object. Even though users shouldn't change that option, it is still safer to consume that from an `UAConfig` objectt.

uaclient/util.py
```diff
@@ -477,3 +477,10 @@ def print_package_list(
...
+    else:
+        return "~/.cache/ubuntu-pro/ubuntu-pro.log"
```
lucasmoura, line 486: I think we can at least define part of this path on the `defaults` module and reuse it here.
This would even be useful for other places where we specify the non-root log path, like the `log` module

uaclient/messages/__init__.py
```diff
@@ -1534,13 +1534,14 @@ def __repr__(self):
...
 
-UNEXPECTED_ERROR = NamedMessage(
+UNEXPECTED_ERROR = FormattedNamedMessage(
     "unexpected-error",
     t.gettext(
         """\
-Unexpected error(s) occurred.
-For more details, see the log: /var/log/ubuntu-advantage.log
-To file a bug run: ubuntu-bug ubuntu-advantage-tools"""
+Unexpected error(s) occurred: {error_msg}
```
lucasmoura, line 1541: I liked the idea of asking the user to look for the error first before opening a bug, but I am still unsure if we should say that directly. What do you think @orndorffgrant @renanrodrigo ?

renanrodrigo: That is a suggested requirement of bug #2600 
I'd say it is good to tell users to investigate first, at least to say we tried hah
Not sure how effective it will really be, but I don't think it hurts.

uaclient/messages/__init__.py
```diff
@@ -1534,13 +1534,14 @@ def __repr__(self):
...
+For more details, see the log: {log_path}
+First, try searching online for solutions to this error.
```
renanrodrigo, line 1543: But if your point is that we could have more.. ahem, sensible wording, then yes we could

a-dubs: Brb gonna go run this through chatGPT to make it more friendly.

uaclient/util.py
a-dubs: Thank you for this suggestion! I was wondering if there was somewhere I could consume this info from, so thank you for pointing me in the right direction

uaclient/util.py
orndorffgrant: I think the log module should have a function that returns the non-root log file location - we can probably use that here

uaclient/messages/__init__.py
orndorffgrant: It could be as simple as adding an "if", e.g. "If you think this is a bug, please run: ubuntu-bug ubuntu-advantage-tools"

a-dubs: Putting this all together. How does the following sound?
```
Consider searching online for solutions to this error.
Or, if you think this is a bug, please run: ubuntu-bug ubuntu-advantage-tools
```

orndorffgrant: I meant that I think we don't need to explicitly ask the user to search online at all (even though I'm the one who suggested this). Seeing it written out feels a little too much like we're pushing too much responsibility on the user.
I do still think that the wording "To file a bug run: ubuntu-bug ubuntu-advantage-tools" or similar is too imperative though. So my idea is to only add a conditional ("if") to that suggestion.
I think the "if" suggests enough that it isn't necessarily a bug and could be worth searching online for.
So I think in total it would be:
```
Unexpected error(s) occurred: {error_msg}
For more details, see the log: {log_path}
If you think this is a bug, please run: ubuntu-bug ubuntu-advantage-tools
```

uaclient/messages/__init__.py
orndorffgrant, line 1541: While we're editing this - do we need the "(s)" after "error" here @lucasmoura @renanrodrigo? (Is there any chance we'll display more than one error?

uaclient/messages/__init__.py
orndorffgrant: Thank you! And sorry for flip-flopping on that idea :upside_down_face:

uaclient/messages/__init__.py
a-dubs: The only place I saw that used this error message and has multiple errors it reports is in contract.py. But it seems like it generates a separate error for each unexpected exception, instead of combining all the messages into one singular error?

renanrodrigo: yeah we can use the singular I guess
If many errors occur, then one error occurred

uaclient/messages/__init__.py
a-dubs: No worries! Thank you for re-evaluating this change's requirements :)

a-dubs:
The error message is being changed per the requirements of issue #2600. This means that translations will need to be updated for the new error message.

uaclient/log.py
```diff
@@ -61,6 +62,17 @@ def format(self, record: logging.LogRecord) -> str:
...
 
+def get_user_or_root_log_file_path() -> str:
+    """
+    Gets the correct log_file path,
+    adjusting for whether the user is root or not.
+    """
+    if util.we_are_currently_root():
+        return UAConfig().log_file
+    else:
+        return "~/.cache/" + defaults.USER_CACHE_SUBDIR + "/ubuntu-pro.log"
```
orndorffgrant, line 73: Since we only call this when non-root, let's re-use the function below
```suggestion
        return get_user_log_file()
```

uaclient/tests/test_log.py
```diff
@@ -158,3 +159,42 @@ def test_valid_json_output(
             assert val[6].get("key") == extra.get("key")
         else:
             assert 7 == len(val)
+
+
+class TestLogHelpers:
+    def test_get_user_or_root_log_file_path(self):
```
orndorffgrant, line 165: This looks like a good set of tests, but let's use parametrization and mock decorators to follow the patterns in the rest of the codebase. Something like this:
```suggestion
    @pytest.mark.parametrize(
        [
            "we_are_currently_root",
            "expected",
        ],
        [
            (True, "cfg_log_file"),
            (False, "user_log_file"),
        ]
    )
    @mock.patch("uaclient.util.we_are_currently_root")
    @mock.patch("uaclient.config.UAConfig.log_file", new_callable=mock.PropertyMock, return_value="cfg_log_file")
    @mock.patch("uaclient.log.get_user_log_file", return_value="user_log_file")
    def test_get_user_or_root_log_file_path(
        self,
        m_we_are_currently_root,
        m_cfg_log_file,
        m_get_user_log_file,
        we_are_currently_root,
        expected
    ):
```

orndorffgrant (commented):
Looking good! Some small nits

uaclient/tests/test_log.py
orndorffgrant: @a-dubs sorry, I just noticed a confusing error in my suggestion above - the mock decorators were in the wrong order. It should be fixed now

a-dubs: thank you! will implement these changes now. Will break the last portion of the test out into a different function because it does not make sense with the parameterization.

a-dubs: nevermind I'm just going to get rid of that test case because it is redundant.

renanrodrigo (approved):

orndorffgrant (approved):
This looks good - thank you!

PR #2886, opened 2023-12-11
ua security isolation for the esm-cache service

panlinux:
## Why is this needed?

## Checklist

 - [x] Changes here need to be documented, and this was done in:  TBD soon, just a quick update to existing docs on how to disable apparmor and security isolations for this new service, but there are no functional changes.

 - [x] Yes
 - [ ] No

I pinged #security and they added some comments already which I responded to.

### DIFF
--- a/debian/apparmor/ubuntu_pro_esm_cache.jinja2
+++ b/debian/apparmor/ubuntu_pro_esm_cache.jinja2
@@ -1,0 +1,276 @@
+{% if ubuntu_codename not in ["xenial", "bionic", "focal"] %}
+abi <abi/3.0>,
+{% endif %}
+include <tunables/global>
+
+profile ubuntu_pro_esm_cache flags=(attach_disconnected) {
+  include <abstractions/base>
+  include <abstractions/nameservice>
+  include <abstractions/openssl>
+  include <abstractions/python>
+  include <abstractions/user-tmp>
+
+  capability chown,
+  capability dac_override,
+  capability dac_read_search,
+  capability fowner,
+  capability kill,
+  capability setgid,
+  capability setuid,
+
+  signal send set=int peer=ubuntu_pro_esm_cache//apt_methods,
+  signal send set=int peer=ubuntu_pro_esm_cache//apt_methods_gpgv,
+
+  /etc/apt/** r,
+  /etc/machine-id r,
+  /etc/ubuntu-advantage/uaclient.conf r,
+
+  /run/ubuntu-advantage/ rw,
+  /run/ubuntu-advantage/** rw,
+
+  /run/systemd/container/ r,
+  /run/systemd/container/** r,
+
+  /usr/bin/apt mrix,
+  /usr/bin/apt-cache mrix,
+  /usr/bin/ischroot mrix,
+  /usr/bin/python3.{1,}[0-9] mrix,
+  /usr/bin/uname mrix,
+
+  /usr/bin/cloud-id Cx -> cloud_id,
+  /{,usr/}bin/ps Cx -> ps,
+  /usr/bin/systemd-detect-virt Px -> ubuntu_pro_esm_cache_systemd_detect_virt,
+  /usr/bin/dpkg Cx -> dpkg,
+  /usr/bin/ubuntu-distro-info Cx -> ubuntu_distro_info,
+  /usr/lib/apt/methods/gpgv Cx -> apt_methods_gpgv,
+  /usr/lib/apt/methods/http Cx -> apt_methods,
+  /usr/lib/apt/methods/https Cx -> apt_methods,
+  /usr/lib/apt/methods/store Cx -> apt_methods,
+
+  /usr/share/dpkg/** r,
+  /usr/share/keyrings/* r,
+
+  /var/cache/apt/** rw,
+
+  /var/lib/apt/** r,
+  /var/lib/dpkg/** r,
+  /var/lib/ubuntu-advantage/** rwk,
+
+  /var/log/ubuntu-advantage.log rw,
+
+  @{PROC}/@{pid}/fd/ r,
+  @{PROC}/1/cgroup r,
+  @{PROC}/version_signature r,
+  @{PROC}/@{pid}/mountinfo r,
+  @{PROC}/@{pid}/status r,
+  @{PROC}/@{pid}/stat r,
+  @{PROC}/sys/kernel/osrelease r,
+
+{% if ubuntu_codename in ["bionic", "xenial"] %}
+  # see https://bugs.python.org/issue40501
+  /sbin/ldconfig rix,
+  /sbin/ldconfig.real rix,
+  @{PROC}/@{pid}/mounts r,
+  /usr/bin/@{multiarch}-gcc-* rix,
+  /usr/bin/@{multiarch}-ld.bfd rix,
+  /usr/lib/gcc/@{multiarch}/*/collect2 rix,
+  /usr/bin/@{multiarch}-objdump rix,
+{% endif %}
+
+  profile ps flags=(attach_disconnected) {
+    include <abstractions/base>
+    include <abstractions/nameservice>
+
+    capability sys_ptrace,
+    ptrace read,
+
+    /{,usr/}bin/ps mrix,
+
+    /dev/tty r,
+
+    @{PROC}/ r,
+    @{PROC}/@{pid}/** r,
+    @{PROC}/uptime r,
+    @{PROC}/sys/kernel/** r,
+  }
+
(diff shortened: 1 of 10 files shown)


features/environment.py
```diff
@@ -465,8 +466,24 @@ def after_scenario(context, scenario):
                 logging.warning("Failed to collect coverage")
 
 
+def _get_apparmor_logs_from_host():
+    # get apparmor DENIED messages from the host
+    with open("/var/log/syslog", "r") as syslog_fd:
```
setharnold, line 471: If `auditd` is running, apparmor DENIED messages will go to `/var/log/audit/audit.log`, not syslog. This looks like a test framework, so it's probably fine.

systemd/esm-cache.service
```diff
@@ -13,3 +13,13 @@ Description=Update the local ESM caches
...
+AppArmorProfile=-ubuntu_pro_esm_cache
+CapabilityBoundingSet=~CAP_SYS_ADMIN
+CapabilityBoundingSet=~CAP_NET_ADMIN
+CapabilityBoundingSet=~CAP_NET_BIND_SERVICE
+CapabilityBoundingSet=~CAP_SYS_PTRACE
+CapabilityBoundingSet=~CAP_NET_RAW
+PrivateTmp=true
+RestrictAddressFamilies=~AF_NETLINK
+RestrictAddressFamilies=~AF_PACKET
+
```
setharnold, line 25: This is awesome. If you've got more time and interest, maybe adding additional seccomp filters would be nice, too. I've got these on one of my services, it feels like a reasonable match for what I think these are doing:

```
SystemCallFilter=@basic-io @file-system @io-event @network-io @process @signal @sync @timer brk mprotect ioctl sysinfo uname
```

I can't recall the exact details but missing syscalls are probably logged to `dmesg` or `auditd` as they are encountered and denied.

setharnold (commented):
Nice; perhaps Georgia will have more thoughts, so I'm just marking this "comment". Thanks

features/environment.py
panlinux: Correct. I'm using this to easily get access to apparmor logs from the github runners. I don't know yet if I will leave it in here, or change it.

systemd/esm-cache.service
panlinux: Oh, I was trying with a whole bunch of restrictions before (see here: https://github.com/canonical/ubuntu-pro-client/blob/9e9df67e5a4e3c339caced0d6d19f44bc38799d8/systemd/apt-news.service#L26). But I encountered some difficulties:
- some tests started failing after those changes, and I couldn't figure out why
- unlike apparmor, there are no nice DENIED messages when something was blocked by these :( So I can only use trial and error (enable, run tests, check results), or some deduction. In both cases, it was taking way too long. I plan to return to these.

debian/apparmor/ubuntu_pro_esm_cache.jinja2
```diff
@@ -0,0 +1,289 @@
...
+
+  signal send set=int peer=ubuntu_pro_esm_cache//apt_methods,
+  signal send set=int peer=ubuntu_pro_esm_cache//apt_methods_gpgv,
+
+  /etc/apt/** r,
+  /etc/machine-id r,
+  /etc/ubuntu-advantage/uaclient.conf r,
+
+  /run/ubuntu-advantage/ rw,
+  /run/ubuntu-advantage/** rw,
```
gegarcia, line 32: these types of rules could be condensed into
```suggestion
  /run/ubuntu-advantage/{,**} rw,
```
it's just a suggestion, I understand it can be less legible

debian/apparmor/ubuntu_pro_esm_cache.jinja2
```diff
@@ -0,0 +1,289 @@
...
+    /sbin/ldconfig.real rix,
+    @{PROC}/@{pid}/mounts r,
+    /usr/bin/@{multiarch}-gcc-* rix,
+    /usr/bin/@{multiarch}-ld.bfd rix,
+    /usr/lib/gcc/@{multiarch}/*/collect2 rix,
+    /usr/bin/@{multiarch}-objdump rix,
+
+    /etc/lsb-release r,
+    @{PROC}/cmdline r,
+    /bin/dash mrix,
```
gegarcia, line 150: could it change based on the system configs?

gegarcia: ah, just saw Seth's comment that this is a test framework, so I guess not

panlinux: What do you mean? The part Seth commented on (if I understood it correctly) was the python code in features/environment.py, that is a test framework. But this profile is for the "real thing": ubuntu-pro-client (a deb) installed on ubuntu releases since xenial.

gegarcia: my bad, I misunderstood. I guess the original comment is still valid then, could another shell be executed instead of dash?

panlinux: The exact sequence of calls when I run `cloud-id` on bionic is:
```
TIME(s) PCOMM            PID    PPID   RET ARGS
6.063   cloud-id         3096   3080     0 /usr/bin/cloud-id
6.166   ldconfig         3097   3096     0 /sbin/ldconfig -p
6.166   ldconfig.real    3097   3096     0 /sbin/ldconfig.real -p
6.177   ldconfig         3098   3096     0 /sbin/ldconfig -p
6.177   ldconfig.real    3098   3096     0 /sbin/ldconfig.real -p
6.201   sh               3099   3096     0 /bin/sh -c uname -p 2> /dev/null
6.202   uname            3100   3099     0 /bin/uname -p
6.242   systemd-detect-  3101   3096     0 /usr/bin/systemd-detect-virt --quiet --container
```
The reason that ended up calling `/bin/dash` is because our `/bin/sh` is a symlink to `/bin/dash`. I have never seen it be anything else.

And the reason `/bin/sh` was called is because it's the shebang of `/sbin/ldconfig`:
```terminal
root@b-pro:~# cat /sbin/ldconfig
#!/bin/sh

if  test $# = 0                                                 \
    && test x"$LDCONFIG_NOTRIGGER" = x                          \
 && test x"$DPKG_MAINTSCRIPT_PACKAGE" != x                      \
 && dpkg-trigger --check-supported 2>/dev/null
then
        if dpkg-trigger --no-await ldconfig; then
                if test x"$LDCONFIG_TRIGGER_DEBUG" != x; then
                        echo "ldconfig: wrapper deferring update (trigger activated)"
                fi
                exit 0
        fi
fi

exec /sbin/ldconfig.real "$@"
```
I don't think another shell would be called, but is there a more general abstraction to allow "all"  shells that I should perhaps consider?

orndorffgrant:
@panlinux Do you want/expect this to be ready for inclusion in ubuntu-pro-client 32? If so, we'd want to merge it this week.

panlinux:
> @panlinux Do you want/expect this to be ready for inclusion in ubuntu-pro-client 32? If so, we'd want to merge it this week.

Yes, that's why I requested reviews

debian/apparmor/ubuntu_pro_esm_cache.jinja2
gegarcia: sorry I missed this notification. yeah, you're right, I don't think any other shell would be called. thanks for checking and for the detailed explanation.

orndorffgrant (approved):
Thanks for the contribution and the explanatory comments!

panlinux:
I pushed a missing pair of `include` statements for local apparmor config snippets. This is standard ubuntu/dh_apparmor behavior, and allows admins to make certain local changes to apparmor profiles without touching the main apparmor file and thus creating a dpkg config prompt in the next upgrade. The files in /etc/apparmor.d/local/ are automatically created at package install by dh_apparmor, but the main profile needs to include them, and that's what I just committed.

panlinux:
Hm, ok, I can't do that here because we check the apparmor profile during build time, and there it expects to find /etc/apparmor.d/local/* already populated, which is obviously not the case in the build environment:
```
[pid  2813] openat(AT_FDCWD, "/etc/apparmor.d/local/ubuntu_pro_apt_news", O_RDONLY) = -1 ENOENT (No such file or directory)
```
I'll revert that commit and live with this bug for now, and ask for guidance from the security team.

renanrodrigo (approved):
LGTM
One of the new VM tests is failing, but that is a test problem - we will update it in a subsequent PR.
This test is not part of CI, so it will not interfere with anything we are doing here.

PR #2887, opened 2023-12-11
Binary package rename (SC-1623)

orndorffgrant:
## Why is this needed?

This PR renames the binary packages created from ubuntu-advantage-tools as agreed. The new package names are "ubuntu-pro-client" and "ubuntu-pro-image-auto-attach".

This PR is packaging heavy and deserves extra scrutiny there.

This PR also adjusts our tests to use the new packages.

A subsequent PR will also include a manual test script for testing the transition

## Test Steps
Test adding this ppa and upgrading: `ppa:orndorffgrant/proclient-rename-test`

A test that can only be done with debs and not with a proper apt source of the full ubuntu-advantage-tools set of binaries:
1. launch a lxd container
2. copy over just ubuntu-pro-client.deb
3. `apt install ubuntu-pro-client.deb`
That should fail saying that ubuntu-advantage-tools >=31 is also required

## Checklist

 - [ ] Yes
 - [ ] No


orndorffgrant:
>     * [x]  Existing messages are being changed.
>       
>       * ⚠️ Please add a comment with justification of why messages are being altered.
>       * If the changes are trivial (e.g. a typo fix), then translations must be preserved.
>       * If the changes are substantial, then we will ask translators to take a look and update translations if they have time, but it will not block this PR.

Only the package name inside a message is being updated. Any translated messages that include it will be updated in this PR.

debian/control
```diff
@@ -32,6 +32,15 @@ Vcs-Browser: https://github.com/canonical/ubuntu-pro-client
 Rules-Requires-Root: no
 
 Package: ubuntu-advantage-tools
+Depends: ubuntu-pro-client, ${misc:Depends}
```
paride, line 35: We probably don't need `${misc:Depends}` here (but it's also not harmful in any way).

debian/control
```diff
@@ -40,27 +49,39 @@ Depends: ${misc:Depends},
...
+ Ubuntu Pro is a suite of additional services provided by Canonical on
+ top of Ubuntu. Whether you’re an enterprise customer deploying systems
+ at scale or want security patching for your personal Ubuntu LTS
+ at home, the Ubuntu Pro Client (pro) is the command-line tool that
+ will help you manage the services you need.
 
 Package: ubuntu-pro-client-l10n
 Architecture: any
-Depends: ${misc:Depends}, ubuntu-advantage-tools (>=30~)
+Depends: ${misc:Depends}, ubuntu-pro-client (>=31~)
```
paride, line 64: We should be able to automate these versioned dependencies by using `${binary:Version}`, see the deb-substvars(5) manpage.

lucasmoura (approved):
As @paride mentioned, we still need to explore `${binary:Version}`, but besides that point, I can confirm that this setup works and we don't detach when upgrading to the new binary package

renanrodrigo (approved):
LGTM overall, works on the local tests I did.
However, as I would not have the expertise for a final approval, I am binding this to Paride+SRU aproval of course :D

debian/control
orndorffgrant: I removed it but then lintian gives ` debhelper-but-no-misc-depends` as a warning so I'll add it back

features/steps/ubuntu_advantage_tools.py
```diff
@@ -312,3 +312,111 @@ def when_i_check_apt_cache_policy(context):
     for step in context.scenario.steps:
         if step.name == APT_POLICY_IS:
             step.text = context.process.stdout
+
+
+@when("I install transition package ubuntu-advantage-tools")
+def when_i_install_transition_uat(context, machine_name=SUT):
```
orndorffgrant, line 318: This is mostly a copy-past of `when_i_install_uat`, so I think we can refactor this to reuse some code.

Can you turn most of this into a step called "When I set up the source for ubuntu-advantage-tools" which would only add PPAs or set up -proposed as necessary (and not install anything yet). Then the next step in this new test would be "When I apt install `ubuntu-advantage-tools`". We could also reuse that function inside of `when_i_install_uat`

orndorffgrant: forgot to tag: @lucasmoura

orndorffgrant:
@paride @basak We are officially using this method for the binary package rename. Please review it here in this PR. When we are happy with it, I'll merge and add it on top of the release PR (#2913).

Also, please note that there is a product desire to totally remove the `ubuntu-advantage-tools` transition package in 26.04 onwards. Please take that future change into account when reviewing.

features/steps/ubuntu_advantage_tools.py
lucasmoura: That's a great idea. I have updated the code

debian/control
```diff
@@ -35,6 +35,15 @@ Vcs-Browser: https://github.com/canonical/ubuntu-pro-client
 Rules-Requires-Root: no
 
 Package: ubuntu-advantage-tools
+Depends: ubuntu-pro-client (= ${binary:Version}), ${misc:Depends}
```
paride, line 38: This Depends together with `Architecture: all` (which is correct for a transitional package) causes the following lintian error:

```
E: ubuntu-advantage-tools source: not-binnmuable-all-depends-any ubuntu-advantage-tools -> ubuntu-pro-client
```

(Run `lintian-explain-tags not-binnmuable-all-depends-any` for more info on it.) Now, this is not an issue on Ubuntu, because in Ubuntu we don't name binNMUs, however I think we should still avoid in in any case.

The easiest way in my opinion is to drop the versioned dependency altogether, we don't need to worry about an ubuntu-pro-client package being "old" because we're just introducing it. It will be the "real" package to have versioned dependencies as needed.

debian/control
```diff
@@ -43,27 +52,40 @@ Depends: ${misc:Depends},
...
- Ubuntu Pro is the professional package of tooling, technology
- and expertise from Canonical, helping organisations around the world
- manage their Ubuntu deployments.
- .
- Subscribers to Ubuntu Pro will find helpful tools for accessing
- services in this package.
+Breaks: ubuntu-advantage-tools (<<31~)
+Replaces: ubuntu-advantage-tools (<<31~)
+# IMPORTANT: ubuntu-pro-client does not "Provide" ubuntu-advantage-tools
+# TODO fill this in with rationale after testing.
```
paride, line 58: I think it's fine not to have a Provides for the replaced package if not needed, but if you have an explanation on why we don't want it (so we avoid adding it in the future), +1 for adding a comment on it under the ubuntu-pro-client binary package stanza.

paride (changes requested):
## On d/control:

The identical descriptions for the transitional packages make lintian slightly unhappy:

```
I: ubuntu-advantage-tools source: duplicate-long-description ubuntu-advantage-pro ubuntu-advantage-tools [debian/control]
I: ubuntu-advantage-tools source: duplicate-short-description ubuntu-advantage-pro ubuntu-advantage-tools [debian/control]
```

Consider overriding that lintian tag, or using slightly different descriptions (e.g. naming the package: `transitional dummy package for <package>`).

## On lintian

This one way I run lintian. From the unpackage source tree (the proposed branch in this case), I build the package creating a .changes file that contains both the binary packages and source package, e.g.

```
DEB_BUILD_OPTIONS=nocheck dpkg-buildpackage -sa -us -uc
```

That `nocheck` is to skip tests, useful to make iterations quicker. Then I run lintian like this:

```
lintian -EvIL +pedantic
```

to get all possible warnings. It will automatically find the relevant `../*.changes` file and check it both the binary packages and source package it refers.

Let's make sure that this branch does not introduce new warnings; or if it does and we are OK with those, let's override them.

paride:
@orndorffgrant added a few review comments, but it should be all small stuff, the general approach looks good.

orndorffgrant:
Thank you @paride! I believe I have addressed all of your comments in the latest commits

paride (approved):
Thanks for addressing all the review comments, this LGTM!

debian/control
```diff
@@ -56,7 +56,21 @@ Recommends: ubuntu-pro-client-l10n
...
 # IMPORTANT: ubuntu-pro-client does not "Provide" ubuntu-advantage-tools
-# TODO fill this in with rationale after testing.
+# At the time of the rename, existing releases with ubuntu-advantage-tools
+# include xenial, bionic, focal, jammy, and mantic.
+# We cannot allow ubuntu-advantage-tools <<31 to be removed on those releases,
+# because it could break ESM (remove sources.list entries in prerm).
+# On those releases, ubuntu-minimal (and ubuntu-cloud-minimal) Depends on
+# ubuntu-advantage-tools, which prevents it from being removed without also
+# removing ubuntu-minimal or ubuntu-cloud-minimal. We consider that to be a
+# sufficient # warning to users that removing ubuntu-advantage-tools is not
```
basak, line 66: (Very very minor) wrap error

basak (approved):
As discussed, this is very complex and so I fear that it will go wrong for some edge case. We will need to rely on good testing, and I understand that's already been discussed and there's a good plan for this. I have not been able to spot any significant issues. A couple of very minor comments are inline.

debian/rules
```diff
@@ -39,50 +58,54 @@ endif
 endif
 
 override_dh_gencontrol:
-	echo extra:Depends=$(APT_PKG_DEPS) $(DISTRO_INFO_DEPS) >> debian/ubuntu-advantage-tools.substvars
+	echo extra:Depends=$(APT_PKG_DEPS) $(DISTRO_INFO_DEPS) $(SELF_DEPS) >> debian/ubuntu-pro-client.substvars
```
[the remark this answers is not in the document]
orndorffgrant: Yep the end result of this PR is currently no change to that part of the line, but it should have quotes. I'll go ahead and add them in a new commit on top.

debian/control
orndorffgrant: Whoops! I'll fix it

PR #2888, opened 2023-12-11
Standardize apt/apt-get calls in tests and messages

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because now we use `apt` instead of `apt-get` to simulate user interaction with the client in the acceptance tests.
We also don't suggest that people run `apt-get` in our hints anymore.

## Test Steps
mostly a refactor, integration tests cover the changes messages

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo:
>     * [x]  Existing messages are being changed.
>       
>       * ⚠️ Please add a comment with justification of why messages are being altered.

That's the spirit of the first commit - fix messages.

>       * If the changes are trivial (e.g. a typo fix), then translations must be preserved.

They are

features/steps/packages.py
```diff
@@ -3,19 +3,91 @@
...
+@when("I install `{package_names}`")
+@when("I install `{package_names}` on the `{machine_name}` machine")
+def when_i_apt_install(context, package_names, machine_name=SUT):
+    when_i_retry_run_command(
         context,
         " ".join(
             [
                 "DEBIAN_FRONTEND=noninteractive",
-                "apt-get",
+                "apt",
```
lucasmoura, line 90: Should we also add  `-o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" ` here ?

lucasmoura (approved):
Just a question on the install step, but let me know if that is not needed

features/steps/packages.py
renanrodrigo: We didn't in any of the steps in the current tests (thus it is not here) but I think it does not hurt. Will add.

orndorffgrant: Those are already here :)

features/steps/packages.py
```diff
@@ -3,19 +3,91 @@
...
+        cmd_list.append("-o APT::Get::Always-Include-Phased-Updates=true")
+
+    when_i_run_command(
+        context,
+        " ".join(cmd_list),
+        "with sudo",
+    )
+
+
+@when("I install `{package_names}`")
```
orndorffgrant, line 82: I had it as "I apt install" to differentiate it from a snap install, do you think that is a useful differentiation?

features/steps/packages.py
```diff
@@ -37,6 +37,39 @@ def when_i_apt_update(context, machine_name=SUT):
     )
 
 
+@when("I perform a {dry} run of {command} upgrade")
+@when("I upgrade all apt packages in the system")
+@when("I upgrade all apt packages in the system using {command}")
+@when("I upgrade all apt packages in the system including {phased} updates")
```
orndorffgrant, line 52: Do we ever not want to include phased updates?

orndorffgrant:
I'm just a little hesitant to endorse the step names, mostly because I selfishly don't want to memorize new phrases for these :upside_down_face: 
We could name them more closely to the commands like:
```
When I apt upgrade
When I apt dist-upgrade
When I apt remove `package1 package2`
```
Let's talk about it in 2024 :)

features/steps/packages.py
renanrodrigo: I would say the standard is to not include them, so we can consider acceptance tests to do what users do.
We install phased updates when testing our apt messages, for practicality purposes.

features/steps/packages.py
renanrodrigo: AHA

features/steps/packages.py
renanrodrigo: yes

orndorffgrant (approved):
ci fails unrelated

PR #2889, opened 2023-12-12
Replace status tests

lucasmoura:
## Why is this needed?
Some of our integration tests are relying on the output of `pro status` to access if the machine is enabled or a specific service is enabled. We already have API features that display which services are enabled. Due to that, we are now updating our tests to rely on the API for this kind of information.

Note that not all `pro status` calls were removed. They were kept if they are evaluating that service status is `n/a` or to see that a change in the system has modified the service description in status (i.e. Livepatch support for an unsupported kernel) 

This PR also includes fixes for the `enabled_services` endpoint:

* `usg` service not being reported there
* services with `warning` status were not show as enabled on the API

## Test Steps
Guarantee that the modified tests are working as expected and the proposed changes are not modifying the original test flow

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/proxy_config.feature
+++ b/features/proxy_config.feature
@@ -4,12 +4,12 @@ Feature: Proxy configuration
     @slow
     Scenario Outline: Attach command when proxy is configured for uaclient
         Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
-        Given a `focal` machine named `proxy`
+        Given a `focal` `lxd-container` machine named `proxy`
         When I run `apt install squid -y` `with sudo` on the `proxy` machine
         And I add this text on `/etc/squid/squid.conf` on `proxy` above `http_access deny all`:
-            """
-            dns_v4_first on\nacl all src 0.0.0.0\/0\nhttp_access allow all
-            """
+        """
+        dns_v4_first on\nacl all src 0.0.0.0\/0\nhttp_access allow all
+        """
         And I run `systemctl restart squid.service` `with sudo` on the `proxy` machine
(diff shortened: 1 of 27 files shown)


features/proxy_config.feature
```diff
@@ -535,15 +533,15 @@ Feature: Proxy configuration
 
     @slow
     Scenario Outline: Attach command when authenticated proxy is configured manually for uaclient
-        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
-        Given a `focal` machine named `proxy`
+        Given a `<release>` `lxd-container` machine with ubuntu-advantage-tools installed
```
orndorffgrant, line 536: I think this line should stay as it was
```suggestion
        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
```

orndorffgrant (approved):
This is a great refactor and improvement! thank you!
Just one small nit then LGTM

features/proxy_config.feature
lucasmoura: @orndorffgrant My rationale for this is if we want to expand this test in the future for VMs, the proxy machine should still be a `lxd-container`. But let me know if you disagree

orndorffgrant: @lucasmoura I totally agree with the change to the `Given` line setting up the proxy, but the "machine with ubuntu-advantage-tools` should be <machine_type> still, right?

lucasmoura: Ohhh good catch, I missed that I changed that on both places. Fixing it now

lucasmoura:
@orndorffgrant done

lucasmoura:
@orndorffgrant I have addressed the problematic mantic test

PR #2890, opened 2023-12-12
feat: warn/confirm with user if enabling fips downgrades the kernel

catmsred:
This change adds a new message prompting the user to confirm if enabling fips downgrades the kernel.

## Why is this needed?
There have been reports of people enabling FIPS (not realizing it will downgrade their kernel), and having their hardware fail as it needed the new kernel. This change adds an additional warning/prompt if a FIPS enablement is determined to downgrade the kernel.

Jira card: SC-1534

## Test Steps
On a pro (but not FIPS) VM, enable FIPS and you should see the new messages.

## Checklist

 - [ ] Changes here need to be documented, and this was done in: [N/A]

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/enable_fips_pro.feature
+++ b/features/enable_fips_pro.feature
@@ -12,11 +12,15 @@ Feature: FIPS enablement in PRO cloud based machines
         Then I verify that `fips` is disabled
         And I verify that `fips-updates` is disabled
         When I run `pro enable <fips-service> --assume-yes` with sudo
-        Then stdout contains substring:
+        Then stdout matches regexp:
         """
         Updating <fips-name> package lists
         Installing <fips-name> packages
-        Updating standard Ubuntu package lists
+        This will downgrade the kernel from .+ to .+\.
+        Warning: Downgrading the kernel may cause hardware failures.  Please ensure the
+                 hardware is compatible with the new kernel version before proceeding.
+
+        Updating standard Ubuntu package lists(\n.*)?
         <fips-name> enabled
         A reboot is required to complete install
         """
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -1,7 +1,8 @@
 import logging
 import os
+import re
 from itertools import groupby
-from typing import List, Optional, Tuple  # noqa: F401
+from typing import Callable, List, Optional, Tuple, Union  # noqa: F401
 
 from uaclient import apt, event_logger, exceptions, messages, system, util
 from uaclient.clouds.identity import NoCloudTypeReason, get_cloud_type
@@ -154,6 +155,56 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
 
         return FIPS_CONDITIONAL_PACKAGES.get(series, [])
 
+    def prompt_if_kernel_downgrade(
+        self,
+        assume_yes: bool = False,
+    ) -> bool:
+        """Check if installing a FIPS kernel will downgrade the kernel
+        and prompt for confirmation if it will.
+        """
+        # Prior to installing packages, check if the kernel is being downgraded
+        # and if so verify that the user wants to continue
+        our_full_kernel_str = (
+            system.get_kernel_info().proc_version_signature_version
+        )
+        if our_full_kernel_str is None:
+            LOG.warning("Cannot gather kernel information")
+            return False
+        our_m = re.search(
+            r"(?P<kernel_version>\d+\.\d+\.\d+)", our_full_kernel_str
+        )
+        fips_kernel_version_str = apt.get_pkg_candidate_version("linux-fips")
+        if our_m is not None and fips_kernel_version_str is not None:
+            our_kernel_version_str = our_m.group("kernel_version")
+            LOG.debug(
+                "Kernel information: cur='%s' and fips='%s'",
+                our_full_kernel_str,
+                fips_kernel_version_str,
+            )
+            if (
+                apt.version_compare(
+                    fips_kernel_version_str, our_kernel_version_str
+                )
+                < 0
+            ):
+                event.info(
+                    messages.KERNEL_DOWNGRADE_WARNING.format(
+                        current_version=our_kernel_version_str,
+                        new_version=fips_kernel_version_str,
+                    )
+                )
+                return util.prompt_for_confirmation(
+                    msg=messages.PROMPT_YES_NO,
+                    assume_yes=self.assume_yes,
+                )
+        else:
+            LOG.warning(
+                "Cannot gather kernel information for '%s' and '%s'",
+                our_full_kernel_str,
+                fips_kernel_version_str,
+            )
+        return True
+
     def install_packages(
         self,
         package_list: Optional[List[str]] = None,
@@ -167,6 +218,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         :param cleanup_on_failure: Cleanup apt files if apt install fails.
         :param verbose: If true, print messages to stdout
         """
+
         if verbose:
             event.info(
                 messages.INSTALLING_SERVICE_PACKAGES.format(title=self.title)
@@ -195,8 +247,14 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
 
         for pkg in desired_packages:
             try:
-                super().install_packages(
-                    package_list=[pkg], cleanup_on_failure=False, verbose=False
+                apt.run_apt_install_command(
+                    packages=[pkg],
+                    override_env_vars={"DEBIAN_FRONTEND": "noninteractive"},
+                    apt_options=[
+                        "--allow-downgrades",
+                        '-o Dpkg::Options::="--force-confdef"',
+                        '-o Dpkg::Options::="--force-confold"',
+                    ],
                 )
             except exceptions.UbuntuProError:
                 event.info(
@@ -478,6 +536,14 @@ class FIPSEntitlement(FIPSCommonEntitlement):
                     {"msg": pre_enable_prompt, "assume_yes": self.assume_yes},
                 )
             ],
+            "pre_install": [
+                (
+                    self.prompt_if_kernel_downgrade,
+                    {
+                        "assume_yes": self.assume_yes,
+                    },
+                )
+            ],
             "post_enable": post_enable,
             "pre_disable": pre_disable,
         }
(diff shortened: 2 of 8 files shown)


uaclient/entitlements/fips.py
```diff
@@ -167,6 +170,46 @@ def install_packages(
...
+        our_full_kernel_str = (
+            system.get_kernel_info().proc_version_signature_version
+        )
+        if our_full_kernel_str is None:
+            LOG.warning("Cannot gather kernel information")
+            return
+        our_m = re.search(
+            r"(?P<kernel_version>\d+\.\d+\.\d+)", our_full_kernel_str
+        )
+        fips_kernel_policy = apt.get_apt_cache_policy_for_package("linux-fips")
```
lucasmoura, line 185: Maybe I am missing something here, but can't we use `apt.get_pkg_version` here ?

uaclient/entitlements/fips.py
```diff
@@ -167,6 +170,46 @@ def install_packages(
...
+        fips_m = re.search(
+            r"Candidate: (?P<kernel_version>\d+\.\d+\.\d+)", fips_kernel_policy
+        )
+        if fips_m is not None and our_m is not None:
```
lucasmoura, line 189: If we extract both version of the fips packages directly, can't we compare then with the `apt.version_compare` ?

uaclient/entitlements/fips.py
```diff
@@ -167,6 +170,46 @@ def install_packages(
...
+            our_version = pkg_resources.parse_version(our_kernel_version_str)
+            fips_version = pkg_resources.parse_version(fips_kernel_version_str)
+            if fips_version < our_version:
+                event.info(
+                    messages.KERNEL_DOWNGRADE_WARNING.format(
+                        current_version=our_kernel_version_str,
+                        new_version=fips_kernel_version_str,
+                    )
+                )
+                if not util.prompt_for_confirmation(
```
lucasmoura, line 201: Maybe we could move this logic into the `messaging` module, just to be consistent with the other prompts we have for `fips`.

My idea is:

1. Move this logic into a separate function that will only check if a kernel downgrade will happen
2. Call this function on the messaging module and if it returns `True`, then we create the promp as a `pre_enable` step.

But let me know if you disagree or if I am missing something about the implementation

uaclient/entitlements/fips.py
catmsred: Does `apt.get_pkg_version` return the currently installed version on the to-be-installed version?  I would be much happier to use something cleaner if it exists.

uaclient/entitlements/fips.py
catmsred: I noticed that this was the standard paradigm when I originally drafted it.  However, for the apt-cache to return the correct fips kernel version, `self.setup_apt_config(silent=silent)` has to run so the kernel downgrade logic effectively needs to fit in the middle of `RepoEntitlement._perform_enable`.  Open to suggestions as to whether this should use a larger refactor or if we want to keep the downgrade checking logic in `FIPSCommonEntitlement.install_packages`.

uaclient/entitlements/fips.py
renanrodrigo: `apt.get_pkg_version` -> installed version
`apt.get_pkg_candidate_version` -> to-be-installed (candidate) version

uaclient/entitlements/fips.py
lucasmoura: That's a fair point and I agree that we can't use `pre_enable` for that, but we do have the `pre_install` check inside [install_packages](https://github.com/canonical/ubuntu-pro-client/blob/main/uaclient/entitlements/repo.py#L430). At this point, `setup_apt_config` should have already run.

The messaging method should handle this correctly, because we construct the messages dicts everytime we call it. Maybe this is something we could change in the future, but I think right now it should work.

But please let me know if I am still missing something here

catmsred: I think pre_install should work -- let me try rewriting in that format and make sure it works

catmsred:
Integration test failures are just proxy tests which I don't think have anything to do with this code

uaclient/entitlements/fips.py
```diff
@@ -195,8 +247,14 @@ def install_packages(
 
         for pkg in desired_packages:
             try:
-                super().install_packages(
-                    package_list=[pkg], cleanup_on_failure=False, verbose=False
+                apt.run_apt_install_command(
```
lucasmoura, line 250: Just double checking, why are we changing this ?

uaclient/entitlements/fips.py
```diff
@@ -154,6 +155,56 @@ def conditional_packages(self):
...
+            LOG.warning("Cannot gather kernel information")
+            return False
+        our_m = re.search(
+            r"(?P<kernel_version>\d+\.\d+\.\d+)", our_full_kernel_str
+        )
+        fips_kernel_version_str = apt.get_pkg_candidate_version("linux-fips")
+        if our_m is not None and fips_kernel_version_str is not None:
+            our_kernel_version_str = our_m.group("kernel_version")
+            LOG.debug(
+                "Kernel information: cur={cur} and fips={fips}".format(
```
lucasmoura, line 180: For the logs, we are using the string interpolation strategy as you can see here:
https://github.com/canonical/ubuntu-pro-client/blob/main/uaclient/entitlements/base.py#L349

Note there is nothing wrong with the logs you have written, but I think we just change them just be consistent with the other logs we have already defined

features/enable_fips_pro.feature
```diff
@@ -12,11 +12,15 @@ Feature: FIPS enablement in PRO cloud based machines
         Then I verify that `fips` is disabled
         And I verify that `fips-updates` is disabled
         When I run `pro enable <fips-service> --assume-yes` with sudo
-        Then stdout contains substring:
+        Then stdout matches regexp:
         """
         Updating <fips-name> package lists
         Installing <fips-name> packages
-        Updating standard Ubuntu package lists
+        This will downgrade the kernel from .+ to .+\.
```
lucasmoura, line 19: Is this test supposed to pass on all releases specified in this `Scenario Outline`  ? I have tested the `bionic` `aws.pro` and the test didn't work

uaclient/entitlements/fips.py
catmsred: Every time super().install_packages is called, the kernel downgrade warning is sent.  In my testing, we generally saw it three times.  Grant suggested that this call should properly be `run_apt_install_command` instead.

features/enable_fips_pro.feature
catmsred: Where is this failing for you?  It worked for all the iterations I ran of it and is currently passing in CI -- the aws.pro bionic failure is only for an unrelated proxy test.

uaclient/entitlements/fips.py
catmsred: I will make the update to use that format but I will note that string interpolation can be significantly harder to read, especially when there are multiple variables to consider.  I would like to see us migrate to format strings e.g. `f"{Kernel information: {our_full_kernel_str=} and {fips_kernel_version_str}"` but that would be a much larger discussion/refactor and code consistency is important.

features/enable_fips_pro.feature
lucasmoura: That's weird. I am getting the following error on bionic:
```
      Assertion Failed: Expected to match regexp:                                                                                                                                                                  
        Updating FIPS package lists                                                                                                                                                                                
        Installing FIPS packages                                                                                                                                                                                   
        This will downgrade the kernel from .+ to .+\.                                                                                                                                                             
        Warning: Downgrading the kernel may cause hardware failures.  Please ensure the                                                                                                                            
                 hardware is compatible with the new kernel version before proceeding.                                                                                                                             
                                                                                                                                                                                                                   
        Updating standard Ubuntu package lists(\n.*)?                                                                                                                                                              
        FIPS enabled                                                                                                                                                                                               
        A reboot is required to complete install                                                                                                                                                                   
      But got:                                                                                                                                                                                                     
        Disabling incompatible service: Livepatch                                                                                                                                                                  
        Updating FIPS package lists                                                                                                                                                                                
        Installing FIPS packages                                                                                                                                                                                   
        Updating standard Ubuntu package lists                                                                                                                                                                     
        FIPS enabled                                                                                                                                                                                               
        A reboot is required to complete install.  
```

But I haven't checked the other releases yet

catmsred: I think I see the issue.  Checking the AWS listings it looks like Canonical only has 16.04 Pro FIPS (not plain pro) so the kernel will not be being downgraded in that instance.  I went back to check the logs in the CI run and it seems this test is getting skipped there.  Not sure why that is happening when it runs locally.

renanrodrigo (approved):
LGTM
All CI failures caused by flaky network connections, and not affected by this PR at all.

features/enable_fips_pro.feature
catmsred: Hi, thanks for the check!  I have since fixed the downgrade conditional -- let me know if you're still seeing unexpected failures?

lucasmoura (commented):
@catmsred I can confirm the tests are now passing for me. Just one more addition to this PR and I will approve it

uaclient/entitlements/fips.py
```diff
@@ -154,6 +155,56 @@ def conditional_packages(self):
 
         return FIPS_CONDITIONAL_PACKAGES.get(series, [])
 
+    def prompt_if_kernel_downgrade(
```
lucasmoura, line 158: I think we just need some unittest for this function now

uaclient/entitlements/tests/test_fips.py
```diff
@@ -109,6 +109,55 @@ def test_conditional_packages(
...
+        self,
+        m_prompt_for_confirmation,
+        m_pkg_candidate_version,
+        fips_version,
+        assume_yes,
+        expected_continue,
+        entitlement,
+    ):
+        """Test kernel downgrades block install if user denies prompt"""
+        m_prompt_for_confirmation = util.prompt_for_confirmation
```
lucasmoura, line 149: I think I am missing something. Why do we need this line here ?

lucasmoura (approved):
Thanks for all the work on this issue @catmsred. I have just one tiny nit here, but let me know if I am missing something

uaclient/entitlements/tests/test_fips.py
catmsred: We do not!  Good catch, that was left over from my experiments for the best way to handle this mock.  I will clean it up and then we should be good to merge.

orndorffgrant:
CI fails are flakes + two approvals = merging
Thank you @catmsred!

PR #2895, opened 2023-12-13
fix: ignore LSNs when fetching CVE related USNs

lucasmoura:
## Why is this needed?
When fixing a CVE, we also fetch the related USNs tied to it. When doing that, we need to discard LSNs since they are not supported at the moment.

## Test Steps
Run the modified integration test to assert that LSNs are being ignore now when fixing CVEs

## Checklist

 - [ ] Yes
 - [x] No


orndorffgrant (approved):

renanrodrigo (approved):

renanrodrigo:
CI Errors are unrelated

PR #2896, opened 2023-12-14
Run fix when esm cache not updated

lucasmoura:
## Why is this needed?
When running a `pro fix` command, we might need to access information on the ESM cache. If the cache is not yet updated, we will not show an accurate representation of the fix in the system.

To avoid that, we are now detecting if we need to update the ESM cache during the plan endpoint and if we can't update the ESM cache (i.e users running as non-root), we will now inform the user on the cli command and also warn them on the output of the plan endpoint API.

## Test Steps
Verify that the modified test is working as expected

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/fix.feature
+++ b/features/fix.feature
@@ -203,7 +203,82 @@ Feature: Ua fix command behaviour
         """
         Error: USN-12345-12 not found.
         """
+        When I verify that running `pro fix USN-5079-2 --dry-run` `as non-root` exits `1`
+        Then stdout matches regexp:
+        """
+        .*WARNING: The option --dry-run is being used.
+        No packages will be installed when running this command..*
+        USN-5079-2: curl vulnerabilities
+        Associated CVEs:
+         - https://ubuntu.com/security/CVE-2021-22946
+         - https://ubuntu.com/security/CVE-2021-22947
+
+        Fixing requested USN-5079-2
+        1 affected source package is installed: curl
+
+        .*WARNING: Unable to update ESM cache when running as non-root,
+        please run sudo apt update and try again if packages cannot be found..*
+
+        \(1/1\) curl:
+        A fix is available in Ubuntu Pro: ESM Infra.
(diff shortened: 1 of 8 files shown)


uaclient/api/u/pro/security/fix/_common/plan/v1.py
```diff
@@ -799,6 +829,26 @@ def get_pocket_short_name(pocket: str):
         return pocket
 
 
+def _should_update_esm_cache(
+    check_esm_cache: bool,
+    esm_cache_updated: bool,
+    cfg: UAConfig,
+):
```
renanrodrigo, line 836: could be explicit that this function returns a `bool`, and add a `return False` after the `if` ends

renanrodrigo (approved):

uaclient/api/u/pro/security/fix/_common/plan/v1.py
orndorffgrant: I agree

features/fix.feature
```diff
@@ -203,7 +203,82 @@ Feature: Ua fix command behaviour
         """
```
orndorffgrant, line 203: Not for this PR, but I'm noticing that this scenario is hundreds of lines long. We should break it up in the future

orndorffgrant (approved):
After @renanrodrigo's comment is addressed I think this is good to go

uaclient/api/u/pro/security/fix/_common/plan/v1.py
lucasmoura: Good catch, I am updating the code

features/fix.feature
lucasmoura: Yep, we also need to reevaluate if we need all of those fix scenarios, because maybe we don't

lucasmoura:

PR #2901, opened 2024-01-02
tests: mock terminal size for terminal-size-dependent tests

orndorffgrant:
## Why is this needed?
In mantic+ the help output is terminal-size dependent even inside pytest. This mocks the terminal size to 80 columns which is what our tests have always been assuming.

## Test Steps

unit tests should pass on mantic

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/cli/tests/test_cli_security_status.py
+++ b/uaclient/cli/tests/test_cli_security_status.py
@@ -74,7 +74,6 @@ class TestActionSecurityStatus:
                 ):
                     main()
         out, _err = capsys.readouterr()
-        print(out)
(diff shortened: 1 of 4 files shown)


uaclient/cli/tests/test_cli_security_status.py
```diff
@@ -74,7 +74,6 @@ def test_security_status_help(
                 ):
                     main()
         out, _err = capsys.readouterr()
-        print(out)
```
renanrodrigo, line 77: whoooooooops

renanrodrigo (approved):

PR #2904, opened 2024-01-03
api: u.pro.services.enable.v1 (SC-1619)

orndorffgrant:
## Why is this needed?

This adds an API for enabling services, which will be useful for other software wanting to enable pro services in a dependable way.

## Test Steps

New behave tests should pass

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/po/pt_BR.po
+++ b/debian/po/pt_BR.po
@@ -3114,9 +3114,9 @@ msgid "failed to enable some services"
 msgstr ""
 
 #: ../../uaclient/messages/__init__.py:2193
-#, fuzzy, python-brace-format
+#, python-brace-format
 msgid "failed to enable {service}"
-msgstr "Habilite {service} com: pro enable {service}"
+msgstr ""
 
 #: ../../uaclient/messages/__init__.py:2199
 msgid "Failed to enable default services, check: sudo pro status"
@@ -3454,9 +3454,8 @@ msgstr ""
 "{values}"
 
 #: ../../uaclient/messages/__init__.py:2558
-#, fuzzy
 msgid "The operation is not supported"
-msgstr "A atualização ainda não está instalada"
+msgstr "A operação não é suportada"
 
 #~ msgid "Detach this machine from Ubuntu Pro services."
 #~ msgstr "Desvincule esta máquina de uma assinatura Ubuntu Pro"
--- a/debian/po/pt_BR.po
+++ b/debian/po/pt_BR.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-01-05 13:05-0300\n"
+"POT-Creation-Date: 2024-01-08 11:51-0500\n"
 "PO-Revision-Date: 2023-09-25 12:29-0400\n"
 "Last-Translator: Lucas Moura <lucas.moura@canonical.com>\n"
 "Language-Team: Brazilian Portuguese <ldpbr-translation@lists.sourceforge."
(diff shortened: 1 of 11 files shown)


uaclient/entitlements/base.py
```diff
@@ -416,6 +416,9 @@ def can_enable(self) -> Tuple[bool, Optional[CanEnableFailure]]:
     def enable(
         self,
         silent: bool = False,
+        enable_required_services: Optional[bool] = None,
+        disable_incompatible_services: Optional[bool] = None,
+        api: bool = False,
```
lucasmoura, line 421: I know this all be refactored in the future, but what if right now, instead of introducing new parameters to the method, we create two additional methods (names are just a suggestion):

* `handle_can_enable`
* `handle_required_components`

Then, we would just need to call those method directly on the `enable` API function without changing too much the main `enable` method. I believe this is better because this solution will not cross the boundaries between the API and the current enable logic.

But I again understand that this all be refactored in the future (even if we adopt my idea), so feel to disagree with this approach.

features/api_enable.feature
```diff
@@ -0,0 +1,169 @@
...
+          "code": "entitlement-not-enabled",
+          "meta": {
+            "reason": {
+              "code": "enable-access-only-not-supported",
+              "title": "ROS ESM All Updates does not support being enabled with --access-only"
+            }
+          },
+          "title": "failed to enable ros-updates"
+        }
+        """
```
lucasmoura, line 101: Should we add a test where we try to enable `landscape` ?

uaclient/api/u/pro/services/enable/v1.py
```diff
@@ -0,0 +1,167 @@
...
+
+    success = False
+    fail_reason = None
+
+    try:
+        with lock.SpinLock(
+            cfg=cfg,
+            lock_holder="u.pro.services.enable.v1",
+        ):
+            success, fail_reason = entitlement.enable(
```
lucasmoura, line 116: Just to double check, if we enable a service and its status is `warning`, should we notify the user somehow ?

orndorffgrant: Talked during standup: Even if a service ends up in a warning state, it was still enabled. The enable endpoint should present warnings that happened during the enable-action, but not about the service status post-enable. To get the full status of a service, a user should be using one of the status endpoints (e.g. u.pro.status.enabled_services.v1)

lucasmoura (approved):

debian/po/pt_BR.po
```diff
@@ -889,6 +889,7 @@ msgid "an error while reaching {url}"
 msgstr "um error ao acessar {url}"
 
 #: ../../uaclient/messages/__init__.py:414
+#: ../../uaclient/messages/__init__.py:1918
```
renanrodrigo, line 892: This is a duplicate message  - should we reuse it?

debian/po/pt_BR.po
```diff
@@ -3431,13 +3447,18 @@ msgstr ""
...
 #, python-brace-format
 msgid "Value provided was not found in {enum_class}'s allowed: value: {values}"
 msgstr ""
 "Valor fornecido não está presente nos valores permitidos de {enum_class}: "
 "{values}"
 
+#: ../../uaclient/messages/__init__.py:2558
+#, fuzzy
+msgid "The operation is not supported"
+msgstr "A atualização ainda não está instalada"
```
renanrodrigo, line 3460: "A operação não é suportada"

features/api_enable.feature
```diff
@@ -0,0 +1,188 @@
...
+        {
+          "disabled": [],
+          "enabled": [
+            "esm-infra"
+          ],
+          "messages": [],
+          "reboot_required": false
+        }
+        """
+        When I run shell command `pro api u.pro.status.enabled_services.v1 | jq .data.attributes.enabled_services ` with sudo
```
renanrodrigo, line 22: We recently landed a PR which brings simpler steps, like `verify that esm-infra is enabled` - we could use that here.

features/api_enable.feature
```diff
@@ -0,0 +1,188 @@
+Feature: u.pro.services.enable
+
+    Scenario Outline: u.pro.services.enable.v1 container services
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+        When I run `apt-get update` with sudo
```
renanrodrigo, line 5: We recently landed a pr which implements `When I apt update` - we could use it here; same for similar calls

features/api_enable.feature
```diff
@@ -0,0 +1,188 @@
...
+          {
+            "name": "realtime-kernel",
+            "variant_enabled": true,
+            "variant_name": "intel-iotg"
+          }
+        ]
+        """
+        Examples:
+           | release | machine_type |
+           | jammy   | lxd-vm       |
```
renanrodrigo, line 188: cursed missing newline

renanrodrigo (approved):
small nits here and there, but overall lgtm

uaclient/api/u/pro/services/enable/v1.py
```diff
@@ -0,0 +1,170 @@
...
+    return _enable(options, UAConfig())
+
+
+def _enable(
+    options: EnableOptions,
+    cfg: UAConfig,
+) -> EnableResult:
+    event.set_event_mode(event_logger.EventLoggerMode.JSON)
+
+    if not _is_attached(cfg).is_attached:
```
lucasmoura, line 84: Should we also handle the situation where a user is running as `non-root` ?

orndorffgrant: added!

orndorffgrant:

lucasmoura (approved):

orndorffgrant:
I'm forging ahead with #2908 and will likely refactor some of this, break into new PRs, etc. So I'll close this since it is just cluttering up our PR list

PR #2905, opened 2024-01-03
update logic that checks if a service is enabled

lucasmoura:
## Why is this needed?
When enabling a repo service, we check to see if it apt url is present at the output of apt-cache policy. We are updating this check to also check if any suite of the service also appear together with the apt url. We need this for some setups where the service share the same apt url.

LP: #2031192

## Test Steps
Run the new integration test

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/po/pt_BR.po
+++ b/debian/po/pt_BR.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-02-05 11:04-0800\n"
+"POT-Creation-Date: 2024-02-12 21:09-0500\n"
 "PO-Revision-Date: 2023-09-25 12:29-0400\n"
 "Last-Translator: Lucas Moura <lucas.moura@canonical.com>\n"
 "Language-Team: Brazilian Portuguese <ldpbr-translation@lists.sourceforge."
@@ -2342,7 +2342,7 @@ msgid "FIPS support requires system reboot to complete configuration."
 msgstr ""
 
 #: ../../uaclient/messages/__init__.py:1323
-#: ../../uaclient/messages/__init__.py:1768
+#: ../../uaclient/messages/__init__.py:1772
 msgid "Reboot to FIPS kernel required"
 msgstr ""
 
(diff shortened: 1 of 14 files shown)


renanrodrigo (approved):
Given the limitations we have for checking package availability, this is a good step forward

features/attach_validtoken.feature
```diff
@@ -278,3 +278,59 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Pro
...
+       """
+       And I append the following on uaclient config:
+       """
+       features:
+         serviceclient_url_responses: "/tmp/response-overlay.json"
+       """
+       And I verify that running `pro attach TOKEN` `with sudo` exits `1`
+       Then I will see the following on stderr:
+       """
+       Error processing entitlements directives for: esm-apps, esm-infra
```
orndorffgrant, line 323: We may want a line somewhere in here specifically stating that this is a backend problem that can't be fixed on the local machine.
Maybe start with "There is a problem with the resource definition provided by {contract_url}"?

lucasmoura: That's a good idea. I have updated the message

uaclient/actions.py
```diff
@@ -67,6 +68,16 @@ def attach_with_token(
...
+    ret, entitlement_clash_info = check_entitlement_directives_are_unique(cfg)
+
+    if not ret and entitlement_clash_info:
+        cfg.machine_token_file.delete()
+        raise exceptions.EntitlementsDirectivesAreNotUnique(
+            url=cfg.contract_url,
+            names=", ".join(sorted(entitlement_clash_info.entitlements)),
+            apt_url=entitlement_clash_info.apt_url,
+            suite=entitlement_clash_info.suite,
+        )
```
orndorffgrant, line 80: We can move this around a bit to avoid a write-then-immediately-delete case.
```suggestion
    ret, entitlement_clash_info = check_entitlement_directives_are_unique(cfg)
    if not ret and entitlement_clash_info:
        raise exceptions.EntitlementsDirectivesAreNotUnique(
            url=cfg.contract_url,
            names=", ".join(sorted(entitlement_clash_info.entitlements)),
            apt_url=entitlement_clash_info.apt_url,
            suite=entitlement_clash_info.suite,
        )
        
    cfg.machine_token_file.write(new_machine_token)
```

uaclient/entitlements/__init__.py
```diff
@@ -217,3 +226,55 @@ def create_enable_entitlements_not_found_error(
...
+
+# This function is just here to simplify our unittests, as
+# mocking isinstance directly doesn't work here
+def _is_repo_entitlement(ent):
+    return isinstance(ent, RepoEntitlement)
+
+
+def check_entitlement_directives_are_unique(
+    cfg: UAConfig,
+) -> Tuple[bool, Optional[EntitlementDirectiveClash]]:
```
orndorffgrant, line 239: Could this function `raise exceptions.EntitlementsDirectivesAreNotUnique` directly?

uaclient/messages/__init__.py
```diff
@@ -2554,3 +2558,13 @@ def __repr__(self):
     "error-updating-esm-cache",
     t.gettext("Error updating ESM services cache: {error}"),
 )
+
+E_ENTITLEMENTS_DIRECTIVES_ARE_NOT_UNIQUE = FormattedNamedMessage(
+    "entitlements-definitions-are-not-unique",
```
orndorffgrant, line 2563: There is a name mismatch here
```suggestion
E_ENTITLEMENTS_DIRECTIVES_ARE_NOT_UNIQUE = FormattedNamedMessage(
    "entitlements-directives-are-not-unique",
```

orndorffgrant: Separately, this message is pretty specific to APT sources and is not applicable to differences in directives in general (e.g. requiredSnaps or additionalPackages), so "entitlement directives are not unique" is a little to generic. Should we rename all of the functions/names in this PR to be more specific to apt sources?

orndorffgrant (commented):
Looking good! Just some small suggestions

debian/po/pt_BR.po
```diff
@@ -3455,18 +3460,28 @@ msgstr ""
...
 msgstr ""
 
+#: ../../uaclient/messages/__init__.py:2565
+#, python-brace-format
+msgid ""
+"There is a problem with the resource directives provided by {url}\n"
+"These entitlements: {names} are sharing the following directives\n"
+" - APT url: {apt_url}\n"
+" - Suite: {suite}\n"
+"Those fields need to be unique for every entitlement."
```
s-makin, line 3482: "Those fields" feels a bit vague to me, which fields specifically need to be unique?

uaclient/entitlements/__init__.py
lucasmoura: Yes, that's a better approach

debian/po/pt_BR.po
renanrodrigo: those are the directives above.
So maybe changing `Those fields` for `These directives` ?

uaclient/messages/__init__.py
lucasmoura: Indeed, that's a good point

debian/po/pt_BR.po
lucasmoura: My idea was to refer to both `APT url` and `Suite`. Do you think I should rewrite the message to make this more explicit ?

lucasmoura:
@orndorffgrant updated

features/airgapped.feature
```diff
@@ -47,3 +47,53 @@ Feature: Performing attach using ua-airgapped
...
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+        # set up the apt mirror configuration
+        Given a `jammy` `<machine_type>` machine named `mirror`
+        When I run `add-apt-repository ppa:yellow/ua-airgapped -y` `with sudo` on the `mirror` machine
+        And I run `apt-get update` `with sudo` on the `mirror` machine
+        And I run `apt-get install apt-mirror get-resource-tokens ua-airgapped -yq` `with sudo` on the `mirror` machine
+        And I download the service credentials on the `mirror` machine
+        And I extract the `esm-infra` credentials from the `mirror` machine
+        And I extract the `esm-apps` credentials from the `mirror` machine
+        And I set the apt-mirror file for `<release>` with the `esm-infra,esm-apps` credentials on the `mirror` machine
```
s-makin, line 61: Does apt-mirror on this line need to be decorated like `apt-mirror` on the next line?

debian/po/pt_BR.po
s-makin: "These directives" would resolve it, because then there's no ambiguity about what fields are being referred to.

Since you defined already that APT url and Suite are the directives, I don't think you need to specify both of them again. The original wording, it's not clear if "those fields" refers to "these entitlements" or "the following directives" (that was the confusion I had)

uaclient/messages/__init__.py
```diff
@@ -2554,3 +2558,13 @@ def __repr__(self):
...
+    t.gettext(
+        "There is a problem with the resource directives provided by {url}\n"
+        "These entitlements: {names} are sharing the following directives\n"
+        " - APT url: {apt_url}\n - Suite: {suite}\n"
+        "Those fields need to be unique for every entitlement."
```
s-makin, line 2568: Same as above, just highlighting for visibility

s-makin (approved):
Only that small s/those fields/these directives/ substitution to make, otherwise LGTM :)

uaclient/actions.py
orndorffgrant: Ah sorry, is there any reason that that `check_entitlement_directives_are_unique` function cannot operate directly on `new_machine_token` here? So the call would become:
```python
    ret, entitlement_clash_info = check_entitlement_directives_are_unique(new_machine_token)
```

orndorffgrant: Oh I see. The logic relies on creating the entitlement instances, which take the `cfg` and read the machine token that way.

This is okay then, no need to follow my suggestion.

uaclient/messages/__init__.py
orndorffgrant: I think the message still needs renamed to match the new exception name
```suggestion
E_ENTITLEMENTS_APT_DIRECTIVES_ARE_NOT_UNIQUE = FormattedNamedMessage(
    "entitlements-apt-directives-are-not-unique",
```

orndorffgrant (commented):
Looks good - just one minor comment and Sally's suggestion to resolve

PR #2907, opened 2024-01-04
Convert repo files to deb822

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because Ubuntu itself will change to deb822 sources by default, and the Client should follow. Not changing means warnings in the future, and adapting early gives us room to facilitate integration when the shift actually happens.

## Test Steps
Except from the file format, everything should work exactly the same. CI should cover all cases.

## Checklist

 - [x] Yes - SRU people + Julian would be nice
 - [ ] No

### DIFF
--- a/lib/convert_list_to_deb822.py
+++ b/lib/convert_list_to_deb822.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+
+"""
+This script is called after running do-release-upgrade in a machine.
+This converts list files to deb822 files when upgrading to Noble.
+"""
+
+import logging
+import os
+import sys
+
+from aptsources.sourceslist import SourceEntry  # type: ignore
+
+from uaclient import entitlements
+from uaclient.apt import _get_sources_file_content
+from uaclient.cli import setup_logging
+from uaclient.config import UAConfig
+from uaclient.system import (
+    ensure_file_absent,
+    get_release_info,
+    load_file,
+    write_file,
+)
+from uaclient.util import set_filename_extension
+
+if __name__ == "__main__":
+    series = get_release_info().series
+    if series != "noble":
+        sys.exit(0)
+
+    setup_logging(logging.DEBUG)
+    cfg = UAConfig()
+
+    for entitlement_class in entitlements.ENTITLEMENT_CLASSES:
+        if not issubclass(
+            entitlement_class, entitlements.repo.RepoEntitlement
+        ):
+            continue
+
+        entitlement = entitlement_class(cfg)
+
+        filename = set_filename_extension(entitlement.repo_file, "list")
+        if os.path.exists(filename):
+            # If do-release-upgrade commented out the file, whether the
+            # repository is not reachable or is considered a third party, then
+            # it will be handled in upgrade_lts_contract. This script only
+            # changes services which are enabled, active and reachable.
+            valid_sources = [
+                SourceEntry(line)
+                for line in load_file(filename).strip().split("\n")
+                if line.strip().startswith("deb")
+            ]
+            if valid_sources:
+                # get this information from the file, to avoid interacting with
+                # the entitlement_config
+                suites = list(set(source.dist for source in valid_sources))
+                repo_url = valid_sources[0].uri
+                include_deb_src = any(
+                    source.type == "deb-src" for source in valid_sources
+                )
+                content = _get_sources_file_content(
+                    suites,
+                    series,
+                    True,
+                    repo_url,
+                    entitlement.repo_key_file,
+                    include_deb_src,
+                )
+                write_file(entitlement.repo_file, content)
+
+            ensure_file_absent(filename)
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -658,6 +658,12 @@ def remove_auth_apt_repo(
 ) -> None:
     """Remove an authenticated apt repo and credentials to the system"""
     system.ensure_file_absent(repo_filename)
+    # Also try to remove old .list files for compatibility with older releases.
+    if repo_filename.endswith(".sources"):
+        system.ensure_file_absent(
+            util.set_filename_extension(repo_filename, "list")
+        )
+
     if keyring_file:
         keyring_file = os.path.join(APT_KEYS_DIR, keyring_file)
         system.ensure_file_absent(keyring_file)
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -63,15 +63,6 @@ Components: main
 Signed-By: {keyrings_dir}/{keyring_file}
 """
 
-ESM_REPO_FILE_CONTENT = """\
-# Written by ubuntu-advantage-tools
-
-deb https://esm.ubuntu.com/{name}/ubuntu {series}-{name}-security main
-# deb-src https://esm.ubuntu.com/{name}/ubuntu {series}-{name}-security main
-
-deb https://esm.ubuntu.com/{name}/ubuntu {series}-{name}-updates main
-# deb-src https://esm.ubuntu.com/{name}/ubuntu {series}-{name}-updates main
-"""
 
 ESM_BASIC_FILE_STRUCTURE = {
     "files": [
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -56,7 +56,7 @@ SERIES_NOT_USING_DEB822 = ("xenial", "bionic", "focal", "jammy", "mantic")
 DEB822_REPO_FILE_CONTENT = """\
 # Written by ubuntu-advantage-tools
 
-Types: deb
+Types: deb{deb_src}
 URIs: {url}
 Suites: {suites}
 Components: main
@@ -506,6 +506,7 @@ def _get_sources_file_content(
     updates_enabled: bool,
     repo_url: str,
     keyring_file: str,
+    include_deb_src: bool = False,
 ) -> str:
     appliable_suites = [suite for suite in suites if series in suite]
     if not updates_enabled:
@@ -518,11 +519,14 @@ def _get_sources_file_content(
             suite for suite in appliable_suites if "-updates" not in suite
         ]
 
+    deb_src = " deb-src" if include_deb_src else ""
+
     content = DEB822_REPO_FILE_CONTENT.format(
         url=repo_url,
         suites=" ".join(appliable_suites),
         keyrings_dir=KEYRINGS_DIR,
         keyring_file=keyring_file,
+        deb_src=deb_src,
     )
 
     return content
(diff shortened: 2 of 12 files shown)


uaclient/tests/test_util.py
```diff
@@ -468,3 +468,31 @@ def test_replace_top_level_logger_name(self, logger_name, new_logger_name):
         assert (
             util.replace_top_level_logger_name(logger_name) == new_logger_name
         )
+
+
+class TestSetFilenameExtension:
+    @pytest.mark.parametrize(
+        "input_string,extension,output_string",
```
panlinux, line 475: you could add a test for a file with multiple dots, like "some.file.txt"

uaclient/tests/test_apt.py
```diff
@@ -347,140 +347,190 @@ def test_errors_on_apt_helper_process_timeout(
...
-            suites=("xenial",),
+            suites=(series,),
             keyring_file="keyring",
         )
 
-        expected_content = (
-            "deb http://fakerepo xenial main\n"
-            "# deb-src http://fakerepo xenial main\n"
-        )
+        if series == "xenial":
```
panlinux, line 381: I know the test is being run with "xenial" and "noble" as series, but perhaps here to use less hardcoding, what do you think about checking `if series in apt.SERIES_NOT_USING_DEB822` instead?

uaclient/tests/test_apt.py
```diff
@@ -347,140 +347,190 @@ def test_errors_on_apt_helper_process_timeout(
...
+            expected_content = (
+                "deb http://fakerepo xenial main\n"
+                "# deb-src http://fakerepo xenial main\n"
```
panlinux, line 384: Maybe replace `xenial` with `{series}` here and add the necessary `.format()`.

uaclient/tests/test_apt.py
```diff
@@ -347,140 +347,190 @@ def test_errors_on_apt_helper_process_timeout(
...
+            src_keyfile = os.path.join(KEYRINGS_DIR, "keyring")
+            dest_keyfile = os.path.join(APT_KEYS_DIR, "keyring")
+            gpg_export_calls = [mock.call(src_keyfile, dest_keyfile)]
+        else:
+            expected_content = (
+                "# Written by ubuntu-advantage-tools\n"
+                "\n"
+                "Types: deb\n"
+                "URIs: http://fakerepo\n"
+                "Suites: noble\n"
```
panlinux, line 395: Also best to use `{series}` instead of the hardcoded `noble`, assuming also you took the suggestion of using `SERIES_NOT_USING_DEB822` from above.

uaclient/tests/test_apt.py
```diff
@@ -347,140 +347,190 @@ def test_errors_on_apt_helper_process_timeout(
...
 
-        expected_content = dedent(
-            """\
-            deb http://fakerepo xenial-one main
-            # deb-src http://fakerepo xenial-one main
-            deb http://fakerepo xenial-updates main
-            # deb-src http://fakerepo xenial-updates main
-        """
-        )
+        if series == "xenial":
```
panlinux, line 449: Also best to use a check using `if series in apt.SERIES_NOT_USING_DEB822` perhaps?

uaclient/tests/test_apt.py
```diff
@@ -347,140 +347,190 @@ def test_errors_on_apt_helper_process_timeout(
...
+            expected_content = dedent(
+                """\
+                deb http://fakerepo xenial-one main
+                # deb-src http://fakerepo xenial-one main
+                deb http://fakerepo xenial-updates main
+                # deb-src http://fakerepo xenial-updates main
```
panlinux, line 455: And here we could use `{series}-` instead of hardcoded `xenial`. Just like `add_auth_apt_repo()` used above fo `suites=(...)`.

uaclient/tests/test_apt.py
```diff
@@ -347,140 +347,190 @@ def test_errors_on_apt_helper_process_timeout(
...
+            """
+            )
+        else:
+            expected_content = dedent(
+                """\
+                # Written by ubuntu-advantage-tools
+
+                Types: deb
+                URIs: http://fakerepo
+                Suites: noble-one noble-updates
```
panlinux, line 465: Also use `{series}` here

uaclient/tests/test_apt.py
```diff
@@ -347,140 +347,190 @@ def test_errors_on_apt_helper_process_timeout(
...
 
-        expected_content = dedent(
-            """\
-            deb http://fakerepo xenial-one main
-            # deb-src http://fakerepo xenial-one main
-            # deb http://fakerepo xenial-updates main
-            # deb-src http://fakerepo xenial-updates main
-        """
-        )
+        if series == "xenial":
```
panlinux, line 512: Also best to use a check using `if series in apt.SERIES_NOT_USING_DEB822` perhaps?

uaclient/tests/test_apt.py
```diff
@@ -347,140 +347,190 @@ def test_errors_on_apt_helper_process_timeout(
...
+        else:
+            expected_content = dedent(
+                """\
+                # Written by ubuntu-advantage-tools
+
+                Types: deb
+                URIs: http://fakerepo
+                Suites: noble-one
+                Components: main
+                Signed-By: /usr/share/keyrings/keyring
```
panlinux, line 530: And user `{series}` instead of the hardcoded values

uaclient/apt.py
```diff
@@ -601,6 +653,12 @@ def remove_auth_apt_repo(
 ) -> None:
     """Remove an authenticated apt repo and credentials to the system"""
     system.ensure_file_absent(repo_filename)
+    # Also try to remove old .list files for compatibility with older releases.
+    # If repo_filename is already .list, this is just a 'noop'
```
panlinux, line 657: Noop, yes, but will trigger a log warning, right? About the file not being there, because the call above already removed it.

lib/convert_list_to_deb822.py
```diff
@@ -0,0 +1,67 @@
...
+
+    for entitlement_class in entitlements.ENTITLEMENT_CLASSES:
+        if not issubclass(
+            entitlement_class, entitlements.repo.RepoEntitlement
+        ):
+            continue
+
+        entitlement = entitlement_class(cfg)
+
+        filename = set_filename_extension(entitlement.repo_file, "list")
```
panlinux, line 42: Just checking, but a normal `do-release-upgrade` will rename "third-party" `.list` files to something else, and I'm presuming it will not consider the entitlement repo_files as third party, correct? Because otherwise at this point these files will have been renamed to something like `.list.distUpgrade` or other.

lib/convert_list_to_deb822.py
```diff
@@ -0,0 +1,67 @@
...
+        if os.path.exists(filename):
+            valid_sources = [
+                SourceEntry(line)
+                for line in load_file(filename).strip().split("\n")
+                if line.strip().startswith("deb")
```
panlinux, line 47: Continuing on the previous comment, `do-release-upgrade` could have disabled a sources.list entry if it was deemed a third-party repository, and thus this `startswith` check wouldn't work.

panlinux:
I didn't quite understand this commit:
```
commit e274a774c026992e678cd2955971650aa8ad41d9
Author: Renan Rodrigo <renanrodrigo@canonical.com>
Date:   Wed Jan 3 19:10:21 2024 -0300

    apt: change private cache list files to deb822
    
    In this case we can do it for all releases, as there is no dependency on
    it outside of the client.
```
 What are the "private cache list files" again, and why does the ubuntu release not matter?

uaclient/tests/test_apt.py
orndorffgrant: Alternatively, we can parametrize the `expected_content` and `gpg_export_calls` alongside the `series`, so we won't need an if statement here at all.

lib/convert_list_to_deb822.py
```diff
@@ -0,0 +1,67 @@
...
+    for entitlement_class in entitlements.ENTITLEMENT_CLASSES:
+        if not issubclass(
+            entitlement_class, entitlements.repo.RepoEntitlement
+        ):
+            continue
+
+        entitlement = entitlement_class(cfg)
+
+        filename = set_filename_extension(entitlement.repo_file, "list")
+        if os.path.exists(filename):
```
lucasmoura, line 43: @renanrodrigo let's just add a comment that `do-release-upgrade` will comment out the list file of other services (i.e. `usg`) and because of that, they not be evaluated here

uaclient/apt.py
orndorffgrant: It will be a double log, but not a warning

panlinux: Sorry, not a warning, but a debug log, which is the default log level of the client, right?

lib/convert_list_to_deb822.py
renanrodrigo: Added

uaclient/tests/test_util.py
renanrodrigo: done

uaclient/tests/test_apt.py
renanrodrigo: I think parametrizing everything makes sense, but @panlinux 's suggestion brings better readability - done that, let me know if this suffices

uaclient/apt.py
renanrodrigo: I added a check before the removal just to make sure we only try to remove the `.list` when removing a `.sources`.
- for SRUs, this guarantees no double logging
- from Noble on the log that the file exists will be there, but we actually do want it to be there.

lib/convert_list_to_deb822.py
renanrodrigo: So when `do-release-upgrade` considers some of our services as third-parties (which it should not do, btw...) the `.list` file is commented out, and the `.list.distUpgrade` file is not a rename, but rather a backup.
As of now, the `update_lts_contract` script will handle this situation, by removing the file and recreating the apt config for the given entitlement - which will be a `.sources` file in noble.
For the future, we can improve this situation by making sure we are always recognized as valid sources.

lib/convert_list_to_deb822.py
renanrodrigo: Not only that - it does the same when the repository is not available. As mentioned, the other post-upgrade script should deal with this case.

renanrodrigo:
@panlinux about the local esm cache files:
We have a private apt cache to store information about esm packages for systems where the services are disabled (the work started in #2342 and continued in subsequent PRs).
This private cache uses the apt machinery, including the repository files, which we are migrating to `.sources` now.
The ubuntu release does not matter because even in `Xenial`, the `.sources` files are fully supported. We want to standardize it now.
It is important to note that even for the regular apt configuration (the "real" one haha) we wanted to change everything to `deb822` - however we need to keep the old `.list` format for the released series because other tools also consume archives from apt, and those tools would not be prepared for the new format, causing regressions.
As no one but ourselves is depending on the local caches, it is easier to migrate everything than keeping both formats for our thing.

panlinux (approved):
+1

PR #2912, opened 2024-01-08
fix: don't include LSN in cve notices list

lucasmoura:
## Why is this needed?
When fetching the CVE notices list, we should exclude a LSNs, as they are not supported for the fix operation

## Test Steps
Run the bionic fix test and guarantee that it is passing

## Checklist

 - [ ] Yes
 - [x] No


lucasmoura:
I wasn't able to reproduce any of the CI failures locally

orndorffgrant (approved):
LGTM thanks!

PR #2916, opened 2024-01-12
Some bits missing in the deb822 transition

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it brings small tweaks needed for the deb822 transition, both in tests and functionality bits.

## Test Steps
This one is interesting... the integration test changes are fine to check, but the other ones can really be tested when noble changes the format in images :D

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/entitlements/base.py
+++ b/uaclient/entitlements/base.py
@@ -1322,5 +1322,5 @@ class UAEntitlement(metaclass=abc.ABCMeta):
         if self._is_sources_list_updated:
             return
         event.info(messages.APT_UPDATING_LIST.format(name="standard Ubuntu"))
-        apt.update_sources_list("/etc/apt/sources.list")
+        apt.update_sources_list(apt.get_system_sources_file())
         self._is_sources_list_updated = True
--- a/debian/ubuntu-advantage-tools.prerm
+++ b/debian/ubuntu-advantage-tools.prerm
@@ -6,6 +6,7 @@ remove_apt_files() {
     # This list should be kept up to date with the list of available apt-repo-based services
     for service in anbox cc-eal cis esm-apps esm-infra fips fips-preview fips-updates realtime-kernel ros ros-updates; do
         rm -f /etc/apt/sources.list.d/ubuntu-${service}.list
+        rm -f /etc/apt/sources.list.d/ubuntu-${service}.sources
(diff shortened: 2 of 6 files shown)


uaclient/entitlements/base.py
```diff
@@ -1322,5 +1323,12 @@ def _update_sources_list(self):
...
         event.info(messages.APT_UPDATING_LIST.format(name="standard Ubuntu"))
-        apt.update_sources_list("/etc/apt/sources.list")
+        old_sources_path = "/etc/apt/sources.list"
+        new_sources_path = "/etc/apt/sources.list.d/ubuntu.sources"
+        sources_file = (
+            new_sources_path
+            if os.path.exists(new_sources_path)
+            else old_sources_path
+        )
+        apt.update_sources_list(sources_file)
```
lucasmoura, line 1333: Just to avoid code duplication, what about if we make this logic the default `sources_file` we if we don't pass anything to `update_sources_list` ?

lucasmoura (approved):
Just an idea to avoid code duplication, let me know what you think

uaclient/entitlements/base.py
renanrodrigo: right
any chance of a situation where we just want to update all sources, pro or not, without specifying a file?
Doesn't exist right now but I wonder if we should account for that

lucasmoura: Good point, what if we create a shell function that just calls `update_sources_list` with the `/etc/` path ? Then we would still be able to extend the `update_sources_list` function to the use case you mentioned, while also avoiding the code duplication.

What do you think ?

renanrodrigo: I thought that would be overcomplicating it - deduplicated with an apt util function to get the file. Let me know if that suffices

renanrodrigo:
@orndorffgrant @lucasmoura added a new commit with a small tweak, double check before clicking

orndorffgrant (approved):

uaclient/entitlements/base.py
lucasmoura: That works for me. Thanks @renanrodrigo

lucasmoura (approved):

Issue #2850, opened 2023-11-24
Bug: API integration tests are difficult to review


lucasmoura:
**Description of the bug**

On a integration test, the expected output of an API endpoint is a just JSON blob in a single line. That is because this is exactly what we produce in the API endpoint itself. However, this JSON blob is extremely difficult to read and it makes the review process a lot harder, since evaluating if the blob has indeed desired output takes a lot of manual work on our team.

**Expected behavior**

There are two things we can consider here:

1. Properly format the JSON output to make it human readable. That will at least make the review process easier
2. Re-evaluate the tests for API endpoints that are too verbose (i.e. fix plan). We need to understand if all the tests there are actually necessary.

**Current behavior**

As mentioned, today we just review the JSON blob directly, which is not ideal.

**To Reproduce**

Just look at any api integration test that is still in use

**System information:**

Not relevant

PR #2917, opened 2024-01-12
tests: simplify API output reviews

lucasmoura:
## Why is this needed?
Currently, we need to review the full API blob on the integration tests. We are now updating how we compare API outputs to make the review process easier.

Fixes: #2850

## Test Steps
Check if the API tests are still passing

## Checklist

 - [ ] Yes
 - [ ] No


lucasmoura:
@renanrodrigo originally, I thought about adding a new step, but we would need to create something completely different  from the `{stream} matches regexp` and my idea was to maintain that pattern to make it clear that we are performing the same type of check.

But if you prefer, we can create a new step that doesn't follow this pattern. I am not oppose to it

orndorffgrant:
I think the names of the steps are good ("The API full output matches regexp") are good, but I agree that defining them via separate step functions would be more maintainable like:
```python
@then("API full output matches regexp")
def then_api_full_matches_regexp(context, stream):
    json_data = json.loads(context.process.stdout.strip())
    content = json.dumps(json_data, indent=2)
    text = context.text.replace("[", "\\[")
    text = text.replace("]", "\\]")
    text = process_template_vars(context, text)
    if re.compile(text).search(content) is None:
        raise AssertionError(
            "Expected to match regexp:\n{}\nBut got:\n{}".format(
                textwrap.indent(text, "  "), textwrap.indent(content, "  ")
            )
)
```
I think this because when I forget where a step is defined I'll often search the codebase for what looks like the most unique part of the step name - so I'd search for "API full output". But that won't get me to the correct step definition in the current implementation

lucasmoura:
@renanrodrigo @orndorffgrant Thanks for the review. I have updated the code with the new steps

orndorffgrant:
There is one test that seems to be a legitimate failure. Otherwise LGTM!

lucasmoura:
@orndorffgrant fixed

Issue #2859, opened 2023-11-29
`Given a` machine does not run apt-get update on it


panlinux:
**Description of the bug**

Tests where an extra machine is created, and where packages are installed via apt-get, can fail to install the packages if the repository information is out of date. I'm seeing this in the [ubuntu_pro.feature](https://github.com/canonical/ubuntu-pro-client/actions/runs/7023435516/job/19110296166?pr=2794) set of tests where a proxy machine is created, and the step to install squid fails on it:
```
The following NEW packages will be installed:
        libdbi-perl libecap3 squid squid-common squid-langpack ssl-cert
      0 upgraded, 6 newly installed, 0 to remove and 0 not upgraded.
      Need to get 3687 kB of archives.
      After this operation, 15.7 MB of additional disk space will be used.
      Get:1 http://us-west2.gce.archive.ubuntu.com/ubuntu focal/main amd64 libecap3 amd64 1.0.1-3.2ubuntu1 [17.4 kB]
      Get:2 http://us-west2.gce.archive.ubuntu.com/ubuntu focal/main amd64 squid-langpack all 20191103-1 [167 kB]
      Ign:3 http://us-west2.gce.archive.ubuntu.com/ubuntu focal-updates/main amd64 squid-common all 4.10-1ubuntu1.7
      Get:4 http://us-west2.gce.archive.ubuntu.com/ubuntu focal-updates/main amd64 libdbi-perl amd64 1.643-1ubuntu0.1 [730 kB]
      Get:5 http://us-west2.gce.archive.ubuntu.com/ubuntu focal/main amd64 ssl-cert all 1.0.39 [17.0 kB]
      Ign:6 http://us-west2.gce.archive.ubuntu.com/ubuntu focal-updates/main amd64 squid amd64 4.10-1ubuntu1.7
      Err:3 http://security.ubuntu.com/ubuntu focal-updates/main amd64 squid-common all 4.10-1ubuntu1.7
        404  Not Found [IP: 104.198.11.147 80]
      Err:6 http://security.ubuntu.com/ubuntu focal-updates/main amd64 squid amd64 4.10-1ubuntu1.7
        404  Not Found [IP: 104.198.11.147 80]
      Fetched 931 kB in 1s (1419 kB/s)
      2023-11-28 19:56:51,985:ERROR:root:stderr: 
      WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
      
[...]

PR #2918, opened 2024-01-15
tests: run apt update when creating instances in behave

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because now we don't need explicit steps to update the apt data, making the apt update calls in behave tests happen only when really testing something.

Also
Fixes: #2859

## Test Steps
All behave should behave the same.

## Checklist
 - [x] I have updated or added any integration tests accordingly AHA

 - [ ] Yes
 - [x] No


orndorffgrant (approved):
Looks good! Just some conflicts to resolve then :rocket:

renanrodrigo:
@orndorffgrant thanks for resolving this

PR #2921, opened 2024-01-17
dev-docs: tutorial for getting started

orndorffgrant:
## Why is this needed?

We previously had no good starting place for someone who wants to contribute to pro-client. This tutorial should get someone up and running from a blank system to running unit tests and then to running basic behave tests.

## Test Steps
Follow the steps in a mantic VM


dev-docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,142 @@
...
+wget https://raw.githubusercontent.com/canonical/pycloudlib/main/pycloudlib.toml.template -O ~/.config/pycloudlib.toml
+```
+
+## Run a behave test
+
+```bash
+tox -e behave -- features/config.feature -D releases=xenial -D machine_types=lxd-container
+```
+
+All of the arguments after the `--` are passed to behave. In this case, we're telling behave to only run the `config.feature` test, and to filter the tests in that file to only those for xenial lxd containers. Note that the `-D` options are specific to our behave tests and are not behave options.
```
s-makin, line 129:
```suggestion
All of the arguments after the `--` are passed to behave. In this case, we're telling behave to only run the `config.feature` test, and to filter the tests in that file to only those for Xenial LXD containers. Note that the `-D` options are specific to our behave tests and are not behave options.
```

dev-docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,142 @@
...
+
+# Interacting with local changes in a container
+With all of the above in place, you can now make changes to the code and run a local version of Ubuntu Pro Client in a lxd container to try out your changes.
```
s-makin, line 132:
```suggestion
With all of the above in place, you can now make changes to the code and run a local version of Ubuntu Pro Client in a LXD container to try out your changes.
```

dev-docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,142 @@
...
+
+Edit `uaclient/version.py` and modify the `get_version()` function to return a fake version string. For example, change the first line of the function to `return "42:42"`.
+
+Now use our helper script to build a deb with your changes, launch a lxd container, install your deb, and drop you into a shell on the container.
```
s-makin, line 136:
```suggestion
Now use our helper script to build a deb with your changes, launch a LXD container, install your deb, and drop you into a shell on the container.
```

s-makin (commented):
Nice, thanks for taking the time to create this! I've left a few nits, mostly just capitalising for consistency. I'll let the others take a look before approving.

dev-docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,142 @@
...
+
+> **Note**
+> This tutorial was last tested on Ubuntu 23.10 Mantic Minotaur (mantic).
+
+# From nothing to unit tests
+Open a terminal.
+
+Install necessary host tooling.
+```bash
+sudo apt install git tox pre-commit intltool libapt-pkg-dev sbuild-launchpad-chroot
```
lucasmoura, line 15: I wonder if we should modify our `Makefile` to do that. But if we decide we should, this can be done later

lucasmoura (approved):
Just a suggestion for later, but I believe this a great tutorial.
Thanks for the effort here @orndorffgrant

a-dubs (changes requested):
`Contributing.md` needs updated. The first link is a dead link, pointing to one of the two deleted markdown files.
THe following needs updated or removed:
```
* [How to build](./dev-docs/howtoguides/building.md)   
```

dev-docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,142 @@
...
+```bash
+tox
+```
+
+> **Note**
+> You can run individual jobs (or "environments" in tox terminology) with the `-e` flag.
+>
+> For example, `tox -e test` will only run the unit tests.
+>
+> Explore `tox.ini` to see what tox environments are available and how they're configured.
```
a-dubs, line 42: Would it be beneficial to include additional examples? I'm torn because it would be nice to know right off the rip that `tox -e do_format` and `tox -e check_format` exists. 
(I never scrolled far enough down in tox.ini before to see that do_format exists so I've been doing all my formatting manually :smiling_face_with_tear:)

But yeah I think those 2 commands as well as `tox -e test` are essential for development and would be worth explicitly giving to the user. Because some new contributers will probably be like me: coming from a background of unfamiliarity with tox, and thus even if they were to read the whole tox.ini file, would not know what commands they actually have available to them.

dev-docs/tutorials/getting-started.md
renanrodrigo: I would say we definitely should do that

dev-docs/tutorials/getting-started.md
renanrodrigo: more than that, we could have a small section explaining style formatting, shellcheck, etc; all tox things deserve it

dev-docs/tutorials/getting-started.md
renanrodrigo, line 129: I'd say it would be nice to highlight, maybe here, all the `-D` options we support, plus the behave native `-i` and `-n`, for usefulness' sake

renanrodrigo:
- Do we want to mention Multipass in this entry level tutorial? I'd say yes - and also mention that it is a little harder to spin lxc Xenial/Bionic VMS yourself.
- Should we differentiate lxd VM and container testing? 
- Should we mention cloud tests, and which secrets/credentials are expected for each cloud?

dev-docs/tutorials/getting-started.md
orndorffgrant: Sure makes sense, we'll just need this for `git` but the rest can be `make install-dev-deps` or something like that

dev-docs/tutorials/getting-started.md
orndorffgrant: do_format and check_format are cloud-init tox commands :), but I like the sentiment that we should list a couple more commands.

I'm wary of adding too much here though, I would rather encourage the reader to explore more on their own here, and have separate detailed docs for certain topics as necessary. But I'll try to add a short explanation of what we use tox for.

orndorffgrant: And @a-dubs for ubuntu-pro-client I recommend setting up `pre-commit` as the tutorial recommends - that will format your code before committing.

dev-docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,142 @@
...
+
+# Getting to basic behave tests
+We use [`behave`](https://behave.readthedocs.io/en/stable/) as a framework for writing integration/acceptance/regression tests. behave tests are in the `features/` folder and generally follow the pattern of:
+1. Launch a VM or container on a platform
```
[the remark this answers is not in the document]
orndorffgrant: I was going for a generic description of what the behave test does, so its kinda vague for a reason. And I don't want to list all of the supported platforms here because it will get out of date. I do like the idea of listing a couple examples in parentheses though to demonstrate what we mean by "platform", I'll add that for sure

dev-docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,142 @@
...
+ping -c 3 ubuntu.com
+# the ping command should succeed
+exit
+# now you should be back on your host
+lxc delete --force testx
+```
+
+> **Note**
+> Docker can interfere with lxd container networking. If you need docker installed alongside lxd, follow the guidance in the lxd documentation to ensure that docker doesn't break lxd networking: https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker
+
```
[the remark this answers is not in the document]
orndorffgrant: It's effectively a requirement to be a "core" pro-client dev, but you're right I'll add a short note.

dev-docs/tutorials/getting-started.md
orndorffgrant: *looks up `-i`* - that's cool

I agree we need documentation on these, but I'm not sure if it makes sense to put it all in this tutorial. I'm pondering creating a "how to work with behave" and/or "pro behave reference manual" type document that is a bit more comprehensive instead.

dev-docs/tutorials/getting-started.md
orndorffgrant: this requires a separate PR to add to the Makefile so I'd like to leave it as is for now.

Breaks: create `make install-dev-deps` target to easily install all dev tooling

orndorffgrant: Breaks: create `make install-dev-deps` target to easily install all dev tooling

orndorffgrant:
> * Do we want to mention Multipass in this entry level tutorial? I'd say yes - and also mention that it is a little harder to spin lxc Xenial/Bionic VMS yourself.

Good idea, I'll add it.

>     * Should we differentiate lxd VM and container testing?

I suppose there's no real reason to, I can add another suggested behave test to try out a lxd-vm one.

>     * Should we mention cloud tests, and which secrets/credentials are expected for each cloud?

I'm trying to make the getting started tutorial as copy-paste-able as possible, and we'd lose that quality too much once we get into cloud secret management. I think setting up the cloud tests might deserve its own how-to-guide - what do you think?

We should at least add instructions on configuring a contract token for behave in this tutorial though.

dev-docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,190 @@
...
+tox
+```
+`tox` manages separate virtual environments for each of the different jobs we run. It will install all of the dependencies for each job into the virtual environment for that job, and then run the job. This ensures that the dependencies for each job don't interfere with each other.
+
+> **Note**
+> You can run individual jobs (or "environments" in tox terminology) with the `-e` flag.
+>
+> For example:
+> - `tox -e test` will run the unit tests.
+> - `tox -e flake8` will run the python linter.
```
s-makin, line 45:
```suggestion
> - `tox -e flake8` will run the Python linter.
```

s-makin (approved):
One small nit, otherwise lgtm

renanrodrigo (approved):
Agree on separate guides for the specificities, makes sense

PR #2930, opened 2024-01-29
Update dev docs to include doc testing

dheyay:
## Why is this needed?

Testing documentation updated to include local testing for updates to docs.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/dev-docs/documentation/contribute_docs.md
+++ b/dev-docs/documentation/contribute_docs.md
@@ -61,6 +61,18 @@ This will build *and serve* the documentation at http://127.0.0.1:8000 -- this
 gives you a live preview of any changes you make (and save) to the
 documentation without needing to continually rebuild.
 
+## Doc testing
+
+Before running the `make` command, make sure to download the Python venv module by running the following command:
+```sh
+apt install python3-venv
+```
+When working with any `docs` branch and making updates, you can run the documentation tests using the Makefile. 
+The `make` command lists all targets to test. You can run individual targets using the command:
+```bash
+make target
+```
+
 ## Docs structure and organisation
 
 We follow the [principles of Diataxis](https://diataxis.fr/) in our
--- a/dev-docs/howtoguides/testing.md
+++ b/dev-docs/howtoguides/testing.md
@@ -246,3 +246,13 @@ a daily ubuntu-advantage-tools image.
 ```sh
 UACLIENT_BEHAVE_REUSE_IMAGE=your-custom-image-id tox -e behave-gcppro-20.04
 ```
+
+## Documentation tests
+
+When working with a `docs` branch and making updates, you can run the documentation tests using the Makefile. The `make` command should display all targets to test.
+
+Before running the `make` command, make sure to download the python venv module by running the following command:
+
+```sh
+apt install python3-venv
+```
--- a/dev-docs/howtoguides/testing.md
+++ b/dev-docs/howtoguides/testing.md
@@ -246,13 +246,3 @@ a daily ubuntu-advantage-tools image.
 ```sh
 UACLIENT_BEHAVE_REUSE_IMAGE=your-custom-image-id tox -e behave-gcppro-20.04
 ```
-
-## Documentation tests
-
-When working with a `docs` branch and making updates, you can run the documentation tests using the Makefile. The `make` command should display all targets to test.
-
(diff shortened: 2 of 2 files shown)


dev-docs/howtoguides/testing.md
```diff
@@ -246,3 +246,13 @@ a daily ubuntu-advantage-tools image.
 ```sh
 UACLIENT_BEHAVE_REUSE_IMAGE=your-custom-image-id tox -e behave-gcppro-20.04
 ```
+
+## Documentation tests
+
+When working with a `docs` branch and making updates, you can run the documentation tests using the Makefile. The `make` command should display all targets to test.
```
lucasmoura, line 252: I think this is actually displaying the `sphinx` commands, but for this section, we might instead highlight the make run command as it will compile the docs and allow a developer to visualize it.

lucasmoura (commented):
Hi @dheyay, thanks for this addition to the dev-docs. I have just a small suggestion here, but let me know what you think

dev-docs/howtoguides/testing.md
dheyay: Makes sense, I was thinking more in the direction if specific targets want to be run but compiling docs is better than showing sphinx commands, will add that change.

orndorffgrant: We already have some documentation on what the Makefile offers and how to build the docs, including `make html` and `make run` in `dev-docs/documentation/contribute_docs.md`. 

I think this PR introduces two great things shat we should have documented:
1. `make`/`make help` for listing options, including the tests that run as part of CI for docs
2.  the python3-venv dependency

Can you add those to the `contribute_docs.md` file instead of `testing.md` @dheyay?

lucasmoura (approved):

dev-docs/documentation/contribute_docs.md
```diff
@@ -61,6 +61,18 @@ This will build *and serve* the documentation at http://127.0.0.1:8000 -- this
 gives you a live preview of any changes you make (and save) to the
 documentation without needing to continually rebuild.
 
+## Doc testing
+
+Before running the `make` command, make sure to download the python venv module by running the following command:
```
s-makin, line 66:
```suggestion
Before running the `make` command, make sure to download the Python `venv` module by running the following command:
```

dev-docs/documentation/contribute_docs.md
```diff
@@ -61,6 +61,18 @@ This will build *and serve* the documentation at http://127.0.0.1:8000 -- this
...
+```sh
+apt install python3-venv
+```
+When working with a `docs` branch and making updates, you can run the documentation tests using the Makefile.
```
s-makin, line 70: I think this applies to both `docs` and `docs-devel`. I'd prefer us to be specific here, because the implication of "a docs branch" is "any" docs branch. Alternatively, I guess we could say "with our docs-related branches". Either would be clearer, so happy to leave the choice up to you :)

dev-docs/documentation/contribute_docs.md
```diff
@@ -61,6 +61,18 @@ This will build *and serve* the documentation at http://127.0.0.1:8000 -- this
...
 documentation without needing to continually rebuild.
 
+## Doc testing
+
+Before running the `make` command, make sure to download the python venv module by running the following command:
+```sh
+apt install python3-venv
+```
+When working with a `docs` branch and making updates, you can run the documentation tests using the Makefile. 
+The `make` command should list all targets to test. You can run individual targets using the command:
```
s-makin, line 71: It definitely *should* do that. Does it? (I don't usually run the tests, so not sure)

s-makin (commented):
Thanks for making this change. Just a couple of small nits to add :)

dev-docs/documentation/contribute_docs.md
dheyay: Yes, when the `make` command is run, it lists all targets that can be run individually.

s-makin: Nice - in that case, could we tweak this to say the command "lists all targets" rather than "should list all targets"? I think it's a bit clearer then what the expected behaviour is

s-makin (approved):
Thanks for the changes :) LGTM

Issue #2926, opened 2024-01-26
Bug: `pro` execs itself during `pro collect-logs`


orndorffgrant:
**Description of the bug**

`pro` execs itself (`pro status --format json`) during `pro collect-logs`. This results in logs that look like an additional command was run.

**Expected behavior**

`pro collect-logs` can call its own internal functions to get status information.

**Current behavior**

`pro collect-logs` execs `pro status --format json`
The logs look like someone ran both commands.

**To Reproduce**

Please include details on how to reproduce the bug.
1. Launch a VM/Container
2. Run `sudo pro collect-logs`
3. look at `/var/log/ubuntu-advantage.log`

**System information:**

- Ubuntu release: all
- Pro Client version: latest

PR #2935, opened 2024-01-31
Updated pro status call in collect-logs

dheyay:
## Why is this needed?
 
`pro collect-logs` calls itself by running `pro status` to write the current status into a json file. This avoids that by using internal functions to achieve the same outcome.

Fixes: [#2926]

## Test Steps

To re-produce behavior and verify changes:

    Launch a VM/Container
    Run sudo pro collect-logs
    look at /var/log/ubuntu-advantage.log

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/actions.py
+++ b/uaclient/actions.py
@@ -1,5 +1,6 @@
 import datetime
 import glob
+import json
 import logging
 import os
 import re
@@ -235,9 +236,6 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
         "cloud-id", "{}/cloud-id.txt".format(output_dir)
     )
     _write_command_output_to_file(
-        "pro status --format json", "{}/ua-status.json".format(output_dir)
-    )
-    _write_command_output_to_file(
         "{} status".format(livepatch.LIVEPATCH_CMD),
         "{}/livepatch-status.txt".format(output_dir),
     )
@@ -268,6 +266,11 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
             "{}/{}.txt".format(output_dir, service),
             return_codes=[0, 3],
         )
+    pro_status, _ = status(cfg=cfg, show_all=False)
+    system.write_file(
+        "{}/pro-status.json".format(output_dir),
+        json.dumps(pro_status),
+    )
 
     state_files = _get_state_files(cfg)
     user_log_files = (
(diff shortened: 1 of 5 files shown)


uaclient/actions.py
```diff
@@ -268,6 +266,11 @@ def collect_logs(cfg: config.UAConfig, output_dir: str):
             "{}/{}.txt".format(output_dir, service),
             return_codes=[0, 3],
         )
+    pro_status, _ = status(cfg=cfg, show_all=False)
```
lucasmoura, line 269: That works, but when calling the `pro status --format json`, there are some extra steps performed on the status dict before converting it to JSON:
https://github.com/canonical/ubuntu-pro-client/blob/main/uaclient/event_logger.py#L44

Another difference is that `pro status --format json` will always produce a JSON response, even if it errors out. However, I think calling the `_write_command_output_to_file` and directly writing the exception that was raised is a better approach.

But I am unsure for the first point I raised. @orndorffgrant @renanrodrigo do you think we should output the exact same JSON structure as `pro status --format json` ?

orndorffgrant: I don't think we need to keep the exact format, since these are just debug artifacts for us to look at.

It does seem that function adds the pro environment variables, so the current PR loses that information.

@dheyay I think we can add a separate artifact that also saves the output of `util.get_pro_environment` in a file

dheyay: Sure, I'll take a look at that

dheyay:
@orndorffgrant @lucasmoura updated and should be ready

lucasmoura (approved):

orndorffgrant (approved):
Looks and works great! Can you just squash the first two commits together, and the second two commits together? Then we'll have two nice commits that each do one thing. Something like this in an interactive rebase:

```
pick a0f5cb23 collect-logs: Updated pro status call and its  unit tests
fixup d1f57b80 Update behave test feature and ua-status reference
pick 71653098 Environment variables file created in logs
fixup 292bbb41 Update behave tests for env_variable file
```

dheyay:
> Looks and works great! Can you just squash the first two commits together, and the second two commits together? Then we'll have two nice commits that each do one thing. Something like this in an interactive rebase:
> 
> ```
> pick a0f5cb23 collect-logs: Updated pro status call and its  unit tests
> fixup d1f57b80 Update behave test feature and ua-status reference
> pick 71653098 Environment variables file created in logs
> fixup 292bbb41 Update behave tests for env_variable file
> ```

Done

Issue #2598, opened 2023-05-26
all recommendations to run `pro` in messaging should include `sudo` if required


orndorffgrant:
We should check all messages that suggest running `pro` and make sure `sudo` is there for ones that require root.

At least one is known: the livepatch kernel support message that suggests `pro disable livepatch`.

orndorffgrant:
fixed in https://github.com/canonical/ubuntu-pro-client/pull/2936

PR #2936, opened 2024-01-31
Update to message recommendations asking to run pro

dheyay:
## Why is this needed?

Ensure 'sudo' is included for commands running `pro` which require root access.
Fixes: #2598

## Checklist

 - [ ] Yes
 - [x] No


lucasmoura:
There are two things missing on this PR:

1) We need to update the existing integration tests that check that message:
https://github.com/canonical/ubuntu-pro-client/blob/main/features/livepatch.feature#L54

2) We need to update the translation files. To do that, just run the `update-pos.sh` script inside the `tools` folder and commit the result. But double check that it has not touched any other message besides the one that was modified.

features/livepatch.feature
```diff
@@ -51,7 +51,7 @@ Feature: Livepatch
         NOTICES
         The current kernel \(5.4.0-(\d+)-kvm, x86_64\) is not supported by livepatch.
         Supported kernels are listed here: https://ubuntu.com/security/livepatch/docs/kernels
-        Either switch to a supported kernel or `pro disable livepatch` to dismiss this warning.
+        Either switch to a supported kernel or `sudo pro disable livepatch` to dismiss this warning.
```
lucasmoura, line 54: We also need to add this change to line 68

orndorffgrant (commented):
Even though it is a logging statement, I think the `pro config set` suggestion in `config.py` around line 588 should have a `sudo` as well.

Otherwise this looks good!

dheyay:
Should be good  to merge now

uaclient/config.py
```diff
@@ -585,7 +585,7 @@ def warn_about_invalid_keys(self):
             LOG.warning('legacy "ua_config" found in uaclient.conf')
             LOG.warning("Please do the following:")
             LOG.warning(
-                "  1. run `pro config set field=value` for each"
+                "  1. run `sudo pro config set field=value` for each"
```
lucasmoura, line 588: It seems we also have that same suggestion on `postinst`. I think we should update it there as well

dheyay: Should be good now @lucasmoura, thanks!

orndorffgrant (approved):
Thank you!

PR #2937, opened 2024-02-01
daemon: wait for cloud-config.service to complete in python

orndorffgrant:
currently based on https://github.com/canonical/ubuntu-pro-client/pull/2887 so only look at the last two commit

## Why is this needed?

When cloud-init gets "ubuntu_advantage" user-data, it may need to
install or upgrade ubuntu-advantage-tools. This happens during
cloud-config.service. The install/upgrade of ubuntu-advantage-tools
triggers a "systemctl start ubuntu-advantage.service".
ubuntu-advantage.service is configured with an
"After=cloud-config.service", so it waits to start until
cloud-config.service becomes "active". In this scenario,
cloud-config.service cannot continue until ubuntu-advantage.service
starts, creating deadlock.

We get around this deadlock by removing the "After=cloud-config.service"
from ubuntu-advantage.service. Instead we are moving the ordering dependency
enforcement into the python code called by the service. This allows
systemd to go ahead and start the service allowing the install of
ubuntu-advantage-tools to finish and cloud-init to continue.
ubuntu-advantage.service will wait until cloud-config.service is
"active" to actually do anything.

LP: #2050022

## Test Steps
Run the added behave tests with `-D install_from=ppa:orndorffgrant/proclient-cloudinit-deadlock-test`

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/lib/daemon.py
+++ b/lib/daemon.py
@@ -1,13 +1,42 @@
 import logging
 import os
 import sys
+import time
 
-from uaclient import http
+from uaclient import http, system
 from uaclient.config import UAConfig
 from uaclient.daemon import poll_for_pro_license, retry_auto_attach
 from uaclient.log import setup_journald_logging
 
 LOG = logging.getLogger("ubuntupro.daemon")
+
+
+# 10 seconds times 120 = 20 minutes
+WAIT_FOR_CLOUD_CONFIG_SLEEP_TIME = 10
+WAIT_FOR_CLOUD_CONFIG_POLL_TIMES = 120
+
+
+def _wait_for_cloud_config():
+    LOG.debug("waiting for cloud-config.service to finish")
+    for i in range(WAIT_FOR_CLOUD_CONFIG_POLL_TIMES + 1):
+        state = system.get_systemd_unit_active_state("cloud-config.service")
+        LOG.debug("cloud-config.service state: %r", state)
+        if state is not None and state == "activating":
+            if i < WAIT_FOR_CLOUD_CONFIG_POLL_TIMES:
+                LOG.debug(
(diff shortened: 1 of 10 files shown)


lib/daemon.py
```diff
@@ -1,15 +1,44 @@
...
+                )
+                time.sleep(WAIT_FOR_CLOUD_CONFIG_SLEEP_TIME)
+            else:
+                LOG.warning(
+                    "cloud-config.service is still activating after "
+                    "20 minutes. continuing anyway"
+                )
+                return
+        else:
+            LOG.debug("cloud-config.service is not activating. continuing")
```
lucasmoura, line 38: should we log the state the service is actually in ? Maybe it is worth detecting an inactive vs failed service

lucasmoura (approved):
I think that's a great solution.

I have just one nit for the logging and I think we need unittests for the new functions. But besides that, LGTM

lib/daemon.py
orndorffgrant: There is a separate log statement above on line 23 that includes the current state

lucasmoura (approved):
I can confirm the tests are working as expected.
However, is there a reason why we are not testing it in xenial ?

orndorffgrant:
I was unable to reproduce the bug on xenial, likely because of the older version of cloudinit on xenial. But let me try a little harder with some more scenarios just to be sure.

orndorffgrant:
I was able to reproduce the deadlock on xenial using an older cloud-init user data field name and the test shows that the solution works on xenial too. I'll merge this unless you notice anything else @lucasmoura

orndorffgrant:
CI fail unrelated

PR #2941, opened 2024-02-07
Notice permissions

dheyay:
## Why is this needed?

Prevents crashes when executing `pro status` as non-root due to encountering temporary files with restrictive permissions in the notices directory. 

Added a check for unreadable files when listing notices from the notices directory.

## Checklist

 - [ ] Yes
 - [x] No


features/notice_permission.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Notice Permission
```
orndorffgrant, line 1: I think this test can go in the existing `unattached_status.feature` file.

features/notice_permission.feature
```diff
@@ -0,0 +1,33 @@
...
+
+    Scenario Outline: Check notice file read permission
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+        When I run `rm -rf /run/ubuntu-advantage` with sudo
+        When I run `mkdir /run/ubuntu-advantage` with sudo
+        When I run `mkdir /run/ubuntu-advantage/notices` with sudo
+        When I run `touch /run/ubuntu-advantage/notices/crasher` with sudo
+        When I run `chmod 0 /run/ubuntu-advantage/notices/crasher` with sudo
```
orndorffgrant, line 9: I suggest doing this for `/var/lib/ubuntu-advantage/notices` as well

features/notice_permission.feature
```diff
@@ -0,0 +1,33 @@
...
+        fips-updates    +yes       +FIPS compliant crypto packages with stable security updates
+        livepatch       +yes      +(Canonical Livepatch service|Current kernel is not supported)
+        ros             +yes       +Security Updates for the Robot Operating System
+        ros-updates     +yes       +All Updates for the Robot Operating System
+
+        For a list of all Ubuntu Pro services, run 'pro status --all'
+
+        This machine is not attached to an Ubuntu Pro subscription.
+        See https://ubuntu.com/pro
+        """
```
orndorffgrant, line 29: This output will change over time and break this test even though the functionality we're testing here will still be working properly.

I suggest we don't check all of this output. Instead, also add a valid notice file in /run/ubuntu-advantage/notices. Then only check something like:
```
Then stdout contains substring:
"""
NOTICES
The notice that you added that we actually expect

"""
```

orndorffgrant (commented):
Looking good! A couple suggestions

features/notice_permission.feature
dheyay: Sure let me take a look at these changes!

uaclient/files/notices.py
```diff
@@ -173,35 +216,20 @@ def list(self) -> List[str]:
...
-                for file_name in os.listdir(notice_directory)
-                if os.path.isfile(os.path.join(notice_directory, file_name))
-            ]
+            notice_file_names = self._get_notice_file_names(notice_directory)
             for notice_file_name in notice_file_names:
+                if not self._is_readable(notice_directory, notice_file_name):
+                    continue
                 notice_file_contents = system.load_file(
                     os.path.join(notice_directory, notice_file_name)
                 )
```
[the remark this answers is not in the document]
dheyay: Yeah that seems like a great idea, thanks!

dheyay:
@orndorffgrant Should be good for another look

uaclient/files/notices.py
```diff
@@ -160,6 +160,56 @@ def remove(self, notice_details: Notice):
...
+        :returns: True if the file is valid, False otherwise.
+        """
+        try:
+            order_id, label = file_name.split("-")
+            for n in Notice:
+                if n.order_id == order_id and n.label == label:
+                    return True
+            return False
+        except ValueError:
+            return False
```
orndorffgrant, line 189: We can make this more restrictive by using the `is_permanent` field. permanent notices should only appear under `/var/lib/...` and non-permanent notices should only appear under `/run/...`

uaclient/files/notices.py
```diff
@@ -160,6 +160,56 @@ def remove(self, notice_details: Notice):
...
+
+    def _is_readable(self, directory: str, file_name: str) -> bool:
+        """Checks if the notice file is readable.
+
+        :param directory: The directory where the notice file is located.
+        :param file_name: The name of the notice file.
+        :returns: True if the file is readable, False otherwise.
+        """
+        return os.access(os.path.join(directory, file_name), os.R_OK)
```
orndorffgrant, line 198: We can delete this now

orndorffgrant (commented):
This is looking great! Just some small nits

uaclient/files/notices.py
dheyay: @orndorffgrant Should be fixed and ready for review

uaclient/files/notices.py
```diff
@@ -160,6 +160,60 @@ def remove(self, notice_details: Notice):
...
+                        or (
+                            directory == defaults.NOTICES_TEMPORARY_DIRECTORY
+                            and not n.is_permanent
+                        )
+                    )
+                ):
+                    return True
+            return False
+        except ValueError:
+            return False
```
orndorffgrant, line 202: I want to point out an alternative way of implementing this to consider.
```python
is_permanent_dir = directory == defaults.NOTICES_PERMANENT_DIRECTORY
valid_file_names = {
    "{}-{}".format(n.order_id, n.label)
    for n in Notice
    if n.is_permanent == is_permanent_dir
}
return file_name in valid_file_names
```

But the current method is good and works too

uaclient/files/notices.py
```diff
@@ -160,6 +160,60 @@ def remove(self, notice_details: Notice):
...
+
+    def _get_default_message(self, file_name: str) -> str:
+        """Gets the default message for a notice file.
+
+        :param file_name: The name of the notice file.
+        :returns: The default message defined in the enum.
+        """
+        order_id, label = file_name.split("-")
+        notice = None
```
orndorffgrant, line 211:
```suggestion
```
minor

features/unattached_status.feature
```diff
@@ -599,3 +599,26 @@ Feature: Unattached status
...
+
+    Scenario Outline: Check notice file read permission
+        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+        When I run `mkdir -p /run/ubuntu-advantage/notices` with sudo
+        When I run `touch /run/ubuntu-advantage/notices/crasher` with sudo
+        When I run `chmod 0 /run/ubuntu-advantage/notices/crasher` with sudo
+        When I run `mkdir -p /var/lib/ubuntu-advantage/notices` with sudo
+        When I run `touch /var/lib/ubuntu-advantage/notices/crasher` with sudo
+        When I run `chmod 0 /var/lib/ubuntu-advantage/notices/crasher` with sudo
+        When I run `touch /var/lib/ubuntu-advantage/notices/10-reboot_required` with sudo
```
orndorffgrant, line 611: Now that we're checking for valid notices, this one doesn't show.
This notice is a non-permanent one
```suggestion
        When I run `touch /run/ubuntu-advantage/notices/10-reboot_required` with sudo
```

orndorffgrant (approved):
Thanks for this! Pre-emptively approving, but please:
1. Squash all the commits - make sure the "Fixes" line is still in the final commit
2. Address the test comment and "minor" comment below. The refactor suggestion is optional.

Then we can land it :rocket:

dheyay:
> Thanks for this! Pre-emptively approving, but please:
> 
>     1. Squash all the commits - make sure the "Fixes" line is still in the final commit
> 
>     2. Address the test comment and "minor" comment below. The refactor suggestion is optional.
> 
> 
> Then we can land it 🚀

Should be good now! @orndorffgrant

orndorffgrant:
CI blocked on noble flake8 failure
I tested the new test locally and it passed

PR #2942, opened 2024-02-08
logs: fix redaction filter

lucasmoura:
## Why is this needed?
The redaction filter was not correctly applied to our current logs. We are now fixing this by adding the redaction filter directly in the log file handler

## Test Steps
Run the modified integration tests and also confirm that when performing an attach operation, that all secrets are redacted in the logs

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/features/attached_commands.feature
+++ b/features/attached_commands.feature
@@ -4,6 +4,9 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
     Scenario Outline: Attached refresh in a ubuntu machine
         Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
+        And I verify that `Bearer ` field is redacted in the logs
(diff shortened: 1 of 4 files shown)


features/attached_commands.feature
```diff
@@ -4,6 +4,9 @@ Feature: Command behaviour when attached to an Ubuntu Pro subscription
     Scenario Outline: Attached refresh in a ubuntu machine
         Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
         When I attach `contract_token` with sudo
+        And I verify that `Bearer ` field is redacted in the logs
```
renanrodrigo, line 7: gold

renanrodrigo (approved):

features/attached_commands.feature
orndorffgrant: :facepalm: we should've added this ages ago

orndorffgrant (approved):
CI fail is unrelated - thank you!

Issue #2809, opened 2023-10-24
Bug: `pro config show` as non-root can be inaccurate


orndorffgrant:
**Description of the bug**

`pro config show` as non-root can be inaccurate when the root user changed a setting

**Expected behavior**

`pro config show` as non-root always shows the correct settings, with secrets redacted

**Current behavior**

`pro config show` as non-root always shows the defaults

**To Reproduce**

Please include details on how to reproduce the bug.
1. Launch a VM/Container
2. Run `sudo pro config set field=value`
3. Run `pro config show` and you won't see the new value

**Additional context**

We can't just make user-config.json world-readable (as in https://github.com/canonical/ubuntu-pro-client/pull/2804) because it may contain proxy secrets. We will likely need a private and public version, similar to our machine-token.json split

PR #2945, opened 2024-02-14
Config show inaccuracy

dheyay:
## Why is this needed?

`pro config show` command inaccurately displays default settings instead of reflecting changes made by the root user. Proxy secrets set by root were also not redacted when displaying to non root user.

Fixes: [#2809]

Issue #2982 raised for migration of user configs when merging next-v32 into main.

## Checklist

- Added behave test to config.feature
- Added unit test to test_config.py

 - [ ] Yes
 - [ ] No

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -305,7 +305,6 @@ create_public_user_config_file() {
         mkdir -p "$destination_dir"
         # Move the user-config.json file to the private directory
         mv "$source_file" "$destination_dir/user-config.json"
-    fi
 
     /usr/bin/python3 -c "
 from uaclient.files import UserConfigFileObject
@@ -316,6 +315,7 @@ try:
 except Exception as e:
     print('Error while creating public user-config file: {}'.format(e))
 "
+    fi
 }
 
 migrate_ubuntu_pro_beta_banner() {
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -296,11 +296,22 @@ create_public_user_config_file() {
     # UserConfigFile we already write the public
     # version of the user-config file with all the
     # sensitive data removed.
+    # We also move the user-config.json file to the private 
+    # directory
+    source_file="/var/lib/ubuntu-advantage/user-config.json"
+    destination_dir="/var/lib/ubuntu-advantage/private"
+    # Check if the source file exists
+    if [ -f "$source_file" ]; then
+        mkdir -p "$destination_dir"
+        # Move the user-config.json file to the private directory
+        mv "$source_file" "$destination_dir/user-config.json"
+    fi
+
     /usr/bin/python3 -c "
-from uaclient.files import UserConfigFileObject, UserConfigData
+from uaclient.files import UserConfigFileObject
 try:
     user_config_file = UserConfigFileObject()
-    content = UserConfigData()
+    content = user_config_file.read()
     user_config_file.write(content)
 except Exception as e:
     print('Error while creating public user-config file: {}'.format(e))
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -291,6 +291,22 @@ machine_token_file.write(content)
 "
 }
 
+create_public_user_config_file() {
+    # When we perform the write operation through
+    # UserConfigFile we already write the public
+    # version of the user-config file with all the
+    # sensitive data removed.
+    /usr/bin/python3 -c "
+from uaclient.files import UserConfigFileObject, UserConfigData
+try:
+    user_config_file = UserConfigFileObject()
+    content = UserConfigData()
+    user_config_file.write(content)
+except Exception as e:
+    print('Error while creating public user-config file: {}'.format(e))
+"
+}
+
 migrate_ubuntu_pro_beta_banner() {
     # This only shipped in 27.11.2~
     if dpkg --compare-versions "$PREVIOUS_PKG_VER" ge "27.11.2~" \
@@ -473,6 +489,10 @@ case "$1" in
       # Rename the gpg keys to -pro-
       if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "29~"; then
           rename_gpg_keys
+      fi
+
+      if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "32~"; then
+          create_public_user_config_file
       fi
 
       if grep -q "^ua_config:" /etc/ubuntu-advantage/uaclient.conf; then
--- a/features/proxy_config.feature
+++ b/features/proxy_config.feature
@@ -216,118 +216,41 @@ Feature: Proxy configuration
       """
       No proxy set in config; however, proxy is configured for: snap, livepatch.
       See https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/howtoguides/configure_proxies.html for more information on pro proxy configuration.
-      
-      Examples: ubuntu release
+
+      Successfully processed your pro configuration.
+      """
+    When I create the file `/var/lib/ubuntu-advantage/user-config.json` with the following:
+      """
+      {
+        "http_proxy": "invalidurl",
+        "https_proxy": "invalidurls"
+      }
+      """
+    And I apt install `python3-pycurl`
+    And I verify that running `pro refresh config` `with sudo` exits `1`
+    Then stderr matches regexp:
+      """
+      \"invalidurl\" is not a valid url. Not setting as proxy.
+      """
+    When I create the file `/var/lib/ubuntu-advantage/user-config.json` with the following:
+      """
+      {
+        "https_proxy": "https://localhost:12345"
+      }
+      """
+    And I verify that running `pro refresh config` `with sudo` exits `1`
+    Then stderr matches regexp:
+      """
+      \"https://localhost:12345\" is not working. Not setting as proxy.
+      """
+
+    Examples: ubuntu release
       | release | machine_type |
       | xenial  | lxd-vm       |
       | bionic  | lxd-vm       |
 
-    @slow
-    Scenario Outline: Attach and config show command when authenticated proxy is configured for uaclient
-        Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
-        Given a `focal` `lxd-container` machine named `proxy`
-        When I apt install `squid apache2-utils` on the `proxy` machine
-        And I run `htpasswd -bc /etc/squid/passwordfile someuser somepassword` `with sudo` on the `proxy` machine
-        And I add this text on `/etc/squid/squid.conf` on `proxy` above `http_access deny all`:
-        """
-        dns_v4_first on\nauth_param basic program \/usr\/lib\/squid\/basic_ncsa_auth \/etc\/squid\/passwordfile\nacl topsecret proxy_auth REQUIRED\nhttp_access allow topsecret
-        """
-        And I run `systemctl restart squid.service` `with sudo` on the `proxy` machine
-        And I run `truncate -s 0 /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        And I verify `/var/log/squid/access.log` is empty on `proxy` machine
-        When I run `pro config set https_proxy=http://someuser:somepassword@$behave_var{machine-ip proxy}:3128` with sudo
-        Then stdout matches regexp:
-        """
-        Setting snap proxy
-        """
-        When I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        Then stdout matches regexp:
-        """
-        .*CONNECT api.snapcraft.io.*
-        """
-        When I run `pro config set http_proxy=http://someuser:somepassword@$behave_var{machine-ip proxy}:3128` with sudo
-        Then stdout matches regexp:
-        """
-        Setting snap proxy
-        """
-        When I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        Then stdout matches regexp:
-        """
-        .*HEAD http://api.snapcraft.io.*
-        """
-        When I attach `contract_token` with sudo
-        And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        Then stdout matches regexp:
-        """
-        .*CONNECT contracts.canonical.com.*
-        """
-        When I run `truncate -s 0 /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        And I verify `/var/log/squid/access.log` is empty on `proxy` machine
-        When I run `pro config set ua_apt_http_proxy=http://someuser:somepassword@$behave_var{machine-ip proxy}:3128` with sudo
-        Then stdout matches regexp:
-        """
-        Setting UA-scoped APT proxy
-        """
-        When I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        Then stdout matches regexp:
-        """
-        .*HEAD http://archive.ubuntu.com.*
-        """
-        When I run `pro config set ua_apt_https_proxy=http://someuser:somepassword@$behave_var{machine-ip proxy}:3128` with sudo
-        Then stdout matches regexp:
-        """
-        Setting UA-scoped APT proxy
-        """
-        When I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        Then stdout matches regexp:
-        """
-        .*CONNECT esm.ubuntu.com.*
-        """
-        When I run `pro refresh config` with sudo
-        And I apt update
-        And I run `cat /var/log/squid/access.log` `with sudo` on the `proxy` machine
-        Then stdout matches regexp:
-        """
-        CONNECT esm.ubuntu.com:443
-        """
-        Then stdout does not match regexp:
-        """
-        .*GET.*ubuntu.com/ubuntu/dists.*
-        """
-        Then stdout does not match regexp:
-        """
-        .*GET.*archive.ubuntu.com.*
-        """
-        Then stdout does not match regexp:
-        """
-        .*GET.*security.ubuntu.com.*
-        """
-        And I verify that running `pro config set ua_apt_https_proxy=http://wronguser:wrongpassword@$behave_var{machine-ip proxy}:3128` `with sudo` exits `1`
-        Then stderr matches regexp:
-        """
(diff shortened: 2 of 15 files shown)


features/config.feature
```diff
@@ -50,3 +50,39 @@ Feature: pro config sub-command
             | xenial  | lxd-container |
             | jammy   | lxd-container |
             | mantic  | lxd-container |
+
+    Scenario Outline: Proxy information is redacted when using pro config show
```
lucasmoura, line 54: @dheyay I think we can already reuse one of the existing proxy tests for this check, just to avoid creating another instance just for this feature

uaclient/config.py
```diff
@@ -698,3 +709,38 @@ def new_f():
...
+    except Exception as e:
+        LOG.warning("Error loading user config", exc_info=e)
+        LOG.warning("Using default config values")
+        return None
+
+
+def write_user_config(user_config: state_files.UserConfigData) -> None:
+    """Write the user-config.json file with the given contents."""
+    state_files.user_config_file_public.write(user_config)
+    state_files.user_config_file_private.write(user_config)
```
lucasmoura, line 746: I think we might use a similar structure to what we have for the `machine-token` file:
https://github.com/canonical/ubuntu-pro-client/blob/main/uaclient/files/files.py#L74

You can see that this class is responsible for handling both the public and private version of the machine token file. This would simplify the config module, as it wouldn't need to know about the existence of the separate files

lucasmoura (commented):
@dheyay we also need a plan to adapt existing users to this new file separation. We will probably need a function on the package `postinst` file that will generate the public config file.

features/config.feature
dheyay: Sure, will find a better place for this test

uaclient/config.py
dheyay: @lucasmoura should be good for another look

uaclient/tests/test_log.py
```diff
@@ -197,7 +197,7 @@ def test_get_user_or_root_log_file_path(
         m_we_are_currently_root.return_value = we_are_currently_root
         result = pro_log.get_user_or_root_log_file_path()
         # ensure mocks are used properly
-        assert m_we_are_currently_root.call_count == 1
+        # assert m_we_are_currently_root.call_count == 2
```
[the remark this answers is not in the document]
dheyay: Error on my side, there is no need for the comment

debian/ubuntu-advantage-tools.postinst
```diff
@@ -291,6 +291,19 @@ machine_token_file.write(content)
 "
 }
 
+create_public_user_config_file() {
```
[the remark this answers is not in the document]
dheyay: Oh I see, will fix that. 
We want the integration test to check that the public user config file is correctly created right?

uaclient/files/user_config_file.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        self._public = DataObjectFile(
+            UserConfigData,
+            UAFile(file_name, directory, private=False),
+            DataObjectFileFormat.JSON,
+            optional_type_errors_become_null=True,
+        )
+
+    def redact_config_data(
+        self, user_config: Optional[UserConfigData]
```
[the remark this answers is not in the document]
dheyay: The input should not be optional here. Will correct this.

uaclient/files/user_config_file.py
```diff
@@ -0,0 +1,139 @@
...
+            LOG.warning("Error reading user config", exc_info=e)
+            return None
+
+    def write(self, content: UserConfigData):
+        if not util.we_are_currently_root():
+            raise exceptions.NonRootUserError()
+
+        self._private.write(content)
+        redacted_content = self.redact_config_data(content)
+        self._public.write(redacted_content or content)
```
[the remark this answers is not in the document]
dheyay: Currently the redact_config_data only redacts config in case the value is not None, else it returns None. I can change that inside the redact function so we can get rid of the `or`.

uaclient/files/user_config_file.py
```diff
@@ -0,0 +1,139 @@
...
+            if util.we_are_currently_root():
+                return self._private.read()
+            return self._public.read()
+        except Exception as e:
+            LOG.warning("Error reading user config", exc_info=e)
+            return None
+
+    def write(self, content: UserConfigData):
+        if not util.we_are_currently_root():
+            raise exceptions.NonRootUserError()
```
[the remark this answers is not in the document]
dheyay: This was added as a precaution in case a non-root user tries setting pro config. It seems that this case is  covered by the cli so this can be removed.

features/proxy_config.feature
```diff
@@ -334,7 +334,18 @@ Feature: Proxy configuration
...
+        When I run `pro config set http_proxy=http://someuser:somepassword@$behave_var{machine-ip proxy}:3128` with sudo
+        When I run `pro config show` with sudo
+        Then stdout matches regexp:
+        """
+        http_proxy              http://someuser:somepassword@$behave_var{machine-ip proxy}:3128
+        """
+        When I run `pro config show` as non-root
+        Then stdout matches regexp:
+        """
+        http_proxy              <REDACTED>
```
orndorffgrant, line 346:
```suggestion
        http_proxy              http://<REDACTED>:<REDACTED>@$behave_var{machine-ip proxy}:3128
```

orndorffgrant: @dheyay I changed my mind. The urlparse function in the stdlib is not nice enough to make it easy to reconstruct the url with redacted parts, and it's not worth the risk of a bug here to try to do it ourselves via regex or otherwise. Redacting the whole proxy url is good.

uaclient/files/user_config_file.py
```diff
@@ -0,0 +1,134 @@
...
+    ) -> UserConfigData:
+        redacted_data = copy.deepcopy(user_config)
+        for field in PROXY_FIELDS:
+            value = getattr(redacted_data, field)
+            if value:
+                setattr(
+                    redacted_data,
+                    field,
+                    "<REDACTED>",
+                )
```
orndorffgrant, line 116: I think we only need to redact if there is a username or password present. `urlparse` can be used to check for for the existence of those fields, then we can just redact the whole URL if one of them is present. What do you think @lucasmoura @dheyay?

dheyay: should be good I think. 

do we have cases where root user set's the proxy value and it does not contain username, password? In that case this should be fine as long as we want the non-root user to be able to see such proxy values. @orndorffgrant

orndorffgrant: Yes that is exactly the scenario I think we should support

debian/ubuntu-advantage-tools.postinst
```diff
@@ -278,6 +278,20 @@ remove_old_systemd_units() {
...
+    destination_dir="/var/lib/ubuntu-advantage/private"
+
+    # Check if the source file exists
+    if [ -f "$source_file" ]; then
+        mkdir -p "$destination_dir"
+        # Move the user-config.json file to the private directory
+        mv "$source_file" "$destination_dir/user-config.json"
+    fi
+}
+
```
orndorffgrant, line 294: No need for this to be a separate function. We should have one function to handle the complete migration of the old user-config.json to the new split user-config.json

debian/ubuntu-advantage-tools.postinst
```diff
@@ -291,6 +305,22 @@ machine_token_file.write(content)
 "
 }
 
+create_public_user_config_file() {
+    # When we perform the write operation through
+    # UserConfigFile we already write the public
+    # version of the user-config file with all the
+    # sensitive data removed.
+    /usr/bin/python3 -c "
+from uaclient.files import UserConfigFileObject, UserConfigData
```
orndorffgrant, line 314:
```suggestion
from uaclient.files import UserConfigFileObject
```

sru/release-32/test-user-config-created.sh
```diff
@@ -0,0 +1,78 @@
...
+echo "###########################################"
+lxc exec $name -- apt-cache policy ubuntu-advantage-tools
+echo -e "###########################################\n"
+
+# Create user-config.json file with the given content
+lxc exec $name -- sh -c "echo '{\"http_proxy\": \"http://someuser:somepassword@example.com:3128\"}' > /var/lib/ubuntu-advantage/user-config.json"
+if lxc exec $name -- test -e /var/lib/ubuntu-advantage/user-config.json; then
+  user_config_contents=$(lxc exec $name -- cat /var/lib/ubuntu-advantage/user-config.json)
+  http_proxy_value=$(echo "$user_config_contents" | jq -r '.http_proxy')
+fi
```
orndorffgrant, line 37: This is cool, but we already know the expected value - we don't need to parse it from the file.
```suggestion
http_proxy_value="http://someuser:somepassword@example.com:3128
# Create user-config.json file with the given content
lxc exec $name -- sh -c "echo '{\"http_proxy\": \"$http_proxy_value\"}' > /var/lib/ubuntu-advantage/user-config.json"
```

sru/release-32/test-user-config-created.sh
```diff
@@ -0,0 +1,78 @@
...
+  lxc exec $name -- apt-get update > /dev/null
+  lxc exec $name -- apt-get install ubuntu-advantage-tools -y > /dev/null
+else
+  lxc file push $install_from $name/new-ua.deb
+  lxc exec $name -- dpkg -i /new-ua.deb > /dev/null
+fi
+# ----------------------------------------------------------------
+
+# Check if user-config.json is moved to the private directory
+if lxc exec $name -- test -e /var/lib/ubuntu-advantage/private/user-config.json; then
```
orndorffgrant, line 55: Currently the test will only test the migration worked if the new files exist. So if the migration fails to even create the new files, then the test will pass :upside_down_face:.
Instead, we should assert that the new files were created and then test the contents of them.
```suggestion
lxc exec $name -- test -e /var/lib/ubuntu-advantage/private/user-config.json
```

sru/release-32/test-user-config-created.sh
```diff
@@ -0,0 +1,78 @@
...
+  private_config_contents=$(lxc exec $name -- cat /var/lib/ubuntu-advantage/private/user-config.json)
+  private_http_proxy_value=$(echo "$private_config_contents" | jq -r '.http_proxy')
+  # Check if the contents are the same as the previous contents
+  if [ "$http_proxy_value" == "$private_http_proxy_value" ]; then
+    echo "Contents of private/user-config.json have not changed"
+  fi
+  # Check if the file permissions are root
+  lxc exec $name -- stat -c %U /var/lib/ubuntu-advantage/private/user-config.json | grep -q "root"
```
orndorffgrant, line 63: This checks that the owner is root, but does not check the rwx permissions of the file. Even the public file will be owned by root
```suggestion
  lxc exec $name -- stat -c %A /var/lib/ubuntu-advantage/private/user-config.json | grep -q "-rw-------"
```

sru/release-32/test-user-config-created.sh
```diff
@@ -0,0 +1,78 @@
...
+# Check if a new public file is created
+if lxc exec $name -- test -e /var/lib/ubuntu-advantage/user-config.json; then
+  public_config_contents=$(lxc exec $name -- cat /var/lib/ubuntu-advantage/user-config.json)
+  public_http_proxy_value=$(echo "$public_config_contents" | jq -r '.http_proxy')
+  # Check if the public_http_proxy_value is "<REDACTED>"
+  if [ "$public_http_proxy_value" = "<REDACTED>" ]; then
+    echo "public_http_proxy_value is <REDACTED>"
+  fi
+  # Check if file permissions are public
+  lxc exec $name -- stat -c %A /var/lib/ubuntu-advantage/user-config.json | grep -q "rw-rw-rw-"
```
orndorffgrant, line 75: I don't think the file should be world-writable - if it is then we should fix that.
```suggestion
  lxc exec $name -- stat -c %A /var/lib/ubuntu-advantage/user-config.json | grep -q "-rw-r--r--"
```

sru/release-32/test-user-config-created.sh
```diff
@@ -0,0 +1,78 @@
...
+  # Check if the contents are the same as the previous contents
+  if [ "$http_proxy_value" == "$private_http_proxy_value" ]; then
+    echo "Contents of private/user-config.json have not changed"
+  fi
+  # Check if the file permissions are root
+  lxc exec $name -- stat -c %U /var/lib/ubuntu-advantage/private/user-config.json | grep -q "root"
+fi
+
+# Check if a new public file is created
+if lxc exec $name -- test -e /var/lib/ubuntu-advantage/user-config.json; then
```
orndorffgrant, line 67:
```suggestion
lxc exec $name -- test -e /var/lib/ubuntu-advantage/user-config.json
```

uaclient/files/user_config_file.py
```diff
@@ -103,6 +103,12 @@ def __init__(self, directory: str = defaults.DEFAULT_DATA_DIR):
             optional_type_errors_become_null=True,
         )
 
+    def get_public_config(self):
+        public_config = self._public.read()
+        if public_config is None:
+            public_config = UserConfigData()
+        return public_config
```
lucasmoura, line 110: Just to be more pythonic, let's turn this into a `property`

uaclient/files/user_config_file.py
```diff
@@ -119,14 +125,15 @@ def redact_config_data(
...
-            return None
+    def read(self) -> UserConfigData:
+        if util.we_are_currently_root():
+            private_config = self._private.read()
+            if private_config is not None:
+                return private_config
+        public_config = self._public.read()
+        if public_config is not None:
+            return public_config
+        return UserConfigData()
```
lucasmoura, line 136: Why are we returning an empty `UserConfigData` object here ? Is it just for us to have the default values in place ?

uaclient/files/user_config_file.py
lucasmoura: And let's also add type hints for this method

uaclient/files/user_config_file.py
dheyay: Yes, to have the default values in place in case there are no files to read so the Config always has a user_config set.

renanrodrigo (approved):
I think this is good after the changes.
I am slightly uncomfortable with the `files/user_config_file` new file comparing to what we have in the files folder... But how that folder is organized is a matter we can talk about later.
+1

sru/release-32/test-user-config-created.sh
```diff
@@ -0,0 +1,74 @@
...
+# Check if user-config.json is moved to the private directory
+lxc exec $name -- test -e /var/lib/ubuntu-advantage/private/user-config.json;
+private_config_contents=$(lxc exec $name -- cat /var/lib/ubuntu-advantage/private/user-config.json)
+private_http_proxy_value=$(echo "$private_config_contents" | jq -r '.http_proxy')
+# Check if the contents are the same as the previous contents
+if [ "$http_proxy_value" == "$private_http_proxy_value" ]; then
+  echo "Contents of private/user-config.json have not changed"
+fi
+# Check if the file permissions are root
+lxc exec $name -- stat -c %A /var/lib/ubuntu-advantage/private/user-config.json | grep -- "-rw-------"
```
lucasmoura, line 61: Let's just add a message before each  grep command to make it easier to visualize in the output which file we are testing at the moment

lucasmoura (approved):
Just a tiny nit, but LGTM.

Thanks for all the work here @dheyay

sru/release-32/test-user-config-created.sh
dheyay: Changes should be added. Thanks again @lucasmoura for all the help

lucasmoura: Awesome. Just another hint, we don't need to execute the `echo` command inside the container, as it is just an "status" message.

dheyay: Oh yeah, I might have missed that. We are just printing to terminal. Should be updated.

uaclient/files/__init__.py
```diff
@@ -1,9 +1,15 @@
 from uaclient.files.data_types import DataObjectFile, DataObjectFileFormat
 from uaclient.files.files import MachineTokenFile, UAFile
+from uaclient.files.user_config_file import (
+    UserConfigData,
```
orndorffgrant, line 4: This isn't used anywhere AFAICT
```suggestion
```

debian/ubuntu-advantage-tools.postinst
```diff
@@ -291,6 +291,33 @@ machine_token_file.write(content)
...
+
+    /usr/bin/python3 -c "
+from uaclient.files import UserConfigFileObject
+try:
+    user_config_file = UserConfigFileObject()
+    content = user_config_file.read()
+    user_config_file.write(content)
+except Exception as e:
+    print('Error while creating public user-config file: {}'.format(e))
+"
```
orndorffgrant, line 318: Let's move this inside the if statement to make sure it only runs if necessary
```suggestion
        /usr/bin/python3 -c "
from uaclient.files import UserConfigFileObject
try:
    user_config_file = UserConfigFileObject()
    content = user_config_file.read()
    user_config_file.write(content)
except Exception as e:
    print('Error while creating public user-config file: {}'.format(e))
"
    fi
```

orndorffgrant (approved):
Just some little things, then let's merge

uaclient/status.py
```diff
@@ -394,9 +394,9 @@ def _get_config_status(cfg) -> Dict[str, Any]:
         "features": cfg.features,
     }
     # LP: #2004280 maintain backwards compatibility
-    ua_config = {}
+    ua_config = user_config_file.user_config.public_config.to_dict()
     for key in UA_CONFIGURABLE_KEYS:
-        if hasattr(cfg, key):
+        if hasattr(cfg, key) and ua_config[key] is None:
             ua_config[key] = getattr(cfg, key)
     ret["config"]["ua_config"] = ua_config
```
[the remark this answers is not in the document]
dheyay: @orndorffgrant  Going through the extra steps here since the default config has all values set to `None`, and these values are shown in the status-cache as None as well instead of the default values. The default values are returned when `getattr()` is called on the `UAConfig` class. This sets the correct default values for keys like `apt_news_url`, `metering_timer`.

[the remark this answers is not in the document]
orndorffgrant: Thanks sorry I missed that! This is good as is

orndorffgrant:
Just adding detail to be clear: This PR introduces https://github.com/canonical/ubuntu-pro-client/issues/2982 which we must fix before release 32. (CC @lucasmoura @renanrodrigo @dheyay)

CI fails are unrelated - merging.

PR #2946, opened 2024-02-15
tests: remove livepatch test for WSL

lucasmoura:
## Why is this needed?
There is no livepatch support for WSL kernels. Because of that, we are removing the WSL machine type from a specific livepatch test

## Test Steps

 - [ ] Yes
 - [x] No


renanrodrigo (approved):

renanrodrigo:
not merging yet - should this go to `main` as we probably will run the test suite for WSL as part of the release validation?

lucasmoura:
Indeed, let me change the base

lucasmoura:
@renanrodrigo the problem is that the WSL test infrastructure is not on main :(

renanrodrigo:
LOL sorry this is good

PR #2947, opened 2024-02-15
tls-in-tls: show auth error for newer pycurl messages

orndorffgrant:
## Why is this needed?

Newer versions of curl/pycurl have a slightly different message for 407 http status codes. The new check only looks for "407" anywhere in the message which works for old and new versions of the message, and is hopefully a bit more future proof.

## Test Steps
`tox -e behave -- features/proxy_config.feature:1302` should pass again

## Checklist

 - [ ] Yes
 - [x] No


dheyay:
Looks good

renanrodrigo (approved):

orndorffgrant:
CI fails unrelated

PR #2949, opened 2024-02-19
Remove lock from the config module

lucasmoura:
## Why is this needed?
We are active working on making the config module more focused on only handling the actual Pro client configuration. Right now, it still handles some functionality that is not directly tied to the configuration.

This is an initial work that removed some unused data paths from the UAConfig class and also completely remove the lock handling from the config module 

## Test Steps
Guarantee that no test breaks after this change and that our lock mechanism is still working.
For example:

1. Launch a LXD container
2. Run `pro attach`
3. On another window, run `pro refresh`
4. Check that `pro refresh` will only start after `pro attach` has finished

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/ubuntu-advantage-tools.postinst
+++ b/debian/ubuntu-advantage-tools.postinst
@@ -114,15 +114,16 @@ redact_ubuntu_release_from_ua_apt_filenames() {
 check_service_is_enabled() {
     service_name=$1
     _RET=$(/usr/bin/python3 -c "
-import os
-import json
-from uaclient.config import UAConfig
-cfg = UAConfig()
-status = cfg.read_cache('status-cache')
-if status:
-    for service in status.get('services', []):
-       if service.get('name', '') == '${service_name}':
-           print(service.get('status', ''))
+from uaclient.files.state_files import status_cache_file
+try:
+    status = status_cache_file.read()
+    if status:
+        for service in status.get('services', []):
+            if service.get('name', '') == '${service_name}':
+                print(service.get('status', ''))
+except Exception:
+    # Assume not enabled
+    pass
 ")
    if [ "${_RET}" = "enabled" ]; then
        return  0
--- a/features/steps/files.py
+++ b/features/steps/files.py
@@ -4,7 +4,7 @@ import tempfile
 
 import yaml
 from behave import then, when
-from hamcrest import assert_that, matches_regexp
+from hamcrest import assert_that, equal_to, matches_regexp
 
 from features.steps.shell import when_i_run_command
 from features.util import SUT, process_template_vars
@@ -160,3 +160,27 @@ def when_i_move_file_from_one_machine_to_another(
     when_i_create_file_with_content(
         context, dest_path, machine_name=dest_machine, text=content
     )
+
+
+@then(
+    "I verify that `{file_path}` is owned by `{owner}` with permission `{permission}`"  # noqa
+)
+def then_i_verify_that_file_is_world_readable(
+    context, file_path, owner, permission, machine_name=SUT
+):
+    when_i_run_command(
+        context,
+        "stat -c '%U:%G' {}".format(file_path),
+        "as non-root",
+        machine_name=machine_name,
+    )
+    assert_that(context.process.stdout.strip(), equal_to(owner))
+
+    when_i_run_command(
+        context,
+        "stat -c '%a' {}".format(file_path),
+        "as non-root",
+        machine_name=machine_name,
+    )
+
+    assert_that(context.process.stdout.strip(), equal_to(permission))
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -78,8 +78,6 @@ event = event_logger.get_event_logger()
 
 class UAConfig:
     data_paths = {
-        "instance-id": DataPath("instance-id", True),
-        "machine-access-cis": DataPath("machine-access-cis.json", True),
         "lock": DataPath("lock", False),
         "status-cache": DataPath("status.json", False),
     }  # type: Dict[str, DataPath]
--- a/uaclient/config.py
+++ b/uaclient/config.py
@@ -77,9 +77,7 @@ event = event_logger.get_event_logger()
 
 
 class UAConfig:
-    data_paths = {
-        "status-cache": DataPath("status.json", False),
-    }  # type: Dict[str, DataPath]
+    data_paths = {}  # type: Dict[str, DataPath]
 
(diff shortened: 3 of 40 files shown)


debian/ubuntu-advantage-tools.postinst
```diff
@@ -114,15 +114,15 @@ redact_ubuntu_release_from_ua_apt_filenames() {
...
-    for service in status.get('services', []):
-       if service.get('name', '') == '${service_name}':
-           print(service.get('status', ''))
+from uaclient.files.state_files import status_cache_file
+try:
+    status = status_cache_file.read()
+    if status:
+        for service in status.get('services', []):
+        if service.get('name', '') == '${service_name}':
+            print(service.get('status', ''))
```
orndorffgrant, line 123: These lines need to be indented more

uaclient/files/files.py
```diff
@@ -35,7 +35,10 @@ def is_private(self) -> bool:
     def is_present(self):
         return os.path.exists(self.path)
 
-    def write(self, content: str):
+    def write(self, content: Any):
+        self._write(str(content))
```
orndorffgrant, line 39: Using `Any` here removes all the help from mypy.

I'd rather have different classes that use different types than use `Any`. For example, rather than `ProJSONFile` _extending_ `UAFile`, it could _have_ a `UAFile` passed to the constructor.

orndorffgrant (commented):
Excited for this one! Just some small comments

debian/ubuntu-advantage-tools.postinst
```diff
@@ -114,15 +114,15 @@ redact_ubuntu_release_from_ua_apt_filenames() {
...
+except Exception:
+    print('error')
```
renanrodrigo, line 125: I would suggest something a bit more descriptive. "Error checking service <name>"?

features/steps/files.py
```diff
@@ -160,3 +160,17 @@ def when_i_move_file_from_one_machine_to_another(
...
+    context, file_path, machine_name=SUT
+):
+    when_i_run_command(
+        context,
+        "stat -c '%a' {}".format(file_path),
+        "with sudo",
+        machine_name=machine_name,
+    )
+
+    assert_that(context.process.stdout.strip(), equal_to("644"))
```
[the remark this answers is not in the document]
renanrodrigo: (as a matter of fact, depending on user+group combination there are chances files would be `664` - I have local examples :D)

uaclient/config.py
```diff
@@ -77,12 +77,7 @@
 
 
 class UAConfig:
-    data_paths = {
-        "instance-id": DataPath("instance-id", True),
-        "machine-access-cis": DataPath("machine-access-cis.json", True),
-        "lock": DataPath("lock", False),
-        "status-cache": DataPath("status.json", False),
-    }  # type: Dict[str, DataPath]
+    data_paths = {}  # type: Dict[str, DataPath]
```
renanrodrigo, line 80: Do we even need to keep this variable around?

uaclient/files/files.py
renanrodrigo: Or, it could extend and overwrite the method :D

uaclient/config.py
lucasmoura: Yes, we do. There are some "hidden" state files on the system still and for the config module to handle them, we need the `data_paths`. I am getting rid of it on the following PR #2966

features/steps/files.py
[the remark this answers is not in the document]
lucasmoura: Since we always use the same user and group for the test, do you think that this is still applicable ?

lucasmoura:

features/steps/files.py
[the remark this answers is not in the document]
orndorffgrant: Idea: how about a step that is
```
Then I verify that `/path/to/file` is owned by `root:root` with permissions `644`
```
?

debian/ubuntu-advantage-tools.postinst
```diff
@@ -114,15 +114,15 @@ redact_ubuntu_release_from_ua_apt_filenames() {
...
-           print(service.get('status', ''))
+from uaclient.files.state_files import status_cache_file
+try:
+    status = status_cache_file.read()
+    if status:
+        for service in status.get('services', []):
+            if service.get('name', '') == '${service_name}':
+                print(service.get('status', ''))
+except Exception:
+    print('Error checking service: ${service_name}')
```
orndorffgrant, line 125: I guess this won't actually print to stdout, the value will just be used in the following if statement. Maybe it should instead be
```suggestion
    # Assume not enabled
    pass
```

lucasmoura:
@orndorffgrant updated

features/steps/files.py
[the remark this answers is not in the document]
renanrodrigo: yes, then we have clear expectation on that, but then the step name should be `matches public root permission expectation` or something like that?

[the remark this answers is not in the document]
orndorffgrant: The "root:root" and "644" don't have to be actual variables, but I think we should be able to read the step and know exactly what is being checked. If I read "matches public root permission expectation", then I don't know much and have to go read the code to figure out what actually is the root permission expectation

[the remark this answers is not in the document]
renanrodrigo: good point. I'm +1 reading owner + permissions, and making it explicit in the step then

renanrodrigo:
LGTM, but some integration test failures seem legit and must be investigated. Maybe a rebase will sort this out?

uaclient/files/files.py
```diff
@@ -64,6 +64,36 @@ def delete(self):
...
 
 
+class ProJSONFile:
+    def __init__(
+        self,
+        name: str,
+        directory: str = defaults.DEFAULT_DATA_DIR,
+        private: bool = True,
+    ):
+        self._file = UAFile(name=name, directory=directory, private=private)
```
orndorffgrant, line 74: You could accept a `UAFile` as an argument instead of drilling arguments
```suggestion
    def __init__(
        self,
        ua-file: UAFile,
    ):
        self._file = ua_file
```

orndorffgrant (approved):
Just some nits and a rebase needed - preemptively approving - thanks!

lucasmoura:

renanrodrigo:
There are still some tests consistently failing on GH but now they are definitely not related to this pull request.
I am merging this for the purpose of moving on - a separate PR will fix the errors.

PR #2950, opened 2024-02-20
Reformat behave tests

dheyay:
## Why is this needed?

Creates a new tox environment which checks for formatting of gherkin files (behave tests), runs with pre-commit.

Reference for gherkin reformatting library - https://pypi.org/project/reformat-gherkin/

## Checklist

 - [ ] Yes
 - [ ] No


.pre-commit-config.yaml
```diff
@@ -1,12 +1,16 @@
 repos:
   - repo: https://github.com/ambv/black
-    rev: 22.3.0  # Also stored in dev-requirements.txt; update both together!
+    rev: 24.2.0  # Also stored in dev-requirements.txt; update both together!
```
renanrodrigo, line 3: The comment says you must update dev-requirements together with this :D

renanrodrigo: But besides that, is there a reason to bump black and isort here?

dev-requirements.txt
```diff
@@ -4,3 +4,4 @@ black==22.3.0
 isort==5.12.0
 pre-commit
 shellcheck-py==0.9.0.6
+reformat-gherkin
```
renanrodrigo, line 7: Should we pin the dependency to a version?

.pre-commit-config.yaml
renanrodrigo: But the question is : is there a reason to bump those deps?

renanrodrigo (approved):
Overall LGTM, standardizing this is great, thanks much
I have a couple questions on the dependencies though

dev-requirements.txt
dheyay: Yes, my bad. I think we should pin it to 3.0.1. Thanks for pointing that out.

orndorffgrant (approved):
LGTM thank you!

PR #2956, opened 2024-02-21
tests: Added lxd vm test for attach validtoken

dheyay:
## Why is this needed?

Adds an integration that attaches on a lxd-vm and asserts that esm-infra, esm-apps, and livepatch are all enabled by default.

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/features/attach_validtoken.feature
+++ b/features/attach_validtoken.feature
@@ -177,6 +177,34 @@ Feature: Command behaviour when attaching a machine to an Ubuntu Pro
       | bionic  | lxd-container | cis        |
       | focal   | lxd-container | usg        |
 
+  Scenario Outline: Auto enable by default and attached disable of livepatch in a lxd vm
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I attach `contract_token` with sudo
+    Then I verify that `esm-infra` is enabled
+    And I verify that `esm-apps` is enabled
+    And I verify that `livepatch` status is `<livepatch_status>`
+    When I run `pro disable livepatch` with sudo
+    Then I verify that running `canonical-livepatch status` `with sudo` exits `1`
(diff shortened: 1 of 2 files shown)


lucasmoura (commented):
That's great @dheyay, but we already have a similar test to this one:
https://github.com/canonical/ubuntu-pro-client/blob/main/features/attached_enable.feature#L517

So I think it might be better improving this one instead

orndorffgrant:
@lucasmoura @dheyay I missed that test when I recommended creating this new one, but I think some changes and a new test are still necessary.

Attaching and auto enabling esm-infra, esm-apps, and livepatch is the most widely used feature of pro-client, so I think it deserves its own test. The "Attached disable of livepatch" in attached_enable.feature only tests the auto-enablement as an aside while actually testing disabling livepatch.

Disabling livepatch does deserve a test, but shouldn't care too much about other services, so I think two things are warranted:
1. Create a new test in `attach_validtoken.feature` like this PR is currently doing. All it should do is attach and check that the expected services are auto-enabled. This should run on all LTS releases in a lxd-vm. This test doesn't need to do any `pro refresh` or extra `pro attach`.
2. Edit "Attached disable of livepatch" test to use --no-auto-enable when attaching. Then it doesn't need to assert that the esm services are enabled.

lucasmoura:
@orndorffgrant I see the point here, but due to the cost of running VM tests, do you think it is worth to have a test that only checks the behavior of disabling livepatch ?

I still think the tests can be combined. Even though we are testing distinct features, they don't clash with each other and this would avoid spinning another VM for an extra test.

But let me know if you disagree

lucasmoura (approved):

orndorffgrant (approved):
Thank you! This just needs conflicts resolved since we merged the gherkin reformat

PR #2960, opened 2024-02-22
deps: pick the proper python-apt based on the Python version

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it enables us to run our unit test suite again.
`python-apt` had an update which breaks installing from source on older releases (and even Jammy is too old for that) so now we are looking at the python version, and cloning it from a git-ubuntu branch which "matches" what you have.

This implies that developers + GH CI are always running one of the supported development environments (latest LTS, latest released, or devel) and that they did NOT change their Python version on that system.

The damage is very low if you decide to change whatever though, so this is very acceptable.

## Test Steps
- `make clean`, then run unit tests on a Jammy machine (py 3.10). All should work. Then check the venv, and see python3-apt is installed from `f043e6` (jammy-updates).
- `make clean`, then run unit tests on a Mantic/Noble machine (py 3.11). All should work. Then check the venv, and see python3-apt is installed from `f058cb` (mantic).
- When Noble migrates to 3.12 this will pick it from (noble) and all should work fine too.

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo:
a victim of the Noble build bug
Should I target next-32?

orndorffgrant:
Yes I think everything should target `next-v32` until the great merge of `main` and `release` occurs.

renanrodrigo:
closing in favor of #2963

PR #2961, opened 2024-02-23
u.pro.services.dependencies.v1

orndorffgrant:
## Why is this needed?

This is part two of prep for the enable api. This introduces the dependencies endpoint as described in the spec.

## Test Steps
New tests should pass

## Checklist

 - [x] Changes here need to be documented, and this was done in: #2962

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/po/pt_BR.po
+++ b/debian/po/pt_BR.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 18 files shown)


lucasmoura (approved):
Even though we are already testing this on the new integration test, I think we should also add the new endpoint here:
https://github.com/canonical/ubuntu-pro-client/blob/main/features/api.feature#L3

Just for consistency reasons

orndorffgrant:
@lucasmoura great catch! added!

renanrodrigo (approved):
LGTM
I honestly am sad about the `EntitlementWithMessage` name, but I have no better idea to suggest so

PR #2962, opened 2024-02-23
docs: api format update + 3 new endpoints

orndorffgrant:
## Why is this needed?
This documents the new api endpoint being introduced in #2961

This is not done. It should also not be merged until after release 31 docs are released.

The enable and disable endpoint docs were generated using a new script found in #3085


docs/references/api.rst
```diff
@@ -1949,6 +1950,66 @@ are:
...
+        - ``uaclient.api.u.pro.services.dependencies.v1.DependenciesResult``
+
+          .. list-table::
+             :header-rows: 1
+
+             * - Field Name
+               - Type
+               - Description
+             * - ``services``
+               - *List[ServiceWithDependencies]*
```
s-makin, line 1989: Does ServiceWithDependencies need to be treated as a single word? The spellchecker doesn't like it very much.

s-makin (commented):
Only one error, otherwise looks good to me!

orndorffgrant:
I believe the linkcheck only failed because that site is temporarily down

orndorffgrant:
Hi @s-makin I added some more content to this one. Please take another look when you have time

docs/references/api.rst
```diff
@@ -227,11 +230,137 @@ This endpoint shows the installed Pro Client version.
...
+          No data present in the result.
+
+      - Raised exceptions
+
+        - ``AlreadyAttachedError``: Raised if running on a machine which is
+          already attached to a Pro subscription.
+        - ``AutoAttachDisabledError``: Raised if ``disable_auto_attach: true``
+          in ``uaclient.conf``.
+        - ``ConnectivityError``: Raised if it is not possible to connect to the
+          Contracts Server.
```
s-makin, line 336: More of a question than a suggestion, but is it "Contracts" server or "Contract" server? Because we have both shown in this page, and I'm not sure which is the correct one.

docs/references/api.rst
```diff
@@ -2178,36 +2427,62 @@ by origin.
...
+                      }
+                    },
+                    ...
+                  ],
+                  "incompatible_with": [
+                    {
+                      "name": "two",
+                      "reason": {
+                        "code": "one-and-two",
+                        "title": "Services One and Two don't go together."
```
s-makin, line 2455: I realise this is a code rather than docs issue, but saying "don't go together" is quite idiomatic English and may not be well understood by those with English as a 2nd+ language. I think it's ok to say "are not compatible" or something similar.

docs/references/api.rst
```diff
@@ -2346,33 +2571,40 @@ This endpoint shows if the machine is attached to a Pro subscription.
...
+
+            * - Field Name
+              - Type
+              - Description
+            * - ``enabled``
+              - ``List[str]``
+              - List of services that were enabled.
+            * - ``disabled``
+              - ``List[str]``
+              - List of services that were disabled
```
s-makin, line 2592: For consistency of appearance, I'd suggest putting full stops at the end of each of these descriptors (since the last one needs one, they should all have one)

docs/references/api.rst
```diff
@@ -2346,33 +2571,40 @@ This endpoint shows if the machine is attached to a Pro subscription.
...
-               - *int*
-               - The number of days remaining for the contract to be valid
-             * - ``is_attached_and_contract_valid``
-               - *bool*
-               - If the machine is attached and the contract is still valid
+      - Raised exceptions:
+
+        - ``NonRootUserError``: When called as non-root user
+        - ``UnattachedError``: When called on a machine that is not attached to a Pro subscription
+        - ``NotSupported``: When called for a service that doesn't support being enabled via API (currently only landscape)
```
s-makin, line 2604:
```suggestion
        - ``NotSupported``: When called for a service that doesn't support being enabled via API (currently only Landscape)
```

s-makin (commented):
Generally LGTM - found a couple of nits and suggestions. My main one is that we mention the "magic token", but we don't define what that is or how it's different from a regular token anywhere. This might lead to confusion so it could be worth defining it.

orndorffgrant:
Thanks @s-makin! Some of this I only moved around to alphabetize and didn't re-assess the content. I'll make the changes you've suggested!

docs/references/api.rst
```diff
@@ -467,27 +570,50 @@ This endpoint checks if a given system should run auto-attach on boot.
...
                - Description
-             * - ``should_auto_attach``
-               - *bool*
-               - True if the system should run auto-attach on boot
+             * - ``user_code``
+               - ``str``
+               - Code the user will see in the UI when confirming the Magic Attach
+             * - ``token``
+               - ``str``
+               - Magic Token used by the tooling to continue the operation
```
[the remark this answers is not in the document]
orndorffgrant: Not sure - I only moved this :upside_down_face:. I'll read through it and adjust

docs/references/api.rst
orndorffgrant: Ah yep I missed it - thanks!

orndorffgrant:
@s-makin updated!

s-makin (approved):
Thanks for the changes! All lgtm :)

PR #2965, opened 2024-02-27
Redact secrets from logs

dheyay:
## Why is this needed?

Redacts secrets from logs using a secret manager, makes sure the specific secrets are redacted from all logs accurately.

## Checklist

 - [ ] Yes
 - [ ] No


uaclient/cli/__init__.py
```diff
@@ -27,7 +27,7 @@
     lock,
 )
 from uaclient import log as pro_log
-from uaclient import messages, security_status
+from uaclient import messages, secretManager, security_status
```
renanrodrigo, line 30: Naming conventions recommend that module names are defined in a single word (I don't have a better suggestion than this one so I'm fine) and, having multiple words, use snake_case.
`secret_manager`?

dheyay: Agreed, have pushed the changes.

uaclient/contract.py
```diff
@@ -263,6 +264,7 @@ def get_magic_attach_token_info(self, magic_token: str) -> Dict[str, Any]:
         response = self.request_url(
             API_V1_GET_MAGIC_ATTACH_TOKEN_INFO, headers=headers
         )
+        secret_manager.secrets.add_secret(magic_token)
```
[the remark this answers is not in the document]
orndorffgrant: Many secrets actually first enter memory in a response from the contract server. Maybe review all of our endpoints we use and cross-reference with that API spec and add more `add_secret` calls throughout this module

uaclient/daemon/__init__.py
```diff
@@ -48,7 +48,7 @@ def setup_logging(console_level, log_level, log_file, logger=None):
     logger.setLevel(log_level)
 
     logger.handlers = []
-    logger.addFilter(pro_log.RedactionFilter())
+    logger.addFilter(pro_log.SecretRedactionFilter())
```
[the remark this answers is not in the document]
orndorffgrant: And we'll want to add this filter in the other place we setup logging, currently in cli/__init__.py

orndorffgrant (commented):
The idea looks good! I added some suggestions on changes and more places to look for secrets

uaclient/actions.py
```diff
@@ -72,6 +73,8 @@ def attach_with_token(
 
     cfg.machine_token_file.write(new_machine_token)
 
+    secret_manager.secrets.add_secret(token)
```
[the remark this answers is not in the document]
dheyay: Thanks for the review @orndorffgrant , really helpful. Will update with changes!

dheyay:
Rebased and updated  
Let me know if I have missed something here @orndorffgrant @renanrodrigo @lucasmoura

uaclient/files/files.py
```diff
@@ -159,7 +166,12 @@ def read(self) -> Optional[dict]:
             file_handler = self.private_file
         else:
             file_handler = self.public_file
-        return file_handler.read()
+        content = file_handler.read()
```
orndorffgrant, line 169: Please also add `content["machineToken"]`

test
```diff
@@ -0,0 +1,40 @@
+
+abi <abi/3.0>,
```
lucasmoura, line 2: why do we need this file here ?

uaclient/contract.py
```diff
@@ -93,8 +94,10 @@ def add_contract_machine(
                 code=response.code,
                 body=response.body,
             )
-
-        return response.json_dict
+        response_json = response.json_dict
+        if "machineToken" in response_json:
+            secret_manager.secrets.add_secret(response_json["machineToken"])
```
lucasmoura, line 99: Can't we use a `get` call here and just ignore the secret if it is empty on the secret manager method ?

lucasmoura: This also apply to all other similar checks in the code

test
dheyay: I actually did not see why this file was created or crept into the commit, removed now.

uaclient/contract.py
dheyay: Thanks for that, also added a check when adding secrets so the `""` string is ignored if the `get` call returns an empty string in these cases.

test
lucasmoura: @renanrodrigo that is the tox issue you mentioned right ?
Regardless, I don't think we should commit this file. It is okay to remove it from this PR

dheyay: Updated

lucasmoura (commented):
@orndorffgrant @renanrodrigo do you think we could have a integration test for this somehow ?

dheyay:
> @orndorffgrant @renanrodrigo do you think we could have a integration test for this somehow ?

I think it's possible but tricky. We would want to find an incident (or create one) where the secret is leaked into logs and not caught by the `RegexSecretRedaction` Filter. Then we can check if the new redaction filter sucessfully redacts that particular secret. This might not be the best way by any means but let me know what you guys think.

lucasmoura:
Yeah, I don't know if we can easily reproduce that. Maybe it is better to just think of the secret manager a good redundancy for secret redaction

lucasmoura (approved):

orndorffgrant:
> Maybe it is better to just think of the secret manager a good redundancy for secret redaction

Yeah that was what I was thinking of it as. The regex filters are fragile to slight changes in logging and this feels a little more robust, though it has gaps of its own. Together, they increase my confidence in our secret redaction.

uaclient/api/u/pro/attach/magic/wait/v1.py
```diff
@@ -54,6 +54,7 @@ def __init__(
 def wait(
     options: MagicAttachWaitOptions,
 ) -> MagicAttachWaitResult:
+    secret_manager.secrets.add_secret(options.magic_token)
```
orndorffgrant, line 57: please move this to the top of the `_wait` function (we use that one directly in internal calls)

uaclient/apt.py
```diff
@@ -546,9 +554,11 @@ def add_auth_apt_repo(
     """
     try:
         username, password = credentials.split(":")
+        secret_manager.secrets.add_secret(password)
     except ValueError:  # Then we have a bearer token
         username = "bearer"
         password = credentials
+        secret_manager.secrets.add_secret(password)
```
orndorffgrant, line 561: small nit
```suggestion
    try:
        username, password = credentials.split(":")
    except ValueError:  # Then we have a bearer token
        username = "bearer"
        password = credentials
    secret_manager.secrets.add_secret(password)
```

uaclient/cli/__init__.py
```diff
@@ -1009,6 +1010,7 @@ def _magic_attach(args, *, cfg, **kwargs):
 
     event.info(messages.CLI_MAGIC_ATTACH_INIT)
     initiate_resp = _initiate(cfg=cfg)
+    secret_manager.secrets.add_secret(initiate_resp.token)
```
orndorffgrant, line 1013: We know we already added this inside the `_initiate` call right above
```suggestion
```

uaclient/contract.py
```diff
@@ -93,8 +94,11 @@ def add_contract_machine(
                 code=response.code,
                 body=response.body,
             )
-
-        return response.json_dict
+        response_json = response.json_dict
+        secret_manager.secrets.add_secret(
+            response_json.get("machineToken", "")
+        )
+        return response_json
```
orndorffgrant, line 101: We should also redact the resourceTokens (same as in machine_token_file.read)
```suggestion
        for token in response_json.get("resourceTokens", []):
            secret_manager.secrets.add_secret(token.get("token", ""))
        return response_json
```

orndorffgrant (commented):
Looking really good! Let's get this in before the next release.

I think we can squash the two commits together.

Also, `get_contract_token_for_cloud_instance` and ` get_resource_machine_access` also return secrets IIRC

dheyay:
Updated @orndorffgrant

orndorffgrant (approved):
Thanks!

PR #2966, opened 2024-02-28
Remove config cache

lucasmoura:
## Why is this needed?
The is no need for the config module to hold a cache anymore, as this cache is not tied to any particular user config.
We are creating dedicated files for each state that the config module handles and placing them on more appropriate modules.

PS: This PR is based on #2949. We need to merge it first

## Test Steps
Guarantee that the cache files are still being created. This can be verified by running the `attach` command and verifying that all necessary files are still created

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -17,7 +17,7 @@ from uaclient.api.u.pro.status.enabled_services.v1 import _enabled_services
 from uaclient.api.u.pro.status.is_attached.v1 import _is_attached
 from uaclient.config import UAConfig
 from uaclient.defaults import ATTACH_FAIL_DATE_FORMAT
-from uaclient.files.state_files import attachment_data_file
+from uaclient.files.state_files import attachment_data_file, machine_id_file
 from uaclient.http import serviceclient
 from uaclient.log import get_user_or_root_log_file_path
 
@@ -658,7 +658,7 @@ def refresh(cfg):
     machine_id = resp.get("machineTokenInfo", {}).get(
         "machineId", system.get_machine_id(cfg)
     )
-    cfg.write_cache("machine-id", machine_id)
+    machine_id_file.write(machine_id)
 
     process_entitlements_delta(
         cfg,
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -165,7 +165,6 @@ class UAContractClient(serviceclient.UAServiceClient):
         machine_token: str,
         resource: str,
         machine_id: Optional[str] = None,
-        save_file: bool = True,
     ) -> Dict[str, Any]:
         """Requests machine access context for a given resource
 
@@ -174,7 +173,6 @@ class UAContractClient(serviceclient.UAServiceClient):
         @param resource: Entitlement name.
         @param machine_id: Optional unique system machine id. When absent,
             contents of /etc/machine-id will be used.
-        @save_file: If the machine access should be saved on the user machine
 
         @return: Dict of the JSON response containing entitlement accessInfo.
         """
@@ -194,10 +192,6 @@ class UAContractClient(serviceclient.UAServiceClient):
             )
         if response.headers.get("expires"):
             response.json_dict["expires"] = response.headers["expires"]
-        if save_file:
-            self.cfg.write_cache(
-                "machine-access-{}".format(resource), response.json_dict
-            )
         return response.json_dict
 
     def update_activity_token(self):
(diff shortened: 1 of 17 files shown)


uaclient/files/files.py
```diff
@@ -64,6 +64,36 @@ def delete(self):
...
+
+        return None
+
+    def delete(self):
+        return self._file.delete()
+
+    @property
+    def is_present(self):
+        return self._file.is_present
+
```
dheyay, line 95: Was there a specific reason for introducing this layer of abstraction other than the json serialization/de serialization ? It would be helpful to understand the motivation behind this change (which i think is a great addition).

Also, do we want to handle any errors when doing when reading and writing through json?

uaclient/files/state_files.py
```diff
@@ -295,3 +295,12 @@ def __init__(self, attached_at: datetime.datetime):
...
 )
+
+
+def delete_state_files():
+    machine_id_file.delete()
+    status_cache_file.delete()
+    attachment_data_file.delete()
+    anbox_cloud_credentials_file.delete()
+    reboot_cmd_marker_file.delete()
+    services_once_enabled_file.delete()
```
dheyay, line 306: Should the `user_config_file` also be deleted with the state files here?

dheyay (approved):
Thanks for the changes @lucasmoura ,  looks great!
I just had a few questions about implementation.

uaclient/system.py
```diff
@@ -282,15 +283,19 @@ def get_machine_id(cfg) -> str:
...
 
-    for path in [ETC_MACHINE_ID, DBUS_MACHINE_ID, fallback_machine_id_file]:
+    for path in [ETC_MACHINE_ID, DBUS_MACHINE_ID]:
         if os.path.exists(path):
             content = load_file(path).rstrip("\n")
             if content:
                 return content
+
+        if fallback_machine_id:
+            return fallback_machine_id
```
orndorffgrant, line 295: I think this needs to be outside the loop to keep consistent behavior
```suggestion
    if fallback_machine_id:
        return fallback_machine_id
```

uaclient/contract.py
```diff
@@ -208,10 +206,6 @@ def get_resource_machine_access(
             )
         if response.headers.get("expires"):
             response.json_dict["expires"] = response.headers["expires"]
-        if save_file:
-            self.cfg.write_cache(
-                "machine-access-{}".format(resource), response.json_dict
-            )
```
orndorffgrant, line 214: Just making sure - anbox doesn't rely on this right? We write the credentials in another way for them to use?

uaclient/files/state_files.py
orndorffgrant, line 306: I don't think we want to delete services_once_enabled

orndorffgrant: No because it holds config value that are relevant while attached and this function is called on detach.

Maybe this function should be renamed to "delete_state_files_on_detach" - or does it need to be a function at all? Should these delete calls just go into the detach function (next to `machine_token_file.delete()`)?

uaclient/lock.py
```diff
@@ -1,24 +1,82 @@
...
+
+    :return: A tuple (pid, string describing lock holder)
+        If no active lock, pid will be -1.
+    """
+
+    try:
+        lock_data_obj = lock_data_file.read()
+    except exceptions.InvalidFileFormatError:
+        raise exceptions.InvalidLockFile(
+            lock_file_path=os.path.join(lock_data_file.path)
```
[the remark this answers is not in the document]
lucasmoura: Good point, I think it can be just the `path` value

uaclient/files/state_files.py
lucasmoura: To be honest, I did that to make mocking in the unittest simpler. or we would need to individually mock all of the files for the happy path of the detach. But I can rename the function if you prefer

uaclient/contract.py
lucasmoura: No, it doesn't. We are storing the tokens for them in a separate file

uaclient/files/state_files.py
lucasmoura: Good point, let me update that

lucasmoura:
@renanrodrigo @dheyay @orndorffgrant I have updated the code

dheyay (approved):

renanrodrigo (approved):
Thanks! Another good step forward with the config

renanrodrigo:
CI errors completely unrelated

PR #2971, opened 2024-02-28
enable cli prompt refactor

orndorffgrant:
## Why is this needed?

This moves all CLI concerns, such as printing messages and prompting for
input, up into the cli module. It also replaces the global event_logger
with a call-specific progress object.

As part of this I added an extra progress message which seemed useful.

## Test Steps

Check small modifications to behave tests. Otherwise everything should still pass

## Checklist

 - [ ] Yes
 - [x] No


uaclient/cli/enable.py
```diff
@@ -11,25 +13,138 @@
...
     CanEnableFailureReason,
 )
 
-event = event_logger.get_event_logger()
 LOG = logging.getLogger(util.replace_top_level_logger_name(__name__))
 
 
+class CLIEnableProgress(api.AbstractProgress):
+    def __init__(self, messaging):
+        self.messaging = messaging
```
lucasmoura, line 29: Any reason why we must create the object if the messages from the entitlement ? I am wondering if it is simpler to always pass the event and payload to `emit`

uaclient/entitlements/repo.py
```diff
@@ -133,26 +134,34 @@ def can_disable(
 
         return result, reason
 
-    def _perform_enable(self, silent: bool = False) -> bool:
+    def enable_steps(self) -> int:
+        will_install = self.packages is not None and len(self.packages) > 0
+        if self.access_only or not will_install:
```
lucasmoura, line 139: Let's add a comment stating which operation we think should happen, just to make it clear why we have those numbers here

uaclient/cli/enable.py
```diff
@@ -38,31 +153,70 @@ def action_enable(args, *, cfg, **kwargs):
...
+            access_only=access_only,
+            extra_args=kwargs.get("extra_args"),
+        )
+        real_name = ent.name
+        if json_output:
+            progress = api.ProgressWrapper()
+        else:
+            progress = api.ProgressWrapper(CLIEnableProgress(ent.messaging))
+
+        progress.total_steps = ent.calculate_total_enable_steps()
```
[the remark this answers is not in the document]
orndorffgrant: We don't, but will need it in the API which will be in the next PR and replace this code. I can remove this if you prefer.

uaclient/cli/enable.py
orndorffgrant: Yep that's a good idea, I'll refactor to do that

uaclient/cli/enable.py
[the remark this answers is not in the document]
lucasmoura: No, it is fine. We can keep it

orndorffgrant:
@lucasmoura @renanrodrigo this is ready for review now. Keep in mind that I will have another PR soon that touches this code again to make it use the new API for enable - that will simplify some things (and complicate others, of course :upside_down_face:)

test
```diff
@@ -0,0 +1,40 @@
+
+abi <abi/3.0>,
```
lucasmoura, line 2: Tox is incorrectly creating this file. We need to address this in a future PR, but let's just remove that file now

uaclient/cli/enable.py
```diff
@@ -11,25 +13,137 @@
...
+        if event == "info":
+            print(payload)
+            return
+        elif event == "message_operation":
+            if not util.handle_message_operations(payload, print):
+                raise exceptions.PromptDeniedError()
+
+
+def prompt_for_dependency_handling(
+    cfg,
```
lucasmoura, line 49: Let's just add type hints here

uaclient/cli/enable.py
```diff
@@ -38,31 +152,76 @@ def action_enable(args, *, cfg, **kwargs):
...
+        s.name for s in _enabled_services(cfg).enabled_services
+    ]
+    all_dependencies = _dependencies(cfg).services
+
     ret = True
-    for ent_name in entitlements_found:
+    for ent_name in entitlements.order_entitlements_for_enabling(
+        cfg, entitlements_found
+    ):
+        real_name = entitlements.get_real_name(cfg, ent_name)
```
lucasmoura, line 183: I think we can create the entitlement object early through the `entitlement_factory` function, which already respects the `valid_names` logic. Then we would need to create it here passing the `called_name` argument to entitlement class. If we do that, then we can fetch the real entitlement name through the `presentation_name` property.

This would allow us to get rid of the `get_real_name` function

uaclient/entitlements/base.py
```diff
@@ -413,7 +447,7 @@ def can_enable(self) -> Tuple[bool, Optional[CanEnableFailure]]:
     # support additional reason types in the future.
     def enable(
         self,
-        silent: bool = False,
+        progress: api.ProgressWrapper,
```
lucasmoura, line 450: Is there a reason why we don't receive the `ProgressWrapper` object when creating the class instead of drilling down to the individual methods ?

uaclient/entitlements/fips.py
```diff
@@ -207,30 +207,31 @@ def prompt_if_kernel_downgrade(
...
-            )
-
         # We need to guarantee that the metapackage is installed.
         # While the other packages should still be installed, if they
         # fail, we should not block the enable operation.
         mandatory_packages = self.packages
-        super().install_packages(
-            package_list=mandatory_packages, verbose=False
-        )
+        if mandatory_packages:
```
lucasmoura, line 224: Is there a situation where this will be empty ?

uaclient/entitlements/fips.py
```diff
@@ -257,11 +258,20 @@ def install_packages(
...
                         service=self.title, pkg=pkg
-                    )
+                    ),
                 )
+        if self._check_for_reboot():
+            notices.add(
+                Notice.FIPS_SYSTEM_REBOOT_REQUIRED,
+            )
+
+    def _check_for_reboot(self) -> bool:
```
lucasmoura, line 272: Why are we overwriting this method here ? Since it is exactly the same as the one defined on the base class, shouldn't we instead adapt this method on the `repo` class instead ?

test
orndorffgrant: Arg sorry I missed this

uaclient/entitlements/fips.py
orndorffgrant: On containers, this will be empty.

uaclient/entitlements/fips.py
orndorffgrant: I changed the implementation in the base class :grimacing: let me know what you think.

lucasmoura: I missed that on the review :facepalm: But why are we returning False by default on the base class now ? It is just because the logic is redundant considering the implementation we have on the repo class ?

uaclient/cli/enable.py
orndorffgrant: Coming back to rewrite `get_real_name` and its actually so simple that I've come around to thinking its not necessary.

I'm still going to keep it separate from the later entitlement instantiation, because that will get removed in the next PR anyway.

orndorffgrant: Updated!

test
orndorffgrant: Done!

uaclient/entitlements/base.py
lucasmoura: I have discussed that together and we agreed that the current solution is simpler

uaclient/entitlements/fips.py
lucasmoura: We have agreed that the necessary implementation here should be on the repo class (which already correctly implements the logic)

lucasmoura:

lucasmoura (approved):

uaclient/entitlements/base.py
```diff
@@ -444,39 +480,33 @@ def enable(
...
-            if not self.handle_required_packages():
+            if not self.handle_required_packages(progress):
                 return False, None
 
-        ret = self._perform_enable(silent=silent)
+        ret = self._perform_enable(progress)
         if not ret:
             return False, None
 
-        msg_ops = self.messaging.get("post_enable", [])
```
renanrodrigo, line 473: we are not emit-ing post enable messages here, similar to pre enable?

uaclient/entitlements/base.py
```diff
@@ -893,7 +905,7 @@ def disable(
                 include other types of reasons in the future.
         """
         msg_ops = self.messaging.get("pre_disable", [])
-        if not util.handle_message_operations(msg_ops):
+        if not util.handle_message_operations(msg_ops, event.info):
```
renanrodrigo, line 908: If I got this correctly, event.info will go away here when `disable` implements our progress beauties

tools/spellcheck-allowed-words.txt
```diff
@@ -23,6 +23,7 @@ ARGS
 aws
 bootable
 Cancelling
+cancelled
```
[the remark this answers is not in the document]
orndorffgrant: technically both okay, `canceled` is more US-ish, `cancelled` is more UK-ish and we're supposed to be UK-ish

uaclient/entitlements/base.py
orndorffgrant: Yes

uaclient/entitlements/base.py
orndorffgrant: When we specced the api for enable, we decided that the api would return the post_enable messages. So I'm moving the handling of the out of the entitlement and into the cli/enable module

uaclient/cli/enable.py
```diff
@@ -11,25 +13,140 @@
...
+    warnings = []
+
+    json_response = {
+        "_schema_version": event_logger.JSON_SCHEMA_VERSION,
+        "result": "success",
+        "needs_reboot": False,
+    }
+
+    json_output = args.format == "json"
+    print_fn = _create_print_function(json_output)
```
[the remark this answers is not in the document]
orndorffgrant: Another thing I seriously considered is splitting enable into two functions: 1 for interactive and 1 for json - it makes it much more readable and removes this hack, but there is a lot of duplicated logic. Thoughts?

Even if we do that I think I'd like to add that on top of the next PR, so I will add a big comment here for now anyway.

uaclient/cli/enable.py
```diff
@@ -11,25 +13,140 @@
...
+
+def _null_print(*args, **kwargs):
+    pass
+
+
+def _create_print_function(json_output: bool):
+    if json_output:
+        return _null_print
+    else:
+        return print
```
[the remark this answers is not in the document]
orndorffgrant: :grimacing:

uaclient/messages/__init__.py
```diff
@@ -2594,3 +2599,8 @@ def __repr__(self):
         "These directives need to be unique for every entitlement."
     ),
 )
+
+E_LANDSCAPE_CONFIG_FAILED = NamedMessage(
+    "landscape-config-failed",
+    t.gettext("""landscape-config command failed"""),
```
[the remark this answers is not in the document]
orndorffgrant: yep, probably just copypasted from somewhere else and didn't change it - will fix

tools/spellcheck-allowed-words.txt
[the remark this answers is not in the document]
renanrodrigo: TIL

uaclient/cli/enable.py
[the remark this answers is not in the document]
renanrodrigo: Let's see how it will look like after the whole thing is implemented.
Change is tough, let's do it slow and steady.

uaclient/entitlements/base.py
renanrodrigo: ack lgtm

renanrodrigo (approved):
rock it

orndorffgrant:
I'll let CI run one more time since I rebased as well

orndorffgrant:
CI fail is #3013

PR #2972, opened 2024-03-01
Fix some integration tests

lucasmoura:
## Why is this needed?
Fix some broken integration tests

## Test Steps
Verify that possible CI errors now are due to flaky tests

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/cloud_pro_clone.feature
+++ b/features/cloud_pro_clone.feature
@@ -18,8 +18,8 @@ Feature: Creating golden images based on Cloud Ubuntu Pro instances
     And I run `pro status --format yaml` with sudo
     Then stdout matches regexp:
       """
-      name: fips-updates
-      status: enabled
+      \s*name: fips-updates
+      \s*status: enabled
       """
     When I reboot the machine
     When I take a snapshot of the machine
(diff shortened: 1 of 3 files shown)


features/cloud_pro_clone.feature
```diff
@@ -18,8 +18,8 @@ Feature: Creating golden images based on Cloud Ubuntu Pro instances
     And I run `pro status --format yaml` with sudo
     Then stdout matches regexp:
       """
-      name: fips-updates
-      status: enabled
+      \s*name: fips-updates
+      \s*status: enabled
```
dheyay, line 22: Hi @lucasmoura , I noticed the use of `\s*` in this regexp matching. Could you help me understand why it's used here and not in other similar situations when matching stdout with regexp?

dheyay (approved):
Looks good! Thank you.

features/cloud_pro_clone.feature
lucasmoura: This is just because we are testing the yaml output and this output is formatted in a way that the extra spaces are required here.

The problem is that the gherkin formatter is removing those spaces from the file and that's why I have used the regex.

dheyay: Ah yes, the gherkin formatter is very picky about extra lines or spaces in output comparisons. Thanks!

orndorffgrant:
`features/logs.feature:55` seems to be consistently failing as well - can you take a look at that one too @lucasmoura?

PR #2976, opened 2024-03-05
Add architectures and packages selectors

dheyay:
## Why is this needed?

We needed APT News selectors to only display messages on systems with certain package versions or certain architectures. This adds selectors for both architectures and packages to filter APT news.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/apt_messages.feature
+++ b/features/apt_messages.feature
@@ -675,6 +675,352 @@ Feature: APT Messages
       | focal   | gcp.generic   | Learn more about Ubuntu Pro on GCP at https://ubuntu.com/gcp/pro                 |
 
   @uses.config.contract_token
+  Scenario Outline: APT news selectors
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I attach `contract_token` with sudo
+    When I apt upgrade including phased updates
+    When I apt autoremove
+    When I apt install `jq`
+    When I run `pro detach --assume-yes` with sudo
+    Given a `mantic` `<machine_type>` machine named `apt-news-server`
+    When I apt install `nginx` on the `apt-news-server` machine
+    When I run `sed -i "s/gzip on;/gzip on;\n\tgzip_min_length 1;\n\tgzip_types application\/json;\n/" /etc/nginx/nginx.conf` `with sudo` on the `apt-news-server` machine
+    When I run `systemctl restart nginx` `with sudo` on the `apt-news-server` machine
+    When I run `pro config set apt_news_url=http://$behave_var{machine-ip apt-news-server}/aptnews.json` with sudo
+    # Testing codename selector
+    When I create the file `/var/www/html/aptnews.json` on the `apt-news-server` machine with the following:
+      """
+      {
+        "messages": [
+          {
+            "begin": "$behave_var{today}",
+            "selectors": {
+              "codenames": ["<wrong_release>"],
+            },
+            "lines": [
+              "one"
+            ]
+          }
+        ]
+      }
+      """
+    When I run `pro refresh messages` with sudo
+    When I apt upgrade
+    Then I will see the following on stdout
+      """
+      Reading package lists...
+      Building dependency tree...
+      Reading state information...
+      Calculating upgrade...
+      0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
+      """
+    When I create the file `/var/www/html/aptnews.json` on the `apt-news-server` machine with the following:
+      """
+      {
+        "messages": [
+          {
+            "begin": "$behave_var{today}",
+            "selectors": {
+              "codenames": ["<release>"]
+            },
+            "lines": [
+              "one",
+              "two"
+            ]
+          }
+        ]
+      }
+      """
+    When I run `pro refresh messages` with sudo
+    When I apt upgrade
+    Then I will see the following on stdout
+      """
+      Reading package lists...
+      Building dependency tree...
+      Reading state information...
+      Calculating upgrade...
+      #
+      # one
+      # two
+      #
+      0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
+      """
+    # Testing architectures selector
+    When I create the file `/var/www/html/aptnews.json` on the `apt-news-server` machine with the following:
+      """
+      {
+        "messages": [
+          {
+            "begin": "$behave_var{today}",
+            "selectors": {
+              "architectures": ["amd64"]
+            },
+            "lines": [
+              "one",
+              "two"
+            ]
+          }
+        ]
+      }
+      """
+    When I run `pro refresh messages` with sudo
+    When I apt upgrade
+    Then I will see the following on stdout:
+      """
+      Reading package lists...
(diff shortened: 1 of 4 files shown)


uaclient/apt_news.py
```diff
@@ -88,6 +100,49 @@ def do_selectors_apply(
...
+
+    if selectors.packages is not None:
+        installed_packages = get_installed_packages()
+        installed_packages_names = [
+            package.name for package in installed_packages
+        ]
+        package_matched = False
+
+        for package in selectors.packages:
+            package_name, version_operator, package_version = package
```
orndorffgrant, line 115: We should check that there are three items in this list before doing this. We can log something and return False if there are not three items.

uaclient/apt_news.py
```diff
@@ -88,6 +100,49 @@ def do_selectors_apply(
...
+                    )
+                    or (
+                        version_operator == ">="
+                        and (
+                            version_comparison == 0 or version_comparison == 1
+                        )
+                    )
+                    or (version_operator == "<" and version_comparison == -1)
+                    or (version_operator == ">" and version_comparison == 1)
+                ):
```
orndorffgrant, line 139: Style suggestion
```suggestion
                if any(
                    [
                        (version_comparison == 0 and version_operator in ["==", "<=", ">="]),
                        (version_comparison < 0 and version_operator in ["<", "<="]),
                        (version_comparison > 0 and version_operator in [">", ">="]),
                    ]
                ):
```

uaclient/apt_news.py
```diff
@@ -88,6 +100,49 @@ def do_selectors_apply(
...
+    if selectors.packages is not None:
+        installed_packages = get_installed_packages()
+        installed_packages_names = [
+            package.name for package in installed_packages
+        ]
+        package_matched = False
+
+        for package in selectors.packages:
+            package_name, version_operator, package_version = package
+            if package_name in installed_packages_names:
```
orndorffgrant, line 116: Optional suggestion: invert this if statement to return early
```suggestion
            if not apt.is_installed(package_name):
                LOG.info("%s not installed", package_name)
                return False
```

orndorffgrant (commented):
Looking good!

Please add cases to the test in apt_messages.feature to cover this new functionality.

uaclient/apt_news.py
dheyay: Looks much better than the original change haha, thanks @orndorffgrant

renanrodrigo:
This is not building because there is an extra comma in `apt_news.py:54`

features/apt_messages.feature
```diff
@@ -669,6 +669,348 @@ Feature: APT Messages
...
+      Reading state information...
+      Calculating upgrade...
+      The following packages have been kept back:
+        libcurl4
+      0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
+      """
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | mantic  | lxd-container |
```
lucasmoura, line 1012: I think we should test this on all LTS releases as well

uaclient/apt_news.py
```diff
@@ -90,6 +103,54 @@ def do_selectors_apply(
...
+                package_name, version_operator, package_version = package
+            except ValueError:
+                LOG.warning("Invalid package selector: %r", package)
+                return False
+            if not is_installed(package_name):
+                LOG.info("Package %s is not installed", package_name)
+                return False
+            installed_package = installed_packages[
+                installed_packages_names.index(package_name)
+            ]
```
lucasmoura, line 128: instead of looking at the packages list, we can instead use the `get_pkg_version` function directly

uaclient/apt_news.py
```diff
@@ -90,6 +103,54 @@ def do_selectors_apply(
...
+                        and version_operator in ["<", "<="]
+                    ),
+                    (
+                        version_comparison > 0
+                        and version_operator in [">", ">="]
+                    ),
+                ]
+            ):
+                is_package_matched = True
+                break
```
lucasmoura, line 149: I think we can update the logic to a `not any` and just return `False` here.
Then we don't need the `is_package_matched` variable anymore

uaclient/apt_news.py
lucasmoura: If we do that, we don't even need the `installed_packages` anymore

dheyay: Thanks @lucasmoura , seems like a good idea.
Updated!

uaclient/apt_news.py
```diff
@@ -90,6 +103,41 @@ def do_selectors_apply(
...
+            try:
+                package_name, version_operator, package_version = package
+            except ValueError:
+                LOG.warning("Invalid package selector: %r", package)
+                return False
+            if not is_installed(package_name):
+                LOG.info("Package %s is not installed", package_name)
+                return False
+            version_comparison = version_compare(
+                get_pkg_version(package_name), package_version
```
lucasmoura, line 121: I think we need to think of the situation where this function returns `None` (when the package doesn't exist yet)

dheyay: Would that not be flagged be the installed package check and return False earlier?

features/apt_messages.feature
dheyay: Updated the test to support all LTS releases @lucasmoura

uaclient/apt_news.py
lucasmoura: Yes, you are totally right. But now, looking at the code, I think it might be better if we don't have that call to `is_installed` at all.

My rationale here is that this function will fetch all installed packages and then perform a search on that list.
For the `get_pkg_version` we directly check the package in the apt cache, which should be faster.

We would then return false only if `get_pkg_version` returns `None`.

What do you think ?

dheyay: Oh yes, actually, that makes a lot of sense. That would be a good change, thanks @lucasmoura

features/apt_messages.feature
```diff
@@ -669,6 +669,352 @@ Feature: APT Messages
...
+            ]
+          }
+        ]
+      }
+      """
+    And I apt install `<package>=<installed_version>`
+    And I run `apt-mark hold <package>` with sudo
+    And I run `pro refresh messages` with sudo
+    And I apt upgrade
+    Then stdout matches regexp:
```
orndorffgrant, line 885: Many of these `stdout matches regexp` steps can be replaced by the stricter `stdout contains substring:` step

orndorffgrant (approved):
This looks great! just a nit on the tests.

And the spec isn't actually approved yet so we shouldn't merge this for now.

dheyay:
Updated @orndorffgrant

orndorffgrant:
The CI failure in the apt news test was strange and I couldn't reproduce it locally. If it happens again, we can add a step that clears the pro apt cache which should prevent that message from occurring during the test.

lucasmoura (approved):

renanrodrigo:
Well I had understood that this would wait for a proper spec approval
But we can iterate in further PRs if things change there I guess (:

PR #2977, opened 2024-03-06
tests: small fixes to behave tests

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because we need our integration tests to pass

## Test Steps
Integration tests

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo:
This was a string that was wrong in proxy tests, but I found another small needed fix and appended the commit here.

features/api_configure_retry_service.feature
```diff
@@ -18,6 +18,8 @@ Feature: api.u.pro.attach.auto.configure_retry_service
         """
         When I run `systemctl enable apitest.service` with sudo
         When I reboot the machine
+        # Cloud init may take a while here
+        And I wait `15` seconds
```
dheyay, line 22: Is there a reason for a 15 second wait or it's just the better option given the amount of time cloud-init takes?

dheyay (approved):
Small question. Looks good otherwise, thanks!

features/api_configure_retry_service.feature
renanrodrigo: completely arbitrary
I am sad about this change because of the complex thought process on deciding the time: "on my machine it takes about 10 seconds, so I will set is to 15". 
And why does it take sooo long only on Bionic? Other series were fine...

dheyay: Makes sense, thanks!

orndorffgrant (approved):

Issue #2903, opened 2024-01-02
Feature: behave tests: measure code coverage


orndorffgrant:
**Please describe the scenario where the new feature would be useful**

We're starting to grade our code quality and code coverage is an important metric. We measure coverage of unit tests but not behave/integration tests.

**Describe the solution you'd like**

python3-coverage can be installed and configured on integration test systems. The coverage results can be pulled and combined for all integration tests using coverage.py.

**Current behavior**

No coverage is measured for integration tests currently.

PR #2979, opened 2024-03-06
Add coverage for behave tests

dheyay:
## Why is this needed?

We did not have code coverage for the behave integration tests. This captures code coverage for the feature files when the tests are run.

### Obtaining coverage
Obtaining coverage is optional. To enable it,  set the following environment variable:
`export UACLIENT_BEHAVE_COLLECT_COVERAGE=True`

The `.coverage` files are saved in the `artifacts/coverage/` directory for each run.
To obtain converage, run the following in the `ubuntu-pro-client` directory:

`python3-coverage combine --rcfile=.coveragerc /artifacts/coverage/*` 
`python3-coverage report`

This PR is for issue: #2903.

Note: 
Coverage reports only being generated for `jammy` and `mantic` currently. `xenial, bionic, focal` are excluded from the integration test coverage due to difference in python and python-coverage versions. Most current method of reporting code coverage does not support older versions.
Breaking change in python-coverage version 5.0.0. [Read more about the change](https://coverage.readthedocs.io/en/latest/whatsnew5x.html)

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[paths]
+source = 
+	uaclient/
+	/usr/lib/python3/dist-packages/uaclient/
--- a/features/environment.py
+++ b/features/environment.py
@@ -412,6 +412,59 @@ def before_scenario(context: Context, scenario: Scenario):
             )
 
 
+def after_scenario(context, scenario):
+    """Collect the coverage files after the scenario is run."""
+    cov_dir = os.path.join(context.pro_config.artifact_dir, "coverage")
+    if not os.path.exists(cov_dir):
+        os.makedirs(cov_dir)
+
+    inner_dir = os.path.join(
+        datetime.datetime.now().strftime("%Y-%m-%d"),
+        "{}".format(os.path.basename(scenario.filename.replace(".", "_"))),
+    )
+    new_artifacts_dir = os.path.join(
+        cov_dir,
+        inner_dir,
+    )
+    if not os.path.exists(new_artifacts_dir):
+        os.makedirs(new_artifacts_dir)
+    if hasattr(context, "machines") and SUT in context.machines:
+        try:
+            scenario_name = (
+                os.path.basename(scenario.filename.replace(".", "_"))
+                + "_"
+                + str(scenario.line)
+            )
+            cov_filename = ".coverage.{}".format(scenario_name)
+            context.machines[SUT].instance.execute(
+                [
+                    "bash",
+                    "-c",
+                    "mv .coverage /tmp/{cov_filename}".format(
+                        cov_filename=cov_filename
+                    ),
+                ],
+                use_sudo=True,
+            )
+            context.machines[SUT].instance.execute(
+                [
+                    "chmod",
+                    "666",
+                    "/tmp/{cov_filename}".format(cov_filename=cov_filename),
+                ],
+                use_sudo=True,
+            )
+
+            dest = os.path.join(new_artifacts_dir, cov_filename)
+            context.machines[SUT].instance.pull_file(
+                "/tmp/{cov_filename}".format(cov_filename=cov_filename), dest
+            )
+            logging.warning("Done collecting coverage.")
+        except Exception as e:
+            logging.error(str(e))
+            logging.warning("Failed to collect coverage")
+
+
 def after_step(context, step):
     """Collect test artifacts in the event of failure."""
     if step.status == "failed":
--- a/features/environment.py
+++ b/features/environment.py
@@ -69,6 +69,7 @@ class UAClientBehaveConfig:
     # environment variable input to the appropriate Python types for use within
     # the test framework
     boolean_options = [
+        "collect_coverage",
         "destroy_instances",
         "ephemeral_instance",
         "snapshot_strategy",
@@ -114,6 +115,7 @@ class UAClientBehaveConfig:
         self,
         *,
         cloud_credentials_path: Optional[str] = None,
+        collect_coverage: bool = False,
         destroy_instances: bool = True,
         ephemeral_instance: bool = False,
         snapshot_strategy: bool = False,
@@ -141,6 +143,7 @@ class UAClientBehaveConfig:
     ) -> None:
         # First, store the values we've detected
         self.cloud_credentials_path = cloud_credentials_path
+        self.collect_coverage = collect_coverage
         self.ephemeral_instance = ephemeral_instance
         self.snapshot_strategy = snapshot_strategy
         self.sbuild_output_to_terminal = sbuild_output_to_terminal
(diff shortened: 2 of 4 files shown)


dheyay:
Integration tests break due to python version difference on ubuntu releases

.coveragerc
```diff
@@ -0,0 +1,5 @@
+[paths]
+source = 
+	uaclient/
+	ubuntu-pro-client/uaclient/
+	/usr/lib/python3/dist-packages/uaclient/
```
orndorffgrant, line 5: I understand `uaclient` and `/usr/lib/...` but why do we need `ubuntu-pro-client/uaclient` ?

features/environment.py
```diff
@@ -412,6 +412,59 @@ def before_scenario(context: Context, scenario: Scenario):
...
 
 
+def after_scenario(context, scenario):
+    """Collect the coverage files after the scenario is run."""
+    cov_dir = os.path.join(context.pro_config.artifact_dir, "coverage")
+    if not os.path.exists(cov_dir):
+        os.makedirs(cov_dir)
+
+    inner_dir = os.path.join(
+        datetime.datetime.now().strftime("%Y-%m-%d"),
```
orndorffgrant, line 422: We may not actually need the directory with a date name. 
Our unit tests just overwrite the existing .coverage file if there is one.
What do you think?

features/environment.py
```diff
@@ -412,6 +412,59 @@ def before_scenario(context: Context, scenario: Scenario):
...
+                + str(scenario.line)
+            )
+            cov_filename = ".coverage.{}".format(scenario_name)
+            context.machines[SUT].instance.execute(
+                [
+                    "bash",
+                    "-c",
+                    "mv .coverage /tmp/{cov_filename}".format(
+                        cov_filename=cov_filename
+                    ),
```
orndorffgrant, line 445: I don't think this needs to execute in bash - or am I missing something?
```suggestion
                    "mv",
                    ".coverage",
                    "/tmp/{}".format(cov_filename),
```

features/environment.py
```diff
@@ -412,6 +412,59 @@ def before_scenario(context: Context, scenario: Scenario):
...
+                    "chmod",
+                    "666",
+                    "/tmp/{cov_filename}".format(cov_filename=cov_filename),
+                ],
+                use_sudo=True,
+            )
+
+            dest = os.path.join(new_artifacts_dir, cov_filename)
+            context.machines[SUT].instance.pull_file(
+                "/tmp/{cov_filename}".format(cov_filename=cov_filename), dest
```
orndorffgrant, line 460: We construct this string three times in a row. Consider putting it in a variable and re-using it

features/environment.py
orndorffgrant: Also, I'd prefer to use an absolute path to the .coverage file

features/steps/machines.py
```diff
@@ -110,6 +110,9 @@ def given_a_machine(
     # make sure the machine has up-to-date apt data
     when_i_apt_update(context, machine_name=machine_name)
 
+    # add coverage
+    when_i_apt_install(context, "python3-coverage", machine_name=machine_name)
```
orndorffgrant, line 114: I think we may want to make the whole act of collecting coverage optional.
You can add a new field to `UAClientBehaveConfig` and then check it throughout like so:
```suggestion
    if context.pro_config.collect_coverage:
        when_i_apt_install(context, "python3-coverage", machine_name=machine_name)
```

features/steps/shell.py
```diff
@@ -64,7 +64,17 @@ def when_i_run_command(
 ):
     command = process_template_vars(context, command)
     prefix = get_command_prefix_for_user_spec(user_spec)
-    full_cmd = prefix + shlex.split(command)
+    if "pro" in command.split():
+        command = command.replace(
+            "pro",
+            "python3 -m coverage run \
+                --source=/usr/lib/python3/dist-packages/uaclient /usr/bin/pro",
```
orndorffgrant, line 71: I'd rather use `# noqa: E501` at the end of the line than break up a string like this to keep it less than 80 characters wide.

features/steps/shell.py
```diff
@@ -64,7 +64,17 @@ def when_i_run_command(
 ):
     command = process_template_vars(context, command)
     prefix = get_command_prefix_for_user_spec(user_spec)
-    full_cmd = prefix + shlex.split(command)
+    if "pro" in command.split():
+        command = command.replace(
+            "pro",
```
orndorffgrant, line 69: this would also turn `echo proposed` into `echo python3 -m coverage run  --source=/usr/lib/python3/dist-packages/uaclient /usr/bin/proposed`

I think we only want to check if `command.split[0] == "pro"`, replace the first item, and then `" ".join` it back together before `shlex.split`ing it.

features/steps/shell.py
```diff
@@ -64,7 +64,17 @@ def when_i_run_command(
...
+            "python3 -m coverage run \
+                --source=/usr/lib/python3/dist-packages/uaclient /usr/bin/pro",
+            1,
+        )
+
+    split_command = shlex.split(command)
+    print(split_command)
+    full_cmd = prefix + split_command
```
orndorffgrant, line 77: If this is working, we can consolidate this back into one line
```suggestion
    full_cmd = prefix + shlex.split(command)
```

features/steps/machines.py
dheyay: Yes, that's a great idea

features/environment.py
dheyay: I feel instead of the date folder, we can just have the coverage folder which has the `.coverage` for each feature and we can collect it from there.

The date folder was there just in case comparing coverage was ever needed, so it can be removed

features/steps/shell.py
dheyay: This print statement slipped through, should be working so will revert to one line

.coveragerc
dheyay: It was added for testing relational paths, removed now

dheyay:
Should be updated now @orndorffgrant

features/environment.py
```diff
@@ -412,6 +415,58 @@ def before_scenario(context: Context, scenario: Scenario):
...
+                )
+                cov_filename = ".coverage.{}".format(scenario_name)
+                tmp_covfile_path = "/tmp/{cov_filename}".format(
+                    cov_filename=cov_filename
+                )
+                context.machines[SUT].instance.execute(
+                    [
+                        "mv",
+                        "/home/ubuntu/.coverage",
+                        f"{tmp_covfile_path}",
```
lucasmoura, line 447: Let's not use f-strings here because we might run our tests on CI that still don't support it :(

features/environment.py
```diff
@@ -412,6 +415,58 @@ def before_scenario(context: Context, scenario: Scenario):
...
+                        "chmod",
+                        "666",
+                        f"{tmp_covfile_path}",
+                    ],
+                    use_sudo=True,
+                )
+
+                dest = os.path.join(new_artifacts_dir, cov_filename)
+                context.machines[SUT].instance.pull_file(
+                    f"{tmp_covfile_path}", dest
```
lucasmoura, line 462: same f-string comment here

features/environment.py
```diff
@@ -412,6 +415,58 @@ def before_scenario(context: Context, scenario: Scenario):
...
+        if not os.path.exists(new_artifacts_dir):
+            os.makedirs(new_artifacts_dir)
+        if hasattr(context, "machines") and SUT in context.machines:
+            try:
+                scenario_name = (
+                    os.path.basename(scenario.filename.replace(".", "_"))
+                    + "_"
+                    + str(scenario.line)
+                )
+                cov_filename = ".coverage.{}".format(scenario_name)
```
lucasmoura, line 439: Any reason why the coverage file needs to start with a dot ?

features/environment.py
dheyay: sure, will just use a string format instead

features/environment.py
dheyay: I think it's naming convention followed universally for coverage files to keep them hidden when generated so it does not clutter the user's file system given it is only used by developers for internal usage. Also, the python coverage library by default looks for .coverage files

dheyay:
Updated @lucasmoura @orndorffgrant

lucasmoura (approved):

features/steps/shell.py
```diff
@@ -64,6 +64,15 @@ def when_i_run_command(
...
     prefix = get_command_prefix_for_user_spec(user_spec)
+
+    if context.pro_config.collect_coverage:
+        if command.split()[0] == "pro":
+            split_command = command.split()
+            split_command[
+                0
+            ] = "python3 -m coverage run --source=/usr/lib/python3/dist-packages/uaclient /usr/bin/pro"  # noqa: E501
+            command = " ".join(split_command)
+
```
renanrodrigo, line 75: Different syntax suggestion, a little clearer to me and avoiding the noqa.
```suggestion
        coverage_command = (
            "python3 -m coverage run "
            "--source=/usr/lib/python3/dist-packages/uaclient /usr/bin/pro"
        )
        split_command = command.split()
        if split_command[0] == "pro":
            command = " ".join([coverage_command] + split_command[1:])

```

renanrodrigo (approved):
LGTM overall, two small suggestions before we merge.

features/steps/shell.py
dheyay: Yeah, this looks better, thanks!

dheyay: Updated @renanrodrigo

Issue #2925, opened 2024-01-24
Bug: `pro config set http_proxy=` doesn't fail and doesn't unset


orndorffgrant:
**Description of the bug**

`pro config set http_proxy=` doesn't behave as expected

**Expected behavior**

It should error because a valid proxy url was not passed.

**Current behavior**

```
# Set
sudo pro config set http_proxy=http://squid.internal:3128
sudo pro config set https_proxy=http://squid.internal:3128
sudo snap get system proxy.https # expected output

# Set to empty: does not error, does not unset
sudo pro config set http_proxy=
sudo pro config set https_proxy=
sudo snap get system proxy.https # unexpected output

# Unset: unsets properly
sudo pro config unset http_proxy
sudo pro config unset https_proxy
sudo snap get system proxy.https # expected output
```

**System information:**

- Pro Client version: v30

PR #2984, opened 2024-03-08
Handle empty value for config set proxy

dheyay:
## Why is this needed?

Currently the user can set config proxy value to empty using `pro config set http_proxy=`
This fixes that, since the given proxy is invalid, an error is shown.

Fixes: [#2925]

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/uaclient/cli/__init__.py
+++ b/uaclient/cli/__init__.py
@@ -690,6 +690,9 @@ def action_config_set(args, *, cfg, **kwargs):
         raise exceptions.InvalidArgChoice(
             arg="<key>", choices=", ".join(config.UA_CONFIGURABLE_KEYS)
         )
+    if not set_value.strip():
+        subparser.print_help()
+        raise exceptions.EmptyConfigValue(arg=set_key)
     if set_key in ("http_proxy", "https_proxy"):
         protocol_type = set_key.split("_")[0]
         if protocol_type == "http":
--- a/uaclient/exceptions.py
+++ b/uaclient/exceptions.py
@@ -335,7 +335,6 @@ class EntitlementNotFoundError(UbuntuProError):
 
 
 class EntitlementsNotEnabledError(UbuntuProError):
-
     exit_code = 4
     _msg = messages.E_ENTITLEMENTS_NOT_ENABLED_ERROR
 
@@ -494,6 +493,10 @@ class InvalidSecurityIssueIdFormat(UbuntuProError):
 
 class InvalidArgChoice(UbuntuProError):
     _formatted_msg = messages.E_CLI_VALID_CHOICES
+
+
+class EmptyConfigValue(UbuntuProError):
+    _formatted_msg = messages.E_CLI_EMPTY_CONFIG_VALUE
 
 
 class GenericInvalidFormat(UbuntuProError):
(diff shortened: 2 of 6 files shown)


dheyay:
Might need updated translations, also open to suggestions for a solution to re-use other error messages than creating a new one

uaclient/messages/__init__.py
```diff
@@ -2425,6 +2425,11 @@ def __repr__(self):
     "invalid-arg-choice", "\n" + t.gettext("{arg} must be one of: {choices}")
 )
 
+E_CLI_VALID_VALUE = FormattedNamedMessage(
```
orndorffgrant, line 2428:
```suggestion
E_CLI_EMPTY_CONFIG_VALUE = FormattedNamedMessage(
```

uaclient/exceptions.py
```diff
@@ -496,6 +495,10 @@ class InvalidArgChoice(UbuntuProError):
     _formatted_msg = messages.E_CLI_VALID_CHOICES
 
 
+class InvalidArgValue(UbuntuProError):
+    _formatted_msg = messages.E_CLI_VALID_VALUE
```
orndorffgrant, line 499:
```suggestion
class EmptyConfigValue(UbuntuProError):
    _formatted_msg = messages.E_CLI_EMPTY_CONFIG_VALUE
```

uaclient/cli/__init__.py
```diff
@@ -690,6 +690,9 @@ def action_config_set(args, *, cfg, **kwargs):
         raise exceptions.InvalidArgChoice(
             arg="<key>", choices=", ".join(config.UA_CONFIGURABLE_KEYS)
         )
+    if not set_value.strip():
+        subparser.print_help()
+        raise exceptions.InvalidArgValue(arg=set_key)
```
orndorffgrant, line 695:
```suggestion
        raise exceptions.EmptyConfigValue(arg=set_key)
```

orndorffgrant (commented):
Thanks!
Let's name the error and message more specifically.
Otherwise, looks good!

dheyay:
Updated @orndorffgrant

renanrodrigo:
I am sad about this nit, but:
could you just adapt the commit message comment to say
`Fixes: #2925`
instead of 
`Fixes Bug: [#2925]`

For CI reasons, we always use that format :/

uaclient/cli/__init__.py
```diff
@@ -690,6 +690,9 @@ def action_config_set(args, *, cfg, **kwargs):
         raise exceptions.InvalidArgChoice(
             arg="<key>", choices=", ".join(config.UA_CONFIGURABLE_KEYS)
         )
+    if not set_value.strip():
+        subparser.print_help()
```
lucasmoura, line 694: Do we need to print the help message here ? Maybe just stating that the user provided an invalid empty value is enough.

What do you think @renanrodrigo ?

renanrodrigo: This is following what we do for all other errors on setting config values. I think this looks good as is.

renanrodrigo (approved):
LGTM

PR #2986, opened 2024-03-12
Merge `release` and `main`

orndorffgrant:
## Why is this needed?

After this we can tag the last commit on `main` as 31.2. With that, our latest release tag will be on `main` which frees up our git branching strategy from the burden of a the parallel history of `release`. We can potentially drop the `release` branch altogether. I'll be following up with proposed release process documentation changes.


orndorffgrant:
That's interesting, not sure if this is what we want. It might make more sense to merge locally and push?

PR #2990, opened 2024-03-13
more accurate reboot-required "no" description

orndorffgrant:
This could technically not always be true. @lucasmoura - you've been in reboot-required discussions, do you think this change is sufficient to resolve https://warthogs.atlassian.net/browse/SC-1416 ?


renanrodrigo (approved):
@lucasmoura approves

PR #2992, opened 2024-03-14
auto-attach: make error message clearer

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because now we tell people to run `auto-attach` after upgrades, and users are seeing the generic error.
When a machine tries to auto-attach and the contract server returns an error, print hints on what could have happened together with the error details.

## Test Steps
Integration test is good, can be tried in any cloud manually.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/po/pt_BR.po
+++ b/debian/po/pt_BR.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-03-18 12:55-0700\n"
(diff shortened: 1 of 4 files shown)


renanrodrigo:
> Please add a comment with justification of why messages are being altered

Updating the message is the whole point of this PR

> If the changes are substantial, then we will ask translators to take a look and update translations if they have time, but it will not block this PR.

No translation for this one yet anyway.

orndorffgrant (approved):
LGTM!

renanrodrigo:
rebased and got rid of conflicts

renanrodrigo:
re rebased and the maybe line is gone; merging

PR #2994, opened 2024-03-14
fix typo

s-makin:
Fix typo in configure_proxies doc


renanrodrigo:
It was Murican!
So @s-makin  if you think you can just open PRs and merge them yourself without any approval
You are completely right
Do it
And thanks for helping us improve docs and polish rough edges

Issue #2996, opened 2024-03-14
Bug: `tox -e test` fails on a system that has `/var/lib/ubuntu-advantage/user-config.json`


orndorffgrant:
**Description of the bug**

`tox -e test` fails on a system that has `/var/lib/ubuntu-advantage/user-config.json` 
There is a leak in the unit tests where they try to read the actual system user-config.json file (and fail).

**Expected behavior**

`tox -e test` should not try to read any files on the system.

**Current behavior**

`tox -e test` fails on a system that has `/var/lib/ubuntu-advantage/user-config.json` 

**To Reproduce**

Please include details on how to reproduce the bug.
1. `sudo pro config set update_messaging_timer=21601`
2. `tox -e test`

**System information:**

- Ubuntu release: mantic
- Pro Client version: 31.2

**Additional context**

Likely a result of https://github.com/canonical/ubuntu-pro-client/pull/2945 - unfortunately CI passed there :/

PR #2998, opened 2024-03-14
tests: Update tests which call config status

dheyay:
Fixes: #2996

## Why is this needed?

This PR solves all of our problems because  a bug was introduced earlier which allowed the unit tests to read a system file, which should not be happening.

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/uaclient/cli/tests/test_cli_attach.py
+++ b/uaclient/cli/tests/test_cli_attach.py
@@ -24,6 +24,7 @@ from uaclient.exceptions import (
     NonRootUserError,
     UbuntuProError,
 )
+from uaclient.files.user_config_file import UserConfigData
 from uaclient.testing.fakes import FakeFile, FakeUbuntuProError
 from uaclient.yaml import safe_dump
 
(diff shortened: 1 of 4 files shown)


uaclient/cli/tests/test_cli_attach.py
```diff
@@ -553,6 +560,7 @@ def test_attach_when_one_service_fails_to_enable(
         FakeConfig,
         event,
     ):
+        m_public_config.return_value = UserConfigData()
```
orndorffgrant, line 563: `return_value` is already set above in the decorator, so no need to do it again here
```suggestion
```

orndorffgrant (approved):
Looks good! Just a nit

dheyay:
@orndorffgrant should be good now!

renanrodrigo (approved):
This is good for now.
@lucasmoura correctly mentioned this should be a fixture, so we don't leak it anywhere in the tests - but that is true for all state files isn't it?

PR #3010, opened 2024-03-21
Add detach API endpoint

lucasmoura:
## Why is this needed?
We are adding an endpoint for the detach functionality.

Notice that we still need to refactor it in the future to make the CLI logic use that endpoint.

## Test Steps
Run the new integration tests

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/api/tests/test_api_u_pro_detach_v1.py
+++ b/uaclient/api/tests/test_api_u_pro_detach_v1.py
@@ -1,0 +1,145 @@
+import mock
+import pytest
+
+from uaclient import entitlements, exceptions, messages
+from uaclient.api.data_types import ErrorWarningObject
+from uaclient.api.u.pro.detach.v1 import DetachResult, _detach
+from uaclient.entitlements.entitlement_status import (
+    CanDisableFailure,
+    CanDisableFailureReason,
+)
+
+M_PATH = "uaclient.api.u.pro.detach.v1."
+
+
+@mock.patch("uaclient.lock.RetryLock.__enter__")
+class TestDetachV1:
+    @mock.patch(M_PATH + "_is_attached")
+    def test_detach_when_unattached(self, m_is_attached, _m_lock_enter):
+        m_is_attached.return_value = mock.MagicMock(is_attached=False)
+        assert DetachResult(disabled=[], reboot_required=False) == _detach(
+            mock.MagicMock()
+        )
+
+    @mock.patch("uaclient.util.we_are_currently_root")
+    def test_detach_when_non_root(
+        self, m_we_are_currently_root, _m_lock_enter
+    ):
+        m_we_are_currently_root.return_value = False
+        with pytest.raises(exceptions.NonRootUserError):
+            _detach(mock.MagicMock())
+
+    @mock.patch("uaclient.timer.stop")
+    @mock.patch("uaclient.daemon.start")
+    @mock.patch("uaclient.files.state_files.delete_state_files")
+    @mock.patch(M_PATH + "_reboot_required")
+    @mock.patch(M_PATH + "_is_attached")
+    @mock.patch(M_PATH + "update_motd_messages")
+    def test_detach(
+        self,
+        m_update_motd_messages,
+        m_is_attached,
+        m_reboot_required,
+        m_delete_state_files,
+        m_daemon_start,
+        m_timer_stop,
+        _m_lock_enter,
+        mock_entitlement,
+    ):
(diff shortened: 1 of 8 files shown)


uaclient/api/u/pro/detach/v1.py
```diff
@@ -0,0 +1,100 @@
...
+        except exceptions.EntitlementNotFoundError:
+            continue
+
+        ent = ent_cls(cfg=cfg, assume_yes=True)
+        # For detach, we should not consider that a service
+        # cannot be disabled because of dependent services,
+        # since we are going to disable all of them anyway
+        can_disable, _ = ent.can_disable(ignore_dependent_services=True)
+        if can_disable:
+            ret, reason = ent.disable(silent=True)
```
dheyay, line 58: Would it be clearer to rename the `ret` variable to something like `is_disabled` for better readability?

uaclient/api/tests/test_api_u_pro_detach_v1.py
```diff
@@ -0,0 +1,141 @@
...
+            name="ent5",
+            can_disable=(True, None),
+            disable=(
+                False,
+                None,
+            ),
+        )
+
+        def ent_factory_side_effect(cfg, name):
+            if name == "ent1":
```
dheyay, line 88: Is there a specific reason why the `m_ent_cls` variables are assigned inside the `ent_factory_side_effect` function, even though they don't seem to be used outside of it? Could we just return the mock entitlement classes directly from the function without assigning them to variables?

dheyay (commented):
Looks good, just had some really small questions. Thanks! @lucasmoura

uaclient/api/tests/test_api_u_pro_detach_v1.py
lucasmoura: Yes, we can. I just used the variables for readability purposes. But if everyone agrees we should define everything inside the side_effect function, I don't see any problem doing that change

uaclient/api/u/pro/detach/v1.py
```diff
@@ -0,0 +1,100 @@
...
+def detach() -> DetachResult:
+    return _detach(UAConfig())
+
+
+def _detach(cfg: UAConfig) -> DetachResult:
+    if not util.we_are_currently_root():
+        raise exceptions.NonRootUserError
+
+    if not _is_attached(cfg).is_attached:
+        raise exceptions.UnattachedError()
```
orndorffgrant, line 42: I think we should have an early return success here with an empty disabled list - similar to how we decided "attach" should act if already attached.

lucasmoura: Maybe I am overthinking this, but in the situation where desktop consumes that API, wouldn't it be better to tell the user that the machine is not attached than always saying to the user that detach "worked".

But let me know if this is overthinking

lucasmoura: We have agreed that this should not be an error. I will update the PR to reflect that

lucasmoura:
@orndorffgrant @dheyay updated

uaclient/api/u/pro/detach/v1.py
```diff
@@ -0,0 +1,102 @@
...
+            ent_cls = entitlements.entitlement_factory(cfg=cfg, name=ent_name)
+        except exceptions.EntitlementNotFoundError:
+            continue
+
+        ent = ent_cls(cfg=cfg, assume_yes=True)
+        # For detach, we should not consider that a service
+        # cannot be disabled because of dependent services,
+        # since we are going to disable all of them anyway
+        can_disable, _ = ent.can_disable(ignore_dependent_services=True)
+        if can_disable:
```
orndorffgrant, line 60: Do we need this check? `disable` will call it and disable dependent services with `assume_yes=True`

orndorffgrant (approved):
Just some final nits

uaclient/api/u/pro/detach/v1.py
lucasmoura: The rationale here is to avoid having warnings for services that should not be disabled anyway, as we are looping on every service we have.

I can remove that check, if we instead loop into only services returned by `_enabled_services`. But let me know what you think ?

orndorffgrant: Ohhh now I understand - I do think the intent would be more clear if we looped over `enabled_services` instead, but this is fine too - up to you.

renanrodrigo (approved):
I checked all the CI errors, and all of them are related to a Contracts Server outage; all in this PR is good

PR #3012, opened 2024-03-25
Fix broken landscape link

s-makin:
The link is not actually "broken", but it does fail link checks due to the string of redirects. I have updated it to a more direct link to the manpage in question.

### DIFF
--- a/docs/howtoguides/enable_landscape.rst
+++ b/docs/howtoguides/enable_landscape.rst
@@ -74,4 +74,4 @@ management.
 
 .. include:: ../links.txt
 
-.. _parameters supported by: https://manpages.ubuntu.com/landscape-config
+.. _parameters supported by: https://manpages.ubuntu.com/manpages/en/man1/landscape-config.1.html
--- a/docs/howtoguides/enable_landscape.rst
+++ b/docs/howtoguides/enable_landscape.rst
@@ -74,4 +74,4 @@ management.
 
 .. include:: ../links.txt
 
-.. _parameters supported by: https://manpages.ubuntu.com/manpages/en/man1/landscape-config.1.html
+.. _parameters supported by: https://manpages.ubuntu.com/manpages/noble/en/man1/landscape-config.1.html


docs/howtoguides/enable_landscape.rst
```diff
@@ -74,4 +74,4 @@ management.
 
 .. include:: ../links.txt
 
-.. _parameters supported by: https://manpages.ubuntu.com/landscape-config
+.. _parameters supported by: https://manpages.ubuntu.com/manpages/noble/en/man1/landscape-config.1.html
```
renanrodrigo, line 77: In this particular case, if we are not running the redirects, how are we going to track the release where the manpage comes from? I see the link now points to `noble` but shouldn't we rely on the not-changing link to redirect to `o`, `p` or `q` etc; when they come out?

s-makin: I agree, I would prefer it to be release-agnostic, but in that case we might a problem with the link checker. The original link is 404'ing in the Pro Handbook docs (I copy this page over exactly. Is there something special about the way we've set up our GH actions that it ignores this URL that I need to port over there?

renanrodrigo: Hmmm no I don't think we do any magic here... I don't know either why the link checker is unhappy with the redirect (as it does recognize redirecting other things) or why this would 404 (I couldn't get a 404 here in many tries with different agents).
In the abscence of something better, we can stick with this link, but then remember to update it every 6 months :D :(

s-makin: That was the weird thing, I also couldn't get it to 404, but the link checker insists that it does. I was actually about to close this PR unmerged and leave it as it was lol.

orndorffgrant: @renanrodrigo @s-makin We have the following in our conf.py
```
linkcheck_ignore = [
    "https://manpages.ubuntu.com/landscape-config"
]
```

orndorffgrant: The redirect is actually a 404 not found page that has javascript that guesses where you wanted to go :shrug:

renanrodrigo: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

Issue #2284, opened 2022-10-06
Conflict with unattended-upgrades should be handled more gracefully


cpaelzer:
If people boot systems they had off for some time or images that are outdated or even in many other cases chances are quite high that right after reboot unattended-upgrades will run.

When the users then want to enable pro they will get something like:

```
ubuntu@x:~$ sudo pro attach <...>
Enabling default service esm-apps
Updating package lists
Stderr: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/

Stdout: Reading package lists...

Stderr: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/

Stdout: Reading package lists...

Stderr: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/

Stdout: Reading package lists...

Stderr: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/

Stdout: Reading package lists...

Enabling default service esm-infra
Updating package lists
Stderr: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/

Stdout: Reading package lists...

Stderr: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/

Stdout: Reading package lists...

Stderr: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/

Stdout: Reading package lists...

Stderr: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/

Stdout: Reading package lists...

Enabling default service livepatch
Canonical livepatch enabled.
Failed to enable default services, check: sudo pro status
```

IMHO We should
- pre-detect if apt is running
- maybe hint at what triggered it (like unattended-upgrades or user foo running apt or ... [depends what we can detect)
- offer an "do you want to try later, or should I wait a bit longer than normal" option?
[...]

PR #3017, opened 2024-03-26
apt: better ux on lock held

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it will improve the experience users have with apt locks being held.

This is blocked on the Progress Reporting implementation, because it would duplicate logic to identify CLI calls unnecessarily.

## Test Steps
n/a

## Checklist
 
- [ ] Yes
- [x] No

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -7,6 +7,7 @@ import os
 import re
 import subprocess
 import tempfile
+import time
 from functools import lru_cache, wraps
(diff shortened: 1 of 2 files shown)


renanrodrigo:
This aims at #2284, and is now waiting on a good time to move forward

renanrodrigo:
This PR is sitting here for a while now, and although the idea kind of makes sense, this effort was unprioritized at this time. I am closing this until we revisit the issue and are able to actually solve it in a better way.

PR #3020, opened 2024-03-27
fix: wait for cloud-init.service to fully activate

### LINKED ISSUE #3000: Bug: pro sometimes runs before cloud-config.service
**Description of the bug**

Due to commit 0ea591f8dd87803908d8493a81d8c29f97b1924b pro does not strictly run after cloud-config.service.  If cloud-config.service has not been started when pro runs, it can complete before cloud-config.service begins and thus the user-specificed pro configuration will be ignored since the instance is already attached.

**Expected behavior**

When cloud-config.service has yet to run, ubuntu-advantage.service should wait until it's finished before running.

**Current behavior**

On instances with pro version `31.2~22.04` and cloud-init version `23.4.4-0ubuntu0~22.04.1`, ubuntu-advantage.service consistently runs prior to cloud-config.service, sufficiently before that cloud-config.service is not in the `activating` state when pro runs.

**To Reproduce**

Using userdata.yaml:

```
#cloud-config

ubuntu_advantage:
  enable: []
```

Set up a VM with the conflicting versions of pro and cloud-init:
```
$ gcloud compute instances create pro-order-bug --image daily-ubuntu-pro-2204-jammy-v20240314 --image-project ubuntu-os-cloud-image-proposed --metadata-from-file=user-data=userdata.yaml
```

Log onto the VM and observe that (a) pro services are enabled against user request and (b) cloud-init reports a warning:
```
$ pro status
SERVICE          ENTITLED  STATUS       DESCRIPTION
anbox-cloud      yes       disabled     Scalable Android in the cloud
esm-apps         yes       enabled      Expanded Security Maintenance for Applications
esm-infra        yes       enabled      Expanded Security Maintenance for Infrastructure
fips-preview     yes       disabled     Preview of FIPS crypto packages undergoing certification with NIST
livepatch        yes       enabled      Canonical Livepatch service
usg              yes       disabled     Security compliance and audit tools

For a list of all Ubuntu Pro services, run 'pro status --all'
Enable services with: pro enable <service>

                Account: ubuntu-catred
           Subscription: ubuntu-catred
            Valid until: Fri Dec 31 00:00:00 9999 UTC

$ cat /var/log/cloud-init.log | grep 'WARNING'
2024-03-15 19:44:47,211 - cc_ubuntu_advantage.py[WARNING]: The instance is already attached to Pro. Leaving enabled services untouched. Ignoring config directives ubuntu_advantage: enable and enable_beta
```

**System information:**

- Ubuntu release: 22.04 (but the ordering problem has been observed in all versions 23.10 and earlier)
- Pro Client version: 31.2

**Additional context**

This is related to LP#2057965, where the same ordering issue caused google-startup-script tests to fail.  As a result, the images where this bug can be observed were not published to the GCE marketplace.  I was not able to easily reproduce the issue after first boot.

catmsred:
Once the systemd constraint on After cloud-config.service was removed, there are situations where cloud-config.service runs so long after pro that the python code that waits for cloud-config.service to finish activating is being bypassed as if cloud-init isn't on the system.  This change forces pro to wait for cloud-config to finish if cloud-init is running at all.

Fixes #3000

```
fix: wait for cloud-init.service to fully activate

Once the systemd constraint on After cloud-config.service was
removed, there are situations where cloud-config.service runs so
long after pro that the python code that waits for
cloud-config.service to finish activating is being bypassed as if
cloud-init isn't on the system.  This change forces pro to wait
for cloud-config to finish if cloud-init is running at all.

Fixes: #3000 

```

## Checklist

 - [ ] I have updated or added any integration tests accordingly [still TODO]

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/tests/test_lib_daemon.py
+++ b/uaclient/tests/test_lib_daemon.py
@@ -17,24 +17,41 @@ class TestWaitForCloudConfig:
         (
             # not activating
             (
-                ["active"],
+                ["active"] * 2,
+                [],
+            ),
+            # inactive (all cloud-init)
+            (
+                ["inactive"] * 2,
                 [],
             ),
             (
-                ["inactive"],
+                [None] * 2,
                 [],
             ),
+            # cloud-config activating, then finishes
+            # cloud-init is active
             (
-                [None],
-                [],
-            ),
-            # activating, then finishes
-            (
-                (["activating"] * 11) + ["active"],
+                (["activating", "active"] * 11) + ["active"] * 2,
                 [mock.call(WAIT_FOR_CLOUD_CONFIG_SLEEP_TIME)] * 11,
             ),
             (
-                (["activating"] * 11) + ["failed"],
+                (["activating", "active"] * 11) + ["failed"] + ["active"],
+                [mock.call(WAIT_FOR_CLOUD_CONFIG_SLEEP_TIME)] * 11,
+            ),
+            # inactive cloud-config, active cloud-init
+            (
+                (["inactive", "active"] * 6)
+                + (["activating", "active"] * 5)
+                + ["active"] * 2,
+                [mock.call(WAIT_FOR_CLOUD_CONFIG_SLEEP_TIME)] * 11,
+            ),
+            # inactive cloud-config, activating cloud-init
+            (
+                (["inactive", "activating"] * 2)
+                + (["inactive", "active"] * 4)
+                + (["activating", "active"] * 5)
+                + ["active"] * 2,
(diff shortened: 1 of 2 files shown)


uaclient/tests/test_lib_daemon.py
```diff
@@ -54,6 +71,7 @@ def test_wait_for_cloud_config(
         active_state_side_effect,
         expected_sleep_calls,
     ):
+        print(active_state_side_effect)
```
renanrodrigo, line 74: This one is not needed anymore (:

PR #3024, opened 2024-03-27
Add API endpoint for the attach operation

lucasmoura:
## Why is this needed?
Add API endpoint for the attach operation. Note that there are still some other additions needed to this work:

* use the endpoint on the CLI
* Guarantee that the token can be passed through cat (i.e. `cat token.json | pro api u.pro.attach.token.full_token_attach.v1 --data`)

## Test Steps
Run the new integration tests

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/api_full_token_attach.feature
+++ b/features/api_full_token_attach.feature
@@ -104,6 +104,36 @@ Feature: Attach API endpoint
     And the machine is attached
     And I verify that `esm-apps` is disabled
     And I verify that `esm-infra` is disabled
+    When I run `pro detach --assume-yes` with sudo
+    And I attach using the API and stdin
+    Then stdout is a json matching the `api_response` schema
+    And the json API response data matches the `attach` schema
+    And API full output matches regexp:
+      """
+      {
+        "_schema_version": "v1",
+        "data": {
+          "attributes": {
+            "enabled": [
+              "esm-apps",
+              "esm-infra"
+            ],
+            "reboot_required": false
+          },
+          "meta": {
+            "environment_vars": []
+          },
+          "type": "FullTokenAttach"
+        },
+        "errors": [],
+        "result": "success",
+        "version": ".*",
+        "warnings": []
+      }
+      """
+    And the machine is attached
+    And I verify that `esm-apps` is enabled
+    And I verify that `esm-infra` is enabled
 
     Examples: ubuntu release
       | release | machine_type  |
--- a/features/api_full_token_attach.feature
+++ b/features/api_full_token_attach.feature
@@ -1,0 +1,113 @@
+Feature: Attach API endpoint
+
+  Scenario Outline: Attach API endpoint
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I verify that running `pro api u.pro.attach.token.full_token_attach.v1 --args token=TOKEN` `as non-root` exits `1`
+    Then stdout is a json matching the `api_response` schema
+    And API full output matches regexp:
+      """
+      {
+        "_schema_version": "v1",
+        "data": {
+          "meta": {
+            "environment_vars": []
+          }
+        },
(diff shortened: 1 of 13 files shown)


orndorffgrant (commented):
This approach looks good! An integration tests looks like it needs updating.

I'm torn on releasing this without the reading args from stdin. Is it small enough to tack onto this PR? If not that's okay

features/api_full_token_attach.feature
```diff
@@ -0,0 +1,83 @@
+Feature: Attach API endpoint
```
renanrodrigo, line 1: Could we have a test here that attaches the machine and uses the option to not enable the services by default, just to assure the functionality on the API?

renanrodrigo (approved):
pending conflicts and a small test request, lgtm

orndorffgrant (commented):
I think the changes in this PR remove the "Ubuntu Pro: ESM Apps enabled" messages that happen during `pro attach`, which is okay with me, but it is failing some behave tests.

uaclient/cli/__init__.py
```diff
@@ -897,6 +897,10 @@ def action_api(args, *, cfg, **kwargs):
     if args.options and args.data:
         raise exceptions.CLIAPIOptionsXORData()
 
+    if args.data and args.data == "-":
+        if not sys.stdin.isatty():
```
orndorffgrant, line 901: I'm curious why wee need the `isatty` check - what does that do?

features/steps/attach.py
```diff
@@ -108,5 +109,22 @@ def when_i_attach_using_the_api_withot_enabling_services(context):
...
+        {{
+          \"token\": \"{}\"
+        }}
+        """.format(
+            context.pro_config.contract_token
+        ),
+    )
+
+    cmd = "sh -c 'cat /tmp/file.json | pro api u.pro.attach.token.full_token_attach.v1 --data -'"  # noqa
     when_i_run_command(context, cmd, "with sudo")
```
orndorffgrant, line 130: I suggest we place the emphasis on getting data from stdin, rather than piping a file, which is just one way of doing that. We also don't need to pipe a file for this test since we have support for setting stdin
```suggestion
@when("I attach using the API by writing token arg to stdin")
def when_i_attach_using_the_api_and_stdin(context):
    when_i_run_command(
        context,
        "pro api u.pro.attach.token.full_token_attach.v1 --data -",
        "with sudo",
        stdin=json.dumps({"token": context.pro_config.contract_token})
    )
```

uaclient/cli/__init__.py
lucasmoura: This is just to guarantee that we will only execute that path if we are being called through a "redirect".
But given your other comment about the integration test, let me see if this logic prevents `stdin` somehow

lucasmoura: I have updated the test and that logic doesn't block the approach

orndorffgrant: I'm still not sure it's necessary - I think it would prevent something like typing:
```
$ sudo pro api.u.pro.attach.token.full_token_attach.v1 --data -
{
"token": token
}
<Ctrl-D>
```
interactively in a shell. which I think would work if we read stdin without this check.

But also that use-case doesn't seem all that important. I'm fine either way

orndorffgrant (approved):

uaclient/cli/__init__.py
renanrodrigo: I would say if someone is typing it to the CLI and <Ctrl-D>ing , they need to type their token there.
If they do, they don't care about not having the token in the CLI at all, but only not having it in history?
I don't know, seems to me that if there is a real use case for this scenario we can remove the check later. But IMHO, this feature makes way more sense for non interactive inputs.

lucasmoura:
CI errors are unrelated

PR #3028, opened 2024-03-28
disable: cli disable progress refactor

orndorffgrant:
## Why is this needed?

This refactors the disable cli command to use the progress api for printing messages to the user, which will allow the api for disable to be implemented without hacking around prompts deep in the call stack.

## Test Steps

CI and tests should all pass

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/cli/cli_util.py
+++ b/uaclient/cli/cli_util.py
@@ -1,7 +1,43 @@
 from functools import wraps
+from typing import Optional
 
-from uaclient import entitlements, exceptions, lock, util
+from uaclient import api, entitlements, exceptions, lock, util
 from uaclient.api.u.pro.status.is_attached.v1 import _is_attached
+
+
+class CLIEnableDisableProgress(api.AbstractProgress):
+    def __init__(self):
+        self.is_interactive = True
+
+    def progress(
+        self,
+        *,
+        total_steps: int,
+        done_steps: int,
+        previous_step_message: Optional[str],
+        current_step_message: Optional[str]
+    ):
+        if current_step_message is not None:
+            print(current_step_message)
+
+    def _on_event(self, event: str, payload):
+        if event == "info":
+            print(payload)
+            return
+        elif event == "message_operation":
+            if not util.handle_message_operations(payload, print):
+                raise exceptions.PromptDeniedError()
+
+
+def _null_print(*args, **kwargs):
+    pass
+
+
+def create_interactive_only_print_function(json_output: bool):
+    if json_output:
+        return _null_print
+    else:
+        return print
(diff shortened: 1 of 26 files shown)


uaclient/entitlements/__init__.py
```diff
@@ -274,3 +272,12 @@ def check_entitlement_apt_directives_are_unique(
                 )
 
     return True
+
+
+def get_title(cfg: UAConfig, ent_name: str, variant=""):
+    try:
+        return entitlement_factory(cfg, ent_name, variant=variant)(
```
lucasmoura, line 279: Just to double check, what is the advantage of this approach ? We are now creating the entitlement object to fetch the title, which is not necessary.

Is that because of we might miss a variant title by using `ENTITLEMENT_NAME_TO_TITLE` ?

uaclient/cli/enable.py
```diff
@@ -216,7 +186,9 @@ def action_enable(args, *, cfg, **kwargs) -> int:
             if json_output:
                 progress = api.ProgressWrapper()
             else:
-                progress = api.ProgressWrapper(CLIEnableProgress())
+                progress = api.ProgressWrapper(
```
lucasmoura, line 189: Even though this is not a problem introduced in this PR, but I noticed that we are now creating the same `entitlement` object on lines 155 and 193. I think we should only do that once

uaclient/cli/cli_util.py
```diff
@@ -1,9 +1,45 @@
...
+        previous_step_message: Optional[str],
+        current_step_message: Optional[str]
+    ):
+        if current_step_message is not None:
+            print(current_step_message)
+
+    def _on_event(self, event: str, payload):
+        if event == "info":
+            print(payload)
+            return
```
lucasmoura, line 26: Just a nit, but I think we can remove the empty return here

uaclient/entitlements/__init__.py
orndorffgrant: It is unfortunately necessary for cis, because the title depends on `self.called_name`

uaclient/cli/enable.py
orndorffgrant: Yep I can consolidate them

orndorffgrant:
@lucasmoura updated!

uaclient/entitlements/__init__.py
renanrodrigo: I cry so much everytime this kind of comment pops out somewhere
We need to kill this horrible bug

lucasmoura (approved):

PR #3032, opened 2024-04-01
refactor: use enable api in cli subcommand

orndorffgrant:
## Why is this needed?

This refactors the enable cli code to use the api for enabling services.
Some adjustments had to be made: 
- the api doesn't support landscape so we need a special case for that
- assume_yes was being set to True by the api, so I extracted that out of the enable logic to be only a cli concern

## Test Steps

All enable tests should work exactly the same as before.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/cli/cli_util.py
+++ b/uaclient/cli/cli_util.py
@@ -25,7 +25,7 @@ class CLIEnableDisableProgress(api.AbstractProgress):
         if event == "info":
             print(payload)
         elif event == "message_operation":
-            if not util.handle_message_operations(payload, print):
+            if not util.handle_message_operations(payload):
                 raise exceptions.PromptDeniedError()
 
 
--- a/uaclient/cli/cli_util.py
+++ b/uaclient/cli/cli_util.py
@@ -7,8 +7,9 @@ from uaclient.files import machine_token
 
 
 class CLIEnableDisableProgress(api.AbstractProgress):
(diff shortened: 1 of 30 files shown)


dheyay (commented):
Looks good to me, pending CI fix

Just the one test failing at - features/attached_enable.feature:187
Other failures in`xenial` and `focal` releases are unrelated to this PR.

uaclient/cli/cli_util.py
```diff
@@ -6,8 +6,9 @@
 
 
 class CLIEnableDisableProgress(api.AbstractProgress):
-    def __init__(self):
-        self.is_interactive = True
+    def __init__(self, *, assume_yes: bool):
+        self.is_interactive = not assume_yes
+        self.assume_yes = assume_yes
```
lucasmoura, line 11: Why do we need both variables here ? Can't we just use `is_interactive` ?

lucasmoura (approved):

uaclient/cli/cli_util.py
orndorffgrant: `progress.is_interactive()` is used for different purposes than `assume_yes` and needs to work on any progress object, even ones that don't have `assume_yes`. So this implementation of `is_interactive` is trivial, but it is useful to keep the two concepts separated.

PR #3033, opened 2024-04-02
api: u.pro.services.disable.v1

orndorffgrant:
## Why is this needed?

This is based on #3015 so that should be merged first.

This introduces u.pro.services.disable.v1 but does not refactor the cli disable command to use it.

## Test Steps

New tests should pass

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/api/u/pro/services/disable/v1.py
+++ b/uaclient/api/u/pro/services/disable/v1.py
@@ -1,0 +1,132 @@
+import logging
+from typing import List, Optional
+
+from uaclient import entitlements, lock, messages, status, util
+from uaclient.api import AbstractProgress, ProgressWrapper, exceptions
+from uaclient.api.api import APIEndpoint
+from uaclient.api.data_types import AdditionalInfo
+from uaclient.api.u.pro.status.enabled_services.v1 import _enabled_services
+from uaclient.api.u.pro.status.is_attached.v1 import _is_attached
+from uaclient.config import UAConfig
+from uaclient.data_types import (
+    BoolDataValue,
+    DataObject,
+    Field,
+    StringDataValue,
+    data_list,
+)
+
+LOG = logging.getLogger(util.replace_top_level_logger_name(__name__))
+
+
+class DisableOptions(DataObject):
+    fields = [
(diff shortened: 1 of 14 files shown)


uaclient/api/u/pro/services/disable/v1.py
```diff
@@ -0,0 +1,127 @@
...
+            disabled=[],
+        )
+
+    ent_cls = entitlements.entitlement_factory(cfg=cfg, name=options.service)
+    entitlement = ent_cls(
+        cfg,
+        assume_yes=True,
+        called_name=options.service,
+        purge=options.purge,
+    )
```
lucasmoura, line 78: Let's verify if the user provided an invalid service

uaclient/api/u/pro/services/disable/v1.py
```diff
@@ -0,0 +1,127 @@
...
+            reason = fail_reason.message
+        else:
+            reason = messages.GENERIC_UNKNOWN_ISSUE
+        raise exceptions.EntitlementNotDisabledError(
+            service=options.service, reason=reason
+        )
+
+    enabled_services_after = _enabled_services_names(cfg)
+
+    progress.finish()
```
lucasmoura, line 108: Should we update the status cache and send the activity ping to the CS here?

orndorffgrant: In standup we decided:
yes update status cache
no don't send activity ping (it will send on a timer anyway)

uaclient/api/u/pro/services/disable/v1.py
orndorffgrant: The exception already raised looks pretty good and I added it to the integration test. It looks like this:
```
        {
          "code": "entitlement-not-found",
          "meta": {
            "entitlement_name": "invalid"
          },
          "title": "could not find entitlement named \"invalid\""
        }
```

orndorffgrant:
@lucasmoura updated!

lucasmoura (approved):

lucasmoura:
CI errors are unrelated

PR #3037, opened 2024-04-02
daemon: remove unnecessary None check for state

lucasmoura:
## Why is this needed?
We had a check to see if the cloud-config.service is not None, which is redundant given the logic we have in place

## Test Steps
Nothing should change here

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/lib/daemon.py
+++ b/lib/daemon.py
@@ -24,13 +24,9 @@ def _wait_for_cloud_config():
         LOG.debug("cloud-init.service state: %r", ci_state)
         # if cloud-config.service is not yet activating but cloud-init is
         # running, wait for cloud-config to start
-        if state is not None and (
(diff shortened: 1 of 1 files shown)


lib/daemon.py
```diff
@@ -24,13 +24,10 @@ def _wait_for_cloud_config():
...
-        if state is not None and (
-            state == "activating"
-            or (
-                state == "inactive"
-                and ci_state is not None
-                and (ci_state == "activating" or ci_state == "active")
-            )
+        if state == "activating" or (
+            state == "inactive"
+            and ci_state is not None
```
orndorffgrant, line 29: this line can be removed too
```suggestion
```

renanrodrigo (approved):

Issue #2706, opened 2023-08-15
Recommend methods of attaching that don't involve secrets on the command line


orndorffgrant:
`pro attach $token` is our default recommendation for attaching, but that involves passing a secret on the command line, which is bad practice.

Where it makes sense, we should change documentation to use either interactive `pro attach` or the file version: `pro attach --attach-config file-with-token.yaml`. Attach with file is preferred - interactive version only sometimes makes sense (e.g. the tutorial)

orndorffgrant:
The main docs have been updated, but we should update the manpage to explicitly recommend `pro attach` and `--attach-config` and maybe not even mention the option to pass on the command line.

PR #3039, opened 2024-04-04
Pro attach docs

s-makin:
## Why is this needed?
It came to my attention that pro attach <token> is no longer the desired or recommended way for users to attach to their subscription except in specific circumstances. 

I have restructured and revamped several pages where this was the instruction we gave, and have also taken the opportunity to bring our docs visually in line with the new Pro Handbook

This includes moving the "Which services are for me?" off the front page and into its own document. 

Fixes: #2706 

 - [ ] Yes
 - [X] No

### DIFF
--- a/docs/explanations/which_services.rst
+++ b/docs/explanations/which_services.rst
@@ -1,0 +1,188 @@
+.. _which-services:
+
+Which services are for me?
+**************************
+
+Ubuntu Pro is a broad subscription designed to meet many different needs. You
+are unlikely to want to use all of these tools and services at the same time,
+so you are free to select the precise stream of updates and versions you want
+to apply on any given machine covered by your Pro subscription.
+
+To help you navigate the services offered through Ubuntu Pro, click on the tab
+that best describes your needs for our suggestions on which services might be
+of interest to you. For example, Pro includes a set of package versions that
+are compliant with FIPS regulations. You will only want these versions on
+machines that *need* to meet FIPS requirements, so you can choose to enable
+that stream specifically on those machines.
+
+.. tab-set::
+
+    .. tab-item:: Commercial
+
+       * **Expanded Security Maintenance (ESM)**
+
+         If you're using any Ubuntu LTS -- Desktop or Server -- in a commercial
+         setting and you rely on 'Universe' packages, or if you're on an older
+         LTS and don't want to upgrade yet, you can get additional security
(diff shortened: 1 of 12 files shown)


docs/explanations/which_services.rst
```diff
@@ -0,0 +1,188 @@
...
+    .. tab-item:: Personal user
+
+       * **Expanded Security Maintenance (ESM)**
+
+         If you're using Pro on an LTS machine (Desktop or Server) you can get
+         additional security support through
+         `Expanded Security Maintenance (ESM) <esm_>`_. We recommend it for everyone
+         using Pro, and it's enabled by default. It includes two services:
+
+         * ``esm-infra`` provides security updates for packages in the Ubuntu
```
lucasmoura, line 116: We are using the same description here as the one on the commercial tab. Maybe we can create a template for it and reuse it on each tab that we need.

I am thinking of an approach like this one:
https://github.com/canonical/ubuntu-pro-client/blob/docs/docs/tutorials/fix_intro.rst?plain=1#L7

This idea also applies to any other section were we are reusing the same description for a service

docs/explanations/which_services.rst
```diff
@@ -0,0 +1,188 @@
...
+         lightweight, it may not be suitable on a very small system with
+         limited memory. Check our `list of supported kernels`_ if you're
+         unsure.
+
+         * Find out
+           :ref:`how to enable Livepatch<manage-livepatch>`.
+
+       Want more information?
+
+    .. tab-item:: Specialty services
```
lucasmoura, line 142: I think we can also advertise Anbox here

docs/explanations/which_services.rst
s-makin: That's a really good point, but I think I'll consider that out of scope for this PR to avoid inflating it further. At some point I'll want to do a general review of every page and modularise everything that can be modularised - it will help to make it easier for other teams to sync the content, now that I know how to do it.

docs/explanations/which_services.rst
s-makin: That's a really good point, but I think that again, I'll leave this as follow-up work for now. I'm planning to invite all teams that have products represented via Pro to review what content we have and for them to modify the wording/add links to key services they want to highlight, etc. We can include Anbox as part of that.

lucasmoura (approved):

PR #3047, opened 2024-04-08
Docs apparmor update esm cache

panlinux:
## Why is this needed?
This updates the devel documentation with another confined service that landed in main: esm-cache.

The troubleshooting steps are basically the same, I just adapted the docs to not be specific to just apt-news, but added a small section on something that is specific to esm-cache which is the multiple profiles it defines.

## Test Steps
The docs have some troubleshooting steps. Feel free to exercise them to verify their correctness.

## Checklist

 - [ ] Yes
 - [ ] No


s-makin (approved):
LGTM :)

lucasmoura (approved):

PR #3049, opened 2024-04-09
Add support to esm-infra-legacy on Trusty

lucasmoura:
## Why is this needed?
This PR adds support to the new esm-infra-legacy service on Trusty.

## Test Steps
1. Launch a Trusty container
2. Set the contract url to staging in /etc/ubuntu-advantage/uaclient.conf
3. Attach by using the special token we have for testing
4. Certify that you can see `esm-infra-legacy`  on status, but the service is not enabled
5. Enable the service
6. Check apt-cache policy for the service url
7. Check that status reports the service as enabled
8. Disable the service
9. Check that esm-infra-legacy appears as disabled in status, but not esm-infra
10. Enable the service again
11. Run detach
12. Check that the machine is unattached now

## Checklist

 - [ ] Yes
 - [x] No


orndorffgrant (commented):
I've only done a basic smoke test since staging is down but looks good so far!

orndorffgrant (approved):
Thanks for the quick fixes! My tests all worked

athos-ribeiro:
Are the CI failures something we want to look at (perhaps you need to pin a different pyyaml version for those tests) ?

Anyway, this LGTM from a packaging point of view.

AFAIU, we are going to wait on an SRU pre-approval. When granted, I am going to build a source package from this branch and upload it. there will be no git-ubuntu involved here. Is this the intended path forward?

athos-ribeiro (approved):

panlinux:
I would also like to see the CI pass, or fixed. Ignoring it should be well explained, if we get to that.

panlinux:
When I follow these steps below, I end up with a system that is attached, esm-infra is enabled, esm-infra-legacy is disabled, but the gpg key for esm-infra is removed:
```
ua attach <token>

# confirm just esm-infra (not legacy) is enabled
ua status

# confirm the esm key is there
apt-key list

# confirm both esm-infra and esm-infra-legacy are enabled
ua enable esm-infra-legacy
ua status

# confirm the same esm key from before is there
apt-key list

# confirm just esm-infra is enabled, and esm-infra-legacy is now disabled
ua disable esm-infra-legacy
ua status

# the esm key is gone
apt-key list

# it warns about the missing key and that it cannot authenticate esm-infra repositories anymore
apt-get update

# confirm screen from esm is available and without the negative pin, i.e., it can be installed
apt-cache policy screen
```

I understand there is some weird interaction between `esm-infra` and `esm-infra-legacy`, but since `esm-infra-legacy` can be enabled on its own right after attach, it's weird that disabling `esm-infra-legacy` leads to the removal of the gpg key for ESM, and thus renders the ESM repository unsafe since no GPG checks are done anymore.

uaclient/entitlements/esm.py
```diff
@@ -12,3 +12,15 @@ class ESMEntitlement(repo.RepoEntitlement):
...
+
+class ESMLegacyEntitlement(repo.RepoEntitlement):
+
+    help_doc_url = "https://ubuntu.com/esm"
+    name = "esm-infra-legacy"
+    title = "Ubuntu Pro: ESM Infra (Legacy)"
+    origin = "UbuntuESM"
+    description = "Expanded Security Maintenance for Infrastructure on Legacy Instances"  # noqa
+    repo_url = "https://esm.ubuntu.com/infra-legacy/"
+    repo_key_file = "ubuntu-advantage-esm-infra-legacy-trusty.gpg"
```
panlinux, line 25: Do we also need to handle this new key file in postinst and postrm?

lucasmoura: Indeed, I have updated `postrm` for the purge scenario.

For postinst, it is not necessary as we don't want to configure "advertisment` for `esm-infra-legacy` as we do for `esm-infra`. So I am not touching that file

lucasmoura:
@panlinux I think the tests run during package build should be the same. I have looked at the output of sbuild and I am seeing:

```
python3 -m pytest                                                                                                                                                                                                  
============================= test session starts ==============================                                                                                                                                   
platform linux -- Python 3.4.3 -- pytest-2.5.1                                                                                                                                                                     
collected 487 items                                                                                                                                                                                                
                                                                                                                                                                                                                   
test-requirements.txt .                                                                                                                                                                                            
uaclient/entitlements/tests/test_base.py .................................                                                                                                                                         
uaclient/entitlements/tests/test_cc.py .......                                                                                                                                                                     
uaclient/entitlements/tests/test_cis.py ..                                                                                                                                                                         
uaclient/entitlements/tests/test_esm.py .....                                                                                                                                                                      
uaclient/entitlements/tests/test_fips.py ..........................................................................................                                                                                
uaclient/entitlements/tests/test_livepatch.py ......................................                                                                                                                               
uaclient/entitlements/tests/test_repo.py ......................                                                                                                                                                    
uaclient/tests/test_apt.py ..............................................................                                                                                                                          
uaclient/tests/test_cli.py ...........................                                                                                                                                                             
uaclient/tests/test_cli_attach.py ...........                                                                                                                                                                      
uaclient/tests/test_cli_detach.py .........                                                                                                                                                                        
uaclient/tests/test_cli_disable.py ......                                                                                                                                                                          
uaclient/tests/test_cli_enable.py .........                                                                                                                                                                        
uaclient/tests/test_cli_refresh.py ....                                                                                                                                                                            
uaclient/tests/test_cli_status.py .......                                                                                                                                                                          
uaclient/tests/test_config.py .......................................................                                                                                                                              
uaclient/tests/test_contract.py .............                                                                                                                                                                      
uaclient/tests/test_gpg.py ..                                                                                                                                                                                      
uaclient/tests/test_serviceclient.py ...                                                                                                                                                                           
uaclient/tests/test_status.py ..................                                                                                                                                                                   
uaclient/tests/test_util.py ............................................................... 
```

Which is consistent to what I see running `pytest` locally (Even though, locally, we only detect 486 test cases :upside_down_face: )

panlinux (approved):
+1 from me (pre-SRU review)

panlinux:
This may be redundant, and it's been a while since I was a sponsor for u-a-t, but we may want a PR in launchpad against the actual package branch that is in trusty (LP being the source of truth).

lucasmoura:
@panlinux I have compared the base branch here, `19.6-trsuty` with the launchpad branch `ubuntu/trusty-devel` and the diff is minimal (.gitignore and a changelog date).

Due to that, and as we are trying to start cutting all of our releases on github (as can be seen [here](https://github.com/canonical/ubuntu-pro-client/blob/docs/dev-docs/howtoguides/release_a_new_version.md)), I was under the assumption that we could use just this branch to cut the release.

But please let me know if you think that doesn't work. I can create the MP in launchpad with needed without a problem

panlinux:
> Due to that, and as we are trying to start cutting all of our releases on github (as can be seen [here](https://github.com/canonical/ubuntu-pro-client/blob/docs/dev-docs/howtoguides/release_a_new_version.md)), I was under the assumption that we could use just this branch to cut the release.

Whatever is easier for you. Just keep in mind that when reviewing what is actually uploaded to unapproved, the diff I will get will be against what is in the Ubuntu Archive, and that is what will matter, because that's what our users will get.

athos-ribeiro:
>  I have compared the base branch here, 19.6-trsuty with the launchpad branch ubuntu/trusty-devel and the diff is minimal (.gitignore and a changelog date).

I suppose we do not want to change the changelog date for an already released version, right?

panlinux:
> > I have compared the base branch here, 19.6-trsuty with the launchpad branch ubuntu/trusty-devel and the diff is minimal (.gitignore and a changelog date).
> 
> I suppose we do not want to change the changelog date for an already released version, right?

Yeah, that sounds unecessary, unless it's blatantly incorrect.

lucasmoura:
I am closing this PR as we have already released this package

PR #3050, opened 2024-04-09
docs: Adding howto guides for CVEs/USNs

dheyay:
## Why is this needed?

This PR solves all of our problems because it adds the following guides to documentation 
 -  How to resolve a given CVE/USN?
 -  How to know if a system is affected by a CVE/USN?

## Checklist

 - [ ] Yes
 - [ ] No


lucasmoura:
This looks great. The only suggestion I have is to add one extra scenario for `--dry-run` where the user would need to enable a Pro service to fix the CVE/USN

PR #3054, opened 2024-04-09
New version v32

lucasmoura:


debian/changelog
```diff
@@ -1,8 +1,66 @@
-ubuntu-advantage-tools (1:1+devel) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (32) noble; urgency=medium
```
orndorffgrant, line 1: This actually won't be released until Optical Orangutan is the new devel release, but since we don't know the real name, I'm not sure what to put here for now.

debian/changelog
```diff
@@ -1,8 +1,66 @@
...
 
-  * wip
+  * d/postinst: ensure migrations happen in correct package postinst (GH: #2982)
+  * d/apparmor:introduce new ubuntu_pro_esm_cache apparmor policy
+  * New upstream release 31 (LP: #2060732)
+    - api:
+      + u.pro.attac.token.full_token_attach.v1: add suppport for attach
```
orndorffgrant, line 7:
```suggestion
      + u.pro.attach.token.full_token_attach.v1: add suppport for attach
```

debian/changelog
```diff
@@ -1,8 +1,66 @@
...
+      - improved cli/log message for unexpected errors (GH: #2600)
+      - properly handle setting empty config values (GH: #2925)
+    - cloud-init: support ubuntu_pro user-data
+    - collect-logs: update default output file to pro_logs.tar.gz (LP: #2033313)
+    - config: create public and private config (GH: #2809)
+    - entitlements:
+      + update logic that checks if a service is enabled (LP: #2031192)
+    - fips: warn/confirm with user if enabling fips downgrades the kernel
+    - fix: warn users if ESM cache cannot be updated (GH: #2841)
+    - landscape: use exception for landscape-config failures
```
orndorffgrant, line 27: I'm not sure if this one matters much to users - I think its more of an implementation refactor
```suggestion
```

orndorffgrant (commented):
Looks good! just minor suggestions

debian/changelog
lucasmoura: @athos-ribeiro any suggestion here ?

athos-ribeiro: how about UNRELEASED for now since this is not going to be uploaded anywhere?

athos-ribeiro: OK. After talking to Lucas offline, I realize that the reasoning behid not using noble here is the concern that the 24.10 (oo) development cycle could open during the SRU process for this new release.

IMHO, let's not worry about it for now. You can keep `noble` here (we do need to push this to noble anyway).

If 24.10 opens for development __after__ this is accepted in -proposed, i suppose it will be picked up in OO. If it otherwise the oo development cycle opens __before__ this is accepted in proposed by the SRU team, then we can circle back and re-adjust (i.e., upload 32 to oo and a backport to noble).

Cc @panlinux

panlinux: OO will open up with whatever is in noble at that time i.e., binary copies. So yes, we will have u-a-t v32 in both noble and OO, and that is fine. +1 for the `noble` name.

debian/changelog
```diff
@@ -1,8 +1,65 @@
-ubuntu-advantage-tools (1:1+devel) UNRELEASED; urgency=medium
+ubuntu-advantage-tools (32) noble; urgency=medium
 
-  * wip
+  * d/postinst: ensure migrations happen in correct package postinst (GH: #2982)
+  * d/apparmor:introduce new ubuntu_pro_esm_cache apparmor policy
+  * New upstream release 32 (LP: #2060732)
+    - api:
+      + u.pro.attach.token.full_token_attach.v1: add suppport for attach
```
orndorffgrant, line 7: very minor
```suggestion
      + u.pro.attach.token.full_token_attach.v1: add support for attach
```

orndorffgrant (approved):

PR #3057, opened 2024-04-10
Add docs for new API endpoints

lucasmoura:
## Why is this needed?
Add documentation for the following API related changes:

* u.pro.detach.v1 endpoint
* u.pro.attach.token.full_token_attach.v1 endpoint
* u.pro.status.is_attached.v1 new fields in the API result

## Test Steps
Verify if the documentation page is correctly rendered and we are not missing any information related to the API changes

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/docs/references/api.rst
+++ b/docs/references/api.rst
@@ -154,6 +154,7 @@ The currently available endpoints are:
 - `u.pro.attach.auto.full_auto_attach.v1`_
 - `u.pro.attach.auto.configure_retry_service.v1`_
 - `u.pro.attach.token.full_token_attach.v1`_
+- `u.pro.detach.v1`_
 - `u.pro.security.fix.cve.execute.v1`_
 - `u.pro.security.fix.usn.execute.v1`_
 - `u.pro.security.fix.cve.plan.v1`_
@@ -712,6 +713,65 @@ a token.
 
            {
                "enabled": ["service1", "service2"],
+               "reboot_required": false
+           }
+
+u.pro.detach.v1
+============================================
+
+This endpoint allow the user to detach the machine from a Pro subscription.
+
+- Introduced in Ubuntu Pro Client Version: ``32~``
+
+.. tab-set::
+
+   .. tab-item:: Python API interaction
+      :sync: python
+
+      - Calling from Python code:
+
+        .. code-block:: python
+
+           from uaclient.api.u.pro.detach.v1 import detach
+
+           result = detach()
+
+      - Expected return object:
+
+        - ``uaclient.api.u.pro.detach.v1.DetachResult``
+
+          .. list-table::
+             :header-rows: 1
+
+             * - Field Name
+               - Type
+               - Description
+             * - ``disabled``
+               - *List[str]*
+               - The services disabled during the detach operation
+             * - ``reboot_required``
+               - *bool*
+               - True if the system requires a reboot after the detach operation
+
+      - Raised exceptions:
+
+        - ``NonRootUserError``: Raised if a non-root user executes this endpoint
+
+   .. tab-item:: CLI interaction
+      :sync: CLI
+
+      - Calling from the CLI:
+
+        .. code-block:: bash
+
+           pro api u.pro.detach.v1
+
+      - Expected attributes in JSON structure:
+
+        .. code-block:: json
+
+           {
+               "disabled": ["service1", "service2"],
                "reboot_required": false
            }
 
--- a/docs/references/api.rst
+++ b/docs/references/api.rst
@@ -153,6 +153,7 @@ The currently available endpoints are:
 - `u.pro.attach.auto.should_auto_attach.v1`_
 - `u.pro.attach.auto.full_auto_attach.v1`_
 - `u.pro.attach.auto.configure_retry_service.v1`_
+- `u.pro.attach.token.full_token_attach.v1`_
 - `u.pro.security.fix.cve.execute.v1`_
 - `u.pro.security.fix.usn.execute.v1`_
 - `u.pro.security.fix.cve.plan.v1`_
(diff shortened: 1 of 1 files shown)


docs/references/api.rst
```diff
@@ -631,6 +633,148 @@ like ``systemctl start ubuntu-advantage.service``.
...
+u.pro.attach.token.full_token_attach.v1
+============================================
+
+This endpoint allow the user to attach to a Pro subscription using
+a token.
+
+- Introduced in Ubuntu Pro Client Version: ``32~``
+- Args:
+
+  - ``token``: The token associated with a Pro subscriptionn
```
s-makin, line 646:
```suggestion
  - ``token``: The token associated with a Pro subscription
```

docs/references/api.rst
```diff
@@ -631,6 +633,148 @@ like ``systemctl start ubuntu-advantage.service``.
...
+
+             * - Field Name
+               - Type
+               - Description
+             * - ``enabled``
+               - *List[str]*
+               - The services enabled during the attach operation
+             * - ``reboot_required``
+               - *bool*
+               - If the system requires a reboot after the attach operation
```
s-makin, line 679: Presumably this is True if the system requires a reboot?

docs/references/api.rst
```diff
@@ -631,6 +633,148 @@ like ``systemctl start ubuntu-advantage.service``.
...
+   .. tab-item:: CLI interaction
+      :sync: CLI
+
+      - Calling from the CLI:
+
+        .. code-block:: bash
+
+           pro api u.pro.attach.token.full_token_attach.v1
+
+        Note that we don't need to explicit pass the token when calling the CLI command.
```
s-makin, line 694:
```suggestion
        Note that we don't need to explicitly pass the token when calling the CLI command.
```

docs/references/api.rst
```diff
@@ -631,6 +633,148 @@ like ``systemctl start ubuntu-advantage.service``.
...
+        If we define a json file (i.e. file.json) with the same attributes as the options for this endpoint:
```
s-makin, line 695:
```suggestion
        If we define a JSON file (i.e. ``file.json``) with the same attributes as the options for this endpoint:
```

docs/references/api.rst
```diff
@@ -631,6 +633,148 @@ like ``systemctl start ubuntu-advantage.service``.
...
+
+        .. code-block:: json
+
+           {
+               "token": "TOKEN",
+               "auto_enable_services": false
+           }
+
+        Then, we can call the API like this:
```
s-makin, line 704:
```suggestion
        Then we can call the API like this:
```

docs/references/api.rst
```diff
@@ -631,6 +633,148 @@ like ``systemctl start ubuntu-advantage.service``.
...
+
+             * - Field Name
+               - Type
+               - Description
+             * - ``disabled``
+               - *List[str]*
+               - The services disabled during the detach operation
+             * - ``reboot_required``
+               - *bool*
+               - If the system requires a reboot after the detach operation
```
s-makin, line 754: Should this also be "True if..."

docs/references/api.rst
```diff
@@ -2228,6 +2382,17 @@ This endpoint shows if the machine is attached to a Pro subscription.
         .. code-block:: bash
 
            pro api u.pro.status.is_attached.v1
+           
+   .. tab-item:: Explanation
+      :sync: explanation
+
+      The *contract_status* field can return 4 different states, they are:
```
s-makin, line 2389:
```suggestion
      The ``contract_status`` field can return 4 different states, they are:
```

docs/references/api.rst
```diff
@@ -2228,6 +2382,17 @@ This endpoint shows if the machine is attached to a Pro subscription.
...
+
+      * **active**: The contract is currently valid.
+      * **grace-period**: The contract is in the grace period. This means that it is expired,
+        there are still some days were the contract will be valid.
```
s-makin, line 2393:
```suggestion
      * **grace-period**: The contract is in the grace period. This means that it is expired,
        but there are still some days where the contract will be valid.
```

s-makin (commented):
Just a few nits :)

docs/references/api.rst
lucasmoura: Yes, do you think we should make this explicit ?

lucasmoura: Yes, I will make it more explicit

lucasmoura:
@s-makin updated

docs/references/api.rst
```diff
@@ -631,6 +633,148 @@ like ``systemctl start ubuntu-advantage.service``.
...
+
+             * - Field Name
+               - Type
+               - Description
+             * - ``enabled``
+               - *List[str]*
+               - The services enabled during the attach operation
+             * - ``reboot_required``
+               - *bool*
+               - True If the system requires a reboot after the attach operation
```
s-makin, line 679:
```suggestion
               - True if the system requires a reboot after the attach operation
```

docs/references/api.rst
```diff
@@ -2219,6 +2363,16 @@ This endpoint shows if the machine is attached to a Pro subscription.
              * - ``is_attached``
                - *bool*
                - If the machine is attached to a Pro subscription
+             * - ``contract_status``
+               - *str*
+               - The current contract status (active, grace-period, active-soon-to-expire, expired).
+                 Please refer to the explanation tab for a description of each state
```
s-makin, line 2369:
```suggestion
                 Please refer to the explanation tab for a description of each state.
```

s-makin (approved):
Just two little nits

lucasmoura:
@s-makin fixed

s-makin (approved):

PR #3064, opened 2024-04-16
livepatch: fix delta check for enableByDefault

lucasmoura:
### Why is this needed?
On the `process_contract_deltas` function, there is a check to see if the entitlement delta has an enableByDefault directive. However, there was a typo on when fetching that value from the delta dict. We are now fixing this typo and adding an unittest for it

## Test Steps
check the new unittest for this change

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/entitlements/tests/test_livepatch.py
+++ b/uaclient/entitlements/tests/test_livepatch.py
@@ -289,6 +289,21 @@ class TestLivepatchProcessContractDeltas:
             setup_calls = []
         assert setup_calls == m_setup_livepatch_config.call_args_list
 
+    @mock.patch(M_PATH + "LivepatchEntitlement.enable")
+    @mock.patch(M_PATH + "LivepatchEntitlement.application_status")
+    def test_livepatch_disable_and_delta_with_enable_by_default(
+        self,
+        m_application_status,
+        m_enable,
+        entitlement,
+    ):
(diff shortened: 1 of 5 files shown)


orndorffgrant (approved):
Let's let CI pass, but looks good to me

uaclient/entitlements/tests/test_livepatch.py
```diff
@@ -289,6 +289,21 @@ def test_livepatch_disable_and_setup_performed_when_resource_token_changes(
...
+        m_application_status,
+        m_enable,
+        entitlement,
+    ):
+        deltas = {"entitlement": {"obligations": {"enableByDefault": True}}}
+        m_enable.return_value = (True, None)
+        application_status = ApplicationStatus.DISABLED
+        m_application_status.return_value = (application_status, "")
+        entitlement.process_contract_deltas({}, deltas, False)
+        assert [mock.call(mock.ANY)] == m_enable.call_args_list
```
a-dubs, line 305: What does this assert actually test/confirm?

lucasmoura: It just confirm that this mocked method was actually called

dheyay (approved):
Looks good

CI failures on integration test builds - 
`xenial` -  are due to an extra package being held back and the apt news output check failing. Seems unrelated to the PR.

PR #3065, opened 2024-04-17
api: adding current status field to plan api

dheyay:
## Why is this needed?

This PR adds a field displaying the `current_status` of the cve to show if a system is currently affected by a cve/usn before the fix plan is executed.
The field is returned as a part of the plan api.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/uaclient/api/u/pro/security/fix/_common/plan/v1.py
+++ b/uaclient/api/u/pro/security/fix/_common/plan/v1.py
@@ -377,6 +377,7 @@ class FixPlanResult(DataObject):
     fields = [
         Field("title", StringDataValue),
         Field("description", StringDataValue, required=False),
+        Field("current_status", StringDataValue, required=False),
         Field("expected_status", StringDataValue),
         Field("affected_packages", data_list(StringDataValue), required=False),
         Field("plan", data_list(FixPlanStep)),
@@ -395,10 +396,12 @@ class FixPlanResult(DataObject):
         error: Optional[FixPlanError],
         additional_data: AdditionalData,
         description: Optional[str] = None,
-        affected_packages: Optional[List[str]] = None
+        affected_packages: Optional[List[str]] = None,
+        current_status: Optional[str] = None
     ):
         self.title = title
         self.description = description
+        self.current_status = current_status
         self.expected_status = expected_status
         self.affected_packages = affected_packages
         self.plan = plan
@@ -429,11 +432,13 @@ class FixPlan:
         title: str,
         description: Optional[str],
         affected_packages: Optional[List[str]] = None,
+        current_status: Optional[str] = None,
     ):
         self.order = 1
         self.title = title
         self.description = description
         self.affected_packages = affected_packages
+        self.current_status = current_status
         self.fix_steps = []  # type: List[FixPlanStep]
         self.fix_warnings = []  # type: List[FixPlanWarning]
         self.error = None  # type: Optional[FixPlanError]
@@ -506,7 +511,7 @@ class FixPlan:
     def register_additional_data(self, additional_data: Dict[str, Any]):
         self.additional_data = AdditionalData(**additional_data)
 
-    def _get_status(self) -> str:
+    def _get_expected_status(self) -> str:
         if self.error:
             return "error"
 
@@ -515,18 +520,19 @@ class FixPlan:
             and isinstance(self.fix_steps[0], FixPlanNoOpStep)
             and self.fix_steps[0].data.status == "system-not-affected"
         ):
-            return str(FixStatus.SYSTEM_NOT_AFFECTED)
+            return FixStatus.SYSTEM_NOT_AFFECTED.value.msg
         elif self.fix_warnings:
-            return str(FixStatus.SYSTEM_STILL_VULNERABLE)
+            return FixStatus.SYSTEM_STILL_VULNERABLE.value.msg
         else:
-            return str(FixStatus.SYSTEM_NON_VULNERABLE)
+            return FixStatus.SYSTEM_NON_VULNERABLE.value.msg
 
     @property
     def fix_plan(self):
         return FixPlanResult(
             title=self.title,
             description=self.description,
-            expected_status=self._get_status(),
+            expected_status=self._get_expected_status(),
+            current_status=self.current_status,
             affected_packages=self.affected_packages,
             plan=self.fix_steps,
             warnings=self.fix_warnings,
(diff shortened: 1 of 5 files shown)


uaclient/api/u/pro/security/fix/_common/plan/v1.py
```diff
@@ -521,12 +524,36 @@ def _get_status(self) -> str:
...
+        system_status = self._get_expected_status()
+        if system_status == "error":
+            return "error"
+        if (
+            len(self.fix_steps) == 1
+            and isinstance(self.fix_steps[0], FixPlanNoOpStep)
+            and (
+                self.fix_steps[0].data.status == "system-not-affected"
+                or self.fix_steps[0].data.status == "cve-already-fixed"
+                or self.fix_steps[0].data.status == "cve-fixed-by-livepatch"
```
lucasmoura, line 537: just a style suggestion, but we can update this for a `in` check:
`self.fix_steps[0].data.status in ("system-not-affected", "cve-already-fixed", "cve-fixed-by-livepatch")`

lucasmoura (approved):
Just a python style suggestion, but if you disagree, no worries at all

uaclient/api/u/pro/security/fix/_common/plan/v1.py
dheyay: Looks better, updated!

orndorffgrant (changes requested):
Let's re-use the existing enum values for `FixStatus`

More importantly, I think we should expose the current state as it is rather than try to derive the current state from the plan and expected state after the fact (that is exactly what we said was too messy/hacky to recommend to users).

So in `_generate_fix_plan` there is a check `if count == 0:`. If that is true then we know the system is not affected, so if true it can do something like `fix_plan.current_status = FixStatus.SYSTEM_NOT_AFFECTED`.

Then later, there is a `if not binary_pkgs: if src_packages:` I _think_ in that situation we'd want `fix_plan.current_status = FixStatus.SYSTEM_NON_VULNERABLE`.

It looks like we're missing logic to determine if the fixed package is already installed but we haven't rebooted yet, so I don't think we'll be able to set `fix_plan.current_status = FixStatus.SYSTEM_VULNERABLE_UNTIL_REBOOT` until we solve that - which might warrant it's own bug.

WDYT @lucasmoura ?

dheyay:
@orndorffgrant I agree, the current status would be more effective if not derived from the plan and can be set in the `_generate_fix_plan` function.

dheyay:
Updated @orndorffgrant, let me know if this works?

lucasmoura:
@orndorffgrant I thinks this is a good idea

uaclient/api/u/pro/security/fix/_common/plan/v1.py
```diff
@@ -389,6 +390,7 @@ def __init__(
         self,
         *,
         title: str,
+        current_status: Optional[str],
```
orndorffgrant, line 393: Since this is optional, lets put it down at the bottom with a default value of `None`

uaclient/api/u/pro/security/fix/_common/plan/v1.py
```diff
@@ -700,6 +707,7 @@ def _fix_plan_cve(issue_id: str, cfg: UAConfig) -> FixPlanResult:
         exceptions.SecurityAPIError,
     ) as e:
         fix_plan = get_fix_plan(title=issue_id)
+        fix_plan.current_status = "error"
```
orndorffgrant, line 710: I think current_status can be a default of `None` in error cases - then we don't need to add this to every possible error scenario - WDYT?

uaclient/api/u/pro/security/fix/_common/plan/v1.py
```diff
@@ -681,6 +687,7 @@ def _fix_plan_cve(issue_id: str, cfg: UAConfig) -> FixPlanResult:
 
     if livepatch_cve_status:
         fix_plan = get_fix_plan(title=issue_id)
+        fix_plan.current_status = str(FixStatus.SYSTEM_NON_VULNERABLE)
```
orndorffgrant, line 690: using `str` adds an extra layer of indirection - I'd rather do `FixStatus.SYSTEM_NON_VULNERABLE.msg`

uaclient/cli/tests/test_cli_fix.py
```diff
@@ -145,6 +145,7 @@ def test_cve_header(self):
         cve = FixPlanResult(
             title="CVE-2020-1472",
             description="Samba vulnerability",
+            current_status="system-not-affected",
```
orndorffgrant, line 148: I think this would be "not-affected" now (?)

orndorffgrant (commented):
Looking good! Some nits

Since #3050 will depend on this new feature, it may need to be split into a PR that can go live immediately and a separate PR that will only be published with version 33

Also you'll need to add documentation for this API endpoint change

uaclient/api/u/pro/security/fix/_common/__init__.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, value: int, msg: str):
     SYSTEM_NOT_AFFECTED = _Value(0, "not-affected")
     SYSTEM_STILL_VULNERABLE = _Value(1, "still-affected")
     SYSTEM_VULNERABLE_UNTIL_REBOOT = _Value(2, "affected-until-reboot")
+    SYSTEM_VULNERABLE = _Value(3, "affected")
```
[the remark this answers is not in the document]
dheyay: The rationale for having this as a different message was that if the user would like to see if their system is affected, a `still-affected` status could be confusing compared to a simple binary - `affected` / `not-affected`. Also, the `still-affected` message is used to convey the expected_status where it makes sense since we provide with a reason as to why the fix/plan did not resolve the cve.

But I understand if having this only adds to complexity and can be substituted by the already existing still-affected message. What do you suggest @orndorffgrant ?

uaclient/api/u/pro/security/fix/_common/plan/v1.py
dheyay: I see why `None` would be a better option. Thanks!

uaclient/cli/tests/test_cli_fix.py
dheyay: This one might've slipped through, it should be not-affected

dheyay:
Updated @orndorffgrant, will also update the documentation and create a PR for the same

orndorffgrant (approved):
One last nit then I think we're good

uaclient/api/u/pro/security/fix/_common/plan/v1.py
```diff
@@ -942,6 +958,7 @@ def _generate_fix_plan(
                 apt.update_esm_caches(cfg)
                 esm_cache_updated = True
             except Exception as e:
+                fix_plan.current_status = "error"
```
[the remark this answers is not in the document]
dheyay: Fixed @orndorffgrant

Issue #3071, opened 2024-04-18
Bug: behave check for apparmor denials isn't specific enough


orndorffgrant:
**Description of the bug**

The check can cause failures in a couple scenarios where it shouldn't:
1. if the test platform is anything other than lxd-container
2. if there is a denial related to a previous test, not the current one

**Expected behavior**

1. if the test platform is anything other than lxd-container
    - apparmor denials should be searched on the test instance rather than the test runner
2. if there is a denial related to a previous test, not the current one
    - for lxd-container runs, logs should be filtered to only look for logs related to the test container

**Current behavior**

Unrelated apparmor denials in either of those situations cause otherwise good tests to fail.

**To Reproduce**

Run multiple tests including one that causes an apparmor denial. All subsequent tests will fail as well.

**System information:**

- Ubuntu release: any
- Pro Client version: 32

PR #3072, opened 2024-04-19
tests: more specific apparmor check

orndorffgrant:
## Why is this needed?

Fixes: #3071

## Test Steps

I temporarily applied this diff https://pastebin.canonical.com/p/68W67tpDYj/
That diff _inconsistently_ causes xenial tests to cause apparmor denials.

Then I ran the following:
```
tox run -e behave -- --logging-level warning --no-junit --no-color features/_version.feature -D check_version=1:1+devel -D releases=xenial,jammy -D machine_types=lxd-container,lxd-vm
```
It might take a couple tries because it doesn't consistently trigger the apparmor denials, but eventually you can get a run where the xenial lxd-container test fails and displays apparmor DENIED logs from your host that are specific to that xenial test container. And you can get a xenial lxd-vm test fail with DENIED logs from the xenial vm. The jammy tests will always pass even if the xenial tests fail during the same run.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -18,6 +18,7 @@ import features.cloud as cloud
 from features.util import (
     BUILDER_NAME_PREFIX,
     SUT,
+    UA_TMP_DIR,
     InstallationSource,
     landscape_reject_all_pending_computers,
     process_template_vars,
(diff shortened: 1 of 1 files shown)


features/environment.py
```diff
@@ -498,22 +499,47 @@ def after_scenario(context, scenario):
...
+            )
+            sut.instance.pull_file("/var/log/syslog", syslog_dest)
+            with open(syslog_dest, "r") as syslog_fd:
+                syslog_messages = syslog_fd.readlines()
+            apparmor_denied = [
+                msg.strip()
+                for msg in syslog_messages
+                if ("DENIED" in msg and "ubuntu_pro_" in msg)
+            ]
+            return apparmor_denied
```
panlinux, line 532: This bit won't work, unfortunately. apparmor denials in the case of a lxd container will not show up in the container, but only in the host.

panlinux: In other words, the apparmor kernel messages are blcoked from viewing from inside the container, and can only be seen on the host. Try running `dmesg` inside a lxd container, even as root.

features/environment.py
```diff
@@ -498,22 +499,47 @@ def after_scenario(context, scenario):
...
-            if ("DENIED" in msg and "ubuntu_pro_" in msg)
-        ]
-    return apparmor_denied
+def _get_relevant_apparmor_logs(context):
+    if hasattr(context, "machines") and SUT in context.machines:
+        sut = context.machines[SUT]
+        if sut.cloud == "lxd-container":
+            # get apparmor DENIED messages from the host
+            with open("/var/log/syslog", "r") as syslog_fd:
+                syslog_messages = syslog_fd.readlines()
```
panlinux, line 508: In the case of a lxd container, yes, we need to check for apparmor errors on the host's /var/log/syslog.

features/environment.py
```diff
@@ -498,22 +499,47 @@ def after_scenario(context, scenario):
...
+                    and sut.instance.name in msg
+                )
+            ]
+            return apparmor_denied
+        else:
+            # get apparmor DENIED messages from the SUT
+            syslog_dest = os.path.join(
+                UA_TMP_DIR, "{}-syslog".format(sut.instance.name)
+            )
+            sut.instance.pull_file("/var/log/syslog", syslog_dest)
```
panlinux, line 524: The syslog file inside a lxd container unfortunately won't contain the apparmor denials. Nor anything else inside the container that I could find. It's always on the host in this case :(

panlinux: Ah, ok, in this case it's NOT a lxd container, so it's correct to grab it from the "inside".

features/environment.py
```diff
@@ -498,22 +499,47 @@ def after_scenario(context, scenario):
...
+            # get apparmor DENIED messages from the host
+            with open("/var/log/syslog", "r") as syslog_fd:
+                syslog_messages = syslog_fd.readlines()
+            apparmor_denied = [
+                msg.strip()
+                for msg in syslog_messages
+                if (
+                    "DENIED" in msg
+                    and "ubuntu_pro_" in msg
+                    and sut.instance.name in msg
```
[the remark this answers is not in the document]
orndorffgrant: Yep, it includes a timestamp

features/environment.py
panlinux: Ah, right, sorry for jumping to an incorrect conclusion

lucasmoura:
CI errors seem unrelated to this change

panlinux (approved):

PR #3075, opened 2024-04-22
Add legacy support for Trusty cloud package

lucasmoura:
## Why is this needed?
We are adding the same change we did on #3049 to the package that is being used on the Trusty Pro cloud images.

## Test Steps
1. Launch an Azure Pro Trusty image
2. Build the package and install it on the image
3. See that esm-infra-legacy doesn't appear on status
4. Detach the machine
5. Run ua status and check that esm-infra-legacy appears on the output
6. Change the contract url to staging
7. Attach the machine with the staging token entitled to esm-infra-legacy
8. Check that esm-infra is enabled, but not esm-infra-legacy
9. Enable esm-infra-legacy and see that it appears as enabled on ua status
10. Disable the esm-infra-legacy and check that esm-infra is still enabled

## Checklist

 - [ ] Yes
 - [x] No


orndorffgrant (approved):
My tests worked and it's essentially equivalent to the other diff, so LGTM

lucasmoura:
Closing this PR as we already have a PPA with the package

PR #3077, opened 2024-04-24
chore: organize dependencies for dev tooling 

orndorffgrant:
## Why is this needed?

- toxenvs now only install the dependencies they need
- dependencies are only stored in txt files where it makes sense
- tiobe gets its own list of dependencies in a special file
- remove unit test leak that added apparmor file to working dir as a file named "test"

## Test Steps

Run all the toxenvs and make sure they still work

## Checklist

 - [ ] Yes
 - [x] No


orndorffgrant:
I'm going to rebase this on `next-v33` and only do the minimal change to `main`

PR #3085, opened 2024-04-25
dev: tool to generate api endpoint documentation

orndorffgrant:
## Why is this needed?

Our api docs are growing, and keeping them up-to-date and consistently styled is becoming tedious. This automates a lot of the work of creating the endpoint documentation by putting the docstrings as metadata with the code itself.

@lucasmoura @dheyay LMK what you think before I apply this strategy to the other existing endpoints.

## Test Steps
Try out the new script (hopefully the docstring at the top of the file is sufficient to start playing with it)

Also review the docs PR #3182 

---


tools/apidocgen.py
```diff
@@ -0,0 +1,176 @@
...
+"""
+
+import sys
+import textwrap
+from importlib import import_module
+
+from uaclient import data_types
+
+NO_ARGS = "- This endpoint takes no arguments."
+NO_EXCEPTIONS = "- No exceptions raised by this endpoint."
```
renanrodrigo, line 14: now may be the time to fix this kinda incorrect information
this is more like "no PARTICULAR exception expected to be raised here"

because see - all endpoints may spit UbuntuProErrors if things go wrong?

orndorffgrant: Yes good catch!

lucasmoura (approved):

tools/apidocgen.py
```diff
@@ -0,0 +1,258 @@
...
+            issubclass(cls, data_types.DataObject)
+            and cls not in collection
+            and cls not in ignore
+        ):
+            collection.append(cls)
+            collect_data_object_classes(cls, collection, ignore=[])
+
+
+def print_endpoint_docs(endpoint_name):
+    module = import_module("uaclient.api." + endpoint_name)
```
lucasmoura, line 136: I think we can have `try..except`  block here just to better inform when we provide an invalid endpoint

lucasmoura (commented):
Just a tiny nit, but LGTM

lucasmoura (approved):

tools/apidocgen.py
orndorffgrant: Added!

renanrodrigo (approved):

Issue #3088, opened 2024-04-26
Bug: apt news package selector list items are ANDed together instead of ORed


orndorffgrant:
**Description of the bug**

Given the package list:
```
          ["xorg-server", "<", "2:21.1.7-3ubuntu2.6"],
          ["xwayland", "<", "2:23.2.0-1ubuntu0.4"]
```
the news message will only be displayed if both of those packages are installed and match the version comparison.

**Expected behavior**

package comparisons should be ORed together.

**Current behavior**

They are ANDed, which makes it harder to specify messages that apply to any of several package situations

**System information:**

- Pro Client version: 32

panlinux:
I verified spec US085, which defined apt-news package and architecture selectors, and it does define the package selector to be an "OR". I also checked that it wasn't just recently modified ;)

PR #3089, opened 2024-04-26
apt-news: package selectors should be ORed together

orndorffgrant:
## Why is this needed?

Sorry @lucasmoura @athos-ribeiro I noticed a difference between this new feature as implemented and the spec

Fixes: #3088

## Test Steps

The new test fails before the change and passes after this change.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/apt_messages.feature
+++ b/features/apt_messages.feature
@@ -952,6 +952,36 @@ Feature: APT Messages
       # two
       #
       """
+    # Still displayed if one package selector fails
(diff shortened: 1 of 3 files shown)


lucasmoura (approved):

athos-ribeiro (commented):
Can we include a unit test covering the new desired behavior? Ideally, the test would break without the proposed change.

orndorffgrant:
Yep good call @athos-ribeiro - just added

dheyay (approved):

athos-ribeiro (approved):

Issue #3068, opened 2024-04-18
Bug: clones not noticed right away


orndorffgrant:
**Description of the bug**

It may take several hours of a cloned machine running before it is noticed by the backend.

**Expected behavior**

Clones should be counted right away after boot.

This may involve running ua-timer on boot (and bypassing the `jobs-status.json` timestamp check), or running a separate service that only does the activity-ping on boot.

**Current behavior**

The activity ping doesn't occur until `ua-timer.service` runs and enough time has passed since the timestamp in `jobs-status.json`.

**To Reproduce**

Please include details on how to reproduce the bug.
1. Launch a VM/Container
2. attach it
3. snapshot it
4. launch a clone
5. see that `activityID` has not changed (therefore it hasn't been counted)

**System information:**

- Ubuntu release: all
- Pro Client version: 32

PR #3098, opened 2024-05-01
Bug: Add service to detect clones on boot

dheyay:
Fixes: #3068

## Why is this needed?

## Checklist

 - [ ] Yes
 - [ ] No


lucasmoura:
@dheyay @orndorffgrant I wonder if we should instead reuse the `ua-reboot-cmds` for this. Even though this service requires a marker file to exist in order to run, I think we can create that marker file whenever we attach the machine. In that sense, a clone would also carry that marker file with it. 

The downside of this approach is that we will now execute this service on every reboot of an attached machine. We could create another marker file to show that we should only run the metering part of the service, but this might be adding to much complexity to the service.

So let me know if we should further discuss this idea.

orndorffgrant:
Upon reviewing I had yet another idea: we could store jobs-status.json in /run which would make just the normal timer (with OnStartupSec=1min) execute all jobs once on startup.
We can discuss more when we have time (in madrid or after)

PR #3102, opened 2024-05-08
ci-deps: bump undici per dependabot recommendation... again

renanrodrigo:
## Why is this needed?

This PR solves all of our problems because it solves a couple low severity security issues in the GH report.
As a matter of fact, this replaces #3043 as CI will never pass there because of the secrets sharing.


orndorffgrant (approved):
Thanks!

PR #3103, opened 2024-05-08
tests: fix APT news tests for bionic

lucasmoura:
## Why is this needed?
On Bionic, the distro-info-data package version is bigger than what we have on ESM. Therefore, if we upgrade when attached, we will get the ESM version, due to the pin file. After we detach and run apt upgrade, distro-info-data will then be upgraded. Since our tests required that we have no packages to upgrade when checking the apt news messages, we are now adding an additional apt upgrade step after we detach just to get rid of this issue.

## Test Steps
Check that the APT News bionic tests are now passing

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/apt_messages.feature
+++ b/features/apt_messages.feature
@@ -79,6 +79,7 @@ Feature: APT Messages
   @uses.config.contract_token
   Scenario Outline: APT Hook advertises esm-infra on upgrade
     Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I remove support for `backports` in APT
     When I apt update
     When I apt upgrade
     When I apt autoremove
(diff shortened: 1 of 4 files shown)


features/apt_messages.feature
```diff
@@ -687,6 +690,9 @@ Feature: APT Messages
     When I apt autoremove
     When I apt install `jq`
     When I run `pro detach --assume-yes` with sudo
+    # We are doing this because ESM pin might prevent some packages to be upgraded (i.e.
+    # distro-info-data)
+    When I apt upgrade
```
orndorffgrant, line 695: That sounds like a larger issue? What exactly is happening here?

orndorffgrant (approved):

Issue #3109, opened 2024-05-13
Bug: esm-cache.service fails if `/etc/os-release` is NOT a symlink to `/usr/lib/os-release`


panlinux:
**Description of the bug**

On systems where `/etc/os-release` is an actual file instead of a symlink to `/usr/lib/os-release`, the apparmor profile `ubuntu_pro_esm_cache` will block access to it. The existing profile only allows access to `/usr/lib/os-release` (via globbing rules written in other profiles that are being included).

Since the target of the symlink is what matters, if `/etc/os-release` is a symlink to `/usr/lib/os-release` (normal/common in ubuntu systems), the existing rules allow that access. But if `/etc/os-release` is an actual file, there is no rule allowing in `ubuntu_pro_esm_cache` it to be read, and the esm-cache.service fails to start.

```
root@j-uat-2065573:~# ll /etc/os-release /usr/lib/os-release 
lrwxrwxrwx 1 root root  19 May 13 19:17 /etc/os-release -> /usr/lib/os-release
-rw-r--r-- 1 root root 386 May 13 16:41 /usr/lib/os-release
root@j-uat-2065573:~# rm /etc/os-release
root@j-uat-2065573:~# cp /usr/lib/os-release /etc/
root@j-uat-2065573:~# ll /etc/os-release /usr/lib/os-release 
-rw-r--r-- 1 root root 386 May 13 19:17 /etc/os-release
-rw-r--r-- 1 root root 386 May 13 16:41 /usr/lib/os-release
root@j-uat-2065573:~# systemctl start esm-cache.service
```
The esm-service.cache status will show this error:
```
May 13 19:17:29 j-uat-2065573 python3[3490]: ["2024-05-13T19:17:29.370", "ERROR", "ubuntupro.lib.esm_cache", "main", 17, "Error updating the cache: [Errno 13] Permission denied: '/etc/os-release'", {}]
```
And here is the corresponding apparmor log:
```
[Mon May 13 19:17:28 2024] audit: type=1400 audit(1715627849.366:49): apparmor="DENIED" operation="open" profile="ubuntu_pro_esm_cache" name="/etc/os-release" pid=3490 comm="python3" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
```

**Expected behavior**

It's not clear under which circumstances `/etc/os-release` might be a file instead of a symlink, but nevertheless reading it should be allowed by the apparmor profile.

**Current behavior**

[...]

PR #3110, opened 2024-05-13
Apparmor allow for /etc/os-release

panlinux:
## Why is this needed?

## Test Steps
Keep `sudo dmesg -wT | grep ubuntu_pro` running in a terminal (in the same VM, if testing in a VM, or in the host, if testing with a LXD container), and then run this on the system being tested (LXD or VM):
```
sudo rm /etc/os-release
sudo cp /usr/lib/os-release /etc
sudo rm -rf /var/lib/apt/periodic/*
sudo systemctl start esm-cache.service
```
there should be no apparmor DENIED message for an access to `/etc/os-release`.

Additionally, for a more surgical test, also run these:
```
sudo rm /etc/os-release
sudo cp /usr/lib/os-release /etc
sudo aa-exec -p ubuntu_pro_esm_cache cat /etc/os-release
```
On a system with the fixed apparmor profile, you should see the contents of /etc/os-release. With the bug, the last command above will return a permission denied error and dmesg will show a corresponding apparmor DENIED error.

## Checklist

 - [ ] Yes
 - [x] No

Launchpad bug: LP: #2065573
Fixes: #3109

### DIFF
--- a/debian/apparmor/ubuntu_pro_apt_news.jinja2
+++ b/debian/apparmor/ubuntu_pro_apt_news.jinja2
@@ -23,6 +23,10 @@ profile ubuntu_pro_apt_news flags=(attach_disconnected) {
   /etc/apt/** r,
   /etc/default/apport r,
   /etc/ubuntu-advantage/* r,
+  # GH: #3109
+  # Allow reading the os-release file (possibly a symlink to /usr/lib).
+  /{etc/,usr/lib/}os-release r,
+
   /usr/bin/python3.{1,}[0-9] mrix,
 {% if ubuntu_codename in ["focal"] %}
   # "import uuid" in focal triggers an uname call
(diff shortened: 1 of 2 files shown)


panlinux:
The xenial failure is an unexpected update being available:
```
  Reading package lists...
  Building dependency tree...
  Reading state information...
  Calculating upgrade...
  The following packages have been kept back:
    libcurl3-gnutls
  0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
But got:
  Reading package lists...
  Building dependency tree...
  Reading state information...
  Calculating upgrade...
  The following security update requires Ubuntu Pro with 'esm-infra' enabled:
    libcurl3-gnutls
  Learn more about Ubuntu Pro for 16.04 at https://ubuntu.com/16-04
  The following packages have been kept back:
    libcurl3-gnutls
  0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
```

lucasmoura (approved):

PR #3112, opened 2024-05-14
apport: update path for timer jobs status file

lucasmoura:
## Why is this needed?
The path for the timer jobs status file is no longer being managed by the config module. Because of that, we need to update our apport script to fetch that data from the correct place. (LP: #2065616)

## Test Steps
Run `ubuntu-bug ubuntu-advantage-tools` and check that it doesn't crash

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (approved):

athos-ribeiro (approved):

panlinux (approved):
+1

PR #3118, opened 2024-05-21
v32 verification test fixups

orndorffgrant:


renanrodrigo (approved):
lgtm - except I would say 52758de7c41e56f7f7414de509be17d574f6f908 could be instead separating the test (or maybe just part of it?) into a separate one if the output checks are complicated. This one was added as part of the apparmor efforts and by the number of denials we have been getting, the more testing there the merrier

Issue #3137, opened 2024-05-31
Bug: apparmor denies /var/lib/dpkg/arch


panlinux:
**Description of the bug**

Seen in the logs from https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/2067728:
```
May 31 18:00:17 loky kernel: audit: type=1400 audit(1717158617.647:261): apparmor="DENIED" operation="open" class="file" profile="ubuntu_pro_esm_cache//dpkg" name="/var/lib/dpkg/arch" pid=8510 comm="dpkg" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
May 31 18:00:17 loky kernel: audit: type=1400 audit(1717158617.678:262): apparmor="DENIED" operation="open" class="file" profile="ubuntu_pro_esm_cache//dpkg" name="/var/lib/dpkg/arch" pid=8511 comm="dpkg" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
May 31 18:00:17 loky kernel: audit: type=1400 audit(1717158617.680:263): apparmor="DENIED" operation="open" class="file" profile="ubuntu_pro_esm_cache//dpkg" name="/var/lib/dpkg/arch" pid=8512 comm="dpkg" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
May 31 18:00:17 loky kernel: audit: type=1400 audit(1717158617.683:264): apparmor="DENIED" operation="open" class="file" profile="ubuntu_pro_esm_cache//dpkg" name="/var/lib/dpkg/arch" pid=8513 comm="dpkg" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
May 31 18:00:18 loky kernel: audit: type=1400 audit(1717158618.556:265): apparmor="DENIED" operation="open" class="file" profile="ubuntu_pro_esm_cache//dpkg" name="/var/lib/dpkg/arch" pid=8712 comm="dpkg" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
May 31 18:00:18 loky kernel: audit: type=1400 audit(1717158618.601:266): apparmor="DENIED" operation="open" class="file" profile="ubuntu_pro_esm_cache//dpkg" name="/var/lib/dpkg/arch" pid=8714 comm="dpkg" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
[...]

PR #3139, opened 2024-06-03
Apparmor dpkg foreign arch

panlinux:
## Why is this needed?
Systems with a `/var/lib/dpkg/arch` file will trigger an apparmor DENIED log entry when the esm-cache service tries to access that file. So far we found the following commands that will attempt that:
- apt-cache policy
- dpkg --print-foreign-architectures

The `/var/lib/dpkg/arch` file can be created, probably among other scenarios, when a subarchitecture is added. For example, on amd64 systems, it's quite common to also have i386 added via the command `sudo dpkg --add-architecture i386`. That is enough to create `/var/lib/dpkg/arch` populated with both am64 and i386, and trigger this bug.

We don't know yet (TBD) the implications of this DENIED error. The output of `apt-cache policy` remains the same, attached or unattached, with and without the apparmor fix.

LP: #2067810
Fixes: #3137

-->

## Test Steps
```
touch /var/lib/dpkg/arch
aa-exec -p ubuntu_pro_esm_cache//dpkg dpkg --print-foreign-architectures
```
Or, for a more realistic scenario on amd64:
```
sudo dpkg --add-architecture i386
sudo rm -rf /var/lib/apt/periodic/*
sudo systemctl start esm-cache.service
```

In both cases, observe `dmesg -wT | grep ubuntu_pro | grep DENIED`.

We should probably add a `/var/lib/dpkg/arch` file to our tests to be sure there is nothing else that needs to access it... I might amend this PR with that, but it does not block immediate reviews and GH test runs.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/apparmor/ubuntu_pro_esm_cache.jinja2
+++ b/debian/apparmor/ubuntu_pro_esm_cache.jinja2
@@ -178,6 +178,9 @@ profile ubuntu_pro_esm_cache flags=(attach_disconnected) {
 
     /{,usr/}bin/dpkg mr,
 
+    # LP: #2067810
+    /var/lib/dpkg/** r,
+
   }
 
   profile ubuntu_distro_info flags=(attach_disconnected) {
@@ -212,6 +215,9 @@ profile ubuntu_pro_esm_cache flags=(attach_disconnected) {
     /{,usr/}lib/apt/methods/store mr,
 
     /usr/share/dpkg/** r,
+
+    # LP: #2067810
+    /var/lib/dpkg/** r,
 
     /var/lib/ubuntu-advantage/apt-esm/** rwk,
 
@@ -251,6 +257,9 @@ profile ubuntu_pro_esm_cache flags=(attach_disconnected) {
     # Note: observed only in xenial tests, but makes sense for all releases
     /etc/apt/** r,
     /var/lib/apt/** r,
+
+    # LP: #2067810
+    /var/lib/dpkg/** r,
 
   }
 
--- a/features/steps/machines.py
+++ b/features/steps/machines.py
@@ -116,6 +116,14 @@ def given_a_machine(
             context, "python3-coverage", machine_name=machine_name
         )
 
+    # trigger GH: #3137
+    when_i_run_command(
+        context,
+        "touch /var/lib/dpkg/arch",
+        "with sudo",
+        machine_name=machine_name,
+    )
+
     if cleanup:
 
         def cleanup_instance():


panlinux:
With the fix, we still see some DENIED entries in the logs for the bionic aws test, but I checked and they come from the current package that is in the image. The test soon upgrades to the package built from this PR, but by then "apt-get update" was run already and triggered apparmor. After the new package is installed, we don't see denials anymore.

I'm now running one more time WITHOUT the fix, and I expect to see many more apparmor DENIED messages.

panlinux:
As expected, we started seeing a bunch of new DENIED apparmor logs due to the fix being removed.

panlinux:
This last push has the fix in place again, and the bug trigger was moved to after the first "apt update" that is run on the fresh SUT, to see if that will get rid of the DENIED log entries we saw before pro was updated in the SUT with the fix.

panlinux:
Zero apparmor failures in these test results, so the fix and the change to trigger the bug seem correct now.

I want to still do one more test, that is to see if what breaks, if anything, without the fix, but with the trigger. To do that, I have to:
- remove the fix
- leave the trigger
- remove the apparmor check in the tests (because there WILL be apparmor messages, and I don't want those to be the reason the test failed).
-

panlinux:
Lots of server timeout errors in these results :/

panlinux:
Analyzing results:

- bionic-aws-pro: known failure `cloud-init userdata has ubuntu-advantage key`, no apparmor DENIED logs
- bionic-gcp-pro: known failure `cloud-init userdata has ubuntu-advantage key.`, no apparmor DENIED logs
- bionic-lxd: DENIED logs matching this bug, but all failures revolve around a server timeout :/
- jammy-lxd: test got stuck at step `And I disable any internet connection on the 'contracts' machine` and was killed. No failures before that
- noble-lxd:
- - step `Unattached status in a ubuntu machine` failed, pro status got a new entry "realtime-kernel"
- - step `Simulate status in a ubuntu machine` failed on this diff between expected/got:
```diff
-realtime-kernel +no        +yes       +no
+realtime-kernel  yes        yes        no
```
- xenial-lxd: DENIED logs matching this bug, but all failures revolve around a server timeout :/

panlinux:
I rebased on top of main and cleaned up the commits. One commit with the fix, and another one triggering the bug, which I added to the step that prepares a machine, so all machines will have the trigger.

renanrodrigo (approved):
LGTM

lucasmoura (approved):

PR #3143, opened 2024-06-05
Make docs and docs-devel consistent

lucasmoura:
## Why is this needed?
After the release, we are now making both docs and docs-devel branches consistent

## Test Steps
Double check if the branches are indeed aligned

## Checklist

 - [ ] Yes
 - [x] No


renanrodrigo (approved):
l g t m

PR #3144, opened 2024-06-05
realtime: auto-select raspi variant when appropriate

orndorffgrant:
## Why is this needed?

This helps users get the right realtime-kernel when running on a raspberry pi. It also informs users of what variant will be installed when it is auto-selected.

## Test Steps

Ideally test on a rpi3/4/5 and test that auto-selection works as expected (select raspi on 4/5 and generic everwhere else)
Or you can trick the rpi5 detection on a multipass vm with something like this:
```
./tools/test-in-multipass.sh noble

# now in vm
pro attach --no-auto-enable

cat > /usr/local/bin/dpkg << EOF
#!/usr/bin/bash
echo arm64
EOF
chmod +x /usr/local/bin/dpkg

sed -i 's/(proc_file_path)/("\/home\/ubuntu\/fake")/' /usr/lib/python3/dist-packages/uaclient/entitlements/realtime.py
echo "Raspberry Pi 5 Model B Rev 1.0" > /home/ubuntu/fake
```

Then `pro enable realtime-kernel` and ensure that the raspi variant is auto-selected.

Also test that raspi is not selected if the proc file is mocked but the arch is not arm64.

## Checklist

 - [ ] Yes
 - [x] No

### DIFF
--- a/debian/po/pt_BR.po
+++ b/debian/po/pt_BR.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-06-06 15:29-0400\n"
+"POT-Creation-Date: 2024-06-10 14:00-0400\n"
 "PO-Revision-Date: 2023-09-25 12:29-0400\n"
 "Last-Translator: Lucas Moura <lucas.moura@canonical.com>\n"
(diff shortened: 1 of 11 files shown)


dheyay (approved):
lgtm

dheyay (approved):

dheyay (commented):
One unit test failing, otherwise looks good

renanrodrigo (approved):
lgtm ty vm!

PR #3145, opened 2024-06-05
One time pro subscription attached to a specific release

dheyay:
## Why is this needed?

This PR introduces a change to the Ubuntu Pro Client to enforce contract-specific restrictions on Ubuntu releases for Pro attachment as per the new `onlySeries` field in the contracts which specify the LTS release a contract is tied to.

## Checklist

 - [ ] Yes
 - [ ] No

### DIFF
--- a/features/subscription_attach_restrictions.feature
+++ b/features/subscription_attach_restrictions.feature
@@ -1,0 +1,55 @@
+@uses.config.contract_token
+Feature: One time pro subscription related tests
+
+  Scenario Outline: Attach fail if subscription is restricted to release
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I set the machine token overlay to the following yaml:
+      """
+      machineTokenInfo:
+          contractInfo:
+              resourceEntitlements:
+                  - type: support
+                    affordances:
+                        onlySeries: <onlyseries>
+      """
+    When I attach `contract_token` with sudo
+    Then stderr contains substring:
+      """
+      Attaching to this contract is only allowed on the Ubuntu <onlyseries> <onlyrelease> release
+      """
+    And the machine is unattached
+
+    Examples: ubuntu release
+      | release | machine_type  | onlyseries | onlyrelease |
+      | xenial  | lxd-container | bionic     | 18.04       |
+      | bionic  | lxd-container | xenial     | 16.04       |
+      | focal   | lxd-container | noble      | 24.04       |
+      | jammy   | lxd-container | focal      | 20.04       |
+      | noble   | lxd-container | jammy      | 22.04       |
+
+  Scenario Outline: check notice visible when attached with onlySeries present
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I set the machine token overlay to the following yaml:
+      """
+      machineTokenInfo:
+          contractInfo:
+              resourceEntitlements:
+                  - type: support
+                    affordances:
+                        onlySeries: <onlyseries>
+      """
+    When I attach `contract_token` with sudo
+    Then the machine is attached
+    When I run `pro status` with sudo
+    Then stdout contains substring:
+      """
+      Limited to release: Ubuntu <onlyseries>
+      """
+
+    Examples: ubuntu release
+      | release | machine_type  | onlyseries |
+      | xenial  | lxd-container | xenial     |
+      | bionic  | lxd-container | bionic     |
+      | focal   | lxd-container | focal      |
+      | jammy   | lxd-container | jammy      |
+      | noble   | lxd-container | noble      |
--- a/features/subscription_attach_restrictions.feature
+++ b/features/subscription_attach_restrictions.feature
@@ -3,53 +3,116 @@ Feature: One time pro subscription related tests
 
   Scenario Outline: Attach fail if subscription is restricted to release
     Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
-    When I set the machine token overlay to the following yaml:
+    When I create the file `/tmp/response-overlay.json` with the following:
       """
-      machineTokenInfo:
-          contractInfo:
-              resourceEntitlements:
-                  - type: support
-                    affordances:
-                        onlySeries: <onlyseries>
+      {
+          "https://contracts.canonical.com/v1/context/machines/token": [
+          {
+            "code": 200,
+            "response": {
+              "machineTokenInfo": {
+                "contractInfo": {
+                  "resourceEntitlements": [
+                    {
+                      "type": "support",
+                      "affordances": {
+                        "onlySeries": "<onlyseries>"
+                      }
+                    }
+                  ]
+                }
+              }
+            }
+        }]
+      }
       """
-    When I attach `contract_token` with sudo
+    And I append the following on uaclient config:
+      """
+      features:
+        serviceclient_url_responses: "/tmp/response-overlay.json"
+      """
+    When I attempt to attach `contract_token` with sudo
     Then stderr contains substring:
       """
-      Attaching to this contract is only allowed on the Ubuntu <onlyseries> <onlyrelease> release
+      Attaching to this contract is only allowed on the Ubuntu <onlyrelease> (<onlyseries_codename>) release
       """
     And the machine is unattached
 
     Examples: ubuntu release
-      | release | machine_type  | onlyseries | onlyrelease |
-      | xenial  | lxd-container | bionic     | 18.04       |
-      | bionic  | lxd-container | xenial     | 16.04       |
-      | focal   | lxd-container | noble      | 24.04       |
-      | jammy   | lxd-container | focal      | 20.04       |
-      | noble   | lxd-container | jammy      | 22.04       |
+      | release | machine_type  | onlyseries | onlyrelease | onlyseries_codename |
+      | xenial  | lxd-container | bionic     | 18.04 LTS   | Bionic Beaver       |
+      | bionic  | lxd-container | xenial     | 16.04 LTS   | Xenial Xerus        |
+      | focal   | lxd-container | noble      | 24.04 LTS   | Noble Numbat        |
+      | jammy   | lxd-container | focal      | 20.04 LTS   | Focal Fossa         |
+      | noble   | lxd-container | jammy      | 22.04 LTS   | Jammy Jellyfish     |
 
-  Scenario Outline: check notice visible when attached with onlySeries present
+  Scenario Outline: Check notice visible when attached with onlySeries present
     Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
-    When I set the machine token overlay to the following yaml:
+    When I create the file `/tmp/response-overlay.json` with the following:
       """
-      machineTokenInfo:
-          contractInfo:
-              resourceEntitlements:
-                  - type: support
-                    affordances:
-                        onlySeries: <onlyseries>
+      {
+          "https://contracts.canonical.com/v1/context/machines/token": [
+          {
+            "code": 200,
+            "response": {
+              "machineTokenInfo": {
+                "accountInfo": {
+                  "name": "testName",
+                  "id": "testAccID"
+                },
+                "contractInfo": {
+                  "id": "testCID",
+                  "name": "testName",
+                  "resourceEntitlements": [
+                    {
+                      "type": "support",
+                      "affordances": {
+                        "onlySeries": "<onlyseries>"
+                      }
+                    }
+                  ]
+                },
+                "machineId": "testMID"
+              }
+            }
+        }],
+        "https://contracts.canonical.com/v1/contracts/testCID/context/machines/testMID": [
+          {
+            "code": 200,
+            "response": {
+              "activityToken": "test-activity-token",
+              "activityID": "test-activity-id",
+              "activityPingInterval": 123456789
+            }
+          }],
+          "https://contracts.canonical.com/v1/contracts/testCID/machine-activity/testMID": [
+          {
+            "code": 200,
+            "response": {
+              "activityToken": "test-activity-token",
+              "activityID": "test-activity-id",
+              "activityPingInterval": 123456789
+            }
+          }]
+      }
+      """
+    And I append the following on uaclient config:
+      """
+      features:
+        serviceclient_url_responses: "/tmp/response-overlay.json"
       """
     When I attach `contract_token` with sudo
     Then the machine is attached
     When I run `pro status` with sudo
     Then stdout contains substring:
       """
-      Limited to release: Ubuntu <onlyseries>
+      Limited to release: Ubuntu <onlyrelease> (<onlyseries_codename>)
       """
 
     Examples: ubuntu release
-      | release | machine_type  | onlyseries |
-      | xenial  | lxd-container | xenial     |
-      | bionic  | lxd-container | bionic     |
-      | focal   | lxd-container | focal      |
-      | jammy   | lxd-container | jammy      |
-      | noble   | lxd-container | noble      |
+      | release | machine_type  | onlyseries | onlyrelease | onlyseries_codename |
+      | xenial  | lxd-container | xenial     | 16.04 LTS   | Xenial Xerus        |
+      | bionic  | lxd-container | bionic     | 18.04 LTS   | Bionic Beaver       |
+      | focal   | lxd-container | focal      | 20.04 LTS   | Focal Fossa         |
+      | jammy   | lxd-container | jammy      | 22.04 LTS   | Jammy Jellyfish     |
+      | noble   | lxd-container | noble      | 24.04 LTS   | Noble Numbat        |
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -16,12 +16,16 @@ from uaclient.files.state_files import (
     timer_jobs_state_file,
 )
 from uaclient.timer.metering import metering_enabled_resources
-from uaclient.timer.update_contract_info import update_contract_info
+from uaclient.timer.update_contract_info import (
+    update_contract_info,
+    validate_release_series,
+)
 from uaclient.timer.update_messaging import update_motd_messages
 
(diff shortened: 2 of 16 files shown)


features/subscription_attach_restrictions.feature
```diff
@@ -0,0 +1,130 @@
...
+                  - type: support
+                    affordances:
+                        onlySeries: <onlyseries>
+      """
+    When I attempt to attach `contract_token` with sudo
+    Then the machine is attached
+
+    Examples: ubuntu release
+      | release | machine_type  | onlyseries |
+      | xenial  | lxd-container | none       |
```
orndorffgrant, line 20: Did I miss a spec? I don't think this value will ever be "none"

features/subscription_attach_restrictions.feature
```diff
@@ -0,0 +1,130 @@
...
+    Then stderr contains substring:
+      """
+      Attaching to this contract is only allowed on the Ubuntu <onlyseries>
+      """
+    And the machine is unattached
+    When I run `pro status` with sudo
+    Then stdout contains substring:
+      """
+      Limited to release: Ubuntu <onlyseries>
+      """
```
orndorffgrant, line 73: We don't want a notice left behind if the attach fails.

features/subscription_attach_restrictions.feature
```diff
@@ -0,0 +1,130 @@
...
+    When I attempt to attach `contract_token` with sudo
+    Then the machine is attached
+
+    Examples: ubuntu release
+      | release | machine_type  | onlyseries |
+      | xenial  | lxd-container | xenial     |
+      | bionic  | lxd-container | bionic     |
+      | focal   | lxd-container | focal      |
+      | jammy   | lxd-container | jammy      |
+      | noble   | lxd-container | noble      |
```
orndorffgrant, line 103: The test right below this one is a superset of this test so we can delete this one.

lib/timer.py
```diff
@@ -120,6 +123,11 @@ def run_interval_seconds(self, cfg: UAConfig) -> int:
     update_contract_info,
     UPDATE_CONTRACT_INFO_INTERVAL,
 )
+validate_release_series_job = TimedJob(
+    "validate_release_series",
+    validate_release_series,
+    UPDATE_CONTRACT_INFO_INTERVAL,
```
orndorffgrant, line 129: I know we want the same value, but we should give this it's own variable to hold the INTERVAL value.

uaclient/actions.py
```diff
@@ -161,6 +162,34 @@ def attach_with_token(
 
     machine_token_file.write(new_machine_token)
 
+    current_series = system.get_release_info().series
+    if current_series:
```
orndorffgrant, line 166: I don't think we need to check if `current_series` is truthy

uaclient/actions.py
```diff
@@ -161,6 +162,34 @@ def attach_with_token(
...
+        support_entitlement = (
+            machine_token_file.entitlements()
+            .get("support", {})
+            .get("entitlement")
+        )
+        if support_entitlement:
+            allowed_series = support_entitlement.get("affordances", {}).get(
+                "onlySeries"
+            )
+            if allowed_series and allowed_series != "none":
```
orndorffgrant, line 176: I don't think allowed_series will ever be the string "none" (though it will usually be `None`)

We can combine some of these checks to reduce nesting

```suggestion
        only_series = (
            machine_token_file.entitlements()
            .get("support", {})
            .get("entitlement", {})
            .get("affordances", {})
            .get("onlySeries", None)
        )
```

uaclient/actions.py
```diff
@@ -161,6 +162,34 @@ def attach_with_token(
...
+                try:
+                    allowed_release = CODENAME_TO_RELEASE.get(allowed_series)
+                except KeyError:
+                    allowed_release = ""
```
orndorffgrant, line 180: `dict.get()` will never raise a KeyError. It will return `None` if the key isn't present by default, but you can change the default value to `""`
```suggestion
                allowed_release = CODENAME_TO_RELEASE.get(allowed_series, "")
```

uaclient/actions.py
```diff
@@ -161,6 +162,34 @@ def attach_with_token(
...
+                notices.add(
+                    notices.Notice.LIMITED_TO_RELEASE,
+                    release_codename=allowed_series,
+                    release=allowed_release,
+                )
```
orndorffgrant, line 185: We only want to show this if the attach was successful

uaclient/timer/update_contract_info.py
```diff
@@ -28,3 +28,37 @@ def update_contract_info(cfg: UAConfig) -> bool:
             )
             return False
     return True
+
+
+def validate_release_series(cfg: UAConfig, show_message=False):
```
orndorffgrant, line 33: a lot of the suggestions above apply to this function as well

uaclient/timer/update_contract_info.py
```diff
@@ -28,3 +28,37 @@ def update_contract_info(cfg: UAConfig) -> bool:
...
+                            allowed_release = ""
+                        print(
+                            messages.PRO_ONLY_ALLOWED_FOR_RELEASE.format(
+                                allowed_series, allowed_release
+                            )
+                        )
+                    LOG.warning(
+                        "Pro subscription is only allowed for the release %s.",
+                        allowed_series,
+                    )
```
orndorffgrant, line 64: We need to say that we are automatically detaching and why. Something like:
```
Detaching Ubuntu Pro. Previously attached subscription was only valid for Ubuntu {release_codename} {release}.
```

uaclient/timer/update_contract_info.py
```diff
@@ -28,3 +28,37 @@ def update_contract_info(cfg: UAConfig) -> bool:
...
+            .get("support", {})
+            .get("entitlement")
+        )
+        if support_entitlement:
+            allowed_series = support_entitlement.get("affordances", {}).get(
+                "onlySeries"
+            )
+            if allowed_series and allowed_series != "none":
+                if allowed_series != current_series:
+                    machine_token_file.delete()
```
orndorffgrant, line 48: This should probably call our `detach` api function

orndorffgrant (commented):
Looks mostly correct! Mostly small modifications and optimizations suggested.

features/subscription_attach_restrictions.feature
dheyay: I may/may not have looked at the `supportLevel` field in the onlySeries contract PR when thinking this. It will never be set to "none". :sweat:

uaclient/timer/update_contract_info.py
dheyay: Updated @orndorffgrant

features/subscription_attach_restrictions.feature
```diff
@@ -0,0 +1,77 @@
...
+    When I set the machine token overlay to the following yaml:
+      """
+      machineTokenInfo:
+          contractInfo:
+              resourceEntitlements:
+                  - type: support
+                    affordances:
+                        onlySeries: <onlyseries>
+      """
+    When I attempt to attach `contract_token` with sudo
```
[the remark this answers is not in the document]
dheyay: I initially added `attempt to` during my testing of the feature to see what's going on, will do. Thanks!

dheyay:
Updated with changes @orndorffgrant, thanks again!

features/subscription_attach_restrictions.feature
```diff
@@ -0,0 +1,55 @@
...
+          contractInfo:
+              resourceEntitlements:
+                  - type: support
+                    affordances:
+                        onlySeries: <onlyseries>
+      """
+    When I attach `contract_token` with sudo
+    Then stderr contains substring:
+      """
+      Attaching to this contract is only allowed on the Ubuntu <onlyseries> <onlyrelease> release
```
renanrodrigo, line 18: In other scenarios related to services where we identify the release and series like this, we use a different format.
`<release> LTS (series_full_name)` (which is the VERSION string format in /etc/os-release)

Could we, instead of
`Ubuntu jammy 22.04`
say
`Ubuntu 22.04 LTS (Jammy Jellyfish)`
for consistency reasons?

features/subscription_attach_restrictions.feature
```diff
@@ -0,0 +1,55 @@
...
+                  - type: support
+                    affordances:
+                        onlySeries: <onlyseries>
+      """
+    When I attach `contract_token` with sudo
+    Then the machine is attached
+    When I run `pro status` with sudo
+    Then stdout contains substring:
+      """
+      Limited to release: Ubuntu <onlyseries>
```
renanrodrigo, line 46: same standardization comment as above

uaclient/messages/__init__.py
```diff
@@ -81,6 +81,11 @@ class TxtColor:
     "Finished upgrade of Ubuntu Pro service configuration"
 )
 
+PRO_ONLY_ALLOWED_FOR_RELEASE = t.gettext(
+    "Detaching Ubuntu Pro. Previously attached subscription \
+was only valid for Ubuntu {release_codename} {release} release."
```
renanrodrigo, line 86: a) same standardization comment
b) should we have an acceptance test with this flow?

uaclient/actions.py
```diff
@@ -161,6 +162,28 @@ def attach_with_token(
...
+        machine_token_file.entitlements()
+        .get("support", {})
+        .get("entitlement", {})
+        .get("affordances", {})
+        .get("onlySeries", None)
+    )
+    if only_series:
+        allowed_release = CODENAME_TO_RELEASE.get(only_series, "")
+        if only_series != current_series:
+            machine_token_file.delete()
```
renanrodrigo, line 176: We just wrote the machine token at this point. Could this flow be executed before the above `write` call?

uaclient/defaults.py
```diff
@@ -83,3 +83,12 @@
     "/etc/apparmor.d/ubuntu_pro_apt_news",
     "/etc/apparmor.d/ubuntu_pro_esm_cache",
 ]
+
+CODENAME_TO_RELEASE = {
```
renanrodrigo, line 87: Eeeehhhh I don't like how this will be maintained...
but I see this as the only reasonable way of accepting my comment about the string format :| so I can live with it.

I see Mantic on the list. I don't expect to have mantic-only contracts. Do you think this map will be used for anything non-LTS related? Otherwise, keep it LTS I guess

uaclient/timer/tests/test_update_contract_info.py
```diff
@@ -84,3 +87,61 @@ def test_contract_failure(
...
+        allowed_series = entitlements["support"]["entitlement"]["affordances"][
+            "onlySeries"
+        ]
+        fake_machine_token_file._entitlements = entitlements
+        validate_release_series(cfg=FakeConfig())
+        if allowed_series:
+            if current_series != allowed_series:
+                assert m_detach.call_count == 1
+        else:
+            assert m_detach.call_count == 0
```
renanrodrigo, line 147: I tried to make it a little simpler, only change if you agree
```suggestion
class TestValidateReleaseSeries:
    @pytest.mark.parametrize("allowed_series", (None, "bionic", "jammy"))
    @mock.patch(
        "uaclient.system.get_release_info",
        return_value=mock.MagicMock(series="jammy"),
    )
    @mock.patch(M_PATH + "detach")
    def test_validate_release_series(
        self,
        m_detach,
        m_get_release_info,
        allowed_series,
        FakeConfig,
        fake_machine_token_file,
    ):
        fake_machine_token_file._entitlements = {
            "support": {
                "entitlement": {"affordances": {"onlySeries": allowed_series}}
            }
        }

        validate_release_series(cfg=FakeConfig())
        if allowed_series:
            if allowed_series != "jammy":
                assert m_detach.call_count == 1
        else:
            assert m_detach.call_count == 0
```

renanrodrigo: wow huge suggestions look horrible in the page

uaclient/defaults.py
dheyay: Yes, I discussed with Grant and we both agree the maintenance of this is not ideal but to get the strings for the messages this seemed like one way.
To add to your comments above, if we want to standardize the strings then this would just become a dict with both the release number and the series full name.  (Also, this should be LTS only so will remove mantic)

uaclient/timer/tests/test_update_contract_info.py
dheyay: Looks much better with the change

uaclient/defaults.py
dheyay: Now that I look at it, we can use the `system.get_distro_info` function instead of having to maintain this string?

dheyay: A less maintainable way to do this would be something like the function `system.get_distro_info` but use the distro information CSV to get the release, series name and the full name. This would be easier to use and would not have the same maintenance issue as this dict. Would just be an added function to system.

Let me know if you think thats a good idea @orndorffgrant @renanrodrigo @lucasmoura

orndorffgrant: I think it's very possible that it could get reused in the future for something else without looking to closely. I think we might as well maintain it for our full set of supported releases

features/subscription_attach_restrictions.feature
orndorffgrant: Thanks, this is good - the current string was taken from the spec so it's my bad for not following our convention.

uaclient/timer/update_contract_info.py
```diff
@@ -28,3 +29,31 @@ def update_contract_info(cfg: UAConfig) -> bool:
...
+                    messages.PRO_ONLY_ALLOWED_FOR_RELEASE.format(
+                        allowed_release.release,
+                        allowed_release.series_codename,
+                    )
+                )
+            LOG.warning(
+                "Detaching Ubuntu Pro. Previously attached subscription was only valid for Ubuntu %s (%s) release.",  # noqa E501
+                allowed_release.release,
+                allowed_release.series_codename,
+            )
```
renanrodrigo, line 59: I usually would not mind that much but let's kill a noqa
```suggestion
            message = messages.PRO_ONLY_ALLOWED_FOR_RELEASE.format(
                allowed_release.release,
                allowed_release.series_codename,
            )
            if show_message:
                print(message)
            LOG.warning(message)
```

uaclient/defaults.py
orndorffgrant: Whoops I posted this reply without seeing anything other than the initial comment and we talked about this out of band - should have refreshed before posting :upside_down_face:

renanrodrigo (approved):
I vote merge on green CI
Thanks much for this one @dheyay !

uaclient/actions.py
orndorffgrant: @dheyay you moved this chunk of code but now we don't need the `delete` call right?

uaclient/timer/update_contract_info.py
```diff
@@ -28,3 +29,26 @@ def update_contract_info(cfg: UAConfig) -> bool:
...
+    current_series = system.get_release_info().series
+    only_series = (
+        machine_token_file.entitlements()
+        .get("support", {})
+        .get("entitlement", {})
+        .get("affordances", {})
+        .get("onlySeries", None)
+    )
+    if only_series:
+        if only_series != current_series:
```
orndorffgrant, line 45: let's combine these
```suggestion
    if only_series and only_series != current_series:
```

uaclient/actions.py
```diff
@@ -159,8 +159,29 @@ def attach_with_token(
...
+            raise exceptions.AttachFailureRestrictedRelease(
+                release=allowed_release.release,
+                series_codename=allowed_release.series_codename,
+            )
+        notices.add(
+            notices.Notice.LIMITED_TO_RELEASE,
+            release=allowed_release.series,
+            series_codename=allowed_release.series_codename,
+        )
```
orndorffgrant, line 182: We should remove this in the detach API function

uaclient/api/u/pro/detach/v1.py
```diff
@@ -97,6 +97,7 @@ def _detach_in_lock(cfg: UAConfig) -> DetachResult:
             else:
                 disabled.append(ent_name)
 
+    notices.remove(notices.LIMITED_TO_RELEASE)
```
orndorffgrant, line 100:
```suggestion
    notices.remove(notices.Notice.LIMITED_TO_RELEASE)
```

uaclient/actions.py
```diff
@@ -159,8 +159,29 @@ def attach_with_token(
...
+        machine_token_file.entitlements()
+        .get("support", {})
+        .get("entitlement", {})
+        .get("affordances", {})
+        .get("onlySeries", None)
+    )
+    if only_series:
+        allowed_release = system.get_distro_info(only_series)
+        if only_series != current_series:
+            machine_token_file.delete()
```
orndorffgrant, line 173: Adding a new comment so it's more visible - I don't think we need this `delete` call anymore - right?

dheyay: We cannot call the detach api when we check for the onlySeries in the attach operation (due to lock issues), so I think we might want to keep this here as a way to detach?

dheyay: While checking the behave test failure, I realized that the test failed because we pushed this chunk of code before the machineTokenFile is written to. What this does is that the contract response (in the case of the test, the response_json) is not written to, so the `machine_token_file.entitlements()` function returns an empty dict.

Possible solution is to directly check the `new_machine_token` variable for the onlySeries or move the `machine_token_file.write()` operation to where it was earlier?

renanrodrigo: @dheyay my idea when I first saw this was to get `onlySeries` from `new_machine_token`, yes. If you think it looks messy, maybe rename the variable to `contract_machine_token` or `received_machine_token` - although for me the current name explains it well and checking information there is completely fine

renanrodrigo:
I think all is good and resolved here. But given @orndorffgrant had the latest reviews, I will wait for him to push the buttons.

orndorffgrant (approved):

PR #3146, opened 2024-06-06
Merge next v33

lucasmoura:
## Why is this needed?
Now that we have released version 32, we need to merge to main all the changed we did for release 33

## Checklist
Check if we are not loosing any work in this merge, specially on the enable refactor, given the main conflict on the marge happened for that work

 - [ ] Yes
 - [ ] No


orndorffgrant (approved):
CI is not happy but we'll fix that in follow-ups. Landing to unblock rebasing other PRs on `main`.

PR #3150, opened 2024-06-11
tests: add scenario to assure custom APT configuration is respected

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it ensures custom APT configuration is respected by the pro client when done globally, whether through proper system configuration or using environment variables.

## Test Steps
CI should be green, as the new integration test runs on X/N lxd-container

## Checklist
 - [x] I have updated or added any integration tests accordingly - that is the whole point
 - [x] Changes here need to be documented, and this was done in: #3151

 - [ ] Yes
 - [x] No

### DIFF
--- a/features/security_status.feature
+++ b/features/security_status.feature
@@ -859,3 +859,105 @@ Feature: Security status command behavior
 
       Ubuntu Pro is not available for non-LTS releases\.
       """
+
+  Scenario Outline: Pass custom APT configuration to the Client for updates information
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    # Get the system up to date
+    When I apt update
+    And I apt upgrade
+    # Install older versions of packages which have alternatives in -updates and -security
+    # This will mess up a little with the system but all should be fine for the test purpose
+    And I apt install `<pkg_in_updates> <pkg_in_security>`
+    And I apt update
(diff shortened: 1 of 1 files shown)


renanrodrigo:
I chose Xenial and Noble because
- Xenial older, Noble newest
- Xenial to Jammy will have APT behaving exactly the same
- Noble carries the new deb822 config files, and we make sure it still works even there

features/security_status.feature
```diff
@@ -859,3 +859,105 @@ Feature: Security status command behavior
...
       """
+
+  Scenario Outline: Pass custom APT configuration to the Client for updates information
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    # Get the system up to date
+    When I apt update
+    And I apt upgrade
+    # Install older versions of packages which have alternatives in -updates and -security
+    # This will mess up a little with the system but all should be fine for the test purpose
+    And I apt install `<packages_to_install>`
```
orndorffgrant, line 870: Nit: please separate this into two variables for the updates update and the security update.

orndorffgrant (approved):
LGTM

features/security_status.feature
renanrodrigo: done

Issue #3126, opened 2024-05-24
Bug: `pro --help` reaches out to the contract server


panlinux:
**Description of the bug**
```
$ pro --help
Error connecting to /v1/resources: 503 <html><body><h1>503 Service Unavailable</h1>
No server is available to handle this request.
</body></html>
```
**Expected behavior**

The help output should be shown without having to reach out to the internet.

**Current behavior**

`pro --help` fails to produce help output in an offline environment, or in an online one if the contract server is failing.

**To Reproduce**

Run `pro --help` in an offline environment:
```
root@n:~# ip l set dev eth0 down
root@n:~# ping 1.1.1.1
ping: connect: Network is unreachable
root@n:~# pro --help
Failed to connect to https://contracts.canonical.com/v1/resources?architecture=amd64&kernel=6.5.0-35-generic&series=noble&virt=lxc
[Errno -3] Temporary failure in name resolution
```
**System information:**

- Ubuntu release: noble
- Pro Client version: 31.2.3, also tested 32

**Additional context**

renanrodrigo:
Hello, @panlinux , thanks for reporting this.

There is work ongoing to fix it.

As of today, `pro help` and similar commands present information about the tool AND about the Pro services.
In the UX rework we are doing, `help` will only contain information about the tool, and will not need to fetch anything from any server.

PR #3172, opened 2024-06-18
help: remove service list from help output

renanrodrigo:
## Why is this needed?

This PR solves all of our problems because we don't call the backends anymore when just running `help`.

OOOOOOH but we are missing the service list now
yes
that's the point
it is not there in the new UX design, and getting rid of it now prevents problems

Fixes: #3126

## Test Steps

Disconnect your system
Run `pro help` or `pro -h` or `pro --help` and see it works and has less content
Integration tests are smaller too

---


dheyay (approved):
Looks goood

orndorffgrant (approved):
thank you thank you this is good

lucasmoura (approved):

PR #3173, opened 2024-06-18
typing: add no_implicit_optional lint lint

aciba90:
## Why is this needed?

This PR solves all of our problems because it enforces a newer and better default for a lint check.

## Test Steps

---


renanrodrigo (approved):
@aciba90 thanks! LGTM

I would suggest changing the base branch to `next-v34` as `main` is currently frozen for release, but I think this is simple enough to make it to `main`.

Issue #3058, opened 2024-04-10
Bug: Dead code for periodically checking for contract updates


orndorffgrant:
**Description of the bug**

We have some dead code that periodically looks for contract updates. 

**Expected behavior**

We should either use the code or delete it. It may also be replaced by a future contract migration feature which would allow the backend to signal a refresh is needed via a flag in the activity ping response.

**Current behavior**

https://github.com/canonical/ubuntu-pro-client/blob/next-v33/uaclient/timer/update_contract_info.py is dead code

**Additional context**

We stopped this job because it was putting too much pressure on the backend.

PR #3174, opened 2024-06-18
Remove dead functionality from the codebase

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because dead code is dead, and should be buried in the past.

The functionality can be restored from git history in the future if we want to see it again. creepy right.

Fixes: #3058 

## Test Steps
Everything should be working normal, from tox to behave, as this is only removing unused functionality.

---


lucasmoura (approved):
:clap: :clap: 

Great job on spotting and removing this code.

I am approving, but let's wait for CI to see if we are not missing anything

orndorffgrant (approved):
Thanks this looks great

PR #3175, opened 2024-06-19
Rewrite the CLI module using OO, and organize argparse

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because this will set a new standard for the Pro Client CLI, which will be leveraged for the new UX implementation.

- Implement commands and arguments as classes
- Simplify separation of concerns, maintainability and testing
- Split the CLI into submodules
- Test help text in behave

## Test Steps

~This PR should contain only refactors.~ Yep check tox and behave, small changes to help text

---


renanrodrigo:
@lucasmoura @orndorffgrant @dheyay I want to do this to ALL commands but please take preliminary looks to make sure you agree with the structure - easier to change earlier when less commands are implemented/affected

uaclient/cli/commands.py
```diff
@@ -0,0 +1,50 @@
...
+    ):
+        self.name = name
+        self.help = help
+        self.description = description
+        self.usage = usage or USAGE_TMPL.format(name=NAME, command=name)
+        self.action = action
+        self.arguments = arguments
+
+    def __call__(self, args: argparse.Namespace, *, cfg: UAConfig, **kwargs):
+        return self.action(args, cfg)
```
orndorffgrant, line 39: Defining `__call__` and passing `self` as the action feels a bit weird to me.

1. Do we need this layer of indirection? What is the benefit of this over setting `action=self.action` during `register`?
2. If we do need this layer - could we accomplish the same via defining a method here instead of `__call__`? It could be something like:
```suggestion
    def call_action(self, args: argparse.Namespace, *, cfg: UAConfig, **kwargs):
        return self.action(args, cfg)
```
 and pass `action=self.call_action` in `register`.

orndorffgrant (commented):
:clap: :clap: :clap: Excellent - big fan

How much should we convert to this at once? Should we just go for all of it? (each command in a separate commit, though)

dheyay (approved):
Looks great, will make the CLI module readable :eyes: :balloon:

uaclient/cli/commands.py
renanrodrigo: obrazek

renanrodrigo: Jokes apart we can change it to not use the dunders, sure we do

I mean I would use more and more dunder methods instead of helpers, here and in the code overall.
I took this one particularly from a suggestion in SO discussions on how to make argparse more OOPythinic and found it to be interesting

renanrodrigo:
@orndorffgrant I want to do this to all commands, yes - maybe incrementally, simple ones in this PR, if I find something more tricky (and I will) I can send it separate for ease of review. As it is a refactor anyway, we can merge without usability changes

uaclient/cli/commands.py
orndorffgrant: :laughing: pythonicity is in the eye of the beholder I guess? :upside_down_face: 
If the extra layer is useful, I don't mind using the dunder all that much. But is it useful? It would only be useful if we wanted to do something additional in the `__call__` method besides drill the args into `self.action`? If we foresee that being something we may want to do in the future, then I'm okay with it too.

orndorffgrant:
> @orndorffgrant I want to do this to all commands, yes - maybe incrementally, simple ones in this PR, if I find something more tricky (and I will) I can send it separate for ease of review. As it is a refactor anyway, we can merge without usability changes

@renanrodrigo makes sense. Maybe can you go ahead and convert all the subcommands that have their own module into this form in this PR? The we'd only have "new" commands in this form and "old" commands in `cli/__init__.py` (and no "in-between" commands in the better but not as good separate module form)

uaclient/cli/commands.py
renanrodrigo: Right nice, I'm up to that. For now I will convert it to a regular method and pass it. If something happens, won't be a huge change.

renanrodrigo:
> @renanrodrigo makes sense. Maybe can you prioritize the subcommands that have their own module? The we'd only have "new" commands in this form and "old" commands in `cli/__init__.py` (and no "in-between" commands in the better-but-not-as-good separate module form)

Not sure - some of those have been separated exactly because they are more complicated somehow. I'll try and see what happens

uaclient/cli/commands.py
renanrodrigo: the `.action` in the tests now...
hahahaha
resolving this one.

uaclient/cli/commands.py
```diff
@@ -0,0 +1,104 @@
...
+class ProCommand:
+    def __init__(
+        self,
+        name: str,
+        help: str,
+        description: str,
+        usage: Optional[str] = None,
+        action: Callable = lambda *args, **kwargs: None,
+        preserve_description: bool = False,
+        argument_groups: List[ProArgumentGroup] = [],
```
orndorffgrant, line 81: nit (that applies in a couple places in this file): default values of mutable lists can lead to confusing behavior. If one call modifies the default value of `argument_groups`, then all calls will get that modified default value. Maybe make it `Iterable[ProArgumentGroup]` and set an empty tuple as the default value?

renanrodrigo: aha good catch I forgot about that

renanrodrigo:
@lucasmoura @orndorffgrant @dheyay this is now ready for review. Please keep in mind:
- the help text is standardized and broken - next PR will target only help outputs
- there are some workarounds that I could improve, but not remove.
- I could rewrite a lot more, but focused on standardizing and organizing argparse magic.

orndorffgrant (approved):
Approving with the understanding that we need at least the most basic form of the follow up --help PR merged. Specifically, the usage line for sub-subcommands (e.g. `pro config show`) is broken after this PR.

And a nit that doesn't block this PR: I'd like to change the help behave tests to test stdout equality rather than using a regexp. We can put "optional arguments" vs "options" into an example table variable per ubuntu release.

PR #3176, opened 2024-06-19
tests: add --force flag to ufw enable

lucasmoura:
## Why is this needed?
Sometimes, the aigapped tests are hanging when we try to disable the internet on a given machine. We believe this is being caused by unhandled prompts on the ufw enable command. We are now adding the --force flag to prevent that

## Test Steps
Guarantee that airgapped tests are now unstuck on CI

---

### DIFF
--- a/features/steps/network.py
+++ b/features/steps/network.py
@@ -73,8 +73,7 @@ def disable_internet_connection(context, machine_name=SUT):
     )
     when_i_run_command(
         context,
-        "ufw enable",
+        "ufw --force enable",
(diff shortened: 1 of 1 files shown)


a-dubs (changes requested):
Tiny typo: The "aigapped tests" in your commit body/description should be "airgapped tests", right?

lucasmoura:
Good catch @a-dubs. I have fixed the commit message

renanrodrigo (approved):
works good, thanks!

PR #3178, opened 2024-06-20
docs: add new security tutorial

s-makin:
## Why is this needed?
It will help users understand better what value Pro gives them in terms of system security.

```
doc: Add security-related tutorial

A common issue is that users struggle to understand what Pro does for them
and how they can use it to keep their system secure. This tutorial walks
through the example of an EOL/ESM system to demonstrate the use of Apps/Infra
and Livepatch, and inspecting/fixing a CVE.
```

## Review and merge
This PR also needs to be reviewed and tested by the Customer Success team before it can be merged.
The steps under `pro fix` cannot be reliably tested at present, due to the 503 issue. We should not merge until that issue is resolved by the web team.


docs/tutorials/security-with-pro.rst
```diff
@@ -0,0 +1,656 @@
...
+    update-manager-core/bionic-updates 1:18.04.12 all [upgradable from: 1:18.04.11.13]
+    update-notifier-common/bionic-updates 3.192.1.21 all [upgradable from: 3.192.1.19]
+
+One of the packages that needs updating is the ``ubuntu-advantage-tools``
+package, which provides the Pro Client, but it's good practice to always keep
+software up-to-date, so let's upgrade all of these packages in our VM:
+
+.. code-block::
+
+   $ sudo apt update && sudo apt upgrade
```
lucasmoura, line 113: Let's use `sudo apt upgrade -y` just to avoid prompting the user

docs/tutorials/security-with-pro.rst
```diff
@@ -0,0 +1,656 @@
...
+
+We might notice that there are three services that are not disabled by default:
+``esm-apps``, ``esm-infra``, and ``livepatch``. These three services are
+considered useful for everyone, and so they are enabled by default when you
+attach.
+
+Enabling ``esm-apps`` and ``esm-infra`` provides us with access to the
+repositories containing the additional security updates we saw earlier.
+
+There is also a warning beside ``livepatch`` with the following notice:
```
lucasmoura, line 280: Should we already link here to the section explaining what Livepatch is ?

docs/tutorials/security-with-pro.rst
```diff
@@ -0,0 +1,656 @@
...
+    Please upgrade the kernel with apt and reboot for continued livepatch support.
+
+This is because the kernel in the default Bionic image is older and needs to be
+upgraded before we can run Livepatch. Luckily, we can apply the ESM security
+updates **and** upgrade our kernel just by running our upgrade command once
+more:
+
+.. code-block:: bash
+
+   sudo apt update && sudo apt upgrade
```
lucasmoura, line 296: Same `-y` comment here

docs/tutorials/security-with-pro.rst
```diff
@@ -0,0 +1,656 @@
...
+
+However, there is always a period of time between a CVE or USN being reported,
+and a fix being released. In this case, we might reasonably want some way to
+find out if a vulnerability is affecting our system.
+
+In an upcoming release of the Pro Client, we will have a command that will show
+us a list of all CVEs that affect our system, and their status. For now, we can
+use the ``pro fix`` command to manually inspect and resolve both CVEs and USNs.
+
+.. note::
```
lucasmoura, line 478: I don't think we support `pro fix` on an airgapped scenario, as we need to send requests to the security API and as far as I know, there is no straightforward way to mirror that

docs/tutorials/security-with-pro.rst
```diff
@@ -0,0 +1,656 @@
...
+
+    ✔ CVE-2023-5764 is resolved.
+
+As we can see, in the version of Ansible we are running (thanks to Pro) we are
+not affected by this CVE. But what happens if we downgrade Ansible to an older
+version that **was** affected by it:
+
+.. code-block:: bash
+
+   sudo apt install ansible/bionic-updates
```
lucasmoura, line 514: Same `-y` comment here

docs/tutorials/security-with-pro.rst
s-makin: Noted, thanks for the info :)

orndorffgrant:
Still blocked on https://github.com/canonical/ubuntu-com-security-api/issues/165 because it prevents the suggested fix commands from working.

docs/tutorials/security-with-pro.rst
```diff
@@ -0,0 +1,611 @@
...
+
+    ✔ CVE-2023-5764 is resolved.
+
+As we can see, in the version of Ansible we are running (thanks to Pro) we are
+not affected by this CVE. But what happens if we downgrade Ansible to an older
+version that **was** affected by it:
+
+.. code-block:: bash
+
+   sudo apt install ansible/bionic-updates -y
```
orndorffgrant, line 472:
```suggestion
   sudo apt install ansible/bionic-updates -y --allow-downgrades
```

orndorffgrant (approved):
@s-makin The security API has been working better lately, so I went through this again. The fix commands were a bit slow but they worked! I only found one small thing needed but I think this is safe to publish now :D

docs/tutorials/security-with-pro.rst
```diff
@@ -0,0 +1,611 @@
...
+
+    Listing... Done
+    [...]
+    python3-update-manager/bionic-updates 1:18.04.12 all [upgradable from: 1:18.04.11.13]
+    shim-signed/bionic-updates 1.37~18.04.13+15.7-0ubuntu1 amd64 [upgradable from: 1.37~18.04.11+15.4-0ubuntu9]
+    ubuntu-advantage-tools/bionic-updates 32.3~18.04 amd64 [upgradable from: 27.14.4~18.04]
+    update-manager-core/bionic-updates 1:18.04.12 all [upgradable from: 1:18.04.11.13]
+    update-notifier-common/bionic-updates 3.192.1.21 all [upgradable from: 3.192.1.19]
+
+One of the packages that needs updating is the ``ubuntu-advantage-tools``
```
lucasmoura, line 99: @s-makin maybe we should starting to advertise the `ubuntu-pro-client` package here instead of the old one.

what do you think @orndorffgrant @renanrodrigo @dheyay ?

lucasmoura (approved):

docs/tutorials/security-with-pro.rst
s-makin: That's a great point, and yes - definitely we should!

PR #3182, opened 2024-06-27
docs: start using auto-generated API docs

orndorffgrant:
## Why is this needed?

This integrates api documentation as generated by #3085

The first two commits are minor fixes.
The third commit moves all auto-generated content into it's own file.
The fourth commit brings in the auto-generated content to replace the old content.

## Test Steps

In particular, look at the fourth commit. Are there any errors? Has any content gone missing?
Look at the rendered version on RTD - does it look okay?

---


docs/references/api_endpoints.txt
```diff
@@ -0,0 +1,3229 @@
+..
+   THIS DOCUMENTATION WAS AUTOMATICALLY GENERATED
+   Do not edit this document directly. Instead, edit the metadata in the api source code.
+
+Available Endpoints
```
s-makin, line 5:
```suggestion
Available endpoints
```

docs/references/api_endpoints.txt
```diff
@@ -0,0 +1,3229 @@
...
+               "should_auto_attach": false
+           }
+
+
+
+
+u.pro.attach.magic.initiate.v1
+==============================
+
+This endpoint initiates the Magic Attach flow, retrieving the User Code to
```
s-makin, line 335: I'm not sure how I feel about us referring to the "Magic Attach" process since this is an internal shorthand and not documented anywhere. I know it's not part of this PR necessarily, but it's reminded me we should address this and change to a name that reflects what the process is/is doing.

docs/references/api_endpoints.txt
```diff
@@ -0,0 +1,3229 @@
+..
+   THIS DOCUMENTATION WAS AUTOMATICALLY GENERATED
+   Do not edit this document directly. Instead, edit the metadata in the api source code.
```
s-makin, line 3: "metadata in the api source code" - where is that?

s-makin (commented):
A couple of little nits, but otherwise LGTM!

I particularly like the arguments being presented in a table now rather than the previous bullet points, it makes it much easier to differentiate them from the API/CLI interactions which are also shown in bullet points. 

As I look at the formatting on the preview of the CLI/API interaction tabs, I'm wondering if it wouldn't be better for us to treat the first level of bullet points as sort of mini-headers instead, so instead of ``* Calling from Python code:`` we do ``**Calling from Python code:**``. In this way, we would better delineate the "sections" in the tab and also make the bullet lists under these mini-headings more obvious. (the current view has the bullets being "empty", which makes them quite hard to see).

docs/references/api_endpoints.txt
orndorffgrant: This line didn't change in this PR so I'm not inclined to change it right now. Also, for someone working with the API, "Magic Attach" is the official name (we just don't call it that at the user-facing CLI level).

docs/references/api_endpoints.txt
orndorffgrant: It is spread across many files. I'll add an example that will hopefully be explanatory enough. Something like:
```
If you need to edit the documentation for `u.pro.version.v1`, you can find the api source code in `uaclient/api/u/pro/version/v1.py` on the `main` branch of our git repository.
```

orndorffgrant:
@s-makin updated!

s-makin (approved):
LGTM! Thanks for the changes :)

renanrodrigo (approved):

PR #3186, opened 2024-07-02
early wip: add generic-hwe variant

orndorffgrant:
## Why is this needed?

## Test Steps

---


orndorffgrant:
This is looking less likely and not worth keeping this tiny diff around for now

PR #3195, opened 2024-07-09
tests: log sbuild error

lucasmoura:
## Why is this needed?
We are now logging the error output of sbuild in case it fails, instead of just alerting that the command failed

## Test Steps
Make a modification that will break sbuild and check we are now logging the full output of the command

---

### DIFF
--- a/features/util.py
+++ b/features/util.py
@@ -1,4 +1,5 @@
 import datetime
+import glob
 import hashlib
 import hmac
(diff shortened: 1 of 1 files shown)


dheyay (approved):
:+1:

orndorffgrant (approved):
Thanks this is helpful!

PR #3196, opened 2024-07-10
New release: 33.1

renanrodrigo:
Fix the problems related to UTF-8 found as part of the v33 release.

---

### DIFF
--- a/features/i18n.feature
+++ b/features/i18n.feature
@@ -227,3 +227,60 @@ Feature: Pro supports multiple languages
       | xenial  | lxd-container |
       | jammy   | lxd-container |
       | mantic  | lxd-container |
+
+  @uses.config.contract_token
+  Scenario Outline: Pro client's commands run successfully in a non-utf8 locale
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    # # Change the locale
+    When I apt install `language-pack-fr`
+    And I run `locale-gen fr_FR` with sudo
+    And I run `update-locale LANG=fr_FR` with sudo
+    And I reboot the machine
+    And I run `cat /etc/default/locale` as non-root
+    Then stdout matches regexp:
+      """
+      LANG=fr_FR
+      """
+    # Attach
+    When I attach `contract_token` with sudo
+    # Refresh command
+    When I run `pro refresh` with sudo
+    Then I will see the following on stdout:
+      """
+      Successfully processed your pro configuration.
+      Successfully refreshed your subscription.
+      Successfully updated Ubuntu Pro related APT and MOTD messages.
+      """
+    # status command
+    When I run `pro status --format json` as non-root
+    Then stdout is a json matching the `ua_status` schema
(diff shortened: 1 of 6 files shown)


dheyay (approved):
:+1:

features/i18n.feature
```diff
@@ -108,25 +108,27 @@ Feature: Pro supports multiple languages
...
     When I apt install `language-pack-fr`
-    And I run `update-locale LANG=fr_FR.UTF-8` with sudo
+    And I run `locale-gen fr_FR` with sudo
+    And I run `update-locale LANG=fr_FR` with sudo
     And I reboot the machine
     And I run `cat /etc/default/locale` as non-root
-    Then stdout matches regexp:
+    Then I will see the following on stdout:
       """
-      LANG=fr_FR.UTF-8
```
lucasmoura, line 116: This test is failing due to:
```
        LANG=fr_FR                                                                                                                                                                                                 
      But got:                                                                                                                                                                                                     
        #  File generated by update-locale                                                                                                                                                                         
        LANG=fr_FR 
```

uaclient/system.py
```diff
@@ -664,8 +663,8 @@ def _subp(
         except UnboundLocalError:
             raise exceptions.ProcessExecutionError(cmd=redacted_cmd)
 
-    out_result = out.decode("utf-8") if out else ""
-    err_result = err.decode("utf-8") if err else ""
+    out_result = out.decode("utf-8", errors="ignore") if out else ""
+    err_result = err.decode("utf-8", errors="ignore") if err else ""
```
lucasmoura, line 667: There is another decode call on the `snap` module. Let's add the same logic there too

features/i18n.feature
renanrodrigo: I had to change the whole thing, see below

renanrodrigo:
@lucasmoura I had to restore the old test to what it was, and added a new test which is a subset of the existing one, but using non-utf8 french.

The reason is: we need to test translations which don't exist, and we are doing it by passing FR.UTF-8. Translations are intentionally left out if utf8 is not supported, so changing it there would make us don't test anything, and we would be regressing our tests. 

Besides that, calling commands like `apt` or `lscpu` with non-utf8 will break because `pycloudlib` is also being strict about utf-8 decode. I have [raised an issue](https://github.com/canonical/pycloudlib/issues/393) against `pycloudlib` to reflect that.

lucasmoura (approved):

renanrodrigo:
@basak that is a very fair point.

We end up piping the stdout/stderr for the subprocess calls anyway, so we are the ones processing the output almost always. Landscape is, indeed, the only exception (so far). We can have more in the future, but we will need to set `pipe_stdouterr` on those new calls if we want them to show anything directly to users in the first place.

So I added a new commit to this branch, explicitly setting the LANG and LC_ALL only in the cases where we don't have direct output. Please let me know if that is reasonable.

lucaskanashiro (approved):
LGTM.

Issue #3189, opened 2024-07-03
Bug: onlySeries release check does not need to run in the timer


orndorffgrant:
**Description of the bug**

The ubuntu release can only really fully change after a reboot. So we should only check the `onlySeries` ubuntu release requirement on reboot.

**Expected behavior**

There should not be a daily timer checking the `onlySeries` requirement. It should be checked on reboot (in a service that does not block boot and only runs when attached).

**Current behavior**

There is a daily job that runs and if `onlySeries` is an affordance on the contract, then it checks the current ubuntu release.

**System information:**

- Ubuntu release: all
- Pro Client version: 33

PR #3198, opened 2024-07-10
Check onlySeries on reboot

dheyay:
## Why is this needed?

Runs onlySeries check on machine reboot instead of a timer service

Fixes: #3189 

## Test Steps

Run the tox integration test for the `features/ubuntu_upgrade` file to check if the onlySeries check runs successfully on reboot

---


systemd/ua-reboot-cmds.service
```diff
@@ -7,7 +7,6 @@
 
 [Unit]
 Description=Ubuntu Pro reboot cmds
-ConditionPathExists=/var/lib/ubuntu-advantage/marker-reboot-cmds-required
 ConditionPathExists=/var/lib/ubuntu-advantage/private/machine-token.json
```
orndorffgrant, line 10: Something like this might be helpful
```suggestion
ConditionPathExists=|/var/lib/ubuntu-advantage/marker-reboot-cmds-required
ConditionPathExists=|/var/lib/ubuntu-advantage/marker-only-series
ConditionPathExists=/var/lib/ubuntu-advantage/private/machine-token.json
```

debian/ubuntu-pro-client.postinst
```diff
@@ -40,6 +40,30 @@ mark_reboot_cmds_as_needed() {
...
+    support_resource.get("support", {})
+    .get("affordances", {})
+    .get("onlySeries", None)
+)
+if only_series:
+    state_files.only_series_check_marker.write(only_series)
+"
+}
+
+create_onlyseries_marker_file
```
dheyay, line 66: @orndorffgrant @lucasmoura I think there is a better way to handle this in the postinst, please feel free to leave suggestions for improvement. Thanks!

dheyay (commented):
Had a little `postinst` related question

debian/ubuntu-pro-client.postinst
orndorffgrant: Aside: this needs to go into the postinst-migrations.sh file

My initial thoughts are:

I think for postinst it would be better to not rely on our MachineTokenFile abstraction. Instead, check if the machine token is there in `sh`. If it is, go into python but read it with the normal python `json` module.

Also finding the support dict can be a bit simpler:
```python
ents = dict.get("machineTokenInfo", {}).get("contractInfo", {}).get("resourceEntitlements", [])
support = next([e for e in ents if e.get("type") == "support"], {})
```
then the python script should just print the value of onlyseries and the `sh` can check if it is nonempty and `touch` the marker file.

dheyay: Updated @orndorffgrant, does this look better?

orndorffgrant:

PR #3206, opened 2024-07-18
Adjustments and tweaks to the help text

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it applies a set of unhacks to our argparse logic, tweaking the help text and preparing fo the new format.

## Test Steps
I changed unit and integration tests, but give it a go and check all help messages are good.
---

### DIFF
--- a/features/help.feature
+++ b/features/help.feature
@@ -288,7 +288,7 @@ Feature: Pro Client help text
     When I run `pro system --help` as non-root
     Then I will see the following on stdout
       """
-      usage: pro system [-h]  ...
+      usage: pro system [-h] {reboot-required} ...
 
       Output system related information related to Pro services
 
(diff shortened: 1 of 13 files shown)


renanrodrigo:
@orndorffgrant I have changed most tests to strictly check for output. The caveats are:
- the API help is very different regarding `OPTIONS` and newlines, so I kept it as a regex for sanity's sake. It looks horrible anyway.
- reformat-gherkin now hates us because of a couple spaces after "Available Commands" that argparse itself is inserting for no reason...
Let me know if those are acceptable.

renanrodrigo:
Breaks: pro <valid command> <--invalid flag> gives usage for the `pro` parser

orndorffgrant (approved):
Approving pending fixing failing integration tests

orndorffgrant:
@lucasmoura The breaks comment above should open an issue we can fix later about the issue where an invalid flag on a subcommand shows the wrong usage string. I don't think we should delay this PR or v34 on it

lucasmoura (approved):

PR #3214, opened 2024-07-24
[docs] Fix weird indentation/formatting

s-makin:
## Why is this needed?
This PR solves all of our problems because...There was an accidental indent on the page that
made the first paragraph render incorrectly.


aciba90 (approved):
LGTM, thanks!

PR #3215, opened 2024-07-24
Reorganise how-to navigation

s-makin:
## Why is this needed?
This PR solves all of our problems because...the LHS navigation menu in the docs needed to be reorganised. 

```
doc: organise how-to navigation

- Add extra drop-downs to gather grouped pages
- Add anchors to pages that had none
- Other housekeeping
```

### DIFF
--- a/docs/howtoguides/attach_index.rst
+++ b/docs/howtoguides/attach_index.rst
@@ -1,0 +1,19 @@
+Attach a machine to a subscription
+**********************************
+
+* :ref:`How to attach <attach>` shows you how to perform the attach process
+* :ref:`How to simulate the attach operation <simulate-attach>` to check a
+  specific token to see what will happen if you attach to it (before
+  attaching anything), you can see:
+* :ref:`How to attach with a config file <attach-with-config>` to pass a
+  specific token to your machine using a configuration file.
+
+.. TOC
+
+.. toctree::
+   :titlesonly:
+   :hidden:
+   
+   How to attach <how_to_attach>
+   How to simulate attaching <how_to_simulate_attach>
+   How to attach with a configuration file <how_to_attach_with_config_file>
--- a/docs/howtoguides/attach_index.rst
+++ b/docs/howtoguides/attach_index.rst
@@ -1,10 +1,10 @@
 Attach a machine to a subscription
 **********************************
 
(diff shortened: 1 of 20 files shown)


docs/howtoguides/how_to_simulate_attach.rst
```diff
@@ -31,3 +55,19 @@ automatically enabled when running an ``attach`` operation.
 
     If you want more information about the **AVAILABLE** and **ENTITLED**
     columns, please refer to :ref:`this explanation <pro-status-output>`.
+
+If there are services that will be auto-enabled that you don't want to enable,
+or you only want to enable specific services, you may want to run ``attach``
+:ref:`using a config file <attach-with-config>` instead of the usual attach
```
dheyay, line 61: Should we add an example of using the attach config file here?

dheyay (approved):
Just had a question, otherwise looks great. Thanks @s-makin!

docs/howtoguides/attach_index.rst
```diff
@@ -0,0 +1,19 @@
+Attach a machine to a subscription
+**********************************
+
+* :ref:`How to attach <attach>` shows you how to perform the attach process
+* :ref:`How to simulate the attach operation <simulate-attach>` to check a
+  specific token to see what will happen if you attach to it (before
+  attaching anything), you can see:
```
orndorffgrant, line 7: is this "you can see" supposed to be there? It is confusing me
```suggestion
* :ref:`How to simulate the attach operation <simulate-attach>` to check a
  specific token to see what will happen if you attach to it (before
  attaching anything)
```

orndorffgrant: Also I think all bullets are supposed to end with a `.` if one of them does, right?

orndorffgrant (commented):
One part confused me but the rest looks great!

docs/howtoguides/how_to_simulate_attach.rst
s-makin: On reflection I'd say probably not here, since it's already covered in the "attach with a config file" page (which has an example config file and the command needed to attach)

docs/howtoguides/attach_index.rst
s-makin: Right on both counts! Thanks for spotting them

Issue #2894, opened 2023-12-13
Question: what data from the system is sent to canonical servers and is any of it identifiable or tied to my identity?


orndorffgrant:
**What is the subject you need more information on?**

what data from the system is sent to canonical servers and is any of it identifiable or tied to my identity?

**Did you check the documentation first?**

After looking for guidance in [the Pro Client documentation](https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/), mark one of the following checkboxes:

 - [ ] There is documentation about this, but it needs improvement
 - [x] There is no documentation about this

orndorffgrant:
We're working on this. It may not end up in our docs, but we will at least point to a public page with an explanation.

PR #3216, opened 2024-07-24
docs: describe all data collection

orndorffgrant:
Fixes: #2894

### DIFF
--- a/docs/explanations/data_collection.rst
+++ b/docs/explanations/data_collection.rst
@@ -18,8 +18,8 @@ These data elements are collected to ensure machines that are attached to a
 particular Ubuntu Pro contract are compliant with the terms of that particular
 contract.
 
-Data sent in order to provide service
-=====================================
+Data sent to provide service
+============================
 
 The following data is not purposefully collected, but is sent to Canonical
 servers in order to provide Ubuntu Pro services.
--- a/docs/explanations/data_collection.rst
+++ b/docs/explanations/data_collection.rst
@@ -11,8 +11,8 @@ This document categorises data collection by method of collection.
 APT package downloads
 =====================
 
-If you have any of the following services enabled, then the following data
-collection method will be in use whenever downloading packages for one of
+If you have any of the following services enabled, then the data collection
+method described below will be in use whenever downloading packages for one of
 these services.
 
 - ``anbox-cloud``
@@ -40,17 +40,17 @@ a request that looks something like this:
 
    https://bearer:$resource_token@esm.ubuntu.com/apps/ubuntu/pool/main/h/hello/hello_2.10-2ubuntu4+esm1_amd64.deb
 
-This request is necessary to download the Pro update and it necessarily
-includes the following data.
+This request is necessary to download the Pro update and includes the
+following data.
 
 - Ubuntu codename (e.g. "Jammy")
 - Package name (e.g. "hello")
 - Package version (e.g. "2.10-2ubuntu4+esm1")
 - Package architecture (e.g. "amd64")
 
-Because this request is necessarily authenticated and the authentication token
-is tied to a particular Ubuntu Pro subscription, this data is inherently tied
-to the Ubuntu Pro subscription that authenticated access to the package.
+Because this request needs to be authenticated and the authentication token is
+tied to a particular Ubuntu Pro subscription, this data is inherently tied to
+the Ubuntu Pro subscription that authenticated access to the package.
 
 Livepatch downloads
 ===================
@@ -61,16 +61,18 @@ download the correct kernel patches:
 - Kernel version (e.g. "6.8.0-38.38-generic")
 - Machine architecture (e.g. "amd64")
 
-Because this request is necessarily authenticated and the authentication token
-is tied to a particular Ubuntu Pro subscription, this data is inherently tied
-to the Ubuntu Pro subscription that authenticated access to the package.
+Similarly to APT package downloads, because this request needs to be
+authenticated and the authentication token is tied to a particular Ubuntu Pro
+subscription, this data is inherently tied to the Ubuntu Pro subscription that
+authenticated access to the package.
 
 
 Machine activity checks
 =======================
 
 Regardless of which services you have enabled, if a machine is attached to an
-Ubuntu Pro subscription, the following data is collected and updated regularly.
+Ubuntu Pro subscription, the following data is collected and updated regularly
+(default: every 6 hours).
 
 - Distribution (e.g. "Ubuntu")
 - Release codename (e.g. "Noble")
--- a/docs/explanations/data_collection.rst
+++ b/docs/explanations/data_collection.rst
@@ -3,12 +3,12 @@ What data does Canonical collect from Ubuntu Pro machines?
 
 For the purposes of delivering Ubuntu Pro services to customers in compliance
 with the terms of the Ubuntu Pro subscription, some system data is sent to
-Canonical servers. This data is sent via a couple different methods, depending
+Canonical servers. This data is sent via a few different methods, depending
 on the service and the purpose of that particular data element.
 
-This document categorizes data collection by method of collection.
+This document categorises data collection by method of collection.
 
-APT Package Downloads
+APT package downloads
 =====================
 
 If you have any of the following services enabled, then the following data
@@ -28,58 +28,59 @@ these services.
 - ``ros-updates``
 - ``usg``
 
-Whenever you ``apt install`` a package from a pro service (or ``apt upgrade``
-to a version of a package from a pro service), ``apt`` will make a GET request
-to ``esm.ubuntu.com`` that includes the package name and version and HTTP
-basic auth credentials that are tied to the Ubuntu Pro Subscription.
+Whenever you ``apt install`` a package from a Pro service (or ``apt upgrade``
+to a version of a package from a Pro service), ``apt`` will make a GET request
+to ``esm.ubuntu.com`` that includes the package name and version, and HTTP
+basic auth credentials that are tied to the Ubuntu Pro subscription.
 
 For example, installing the ``hello`` package from ``esm-apps`` will result in
 a request that looks something like this:
 
 .. code-block:: text
+
    https://bearer:$resource_token@esm.ubuntu.com/apps/ubuntu/pool/main/h/hello/hello_2.10-2ubuntu4+esm1_amd64.deb
 
 This request is necessary to download the Pro update and it necessarily
 includes the following data.
 
-- ubuntu codename (e.g. "jammy")
-- package name (e.g. "hello")
-- package version (e.g. "2.10-2ubuntu4+esm1")
-- package architecture (e.g. "amd64")
+- Ubuntu codename (e.g. "Jammy")
+- Package name (e.g. "hello")
+- Package version (e.g. "2.10-2ubuntu4+esm1")
+- Package architecture (e.g. "amd64")
 
 Because this request is necessarily authenticated and the authentication token
 is tied to a particular Ubuntu Pro subscription, this data is inherently tied
 to the Ubuntu Pro subscription that authenticated access to the package.
 
-Livepatch Downloads
+Livepatch downloads
 ===================
 
 If you have ``livepatch`` enabled, then the following data is sent in order to
 download the correct kernel patches:
 
-- kernel version (e.g. "6.8.0-38.38-generic")
-- machine architecture (e.g. "amd64")
+- Kernel version (e.g. "6.8.0-38.38-generic")
+- Machine architecture (e.g. "amd64")
 
 Because this request is necessarily authenticated and the authentication token
 is tied to a particular Ubuntu Pro subscription, this data is inherently tied
 to the Ubuntu Pro subscription that authenticated access to the package.
 
 
-Machine Activity Checks
+Machine activity checks
 =======================
 
 Regardless of which services you have enabled, if a machine is attached to an
 Ubuntu Pro subscription, the following data is collected and updated regularly.
 
-- distribution (e.g. "Ubuntu")
-- release codename (e.g. "noble")
-- kernel version (e.g. "6.8.0-38.38-generic")
-- machine architecture (e.g. "amd64")
-- is the machine a desktop (e.g. "true")
-- virtualization type (e.g. "docker")
-- services enabled (e.g. "ros" and "realtime-kernel generic variant")
-- time the machine was attached (e.g. "2024-07-24T13:54:07+00:00")
-- version of ``ubuntu-pro-client`` (e.g. "33.2~24.04")
+- Distribution (e.g. "Ubuntu")
+- Release codename (e.g. "Noble")
+- Kernel version (e.g. "6.8.0-38.38-generic")
+- Machine architecture (e.g. "amd64")
+- Is the machine a desktop? (e.g. "true")
+- Virtualisation type (e.g. "Docker")
+- Services enabled (e.g. "ros" and "realtime-kernel generic variant")
+- When the machine was attached (e.g. "2024-07-24T13:54:07+00:00")
+- Version of ``ubuntu-pro-client`` (e.g. "33.2~24.04")
 
 These data elements are collected to ensure machines that are attached to a
 particular Ubuntu Pro contract are compliant with the terms of that particular
--- a/docs/explanations/data_collection.rst
+++ b/docs/explanations/data_collection.rst
@@ -1,15 +1,31 @@
-What data does Canonical collect from Ubuntu Pro machines?
-**********************************************************
+What data is collected from active Ubuntu Pro machines?
+*******************************************************
 
-Some system data is sent to Canonical servers for the purpose of delivering
-Ubuntu Pro services in compliance with the terms of the Ubuntu Pro subscriptio
-. This data is sent via a few different methods, depending on the service and
-the purpose of that particular data element.
+If a machine is attached to an Ubuntu Pro subscription, the following data is
+collected and updated regularly (default: every 6 hours).
 
-This document categorises data collection by method of collection.
+- Distribution (e.g. "Ubuntu")
+- Release codename (e.g. "Noble")
+- Kernel version (e.g. "6.8.0-38.38-generic")
+- Machine architecture (e.g. "amd64")
+- Is the machine a desktop? (e.g. "true")
+- Virtualisation type (e.g. "Docker")
+- Services enabled (e.g. "ros" and "realtime-kernel generic variant")
+- When the machine was attached (e.g. "2024-07-24T13:54:07+00:00")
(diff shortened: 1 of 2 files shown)


docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
+What data does Canonical collect from Ubuntu Pro machines?
```
s-makin, line 1: I'd probably suggest changing this header to "what data does Canonical collect through the Ubuntu Pro Client"

The reason being that the contents of the page are relevant to people who haven't attached yet (and want the info to know if they want to attach), or who have detached and aren't using Pro anymore (can we consider their machine an Ubuntu Pro machine?)

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+to ``esm.ubuntu.com`` that includes the package name and version and HTTP
+basic auth credentials that are tied to the Ubuntu Pro Subscription.
+
+For example, installing the ``hello`` package from ``esm-apps`` will result in
+a request that looks something like this:
+
+.. code-block:: text
+   https://bearer:$resource_token@esm.ubuntu.com/apps/ubuntu/pool/main/h/hello/hello_2.10-2ubuntu4+esm1_amd64.deb
+
+This request is necessary to download the Pro update and it necessarily
```
a-dubs, line 42: the "necessarily" after already saying "necessary" early in the sentence feels _unnecessary_ (lol). But beyond that, it also makes the sentence super clunky. I'd suggest rectifying this by dropping the "it necessarily" so that the sentence reads:
"This request is necessary to download the Pro update and includes the following data."

a-dubs (changes requested):
left some suggested changes! thanks grant!

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+Because this request is necessarily authenticated and the authentication token
+is tied to a particular Ubuntu Pro subscription, this data is inherently tied
+to the Ubuntu Pro subscription that authenticated access to the package.
+
+
+Machine Activity Checks
+=======================
+
+Regardless of which services you have enabled, if a machine is attached to an
+Ubuntu Pro subscription, the following data is collected and updated regularly.
```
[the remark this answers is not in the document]
orndorffgrant: It is configurable but I'll state the default

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
+What data does Canonical collect from Ubuntu Pro machines?
+**********************************************************
+
+For the purposes of delivering Ubuntu Pro services to customers in compliance
+with the terms of the Ubuntu Pro subscription, some system data is sent to
+Canonical servers. This data is sent via a couple different methods, depending
```
s-makin, line 6: I'd flip the order of this around to make what we're doing explicitly stated at the start, i.e.:
"Some system data is sent to Canonical servers for the purpose of delivering Ubuntu Pro services in compliance with the terms of the Ubuntu Pro subscription."

s-makin, line 6:
```suggestion
For the purposes of delivering Ubuntu Pro services to customers in compliance
with the terms of the Ubuntu Pro subscription, some system data is sent to
Canonical servers. This data is sent via a few different methods, depending
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+on the service and the purpose of that particular data element.
+
+This document categorizes data collection by method of collection.
```
s-makin, line 9:
```suggestion
This document categorises data collection by method of collection.
```

s-makin: (sorry!)

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+
+APT Package Downloads
```
s-makin, line 11:
```suggestion
APT package downloads
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+=====================
+
+If you have any of the following services enabled, then the following data
+collection method will be in use whenever downloading packages for one of
+these services.
```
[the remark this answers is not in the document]
s-makin: I agree this is better - what follows is not a description of the methods, but the list of services, so the follows should be attached to the list of services.

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- ``esm-infra``
+- ``fips``
+- ``fips-preview``
+- ``fips-updates``
+- ``realtime-kernel``
+- ``ros``
+- ``ros-updates``
+- ``usg``
+
+Whenever you ``apt install`` a package from a pro service (or ``apt upgrade``
```
s-makin, line 31:
```suggestion
Whenever you ``apt install`` a package from a Pro service (or ``apt upgrade``
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+to a version of a package from a pro service), ``apt`` will make a GET request
```
s-makin, line 32:
```suggestion
to a version of a package from a Pro service), ``apt`` will make a GET request
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+to ``esm.ubuntu.com`` that includes the package name and version and HTTP
```
s-makin, line 33:
```suggestion
to ``esm.ubuntu.com`` that includes the package name and version, and HTTP
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+basic auth credentials that are tied to the Ubuntu Pro Subscription.
```
s-makin, line 34:
```suggestion
basic auth credentials that are tied to the Ubuntu Pro subscription.
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+
+For example, installing the ``hello`` package from ``esm-apps`` will result in
+a request that looks something like this:
+
+.. code-block:: text
+   https://bearer:$resource_token@esm.ubuntu.com/apps/ubuntu/pool/main/h/hello/hello_2.10-2ubuntu4+esm1_amd64.deb
```
s-makin, line 40:
```suggestion
.. code-block:: text

   https://bearer:$resource_token@esm.ubuntu.com/apps/ubuntu/pool/main/h/hello/hello_2.10-2ubuntu4+esm1_amd64.deb
```

s-makin: This is causing the preview to fail to build

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+
+This request is necessary to download the Pro update and it necessarily
+includes the following data.
+
+- ubuntu codename (e.g. "jammy")
```
s-makin, line 45:
```suggestion
- Ubuntu codename (e.g. "Jammy")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- package name (e.g. "hello")
```
s-makin, line 46:
```suggestion
- Package name (e.g. "hello")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- package version (e.g. "2.10-2ubuntu4+esm1")
```
s-makin, line 47:
```suggestion
- Package version (e.g. "2.10-2ubuntu4+esm1")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- package architecture (e.g. "amd64")
```
s-makin, line 48:
```suggestion
- Package architecture (e.g. "amd64")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Because this request is necessarily authenticated and the authentication token
```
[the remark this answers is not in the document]
s-makin: Should be "needs to be authenticated," or "needs authentication,"

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+is tied to a particular Ubuntu Pro subscription, this data is inherently tied
+to the Ubuntu Pro subscription that authenticated access to the package.
+
+Livepatch Downloads
```
s-makin, line 54:
```suggestion
Livepatch downloads
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+===================
+
+If you have ``livepatch`` enabled, then the following data is sent in order to
+download the correct kernel patches:
+
+- kernel version (e.g. "6.8.0-38.38-generic")
```
s-makin, line 60:
```suggestion
- Kernel version (e.g. "6.8.0-38.38-generic")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- machine architecture (e.g. "amd64")
```
s-makin, line 61:
```suggestion
- Machine architecture (e.g. "amd64")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Because this request is necessarily authenticated and the authentication token
```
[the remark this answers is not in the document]
s-makin: Same as above

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+is tied to a particular Ubuntu Pro subscription, this data is inherently tied
+to the Ubuntu Pro subscription that authenticated access to the package.
```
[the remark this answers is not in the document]
s-makin: I would suggest the same

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+
+
+Machine Activity Checks
```
s-makin, line 68:
```suggestion
Machine activity checks
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+=======================
+
+Regardless of which services you have enabled, if a machine is attached to an
```
s-makin, line 71: I would reorder this as well:
"If your machine is attached to an Ubuntu Pro subscription, the following data is collected and updated regularly (regardless of which services are enabled or disabled).

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+Ubuntu Pro subscription, the following data is collected and updated regularly.
+
+- distribution (e.g. "Ubuntu")
```
s-makin, line 74:
```suggestion
- Distribution (e.g. "Ubuntu")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- release codename (e.g. "noble")
```
s-makin, line 75:
```suggestion
- Release codename (e.g. "Noble")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- kernel version (e.g. "6.8.0-38.38-generic")
```
s-makin, line 76:
```suggestion
- Kernel version (e.g. "6.8.0-38.38-generic")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- machine architecture (e.g. "amd64")
```
s-makin, line 77:
```suggestion
- Machine architecture (e.g. "amd64")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- is the machine a desktop (e.g. "true")
```
s-makin, line 78:
```suggestion
- Is the machine a desktop? (e.g. "true")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- virtualization type (e.g. "docker")
```
s-makin, line 79:
```suggestion
- Virtualisation type (e.g. "Docker")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- services enabled (e.g. "ros" and "realtime-kernel generic variant")
```
s-makin, line 80:
```suggestion
- Services enabled (e.g. "ros" and "realtime-kernel generic variant")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- time the machine was attached (e.g. "2024-07-24T13:54:07+00:00")
```
s-makin, line 81:
```suggestion
- When the machine was attached (e.g. "2024-07-24T13:54:07+00:00")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+- version of ``ubuntu-pro-client`` (e.g. "33.2~24.04")
```
s-makin, line 82:
```suggestion
- Version of ``ubuntu-pro-client`` (e.g. "33.2~24.04")
```

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,86 @@
...
+
+These data elements are collected to ensure machines that are attached to a
+particular Ubuntu Pro contract are compliant with the terms of that particular
+contract.
```
s-makin, line 86: Could any of this info/data be considered as personally identifiable?

Do we know roughly how long is data kept for?

s-makin (commented):
Looks good! Some suggestions

docs/explanations/data_collection.rst
orndorffgrant: Two of the sections are data not collected directly through the Ubuntu Pro Client, but the collection happens because you used the pro-client to attach to Pro. So I'm not sure "collect through the Ubuntu Pro Client" is quite accurate.

A detached machine shouldn't be considered an Ubuntu Pro machine - they are in the same bucket as never attached machines for the purpose of this doc.
None of this data is collected for unattached machines. And I don't think the current title would prevent someone wondering about data collection from looking at it. We could rename it to "What data does Canonical collect from Ubuntu machines that are attached to an Ubuntu Pro subscription" but that felt unnecessarily verbose

IDK though, I don't have a strong opinion on the title here, just wanted to list my thoughts before changing it. What do you think?

s-makin: I think that depends on how long we keep data for. Is machine data purged when the machine is detached?

orndorffgrant: Data is not purged on detach - so a detached machine would have had data collected while it was Ubuntu Pro, and after it is detached, no more data will be collected, but data will exist on the backend for some time.

docs/explanations/data_collection.rst
orndorffgrant: All of the document only applies if the machine is attached (ignoring the caveat of data retention for now), so this sentence was supposed to emphasize the irrelevance of which services are enabled for this section. The other sections open with which services need to be enabled for the section to be relevant, so I wanted to do the same here.

I'm happy to change it, but want to check if you would edit your suggestion given that context?

docs/explanations/data_collection.rst
orndorffgrant: I don't _think_ any of this counts as personally identifiable on it's own, but it is connected to an Ubuntu Pro account on the backend via a machine id.

> Do we know roughly how long is data kept for?

I don't know the answer to this one. Tagging @pandrey2003 and @alnvdl-work

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,89 @@
+What data does Canonical collect from Ubuntu Pro machines?
+**********************************************************
+
+Some system data is sent to Canonical servers for the purpose of delivering
+Ubuntu Pro services in compliance with the terms of the Ubuntu Pro subscriptio
```
lucasmoura, line 5: s/subscriptio/subscription

docs/explanations/data_collection.rst
```diff
@@ -0,0 +1,89 @@
...
+. This data is sent via a few different methods, depending on the service and
+the purpose of that particular data element.
+
+This document categorises data collection by method of collection.
+
+APT package downloads
```
lucasmoura, line 11: I can be wrong here, but it seems that the `APT packages downloads` and `Livepatch downloads` are not directly tied to data collection per se. It seems more like `data used per service` than a collection of some sorts.

I think there is still value for those sections, but I would not put them under `data collection`. We could create something like `service data needs` for them

s-makin: I think that's a good idea :)

docs/explanations/data_collection.rst
s-makin: If data isn't purged on detach, then I think there is a difference between detached and never-attached machines since users who care about this stuff want to know what we collect, why, and how long we expect to keep that data for.

I think tweaking the title to say "What data is collected from active Ubuntu Pro machines?" would be enough to satisfy the distinction, especially if we can also provide info on how long it takes before collected info is purged (although I wouldn't consider that a blocker).

docs/explanations/data_collection.rst
s-makin: That subtlety passed me by completely lol, but it makes sense and I'm happy for you leave this as it is.

docs/explanations/data_collection.rst
s-makin: I won't consider this as a blocker to getting this merged. We can always add a section later at the bottom here about data retention.

docs/explanations/data_collection.rst
orndorffgrant: Yeah that's a good point, this is data that is sent to canonical servers for the purposes of using the services. I'll rework the structure of this a bit to make that more clear.

docs/explanations/data_collection.rst
orndorffgrant: whoops :upside_down_face:

docs/explanations/data_collection.rst
orndorffgrant: @s-makin I've restructured the document and headers around this. All the content is the same, but I'm not sure about the new header "Data sent in order to provide service" - do you have any better ideas?

PR #3219, opened 2024-07-24
Remove docs for format option

lucasmoura:
## Why is this needed?
We are removing any mention to either `pro security-status --format` and `pro status --format` from our docs.
That is because we don't want users to rely on that information, but instead on the APIs.

For `security-status`, we can completely replace the use `--format` for the APIs, but this is not true for status.
Still, we are going to have endpoints for status that will allow that information to be fetched, but in the meantime, we shouldn't tell users to rely on `pro status --format` for this.


docs/explanations/how_to_interpret_the_security_status_command.rst
```diff
@@ -208,63 +208,48 @@ source):
...
-package sources in the system:
+If you need a machine readable version of ``pro security-status``, you can use
+these API endpoint to achieve that:
+
+* :ref:`u.pro.packages.summary.v1 <references/api:u.pro.packages.summary.v1>`
+* :ref:`u.pro.packages.updates.v1 <references/api:u.pro.packages.updates.v1>`
+* :ref:`u.pro.status.is_attached.v1 <references/api:u.pro.status.is_attached.v1>`
+* :ref:`u.pro.status.enabled_services.v1 <references/api:u.pro.status.enabled_services.v1>`
+* :ref:`u.pro.security.status.livepatch_cves.v1 <references/api:u.pro.security.status.livepatch_cves.v1>`
+
```
orndorffgrant, line 219: Sorry, but I think we should delete everything after this line. If any of this detail is useful it can be added to the api reference IMO

s-makin: I think it makes sense for that to be the eventual goal, but can we have this as an interim step while we queue up the follow-up to fold this content into the API docs? Or would you prefer we do it now?

orndorffgrant: I'm okay with this as an interim step!

s-makin (approved):
Let's make "gathering all API-worthy content from the docs and putting it into the API docs" an item of future work - happy for this to be merged as-is

Issue #2655, opened 2023-07-17
docs: document data sent to server for all API calls


orndorffgrant:
We have very little information documented about the machine data being sent to the server. This might deserve an Explanation document of what we send and why.

orndorffgrant:
This also includes documenting which api endpoints require network access, preferably with details of which endpoints they talk to.

orndorffgrant:
While we're at it, lets add an api endpoint document template to dev-docs

PR #3221, opened 2024-07-24
apidocs: include network requirements per api endpoint

orndorffgrant:
## Why is this needed?

This adds network requirements to the api doc metadata and renders it using `apidocgen.py`.

Results in #3223

## Test Steps

Regenerate the api docs and double check the new network information for each endpoint.

---


orndorffgrant:
@lucasmoura @dheyay can you adjudicate this approach to #2655 before I add the data to all endpoints?

orndorffgrant:
Reducing the scope to just saying yes/no for network access rather than trying to enumerate all the reasons for each endpoint.

lucasmoura (approved):

PR #3228, opened 2024-07-25
docs: rename pro package name

lucasmoura:
## Why is this needed?
rename the package from ubuntu-advantage-pro to ubuntu-pro-auto-attach, and ubuntu-advantage-tools to ubuntu-pro-client, as those are the new package names


orndorffgrant (approved):

s-makin (approved):
LGTM :)

PR #3230, opened 2024-07-26
debian: add more scenarios to dep8 tests

lucasmoura:
## Why is this needed?
We are now calling two API endpoints on the dep8 tests. Those endpoints don't trigger any network call and their main use is to verify if any python3-apt change would change the way we interact with that library, as those API endpoints heavily rely on apt functionality.

## Test Steps
1. Build the package
2. Use the .dsc file to run this command:

```bash
autopkgtest --no-built-binaries --apt-upgrade --setup-commands setup-testbed --shell-fail ubuntu-advantage-tools_1+devel.dsc -- lxd ubuntu-daily:<RELEASE>
```

3. Check that this passes and that the API endpoints appears in the logs 

---

### DIFF
--- a/debian/tests/usage
+++ b/debian/tests/usage
@@ -2,5 +2,11 @@
 
 set -ex
 
-ua --help | grep --silent services
-ua version
+pro --help | grep --silent services
+pro version
+
+# These calls are used to ensure that the
+# way we interact with APT is still working
+# if python3-apt receives an upgrade
+pro api u.pro.packages.summary.v1
+pro api u.pro.packages.updates.v1


debian/tests/usage
```diff
@@ -2,5 +2,11 @@
 
 set -ex
 
-ua --help | grep --silent services
-ua version
+pro --help | grep --silent services
+pro version
+
+# This calls are used to unsure that the
```
dheyay, line 8: Minor nit
```suggestion
# These calls are used to ensure that the
```

dheyay (changes requested):

debian/tests/usage
lucasmoura: Good catch

lucasmoura:
@dheyay fixed

dheyay (approved):

renanrodrigo:
Question: are those two real use case calls enough to call our tests non-superficial? I'd say no, but wonder what you think about it

lucasmoura:
@renanrodrigo Not at all, the idea is just to collect extra signal here on possible `python3-apt` changes. I have discussed with the team, and even though we agree this is not yet great, it is already an improvement

orndorffgrant:

PR #3234, opened 2024-07-26
docs: minor updates

dheyay:
## Why is this needed?

This PR was to review and update docs for any missing internal references. So far everything looks great, I could only find one instance on review which could use an internal reference to a connected page.

Also includes a very tiny bug fix


s-makin (approved):
LGTM, thanks for catching those!

PR #3235, opened 2024-07-29
apt-hook: redirect errors away from users

orndorffgrant:
## Why is this needed?

When warnings or errors appear in the output of apt update or apt upgrade that are related to our apt hook, it makes pro look bad. This is true even if the cause of the error is entirely unrelated to pro (see LP#2055239).

Here we are redirecting all output that should never appear to users to files that are collected when you run `pro collect-logs`. This way, we don't lose the messages entirely, and they don't disrupt users.

## Test Steps

```
env SHELL_BEFORE=1 ./tools/test-in-lxd.sh noble

# Reproduce the issue before upgrading
apt update                     # no warnings
systemctl enable bug32939      # trigger https://github.com/systemd/systemd/issues/32959
apt update                     # see warnings that make us look bad
exit

# new version gets installed after exit and lxc shell is re-started

apt update                     # no warnings
systemctl enable bug32939      # trigger https://github.com/systemd/systemd/issues/32959
apt update                     # still no warnings
pro collect-logs
tar -xvf pro_logs.tar.gz
cat logs/apt-news.service.txt  # see the warning
```

---


lucasmoura (approved):

orndorffgrant:

orndorffgrant:
merged in #3244

PR #3240, opened 2024-07-31
cli: add custom help to the main 'pro' parser

renanrodrigo:
## Why is this needed?

This PR solves all of our problems because it implements the `help` structure defined for the new UX.

For all other commands we rely on argparse defaults, but for the main --help entry we want to present the available commands in a customized way.

Some commands will have their own categories, and some will not even appear in help.

## Test Steps
Run help.
There is an integration test to cover the whole thing.

---

### DIFF
--- a/uaclient/cli/__init__.py
+++ b/uaclient/cli/__init__.py
@@ -3,6 +3,8 @@
 import argparse
 import logging
 import sys
+from collections import OrderedDict
+from typing import List, NamedTuple  # noqa: F401
 
 from uaclient import (
     apt,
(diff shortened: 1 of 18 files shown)


uaclient/cli/__init__.py
```diff
@@ -67,23 +95,57 @@ def print_help_for_command(self, command: str):
...
 
+    def format_help(self):
+        if self.use_main_help:
+            return super().format_help()
+        help_output = self.format_usage()
+
+        for category, items in self.help_entries.items():
+            help_output += "\n"
+            help_output += "{}:".format(category)
+            help_output += "\n"
```
lucasmoura, line 106: Just a nit, but this can be achieved in a single line

uaclient/cli/__init__.py
```diff
@@ -55,8 +57,34 @@
...
 
+HelpEntry = NamedTuple(
+    "HelpEntry", [("position", int), ("name", str), ("help_string", str)]
+)
+
 
 class ProArgumentParser(argparse.ArgumentParser):
+    help_entries = OrderedDict(
+        [
+            (messages.CLI_HELP_HEADER_QUICK_START, []),
```
lucasmoura, line 68: Even though I get that idea, but I think we should use a `Enum` here instead. Each Enum value could have a message associated with it, but I think it creates a better separation between the helper categories and the messages module

dheyay (approved):
Lgtm :+1:

uaclient/cli/__init__.py
orndorffgrant: I kind of like seeing the newlines on their own lines. I think it makes it a bit easier to understand what the end result will look like.

uaclient/cli/__init__.py
orndorffgrant: I think I agree - an enum would make it clearer what are allowed values when creating editing or adding sub commands.

orndorffgrant:

uaclient/cli/__init__.py
renanrodrigo: will do

uaclient/cli/__init__.py
renanrodrigo: +1 this was intentional

PR #3241, opened 2024-07-31
api: Do not consider unattended-upgrade disabled when any config is 

aleasto:
## Why is this needed?

There are many configs in the 'Unattended-Upgrade::' prefix which can be disabled while still keeping the main functionality enabled. For example:
 - Unattended-Upgrade::OnlyOnACPower "0" Allows upgrades even on battery power
 - Unattended-Upgrade::Skip-Updates-On-Metered-Connections "0" Allows upgrades even on metered connections

 ... and many more.

You may `grep Unattended-Upgrade:: /usr/bin/unattended-upgrade` for a list of keys that may make sense to disable while keeping unattended-upgrades enabled.

## Test Steps
`echo 'Unattended-Upgrade::OnlyOnACPower "0";' | sudo tee /etc/apt/apt.conf.d/99test`
`pro api u.unattended_upgrades.status.v1 | python3 -m json.tool`

without this change it reports:
```
            "unattended_upgrades_disabled_reason": {
                "code": "unattended-upgrades-cfg-value-turned-off",
                "msg": "Unattended-Upgrade::OnlyOnACPower is turned off"
            },
```

---

### DIFF
--- a/uaclient/api/u/unattended_upgrades/status/v1.py
+++ b/uaclient/api/u/unattended_upgrades/status/v1.py
@@ -160,7 +160,8 @@ def _is_unattended_upgrades_running(
     if not systemd_apt_timer_enabled:
         return (False, messages.UNATTENDED_UPGRADES_SYSTEMD_JOB_DISABLED)
 
-    for key, value in unattended_upgrades_cfg.items():
+    for key in UNATTENDED_UPGRADES_CONFIG_KEYS:
+        value = unattended_upgrades_cfg.get(key)
         if not value:
             return (
                 False,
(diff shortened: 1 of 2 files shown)


uaclient/api/u/unattended_upgrades/status/v1.py
```diff
@@ -224,10 +224,7 @@ def _status(cfg: UAConfig) -> UnattendedUpgradesStatusResult:
     unattended_upgrades_last_run = _get_unattended_upgrades_last_run()
 
     unattended_upgrades_cfg = get_apt_config_values(
-        set(
```
lucasmoura, line 227: That's indeed an issue and you are right that we should not consider all config options set on `unatteded-upgrade`  to see if it is enabled or not.

However, one of the API fields we have is the `raw_config` one. On that field, we should capture all of the configs set for `unattended-upgrade`. By removing this extra line here, we loose visibility of `unattended-upgrade` config set through apt in `raw_config`

What if we instead change the `_is_unattended_upgrades_running` to only check the keys `UNATTENDED_UPGRADES_CONFIG_KEYS` ? In that sense, we would not affect the `raw_config` and we will also fix the issue that you found in the API

lucasmoura (commented):
Thanks for this contribution @aleasto 
I only have one comment here, but you definitely found a bug here

aleasto:
Done: https://github.com/canonical/ubuntu-pro-client/pull/3249

PR #3246, opened 2024-08-01
Onlyseries on reboot

dheyay:
## Why is this needed?

Runs onlySeries check on machine reboot instead of a timer service

Reopened PR - #3198 

Fixes: #3189 

## Test Steps

Run the tox integration test for the `features/ubuntu_upgrade` file to check if the onlySeries check runs successfully on reboot

---


lib/postinst-migrations.sh
```diff
@@ -71,4 +71,24 @@ except Exception as e:
...
+        /usr/bin/python3 -c "
+import json
+with open('/var/lib/ubuntu-advantage/private/machine-token.json') as machine_token_file:
+    content = json.load(machine_token_file)
+    ents = content.get('machineTokenInfo', {}).get('contractInfo', {}).get('resourceEntitlements', [])
+    support = next((e for e in ents if e.get('type') == 'support'), {})
+    only_series = (
+        support.get('affordances', {})
+        .get('onlySeries', None)
+    )
```
lucasmoura, line 86: What if we instead create a method for `MachineTokenFile` that returns the `onlySeries` value ? This will make the code easier to test and make our postinst simpler

lib/reboot_cmds.py
```diff
@@ -77,26 +78,30 @@ def refresh_contract(cfg: config.UAConfig):
 
 
 def main(cfg: config.UAConfig) -> int:
```
lucasmoura, line 80: what if separate main into two blocks:

One that executes the previous behavior tied to the marker file
One that executes the `onlySeries` part

In that sense, we make that separation clearer and we simplify the code a bit. What do you think ?

lib/postinst-migrations.sh
dheyay: Yes, that is something I was also thinking about when doing this, it would make life easier across anything using `onlySeries` too
Will do

lib/reboot_cmds.py
dheyay: The only reason I was a little skeptical about breaking it into two was the `lock` which is already being held by `pro-reboot-cmds`.

I am unsure of having two locks for a similar flow of events since the onlySeries bit is just a function call?  :sweat_smile: 
Let me know what you think!

lib/postinst-migrations.sh
renanrodrigo: +1 move as much as possible to the .py file

lib/reboot_cmds.py
dheyay: Updated @lucasmoura @renanrodrigo

lib/postinst-migrations.sh
```diff
@@ -71,4 +71,18 @@ except Exception as e:
...
+    # Create an only series marker file if not present
+    # This is to ensure that the only-series contract check runs correctly
+    if [ ! -f "/var/lib/ubuntu-advantage/private/machine-token.json" ]; then
+        /usr/bin/python3 -c "
+from uaclient.files import machine_token
+machine_token_file = machine_token.get_machine_token_file()
+only_series = machine_token_file.only_series()
+    if only_series:
+        print(only_series)
+"
```
lucasmoura, line 83: Let's wrap this on `try..except` just to be extra safe, as we don't want the postinst to fail if somehow that code fails

dheyay: Updated @lucasmoura

lib/postinst-migrations.sh
```diff
@@ -71,4 +71,21 @@ except Exception as e:
...
+    # Create an only series marker file if not present
+    # This is to ensure that the only-series contract check runs correctly
+    if [ ! -f "/var/lib/ubuntu-advantage/private/machine-token.json" ]; then
+        /usr/bin/python3 -c "
+from uaclient.files import machine_token
+try:
+    machine_token_file = machine_token.get_machine_token_file()
+    only_series = machine_token_file.only_series()
+        if only_series:
+            print(only_series)
```
renanrodrigo, line 83: Tested a direct upgrade on Xenial to pass the guards:
There is an indentation error here.
```
Setting up ubuntu-advantage-tools (32.3.1~16.04) ...
  File "<string>", line 6
    if only_series:
    ^
IndentationError: unexpected indent
```
Besides that, do we really need to print this?

lib/postinst-migrations.sh
```diff
@@ -71,4 +71,21 @@ except Exception as e:
...
+except Exception as e:
+    print('Error while checking only-series field on contract: {}'.format(e))
+"
+        if [ -n "$only_series" ]; then
```
renanrodrigo, line 87: `$only_series` is not being defined anywhere, so this will always be empty string. So, again, the marker file never gets to be created.

Shouldn't we, instead of printing, create the marker file in the python code itself? Using `state_files.only_series_check_marker_file`?

renanrodrigo: The current integration test goes from series to series, but it would be nice to add a manual test to cover upgrading from older versions (with and without the overlay to force the only series)

renanrodrigo (changes requested):

PR #3247, opened 2024-08-01
Review v34

orndorffgrant:
This changeset represents the diff to be uploaded as version 34 of the `ubuntu-pro-client` package.

* **Pro Client Representative:** @orndorffgrant 
* **Sponsor Reviewer:** @sergiodj 
* **SRU Reviewer:** @panlinux 

## Notes:

This PR is based on tag 33.2 which should be exactly equivalent to what was uploaded as 33.2 to ubuntu. You should check this using `git-ubuntu` to make sure I didn't miss something.


renanrodrigo:
@orndorffgrant `translation-updates` is failing here - are we missing some `update-pos` run?

orndorffgrant:
> @orndorffgrant `translation-updates` is failing here - are we missing some `update-pos` run?

Thanks sorry I didn't notice that - updated now.

sergiodj (commented):
Thanks, @orndorffgrant.

I'm still reviewing the PR, but decided to post the comments I had so far so that you don't need to wait for my full review.  Most of the comments are really nitpicks; there's just one question about the invocation of `tools/check-versions-are-consistent.py` inside `d/rules` that I'd like to clarify.

orndorffgrant:
@sergiodj Thank you for the review! I've addressed all of the feedback (except the open question for @panlinux) in a PR here: #3263. Please review that PR when you have time.

sergiodj:
> @sergiodj Thank you for the review! I've addressed all of the feedback (except the open question for @panlinux) in a PR here: #3263. Please review that PR when you have time.

Thanks, @orndorffgrant!  I'll take a look at the PR tomorrow :-).

debian/rules
```diff
@@ -30,6 +30,10 @@ override_dh_auto_build:
 
 override_dh_auto_test:
 ifeq (,$(filter nocheck,$(DEB_BUILD_OPTIONS)))
+	# fail early if version.py wasn't updated to match d/changelog's version
+	# only take into account the actual upstream version, discarding ubuntu and
+	# backport suffixes
+	python3 tools/check-versions-are-consistent.py
```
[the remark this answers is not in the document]
panlinux: Agreed, this was mentioned in the previous SRU and deemed not critical back then, but we should change it. +1 to move it to override_dh_auto_build:
```diff
--- a/debian/rules
+++ b/debian/rules
@@ -30,16 +30,16 @@ endif
        dh $@ --with python3,bash-completion,systemd --buildsystem=pybuild
 
 override_dh_auto_build:
+       # fail early if version.py wasn't updated to match d/changelog's version
+       # only take into account the actual upstream version, discarding ubuntu and
+       # backport suffixes
+       python3 tools/check-versions-are-consistent.py
        dh_auto_build
        make -C apt-hook build
        make -C debian/po build
 
 override_dh_auto_test:
 ifeq (,$(filter nocheck,$(DEB_BUILD_OPTIONS)))
-       # fail early if version.py wasn't updated to match d/changelog's version
-       # only take into account the actual upstream version, discarding ubuntu and
-       # backport suffixes
-       python3 tools/check-versions-are-consistent.py
 # Hooks will only be delivered on LTS instances
 ifeq (LTS,$(findstring LTS,$(VERSION)))
        make -C apt-hook test
```
Just tested and it correctly fails if there is a version mismatch:
```
make[1]: Entering directory '/home/ubuntu/git/packages/ubuntu-advantage-tools/ubuntu-advantage-tools'
# fail early if version.py wasn't updated to match d/changelog's version
# only take into account the actual upstream version, discarding ubuntu and
# backport suffixes
python3 tools/check-versions-are-consistent.py
version.py says "34" but changelog says "34.1"
make[1]: *** [debian/rules:36: override_dh_auto_build] Error 1
make[1]: Leaving directory '/home/ubuntu/git/packages/ubuntu-advantage-tools/ubuntu-advantage-tools'
make: *** [debian/rules:30: binary] Error 2
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2
```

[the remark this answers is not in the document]
sergiodj: Thanks, @panlinux.

sergiodj:
Given @panlinux's comments, I'm going to wait until `d/rules` is updated accordingly before approving this PR :).

orndorffgrant:
@sergiodj - just updated this with @panlinux's suggested diff :)

sergiodj:
> @sergiodj - just updated this with @panlinux's suggested diff :)

Thanks!  I will take another quick look.

sergiodj (approved):
Thanks, @orndorffgrant.

LGTM, +1.

orndorffgrant:
@sergiodj Sorry - we found two small things that cannot wait until v35. Please review the last three commits on this PR when you have time.

sergiodj:
Thanks for the heads up, @orndorffgrant.

I looked at the two commits and they seem OK.  Still LGTM, +1.

debian/changelog
```diff
@@ -688,7 +707,7 @@ ubuntu-advantage-tools (27.11~22.10.1) kinetic; urgency=medium
     - pro:
       + Ubuntu Pro is released as a product
       + make `pro` the recommended executable for the client
-      + client, apt and motd messages updated/rewritten to show Pro 
+      + client, apt and motd messages updated/rewritten to show Pro
```
panlinux, line 710: Here and above: don't change old changelog entries

apt-hook/20apt-esm-hook.conf
```diff
@@ -1,7 +1,7 @@
 APT::Update::Pre-Invoke {
-	"[ ! -e /run/systemd/system ] || [ $(id -u) -ne 0 ] || systemctl start --no-block apt-news.service esm-cache.service || true";
+	"[ ! -e /run/systemd/system ] || [ $(id -u) -ne 0 ] || systemctl start --no-block apt-news.service esm-cache.service >/dev/null 2>&1 || true";
 };
 
 binary::apt::AptCli::Hooks::Upgrade {
-	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-json-hook ] || /usr/lib/ubuntu-advantage/apt-esm-json-hook || true";
+	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-json-hook ] || /usr/lib/ubuntu-advantage/apt-esm-json-hook 2> /var/lib/ubuntu-advantage/apt-upgrade-hook-err.txt || true";
```
panlinux, line 6: I'll take a closer look at this on Monday. The questions that come to mind at the moment are (and I might have answers for them myself when I next look into it, I'm just writing them down):
- even though LP: #2055239 is/was annoying, it helped to chase down this snap bug, so in the end it was useful. Are we sure we want to hide error messages?
- why are errors from apt-news and esm-cache sent to /dev/null, but the ones from apt-esm-json-hook not?
- why a .txt extension, and not .log?
- the way this is being done, the .txt file will be overwritten each time. If there was an error message in in before, it will be gone.
- conversely, if we append to the file instead of overwriting to it, we should think about its size growing without bounds
- why in `/var/lib/ubuntu-advantage` and not in `/var/log/<something>`?
- don't we have to clean this .txt file up on package removal?

panlinux: I checked some other scripts in /etc/apt/apt.conf.d/. Some redirect everything to /dev/null, others just redirect stdout, others just stderr, none to a file.

Have you considered using a proper log? Perhaps even `| logger ...`? For both perhaps?

debian/changelog
orndorffgrant: Sorry, that was my editor auto-removing trailing whitespace on save. I'll add back the trailing whitespace. Or is there any way we can correct this?

apt-hook/20apt-esm-hook.conf
orndorffgrant: Thanks for thinking through this! I have some responses, but ultimately we need to make a call on the tradeoffs.

> I'll take a closer look at this on Monday. The questions that come to mind at the moment are (and I might have answers for them myself when I next look into it, I'm just writing them down):
> 
>     * even though LP: #2055239 is/was annoying, it helped to chase down this snap bug, so in the end it was useful. Are we sure we want to hide error messages?

*systemd (not snap)
I'd argue that it was only useful/necessary to fix LP: #2055239 because the messages from this hook were so annoying. IMO it's not that harmful a bug except for causing these messages that were very visible and made us look bad.

> 
>     * why are errors from apt-news and esm-cache sent to /dev/null, but the ones from apt-esm-json-hook not?
> 

The thing I had in mind with these redirections is: We still want to get this information if someone runs `pro collect-logs`. `pro collect-logs` already collects the status and logs of those systemd services - warnings/errors should be present there. The new log-like file was introduced so that we could collect apt-esm-json-hook info in `pro collect-logs`.

>     * why a .txt extension, and not .log?

As currently implemented it will only ever contain the most recent `stderr` output, so it's not really a log per se IMO, but I could be convinced otherwise.

> 
>     * the way this is being done, the .txt file will be overwritten each time. If there was an error message in in before, it will be gone.

Yes, I considered this a reasonable tradeoff for the simplicity of the approach. I worry about adding additional complexity and therefore ways that these hooks could fail.

> 
>     * conversely, if we append to the file instead of overwriting to it, we should think about its size growing without bounds

Yes, this is some of the complexity I avoided/punted on with the current approach

> 
>     * why in `/var/lib/ubuntu-advantage` and not in `/var/log/<something>`?

Same reason for not giving it a `.log` extension, but I could be convinced either way.

> 
>     * don't we have to clean this .txt file up on package removal?

Good point. So far we haven't removed anything in /var/lib/ubuntu-advantage on removal (but we do remove it entirely on purge). We could start removing this on removal though.

> I checked some other scripts in /etc/apt/apt.conf.d/. Some redirect everything to /dev/null, others just redirect stdout, others just stderr, none to a file.
> 
> Have you considered using a proper log? Perhaps even `| logger ...`? For both perhaps?

TIL about `logger` (thanks!). I'm only worried about unnecessary complexity and unnecessary visibility of errors. These aren't exactly critical services, so extra warnings/errors in syslog may be bothersome without much benefit. I could be convinced otherwise though. Do you think this would be useful and worth the tradeoffs?

panlinux: This whole approach sounds like it could use a second pass. As stated above, the moment I saw this, all these different questions popped into my mind. It just feels weird. If we are interested in possible errors messages, they should be logged. Writing them to some mysterious file in /var/lib is unexpected, and we will really only get the output from the last run, so I question its usefulness. Maybe start with a list:
* What are the kind of error messages we expect from each service?
* Is it bad if users see these messages? Why?
* Is it bad if the users DO NOT see these messages?
* apt-news, esm-cache, already log to ubuntu-advantage.log. Maybe the apt-es-json-hook should too
* I understand that the errors from apt-news did not come from apt-news per se, but from systemd/systemctl
* I also understand the extra complexity in getting apt-esm-json-hook to natively log somewhere (c++, hey). `logger` could help, but it's still extra complexity and not something to be rushed in at the eleventh hour.
* What are the consequences of NOT doing these changes at this time? Just more annoyance at that never-ending systemd/snap bug?
*

orndorffgrant: Just to close this out. We discussed this out of band and decided on the fix implemented in #3288

orndorffgrant:
@panlinux I merged the fixup PR so those commits are here now. Let us know if you have any more feedback.

panlinux (approved):
+1 from me (pre-SRU)

orndorffgrant:
Thank you!!

@sergiodj v34 is now ready for upload to oracular (from `main`) and to the unapproved queue for each of x, b, f, j, n from the following branches:

* https://github.com/canonical/ubuntu-pro-client/tree/release-34-xenial
* https://github.com/canonical/ubuntu-pro-client/tree/release-34-bionic
* https://github.com/canonical/ubuntu-pro-client/tree/release-34-focal
* https://github.com/canonical/ubuntu-pro-client/tree/release-34-jammy
* https://github.com/canonical/ubuntu-pro-client/tree/release-34-noble

Please double check that each of those branches are identical to `main` except for an additional commit with a correctly formatted changelog entry with correct version number for the given ubuntu release.

Let me know if you have any questions or concerns.

orndorffgrant:
@sergiodj (for monday at this point :upside_down_face:)
Sorry for the confusion around that last bugfix getting in. I've updated the changelog and translations here, and went ahead and updated all the SRU branches linked in my previous comment. Please do double check everything as usual.

orndorffgrant:
This was successfully released today!
```
grant@lemur ~/c/w/v/3/verification-2074211> rmadison ubuntu-pro-client
 ubuntu-pro-client | 31.2.3   | noble           | amd64, arm64, armhf, ppc64el, riscv64, s390x
 ubuntu-pro-client | 34~16.04 | xenial-proposed | amd64, arm64, armhf, i386, ppc64el, s390x
 ubuntu-pro-client | 34~16.04 | xenial-updates  | amd64, arm64, armhf, i386, ppc64el, s390x
 ubuntu-pro-client | 34~18.04 | bionic-proposed | amd64, arm64, armhf, i386, ppc64el, s390x
 ubuntu-pro-client | 34~18.04 | bionic-updates  | amd64, arm64, armhf, i386, ppc64el, s390x
 ubuntu-pro-client | 34~20.04 | focal-proposed  | amd64, arm64, armhf, ppc64el, riscv64, s390x
 ubuntu-pro-client | 34~20.04 | focal-updates   | amd64, arm64, armhf, ppc64el, riscv64, s390x
 ubuntu-pro-client | 34~22.04 | jammy-proposed  | amd64, arm64, armhf, ppc64el, riscv64, s390x
 ubuntu-pro-client | 34~22.04 | jammy-updates   | amd64, arm64, armhf, ppc64el, riscv64, s390x
 ubuntu-pro-client | 34~24.04 | noble-proposed  | amd64, arm64, armhf, ppc64el, riscv64, s390x
 ubuntu-pro-client | 34~24.04 | noble-updates   | amd64, arm64, armhf, ppc64el, riscv64, s390x
 ubuntu-pro-client | 34       | oracular        | amd64, arm64, armhf, ppc64el, riscv64, s390x
```

Thanks everyone!

PR #3249, opened 2024-08-02
api: Do not report unattended-upgrade disabled if any config is false

aleasto:
## Why is this needed?

There are many configs in the 'Unattended-Upgrade::' prefix which can be disabled while still keeping the main functionality enabled. For example:
 - Unattended-Upgrade::OnlyOnACPower "0" Allows upgrades even on battery power
 - Unattended-Upgrade::Skip-Updates-On-Metered-Connections "0" Allows upgrades even on metered connections

 ... and many more.

You may `grep Unattended-Upgrade:: /usr/bin/unattended-upgrade` for a list of keys that may make sense to disable while keeping unattended-upgrades enabled.

## Test Steps
`echo 'Unattended-Upgrade::OnlyOnACPower "0";' | sudo tee /etc/apt/apt.conf.d/99test`
`pro api u.unattended_upgrades.status.v1 | python3 -m json.tool`

without this change it reports:
```
            "unattended_upgrades_disabled_reason": {
                "code": "unattended-upgrades-cfg-value-turned-off",
                "msg": "Unattended-Upgrade::OnlyOnACPower is turned off"
            },
```

---

### DIFF
--- a/uaclient/api/tests/test_api_u_unattended_upgrades_status_v1.py
+++ b/uaclient/api/tests/test_api_u_unattended_upgrades_status_v1.py
@@ -45,23 +45,63 @@ class TestIsUnattendedUpgradesRunning:
             (False, {}, False, UNATTENDED_UPGRADES_SYSTEMD_JOB_DISABLED),
             (
                 True,
-                {"test": []},
+                {
+                    "APT::Periodic::Enable": "1",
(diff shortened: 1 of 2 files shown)


lucasmoura (approved):
Thanks for fixing this issue @aleasto

aleasto:
Added test case for `"Unattended-Upgrade::OnlyOnACPower": "0"`

PR #3250, opened 2024-08-02
cli: add custom help to the main 'pro' parser

renanrodrigo:
## Why is this needed?

This PR solves all of our problems because it implements the `help` structure defined for the new UX.

For all other commands we rely on argparse defaults, but for the main --help entry we want to present the available commands in a customized way.

Some commands will have their own categories, and some will not even appear in help.

This is the same as #3240.

## Test Steps
Run help.
There is an integration test to cover the whole thing.

---

### DIFF
--- a/features/help.feature
+++ b/features/help.feature
@@ -7,29 +7,36 @@ Feature: Pro Client help text
       """
       usage: pro [-h] [--debug] [--version] <command> ...
 
+      Quick start commands:
+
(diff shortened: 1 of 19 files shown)


renanrodrigo:
@orndorffgrant @lucasmoura Enum added, feel free to take another look

lucasmoura (approved):

dheyay (approved):

PR #3251, opened 2024-08-02
lintian: adjust overrides for oracular

orndorffgrant:
## Why is this needed?

lintian was erroring on oracular builds + had mismatched warning overrides. This fixes that.

## Test Steps

build and run lintian for oracular

---


dheyay (approved):
:+1:

PR #3255, opened 2024-08-06
tests: drop amc snap from list of required snaps

morphis:
## Why is this needed?

The anbox-cloud product no longer asks to install the amc snap so we should not check for it. The amc command has been merged into the anbox-cloud-appliance snap.

The change isn't yet live in staging or production, please coordinate with @morphis before merging.

---

### DIFF
--- a/features/anbox.feature
+++ b/features/anbox.feature
@@ -73,7 +73,6 @@ Feature: Enable anbox on Ubuntu
       """
       https://archive.anbox-cloud.io/stable <release>/main amd64 Packages
       """
-    And I check that snap `amc` is not installed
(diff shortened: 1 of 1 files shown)


lucasmoura:
Thanks for this PR @morphis !

I think we just need to remove another line that checks for the existence of amc:
`And I check that snap amc is installed` in this line [here](https://github.com/canonical/ubuntu-pro-client/blob/next-v35/features/anbox.feature#L109)

morphis:
Thanks @lucasmoura ! Fixed and updated the PR.

orndorffgrant (approved):

PR #3256, opened 2024-08-06
api: Do not crash when a Unattended-Upgrade config is missing

aleasto:
## Why is this needed?

The keys in UNATTENDED_UPGRADES_CONFIG_KEYS are considered to be mandatory in order to report unattended-upgrades as enabled.

When parsing the config files, get_apt_config_values will set missing keys to None rather than not populating the dict. Thus unattended_upgrades_cfg.get(key, fallback) for a missing key returns None rather than the fallback. Finally `int(None)` fails.

## Test Steps
```
sudo sed -i "/APT::Periodic::Unattended-Upgrade/s/^/#/" /etc/apt/apt.conf.d/20auto-upgrades
pro api u.unattended_upgrades.status.v1
```
should not fail.

---

### DIFF
--- a/uaclient/api/tests/test_api_u_unattended_upgrades_status_v1.py
+++ b/uaclient/api/tests/test_api_u_unattended_upgrades_status_v1.py
@@ -86,6 +86,19 @@ class TestIsUnattendedUpgradesRunning:
                 True,
                 {
                     "APT::Periodic::Enable": "1",
+                    "APT::Periodic::Update-Package-Lists": None,
+                    "APT::Periodic::Unattended-Upgrade": "1",
+                    "Unattended-Upgrade::Allowed-Origins": ["foo"],
+                },
+                False,
+                UNATTENDED_UPGRADES_CFG_VALUE_TURNED_OFF.format(
+                    cfg_name="APT::Periodic::Update-Package-Lists"
+                ),
+            ),
+            (
+                True,
+                {
+                    "APT::Periodic::Enable": "1",
                     "APT::Periodic::Update-Package-Lists": "1",
                     "APT::Periodic::Unattended-Upgrade": "1",
                     "Unattended-Upgrade::Allowed-Origins": ["foo"],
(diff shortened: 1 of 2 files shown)


aleasto:
Should this case return `UNATTENDED_UPGRADES_CFG_LIST_VALUE_EMPTY` or `UNATTENDED_UPGRADES_CFG_VALUE_TURNED_OFF` though?

lucasmoura (approved):
Great catch @aleasto!
Thanks for fixing it.

Now regarding the question about the message we return, since the variable is not there, I would assume it would be empty instead of turned off. But I don't have a strong opinion here, let me know if you think the `turned off` message is better

aleasto:
I think the "empty" message was specifically targeted at lists (basically just `Allowed-Origins`), and indeed the `unattended-upgrades-cfg-list-value-empty` identifier would suggest that, but it does fit configs whose semantics are disabled-by-default too.

orndorffgrant:
At least for the cases being touched here, if the value is not specified in the configuration, then it defaults to "0" (disabled), so they are turned off. The method of turning the config value off (either by specifying it as off or by omission) doesn't matter. The thing our api needs to report is the _effective value_. So I vote for `UNATTENDED_UPGRADES_CFG_VALUE_TURNED_OFF`.

It appears the current code that decides on the disable reason doesn't take into account default values though and would list this situation as `UNATTENDED_UPGRADES_CFG_LIST_VALUE_EMPTY`, so that sounds like an additional small bug.

aleasto:
> It appears the current code that decides on the disable reason doesn't take into account default values though

Indeed it does not, but all configs in `UNATTENDED_UPGRADES_CONFIG_KEYS` have disabled-by-default semantics except `APT::Periodic::Enable` which we are patching here: 
https://github.com/canonical/ubuntu-pro-client/blob/9bf7ff33b5a849344a98c4a66c04d098ad5de70a/uaclient/api/u/unattended_upgrades/status/v1.py#L234-L239

aleasto:
I've added another commit to this PR that makes missing configs report as VALUE_TURNED_OFF.

lucasmoura (approved):

renanrodrigo (approved):

orndorffgrant:
Excellent - thank you!

PR #3258, opened 2024-08-07
api: add access and config endpoints

dheyay:
## Why is this needed?

This PR adds the status apis:

`u.pro.access.v1`
`u.pro.config.v1`

## Test Steps

---


uaclient/cli/config.py
```diff
@@ -35,6 +36,7 @@ def action_config_show(args, *, cfg, **kwargs):
     :return: 0 on success
     :raise UbuntuProError: on invalid keys
     """
+    uaconfig = _config(cfg)
```
lucasmoura, line 39: just for consistency reasons, lets call this `proconfig` instead

uaclient/api/u/pro/subscription/v1.py
```diff
@@ -0,0 +1,265 @@
...
+    machine_token_file = machine_token.get_machine_token_file(cfg)
+    machineTokenInfo = machine_token_file.machine_token["machineTokenInfo"]
+
+    # Contract Info
+    contractInfo = machineTokenInfo["contractInfo"]
+    tech_support_level = UserFacingStatus.INAPPLICABLE.value
+    expires = (
+        machine_token_file.contract_expiry_datetime
+        if contractInfo.get("effectiveTo")
+        else contractInfo.get("effectiveTo", None)
```
lucasmoura, line 172: isn't the `contract_expiry_datetime` property already performing this check ?

uaclient/api/u/pro/subscription/v1.py
```diff
@@ -0,0 +1,265 @@
...
+    )
+    effective = contractInfo.get("effectiveFrom", None)
+
+    support = (
+        machine_token_file.entitlements().get("support", {}).get("entitlement")
+    )
+    if support:
+        supportLevel = support.get("affordances", {}).get("supportLevel")
+        if supportLevel:
+            tech_support_level = supportLevel
```
lucasmoura, line 182: maybe it is better to create a property directly on `machine_token_file` for this. My rationale is that all logic need to extract information from the machine token file should live in that class

uaclient/api/u/pro/subscription/v1.py
```diff
@@ -0,0 +1,265 @@
...
+
+def subscription() -> SubscriptionResult:
+    return _subscription(UAConfig())
+
+
+def _subscription(cfg: UAConfig) -> SubscriptionResult:
+    """
+    Returns the Ubuntu Pro subscription information for the machine.
+    """
+    if not _is_attached(cfg).is_attached:
```
lucasmoura, line 141: should we have a `is_attached` field here ? Just to easily allow users to know that they are running on an unattached machine

uaclient/api/u/pro/config/v1.py
```diff
@@ -0,0 +1,145 @@
...
+            required=False,
+            doc="Update messaging timer",
+        ),
+    ]
+
+    def __init__(
+        self,
+        *,
+        http_proxy: Optional[str] = None,
+        https_proxy: Optional[str] = None,
```
lucasmoura, line 72: I think it is worth having integration tests where we set proxy values and see how they behave on `non-root` vs `root` users

uaclient/api/u/pro/subscription/v1.py
```diff
@@ -0,0 +1,265 @@
...
+        id=machine_token_file.account["id"],
+        name=machine_token_file.account["name"],
+    )
+
+    # Subscription Result
+    activity_id = (
+        machine_token_file.activity_id
+        if machine_token_file.activity_id
+        else ""
+    )
```
lucasmoura, line 210: this can be a `machine_token_file.activity_id or ""`

lucasmoura:
@dheyay let's also add a quick description on the commit message telling what each API should do

uaclient/api/u/pro/config/v1.py
dheyay: Yes, makes sense. Will add those to check if the config output is correct in both cases

uaclient/api/u/pro/subscription/v1.py
```diff
@@ -0,0 +1,265 @@
...
+            activity_id="",
+            machine_is_attached=False,
+        )
+
+    machine_token_file = machine_token.get_machine_token_file(cfg)
+    machineTokenInfo = machine_token_file.machine_token["machineTokenInfo"]
+
+    # Contract Info
+    contractInfo = machineTokenInfo["contractInfo"]
+    tech_support_level = UserFacingStatus.INAPPLICABLE.value
```
orndorffgrant, line 178: We shouldn't use any `UserFacingStatus` values in the api results.
If there is not tech support level in the machine token info, then we should return null

uaclient/api/u/pro/subscription/v1.py
```diff
@@ -0,0 +1,265 @@
...
+    "example_python": """
+from uaclient.api.u.pro.subscription.v1 import subscription
+
+result = subscription()
+""",  # noqa: E501
+    "result_class": SubscriptionResult,
+    "exceptions": [],
+    "example_cli": "pro api u.pro.subscription.v1",
+    "example_json": """
+{
```
orndorffgrant, line 238: indentation is wrong for this example json

uaclient/api/u/pro/config/v1.py
```diff
@@ -0,0 +1,146 @@
...
+from uaclient.api.u.pro.config.v1 import config
+
+result = subscription()
+""",  # noqa: E501
+    "result_class": ConfigInfo,
+    "exceptions": [],
+    "example_cli": "pro api u.pro.config.v1",
+    "example_json": """
+{
+"attributes": {
```
orndorffgrant, line 128: indentation is wrong for this example json

uaclient/cli/config.py
```diff
@@ -43,7 +45,7 @@ def action_config_show(args, *, cfg, **kwargs):
             )
         print(
             "{key} {value}".format(
-                key=args.key, value=getattr(cfg, args.key, None)
+                key=args.key, value=getattr(pro_config, args.key, None)
```
orndorffgrant, line 48: We can refactor this function more.
First turn `pro_config` into a dict with `.to_dict(keep_none=True)`.
Instead of checking if the key is in `UA_CONFIGURABLE_KEYS`, check if it's in the dict.
Instead of using `getattr`, use `.get` on the dict.

uaclient/cli/config.py
```diff
@@ -52,11 +54,11 @@ def action_config_show(args, *, cfg, **kwargs):
...
 
     for key in config.UA_CONFIGURABLE_KEYS:
-        print(row_tmpl.format(key=key, value=getattr(cfg, key, None)))
+        print(row_tmpl.format(key=key, value=getattr(pro_config, key, None)))
 
-    if (cfg.global_apt_http_proxy or cfg.global_apt_https_proxy) and (
-        cfg.ua_apt_http_proxy or cfg.ua_apt_https_proxy
-    ):
+    if (
+        pro_config.global_apt_http_proxy or pro_config.global_apt_https_proxy
```
orndorffgrant, line 60: Even if you use my `to_dict` suggestion above, still do this direct attribute reference for this if statement.
The cases are:
1. if key is a variable, use the dict
2. if the key we care about is known at authoring time, use the attribute on the object

uaclient/api/u/pro/subscription/v1.py
```diff
@@ -0,0 +1,265 @@
...
+        account=account,
+        machine_id=machineTokenInfo["machineId"],
+        activity_id=activity_id,
+        machine_is_attached=machine_is_attached,
+    )
+
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="subscription",
```
orndorffgrant, line 221:
```suggestion
    name="Subscription",
```

uaclient/api/u/pro/config/v1.py
```diff
@@ -0,0 +1,146 @@
...
+def config() -> ConfigInfo:
+    return _config(UAConfig())
+
+
+def _config(cfg: UAConfig) -> ConfigInfo:
+    """This endpoint returns the current user configuration"""
+    pro_config = user_config_file.user_config.read().to_dict()
+    for key in UA_CONFIGURABLE_KEYS:
+        if hasattr(cfg, key) and pro_config[key] is None:
+            pro_config[key] = getattr(cfg, key)
```
orndorffgrant, line 103: I think this can actually be
```suggestion
    pro_config = {}
    for key in UA_CONFIGURABLE_KEYS:
        if hasattr(cfg, key):
            pro_config[key] = getattr(cfg, key)
```

dheyay: Updated @orndorffgrant

lucasmoura (commented):
Let's also import both of these endpoints in this test here:

orndorffgrant (commented):
My only real critique is a git commit history thing. The first commit brings in a file called `api_status.feature` and the next commit deletes it. I think it can be removed entirely from the first commit.

Also the new api .feature files can go in the new `features/api` folder now.

And finally, should we delay this until the subscription naming conversation is resolved?

dheyay:
@orndorffgrant Yes, I was of the opinion this was delayed until there was a resolution on the naming convention. I can fix the conflicts up and commit history as well until then.

orndorffgrant:
Thanks this looks good! We can just hold until we decide on the name.

features/api/access.feature
```diff
@@ -0,0 +1,90 @@
...
+        contractInfo:
+          id: "TestCId"
+          name: "TestCName"
+          products: []
+        accountInfo:
+          id: "TestAAId"
+          name: "TestAAName"
+        machineId: "TestMId"
+      """
+    When I run `pro api u.pro.access.v1` with sudo
```
lucasmoura, line 19: @dheyay Let's add a non-root check on this test too

dheyay: Updated @lucasmoura

orndorffgrant (approved):
Thank you! I'll just let CI run the tests and double check the results before merging.

lucasmoura (approved):

features/api/api.feature
```diff
@@ -22,6 +24,7 @@ Feature: Client behaviour for the API endpoints
     When I run `python3 -c "from uaclient.api.u.pro.status.enabled_services.v1 import enabled_services"` as non-root
     When I run `python3 -c "from uaclient.api.u.pro.status.is_attached.v1 import is_attached"` as non-root
     When I run `python3 -c "from uaclient.api.u.pro.status.notices.v1 import notices"` as non-root
+    When I run `python3 -c "from uaclient.api.u.pro.subscription.v1 import subscription"` as non-root
```
renanrodrigo, line 27: Not sure this should stay

uaclient/api/api.py
```diff
@@ -34,6 +36,7 @@
     "u.pro.status.enabled_services.v1",
     "u.pro.status.is_attached.v1",
     "u.pro.status.notices.v1",
+    "u.pro.subscription.v1",
```
renanrodrigo, line 39: same

uaclient/cli/config.py
renanrodrigo: Do we want to implement these changes?

uaclient/cli/config.py
renanrodrigo: ^

uaclient/cli/config.py
dheyay: I remember we decided to keep it the same since the pro_config variable dict does not have the `apt_http_proxy` and `apt_https_proxy` values.

We can keep the pro_config rename for the variable though since we are now calling the api for those values?@renanrodrigo

renanrodrigo: whatever you have decided - I asked because there was a thumbs up and no conclusion in the thread. I'm fine

renanrodrigo (approved):

renanrodrigo:
CI failures unrelated

PR #3259, opened 2024-08-08
d/control: breaks old python3-minimal on noble

orndorffgrant:
## Why is this needed?

This forces the fix for LP#2075337 to be applied when upgrading to ubuntu-pro-client only on noble.

LP: #2075337

## Test Steps

Test that the `debian/control` file for noble has the extra `Breaks` and builds for other releases don't

---


dheyay (approved):
:+1:

lucasmoura (approved):

renanrodrigo (approved):
tested both in noble and older releases, behaves as expected

PR #3260, opened 2024-08-08
Vulnerabilities api

lucasmoura:
## Why is this needed?
We are adding support for the:

* `u.pro.security.vulnerabilities.cve.v1`
* `u.pro.security.vulnerabilities.usn.v1`
* `u.pro.packages.updates_with_cves.v1`

These API endpoints will list which CVEs/USNs are affecting the current system, while the `updates_with_cves` will tell which CVEs are tied to a given package update.

## Test Steps
Run the integration test for each API endpoint


uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,325 @@
...
+                            "pocket"
+                        )
+                        fix_version = (
+                            source_version[vuln_fixed_version]
+                            .get("binary_packages", {})
+                            .get(pkg_name, "")
+                        )
+                    except KeyError:
+                        # There is bug in the data where some sources are
+                        # not present. The Security team is already aware
```
orndorffgrant, line 304: Do we still need this? If we do is there a reference to a bug or something that will be updated when it is fixed?

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,325 @@
...
+        self.vulnerabilities = {}
+
+    def get_package_vulnerabilities(self, affected_pkg: Dict[str, Any]):
+        raise NotImplementedError
+
+    def parse_data(
+        self,
+        vulnerabilities_data: Dict[str, Any],
+        installed_pkgs_by_source: Dict[str, Dict[str, str]],
+    ):
```
orndorffgrant, line 247: Consider looking at ways to make this long function a bit more readable for future us.

uaclient/api/u/pro/security/vulnerabilities/cve/v1.py
```diff
@@ -0,0 +1,383 @@
...
+            doc="The date of the last apt update operation in the system",
+        ),
+    ]
+
+    def __init__(
+        self,
+        *,
+        cves: List[CVEVulnerabilityResult],
+        vulnerability_data_published_at: datetime.datetime,
+        apt_updated_at: Optional[datetime.datetime] = None
```
orndorffgrant, line 227: I think this should be None when using a manifest file

uaclient/api/u/pro/security/vulnerabilities/cve/v1.py
```diff
@@ -0,0 +1,383 @@
...
+    )
+
+    block_fixable_cves = False
+    block_unfixable_cves = False
+
+    if options.unfixable:
+        block_fixable_cves = True
+
+    if not options.unfixable and not options.all:
+        block_unfixable_cves = True
```
orndorffgrant, line 278:
```suggestion
    if options.unfixable:
        block_fixable_cves = True
        block_unfixable_cves = False
    elif not options.all:
        block_fixable_cves = False
        block_unfixable_cves = True
    else:
        block_fixable_cves = False
        block_unfixable_cves = False
    
```
minor readability suggestion

uaclient/api/u/pro/security/vulnerabilities/cve/v1.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+def vulnerabilities(
+    options: CVEVulnerabilitiesOptions,
+) -> CVEVulnerabilitiesResult:
+    return _vulnerabilities(options, UAConfig())
+
+
+def _vulnerabilities(
+    options: CVEVulnerabilitiesOptions,
```
orndorffgrant, line 248: Let's throw an exception if unfixable and all are both true

uaclient/api/u/pro/security/vulnerabilities/cve/v1.py
```diff
@@ -0,0 +1,383 @@
...
+        ):
+            continue
+
+        cves.append(
+            CVEVulnerabilityResult(
+                name=cve_name,
+                description=cve["description"],
+                published_at=datetime.datetime.strptime(
+                    cve["published_at"], "%Y-%m-%dT%H:%M:%S"
+                ),
```
orndorffgrant, line 304: Try out using parse_rfc3339 here and if it isn't too slow, lets use it.

uaclient/api/u/pro/security/vulnerabilities/cve/v1.py
```diff
@@ -0,0 +1,383 @@
...
+
+endpoint = APIEndpoint(
+    version="v1",
+    name="CVEVulnerabilities",
+    fn=_vulnerabilities,
+    options_cls=CVEVulnerabilitiesOptions,
+)
+
+_doc = {
+    "introduced_in": "34",
```
orndorffgrant, line 340:
```suggestion
    "introduced_in": "35",
```

uaclient/data_types.py
```diff
@@ -112,6 +112,24 @@ def from_value(val: Any) -> int:
...
+
+    python_type_name = "float"
+
+    @staticmethod
+    def from_value(val: Any) -> float:
+        if not isinstance(val, float):
+            raise IncorrectTypeError(
+                expected_type="float", got_type=type(val).__name__
+            )
+        return val
```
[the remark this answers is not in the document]
orndorffgrant: This would apply if there is ever a cvss score that is e.g. `3` not `3.0`

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,325 @@
...
+        raw_json_data = download_xz_file_from_url(data_url)
+
+        json_data = json.loads(raw_json_data.decode("utf-8"))
+        if we_are_currently_root():
+            self._save_cache(series, json_data, last_published_date)
+
+        return json_data
+
+
+def _get_source_package_from_vulnerabilities_data(
```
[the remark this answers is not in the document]
orndorffgrant: Discussed - probably not worth it

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,325 @@
...
+from uaclient.defaults import (
+    VULNERABILITY_CACHE_PATH,
+    VULNERABILITY_DATA_CACHE,
+    VULNERABILITY_DATA_TMPL,
+    VULNERABILITY_PUBLISH_DATE_CACHE,
+)
+from uaclient.entitlements.fips import FIPSEntitlement, FIPSUpdatesEntitlement
+from uaclient.http import download_xz_file_from_url, readurl
+from uaclient.system import get_release_info, load_file, write_file
+from uaclient.util import we_are_currently_root
```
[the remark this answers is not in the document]
renanrodrigo: I would say importing individual functions work better in some cases, but for those three I agree - most of the functionality there is not explicit anyway

renanrodrigo (approved):
Thanks for implementing this - nice work.

I had a second look now and my two or three comments are covered by @orndorffgrant's detailed review - thanks for that too.

Pre-approving here pending existing reviews

features/api_updates_with_cves.feature
```diff
@@ -0,0 +1,140 @@
...
+              "provided_by": "esm-infra",
+              "related_cves": [
+                "CVE-2021-28831",
+                "CVE-2022-48174"
+              ],
+              "status": "pending_attach",
+              "version": ".*"
+            }
+          ],
+          "vulnerability_data_published_at": ".*"
```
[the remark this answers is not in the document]
lucasmoura: We can, but I don't think there is much value using that in the test, since this would capture the data of the static vulnerability data we are using

uaclient/data_types.py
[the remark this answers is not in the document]
lucasmoura: Great catch and I can confirm that we have the `3` situation in the data. I am using your patch then

uaclient/http/__init__.py
```diff
@@ -327,6 +330,27 @@ def _parse_https_proxy(https_proxy) -> Optional[ParseResult]:
...
+
+    https_proxy = get_configured_web_proxy().get("https")
+    if should_use_pycurl(https_proxy, url):
+        resp = _readurl_pycurl_https_in_https(
+            request.Request(url),
+            timeout=timeout,
+            https_proxy=https_proxy,
+        )
+        decompressor = lzma.LZMADecompressor()
+        return decompressor.decompress(resp.body)  # type: ignore
```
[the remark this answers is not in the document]
lucasmoura: Indeed, I have compared both approaches and I don't see any speedup by creating the decompressor instance. Changing to your suggestion then

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,325 @@
...
+            self._get_cache_published_date_path(series), last_published_date
+        )
+
+    def _parse_published_date(self, published_date: str):
+        format_str = "%a, %d %b %Y %H:%M:%S GMT"
+        return datetime.strptime(published_date, format_str)
+
+    def _get_published_date(self, data_url):
+        resp = readurl(url=data_url, method="HEAD")
+        return resp.headers["last-modified"]
```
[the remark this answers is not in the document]
lucasmoura: I think they need to be separated, since the data we parse can come from two distinct places (cache or HEAD request)

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,325 @@
...
+    PKG_RE = re.compile(r"^(?P<pkg>[\w\-\.\+]+)(:\w+)?\s+(?P<version>.+)$")
+
+    @staticmethod
+    def valid(manifest_file: str):
+        with open(manifest_file, "r") as f:
+            for line in f.readlines():
+                if not ProManifestSourcePackage.PKG_RE.match(line):
+                    return False
+
+        return True
```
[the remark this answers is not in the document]
lucasmoura: The problem here is that we would then need to know beforehand the manifest file pattern we expect. Today, this function exist so we can search for all of the manifest formats we support until we find one that matches.

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,325 @@
...
+
+        with open(manifest_file, "r") as f:
+            for line in f.readlines():
+                re_match = ProManifestSourcePackage.PKG_RE.match(line)
+                if re_match:
+                    match_groups = re_match.groupdict()
+                    pkg = match_groups["pkg"]
+
+                    if pkg == "snap":
+                        continue
```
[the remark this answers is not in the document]
lucasmoura: Yes, they should match due to optional group we have on the middle of the regex:
`^(?P<pkg>[\w\-\.\+]+)(:\w+)?\s+(?P<version>.+)$`

And we also have tests with snap values on it

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
lucasmoura: It seems that we do, but even though I will talk with the Security team about this, I still think we should keep this just to be safe in the future

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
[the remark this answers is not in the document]
orndorffgrant: Cool okay - I think it's actually the `.+` that will capture the additional columns

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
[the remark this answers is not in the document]
orndorffgrant: Ah so in the future this would return something other than True/False? Should it be named something like `detect_manifest_format` then and return an enum?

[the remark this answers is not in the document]
lucasmoura: No, the idea is that we will support multiple manifest files and this is the function we will use to see if the provide manifest matches one of the supported ones. But let me know if you think we should change it

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,352 @@
...
+
+                        vulnerabilities[vuln_name]["affected_packages"].append(
+                            {
+                                "name": pkg_name,
+                                "current_version": pkg_version,
+                                "fix_version": fix_version,
+                                "status": vuln_status,
+                                "fix_available_from": pocket,
+                            }
+                        )
```
lucasmoura, line 350: @panlinux @basak @dodys please take a look on the logic where we say if a CVE affects the system or not. I just want to double check if I am not missing anything

uaclient/api/u/pro/packages/updates_with_cves/v1.py
```diff
@@ -0,0 +1,353 @@
...
+
+            for cve_name, cve_info in pkg_cves.items():
+                cve_fixed_version = cve_info["source_fixed_version"]
+                if not cve_fixed_version:
+                    continue
+
+                if version_compare(current_pkg_version, cve_fixed_version) < 0:
+                    if version_compare(update_version, cve_fixed_version) >= 0:
+                        related_cves.append(cve_name)
+                        all_cves.add(cve_name)
```
lucasmoura, line 259: @panlinux @basak @dodys please take a look on the logic where we say if a CVE will be fixed by a package upgrade. I just want to be sure I am not missing anything here

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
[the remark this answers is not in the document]
orndorffgrant: Okay, but if all this function does is say "Yes this matches one of the supported formats but I won't tell you which one", then the function isn't very helpful, because `parse` will have to try to parse all of the formats anyway. We could accomplish the same thing directly in `parse`:
Try format 1 - error? move on to format 2 - error? move on..
If trying to parse all formats fails, then actually raise the error.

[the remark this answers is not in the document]
orndorffgrant: Ah sorry, I lost context between comments, writing a new one...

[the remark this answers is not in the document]
orndorffgrant: We don't need the `valid` method if we're willing to change `parse` and the calling code a bit.

Here is semi-pseudocode

on the manifest parsing class
```python
def parse():
    try:
        parse_format()
    except Exception:
        raise Unsupported
```

the calling code that tries all manifest formats
```python
        manifest_pkgs = None
        for manifest_parser_cls in self.SUPPORTED_MANIFESTS:
            try:
                manifest_pkgs = manifest_parser_cls.parse(self.manifest_file)
                break
            except Unsupported:
                pass
        if manifest_pkgs is None:
            raise UnsupportedManifestFile
```

features/files/security_issues_xenial.xz
basak, line 1: Is it worth considering storing this file uncompressed, and dynamically compressing it when needed for tests? git will handle compression anyway, and this way if the input needs to change then normal diffs and merges will work.

Also, with the xz-utils thing as an example, avoiding storage of binary blobs is (IMHO) generally preferable.

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -0,0 +1,352 @@
...
+                            .get("binary_packages", {})
+                            .get(pkg_name, "")
+                        )
+                    except KeyError:
+                        # There is bug in the data where some sources are
+                        # not present. The Security team is already aware
+                        # of this issue and they are handling it
+                        continue
+
+                    if apt.version_compare(fix_version, pkg_version) > 0:
```
basak, line 335: I think you're comparing the installed binary package version against the source version that a CVE is reported to be fixed in. But these don't necessarily match. I think the correct logic here would be to check against the source version of each installed binary package, but I need to consider this in more detail to ensure that this would always be correct.

I think that https://www.debian.org/doc/debian-policy/ch-controlfields.html#source is saying that if the binary and source versions do not match, then there will be a field of the form `Source: foo (version)` in the binary package metadata. So the logic should be to look for that, and if the field doesn't have brackets or doesn't exist, then use binary package version.

basak (commented):
I'm not going to be able to review this properly until Wednesday next week, but noticed something I think might be useful to mention early. Note that it isn't fully researched and my comment could be wrong or incomplete!

I'll do a full review when I can next week.

features/files/security_issues_xenial.xz
lucasmoura: That's a great point. I will make this change

uaclient/api/u/pro/packages/updates_with_cves/v1.py
dodys: @lucasmoura 
`current_pkg_version` seems to be storing the binary pkg version right? Not the source pkg version
And then `cve_fixed_version` is storing the `source_fixed_version`.
If that is indeed what it is doing, then you will face issues, as binary pkgs might have a different version that its source. You might want to either compare source version against source version or compare binary version against binary version.

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
lucasmoura: Thanks for looking into this @basak, but I am actually doing the comparison exactly like you suggested. If you take a look at the security data, you can see that we the following pattern:

```
      "source_versions": {
        "0.0.25b-1": {
          "binary_packages": {
            "pkg1": "0.0.25b-2"
          },
          "pocket": "release"
        },
      },      
      "cves": {
        "CVE-2022-40281": {
          "source_fixed_version": "0.0.25b-1",
          "status": "vulnerable"
        }
      }
    },
 ```
 
 What I am doing to compare the affected packages given a CVE is then:
 
 1) Get the  `source_fixed_version` from the cve (in this example, `0.0.25b-1`)
 2) Use the `source_fixed_version`  to find the right source object on the `source_versions` object
 3) Get the binary package version from the `source_versions` entry related to the `source_fixed_version`
 4) Compare the binary package version from `source_versions` with the version installed on the system.
 
 Hopefully, this should address the mismatch problem you mentioned. But let me know if you disagree

File: uaclient/api/u/pro/security/vulnerabilities/_common/v1.py (hunk 14 above)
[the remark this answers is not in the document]
Comment (lucasmoura) [reply uaclient/api/u/pro/security/vulnerabilities/_common/v1.py]: Indeed, I think that works. Let me update the code

File: uaclient/api/u/pro/security/vulnerabilities/cve/v1.py (hunk 3 above)
Comment (lucasmoura) [reply uaclient/api/u/pro/security/vulnerabilities/cve/v1.py:229]: Done

File: uaclient/api/u/pro/security/vulnerabilities/cve/v1.py (hunk 5 above)
Comment (lucasmoura) [reply uaclient/api/u/pro/security/vulnerabilities/cve/v1.py:250]: Fixed as well

**Comment by lucasmoura** [comment]
@orndorffgrant updated

File: uaclient/api/u/pro/packages/updates_with_cves/v1.py (hunk 17 above)
Comment (lucasmoura) [reply uaclient/api/u/pro/packages/updates_with_cves/v1.py]: @dodys fixed

File: uaclient/http/__init__.py [hunk 19]
```diff
@@ -230,6 +232,7 @@ def _readurl_pycurl_https_in_https(
     req: request.Request,
     timeout: Optional[int] = None,
     https_proxy: Optional[str] = None,
+    decode_response: bool = True,
```
orndorffgrant, line 235: I think this isn't used

uaclient/http/__init__.py
```diff
@@ -362,10 +385,12 @@ def readurl(
     else:
         resp = _readurl_urllib(req, timeout=timeout)
 
+    decoded_body = resp.body.decode("utf-8")
```
orndorffgrant, line 388: Let's just ignore errors in this decode call

orndorffgrant (approved):
Pre-emptively approving, just needs conflicts resolved and we can let CI run and then merge.

PR #3262, opened 2024-08-12
api: u.pro.services.list.v1

dheyay:
## Why is this needed?

Adds the following api endpoint:
`u.pro.services.list.v1`

## Test Steps

---


features/api_services_list.feature
```diff
@@ -0,0 +1,235 @@
...
+    Examples: ubuntu release
+      | release | machine_type  |
+      | xenial  | lxd-container |
+      | bionic  | lxd-container |
+      | focal   | lxd-container |
+      | jammy   | lxd-container |
+      | noble   | lxd-container |
+
+  @uses.config.contract_token
+  Scenario Outline: Get services list when unattached
```
lucasmoura, line 120: s/unattached/attached

lucasmoura: Additionally, I think we can consolidate both tests, just for performance reasons

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,217 @@
...
+            desc=ent.description,
+            available=available,
+            entitled=None,
+            status=None,
+        )
+        services_list.append(service)
+    return ServiceListResult(services=services_list)
+
+
+def _get_service_status(cfg, ent, inapplicable_resources):
```
lucasmoura, line 134: I know this problem was not introduced in this PR, but I think an entitlement should be able to tell its status. We have logic on different places that check if a service is enabled. I think we should consolidate then on the entitlement class. What do you think @orndorffgrant @renanrodrigo ?

renanrodrigo: agreed, but in a separate changeset imho

features/api_services_list.feature
```diff
@@ -0,0 +1,198 @@
+Feature: Services list api
+
+  @uses.config.contract_token
+  Scenario Outline: Get services list when unattached/attached
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I run `pro api u.pro.services.list.v1` with sudo
```
lucasmoura, line 6: Let's also guarantee that the API runs fine when running as `non-root`.
Let's add that to the attached scenario as well

features/api_services_list.feature
```diff
@@ -0,0 +1,198 @@
+Feature: Services list api
+
+  @uses.config.contract_token
+  Scenario Outline: Get services list when unattached/attached
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
```
lucasmoura, line 5: Let's also include this API under the main api test where we import all other api endpoints:

lucasmoura (commented):
@dheyay Let's also add a description to the commit explaining what the API does

dheyay:
Updated @lucasmoura

features/api_services_list.feature
```diff
@@ -0,0 +1,395 @@
+Feature: Services list api
+
+  @uses.config.contract_token
+  Scenario Outline: Get services list when unattached/attached as non-root
```
renanrodrigo, line 4: Given that the test is expecting the api to work exactly the same, it can be simplified into a single scenario, where you add a line before checking the actual text. We have something like
`And I verify root and non-root ``<command>`` calls have the same output`
So you DRY

features/api_services_list.feature
```diff
@@ -0,0 +1,395 @@
...
+      }
+      """
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | xenial  | lxd-container |
+      | bionic  | lxd-container |
+      | focal   | lxd-container |
+      | jammy   | lxd-container |
+      | noble   | lxd-container |
```
renanrodrigo, line 198: Please add oracular to the list so we can test this works on interim releases too

features/api_services_list.feature
```diff
@@ -0,0 +1,395 @@
...
+              "available": true,
+              "description": "Expanded Security Maintenance for Infrastructure",
+              "entitled": true,
+              "name": "esm-infra"
+            },
+            {
+              "available": false,
+              "description": "NIST-certified FIPS crypto packages",
+              "entitled": true,
+              "name": "fips"
```
renanrodrigo, line 133: Service availability may vary between releases.
I would say the specific values here could use columns in the `Examples` table below, where you set them.
This test fails on `focal`, for instance, where FIPS is available.

features/api_services_list.feature
```diff
@@ -0,0 +1,395 @@
...
+              "available": false,
+              "description": "All Updates for the Robot Operating System",
+              "entitled": true,
+              "name": "ros-updates"
+            },
+            {
+              "available": false,
+              "description": "Security compliance and audit tools",
+              "entitled": true,
+              "name": "usg"
```
renanrodrigo, line 181: this is sorted differently for `attached` and `unattached` - is there a reason?

features/api_services_list.feature
```diff
@@ -0,0 +1,395 @@
+Feature: Services list api
+
+  @uses.config.contract_token
+  Scenario Outline: Get services list when unattached/attached as non-root
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I run `pro api u.pro.services.list.v1` as non-root
+    Then API data field output matches regexp:
```
renanrodrigo, line 7: why the regex?
Could we go `Then API data field output is:` here?

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,197 @@
...
+            name=resource.get("presentedAs", resource["name"]),
+            desc=ent.description,
+            available=available,
+            entitled=None,
+        )
+        services_list.append(service)
+    return ServiceListResult(services=services_list)
+
+
+def _get_service_status(cfg, ent, inapplicable_resources):
```
renanrodrigo, line 124: What bothers me is that this function is doing more than it tells you
It doesn't return the status, but rather the full ServiceInfo object.
Making it clearer would help us.

renanrodrigo: On top of that, type hints would be nice on all functions so we see this clearer

renanrodrigo:
> - [x]  No unit test changes are necessary for this change

Well you are technically adding a unit. Why wouldn't we need the test?

features/api_services_list.feature
dheyay: I am the reason :sob:

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,197 @@
...
+    for resource in resources:
+        try:
+            ent = entitlement_factory(cfg=cfg, name=resource.get("name", ""))
+        except exceptions.EntitlementNotFoundError:
+            continue
+
+        services_list.append(
+            _get_service_status(cfg, ent, inapplicable_resources)
+        )
+        services_list.sort(key=lambda x: x.name)
```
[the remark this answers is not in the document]
dheyay: The serviceInfo class has the property `name` which is equivalent to the `presentation_name`

uaclient/api/u/pro/services/list/v1.py
dheyay: Will rename it to something like `_get_service_info` instead which is more accurate for it's behavior. (ofc also adding type hints)

dheyay:
Updated @renanrodrigo @lucasmoura

features/api_services_list.feature
```diff
@@ -0,0 +1,184 @@
...
+      {"name": "fips-updates", "available": false, "entitled": null}
+      {"name": "landscape", "available": false, "entitled": null}
+      {"name": "livepatch", "available": false, "entitled": null}
+      {"name": "realtime-kernel", "available": false, "entitled": null}
+      {"name": "ros", "available": false, "entitled": null}
+      {"name": "ros-updates", "available": false, "entitled": null}
+      """
+    When I attach `contract_token` with sudo
+    And I verify root and non-root `pro api u.pro.services.list.v1` calls have the same output
+    When I verify root and non-root `pro api u.pro.services.list.v1` calls have the same output
```
lucasmoura, line 92: I think we can remove one of those

features/api_services_list.feature
```diff
@@ -0,0 +1,184 @@
+Feature: Services list api
+
+  @uses.config.contract_token
+  Scenario Outline: Get services list when unattached/attached as non-root
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I verify root and non-root `pro api u.pro.services.list.v1` calls have the same output
+    When I run `pro api u.pro.services.list.v1` as non-root
+    Then stdout is a json matching the `api_response` schema
```
lucasmoura, line 8: Shouldn't we be using `api_service_list_response` here ?

dheyay: Fixed, also checking the correct schema now

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,201 @@
...
+    inapplicable_resources = {
+        resource["name"]: resource.get("description")
+        for resource in sorted(resources, key=lambda x: x.get("name", ""))
+        if not resource.get("available")
+    }
+    for resource in resources:
+        try:
+            ent = entitlement_factory(cfg=cfg, name=resource.get("name", ""))
+        except exceptions.EntitlementNotFoundError:
+            continue
```
lucasmoura, line 92: shouldn't we have the same log message here for an unknown service ?

lucasmoura (approved):

uaclient/api/u/pro/services/list/v1.py
dheyay: Yes, updated, the message should be there for both attached and unattached

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,206 @@
...
+endpoint = APIEndpoint(
+    version="v1",
+    name="ServicesList",
+    fn=_get_service_list,
+    options_cls=None,
+)
+
+_doc = {
+    "introduced_in": "35",
+    "requires_network": False,
```
orndorffgrant, line 160:
```suggestion
    "requires_network": True,
```

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,206 @@
...
+    "example_python": """
+from uaclient.api.u.pro.services.list.v1 import get_service_list
+
+result = get_service_list()
+""",  # noqa: E501
+    "result_class": ServiceListResult,
+    "exceptions": [],
```
orndorffgrant, line 167: At least  `ContractAPIError` should be here since we talk to that. Maybe look for others as well?

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,206 @@
...
+class ServiceListResult(DataObject, AdditionalInfo):
+    fields = [
+        Field("services", data_list(ServiceInfo), doc="List of services"),
+    ]
+
+    def __init__(self, *, services: List[ServiceInfo]):
+        self.services = services
+
+
+def get_service_list() -> ServiceListResult:
```
orndorffgrant, line 67: generally the public function name has matched the final word in the endpoint string. So to follow that pattern, this would be `list()`

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,206 @@
...
+def _get_service_info(
+    cfg: UAConfig, ent: UAEntitlement, inapplicable_resources: Dict
+) -> ServiceInfo:
+    contract_status = ent.contract_status()
+    entitled = None
+    if contract_status == ContractStatus.UNENTITLED:
+        entitled = False
+    elif contract_status == ContractStatus.ENTITLED:
+        entitled = True
+    available = True if ent.name not in inapplicable_resources else False
```
orndorffgrant, line 140: I think we should generally be using `ent.applicability_status()` for the value of `available` - right @lucasmoura @renanrodrigo ?

lucasmoura: I think @dheyay is just replicating the logic we have on status here. However, I think it is worth testing out replacing this logic with `applicability_status`

dheyay: It works with the `applicability_status` logic and does not need the additional inapplicable_resources. Updated the implementation and the endpoint/function names @orndorffgrant

renanrodrigo (approved):
So I still think we could have a couple unit tests for the basic functionality here. The API implementation file has 0% coverage
I'm fine if we merge this without the test for now though, as we have precedents - we already have API files at 0%.

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,212 @@
...
+    entitled = None
+    if contract_status == ContractStatus.UNENTITLED:
+        entitled = False
+    elif contract_status == ContractStatus.ENTITLED:
+        entitled = True
+    available = (
+        True
+        if ent.applicability_status()[0] == ApplicabilityStatus.APPLICABLE
+        else False
+    )
```
orndorffgrant, line 141: No need for an if-else
```suggestion
    available = ent.applicability_status()[0] == ApplicabilityStatus.APPLICABLE
```

uaclient/api/u/pro/services/list/v1.py
```diff
@@ -0,0 +1,212 @@
...
+def _get_service_info(
+    cfg: UAConfig,
+    ent: UAEntitlement,
+) -> ServiceInfo:
+    contract_status = ent.contract_status()
+    entitled = None
+    if contract_status == ContractStatus.UNENTITLED:
+        entitled = False
+    elif contract_status == ContractStatus.ENTITLED:
+        entitled = True
```
orndorffgrant, line 136: I think can be simplified in the same way
```suggestion
    entitled = ent.contract_status() == ContractState.ENTITLED
```

dheyay:
Updated with a unit test @orndorffgrant @renanrodrigo

features/api/api.feature
```diff
@@ -17,6 +17,7 @@ Feature: Client behaviour for the API endpoints
     When I run `python3 -c "from uaclient.api.u.pro.security.status.livepatch_cves.v1 import livepatch_cves"` as non-root
     When I run `python3 -c "from uaclient.api.u.pro.security.status.reboot_required.v1 import reboot_required"` as non-root
     When I run `python3 -c "from uaclient.api.u.pro.services.dependencies.v1 import dependencies"` as non-root
+    When I run `python3 -c "from uaclient.api.u.pro.services.list.v1 import get_service_list"` as non-root
```
orndorffgrant, line 20: This needs to be updated to reflect the new function name
```suggestion
    When I run `python3 -c "from uaclient.api.u.pro.services.list.v1 import list"` as non-root
```

orndorffgrant (approved):
One minor update needed - preemptively approving

PR #3264, opened 2024-08-13
api: u.pro.token_info.v1

dheyay:
## Why is this needed?

Adds the token simulation endpoint:
`u.pro.token_info.v1`

## Test Steps

---

### DIFF
--- a/uaclient/api/api.py
+++ b/uaclient/api/api.py
@@ -33,6 +33,7 @@ VALID_ENDPOINTS = [
     "u.pro.services.list.v1",
     "u.pro.status.enabled_services.v1",
     "u.pro.status.is_attached.v1",
+    "u.pro.token_info.v1",
     "u.apt_news.current_news.v1",
     "u.security.package_manifest.v1",
     "u.unattended_upgrades.status.v1",
--- a/uaclient/api/tests/test_api_u_pro_token_info_v1.py
+++ b/uaclient/api/tests/test_api_u_pro_token_info_v1.py
@@ -1,0 +1,77 @@
+import mock
+
+from uaclient import util
+from uaclient.api.u.pro.token_info.v1 import (
+    AccountInfo,
+    ContractInfo,
+    ServiceInfo,
+    TokenInfoOptions,
+    TokenInfoResult,
+    _get_token_info,
+)
+
+RESPONSE_CONTRACT_INFO = {
+    "accountInfo": {
+        "createdAt": util.parse_rfc3339_date("2019-06-14T06:45:50Z"),
+        "id": "some_id",
+        "name": "Name",
+        "type": "paid",
+    },
+    "contractInfo": {
+        "createdAt": util.parse_rfc3339_date("2021-05-21T20:00:53Z"),
+        "createdBy": "someone",
+        "effectiveTo": util.parse_rfc3339_date("9999-12-31T00:00:00Z"),
+        "id": "some_id",
+        "name": "Name",
+        "products": ["uai-essential-virtual"],
+        "resourceEntitlements": [
+            {
+                "type": "esm-infra",
+                "entitled": True,
+                "obligations": {"enableByDefault": True},
+            },
+            {
+                "type": "livepatch",
+                "entitled": True,
+                "obligations": {"enableByDefault": False},
+            },
+        ],
+    },
+}
+
+RESPONSE_AVAILABLE_SERVICES = [
+    {"name": "livepatch", "available": True},
+]
+
+
+@mock.patch(
+    "uaclient.api.u.pro.token_info.v1.get_available_resources",
+    return_value=RESPONSE_AVAILABLE_SERVICES,
+)
+@mock.patch(
+    "uaclient.api.u.pro.token_info.v1.get_contract_information",
+    return_value=RESPONSE_CONTRACT_INFO,
+)
+class TestTokenInfo:
+
+    def test_token_info_output(
+        self, m_get_contract_information, m_get_available_resources, FakeConfig
+    ):
+        cfg = FakeConfig()
+        token_info = _get_token_info(
+            options=TokenInfoOptions(token="contract_token"), cfg=cfg
+        )
+        expected_info = TokenInfoResult(
+            account=AccountInfo(id="some_id", name="Name"),
+            contract=ContractInfo(id="some_id", name="Name"),
+            services=[
+                ServiceInfo(
+                    name="livepatch",
+                    description="Canonical Livepatch service",
+                    entitled=True,
+                    auto_enabled=False,
+                    available=True,
+                ),
+            ],
+        )
+        assert token_info == expected_info
--- a/uaclient/api/tests/test_api_u_pro_token_info_v1.py
+++ b/uaclient/api/tests/test_api_u_pro_token_info_v1.py
@@ -1,6 +1,9 @@
-import mock
+import copy
 
-from uaclient import util
+import mock
+import pytest
+
+from uaclient import exceptions, util
 from uaclient.api.u.pro.token_info.v1 import (
     AccountInfo,
     ContractInfo,
@@ -44,16 +47,15 @@ RESPONSE_AVAILABLE_SERVICES = [
 ]
 
 
-@mock.patch(
-    "uaclient.api.u.pro.token_info.v1.get_available_resources",
-    return_value=RESPONSE_AVAILABLE_SERVICES,
-)
-@mock.patch(
-    "uaclient.api.u.pro.token_info.v1.get_contract_information",
-    return_value=RESPONSE_CONTRACT_INFO,
-)
 class TestTokenInfo:
-
+    @mock.patch(
+        "uaclient.api.u.pro.token_info.v1.get_available_resources",
+        return_value=RESPONSE_AVAILABLE_SERVICES,
+    )
+    @mock.patch(
+        "uaclient.api.u.pro.token_info.v1.get_contract_information",
+        return_value=RESPONSE_CONTRACT_INFO,
+    )
     def test_token_info_output(
         self, m_get_contract_information, m_get_available_resources, FakeConfig
     ):
@@ -63,7 +65,12 @@ class TestTokenInfo:
         )
         expected_info = TokenInfoResult(
             account=AccountInfo(id="some_id", name="Name"),
-            contract=ContractInfo(id="some_id", name="Name"),
+            contract=ContractInfo(
+                id="some_id",
+                name="Name",
+                effective=None,
+                expires=util.parse_rfc3339_date("9999-12-31T00:00:00Z"),
+            ),
             services=[
                 ServiceInfo(
                     name="livepatch",
@@ -75,3 +82,57 @@ class TestTokenInfo:
             ],
         )
         assert token_info == expected_info
+
+    @pytest.mark.parametrize(
+        "expected_error,expected_err_message,contract_field,date_value",
+        (
+            (
+                exceptions.TokenForbiddenExpired,
+                (
+                    'Contract "some_id" expired on December 31, 2019\n'
+                    "Visit https://ubuntu.com/pro/dashboard to manage "
+                    "contract tokens."
+                ),
+                "effectiveTo",
+                util.parse_rfc3339_date("2019-12-31T00:00:00Z"),
+            ),
+            (
+                exceptions.TokenForbiddenNotYet,
+                (
+                    'Contract "some_id" is not effective until December 31, '
+                    "9999\n"
+                    "Visit https://ubuntu.com/pro/dashboard to manage "
+                    "contract tokens."
+                ),
+                "effectiveFrom",
+                util.parse_rfc3339_date("9999-12-31T00:00:00Z"),
+            ),
+        ),
+    )
+    @mock.patch(
+        "uaclient.api.u.pro.token_info.v1.get_available_resources",
+        return_value=RESPONSE_AVAILABLE_SERVICES,
+    )
+    @mock.patch("uaclient.api.u.pro.token_info.v1.get_contract_information")
+    def test_attach_forbidden_error(
+        self,
+        m_get_contract_information,
+        m_get_available_resources,
+        expected_error,
+        expected_err_message,
+        contract_field,
+        date_value,
+        capsys,
+        FakeConfig,
+    ):
+        exp_contract = copy.deepcopy(RESPONSE_CONTRACT_INFO)
+        exp_contract["contractInfo"][contract_field] = date_value
+        m_get_contract_information.return_value = exp_contract
+        cfg = FakeConfig()
+        try:
+            _get_token_info(
+                options=TokenInfoOptions(token="contract_token"), cfg=cfg
+            )
+        except expected_error as e:
+            err_message = str(e)
+            assert err_message == expected_err_message
--- a/uaclient/api/u/pro/token_info/v1.py
+++ b/uaclient/api/u/pro/token_info/v1.py
@@ -1,5 +1,6 @@
 import logging
-from typing import Any, Dict, List
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional
 
 from uaclient import exceptions, util
 from uaclient.api.api import APIEndpoint
@@ -9,10 +10,12 @@ from uaclient.contract import get_available_resources, get_contract_information
 from uaclient.data_types import (
     BoolDataValue,
     DataObject,
+    DatetimeDataValue,
     Field,
     StringDataValue,
     data_list,
 )
+from uaclient.defaults import ATTACH_FAIL_DATE_FORMAT
 from uaclient.entitlements import entitlement_factory
 
 LOG = logging.getLogger(util.replace_top_level_logger_name(__name__))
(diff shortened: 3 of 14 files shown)


uaclient/api/api.py
```diff
@@ -90,6 +91,7 @@ def _process_data(
 
         kwargs[k] = v
 
+    print(kwargs, warnings)
```
lucasmoura, line 94: lets remove this line

uaclient/api/u/pro/token_info/v1.py
```diff
@@ -0,0 +1,237 @@
...
+        Field(
+            "token", StringDataValue, doc="Token to be used for the request"
+        ),
+    ]
+
+    def __init__(self, token: str):
+        self.token = token
+
+
+class accountInfo(DataObject):
```
lucasmoura, line 33: I think we need to rename the classes so that first letter is an `upcase`

dheyay: Updated @lucasmoura

features/api_token_info.feature
```diff
@@ -0,0 +1,151 @@
...
+            "name": "UA Client Test"
+          },
+          "contract": {
+            "id": ".*",
+            "name": "server-team-ua-client-ci-uaa"
+          },
+          "services": [
+            {
+              "auto_enabled": "yes",
+              "available": "no",
```
orndorffgrant, line 22: (Applies to service list too)
`available` should be a bool

features/api_token_info.feature
```diff
@@ -0,0 +1,151 @@
...
+              "description": "Scalable Android in the cloud",
+              "entitled": "yes",
```
orndorffgrant, line 24: (Applies to service list too)
`entitled` should be a bool

features/api_token_info.feature
```diff
@@ -0,0 +1,151 @@
...
+            "id": ".*",
+            "name": "UA Client Test"
+          },
+          "contract": {
+            "id": ".*",
+            "name": "server-team-ua-client-ci-uaa"
+          },
+          "services": [
+            {
+              "auto_enabled": "yes",
```
orndorffgrant, line 21: `auto_enabled` should be a bool

uaclient/api/api.py
```diff
@@ -19,6 +19,7 @@
     "u.pro.attach.magic.wait.v1",
     "u.pro.attach.token.full_token_attach.v1",
     "u.pro.detach.v1",
+    "u.pro.token_info.v1",
```
orndorffgrant, line 22: Let's keep this in alphabetical order (with the exception of `version` at the top and all non `pro` things at the bottom)

uaclient/api/u/pro/token_info/v1.py
```diff
@@ -0,0 +1,246 @@
...
+        account: AccountInfo,
+        contract: ContractInfo,
+        services: List[ServiceInfo],
+    ):
+        self.account = account
+        self.contract = contract
+        self.services = services
+
+
+def _get_entitlement_information_bool(
```
lucasmoura, line 96: Why do we have the `bool` in the name of the function ?

uaclient/api/u/pro/token_info/v1.py
```diff
@@ -0,0 +1,246 @@
...
+                ),
+            }
+    return {"entitled": False, "auto_enabled": False}
+
+
+def get_token_info(options: TokenInfoOptions) -> TokenInfoResult:
+    return _get_token_info(options=options, cfg=UAConfig())
+
+
+def _get_token_info(
```
lucasmoura, line 115: Can we reuse this API for `status --simulate-with-token` ?
That is because the logic here is really similar to what we have on status and that would avoid code duplication

features/api_token_info.feature
```diff
@@ -0,0 +1,151 @@
+Feature: api.u.pro.token_info.v1
+
+  @uses.config.contract_token
+  Scenario Outline: Get token info from api
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I run `pro api u.pro.token_info.v1 --args token=$behave_var{contract_token}` with sudo
```
lucasmoura, line 6: Let`s either turn this into a `non-root` test or add a new case where we run it as `non-root`

features/api_token_info.feature
```diff
@@ -0,0 +1,151 @@
+Feature: api.u.pro.token_info.v1
+
+  @uses.config.contract_token
+  Scenario Outline: Get token info from api
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
```
lucasmoura, line 5: Let's also import this API in this test:

uaclient/api/u/pro/token_info/v1.py
dheyay: There is currently already a function with the same name in the `status.py` file which returns the information result in strings. I have a better idea to deal with this, maybe we can just have one function instead of two.

uaclient/api/u/pro/token_info/v1.py
dheyay: I was looking to do that initially but because of how the difference between the structure of we currently return `simulated_status` and how the api returns it, we might break the way we generate status.json.

Integrating the api is also diffifcult since we will be using the same function calls either ways to keep the current structure and everything will just end up being called twice. 
Is there a better way to handle this?

features/api_token_info.feature
renanrodrigo: [same as here](https://github.com/canonical/ubuntu-pro-client/pull/3262/files#r1724896035) - we do have a step to check `root` and `non-root` are the same

features/api_token_info.feature
```diff
@@ -0,0 +1,151 @@
...
+              "description": "NIST-certified FIPS crypto packages",
+              "entitled": true,
+              "name": "fips"
+            },
+            {
+              "auto_enabled": false,
+              "available": false,
+              "description": "Preview of FIPS crypto packages undergoing certification with NIST",
+              "entitled": true,
+              "name": "fips-preview"
```
renanrodrigo, line 67: [similar to this one](https://github.com/canonical/ubuntu-pro-client/pull/3262/files#r1724908133) - availability varies between releases.

features/api_token_info.feature
```diff
@@ -0,0 +1,151 @@
...
+      }
+      """
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | xenial  | lxd-container |
+      | bionic  | lxd-container |
+      | focal   | lxd-container |
+      | jammy   | lxd-container |
+      | noble   | lxd-container |
```
renanrodrigo, line 126: Should we test `oracular`? I'd say yes

uaclient/api/u/pro/token_info/v1.py
renanrodrigo: +1 different function. Given you are not importing the `status.py` one here there is no name conflict, just go with the simplest possible.
It does bother me that a function caled `get_..._bool` returns a `Dict` :|

uaclient/api/tests/test_api_u_pro_token_info_v1.py
```diff
@@ -0,0 +1,48 @@
...
+
+from uaclient.api.u.pro.token_info.v1 import (
+    AccountInfo,
+    ContractInfo,
+    ServiceInfo,
+    TokenInfoOptions,
+    TokenInfoResult,
+    _get_token_info,
+)
+from uaclient.cli.tests.test_cli_status import RESPONSE_CONTRACT_INFO
```
renanrodrigo, line 11: eeek importing constants between tests smells
keep it if people don't mind - I mind a little bit

uaclient/api/u/pro/token_info/v1.py
renanrodrigo: I would say this could be reused there, yes.

One thing that I am missing here is the expiration check - in the simulated status, we error out when the contract is expired or not yet valid `status.py:465-493`. The spec states that this situation should cause the API to return an error too.

Once that is in place, calling the API in status will:
- delegate the token checks to the API code
- return the services list in a state we can use to build the outputs

I understand it is not straightforward as other endpoints would be, but I do think we should do our best to reuse things from here.

uaclient/api/tests/test_api_u_pro_token_info_v1.py
dheyay: Yeah, we can not import the constant.

uaclient/api/u/pro/token_info/v1.py
dheyay: I have changed the function names to be more appropriate to their use cases. something like
`user_facing_entitlement_information` and `application_entitlement_information`

uaclient/api/u/pro/token_info/v1.py
dheyay: The part I am a little skeptical about is the gain is not much here given the token_info has less information than needed for the `simulate_status`. Even if we delegate the token checks, we would still require the `get_contract_information()` output either ways since `token_info` does not have all the contract and account information we need. 

A better way to handle this is maybe ->

- Use a different function in the api file to get the contract and account information, use that function in both the api and the `simulate_status` function OR add a flag to api function to return `all` information which defaults to False (just the `_get_token_info` function)
- Move the `expiration` and contract check to the API and use the API in `simulate_status`

This could completely replace the `simulate_status` function code and we can just the api there. What do you guys think about this? @lucasmoura @renanrodrigo

dheyay: Adding to the above comment -
Currently the fields that are included in the `simualte_status` response which are not present in the api are:
```
contract: {
  "created_at",
  "products"
}
account: {
  "created_at",
  "external_account_ids"
}
"expires"
"effectiveTo"
```

We do not want to query the `subscription` api with the token, as @renanrodrigo said that should just work when asking for subscription information.

uaclient/api/u/pro/token_info/v1.py
dheyay: updated

features/api_token_info.feature
```diff
@@ -0,0 +1,125 @@
...
+    Examples: ubuntu release
+      | release  | machine_type  |
+      | xenial   | lxd-container |
+      | bionic   | lxd-container |
+      | focal    | lxd-container |
+      | jammy    | lxd-container |
+      | noble    | lxd-container |
+      | oracular | lxd-container |
+
+  Scenario Outline: Error when no token is provided
```
lucasmoura, line 102: I don't think we need a different container for this, this can be the first check on the previous scenario

uaclient/api/u/pro/token_info/v1.py
lucasmoura: @dheyay I agree that some of those fields are not necessary for this particular API, but maybe including the `expires` and `effectiveTo` are actually a good idea, since I believe users might be interested in that information.

But I also agree that then, this new API is not enough to be used by `--simulate-with-token`

lucasmoura:
@dheyay Looking at CI, I am seeing that the `all API endpoints can be imported individually` is failing everywhere, while `Get token info from api ` is failing on Jammy

dheyay:
@lucasmoura sorry about that, fixed both tests.

uaclient/api/u/pro/token_info/v1.py
dheyay: @lucasmoura @renanrodrigo @orndorffgrant Should we add `expires` and `effectiveTo` to the token_info api

uaclient/api/u/pro/token_info/v1.py
```diff
@@ -0,0 +1,293 @@
...
+    account = AccountInfo(
+        id=account_info.get("id", ""), name=account_info.get("name", "")
+    )
+
+    # Check contract expiration
+    effective, expires = None, None
+    now = datetime.now(timezone.utc)
+    if contract_info.get("effectiveTo"):
+        expiration_datetime = contract_info.get("effectiveTo")
+        expires = expiration_datetime
```
lucasmoura, line 151: Why are we creating this variable ?

uaclient/api/u/pro/token_info/v1.py
```diff
@@ -0,0 +1,293 @@
...
+        delta = expiration_datetime - now
+        if delta.total_seconds() <= 0:
+            raise exceptions.AttachForbiddenExpired(
+                contract_id=contract_info.get("id", ""),
+                date=expiration_datetime.strftime(ATTACH_FAIL_DATE_FORMAT),
+            )
+    if contract_info.get("effectiveFrom"):
+        effective_datetime = contract_info.get("effectiveFrom")
+        effective = effective_datetime
```
lucasmoura, line 160: Same variable question here

uaclient/api/u/pro/token_info/v1.py
```diff
@@ -0,0 +1,293 @@
...
+            raise exceptions.AttachForbiddenNotYet(
+                contract_id=contract_info.get("id", ""),
+                date=effective_datetime.strftime(ATTACH_FAIL_DATE_FORMAT),
+            )
+
+    contract = ContractInfo(
+        id=contract_info.get("id", ""),
+        name=contract_info.get("name", ""),
+        effective=effective,
+        expires=expires,
```
lucasmoura, line 172: I think the API should return the datetime object directly for those fields, not just return the string representation

uaclient/api/u/pro/token_info/v1.py
dheyay: Updated with the new fields @lucasmoura

features/api_token_info.feature
```diff
@@ -0,0 +1,113 @@
+Feature: api.u.pro.token_info.v1
+
+  @uses.config.contract_token
```
lucasmoura, line 3: Let's add an integration test where we try to see an expired token. We can use the staging expired token for this.

dheyay: Apologies for the oversight, only added a unit test for this. An integration test too would make sense. Will add it

dheyay: @lucasmoura Updated with the new integration test and added error messages/exceptions specific to token_info and simulate status too

features/api_token_info.feature
```diff
@@ -143,9 +104,38 @@ Feature: api.u.pro.token_info.v1
...
+      | bionic   | lxd-container |
+      | focal    | lxd-container |
+      | jammy    | lxd-container |
+      | noble    | lxd-container |
+      | oracular | lxd-container |
+
+  @uses.config.contract_token_staging_expired
+  Scenario Outline: Get token info from api
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I run `sed -i 's/contracts.can/contracts.staging.can/' /etc/ubuntu-advantage/uaclient.conf` with sudo
```
lucasmoura, line 118: Here we can an already existing step called `I change contract to staging with sudo` to make that change

features/api_token_info.feature
```diff
@@ -143,9 +104,38 @@ Feature: api.u.pro.token_info.v1
...
+      | release  | machine_type  |
+      | xenial   | lxd-container |
+      | bionic   | lxd-container |
+      | focal    | lxd-container |
+      | jammy    | lxd-container |
+      | noble    | lxd-container |
+      | oracular | lxd-container |
+
+  @uses.config.contract_token_staging_expired
+  Scenario Outline: Get token info from api
```
lucasmoura, line 116: Let's rename this scenario to better reflect what we are testing

lucasmoura (approved):
Just a couple of nits, but LGTM.

Let's just wait to see if any unexpected errors appear on CI

features/api_token_info.feature
dheyay: Thanks a lot for the constant help with the review @lucasmoura :sweat_smile:

uaclient/messages/__init__.py
```diff
@@ -2164,6 +2164,24 @@ def __repr__(self):
...
 )
 
+E_TOKEN_FORBIDDEN_EXPIRED = FormattedNamedMessage(
+    "token-forbidden-expired",
+    t.gettext(
+        """\
+Contract "{{contract_id}}" expired on {{contract_expiry_date}}
+Visit {url} to manage contract tokens."""
+    ).format(url=urls.PRO_DASHBOARD),
+)
```
renanrodrigo, line 2174: sorry I see you have new messages now

for consistency (and to ease the translation a bit), could we make the existing ones extend those?
something like
```
E_ATTACH_FORBIDDEN_EXPIRED = FormattedNamedMessage(
    "attach-forbidden-expired",
    t.gettext( "Attach denied:")
    + "\n"
    + E_TOKEN_FORBIDDEN_EXPIRED
)
```

dheyay: That's a good suggestion. Will do.

dheyay: Updated, cleaned up commit history as well (also rebased)

orndorffgrant (approved):
Let's let CI pass and double check it - then this should be good - thanks!

renanrodrigo (approved):
No failures related to this PR.
@dheyay just rebase it again please, so we can merge

Issue #3132, opened 2024-05-27
pro security-status emits warning if ESM cache doesn't exist


lucasmoura:
**Description of the bug**

On Noble, if the `pro security-status` command runs before the ESM cache is created, we will see the following warning on stderr:

```
/usr/lib/python3/dist-packages/uaclient/apt.py:299: Warning: W:Unable to read /var/lib/ubuntu-advantage/apt-esm/etc/apt/apt.conf.d/ - DirectoryExists (2: No such file or directory)
  apt_pkg.init()
/usr/lib/python3/dist-packages/uaclient/apt.py:281: Warning: W:Unable to read /var/lib/ubuntu-advantage/apt-esm/etc/apt/preferences.d/ - DirectoryExists (2: No such file or directory)
  apt_pkg.init_system()
  ```

**Expected behavior**

This warning should not appear, even if the ESM cache doesn't exit

**To Reproduce**

Please include details on how to reproduce the bug.
1. Launch a Noble container
2. Run `pro security-status | cat`

**System information:**

- Ubuntu release:  Noble, Oracular
- Pro Client version: 32.2

PR #3267, opened 2024-08-15
Remove Mantic from the codebase+tests

renanrodrigo:
## Why is this needed?

This PR solves all of our problems because mantic is EOL and we want to remove it. To cover the gap left behind, some tests got noble and some tests got oracular versions instead.

To sort out some tests, this
Fixes: #3132

## Test Steps
All about integration tests here - every new added noble/oracular test should pass.

---

### DIFF
--- a/uaclient/upgrade_lts_contract.py
+++ b/uaclient/upgrade_lts_contract.py
@@ -42,8 +42,8 @@ current_codename_to_past_codename = {
     "bionic": "xenial",
     "focal": "bionic",
     "jammy": "focal",
-    "mantic": "lunar",
     "noble": "jammy",
+    "oracular": "noble",
 }
 
 LOG = logging.getLogger(util.replace_top_level_logger_name(__name__))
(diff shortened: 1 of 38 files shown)


renanrodrigo:
Agreed with @lucasmoura that this should only land after #3268 is merged, and we rebase on top of that

dheyay (approved):
Tentative approval since we this is dependent on the other tests PR to be merged first.
Also, I just had a question about the inclusion of `oracular` in some tests, are the tests to which the non-lts release is added, selected on some metric?

Thanks @renanrodrigo

uaclient/upgrade_lts_contract.py
```diff
@@ -42,7 +42,6 @@
     "bionic": "xenial",
     "focal": "bionic",
     "jammy": "focal",
-    "mantic": "lunar",
```
orndorffgrant, line 45: should we add oracular while we're here?

orndorffgrant (approved):
Just a suggestion that we add oracular to a couple more places where we're removing mantic - preemptively approving. Thanks for doing this one - this is tedious!

uaclient/upgrade_lts_contract.py
renanrodrigo: added

renanrodrigo:
Changes here are not related to the changeset - the same fix problems as always + azure fips

PR #3268, opened 2024-08-15
Refactor and consolidate Pro client tests

lucasmoura:
## Why is this needed?
We are now consolidating and refactoring several tests on the Pro client. The focus on this PR is to:

* Create feature file for specific Pro functionalities (i.e. cli)
* Consolidate several tests under a single scenario
* Remove tests that are not longer needed or that are covered by other tests 

## Test Steps
Guarantee that tests are still working as expected and that we have not missed any coverage by removing the selected tests

---


orndorffgrant:
This is already massive and I haven't reviewed it yet, but an idea that we might as well do now while re-organizing everything.

Instead of prefixing files with `cli_` or `api_` can we put them in folders? so `features/cli/enable.feature` for example?

features/cli/cli_detach.feature
```diff
@@ -0,0 +1,171 @@
+@uses.config.contract_token
```
renanrodrigo, line 1: If I can make a suggestion, let's get rid of this tag?

I wouldn't say we are running tests without setting a contract token anyway, now that it is way easier using the `protest.yaml` file.
And if we do, I'd say the error is pretty straightforward in telling us to add a token.
So this is not being used for filtering, and enforcing is useless. Besides that, right now it's not standardized through the codebase - here and there you will find a test that needs a token but lacks the tag.

features/cli/cli_enable.feature
```diff
@@ -0,0 +1,631 @@
...
+
+    Examples: ubuntu release
+      | release | machine_type  | valid_services                                                                                                                             |
+      | xenial  | lxd-container | anbox-cloud, cc-eal, cis, esm-apps, esm-infra, fips, fips-preview,\nfips-updates, landscape, livepatch, realtime-kernel, ros, ros-updates. |
+      | bionic  | lxd-container | anbox-cloud, cc-eal, cis, esm-apps, esm-infra, fips, fips-preview,\nfips-updates, landscape, livepatch, realtime-kernel, ros, ros-updates. |
+      | focal   | lxd-container | anbox-cloud, cc-eal, esm-apps, esm-infra, fips, fips-preview, fips-updates,\nlandscape, livepatch, realtime-kernel, ros, ros-updates, usg. |
+      | jammy   | lxd-container | anbox-cloud, cc-eal, esm-apps, esm-infra, fips, fips-preview, fips-updates,\nlandscape, livepatch, realtime-kernel, ros, ros-updates, usg. |
+      | noble   | lxd-container | anbox-cloud, cc-eal, esm-apps, esm-infra, fips, fips-preview, fips-updates,\nlandscape, livepatch, realtime-kernel, ros, ros-updates, usg. |
+
+  Scenario Outline: Attached enable of ESM services in a ubuntu machine
```
renanrodrigo, line 302: Question - should tests like this have their own `esm.feature` file, leaving this one only testing the generic enable details? (fine if not)

renanrodrigo (commented):
May I suggest that we remove the prefix of the file names inside the folders?
So instead of `features/api/api_detach.feature` we get `features/api/detach.feature`

Not important though

features/cli/cli_enable.feature
orndorffgrant: I think it probably should, but I want to get this already massive PR in and it's already a great improvement.

features/cli_refresh.feature
```diff
@@ -46,3 +46,30 @@ Feature: CLI refresh command
       | jammy   | wsl           |
       | mantic  | lxd-container |
       | noble   | lxd-container |
+
+  @wip
```
orndorffgrant: wip can be removed :)

orndorffgrant (commented):
This is fantastic :+1: :+1: :+1: I have just one request before merging

The Scenario previously labeled `Auto enable by default and attached disable of livepatch in a lxd vm` was the only test that explicitly tested our most used function: `pro attach` and check services were auto enabled (some tests _implicitly_ expect this, but this test was actually checking for the correct auto-enable behavior). After this PR, it was renamed to "Attached disable of livepatch in a lxd vm" and put in livepatch.feature.

The test was combined with a livepatch disable test to avoid creating an extra vm, but I think we need to have a simple attach+auto-enable test in `cli/attach.feature`.

Something like
```
  Scenario Outline: Attached disable of livepatch in a lxd vm
    Then I verify that `esm-apps` is enabled
    And I verify that `esm-infra` is enabled
    And I verify that `livepatch` status is `<livepatch_status>`

    Examples: ubuntu release
      | release | machine_type | livepatch_status |
      | xenial  | lxd-vm       | warning          |
      | bionic  | lxd-vm       | enabled          |
      | focal   | lxd-vm       | enabled          |
      | jammy   | lxd-vm       | enabled          |
      | noble   | lxd-vm       | enabled          |

```

lucasmoura:
@orndorffgrant updated

orndorffgrant (approved):

renanrodrigo (approved):

PR #3270, opened 2024-08-16
tests: change how to disable network on airgapped

lucasmoura:
## Why is this needed?
Change how we disable the network connection on the airgapped container, since the ufw approach was occasionally hanging in CI

## Test Steps
Check that the airgapped container cannot reach the network and check that the airgapped tests keep working and no longer hangs on CI

### DIFF
--- a/features/steps/network.py
+++ b/features/steps/network.py
@@ -18,62 +18,14 @@ def block_access_to_address(context, website):
 def disable_internet_connection(context, machine_name=SUT):
     when_i_run_command(
         context,
-        "ufw default deny incoming",
+        "ip route del default",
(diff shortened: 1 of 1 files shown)


renanrodrigo (commented):
I tested this locally (delete the route + change the config) and I still have IPV6 connectivity

renanrodrigo:
@lucasmoura what do we want to do here now that the test is (likely) running again?
If we still want to change it, I would avoid trying to trick IPV6 (given mixed results), and try to remove all ipv6 default routes insted.

lucasmoura:
Closing this one as the airgapped have been working for quite some time now

PR #3272, opened 2024-08-20
test(network): add behave tests to ensure network errors are handled

a-dubs:
## Why is this needed?

This PR adds functionality that ensures that the various networking exception handling issues that have been raised on various LP bugs, which were resolved during the work for #2647 was completed, are properly caught, handled, and raised. 

## Test Steps
Run the new behave tests locally:
```
tox -e behave -- features/network_failures.feature
```

---

### DIFF
--- a/features/network_failures.feature
+++ b/features/network_failures.feature
@@ -1,0 +1,91 @@
+@uses.config.contract_token
+Feature: Ensure network errors are handled gracefully across various services
+
+  Scenario Outline: Various HTTP errors are handled gracefully on attaching contract token
+    # This test simulates various HTTP errors by mocking the response from the serviceclient
+    # when trying to attach contract token
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    # 400 Bad Request
+    When I create a response overlay for `/v1/context/machines/token` with response code `400` and error message `Bad Request`
+    And I append the following on uaclient config:
+      """
+      features:
+        serviceclient_url_responses: "/tmp/response-overlay.json"
+      """
+    When I attempt to attach `contract_token` with sudo
+    Then stderr contains substring:
+      """
+      Error connecting to /v1/context/machines/token: 400 {"error": "Bad Request"}
+      """
+    Then the machine is unattached
+    # 404 Not Found
+    When I create a response overlay for `/v1/context/machines/token` with response code `404` and error message `Not Found`
+    And I append the following on uaclient config:
+      """
+      features:
+        serviceclient_url_responses: "/tmp/response-overlay.json"
+      """
+    When I attempt to attach `contract_token` with sudo
+    Then stderr contains substring:
+      """
+      Error connecting to /v1/context/machines/token: 404 {"error": "Not Found"}
+      """
+    Then the machine is unattached
+    # 503 Bad Gateway
+    When I create a response overlay for `/v1/context/machines/token` with response code `503` and error message `Bad Gateway`
+    And I append the following on uaclient config:
+      """
+      features:
+        serviceclient_url_responses: "/tmp/response-overlay.json"
+      """
+    When I attempt to attach `contract_token` with sudo
+    Then stderr contains substring:
+      """
+      Error connecting to /v1/context/machines/token: 503 {"error": "Bad Gateway"}
+      """
+    Then the machine is unattached
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | xenial  | lxd-container |
+      | noble   | lxd-container |
+
+  Scenario Outline: Network errors for attaching contract token are handled gracefully
+    # This test simulates network failure by disabling internet connection
+    # and then trying to attach contract token
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I disable any internet connection on the machine
+    And I attempt to attach `contract_token` with sudo
+    Then stderr contains substring:
+      """
+      Failed to attach machine. See https://ubuntu.com/pro/dashboard
+      """
+    Then the machine is unattached
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | xenial  | lxd-container |
+      | noble   | lxd-container |
+
+  Scenario Outline: Network errors for enabling Realtime kernel and Livepatch are handled gracefully
+    # This test simulates network failure by disabling internet connection
+    # and then trying to enable realtime-kernel or livepatch
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I attach `contract_token` with sudo and options `--no-auto-enable`
+    Then the machine is attached
+    Then I verify that `<service>` is disabled
+    When I disable any internet connection on the machine
+    And I verify that running `pro enable <service> --assume-yes` `with sudo` exits `1`
+    Then stderr contains substring:
+      """
+      Failed to connect to https://contracts.canonical.com/v1/contracts/
+      """
+    Then I verify that `<service>` is disabled
+
+    # Realtime kernel is not supported on LXD containers so we must use a VM
+    Examples: ubuntu release
+      | release | machine_type  | service         |
+      | xenial  | lxd-vm        | realtime-kernel |
+      | noble   | lxd-vm        | realtime-kernel |
+      | xenial  | lxd-container | livepatch       |
+      | noble   | lxd-container | livepatch       |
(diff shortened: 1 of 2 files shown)


features/network_failures.feature
```diff
@@ -0,0 +1,92 @@
...
+              "error": <error_message>
+            }
+          }
+        ]
+      }
+      """
+    And I append the following on uaclient config:
+      """
+      features:
+      serviceclient_url_responses: "/tmp/response-overlay.json"
```
dheyay, line 24:
```suggestion
      features:
        serviceclient_url_responses: "/tmp/response-overlay.json"
```

Small indent issue causes test to fail, never sets the overlay and the machine ends up attached

features/network_failures.feature
```diff
@@ -0,0 +1,92 @@
...
+      """
+      Error connecting to <endpoint>: <response_code> {"error": <error_message>}
+      """
+    Then the machine is unattached
+
+    Examples: ubuntu release
+      | release | machine_type  | endpoint                   | response_code | error_message |
+      | jammy   | lxd-container | /v1/context/machines/token | 400           | "Bad Request" |
+      | jammy   | lxd-container | /v1/context/machines/token | 404           | "Not Found"   |
+      | jammy   | lxd-container | /v1/context/machines/token | 503           | "Bad Gateway" |
```
dheyay, line 37: We might not need the `endpoint` field, it is currently not in use in the test

features/network_failures.feature
```diff
@@ -0,0 +1,92 @@
...
+    Then I verify that running `pro enable livepatch` `with sudo` exits `1`
+    Then stderr contains substring:
+      """
+      Failed to connect to https://contracts.canonical.com/v1/contracts/
+      """
+    Then I verify that `livepatch` is disabled
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | jammy   | lxd-container |
```
dheyay, line 92: Same comment as above for different releases

dheyay (changes requested):
Looks good :100: , some minor nits

features/network_failures.feature
```diff
@@ -0,0 +1,92 @@
...
+    And I attempt to attach `contract_token` with sudo
+    Then stderr contains substring:
+      """
+      Failed to attach machine. See https://ubuntu.com/pro/dashboard
+      """
+    Then the machine is unattached
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | jammy   | lxd-container |
```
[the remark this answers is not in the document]
a-dubs: How about noble and focal? And then we just skip jammy? I feel like that covers our new & old bases well without drastically increasing the number of tests we have to run?

features/network_failures.feature
a-dubs: 😭😭 good catch. Thanks. (I didn't actually know indentation mattered like that)

features/network_failures.feature
a-dubs: True! I think I originally used it but no longer do.

features/network_failures.feature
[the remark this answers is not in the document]
dheyay: I'd say include all releases from `xenial` upto `noble` given we run the tests by `release` and `machine-type` so this can show up in every run and no release goes unchecked for this test

features/network_failures.feature
renanrodrigo: @a-dubs this is a yaml file, the responses key is nested under features:

a-dubs: oh wait i was being dumb yeah it is a yaml file. i thought the `.feature` file was throwing an error due to indentation lol.

a-dubs:
okay ready for re-review now :saluting_face: @dheyay

features/network_failures.feature
```diff
@@ -0,0 +1,104 @@
...
+    Then the machine is unattached
+
+    Examples: ubuntu release
+      | release | machine_type  | response_code | error_message |
+      | focal   | lxd-container | 400           | "Bad Request" |
+      | focal   | lxd-container | 404           | "Not Found"   |
+      | focal   | lxd-container | 503           | "Bad Gateway" |
+      | jammy   | lxd-container | 400           | "Bad Request" |
+      | jammy   | lxd-container | 404           | "Not Found"   |
+      | jammy   | lxd-container | 503           | "Bad Gateway" |
```
dheyay, line 40: Here I was thinking, instead of spinning up a new container for each message, maybe we could just have all the error message overlays created in the test itself. This could make the test just run all the error messages for a single release rather than create a new container per check. 

However, this approach also works and is good by me. Any thoughts? @a-dubs

a-dubs: I had that same thought.....

dheyay: We can do that here then, would make it easier to test too :eyes:

renanrodrigo:
@a-dubs seems `flake8` is not fond of your message strings

features/network_failures.feature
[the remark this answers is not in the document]
orndorffgrant: @a-dubs Limiting this test to run on a small number of releases seems reasonable. In these cases we typically do the oldest lts and newest lts, so xenial and noble

a-dubs:
Tests have been constrained to only xenial and noble since the functionality being tested by this is unlikely to break and only needs tested on the oldest and newest LTS releases.

dheyay (approved):
Looks good to me, would just need to squash commits before merging

renanrodrigo:
took the liberty to rebase this after merging #3267 - I'll take a last look and probably merge it

features/network_failures.feature
```diff
@@ -0,0 +1,109 @@
...
+
+  Scenario Outline: Network errors for attaching contract token are handled gracefully
+    # This test simulates network failure by disabling internet connection
+    # and then trying to attach contract token
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I disable any internet connection on the machine
+    And I attempt to attach `contract_token` with sudo
+    Then stderr contains substring:
+      """
+      Failed to attach machine. See https://ubuntu.com/pro/dashboard
```
renanrodrigo, line 61: I think this actually needs to be fixed for us to consider the respective LP bug closed. The problem here is that you don't have internet connection:
- there is no indication that lack of connection was the problem
- how are people supposed to access their dashboard if the network is down?

features/network_failures.feature
```diff
@@ -0,0 +1,109 @@
...
+    Then stderr contains substring:
+      """
+      Failed to connect to https://contracts.canonical.com/v1/contracts/
+      """
+    Then I verify that `livepatch` is disabled
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | xenial  | lxd-container |
+      | noble   | lxd-container |
```
renanrodrigo, line 109: See how this test is similar to the upper one, except for the service to be disabled?
You can create another column in the table below with "service to disable" so it looks like this:

```suggestion
    Examples: ubuntu release
      | release | machine_type  | service         |
      | xenial  | lxd-container | livepatch       |
      | xenial  | lxd-container | realtime-kernel |
      | noble   | lxd-container | livepatch       |
      | noble   | lxd-container | realtime-kernel |
```

Feel free to ignore this though - if you think this is good as is, keep it.

features/network_failures.feature
renanrodrigo: @a-dubs if you want, we can land the tests and then you change this message on a next step?

a-dubs: Oh shoot yeah you're right. That's not exactly a very useful or specific message. So you're right we definitely need to update the message to consider those bugs closed. As such, I think it would be best if I put this back in draft state and make the changes to the source code such that the source code changes and tests are bundled together into one logical commit, so that when I say "fixes lp #XXXX", I mean **_fixes_**

features/network_failures.feature
a-dubs: At first glance, I was initially onboard because i love reducing duplicate code where reasonable, but then I noticed the `and stdin 'y\ny'` that only exists in the realtime kernel tests and not the livepatch tests.

renanrodrigo: good catch forget what I said

orndorffgrant: Instead, would using `--assume-yes` on the enable call allow you to avoid the `and stdin` so that they can be the same for livepatch and realtime?

a-dubs: Oh you can do that??? That would be way better!

a-dubs:
@dheyay @renanrodrigo this should be ready for review now. let's just get this in and then i'll work on actually improving the exception handling and messaging for networking errors

PR #3274, opened 2024-08-22
d/rules: check version consistency at build time

orndorffgrant:
## Why is this needed?

This addresses v34 feedback described here: https://github.com/canonical/ubuntu-pro-client/pull/3247#discussion_r1710257188

## Test Steps

Make a temporary change to either the changelog version or version.py version and run
```
./tools/build.sh xenial
```

---

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -30,16 +30,16 @@ endif
 	dh $@ --with python3,bash-completion,systemd --buildsystem=pybuild
 
 override_dh_auto_build:
+	# fail early if version.py wasn't updated to match d/changelog's version
+	# only take into account the actual upstream version, discarding ubuntu and
+	# backport suffixes
+	python3 tools/check-versions-are-consistent.py
 	dh_auto_build
 	make -C apt-hook build
 	make -C debian/po build
 
 override_dh_auto_test:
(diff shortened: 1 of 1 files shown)


debian/rules
```diff
@@ -30,16 +30,16 @@ endif
...
+	# fail early if version.py wasn't updated to match d/changelog's version
+	# only take into account the actual upstream version, discarding ubuntu and
+	# backport suffixes
+	python3 tools/check-versions-are-consistent.py
 	dh_auto_build
 	make -C apt-hook build
 	make -C debian/po build
 
 override_dh_auto_test:
 ifeq (,$(filter nocheck,$(DEB_BUILD_OPTIONS)))
```
dheyay, line 42: Not quite sure but do we need the `ifeq` if we no conditional run inside?

dheyay (approved):
Just a small question, good otherwise

renanrodrigo (approved):

debian/rules
renanrodrigo: there is a filter for the `nocheck` option. If this is present, there is no testing when building at all

orndorffgrant: Yes, and the lack of indentation makes it hard to read but there is code that only runs inside this `ifeq`

PR #3275, opened 2024-08-22
tests: make test_gpg only test the gpg unit

orndorffgrant:
## Why is this needed?

The existing test was overcomplicated and tested more than the unit it needs to test.

## Test Steps

Unit tests should pass. `gpg.py` should still have complete coverage.

---


renanrodrigo (approved):
~-1 removing binaries where we could hide malicious code now that actual GPG checks are not performed in testing
How are we going to trick people~
LGTM

PR #3276, opened 2024-08-23
Better secret cleanup instructions in docker docs

orndorffgrant:
purging ubuntu-advantage-tools with autoremove only "removes" ubuntu-pro-client which doesn't remove secrets.
This is an unexpected consequence of the package rename.
Also start recommending to detach to be safe.


s-makin (approved):
LGTM :)

PR #3278, opened 2024-08-23
tests: update Unattended-Upgrade::Allowed-Origins

lucasmoura:
## Why is this needed?
We are now updating the expected value of `Unattended-Upgrade::Allowed-Origins` when that field is not present. We have recently changed the default value from null to an empty list and we are now reflecting that on the test

## Test Steps
Check that the modified integration test is working

---


dheyay (approved):
:+1:

PR #3279, opened 2024-08-23
fix: update endpoint for fetching USNs

lucasmoura:
## Why is this needed?
Use cves parameter instead of details to fetch USNs from the security endpoint

## Test Steps
Check that the fix integration tests are now working as expected

---


orndorffgrant (approved):
This lets `pro fix` work sometimes, which is better than never.

renanrodrigo (approved):

PR #3281, opened 2024-08-26
Collect cpu type

lucasmoura:
## Why is this needed?
We are now collecting the CPU type of a machine and sending it to the Pro backend during our timer jobs and attach operation. Additionally, note that the CPU type will only be collected for **attached** users.

## Test Steps
I have created unit tests for all architectures we support and I when can test it out on LXD container by directly importing the `get_cpu_type` function and calling it there to see that we are getting the right data.

---


uaclient/messages/__init__.py
```diff
@@ -2692,3 +2692,8 @@ def __repr__(self):
     "invalid-url",
     t.gettext("Invalid URL: {url}"),
 )
+
+E_UNKNONW_PROCESSOR_TYPE = FormattedNamedMessage(
```
orndorffgrant, line 2696: typo
```suggestion
E_UNKNOWN_PROCESSOR_TYPE = FormattedNamedMessage(
```

uaclient/system/cpu_type.py
```diff
@@ -0,0 +1,179 @@
...
+
+def is_s390_processor(processor_type: str):
+    return processor_type.startswith("s390")
+
+
+def is_arm_processor(processor_type: str):
+    return processor_type.startswith("arm")
+
+
+def get_cpu_type_collector(processor_type):
```
orndorffgrant, line 159: This is essentially the factory pattern, should we call this `cpu_type_collector_factory`?

uaclient/system/cpu_type.py
```diff
@@ -0,0 +1,179 @@
...
+    elif is_s390_processor(processor_type):
+        return S390CPUTypeCollector()
+    elif is_arm_processor(processor_type):
+        return ArmCPUTypeCollector()
+    else:
+        raise exceptions.UnknownProcessorType(processor_type=processor_type)
+
+
+def get_cpu_type():
+    processor_type = system.get_dpkg_arch()
```
orndorffgrant, line 173: dpkg arch is not necessarily the arch of the cpu.
It would be more technically accurate to use `os.uname().machine` I think

lucasmoura: @cpaelzer just double checking with you if that is a preferred option, since we had that `uname` discussion in the past

orndorffgrant: We discussed out of band and decided dpkg arch is okay

uaclient/system/cpu_type.py
```diff
@@ -0,0 +1,179 @@
...
+                or ""
+            )
+
+            fallback_name += "{}:{}".format(
+                field,
+                field_value,
+            )
+
+            if field != fallback_fields[-1]:
+                fallback_name += "; "
```
orndorffgrant, line 50: This is here and then there is at least one hardcoded string that is supposed to fit the same format. Can we have a single implementation of the serialization?
```python
def cpu_type_serialize(d: Dict) -> str:
    # turn the dict into the string
```

and then use that function for all cases where we're returning the serialized string?

orndorffgrant (approved):
Some minor suggestions

uaclient/system/cpu_type.py
```diff
@@ -0,0 +1,179 @@
...
+
+def cpu_type_collector_factory(processor_type):
+    if is_intel_or_amd_processor(processor_type):
+        return X86CPUTypeCollector()
+    elif is_ppc_processor(processor_type):
+        return PPCCPUTypeCollector()
+    elif is_s390_processor(processor_type):
+        return S390CPUTypeCollector()
+    elif is_arm_processor(processor_type):
+        return ArmCPUTypeCollector()
```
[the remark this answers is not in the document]
lucasmoura: I think that is better, let me adapt the code to use it

uaclient/system/cpu_type.py
lucasmoura: Just double checking, you want us to first build the dictionary of fallback fields and just use this function to consolidate the dict into a string ?

orndorffgrant: Yes, because fundamentally what we're doing is serializing key-value pairs into a string (though "serialize" is a strong word here because it's not 100% reversible). So we should have a clear single implementation of that serialization that is tested and used anywhere it is needed (including in this function).

renanrodrigo: s/serialize/stringify/

uaclient/system/cpu_type.py
```diff
@@ -0,0 +1,187 @@
...
+        )
+        cpu_model = (
+            self._get_field_value(
+                cpu_info=cpu_info,
+                re_pattern=self.CPU_TYPE_REGEX_PATTERN.format("Model"),
+            )
+            or ""
+        )
+
+        return "Type:{}; Model:{}".format(cpu_type, cpu_model)
```
orndorffgrant, line 130: Now here we can call
```suggestion
        return self._cpu_type_stringify({"type": cpu_type, "model": cpu_model})
```

lucasmoura:
CI failures are unrelated to this PR

PR #3283, opened 2024-08-27
feat(auto-attach): skip if no codes on aws (CPC-4926)

aciba90:
## Why is this needed?

See CPC0069.

```
feat(auto-attach): skip if no codes on aws (CPC-4926)

On aws, ua-auto-attach.service will only try to auto-attach if the instance identity document[0] contains any billing product or marketplace product code.

Refs:

- [0] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html#verify-iid
- [1] CPC-4926
- [2] CPC0069

Signed-off-by: Alberto Contreras <alberto.contreras@canonical.com>
```

## Test Steps

```
tox -e behave -- features/ubuntu_pro.feature:191 -D releases=jammy -D machine_types=aws.pro
```

---

### DIFF
--- a/features/ubuntu_pro.feature
+++ b/features/ubuntu_pro.feature
@@ -174,21 +174,83 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
 
     Examples: ubuntu release
       | release | machine_type |
-      | xenial  | aws.pro      |
       | xenial  | azure.pro    |
       | xenial  | gcp.pro      |
-      | bionic  | aws.pro      |
       | bionic  | azure.pro    |
       | bionic  | gcp.pro      |
-      | focal   | aws.pro      |
       | focal   | azure.pro    |
       | focal   | gcp.pro      |
-      | jammy   | aws.pro      |
       | jammy   | azure.pro    |
       | jammy   | gcp.pro      |
-      | noble   | aws.pro      |
       | noble   | azure.pro    |
       | noble   | gcp.pro      |
+
+  Scenario Outline: Auto-attach service works on Pro Machine on aws.pro
+    Given a `<release>` aws.pro machine with ubuntu-advantage-tools installed
+    When I run `systemctl start ua-auto-attach.service` with sudo
+    And I create the file `/etc/ubuntu-advantage/uaclient.conf` with the following:
+      """
+      contract_url: 'https://contracts.canonical.com'
+      data_dir: /var/lib/ubuntu-advantage
+      log_level: debug
+      log_file: /var/log/ubuntu-advantage.log
+      """
+    And I reboot the machine
+    And I run `pro status --wait` with sudo
+    And I run `pro security-status --format json` with sudo
+    Then stdout matches regexp:
+      """
+      "attached": true
+      """
+    When I run `journalctl --no-pager -b -u ua-auto-attach.service` with sudo
+    Then stdout matches regexp:
+      """
+      Auto-attaching: product code found on AWS
+      """
+    Then stdout does not match regexp:
+      """
+      Skipping auto-attach. Reason: No billingProduct nor marketplaceProductCode on AWS."
+      """
+
+    Examples: ubuntu release
+      | release |
+      | xenial  |
+      | bionic  |
+      | focal   |
+      | jammy   |
+      | noble   |
+
+  Scenario Outline: Auto-attach service works on Pro Machine on aws.generic
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I install ubuntu-pro-auto-attach
+    When I reboot the machine
+    And I run `pro status --wait` with sudo
+    Then stdout matches regexp:
+      """
+      This machine is not attached to an Ubuntu Pro subscription.
+      """
+    When I verify that running `systemctl status ua-auto-attach.service` `as non-root` exits `3`
+    Then stdout matches regexp:
+      """
+      \s*Main PID: \d+ \(code=exited, status=0/SUCCESS\)
+      """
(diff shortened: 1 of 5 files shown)


aciba90:
Force-pushed to fix the integration test that I broke. Other failing integration tests look like unrelated to this change.

lucasmoura:
@aciba90 I thinking I am missing more context around this change. It seems that there is a type of image that we shouldn't auto-attach and this is fixing it. However, which type of image is that and shouldn't our integration test actually verify if auto-attach doesn't work on that image ?

aciba90:
> @aciba90 Shouldn't our integration test actually verify if auto-attach work on the image type defined on the CPC spec ?
> Because by testing it directly on the already used AWS Pro image, we are just guaranteeing that your change doesn't break the process there.
> 
> Additionally, it would be great if we can guarantee that this also doesn't work on generic AWS images

Thanks, @lucasmoura, for pointing that out.

At the moment, we have AWS generic and pro images. The difference is that the latter one comes with `ubuntu-pro-auto-attach` preinstalled. What this PR will allow in the future is to be able to have a single image capable of acting as pro and non-pro. Thus, in the future, both images will contain `ubuntu-pro-auto-attach` preinstalled. Thus, testing current AWS Pro and current AWS Generic + `ubuntu-pro-auto-attach` does cover all cases. I have added a supplemental test covering the later case.

features/ubuntu_pro.feature
```diff
@@ -174,22 +174,84 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
...
       | focal   | azure.pro    |
       | focal   | gcp.pro      |
-      | jammy   | aws.pro      |
       | jammy   | azure.pro    |
       | jammy   | gcp.pro      |
-      | noble   | aws.pro      |
       | noble   | azure.pro    |
       | noble   | gcp.pro      |
 
+  Scenario Outline: Auto-attach service works on Pro Machine on aws.pro
```
a-dubs, line 188: so this test ensures backwards compatability with existing pro images? and `Scenario Outline: Auto-attach service works on Pro Machine on aws.generic` tests the new functionality right?

uaclient/clouds/aws.py
```diff
@@ -43,10 +42,33 @@ def _get_imds_url_response(self):
...
+            ii_doc = json.loads(
+                self._get_imds_url_response(url, headers=headers)
+            )
+        except json.JSONDecodeError as e:
+            LOG.debug("Error decoding instance identity document: %s", e)
+            return {}
+        return ii_doc
+
+    @property  # type: ignore
+    def is_likely_pro(self) -> bool:
```
a-dubs, line 65: :rofl: good function name haha

uaclient/clouds/aws.py
```diff
@@ -43,10 +42,33 @@ def _get_imds_url_response(self):
...
     def identity_doc(self) -> Dict[str, Any]:
-        imds_url_response = self._get_imds_url_response()
+        headers = self._request_imds_v2_token_headers()
+        url = IMDS_URL.format(self._ip_address)
+        imds_url_response = self._get_imds_url_response(url, headers=headers)
         secret_manager.secrets.add_secret(imds_url_response)
         return {"pkcs7": imds_url_response}
 
+    @util.retry(exceptions.CloudMetadataError, retry_sleeps=[0.5, 1, 1])
+    def _get_ii_doc(self) -> Dict:
```
a-dubs, line 52: also could you add docstrings to this function and the below `is_likely_pro` function?

a-dubs (changes requested):
left a question, and a small nit for docstrings

features/ubuntu_pro.feature
aciba90: `Scenario Outline: Auto-attach service works on Pro Machine on aws.pro` and
`Scenario Outline: Auto-attach service works on Pro Machine on aws.generic` test that pro images and generic images behave correctly with `ubuntu-advantage-pro` installed. 

This service is pre-installed in every Ubuntu Pro image. And at some point in the future, likely, AWS generic images will start shipping `ubuntu-advantage-pro` and not auto-attaching, thanks to this PR.

aciba90: For more context see the attached jira and spec.

uaclient/clouds/aws.py
aciba90: Do you have any better suggestion? I think it represents right the feature it holds.

uaclient/clouds/aws.py
aciba90: Done.

uaclient/clouds/aws.py
renanrodrigo: no this was an actual compliment, keep it

features/ubuntu_pro.feature
```diff
@@ -174,22 +174,84 @@ Feature: Command behaviour when auto-attached in an ubuntu PRO image
...
+      | release |
+      | xenial  |
+      | bionic  |
+      | focal   |
+      | jammy   |
+      | noble   |
+
+  Scenario Outline: Auto-attach service works on Pro Machine on aws.generic
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I install ubuntu-advantage-pro
```
lucasmoura, line 225: @aciba90 let's instead install the `ubuntu-pro-auto-attach` package. The `ubuntu-advantage-pro` package is now just a transition package to this new one

aciba90: Done, thanks for the suggestion!

lucasmoura (approved):

renanrodrigo (approved):

PR #3285, opened 2024-08-27
Add arm64 tag to tests

dheyay:
## Why is this needed?

@arm64 tag added to basic tests for running weekly on new arm64 jenkins job

## Test Steps

---

### DIFF
--- a/features/api/api.feature
+++ b/features/api/api.feature
@@ -1,5 +1,6 @@
 Feature: Client behaviour for the API endpoints
 
+  @arm64
   Scenario Outline: all API endpoints can be imported individually
     Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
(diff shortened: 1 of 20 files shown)


dheyay:
Currently I have only added some tests as arm-64 tagged, also they are exclusively only `lxd-container` tests. the idea is to get these working and see how it goes, once these are working, we can expand to more tests and also vm tests.

renanrodrigo (approved):
@dheyay as we spoke before, I think we good to apply the changes but now you need some rebase - will merge as soon as done

renanrodrigo (changes requested):
Seems you brought back some files from the dead
And there are merge conflicts in the diff

I would say it's valid to just start from scratch and re-tag your tests instead of trying to solve conflicts at this point

dheyay:
@renanrodrigo I was working on that as soon as I realized this :sweat_smile:

dheyay:
Closing PR, created a new one with no conflicts.

PR #3288, opened 2024-09-03
Release v34 fixups

orndorffgrant:
## Why is this needed?

This addresses feedback on v34. It also adds an additional LP bug for one of our fixes as I agreed to track in on the higher visibility bug in addition to the generic bug.

---


dheyay (approved):

debian/changelog
```diff
@@ -44,7 +44,7 @@ ubuntu-advantage-tools (33) oracular; urgency=medium
     - contracts: add support for contracts which target a specific series
     - fips: change enable functionality to ensure all packages with a FIPS
       candidate are upgraded to the FIPS version (GH: #2667)
-    - fix:
+    - fix:
```
panlinux, line 47: Since this is up for review... Extra whitespace at the end of the line? ;)

debian/changelog
```diff
@@ -707,7 +707,7 @@ ubuntu-advantage-tools (27.11~22.10.1) kinetic; urgency=medium
     - pro:
       + Ubuntu Pro is released as a product
       + make `pro` the recommended executable for the client
-      + client, apt and motd messages updated/rewritten to show Pro
+      + client, apt and motd messages updated/rewritten to show Pro
```
panlinux, line 710: Since this is up for review... Extra whitespace at the end of the line? ;)

apt-hook/20apt-esm-hook.conf
```diff
@@ -3,5 +3,5 @@ APT::Update::Pre-Invoke {
 };
 
 binary::apt::AptCli::Hooks::Upgrade {
-	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-json-hook ] || /usr/lib/ubuntu-advantage/apt-esm-json-hook 2> /var/lib/ubuntu-advantage/apt-upgrade-hook-err.txt || true";
+	"[ ! -f /usr/lib/ubuntu-advantage/apt-esm-json-hook ] || /usr/lib/ubuntu-advantage/apt-esm-json-hook 2>> /var/log/ubuntu-advantage-apt-hook.log || true";
```
panlinux, line 6: I checked and `debian/ubuntu-pro-client.logrotate` includes this new log file in its existing globbing rule.

debian/changelog
orndorffgrant: @panlinux this is in response to your comment here :) https://github.com/canonical/ubuntu-pro-client/pull/3247#discussion_r1739416521

panlinux: Ah, you are undoing that change, ok, I should diff against pkg/ubuntu/devel

apt-hook/20apt-esm-hook.conf
panlinux: Did some testing, forcing a logrotate while the log file was empty, and later populating it with some garbage. It worked as expected. Here is when I forcibly populated the file, and then issue logrotate a few times:
```
$ ls -la /var/log/ubuntu-advantage*
-rw-r--r-- 1 root root    0 Sep  4 18:43 /var/log/ubuntu-advantage-apt-hook.log
-rw-r--r-- 1 root root    6 Sep  4 18:43 /var/log/ubuntu-advantage-apt-hook.log.1
-rw-r--r-- 1 root root   26 Sep  4 18:43 /var/log/ubuntu-advantage-apt-hook.log.2.gz
-rw-r--r-- 1 root root    0 Sep  4 18:42 /var/log/ubuntu-advantage.log
-rw-r--r-- 1 root root 2730 Sep  4 18:42 /var/log/ubuntu-advantage.log.1
-rw-r----- 1 root root 7737 Sep  4 18:38 /var/log/ubuntu-advantage.log.2.gz
```
And then the files are zero-sized, as it is in the snapshot above, logrotate correctly does nothing to them.

debian/changelog
panlinux: Did the diff against pkg/ubuntu/devel, no noise, just the new d/changelog entry, all good.

apt-hook/20apt-esm-hook.conf
panlinux, line 6: We need to include the removal of `/var/log/ubuntu-advantage-apt-hook.log*` in `debian/ubuntu-pro-client.postrm`

panlinux (changes requested):
We need to include the removal of `/var/log/ubuntu-advantage-apt-hook.log*` in `debian/ubuntu-pro-client.postrm`

orndorffgrant:
@panlinux I'm aware of that postinst snippet. It is old cruft in its current form. We may want to remove it and add something similar to ubuntu-pro-client.postinst, but that is not urgent and its current form is not harmful. I think it's okay to leave as is for this release.

orndorffgrant:
> We need to include the removal of `/var/log/ubuntu-advantage-apt-hook.log*` in `debian/ubuntu-pro-client.postrm`

Added!

panlinux:
rechecking...

panlinux (approved):
+1

PR #3292, opened 2024-09-05
Ignore corrupt jobs-state.json file

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because we now ignore invalid jobs-state.json files if they are corrupted and cannot be decoded, avoiding the oops which stopped phasing on noble (see LP bug)

LP: #2078737

## Test Steps
Unit tests cover behavior
for the functionality:
- insert some weird characters in your jobs-state.json file
- `systemctl start ua-timer.service`
- see it degrade
- apply this patch
- repeat
- see it happy

---

### DIFF
--- a/lib/timer.py
+++ b/lib/timer.py
@@ -8,7 +8,10 @@ from typing import Callable, Optional
 
 from uaclient import http, log
 from uaclient.config import UAConfig
-from uaclient.exceptions import InvalidFileFormatError
+from uaclient.exceptions import (
+    InvalidFileEncodingError,
+    InvalidFileFormatError,
+)
 from uaclient.files import machine_token
 from uaclient.files.state_files import (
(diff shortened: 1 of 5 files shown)


renanrodrigo:
Ok Github is failing to build the package on Xenial because it is using an outdated version of the Python 3.5 `stdlib` in the chroot. This is reproducible locally. May have to comment the test out.

renanrodrigo:
As agreed I just removed the extra unit test (as we never had it anyway) and there is a unit test covering the new exception in the timer job. This is now good to go.

panlinux (approved):
+1 from SRU perspective

sergiodj (approved):
+1 from me as well.

orndorffgrant (approved):
Thanks! This works and looks good.

PR #3231, opened 2024-07-26
docs: add documentation for dep8 tests

lucasmoura:
## Why is this needed?
We are now documenting our rationale for our dep8 tests while also documenting how to run them locally

PS:We need to first merge #3230, as this documentation already consider the updated test

- [ ] *(un)check this to re-run the checklist action*

PR #3293, opened 2024-09-05
debian: add more scenarios to dep8 tests

lucasmoura:
## Why is this needed?
We are now calling two API endpoints on the dep8 tests. Those endpoints don't trigger any network call and their main use is to verify if any python3-apt change would change the way we interact with that library, as those API endpoints heavily rely on apt functionality.

This is being documented on: #3231 

## Test Steps
1. Build the package
2. Use the .dsc file to run this command:

```bash
autopkgtest --no-built-binaries --apt-upgrade --setup-commands setup-testbed --shell-fail ubuntu-advantage-tools_1+devel.dsc -- lxd ubuntu-daily:<RELEASE>
```

3. Check that this passes and that the API endpoints appears in the logs 

---

### DIFF
--- a/debian/tests/usage
+++ b/debian/tests/usage
@@ -2,5 +2,11 @@
 
 set -ex
 
-ua --help | grep --silent services
-ua version
+pro --help | grep --silent services
+pro version
+
+# These calls are used to ensure that the
+# way we interact with APT is still working
+# if python3-apt receives an upgrade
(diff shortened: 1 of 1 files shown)


renanrodrigo (commented):
LGTM!

renanrodrigo (approved):

PR #3294, opened 2024-09-05
Add confg values for the new CLI

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because users should be able to set color and suggestion output in the new CLI using the Client configuration. And the possible values should be true or false.

We also add a class to serve as global configuration for the CLI specific options. Besides the colors and suggestions, it checks for utf-8 support. It adds new constraints to use colors, so commands don't read directly from the UAConfig, but combine this with this class instead.

The class has methods which can be dinamically executed based on CLI arguments after parsing those - because it's information we don't have from the system, but rather from the user.

## Test Steps
Unit + integration was addressed

---

### DIFF
--- a/uaclient/cli/formatter.py
+++ b/uaclient/cli/formatter.py
@@ -1,0 +1,38 @@
+import os
+import sys
+
+from uaclient.config import UAConfig
+
+
+# Class attributes and methods so we don't need singletons or globals for this
+class ProOutputFormatterConfig:
+    use_utf8 = True
+    use_color = True
+    show_suggestions = True
+
+    # Initializing the class after the import is useful for unit testing
+    @classmethod
+    def init(cls, cfg: UAConfig):
+        cls.use_utf8 = (
+            sys.stdout.encoding is not None
+            and "UTF-8" in sys.stdout.encoding.upper()
+        )
+
+        cls.use_color = (
+            sys.stdout.isatty()
+            and os.getenv("NO_COLOR") is None
+            and cfg.cli_color
+        )
+
+        cls.show_suggestions = cfg.cli_suggestions
+
+    @classmethod
+    def disable_color(cls) -> None:
+        cls.use_color = False
+
+    @classmethod
+    def disable_suggestions(cls) -> None:
+        cls.show_suggestions = False
+
+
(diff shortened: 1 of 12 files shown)


dheyay (approved):
:+1:

renanrodrigo:
@dheyay thanks for approving! 
I have been working on the next steps, and considered this PR deserves a new commit, with CLI specific configuration. PTAL. Explanation in the PR description.

uaclient/cli/formatter.py
```diff
@@ -0,0 +1,38 @@
...
+    @classmethod
+    def init(cls, cfg: UAConfig):
+        cls.use_utf8 = (
+            sys.stdout.encoding is not None
+            and "UTF-8" in sys.stdout.encoding.upper()
+        )
+
+        cls.use_color = (
+            sys.stdout.isatty()
+            and os.getenv("NO_COLOR") is None
```
lucasmoura, line 23: what if the user sets this to `NO_COLOR=0` ?

lucasmoura (approved):

uaclient/cli/formatter.py
renanrodrigo: From: https://no-color.org/

```
Command-line software which adds ANSI color to its output by default should check for a NO_COLOR environment variable that, when present and not an empty string (regardless of its value), prevents the addition of ANSI color.
```

renanrodrigo: btw the same website says that `NO_COLOR` should be the last checked thing: that if configuration or CLI says show color then you show color regardless.
our CLI will have a no-color boolean option instead of a color=? option that you can set to things like 'yes, enabled, disabled, auto, blablabla' so we good CLI wise
but the configuration is not being respected: `NO_COLOR` env means no color, I'm ANDing the cases.

If you think this is a problem, we can always rename the config defaults from `cli_color = True` to `disable_color = False` and we would get a reasonable behavior.

renanrodrigo: (by reasonable I mean that any of the triggers would disable the colors, regardless of order, and none would need precedence)

PR #3295, opened 2024-09-06
fix api error schema

orndorffgrant:
## Why is this needed?

This has been documented wrong for a long time and we just noticed :(

### DIFF
--- a/docs/references/api.rst
+++ b/docs/references/api.rst
@@ -28,9 +28,9 @@ standard structure, as can be seen below:
            },
            "type": "<response_type>"
        },
-       "errors": [], 
+       "errors": [],
        "result": "<success|failure>",
-       "version": "<client version>", 
+       "version": "<client version>",
        "warnings": []
    }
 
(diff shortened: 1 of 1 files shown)


s-makin (approved):
LGTM

renanrodrigo (approved):

s-makin:
The IRC link is being complained about by the linkchecker. Did something happen to the IRC channel?

renanrodrigo:
@s-makin the problem is that this is using Kiwi IRC (an online client) and that is the problem. The channel is up, this is a shortcut to click+access it if you don't have a client yourself.

Given that, yes, it seems to be off. Even just `kiwiirc.com` is off.

s-makin:
> @s-makin the problem is that this is using Kiwi IRC (an online client) and that is the problem. The channel is up, this is a shortcut to click+access it if you don't have a client yourself.
> 
> Given that, yes, it seems to be off. Even just `kiwiirc.com` is off.

Is there an alternative we can provide that does work? Or should we just remove the link entirely?

PR #3296, opened 2024-09-06
chore: use python-apt version from noble-updates in tox

renanrodrigo:
## Why is this needed?

This PR solves all of our problems because we want the latest compatible python3-apt in tests.

## Test Steps
`tox` on Noble

---


dheyay (approved):
:+1:

Issue #3184, opened 2024-07-02
Bug: pinned packages are reported as available updates in the Pro API


renanrodrigo:
**Description of the bug**

If a package is installed from ESM, but there is a new version in the archive, the `u.pro.packages.updates.v1` API will report this new version as an available update. This is not wrong, as it is an available update indeed, but APT will not pick it up due to pinning, causing
a) a reported update that never gets installed, and
b) customer confusion because of a)

**Expected behavior**

The Pro Client could identify, using the `status` field of the update, that it won't be installed by default somehow.

**Current behavior**

`"status": "upgrade_available"` and lots of questions.

**To Reproduce**

Mocking is the easiest way, as the security team should be taking care of those corner cases, and the one package that triggered this issue - it happened recently with `distro-info-data` - is already fixed (copied to ESM with a higher version so it's installable).

**System information:**

irrelevant

PR #3297, opened 2024-09-06
Apt pinned packages

dheyay:
## Why is this needed?

Fix for #3184
Documentation update PR: #3328 

## Test Steps

---

### DIFF
--- a/uaclient/security_status.py
+++ b/uaclient/security_status.py
@@ -17,6 +17,7 @@ from uaclient.apt import (
     get_apt_cache_datetime,
     get_apt_pkg_cache,
     get_esm_apt_pkg_cache,
+    get_pkg_candidate_version,
 )
 from uaclient.config import UAConfig
 from uaclient.entitlements import ESMAppsEntitlement, ESMInfraEntitlement
@@ -39,6 +40,7 @@ ESM_SERVICES = ("esm-infra", "esm-apps")
 class UpdateStatus(Enum):
     "Represents the availability of a security package."
     AVAILABLE = "upgrade_available"
+    AVAILABLE_NOT_PREFERRED = "upgrade_available_not_preferred"
     UNATTACHED = "pending_attach"
     NOT_ENABLED = "pending_enable"
     UNAVAILABLE = "upgrade_unavailable"
@@ -118,7 +120,20 @@ def get_origin_for_installed_package(
     return "third-party"
 
 
-def get_update_status(service_name: str, ua_info: Dict[str, Any]) -> str:
+@lru_cache(maxsize=None)
+def _is_candidate_version(pkg: str, version: str) -> bool:
+    """Returns True if the package version is a candidate version."""
+    candidate_version = get_pkg_candidate_version(pkg, check_esm_cache=False)
+    if candidate_version:
+        return version == candidate_version
+    return False
+
+
+def get_update_status(
+    service_name: str,
+    ua_info: Dict[str, Any],
+    version: apt_pkg.Version,
+) -> str:
     """Defines the update status for a package based on the service name.
 
     For ESM-[Infra|Apps] packages, first checks if Pro is attached. If this is
@@ -127,7 +142,14 @@ def get_update_status(service_name: str, ua_info: Dict[str, Any]) -> str:
     if service_name in ("standard-security", "standard-updates") or (
         ua_info["attached"] and service_name in ua_info["enabled_services"]
     ):
-        return UpdateStatus.AVAILABLE.value
+        is_candidate = _is_candidate_version(
+            version.parent_pkg.name, version.ver_str
+        )
+        return (
+            UpdateStatus.AVAILABLE.value
+            if is_candidate
+            else UpdateStatus.AVAILABLE_NOT_PREFERRED.value
+        )
     if not ua_info["attached"]:
         return UpdateStatus.UNATTACHED.value
     if service_name in ua_info["entitled_services"]:
@@ -262,8 +284,8 @@ def create_updates_list(
 ) -> List[Dict[str, Any]]:
     updates = []
     for service, version_list in upgradable_versions.items():
-        status = get_update_status(service, ua_info)
         for version, origin in version_list:
+            status = get_update_status(service, ua_info, version)
             updates.append(
                 {
                     "package": version.parent_pkg.name,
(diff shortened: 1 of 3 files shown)


uaclient/security_status.py
```diff
@@ -154,6 +157,7 @@ def filter_updates(
         for package in packages:
             # We only care about installed packages here
             if package.current_ver:
+                candidate_version = policy.get_candidate_ver(package)
```
renanrodrigo, line 160: Sorry I know this is still a WIP.

I'd say this function is meant to filter the available updates for a package list, and the candidate logic should not live there.
I'd say the `create_updates_list` function is the one which should have knowledge about the candidates, so it can be passed to `get_update_status` and treated there as a separate case, instead of a separate variable drilled down.
I know depcaches have candidates for packages - I suggest a small helper function:
`_is_version_the_candidate(version: apt_pkg.Version) -> bool:`
which would check the parent package candidate and see it it matches the version, or something close to this.
Also, consider `lru_cache`-ing the logic which finds a candidate for a package, so you don't have to invoke the depcache multiple times for the same package (although this should be minimal / fine for this step)

uaclient/security_status.py
```diff
@@ -175,24 +189,37 @@ def filter_updates(
...
                                 )
                                 if service:
+                                    upgrade_status = (
+                                        None
+                                        if version == esm_candidate_version
+                                        else UpdateStatus.AVAILABLE_LOWER_PRIORITY.value
+                                    )
                                     result[service].append(
-                                        (version, origin.site)
+                                        (version, origin.site, upgrade_status)
```
renanrodrigo, line 222: When the user does not have a service enabled, the upgrade will never be available
This part is not needed afaics

uaclient/security_status.py
dheyay: Good cache (catch) :eyes: 
Yeah since we discussed not going the policy route, it makes sense to get a different function and not populate the filter function with this task.

uaclient/security_status.py
```diff
@@ -118,15 +120,35 @@ def get_origin_for_installed_package(
...
 
 
-def get_update_status(service_name: str, ua_info: Dict[str, Any]) -> str:
+@lru_cache(maxsize=None)
+def _is_candidate_version(pkg: str, version: str) -> bool:
+    """Returns True if the package version is a candidate version."""
+    candidate_version = get_pkg_candidate_version(pkg, check_esm_cache=False)
+    if candidate_version:
+        return version == candidate_version
+    return False
```
lucasmoura, line 129: Maybe we can replace this for a one liner:
`return version == candidate_version`

Or is there a situation where `version` is None too ?

lucasmoura (commented):
@dheyay @renanrodrigo do you think we could come up with an integration test for this ?

uaclient/security_status.py
dheyay: The `get_pkg_candidate_version` function has the return type of `Optional[str]` so I believe there is a case where the candidate version returned can be `None`

orndorffgrant (changes requested):

uaclient/security_status.py
```diff
@@ -256,20 +258,40 @@ def get_livepatch_fixed_cves() -> List[Dict[str, Any]]:
     return []
 
 
+@lru_cache(maxsize=None)
+def _is_candidate_version(pkg: str, version: str) -> bool:
+    """Returns True if the version is a candidate version."""
+    candidate_version = get_pkg_candidate_version(pkg, check_esm_cache=True)
+    # print(f"candidate_version: {candidate_version}, version: {version}")
```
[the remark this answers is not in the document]
orndorffgrant: I didn't even notice this was a stale review comment there - sorry.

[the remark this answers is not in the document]
dheyay: Removed this while editing currently, should be good, thanks :smile:

uaclient/security_status.py
```diff
@@ -118,15 +120,35 @@ def get_origin_for_installed_package(
...
+    if service_name in ("standard-security", "standard-updates"):
+        is_candidate = _is_candidate_version(
+            version.parent_pkg.name, version.ver_str
+        )
+        return (
+            UpdateStatus.AVAILABLE.value
+            if is_candidate
+            else UpdateStatus.AVAILABLE_NOT_PREFERRED.value
+        )
+    if ua_info["attached"] and service_name in ua_info["enabled_services"]:
```
renanrodrigo, line 151: Is there a particular reason to apply the logic only to the archive, but not to possible underpreferred ESM versions?
I mean, even if those are not a planned thing, they may happen and the criteria is the same - anything available but not a candidate should be in the new state

uaclient/security_status.py
renanrodrigo: Candidate will be None when you request a package that is not in the Cache, or a package which is in the Cache but has no installable version (or is pinned to never, or things like that) so yeah for the sake of corner cases, keep the check

renanrodrigo:
Writing the integration test would be hard (as you probably figured out by now) because the situation where we have unpreferred updates is (or should be) temporary (see the issue - the package which triggered it is fixed already). Either we fake it all (no reason to have an integration test then, just the unit test will do) or we reason about it and move on.

In the second case, I would still go for a manual test, even if i involves manipulating apt caches, just to make sure.

uaclient/security_status.py
dheyay: Updated this as per standup discussion :smile:

uaclient/security_status.py
```diff
@@ -118,7 +120,21 @@ def get_origin_for_installed_package(
     return "third-party"
 
 
-def get_update_status(service_name: str, ua_info: Dict[str, Any]) -> str:
+@lru_cache(maxsize=None)
+def _is_candidate_version(pkg: str, version: str) -> bool:
+    """Returns True if the package version is a candidate version."""
+    candidate_version = get_pkg_candidate_version(pkg, check_esm_cache=False)
+    print(pkg, " ", version, " ", candidate_version)
```
orndorffgrant, line 127:
```suggestion
```

orndorffgrant (commented):
Thanks this looks good! Approval except for below:

From the checklist:

> - [x] No documentation updates are necessary for this change

This introduces a new enum value, which must at least be described here: https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/explanations/how_to_interpret_the_security_status_command/#u-pro-packages-updates-v1

lucasmoura (approved):

orndorffgrant (approved):

PR #3304, opened 2024-09-13
cli: add a wrap function which considers colored text

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because with this we are able to wrap text even if there are color codes in it, using the full space on screen. This is mostly useful for tables/blocks.

This is based on #3294 and should land after that.

## Test Steps
unit tests added for the functionality

---


orndorffgrant (approved):
pre-emptive approval for after cli config pr merged

PR #3306, opened 2024-09-13
tests: don't install ubuntu-advantage-tools on Noble

lucasmoura:
## Why is this needed?
On Noble, the ubuntu-advantage-tools package is not installed by default, that means that once we install the package in our tests, all of the previous migrations would run. This was causing some issue on our Pro cloud integration tests. We configure those tests to disable the auto_attach feature on startup. However, one of the migrations was overriding our config file momentarily, allowing our daemon to run and attach the machine. Since we want to avoid that behavior, we are now masking the daemon before the package install and unmasking it just after the operation

## Test Steps
Run the GCP Pro Noble tests and double check they are working now

---

### DIFF
--- a/features/steps/ubuntu_advantage_tools.py
+++ b/features/steps/ubuntu_advantage_tools.py
@@ -84,7 +84,7 @@ def when_i_install_uat(context, machine_name=SUT):
             instance.push_file(deb_path, instance_tmp_path)
             to_install.append(instance_tmp_path)
         if is_pro:
-            for deb_name, deb_path in debs.cloud_pro_image_debs(series):
+            for deb_name, deb_path in debs.cloud_pro_image_debs():
                 instance_tmp_path = "/tmp/behave_{}.deb".format(deb_name)
                 instance.push_file(deb_path, instance_tmp_path)
                 to_install.append(instance_tmp_path)
@@ -151,7 +151,7 @@ def when_i_have_the_debs_under_test(context, series, dest):
             sbuild_output_to_terminal=context.pro_config.sbuild_output_to_terminal,  # noqa: E501
         )
 
-        for deb_name, deb_path in debs.all_debs():
+        for deb_name, deb_path in debs.all_debs(series):
             context.machines[SUT].instance.push_file(
                 deb_path, "{}/{}.deb".format(dest, deb_name)
             )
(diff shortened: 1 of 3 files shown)


orndorffgrant:
Thanks I like this approach better!
Is there something missing though? I see some calls have changed to `debs.non_cloud_pro_image_debs(series)` but the definition hasn't changed?

orndorffgrant (approved):

features/util.py
```diff
@@ -52,12 +52,18 @@ class ProDebPaths:
...
+                ("ubuntu-advantage-tools", self.ubuntu_advantage_tools),
+                ("ubuntu-pro-client-l10n", self.ubuntu_pro_client_l10n),
+            ]
+        else:
+            return [
+                ("ubuntu-pro-client", self.ubuntu_pro_client),
+                ("ubuntu-pro-client-l10n", self.ubuntu_pro_client_l10n),
+            ]
 
     def cloud_pro_image_debs(self) -> List[Tuple[str, str]]:
```
orndorffgrant, line 68: Above calls `debs.cloud_pro_image_debs(series)` so I think we need a similar modification to accept `series` down here but for `ubuntu-advantage-pro`

features/steps/ubuntu_advantage_tools.py
```diff
@@ -97,24 +97,32 @@ def when_i_install_uat(context, machine_name=SUT):
...
         )
         to_install = []
-        for deb_name, deb_path in debs.non_cloud_pro_image_debs():
+        for deb_name, deb_path in debs.non_cloud_pro_image_debs(series):
             instance_tmp_path = "/tmp/behave_{}.deb".format(deb_name)
             instance.push_file(deb_path, instance_tmp_path)
             to_install.append(instance_tmp_path)
         if is_pro:
-            for deb_name, deb_path in debs.cloud_pro_image_debs():
+            for deb_name, deb_path in debs.cloud_pro_image_debs(series):
```
orndorffgrant, line 105: Or change this back
```suggestion
            for deb_name, deb_path in debs.cloud_pro_image_debs():
```

lucasmoura:
@orndorffgrant fixed

orndorffgrant:
@lucasmoura There are just two mypy errors of other callsites that need to be updated

lucasmoura:
:facepalm: Now it should be fixed @orndorffgrant

PR #3307, opened 2024-09-15
cli: add Table class

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because the new UX is full of tables, and this helps with formatting the tabular information and printing it to screen in a standardized way.

This is based on top of #3304 and should be merged after that.

## Test Steps
Unit tests cover the functionality. This will be widely used in the CLI so integration tests will include these changes, but for specific use cases.

---

### DIFF
--- a/uaclient/cli/formatter.py
+++ b/uaclient/cli/formatter.py
@@ -1,9 +1,10 @@
 import os
 import re
 import sys
-from typing import List
+from typing import List, Optional
 
 from uaclient.config import UAConfig
+from uaclient.messages import TxtColor
 
 COLOR_FORMATTING_PATTERN = r"\033\[.*?m"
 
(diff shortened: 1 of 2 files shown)


uaclient/cli/formatter.py
```diff
@@ -0,0 +1,175 @@
...
+    def ljust(string: str, total_length: int) -> str:
+        str_length = len_no_color(string)
+        if str_length >= total_length:
+            return string
+        return string + " " * (total_length - str_length)
+
+    def _validate_content(self) -> List[List[str]]:
+        if not self.headers and not self.rows:
+            raise ValueError(
+                "Empty table not supported. Please provide headers or rows."
```
lucasmoura, line 112: I think we need to move those exception messages to the `messages` module

uaclient/cli/formatter.py
```diff
@@ -0,0 +1,175 @@
...
+                "Empty row not supported. Please provide content for each row."
+            )
+
+        all_content = []
+        if self.headers:
+            all_content.append(self.headers)
+        if self.rows:
+            all_content.extend(self.rows)
+
+        expected_length = len(all_content[0])
```
lucasmoura, line 126: Do we need the `all_content` variable ? Can't the we do an if block like this:
`if not all(len(row) == len(self.headers) for row in self.rows)`

lucasmoura (approved):

uaclient/cli/formatter.py
renanrodrigo: The reason I didn't put those there is that this should not be translated.These tables should never depend on user input, and the network input that lands here needs to be sanitized in code.
Those issues are more like warnings to us, developers, to build the stuff the correct way.

If you disagree, we can address it in a separate PR

uaclient/cli/formatter.py
renanrodrigo: Not really, because tables may not have a header or not have rows - that's the reason I'm building the content there (and we calculate the columns width later based on this). Of course even with this restriction we could use a single block, but the variable helps with readability.
I just merged both into a single function, and I think it's better now. I kept the variable there for the same readability purposes.
If you disagree, please tell me and I can fix it in a future iteration.
Functionality is the same so we unblock you, @lucasmoura

Issue #3299, opened 2024-09-10
Bug: another process is running APT


VariableDeclared:
**Description of the bug**

If process fails, an APT error is hidden by the try/except block on [apt.py(450)](https://github.com/canonical/ubuntu-pro-client/blob/46c87bab8ca76d90962158e407d0132534f6ad2d/uaclient/apt.py#L450) 

**Expected behavior**

The APT error should be bubbled up, or show when `--debug` is passed.

**Current behavior**

The APT error will be hidden, preventing the user from understanding what has happened. In my customers case - this prevented them from spotting a mis-match in Ubuntu URL provided by the contracting server.

**To Reproduce**

1. Launch a client,
2. Register with offline contracting server
3. Attempt to enable esm-infra - ensure the esm repo URL is incorrect on the contracting server [ensure server is correct but URI is incorrect]

**System information:**

Ubuntu 24.04

**Additional context**

None

PR #3310, opened 2024-09-16
Update apt error logging

dheyay:
Fixes: #3299

## Why is this needed?

## Test Steps

---

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -153,6 +153,7 @@ def assert_valid_apt_credentials(repo_url, username, password):
                 retry_sleeps=APT_RETRIES,
             )
     except exceptions.ProcessExecutionError as e:
+        LOG.error("Error running apt-helper: %s", str(e))
         if e.exit_code == 100:
             stderr = str(e.stderr).lower()
             if re.search(r"401\s+unauthorized|httperror401", stderr):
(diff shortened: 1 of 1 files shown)


orndorffgrant (changes requested):

dheyay:
Rebased on  main @orndorffgrant

orndorffgrant (commented):
From the checklist

> - [x] I've properly referenced all bugs that this PR fixes

The bug is not properly referenced in the commit message.

Otherwise, looks great!

uaclient/apt.py
```diff
@@ -377,6 +379,7 @@ def run_apt_update_command(
             cmd=["apt-get", "update"], override_env_vars=override_env_vars
         )
     except exceptions.APTProcessConflictError:
+        LOG.warning("Error running apt update")
```
lucasmoura, line 382: Should we capture the exception and log it here too ?

uaclient/apt.py
```diff
@@ -448,6 +452,7 @@ def run_apt_install_command(
             override_env_vars=override_env_vars,
         )
     except exceptions.APTProcessConflictError:
```
lucasmoura, line 454: Same suggestion here

lucasmoura (approved):
Just two nits, but LGTM

uaclient/apt.py
dheyay: I think this one is covered by the logging on line `455`

uaclient/apt.py
dheyay: Good catch! Will update this one

orndorffgrant (approved):

PR #3313, opened 2024-09-17
tests: give jammy->noble upgrade its own test

orndorffgrant:
## Why is this needed?

jammy -> noble needs special treatment now that do-release-upgrade disables -proposed. We need a special step to ensure that the package is actually upgraded to the noble version. We also need to mask the apparmored services so that they don't cause denials during the upgrade to the noble version.

## Test Steps

Run the new test

---

### DIFF
--- a/features/ubuntu_upgrade.feature
+++ b/features/ubuntu_upgrade.feature
@@ -45,12 +45,73 @@ Feature: Upgrade between releases when uaclient is attached
       """
 
     Examples: ubuntu release
-      | release | machine_type  | next_release | prompt | devel_release   | service1  | service1_status | service2 | service2_status | before_cmd     |
-      | xenial  | lxd-container | bionic       | lts    |                 | esm-infra | enabled         | esm-apps | enabled         | true           |
-      | bionic  | lxd-container | focal        | lts    |                 | esm-infra | enabled         | esm-apps | enabled         | true           |
-      | bionic  | lxd-container | focal        | lts    |                 | usg       | enabled         | usg      | enabled         | pro enable cis |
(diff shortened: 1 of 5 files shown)


lucasmoura (approved):

features/ubuntu_upgrade.feature
```diff
@@ -45,12 +45,63 @@ Feature: Upgrade between releases when uaclient is attached
...
-      | focal   | lxd-container | jammy        | lts    |                 | esm-infra | enabled         | esm-apps | enabled         | true           |
-      | jammy   | lxd-container | noble        | lts    | --devel-release | esm-infra | enabled         | esm-apps | enabled         | true           |
+      | release | machine_type  | next_release | prompt | devel_release | service1  | service1_status | service2 | service2_status | before_cmd     |
+      | xenial  | lxd-container | bionic       | lts    |               | esm-infra | enabled         | esm-apps | enabled         | true           |
+      | bionic  | lxd-container | focal        | lts    |               | esm-infra | enabled         | esm-apps | enabled         | true           |
+      | bionic  | lxd-container | focal        | lts    |               | usg       | enabled         | usg      | enabled         | pro enable cis |
+      | focal   | lxd-container | jammy        | lts    |               | esm-infra | enabled         | esm-apps | enabled         | true           |
+
+  @slow @upgrade
```
orndorffgrant, line 54: reminder to self to add a comment here explaining the downsides of this approach and what we would ideally do instead

orndorffgrant:
@lucasmoura I added a comment and also removed some bionic wsl tests in accordance with our discussion a while back (and also bumped cloudlib - that was necessary for azure noble pro)

renanrodrigo (approved):

PR #3318, opened 2024-09-20
api: update cache for vulnerabilities

lucasmoura:
## Why is this needed?
We are now caching the vulnerability result as well. That means that if we detect that there is now new vulnerability JSON data to be used and no dpkg related changes on the system, we can simply reuse the old results.

## Test Steps
1. Verify that the vulnerabilities integration tests are still working 
2. Verify that the result cache is being produced
3. Verify when running two API commands in a row, the second one is way faster than the first
---

### DIFF
--- a/uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
+++ b/uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
@@ -1,10 +1,10 @@
 import abc
+import datetime
 import enum
 import json
 import os
 import re
-from datetime import datetime
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, NamedTuple, Optional, Tuple
 from urllib.parse import urljoin
 
 from uaclient import apt, exceptions, http, system, util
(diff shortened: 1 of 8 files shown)


uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -50,25 +64,28 @@ def __init__(
         cfg: UAConfig,
         data_file: Optional[str] = None,
         series: Optional[str] = None,
+        update_data: bool = True,
```
renanrodrigo, line 67: heh you did it here

uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -79,78 +96,112 @@ def _save_cache_publish_date(
...
+        cache_date_file = self._get_cache_file()
+        return cache_date_file.read() is not None
 
     def _is_cache_valid(
         self,
-        series: str,
         last_published_date: str,
         cache_date_file: DataObjectFile,
-    ) -> bool:
+    ) -> Tuple[bool, Optional[datetime.datetime]]:
```
renanrodrigo, line 125: I would suggest some comment here explaining what is this datetime's meaning

renanrodrigo (approved):

lucasmoura:
@renanrodrigo updated

orndorffgrant (changes requested):

orndorffgrant (commented):
As discussed, we should document the architecture of these caching mechanisms, when they are used, and when they are invalidated in some kind of dev-docs explanation.

lucasmoura:
Since we are now adding two more caches, I have created this diagram to simplify the visualization on how those caches are being used:

image

lucasmoura:
@orndorffgrant updated

orndorffgrant (approved):

PR #3321, opened 2024-09-20
Update Anbox page with version link

s-makin:
## Why is this needed?
The Anbox page should include 24.04 LTS availability.
I have reworded the version note box to point to the official 
requirements page that lists the LTS versions Anbox is available on.
This will mean we don't need to update the page every time a new
LTS is released.

```
doc: update Anbox Cloud requirements

Requirements were out of date, now include a link to the official
documentation as well.
```


catmsred (approved):
Looks good to me!

PR #3323, opened 2024-09-23
Printing standardized output for CLI commands

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it brings some tweaks to the Table implementation, includes the indented Block and SuggestionBlock implementation and adds a helper function which considers the formatter config when returning text.
This is the last base PR for the CLI before the commands start to appear.

## Test Steps
Unit tests cover the functionality
Integration and such will come with the commands

---

### DIFF
--- a/uaclient/cli/formatter.py
+++ b/uaclient/cli/formatter.py
@@ -1,7 +1,8 @@
 import os
 import re
 import sys
-from typing import List, Optional
+import textwrap
+from typing import Any, List, Optional
 
 from uaclient.config import UAConfig
 from uaclient.messages import TxtColor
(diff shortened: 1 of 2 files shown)


uaclient/cli/formatter.py
```diff
@@ -133,9 +130,15 @@ def _get_column_sizes(self) -> List[int]:
         return column_sizes
 
     def __str__(self) -> str:
+        return self.to_string()
+
+    def to_string(self, length: Optional[int] = None) -> str:
```
lucasmoura, line 154: can we rename this variable to `line_lenght` ? Just to make the meaning of this variable more explicit

uaclient/cli/formatter.py
```diff
@@ -182,3 +183,55 @@ def _fill_row(self, row: List[str]) -> str:
...
+        title: Optional[str] = None,
+        content: Optional[List[Any]] = None,
+    ):
+        self.title = title
+        self.content = content if content is not None else []
+
+    def __str__(self) -> str:
+        return self.to_string()
+
+    def to_string(self, length: Optional[int] = None) -> str:
```
lucasmoura, line 222: same `line_length` comment here

uaclient/cli/formatter.py
```diff
@@ -133,11 +133,12 @@ def _get_column_sizes(self) -> List[int]:
         return column_sizes
 
     def __str__(self) -> str:
+        rows = self.rows
```
lucasmoura, line 158: why do we need the `rows` variable if the fix is applied on `new_rows.append` call ?

lucasmoura (approved):

orndorffgrant (changes requested):

renanrodrigo:
Hopefully rebased this correctly against `main`

uaclient/cli/formatter.py
renanrodrigo: done

uaclient/cli/formatter.py
renanrodrigo: done

uaclient/cli/formatter.py
renanrodrigo: Because the new_rows thing is only done if the table exceeds the line length. If there is no line exceeding the available space, then the whole thing short-circuits to using the actual content in self.rows, without change.

renanrodrigo:
@lucasmoura I have created the ABC to represent Blocks and Tables and something else. That suffices for not caring about the isinstance call if we want to add new representations or derivatives of those in the future - so I think this is enough for the call; no need to convert anything nonstandard to an instance of that class. PTALALMK

renanrodrigo:
bad, bad `fix` CI

orndorffgrant (approved):
Thanks!

PR #3324, opened 2024-09-24
docs: update support matrix

s-makin:
docs: update support matrix

Update support matrix to include Noble and Oracular

### DIFF
--- a/docs/references/support_matrix.rst
+++ b/docs/references/support_matrix.rst
@@ -49,3 +49,12 @@ Here is a list of platforms and releases that Ubuntu Pro Client, through the
       - amd64, arm64, armhf, ppc64el, riscv64, s390x
       - TBD
       - Active SRU of all features
+    * - Noble
+      - amd64, arm64, armhf, ppc64el, riscv64, s390x
+      - TBD
+      - Active SRU of all features
+    * - Oracular
+      - amd64, arm64, armhf, ppc64el, riscv64, s390x
+      - TBD
+      - Active SRU of all features
+
--- a/docs/references/support_matrix.rst
+++ b/docs/references/support_matrix.rst
@@ -40,18 +40,18 @@ Here is a list of platforms and releases that Ubuntu Pro Client, through the
     * - Kinetic
       - amd64, arm64, armhf, ppc64el, riscv64, s390x
       - 27.11.2
-      - Active SRU of all features
+      - 28.1
     * - Lunar
       - amd64, arm64, armhf, ppc64el, riscv64, s390x
       - 27.14.4
-      - Active SRU of all features
+      - 30
     * - Mantic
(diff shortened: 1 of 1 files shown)


docs/references/support_matrix.rst
```diff
@@ -49,3 +49,12 @@ Here is a list of platforms and releases that Ubuntu Pro Client, through the
       - amd64, arm64, armhf, ppc64el, riscv64, s390x
       - TBD
       - Active SRU of all features
+    * - Noble
+      - amd64, arm64, armhf, ppc64el, riscv64, s390x
+      - TBD
```
renanrodrigo, line 54: Itinial release on Noble is `31.2.3`

docs/references/support_matrix.rst
```diff
@@ -49,3 +49,12 @@ Here is a list of platforms and releases that Ubuntu Pro Client, through the
       - amd64, arm64, armhf, ppc64el, riscv64, s390x
       - TBD
```
renanrodrigo, line 50: Initial release on Mantic is `29.4`

docs/references/support_matrix.rst
```diff
@@ -49,3 +49,12 @@ Here is a list of platforms and releases that Ubuntu Pro Client, through the
...
       - Active SRU of all features
```
renanrodrigo, line 51: Final release on Mantic is `32.3.1`

docs/references/support_matrix.rst
renanrodrigo: I can't highlight because it falls off the changes of the PR, but
Final release on Lunar is `30`
Final release on Kinetic is `28.1`

renanrodrigo (commented):
Thanks for adding the newest relases, @s-makin !
Given we are touching this, could we add the dates we know for some previous releases? Our big fault not to have them there yet :Z

s-makin:
> Thanks for adding the newest relases, @s-makin ! Given we are touching this, could we add the dates we know for some previous releases? Our big fault not to have them there yet :Z

Perfect, thanks for the info, I'll add it :)

lucasmoura:
I think we can remove the unsupported ones

s-makin:
> I think we can remove the unsupported ones

I've removed the non-LTS releases as requested. I also moved the common architectures outside the table to make it a bit more readable as the columns don't need to wrap anymore (hopefully). Let me know what you think - if you prefer them to remain inside the table I'll put them back! :)

s-makin:
> I think I personally would skim right passed the sentence about architectures supported on all releases and be confused by the short arch lists in the table 🙃 - but I don't have a strong opinion

Noted - I'll put them back then

renanrodrigo:
@orndorffgrant right so how historical this should be? should we insert entries for older-than-kinetic interims?

PR #3325, opened 2024-09-24
Add debug logging of masked exception

MggMuggins:
## Why is this needed?
Customers very occasionally run into this exception handling logic; the real error is eaten in favor of the unhelpfully generic `Another process is running APT`. A log entry would save us some consternation.

(Not sure which branch is appropriate here, happy to reopen if needed).

## Test Steps
Based on our records this happens most often when internal repos aren't mirroring i386 packages; I don't have a reproducer handy to double check. Please let me know if you need me to chase this down.

---

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -419,7 +419,8 @@ def update_sources_list(sources_list_path: str):
             with lock:
                 cache.update(fetch_progress, sources_list, 0)
         # No apt_pkg.Error on Xenial
-        except getattr(apt_pkg, "Error", ()):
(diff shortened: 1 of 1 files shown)


orndorffgrant:
Thank you for the contribution @MggMuggins - we appreciate it and it is super welcome :D

In this particular case, we already have a PR with an equivalent change as a subset here: https://github.com/canonical/ubuntu-pro-client/pull/3310/files so we'll just move forward with that one if that is okay with you.

Please re-open if I missed something and that other PR is not sufficient.

MggMuggins:
Whoops, nice catch. That'll do nicely, thanks!

PR #3326, opened 2024-09-24
chore: archive old sru verification scripts

orndorffgrant:
## Why is this needed?

This archives sru scripts that we don't need any more.

## Test Steps

N/A

---


renanrodrigo (approved):
hmm including a sneaky change to CI in an unrelated PR...

I like it

PR #3328, opened 2024-09-30
Update security status enum documentation

dheyay:
## Why is this needed?

Documentation update for the PR #3297 

## Test Steps

---

### DIFF
--- a/docs/explanations/how_to_interpret_the_security_status_command.rst
+++ b/docs/explanations/how_to_interpret_the_security_status_command.rst
@@ -320,6 +320,9 @@ will contain the following attributes:
 * **status**: The status for this update. It will be one of:
 
   * **"upgrade_available"**: The package can be upgraded right now.
+  * **"upgrade_available_not_preferred"**: The package can be upgraded but the
+    upgrade will not automatically selected by APT or other tools due to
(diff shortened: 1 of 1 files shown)


docs/explanations/how_to_interpret_the_security_status_command.rst
```diff
@@ -320,6 +320,8 @@ will contain the following attributes:
 * **status**: The status for this update. It will be one of:
 
   * **"upgrade_available"**: The package can be upgraded right now.
+  * **"upgrade_available_not_preferred"**: The package can be upgraded but the
+    upgrade is not the preferred candidate version.
```
renanrodrigo, line 324: I'd say we can explicitly say that `the package can be upgraded, but this upgrade will not be automatically selected by APT or other tools - due to lower pinning or to the existence of another preferred update.

renanrodrigo (commented):
I  think this should target `docs-devel`, as this will only be valid documentation when the fix is released, together with v35

docs/explanations/how_to_interpret_the_security_status_command.rst
dheyay: Updated!

lucasmoura (approved):

PR #3332, opened 2024-10-02
WIP: Convert devdocs to rST and promote to docs

s-makin:
## Why is this needed?
This PR solves all of our problems because...we need a full review and reorg of the devdocs. 

I've done the initial pass to convert and make sure it renders.
Between us, we now need to do a full audit to decide:
1) What content needs to be updated/deleted/added
2) How we want to organise it, e.g., do we want to have a maintainers subsection for "our" docs (e.g. the stuff about releases), and a contributors subsection for general development docs
3) If there's anything we don't want to surface through the docs to users (which can stay in GH).

When reviewing, please think about for each page:
* Does the page need to be exposed to the user
* Does the content need to be on a separate page or can it be folded into another (some pages are *very* short)
* Does the page belong in the development section, or should it be in the actual docs?
* What order should the pages be shown in?
* (etc)

```
doc: convert devdocs to rST and promote to docs
```


orndorffgrant:
@lucasmoura and I went over this and came up with the following list of things to do before we're comfortable publishing this.
* Delete the following pages:
  * Terminology
  * Architecture
  * Enabling a service
  * Directory layout
  * Daily Builds
* Update "How to run integration tests": (@orndorffgrant can do this)
* Move the following to "explanations" 
  * "Pull request review policy"
  * "Early review sign-off"
  * "What happens during attach" (also change it's reference to `pro attach TOKEN` to just `pro attach`)
* Re-title "systemd units" to something like "Auto attach mechanisms" and update `ubuntu-advantage-pro` package name to `ubuntu-pro-auto-attach` within it

s-makin:
Closing this PR as it is replaced by this one https://github.com/canonical/ubuntu-pro-client/pull/3353

Issue #3331, opened 2024-10-02
Bug: Trying to enable landscape says it's not available for releases older than Noble


renanrodrigo:
**Description of the bug**

Landscape is available for all supported releases, but it was only defined as a pro service from 23.10 (mantic).
This means that trying to enable Landscape using Pro works fine on Noble, for instance, but on other releases it gives misleading information, telling the user that Landscape is not available to the specific release.

We need to implement better communication, informing users that Landscape is actually available, just not through Pro.

**Expected behavior**

```
$ sudo pro enable landscape
Landscape cannot be enabled via Pro Client on Ubuntu 22.04 and earlier.
Please manually install Landscape: https://ubuntu.com/landscape/install
```

**Current behavior**

```
$ sudo pro enable landscape
Landscape is not available for Ubuntu 22.04 LTS (Jammy Jellyfish).
```

**To Reproduce**

1. Launch a Focal/Jammy VM/Container
2. Attach to a subscription
3. Try to enable Landscape

**System information:**

- Ubuntu release: xenial(?), bionic(?), focal, jammy
- Pro Client version: 34

orndorffgrant:
We need to verify which releases we put the new message on (do we include xenial?)

Also https://ubuntu.com/landscape/install is a confusing landing page for installing the client, I'll reach out to landscape for a better link

PR #3333, opened 2024-10-03
Update landscape inapplicability message

dheyay:
## Why is this needed?

Fixes: #3331 

## Test Steps

---

### DIFF
--- a/uaclient/entitlements/landscape.py
+++ b/uaclient/entitlements/landscape.py
@@ -3,7 +3,10 @@ from typing import Any, Dict, Optional, Tuple
 
 from uaclient import api, event_logger, exceptions, messages, system, util
 from uaclient.entitlements.base import UAEntitlement
-from uaclient.entitlements.entitlement_status import ApplicationStatus
+from uaclient.entitlements.entitlement_status import (
+    ApplicabilityStatus,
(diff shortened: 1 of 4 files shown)


uaclient/entitlements/landscape.py
```diff
@@ -74,6 +77,23 @@ def application_status(
...
+            affordance = self.entitlement_cfg["entitlement"].get(
+                "affordances", {}
+            )
+            affordance_series = affordance.get("series", None)
+            current_release = system.get_release_info()
+            if current_release not in affordance_series:
+                return (
+                    ApplicabilityStatus.INAPPLICABLE,
+                    messages.LANDSCAPE_INAPPLICABLE,
+                )
```
orndorffgrant, line 94: Let's make it exactly the same as the super() impl of this check just to be safe
```
        if (
            self.affordance_check_series
            and affordance_series is not None
            and system.get_release_info().series not in affordance_series
        ):
            return (
                ApplicabilityStatus.INAPPLICABLE,
                messages.LANDSCAPE_INAPPLICABLE
            )

```

orndorffgrant (approved):
Thanks! Just tested and it works well.

PR #3337, opened 2024-10-07
Drop python3-pkg-resources dependency on Oracular

lucasmoura:
## Why is this needed?
We don't need this package dependency on Oracular and for the upcoming releases. Since we are on ubuntu-minimal, we are now removing that dependency from Oracular forward to not bring an unneeded dependency into our minimal images.

LP: #2083665

## Test Steps
Check that the control file for packages before Oracular include `python3-pkg-resources` and the Oracular package doesn't

---


debian/rules
```diff
@@ -49,7 +61,7 @@ endif
 endif
 
 override_dh_gencontrol:
-	echo "extra:Depends=$(APT_PKG_DEPS) $(DISTRO_INFO_DEPS)" >> debian/ubuntu-pro-client.substvars
+	echo "extra:Depends=$(APT_PKG_DEPS) $(DISTRO_INFO_DEPS) ${PYTHON3_PKG_RESOURCES}" >> debian/ubuntu-pro-client.substvars
```
orndorffgrant, line 64: Is there a reason we're using curly braces for this one when we use parentheses for the other similar variables?

renanrodrigo (approved):

debian/rules
lucasmoura: Good point, they are now consistent

Issue #3335, opened 2024-10-04
Bug: version check script in tools doesn't handle X.Y.Z style versions


panlinux:
**Description of the bug**

When we build Pro with a version like 34.1.1, the script hilariously fails like this:
```
# fail early if version.py wasn't updated to match d/changelog's version
# only take into account the actual upstream version, discarding ubuntu and
# backport suffixes
python3 tools/check-versions-are-consistent.py
version.py says "34.1.1" but changelog says "34.1.1"
```

**Expected behavior**

It should correctly compare the version in version.py and debian/changelog.

**Current behavior**
It fails if the version has three components (X.Y.Z).

**To Reproduce**
Change version.py and debian/changelog to 34.1.1 and build the deb package.

**System information:**

- Ubuntu release: oracular
- Pro Client version: 34.1.1
- 
**Additional context**

Patch that was uploaded to oracular:
```diff
--- a/tools/check-versions-are-consistent.py
+++ b/tools/check-versions-are-consistent.py
@@ -20,7 +20,7 @@ changelog_version = (
 # ppa test builds: 32.3~22.04~ppa1 -> 32.3
 # devel: 1:1+devel -> 1:1+devel
 #
-m = re.match(r"((\d+:\d+\+devel)|\d+(\.\d+)?)", changelog_version)
+m = re.match(r"((\d+:\d+\+devel)|\d+(\.\d+)*)", changelog_version)
 if m:
     base_changelog_version = m.group()
 else:
```

Note the script should also print a more useful error message, i.e., include `base_changelog_version`.

PR #3338, opened 2024-10-07
Version script and apparmor fix for v-35

dheyay:
## Why is this needed?

Fixes: #3335 

Also includes fix for the following apparmor issue seen on CI:

```
WARNING:root:XXX apparmor DENIED begin
WARNING:root:Oct  6 21:15:35 fv-az1024-661 kernel: [ 1566.602397] audit: type=1400 audit(1728249335.033:2316): apparmor="DENIED" operation="capable" class="cap" namespace="root//lxd-pro-jammy-sut-1006-211402949805_<var-snap-lxd-common-lxd>" profile="ubuntu_pro_apt_news" pid=41839 comm="python3" capability=0  capname="chown"
WARNING:root:XXX apparmor DENIED end
```

## Test Steps

---

### DIFF
--- a/debian/apparmor/ubuntu_pro_apt_news.jinja2
+++ b/debian/apparmor/ubuntu_pro_apt_news.jinja2
@@ -20,6 +20,8 @@ profile ubuntu_pro_apt_news flags=(attach_disconnected) {
   # GH: 3079
   capability dac_override,
   capability kill,
+  capability chown,
+  capability fowner,
 
   /etc/apt/** r,
   /etc/default/apport r,


debian/apparmor/ubuntu_pro_apt_news.jinja2
```diff
@@ -20,6 +20,8 @@ profile ubuntu_pro_apt_news flags=(attach_disconnected) {
   # GH: 3079
   capability dac_override,
   capability kill,
+  capability chown,
+  capability fowner,
```
panlinux, line 24: Both of these were needed for `chown`?

dheyay: Yes, looks like it, but could use a second opinion in case this goes against recommended practice.

When testing with just `capability chown`, the tests on `apt_messages.feature` failed with another denial related to `fowner`. Adding both works

panlinux: Ok, thanks, that's good enough

panlinux (approved):
+1

lucasmoura (approved):

renanrodrigo (approved):

PR #3339, opened 2024-10-07
fix: use USNs directly from the CVE response

lucasmoura:
## Why is this needed?
We need to fetch the USNs related to a CVE to extract the ubuntu version of the packages affected by the CVE. In the past, we where running another request to fetch all USNs related to the CVEs. To make this process less demanding on the API side, we are now relaying on the related USNs already delivered by the CVE endpoint instead

## Test Steps
Check that CVE related tests for the fix command are still passing

---

### DIFF
--- a/uaclient/api/u/pro/security/fix/_common/plan/v1.py
+++ b/uaclient/api/u/pro/security/fix/_common/plan/v1.py
@@ -742,7 +742,7 @@ def _get_cve_data(
 ) -> Tuple[CVE, List[USN]]:
     try:
         cve = client.get_cve(cve_id=issue_id)
-        usns = client.get_notices(cves=issue_id)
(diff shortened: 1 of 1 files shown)


renanrodrigo:
@lucasmoura I think this change is fine and I'd like to merge it, but I see that fix tests still fail with 503s on CI here. What should we do about this?

lucasmoura:
Yes, I know, but this fix is just an attempt to improve the CVE situation, it will not fully solve the issue, unfortunately

renanrodrigo:
I'm going to merge this one. Although it does not solve the problems, it reduces the number of API calls in a fix request, which is a nicety per se,

renanrodrigo (approved):

PR #3341, opened 2024-10-08
esm: add snapshot urls to apt auth

orndorffgrant:
## Why is this needed?

esm-apps and esm-infra now have apt snapshots available; however, without the apt auth configured properly, downloads from the snapshot servers will fail. This ensures new enablements will include the snapshot urls, and it migrates existing enablements to include the snapshot urls on upgrade.

## Test Steps

Run the new integration tests
Consider how I did the one-time sru test in a feature file

---

### DIFF
--- a/features/esm.feature
+++ b/features/esm.feature
@@ -1,0 +1,106 @@
+Feature: ESM Resource Specificities
+
+  Scenario Outline: esm apt auth includes snapshot urls
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I attach `contract_token` with sudo and options `--no-auto-enable`
+    When I run `pro enable esm-infra esm-apps` with sudo
+    When I run `cat /etc/apt/auth.conf.d/90ubuntu-advantage` with sudo
+    Then stdout contains substring:
+      """
+      machine esm.ubuntu.com/infra/ubuntu/ login bearer password m
+      """
+    Then stdout contains substring:
+      """
+      machine snapshot.infra-security.esm.ubuntu.com/infra/ubuntu/ login bearer password m
+      """
+    Then stdout contains substring:
+      """
+      machine snapshot.infra-updates.esm.ubuntu.com/infra/ubuntu/ login bearer password m
+      """
+    Then stdout contains substring:
+      """
+      machine esm.ubuntu.com/apps/ubuntu/ login bearer password m
+      """
+    Then stdout contains substring:
+      """
+      machine snapshot.apps-security.esm.ubuntu.com/apps/ubuntu/ login bearer password m
+      """
+    Then stdout contains substring:
+      """
+      machine snapshot.apps-updates.esm.ubuntu.com/apps/ubuntu/ login bearer password m
+      """
+
+    Examples: ubuntu release
+      | release | machine_type  |
+      | xenial  | lxd-container |
+      | bionic  | lxd-container |
+      | focal   | lxd-container |
+      | jammy   | lxd-container |
+      | noble   | lxd-container |
+
+  @oneoff
+  Scenario Outline: attached upgrade to v35 adds snapshot urls for esm
+    Given a `<release>` `<machine_type>` machine
+    When I apt update
+    When I apt upgrade
+    When I attach `contract_token` with sudo and options `--no-auto-enable`
+    When I run `pro enable esm-infra esm-apps` with sudo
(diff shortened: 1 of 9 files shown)


lucasmoura (commented):
It seems that some tests are failing due to this change:
- `APT auth file is edited correctly on enable`
- `Attached detach in an ubuntu machine`

Those seems like valid failures and it is worth double checking them

features/esm.feature
```diff
@@ -0,0 +1,106 @@
...
+      """
+    Then stdout does not contain substring:
+      """
+      machine snapshot.apps-security.esm.ubuntu.com/apps/ubuntu/ login bearer password m
+      """
+    Then stdout does not contain substring:
+      """
+      machine snapshot.apps-updates.esm.ubuntu.com/apps/ubuntu/ login bearer password m
+      """
+    When I install ubuntu-advantage-tools
```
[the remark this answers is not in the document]
orndorffgrant: That's the step name that installs either ubuntu-advantage-tools or ubuntu-pro-client depending on the ubuntu release :upside_down_face:

orndorffgrant:
@lucasmoura updated so those tests should pass now

lucasmoura (approved):

uaclient/entitlements/esm.py
```diff
@@ -97,6 +97,10 @@ class ESMAppsEntitlement(ESMBaseEntitlement):
     description = messages.ESM_APPS_DESCRIPTION
     help_text = messages.ESM_APPS_HELP_TEXT
     repo_key_file = "ubuntu-pro-esm-apps.gpg"
+    snapshot_urls = [
+        "snapshot.apps-security.esm.ubuntu.com/apps/ubuntu/",
+        "snapshot.apps-updates.esm.ubuntu.com/apps/ubuntu/",
```
renanrodrigo, line 102: I can't say I like those values hardcoded here, as changing them in the future will require another SRU of Pro to conclude, and more migrations.

Fetching it from APT should be possible but I don't know any good way of doing it without hacking the metadata from the release file, and that's a no.


Given the APT url for a service already comes from the contracts server, do you see any harm in getting that from CS as well? If the answer is no because of the time we have, could we please consider that for the near future, before other services have their own snapshots?

renanrodrigo (approved):
LGTM with a single concern, which can be addressed later if you wish

renanrodrigo:
CI checked, errors unrelated.

orndorffgrant:
@renanrodrigo updated based on our discussion to use snapshot urls from the backend on normal enablement and use the hardcoded values for the migration.

lib/add_esm_snapshot_auth.py
```diff
@@ -0,0 +1,59 @@
...
+
+def add_esm_snapshot_auth(cfg):
+    if not _is_attached(cfg).is_attached:
+        LOG.info("Not attached. Ending.")
+        return
+    for esm, urls in [
+        (ESMAppsEntitlement(cfg), ESM_APPS_SNAPSHOT_URLS),
+        (ESMInfraEntitlement(cfg), ESM_INFRA_SNAPSHOT_URLS),
+    ]:
+        print(esm.apt_url)
```
lucasmoura, line 28: Let's remove this print here

lib/add_esm_snapshot_auth.py
```diff
@@ -0,0 +1,59 @@
...
+]
+ESM_INFRA_SNAPSHOT_URLS = [
+    "snapshot.infra-security.esm.ubuntu.com/infra/ubuntu/",
+    "snapshot.infra-updates.esm.ubuntu.com/infra/ubuntu/",
+]
+
+
+def add_esm_snapshot_auth(cfg):
+    if not _is_attached(cfg).is_attached:
+        LOG.info("Not attached. Ending.")
```
lucasmoura, line 22: Can we just log that we are not adding the `esm snapshot` url here ?

lucasmoura (approved):

renanrodrigo (approved):

lib/add_esm_snapshot_auth.py
orndorffgrant: :upside_down_face:

lib/add_esm_snapshot_auth.py
orndorffgrant: Yep good call

PR #3344, opened 2024-10-08
Separate CIS and USG in the client

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it implements the changes needed to turn USG into an actual resource on the client side. This aligns client behaviour with the implementation done for the backend

## Test Steps
kept behavior:
- enabling CIS should work on X, B
- enabling USG should work on J
- both should work on F, but USG gets enabled in both cases , with a message telling about the transition

new behavior:
- upgrading from older versions of pro client where CIS was enabled as USG on F, J should enable the actual USG resource

Integration tests TBD
---

### DIFF
--- a/lib/postinst-migrations.sh
+++ b/lib/postinst-migrations.sh
@@ -45,6 +45,7 @@
 set -e
 
 PREVIOUS_PKG_VER=$1
+UBUNTU_RELEASE=$2
 
 if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "32~"; then
     # When we perform the write operation through
@@ -86,5 +87,16 @@ if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "35~"; then
     # is updated to include the snapshot urls for esm-infra and esm-apps
     if [ -f "/var/lib/ubuntu-advantage/private/machine-token.json" ]; then
         /usr/bin/python3 /usr/lib/ubuntu-advantage/add_esm_snapshot_auth.py
+    fi
+
+    # On Focal and Jammy, the CIS service transitioned to USG from v35
+    # This changes the lists and authentication files to represent that
+    # in case CIS is enabled
+    if [ "$UBUNTU_RELEASE" = "20.04" ] || [ "$UBUNTU_RELEASE" = "22.04" ]; then
+        if [ -f /etc/apt/sources.list.d/ubuntu-cis.list ]; then
+            mv /etc/apt/sources.list.d/ubuntu-cis.list /etc/apt/sources.list.d/ubuntu-usg.list
+            sed -i 's#/cis/ubuntu#/usg/ubuntu#' /etc/apt/sources.list.d/ubuntu-usg.list
+            sed -i 's#/cis/ubuntu#/usg/ubuntu#' /etc/apt/auth.conf.d/90ubuntu-advantage
+        fi
     fi
 fi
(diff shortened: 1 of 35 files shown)


uaclient/cli/enable.py
```diff
@@ -449,6 +447,13 @@ def action_enable(args, *, cfg, **kwargs) -> int:
     enabled_services = _enabled_services(cfg).enabled_services
     all_dependencies = _dependencies(cfg).services
 
+    # On Focal, CIS transitioned to USG.
+    # If a user tries to enable CIS on Focal, they should get USG instead.
+    if "cis" in names and system.get_release_info().series == "focal":
+        entitlements_found.append("usg")
+        entitlements_not_found.remove("cis")
+        interactive_only_print(messages.CIS_IS_NOW_USG)
```
lucasmoura, line 455: I wonder if this logic should sit on the entitlements module instead. The rationale is that this should also apply to the enable API, which uses the `entitlement_factory` function. I know that this function is not used here, but I think the entitlement module should be the one aware of this situation, and not the CLI module

orndorffgrant (commented):
Looking good! A few little things.
I guess we can't really test yet until the resource is up on the backend

uaclient/cli/enable.py
renanrodrigo: The problem is the still somehow - as we need to do this during the enablement step, but not for other commands.
If we try to be always smart, the workaround will grow bigger than we want to maintain imho.
I'm leaning towards adding a similar snippet in the API call and live with it.

orndorffgrant: I think that would be okay. I think ideally we would still install the cis packages on focal too, though, unless product has agreed that it's okay to break that?

renanrodrigo: So I have changed the USG class to accomodate the message+packageset changes for CIS on Focal, and applied the logic for both enablement flows.

lib/postinst-migrations.sh
```diff
@@ -81,4 +82,15 @@ if dpkg --compare-versions "$PREVIOUS_PKG_VER" lt "35~"; then
     if [ -f "/var/lib/ubuntu-advantage/private/machine-token.json" ]; then
         touch "/var/lib/ubuntu-advantage/marker-only-series-check"
     fi
+
+    # On Focal and Jammy, the CIS service transitioned to USG from v35
+    # This changes the lists and authentication files to represent that
+    # in case CIS is enabled
+    if [ "$VERSION_ID" = "20.04" ] || [ "$VERSION_ID" = "22.04" ]; then
```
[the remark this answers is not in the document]
renanrodrigo: The only thing missing in this PR now is integration tests. I have manually tested it many times up to the point where we can without the backends, and will have the behave stuff in place asap

renanrodrigo:
@orndorffgrant @lucasmoura this is done feature-wise, just missing the beave testing.
If you want to merge this to finalize the release code, I can send a separate PR with just the tests.

orndorffgrant (approved):
I'm okay merging this ahead of integration tests, but would like @lucasmoura to approve as well before we do

lucasmoura (approved):

uaclient/api/u/pro/services/enable/v1.py
```diff
@@ -170,16 +171,27 @@ def _enable(
...
         raise exceptions.NotSupported()
 
+    # On Focal, CIS transitioned to USG.
+    # If a user tries to enable CIS on Focal, they should get USG instead.
+    if (
+        service_to_enable == "cis"
+        and system.get_release_info().series == "focal"
+    ):
+        service_to_enable = "usg"
+        USGEntitlement.cis_version = True
```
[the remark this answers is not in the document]
renanrodrigo: I'm hate hating this for years now I hate this part a little less

PR #3346, opened 2024-10-10
Update vulnerability list applied fixes message

lucasmoura:
## Why is this needed?
Instead of showing the count of applied fixes per package, we are now switching to count be vulnerability instead. The
main reason is to make the count consistent with the other summary information we have on pro vulnerability list

## Test Steps

---

### DIFF
--- a/uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
+++ b/uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
@@ -4,6 +4,7 @@ import enum
 import json
 import os
 import re
+from collections import defaultdict
 from functools import lru_cache
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple
 from urllib.parse import urljoin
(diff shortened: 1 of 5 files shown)


uaclient/api/u/pro/security/vulnerabilities/_common/v1.py
```diff
@@ -328,6 +328,47 @@ def _get_vulnerability_fix_status(
 )
 
 
+class VulnerabilitiesAlreadyFixed:
+    def __init__(self):
+        self.vulns = {}
```
renanrodrigo, line 333:
```suggestion
        self.vulns = defaultdict(lambda: {"issues": set(), "priority_count": defaultdict(int)})
```

renanrodrigo: (lonely comment left here based on conversation context in other channel)
This is a suggestion to simplify the add function

renanrodrigo (approved):

PR #3348, opened 2024-10-15
cli: add spinning wheel to vulnerability list

lucasmoura:
## Why is this needed?
Add a spinning wheel when users are running the pro vulnerability list command. This should better inform users that some processing is happening in the background

## Test Steps
Check the vulnerability list integration test still works and that the spinning wheel can be seen when running the command on tty terminal

---


orndorffgrant (approved):
Pre-approving, pending the outcome of the discussion around potentially retrying the head requests

PR #3351, opened 2024-10-17
Review 35 feedback

lucasmoura:
## Why is this needed?
Address the review feedback from @bryceharrington 

## Test Steps
Check that no integration tests are now failing due to this change

---

### DIFF
--- a/uaclient/api/u/pro/config/v1.py
+++ b/uaclient/api/u/pro/config/v1.py
@@ -101,7 +101,7 @@ class ConfigInfo(DataObject, AdditionalInfo):
                 " 'available'. If set to 'on', the guest will attach using the"
                 " host's Pro access automatically on launch. If set to 'off',"
                 " the guest will not be allowed to attach using the host's Pro"
-                " lx access. If set to 'available', the guest will be allowed"
+                " access. If set to 'available', the guest will be allowed"
                 " to attach using the host's Pro access, but it will not be"
                 " automatic."
             ),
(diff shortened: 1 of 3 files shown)


renanrodrigo (approved):

uaclient/api/u/pro/config/v1.py
```diff
@@ -101,7 +101,7 @@ class ConfigInfo(DataObject, AdditionalInfo):
                 " 'available'. If set to 'on', the guest will attach using the"
                 " host's Pro access automatically on launch. If set to 'off',"
                 " the guest will not be allowed to attach using the host's Pro"
-                " lx access. If set to 'available', the guest will be allowed"
```
orndorffgrant, line 104: I think the more correct statement would be "...allowed to attach using the host's Pro access...". So I think we shuold remove "lx"/"lxd" entirely

uaclient/http/__init__.py
```diff
@@ -398,6 +398,8 @@ def readurl(
             json_dict = json_body
         elif isinstance(json_body, list):
             json_list = json_body
+        else:
+            LOG.warning("unexpected JSON response: %s", str(json_body))
```
orndorffgrant, line 402: There is just a typo in the commit message of this one. It says "we will **not** log..." but should say "we will now log..."

orndorffgrant (approved):
pre-emptively approving pending my minor suggestions

PR #3353, opened 2024-10-18
Move dev-docs back to main branch

s-makin:
# Why is this needed?

We wanted to take a close look at the devdocs, which we did in [PR 3332](https://github.com/canonical/ubuntu-pro-client/pull/3332).

It's been agreed that rather than publishing the dev docs as a "contributing" or "development" section in the docs, we want to move them back to the main branch instead for two reasons:

* They don't need to be kept updated alongside the Client release schedule
  and so don't need to be in the docs branch
* Having them in the docs branch makes it harder to keep them up to date
  with process/policy changes because they're too far removed from the code.
  By moving them back, it should be easier to keep them up to date

# Please Squash this PR with this commit message

```
doc: move development docs to main branch
```


orndorffgrant:
Somehow all the pages got deleted in the last commit :upside_down_face: 
Maybe revert that commit and try the updates to the contributing page again?

s-makin:
> Somehow all the pages got deleted in the last commit 🙃 Maybe revert that commit and try the updates to the contributing page again?

Turns out I forgot to commit when I added them to the new place :upside_down_face:

orndorffgrant:
Replaced by #3369

PR #3365, opened 2024-11-05
Fix apt-hook text output format on recent releases (resolute + )

dheyay:
## Why is this needed?

This PR solves all of our problems because it correctly formats apt news and apt messages for oracular.

Fixes: LP #2086753

## Test Steps

---

### DIFF
--- a/apt-hook/json-hook.cc
+++ b/apt-hook/json-hook.cc
@@ -3,7 +3,6 @@
 #include <iostream>
 #include <json-c/json.h>
 #include <libintl.h>
-#include <sstream>
 #include <string>
 #include <vector>
 
@@ -96,6 +95,32 @@ bool version_from_origin_and_archive_ends_with(json_object *version, std::string
     }
 
     return false;
+}
+
+release_info get_release_info() {
+    release_info info = {"", "", false};
+    std::ifstream machine_series_file("/etc/os-release");
+    if (machine_series_file.is_open()) {
+        std::string currentLine;
+        while(std::getline(machine_series_file, currentLine)) {
+            size_t position = currentLine.find("="); // Find position of equals sign as separator
+            std::string key = currentLine.substr(0, position);
+            std::string value = currentLine.substr(position + 1);
+
+            if (!value.empty() && value.front() == '"' && value.back() == '"') {
+                value = value.substr(1, value.length() - 2);
+            }
+
+            if (key=="VERSION_ID") {
+                info.release = value;
+            } else if (key == "VERSION_CODENAME") {
+                info.series = value;
+            } else if (key == "PRETTY_NAME" && value.find("LTS") != std::string::npos) {
+                info.is_lts = true;
+            }
+        }
+    }
+    return info;
 }
 
 bool count_security_packages_from_apt_stats_json(json_object *stats, security_package_counts &result) {
@@ -193,7 +218,7 @@ bool collect_pro_packages_from_pre_prompt_json(json_object *pre_prompt, std::vec
             continue;
         }
         std::string package_mode(json_object_get_string(tmp));
-        
+
         has_key = json_object_object_get_ex(package, "name", &tmp);
         if (!has_key) {
             continue;
@@ -234,21 +259,29 @@ bool collect_pro_packages_from_pre_prompt_json(json_object *pre_prompt, std::vec
 }
 
 #define MAX_COUNT_MESSAGE_LEN 256
-std::string create_count_message(security_package_counts &counts) {
+std::string create_count_message(security_package_counts &counts, release_info info) {
     char buf[MAX_COUNT_MESSAGE_LEN] = {0};
+    std::string lts_text = info.is_lts ? "LTS " : "";
 
     if (counts.esm_apps == 0) {
         if (counts.esm_infra == 0) {
             if (counts.standard == 0) {
                 return "";
             } else if (counts.standard == 1) {
-                return std::string(gettext("1 standard LTS security update"));
+                std::snprintf(
+                    buf,
+                    MAX_COUNT_MESSAGE_LEN,
+                    gettext("1 standard %ssecurity update"),
+                    lts_text.c_str()
+                );
+                return std::string(buf);
             } else if (counts.standard > 1) {
                 std::snprintf(
                     buf,
                     MAX_COUNT_MESSAGE_LEN,
-                    gettext("%lu standard LTS security updates"),
-                    counts.standard
+                    gettext("%lu standard %ssecurity updates"),
+                    counts.standard,
+                    lts_text.c_str()
                 );
                 return std::string(buf);
             }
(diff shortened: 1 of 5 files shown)


apt-hook/json-hook.cc
```diff
@@ -236,18 +262,20 @@ bool collect_pro_packages_from_pre_prompt_json(json_object *pre_prompt, std::vec
...
             } else if (counts.standard == 1) {
-                return std::string(gettext("1 standard LTS security update"));
+                return std::string(gettext("1 standard ")) + lts_text + gettext("security update");
             } else if (counts.standard > 1) {
                 std::snprintf(
                     buf,
                     MAX_COUNT_MESSAGE_LEN,
-                    gettext("%lu standard LTS security updates"),
+                    (std::string(gettext("%lu standard ")) + lts_text + gettext("security updates")).c_str(),
                     counts.standard
```
orndorffgrant, line 279: We should insert the "LTS " via string formatting, not concatenation. We can use the existing `std::snprintf` calls already being used to format in the counts. Something like this
```suggestion
                    gettext("%lu standard %ssecurity updates"),
                    counts.standard,
                    lts_text
```

(This applies to all of the similar cases in this function)

apt-hook/json-hook.hh
```diff
@@ -13,10 +13,16 @@ struct security_package_counts {
     long unsigned int esm_infra;
     long unsigned int esm_apps;
 };
+struct release_info {
+    std::string release; // e.g. "24.04"
+    std::string series;  // e.g. "noble"
```
orndorffgrant, line 18: I don't think `series` is currently used anywhere.

orndorffgrant (commented):
Sorry for the very long delay.

apt-hook/json-hook.cc
```diff
@@ -690,9 +722,15 @@ int run()
...
+            release_info info = get_release_info();
             std::string message = create_count_message(counts);
             if (message != "") {
-                std::cout << message << std::endl;
+                if (info.release != "" && info.release >= "24.10") {
+                    std::cout << "  " << message << std::endl;
+                }
+                else {
+                    std::cout << message << std::endl;
+                }
```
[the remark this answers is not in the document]
dheyay: We talked about this in our pair programming session. We can just keep as is, based on our discussion.

apt-hook/json-hook.hh
dheyay: We will use the series variable in the esm_infra_series function

lucasmoura (commented):
@dheyay I know that you are adding a test for APT news, but I wonder if we should add an orocular integration test for the `json-hook` changes

dheyay:
> @dheyay I know that you are adding a test for APT news, but I wonder if we should add an orocular integration test for the `json-hook` changes

@lucasmoura I would say we can probably do without a dedicated test since the updated test acts like a dedicated oracular test for the json-hook. It checks the formatting for the apt-news message as well as the summary for the number of packages.

Let me know if that works, I would not mind adding a short test to check for formatting of updates/packages as well.

dheyay:
Updated with the test @lucasmoura @orndorffgrant

uaclient/apt_news.py
```diff
@@ -263,11 +263,19 @@ def local_apt_news(cfg: UAConfig) -> Optional[str]:
...
+    if system.get_release_info().series == "oracular":
+        prefix = "\nAPT news:"
+        for line in lines:
+            formatted_lines.append("  " + line)
+        return "{0}\n{1}\n\n".format(prefix, "\n".join(formatted_lines))
+    else:
+        prefix = "#"
+        for line in lines:
+            formatted_lines.append(prefix + " " + line)
+        return "{0}\n{1}\n{0}\n".format(prefix, "\n".join(formatted_lines))
```
orndorffgrant, line 278: I think this is still less clear than something like the original. For example:

```suggestion
    if system.get_release_info().series == "oracular":
        result = "\nAPT news:"
        for line in lines:
            result += "  {}\n".format(line)
        result += "\n"
        return result
    else:
        result = "#\n"
        for line in lines:
            result += "# {}\n".format(line)
        result += "#\n"
        return result
```

apt-hook/json-hook.cc
```diff
@@ -1,3 +1,4 @@
+#include <cstdio>
```
orndorffgrant, line 1: Why does this need to be included?

apt-hook/json-hook.cc
```diff
@@ -429,8 +499,9 @@ std::string create_count_message(security_package_counts &counts) {
                 std::snprintf(
                     buf,
                     MAX_COUNT_MESSAGE_LEN,
-                    gettext("%lu standard LTS security updates, %lu esm-infra security updates and %lu esm-apps security updates"),
+                    gettext("%lu standard %ssecuriy updates, %lu esm-infra security updates and %lu esm-apps security updates"),
```
orndorffgrant, line 502:
```suggestion
                    gettext("%lu standard %ssecurity updates, %lu esm-infra security updates and %lu esm-apps security updates"),
```

orndorffgrant (commented):
Please run the hook unit tests: `make -C apt-hook test` (Github CI is failing on package builds because the build runs those tests).

Also:

> - [x] I've properly referenced all bugs that this PR fixes

is not accurate since the comment displays "None" under "Bug references". Please reference the LP bug in a commit in the format "LP: #12345678" (No "Fixes").

Also please clean up the commits in the PR. These can probably all be squashed into one.

apt-hook/json-hook.cc
dheyay: Does not need to be there, removed now

uaclient/apt_news.py
dheyay: the string formatting was the issue, making it a hard read here, apologies. Fixed!

orndorffgrant (commented):
I'm still running all the affected integration tests, but noticed two things.

uaclient/apt_news.py
```diff
@@ -263,8 +263,17 @@ def local_apt_news(cfg: UAConfig) -> Optional[str]:
 
 
 def format_news_for_apt_update(news: str) -> str:
+    lines = news.split("\n")
+
+    if system.get_release_info().release >= "24.10":
+        result = "\nAPT news:"
```
orndorffgrant, line 269: One test failed: features/apt_messages.feature:755 with

```
        Reading package lists...
        Building dependency tree...
        Reading state information...
        Calculating upgrade...

        APT news:
          one

        Summary:
          Upgrading: 0, Installing: 0, Removing: 0, Not Upgrading: 0
      But got:
        Reading package lists...
        Building dependency tree...
        Reading state information...
        Calculating upgrade...

        APT news:  one

        Summary:
          Upgrading: 0, Installing: 0, Removing: 0, Not Upgrading: 0

```

I think this will solve it
```suggestion
        result = "\nAPT news:\n"
```

dheyay: I was just running these locally, haha, thanks for the find!

apt-hook/json-hook.test.cc
```diff
@@ -8,42 +8,63 @@ BOOST_AUTO_TEST_SUITE(JSON_Hook)
...
 
-void count_message_test(int standard_count, int infra_count, int apps_count, std::string expected_message) {
+void count_message_test(int standard_count, int infra_count, int apps_count, bool is_lts, std::string expected_message) {
     security_package_counts counts;
     counts.standard = standard_count;
     counts.esm_infra = infra_count;
     counts.esm_apps = apps_count;
-    std::string message = create_count_message(counts);
+    // Set is_lts to true to match existing test expectations
+    // Non security package counts do not have the LTS text
```
orndorffgrant, line 17:
```suggestion
```

orndorffgrant (approved):
Thanks!

@julian-klode Can you also do a quick review of this?

This PR shows reformatted pro messages and APT News to fit with the new APT UX on oracular onward. It determines what style to show by the release in /etc/os-release.

I know that ideally the style would be determined by APT telling the hook what style it is using via jsonrpc params, but we'd like to get something basic done in our v35 release in February.

julian-klode (approved):
This looks good. It might be worth thinking about realigning 

"%d standard security updates"

to 

"Standard security updates: %d"

to match the new summary format ordering ("X upgraded" became "Upgrading: X").

julian-klode:
You can now also query for plucky onward the `binary::apt::APT::Output-Version` option in the `options` array that's part of the hello handshake.

It is an integer (possibly stored in a string), 30 is the new format (it's essentially major*10+minor.

renanrodrigo:
@dheyay Thanks much for your work. As @julian-klode says in plucky we get to read the option, it's better if we fix it for plucky onward. I will take this PR when v35 is released and this will be part of the next release.

renanrodrigo:
@yujinhong-canonical this is a good candidate for a future release.

PR #3364, opened 2024-11-01
wsl: do not enable livepatch on wsl

orndorffgrant:
## Why is this needed?

Fixes: #3156

FYI @CarlosNihelton

## Test Steps

Run the new test.

---

- [x] *(un)check this to re-run the checklist action*

PR #3375, opened 2024-12-18
Bring v36 into v35

orndorffgrant:
## Why is this needed?

Due to delays in v35, we will include the (tiny) things that were going to be in v36 in v35 instead. This brings those couple commits into `main` to be released in v35 when ready. After this PR, we can delete `next-v36` and set `main` as the default branch again.

The only substantive change is from #3364 and is a bugfix, not a new feature. The rest are docs or tooling.

## Test Steps

CI should look okay.

---


renanrodrigo (approved):
mostly docs and the (very) needed adjustment for wsl - lgtm

PR #3376, opened 2024-12-20
Pro lxd dir change

orndorffgrant:
## Why is this needed?

We don't want to introduce the new `interfaces` directory under the old name `/var/lib/ubuntu-advantage` dir. So, instead, it will be the first item in the new `/var/lib/ubuntu-pro`.

## Test Steps

`lxd_guest_attach` configuration tests should still pass

---


orndorffgrant:
@tomponline @markylaing Heads up. This will be the only change to the pro-lxd-auto-attach feature on our side before we release it. This will require a small adjustment on your side. Release date still TBD though.

tomponline:
thanks @orndorffgrant 

@markylaing please can you create a jira task to re-introduce lxd pro client support with the updated location once we have some idea on the release date. @orndorffgrant if you can keep as advised of its progress that would be great. Thanks

lucasmoura (approved):

PR #3378, opened 2025-01-09
Remove features that we don't want in v35

orndorffgrant:
## Why is this needed
Because main should have only what is going to be released.

---


renanrodrigo:
@orndorffgrant after the CVE work lands I'll take this one

orndorffgrant:
@renanrodrigo reminder that this PR still needs to remove the status APIs as well.

renanrodrigo:
I reverted the status API PRs, and will double check if there is something else to remove. If something is at the top of mind, please tell me here @orndorffgrant @lucasmoura @dheyay

orndorffgrant:
This looks like it is removing all the right things. I think we should be able to get this PR to green CI (maybe minus the fix tests? what is the state of those?)

renanrodrigo:
you are right - this should be good minus fix
fix is a nightmare
status is: nada acontece feijoada

lucasmoura (approved):

PR #3386, opened 2025-01-30
Update vulnerabilities CVE endpoint

lucasmoura:
## Why is this needed?
We have redesigned the vulnerabilities CVE endpoint to per-package instead of per-cve. Now, each installed package will list all of the CVEs that it is affected by.

## Test Steps
1. Run the new integration test created for the endpoint
2. Check that the vulnerability caches are still behaving correctly

---


uaclient/messages/__init__.py
```diff
@@ -2868,6 +2868,11 @@ def __repr__(self):
     t.gettext("Got value with incorrect type at index {index}:\n{nested_msg}"),
 )
 
+E_INCORRECT_DICT_ELEMENT_TYPE_ERROR_MESSAGE = FormattedNamedMessage(
+    "incorrect-dict-element-type",
+    t.gettext("Got value with incorrect type for {key}:{value}\n{nested_msg}"),
+)
+
```
orndorffgrant, line 2875: I know we have a specific use case for this that doesn't include anything sensitive. But it is good practice to avoid printing the content of the data being validated. So I suggest changing this to only print the key and the type of the value (not the actual value).

features/api/vulnerabilities_cve.feature
```diff
@@ -52,6 +52,9 @@ Feature: Client behaviour for CVE vulnerabilities API
               "notes": [],
               "priority": "low",
               "published_at": ".*",
+              "related_packages": [
+                "accountsservice"
+              ],
```
orndorffgrant, line 57: This isn't in the spec.

If this data is needed by the CLI, we need to query it specifically for that use case. We shouldn't put it in the API.

uaclient/data_types.py
```diff
@@ -282,6 +353,8 @@ def to_dict(self, keep_none: bool = True) -> dict:
                 new_val = val.to_dict(keep_none)
             elif isinstance(val, list):
                 new_val = data_list_to_list(val, keep_none)
+            elif isinstance(val, dict):
+                new_val = data_dict_to_dict(val, keep_none)
```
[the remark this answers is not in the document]
lucasmoura: Sorry, I think I am missing your point, why `data_list` and  `data_dict` should handle that use case  ? Shouldn't we only treat their respective type on those functions ? Or am I totally missing your point ?

features/api/vulnerabilities_cve.feature
```diff
@@ -0,0 +1,382 @@
...
+      """
+      features:
+        serviceclient_url_responses: "/tmp/response-overlay.json"
+      """
+    And I run `pro api u.pro.security.vulnerabilities.cve.v1` as non-root
+    Then API data field output matches regexp:
+      """
+      {
+        "attributes": {
+          "apt_updated_at": ".*",
```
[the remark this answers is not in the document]
lucasmoura: We just that information on the output of `pro cve`, that's why I have added it. Although I can get that information directly from the system if needed. Do you see any value of keeping this attribute on the API output ? If not, I think we can drop it.

features/api/vulnerabilities_cve.feature
```diff
@@ -0,0 +1,382 @@
...
+              "description": "An issue exists AccountService 0.6.37 in the\nuser_change_password_authorized_cb() function in user.c which could let a\nlocal users obtain encrypted passwords.",
+              "notes": [],
+              "priority": "low",
+              "published_at": ".*",
+              "related_usns": [
+                {
+                  "name": "USN-6687-1",
+                  "title": ""
+                }
+              ]
```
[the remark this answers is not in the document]
lucasmoura: True, but that one we need for `pro cve` and there is no other way to fetch that information :(

features/api/vulnerabilities_cve.feature
```diff
@@ -0,0 +1,382 @@
...
+                {
+                  "fix_origin": "esm-infra",
+                  "fix_status": "fixed",
+                  "fix_version": "0.6.40-2ubuntu11.6\+esm1",
+                  "name": "CVE-2023-3297"
+                }
+              ]
+            }
+          },
+          "vulnerability_data_published_at": ".*"
```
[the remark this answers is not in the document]
lucasmoura: Same `pro cve` need

features/api/vulnerabilities_cve.feature
```diff
@@ -0,0 +1,382 @@
+Feature: Client behaviour for CVE vulnerabilities API
+
+  @uses.config.contract_token
+  Scenario Outline: CVE vulnerabilities for xenial machine
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I attach `contract_token` with sudo
+    # Check we can download and parse the JSON data
+    And I run `pro api u.pro.security.vulnerabilities.cve.v1` as non-root
```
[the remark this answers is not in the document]
lucasmoura: Good catch, let me fix it

features/api/vulnerabilities_cve.feature
lucasmoura: I don't know if querying that information outside of this endpoint is a good approach, but we can better discuss it our next meeting.

renanrodrigo:
rebased to accomodate commits from #3384

renanrodrigo:
Xenial builds now broken on a dict comparison - maybe ordering?

uaclient/api/u/pro/security/cves/cve/v1.py
```diff
@@ -1,13 +1,12 @@
 import datetime
```
orndorffgrant, line 1: During the rename, this file was left behind. I believe all of `uaclient/api/u/pro/security/cves/cve/*` can be deleted

uaclient/api/u/pro/security/cves/_common/v1.py
```diff
@@ -182,9 +182,6 @@ def get_published_date(self):
         return vulnerability_json_data["published_at"]
```
orndorffgrant, line 182: You don't need to make a change, but I suppose the `_common` structure is not needed anymore since there is only one API endpoint.

uaclient/cli/vulnerability/update.py
```diff
@@ -1,7 +1,5 @@
 from uaclient import messages
```
orndorffgrant, line 1: why was `update` left behind if we're deleting the other two vulnerability commands in this PR?

uaclient/api/u/pro/security/cves/_common/v1.py
lucasmoura: True, I will keep as is just as a reminder that this feature could be extended in the future

lucasmoura:
@orndorffgrant fixed

features/api/cves.feature
```diff
@@ -0,0 +1,318 @@
...
+        {
+          "code": "invalid-option-combination",
+          "meta": {
+            "option1": "unfixable",
+            "option2": "fixable"
+          },
+          "title": "Error: Cannot use unfixable together with fixable."
+        }
+      ]
+      """
```
renanrodrigo, line 22: Well I see the point, but would it make sense for a user perspective to just treat it as both and show them all?
I mean, if we are showing all by default and someone asks for fixable and unfixable, that's not conflicting, but rather complementary. I checked the spec, there is no special case defined for this.

lucasmoura: Good point, I am now supporting both flags together

features/cli/help.feature
```diff
@@ -465,91 +465,12 @@ Feature: Pro Client help text
...
-    When I run `pro vulnerability update --help` as non-root
-    Then I will see the following on stdout
-      """
-      usage: pro vulnerability update [-h]
-
-      Updates the vulnerability data stored in the machine.
-      If no vulnerability data exists yet, the command will download it
-
-      <options_string>:
+      optional arguments:
```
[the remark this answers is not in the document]
lucasmoura: This tests will be deleted on the CLI PR, that's why I am not applying this change here.

lucasmoura:
@renanrodrigo updated

features/api/api.feature
```diff
@@ -31,9 +31,7 @@ Feature: Client behaviour for the API endpoints
     When I run `python3 -c "from uaclient.api.u.unattended_upgrades.status.v1 import status"` as non-root
     When I run `python3 -c "from uaclient.api.u.apt_news.current_news.v1 import current_news"` as non-root
     When I run `python3 -c "from uaclient.api.u.pro.detach.v1 import detach"` as non-root
-    When I run `python3 -c "from uaclient.api.u.pro.security.vulnerabilities.cve.v1 import vulnerabilities"` as non-root
-    When I run `python3 -c "from uaclient.api.u.pro.security.vulnerabilities.usn.v1 import vulnerabilities"` as non-root
-    When I run `python3 -c "from uaclient.api.u.pro.packages.updates_with_cves.v1 import updates_with_cves"` as non-root
+    When I run `python3 -c "from uaclient.api.u.pro.security.cves.v1 import cves"` as non-root
```
renanrodrigo, line 34: I think the function name in the file needs to change as well?

renanrodrigo:
@lucasmoura xenial build broke again :/

uaclient/api/u/pro/security/cves/v1.py
```diff
@@ -282,141 +252,119 @@ def _post_process_vulnerability_info(
...
+    elif options.fixable and not is_fixable:
+        return False
+    return True
+
+
 def vulnerabilities(
     options: CVEVulnerabilitiesOptions,
-) -> CVEVulnerabilitiesResult:
+) -> PackageVulnerabilitiesResult:
     return _vulnerabilities(options, UAConfig())
```
orndorffgrant, line 268: I think @renanrodrigo is right that these need renamed to align with our convention of the function name matching the module name.
Also really the options and result objects should be named similarly.
```suggestion
def cves(
    options: CVEsOptions,
) -> CVEsResult:
    return _cves(options, UAConfig())
```

uaclient/api/u/pro/security/cves/v1.py
```diff
@@ -282,141 +252,119 @@ def _post_process_vulnerability_info(
...
-        ),
-        cve_vulnerabilities_result.vulnerabilities_info.get(
-            "applied_fixes_count", {}
-        ),
-    )
-
-
 endpoint = APIEndpoint(
     version="v1",
     name="CVEVulnerabilities",
```
orndorffgrant, line 370: Same thing but down here
```suggestion
    name="CVEs",
```

orndorffgrant (approved):

renanrodrigo (approved):

PR #3387, opened 2025-01-31
Redesign pro vulnerability list command

lucasmoura:
## Why is this needed?
This is PR introduces the redesign of `pro vulnerability list` into `pro cves`. The command still presents a lists of CVEs that
affect the system, but the list is now indexed by package and we there is no longer a summary information at
the end

PS: This PR depends on #3384 and #3386 

## Test Steps
Check that the new integration test is working as expected

---


uaclient/messages/__init__.py
```diff
@@ -1294,66 +1294,27 @@ class TxtColor:
...
-)
-CLI_VULNERABILITY_LIST_FIXES_AVAILABLE_COUNT = t.gettext(
-    "fixable via {pocket}"
+CLI_CVES_UNFIXABLE = t.gettext(
+    "List only vulnerabilities without a fix available"
 )
-CLI_VULNERABILITY_LIST_UNFIXABLE_AVAILABLE_COUNT = t.gettext(
-    "unfixable vulnerabilities found"
+CLI_CVES_FIXABLE = t.gettext("List only vulnerabilities with a fix available")
+CLI_CVES_NOT_AFFECTED = t.gettext("No {issue} found that affects this system")
```
orndorffgrant, line 1255: This is redefined below, so I think we can delete this one.
```suggestion
```

uaclient/cli/cves.py
```diff
@@ -0,0 +1,143 @@
...
+            option1="unfixable", option2="fixable"
+        )
+
+    try:
+        _list_cves(args=args, cfg=cfg)
+    except BrokenPipeError:
+        sys.stderr.close()
+
+
+vulnerability_list_args_group = ProArgumentGroup(
```
orndorffgrant, line 105: nit
```suggestion
cves_args_group = ProArgumentGroup(
```

orndorffgrant: Alternatively, now that this argument group doesn't need to be reused, it can be defined inline below at `argument_groups=...`

uaclient/cli/cves.py
```diff
@@ -0,0 +1,143 @@
...
+                    (
+                        package_name,
+                        cve_info.priority,
+                        cve.fix_origin or "-",
+                        create_link(
+                            text=cve.name,
+                            url="https://ubuntu.com/security/{}".format(
+                                cve.name
+                            ),
+                        ),
```
orndorffgrant, line 41: Nit. `create_link` arguably is formatting for these values similar to the coloring of the priority. Maybe the `create_link` call belongs in `_format_cve_rows` instead of here?

orndorffgrant (commented):
Initial code review

uaclient/cli/cve.py
```diff
@@ -0,0 +1,172 @@
...
+
+                if binary_pkg_cve_info:
+                    rows.append(
+                        AffectedPackage(
+                            name=binary_pkg,
+                            fix_status=binary_pkg_cve_info.fix_status,
+                            fix_origin=binary_pkg_cve_info.fix_origin,
+                            fix_version=binary_pkg_cve_info.fix_version,
+                        )
+                    )
```
orndorffgrant, line 51: minor suggestion
```suggestion
                    if cve.name == cve_name:
                        rows.append(
                            AffectedPackage(
                                name=binary_pkg,
                                fix_status=cve.fix_status,
                                fix_origin=cve.fix_origin,
                                fix_version=cve.fix_version,
                            )
                        )
                        break
```

uaclient/cli/cve.py
```diff
@@ -0,0 +1,172 @@
...
+
+    return formatted_rows
+
+
+def action_cve(args, *, cfg, **kwargs):
+    cve_vulnerabilities = _get_cve_vulnerabilities(args, cfg=cfg)
+
+    if args.cve not in cve_vulnerabilities.cves:
+        print(messages.CLI_CVE_NOT_AFFECTED.format(issue=args.cve))
+        return
```
orndorffgrant, line 87: I think we may still want to show info for CVEs that don't affect the system. But `affected-packages:` will be empty.
This could make things harder :grimacing:

uaclient/cli/cve.py
```diff
@@ -0,0 +1,172 @@
...
+
+    cve_info = cve_vulnerabilities.cves[args.cve]
+    affected_pkgs_rows = _format_affected_pkgs(
+        _get_affected_pkgs(cve_vulnerabilities, cve_info, args.cve)
+    )
+    affected_pkgs_table = Table(rows=affected_pkgs_rows)
+
+    print("name:               {}".format(args.cve))
```
orndorffgrant, line 95: There is too much space between the columns in the first key-value section:
```diff
- name:               CVE-2022-3296
- public-url:         https://ubuntu.com/security/CVE-2022-3296
- published-at:       2022-09-25
- cve-cache-date:     2025-02-06
- apt-cache-date:     2025-02-07
- priority:           low
- cvss-score:         7.8
- cvss-severity:      high
+ name:            CVE-2022-3296
+ public-url:      https://ubuntu.com/security/CVE-2022-3296
+ published-at:    2022-09-25
+ cve-cache-date:  2025-02-06
+ apt-cache-date:  2025-02-07
+ priority:        low
+ cvss-score:      7.8
+ cvss-severity:   high
```

uaclient/cli/cve.py
```diff
@@ -0,0 +1,172 @@
...
+            cli_util.colorize_priority(cve_info.priority)
+        )
+    )
+    print("cvss-score:         {}".format(cve_info.cvss_score))
+    print("cvss-severity:      {}".format(cve_info.cvss_severity))
+    print(
+        "description: |\n  {}".format(
+            cve_info.description.replace("\n", "\n  ")
+        )
+    )
```
orndorffgrant, line 127: Generally, I think this function will be easier to read if each new line got its own `print` call. So for this piece, it might look more like:
```suggestion
    print("description: |")
    for line in cve_info.description.split("\n"):
        print("  {}".format(line))
```
Similar refactors could be applied to `notes`, `affected_packages`, and `related_usns`

Separately, is it safe to rely on the newlines already in the description? Or should we rewrap it ourselves to max 80 characters?

orndorffgrant:
We should also add `cves` and `cve` to the `ua.bash` autocomplete SUBCMDS list

uaclient/cli/cve.py
lucasmoura: But then the API would need to return that information as well and I don't think that's worth it. However, maybe
we display the security page for that CVE in the message. Then the user will at least know where to find more information about it. What do you think ?

uaclient/cli/cve.py
lucasmoura: Yeah, let's be conservative and wrap that to 80 characters

lucasmoura:
@orndorffgrant updated

uaclient/cli/cve.py
```diff
@@ -0,0 +1,184 @@
...
+    if cve_info.notes:
+        print("notes:")
+        print(
+            "{}".format(
+                "\n".join(
+                    "  - {}".format(note.replace("\n", "\n    "))
+                    for note in cve_info.notes
+                )
+            )
+        )
```
orndorffgrant, line 144: My previous comment was going for something more like:
```suggestion
        for note in cve_info.notes:
            print("  - {}".format(note.replace("\n", "\n    "))
```
but this is okay too - not a strong opinion.

If we stick with the `"\n".join` approach, do we need the `"{}".format`?
```suggestion
        print(
            "\n".join(
                "  - {}".format(note.replace("\n", "\n    "))
                for note in cve_info.notes
            )
        )
```

Separately, similar to `description`, should we use `textwrap`? We can set `initial_indent="  - ", subsequent_indent="    "`

uaclient/cli/cve.py
orndorffgrant: I think the CLI for `pro cve` cannot rely on the API, because I think it needs to have consistent output for CVEs that don't affect the machine.

uaclient/cli/cve.py
renanrodrigo: (quick vote for the first suggestion)

lucasmoura:
@orndorffgrant @renanrodrigo fixed

uaclient/cli/cve.py
```diff
@@ -0,0 +1,174 @@
...
+                    subsequent_indent="  ",
+                )
+            )
+        )
+    )
+
+    if cve_info.notes:
+        print("notes:")
+        for note in cve_info.notes:
+            print("  - {}".format(note.replace("\n", "\n    ")))
```
orndorffgrant, line 138: Last nit suggestion on this section I promise :)

```suggestion
    print("description: |")
    print(
        textwrap.fill(
            cve_info.description,
            width=defaults.PRINT_WRAP_WIDTH,
            break_long_words=False,
            break_on_hyphens=False,
            initial_indent="  ",
            subsequent_indent="  ",
        )
    )

    if cve_info.notes:
        print("notes:")
        for note in cve_info.notes:
            print(
                textwrap.fill(
                    note,
                    width=defaults.PRINT_WRAP_WIDTH,
                    break_long_words=False,
                    break_on_hyphens=False,
                    initial_indent="  - ",
                    subsequent_indent="    ",
                )
            )
```

lucasmoura:
@orndorffgrant updated

renanrodrigo:
rebased on main after #3386 landed

tools/ua.bash
```diff
@@ -4,7 +4,7 @@
 
 API_ENDPOINTS=$(/usr/bin/python3 -c 'from uaclient.api.api import VALID_ENDPOINTS; print(" ".join(VALID_ENDPOINTS))')
 SERVICES="anbox-cloud cc-eal cis esm-apps esm-infra fips fips-updates landscape livepatch realtime-kernel ros ros-updates usg"
-SUBCMDS="--debug --help --version api attach auto-attach collect-logs config detach disable enable fix help refresh security-status status system version"
+SUBCMDS="--debug --help --version api attach auto-attach collect-logs config cve cves detach disable enable fix help refresh security-status status system version"
```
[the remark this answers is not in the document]
lucasmoura: Done

renanrodrigo (approved):

uaclient/cli/cve.py
```diff
@@ -0,0 +1,186 @@
...
+            cve_vulnerabilities.apt_updated_at.strftime("%Y-%m-%d")
+        )
+    )
+    print(
+        "priority:        {}".format(
+            cli_util.colorize_priority(cve_info.priority)
+        )
+    )
+    print("cvss-score:      {}".format(cve_info.cvss_score))
+    print("cvss-severity:   {}".format(cve_info.cvss_severity))
```
orndorffgrant, line 121: I found some cases where these say "None". I think these shouldn't be printed when their value is `None`

uaclient/cli/cve.py
```diff
@@ -0,0 +1,186 @@
...
+
+    if args.cve not in cve_vulnerabilities.cves:
+        print("name:            {}".format(args.cve))
+        print(
+            "public-url:      https://ubuntu.com/security/{}".format(args.cve)
+        )
+        print("description: |")
+        print("  " + messages.CLI_CVE_NOT_AFFECTED.format(issue=args.cve))
+        print("affected-packages: null")
+        return
```
orndorffgrant, line 88: In testing this I have a couple comments:

I don't think `description` should ever have anything other than the description of the CVE.

I think I prefer `affected-packages: []` instead of `null` but that's not a strong opinion.

More importantly, there are actually three cases:
1. CVE affects system and in API response
2. CVE does not affect system but all data in the json file we have
3. CVE not even in the JSON file.

Case 1 looks great already.

Case 2 can be something like
```python
    if args.cve not in cve_vulnerabilities.cves:
        from uaclient.api.u.pro.security.cves._common.v1 import VulnerabilityData
        cve_data = VulnerabilityData(cfg).get("security_issues", {}).get("cves", {}).get(args.cve)
```
From there we have most of the data needed to fill out all the fields in this output. There is just the minor issue of turning that dict into the `CVEInfo` object used below.

If after the above snippet `cve_data` is None, then we have case 3. In that case we can do your original plan of saying something like "CVE not present in security data for jammy. You may be able to find more information at https://ubuntu.com/security/CVE-1234-1234".

renanrodrigo:
Merging for the sake of starting the v35 efforts - if there are future tweaks to the vulnerability commands, they can land later

Issue #3388, opened 2025-02-05
docs: Please update FIPS enablement to use fips-updates


henrycoggillcnc:
*Please describe the question or issue you're facing with "How to manage FIPS - Ubuntu Pro Client documentation".*

The steps:
 $ sudo pro enable fips
 $ sudo pro disable fips

should be updated to:
$ sudo pro enable fips-updates
$ sudo pro disable fips-updates

---
*Reported from: https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/howtoguides/enable_fips/*

renanrodrigo:
@henrycoggillcnc thanks for reporting - I will update that.
Quick question - should we add a note in this very page for this change? I mean, mention that `fips` exists but people should be using `fips-updates` instead, for the security patches?

henrycoggillcnc:
That's a reasonable suggestion, yes.

renanrodrigo:
Fixed in #3389

PR #3389, opened 2025-02-10
docs: suggest fips-updates instead of fips

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because we now recommend `fips-updates`, as `fips` is deprecated from Jammy

Fixes: #3388 

---

### DIFF
--- a/docs/howtoguides/enable_fips.rst
+++ b/docs/howtoguides/enable_fips.rst
@@ -5,7 +5,17 @@ How to manage FIPS
 
 .. note::
 
-    FIPS is supported on Ubuntu `16.04, 18.04 and 20.04`_ releases.
+    The latest FedRAMP policy on the use of cryptographic modules relaxes some
+    of the past restrictions that prevented organizations from applying
+    critical security updates. Because of that, we don't recommend enabling
+    ``fips`` anymore - please use ``fips-updates`` instead. Read more in
+    `our blog post`_. Jammy Jellyfish (22.04) and later releases will not
+    have the legacy ``fips`` service available.
+
+.. note::
+
+    Check `the FIPS documentation`_ to verify which Ubuntu releases have FIPS
+    available.
 
 .. caution::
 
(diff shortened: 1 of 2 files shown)


renanrodrigo:
@henrycoggillcnc please take a look and see if the text fits

docs/howtoguides/enable_fips.rst
```diff
@@ -25,24 +26,32 @@ To enable FIPS, run:
...
 a reboot into the FIPS-certified kernel.
 
+.. note::
+
+    ``fips-updates`` is the recommended way of enabling FIPS, as it will bring
+    the FIPS-compliant packages with security updates and coverage. There is
+    also a ``fips`` service, which has only the certified packages, *without
+    further security updates*, but that is deprecated from Ubuntu 22.04 (Jammy
+    Jellyfish) onward.
+
```
s-makin, line 51: The content above this note is really confusing, because there is a difference between fips and fips-updates, and we're installing but referring to the other. 

I think this note needs to be the introduction to the page as body text, and not a note, because that makes it clearer why there is this confusion.

Better would be to move the note (as mentioned) but also change the references to FIPS throughout the document to "FIPS with updates" or however we want to call it.

docs/howtoguides/enable_fips.rst
```diff
@@ -54,7 +63,7 @@ If you wish to disable FIPS, you can use the following command:
 
 .. code-block:: bash
 
-    sudo pro disable fips
+    sudo pro disable fips-updates
```
s-makin, line 66: Does the difficulty in removing fips also apply to fips-updates?

docs/howtoguides/enable_fips.rst
renanrodrigo: so this is a product decision. From Jammy onwards, `fips` does not even exist, and the concept of "enabling FIPS" is achieved using the `fips-updates` service. We don't want people enabling `fips`. @henrycoggillcnc can explain it better if needed

docs/howtoguides/enable_fips.rst
renanrodrigo: yes

renanrodrigo:
@s-makin I have added the link to our blog post about the new standards to the docs, and put the note in the start of the page for visibility. PTAL.

s-makin (approved):
Thanks for the update, this is exactly what I was looking for :) lgtm

renanrodrigo:
The spellcheck errors in this PR are here due to the past, and wlll be fixed in the future. No regressions introduced. Thanks all

PR #3394, opened 2025-02-18
Bump @octokit/request-error and @actions/github

renanrodrigo:
## Why is this needed?
This brings the dependabot changes from #3391 #3392 and #3393

## Test Steps
Check if CI still works, checks should be ok

---


renanrodrigo:
Merging this as a chore

PR #3395, opened 2025-02-19
CVE features improvements

lucasmoura:
## Why is this needed?
We are now improving the error message for the case where a CVE is not in the vulnerability data on `pro cve` and also improving how we fetch the vulnerability data by relying on the etag of the resource.

## Test Steps
Guarantee that all cve related feature tests are still passing


orndorffgrant (approved):
LGTM Tested and it works well. Thanks

renanrodrigo (approved):
Seems spelchecking hates etags, we can fix that
besides that all good

renanrodrigo:
There I fixed it. Merging.

PR #3397, opened 2025-02-20
Bring releases 34.x changes from launchpad

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it syncs the LP changes with what we have here, to facilitate reviews and uploads.

A couple changes appear only in the changelog (python3-pkg-resources dependency and version check) because they have been applied in a better way in the current codebase, the LP versions were workarounds.

I also updated the actual changelog for v35 here.

## Test Steps
Make sure CI works as it was working before

---

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -46,6 +46,41 @@ ubuntu-advantage-tools (35) oracular; urgency=medium
 
  -- Lucas Moura <lucas.moura@canonical.com>  Tue, 08 Oct 2024 16:50:42 -0300
 
+ubuntu-advantage-tools (34.1.3) plucky; urgency=medium
+
+  * apt-hook: set C++ standards version to c++17 for APT 2.9.30 compatibility
+    (LP: #2098862)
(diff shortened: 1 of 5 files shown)


lucasmoura (approved):

debian/changelog
```diff
@@ -29,22 +20,59 @@ ubuntu-advantage-tools (35) oracular; urgency=medium
...
+      + aws: skip operation if no product codes found
+      + gcp: add minimal image license codes
     - cli:
       + add support for vulnerability commands:
-        * pro vulnerability list: List fixable vulnerabilities in the machine
-        * pro vulnerability show: Show information about an affected
-          vulnerability
-        * pro vulnerability update: Update the vulnerability data on the
-          machine
+        * pro cves: List fixable cves in the machine
```
orndorffgrant, line 28: Default includes unfixable now
```suggestion
        * pro cves: List cves in the machine
```

Issue #3390, opened 2025-02-11
Bug: Missing the openssl-fips-module-3 package when enabling FIPS in 22.04 containers


zhijie-yang:
**Description of the bug**

When enabling the FIPS service in the 22.04 containers with the Pro Client, the `openssl-fips-module-3` package is not installed automatically by the enablement of the service. This leads to broken cryptography runtime libraries.

**Expected behavior**

The `openssl-fips-module-3` packages should be installed by the Pro Client automatically as the FIPS cryptography provider when enabling the FIPS services inside containers.

**Current behavior**

The following packages, except `openssl-fips-module-3` are installed when FIPS is enabled in a 22.04 container.
```
libgcrypt20
libgcrypt20-hmac
libgmp10
libgnutls30
libhogweed6
libnettle8
libssl3
openssl
```

**To Reproduce**

1. Launch a normal Ubuntu 22.04 VM in multipass.
2. (In the VM) Install docker with `sudo snap install docker --classic` (and optionally set the user and group for docker)
3. (In the VM) Attach the Pro Services with `sudo pro attach <token> --no-auto-enable`
4. (In the VM) Enable the FIPS service with `sudo pro enable fips-updates --assume-yes`
5. Reboot the VM
6. (In the VM) `docker run --rm -it ubuntu:22.04`
7. (In the docker container) `apt update && apt install ubuntu-advantage-tools ca-certificates` (I would also suggest making `ca-certificates` a recommended dependency of `ubuntu-advantage-tools`, as the Pro Client needs to verify the certificates of the Pro Servers.)
8. (In the docker container) Attach the Pro Services and enable the FIPS service using the same commands in steps 3 and 4 without `sudo`.
9. (In the docker container) Run `dpkg -l | grep "Fips"`, the `openssl-fips-module-3` is missing.
10. (In the docker container) Run `openssl speed` should fail with the following errors:
```
root@8e79bdc84521:/# openssl speed
Doing rand for 3s on 16 size blocks: 302004 rand's in 2.34s
Doing rand for 3s on 64 size blocks: 297073 rand's in 2.40s
Doing rand for 3s on 256 size blocks: 321039 rand's in 2.47s
Doing rand for 3s on 1024 size blocks: 314125 rand's in 2.48s
Doing rand for 3s on 8192 size blocks: 324391 rand's in 2.47s
Doing rand for 3s on 16384 size blocks: 332565 rand's in 2.44s
[...]

PR #3398, opened 2025-02-24
fips: include `openssl-fips-module-3` when enabling FIPS on Jammy

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it

Fixes: #3390 

This is a temporary fix for the bug, as it only includes the missing package when calculating what to install/upgrade on Jammy.
There are ongoing conversations with @Ek0n about creating a metapackage for containers, only with the userspace dependencies, and this will be a definitive solution in the future.

## Test Steps
Enable `fips-updates` or `fips-preview` on Jammy, and verify `openssl-fips-module-3` is present.

---

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -307,7 +307,7 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         install_all_updates_override = util.is_config_value_true(
             config=self.cfg.cfg, path_to_value="features.fips_auto_upgrade_all"
         )
-        # noble onward automatically uses new auto-upgrade logic
+        # jammy onward automatically uses new auto-upgrade logic
         hardcoded_release = system.get_release_info().series in {
             "xenial",
             "bionic",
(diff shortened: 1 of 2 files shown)


orndorffgrant:
Should we instead fix the hack where we ignore `additionalPackages` entirely? We could only ignore a defined list of packages (ubuntu-fips, ubuntu-aws-fips, etc.). Still not great I know, but we'll need something like that (and/or an override based on container-ness) for the future noble fix.

renanrodrigo:
@orndorffgrant not worth fixing the hack now, because the certs team will be working (starting this cycle) on the userspace metapackage - and that *should* be ready before fips gets to noble.
handpicking the ignore list is the same as picking the depends list anyway (:

uaclient/entitlements/fips.py
```diff
@@ -325,6 +325,10 @@ def install_all_available_fips_upgrades(
             )
         ]
 
+        # See GH #3390
+        if system.get_release_info().series == "jammy":
```
lucasmoura, line 329: Any reason why we are not creating a jammy entry under `FIPS_CONTAINER_CONDITIONAL_PACKAGES` ?

renanrodrigo:
#3402 created as a follow-up

PR #3399, opened 2025-02-24
usg: install the `usg` package when enabling

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because when the user runs `pro enable usg`, it is expected by product that `usg` is installed in the system.

In time: I know this could be done via contracts, setting `additionalPackages` there - but we would still need a workaround for `focal`, so change by change I preferred this one.
If in the future having this hardcoded bite us back, we can consider changing the resource in the backend.
Note that the USG resource already has this in place, it's just the CIS resource that will not return the additionalPackages we expect. But that is the resource this client will be using.

## Test Steps
enable usg on a machine
verify the usg package is installed

---

### DIFF
--- a/features/usg.feature
+++ b/features/usg.feature
@@ -10,10 +10,13 @@ Feature: Enable usg on Ubuntu
       One moment, checking your subscription first
       Configuring APT access to Ubuntu Security Guide
       Updating Ubuntu Security Guide package lists
+      Updating standard Ubuntu package lists
+      Installing Ubuntu Security Guide packages
       Ubuntu Security Guide enabled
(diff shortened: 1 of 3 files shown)


dheyay (approved):

orndorffgrant (approved):
Okay because it is really just correcting the already existing hack. And long term we want to get rid of the hack altogether.

renanrodrigo:
@orndorffgrant yep - for the future this hack should be removed, the proper `usg` entitlement should be used and we good

lucasmoura (approved):

PR #3401, opened 2025-02-26
docs: document cve related commands

lucasmoura:
## Why is this needed?

This PR documents the all the CVE related commands that will be released in the next version of the Pro client.


docs/explanations.rst
```diff
@@ -59,6 +59,22 @@ Messaging
    
    Messages... <explanations/index_messages>
 
+CVE Commands
```
s-makin, line 62:
```suggestion
CVE commands
```

docs/explanations.rst
```diff
@@ -59,6 +59,22 @@ Messaging
...
+=============
+
+Here we'll show you the information related to CVE related commands.
```
s-makin, line 65:
```suggestion
Here we'll show information about CVE-related commands.
```

docs/explanations.rst
```diff
@@ -59,6 +59,22 @@ Messaging
...
+On any of the listed commands, we consider a CVE vulnerability as **fixable**,
+if and only if, at least **one** affected installed package associated with it can
+be fixed. That means that a vulnerability is unfixable if no affected installed packages
+related to the vulnerability can be fixed at the moment.
+
+..  toctree::
+    :maxdepth: 1
+
+    explanations/how_to_interpret_output_of_pro_cves.md
+    explanations/how_to_interpret_output_of_pro_cve.md
```
s-makin, line 76: I don't think it makes sense to have two separate explanation pages for this - can we combine these into a single how-to that can disambiguate between the two commands? They're so similarly named that I worry having the two presented side-by-side like this is going to confuse users.

If that's out of scope, I would at least recommend not calling them "how to interpret...", but rather naming the pages according to the intended use of the command.

docs/explanations/how_to_interpret_output_of_pro_cve.rst
```diff
@@ -0,0 +1,124 @@
+How to interpret the output of pro cve
```
s-makin, line 1: Shouldn't all of this content be in the CLI reference page? (same for the second page as well)

s-makin (commented):
Some comments - I think that rather that presenting two explanation guides of content that really belongs in the CLI reference page, I would propose a single how-to guide that discusses using both commands and differentiating between the two. The naming for both is so similar that splitting them apart is going to lead to users not being able to find what they're looking for without a lot of clicking (which increases their annoyance)

lucasmoura:
@s-makin sorry for the delay. I have updated the PR with your suggestions

docs/explanations/how_to_interpret_output_of_pro_cve.rst
lucasmoura: I don't think all of them, due to the size of it. But I have added the missing entries on the CLI reference page

s-makin (approved):
Thanks, lgtm :) there are some unrelated broken links and some spellings triggering the link/spelling checker but they can be dealt with separately

PR #3407, opened 2025-03-11
Add Anbox Cloud to recommended services

keirthana:
## Why is this needed?
Anbox Cloud is a service available via Ubuntu Pro and hence it should be included in the services topic.

### DIFF
--- a/docs/explanations/which_services.rst
+++ b/docs/explanations/which_services.rst
@@ -19,25 +19,6 @@ that stream specifically on those machines.
 
     .. tab-item:: Commercial
 
-       * **Anbox Cloud**
-
-         Canonical Anbox Cloud makes large-scale Android development easier by
-         running Android in secure, scalable containers or virtual machines.
-
-         Compared to other Android emulation solutions, Anbox Cloud can provide
-         at least twice the density and can serve up to 100 Android instances per
-         server.
-
-         Due to its highly scalable nature and performance optimization, Anbox Cloud enables
-         delivering device-agnostic mobile applications. Popular use cases of Anbox Cloud
-         include mobile game streaming services, corporate application streaming,
-         application automation and testing.
-
-         * `Try the appliance`_ variant offered by Anbox Cloud.
-
-         * Find out more about Anbox Cloud in the `official documentation`_.
-
-
        * **Expanded Security Maintenance (ESM)**
 
          If you're using any Ubuntu LTS -- Desktop or Server -- in a commercial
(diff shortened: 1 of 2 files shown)


docs/explanations/which_services.rst
```diff
@@ -19,6 +19,25 @@ that stream specifically on those machines.
 
     .. tab-item:: Commercial
 
+       * **Anbox Cloud**
```
s-makin, line 22: I don't think it should be above ESM and Livepatch. I don't have strong feelings about it being above or below landscape, but ESM should be at the top of the list.

docs/explanations/which_services.rst
```diff
@@ -19,6 +19,25 @@ that stream specifically on those machines.
...
+         Compared to other Android emulation solutions, Anbox Cloud can provide
+         at least twice the density and can serve up to 100 Android instances per
+         server.
+
+         Due to its highly scalable nature and performance optimization, Anbox Cloud enables
+         delivering device-agnostic mobile applications. Popular use cases of Anbox Cloud
+         include mobile game streaming services, corporate application streaming,
+         application automation and testing.
+
+         * `Try the appliance`_ variant offered by Anbox Cloud.
```
s-makin, line 36: What's the appliance variant? (If I don't know, a user won't either)

We already have a "how to enable" guide for Anbox cloud - is that different than what you're linking to here? If so, we should also have a link to the "how to enable" guide

s-makin (commented):
Generally lgtm, although I think the value proposition could be stated a bit more strongly. It's not clear exactly what the general benefit is in a commercial setting (unless a company is developing android apps) - should this more rightly sit under "specialist"?

You will need to move the content below ESM + livepatch as well. These are the most generally applicable products with the highest value proposition for readers, so they take precedence over products that are applicable to a subset of users.

docs/explanations/which_services.rst
orndorffgrant: IMO this should go under "Specialty Services"

morphis: For ordering, it already appears top in `pro status` (I guess because of alphabetical ordering)

What is the definition of "Specialty Services" (Anbox Cloud isn't really a service, it's an entire software stack) and "Commercial"?

orndorffgrant: The "Commercial" heading is for services/software that are generally useful to any company ("Commercial" is used to distinguish from "Personal" use). Any company that uses Ubuntu could benefit from ESM, Livepatch and Landscape.

The "Specialty Services" heading (which maybe we could change to "Specialty Products" if that works better) is for services/software that you only need if you're doing something specific. E.g. you only need ROS if you're doing robotics. You only need realtime-kernel in certain scenarios. You only need anbox-cloud if you need scalable android emulation.

We could put Anbox at the top of "Specialty Products".

Also open to suggestions on the category names, but @s-makin gets the final say

keirthana: That clarification helps and with that in mind, it makes a lot of sense to have Anbox in Specialty Products. I will wait for Sally to voice her opinion on the naming because the services part is the one that trips us for Anbox.

keirthana: Thanks, have moved it to specialty products.

docs/explanations/which_services.rst
keirthana: Oh, I forgot about that. It is pretty much the first part of installation. I have rephrased to point to the how-to enable guide and then to official documentation.

s-makin (approved):
All lgtm :) thanks for your work on this one @keirthana

orndorffgrant (approved):
Looks great!
Sorry @s-makin it looks like you may have lost permissions to merge these :/ I'll try to fix that, but in the mean time, merging here to unblock this.

PR #3409, opened 2025-03-19
postrm: remove /var/lib/ubuntu-pro cache dir on purge

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because there is a new `/var/lib` directory around, and it should go away when the package is purged.

Purging `ubuntu-advantage-tools` is not really supported but we can make sure it is doing the right thing

## Test Steps
- `sudo apt purge ubuntu-advantage-tools`
- make sure `/var/lib/ubuntu-pro` is gone for good

---


panlinux (approved):
+1

orndorffgrant (approved):
:pray:

PR #3411, opened 2025-03-21
docs: standardize the page header to match the Ubuntu Pro Handbook

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it addresses SD-1780 - this is part of the effort to implement a seamless-like experience for users reading the Ubuntu Pro Client docs and the Ubuntu Pro Handbook

- added links to the handbook and to the CLI docs to the header, highlighting the CLI one here
- Removed cis and added anbox to the pro services dropdown
- JS and CSS to make dropdowns work and adjust header size (ugly ugly but that's what it is)

Please review and merge this together with https://github.com/canonical/ubuntu-pro-docs/pull/61

## Test Steps
Build docs and see the magic, ahem

---

### DIFF
--- a/docs/_static/css/header.css
+++ b/docs/_static/css/header.css
@@ -30,7 +30,7 @@ ul.p-navigation__links {
   margin-left: 0;
   margin-top: auto;
   margin-bottom: auto;
-  max-width: 800px;
+  max-width: 1000px;
   width: 100%;
 }
 
@@ -73,26 +73,18 @@ ul.p-navigation__links .p-navigation__link:hover {
 }
 
 ul.p-navigation__links .p-dropdown__link:hover {
-  background-color: var(--color-sidebar-item-background--hover);
+  background-color: #2b2b2b;
 }
 
 ul.p-navigation__links .p-navigation__sub-link {
-  background: var(--color-background-primary);
+  background: #333333;
   padding: .5rem 0 .5rem .5rem;
   font-weight: 300;
 }
 
-ul.p-navigation__links .more-links-dropdown li a {
-  border-left: 1px solid var(--color-sidebar-background-border);
-  border-right: 1px solid var(--color-sidebar-background-border);
-}
-
-ul.p-navigation__links .more-links-dropdown li:first-child a {
-  border-top: 1px solid var(--color-sidebar-background-border);
-}
-
-ul.p-navigation__links .more-links-dropdown li:last-child a {
-  border-bottom: 1px solid var(--color-sidebar-background-border);
+ul.p-navigation__links .pro-services-dropdown li a {
+  color: #ffffff;
+  text-decoration: none;
 }
 
 ul.p-navigation__links .p-navigation__logo {
@@ -103,7 +95,7 @@ ul.p-navigation__links .p-navigation__logo img {
   width: 40px;
 }
 
-ul.more-links-dropdown {
+ul.pro-services-dropdown {
   display: none;
   overflow-x: visible;
   height: 0;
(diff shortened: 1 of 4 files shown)


s-makin (commented):
I'm just going to leave the review on one of these, since it's the same feedback for both.

Overall it looks great! The muted orange underline is understated and sleek, and looks closer to what we see on the product pages.

Looking at the two side-by-side, there are a couple of things that I think we should address to make sure this looks as good as it can before we merge.

1) With all the stuff in the header, the "Pro services" menu gets wider when you hover over it because it matches the width of the ESM line and you get a weird overlap effect on the "More resources" menu element. We *could* fix the widths, but the more I look, the more I think we should just remove the "more resources" element completely. It doesn't really deserve a spot in the header just for the FAQ and Discourse. Let's have just the Handbook, CLI, and Pro services in the header.

* Let's change the names of the projects in the header. Can we make "Handbook" something like "Pro overview" and "CLI" something like "Pro Client" or "Pro command line"? I think those will be more intuitively understandable for new Pro users.

* Can we make the background of the dropdown menu the same dark grey with white text as the header? This will better match the marketing pages and should provide enough visual distinction in both light and dark mode from the text under the menu. The hover-over can stay as it is (light grey with black text).

* I'm not sure why, but in the previews the links to the Pro services from the header menu in the Pro Handbook docs don't work, they just point back to the start page. They work in the Pro Client docs, so it's a bit weird.

* I noticed when I looked at them together that the Pro Handbook has an orange "go" button next to the search bar on the left, but the Pro Client doesn't. I'm not too bothered if we add one to the Pro Client or remove the one from the Pro Handbook - whichever is easier to implement.

renanrodrigo:
@s-makin PRs updated.

> With all the stuff in the header, the "Pro services" menu gets wider when you hover over it because it matches the width of the ESM line and you get a weird overlap effect on the "More resources" menu element. We could fix the widths, but the more I look, the more I think we should just remove the "more resources" element completely. It doesn't really deserve a spot in the header just for the FAQ and Discourse. Let's have just the Handbook, CLI, and Pro services in the header.

We may still have to adjust widths, as there is still some 'weird movement' when you click the menu... Let me know if that is too ugly, or if it is acceptable.

> Can we make the background of the dropdown menu the same dark grey with white text as the header? This will better match the marketing pages and should provide enough visual distinction in both light and dark mode from the text under the menu. The hover-over can stay as it is (light grey with black text).

Did my best, also removed borders which were weird on the handbook side.

> I'm not sure why, but in the previews the links to the Pro services from the header menu in the Pro Handbook docs don't work, they just point back to the start page. They work in the Pro Client docs, so it's a bit weird.

ahem, maybe some developer made some mistake, *cough cough*, I fixed it.

> I noticed when I looked at them together that the Pro Handbook has an orange "go" button next to the search bar on the left, but the Pro Client doesn't. I'm not too bothered if we add one to the Pro Client or remove the one from the Pro Handbook - whichever is easier to implement.

UNFAIR: this is not on the header but on the sidebar. And my card says "header". I updated it anyway.

s-makin:
> We may still have to adjust widths, as there is still some 'weird movement' when you click the menu... Let me know if that is too ugly, or if it is acceptable.

Without the 2nd dropdown menu, you don't really notice the weird movement very much (only if you're looking out for it). At any rate, it looks way better and more deliberate with just the three - having a dropdown just for two links was a bit overkill, thanks for getting rid of it.
 
> Did my best, also removed borders which were weird on the handbook side.

This looks 1000% better, thanks!
 
> ahem, maybe some developer made some mistake, _cough cough_, I fixed it.

It's like it never happened ;)

> UNFAIR: this is not on the header but on the sidebar. And my card says "header". I updated it anyway.

It is unfair I know. Thanks for fixing it anyway :)

PR #3412, opened 2025-03-21
cli: do not overwrite file on collect-logs

lucasmoura:
## Why is this needed?
We no longer override any file when an user runs the collect-logs command

## Test Steps
Check that the modified integration tests are passing

---

### DIFF
--- a/features/cli/help.feature
+++ b/features/cli/help.feature
@@ -92,11 +92,11 @@ Feature: Pro Client help text
       Collect logs and relevant system information into a tarball.
       This information can be later used for triaging/debugging issues.
 
-      <options_string>:
+      optional arguments:
         -h, --help            show this help message and exit
         -o OUTPUT, --output OUTPUT
                               tarball where the logs will be stored. (Defaults to
-                              ./pro_logs.tar.gz)
+                              ./pro_logs.tar.gz).
       """
     When I run `pro api --help` as non-root
     Then stdout matches regexp:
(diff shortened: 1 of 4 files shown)


renanrodrigo (approved):
lgtm, 1 comment

uaclient/messages/__init__.py
```diff
@@ -942,7 +942,10 @@ class TxtColor:
 This information can be later used for triaging/debugging issues."""
 )
 CLI_COLLECT_LOGS_OUTPUT = t.gettext(
-    "tarball where the logs will be stored. (Defaults to " "./pro_logs.tar.gz)"
+    """\
+tarball where the logs will be stored. (Defaults to ./pro_logs.tar.gz).
+Notice that the file must not already exist in the system, otherwise
```
renanrodrigo, line 947: I think this message can stay as-is, we don't need to be explicit about the file existing, because the error message is explicit

lucasmoura:
@renanrodrigo updated

features/cli/help.feature
```diff
@@ -92,11 +92,11 @@ Feature: Pro Client help text
       Collect logs and relevant system information into a tarball.
       This information can be later used for triaging/debugging issues.
 
-      <options_string>:
+      optional arguments:
```
renanrodrigo, line 95: This change should not be part of this PR - I will fix it for the merge.

renanrodrigo (approved):

PR #3413, opened 2025-03-25
cli: update Anbox installation output message

keirthana:
## Why is this needed?
The Anbox Cloud appliance installation message is missing a step before initialising it. Adding this to the CLI output so that users can follow the correct sequence of commands.

## Test Steps
1. Enable `anbox-cloud` on Ubuntu Pro
2. Verify if the output contains the updated text.

---

### DIFF
--- a/features/anbox.feature
+++ b/features/anbox.feature
@@ -96,11 +96,11 @@ Feature: Enable anbox on Ubuntu
       The `prepare-node-script` command lets you preview a script that installs some additional packages,
       kernel modules and GPU driver packages, if a GPU is available:
 
-      $ anbox-cloud-appliance prepare-node-script
+      $ anbox-cloud-appliance prepare-node-script > prepare.sh
 
-      Once you have previewed the script, apply it to complete the installation:
+      Preview the script and when ready, apply it to complete the installation:
 
-      $ anbox-cloud-appliance prepare-node-script | sudo bash -ex
+      $ sudo bash -ex prepare.sh
 
       Once installed, to initialise Anbox Cloud, run:
 
--- a/features/anbox.feature
+++ b/features/anbox.feature
@@ -90,7 +90,19 @@ Feature: Enable anbox on Ubuntu
       Configuring APT access to Anbox Cloud
       Updating Anbox Cloud package lists
       Anbox Cloud enabled
-      To finish setting up the Anbox Cloud Appliance, run:
+      To finish setting up the Anbox Cloud Appliance, run the following commands
+      sequentially:
+
+      The `prepare-node-script` command lets you preview a script that installs some additional packages,
+      kernel modules and GPU driver packages, if a GPU is available:
+
+      $ anbox-cloud-appliance prepare-node-script
+
+      Once you have previewed the script, apply it to complete the installation:
+
+      $ anbox-cloud-appliance prepare-node-script | sudo bash -ex
+
+      Once installed, to initialise Anbox Cloud, run:
 
       $ sudo anbox-cloud-appliance init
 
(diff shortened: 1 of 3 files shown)


uaclient/messages/__init__.py
```diff
@@ -1294,7 +1294,19 @@ class TxtColor:
 ).format(url=urls.ANBOX_HOME_PAGE)
 ANBOX_RUN_INIT_CMD = t.gettext(
     """\
-To finish setting up the Anbox Cloud Appliance, run:
+To finish setting up the Anbox Cloud Appliance, run the following commands
+sequentially:
+
+The following command lets you preview a script that installs some additional packages,
```
renanrodrigo, line 1300: Not sure about the best way to reword it, but maybe avoid repeating "following command"?

renanrodrigo (commented):
Hello, @keirthana , thanks for fixing this message.

The `features/anbox.feature` file contains a test which checks this message, and that needs to be updated too.

uaclient/messages/__init__.py
keirthana: @renanrodrigo Thank you, I have reworded it and also updated the `anbox.feature` file. Please have a look.

features/anbox.feature
```diff
@@ -90,7 +90,19 @@ Feature: Enable anbox on Ubuntu
...
+      sequentially:
+
+      The `prepare-node-script` command lets you preview a script that installs some additional packages,
+      kernel modules and GPU driver packages, if a GPU is available:
+
+      $ anbox-cloud-appliance prepare-node-script
+
+      Once you have previewed the script, apply it to complete the installation:
+
+      $ anbox-cloud-appliance prepare-node-script | sudo bash -ex
```
morphis, line 103: Maybe we should do the following for a more secure way to not have the app changing what has been reviewed, e.g.

```
anbox-cloud-appliance prepare-node-script > prepare.sh
# Now review the script and once done, apply via
sudo bash -ex prepare.sh
```
@keirthana wdyt?

morphis (changes requested):

features/anbox.feature
keirthana: Thanks @morphis, that is indeed better. I have applied the feedback.

renanrodrigo (approved):
lgtm!

To be merged when we sort out the last release bits

keirthana:
> @keirthana hunspell is unhappy about `initialise` I remember seeing something about switching to en_US - and tbh the client messages are en_US anyway
> 
> I would say this needs to read `initialize` instead?

Thanks @renanrodrigo I thought the Ubuntu Pro client docs were still on en_GB. I have fixed it now. 

I also fixed the URL for the initialize section in the `urls.py` which makes the formatting check complain. I don't know if it complains because there is a limit on the number of characters or something else is amiss. Could you help?

keirthana:
> > @keirthana hunspell is unhappy about `initialise` I remember seeing something about switching to en_US - and tbh the client messages are en_US anyway
> > I would say this needs to read `initialize` instead?
> 
> Thanks @renanrodrigo I thought the Ubuntu Pro client docs were still on en_GB. I have fixed it now.
> 
> I also fixed the URL for the initialize section in the `urls.py` which makes the formatting check complain. I don't know if it complains because there is a limit on the number of characters or something else is amiss. Could you help?

I tried reformatting the file but now style check complains that the line is too long. :woman_shrugging:

renanrodrigo:
@keirthana yep it happens - that's a big URL heh

please keep whatever black defined, and add `  # noqa: E501` (mind the couple spaces) after the end of the string, so it's ignored by the style check

keirthana:
> @keirthana yep it happens - that's a big URL heh
> 
> please keep whatever `black` defined, and add ` # noqa: E501` (mind the couple spaces) after the end of the string, so it's ignored by the style check

Thanks, that helped!

renanrodrigo:
Thanks!

PR #3414, opened 2025-03-26
wip: minimal patch for generic-hwe variant for realtime-kernel

orndorffgrant:
## Why is this needed?

## Test Steps

---


orndorffgrant:
This was useful as a demonstration of what this could look like but will not be happening any time soon. Instead `pro enable realtime-kernel --access-only` will be used for extra alternative kernels like this.

PR #3421, opened 2025-04-08
lxd: create config on /var/lib/ubuntu-advantage

lucasmoura:
## Why is this needed?
We are creating the lxd config files under /var/lib/ubuntu-advantage to avoid pontential race conditions with the lxd file watcher

## Test Steps
Guarantee that the `pro config set lxd_guest_attach=<VALUE>` command creates the interface folder on `/var/lib/ubuntu-advantage`

---


renanrodrigo (approved):
lgtm
thanks!

PR #3422, opened 2025-04-09
changelog: release 35.1ubuntu0

renanrodrigo:
This is just the changelog entry for v35.1


lucasmoura (approved):

renanrodrigo:
`version.py says "35.1ubuntu0" but changelog says "35.1ubuntu0"`
Sigh... seems we need a change to the version script too. on that.

PR #3426, opened 2025-04-11
relax onlySeries directive check

lucasmoura:
## Why is this needed?
If a token is associated with a `onlySeries` directive, we will allow the `onlySeries` release and all previous releases to attach using the token, instead of only the release specified by `onlySeries`

## Test Steps

---


dheyay (approved):
Looks good!

renanrodrigo (approved):

PR #3428, opened 2025-04-20
use openssl abstraction in apparmor profile

h0tw1r3:
## Why is this needed?

ubuntu_pro_esm_cache apparmor profile is denying access to custom ssl engine configurations.

## Test Steps

1. Modify `/etc/ssl/openssl.cnf` to include another file supported by the apparmor openssl abstraction, example `.include /etc/ssl/engines.d/custom.cnf`.

2. Run the cache service: `systemctl start esm-cache.service`

3. Review the kernel log (dmesg), there will be a apparmor deny message like:
   > [1831.626508] audit: type=1400 audit(1745170012.847:58): apparmor="DENIED" operation="open" profile="ubuntu_pro_esm_cache//cloud_id" name="/etc/ssl/engines.d/custom.cnf" pid=19229 comm="cloud-id" requested_mask="r" denied_mask="r" fsuid=0 ouid=0

---

### DIFF
--- a/debian/apparmor/ubuntu_pro_esm_cache.jinja2
+++ b/debian/apparmor/ubuntu_pro_esm_cache.jinja2
@@ -118,6 +118,7 @@ profile ubuntu_pro_esm_cache flags=(attach_disconnected) {
   profile cloud_id flags=(attach_disconnected) {
     include <abstractions/base>
     include <abstractions/nameservice>
+    include <abstractions/openssl>
     include <abstractions/python>
 
     ptrace read peer=unconfined,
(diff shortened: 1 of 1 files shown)


panlinux (approved):
+1, this makes sense

renanrodrigo:
Thanks for the review @panlinux !

SO, @h0tw1r3 , could you please sign the Canonical CLA so we can accept your contribution?

h0tw1r3:
I did, but it looks like the cla check doesn't support GitHub no reply addresses :(

renanrodrigo:
Yeah it does not...
However, it accepts manual checks
cla
So we good - approving it.

renanrodrigo (approved):

PR #3434, opened 2025-04-30
Adjus docs branch for v35.1

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because it shapes our docs so we can release them together with the v35.1 client.

- I have corrected the spellcheck errors.
- I have changed the spelling from en_GB to en_US, as this is the new standard for the company - we should be following it everywhere
- I polished the custom wordlist a bit

## Test Steps
Check CI below, see the rendered results

---


docs/howtoguides/fix_index.rst
```diff
@@ -29,5 +29,5 @@ You can also resolve multiple vulnerabilities at once using the API.
    Preview pro fix results <fix_how_to_know_what_the_fix_command_would_change>
    Skip fixing related USNs <fix_how_to_not_fix_related_usns>
    Check if your system is affected by multiple CVEs <how_to_know_if_system_affected_by_a_list_of_cves>
-   Preview result of fixing multiple CVEs <fix_how_to_better_visualise_fixing_multiple_cves>
+   Preview result of fixing multiple CVEs <fix_how_to_better_visualize_fixing_multiple_cves>
```
s-makin, line 32: If you're changing the file name you will also want to set up a redirect

s-makin (commented):
Generally lgtm, there's just one small issue but then it should be good to go

docs/howtoguides/fix_index.rst
renanrodrigo: Added, please check if that's the proper way of doing it

lucasmoura (approved):

docs/howtoguides/fix_index.rst
s-makin: Seems to be working fine in the preview :+1:

PR #3457, opened 2025-06-09
tests: fix collect-logs test

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because we like to see tests passing.

This test was overwriting the target file for collect-logs, which is not supported anymore.

Fixes: #3451

## Test Steps
Run the modified test.

---


dheyay (approved):
:1st_place_medal:

PR #3461, opened 2025-06-11
ci: remove actions that are no longer helpful

orndorffgrant:
## Why is this needed?

This PR solves all of our problems because these actions haven't been helpful in recent times. They are mostly creating noise. I think we'll be okay without them.


.github/actions/pr-checklist/index.js
```diff
@@ -1,338 +0,0 @@
...
-            owner: context.issue.owner,
-            repo: context.issue.repo,
-            issue_number: context.issue.number,
-            body: newBody,
-        });
-        client.rest.reactions.createForIssue({
-            owner: context.issue.owner,
-            repo: context.issue.repo,
-            issue_number: context.issue.number,
-            content: "eyes"
```
renanrodrigo, line 325: We'll have to put this code in a separate action as it's actually very important for the repo
But that can be done later

renanrodrigo (approved):
LGTM
@lucasmoura never used these anyway HAHAHA
JK it's ok to remove it

PR #3469, opened 2025-06-26
Tests on plucky

renanrodrigo:
## Why is this needed?
This PR solves all of our problems because if we want to SRU to plucky we need to test on plucky.

## Test Steps
Run the integration tests.

There are a couple things failing in this new set:
- landscape tests, because landscape client has different output now, the test needs adjustment
- apt news tests, because it checks for the whole apt output (as it should), but on plucky some phrases are duplicated... (because of the new apt experience).

I am sending this now, and will fix those details when running the tests for the validation.


dheyay (approved):
:+1: :+1: :100:  Looks good, thanks!

PR #3482, opened 2025-08-11
Update wsl tests to only cover ESM service

lucasmoura:
## Why is this needed?
Only the ESM services are available on wsl. The tests are being updated to reflect that as well

## Test Steps


renanrodrigo (approved):
@lucasmoura Thank you very much for adjusting those!

PR #3484, opened 2025-08-21
Add support for esm infra legacy

lucasmoura:
## Why is this needed?
This PR adds support for both `esm-infra-legacy` and `esm-apps-legacy` on the Pro client.
Additionally, it also updated the `upgrade_lts_contract` script to just remove services that are no longer supported on a given release, as the old logic was not working as expected and refactoring would still not allow services to be disabled during a `do-release-upgrade`

## Test Steps
Run the new/modified integration tests


renanrodrigo (commented):
a few comments skimming through

lucasmoura:
@renanrodrigo good catch on the `refresh-keyrings` script, as it should not be modified here. I have updated the PR to address that while also fixing the behave version change issue and adding tests for autocomplete as well

PR #3490, opened 2025-09-09
tests: turn future onlyseries unit test into a behave test

renanrodrigo:
This unit test is breaking builds in environmentes without distro-info-data. Bringing the package as a build dependency for a single unit test is overkill.

TBH this only shows how our unit tests cry for refactoring :/


alnvdl-work (approved):
LGTM, thanks!

Issue #3488, opened 2025-09-05
Bug: downgrade warning shows incorrect kernel version on cloud images


magalilemes:
**Description of the bug**

When enabling fips-updates on AWS, Azure or Google cloud, the message warning about the kernel downgrade shows the wrong version for the kernel that is going to be installed. The correct cloud-specific kernel is installed, but the message always displays the kernel version of the generic-based kernel, which is not the one being installed.
I observed this behavior on both Jammy and Noble.

**Expected behavior**

For instance, on a Noble image for AWS, when running `pro enable fips-updates`, the message should state
`This will downgrade the kernel from 6.14.0 to 6.8.0-1036.38+fips1.`, since `6.8.0-1036.38+fips1` is the current AWS-specific kernel version that is going to be installed.

**Current behavior**

The current message, regardless of the cloud environment, is `This will downgrade the kernel from 6.14.0 to 6.8.0-79.79+fips1.`, where `6.8.0-79.79+fips1` is actually the generic-based FIPS kernel version.

**To Reproduce**

- Launch a Noble/Jammy instance on one of AWS, Azure or Google cloud.
- Having the instance attached to a Pro subscription, run `pro enable fips-updates`.
- Observe the message displays the wrong kernel version for downgrade, yet the correct kernel, with a different version, is still installed.

**System information:**

- Ubuntu release: 24.04.3 LTS (Noble Numbat); 22.04.5 LTS (Jammy Jellyfish)
- Pro Client version: 36ubuntu0\~24.04; 36ubuntu0\~22.04

renanrodrigo:
Thanks for the report, @magalilemes !

I could verify this is a valid bug. Will bring it to the next cycle's planning.

PR #3492, opened 2025-09-10
fips: show correct kernel version in the downgrade message on clouds

renanrodrigo:
## Why is this needed?
It 
Fixes: #3488

'linux-fips' was hardcoded as the package to look at. Instead, we look into whatever the backend tells us is the correct package which will be installed.

This relies on three facts that must remain true:
- There is only one metapackage for each fips flavor on VMs
- This metapackage is called ubuntu-fips or ubuntu-<something>
- The linux kernel to be installed is called linux-<something>

If something derails, the worse that can happen is, again, a wrong version in the CLI. So not much to worry about.

## Test Steps
Behave tests for this would be hell, because kernel versions tend to change and keeping track of it without mocking would be horrible.

Unit tests would be possible but unfortunately too convoluted with the current codebase. Lots of refactors before we could make a unit test really useful here.

So the best way is manually testing with the steps outlined in the linked bug, with and without this change, and confirm the fix.

### DIFF
--- a/uaclient/entitlements/fips.py
+++ b/uaclient/entitlements/fips.py
@@ -248,7 +248,17 @@ class FIPSCommonEntitlement(repo.RepoEntitlement):
         our_m = re.search(
             r"(?P<kernel_version>\d+\.\d+\.\d+)", our_full_kernel_str
         )
-        fips_kernel_version_str = apt.get_pkg_candidate_version("linux-fips")
+        # This relies on the fact that fips has a single metapackage in
+        # additionalPackages. If more packages are added to additionalPackages
+        # for some reason, we fall back to "ubuntu-fips".
+        fips_package = "ubuntu-fips"
+        if len(self.packages) == 1:
+            fips_package = self.packages[0]
+
+        # We are interested in the actual kernel version
+        fips_kernel_version_str = apt.get_pkg_candidate_version(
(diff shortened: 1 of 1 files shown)


uaclient/entitlements/fips.py
```diff
@@ -248,7 +248,17 @@ def prompt_if_kernel_downgrade(self, *, assume_yes: bool) -> bool:
         our_m = re.search(
             r"(?P<kernel_version>\d+\.\d+\.\d+)", our_full_kernel_str
         )
-        fips_kernel_version_str = apt.get_pkg_candidate_version("linux-fips")
+        # This relies on the fact that fips has a single metapackage in
+        # additionalPackages. I more packages are added to additionalPackages
```
alnvdl-work, line 252: I**f** more packages...

alnvdl-work:
BTW, this also hints that maybe the client is doing things it shouldn't be doing. While it's nice to display what is going to happen, that's just a side-effect of the changes of the steps performed by the client.

If the client has to capture and communicate all these side-effects, its complexity is bound to grow with the complexity of the systems it is integrating with...

For a future redesign, we might consider just communicating the steps that will be performed (e.g., install meta-package `x`), and let the downstream consequences of that be represented/explained by other systems.

PR #3497, opened 2025-09-19
apparmor: adjust apparmor profiles to cope with coreutils paths

gegarcia:
Due to Rust coreutils in questing, some AppArmor policy is affected due to the change in path and symlinks.
Instead of using the @{coreutil_dirs} variable, which is available only on apparmor 5.0.0~alpha1-0ubuntu7, we are using the hardcoded path. This is because this project is usually backported to older releases and that will maintain compatibility.

Fixes: https://bugs.launchpad.net/bugs/2123870

Tested with version in PPA 
https://launchpad.net/~georgiag/+archive/ubuntu/apparmor-coreutils


renanrodrigo:
Hello, @gegarcia , thanks for the change!

I am having a hard time trying to get some apparmor denials so I can test this. How did you perform the testing? Is it possible to generate some invalid call and then see it is not DENIED or something after the change?

PR #3511, opened 2025-10-23
Release37 test fixes

renanrodrigo:
## Why is this needed?
This is a set of changes done to tests to make sure they pass as part of the release validation.
All of the changes here should have been applied before, either when features changed or when side effects changed the expected outputs.

There is also a commit adding some tests for `questing`.

## Test Steps
Run the CI.


renanrodrigo:
Please don't merge this before the SRU for version 37.1 is completed.

renanrodrigo:
Updated the PR opening the next dev version - merging

PR #3512, opened 2025-10-23
Fix logic to check the APT lock

renanrodrigo:
## Why is this needed?
We need to wait for the lock to be released, but can't hold it while
performing the operations in the hook, because we subprocess to APT,
which needs the lock to run.

Releasing the lock as soon as we get it is good enough, and then we let
APT itself deal with it.

LP: #2129712

## Test Steps
Make sure all upgrade acceptance tests are passing.


renanrodrigo:
Review for the point release should be done in this PR. This should only get merged to `main` once uploaded to Ubuntu.

ThiagoSDQ (approved):

PR #3526, opened 2026-03-22
fix: suppress upgrade hint for phased packages

### LINKED ISSUE #3192: Bug: New version hints ignore phasing
**Description of the bug**

I love that we suggest to upgrade if there is a new version, it is helpful to avoid failing due to known issues.
But I realized we might be slightly too aggressive.

In case the update is still held back by phasing we are already warning.
And while that warning is nice, it is also always a bit of critique like "you haven't updated your things"

**Expected behavior**

While still held back by phasing of the package do not yet hint at please updating it.

Since unattended upgrades is - by default - only installing security fixes users will still see it.
But it avoids the following flow
1. run pro, see the hint
2. run apt upgrade see it is held back
3. uncertainty if they should really install something that is held back

As a bonus, someone updating non -security regularly by unattended-upgrades or other means will more or less never see the suggestion cluttering the output.

**Current behavior**

Held back on upgrades

```
paelzer@Keschdeichel 07:49:30   ~  sudo apt upgrade
Reading package lists... Done
Building dependency tree... Done
Calculating upgrade... Done
#
# OpenSSH CVE-2024-6387 has been fixed for 22.04 LTS, 23.10 and 24.04 LTS.
# RegreSSHion: Possible RCE Due To A Race Condition In Signal Handling.
# For more details see: https://ubuntu.com/security/notices/USN-6859-1.
#
The following upgrades have been deferred due to phasing:
  file-roller libclang-cpp18 libclang1-18 libllvm18 ubuntu-pro-client ubuntu-pro-client-l10n
0 upgraded, 0 newly installed, 0 to remove and 6 not upgraded.
```

Seen as in phasing
```
paelzer@Keschdeichel 07:49:47   ~  apt-cache policy ubuntu-pro-client
ubuntu-pro-client:
  Installed: 32.3~24.04
  Candidate: 32.3.1~24.04
  Version table:
     32.3.1~24.04 500 (phased 60%)
        500 http://de.archive.ubuntu.com/ubuntu noble-updates/main amd64 Packages
 *** 32.3~24.04 100
        100 /var/lib/dpkg/status
     31.2.3 500
        500 http://de.archive.ubuntu.com/ubuntu noble/main amd64 Packages
```

We hint at "you should upgrade"

```
paelzer@Keschdeichel 07:49:58   ~  pro status
SERVICE          ENTITLED  STATUS       DESCRIPTION
...

[info] A new version is available: 32.3.1~24.04
Please run:
    sudo apt install ubuntu-pro-client
to get the latest bug fixes and new features.
```

**To Reproduce**

1. Get the situation that the pro client is in -updates but phasing and not yet to be installed on this system
2. run any pro command

**System information:**

- Ubuntu release: 24.04
- Pro Client version: 32.3~24.04

JFantasy:
## Problem

When a newer `ubuntu-pro-client` version exists but is held back by Debian's phased rollout, `pro status` still displays:

```
[info] A new version is available: 32.3.1~24.04
Please run: sudo apt install ubuntu-pro-client
```

This is misleading — the user cannot actually install the update yet, as `apt-cache policy` shows it as `(phased 60%)`.

Fixes #3192

## Root Cause

`get_pkg_candidate_version()` in `uaclient/apt.py` calls `dep_cache.get_candidate_ver(package)` which returns the candidate version regardless of phasing status.

## Fix

After retrieving the candidate, check `dep_cache.phasing_applied(package)` — the official `apt_pkg` API that returns `False` when the phased update is not yet applicable to this machine. If not applicable, return `None` to suppress the hint.

## Tests

Added two new test cases to `TestGetPkgCandidateversion`:
- `test_get_pkg_candidate_version_when_phased` — verifies `None` is returned when phasing is not applied
- `test_get_pkg_candidate_version_when_phasing_complete` — verifies the version is returned normally when phasing is complete

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -342,6 +342,9 @@ def get_pkg_candidate_version(
         if not candidate:
             return None
 
+        if not dep_cache.phasing_applied(package):
+            return None
+
         candidate_version = candidate.ver_str
 
     if not check_esm_cache:
--- a/uaclient/tests/test_apt.py
+++ b/uaclient/tests/test_apt.py
@@ -1346,6 +1346,42 @@ class TestGetPkgCandidateversion:
         actual_value = get_pkg_candidate_version("pkg1", check_esm_cache=True)
         assert "1.2" == actual_value
 
+    @mock.patch("uaclient.apt.apt_pkg.DepCache")
+    @mock.patch("uaclient.apt.get_apt_pkg_cache")
+    def test_get_pkg_candidate_version_when_phased(
+        self,
+        m_apt_cache,
+        m_dep_cache,
+    ):
+        """Return None when candidate exists but phasing is not yet applied."""
+        m_pkg = mock.MagicMock()
+        m_apt_cache.return_value = {"pkg1": m_pkg}
+        m_dep_cache.return_value.get_candidate_ver.return_value = (
+            mock.MagicMock(ver_str="1.2")
+        )
+        m_dep_cache.return_value.phasing_applied.return_value = False
+
+        actual_value = get_pkg_candidate_version("pkg1")
+        assert actual_value is None
+
+    @mock.patch("uaclient.apt.apt_pkg.DepCache")
+    @mock.patch("uaclient.apt.get_apt_pkg_cache")
+    def test_get_pkg_candidate_version_when_phasing_complete(
+        self,
+        m_apt_cache,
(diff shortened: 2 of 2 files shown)


renanrodrigo:
@yujinhong-canonical I would like to close this PR (and the issue) but you have the final word. Please take a look and let me know what you think.

PR #3530, opened 2026-04-10
test: add check for DMI path

yujinhong-canonical:
## Why is this needed?
This PR updates the AppArmor firmware verification script to include  `/sys/firmware/dmi/entries/0-0/raw` verfication.

## Test Steps
Build debian files with:
- post-fix: main
- pre-fix device-tree: comment out line 80 (`/sys/firmware/devicetree/base/model r,`) from [debian/apparmor/ubuntu_pro_esm_cache.jinja2](https://github.com/canonical/ubuntu-pro-client/blob/main/debian/apparmor/ubuntu_pro_esm_cache.jinja2).
- pre-fix dmi: comment out line 331 (`/sys/firmware/dmi/entries/0-0/raw r,`) from [debian/apparmor/ubuntu_pro_esm_cache.jinja2](https://github.com/canonical/ubuntu-pro-client/blob/main/debian/apparmor/ubuntu_pro_esm_cache.jinja2).

How to build deb:
```
./tools/build.sh noble
```

Verify device-tree check failure
```
./sru/release-37/test-apparmor-firmware-access.sh noble /path/to/built/pre-fix/device-tree/ubuntu-pro-client.deb
```
Verify dmi check failure
```
./sru/release-37/test-apparmor-firmware-access.sh noble /path/to/built/pre-fix/dmi/ubuntu-pro-client.deb
```
Verify success detection
```
./sru/release-37/test-apparmor-firmware-access.sh noble /path/to/built/post-fix/ubuntu-pro-client.deb
```

### DIFF
--- a/sru/release-37/test-apparmor-firmware-access.sh
+++ b/sru/release-37/test-apparmor-firmware-access.sh
@@ -17,7 +17,6 @@ function on_err {
 }
 trap on_err ERR
 
-
 lxc launch ubuntu-daily:$series $name -c security.nesting=true -c security.privileged=true
 sleep 5
 
@@ -27,7 +26,7 @@ lxc exec $name -- apt-get install -y ubuntu-advantage-tools > /dev/null
 echo -e "\n* Current ubuntu-advantage-tools installed"
 echo "###########################################"
 lxc exec $name -- apt-cache policy ubuntu-advantage-tools
-echo -e "###########################################\n"
+echo "###########################################\n"
 
 APPARMOR_PROFILE=/etc/apparmor.d/ubuntu_pro_esm_cache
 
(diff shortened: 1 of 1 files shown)


sru/release-37/test-apparmor-firmware-access.sh
```diff
@@ -81,21 +81,30 @@ sleep 3
...
-echo "Attempting to read /sys/firmware/devicetree/base/model under profile: $APPARMOR_PROFILE"
-
-# Capture the attempt. We use 'aa-exec' to force the security profile.
-if lxc exec $name -- aa-exec -p ubuntu_pro_esm_cache python3 -c "open('/sys/firmware/devicetree/base/model').read()" > /dev/null 2>&1; then
+echo "Attempting to read firmware paths under profile: $APPARMOR_PROFILE"
+
+# Check both the Device Tree and DMI paths mocked earlier
+PATHS_TO_TEST="/sys/firmware/devicetree/base/model /sys/firmware/dmi/entries/0-0/raw"
+
+if lxc exec $name -- aa-exec -p ubuntu_pro_esm_cache python3 -c "
```
srunde3: We don't want to execute this under the same profile; it was added to a different profile. That's why this test fails, and that's probably what prompted you to make the change in https://github.com/canonical/ubuntu-pro-client/pull/3529.

Notice this line:

```
aa-exec -p ubuntu_pro_esm_cache
```

That means each test uses the `ubuntu_pro_esm_cache` profile. We only added the `dmi/...` to the `ubuntu_pro_esm_cache_systemd_detect_virt` profile, because it's only during `systemd-detect-virt` that we need to access those DMI files (at least according to the original bug report).

We need to run the check using the `ubuntu_pro_esm_cache_systemd_detect_virt` profile.

srunde3 (changes requested):

srunde3 (commented):
I'm not understanding the intent behind these changes given the context; are you changing the entire test approach? I see we're symlinking `cat` into `systemd-detect-virt`, for example.

I had understood that we would just be adding an additional `aa-exec` call; could you elaborate on this new approach?

yujinhong-canonical:
> I'm not understanding the intent behind these changes given the context; are you changing the entire test approach? I see we're symlinking `cat` into `systemd-detect-virt`, for example.
> 
> I had understood that we would just be adding an additional `aa-exec` call; could you elaborate on this new approach?

As mentioned in our DMs, this is a workaround because the `ubuntu_pro_esm_cache_systemd_detect_virt systemd-detect-virt` profile is a lot more restrictive than the `ubuntu_pro_esm_cache` profile. Unlike the `ubuntu_pro_esm_cache` profile which allows the execution of python, `ubuntu_pro_esm_cache_systemd_detect_virt systemd-detect-virt` only allows the execution of one file, which is `systemd-detect-virt`. So, I could not use the same approach for the dmi path read access verification.

I attempted adding an additional `aa-exec` call under the `ubuntu_pro_esm_cache_systemd_detect_virt` profile in [this commit](https://github.com/canonical/ubuntu-pro-client/pull/3530/changes/635ed7709e4ff3e056c512b5608ee89a1c07cfa7). This test fails before it even gets to opening the dmi file.

Here is how I verify read access of the device tree: `aa-exec -p ubuntu_pro_esm_cache python3 -c "open('/sys/firmware/devicetree/base/model').read()"`. I cannot do the same for the dmi path, because python is not allowed in `systemd-detect-virt`. Neither is cat, which could have been an alternative. This is why I temporarily subbed in the cat binary as `systemd-detect-virt`, the only executable file under this profile.

srunde3 (approved):
Talked with Yujin in MM about the difficulties of emulating this behavior with LXC; I think this "hacky" test is OK. We're basically narrowly testing the declarations in the AppArmor config. We don't have the right test bed to be able to properly test how `systemd-detect-virt` behaves on various architectures and virtualizations.

If that changes in the future, we can revisit.

renanrodrigo (approved):

srunde3:
bypassing merge rules

PR #3533, opened 2026-04-17
fix: map aws-gov to aws for correct FIPS flavor

yujinhong-canonical:
## Why is this needed?
Fixes [LP: 2144693](https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/2144693) by aliasing `aws-gov` to `aws` within the pro client.

## Reproduce bug:
From host machine:
```
multipass launch --name fips-repro noble
multipass shell fips-repro
```
From fips-repo:
Install AWS kernel
```
sudo apt update && sudo apt install -y linux-aws
sudo reboot
```
Mock aws-gov
```
# 1. Overwrite the result files
echo "aws-gov" | sudo tee /run/cloud-init/cloud-id-result
echo "aws-gov" | sudo tee /run/cloud-init/cloud-id

# 2. Update the instance metadata JSON
# Note: If this is a fresh Multipass instance, we change 'multipass' to 'aws-gov'
sudo sed -i 's/multipass/aws-gov/g' /run/cloud-init/instance-data.json

# 3. Clear cache
sudo rm -rf /var/lib/ubuntu-advantage/private/machine-access-cache
```
Verify the mock
```
# Check that it says aws-gov
cloud-id
```

Try fips enablement:
```
sudo pro enable fips-updates
``` 

## Test Code
From host machine:
Build the deb based on this PR
```
./tools/build.sh noble
```
Delploy client
```
multipass launch --name fips-test noble
multipass transfer /path/to/the/built/debs/*.deb fips-test:/tmp/
multipass shell fips-test
```
From fips-test:
Install client
```
sudo apt install ./*.deb
```
Install AWS kernel
```
sudo apt update && sudo apt install -y linux-aws
sudo reboot
```
Mock aws-gov
```
# 1. Overwrite the result files
echo "aws-gov" | sudo tee /run/cloud-init/cloud-id-result
echo "aws-gov" | sudo tee /run/cloud-init/cloud-id

# 2. Update the instance metadata JSON
# Note: If this is a fresh Multipass instance, we change 'multipass' to 'aws-gov'
sudo sed -i 's/multipass/aws-gov/g' /run/cloud-init/instance-data.json

# 3. Clear cache
sudo rm -rf /var/lib/ubuntu-advantage/private/machine-access-cache
```
Verify the mock
```
# Check that it says aws-gov
cloud-id
```
Try fips enablement:
```
sudo pro enable fips-updates
```

### DIFF
--- a/uaclient/apt_news.py
+++ b/uaclient/apt_news.py
@@ -113,9 +113,12 @@ def do_selectors_apply(
             return False
 
     if selectors.clouds is not None:
+        from uaclient.clouds.identity import cloud_type_to_contract_cloud_type
+
         cloud_id, fail = get_cloud_type()
         if fail is not None:
             return False
+        cloud_id = cloud_type_to_contract_cloud_type(cloud_id)
         if cloud_id not in selectors.clouds:
             return False
 
--- a/uaclient/cli/fix.py
+++ b/uaclient/cli/fix.py
@@ -344,7 +344,10 @@ def _run_ua_attach(cfg: UAConfig, token: str) -> bool:
 
 def _inform_ubuntu_pro_existence_if_applicable() -> None:
     """Alert the user when running Pro on cloud with PRO support."""
+    from uaclient.clouds.identity import cloud_type_to_contract_cloud_type
+
     cloud_type, _ = get_cloud_type()
+    cloud_type = cloud_type_to_contract_cloud_type(cloud_type)
     if cloud_type in PRO_CLOUD_URLS.keys():
         print(
             messages.SECURITY_USE_PRO_TMPL.format(
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -23,6 +23,7 @@ PRO_CLOUD_URLS = {
 }
 
 CONTRACT_CLOUD_TYPE_ALIASES = {
+    "aws-china": "aws",
     "aws-gov": "aws",
 }
 
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -23,7 +23,6 @@ PRO_CLOUD_URLS = {
 }
 
 CONTRACT_CLOUD_TYPE_ALIASES = {
-    "aws-china": "aws",
     "aws-gov": "aws",
 }
 
--- a/uaclient/clouds/identity.py
+++ b/uaclient/clouds/identity.py
@@ -22,10 +22,21 @@ PRO_CLOUD_URLS = {
     "gce": messages.urls.PRO_ON_GCP_HOME_PAGE,
 }
 
+CONTRACT_CLOUD_TYPE_ALIASES = {
+    "aws-china": "aws",
+    "aws-gov": "aws",
+}
+
 
 class NoCloudTypeReason(Enum):
     NO_CLOUD_DETECTED = 0
     CLOUD_ID_ERROR = 1
+
+
+def cloud_type_to_contract_cloud_type(
+    cloud_type: Optional[str],
+) -> Optional[str]:
+    return CONTRACT_CLOUD_TYPE_ALIASES.get(cloud_type, cloud_type)
 
 
 def get_instance_id() -> Optional[str]:
--- a/uaclient/contract.py
+++ b/uaclient/contract.py
@@ -251,9 +251,11 @@ class UAContractClient(serviceclient.UAServiceClient):
 
         @return: Dict of the JSON response containing the contract-token.
         """
+        from uaclient.clouds.identity import cloud_type_to_contract_cloud_type
+
         response = self.request_url(
             API_V1_GET_CONTRACT_TOKEN_FOR_CLOUD_INSTANCE.format(
-                cloud_type=cloud_type
+                cloud_type=cloud_type_to_contract_cloud_type(cloud_type)
             ),
             data=data,
         )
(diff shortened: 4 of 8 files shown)


uaclient/clouds/identity.py
```diff
@@ -22,12 +22,23 @@
...
+    "aws-gov": "aws",
+}
+
 
 class NoCloudTypeReason(Enum):
     NO_CLOUD_DETECTED = 0
     CLOUD_ID_ERROR = 1
 
 
+def cloud_type_to_contract_cloud_type(
```
renanrodrigo, line 36: can't this be done directly in `get_cloud_type`?
I don't think we need a separate function for this - just convert it before returning there

srunde3: Agree with Renan on this point given the implementation.

(Sidebar about an alternative):

You might be able to justify a function if you want this to be the "public API" (w.r.t. the module) for getting cloud types from a contract cloud type. That would give you more flexibility in the future if you wanted to e.g., match on a prefix. In that case I think the `cloud_type` should not be optional, though.

uaclient/cli/fix.py
```diff
@@ -344,7 +344,10 @@ def _run_ua_attach(cfg: UAConfig, token: str) -> bool:
 
 def _inform_ubuntu_pro_existence_if_applicable() -> None:
     """Alert the user when running Pro on cloud with PRO support."""
+    from uaclient.clouds.identity import cloud_type_to_contract_cloud_type
```
srunde3: If it's possible, it is nice to avoid inline imports. I admit that sometimes it's difficult to unwind what's already there, though.

uaclient/clouds/identity.py
yujinhong-canonical: As discussed with @rr during meeting, a distinction between `aws` and `aws-gov` in cloud type is currently necessary, as there are [functions](https://github.com/canonical/ubuntu-pro-client/blob/88330d0a7d85ef5f7e2ac602740033105edd00f4/uaclient/clouds/identity.py) that require this. As are trying to move away from differentiating between `aws` cloud types, these legacy functions can be refactored later on. However, for the scope of this bug fix, we will keep the distinction as is and keep `get_cloud_type_to_contract_cloud_type`.

uaclient/cli/fix.py
yujinhong-canonical: @renanrodrigo since an official PPA has already been published for this fix, should I refrain from making this change in the repo, or can I go ahead and do so?

Since this fix is confirmed to be working, we can go ahead and merge this into main if we are not making this change.

renanrodrigo: meh
I'd say merge it, but open a bug stating this all needs to be improved, so you don't forget to revisit this in the future

srunde3:
@yujinhong-canonical could you take a look at the static analysis failures? I think the integration tests typically fail but I didn't think static analysis did.

srunde3:
@yujinhong-canonical Also could you please make a backlog item to get the integration tests running again? It looks like it's just an issue with Landscape credentials--getting it fixed will improve our velocity.

CC @renanrodrigo , maybe you already know which Landscape account we need to regenerate creds for? At the very least, maybe you know what sort of set up was needed on this account. I can assist in making a new one if we need to start from scratch.

uaclient/clouds/identity.py
```diff
@@ -22,12 +22,23 @@
...
+    cloud_type: Optional[str],
+) -> Optional[str]:
+    return CONTRACT_CLOUD_TYPE_ALIASES.get(cloud_type, cloud_type)
```
srunde3, line 39: It's odd to have a function that *just* wraps a dictionary. Receiving and returning `None` is also a little awkward here; I think it's better to check for `None` before calling this.

Even better would be providing some structure to our cloud types, e.g., with enums. This might take some more significant refactoring though.

uaclient/tests/test_contract.py
```diff
@@ -1684,6 +1684,50 @@ def test_different_overrides_applied_together(
 
 @mock.patch("uaclient.http.serviceclient.UAServiceClient.request_url")
 class TestRequestAutoAttach:
+    def test_request_normalizes_aws_gov_cloud_type(
```
srunde3, line 1687: nit: could these be parameterized and consolidated?

uaclient/apt_news.py
```diff
@@ -113,9 +113,12 @@ def do_selectors_apply(
             return False
 
     if selectors.clouds is not None:
+        from uaclient.clouds.identity import cloud_type_to_contract_cloud_type
+
         cloud_id, fail = get_cloud_type()
         if fail is not None:
             return False
+        cloud_id = cloud_type_to_contract_cloud_type(cloud_id)
```
srunde3, line 121: similar to above; maybe we can avoid an inline import.

uaclient/contract.py
```diff
@@ -251,9 +251,11 @@ def get_contract_token_for_cloud_instance(
 
         @return: Dict of the JSON response containing the contract-token.
         """
+        from uaclient.clouds.identity import cloud_type_to_contract_cloud_type
+
         response = self.request_url(
             API_V1_GET_CONTRACT_TOKEN_FOR_CLOUD_INSTANCE.format(
-                cloud_type=cloud_type
+                cloud_type=cloud_type_to_contract_cloud_type(cloud_type)
```
srunde3, line 258: same as above

uaclient/contract.py
```diff
@@ -925,7 +927,10 @@ def apply_contract_overrides(
 
     :param orig_access: Dict with original entitlement access details
     """
-    from uaclient.clouds.identity import get_cloud_type
+    from uaclient.clouds.identity import (
+        cloud_type_to_contract_cloud_type,
+        get_cloud_type,
+    )
```
srunde3, line 933: same as above; there's already an inline import here, so maybe it's difficult to unwind it. See if you have better luck :-)

srunde3 (approved):
Looks good to ship as-is since we already have this live and confirmed working. Let's make the clean-up changes in a follow-up PR.

srunde3:
Forced through since failing CI was unrelated to this PR.

PR #3538, opened 2026-05-19
chore: fix mypy static analysis

srunde3:
## Why is this needed?

CI is broken due to invalid mypy typing. Most of these are situations in which a base class intends for a subclass to define a field and uses `None` as a sentinel value. This makes the field optional, and it complicates the typing a bit.

Where possible I've just marked the types as optional. In some cases, all the subclasses properly override the attribute, so we can replace `None` with `""` to simplify typing.

I've added test coverage for `UbuntuProError` and have added a safe default in case `_msg` and `_formatted_msg` aren't defined; this simplifies the typing. I couldn't find any examples where these weren't overridden by a leaf class anyway. This change ensures that in the worst case, the user gets a slightly less helpful error message instead of a runtime error.

## Test Steps

See the CI for "Static Analysis"

### DIFF
--- a/uaclient/api/u/pro/security/cves/_common/v1.py
+++ b/uaclient/api/u/pro/security/cves/_common/v1.py
@@ -205,7 +205,7 @@ class VulnerabilitiesAlreadyFixed:
 
 
 class VulnerabilityParser(metaclass=abc.ABCMeta):
-    vulnerability_type = None  # type: str
+    vulnerability_type: str
 
     @abc.abstractmethod
     def get_package_vulnerabilities(
--- a/uaclient/api/u/pro/security/cves/_common/v1.py
+++ b/uaclient/api/u/pro/security/cves/_common/v1.py
@@ -205,10 +205,9 @@ class VulnerabilitiesAlreadyFixed:
 
 
 class VulnerabilityParser(metaclass=abc.ABCMeta):
-    @property
-    @abc.abstractmethod
-    def vulnerability_type(self) -> str:
-        pass
+
+    # subclasses must override this
+    vulnerability_type = ""  # type: str
 
     @abc.abstractmethod
     def get_package_vulnerabilities(
--- a/uaclient/api/u/pro/security/cves/_common/v1.py
+++ b/uaclient/api/u/pro/security/cves/_common/v1.py
@@ -205,7 +205,10 @@ class VulnerabilitiesAlreadyFixed:
 
 
 class VulnerabilityParser(metaclass=abc.ABCMeta):
-    vulnerability_type = None  # type: str  # type: ignore[assignment]
+    @property
+    @abc.abstractmethod
+    def vulnerability_type(self) -> str:
+        pass
 
     @abc.abstractmethod
     def get_package_vulnerabilities(
--- a/uaclient/api/u/pro/security/cves/_common/v1.py
+++ b/uaclient/api/u/pro/security/cves/_common/v1.py
@@ -205,7 +205,7 @@ class VulnerabilitiesAlreadyFixed:
 
 
 class VulnerabilityParser(metaclass=abc.ABCMeta):
-    vulnerability_type: str
+    vulnerability_type = None  # type: str  # type: ignore[assignment]
 
     @abc.abstractmethod
     def get_package_vulnerabilities(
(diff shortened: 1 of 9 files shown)


uaclient/api/u/pro/security/cves/_common/v1.py
```diff
@@ -205,7 +205,10 @@ def to_dict(self):
 
 
 class VulnerabilityParser(metaclass=abc.ABCMeta):
-    vulnerability_type = None  # type: str
+    @property
+    @abc.abstractmethod
+    def vulnerability_type(self) -> str:
+        pass
```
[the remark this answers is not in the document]
srunde3: Mypy passes locally and we already use this pattern in `UAEntitlement`.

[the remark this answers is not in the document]
srunde3: I've chosen to revert this anyways, even though it passed. It's a little more straightforward and it's similar to my other fix for `UAEntitlement`.

uaclient/exceptions.py
```diff
@@ -45,17 +45,21 @@ class UbuntuProError(Exception):
...
         if self._formatted_msg is not None:
-            self.named_msg = self._formatted_msg.format(
-                **kwargs
-            )  # type: messages.NamedMessage
+            self.named_msg = self._formatted_msg.format(**kwargs)
         else:
+            if self._msg is None:
+                raise AttributeError(
+                    "{} must define either _msg or _formatted_msg".format(
+                        self.__class__.__name__
```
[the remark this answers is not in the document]
srunde3: good catch, bot

uaclient/api/u/pro/security/cves/_common/v1.py
```diff
@@ -205,7 +205,9 @@ def to_dict(self):
 
 
 class VulnerabilityParser(metaclass=abc.ABCMeta):
-    vulnerability_type = None  # type: str
+
+    # subclasses must override this
+    vulnerability_type = ""  # type: str
```
akinshipe, line 210: Observation: I see there is only one concrete class of this abstract class. Not sure why we needed polymorphism here in the first place.

uaclient/clouds/identity.py
```diff
@@ -36,6 +36,8 @@ class NoCloudTypeReason(Enum):
 def cloud_type_to_contract_cloud_type(
     cloud_type: Optional[str],
 ) -> Optional[str]:
+    if cloud_type is None:
+        return None
```
akinshipe, line 40: Honestly, we can't blame Mypy for not knowing that `CONTRACT_CLOUD_TYPE_ALIASES.get(None, None)` will return None.

uaclient/exceptions.py
```diff
@@ -45,18 +45,23 @@ class UbuntuProError(Exception):
...
+            self.named_msg = self._formatted_msg.format(**kwargs)
+        elif self._msg is not None:
             self.named_msg = self._msg
+        else:
+            # This should not be reached in practice; subclasses should
+            # override either _msg or _formatted_msg. Provide a safe default
+            # just in case.
+            self.named_msg = messages.NamedMessage(
+                "ubuntu-pro-error", "an unexpected error occurred."
+            )
```
akinshipe, line 64: Nit: Not sure if this is worth implementing. But we could have used the dunder method `__init_subclass__` hook to enforce that one of the fields is defined by a subclass. And that a child class will not be used without providing one. This approach is fine as well.

akinshipe (approved):
LGTM. Thanks for the work

uaclient/clouds/identity.py
srunde3: Mypy is trying its best :smiling_face_with_tear:

srunde3:
bypassed since docs check is failing.

uaclient/clouds/identity.py
yujinhong-canonical: Question (out of curiosity): I tracked the commit history to check what broke the static analysis CI check. I'm pretty sure I broke it in [the fips bug fix PR](https://github.com/canonical/ubuntu-pro-client/pull/3533) 😅 , for this reason that Ola mentioned above. But seeing the other fixes you made, it makes me question, how did mypy not break before?

yujinhong-canonical:
Late to the train here, but all changes LGTM! Thank you @akinshipe for the quick review and for helping me understand mypy :smile:

PR #3539, opened 2026-05-21
fix: add cap perfmon to cloud_id AppArmor profile #LP: 2153472

srunde3:
## Why is this needed?

This fixes AppArmor denials seen in Azure VMs that attach to Pro on Noble.

## Test Steps

See SRU note. The CPC team has manually tested this fix and has confirmed that it addresses the bug.


AleksaSvitlica (approved):
I can confirm the reproduction steps in the markdown file are what I used to reproduce this issue.

Spencer provides a `.deb` of the `ubuntu-pro-client` with this change and when I installed that in the VM I was able to confirm that running `apt update` no longer produced the log.

renanrodrigo (approved):
LGT CPC
LGTM

srunde3:
Forcing through since CI is broken.

PR #3547, opened 2026-06-02
refactor: introduce package categorization objects for pro/security/fix/_common/plan/v1

srunde3:
## Why is this needed?

This is some prerequisite work before introducing a bug fix. This separates the calculations of package state (installed, available, minimum required fix version, etc.) from the calculations of how these states affect the fix plan. Calculations of package state are encapsulated into "Categorization" objects to make the code a bit more readable.

The bug fix will be adding a new logical branch that depends on the package state.

## Test Steps

Unit tests should be sufficient for this. It does not introduce any new behaviors and does not modify the unit tests.


renanrodrigo (approved):
Thanks for the work, Spencer!

I feel sad that this whole fix plan/execute got another layer of abstraction.
I also feel sad that some of it is as confusing as before, just explicitly confusing (like "installed source packages", which effectively isn't a thing...)

That said

I am happy with this changeset, because the goal is well met: separation of concerns is good and makes sense. And I do believe it will be easier to track and fix bugs given the context.

uaclient/api/u/pro/security/fix/_common/plan/v1.py
```diff
@@ -49,6 +54,119 @@
...
+        released_binaries_for_source: Dict[str, Dict[str, str]],
+    ) -> "BinaryPackageCategorization":
+        if binary_package in released_binaries_for_source:
+            required_fixed_version = released_binaries_for_source.get(
+                binary_package, {}
+            ).get(
+                "version", ""
+            )  # type: Optional[str]
+        else:
+            required_fixed_version = None
```
yujinhong-canonical, line 93: Nit: The second argument of `.get(binary_package, {})` will not be used, since `.get()` is inside this `if` statement.

Could this code block be simplified to something like this?

```
required_fixed_version = released_binaries_for_source.get(
    binary_package, {}
).get("version", None)
```

uaclient/api/u/pro/security/fix/_common/plan/v1.py
```diff
@@ -49,6 +54,119 @@
...
+            source_package=source_package,
+            binary_package=binary_package,
+            installed_version=installed_version,
+            required_fixed_version=required_fixed_version,
+        )
+
+    @property
+    def needs_upgrade_to_fix(self) -> bool:
+        """Whether this binary package needs an upgrade in order to fix a
+        vulnerability it is affected .
```
yujinhong-canonical: nit: delete space between "affected" and "."

yujinhong-canonical (commented):
LGTM! Confirming that all unit tests passed as well.

uaclient/api/u/pro/security/fix/_common/plan/v1.py
srunde3: yep, good point! It could almost be simplified to that, but we'd lose our distinction between `binary_package in released_binaries_for_source` or not. We want `None` only when it's not present. If it's present but with no `"version"` data, we want `""`.

srunde3:
We may revisit this in the future. We are focusing on an SRU for the next couple months and do not want to complicate the backport process.

PR #3557, opened 2026-06-22
ci: run resolute and stonking in behave test workflow

srunde3:
## Why is this needed?

This ensures that as tests are added, they are gated in CI.

Tests for Resolute are being added; there may not be any ready to run when this PR goes through. There also aren't any tests for Stonking yet.

This PR also makes a small refactor to indicate that we `--skip-proposed` for the development release, not arbitrary releases. This should make it more clear when something needs to be updated.

## Test Steps

None - see changes working CI. Should see "passed" for resolute and stonking since we don't have any tests.

### DIFF
--- a/.github/workflows/ci-integration.yaml
+++ b/.github/workflows/ci-integration.yaml
@@ -14,6 +14,9 @@ defaults:
   run:
     shell: sh -ex {0}
 
+env:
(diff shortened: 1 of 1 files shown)


yujinhong-canonical (approved):
LGTM! Thank you for the work!

PR #3558, opened 2026-06-24
test: run features/cli tests on resolute

srunde3:
## Why is this needed?

Update tests in `features/cli` to run on Resolute. Drop plucky where present.

For the most part this is just extending whatever is there. Three issues discovered while updating tests:

1. AppArmor issue for a locales file
2. Ansible not in ESM; test structure assumes presence
3. Even though Resolute doesn't have CIS/USG ready, we need to update the `presentedAs` in the back end

All issues are tracked in Jira and have acceptance criteria to update the relevant tests.

## Test Steps

```sh
UACLIENT_BEHAVE_INSTALL_FROM=local tox -e behave -- features/cli -D releases=resolute -D machine_types=lxd-container,lxd-vm
```


features/cli/status.feature
```diff
@@ -623,27 +621,67 @@ Feature: CLI status command
...
     Examples: ubuntu release
-      | release | machine_type  |
-      | noble   | lxd-container |
+      | release  | machine_type  |
+      | resolute | lxd-container |
+
+  @uses.config.contract_token @arm64
+  Scenario Outline: Attached status in the latest LTS ubuntu machine - resolute
+    Given a `<release>` `<machine_type>` machine with ubuntu-advantage-tools installed
+    When I attach `contract_token` with sudo
```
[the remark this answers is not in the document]
srunde3: This is deliberate; this runs against arm64. These were previously separated out for noble as well.

features/cli/attach.feature
```diff
@@ -412,3 +414,7 @@ Feature: CLI attach command
       | focal   | lxd-vm       | enabled          |
       | jammy   | lxd-vm       | enabled          |
       | noble   | lxd-vm       | enabled          |
+
+# TODO: Re-enable resolute once cloud_id AppArmor profile
```
renanrodrigo, line 418: Same comment from other PR about how to track the tests to enable

features/cli/config.feature
```diff
@@ -59,7 +59,7 @@ Feature: CLI config command
       usage: pro config [-h] {show,set,unset} ...
       pro config: error: argument command: invalid choice: 'invalid' (choose from 'show', 'set', 'unset')
       """
-    Then if `<release>` in `plucky` and stderr contains substring:
+    Then if `<release>` in `resolute` and stderr contains substring:
```
renanrodrigo, line 62: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa this step

I hate how this is an ugly workaround, but it could be reworded.

Not now, but consider this for the future.

features/cli/enable.feature
```diff
@@ -300,6 +302,9 @@ Feature: CLI enable command
       | jammy   | lxd-container | anbox-cloud, cc-eal, esm-apps, esm-apps-legacy, esm-infra, esm-infra-legacy,\nfips, fips-preview, fips-updates, landscape, livepatch, realtime-kernel, ros,\nros-updates, usg. |
       | noble   | lxd-container | anbox-cloud, cc-eal, esm-apps, esm-apps-legacy, esm-infra, esm-infra-legacy,\nfips, fips-preview, fips-updates, landscape, livepatch, realtime-kernel, ros,\nros-updates, usg. |
 
+  # TODO: Re-enable resolute once ua-contracts resource definitions are
```
renanrodrigo: For these tests, there could be a "short version" for resolute only using the services that are available (esm, landscape, livepatch) until new services arrive - these new services may take quite a while to be available, and there is no reason to lose coverage.

features/cli/status.feature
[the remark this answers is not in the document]
renanrodrigo: Actually
The arm64 tag is used to identify which tests are executed on arm64 runs, but they don't need to be duplicated.
If there was a separation for Noble, I would say some noble services availability were different on arm64?
I think this can be a single test with the tag.

renanrodrigo (commented):
Thanks for the work, Spencer!

Some notes and questions before we call this done:
- Attach command with attach config test in attach.feature covers no supported LTS right now. Consider adding resolute there - I understand no CIS so services need to be adapted.
- Attach command in an generic cloud images in attach.feature should test resolute too.
- I'd like to see *some* security_status test for resolute. IIRC tests are not executed if they don't have the examples table below, so the existing tests need to be changed to have that table too... (like the Focal test has). Because this is a very release-specific command, it's likely that a whole new test needs to be in place.

features/cli/attach.feature
srunde3: This one is tracked here: https://warthogs.atlassian.net/browse/UPRO-1218

srunde3:
@renanrodrigo regarding your comments:

- This test is difficult to adapt - cis/usg are the only "entitled but not enabled-by-default" services it seems. Landscape sort of fits this category but it needs extra config, so it breaks the test.
- I'm skipping clouds for now - will update in a later PR
- I added a security status test for supported LTS, using Resolute as the first example.

renanrodrigo (approved):
Changeset is sane
Further items properly tracked
LGTM +1

PR #3564, opened 2026-07-03
docs: how to migrate attached machines

Leinadium:
## Why is this needed?
Adding a tutorial about migrating attached machines, and explanations about the behavior of machine activity count during these operations.

### DIFF
--- a/docs/howtoguides/how_to_migrate_attached_machines.rst
+++ b/docs/howtoguides/how_to_migrate_attached_machines.rst
@@ -1,7 +1,7 @@
 .. _how-to-migrate-attached-machines:
 
 How to manage Ubuntu Pro attachment when moving machines
-*********************************************
+********************************************************
 
 This guide explains what happens to your Ubuntu Pro attachment when you move,
 clone, rebuild, or delete a machine, and what actions (if any) you need to
--- a/docs/howtoguides/how_to_migrate_attached_machines.rst
+++ b/docs/howtoguides/how_to_migrate_attached_machines.rst
@@ -1,0 +1,122 @@
+.. _how-to-migrate-attached-machines:
+
+How to migrate an Ubuntu Pro attached machine
+*********************************************
+
+This guide explains what happens to your Ubuntu Pro attachment when you move,
+clone, rebuild, or delete a machine, and what actions (if any) you need to
+take in each case.
+
+Ubuntu Pro attachment information is stored locally on the machine. The
+subscription's active machine count is based on activity. If a machine is
+stopped or deleted, it will eventually be removed from the active count
+automatically. No manual detach is required.
+
+.. note::
+
+   Run ``sudo pro status`` or ``sudo pro refresh`` at any time to verify the
+   current attachment state of a machine.
+
+Moving a VM to another host or hypervisor
+-----------------------------------------
+
+When you move an attached VM to a different host, hypervisor, or virtual
+environment (for example, a live migration or export/import), the Pro
+attachment travels with it. The attachment data is part of the filesystem,
+so no re-attachment is needed after the move.
+
+#. Move the VM using your platform's standard tooling.
+#. After the move, verify the attachment is intact:
+
+   .. code-block:: bash
+
+      sudo pro status
+
+No further action is required. The machine continues to be counted as one
+active machine on your subscription, as it was before.
+
+Cloning an attached VM
+-----------------------
+
+When you clone an attached VM, the clone inherits the attachment from the
+original. Both the original and the clone will be counted as separate active
+machines on your subscription.
+
+If you intend to **replace** the original with the clone (for example, for
+testing or migration), stop or delete the original. It will eventually be
+removed from the active count automatically.
+
+#. Clone the VM using your platform's standard tooling.
+#. Verify the attachment on the new clone:
+
+   .. code-block:: bash
+
+      sudo pro status
+
+#. If you no longer need the original, stop or delete it. No detach is needed
+   before deletion.
+
+.. note::
+
+   If you are creating a **reusable template or golden image** from an attached
+   VM, you should not leave the image in an attached state. See
+   :ref:`create_pro_golden_image` for the recommended approach to creating
+   Ubuntu Pro golden images.
+
+Rebuilding a machine with a fresh Ubuntu installation
+------------------------------------------------------
+
+A fresh Ubuntu installation does not carry over any previous Pro attachment.
+You will need to attach the rebuilt machine to your subscription:
+
+.. code-block:: bash
+
+   sudo pro attach <your-pro-token>
+
+Your Pro token is available from the `Ubuntu Pro Dashboard <Pro_dashboard_>`_.
+
+If the previous machine is no longer in use, there is no need to detach it
+first. It will eventually be removed from the active count automatically.
+
+Decommissioning or deleting a machine
+--------------------------------------
+
+You do not need to detach a machine before decommissioning or deleting it. Once
+the machine is gone and no longer communicates with the Ubuntu Pro service, it
+will eventually be removed from the active count automatically.
+
+You can detach before deleting:
+
+.. code-block:: bash
+
+   sudo pro detach
+
+However, this step is entirely optional.
+
+What if the old machine is already gone?
+----------------------------------------
+
+If a machine was deleted without being detached first, no action is required.
+The machine will eventually be removed from the active count automatically.
+
+Verifying the result
+--------------------
+
+After any migration operation, use the following commands to confirm the state
+of your attachment:
+
+- Check the current attachment status and enabled services:
+
+  .. code-block:: bash
+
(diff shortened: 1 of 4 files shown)


docs/howtoguides/how_to_migrate_attached_machines.rst
```diff
@@ -0,0 +1,122 @@
...
+How to migrate an Ubuntu Pro attached machine
+*********************************************
+
+This guide explains what happens to your Ubuntu Pro attachment when you move,
+clone, rebuild, or delete a machine, and what actions (if any) you need to
+take in each case.
+
+Ubuntu Pro attachment information is stored locally on the machine. The
+subscription's active machine count is based on activity. If a machine is
+stopped or deleted, it will eventually be removed from the active count
```
YanisaHS, line 12: it will "eventually" be removed - can you provide something more specific? I noticed it in a few other places throughout this doc too.

If it's too complicated, you can just leave it as is (with "eventually"), but if it's pretty straightforward, it'd be useful to add this information (e.g., if there's a job that runs weekly and cleans up inactive machines, something like that)

docs/howtoguides/how_to_migrate_attached_machines.rst
```diff
@@ -0,0 +1,122 @@
...
+#. Verify the attachment on the new clone:
+
+   .. code-block:: bash
+
+      sudo pro status
+
+#. If you no longer need the original, stop or delete it. No detach is needed
+   before deletion.
+
+.. note::
```
YanisaHS: Should this note be at the top of the section instead (line 40) ?

I don't feel that strongly about this, but at first pass to me this section read like cloning an attached VM is like creating a template, but it seems like that's not the recommended approach for that outcome.

docs/howtoguides/how_to_migrate_attached_machines.rst
```diff
@@ -0,0 +1,122 @@
...
+
+   sudo pro detach
+
+However, this step is entirely optional.
+
+What if the old machine is already gone?
+----------------------------------------
+
+If a machine was deleted without being detached first, no action is required.
+The machine will eventually be removed from the active count automatically.
```
YanisaHS:
```suggestion
```
This seems clear already in the Decommissioning section

docs/howtoguides/how_to_migrate_attached_machines.rst
```diff
@@ -0,0 +1,122 @@
...
+the machine is gone and no longer communicates with the Ubuntu Pro service, it
+will eventually be removed from the active count automatically.
+
+You can detach before deleting:
+
+.. code-block:: bash
+
+   sudo pro detach
+
+However, this step is entirely optional.
```
YanisaHS:
```suggestion
You can manually detach before deleting, but this is optional:

.. code-block:: bash

   sudo pro detach
```
To put the optional part before the actual code to run (so it's clear it's optional before they actually run it)

docs/howtoguides/how_to_migrate_attached_machines.rst
```diff
@@ -0,0 +1,122 @@
...
+   If you are creating a **reusable template or golden image** from an attached
+   VM, you should not leave the image in an attached state. See
+   :ref:`create_pro_golden_image` for the recommended approach to creating
+   Ubuntu Pro golden images.
+
+Rebuilding a machine with a fresh Ubuntu installation
+------------------------------------------------------
+
+A fresh Ubuntu installation does not carry over any previous Pro attachment.
+You will need to attach the rebuilt machine to your subscription:
```
YanisaHS:
```suggestion
If you reinstall a fresh Ubuntu image, the new installation does not carry over the previous Pro attachment.
You will need to attach the rebuilt machine to your subscription:
```
If this is correct? The clarification I'm trying to make is that it's a _reinstallation_ of Ubuntu on the existing VM (or whatever). Feel free to reword or reject - I don't feel that strongly on this one.

docs/howtoguides/how_to_migrate_attached_machines.rst
```diff
@@ -0,0 +1,122 @@
+.. _how-to-migrate-attached-machines:
+
+How to migrate an Ubuntu Pro attached machine
```
YanisaHS: After reading this, I don't think "migration" is the right framing anymore. How about

"How to manage Ubuntu Pro attachment when moving or replacing machines" (or just "when moving machines" instead of "when moving or replacing machines")

and the sidebar could stay the same

YanisaHS (changes requested):

docs/howtoguides/how_to_migrate_attached_machines.rst
Leinadium: Previously it was "24 hours", but this number is not that strict (might fluctuate). Actually, in the Pro dashboard, this number has a "beta" tag, so users don't expect this number to reflect the current reality.

So to prevent any misunderstandings, Thiago suggested me to just use "eventually". I can change to "around 24h" or something similar, if you feel it might be better.

docs/howtoguides/how_to_migrate_attached_machines.rst
Leinadium: Note moved.

docs/howtoguides/how_to_migrate_attached_machines.rst
Leinadium: Makes sense, applying the suggestion

docs/howtoguides/how_to_migrate_attached_machines.rst
Leinadium: Seems fair. Changing the title to `How to manage Ubuntu Pro attachment when moving machines`

YanisaHS:
Your build is failing btw - I've copied the message here (but you can also get it locally by using `make run`)

```
/home/docs/checkouts/readthedocs.org/user_builds/canonical-ubuntu-pro-client/checkouts/3564/docs/howtoguides/how_to_migrate_attached_machines.rst:4: WARNING: Title underline too short.
```

Can you tag me when the build is successful again and I'll verify it previews ok?

docs/howtoguides/how_to_migrate_attached_machines.rst
YanisaHS: Got it, let's leave it as "eventually" then if it fluctuates.

Leinadium:
@YanisaHS the build is now passing. The fix was easy. but I was wondering why the other checks weren't passing. Turns out it was caused by some instability, and they are now passing too.

YanisaHS (approved):

srunde3 (approved):
LGTM, thanks!

docs/howtoguides/how_to_migrate_attached_machines.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Cloning an attached VM
+-----------------------
+
+When you clone an attached VM, the clone inherits the attachment from the
+original. Both the original and the clone will be counted as separate active
+machines on your subscription.
+
+If you intend to **replace** the original with the clone (for example, for
+testing or migration), stop or delete the original. It will eventually be
```
ingarisle, line 46: It will only be removed once the "replaced" machine is deleted right? Maybe this could be a bit clearer for customers who worry about their machine count and want to understand exactly when and how the duplicate would be removed

docs/howtoguides/how_to_migrate_attached_machines.rst
```diff
@@ -0,0 +1,114 @@
...
+You will need to attach the rebuilt machine to your subscription:
+
+.. code-block:: bash
+
+   sudo pro attach <your-pro-token>
+
+Your Pro token is available from the `Ubuntu Pro Dashboard <Pro_dashboard_>`_.
+
+If the previous machine is no longer in use, there is no need to detach it
+first. It will eventually be removed from the active count automatically.
```
ingarisle, line 79: do we have a rough time estimate of how long it takes until the machine is removed automatically?

ingarisle (commented):
Looks great and will be very helpful for customer! I've added two small comments on the automatic removal of machines where we potentially could add some more detail.

docs/howtoguides/how_to_migrate_attached_machines.rst
Leinadium: Me and Yanisa [briefly discussed it here](https://github.com/canonical/ubuntu-pro-client/pull/3564#discussion_r3531113162). We decided to not put exact numbers on the documentation, as the numbers fluctuate a little.

docs/howtoguides/how_to_migrate_attached_machines.rst
Leinadium: Yes, it is removed once the old machine is not used anymore or deleted. I explained that the only required steps is to delete the old machine.
The exact "when" is explained in the next comment: it is not an exact number, but *around* 24 hours.

PR #3565, opened 2026-07-06
chore: add AGENTS.md

srunde3:
## Why is this needed?

`AGENTS.md` is the standard context for agent sessions. This PR introduces a minimal file including test commands, and an overview of the repo.

Treat this as an architectural change and review it as such; there isn't a concrete test to show that this is the "correct" agents file. Review that the content here is correct; I'm not too concerned about making the file complete yet; we can add as we go.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,68 @@
+# AGENTS.md
+
+## Project overview
+
+`ubuntu-pro-client` (the `pro` command, package `ubuntu-advantage-tools`) is a Python client that attaches Ubuntu machines to Ubuntu Pro and manages its services (ESM, Livepatch, FIPS, etc.). The `main` branch supports very old Ubuntu releases (back to 16.04 Xenial), so avoid modern-only syntax and dependencies. Xenial runs on Python 3.5.
+
+- Target Python is the version shipped with supported Ubuntu releases; do not assume the latest Python features are available.
+- Only use dependencies already declared in `requirements*.txt` / `setup.py`. Do not add new runtime dependencies without strong justification.
+
+## Repository layout
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,66 @@
+# AGENTS.md
+
+## Project overview
+
+`ubuntu-pro-client` is a Python client that attaches Ubuntu machines to Ubuntu Pro and manages its services (ESM, Livepatch, FIPS, etc.). The `main` branch supports very old Ubuntu releases (back to 16.04 Xenial), so avoid modern-only syntax and dependencies. Xenial runs on Python 3.5.
+
```
akinshipe, line 16: Maybe add an instruction Terminology section at the top and then use the terminologies in the rest of the document. Below are terminologies for Agents

```
Instruction terminology

The keywords MUST, MUST NOT, SHOULD, SHOULD NOT and MAY describe the strength and meaning of instructions in this file and other agent files in this repo.

MUST — This is a mandatory requirement. Follow it unless a higher-priority instruction from the user or execution environment directly conflicts with it.
MUST NOT — This action is prohibited.
SHOULD — This is the expected default. Deviate only when there is a clear, task-specific reason, and explain the deviation in the final response.
SHOULD NOT — Avoid this action unless there is a clear, task-specific reason to do otherwise.
MAY — This action is optional. Use judgment based on the task and surrounding code.
```

A good example of where to use this will be here:

`The main branch supports very old Ubuntu releases (back to 16.04 Xenial), so MUST avoid modern-only syntax and dependencies. Xenial runs on Python 3.5.`

akinshipe (approved):

AGENTS.md
srunde3: I like this idea - I've incorporated it.

PR #3568, opened 2026-07-09
ci: fix broken azure-mgmt versions against pycloudlib; use Standard_B2s_v2 images on Azure

srunde3:
## Why is this needed?

The newest version of the Azure network management library is not compatible with pycloudlib. I've reported upstream as well: https://github.com/canonical/pycloudlib/issues/511

I'm also lumping in a fix for another known issue with Azure compute management: https://github.com/canonical/pycloudlib/issues/501

Lastly, Azure (or at least Canonical's org) also doesn't stock `Standard_B2s` images AFAIK; they're all `Standard_B2s_v2`.

## Test Steps

You'll need Azure credentials to reproduce and verify that this fixes the problem. Run a test like so, with and without this fix:

```sh
export UACLIENT_BEHAVE_CONTRACT_TOKEN=<YOUR TOKEN>
UACLIENT_BEHAVE_INSTALL_FROM=local && tox -r -e behave -- features/cli/attach.feature -D releases=focal -D machine_types=azure.generic
```

### DIFF
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -536,7 +536,7 @@ class Azure(Cloud):
         )
         inst = self.api.launch(
             image_id=image_name,
-            instance_type="Standard_B2s",
+            instance_type="Standard_B2s_v2",
             user_data=user_data,
             inbound_ports=inbound_ports,
         )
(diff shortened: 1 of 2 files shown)


yujinhong-canonical (approved):
LGTM!

PR #3569, opened 2026-07-10
fix: xenial container network setup on recent-LTS hosts

yujinhong-canonical:
## Why is this needed?
Launching a Xenial LXD container during behave tests on a recent-LTS host produces a broken container with no IPv4 address. The host AppArmor policy prevents Xenial's `dhclient` from completing DHCP, so the container never gets network.

This PR disables the in-container dhclient AppArmor profile and re-acquires the lease, restoring connectivity.

## Test Steps
Prepare a host machine that is Noble (24.04) or more recent.
Run a feature test on Xenial using `lxd-container`, and verify that the test no longer fails during token attachment.
```
UACLIENT_BEHAVE_CONTRACT_TOKEN=<your-token> \
tox -e behave -- features/cli/attach.feature -D releases=xenial -D machine_types=lxd-container
```

### DIFF
--- a/features/steps/machines.py
+++ b/features/steps/machines.py
@@ -98,6 +98,23 @@ def given_a_machine(
     )
 
     if series == "xenial":
+        # On recent host kernels, the host AppArmor policy denies xenial's
+        # ISC dhclient from creating the raw AF_PACKET (LPF) socket it needs
+        # ("Open a socket for LPF: Permission denied"), so the container never
+        # gets an IPv4 DHCP lease and has no outbound network/DNS. Disable the
+        # in-container dhclient AppArmor profile and re-acquire the lease so
+        # the machine can reach the archive and contracts endpoints.
+        when_i_run_command(
+            context,
+            "bash -c '"
+            "ln -sf /etc/apparmor.d/sbin.dhclient /etc/apparmor.d/disable/ 2>/dev/null; "  # noqa: E501
+            "apparmor_parser -R /etc/apparmor.d/sbin.dhclient 2>/dev/null; "
+            "dhclient -v eth0'",
+            "with sudo",
+            verify_return=False,
+            machine_name=machine_name,
+        )
+
         # Upgrading open-iscsi to esm version on xenial restarts this service
         # This sometimes causes resource errors on github action runners
         when_i_run_command(
--- a/features/steps/machines.py
+++ b/features/steps/machines.py
@@ -98,22 +98,24 @@ def given_a_machine(
     )
 
     if series == "xenial":
-        # On recent host kernels, the host AppArmor policy denies xenial's
-        # ISC dhclient from creating the raw AF_PACKET (LPF) socket it needs
-        # ("Open a socket for LPF: Permission denied"), so the container never
-        # gets an IPv4 DHCP lease and has no outbound network/DNS. Disable the
-        # in-container dhclient AppArmor profile and re-acquire the lease so
-        # the machine can reach the archive and contracts endpoints.
-        when_i_run_command(
-            context,
-            "bash -c '"
-            "ln -sf /etc/apparmor.d/sbin.dhclient /etc/apparmor.d/disable/ 2>/dev/null; "  # noqa: E501
-            "apparmor_parser -R /etc/apparmor.d/sbin.dhclient 2>/dev/null; "
-            "dhclient -v eth0'",
-            "with sudo",
-            verify_return=False,
-            machine_name=machine_name,
-        )
+        if machine_type == "lxd-container":
+            # On recent host kernels (Noble+), the host AppArmor policy denies
+            # xenial's ISC dhclient the raw AF_PACKET (LPF) socket it needs
+            # ("Open a socket for LPF: Permission denied"), so the container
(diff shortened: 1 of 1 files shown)


features/steps/machines.py
```diff
@@ -98,6 +98,23 @@ def given_a_machine(
     )
 
     if series == "xenial":
```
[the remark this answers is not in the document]
yujinhong-canonical: Addressed this in [this commit](https://github.com/canonical/ubuntu-pro-client/pull/3569/changes/67b37788d82a39b7e6f1ab49528721b3086f1a3b). I chose to nest the second statement instead, because the code after this applies to all xenial scenarios, not just `lxd-container`.

features/steps/machines.py
```diff
@@ -98,6 +98,23 @@ def given_a_machine(
...
+        when_i_run_command(
+            context,
+            "bash -c '"
+            "ln -sf /etc/apparmor.d/sbin.dhclient /etc/apparmor.d/disable/ 2>/dev/null; "  # noqa: E501
+            "apparmor_parser -R /etc/apparmor.d/sbin.dhclient 2>/dev/null; "
+            "dhclient -v eth0'",
+            "with sudo",
+            verify_return=False,
+            machine_name=machine_name,
+        )
```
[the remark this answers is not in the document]
yujinhong-canonical: Addressed this in [the same commit](https://github.com/canonical/ubuntu-pro-client/pull/3569/changes/67b37788d82a39b7e6f1ab49528721b3086f1a3b) as the previous copilot comment. I chose to add the timeout, but kept the rest as is.

features/steps/machines.py
```diff
@@ -98,6 +98,25 @@ def given_a_machine(
...
+            # xenial's ISC dhclient the raw AF_PACKET (LPF) socket it needs
+            # ("Open a socket for LPF: Permission denied"), so the container
+            # never gets an IPv4 DHCP lease and has no outbound network/DNS.
+            # Disable the in-container dhclient AppArmor profile and re-acquire
+            # the lease so the machine can reach the archive and contracts
+            # endpoints.
+            when_i_run_command(
+                context,
+                "bash -c '"
+                "ln -sf /etc/apparmor.d/sbin.dhclient /etc/apparmor.d/disable/ 2>/dev/null; "  # noqa: E501
```
srunde3, line 112: is AppArmor for sure the culprit here? I was able to get an IPv4 address on the container with just the `dhclient` bit of this

yujinhong-canonical: What do you mean by just the `dhclient` bit of this? Also, were you able to get an IPv4 address before applying this PR?

yujinhong-canonical: Do you mean just this: `"timeout 60 dhclient -v eth0'",`?

srunde3: Right:

```sh
root@test-xenial:~# ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
92: eth0@if93: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 00:16:3e:30:2e:f3 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet6 fe80::216:3eff:fe30:2ef3/64 scope link 
       valid_lft forever preferred_lft forever
```

And then

```sh
root@test-xenial:~# dhclient eth0 
root@test-xenial:~# ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
92: eth0@if93: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 00:16:3e:30:2e:f3 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.194.61.69/24 brd 10.194.61.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::216:3eff:fe30:2ef3/64 scope link 
       valid_lft forever preferred_lft forever
```

yujinhong-canonical: Hmm that is interesting, so looks like `dhclient eth0` was enough to get an IP address to the xenial container.

However, when I tested with AppArmor lines commented out, leaving just this line (`"timeout 60 dhclient -v eth0'",`), this failed all four tests during attach, even though my token was correct. Placing the AppArmor lines back in fixed the issue.

Let me investigate this further.

srunde3: IIRC Xenial containers need cgroups v1; this might be the issue we're running into.

PR #3572, opened 2026-07-14
launch xenial lxd vm without secureboot on newer lxd

yujinhong-canonical:
## Why is this needed?
Launching a xenial lxd-vm on a Noble host fails at creation: `Failed creating instance record: "security.secureboot" is not supported for "virtual-machine"`
Recent LXD versions dropped the security.secureboot key for VMs. I've changed the code so that when LXD rejects secureboot, the launch now retries without it. Only xenial is affected, since it's the only series that sets the key.

## Test Steps
Prepare a host machine that is Noble (24.04) or more recent.
Run a feature test on Xenial using `lxd-vm`, and verify that the test no longer fails during token attachment.
```
UACLIENT_BEHAVE_CONTRACT_TOKEN=<your-token> \
tox -e behave -- features/cli/attach.feature -D releases=xenial -D machine_types=lxd-vm
```

### DIFF
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -704,42 +704,21 @@ class _LXD(Cloud):
         )
 
         if self.name == "lxd-virtual-machine" and series == "xenial":
-            # Livepatch won't apply patches on Xenial with secure boot enabled
-            config_dict = {"security.secureboot": False}
+            # Livepatch won't apply patches on Xenial with secure boot enabled.
+            # "boot.mode" is the config key used by LXD 6.x (the version we
+            # develop against on Noble); it replaces the older
+            # "security.secureboot" key.
+            config_dict = {"boot.mode": "uefi-nosecureboot"}
         else:
             config_dict = {}
 
-        try:
-            inst = self.api.launch(
-                name=instance_name,
-                image_id=image_name,
-                user_data=user_data,
-                ephemeral=ephemeral,
-                config_dict=config_dict,
-            )
-        except RuntimeError as e:
-            # Some LXD versions no longer support the "security.secureboot"
-            # config key for virtual-machines and reject it at instance
-            # creation time. It is only set to allow Livepatch on Xenial, so
-            # fall back to launching without it when it is not supported.
-            if "security.secureboot" in config_dict and (
-                "security.secureboot" in str(e)
-            ):
-                logging.warning(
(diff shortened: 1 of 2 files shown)


features/cloud.py
```diff
@@ -709,13 +709,37 @@ def _create_instance(
...
             config_dict = {}
 
-        inst = self.api.launch(
-            name=instance_name,
-            image_id=image_name,
-            user_data=user_data,
-            ephemeral=ephemeral,
-            config_dict=config_dict,
-        )
+        try:
```
srunde3: I'd argue that LXD is a hard dependency of our test suite, and we can specify which version of LXD we develop against. On noble, that's probably `6.x`:

```sh
$ lxc --version                                                                                                                           
6.9
```

What if we just update the branch above to use the newer config for secureboot? https://canonical.com/lxd/docs/latest/reference/instance_options/#boot-related-options

yujinhong-canonical: I understand your point, we can just expect to always be testing with lxd version 6.9 or above, so we no longer have to consider the scenarios where we need secureboot to exist. I made this change.

yujinhong-canonical: Thank you for the suggestion and the link to the documentation! This approach is a cleaner implementation than what I originally had.

features/cloud.py
```diff
@@ -704,8 +704,11 @@ def _create_instance(
         )
 
         if self.name == "lxd-virtual-machine" and series == "xenial":
-            # Livepatch won't apply patches on Xenial with secure boot enabled
-            config_dict = {"security.secureboot": False}
+            # Livepatch won't apply patches on Xenial with secure boot enabled.
```
srunde3: I don't think this is the right spot to mention that we use LXD 6; this should go into contributing docs or anywhere else that we mention dependencies for development.

At this level we just care about documenting that this is necessary for Xenial containers. Based on your findings, it's necessary for all Xenial containers, and not just as a workaround for Livepatch tests

yujinhong-canonical: Addressed. Will see if there is an appropriate place for this info.

srunde3 (approved):
🚢 🇮🇹

features/cloud.py
yujinhong-canonical: I just added this info to `dev-docs/tutorial/getting-started` under "Setting up LXD".

PR #3573, opened 2026-07-14
feat(GCE): Add Marketplace Pro licenses to uaclient/clouds/gce.py

kajiya3141:
We noticed in [cloud-init issue #6924](https://github.com/canonical/cloud-init/issues/6924) that we should probably add the (different) Google Pro Marketplace licenses to this list as well. The likelihood of a user doing a in-place upgrade on Marketplace is kind of low but not inpossible, plus we now know that this list is also used as a "fallback" for Pro auto-attach on boot, so it's best to keep it up to date.


srunde3 (approved):
Confirmed that these license IDs are correct against the back end. LGTM

PR #3574, opened 2026-07-14
Add resolute badge in readme

alanmcanonical:
## Why is this needed?

- resolute badge is missing from README.md


srunde3 (approved):
Thank you!

PR #3583, opened 2026-07-23
test: add Resolute to "disable with purge" scenario

srunde3:
## Why is this needed?

This coverage was skipped the first time around -- mistakenly, it didn't seem that Ansible was suitable a candidate package for this test flow. That seems not to be the case, it works fine.

Eventually I'd like to refactor these test files more, so that it's clear which qualities of each package are relevant to the test. I included a small refactor here towards that effort.

I also included `@uses.config.contract_token` to other tests in the file that were missing this annotation.

## Test Steps

Export your token, then run:

```sh
UACLIENT_BEHAVE_INSTALL_FROM=local tox -e behave -- features/cli/disable.feature:267
```

### DIFF
--- a/features/cli/disable.feature
+++ b/features/cli/disable.feature
@@ -242,6 +242,7 @@ Feature: CLI disable command
       | focal   | lxd-container | focal            | ansible | universe          |
       | jammy   | lxd-container | jammy            | ansible | universe          |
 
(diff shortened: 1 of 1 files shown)


yujinhong-canonical (approved):
LGTM, the tests also all pass.

PR #3591, opened 2026-07-31
ci(behave): retry failing features/ tests up to 3 times

srunde3:
## Why is this needed?

Ensure suites are not failing due to flaky tests by retrying only the failing tests, up to 3 times.

A single pass is a green signal, so we don't retry a test once it passes once.

If there are >= 10 failures (arbitrary, picked this to start) the retry skips; we just consider that a terminal failure.

## Test Steps

Runs in CI. You can run the tool test suite locally:

```sh
tox -e test -- tools/tests/test_run_behave_with_retries.py
```


renanrodrigo:
@yujinhong-canonical this one will help you in tests I guess. no need to merge it now if there is any problem/uncertainty, but a look is worth it

PR #3596, opened 2026-08-13
ci: remove create-issue action

renanrodrigo:
This is a custom action created for fun, but rarely used for real and constantly bringing JS security concerns.


.github/actions/create-issue/action.yaml
```diff
@@ -1,9 +0,0 @@
-name: 'Create issues on PR merge'
-description: 'Looks for "Breaks:" in a PR comment and opens an issue when found'
```
srunde3, line 2: Was this in the Riga talks? This seems vaguely familiar...

srunde3 (approved):

.github/actions/create-issue/action.yaml
renanrodrigo: yep there is where I presented it.
I may create a real action at some point in time, or just let it be history.

In any case, this repo doesn't need it anymore

PR #3604, opened 2026-08-20
fix: use explicit apt auth files/dirs in assert_valid_apt_credentials

srunde3:
## Why is this needed?

Trusty does not always pick up on the tempfile used for the ESM repo authentication. The credential is correct, but the `ua enable esm-infra-legacy` command fails because `assert_valid_apt_credentials` fails.

## Test Steps

Spin up a trusty machine. You'll also need an Ubuntu Pro token that has legacy products entitled

```sh
sudo apt update && sudo apt install --upgrade-only ubuntu-advantage-tools
sudo ua attach <TOKEN>
sudo ua enable esm-infra-legacy
```

This should repro the issue. Next, build:

```sh
sudo sbuild-adduser "$USER"
mk-sbuild trusty
sbuild --dist=trusty .
```

Push this build into your machine and install it. Try enabling again:

```sh
sudo dpkg -i ./your/new/build
sudo ua enable esm-infra-legacy
```

I had good luck with VirtualBox.

More comprehensive UATs will be performed prior to a release. This review should verify that the bug is fixed and identify any obvious regressions.

### DIFF
--- a/uaclient/apt.py
+++ b/uaclient/apt.py
@@ -68,8 +68,6 @@ def assert_valid_apt_credentials(repo_url, username, password):
     if not os.path.exists("/usr/lib/apt/apt-helper"):
         return
     protocol, repo_path = repo_url.split("://")
-    apt_auth_file = get_apt_auth_file_from_apt_config()
-    auth_path = apt_auth_file + "-validation"
     if not repo_path.endswith("/"):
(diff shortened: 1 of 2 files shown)


yujinhong-canonical (approved):
LGTM, confirming that the issue was reproduced and fixed. Thank you!

PR #3608, opened 2026-08-27
fix: bionic daily image no longer in azure

yujinhong-canonical:
## Why is this needed?
This fixes feature tests on bionic azure cases.

## Test Steps
```
UACLIENT_BEHAVE_CONTRACT_TOKEN=<token> \
tox -e behave --   features/apt_messages.feature -D releases=bionic -D machine_types=azure.generic
```
-->

### DIFF
--- a/features/cloud.py
+++ b/features/cloud.py
@@ -461,6 +461,47 @@ class Azure(Cloud):
         # instead of the instance id
         return instance.name
 
+    def locate_image_name(
+        self,
+        series: str,
+        machine_type: str,
+        daily: bool = True,
+        include_deprecated: bool = False,
+    ) -> str:
+        """Locate and return the image name to use for vm provision.
+
+        :param series:
+            The ubuntu release to be used when locating the image name
+        :machine_type:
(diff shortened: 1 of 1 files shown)


features/cloud.py
```diff
@@ -461,6 +461,47 @@ def get_instance_id(
...
+            logging.debug("looking up daily image for {}".format(series))
+            return self.api.daily_image(
+                release=series,
+                image_type=image_type,
+                include_deprecated=include_deprecated,
+            )
+        else:
+            # pycloudlib.Azure.released_image only accepts `release`
+            logging.debug("looking up released image for {}".format(series))
+            return self.api.released_image(release=series)
```
srunde3, line 503: Since `daily=false` for bionic, that means we don't pass the `machine_type` (i.e., the fact that we want a FIPS image) along; did you confirm that we still get a FIPS image with this change? It seems like we wouldn't...

srunde3 (changes requested):
We typically abstract our interactions with clouds via `pycloudlib`; any change related to availability of images in Azure should go there if possible.

PR #3611, opened 2026-09-04
chore: remove tox-uv from tox.ini deps

srunde3:
## Why is this needed?

`tox-uv` was not really necessary to introduce Workshop. It just allows `tox` to use `uv` to manage the environments instead of `pip`.

It also has an annoying implication for feature tests: `tox` has a single-instance lock during environment provisioning. If we call `tox -e behave ...` with a `tox.ini` that contains a dependency the caller doesn't have installed (e.g., `tox-uv`, if you didn't install that along `tox`), then `tox` will spin up an environment to install those deps in and will re-run the requested command. However, that command is run within the single-instance lock.

You can also get around this issue if you install `tox-uv` wherever you invoke `tox` from, e.g., `pipx`. But it's not strictly necessary to have `tox-uv` at all, so I'd prefer to just remove it from the `tox.ini`.

## Test Steps

You can try this yourself. Run this test from `main` (export your token first):

```sh
UACLIENT_BEHAVE_INSTALL_FROM=archive tox -e behave -- features/_version.feature:148
```

Then try running it again in another terminal, concurrently. The second will only start after the first finishes. Then, check out this branch and try the whole thing again. They should run concurrently.

### DIFF
--- a/.github/workflows/ci-base.yaml
+++ b/.github/workflows/ci-base.yaml
@@ -41,7 +41,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo DEBIAN_FRONTEND=noninteractive apt-get -qy update
-          sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox libapt-pkg-dev
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox libapt-pkg-dev python3-apt
       - name: Git checkout
         uses: actions/checkout@v3
       - name: Unit
--- a/.github/workflows/ci-base.yaml
+++ b/.github/workflows/ci-base.yaml
@@ -41,6 +41,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo DEBIAN_FRONTEND=noninteractive apt-get -qy update
+          # tox test env inherits the distro python3-apt extension.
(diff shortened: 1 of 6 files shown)


yujinhong-canonical (commented):
While reviewing this PR, I tried running two tests at the same time on different terminals. I would notice that the first test I run fails, although it succeeds when you run it on its own. Have you had this issue?

srunde3:
> While reviewing this PR, I tried running two tests at the same time on different terminals. I would notice that the first test I run fails, although it succeeds when you run it on its own. Have you had this issue?

Nope, I had not seen this. What's the failure? And which test?

yujinhong-canonical:
> > While reviewing this PR, I tried running two tests at the same time on different terminals. I would notice that the first test I run fails, although it succeeds when you run it on its own. Have you had this issue?
> 
> Nope, I had not seen this. What's the failure? And which test?

I ran autocomplete.feature for bionic and focal lxd-container, almost right after another. Both failed this time.

Terminal 1:
```
UACLIENT_BEHAVE_INSTALL_FROM=local tox -e behave --   features/autocomplete.feature -D releases=bionic -D machine_types=lxd-container
```
Terminal 2:
```
UACLIENT_BEHAVE_INSTALL_FROM=local tox -e behave --   features/autocomplete.feature -D releases=focal -D machine_types=lxd-container
```

PR #3619, opened 2026-09-19
Fix misleading failure message for already-enabled services

PRINCEGOTI12:
Fixes #3525

Running `pro enable` for an already-enabled service prints both “already enabled - nothing to do” and “Could not enable”, giving conflicting information.

This change removes the generic failure message from the already-enabled branch while preserving the existing return behaviour and error reporting for genuine failures.

Adds a regression test that verifies only the already-enabled message is printed and the return result remains unchanged.

Validation:

* All 33 tests in `uaclient/cli/tests/test_cli_enable.py` passed.
* The new regression test passed independently.
* `git diff --check` passed.


PRINCEGOTI12:
I have signed the Canonical CLA through my GitHub account PRINCEGOTI12 and received the success confirmation.

I tried rerunning the failed CLA check with GitHub CLI, but it returned: “run 35437083216 cannot be rerun; its workflow file may be broken.”

Could a maintainer please rerun the CLA check or help verify that my signature is linked to my GitHub account? My commit uses my GitHub noreply email.

